PR #2111, opened 2025-12-05
[MoE] Add node limited routing support

shuhuayu:
As titled, added node-limited routing support via two-layer routing.  First, group experts into `num_groups` groups, and experts in the same group should reside on the same node to utilize fast intra-node communication. Second, pick the `top_k_group` by the top 2 expert scores' sum in each group. Third, pick `top_k` experts within the selected `top_k_groups`. 

Reference: https://github.com/huggingface/transformers/blob/4c9fde2a2a3aece0bcf1be93f696e88297da9397/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py#L212

Test on one node using DeepSeek V3 debug model with MoE arguments `num_experts=8,
            num_shared_experts=2,
            num_groups=4,
            top_k_group=2,
            top_k=3`.
Pasted Graphic

### DIFF
--- a/torchtitan/models/moe/moe.py
+++ b/torchtitan/models/moe/moe.py
@@ -239,69 +239,46 @@ class TokenChoiceTopKRouter(nn.Module):
         top_scores = scores.gather(dim=1, index=selected_experts_indices)  # [N,K]
         return selected_experts_indices, top_scores
 
-    def _node_limited_routing(
+    def _get_node_limited_routing_scores(
         self,
-        scores: torch.Tensor,
-        expert_bias: torch.Tensor | None,
+        scores_for_choice: torch.Tensor,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        """Select top_k experts, optionally limiting to a subset of expert groups.
-
-        If num_expert_groups is set, applies node-limited routing:
-        1. Select num_limited_groups groups based on group scores
-        2. Select top_k experts only from those groups
-
-        If expert_bias is provided, it is added to scores for selection, but
-        the returned top_scores are always from the original (unbiased) scores.
+        """Select num_limited_groups groups based on group scores,
+            and set expert scores in non-selected groups as -inf
 
         Args:
-            scores: Router scores after sigmoid or softmax, shape (bs*slen, num_experts)
-            expert_bias: Optional bias for load balancing, shape (num_experts,)
+            scores_for_choice: Router scores with expert_bias (if any), shape (bs*slen, num_experts)
 
         Returns:
-            tuple of (selected_experts_indices, top_scores)
-                - selected_experts_indices: shape (bs*slen, top_k)
-                - top_scores: shape (bs*slen, top_k)
+            scores_for_choice: shape (bs*slen, num_experts)
         """
-        scores_for_choice = scores if expert_bias is None else scores + expert_bias
-
-        # Apply node-limited routing mask if configured
-        if self.num_expert_groups is not None:
-            if self.num_limited_groups is None:
-                raise ValueError(
-                    "num_limited_groups must be set when num_expert_groups is set"
-                )
-            if self.num_experts % self.num_expert_groups != 0:
-                raise ValueError(
-                    f"num_experts ({self.num_experts}) must be divisible by num_expert_groups ({self.num_expert_groups})"
-                )
-            experts_per_group = self.num_experts // self.num_expert_groups
-            if experts_per_group < 2:
-                raise ValueError(
-                    f"experts_per_group ({experts_per_group}) must be >= 2"
-                )
-            scores_grouped = scores_for_choice.view(
-                -1, self.num_expert_groups, experts_per_group
+        if self.num_limited_groups is None:
+            raise ValueError(
+                "num_limited_groups must be set when num_expert_groups is set"
             )
-            group_scores = scores_grouped.topk(2, dim=-1)[0].sum(dim=-1)
-            group_idx = torch.topk(
-                group_scores, k=self.num_limited_groups, dim=-1, sorted=False
-            )[1]
-            group_mask = torch.ones_like(group_scores, dtype=torch.bool)
-            group_mask.scatter_(1, group_idx, False)  # False = selected groups (keep)
-            # Mask out experts from non-selected groups
-            scores_for_choice = scores_grouped.masked_fill(
-                group_mask.unsqueeze(-1), float("-inf")
-            ).view(-1, self.num_experts)
+        if self.num_experts % self.num_expert_groups != 0:
+            raise ValueError(
+                f"num_experts ({self.num_experts}) must be divisible by num_expert_groups ({self.num_expert_groups})"
+            )
+        experts_per_group = self.num_experts // self.num_expert_groups
+        if experts_per_group < 2:
+            raise ValueError(f"experts_per_group ({experts_per_group}) must be >= 2")
+        scores_grouped = scores_for_choice.view(
+            -1, self.num_expert_groups, experts_per_group
+        )
+        top2_scores_in_group, _ = scores_grouped.topk(2, dim=-1)
+        group_scores = top2_scores_in_group.sum(dim=-1)
+        _, group_idx = torch.topk(
+            group_scores, k=self.num_limited_groups, dim=-1, sorted=False
+        )
+        group_mask = torch.ones_like(group_scores, dtype=torch.bool)
+        group_mask.scatter_(1, group_idx, False)  # False = selected groups (keep)
+        # Mask out experts from non-selected groups
+        scores_for_choice = scores_grouped.masked_fill(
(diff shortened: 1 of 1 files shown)


torchtitan/models/moe/moe.py
```diff
@@ -26,7 +26,9 @@ class MoEArgs:
     route_scale: float = 1.0
     score_before_experts: bool = True
 
-    # token-choice
+    # token-choice with node limited routing support
+    num_groups: int | None = None  # must be a divisor of num_experts
+    top_k_group: int | None = None
```
wwwjn:
```suggestion
    top_k_groups: int | None = None
```

torchtitan/models/moe/moe.py
```diff
@@ -225,6 +239,60 @@ def _debug_force_load_balance_routing(
...
+        # Mask out experts from non-selected groups
+        scores_for_choice = scores_for_choice.masked_fill(
+            ~score_mask.bool(), float("-inf")
+        )
+        # Select top_k experts from masked scores
+        selected_experts_indices = torch.topk(
+            scores_for_choice, k=self.top_k, dim=-1, sorted=False
+        )[1]
+        # Get actual scores (without bias) for the selected experts
+        top_scores = scores.gather(1, selected_experts_indices)
```
tianyu-l: let's unify this part with other two paths (even no expert bias we can do topk + gather)

torchtitan/models/moe/moe.py
tianyu-l: set the default for 671B model according to https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/configs/config_671B.json

tianyu-l: Are the names coming from some repo? I saw that deepseek repo calls them `n_expert_groups` and `n_limited_groups`.

I think we can combine the convention and call them `num_expert_groups` and `num_limited_groups`. WDYT?

torchtitan/models/moe/moe.py
```diff
@@ -257,7 +325,17 @@ def forward(
...
         # NOTE: The expert_bias is only used for routing. The gating value
         #       top_scores is still derived from the original scores.
-        if expert_bias is not None:
+        if self.num_groups is not None:
+            assert (
+                self.top_k_group is not None
+            ), "top_k_group must be set when num_groups is set"
+            assert (
+                self.num_experts % self.num_groups == 0
+            ), f"num_experts ({self.num_experts}) must be divisible by num_groups ({self.num_groups})"
```
tianyu-l: Instead of doing assert, let's raise ValueError since they are more like error from users.

torchtitan/models/moe/moe.py
tianyu-l: nit: let's put these two fields below `top_k` which is a more "important" arg.

tianyu-l:
need to remove https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/deepseek_v3/model/args.py#L70-L71

torchtitan/models/moe/moe.py
shuhuayu: Sounds great! I changed to `num_expert_groups` and `num_limited_groups`, which have clearer meaning  from naming. The previous names are from huggingface's implementations.

torchtitan/models/moe/moe.py
shuhuayu: Great suggestion! Merged into a unified `_node_limited_routing` method.

shuhuayu: Also changed to a smaller mask to save memory.

torchtitan/models/moe/moe.py
```diff
@@ -225,6 +239,70 @@ def _debug_force_load_balance_routing(
...
+
+        selected_experts_indices = torch.topk(
+            scores_for_choice, k=self.top_k, dim=-1, sorted=False
+        )[1]
+
+        # NOTE: The expert_bias is only used for routing. The gating value
+        #       top_scores is still derived from the original scores.
+        top_scores = scores.gather(dim=1, index=selected_experts_indices)
+
+        return selected_experts_indices, top_scores
```
tianyu-l: this should stay outside node_limited_routing method. If you worry about naming, you could call it something like `_get_node_limited_routing_scores`

torchtitan/models/moe/moe.py
```diff
@@ -180,9 +182,17 @@ class TokenChoiceTopKRouter(nn.Module):
     """This class implements token-choice routing. In token-choice top-K routing, each token is
         routed to top K experts based on the router scores.
 
+    Optionally supports node-limited (group-limited) routing where experts are divided into groups
+    (e.g., by node), and only num_limited_groups groups are considered before selecting top_k experts.
+    This reduces cross-node communication in distributed settings.
```
wwwjn, line 187: This is not true for NCCL native a2a, and only true with DeepEP?

torchtitan/models/moe/moe.py
```diff
@@ -225,6 +239,70 @@ def _debug_force_load_balance_routing(
...
+                raise ValueError(
+                    f"experts_per_group ({experts_per_group}) must be >= 2"
+                )
+            scores_grouped = scores_for_choice.view(
+                -1, self.num_expert_groups, experts_per_group
+            )
+            group_scores = scores_grouped.topk(2, dim=-1)[0].sum(dim=-1)
+            group_idx = torch.topk(
+                group_scores, k=self.num_limited_groups, dim=-1, sorted=False
+            )[1]
```
wwwjn:
```suggestion
            _, group_idx = torch.topk(
                group_scores, k=self.num_limited_groups, dim=-1, sorted=False
            )
```

For readability, it's easy to forgot why we need to slice topk()'s result

torchtitan/models/moe/moe.py
```diff
@@ -225,6 +239,70 @@ def _debug_force_load_balance_routing(
...
+            group_mask = torch.ones_like(group_scores, dtype=torch.bool)
+            group_mask.scatter_(1, group_idx, False)  # False = selected groups (keep)
+            # Mask out experts from non-selected groups
+            scores_for_choice = scores_grouped.masked_fill(
+                group_mask.unsqueeze(-1), float("-inf")
+            ).view(-1, self.num_experts)
+
+        selected_experts_indices = torch.topk(
+            scores_for_choice, k=self.top_k, dim=-1, sorted=False
+        )[1]
```
wwwjn:
```suggestion
        _, selected_experts_indices = torch.topk(
            scores_for_choice, k=self.top_k, dim=-1, sorted=False
        )
```

torchtitan/models/moe/moe.py
```diff
@@ -225,6 +239,70 @@ def _debug_force_load_balance_routing(
...
+                )
+            experts_per_group = self.num_experts // self.num_expert_groups
+            if experts_per_group < 2:
+                raise ValueError(
+                    f"experts_per_group ({experts_per_group}) must be >= 2"
+                )
+            scores_grouped = scores_for_choice.view(
+                -1, self.num_expert_groups, experts_per_group
+            )
+            group_scores = scores_grouped.topk(2, dim=-1)[0].sum(dim=-1)
```
wwwjn:
```suggestion
            top2_scores_in_group, _ = scores_grouped.topk(2, dim=-1)
            group_scores = top2_scores_in_group.sum(dim=-1)
```

 wdyt?

torchtitan/models/moe/moe.py
shuhuayu: I think this depends how the communication is implemented. The node-limited routing itself works the same for both all-to-all communication, i.e., enforcing communication across less nodes.

torchtitan/models/moe/moe.py
shuhuayu: Incorporated, thx!

torchtitan/models/moe/moe.py
shuhuayu: Make sense. Moved these lines into the `forward` method.

torchtitan/models/moe/moe.py
```diff
@@ -225,6 +239,57 @@ def _debug_force_load_balance_routing(
...
+            scores: Router scores after sigmoid or softmax, shape (bs*slen, num_experts)
+            expert_bias: Optional bias for load balancing, shape (num_experts,)
+
+        Returns:
+            scores_for_choice: shape (bs*slen, num_experts)
+        """
+        scores_for_choice = scores if expert_bias is None else scores + expert_bias
+
+        # Apply node-limited routing mask if configured
+        if self.num_expert_groups is not None:
```
tianyu-l: this should happen outside, i.e.
```
if node_limited:
  _get_node_limited_routing_scores()
else:
  ...
```

shuhuayu: Nice suggestion! Moved this condition check to the `forward` method and now this method will process `scores_for_choice` only if node-limited routing is configured, which better aligns with its name.

PR #2112, opened 2025-12-05
Support rope cache indexing using positions

acisseJZhong:
Add support to indexing rope cache using `position_ids`, this might be needed during 
1. inference, where we passed in `position_ids` into transformer forward 
2. CP load balancing where we need to index rope cache given positions ids 

Test: 
running dpskv3 16b base
image

also tested in https://github.com/wwwjn/torchtitan/pull/1/files when passing position_ids
image

### DIFF
--- a/torchtitan/models/deepseek_v3/infra/parallelize.py
+++ b/torchtitan/models/deepseek_v3/infra/parallelize.py
@@ -225,6 +225,8 @@ def apply_non_moe_tp(
     for transformer_block in model.layers.values():
         layer_plan = {
             "attention_norm": SequenceParallel(),
+            # NOTE: when the fourth argument (positions) is not None, its input layout
+            # and desired input layout should be Replicate()
             "attention": prepare_module_input(
                 input_layouts=(Shard(1), Replicate(), None, None),
                 desired_input_layouts=(Replicate(), Replicate(), None, None),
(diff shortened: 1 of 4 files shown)


torchtitan/models/deepseek_v3/infra/parallelize.py
```diff
@@ -226,8 +226,8 @@ def apply_non_moe_tp(
         layer_plan = {
             "attention_norm": SequenceParallel(),
             "attention": prepare_module_input(
-                input_layouts=(Shard(1), Replicate(), None),
-                desired_input_layouts=(Replicate(), Replicate(), None),
+                input_layouts=(Shard(1), Replicate(), None, None),
```
tianyu-l, line 231: Note that when `positions` is not `None`, this is making implicit assumption that `positions` has the the expected sharding when it's used, namely
sharded on batch dim by DP, replicate on TP mesh, sharded on seq dim by CP

I don't have a good solution right now -- making it `Replicate` by default will fail here when `positions` is `None` https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/parallel/style.py#L521
but clearly this is leaving a footgun. I'd suggest we add a comment for now.

acisseJZhong: yeah when testing positions [1, seq_len] and [bz, seqlen] in dp4tp2, I need to manually change both layouts to `Replicate()`. But for the default case it should be `None`. 

For CP, I think we need to manually change https://fburl.com/v2rn2s48. 
For FSDP, not sure how the sharding info is specified today but looks like it's already handled?

will add a comment for now

acisseJZhong:
> We should try both (1, seq_len) and (batch_size, seq_len) inputs (they could be the trivial 0 -> seq_len - 1 ids).

updated the loss graph, I also tested both cases with vllm inference as well, and the text output is the same.

PR #2113, opened 2025-12-05
perf(pipeline): implement auto-partition algorithm

TXacs:
# Auto-Partition in torchtitan

## Overview

This PR provides an automatic partitioning method that considers the computation cost of embedding layers. 
Thsi method involves calculating the floating-point operations (FLOPs) of the embedding layers and constructing an array that incorporates the FLOPs of both the transformer and embedding layers. Subsequently, a heuristic algorithm is employed to identify a balanced pipeline partition.

## Solution Architecture
1. Dynamic Cost Analysis
2. Adaptive Partitioning Algorithm
3. Workload Balancing

## Performance

Hardware configuration: 4x RTX 3090 24GB, pipeline parallelism dimension is 4.

### llama3 配置对比
|  hidden size|  layers |  autopipe TPS|  default TPS|  Speedup     |
|  ---------- |  ----   |  ----------  |  -----------|  ----------- |
|  dim=256    |  6      |  31,094      |  29,549     |  +5.2%       |
|  dim=256    |  12     |  21,803      |  21,923     |  -0.5%       |
|  dim=2048   |  12     |  3,348       |  2,616      |  +28.0%      |
|  dim=4096   |  12     |  981         |  761        |  +28.9%      |

### deepseekv3(without moe) 配置对比

|  hidden size|  layers |  autopipe TPS|  default TPS|  Speedup     |
|  ---------- |  ----   |  ----------  |  -----------|  ----------- |
|  dim=256    |  6      |  13,373      |  13,059     |  +2.4%       |
|  dim=256    |  12     |  7,714       |  6,859      |  +12.5%      |
|  dim=2048   |  12     |  4,331       |  3,810      |  +13.7%      |
|  dim=4096   |  12     |  2,888       |  2,561      |  +12.8%      |
|  dim=4096   |  16     |  2,207       |  2,008      |  +9.9%       |
|  dim=8192   |  16     |  4,331       |  3,935      |  +10.1%      |


torchtitan/experiments/autopartition/infra/cpp/autopipe.cpp
tianyu-l, line 1: This looks interesting -- how much benefit you'd get from having a c++ implementation, compared with a python one?

tianyu-l (commented):
Thanks. Is it true that the only "real" deltas are
- autopipe.cpp
- pipeline_parallel.py
- profiler.py

TXacs:
> Thanks. Is it true that the only "real" deltas are
> 
> * autopipe.cpp
> * pipeline_parallel.py
> * profiler.py

Yes，actually, `profile.py` also uses the file from DeepSpeed. It would be even better if TorchTitan could provide a more authoritative FLOPs calculation method in the future, so that we could also adapt it for MoE models.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,598 @@
...
+            inputs = layer(*inputs)
+        mflops = prof.get_total_flops() / 10**6  # Convert to million FLOPS
+        mflops_list.append(round(mflops))
+        prof.end_profile()
+
+    logger.info(f"Autopipe partitioning with mflops: {mflops_list}")
+
+    parts = pipeline(
+        mflops_list,
+        [i * 3 for i in mflops_list],  # Assume backward is 3x forward
```
H-Huang: why is it assumed to be 3x?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,598 @@
...
+        tokenizer=tokenizer,
+        job_config=job_config,
+    )
+    iterator = iter(dataloader)
+    inputs = next(iterator)[0].values()
+
+    # Profile each layer's FLOPS
+    mflops_list = []
+    for _, layer in enumerate(model):
+        prof = FlopsProfiler(layer)
```
H-Huang: I guess the `FlopsProfiler` does not estimate the backward flops?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
McmillanTAC: Similar to the computation time, in which the backward computation time is twice that of the forward computation time, by default, we assume that the floating-point operations (FLOPs) required for backward computation are twice those for forward computation. During the modeling process, we take the recomputation technique into account. This technique inserts an additional forward computation before the backward computation. Consequently, we set the default FLOPs for backward computation to be three times those of forward computation.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
McmillanTAC: You are correct that the `FlopsProfiler` does not estimate backward flops. Instead, we use a heuristic rule by default: given the adoption of recomputation techniques in the backward computation, backward flops are set to three times those of the forward computation.

tianyu-l: curious why three times? IIUC the convention was two times.

McmillanTAC: During the modeling process, we employ the recomputation technique, which introduces an additional forward pass during backpropagation. To control the computational overhead of backpropagation, we can adjust the implementation by setting a scaling factor—typically 2 or 3—depending on whether recomputation is enabled. This adjustment is orthogonal to the pipeline partitioning algorithm.

TXacs: > I guess the `FlopsProfiler` does not estimate the backward flops?

We have completely rebuilt the FLOPs acquisition logic—reverse FLOPs for each layer can now be computed according to the ACConfig strategy, and the AutoPipe algorithm has been re-implemented in pure Python.

McmillanTAC: Following the previous review, we have further refined this PR with the following key updates:
1. Refactored the core algorithm from C++ to pure Python, enhancing maintainability and better aligning with the PyTorch ecosystem.
2. Integrated PyTorch's official torch.utils.flop_counter.FlopCounterMode, replacing the third-party profiler to improve integration and reduce external dependencies.
3. Refined the FLOPs calculation logic​ to dynamically determine backward pass costs based on the activation checkpointing configuration, moving beyond a fixed multiplier for greater accuracy.

We welcome any additional suggestions to move this forward. Thanks again for your review.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
tianyu-l, line 1: @H-Huang could you help review the core algorithm part in pipeline_parallel.py and autopipe.py?

torchtitan/experiments/autopartition/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Overview
+
+This folder provides an automatic partitioning method that considers the computation cost of embedding layers. 
+This method involves calculating the floating-point operations (FLOPs) of the embedding layers and constructing an array that incorporates the FLOPs of both the transformer and embedding layers. Subsequently, a heuristic algorithm is employed to identify a balanced pipeline partition.
+
+## Quick Start
+
+### Compile
+
+First, we need to compile `autopipe.cpp`.
```
wconstab: is this comment outdated? i do not see any .cpp files in the PR anymore.

McmillanTAC: Yes, this comment is outdated. The PR has migrated AutoPipe logic from C++ to pure Python, so there are no .cpp files involved anymore. We should remove or update this line to reflect the current implementation. Sorry for the confusion caused by the outdated documentation!

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,647 @@
...
+    apply_ac(
+        copied_model,
+        job_config.activation_checkpoint,
+        model_compile_enabled=job_config.compile.enable,
+        use_flex_attn=use_flex_attn,
+        op_sac_save_list=_op_sac_save_list,
+    )
+    seq_modules = _build_module_for_profile(copied_model, flatten_module_names)
+
+    parts = autopipe_partition(seq_modules, parallel_dims.pp, job_config)
```
H-Huang: why do we do the `autopipe_partition` after the splitting is already done? I propose there should be 3 different ways of doing things:

1. layer-wise static partitioning (default)
2. config based (user sets layers per stage)
3. autopipe partition (this new implementation)

The option to choose should be set by configuration. Further, if the autopipe way is determined to be strictly better than it should replace the default.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,647 @@
...
+    )
+    iterator = iter(dataloader)
+    inputs = list(next(iterator)[0].values())
+
+    mflops_fwd, mflops_bwd = layerwise_flops(model, inputs)
+
+    logger.info(f"Autopipe partitioning with mflops: {mflops_fwd}, {mflops_bwd}")
+
+    # Partition layers by forward and backward flops
+    parts = pipeline(
```
H-Huang: nit: this `pipeline` name conflicts with naming provided in pytorch under `torch.distributed.pipelining` so maybe should have a different name

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,647 @@
...
+
+    Returns:
+        List of integers representing the number of layers assigned to each stage.
+    """
+
+    # Prepare input for profiling
+    tokenizer = build_hf_tokenizer(job_config)
+
+    # build dataloader
+    dataloader = build_text_dataloader(
```
H-Huang, line 111: it's not ideal that we have to create a new dataloader for each rank. Can we not use dummy data as the example inputs instead?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,647 @@
...
+    "pipeline_module_split",
+]
+
+def layerwise_flops(model, x, backward=True):
+    """Return forward and backward FLOPs (float) for each layer of the model."""
+    fwd_mflops, bwd_mflops = [], []
+
+    for layer_idx, layer in enumerate(model):
+        # forward
+        with FlopCounterMode(display=False) as mode:
```
H-Huang, line 64: Hm, this may be an inherent limitation. IIUC `FlopCounterMode` only counts flops for registered pytorch ops. So it would not capture things like FlexAttention which uses fused kernels, it may also not cover torch.compile related optimizations. Do you happen to know @sanketpurandare? How did you do things for your runtime estimators?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
TXacs: Thanks for reviewing the code. We considered that different models have different data formats, which is why using the original data loader can help us avoid many issues.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
McmillanTAC: Thanks for the insightful comment! You're absolutely right. `FlopCounterMode` has limitations with fused kernels and torch.compile related optimizations. However, our approach is modular and extensible, so it can integrate more accurate cost models as they emerge. We treat FLOPs as a baseline, not the final answer. We're open to incorporating better estimators. Would love to hear your thoughts on how you bridge this gap!

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
TXacs: Yeah，It's already fixed in [dc9ebab](https://github.com/pytorch/torchtitan/pull/2113/commits/dc9ebab9f5c938ed55b3239f6eb587d8c2d2fb5e)

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
TXacs: We changed it in [dc9ebab](https://github.com/pytorch/torchtitan/pull/2113/commits/dc9ebab9f5c938ed55b3239f6eb587d8c2d2fb5e)

torchtitan/experiments/autopartition/infra/autopipe.py
```diff
@@ -0,0 +1,359 @@
...
+    @functools.cache
+    def _hash(p):
+        h = 0
+        for inner in p:
+            for v in inner:
+                h ^= (v + 0x9E3779B9) + (h << 6) + (h >> 2)
+        return h
+
+    # C++ VectorEqual
+    def _eq(a, b):
```
H-Huang: not used

torchtitan/experiments/autopartition/infra/autopipe.py
```diff
@@ -0,0 +1,359 @@
...
+    dp: list[list[int]],
+) -> dict:
+    """
+    C++ find_best_partition
+    return: {"partition": [[...], ...], "cost": int, "critical_stage": int}
+    """
+
+    # Hash func: C++ VectorHash
+    @functools.cache
+    def _hash(p):
```
H-Huang: python has its own hash standard library you should use

torchtitan/experiments/autopartition/infra/autopipe.py
```diff
@@ -0,0 +1,359 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import functools
+from collections import deque
+
+COMM_OVERHEAD = 0
```
H-Huang, line 9: why is this 0?

torchtitan/experiments/autopartition/infra/autopipe.py
```diff
@@ -0,0 +1,359 @@
...
+def _calc_stage_times(
+    partition: list[list[int]],
+    block_time: list[list[int]],
+    fwd: list[int],
+    bwd: list[int],
+    last_mb: list[int],
+):
+    """C++ calculate_stage_times"""
+    num_stages = len(partition)
+    num_micro = num_stages * 2
```
H-Huang: num_stages and num_micro should both be hyperparameters, this assumes num_micro is always 2 * stages?

torchtitan/experiments/autopartition/infra/autopipe.py
```diff
@@ -0,0 +1,359 @@
...
+
+    last_bwd_start = _cooldown(num_stages, critical, steady_time, fwd, bwd)
+    flush = last_bwd_start
+    for stage in range(critical, 0, -1):
+        flush += bwd[stage + 1] + COMM_OVERHEAD
+    flush += bwd[1]
+    return flush, critical
+
+
+# ---------- 最优分区搜索 ----------
```
H-Huang: can the chinese comments be translated to english for consistency?

torchtitan/experiments/autopartition/infra/autopipe.py
McmillanTAC: Thank you very much for your careful review! This function is currently unused, and its presence was an oversight on our part. We sincerely apologize for the confusion and appreciate you pointing it out. We’ll remove the unused function in the next update.

torchtitan/experiments/autopartition/infra/autopipe.py
McmillanTAC: Thank you for the feedback!  We will update the code accordingly to use the official implementation.

torchtitan/experiments/autopartition/infra/autopipe.py
McmillanTAC: We set the communication overhead to 0 because the asynchronous communication is fully overlapped with computation. This effectively hides the communication latency, making the observed overhead negligible in our experimental setup.

torchtitan/experiments/autopartition/infra/autopipe.py
McmillanTAC: The num_stages used in this context is a local variable, while the global num_stages is indeed already treated as a hyperparameter. As for num_micro, we currently default it to 2 * stages because this configuration ensures the pipeline naturally forms three distinct phases: warmup, steady, and cooldown. We agree that num_micro should be configurable as a hyperparameter to support more flexible pipeline configurations. We’ll update the code accordingly.

torchtitan/experiments/autopartition/infra/autopipe.py
McmillanTAC: We’ll translate all Chinese comments into English to maintain consistency across the codebase. Thanks for pointing this out!

McmillanTAC:
> I think the PR should be split up into two parts:
> 
> 1. move autopipe.py to [pytorch/pytorch](https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/_utils.py) under torch/distributed/pipelining/_utils.py
> 2. Update pipeline_parallel.py to import this function from pytorch and perform the split (based on configuration)
> 
> The rest of the changes in this PR can be removed.

Thank you for the helpful suggestion! Could you kindly clarify the expected steps for moving autopipe.py into the PyTorch repository?

H-Huang:
> Could you kindly clarify the expected steps for moving autopipe.py into the PyTorch repository?

You will need to clean up the file to be more generalizable `torch.distributed.pipelining`. The algorithm with the DP solver is fine (no model / torch dependency), but there are a few hardcoded assumptions, namely:
- single stage schedule (1F1B)
- fixed number of microbatches based on stages
- COMM_OVERHEAD
- no support for split backward schedules (zero bubble)

I don't think these need to all be solved as once, but if you can get a proof of concept upstreamed to PyTorch that is configurable (remove some of these hardcoded assumptions), then torchtitan can just perform the plumbing to import this and use it.

So concretely, you can upstream the autopipe algorithm to `pytorch/torch/distributed/pipelining`, you can all it `_partition.py` and also add a new test file for it to `pytorch/test/distributed/pipelining` for include unit tests.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
acisseJZhong: why data format matters for FLOP counting? dummy data should be sufficient as long as it has the correct shape and dtype?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+
+    Args:
+        model: The original model to be profiled
+        flatten_module_names: List of module names to include in the sequential model
+
+    Returns:
+        nn.Sequential: A sequential module containing the specified modules
+    """
+    module_seq = nn.Sequential()
+    base_model = copy.deepcopy(model)
```
acisseJZhong, line 147: this seems redundant to L149?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+                else:
+                    # No layers from this structure needed, set to empty structure
+                    if isinstance(module_value, nn.ModuleDict):
+                        setattr(model, module_name, nn.ModuleDict())
+                    elif isinstance(module_value, nn.ModuleList):
+                        setattr(model, module_name, nn.ModuleList())
+            # Handle simple module attributes (e.g., "linear", "norm")
+            elif module_name not in modules_to_keep:
+                # Replace with None
+                setattr(model, module_name, None)
```
acisseJZhong, line 599: can we reuse logic in _build_module_for_profile? or extract common logic as helper function with detailed comments of what it's doing?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+    pp_schedule = build_pipeline_schedule(job_config, stages, loss_fn)
+
+    # This is used in the train loop to determine whether to pass in the input_ids and labels
+    has_first_stage = False
+    has_last_stage = False
+    for stage in stages:
+        if stage.is_first:
+            has_first_stage = True
+        if stage.is_last:
+            has_last_stage = True
```
acisseJZhong, line 329: has_first_stage = any(stage.is_first for stage in stages)
has_last_stage = any(stage.is_last for stage in stages)

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+        loss_fn=rescale_accumulated_loss(loss_fn, n_microbatches),
+        scale_grads=False,
+    )
+    logger.info(
+        f"Using pipeline schedule {job_config.parallelism.pipeline_parallel_schedule} "
+        f"with {n_microbatches} microbatches and {num_total_stages} stages."
+    )
+
+    if pp_schedule_csv:
+        assert schedule_class in [
```
acisseJZhong, line 391: why do we assert this given L352?

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+        raise ValueError(
+            f"Number of stages ({num_stages}) cannot be greater than effective layers ({num_effective_layers})"
+        )
+
+    # Calculate layers per stage (distribute evenly)
+    layers_per_stage = num_effective_layers // num_stages
+    extra_layers = num_effective_layers % num_stages
+
+    # Feasibility check: Ensure at least 1 layer in each PP stage
+    if layers_per_stage == 0:
```
acisseJZhong, line 447: this check becomes unnecessary as we already checked in L434.

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+        raise ValueError(
+            f"Configuration would result in empty stages. "
+            f"With {num_stages} stages and {num_effective_layers} effective layers "
+            f"(num_layers={num_layers} + input_weight={input_weight} + output_weight={output_weight}), "
+            f"each stage would get {layers_per_stage} layers on average. "
+            f"Reduce num_stages or increase num_layers/weights."
+        )
+
+    # Balance check: Ensure weights don't exceed minimum layers per stage
+    if input_weight > layers_per_stage:
```
acisseJZhong, line 457: also raise Error when input_weight < 0 or output_weight < 0

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+        generate_llm_fqn_per_model_part(2, 3, input_weight=2, output_weight=2)
+        treats embeddings as 2 layers and norm+output as 2 layers for distribution
+    """
+    if num_stages < 1:
+        raise ValueError("Number of stages must be at least 1")
+
+    if num_stages == 1:
+        # Single stage gets everything
+        layer_names = [f"layers.{i}" for i in range(num_layers)]
+        return [["tok_embeddings"] + layer_names + ["norm", "output"]]
```
acisseJZhong, line 432: the naming `tok_embeddings`, `norm`, `output` seems tied to specific model, can we generalize?

McmillanTAC:
> > Could you kindly clarify the expected steps for moving autopipe.py into the PyTorch repository?
> 
> You will need to clean up the file to be more generalizable `torch.distributed.pipelining`. The algorithm with the DP solver is fine (no model / torch dependency), but there are a few hardcoded assumptions, namely:
> 
> * single stage schedule (1F1B)
> * fixed number of microbatches based on stages
> * COMM_OVERHEAD
> * no support for split backward schedules (zero bubble)
> 
> I don't think these need to all be solved as once, but if you can get a proof of concept upstreamed to PyTorch that is configurable (remove some of these hardcoded assumptions), then torchtitan can just perform the plumbing to import this and use it.
> 
> So concretely, you can upstream the autopipe algorithm to `pytorch/torch/distributed/pipelining`, you can all it `_partition.py` and also add a new test file for it to `pytorch/test/distributed/pipelining` for include unit tests.

Thank you very much for the detailed guidance! We'll proceed by refactoring autopipe.py to make it more general-purpose and aligned with the design of torch.distributed.pipelining. We'll also add a new module (tentatively named _partition.py) under `pytorch/torch/distributed/pipelining` along with a corresponding unit test file in `pytorch/test/distributed/pipelining`.
We appreciate your support and will aim to submit a clean, minimal proof-of-concept PR to PyTorch soon!

torchtitan/experiments/autopartition/infra/pipeline_parallel.py
```diff
@@ -0,0 +1,650 @@
...
+
+    Returns:
+        nn.Sequential: A sequential module containing the specified modules
+    """
+    module_seq = nn.Sequential()
+    base_model = copy.deepcopy(model)
+
+    for module_name in flatten_module_names:
+        # Create a copy of the base model for each module
+        model_copy = copy.deepcopy(base_model)
```
acisseJZhong, line 151: I saw quite a few deepcopys inside for loop to copy the entire model and then prune to only have target layers, will this cause to OOM for larger models? Can we think about other ways to do this?

TXacs: > I saw quite a few deepcopys inside for loop to copy the entire model and then prune to only have target layers, will this cause to OOM for larger models? Can we think about other ways to do this?

This is an good question. We need to obtain the complete forward and backward inputs and outputs at the model layer level. However, extracting a single layer from the model's attribution cannot provide its full computational graph. Therefore, we adopt TorchTitan's `pipeline_module_split` approach: By capturing each layer's complete computational graph before initializing the model parameters, we can accurately compute FLOPs for each layer. During the current function, parameters remain uninitialized, thus preventing OOM issues that would otherwise occur with large models.

acisseJZhong:
Hi @TXacs could you address @H-Huang comments on moving files into pytorch vs torchtitan, and rebase latest torchtitan main? This could make the code easier to review. 

Also, wondering is this auto-partition idea coming from a paper we could reference? If not, could you explain the logic in `autopipe.py` in more details? Right now it's a little bit hard for me to tell the reasoning behind this partition algorithm. Would be better if you could put an RFC for broader audience to understand. Thanks!

TXacs:
> Hi @TXacs could you address @H-Huang comments on moving files into pytorch vs torchtitan, and rebase latest torchtitan main? This could make the code easier to review.
> 
> Also, wondering is this auto-partition idea coming from a paper we could reference? If not, could you explain the logic in `autopipe.py` in more details? Right now it's a little bit hard for me to tell the reasoning behind this partition algorithm. Would be better if you could put an RFC for broader audience to understand. Thanks!

Yes, we are already preparing the code for submission to PyTorch. We have conducted testing with Torchtitan, and the results are included in this PR under `torchtitan/experiments/autopartition/README.md`. The auto-partition logic is based on the method described in [this paper](https://ieeexplore.ieee.org/document/9912711).

McmillanTAC:
Hi @tianyu-l, @H-Huang, @acisseJZhong, and other reviewers!

I wanted to give you a quick update on this PR. As suggested, we've submitted a corresponding PR to PyTorch to integrate the auto-partition algorithm into `torch.distributed.pipelining`.
The implementation has been refactored to be more generalizable and address the hardcoded assumptions mentioned in your feedback:
* Made the algorithm configurable for different pipeline schedules
* Removed hardcoded microbatch assumptions
* Properly handled communication overhead considerations
* Added comprehensive unit tests

Thank you again for your thorough reviews and helpful guidance! Please let me know if you have any additional feedback on either PR.

acisseJZhong:
@TXacs  @McmillanTAC can you rebase this PR on top of Torchtitan main to make it easier for review. Also is the core infra changes on titan side mainly in `partition.py` and `autopipe.py`? Could you either put an RFC or detailed docstring and comments for us to review? Right now the partition has over 3k lines without any comments, which makes extremely hard for us to review carefully and land this in Titan :(

Regarding pytorch side changes, @sanketpurandare has this huge PR landing soon https://github.com/pytorch/pytorch/pull/177727, so would be nice to consider this.

TXacs:
> @TXacs @McmillanTAC can you rebase this PR on top of Torchtitan main to make it easier for review. Also is the core infra changes on titan side mainly in `partition.py` and `autopipe.py`? Could you either put an RFC or detailed docstring and comments for us to review? Right now the partition has over 3k lines without any comments, which makes extremely hard for us to review carefully and land this in Titan :(
> 
> Regarding pytorch side changes, @sanketpurandare has this huge PR landing soon [pytorch/pytorch#177727](https://github.com/pytorch/pytorch/pull/177727), so would be nice to consider this.

Sorry for the late reply. We'd be happy to rebase and add docstrings for documentation. 

Thank you very much for your suggestions — we'll keep an eye on this PR.

PR #2114, opened 2025-12-05
Implement ciflow/rocm on Torchtitan

akashveramd:
In this PR, I implemented ciflow/rocm on Torchtitan. The changes are part of integration_test_8gpu_features.yaml. The workflow still supports running on pull_request (without any PR label) for CUDA. However, along with push to main and cron schedule, with the ciflow/8gpu label added to PR, the workflow runs for both CUDA & ROCm.

### DIFF
--- a/.github/workflows/set-matrix.yaml
+++ b/.github/workflows/set-matrix.yaml
@@ -16,7 +16,6 @@ jobs:
       # Event flags evaluated by github actions before the step runs:
       IS_MAIN_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
       IS_SCHEDULE:  ${{ github.event_name == 'schedule' }}
-      IS_PR:        ${{ github.event_name == 'pull_request' }}
       IS_8GPU_TAG:  ${{ startsWith(github.ref, 'refs/tags/ciflow/8gpu/') }}
       TRIGGERED_8GPU_LABEL: ${{ github.event_name == 'pull_request' && github.event.action == 'labeled' }}
 
@@ -46,28 +45,24 @@ jobs:
           # Use default value as 'false' for unset environment variables
           IS_MAIN_PUSH="${IS_MAIN_PUSH:-false}"
           IS_SCHEDULE="${IS_SCHEDULE:-false}"
-          IS_PR="${IS_PR:-false}"
           IS_8GPU_TAG="${IS_8GPU_TAG:-false}"
           TRIGGERED_8GPU_LABEL="${TRIGGERED_8GPU_LABEL:-false}"
 
           # Decide which matrix entries to include based on event type
           # Runs ROCm only for push tag OR when PR label gets triggered
           if [[ "$IS_8GPU_TAG" == "true" || "$TRIGGERED_8GPU_LABEL" == "true" ]]; then
-            echo "Runs ROCm only for push tag OR when PR label gets triggered"
             cat > matrix.json <<JSON
           {"include": [$ROCM_MATRIX]}
           JSON
 
           # Runs CUDA and ROCm for normal PR (if PR label is present) OR for push to main, cron schedule
           elif [[ ("$IS_MAIN_PUSH" == "true" || "$IS_SCHEDULE" == "true") ]]; then
-            echo "Runs CUDA and ROCm for normal PR (if PR label is present) OR for push to main, cron schedule"
             cat > matrix.json <<JSON
           {"include": [$CUDA_MATRIX,$ROCM_MATRIX]}
           JSON
 
           # Runs CUDA only as default (includes normal PR, if PR label is NOT present)
           else
-            echo "Runs CUDA only as default (includes normal PR, if PR label is NOT present)"
             cat > matrix.json <<JSON
           {"include": [$CUDA_MATRIX]}
           JSON


.github/pytorch-probot.yml
```diff
@@ -0,0 +1,5 @@
+tracking_issue: 24422
+ciflow_tracking_issue: 64124
+ciflow_push_tags:
+- ciflow/8gpu
+- ciflow/rocm
```
wwwjn: Is this label used for rocm only PRs? Seems this tag is not used in the github workflow yet

.github/pytorch-probot.yml
```diff
@@ -0,0 +1,5 @@
+tracking_issue: 24422
+ciflow_tracking_issue: 64124
+ciflow_push_tags:
+- ciflow/8gpu
```
wwwjn: Thanks! So with this change, we need to label those PR we want to run CUDA + ROCm tests?

.github/pytorch-probot.yml
```diff
@@ -0,0 +1,5 @@
+tracking_issue: 24422
+ciflow_tracking_issue: 64124
```
wwwjn: What does the number 24422 and 64124 mean here?

tianyu-l:
> with the ciflow/8gpu label added to PR, the workflow runs for both CUDA & ROCm.

Can this label be automatically added? If not, what happens if it is not manually added -- is rocm test skipped?

.github/pytorch-probot.yml
```diff
@@ -0,0 +1,5 @@
+tracking_issue: 24422
```
huydhn: You don't need any of `tracking_issue` or `ciflow_tracking_issue` here.  They are only for PyTorch
```suggestion
```

.github/pytorch-probot.yml
huydhn:
```suggestion
```

.github/pytorch-probot.yml
huydhn: +1, no need to add new `ciflow` tags that aren't used anywhere, we can always add them later

.github/pytorch-probot.yml
akashveramd: Removed these https://github.com/pytorch/torchtitan/pull/2114#discussion_r2592373934

.github/pytorch-probot.yml
akashveramd: @wwwjn: CUDA tests can run with pull_request or label, but not for both so as to avoid multiple runs. ROCm tests will only run with label.

.github/pytorch-probot.yml
akashveramd: Removed ciflow/rocm tag. Currently it's not used anywhere. ROCm PR is using ciflow/8gpu.

akashveramd:
> > with the ciflow/8gpu label added to PR, the workflow runs for both CUDA & ROCm.
> 
> Can this label be automatically added? If not, what happens if it is not manually added -- is rocm test skipped?

If the label is not added manually then the ROCm tests will not run. CUDA tests can still run normally.
@huydhn: Please correct me if I am wrong, but we wanted to add label manually by code owners due to security risk, similar to how it works in PyTorch?

huydhn:
> If the label is not added manually then the ROCm tests will not run. CUDA tests can still run normally. @huydhn: Please correct me if I am wrong, but we wanted to add label manually by code owners due to security risk, similar to how it works in PyTorch?

For the context, `ciflow` bot only works if the PR author has write permission to the repo or if the CI has already been approved to run https://github.com/pytorch/test-infra/blob/main/torchci/lib/bot/ciflowPushTrigger.ts#L176-L183.  This policy applies to every repos under PyTorch.

* In the first case, it's perfectly ok to automatically add `ciflow` label.
* In the second case, even if you have the mechanism to add `ciflow` automatically, the bot will remove the label from the PR if the author is an unknown contributor with only read permission.  When this happens, the reviewer needs to review the PR, approve CI run, then add the label manually anyway.  This is the manual process that I'm referring to.

So, let's be clear that adding the label automatically is doable, but only in the first case for privilege users.  The mechanism to automatically add the label is independent from the change in this PR, so you can do it in a separate PR later.  We need:

1. A `labeler.yml` rule like https://github.com/pytorch/pytorch/blob/main/.github/labeler.yml#L11.  This works in most cases, e.g. the PR change a file and `ciflow` label is added
2. For more complex logic like using regex, we might need to update `autoLabelbot` at https://github.com/pytorch/test-infra/blob/main/torchci/lib/bot/autoLabelBot.ts#L26

One note is that when auto label rule is in place, having both the `ciflow` push trigger and `pull_request` triggers feel redundant, and the latter can be removed.

.github/pytorch-probot.yml
```diff
@@ -0,0 +1,2 @@
+ciflow_push_tags:
+- ciflow/8gpu
```
tianyu-l: what does `ciflow_push_tags` do?
why it's called `8gpu`?

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -28,20 +30,22 @@ jobs:
   # Step 1: Dynamically compute the matrix based on conditions
   set-matrix:
     runs-on: ubuntu-latest
+    env:
+      HAS_8GPU_LABEL: ${{ (github.event_name == 'pull_request') && contains(github.event.pull_request.labels.*.name, 'ciflow/8gpu') }}
```
tianyu-l: do we have to introduce this variable or can we just inline it below?

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -28,20 +30,22 @@ jobs:
...
     outputs:
       matrix: ${{ steps.set.outputs.matrix }}
     steps:
       - id: set
         run: |
           # Decide which matrix entries to include based on event type
-          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/main" ]] || [[ "${{ github.event_name }}" == "schedule" ]]; then
-          # Include both CUDA and ROCm
+          # Runs CUDA and ROCm for push to main, cron schedule and PR label
+          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/main" ]] || [[ "${{ github.event_name }}" == "schedule" ]] || [[ "${HAS_8GPU_LABEL}" == "true" ]]; then
```
tianyu-l: am I understanding it correctly:
- with this change, if we want to run rocm test with a PR, we'd need to manually add the ciflow/8gpu tag (potentially after approving the test if PR author doesn't have write permission), and it won't disturb the already running tests.
- without this line, adding the ciflow/8gpu tag is a no-op

.github/pytorch-probot.yml
akashveramd: The `ciflow_push_tags` tells the bot which labels should map to git tags that trigger CI workflows.

It's called `8gpu` following Huy's suggestion on slack https://pytorch.slack.com/archives/C0A0E8ARXPZ/p1764843858514179

.github/workflows/integration_test_8gpu_features.yaml
akashveramd: I am currently using it in a couple of places; hence we may assign it to a variable.

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -3,9 +3,12 @@ name: 8 GPU Feature Tests
 on:
   push:
     branches: [ main ]
+    tags:
+      - ciflow/8gpu/*
     paths-ignore:
       - 'torchtitan/experiments/**'
   pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
tianyu-l: what does this do?

.github/pytorch-probot.yml
tianyu-l: It seems adding `ciflow/8gpu` label in a PR will automatically create a tag `ciflow/8gpu/[PR_number]`. I say so because I observed https://github.com/pytorch/torchtitan/releases/tag/ciflow%2F8gpu%2F2125

What is this `ciflow_push_tags` for? Is it for letting the bot know that the above happened and which tag to look at?

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -3,9 +3,12 @@ name: 8 GPU Feature Tests
 on:
   push:
     branches: [ main ]
+    tags:
+      - ciflow/8gpu/*
```
tianyu-l, line 7: What's this for -- why do we need both `pull_request` and `tags`?

tianyu-l (commented):
Is it true that
- for each PR a codeowner needs to explicitly add the `ciflow/8gpu` label, so that rocm test could be run?
- if a codeowner doesn't do this, the PR can be merged without rocm test passing?
- after CUDA test is run / finished, if we add the label, we can run rocm test without rerun the CUDA test?

.github/workflows/integration_test_8gpu_features.yaml
akashveramd: This runs the PR for these events. By-default PR runs for opened, synchronize & reopened when no types are used. However, I added this specifically for labeled, unlabeled types, since I want ROCm workflow to run when the label is added. If I just keep labeled, unlabeled then it'll override the default which we don't want.

tianyu-l: when we do label / unlabel, does it trigger another set of CUDA tests?

akashveramd: No. It's used for ROCm only.

.github/pytorch-probot.yml
akashveramd: That is correct. As per my understanding, when we add PR label, it'll automatically create & push tag. Because the ciflow_push_tags config tells the bot to create & push the **matching** tags so that tag-based CI workflows can run.

.github/workflows/integration_test_8gpu_features.yaml
akashveramd: As per my understanding the PR workflows and tag workflows are totally independent. Tags provides CI flow, meaning tags can be pushed to trigger CI runs on specific commits even after the PR is closed. They can also be used for versioning releases.
pull_request are used to run workflows when the PR is open.

.github/workflows/integration_test_8gpu_features.yaml
akashveramd: The code is outdated and refactored now.

.github/labeler.yml
```diff
@@ -0,0 +1,6 @@
+"ciflow/8gpu":
+- .ci/docker/**
+- .github/workflows/**
+- scripts/**
+- tests/**
+- torchtitan/**
```
tianyu-l: what are these for?

.github/workflows/integration_test_8gpu_features.yaml
tianyu-l: If I add a label to a PR, which is not ciflow/8gpu, would tests be triggered (again)?

.github/workflows/set-matrix.yaml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set.outputs.matrix }}
+    env:
+      # Event flags evaluated by github actions before the step runs:
+      IS_MAIN_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+      IS_SCHEDULE:  ${{ github.event_name == 'schedule' }}
+      IS_PR:        ${{ github.event_name == 'pull_request' }}
+      HAS_8GPU_LABEL: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ciflow/8gpu') }}
+      IS_8GPU_TAG:  ${{ startsWith(github.ref, 'refs/tags/ciflow/8gpu/') }}
```
tianyu-l, line 19: why do we need this, if we already have the one above?

akashveramd: Tags and labels are independent and different events. Hence, created a separate variable for it.

huydhn:
In the `pull_request` trigger run https://github.com/pytorch/torchtitan/actions/runs/20124461053, should there be only CUDA job?

akashveramd:
> In the `pull_request` trigger run https://github.com/pytorch/torchtitan/actions/runs/20124461053, should there be only CUDA job?

@huydhn: For only `pull_request` trigger it'll run CUDA job only. However, for sanity testing, I manually added the ciflow/8gpu PR label and because of that I believe it ran both CUDA & ROCm jobs. Is that correct?

.github/labeler.yml
akashveramd: If PR modifies any of these locations, then GitHub automatically adds the label.

akashveramd:
> Is it true that
> 
> * for each PR a codeowner needs to explicitly add the `ciflow/8gpu` label, so that rocm test could be run?
> * if a codeowner doesn't do this, the PR can be merged without rocm test passing?
> * after CUDA test is run / finished, if we add the label, we can run rocm test without rerun the CUDA test?

@tianyu-l: Here are my responses-
> for each PR a codeowner needs to explicitly add the `ciflow/8gpu` label, so that rocm test could be run?
 -  No, for the trusted authors who has write permissions the `ciflow/8gpu` label will be added automatically which will run ROCm tests. This is WIP. For unknown authors with only read permission, the reviewer needs to review the PR, approve CI run, then add the label manually. Which then will run the ROCm tests. However, on cron schedule and when push to main the ROCm tests will run irrespective of PR label.
 
 > if a codeowner doesn't do this, the PR can be merged without rocm test passing?
 - Yes, for unknown authors. However, isn't it the same for PyTorch, i.e. ROCm workflow needs PR labels to run?
    cc: @huydhn 
 
 > after CUDA test is run / finished, if we add the label, we can run rocm test without rerun the CUDA test?
 -  Yes, CUDA tests won't re-run when the label is added. But something I have noticed is when I manually add the label after the CUDA workflow has completed running, the CUDA workflow will disappear from the checks. I believe it's happening because CUDA & ROCm shares the same workflow. I'll have to see how it works when automatic label is in place.

.github/pytorch-probot.yml
wwwjn: Shat is the bot doing? Is it doing the auto labeling?

wwwjn (commented):
Hi @akashveramd , trying to catch up what will happen after this PR:

As a person with write access to torchtitan:
- My PR should be automatically taged with `ciflow/8gpu` (I saw we have labeler.yml and a bot, is it helping our doing the auto labeling?)
- My PR needs to pass both rocm tests and CUDA tests to get merged, otherwise it will be blocked

As an unauthorized user:
- If my PR didn't get taged with `ciflow/8gpu`, I just need to pass CUDA tests to get my PR merged
- If my PR get labeled with `ciflow/8gpu`, I have to pass both rocm and CUDA tests before get merged

As a repo maintainer:
- We should review and add `ciflow/8gpu` tag for unauthorized user's PR that we think it's safe to run CI tests, and approve CI workflow to run

Is this a correct understanding / summary?

.github/pytorch-probot.yml
akashveramd: The .github/pytorch-probot.yml connects labels to tags. When it sees ciflow/8gpu label in PR, it creates the tag. The automatic labeling is handled by .github/labeler.yml. However, currently I don't see the automatic label getting added when I make changes to PR. @wwwjn Do you know if the totrchtitan repo is setup correctly to put this label automatically or do I need any additional changes?

PR #2115, opened 2025-12-05
Enables parsing of --compile.components through CLI

syed-ahmed:
Without this PR, I'm not able to pass `--compile.components=model,loss`. Tested using `python -m torchtitan.config.manager --compile.components=model,loss`.

### DIFF
--- a/torchtitan/config/job_config.py
+++ b/torchtitan/config/job_config.py
@@ -656,9 +656,7 @@ class Compile:
     enable: bool = False
     """Whether to apply torch.compile"""
 
-    components: list[str] = field(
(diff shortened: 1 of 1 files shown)


fegin:
You can try --compile.components "loss" "model". This works for me.

syed-ahmed:
Thanks @fegin ! That works for me. I'm happy with either. May be `--compile.components=model,loss` is more uniform with the other options, but keeping `--compile.components model loss` does inform the user what exact components are supported. So totally up to you if you wanna merge or close it.

tianyu-l:
> May be --compile.components=model,loss is more uniform with the other options

sounds right to me

PR #2116, opened 2025-12-05
[Not Ready]Let CUDA and ROCm read different loss result

fegin:
As title.


fegin:
Waiting for https://github.com/pytorch/torchtitan/pull/2118 so that I can trigger ROCm CI to verify.

PR #2117, opened 2025-12-05
Remove caching for attention masks

wwwjn:
We remove the lru_cache for attention masks, because in get_attention_mask() function, `and_masks(*mask_mods)` will return different object id. `create_attention_mask` will use all parameters as cache key, and new object id will always cause cache miss.

Before the change: (llama3 debugmodel_flex_attn)
Screenshot 2025-12-09 at 1 27 45 PM

After the change:
Screenshot 2025-12-09 at 1 29 56 PM

### DIFF
--- a/torchtitan/models/attention.py
+++ b/torchtitan/models/attention.py
@@ -163,22 +163,19 @@ class ScaledDotProductAttentionWrapper(torch.nn.Module):
             return F.scaled_dot_product_attention(q, k, v, scale=scale, is_causal=True)
 
 
-# We cannot do inner function/closure because we won't be able to cache it --
-# if we an inner function, a new closure will be created every time
-# `get_causal_mask_mod` is called.
-def _causal_mask(
-    b: torch.Tensor, h: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
-) -> torch.Tensor:
-    """Causal mask that prevents attention to future tokens."""
-    return q_idx >= kv_idx
-
-
 def get_causal_mask_mod() -> _mask_mod_signature:
     """Returns a causal mask modifier for flex attention.
 
     Returns:
         A mask modifier function that implements causal masking.
     """
+
+    def _causal_mask(
+        b: torch.Tensor, h: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
+    ) -> torch.Tensor:
+        """Causal mask that prevents attention to future tokens."""
+        return q_idx >= kv_idx
+
     return _causal_mask
 
 


torchtitan/models/attention.py
```diff
@@ -253,13 +277,8 @@ def get_sliding_window_mask_mod(window_size: int) -> _mask_mod_signature:
...
 
-    def sliding_window_mod(
-        b: torch.Tensor, h: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
-    ) -> torch.Tensor:
-        # Window mask: can only attend within the window
-        # q_idx - kv_idx < window_size ensures we look at most window_size-1 tokens back
-        return (kv_idx <= q_idx) & (q_idx - kv_idx < window_size)
-
+    # Use functools.partial to bind window_size while keeping the function cacheable
+    sliding_window_mod = functools.partial(_sliding_window_mask, window_size)
```
tianyu-l: I might have missed this point -- calling `partial` the 2nd time with the same `window_size` may not give cache hit, and if so this approach won't work. Could you verify?

cc @fegin

wwwjn: You are right, I missed it. I checked the object id when calling the `partial` 2nd time with the same window_size, and the id are different.

So I changed to using `functiontool.lru_cache` to do explicit caching, this would use `window_size` as cache key. I  verified the ids are the same. Wdyt about this solution?

torchtitan/models/attention.py
```diff
@@ -234,19 +234,9 @@ def blocked_mask_mod(
     return blocked_mask_mod
 
 
+@functools.lru_cache(4)
 def get_sliding_window_mask_mod(window_size: int) -> _mask_mod_signature:
-    """Creates a sliding window mask that only attends to tokens within a fixed window size.
-
-    This implements causal sliding window attention where each token can only attend to:
```
tianyu-l, line 240: can we add the comments back?

torchtitan/models/attention.py
```diff
@@ -234,19 +234,9 @@ def blocked_mask_mod(
     return blocked_mask_mod
 
 
+@functools.lru_cache(4)
```
tianyu-l: I'm worried that for the only use case
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/gpt_oss/model/model.py#L364
an `and_masks` will be called for each iteration, whose results will have a different object id for each call (at the beginning of each layer).

In general, the caching mechanism around masks sounds not very robust. The per-iteration overhead might be fine if caching is removed. Should we remove all caching altogether? cc @fegin

wwwjn: I see, looks like we are using and_masks() for almost all models now. 

If the 1and_mask` returns a new object id for each call, then the lru cache around `create_attention_mask` will always miss (because the returned if from add_mask will be part of cache key)

tianyu-l: That's my guess. Could you verify? If so, I'd recommend we remove all the lru cache annotation altogether for better readability and some memory saving (because we no longer maintain a cache which is never used).

wwwjn: Yes I already verified that and_mask will return different object id everytime. I will move towards removing all cache annotation.

torchtitan/models/attention.py
```diff
@@ -235,18 +234,7 @@ def blocked_mask_mod(
...
-    - Itself (current token)
-    - Up to `window_size - 1` previous tokens
-    Args:
-        window_size: The maximum number of tokens to attend to (including current token).
-                    Must be >= 1. A window_size of 1 means attend only to self.
-
-    Returns:
-        A mask modifier function that implements causal sliding window masking.
-    """
+    """Creates a sliding window mask that only attends to tokens within a fixed window size"""
```
tianyu-l: can we add the deleted comments back?

fegin:
Yes, we should remove the lru_cache. It was added before we use add_mask to hope that we can cache some masks (e.g., causal mask). This very hard to achieve in the API level. It may be better to let users decide what to cache.

PR #2118, opened 2025-12-05
Remove the hack for SAC + FlexAttention

fegin:
* __->__ #2118

PyTorch can now support torch.compile inside the SAC region even if torch.compile is not used to wrap SAC. This PR removes the workaround to ensure torch.compile works with Flex

### DIFF
--- a/torchtitan/models/attention.py
+++ b/torchtitan/models/attention.py
@@ -98,6 +98,7 @@ class FlexAttentionWrapper(torch.nn.Module):
 
     _compiled_flex_attn: ClassVar[Callable] = torch.compile(
         flex_attention,
+        # This options also encapsulate max-autotune-no-cudagraphs.
         options={
             "wrap_inductor_compiled_regions": True,
             "max_autotune": True,
--- a/torchtitan/models/llama3/train_configs/debug_model.toml
+++ b/torchtitan/models/llama3/train_configs/debug_model.toml
@@ -62,7 +62,7 @@ async_mode = "disabled"  # ["disabled", "async", "async_with_pinned_mem"]
 
 [activation_checkpoint]
 mode = "selective"  # ["none", "selective", "full"]
-selective_ac_option = 'op'  # 'int' = ac every positive int layer or 'op', ac based on ops policy
+selective_ac_option = '2'  # 'int' = ac every positive int layer or 'op', ac based on ops policy
 
 [compile]
 enable=false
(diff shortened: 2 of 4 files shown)


torchtitan/models/llama3/train_configs/debug_model.toml
```diff
@@ -62,7 +62,7 @@ async_mode = "disabled"  # ["disabled", "async", "async_with_pinned_mem"]
 
 [activation_checkpoint]
 mode = "selective"  # ["none", "selective", "full"]
-selective_ac_option = '2'  # 'int' = ac every positive int layer or 'op', ac based on ops policy
+selective_ac_option = 'op'  # 'int' = ac every positive int layer or 'op', ac based on ops policy
```
tianyu-l: Is this intended? I think it's OK to change this but want to confirm.

tianyu-l (approved):
LGTM. I wonder if you have verified it "works", i.e. it doesn't invalidate SAC / compile anymore?

torchtitan/models/llama3/train_configs/debug_model.toml
fegin: oops, accidentally commit this.

fegin:
Yes, I verified it works with llama3 and llama4. FlexAttention is compiled within SAC.

torchtitan/models/llama3/train_configs/debug_model.toml
soulitzer: Change to config intentional?

torchtitan/models/attention.py
```diff
@@ -97,7 +97,13 @@ class FlexAttentionWrapper(torch.nn.Module):
     """
 
     _compiled_flex_attn: ClassVar[Callable] = torch.compile(
-        flex_attention, mode="max-autotune-no-cudagraphs"
+        flex_attention,
+        options={
+            "wrap_inductor_compiled_regions": True,
+            "max_autotune": True,
+            "coordinate_descent_tuning": True,
```
soulitzer, line 105: Noob question: is this coordinate_descent_tuning also part of the "mode=max-autotune-no-cudagraphs"  -> "options={...}" change?

tianyu-l: forgot to ask: what's the context of this change to "options={}"?

fegin: Yes, https://github.com/pytorch/pytorch/blob/cf7bab873fa55051e1806f8db0c3f90dea452ac5/torch/_inductor/__init__.py#L361

We cannot do mode and options at the same. torch.compile forbid it. According to here, `max-autotune-no-cudagraphs` equals to these three options.

PR #2119, opened 2025-12-05
Run vLLM inference using torchtitan model definition (single GPU)

wwwjn:
As titled, put it in deterministic RL folder


torchtitan/experiments/deterministic_vllm_rl/models/attention.py
```diff
@@ -194,3 +199,388 @@ def backward(ctx, grad_output):
         output = output.transpose(1, 2)
 
         return output
+
+
+class VLLMPagedFlashAttention(torch.nn.Module):
```
tianyu-l: What's this for? I thought we only run inference.

torchtitan/experiments/deterministic_vllm_rl/README.md
```diff
@@ -211,29 +211,110 @@ This implementation uses the same kernels for both rollouts (vLLM) and training
...
+```
+
+## TODOs
+1. Rewrite attention part to use vllm.Attention() with backward as the only attention path.
+2. Integrate with simple_rl.py to run end-to-end RL with one canonical model definition.
+3.  Leverage batch-invariant kernels into model definition.
+
+
+
+# Project Structure
```
tianyu-l: let's not maintain this

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
```diff
@@ -3,366 +3,370 @@
...
-        # Use vLLM's RMSNorm with gradient support for training
-        return rms_norm_with_gradients(x, self.weight, self.eps)
 
-    def reset_parameters(self):
-        nn.init.ones_(self.weight)
+logger = init_logger(__name__)
 
 
-class FeedForwardVLLMCompat(nn.Module):
+class TorchTitanQwen3ForCausalLM(nn.Module):
```
tianyu-l: we should aim for at most 1 general wrapper for all models -- we shouldn't have 1 wrapper for each model.

torchtitan/models/qwen3/infra/parallelize.py
```diff
@@ -242,14 +243,23 @@ def apply_non_moe_tp(
...
+            ),  # NOTE(jianiw): Try using Tensor instead of DTensor
             "attention.wk": colwise_parallel(use_local_output=False),
-            "attention.wv": colwise_parallel(use_local_output=False),
-            "attention.q_norm": SequenceParallel(sequence_dim=2),
-            "attention.k_norm": SequenceParallel(sequence_dim=2),
+            "attention.wv": colwise_parallel(use_local_output=True),
+            "attention.q_norm": SequenceParallel(sequence_dim=2, use_local_output=True),
+            "attention.k_norm": SequenceParallel(sequence_dim=2, use_local_output=True),
+            # NOTE(jianiw): manually convert the output to be Shard(1)
+            "attention.inner_attention": PrepareModuleOutput(
```
tianyu-l: What's this for?

tianyu-l (commented):
left some comments, ignore if you still plan to change

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
```diff
@@ -3,366 +3,370 @@
...
-        self.attention.init_weights(self.weight_init_std)
-        self.feed_forward.init_weights(self.weight_init_std)
+        # Store positions in forward context for attention layers
+        # Also convert positions to 2D format
+        # TODO: The position id information is not properly used yet
+        if positions is not None:
+            positions_2d = positions.unsqueeze(0)  # [total_tokens] -> [1, total_tokens]
 
+        # Get embeddings from 2D tokens
+        h = self.model.tok_embeddings(tokens_2d)  # [1, total_tokens, hidden_size]
```
acisseJZhong: qq: curious are the input tokens always have bz=1?

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
wwwjn: By design this class should be able to run both training and inference in the future, so we have one single attention class. Now it's only used when running inference. I will clean up and remove non-inference related stuff. cc @zhxchen17

torchtitan/models/qwen3/infra/parallelize.py
wwwjn: I patched inner_attetion to be vllm.Attention(), but it doesn't compatible with DTensor - So I changed this part to make then local_tensor. I will make a copy / patch of the tp plan instead of change the main model code

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
wwwjn: Yes that's the goal! The current wrapper is specific to Qwen3 model as we access each model layers by name, let me think about how to design the interface

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
wwwjn: I'm not 100% sure about this part, but I guess no?  I only tried single prompts for now but if batched inference is enabled, I will leave a ToDo here for now

torchtitan/models/qwen3/model/model.py
```diff
@@ -59,24 +59,54 @@ def rotate_half(x: torch.Tensor) -> torch.Tensor:
...
 
     Returns:
         torch.Tensor: Reshaped frequency tensor.
     """
     ndim = x.ndim
     assert ndim > 1
     _, seqlen, _, head_dim = x.shape
+
+    # Extend rope_cache if needed (e.g., during vLLM profiling with 2x max_seq_len)
+    if seqlen > rope_cache.shape[0]:
```
wwwjn: NOTE: Qwen3 0.6B max_seq_len is 4096, but during prefill stage, vllm will do`Chunked prefill is enabled with max_num_batched_tokens=16384.`. But this patch needs to be move into vLLM model wrapper

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
zhxchen17: Suppose each model author defines the model in slightly different ways, then it seems to up to the ppl who do RL to make the adaptation work like we prototyped here. 

So I guess we need some sort of contract baked in the authoring time. e.g. Models should annotate/implement their attention layers in certain ways (having some sort of base class or special methods?)

tianyu-l: Agreed with @zhxchen17.
I think we can start by adding a `BaseTransformer` class (in torchtitan/protocols/model.py) with the standard layers defined, so that other model (text ones) can inherit, e.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama4/model/model.py#L425

but this won't stop model from not using the layers (e.g. tok_embeddings). We can make them less error-prone by removing the `forward` function in subclasses since they are identical.

torchtitan/experiments/rl/unified/infer.py
```diff
@@ -0,0 +1,117 @@
...
+
+# Import and register the TorchTitan vLLM plugin
+from torchtitan.experiments.deterministic_vllm_rl.register import register
+
+# Register TorchTitan models with vLLM.
+# NOTE(jianiw): We could use plug-in system instead: https://docs.vllm.ai/en/latest/design/plugin_system/
+register()
+
+
+def parse_args():
```
tianyu-l, line 20: not urgent, but we should use "our" config system in the long term

wwwjn: Curretnly the entry point is vllm engine, so we are taking the config from whatever vllm engine passed to us. Let me check vllm engine see if there's anything we could do

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
```diff
@@ -3,366 +3,436 @@
...
+            )
 
-    def __init__(self, model_args: Qwen3ModelArgs):
-        super().__init__()
-        self.n_heads = model_args.n_heads
-        self.n_kv_heads = (
-            model_args.n_heads
-            if model_args.n_kv_heads is None
-            else model_args.n_kv_heads
+    def _replice_with_vllm_paged_attention(self, model_args):
```
acisseJZhong: can we expose this function as a util function? It could also be reused for the trainer model

torchtitan/experiments/deterministic_vllm_rl/models/qwen3/model_vllm_compat.py
wwwjn: Cool! The refactor of making a `BaseTransformer` class can be handled in a separate PR, this would need an alignment between `rope_cache` and `freqs_cis` naming as well

Also I was thinking how to design the 1 single general wrapper of vllm model. 
There are 4 things are model specific and need to be pluged into the general wrapper class, before register model to vllm:
1. model class itself
2. model_args class
3. model's parallel plan (mainly TP. PP needs to be considered separately)
4. state dict adapter

I refactored the code into a basic wrapper and plugin components by inheriting when registering the model. Wdyt?

tianyu-l: > this would need an alignment between rope_cache and freqs_cis naming as well

Why? Is it because we put freqs_cis generation outside model for CP? I think @fegin has a plan to move it back into the model.

wwwjn: No it's not related to CP, I was thinking we need to access `self.rope_cache` / `self.freqs_cis` during the forward of `BaseTransformer`: https://github.com/pytorch/torchtitan/blob/f1d41a1179a62e5274c565b080282ba7ccfeab2c/torchtitan/models/qwen3/model/model.py#L573

torchtitan/experiments/deterministic_vllm_rl/README.md
```diff
@@ -211,37 +211,83 @@ This implementation uses the same kernels for both rollouts (vLLM) and training
...
+
+```
+# Set CUDA version environment variable
+export CUDA_HOME=/usr/local/cuda-12.4
+export PATH=/usr/local/cuda-12.4/bin:$PATH
+export LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64:$LD_LIBRARY_PATH
+
+# Clean previous build
+rm -rf build dist *.egg-info
+pip uninstall -y vllm
```
acisseJZhong: nit: uv pip install?

torchtitan/experiments/rl/unified/vllm_wrapper.py
```diff
@@ -0,0 +1,376 @@
...
+
+        for layer_name, layer in self.model.layers.items():
+            if not hasattr(layer, "attention"):
+                raise ValueError(f"Layer {layer_name} must have .attention attribute")
+
+            # Create vLLM paged attention
+            vllm_attn = VLLMPagedFlashAttention(
+                hidden_size=model_args.dim,
+                num_heads=model_args.n_heads,
+                num_kv_heads=model_args.n_heads,  # Use n_heads (already replicated)
```
acisseJZhong, line 126: qq: why not model_args.n_kv_heads?

torchtitan/experiments/deterministic_vllm_rl/models/base_wrapper.py
```diff
@@ -0,0 +1,376 @@
...
+            rope_attr = None
+
+        # Extend RoPE cache if needed (vLLM profiling may use 2x max_seq_len)
+        if positions is not None:
+            max_position = positions.max().item()
+        else:
+            max_position = 0
+
+        rope_cache = self._extend_rope_cache_if_needed(rope_attr, max_position)
+        rope_cache = rope_cache[positions]
```
acisseJZhong: you could delete this, and directly use `h = layer(h, rope_cache, attention_masks=None, positions=positions)` below

torchtitan/experiments/deterministic_vllm_rl/models/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    logger.info(
+        f"Created ParallelDims from vLLM config: "
+        f"DP={parallel_dims.dp_replicate}, TP={parallel_dims.tp}, "
+        f"CP={parallel_dims.cp}, PP={parallel_dims.pp}"
+    )
+
+    return parallel_dims
+
+
+def build_device_mesh_and_parallelize(
```
acisseJZhong: curious how will this function be used? Won't VLLM engine handle TP=2 for us?

torchtitan/experiments/deterministic_vllm_rl/infer.py
```diff
@@ -0,0 +1,118 @@
...
+    # 4. Build device mesh and apply parallelization via build_device_mesh_and_parallelize()
+    # 5. Load model weights and prepare for inference
+    print("Initializing vLLM engine...")
+    llm = LLM(
+        model=args.model,  # Model checkpoint path
+        hf_overrides=hf_overrides,
+        dtype="bfloat16",
+        trust_remote_code=True,
+        enforce_eager=True,  # Use eager mode
+        enable_prefix_caching=False,  # Disable kv cache for now
```
tianyu-l: what's the obstacle for KV cache, now that @acisseJZhong 's positions PR has landed.

torchtitan/experiments/deterministic_vllm_rl/infer.py
```diff
@@ -0,0 +1,118 @@
...
+    # The LLM initialization will internally:
+    # 1. Load TrainSpec for Qwen3 (from register())
+    # 2. Create TorchTitanVLLMModel instance
+    # 3. Process parallelism settings via process_parallelism_settings()
+    # 4. Build device mesh and apply parallelization via build_device_mesh_and_parallelize()
+    # 5. Load model weights and prepare for inference
+    print("Initializing vLLM engine...")
+    llm = LLM(
+        model=args.model,  # Model checkpoint path
+        hf_overrides=hf_overrides,
```
tianyu-l: can we remove the intermediate `hf_overrides` and inline the dict creation here? It's only 1 element dict.

torchtitan/experiments/deterministic_vllm_rl/infer.py
```diff
@@ -0,0 +1,118 @@
...
+
+def main():
+    args = parse_args()
+
+    print("=" * 80)
+    print("INITIALIZING vLLM WITH TORCHTITAN QWEN3 MODEL ")
+    print("=" * 80)
+    print(f"Model: {args.model}")
+    print(f"Tensor Parallel Size: {args.tensor_parallel_size}")
+    print()
```
tianyu-l: can we use logger instead, similar to train.py and avoid some of the verbose "=" logging?

torchtitan/experiments/deterministic_vllm_rl/infer.py
```diff
@@ -0,0 +1,118 @@
...
+
+from vllm import LLM, SamplingParams
+
+# Import models module - this automatically registers TorchTitan models with vLLM
+from torchtitan.experiments.deterministic_vllm_rl import models  # noqa: F401
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Run TorchTitan Qwen3 model inference with vLLM",
```
tianyu-l: not specific to Qwen3?

torchtitan/experiments/rl/unified/infer.py
tianyu-l: wait how is it related to vllm config system? You are just using them as is in `args = parse_args()`.

torchtitan/experiments/rl/unified/vllm_wrapper.py
```diff
@@ -0,0 +1,376 @@
...
+    - HF config to TorchTitan model args mapping
+    - Attention replacement with vLLM paged attention
+    - Tensor parallelism setup
+    - Weight loading from HF checkpoints
+    - vLLM forward/compute_logits interface
+    """
+
+    is_text_generation_model = True  # Required for vLLM runner validation
+    supports_pp = False  # Pipeline parallelism not supported yet
+    supports_multimodal = False
```
tianyu-l, line 53: where are these fields used?

torchtitan/experiments/deterministic_vllm_rl/models/base_wrapper.py
```diff
@@ -0,0 +1,376 @@
...
+        self,
+        hidden_states: torch.Tensor,
+        sampling_metadata=None,
+    ) -> torch.Tensor | None:
+        """Compute logits from hidden states."""
+        h = self.model.norm(hidden_states)
+        logits = self.model.output(h)
+
+        if isinstance(logits, DTensor):
+            logits = logits.full_tensor()
```
tianyu-l: Have you verified this for TP, or is it only working for single gpu? If it's the latter, add a TODO around such conversions.

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
```diff
@@ -194,3 +198,188 @@ def backward(ctx, grad_output):
...
+            cache_config = (
+                config.cache_config if hasattr(config, "cache_config") else None
+            )
+
+            # Generate unique prefix for this attention layer
+            # vLLM expects format "layers.X" for layer index extraction
+            layer_idx = next(VLLMPagedFlashAttention._layer_counter)
+            prefix = f"layers.{layer_idx}"
+
+            self.vllm_attn = Attention(
```
tianyu-l: why can't we always use `VLLMPagedFlashAttention`?

torchtitan/experiments/rl/unified/__init__.py
```diff
@@ -6,8 +6,77 @@
...
+    # Create dynamic model class directly from TrainSpec components
+    class TorchTitanVLLMModelFromSpec(TorchTitanVLLMModel):
+        """Dynamically created vLLM model from TrainSpec."""
+
+        def __init__(self, *, vllm_config, prefix=""):
+            super().__init__(
+                model_cls=train_spec.model_cls,
+                model_args_cls=model_args_cls,
+                state_dict_adapter=train_spec.state_dict_adapter,
+                parallelize_fn=train_spec.parallelize_fn,
```
tianyu-l, line 60: It seems we need these fields and the wrappers `TorchTitanVLLMModel` / `TorchTitanVLLMModelFromSpec` because we rely on vllm's `LLM()` api to create the model.

This is hacky and making things complicated as we are dumping a lot of logic (originally in train.py and checkpoint.py) to the model code itself.

I feel this is unnecessary if our end goal is to use the engine part of vLLM, not the model init part.

torchtitan/experiments/rl/unified/vllm_wrapper.py
tianyu-l, line 1: Not sure what part is "base", could you explain? Maybe call it vllm_wrapper.py?

tianyu-l (changes requested):

torchtitan/experiments/rl/unified/vllm_wrapper.py
```diff
@@ -0,0 +1,376 @@
...
+        # Replace attention with vLLM paged attention
+        self._replace_with_vllm_paged_attention(model_args)
+
+        # Create ParallelDims from vLLM config and apply parallelization
+        # NOTE: We need to apply parallelize within model.__init__ because w
+        parallel_dims = create_parallel_dims_from_vllm_config(vllm_config)
+        if parallel_dims.tp_enabled:
+            self.world_mesh = parallel_dims.world_mesh
+            tp_mesh = self.world_mesh["tp"]
+            parallelize_fn(
```
acisseJZhong, line 94: wondering why do we parallelize the model during init? This model is used during vllm inference, and I thought VLLM has its own TP impl?

torchtitan/experiments/rl/unified/vllm_wrapper.py
wwwjn: Yes, these fields are required by vLLM engine

torchtitan/experiments/rl/unified/vllm_wrapper.py
wwwjn: We want to apply our TP instead of using vLLM TP implementation, we can not have direct access to the model later once LLM() initialized, so we can not apply TP later

torchtitan/experiments/rl/unified/vllm_wrapper.py
wwwjn: I was thinking a proper name of this class, because technically it's not a model, we are actually wrapping torchtitan model definition into a `model_runner ` concept: https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/worker_base.py#L85

torchtitan/experiments/rl/unified/__init__.py
wwwjn: >  dumping a lot of logic (originally in train.py and checkpoint.py) to the model code itself

Agreed, the main blocker is that we need to have control of how Worker instantiate a model. 

According to vllm design , this class is not only a model nn.module,  but a `model_runner` https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/worker_base.py#L85, that's why it has `load_weights` function

torchtitan/experiments/rl/unified/vllm_wrapper.py
wwwjn: Good question! This is because the VLLMPagedFlashAttention is used to patch `inner_attention`, which already after `repeat_kv`: [VLLMPagedFlashAttention](https://github.com/pytorch/torchtitan/blob/fbafd44da2baef0afac58989f07d799c4251bdef/torchtitan/models/qwen3/model/model.py#L259)

torchtitan/experiments/deterministic_vllm_rl/models/utils.py
wwwjn: > VLLM engine handle TP=2 for us?

vllm applies TP by patching each module: https://docs.vllm.ai/en/latest/contributing/model/basic/#3-optional-implement-tensor-parallelism-and-quantization-support, which also changes model definition and happens during model initialization


This function is calling `parallelize_qwen3` function from `qwen3/infra/parallelize.py`, according to parallel_dims, it will apply different parallism

torchtitan/experiments/rl/unified/infer.py
wwwjn: This `args` is only for `infer.py` script, it will pass args into vllm engine LLM() , and vllm engine will create a VLLMConfig instance internally, and pass to our model wrapper

torchtitan/experiments/deterministic_vllm_rl/infer.py
wwwjn: Yes, let me refactor this `infer.py` more close to our `train.py`, it was started from a simple test script

torchtitan/experiments/rl/unified/infer.py
tianyu-l: > it will pass args into vllm engine LLM()

I don't think it's passing the `args` to LLM(). What would be different if we use our config manager to construct `args`?

torchtitan/experiments/deterministic_vllm_rl/infer.py
wwwjn: Removed this line, it was legacy

torchtitan/experiments/deterministic_vllm_rl/models/base_wrapper.py
wwwjn: Solved! Thank you!

torchtitan/experiments/deterministic_vllm_rl/models/base_wrapper.py
wwwjn: Currently this PR only works for single GPU, and I will add TP in a following PR

torchtitan/experiments/deterministic_vllm_rl/unified/__init__.py
```diff
@@ -6,8 +6,77 @@
...
+        train_spec: TorchTitan TrainSpec containing model components
+        model_name: Name to register in vLLM (e.g., "Qwen3TorchTitanForCausalLM")
+
+    """
+    from vllm.model_executor.models.registry import ModelRegistry
+
+    # Extract model_args from TrainSpec
+    # TrainSpec has model_args as a Mapping, get the first value
+    if isinstance(train_spec.model_args, dict):
+        model_args_cls = type(next(iter(train_spec.model_args.values())))
```
tianyu-l: This is taking the "debugmodel". Any reason why we don't make this general?

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
```diff
@@ -194,3 +198,170 @@ def backward(ctx, grad_output):
...
+    Wrapper around vLLM's Attention with custom backward pass.
+
+    Forward: Uses vLLM's optimized Attention layer (flash attention kernels)
+    Backward: Custom implementation for deterministic gradients
+
+    TODO: This class need to be future refined.
+    """
+
+    # Class variable for auto-generating unique layer names (thread-safe)
+    _layer_counter = itertools.count()
```
tianyu-l: Why we do this instead of passing an arg in during construction. 

Also worrying about PP compatibility. Maybe not urgent.

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
```diff
@@ -194,3 +198,170 @@ def backward(ctx, grad_output):
...
+        head_dim: int,
+        scale: float | None = None,
+        causal: bool = True,
+    ) -> None:
+        super().__init__()
+
+        self.hidden_size = hidden_size
+
+        # Handle tensor parallelism: adjust num_heads and num_kv_heads for TP
+        # NOTE(jianiw): As we use local tensor for this region, we need to manually
```
tianyu-l: reverted?

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
```diff
@@ -194,3 +198,170 @@ def backward(ctx, grad_output):
...
+
+        from vllm.config import get_current_vllm_config
+        from vllm.logger import init_logger
+
+        logger = init_logger(__name__)
+        vllm_config = get_current_vllm_config()
+        tp_size = vllm_config.parallel_config.tensor_parallel_size
+
+        if tp_size > 1:
```
tianyu-l: Did you copy these code from somewhere?

If we use torchtitan trainer, we shouldn't define them this way.

If it is only for inference, let's name it that way. But then I'm curious, wouldn't vllm already provide such a class?

torchtitan/experiments/deterministic_vllm_rl/unified/utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+logger = init_logger(__name__)
+
+
+def create_parallel_dims_from_vllm_config(vllm_config: VllmConfig) -> ParallelDims:
+    """
+    Create ParallelDims from vLLM config.
+
+    Maps vLLM parallelism settings to TorchTitan's ParallelDims dataclass.
```
tianyu-l: add comment on why we need this method -- it is used for creating a vllm-compatible model from titan model definition, because vllm doesn't separate model creation and parallelism application and instead requires parallelization to be done inside model constructor.

torchtitan/experiments/deterministic_vllm_rl/models/utils.py
tianyu-l: function not used? if so let's remove

torchtitan/experiments/rl/unified/vllm_wrapper.py
```diff
@@ -0,0 +1,377 @@
...
+
+        # TODO: This part is to work with TP integration
+        if isinstance(logits, DTensor):
+            logits = logits.to_local()
+
+        return logits
+
+    def load_weights(self, weights_iter):
+        """
+        Load weights from HF checkpoint using the provided state dict adapter.
```
tianyu-l, line 286: comment why we need such a method. I guess it's because vllm engine would call it to load weights before running.

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
wwwjn: This part is from llama4, and my understanding is vllm need num_local_heads instead of  num_heads. Currently num_heads are passed from model args

wwwjn:
> After reading the code, I prefer we separate the inference code in this PR from the existing RL work, e.g. we can create a folder called `inference`. This will make our mental model more clear.

I want to discuss where we should put this work, long-term wise we could also make it part of RL system, so naming it `inference` at the beginning we might need to change the name later.

> My main question is: why we need `VLLMPagedFlashAttention` class? I feel there must be something in vllm that already does this.

~~After offline discussion yesterday, i think we would directly use vllm.Attention() in inference, and use `VLLMCompatibleFlashAttention` during training,  and I will update and remove `VLLMPagedFlashAttention` class.~~
- ~~Note: `VLLMCompatibleFlashAttention` added backward for vllm's flash attention kernel~~


Update: After rethinking, the attention plan should be : cc @zhxchen17 
- During training: Directly patch to `VLLMCompatibleFlashAttention`
- During inference, use `vllm.Attention()`, but we still need a wrapper around vllm.Attention() during inference (`VLLMPagedFlashAttention`). Because: 
    - Need to correctly handle reshape/view before
    - When TP is applied, it needs to handle DTensor <-> plain tensor conversion before sending into vllm.Attention()

tianyu-l:
> I want to discuss where we should put this work, long-term wise we could also make it part of RL system, so naming it inference at the beginning we might need to change the name later.

This is under very active development, so renaming in the near future shouldn't be a concern.

Since we are moving towards a more modern approach (running torchtitan model during inference), my ask is to clearly separate them (the so-called vllm-compat model and the model in this PR) for us and for anyone who's going to work on it. I want to avoid the situation where people spend long time to tell what's the difference between this and that. Having a separate folder sounds the most natural thing to do. (We discussed if we want to have another experiment, but that sounds a bit too separate, idk).

You can give it a more longer-term name than `inference` (e.g. `unified`?) to better represent its nature.

torchtitan/experiments/rl/unified/infer.py
wwwjn: We could also use our job_config + config_manager to handle the inference args. For now I didn't use it as we are only using 5 parameters for the simple script, I will refactor it to use the same config system as titan

torchtitan/experiments/deterministic_vllm_rl/unified/__init__.py
wwwjn: Previously, this is taking the class of model_args (Qwen3ModelArgs) and initialized a model_args class from hugging face configs. I've refactor this to initialize args from torchtitan model_args.

wwwjn:
> > I want to discuss where we should put this work, long-term wise we could also make it part of RL system, so naming it inference at the beginning we might need to change the name later.
> 
> This is under very active development, so renaming in the near future shouldn't be a concern.
> 
> Since we are moving towards a more modern approach (running torchtitan model during inference), my ask is to clearly separate them (the so-called vllm-compat model and the model in this PR) for us and for anyone who's going to work on it. I want to avoid the situation where people spend long time to tell what's the difference between this and that. Having a separate folder sounds the most natural thing to do. (We discussed if we want to have another experiment, but that sounds a bit too separate, idk).
> 
> You can give it a more longer-term name than `inference` (e.g. `unified`?) to better represent its nature.

Thank you , that's very helpful! There are 2 different approaches under vllm_deterministic_rl folder: 1) create vllm-compatible model, 2) unified model definition. So I split them into 2 separate folders with an brief intro in README

torchtitan/experiments/deterministic_vllm_rl/models/attention.py
wwwjn: Removed this part by directly passing the layer_id into the constructor

torchtitan/experiments/deterministic_vllm_rl/README.md
```diff
@@ -1,262 +1,12 @@
-# Deterministic RL Training with vLLM
+# Deterministic vLLM RL Training
```
tianyu-l: vLLM RL training doesn't sound right, maybe revert

torchtitan/experiments/deterministic_vllm_rl/__init__.py
tianyu-l, line 1: do we need this file? if not let's delete it for now

torchtitan/experiments/deterministic_vllm_rl/unified/vllm_wrapper.py
```diff
@@ -0,0 +1,331 @@
...
+from torchtitan.models.qwen3.model.model import precompute_rope_cache
+from torchtitan.protocols.model import BaseModelArgs, ModelProtocol
+from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
+
+from .utils import create_parallel_dims_from_vllm_config
+
+
+logger = init_logger(__name__)
+
+ParallelizeFunction: TypeAlias = Callable[..., nn.Module]
```
tianyu-l: can we import from protocols folder instead of defining it again?

torchtitan/experiments/rl/README.md
```diff
@@ -1,262 +1,12 @@
...
-
-- [vLLM Documentation](https://docs.vllm.ai/)
-- [Flash Attention Paper](https://arxiv.org/abs/2205.14135)
-- [PPO Algorithm](https://arxiv.org/abs/1707.06347)
-- [GRPO: Group Relative Policy Optimization](https://arxiv.org/abs/2402.03300)
-
-## License
-
-This code is licensed under the BSD-style license found in the LICENSE file in the TorchTitan repository root directory.
+2. unified/ - Unified approach
```
tianyu-l, line 10: say something like work in progress and only supports inference for now?

torchtitan/experiments/rl/README.md
tianyu-l, line 1: I somehow feel, deterministic_vllm_rl, as a folder name looks too long.
Do you think we can just call it `rl`?

PR #2129, opened 2025-12-09
[Autoparallel] Add local_map variant of DSv3 and 2D mesh AP

xmfan:
Stacked PRs:
 * __->__#2129

--- --- ---

[Autoparallel] Add local_map variant of DSv3 and 2D mesh AP

Currently, the AP experiment monkey patches Titan's main DSv3 implementation. But this is prone to breakage from both model definition changes in titan and from HOP/partitioner related changes in core. When these breaks happen, people are usually blocked until I find the root cause.

I'm going on PTO for the rest of the year, so I'm adding an integration to AP's DSv3 model in an attempt to make the development more stable for the upcoming PP integration.

Test: https://gist.github.com/xmfan/db15fda1e1bc1df7cd523005fe0baf33

### DIFF
--- a/torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
+++ b/torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
@@ -1,0 +1,182 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import time
+
+import torch
+from autoparallel.api import AutoParallel
+from autoparallel.auto_bucketing import configure_inductor_for_autobucketing
+
+from torch.distributed.tensor.placement_types import Shard
+from torchtitan.config import JobConfig
+from torchtitan.distributed import ParallelDims
+
+from torchtitan.tools.logging import logger
+
+
+# TODO: Autoparallel should transparently wrap the original nn.Module
+# but I don't know how to do that.
+def set_torchtitan_fields(orig, new):
+    assert isinstance(new.layers, torch.nn.ModuleDict)
+    for block in new.layers.values():
+        block.moe_enabled = hasattr(block, "moe")
+
+
+def parallelize_deepseekv3(
+    model,
+    parallel_dims: ParallelDims,
+    job_config: JobConfig,
+):
+    """
+    Apply Autoparallel to the model
+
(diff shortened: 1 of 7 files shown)


torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
```diff
@@ -0,0 +1,178 @@
...
+    world_mesh = parallel_dims.world_mesh
+
+    # Update me when changing dsv3.py
+    assert world_mesh.ndim == 2, "AP dsv3.py's local_map is specialized on 2 dims"
+
+    # Provide AP MoE with mesh
+    for layer in model.layers.values():
+        if layer.moe_enabled:
+            layer.moe.mesh = world_mesh
+            layer.moe.axis_name = "dp_shard_in_ep"
```
sanketpurandare, line 64: This seems okay for now, but when we enable TP, this should change so just add a comment that modify this when enabling TP.

torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
```diff
@@ -0,0 +1,178 @@
...
+        dynamic=True,
+    ) as autop:
+        autop.add_parameter_memory_constraint(low=None, high=None)
+
+        x_sharding = (Shard(0), Shard(0))
+        loss_parallel_enabled = (
+            parallel_dims.tp_enabled
+            and not job_config.parallelism.disable_loss_parallel
+        )
+        assert not loss_parallel_enabled
```
sanketpurandare, line 104: Instead of failing here why not disable loss parallel and give a warning?

sanketpurandare (approved):
Just some minor comments that can be addressed optionally

torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
xmfan: warnings are easy to miss, you won't know loss parallel is disable if you don't spot it before titan starts dumping its logs

torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
xmfan: I added an assert to check for dp2ep specifically, because i don't think we're handling the mesh setup well...

PR #2130, opened 2025-12-09
Fix `torch.compile` recompilation issue with HF modeling + TP

3outeille:
Fixing the bug https://github.com/huggingface/torchtitan/issues/6

TODO: need to apply change in `transformers V5`. That requires to wait for V5 to be a bit stable before switch torchtitan transformers modeling backend to v5 (as for now, it relies on 4.57.1)

# Issue

```
[rank3]:/fsx/ferdinandmom/ferdinand-hf/huggingface/torchtitan/env_torchtitan_official/lib/python3.12/site-packages/torch/_inductor/compile_fx.py:321: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
[rank3]:  warnings.warn(
[rank3]:[rank3]:W1209 10:08:01.582000 2539247 torch/_dynamo/convert_frame.py:1564] [0/8] torch._dynamo hit config.recompile_limit (8)
[rank3]:[rank3]:W1209 10:08:01.582000 2539247 torch/_dynamo/convert_frame.py:1564] [0/8]    function: 'forward' (/fsx/ferdinandmom/ferdinand-hf/huggingface/torchtitan/env_torchtitan_official/lib/python3.12/site-packages/torch/distributed/algorithms/_checkpoint/checkpoint_wrapper.py:145)
[rank3]:[rank3]:W1209 10:08:01.582000 2539247 torch/_dynamo/convert_frame.py:1564] [0/8]    last reason: 0/7: ___dict_contains(148, self._modules['_checkpoint_wrapped_module']._modules['self_attn']._forward_pre_hooks_with_kwargs)  # if hook_id in self._forward_pre_hooks_with_kwargs:  # nn/modules/module.py:1815 in inner
[rank3]:[rank3]:W1209 10:08:01.582000 2539247 torch/_dynamo/convert_frame.py:1564] [0/8] To log all recompilation reasons, use TORCH_LOGS="recompiles".
[rank3]:[rank3]:W1209 10:08:01.582000 2539247 torch/_dynamo/convert_frame.py:1564] [0/8] To diagnose recompilation issues, see https://pytorch.org/docs/main/compile/programming_model.recompilation.html
[rank3]:[rank3]: Traceback (most recent call last):
```

# Fix

- Apply + current PR changes + `transformers` at `modeling_llama.py`, change
```diff
       hidden_states, _ = self.self_attn(
-		   hidden_states=hidden_states,
+          hidden_states,
            attention_mask=attention_mask,
            position_ids=position_ids,
            past_key_values=past_key_values,
            use_cache=use_cache,
            cache_position=cache_position,
            position_embeddings=position_embeddings,
            **kwargs,
        )
```

- `./tooling_dev/debug_local.sh debugperf_large --compile`

image

# Explanation

- When `torch.compile` traces your model, it creates a compiled graph along with guards. Guards are conditions that must be true for that graph to be reused. If guard fails, `torch.compile` will recompiles.
- in `modeling_llama.py`, the `self.attn(hidden_states=hidden_states)` is called with `kwargs`
- In torchtitan, if you apply TP, it will apply `register_forward_pre_hook` . However, depending on if you use `kwargs` or not, it will call different function (cf https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/parallel/style.py#L576). 
	- In our case, it will call `module.register_forward_pre_hook(lambda _, inputs, kwargs: some_fn(inputs, kwargs), with_kwargs=True`
- but calling this function is problematic as it will trigger `if hook_id in self._forward_pre_hooks_with_kwargs:` (cf https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L1808)
	- This means that using `kwargs` will results in different `hook_id` , hence the error ` ___dict_contains(148, self._modules['_checkpoint_wrapped_module']._modules['self_attn']._forward_pre_hooks_with_kwargs)` 
- When we don't use`kwargs`,  `self._forward_pre_hooks_with_kwargs`  will always be empty (cf https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L1679C13-L1679C48) so the if check is not triggered, so each attention layer has same `hook_id`, thus no recompile

### DIFF
--- a/torchtitan/experiments/transformers_modeling_backend/infra/parallelize.py
+++ b/torchtitan/experiments/transformers_modeling_backend/infra/parallelize.py
@@ -190,8 +190,8 @@ def apply_non_moe_tp(
         layer_plan = {
             "input_layernorm": SequenceParallel(),
             "self_attn": prepare_module_input(
-                input_kwarg_layouts={"hidden_states": Shard(1)},
-                desired_input_kwarg_layouts={"hidden_states": Replicate()},
+                input_layouts=(Shard(1),),
+                desired_input_layouts=(Replicate(),),
             ),
             "post_attention_layernorm": SequenceParallel(),
         }


torchtitan/experiments/transformers_modeling_backend/__init__.py
```diff
@@ -23,6 +23,26 @@
 
 
 flavors = {
+    "debugperf": HFTransformerModelArgs(
```
wwwjn, line 26: Should we remove  these 2 test models?

tianyu-l, line 26: what's the difference between debugperf / debugperf_large and debugmodel? Can we just keep one of them?

torchtitan/experiments/transformers_modeling_backend/tooling_dev/check_checkpoint_correctness.py
tianyu-l, line 1: Do we need to ship this folder to fix the issue? It's about 2k LoC complexity.

tianyu-l (changes requested):

torchtitan/experiments/transformers_modeling_backend/tooling_dev/check_checkpoint_correctness.py
wwwjn: +1 , I think we could remove these test scripts to keep code simple

torchtitan/models/llama3/__init__.py
```diff
@@ -27,6 +27,17 @@
 
 
 llama3_args = {
+    "debugperf": TransformerModelArgs(
```
wwwjn, line 30: same here, could we remove these 2 models here?

torchtitan/experiments/transformers_modeling_backend/job_config.py
```diff
@@ -11,6 +11,8 @@
 class HFTransformers:
     model: str = ""
     """HuggingFace model ID (e.g., 'Qwen/Qwen3-4B-Instruct-2507')"""
+    tie_word_embeddings: bool = False
```
wwwjn, line 14: Putting `tie_word_embeddings` into job config is a little bit confusing, and seems not related to this error?

IIUC this is a field is decided by model architecture, and not decided by each training run. So previously we put  Qwen3's weight tying config into model_args: https://github.com/pytorch/torchtitan/blob/fbafd44da2baef0afac58989f07d799c4251bdef/torchtitan/models/qwen3/model/args.py#L43

3outeille:
It is a known issue: cf https://github.com/pytorch/pytorch/issues/170110. In this PR, we make changes so that we bypass the use of. `register_forward_hook(..., with_kwargs=True)` (which trigger too many recompilation with `torch.compile`)

PR #2131, opened 2025-12-09
[forge] allow torchforges to set checkpoint base folder

rakkit:
this PR 
1) allowing Torchforge to decide where to put the checkpoint and wandb, etc, instead of the  "current" folder
~~allowing Torchforge to decide to print / log the configs~~

### DIFF
--- a/torchtitan/experiments/forge/engine.py
+++ b/torchtitan/experiments/forge/engine.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import logging
 import os
 from typing import Generator
 
@@ -22,9 +21,6 @@ from torchtitan.tools import utils
 
 from .job_config import ForgeJobConfig
 from .train_spec import ForgeTrainSpec, get_train_spec
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
 
 
 class ForgeEngine(torch.distributed.checkpoint.stateful.Stateful):


torchtitan/experiments/forge/engine.py
```diff
@@ -86,6 +92,13 @@ def __init__(self, job_config: ForgeJobConfig):
             world_size=world_size,
         )
 
+        # Logging needs to happen after distributed initialized
+        if job_config.job.print_config:
+            dict_config = OmegaConf.to_container(job_config, resolve=True)
```
wwwjn: Can you help me understand what's the benefit of introducing extra dependency `OmegaConf`

rakkit: `torchforge` uses `yaml` & `omegaconf` to manage the configs. for some weird reason, when `job_config` is passed to the engine, it becomes an omegaconf object instead of a `dataclass.`

tianyu-l: Let's be careful introducing this dependency. We haven't but should set up some test in torchtitan for this engine, and we don't want to be forced to have this dependency.

If it's for logging purpose, it can be done in forge?

rakkit: Good point, for now what `forge` doing is 
```python
     class ForgeSFTRecipe(ForgeActor, ForgeEngine):
      ...
      ...
    def __init__(self, config: DictConfig):
        job_config = ForgeJobConfig().to_dict()
        # Hack to deal with literal types from titan
        job_config = OmegaConf.merge(job_config, config)
        super().__init__(job_config)
        
```

it make sense on the other side to convert `job_config` back to `dataclass`

rakkit:
sry for the confusion. I realized it's annoying to do this convert. I will move job config logging parts to Forge.

this pr is only about set up the checkpoint's `base_folder` now.

torchtitan/experiments/forge/engine.py
```diff
@@ -22,6 +23,9 @@
 from .job_config import ForgeJobConfig
 from .train_spec import ForgeTrainSpec, get_train_spec
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
tianyu-l: should remove?

rakkit: fixed.

tianyu-l (approved):
sgtm

PR #2132, opened 2025-12-09
PyTorch depends on psutil

fegin:
* __->__ #2132

TorchTitan should also depends on psutil.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
     "fsspec",
     "tyro",
     "tensorboard",
+    "psutil",
 ]
 dynamic = ["version"]
 


wwwjn (commented):
Thank you! I want to understand pytorch dependency more:
I see psutil listed in the dev dependency group, does it means it only used when we build pytorch dev and it's not a core/minital dependency to install pytorch? https://github.com/pytorch/pytorch/blob/7d4056596c0354083e7f66df596c8e8b5405803f/pyproject.toml#L46

And we would also need to list it in torchtitan pyproject.tom: https://github.com/pytorch/torchtitan/blob/refs/heads/main/pyproject.toml#L26 , as `psutil` will become torchtitan's core dependency?

fegin:
That's a good question. It's in requirements.txt but marked as dev in pyproject.toml. I don't know the reason and what's the relation there. In the ideally case, psutil should be installed when we install PyTorch. 

As for TorchTitan, since we use DCP and now DCP uses psutil, we have to include it as the core. Same as fsspec.

fegin:
Discussed offline

wwwjn (approved):

Issue #2124, opened 2025-12-08
DeepSeekV3 model component is not compilable when using Interleaved1F1B pipeline parallelism


syed-ahmed:
### Bug description

I'm seeing the following error for deepSeek-v3-671b:
```
  AssertionError: Attempt to trace forbidden callable <function mark_dynamic at 0x7f06cdd66d40>
  
  from user code:
     File "/workspace/torchtitan/torchtitan/models/llama4/infra/parallelize.py", line 592, in _run_experts_grouped_mm_dynamic
      torch._dynamo.mark_dynamic(x, 0)
  
  Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

Repro using debug model:
```
torchrun --nproc_per_node=8 --rdzv_backend c10d --rdzv_endpoint=localhost:0 --local-ranks-filter 0 --role rank --tee 3 -m torchtitan.train --job.config_file ./torchtitan/models/deepseek_v3/train_configs/debug_model.toml --model.flavor="debugmodel_flex_attn" \
--parallelism.expert_parallel_degree=4 \
--compile.enable \
--compile.components=model,loss \
--activation_checkpoint.mode=full \
--parallelism.pipeline_parallel_degree=2 \
--parallelism.data_parallel_shard_degree=4 \
--parallelism.pipeline_parallel_first_stage_less_layers=1 \
--parallelism.pipeline_parallel_last_stage_less_layers=1 \
--training.seq_len=4096 \
--training.local_batch_size=4 \
--parallelism.pipeline_parallel_schedule="Interleaved1F1B"
```

Workaround:
- Don't compile model: `--compile.components=loss`
- Revert https://github.com/pytorch/torchtitan/pull/2007
Likely because of the @forbin_in_graph: https://github.com/pytorch/pytorch/issues/167269

### Versions

```
Collecting environment information...
PyTorch version: 2.10.0a0+git8bb12cb
Is debug build: False
CUDA used to build PyTorch: 13.1
ROCM used to build PyTorch: N/A

OS: Ubuntu 24.04.3 LTS (x86_64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version: Could not collect
CMake version: version 3.31.6
Libc version: glibc-2.39

[...]

PR #2135, opened 2025-12-10
Fix apply_compile called multiple times in PP initialization

xmfan:
Stacked PRs:
 * __->__#2135

--- --- ---

PP initialization calls apply_compile multiple times, once per pp stage. But apply_compile does some global patching. So I add `already_patched` to avoid patching the same method multiple times.

If we patch multiple times, the second time will wrap `_run_experts_grouped_mm_dynamic` in a torch.compile(fullgraph=True) leading to the error in the issue below.

FIXES https://github.com/pytorch/torchtitan/issues/2124

### DIFF
--- a/tests/unit_tests/test_compile_moe.py
+++ b/tests/unit_tests/test_compile_moe.py
@@ -1,0 +1,81 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+
+import torch
+import torch.nn as nn
+
+from torchtitan.config.job_config import Compile as CompileConfig
+from torchtitan.models.llama4.infra.parallelize import apply_compile
+
+
+class TransformerBlock(nn.Module):
+    def __init__(self, dim=512):
+        super().__init__()
+        self.attention = nn.Linear(dim, dim, bias=False)
+        self.mlp = nn.Linear(dim, dim, bias=False)
+        self.moe_enabled = False
+
+    def forward(self, x):
+        x = self.attention(x)
+        x = self.mlp(x)
+        return x
+
+
+class TinyModel(nn.Module):
+    def __init__(self, num_layers=2, dim=512):
+        super().__init__()
+        self.layers = nn.ModuleDict(
+            {str(i): TransformerBlock(dim) for i in range(num_layers)}
+        )
+
+    def forward(self, x):
+        for layer in self.layers.values():
+            x = layer(x)
+        return x
+
+
+class TestApplyCompile(unittest.TestCase):
+    def test_patched_once(self):
+        """
+        Calls apply_compile multiple times, as in the case with PP.
+        But patches should only happen once
+        """
+        unused_model1 = TinyModel(num_layers=2, dim=128)
+        unused_model2 = TinyModel(num_layers=2, dim=128)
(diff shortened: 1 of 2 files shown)


torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -572,27 +572,34 @@ def apply_compile(model: nn.Module, compile_config: CompileConfig, ep_enabled: b
...
-    moe_module._run_experts_grouped_mm = torch.compile(
-        moe_module._run_experts_grouped_mm,
-        backend=compile_config.backend,
-        fullgraph=True,
+    # Patch some globals only once (apply_compile is called multiple times for PP setup)
+    already_patched = (
+        "_run_experts_grouped_mm_dynamic"
+        in moe_module._run_experts_grouped_mm.__qualname__
     )
+    if not already_patched:
```
tianyu-l, line 580: This sounds a temp workaround. Will there be a "permanent" solution?

xmfan: do you mean (1) the need to mark dynamic or (2) the need to define a global patched method?

(1) afaik marking dynamic is the permanent solution to avoid an initial recompile
(2) patching was chosen to avoid writing this into the model code. Two alternatives:
  - we could mark dynamic the outputs of token dispatch when ep is enabled
  - we could have a global parallelize function for pp to put code that can only run once

torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -572,27 +572,34 @@ def apply_compile(model: nn.Module, compile_config: CompileConfig, ep_enabled: b
...
+            # keep function logic in sync with `already_patched` above
+            def _run_experts_grouped_mm_dynamic(
+                w1: torch.Tensor,
+                w2: torch.Tensor,
+                w3: torch.Tensor,
+                x: torch.Tensor,
+                num_tokens_per_expert: torch.Tensor,
+            ) -> torch.Tensor:
+                # dynamic number of tokens in expert parallel
+                torch._dynamo.mark_dynamic(x, 0)
```
tianyu-l, line 599: maybe not relevant to this PR: how are you going to deal with dynamism in aot approach?

xmfan: depends on the finalized API, you could do explicit dynamic shapes annotations like here, and error in guards evaluation when unexpected dynamic shapes are encountered

xmfan:
@tianyu-l 
> I'm a bit confused why we were not hitting this error in https://github.com/pytorch/torchtitan/blob/main/tests/integration_tests/models.py#L96

When I run the CI command locally without this PR, I do hit the error. `CONFIG_FILE=./tests/integration_tests/base_config.toml NGPU=8 LOG_RANK=0,1,2,3,4,5,6,7 ./run_train.sh --model.name llama4 --parallelism.pipeline_parallel_degree 2 --parallelism.pipeline_parallel_schedule Interleaved1F1B --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 4 --parallelism.expert_tensor_parallel_degree 1 --compile.enable`.

The job also ran on the original PR: https://github.com/pytorch/torchtitan/actions/runs/19188831247/job/54860202625, so idk

tests/unit_tests/test_compile_moe.py
tianyu-l, line 1: please add comment for this file, and consider having a more descriptive name e.g. test_compile_moe.py

PR #2136, opened 2025-12-10
Enable static type checking with Pyrefly

rchen152:
Enables static type checking of torchtitan with [pyrefly](https://github.com/facebook/pyrefly). Type checking the code helps catch bugs earlier in the development cycle.

* Adds pyrefly to CI, as part of the linting workflow.
* Addresses ~100 type errors that can be fixed via local code changes and updates to type annotations, and silences the rest with `# pyrefly: ignore` suppression comments. Note that https://github.com/pytorch/torchtitan/commit/325efd946f1cbea85e503f9e684b8c879891fc1a contains all of the non-comment changes.

### DIFF
--- a/scripts/estimate/estimation.py
+++ b/scripts/estimate/estimation.py
@@ -128,18 +128,23 @@ def estimate_memory(job_config: JobConfig):
             lambda *args, **kwargs: model_converters.post_optimizer_hook(model)
         )
 
+        # pyrefly: ignore [missing-attribute]
         logger.info(f"Vocab size: {model_args.vocab_size}")
         # Create a dummy batch instead of loading from a dataset
         batch = (
             torch.randint(
                 0,
+                # pyrefly: ignore [missing-attribute]
                 model_args.vocab_size,
+                # pyrefly: ignore [missing-attribute]
                 (job_config.training.local_batch_size, model_args.max_seq_len),
                 device="cuda",
             ),
             torch.randint(
                 0,
+                # pyrefly: ignore [missing-attribute]
                 model_args.vocab_size,
+                # pyrefly: ignore [missing-attribute]
                 (job_config.training.local_batch_size, model_args.max_seq_len),
                 device="cuda",
             ),
--- a/torchtitan/protocols/model.py
+++ b/torchtitan/protocols/model.py
@@ -6,7 +6,7 @@
 
 from abc import abstractmethod
 from dataclasses import dataclass
-from typing import Protocol, TYPE_CHECKING
+from typing import Protocol
 
 import torch
 import torch.nn as nn
@@ -17,10 +17,6 @@ from torchtitan.components.tokenizer import BaseTokenizer
 
 from torchtitan.config import JobConfig
 from torchtitan.models.attention import VarlenMetadata
-
-if TYPE_CHECKING:
-    from torchtitan.models.flux.model.autoencoder import AutoEncoderParams
-    from torchtitan.models.moe.moe import MoEArgs
 
 
 AttentionMasksType = dict[str, BlockMask] | BlockMask | VarlenMetadata
@@ -35,17 +31,6 @@ class BaseModelArgs:
     """
 
     _enforced: str = "This field is used to enforce all fields have defaults."
-
-    if TYPE_CHECKING:
-        # Declare fields that exist on BaseModelArgs subclasses to help the type checker.
-        # These aren't added at runtime.
-        autoencoder_params: AutoEncoderParams = AutoEncoderParams()
-        enable_weight_tying: bool = False
-        max_seq_len: int = 0
-        moe_args: MoEArgs = MoEArgs()
-        n_heads: int = 0
-        n_layers: int = 0
-        vocab_size: int = 0
 
     @abstractmethod
     def update_from_config(self, job_config: JobConfig, **kwargs) -> None:
--- a/torchtitan/models/deepseek_v3/model/state_dict_adapter.py
+++ b/torchtitan/models/deepseek_v3/model/state_dict_adapter.py
@@ -130,9 +130,11 @@ class DeepSeekV3StateDictAdapter(MoEStateDictAdapter):
                     # keep this path for offline conversion
                     split_values = self._split_experts_weights(
                         value,
+                        # pyrefly: ignore [missing-attribute]
                         self.model_args.moe_args.num_experts,
                     )
 
+                    # pyrefly: ignore [missing-attribute]
                     for expert_num in range(0, self.model_args.moe_args.num_experts):
                         new_key = new_abstract_key.format(layer_num, expert_num)
                         hf_state_dict[new_key] = split_values[expert_num].squeeze()
@@ -189,6 +191,7 @@ class DeepSeekV3StateDictAdapter(MoEStateDictAdapter):
                         expert_weights_by_layer,
                         titan_abstract_key,
                         layer_num,
+                        # pyrefly: ignore [missing-attribute]
                         self.model_args.moe_args.num_experts,
                     )
 
--- a/torchtitan/models/flux/train.py
+++ b/torchtitan/models/flux/train.py
@@ -50,6 +50,7 @@ class FluxTrainer(Trainer):
         self.autoencoder = load_ae(
             # pyrefly: ignore [missing-attribute]
             job_config.encoder.autoencoder_path,
+            # pyrefly: ignore [missing-attribute]
             model_args.autoencoder_params,
             device=self.device,
             dtype=self._dtype,
--- a/torchtitan/models/qwen3/model/state_dict_adapter.py
+++ b/torchtitan/models/qwen3/model/state_dict_adapter.py
@@ -87,9 +87,11 @@ class Qwen3StateDictAdapter(MoEStateDictAdapter):
                     # keep this path for offline conversion
                     split_values = self._split_experts_weights(
                         value,
+                        # pyrefly: ignore [missing-attribute]
                         self.model_args.moe_args.num_experts,
                     )
 
+                    # pyrefly: ignore [missing-attribute]
                     for expert_num in range(self.model_args.moe_args.num_experts):
                         new_key = new_abstract_key.format(layer_num, expert_num)
                         hf_state_dict[new_key] = split_values[expert_num].squeeze()
@@ -107,6 +109,7 @@ class Qwen3StateDictAdapter(MoEStateDictAdapter):
             else:
                 if key not in to_hf_map:
                     continue
+                # pyrefly: ignore [missing-attribute]
                 if self.model_args.enable_weight_tying and key == "output.weight":
                     continue
                 new_key = to_hf_map[key]
@@ -124,6 +127,7 @@ class Qwen3StateDictAdapter(MoEStateDictAdapter):
         expert_weights_by_layer = {}  # {layer: {abstract_key: {expert_id: tensor}}}
 
         if (
+            # pyrefly: ignore [missing-attribute]
             self.model_args.enable_weight_tying
             and "lm_head.weight" not in hf_state_dict
         ):
@@ -159,6 +163,7 @@ class Qwen3StateDictAdapter(MoEStateDictAdapter):
                         expert_weights_by_layer,
                         titan_abstract_key,
                         layer_num,
+                        # pyrefly: ignore [missing-attribute]
                         self.model_args.moe_args.num_experts,
                     )
 
--- a/torchtitan/models/utils.py
+++ b/torchtitan/models/utils.py
@@ -400,6 +400,7 @@ def get_dense_model_nparams_and_flops(
     # 4. we follow the convention and do not account for sparsity in causal attention
     num_flops_per_token = (
         6 * (nparams - nparams_embedding)
+        # pyrefly: ignore [missing-attribute]
         + 6 * model_args.n_layers * model_args.n_heads * head_dims * seq_len
     )
 
@@ -455,6 +456,7 @@ def get_moe_model_nparams_and_flops(
     nparams_sparse_active = (
         nparams_moe_router
         + nparams_shared_experts
+        # pyrefly: ignore [missing-attribute]
         + nparams_experts * model_args.moe_args.top_k // model_args.moe_args.num_experts
     )
 
@@ -465,6 +467,7 @@ def get_moe_model_nparams_and_flops(
 
     num_flops_per_token = (
         6 * (nparams_dense - nparams_embedding + nparams_sparse_active)
+        # pyrefly: ignore [missing-attribute]
         + 6 * model_args.n_layers * model_args.n_heads * head_dims * seq_len
     )
 


.github/scripts/setup_pyrefly.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Run this script from the repository root.
+
+# Accept optional flags to pass to uv pip install commands
+# Usage: .github/scripts/setup_pyrefly.sh [flags]
+UV_INSTALL_FLAGS="$@"
+
+pip install uv
```
tianyu-l: Is `uv` a must? Currently torchtitan doesn't have uv scripts elsewhere.

.github/scripts/setup_pyrefly.sh
```diff
@@ -0,0 +1,22 @@
...
+uv pip install \
+  torchft-nightly \
+  -r .ci/docker/requirements.txt \
+  -r .ci/docker/requirements-dev.txt \
+  -r .ci/docker/requirements-flux.txt \
+  -r .ci/docker/requirements-transformers-modeling-backend.txt \
+  -r .ci/docker/requirements-vlm.txt \
```
tianyu-l: We probably shouldn't mix these repo dependency install with type checking install. We should maybe put pyrefly install into `.requirements-dev.txt`

.github/workflows/lint.yaml
```diff
@@ -35,3 +35,7 @@ jobs:
         uses: tj-actions/changed-files@d6e91a2266cdb9d62096cebf1e8546899c6aa18f # v45.0.6
       - name: Lint modified files
         run: pre-commit run --files ${{ steps.changed-files.outputs.all_changed_files }}
+      - name: Install type checking dependencies
+        run: .github/scripts/setup_pyrefly.sh --system
```
tianyu-l: similarly, we shouldn't mix pyrefly install with others.

CONTRIBUTING.md
```diff
@@ -15,7 +15,12 @@ We actively welcome your pull requests.
 3. If you've changed APIs, update the documentation.
 4. Ensure the test suite passes.
 5. Make sure your code lints (`pre-commit run --all-files`).
-6. If you haven't already, complete the Contributor License Agreement ("CLA").
+6. Make sure your code type checks:
+   ```
+   .github/scripts/setup_pyrefly.sh  # one-time setup
```
tianyu-l: similarly, should follow `pre-commit` style

scripts/checkpoint_conversion/convert_from_hf.py
```diff
@@ -17,15 +17,18 @@
 @torch.inference_mode()
 def convert_from_hf(input_dir, output_dir, model_name, model_flavor):
     if model_name == "flux":
+        # pyrefly: ignore [missing-import]
         import torchtitan.experiments.flux  # noqa: F401
```
fegin: This import path no longer exist. So this is a bug. Can you try "torchtitan.models.flux"? If it doesn't work, we can leave a comment here. cc., @wwwjn

scripts/checkpoint_conversion/convert_to_hf.py
```diff
@@ -30,8 +30,10 @@ def convert_to_hf(
 
     with torch.device("cpu"):
         model = train_spec.model_cls(model_args)
+    # pyrefly: ignore [bad-argument-type]
```
fegin, line 33: TODO for me: Change ModelProtocol to include nn.Module as the base class.

scripts/checkpoint_conversion/numerical_tests_example.py
```diff
@@ -25,7 +25,7 @@ def loss_fn(logits1, logits2):
     probs2 = F.softmax(logits2, dim=-1)
 
     # Calculate KL Divergence
-    kl_loss = F.kl_div(probs1, probs2, "mean")
+    kl_loss = F.kl_div(probs1, probs2, reduction="mean")
```
fegin, line 28: This is interesting, so what's the error pyrefly reports without this change?

scripts/estimate/estimation.py
```diff
@@ -98,25 +98,33 @@ def estimate_memory(job_config: JobConfig):
...
         model.train()
 
         # build optimizer after applying parallelisms to the model
+        # pyrefly: ignore [bad-argument-type]
         optimizers = build_optimizers([model], job_config.optimizer, parallel_dims)
         lr_schedulers = build_lr_schedulers(
-            optimizers.optimizers, job_config.lr_scheduler, job_config.training.steps
+            # pyrefly: ignore [bad-argument-type]
+            optimizers.optimizers,
+            job_config.lr_scheduler,
```
fegin, line 120: TODO for me: `build_lr_schedulers` has an incorrect typing.

torchtitan/components/ft/manager.py
```diff
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import importlib
+import importlib.util
```
fegin, line 7: Should we remove `import importlib` above this line?

torchtitan/components/checkpoint.py
```diff
@@ -394,13 +403,15 @@ def dcp_save(
...
                 checkpoint_id=checkpoint_save_id,
+                # pyrefly: ignore [bad-argument-type]
                 process_group=self.pg,
             )
         elif async_mode == AsyncMode.ASYNC_WITH_PINNED_MEM:
             ret = dcp.async_save(
                 state_dict,
                 storage_writer=storage_writer,
                 checkpoint_id=checkpoint_save_id,
+                # pyrefly: ignore [bad-argument-type]
```
fegin, line 415: TODO for me: why self.pg has incorrect typing or why async_save doesn't accept the PG typing.

torchtitan/components/dataloader.py
```diff
@@ -41,6 +41,7 @@ def __iter__(self):
         ...
 
 
+# pyrefly: ignore [inconsistent-inheritance]
```
fegin, line 44: TODO for me: this seems to be an inconsistency issue between checkpoint Stateful and dataloader StatefulDataLoader, which we may not be able to change.

torchtitan/components/optimizer.py
```diff
@@ -8,6 +8,7 @@
 from typing import Any, Generic, Iterator, TypeVar
 
 import torch
+import torch.distributed.tensor
```
fegin, line 11: TODO for me: change the reference of `orch.distributed.tensor.DTensor` to `DTensor`.

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -198,7 +198,9 @@ def build_pipeline_schedule(
             f"of stages ({num_total_stages}) which may result in a bubble in the pipeline."
         )
 
+    # pyrefly: ignore [bad-instantiation]
     schedule = schedule_class(
+        # pyrefly: ignore [bad-argument-type]
```
fegin, line 205: typing issue of `get_schedule_class`, cc., @wconstab @H-Huang

torchtitan/distributed/utils.py
```diff
@@ -205,11 +214,11 @@ def create_context_parallel_ctx(
...
         from torch.distributed.tensor.experimental import context_parallel
         from torch.distributed.tensor.experimental._attention import set_rotate_method
-    except ImportError:
-        print(
+    except ImportError as e:
+        raise ValueError(
             f"PyTorch version {torch.__version__} does not include the experimental "
             "Context Parallel API. Please update to a newer version."
-        )
+        ) from e
```
fegin, line 222: TODO for me: `set_rotate_method` has been in the main PyTorch for a while. We should be able to remove this try/except.

fegin (commented):
This is very helpful. This pyrefly PR already catches at least one bug. We should definitely enforce stricter type checking.

scripts/checkpoint_conversion/convert_from_hf.py
wwwjn: Haha it caches my bug lol, I will fix this import path , thank you!

.github/scripts/setup_pyrefly.sh
rchen152: It's not necessary; I've removed the `uv` dependency.

.github/scripts/setup_pyrefly.sh
rchen152: Done. FYI the reason I originally made this a stand-alone script was so that I could install nightly torch and torchao from https://download.pytorch.org/whl/nightly/cu128 in a separate step. With all the dependencies in `requirements-dev.txt`, we're now installing those deps from pypi, which changes the type checking results a bit. I had to add some new suppressions.

.ci/docker/requirements-dev.txt
```diff
@@ -1,6 +1,16 @@
+datasets >= 3.6.0
+einops
 expecttest==0.1.6
+pillow
 pytest==7.3.2
 pytest-cov
 pre-commit
+pyrefly==0.45.1
 tomli-w >= 1.1.0
+torch
```
wwwjn: We can not install pytorch using pip install, as torchtitan depends on pytorch-nightly: https://github.com/pytorch/torchtitan/blob/refs/heads/main/README.md#nightly-builds

scripts/checkpoint_conversion/convert_from_hf.py
```diff
@@ -17,15 +17,18 @@
...
     # initialize model to allocate memory for state dict
     train_spec = train_spec_module.get_train_spec(model_name)
     model_args = train_spec.model_args[model_flavor]
 
     with torch.device("cpu"):
         model = train_spec.model_cls(model_args)
+    # pyrefly: ignore [bad-argument-type]
     model = ModelWrapper(model)
 
+    # pyrefly: ignore [not-callable]
```
wwwjn, line 28: > pyrefly: ignore [bad-argument-type]

I feel like we need to fix/refactor the code if it doesn't pass pyrefly check, instead of leaving a lot of comments to suppress the error

CONTRIBUTING.md
rchen152: Done, pyrefly is now run as a `pre-commit` hook.

Just FYI, the pyrefly hook works a little differently from the others - you have to install all the dependencies yourself (including pyrefly) before running it. This is because a type checker needs to be able to see a project's dependencies, so we use `language: system` to use the local environment rather than an isolated environment managed by pre-commit.

scripts/checkpoint_conversion/numerical_tests_example.py
rchen152: This is the error:
```
ERROR Argument `Literal['mean']` is not assignable to parameter `size_average` with type `bool | None` in function `torch.nn.functional.kl_div` [bad-argument-type]
  --> scripts/checkpoint_conversion/numerical_tests_example.py:28:40
   |
28 |     kl_loss = F.kl_div(probs1, probs2, "mean")
   |                                        ^^^^^^
   |
```

.ci/docker/requirements-dev.txt
rchen152: @wwwjn I originally had a separate `setup_pyrefly.sh` script that installed nightly torch and torchao, which I removed based on @tianyu-l's comment that it would be preferable to have all of the dependencies pyrefly needs in `requirements-dev.txt`. Do you have any suggestions for how to resolve this?

wwwjn: Thank you! I think other dependencies are fine, but for torch and torchao, can we move these 2 package to lint.yaml, and use commands like `pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall` to install them separately?

scripts/checkpoint_conversion/convert_from_hf.py
rchen152: @wwwjn Adding ignores to get the code to a clean state and then incrementally fixing them is the standard way to enable a type checker on a large code base. I already made fixes to remove 100+ `pyrefly: ignore`s, but the remaining issues look fairly tricky to resolve and IMO would be better tackled as follow-ups rather than trying to combine larger and more risky refactors with enabling the type checker.

.ci/docker/requirements-dev.txt
rchen152: Done, thanks!

scripts/checkpoint_conversion/convert_from_hf.py
rchen152: Changing the import worked, thanks!

scripts/checkpoint_conversion/convert_to_hf.py
rchen152: @fegin Would you like me to put these TODOs in the code?

rchen152:
Thanks for all the comments! I believe this is ready for another round of review.

.github/scripts/setup_pyrefly.sh
rchen152: I ended up moving torch and torchao out again to a separate install step - all the other deps are in requirements-dev.txt

.ci/docker/requirements-dev.txt
tianyu-l, line 1: we shouldn't include those already in requirements.txt in this file

.ci/docker/requirements-dev.txt
```diff
@@ -1,6 +1,14 @@
...
+einops
 expecttest==0.1.6
+pillow
 pytest==7.3.2
 pytest-cov
 pre-commit
+pyrefly==0.45.1
 tomli-w >= 1.1.0
+torchdata >= 0.8.0
+torchft-nightly
```
tianyu-l: for now we can do similar things to torchao
@fegin we should move `torchtft` to experiments

.github/workflows/lint.yaml
```diff
@@ -28,7 +28,8 @@ jobs:
         run: python -m pip install --upgrade pip
       - name: Install lint utilities
         run: |
-          python -m pip install pre-commit
+          python -m pip install -r requirements-dev.txt
+          python -m pip install --force-reinstall --pre --index-url https://download.pytorch.org/whl/nightly/cu126 torch torchao
```
tianyu-l: Can user not run pyrefly / pre-commit without installing the optional torchao package?

CONTRIBUTING.md
```diff
@@ -5,6 +5,7 @@ possible. Contributions should follow the [Contributing Guidelines](#contributin
 ### Setup
 ```
 pip install -r requirements-dev.txt
+pip install --force-reinstall --pre --index-url https://download.pytorch.org/whl/nightly/cu126 torch torchao
```
tianyu-l: This used to be optional.

scripts/checkpoint_conversion/convert_from_hf.py
```diff
@@ -17,15 +17,17 @@
 @torch.inference_mode()
 def convert_from_hf(input_dir, output_dir, model_name, model_flavor):
     if model_name == "flux":
-        import torchtitan.experiments.flux  # noqa: F401
+        import torchtitan.models.flux  # noqa: F401
```
tianyu-l: we no longer need this if-import, I believe

.github/workflows/lint.yaml
rchen152: I updated the pyrefly config so that torchao is no longer required - pyrefly now uses `ignore-missing-imports` for torchao, so it'll replace the import with `Any` rather than erroring when it's not present.

.ci/docker/requirements-dev.txt
fegin: Well, I think TorchFT people should start to do that as we discussed before, cc., @tushar00jain

scripts/checkpoint_conversion/convert_from_hf.py
fegin: @wwwjn It's actually our responsibility to fix the typing as we understand the code. We should have follow-up BE PRs to address these ignore.

scripts/checkpoint_conversion/convert_to_hf.py
fegin: No need, we should go through these ignore later one by one anyway, thanks!

scripts/checkpoint_conversion/convert_from_hf.py
wwwjn: Agreed, let's do some BE work later

.ci/docker/requirements-dev.txt
```diff
@@ -1,6 +1,9 @@
+einops
```
tianyu-l: @wwwjn is this and pillow for FLUX? maybe we should put them in core requirements.txt

CONTRIBUTING.md
```diff
@@ -4,7 +4,8 @@ possible. Contributions should follow the [Contributing Guidelines](#contributin
 
 ### Setup
 ```
-pip install -r requirements-dev.txt
+pip install -r requirements.txt -r requirements-dev.txt
+pip install --force-reinstall --pre --index-url https://download.pytorch.org/whl/nightly/cu126 torch
```
tianyu-l: Let's remove this for two reasons:
1. it's not the only way to install torch
2. it's already covered by install tips for using torchtitan in README

We can add a link to that section if you feel necessary

pyproject.toml
```diff
@@ -62,3 +62,7 @@ include = ["torchtitan*"]
 [tool.pytest.ini_options]
 addopts = ["--showlocals"]  # show local variables in tracebacks
 testpaths = ["tests"]
+
+[tool.pyrefly]
```
tianyu-l, line 68: can I ask why we'd put this in pyproject?

torchtitan/protocols/model.py
```diff
@@ -18,6 +18,10 @@
 from torchtitan.config import JobConfig
 from torchtitan.models.attention import VarlenMetadata
 
+if TYPE_CHECKING:
+    from torchtitan.models.flux.model.autoencoder import AutoEncoderParams
+    from torchtitan.models.moe.moe import MoEArgs
```
tianyu-l: doing this doesn't sound clean and can easily cause import cycles?

Can we ignore for now? Or let's define things like `BaseMoEModelArgs` to inherit `BaseModelArgs` if needed

fegin: We should use # pyrefly: ignore for this file now. Fixing the typing for this file worths a PR.

.ci/docker/requirements-dev.txt
fegin: @tianyu-l These two are used by flux and it is annoying they are not in the core requirements.txt when I verified my PRs. So let's move them to the core.

@rchen152 Can you move `einops` and `pillow` to the core requirements.txt since you are adding them?

tianyu-l: also pyproject.toml dependency list

wwwjn: > @wwwjn is this and pillow for FLUX? maybe we should put them in core requirements.txt

Yes these are for FLUX, FLUX will always need a separate requirements.txt because of `transformers`

pyproject.toml
rchen152: Pyrefly configuration can go either in pyproject.toml or in a separate pyrefly.toml file. There isn't a particular advantage to one approach over the other, so I appended to pyproject.toml because I saw it already had sections for other tools. Happy to break this out into a separate pyrefly.toml file if you like.

tianyu-l: I don't know much about pyproject.toml. I had thought it's more related to packaging, e.g. for people who do `pip install torchtitan`.

Would CI / local development without pip install torchtitan respect the pyrefly config here?

rchen152: Yep, as long as pyrefly is run inside the torchtitan repo, it'll find the pyrefly config in pyproject.toml.

tianyu-l (approved):
lgtm

PR #2138, opened 2025-12-10
Add repeated_subgraphs option in AutoParallel example

fmassa:
Using `repeated_subgraphs` enable AutoParallel to enforce the same solution to repeated regions of the model, making the solver be much faster. For 32 layers, we obtained roughtly 21x speedup compared to without repeated subgraphs.


xmfan (approved):

xmfan:
`psutil` issue should be fixed with next nightly now that https://github.com/pytorch/pytorch/pull/169985 landed 3h ago

Issue #2140, opened 2025-12-11
Validation breaking with FlashAttention


francesco-bertolotti:
### Bug description

Enabling validation with flash-attention enabled leads to a dynamo error:

```
    File "torchtitan/components/validate.py", line 158, in validate
      predictions = model_parts[0](inputs)
    File "torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
      return self._call_impl(*args, **kwargs)
             ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
    File ".venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1882, in _call_impl
      return inner()
    File "venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1830, in inner
      result = forward_call(*args, **kwargs)
    File "torchtitan/models/qwen3/model/model.py", line 536, in forward
      h = layer(h, self.rope_cache, attention_masks)
    File "venv/lib/python3.13/site-packages/torch/_dynamo/eval_frame.py", line 441, in __call__
      return super().__call__(*args, **kwargs)
             ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
    File ".venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
      return self._call_impl(*args, **kwargs)
             ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
    File ".venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1882, in _call_impl
      return inner()
    File ".venv/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1830, in inner
      result = forward_call(*args, **kwargs)
    File ".venv/lib/python3.13/site-packages/torch/_dynamo/eval_frame.py", line 939, in compile_wrapper
      raise e.with_traceback(None) from e.__cause__  # User compiler error
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  torch._dynamo.exc.Unsupported: Observed exception
    Explanation: Dynamo found no exception handler at the top-level compiled function when encountering an exception. Exception will propagate outside the compiled region.
    Hint: Dynamo has detected that tracing the code will result in an error when running in eager. Please double check that your code doesn't contain a similar error when actually running eager/uncompiled.
[...]

PR #2142, opened 2025-12-11
fixed validation error when using flash attention

francesco-bertolotti:
This PR addresses the issue https://github.com/pytorch/torchtitan/issues/2140

Briefly, the bug is related to the validation step.
The validate method did not pass the attention mask to the model.
When using flash attention, this leads to an error.

### DIFF
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -4,10 +4,11 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Generator, Any
+from typing import Any, Generator
 
 import torch
 import torch.nn as nn
+import torchtitan.protocols.train_spec as train_spec_module
 from torch.distributed.pipelining.schedules import _PipelineSchedule
 from torchtitan.components.dataloader import BaseDataLoader
 from torchtitan.components.loss import LossFunction
@@ -18,7 +19,6 @@ from torchtitan.distributed import ParallelDims, utils as dist_utils
 from torchtitan.hf_datasets.text_datasets import build_text_validation_dataloader
 from torchtitan.tools import utils
 from torchtitan.tools.logging import logger
-import torchtitan.protocols.train_spec as train_spec_module
 
 
 class BaseValidator:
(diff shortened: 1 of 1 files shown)


torchtitan/components/validate.py
```diff
@@ -105,6 +113,13 @@ def validate(
             ):
                 break
 
+            # prepare attention mask
+            extra_kwargs : dict[str, Any] = {"attention_masks" : model_parts[0].get_attention_masks(
```
wwwjn: For readability, can you make this into a function similar as: https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/train.py#L416C9-L416C33

francesco-bertolotti:
The new commit propagated the fix to the pipeline parallelism case. Further, I have encapsulated the creation of the attention mask in a method.

torchtitan/components/validate.py
```diff
@@ -76,12 +78,43 @@ def __init__(
...
+        model_args.update_from_config(job_config)
+        self.model_args = model_args
+
         if self.job_config.validation.steps == -1:
             logger.warning(
                 "Setting validation steps to -1 might cause hangs because of "
                 "unequal sample counts across ranks when dataset is exhausted."
             )
 
+    def post_dataloading_process(
```
tianyu-l, line 92: can we reuse the one in train.py?

tianyu-l (changes requested):

torchtitan/components/validate.py
francesco-bertolotti: In principle, it would be possible, but I would need to make some modifications since their signatures differ. In particular, the trainer accesses `model_parts` through `self`.

One possible solution is to turn `post_dataloading_process` into a utility function that both the trainer and validator can call. However, this would mean you can no longer modify the behavior through inheritance—unless the current `post_dataloading_process` simply becomes a wrapper around the new utility function.|

If you have any suggestions, I can add a commit.

tianyu-l: I see.
If you want to "modify the behavior", we probably should have a `PostDataloadingProcessor` class.
If not, I think it's fine to call a util directly in both trainer and validator.

WDYT?
cc @fegin

fegin: I'm thinking this as well, I'm seeing the duplicated code in trainer and validator, `model_parts` is not a big problem, you can add it. See https://github.com/pytorch/torchtitan/pull/2144. But the duplicated logic is pretty concerning.

My initial proposal is to have an util function in `distributed.utils` so that both methods can call it.  The only worry is that I don't know whether `distributed.utils` is the best place to put it. It is not now because the util function purely unwrap the input but the util function will encapsulate the CP logic after https://github.com/pytorch/torchtitan/pull/2144. So `distributed.utils` seems to be a good place.


@francesco-bertolotti  you can do it as your PR may be landed first. I can rebase on top of your change.

tianyu-l: agreed we should share code, but not sure about `distributed.utils`. Let's discuss offline.

tianyu-l (approved):
unblocking

tianyu-l:
@francesco-bertolotti CPU unit test failed, could you take a look?

francesco-bertolotti:
Hi @tianyu-l,

I looked into the error, and it turns out to be a circular import issue.

In `validate.py`, I need access to `model_args`. Since `Validator` doesn’t have direct access to it, I currently construct it via `get_train_spec`. However, importing `get_train_spec` triggers an import of `BaseValidator`, which lives in the same module as `Validator`, leading to the circular dependency.

I see two possible ways to resolve this:

1. Pass `model_args` directly to `Validator`, eliminating the need to import `get_train_spec`.
2. Move `BaseValidator` into its own module.

I’m leaning slightly toward changing the `Validator` signature, as it feels cleaner overall, though it would be a breaking change.

Curious to hear your thoughts.

tianyu-l:
It seems this is being fixed in https://github.com/pytorch/torchtitan/pull/2144, with the first approach you mentioned.

Are you OK if we consolidate effort there?

francesco-bertolotti:
Hi @tianyu-l , 

Since I do not think this PR is going to get merge can I close it?

PR #2144, opened 2025-12-11
[CP] Refactor Context Parallel to use new PyTorch CP APIs

fegin:
* #2145
* __->__ #2144

**Summary**

1. Refactored CP Dispatching:
  - New apply_cp() function uses PyTorch's _ContextParallel parallelization plan to dispatch attention call.
  - Enables CP dispatcher for SDPA attention type inside apply_cp()
2. New CP Data Sharding Approach:
  - Added a cp_shard() helper function that wraps PyTorch's _context_parallel_shard API
  - Uses _HeadTailLoadBalancer for SDPA attention load balancing
  - FlexAttention CP support deferred to a future PR
  - CP sharding now happens explicitly in post_dataloading_process() where inputs, labels, and positions are sharded
  - The new positions argument allows us to not shard the freqs_cis.

Note that this PR require https://github.com/pytorch/pytorch/pull/170200

**Test**
```
-> % python3 scripts/loss_compare.py . chienchin/loss_compare --baseline-options="--parallelism.context_parallel_degree=8" --test-options="--parallelism.context_parallel_degree=8" --steps=100 --assert-equal

pick 5903566a Improve the loss_compare.sh logic

[LOSS_COMPARE]
[LOSS_COMPARE] Asserting losses are equal...
[LOSS_COMPARE] Baseline log: /tmp/baseline_training.log
[LOSS_COMPARE] Test log: /tmp/test_training.log
[LOSS_COMPARE] Extracted 100 steps from baseline log
[LOSS_COMPARE] Extracted 100 steps from test log
test_losses_equal
(__main__.assert_losses_equal.<locals>.LossEqualityTest.test_losses_equal)
... ok

----------------------------------------------------------------------
Ran 1 test in 0.000s

OK
[LOSS_COMPARE] All losses are equal. Assertion passed!
[LOSS_COMPARE] ==========================================
[LOSS_COMPARE] LOSS COMPARISON ANALYSIS
[LOSS_COMPARE] ==========================================

[LOSS_COMPARE] Step-by-step loss comparison:
[LOSS_COMPARE] Step    Baseline Loss    Test Loss   Difference
[LOSS_COMPARE] ----    -------------    ---------   ----------
[LOSS_COMPARE] 1       8.1309           8.1309           0.000000
[LOSS_COMPARE] 2       7.8268           7.8268           0.000000
[LOSS_COMPARE] 3       7.2284           7.2284           0.000000
[LOSS_COMPARE] 4       6.4669           6.4669           0.000000
[LOSS_COMPARE] 5       5.4017           5.4017           0.000000
[LOSS_COMPARE] 6       4.7656           4.7656           0.000000
[LOSS_COMPARE] 7       4.3587           4.3587           0.000000
[LOSS_COMPARE] 8       4.0938           4.0938           0.000000
[LOSS_COMPARE] 9       4.4019           4.4019           0.000000
[LOSS_COMPARE] 10      3.7451           3.7451           0.000000
....
[LOSS_COMPARE] 90      2.802            2.802            0.000000
[LOSS_COMPARE] 91      2.7207           2.7207           0.000000
[LOSS_COMPARE] 92      2.7454           2.7454           0.000000
[LOSS_COMPARE] 93      2.6992           2.6992           0.000000
[LOSS_COMPARE] 94      2.743            2.743            0.000000
[LOSS_COMPARE] 95      2.7534           2.7534           0.000000
[LOSS_COMPARE] 96      2.8403           2.8403           0.000000
[LOSS_COMPARE] 97      2.783            2.783            0.000000
[LOSS_COMPARE] 98      3.0892           3.0892           0.000000
[LOSS_COMPARE] 99      2.7905           2.7905           0.000000
[LOSS_COMPARE] 100     2.733            2.733            0.000000
[LOSS_COMPARE]
[LOSS_COMPARE] Summary statistics:
[LOSS_COMPARE] Average baseline loss:  3.1414940000000002
[LOSS_COMPARE] Average test loss: 3.1414940000000002
[LOSS_COMPARE] Average difference:     0.000000
[LOSS_COMPARE]
[LOSS_COMPARE] Loss comparison complete. No results saved (no output
folder specified).
```

**TODO**
  - This PR will invalidate torch.compile + CP due to https://github.com/pytorch/pytorch/issues/170110. We will have to wait for Dynamo to fix the issue or refactor nn.Module core logic to avoid check hook_id.

### DIFF
--- a/torchtitan/models/llama3/infra/parallelize.py
+++ b/torchtitan/models/llama3/infra/parallelize.py
@@ -243,7 +243,6 @@ def apply_tp(
             # pyrefly: ignore [bad-argument-type]
             module=transformer_block,
             device_mesh=tp_mesh,
-            # pyrefly: ignore [bad-argument-type]
             parallelize_plan=layer_plan,
         )
 
--- a/torchtitan/models/llama4/infra/parallelize.py
+++ b/torchtitan/models/llama4/infra/parallelize.py
@@ -309,7 +309,6 @@ def apply_non_moe_tp(
             # pyrefly: ignore [bad-argument-type]
             module=transformer_block,
             device_mesh=tp_mesh,
-            # pyrefly: ignore [bad-argument-type]
             parallelize_plan=layer_plan,
         )
 
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -149,7 +149,6 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
 
         # Build the collection of model converters. No-op if `model.converters` empty
         model_converters = build_model_converters(job_config, parallel_dims)
-        # pyrefly: ignore [bad-argument-type]
         model_converters.convert(model)
 
         # metrics logging
@@ -167,7 +166,6 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
         (
             model_param_count,
             self.metrics_processor.num_flops_per_token,
-            # pyrefly: ignore [bad-argument-type]
         ) = model_args.get_nparams_and_flops(model, job_config.training.seq_len)
 
         logger.info(
--- a/scripts/checkpoint_conversion/convert_from_hf.py
+++ b/scripts/checkpoint_conversion/convert_from_hf.py
@@ -23,7 +23,6 @@ def convert_from_hf(input_dir, output_dir, model_name, model_flavor):
     with torch.device("cpu"):
         # pyrefly: ignore[bad-instantiation]
         model = train_spec.model_cls(model_args)
-    # pyrefly: ignore [bad-argument-type]
     model = ModelWrapper(model)
 
     # pyrefly: ignore[bad-instantiation, not-callable]
--- a/scripts/checkpoint_conversion/convert_to_hf.py
+++ b/scripts/checkpoint_conversion/convert_to_hf.py
@@ -31,7 +31,6 @@ def convert_to_hf(
     with torch.device("cpu"):
         # pyrefly: ignore[bad-instantiation]
         model = train_spec.model_cls(model_args)
-    # pyrefly: ignore [bad-argument-type]
     model = ModelWrapper(model)
 
     # pyrefly: ignore[bad-instantiation, not-callable]
--- a/scripts/checkpoint_conversion/numerical_tests_example.py
+++ b/scripts/checkpoint_conversion/numerical_tests_example.py
@@ -78,13 +78,10 @@ def forward_tt(config_path, checkpoint_path, test_set):
 
     # materalize model
     device = torch.device(device_type)
-    # pyrefly: ignore [missing-attribute]
     model.to_empty(device=device)
     model.init_weights(buffer_device=device)
-    # pyrefly: ignore [missing-attribute]
     model.eval()
 
-    # pyrefly: ignore [bad-argument-type]
     modelWrapper = ModelWrapper(model)
     state_dict = modelWrapper._get_state_dict()
 
@@ -100,7 +97,6 @@ def forward_tt(config_path, checkpoint_path, test_set):
             input_ids = input_ids.unsqueeze(0)
 
         # obtains the logits of only the last token in the predictions
-        # pyrefly: ignore [not-callable]
         predictions = model(input_ids)[:, -1, :].unsqueeze(1)
         output_list.append(predictions)
 
--- a/scripts/estimate/estimation.py
+++ b/scripts/estimate/estimation.py
@@ -99,21 +99,17 @@ def estimate_memory(job_config: JobConfig):
 
         # Build the collection of model converters. No-op if `model.converters` empty
         model_converters = build_model_converters(job_config, parallel_dims)
-        # pyrefly: ignore [bad-argument-type]
         model_converters.convert(model)
 
         # apply PT-D DP/TP parallelisms and activation checkpointing
         train_spec.parallelize_fn(model, parallel_dims, job_config)
 
-        # pyrefly: ignore [missing-attribute]
         model.to_empty(device="cuda")
         if not active_fake_mode():
             model.init_weights()
-        # pyrefly: ignore [missing-attribute]
         model.train()
 
         # build optimizer after applying parallelisms to the model
-        # pyrefly: ignore [bad-argument-type]
         optimizers = build_optimizers([model], job_config.optimizer, parallel_dims)
         lr_schedulers = build_lr_schedulers(
             # pyrefly: ignore [bad-argument-type]
@@ -125,7 +121,6 @@ def estimate_memory(job_config: JobConfig):
         # e.g. calculate float8 dynamic amax/scale for all-parameter for FSDP2
         # where it issues a single all-reduce for all parameters at once for better performance
         optimizers.register_step_post_hook(
-            # pyrefly: ignore [bad-argument-type]
             lambda *args, **kwargs: model_converters.post_optimizer_hook(model)
         )
 
@@ -150,7 +145,6 @@ def estimate_memory(job_config: JobConfig):
                 device="cuda",
             ),
         )
-        # pyrefly: ignore [bad-argument-type]
         fsdp_memtracker = FSDPMemTracker(mod=model, optm=optimizers.optimizers[0])
         fsdp_memtracker.track_inputs(batch)
 
@@ -160,7 +154,6 @@ def estimate_memory(job_config: JobConfig):
                 input_ids, labels = batch
                 # train step
                 with train_context():
-                    # pyrefly: ignore [not-callable]
                     pred = model(input_ids)
                     loss = loss_fn(pred, labels)
                     del pred
@@ -168,7 +161,6 @@ def estimate_memory(job_config: JobConfig):
 
                 # clip gradients
                 torch.nn.utils.clip_grad_norm_(
-                    # pyrefly: ignore [missing-attribute]
                     model.parameters(),
                     job_config.training.max_norm,
                     foreach=True,
--- a/scripts/generate/test_generate.py
+++ b/scripts/generate/test_generate.py
@@ -135,7 +135,6 @@ def test_generate(
 
         # apply_tp (with Sequence Parallel) on unevenly sharded
         # sequences would require https://github.com/pytorch/torchtitan/pull/686
-        # pyrefly: ignore [bad-argument-type]
         apply_tp_minus_sp(model, parallel_dims.get_mesh("tp"))
     else:
         parallel_dims = ParallelDims(
@@ -158,14 +157,11 @@ def test_generate(
     )
 
     # materalize model
-    # pyrefly: ignore [missing-attribute]
     model.to_empty(device=device_type)
     with torch.no_grad():
         model.init_weights()
-    # pyrefly: ignore [missing-attribute]
     model.eval()
 
-    # pyrefly: ignore [missing-attribute]
     state_dict = model.state_dict()
 
     # Checkpoint Loading
--- a/tests/unit_tests/test_train_spec.py
+++ b/tests/unit_tests/test_train_spec.py
@@ -26,9 +26,9 @@ from torchtitan.protocols.train_spec import (
 )
 
 
-class FakeModel(nn.Module, ModelProtocol):
+class FakeModel(ModelProtocol):
     def __init__(self, model_args: BaseModelArgs) -> None:
-        super().__init__()
+        super().__init__(model_args)
         self.linear = nn.Linear(8, 8)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
--- a/torchtitan/components/checkpoint.py
+++ b/torchtitan/components/checkpoint.py
@@ -214,7 +214,6 @@ class CheckpointManager:
         )
 
         if self.ft_manager and not self.enable_ft_dataloader_checkpoints:
-            # pyrefly: ignore [deprecated]
             logger.warning(
                 "Fault tolerance is enabled but enable_ft_dataloader_checkpoints is False. "
                 "This means replicas can retrain over the same data multiple times, which can result in overfitting."
--- a/torchtitan/components/ft/manager.py
+++ b/torchtitan/components/ft/manager.py
@@ -165,5 +165,4 @@ def maybe_semi_sync_training(
             raise ValueError(
                 f"Unknown training method: {semi_sync_method}, only 'diloco' and 'local_sgd' are supported."
             )
-    # pyrefly: ignore [no-matching-overload]
     return nullcontext()
--- a/torchtitan/components/quantization/mx.py
+++ b/torchtitan/components/quantization/mx.py
@@ -57,19 +57,14 @@ class MXLinearConverter(QuantizationConverter):
             MXLinearConfig as TorchAOMXLinearConfig,
         )
 
-        # pyrefly: ignore [bad-assignment]
         mx_job_config: TorchAOMXLinearConfig = job_config.quantize.linear.mx
-        # pyrefly: ignore [missing-attribute]
         config = TorchAOMXLinearConfig.from_recipe_name(mx_job_config.recipe_name)
-        # pyrefly: ignore [missing-attribute]
         config.mxfp8_dim1_cast_kernel_choice = MXFP8Dim1CastKernelChoice[
             mx_job_config.mxfp8_dim1_cast_kernel_choice.upper()
         ]
-        # pyrefly: ignore [missing-attribute]
         self.filter_fqns = mx_job_config.filter_fqns
         self.config = config
         self.enabled = True
-        # pyrefly: ignore [missing-attribute]
         logger.info(f"MX training active with recipe {mx_job_config.recipe_name}")
 
     def convert(self, model: nn.Module):
--- a/torchtitan/components/tokenizer.py
+++ b/torchtitan/components/tokenizer.py
@@ -145,13 +145,10 @@ class HuggingFaceTokenizer(BaseTokenizer):
                 tokenizer = Tokenizer(bpe_model)
 
                 # Configure GPT-2 style components for proper space handling
-                # pyrefly: ignore [read-only]
                 tokenizer.pre_tokenizer = pre_tokenizers.ByteLevel(
                     add_prefix_space=False
                 )
-                # pyrefly: ignore [read-only]
                 tokenizer.decoder = decoders.ByteLevel()
-                # pyrefly: ignore [read-only]
                 tokenizer.post_processor = processors.ByteLevel(trim_offsets=True)
 
                 return tokenizer
--- a/torchtitan/distributed/tensor_parallel.py
+++ b/torchtitan/distributed/tensor_parallel.py
@@ -22,9 +22,6 @@ def maybe_enable_async_tp(job_config: JobConfig, tp_mesh: DeviceMesh):
             "Async TP requires 'model' in --compile.components and --compile.enable"
         )
 
-    from torch.distributed._symmetric_memory import enable_symm_mem_for_group
-
     torch._inductor.config._micro_pipeline_tp = True
-    enable_symm_mem_for_group(tp_mesh.get_group().group_name)
 
     logger.info("Async TP is enabled")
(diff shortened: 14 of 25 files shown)


torchtitan/models/llama3/infra/parallelize.py
```diff
@@ -202,11 +206,13 @@ def apply_tp(
     # NOTE: At the cost of model code change, we can accelerate Sequence Parallel
     #       by folding (and unfolding) the batch dimension and the sequence dimension.
     #       Examples can be found at https://github.com/pytorch/torchtitan/pull/437
+
     for transformer_block in model.layers.values():
         layer_plan = {
             "attention_norm": SequenceParallel(),
             # NOTE: when the fourth argument (positions) is not None, its input layout
-            # and desired input layout should be Replicate()
+            # and desired input layout is still None as we don't convert freqs_cis to
```
acisseJZhong, line 220: thanks for the nice catch! wondering do you want to change other models' `parallelize`? For deepseek v3, we should set to `input_layouts=(Shard(1), None, None, Replicate())` if cp_enabled = True.

fegin: Done, thanks for the reminder.

torchtitan/models/llama3/infra/parallelize.py
```diff
@@ -330,3 +336,42 @@ def apply_ddp(
     replicate(model, device_mesh=dp_mesh, bucket_cap_mb=100)
 
     logger.info("Applied DDP to the model")
+
+
+def apply_cp(
```
tianyu-l: 1. should we put this function to distributed/context_parallel.py?
2. should we apply this to all models?

torchtitan/distributed/utils.py
```diff
@@ -515,3 +513,37 @@ def _clip_grad_norm_with_ep(
...
+    from torch.distributed.tensor.experimental._attention import (
+        _context_parallel_shard,
+        _HeadTailLoadBalancer,
+    )
+    from torch.nn.attention.flex_attention import BlockMask
+
+    INPUT_SEQ_DIM = 1
+    seq_len = inputs[0].size(INPUT_SEQ_DIM)
+    cp_world_size = cp_mesh.size(0)
+    if isinstance(attention_masks, BlockMask):
```
tianyu-l: error out when it's varlen metadata?

torchtitan/distributed/utils.py
```diff
@@ -515,3 +513,37 @@ def _clip_grad_norm_with_ep(
...
+        raise ValueError(
+            "FlexAttention CP is not supported yet. Will come in the next PR."
+        )
+    else:
```
tianyu-l: error out when it's varlen metadata?

torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -241,7 +241,8 @@ def apply_non_moe_tp(
         layer_plan = {
             "attention_norm": SequenceParallel(),
             # NOTE: when the fourth argument (positions) is not None, its input layout
-            # and desired input layout should be Replicate()
+            # and desired input layout is still None as we don't convert freqs_cis to
```
tianyu-l, line 282: maybe should change this after @wwwjn 's PR?

torchtitan/train.py
```diff
@@ -475,29 +490,10 @@ def forward_backward_step(
         inputs, labels, extra_inputs, extra_kwargs = self.post_dataloading_process(
             input_dict, labels
         )
-        # apply context parallelism if cp is enabled
-        # ensure CP handles the separate freqs_cis buffer for each pp stage
```
tianyu-l, line 484: after this PR, we seem no longer needing `freqs_cis` as model input. IIRC we modified the freqs_cis-related model code logic previous to make it model input. Shall we revert those logic?

torchtitan/models/llama4/infra/parallelize.py
wwwjn: The PR #2149 has composability issue when TP + PP is applied, and I'm trying to discuss how to fix. I guess we could also land this PR if it's ready and I could rebase

torchtitan/models/llama3/infra/parallelize.py
fegin: 1. Yes.

~~2. Do we actually verify CP for all models? I think llama3 and llama4, yes. I'm thinking to re-enable CP model by model using this refactor chance.~~
~~nvm, I added to the core models. I'll leave Flux for another PR. c.c,~~ @wwwjn

torchtitan/train.py
fegin: That's a good point.

fegin: We now use max_seq_len and reassign max_seq_len during mode_args initialization. I think this doesn't count as a hack. So I only remove the legacy TODO (which I don't think we need anymore).

torchtitan/models/llama4/infra/parallelize.py
fegin: I don't have a strong opinion. Let's land whichever PR ready first.

torchtitan/components/validate.py
```diff
@@ -89,6 +91,71 @@ def __init__(
                 "unequal sample counts across ranks when dataset is exhausted."
             )
 
+    def post_dataloading_process(
+        self,
+        input_dict: dict[str, torch.Tensor],
+        labels: torch.Tensor,
+        device: torch.device,
```
tianyu-l: can we just use `input_dict["inputs"].device`?

torchtitan/components/validate.py
```diff
@@ -89,6 +91,71 @@ def __init__(
                 "unequal sample counts across ranks when dataset is exhausted."
             )
 
+    def post_dataloading_process(
+        self,
+        input_dict: dict[str, torch.Tensor],
+        labels: torch.Tensor,
```
tianyu-l, line 94: I feel we should consolidate this into `input_dict["labels"]`

torchtitan/components/validate.py
```diff
@@ -89,6 +91,71 @@ def __init__(
...
+                additional keys for auxiliary inputs (e.g., position ids).
+            labels: Target labels for the batch.
+            device: Device to use for creating new tensors (e.g., positions).
+            model_parts: List of model parts for accessing model methods.
+
+        Returns:
+            A tuple of (inputs, labels, extra_inputs, extra_kwargs) where:
+                - inputs: Main input tensor extracted from input_dict["input"].
+                - labels: Target labels (potentially modified by CP sharding).
+                - extra_inputs: Dict of auxiliary input tensors (all keys except
```
tianyu-l, line 115: maybe we can consolidate this into `input_dict` as well?

torchtitan/components/validate.py
```diff
@@ -89,6 +91,71 @@ def __init__(
...
+            while extra_inputs are only available to the first stage.
+        """
+        inputs = input_dict["input"]
+        extra_inputs = {k: v for k, v in input_dict.items() if k != "input"}
+        # For arguments, like attention_masks, we have to put them in a separate
+        # dict as extra_inputs are not forwarded to other stages in PP, but
+        # extra_kwargs are.
+        extra_kwargs: dict[str, Any] = {}
+
+        attn_type = getattr(self.model_args, "attn_type", "sdpa")
```
tianyu-l: It seems we introduce model_args to validator because of this line.

This condition is checked again in `get_attention_masks`. Can we remove it here? We can `return None` instead of throw in `get_attention_masks` when attn_type is sdpa.

The argument is that validator is not supposed to know ModelArgs details.

torchtitan/distributed/utils.py
```diff
@@ -538,3 +533,42 @@ def _clip_grad_norm_with_ep(
     torch.nn.utils.clip_grads_with_norm_(non_ep_params, max_norm, total_norm, foreach)
 
     return total_norm
+
+
+def cp_shard(
```
tianyu-l: can this function also go to context_parallel.py? maybe we can consolidate with `get_context_parallel_inputs` since both are not long.

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,116 @@
...
+    _ContextParallel,
+    _enable_context_parallel_dispatcher,
+)
+from torch.distributed.tensor.parallel import parallelize_module
+
+from torchtitan.distributed import utils as dist_utils
+from torchtitan.tools.logging import logger
+
+
+def apply_cp(
```
tianyu-l: This name sounds too generic for what it's assuming, e.g. you needed a different one for flux.

Maybe `apply_cp_to_transformer_blocks` and send in `model.layers.values()`?

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,116 @@
...
+    Apply context parallelism to the model.
+
+    Context Parallelism (CP) splits the sequence dimension across devices to enable
+    training with longer sequences. This function applies CP to the attention modules
+    of all transformer blocks in the model.
+
+    Args:
+        model: The transformer model with layers containing attention modules
+        cp_mesh: Device mesh for context parallel dimension
+        use_flex_attn: Whether the model uses FlexAttention (True) or SDPA (False)
```
wwwjn: n00b q: We don't consider varlen + CP here? why is that?

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Note:
+        - For FlexAttention: CP plan uses FLEX attention type
+        - For SDPA: Enables CP dispatcher and uses SDPA attention type
+        - Applies to transformer_block.attention.inner_attention for each layer
+    """
+    # Apply context parallelism to every transformer block
+    # TODO: make seq_sim configurable once the implementation doesn't assume 2
```
wwwjn:
```suggestion
    # TODO: make seq_dim configurable once the implementation doesn't assume 2
```

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,116 @@
...
+    # internally.
+    if use_flex_attn:
+        cp_plan = _ContextParallel(
+            seq_dim=2, attention_type=_ContextParallel.AttentionType.FLEX
+        )
+    else:
+        # This is currently required as DTensor dispatcher is not enabled to
+        # dispatch SDPA to CP implementation. We don't disable the CP
+        # dispatching in TorchTitan as it is not needed. But there is a
```
wwwjn: This comment is a little bit confusing to me - DTensor dispatcher is not enabled to dispatch SDPA to CP implementation, so we explicitly enable it by calling `_enable_context_parallel_dispatcher`. But why "we don't disable the CP dispatching in torchtitan"?

torchtitan/models/deepseek_v3/infra/parallelize.py
```diff
@@ -110,6 +112,10 @@ def parallelize_deepseekv3(
             dual_pipe_v=dual_pipe_v,
         )
 
+    if parallel_dims.cp_enabled:
+        use_flex_attn = attn_type == "flex"
```
wwwjn: So for now, it attn_type == "varlen" will fall to SDPA branch in `apply_cp`?

torchtitan/components/validate.py
fegin: Our dataloader signature return two variables input_dict and labels. So unless we consolidate the dataloader type as well, I think it is more consistent to keep this API signature the same as other APIs.

torchtitan/components/validate.py
fegin: I remember `extra_inputs` are being  a seperate dict because PP doesn't forward them across stages. I admit it is quite annoying to have separate inputs. However, it is more clear to have two separate dicts, because one will be forwarded while another one will only consumed by the first stage. We should explore how to make PP input signature more user friendly. cc., @H-Huang

torchtitan/components/validate.py
fegin: I use try/except to remove the need of model_args. When `get_attention_masks`. we intended to keep the return type simple. That's why exception is raised.

torchtitan/distributed/context_parallel.py
fegin: varlen + CP is not supported yet (and not planned yet).

torchtitan/models/deepseek_v3/infra/parallelize.py
fegin: It will error out earlier than this place (same file).

torchtitan/components/validate.py
tianyu-l: Right, I think we should also consolidate them in data loader, but maybe in a future PR.

torchtitan/models/llama3/infra/parallelize.py
```diff
@@ -208,7 +217,8 @@ def apply_tp(
         layer_plan = {
             "attention_norm": SequenceParallel(),
             # NOTE: when the fourth argument (positions) is not None, its input layout
-            # and desired input layout should be Replicate()
+            # and desired input layout is still None as we don't convert freqs_cis to
+            # a DTensor for llama3.
```
tianyu-l, line 221: could you remind me why the treatment for some model (e.g. llama3) is different from others (e.g. dsv3)? is it because some of freqs_cis are dtensors and others are not? shall we make them consistent all as DTensors? maybe leave a TODO for @wwwjn

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,149 @@
...
+    return inputs, labels, extra_kwargs
+
+
+def cp_shard(
+    cp_mesh: DeviceMesh,
+    inputs: tuple[torch.Tensor, ...],
+    attention_masks: AttentionMasksType | None,
+    disable_load_balancer: bool = False,
+):
+    INPUT_SEQ_DIM = 1
```
tianyu-l: Is there plan to lift this constraint?

Let's add a docstring for this public method.

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,149 @@
...
+    seq_len = inputs[0].size(INPUT_SEQ_DIM)
+    cp_world_size = cp_mesh.size(0)
+    if attention_masks is not None:
+        raise ValueError(
+            "FlexAttention CP is not supported yet. Will come in the next PR."
```
tianyu-l, line 172: Based on the loss curves I observed, we might want to make HeadTail load balancer also available for Flex.

torchtitan/components/validate.py
tianyu-l: try/catch doesn't sound right here -- e.g. if `model_args.attn_type` is flex but this method throws NotImplemented, then there's no way the computation can be correct.

If the return type is meant to be optional, it's type should be optional. In the long term we can try to think of a better way to encapsulate the logic between flex and sdpa.

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,149 @@
...
+from torch.distributed.tensor.parallel import parallelize_module
+
+from torchtitan.protocols.model import AttentionMasksType
+from torchtitan.tools.logging import logger
+
+
+def apply_cp_to_attention_module(
+    attention_modules: Sequence[nn.Module],
+    cp_mesh: DeviceMesh,
+    use_flex_attn: bool,
```
tianyu-l: Let's pass the type explicitly, instead of relying on boolean. It's easy to omit / hard to add other cases, e.g. varlen.

torchtitan/components/validate.py
fegin: You are right that the current code cannot differentiate Flex without get_attention_mask implementation. But imo, it is not an issue of try/except but we should throw a correct exception, which is TypeError (according to Python docs, TypeError is used for `This exception may be raised by user code to indicate that an attempted operation on an object is not supported`).

In any case, it depends on the definition of `get_attention_masks`. If it is meant to be called only by flex attention or varlen attention, then its return type is not optional and `TypeError` is a valid exception in such a case. On the other hand, if it is meant to be called even when the attention type is SDPA, then None makes sense.

Both implementations are legit IMO. But since this function is more like an adapter function, making return type of `get_attention_masks` optional seems to leak less information to the trainer, atm. If that is a more clear semantics, I can change `get_attention_masks` to return an optional type.

torchtitan/models/llama3/infra/parallelize.py
fegin: >  is it because some of freqs_cis are dtensors and others are not? shall we make them consistent all as DTensors?

https://github.com/pytorch/torchtitan/pull/2149 will fix the inconsistency. I left a TODO here.

torchtitan/distributed/context_parallel.py
fegin: I'll make it configurable in the FlexCP PR. That's the PR that PTRR is introduced.

fegin:
I changed the exception to TypeError anyway since NotImplementedError is not right for SDPA as mentioned in the above reply. But feel free to request the change to get_attention_masks if people think that API should return an optional mask.

cc., @tianyu-l

torchtitan/components/validate.py
tianyu-l: we already have two `extra_inputs` and `extra_kwargs`, so we should be able to merge inputs at least to one of them?

doesn't need to be updated in this PR.

torchtitan/experiments/forge/example_train.py
```diff
@@ -152,42 +153,57 @@ def batch_generator(
...
-
-        if getattr(self.model_args, "attn_type", "sdpa") == "flex":
-            extra_kwargs["attention_masks"] = model_parts[0].get_attention_masks(
+        extra_inputs = {k: v for k, v in input_dict.items() if k != "input"}
+        # For arguments, like attention_masks, we have to put them in a separate
+        # dict as extra_inputs are not forwarded to other stages in PP, but
+        # extra_kwargs are.
+        extra_kwargs: dict[str, Any] = {}
+
+        try:
```
tianyu-l, line 166: nit: here we have access to `self.model_args` so don't need to try-catch

torchtitan/distributed/context_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+        # For SDPA, we use the _HeadTailLoadBalancer.
+        load_balancer = (
+            None
+            if disable_load_balancer
+            else _HeadTailLoadBalancer(seq_len, cp_world_size, cp_mesh.device_type)
+        )
+
+    inputs = cast(
+        tuple[torch.Tensor, ...],
+        _context_parallel_shard(
```
tianyu-l, line 184: Last time I mentioned the idea of doing reverse permute after all-gather KV so that with vs. without load balancing, forward numerics are bitwise identical. That would require change on `_create_cp_block_mask`. 

What's your latest thought? I guess it's not absolutely necessary for torchtitan CP right now, but maybe good to keep in mind.

fegin:
Landing the PR, https://github.com/pytorch/pytorch/pull/172519 is the candidate fix to the compile issue.

torchtitan/distributed/context_parallel.py
fegin: The only concern is that are we going to make the reversing option as the only option or do we want to make it as configurable? It can be error-prone to make it as configurable as there may be more than one place that is order-aware (BlockMask creation and allgather) -- users need to ensure the consistent option. Or we can make it reverse anyway, but the performance indication is not clear to me.

torchtitan/experiments/forge/example_train.py
fegin: Prefer to keep the consistent style for the same logic.

torchtitan/components/validate.py
fegin: PP treats `extra_inputs` and `extra_kwargs` differently. `extra_inputs` will not be forwarded to next stages while `extra_kwargs` will be forwarded to next stages.  

We should aim to merge "inputs" and "extra_inputs" if possible.

Lucaskabela:
Only note on https://github.com/pytorch/pytorch/pull/172519 is that it will require patching `skip_nnmodule_hook_guards` or `inline_inbuilt_nn_modules` in the dynamo config to have correctness for the hooks if the hooks differ per layer

torchtitan/distributed/context_parallel.py
tianyu-l: I think we should make it the only option, as long as perf is OK. This needs to be tested. Not urgent though.

PR #2145, opened 2025-12-11
[CP] Enable FlexCP for llama3

fegin:
* __->__ #2145

Summary:

Continue the previous PR, this PR enable FlexAttention + CP for llama3. FlexCP will use PTRRLoadBalancer.

Note that this PR requires https://github.com/pytorch/pytorch/pull/170201


torchtitan/distributed/utils.py
```diff
@@ -523,16 +523,15 @@ def cp_shard(
...
         _context_parallel_shard,
         _HeadTailLoadBalancer,
+        _PTRRLoadBalancer,
     )
     from torch.nn.attention.flex_attention import BlockMask
 
     INPUT_SEQ_DIM = 1
     seq_len = inputs[0].size(INPUT_SEQ_DIM)
     cp_world_size = cp_mesh.size(0)
     if isinstance(attention_masks, BlockMask):
```
tianyu-l: If it's a dict, then we use HeadTail by default? If there's longer term plan, shall we add a TODO here?

torchtitan/distributed/utils.py
```diff
@@ -546,4 +545,22 @@ def cp_shard(
...
+    if attention_masks is not None:
+        masks = (
+            [attention_masks]
+            if isinstance(attention_masks, BlockMask)
+            else list(attention_masks.values())
+        )
+        masks = _context_parallel_shard(
+            mesh=cp_mesh,
+            buffers=masks,
+            seq_dims=(2,) * len(masks),
```
tianyu-l: pls add a comment why it's 2 here

torchtitan/distributed/utils.py
```diff
@@ -546,4 +545,22 @@ def cp_shard(
         load_balancer=load_balancer,
     )
 
+    if attention_masks is not None:
```
tianyu-l: what if it's varlen metadata

IIUC you error out in arg.py, but to be safe it's good if we also error out here

fegin: After rebasing, there is a assert here to avoid type checking error.

torchtitan/models/attention.py
```diff
@@ -116,7 +117,8 @@ def forward(
         k: torch.Tensor,
         v: torch.Tensor,
         *,
-        block_mask: BlockMask,
+        score_mod: _score_mod_signature | None = None,
```
tianyu-l, line 122: arg not used anywhere

torchtitan/distributed/context_parallel.py
```diff
@@ -121,20 +123,30 @@ def cp_shard(
...
-        load_balancer = (
-            None
-            if disable_load_balancer
-            else _HeadTailLoadBalancer(seq_len, cp_world_size, cp_mesh.device_type)
-        )
+
+    load_balancer = None
+    if not disable_load_balancer:
+        if isinstance(attention_masks, BlockMask):
+            load_balancer = _PTRRLoadBalancer(attention_masks, cp_world_size)
```
tianyu-l: As mentioned in the previous PR, I wonder if we should keep HeadTail as an option for Flex. I've seen some loss curves with HeadTail yielding slightly better results, but maybe it's biased. Plus it should be a straightforward config to add.

torchtitan/models/attention.py
fegin: We need score_mod to ensure the function signature is the same as flex_attention so that the module hooks know how to allgather K, V.

torchtitan/config/job_config.py
```diff
@@ -420,6 +420,14 @@ class Parallelism:
     context_parallel_degree: int = 1
     """Context parallelism degree. 1 means disabled."""
 
+    context_parallel_load_balancer: str | None = "headtail"
+    """
+    Load balancer type for context parallelism. Options:
+    - "headtail": Use HeadTailLoadBalancer for SDPA
+    - "ptrr": Use PTRRLoadBalancer for FlexAttention
+    - None or "": Disable load balancing
```
tianyu-l: Let's follow titan convention and use `None` exclusively to represent "disable".

tianyu-l (approved):
one nit

rakkit:
@fegin noob question, does it means sliding-window attention (FlexAttn backend) + CP can also work?

PR #1802, opened 2025-10-06
[VLM] Refactor special tokens by subclassing Tokenizer

lkhphuc:
Follow up on the previous comment https://github.com/pytorch/torchtitan/pull/1615#discussion_r2369059607, in this PR I refactor the bit about special tokens by subclassing the tokenizer and add the special tokens as class attribute there.

Also, we are adding the special tokens dynamically to the tokenizer, and the tokenizer will create new tokens if they are not already exists in the tokenizer's vocab.

This make it cleaner later when we provide the script to load pretrained LLM, so that we don't have to instruct users to separately download and modify the tokenizer's config manually.

PR #2148, opened 2025-12-12
Staging SFT training

rakkit:
### TL;DR
Adds SFT training to Torchtitan plus a small `greedy_packing` addition. 

### Changes
- Added SFT dataset config to job config
- Updated attention + modifying `get_attention_masks` to support SFT masks (only landed on Llama3 now)
- Temporarily use `HFTokenizer`,  need to fix this later
- Added SFT dataset/dataloader that returns `input_ids`, `labels` (user tokens masked), `attention_masks`, `position_ids`

###  TODO
- Need to verify PP / CP
- Val dataset not working yet (may depend on https://github.com/pytorch/torchtitan/pull/2142)

### Run
```bash
CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh \
  --training.running_sft_training \
  --model.flavor=debugmodel_varlen_attn \
  --training.dataset_path=openai/gsm8k \
  --sft_data_config.dataset_subset=main
  ```
  
image

## more test
```
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" NGPU=4 ./run_train.sh \
 --training.running_sft_training \ --model.flavor=8B_varlen \ --training.dataset_path=openai/gsm8k \
 --sft_data_config.dataset_subset=main \
 --model.hf_assets_path="$Home/Meta-Llama-3.1-8B-Instruct/" \
 --training.local_batch_size=8 \ --activation_checkpoint.mode=full \
 --training.seq_len=2048 \ --training.steps=100 \
 --lr_scheduler.warmup_steps=10 \  --debug.seed 10 \
 --sft_data_config.pad_mode=right_padding \
 --metrics.enable_wandb \
 ```
 (`torch                    2.10.0.dev20251124+cu129` and i am using cudnn attention )
image

W B Chart 12_12_2025, 8_32_32 PM

compile does not work for no-padding because the seq-len for each training step keeps changing. We could pad the buffer to `seqlen` when turning on `greedy_packing`. (its `packing_on++`) to make compile happy. 

image

### DIFF
--- a/torchtitan/experiments/sft/auto_tokenizer.py
+++ b/torchtitan/experiments/sft/auto_tokenizer.py
@@ -1,0 +1,90 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torchtitan.components.tokenizer import BaseTokenizer
+from torchtitan.config import JobConfig
+from torchtitan.tools.logging import logger
+from transformers import AutoTokenizer as HF_AutoTokenizer
+
+
+class HuggingFaceAutoTokenizer(BaseTokenizer):
+    def __init__(
+        self,
+        tokenizer_path: str,
+        eos_token: str,
+        pad_token_id: int,
+        pad_token: str | None = None,
+    ):
+        self.tokenizer = HF_AutoTokenizer.from_pretrained(
+            tokenizer_path, eos_token=eos_token, use_fast=True
+        )
+        self.tokenizer_path = tokenizer_path
+        self.chat_template = self.tokenizer.chat_template
+
+        self.vocab_size = len(self.tokenizer)
+
+        assert (
+            pad_token_id < self.vocab_size
+        ), f"PAD token ID is out of range: {pad_token_id} >= {self.vocab_size}"
+        assert (
+            pad_token_id != self.tokenizer.eos_token_id
+        ), "PAD token ID is the same as EOS token ID, this can cause problems with varlen/flex attention in dynamic packing."
+
+        self.eos_id = self.tokenizer.eos_token_id
+        self.bos_id = self.tokenizer.bos_token_id
+        self.bos_token = self.tokenizer.bos_token
+        self.eos_token = self.tokenizer.eos_token
+
+        self.pad_id = pad_token_id
+
+        self.maybe_original_token_at_pad_id = self.tokenizer._convert_id_to_token(
+            pad_token_id
+        )
+        if pad_token is not None:
+            self.pad_token = pad_token
+        else:
+            self.pad_token = self.maybe_original_token_at_pad_id
+
+        logger.info(
+            f"[SFT AutoTokenizer] Using EOS token: {self.eos_token} - EOS ID: {self.eos_id} "
+            f"[SFT AutoTokenizer] Using PAD token: {self.pad_token} - PAD ID: {self.pad_id}"
+        )
+
+    def apply_chat_template(self, messages: list[dict], **kwargs):
+        return self.tokenizer.apply_chat_template(messages, **kwargs)
+
+    def encode(self, text: str, *args, **kwargs):
+        return self.tokenizer.encode(text, *args, **kwargs)
+
+    def decode(self, tokens: list[int], *args, **kwargs):
+        decoded = self.tokenizer.decode(tokens, *args, **kwargs)
+        # this is an ad-hoc for debugging purpose
+        if self.pad_id in tokens:
+            decoded = decoded.replace(
+                self.maybe_original_token_at_pad_id, self.pad_token
+            )
+        return decoded
+
+    def get_vocab_size(self):
+        return self.tokenizer.vocab_size
+
+    def __call__(self, text: str, *args, **kwargs):
+        return self.tokenizer(text, *args, **kwargs)
+
+
+def build_auto_tokenizer(job_config: JobConfig) -> HuggingFaceAutoTokenizer:
+    eos_token = job_config.sft_config.eos_token
+    pad_token_id = job_config.sft_config.pad_token_id
+    pad_token = job_config.sft_config.pad_token
+    assert (
+        eos_token is not None and pad_token_id is not None
+    ), "EOS and PAD token IDs must be provided"
+    return HuggingFaceAutoTokenizer(
+        job_config.model.hf_assets_path,
+        eos_token=eos_token,
+        pad_token_id=pad_token_id,
+        pad_token=pad_token,
+    )
--- a/torchtitan/experiments/sft/job_config.py
+++ b/torchtitan/experiments/sft/job_config.py
@@ -1,0 +1,80 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass, field
+from typing import Literal
+
+
+@dataclass
+class SFTConfig:
+    split: str = "train"
+    """Split to use"""
+    dataset_subset: str | None = None
+    """Subset to use"""
+    stream_dataset: bool = True
+    """Whether to stream the dataset"""
+    """
+    At moment we support two types of datasets:
+    # about dataset
+    1. **Multi-turn (`True`):** Loads data from the column specified by `messages_key`.
+        The content must be a list of dictionaries containing 'role' and 'content'.
+        Example:
+        ```
+        [
+            {
+                "role": "user",
+                "content": "Hello, how are you?"
+            },
+            {
+                "role": "assistant",
+                "content": "I am good, thank you!"
+            }
+        ]
+        ```
+    2. **Single-turn (`False`):** Loads data from two separate columns
+        specified by `prompt_key` and `response_key`.
+
+    """
+
+    apply_chat_template: bool = False
+    """
+    If True, we will apply chat template to the messages, otherwise we will use the raw messages.
+    It is important that chat_templated should be properly configured for the tokenizer and
+    it can render the messages for multi-turn mode.
+    """
+    pad_mode: Literal["right_padding", "greedy_packing"] = "greedy_packing"
+    """
+    Padding mode to use.
+    - "right": One batch contains only one unique sequence, the sequence will be padded to the right to reach the max length.
+    - "greedy_packing": One batch contains multiple short sequences and then be padded to the right to reach the max length.
+    """
+    chat_template_kwargs: dict = field(default_factory=dict)
+    """
+    When apply chat template, we will pass these kwargs to the tokenizer.apply_chat_template function.
+    It has to be HF's apply_chat_template kwargs.
+    """
+    ignore_input_ids_mismatch: bool = False
+    """
+    Whether to ignore the input_ids mismatch when apply chat template.
+    """
+
+    eos_token: str | None = None
+    """
+    EOS token to use.
+    """
+    pad_token_id: int | None = None
+    """
+    PAD token ID to use.
+    """
+    pad_token: str | None = None
+    """
+    PAD token to override the tokenizer's pad token.
+    """
+
+
+@dataclass
+class JobConfig:
+    sft_config: SFTConfig = field(default_factory=SFTConfig)
--- a/torchtitan/experiments/sft/llama3/__init__.py
+++ b/torchtitan/experiments/sft/llama3/__init__.py
@@ -1,0 +1,40 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+# from torchtitan.components.loss import build_cross_entropy_loss
+from torchtitan.components.lr_scheduler import build_lr_schedulers
+from torchtitan.components.optimizer import build_optimizers
+from torchtitan.distributed.pipeline_parallel import pipeline_llm
+
+from torchtitan.experiments.sft.auto_tokenizer import build_auto_tokenizer
+
+from torchtitan.experiments.sft.sft_text_datasets import (
+    build_sft_text_dataloader,
+    build_sft_validation_dataloader,
+)
+
+from torchtitan.experiments.vlm.infra.loss import build_token_imbalance_ce_loss
+
+from torchtitan.models.llama3 import llama3_args, Transformer
+from torchtitan.models.llama3.infra.parallelize import parallelize_llama
+from torchtitan.models.llama3.model.state_dict_adapter import Llama3StateDictAdapter
+from torchtitan.protocols.train_spec import TrainSpec
+
+
+def get_train_spec() -> TrainSpec:
+    return TrainSpec(
+        model_cls=Transformer,
+        model_args=llama3_args,
+        parallelize_fn=parallelize_llama,
+        pipelining_fn=pipeline_llm,
+        build_optimizers_fn=build_optimizers,
+        build_lr_schedulers_fn=build_lr_schedulers,
+        build_dataloader_fn=build_sft_text_dataloader,
+        build_tokenizer_fn=build_auto_tokenizer,
+        # build_cross_entropy_loss will averaged on pad tokens unexpectedly
+        build_loss_fn=build_token_imbalance_ce_loss,
+        build_validator_fn=build_sft_validation_dataloader,
+        state_dict_adapter=Llama3StateDictAdapter,
+    )
--- a/torchtitan/experiments/sft/sft_text_datasets.py
+++ b/torchtitan/experiments/sft/sft_text_datasets.py
@@ -1,0 +1,622 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# code are heavily borrowed from
+# [1] https://github.com/volcengine/verl/blob/main/verl/utils/dataset/multiturn_sft_dataset.py
+# [2] https://github.com/OpenRLHF/OpenRLHF/blob/main/openrlhf/datasets/sft_dataset.py#L35
+# [3] https://github.com/volcengine/verl/blob/main/verl/utils/dataset/sft_dataset.py#L33
+from dataclasses import field
+from functools import partial
+from typing import Any, Callable, Optional
+
+import torch
+import torch.nn.functional as F
+
+from datasets import Dataset, load_dataset
+from datasets.distributed import split_dataset_by_node
+from torch.distributed.checkpoint.stateful import Stateful
+from torch.utils.data import IterableDataset
+
+from torchtitan.components.dataloader import ParallelAwareDataloader
+
+from torchtitan.components.loss import IGNORE_INDEX
+from torchtitan.components.tokenizer import BaseTokenizer
+from torchtitan.config.job_config import JobConfig
+from torchtitan.experiments.sft.job_config import SFTConfig
+from torchtitan.tools.logging import logger
+
+
+def _build_multi_turn_messages_from_row_dict(
+    row_dict: dict,
+    messages_key: str = "messages",
+    tools_key: str = "tools",
+    thinking_key: str = "thinking",
+):
+    """Build multi-turn messages from row dictionary."""
+    # return the message, tools, and enable_thinking
+    message = row_dict[messages_key]
+    tools = row_dict.get(tools_key, None)
+    enable_thinking = row_dict.get(thinking_key, None)
+    if isinstance(enable_thinking, str):
+        enable_thinking = enable_thinking == "on"
+    return message, tools, enable_thinking
+
+
+def _build_prompt_response_messages_from_row_dict(
+    row_dict: dict,
+    prompt_key: str = "prompt",
+    response_key: str = "response",
+):
+    """Build one turn messages from row dictionary."""
+    # return the message, tools, and enable_thinking
+    message = [
+        {"role": "user", "content": row_dict[prompt_key]},
+        {"role": "assistant", "content": row_dict[response_key]},
+    ]
+    return message, None, None
+
+
+DATASET_MESSAGE_BUILDERS = {
+    "multi_turn": _build_multi_turn_messages_from_row_dict,
+    "prompt_response": _build_prompt_response_messages_from_row_dict,
+    "question_answer": partial(
+        _build_prompt_response_messages_from_row_dict,
+        prompt_key="question",
+        response_key="answer",
+    ),
+}
+
+
+"""
+From DataFrame rows to a "sequence"
+
+1) [Get row data]
+   Iterate over the DataFrame to obtain a per-row dictionary (row_dict).
+
+2) [Build messages]
+   Given row_dict, call `_build_messages` to convert the original data into `messages`.
+
+   Here, `messages` is a list of dicts with at least the keys "role" and "content", e.g.
+   [
+       {"role": "user", "content": "Hello, how are you?"},
+       {"role": "assistant", "content": "I am good, thank you!"}
+   ]
+
+   We currently support two ways to parse the original data into `messages`:
+
+   2.1) [QA mode]
+        If this is NOT multi-turn mode, we build messages by explicitly assigning "user"
+        and "assistant" roles, reading from `prompt_key` and `response_key`:
+        [
+            {"role": "user", "content": row_dict[prompt_key]},
+            {"role": "assistant", "content": row_dict[response_key]},
+        ]
+
+   2.2) [Multi-turn mode]
+        If this IS multi-turn mode (even if there is only one turn), we read the full
+        message list from the column `messages_key`. The expected input format is:
+        {
+            messages_key: [
+                {"role": "user", "content": "Hello, how are you?"},
+                {"role": "assistant", "content": "I am good, thank you!"},
+                {"role": "user", "content": "What is your name?"},
+            ]
+        }
+
+   2.3) [Not landed yet]
+        If your data format is completely different, you can implement your own
+        `_build_messages` function. It must return a `messages` list in the same format.
+
+3) [Convert messages to tokens]
+   After we have `messages`, we tokenize them by calling `_process_one_row`.
+
+   3.1) If there is only ONE message, we tokenize it via `_process_single_message`:
+        - Either tokenize the message["content"] directly, OR
+        - Apply the tokenizer's chat template (requires the chat template to be configured).
+
+   3.2) If there are MULTIPLE messages, we call `_process_single_message` for each message.
+        There is one special case:
+        When `apply_chat_template=True`, each call to `_process_single_message` may add a
+        [system prompt]. That can produce:
+        [System] [User] [System] [Assistant] [System] [User] [System] [Assistant] ...
+        which is not what we want for a single conversation. Therefore, we include special
+        handling to remove the unexpected [system prompt] and [generation prompt] added
+        from the second message onward.
+"""
+
+"""
+From one or more "sequences" to a "batch"
+
+We offer two modes to handle padding and truncation when forming a batch.
+Padding is needed because sequences may have different lengths, and PyTorch
+cannot stack variable-length sequences into a single tensor directly.
+
+Example:
+    batch = [
+        sequence_1,
+        sequence_2,
+        sequence_3,
+    ]
+
+### Right padding
+The simplest approach is to right-pad each sequence to a target length using
+a special `<pad_token>`. The batch then looks like:
+    batch = [
+        [sequence_1 + several <pad_token>],
+        [sequence_2 + several <pad_token>],
+        [sequence_3 + several <pad_token>],
+    ]
+
+After right padding, all sequences have the same length. We set the
+`attention_mask` for `<pad_token>` positions to 0 so that attention
+implementations such as `varlen_attn` or `flex_attn` can ignore them.
+
+### Greedy packing
+Alternatively, we can use greedy packing to reduce padding waste by packing
+multiple sequences into a single fixed-length row:
+    batch = [
+        [sequence_1, sequence_2, sequence_3],
+        [sequence_4, sequence_5, <pad_token>],
+    ]
+    **We need strictly compact padding without no gap between sequences.**
+
+We still right-pad each packed row to the target length with `<pad_token>`.
+To prevent cross-document attention, we must also infer (or track) the
+boundaries between sequences within each packed row.
+"""
+
+
+def extract_system_prompt_and_generation(tokenizer):
+    """Derive system and generation prompt token chunks from the tokenizer's chat template."""
+    token1 = tokenizer.apply_chat_template(
+        [{"role": "user", "content": ""}], add_generation_prompt=False, tokenize=True
+    )
+    token2 = tokenizer.apply_chat_template(
+        [{"role": "user", "content": ""}] * 2,
+        add_generation_prompt=False,
+        tokenize=True,
+    )
+    # get system prompt tokens
+    system_prompt = token1[: -(len(token2) - len(token1))]
+    # get generate prompt tokens
+    token3 = tokenizer.apply_chat_template(
+        [{"role": "user", "content": ""}], add_generation_prompt=True, tokenize=True
+    )
+    generate_prompt = token3[len(token1) :]
+
+    return system_prompt, generate_prompt
+
+
+class SFTDataset(IterableDataset, Stateful):
+    """
+    Iterable dataset that tokenizes a conversational stream for SFT training.
+    """
+
+    def __init__(
+        self,
+        dataset,
+        tokenizer: BaseTokenizer,
+        message_builder: Callable,
+        seq_len: int = 2048,
+        dp_rank: int = 0,
+        dp_world_size: int = 1,
+        infinite: bool = False,
+        sft_config: SFTConfig = field(default_factory=SFTConfig()),
+    ) -> None:
+        super().__init__()
+        self.tokenizer = tokenizer
+        self._data = split_dataset_by_node(dataset, dp_rank, dp_world_size)
+        self.message_builder = message_builder
+        self.infinite = infinite
+
+        self.sft_config = sft_config
+
+        self.pad_mode = sft_config.pad_mode
+        self.ignore_input_ids_mismatch = sft_config.ignore_input_ids_mismatch
+
+        self.max_length = seq_len
+        self.apply_chat_template_kwargs = sft_config.chat_template_kwargs
+
+        self.apply_chat_template = sft_config.apply_chat_template
+        if self.apply_chat_template:
+            assert self.tokenizer.chat_template is not None, (
+                f"Chat template is not set for the tokenizer {self.tokenizer.tokenizer_path}, "
+                f"please set it in the tokenizer config file"
+            )
+            (
+                self.system_prompt,
+                self.generation_prompt,
+            ) = extract_system_prompt_and_generation(self.tokenizer)
+        else:
+            self.system_prompt = torch.tensor([], dtype=torch.long)
+            self.generation_prompt = torch.tensor([], dtype=torch.long)
+
+        logger.info(
+            f"[sft_text_datasets.py] Infer system_prompt: {self.tokenizer.decode(self.system_prompt)}"
+        )
+        logger.info(
+            f"[sft_text_datasets.py] Infer generation_prompt: {self.tokenizer.decode(self.generation_prompt)}"
+        )
+
+        self.pad_id = self.tokenizer.pad_id
+        self.pad_token = self.tokenizer.pad_token
+        self.eos_id = self.tokenizer.eos_id
+
+        self.buffer_max_length = self.max_length
+        # Stateful variables
+        self._sample_idx = 0
+        self._buffer = self._reset_buffer()
+
+    def _reset_buffer(self):
+        """Reset the greedy packing buffer to empty tensors."""
+        return {
+            "input_ids": [],
+            "position_ids": [],
+            "labels": [],
+            "segment_lens": [],
+            "current_len": 0,
+        }
+
+    def _get_data_iter(self):
+        """Return an iterator over the local partition, resuming for map-style datasets."""
+        # For map-style datasets, resume by skipping to the correct index
+        # For iterable-style datasets, the underlying iterator already points to the correct index
+        if isinstance(self._data, Dataset):
+            if self._sample_idx == len(self._data):
+                return iter([])
+            else:
+                return iter(self._data.skip(self._sample_idx))
+
+        return iter(self._data)
+
+    def load_state_dict(self, state_dict):
+        self._buffer = state_dict["buffer"]
+
+        if isinstance(self._data, Dataset):
+            self._sample_idx = state_dict["sample_idx"]
+        else:
+            assert "dataset" in state_dict
+            self._data.load_state_dict(state_dict["dataset"])
+
+    def state_dict(self):
+        _state_dict = {"buffer": self._buffer}
+
+        if isinstance(self._data, Dataset):
+            _state_dict["sample_idx"] = self._sample_idx
+        else:
+            # Save the iterable dataset's state to later efficiently resume from it
+            # https://huggingface.co/docs/datasets/v3.5.0/en/stream#save-a-dataset-checkpoint-and-resume-iteration
+            _state_dict["dataset"] = self._data.state_dict()
+
+        return _state_dict
+
+    def _process_single_message(
+        self,
+        index: int,
+        message: dict[str, Any],
+        tools: Optional[list[dict[str, Any]]] = None,
+        enable_thinking: Optional[bool] = None,
+    ) -> tuple[list[int], list[int], list[int]]:
+        """Tokenize one conversation turn while applying template overrides."""
+        apply_chat_template_kwargs = {**self.apply_chat_template_kwargs}
+        if enable_thinking is not None:
+            apply_chat_template_kwargs["enable_thinking"] = enable_thinking
+        if self.apply_chat_template:
+            inputs = self.tokenizer.apply_chat_template(
+                [message],
+                tools=tools,
+                add_generation_prompt=False,
+                tokenize=True,
+                return_dict=True,
+                return_attention_mask=False,
+                return_tensors="pt",
+                **apply_chat_template_kwargs,
+            )
+            inputs = dict(inputs)
+            input_ids = inputs.pop("input_ids")[0]
+        else:
+            content = message["content"]
+            if isinstance(content, list):
+                content = "".join(
+                    [item["text"] for item in content if item["type"] == "text"]
+                )
+            enc = self.tokenizer(
+                content,
+                add_special_tokens=False,
+                return_tensors="pt",
+                return_attention_mask=False,
+            )
+            input_ids = enc["input_ids"][0]
+
+            if message["role"] == "assistant":
+                input_ids = torch.cat([input_ids, input_ids.new_tensor([self.eos_id])])
+
+        # remove system prompt if exists
+        if index != 0 and message["role"] != "system":
+            input_ids = input_ids[len(self.system_prompt) :]
+
+        if message["role"] == "assistant":
+            loss_mask = torch.ones_like(input_ids)
+            # mask out generation prompt if assistant message
+            loss_mask[: len(self.generation_prompt)] = 0
+        else:
+            loss_mask = torch.zeros_like(input_ids)
+
+        return input_ids, loss_mask
+
+    def sanity_check(
+        self,
+        input_ids: torch.Tensor,
+        messages: list[dict],
+        tools: list[dict],
+        enable_thinking: bool,
+    ):
+        """Ensure concatenated per-turn templates match a single-shot template invocation."""
+        if not self.apply_chat_template:
+            return
+        apply_chat_template_kwargs = {**self.apply_chat_template_kwargs}
+        if enable_thinking is not None:
+            apply_chat_template_kwargs["enable_thinking"] = enable_thinking
+        inputs = self.tokenizer.apply_chat_template(
+            messages,
+            tools=tools,
+            add_generation_prompt=False,
+            tokenize=True,
+            return_dict=True,
+            return_tensors="pt",
+            **apply_chat_template_kwargs,
+        )
+
+        error_message = (
+            "MultiTurnSFTDataset apply_chat_template to each turn separately and concat `input_ids` "
+            "as a whole sequence, which may not equal to apply_chat_template to whole messages at once.\n"
+            "For example, Qwen Thinking series models add <think></think> tags to last turn, please check "
+            "your tokenizer chat template settings.\n"
+            "Set `ignore_input_ids_mismatch=True` to ignore input_ids mismatch and use the concatenated "
+            "input_ids as the final input_ids. "
+        )
+
+        if not torch.equal(input_ids, inputs["input_ids"].squeeze(0)):
+            if self.ignore_input_ids_mismatch:
+                logger.warning_once(error_message)
+            else:
+                raise AssertionError(error_message)
+
+    def _process_one_row(self, row_dict: dict):
+        """Convert a dataset row into model-ready tensors with causal labels."""
+        messages, tools, enable_thinking = self.message_builder(row_dict)
+
+        # tokenize each message
+        input_ids, loss_mask = [], []
+        for i, message in enumerate(messages):
+            _input_ids, _loss_mask = self._process_single_message(
+                index=i,
+                message=message,
+                tools=tools if i == 0 else None,
+                enable_thinking=enable_thinking,
+            )
+            input_ids.append(_input_ids)
+            loss_mask.append(_loss_mask)
+
+        input_ids = torch.cat(input_ids, dim=0)
+        loss_mask = torch.cat(loss_mask, dim=0)
+
+        self.sanity_check(input_ids, messages, tools, enable_thinking)
+
+        # when chat template is applied, append the EOS token to the input_ids and loss_mask
+        if self.apply_chat_template:
+            input_ids = torch.cat(
+                [input_ids, input_ids.new_tensor([self.eos_id])], dim=0
+            )
+            loss_mask = torch.cat([loss_mask, loss_mask.new_tensor([0])], dim=0)
+
+        position_ids = torch.arange(input_ids.shape[0], dtype=torch.long)  # (seq_len,)
+
+        # comment out these two lines to log the actual text for debugging purpose
+        # actaul_text = self.tokenizer.decode(input_ids, skip_special_tokens=False)
+        # logger.info(f"actual_text: {actaul_text} ||-> last mask : {loss_mask[-3:]}")
+
+        # handle padding
+        sequence_length = input_ids.shape[0]
+        target_length = self.max_length + 1
+
+        # Calculate valid length (unpadded) of the sequence for the model
+        # Note: We slice input_ids[:-1] later, so the valid length for training is len - 1
+        # If truncated, it is target_length - 1
+        if self.pad_mode == "right_padding":
+            if sequence_length < target_length:
+                # Pad sequences
+                pad_token_id = self.pad_id
+                padded_input_ids = torch.full(
+                    (target_length - sequence_length,),
+                    pad_token_id,
+                    dtype=input_ids.dtype,
+                )
+                padded_loss_mask = torch.zeros(
+                    (target_length - sequence_length,), dtype=loss_mask.dtype
+                )
+
+                input_ids = torch.cat((input_ids, padded_input_ids))
+                loss_mask = torch.cat((loss_mask, padded_loss_mask))
+                position_ids = F.pad(
+                    position_ids, (0, target_length - sequence_length), value=0
+                )
+            elif sequence_length > target_length:
+                # "right_trunc":
+                input_ids = input_ids[:target_length]
+                loss_mask = loss_mask[:target_length]
+                position_ids = position_ids[:target_length]
+
+        elif self.pad_mode == "greedy_packing":
+            # notice the actual packing logic happens in the `_yield_buffer` function.
+            # truncate if longer than max_length (respect truncation setting)
+            if len(input_ids) > target_length:
+                input_ids = input_ids[:target_length]
+                loss_mask = loss_mask[:target_length]
+                position_ids = position_ids[:target_length]
+            # In GREEDY_PACKING mode, keep a real attention mask (all ones).
+            # Collate will pad it later in `collate_sft_batch`.
+        else:
+            raise ValueError(f"Unknown pad mode {self.pad_mode}")
+
+        labels = input_ids[1:].clone()
+        labels[loss_mask[1:] == 0] = IGNORE_INDEX
+        input_ids = input_ids[:-1]
+        position_ids = position_ids[:-1]
+
+        return input_ids, labels, position_ids
+
+    def _greedy_pack_buffer(self):
+        if not self._buffer["input_ids"]:
+            return None
+
+        # Concatenate buffer
+        input_ids = torch.cat(self._buffer["input_ids"], dim=0)
+        labels = torch.cat(self._buffer["labels"], dim=0)
+        positions = torch.cat(self._buffer["position_ids"], dim=0)
+
+        L = int(input_ids.numel())
+        T = int(self.buffer_max_length)
+        if L < T:
+            pad_len = T - L
+            input_ids = F.pad(input_ids, (0, pad_len), value=self.pad_id)
+            positions = F.pad(positions, (0, pad_len), value=0)
+            labels = F.pad(labels, (0, pad_len), value=IGNORE_INDEX)
+
+        return {
+            "input": input_ids,
+            "positions": positions,
+        }, labels
+
+    def __iter__(self):
+        while True:
+            for sample in self._get_data_iter():
+                input_ids, labels, positions = self._process_one_row(sample)
+                new_len = input_ids.shape[0]
+
+                if self.pad_mode == "right_padding":
+                    # Yield consistent dict structure immediately
+                    return_dict = {
+                        "input": input_ids,
+                        "positions": positions,
+                    }
+                    yield return_dict, labels
+                    self._sample_idx += 1
+                    continue
+
+                if self._buffer["current_len"] + new_len > self.buffer_max_length:
+                    if self._buffer["current_len"] > 0:
+                        yield self._greedy_pack_buffer()
+
+                    self._buffer = self._reset_buffer()
+                    self._buffer["input_ids"].append(input_ids)
+                    self._buffer["position_ids"].append(positions)
+                    self._buffer["labels"].append(labels)
+                    self._buffer["current_len"] = new_len
+                else:
+                    self._buffer["input_ids"].append(input_ids)
+                    self._buffer["position_ids"].append(positions)
+                    self._buffer["labels"].append(labels)
+                    self._buffer["current_len"] += new_len
+                self._sample_idx += 1
+
+            if not self.infinite:
+                logger.warning("Dataset has run out of data")
+                break
+            else:
+                # Reset offset for the next iteration
+                self._sample_idx = 0
+                logger.warning("Dataset is being re-looped")
+                # Ensures re-looping a dataset loaded from a checkpoint works correctly
+                if not isinstance(self._data, Dataset):
+                    if hasattr(self._data, "set_epoch") and hasattr(
+                        self._data, "epoch"
+                    ):
+                        self._data.set_epoch(self._data.epoch + 1)
+
+
+def build_sft_text_dataloader(
+    dp_world_size: int,
+    dp_rank: int,
+    tokenizer: BaseTokenizer,
+    job_config: JobConfig,
+    infinite: bool = True,
+) -> ParallelAwareDataloader:
+    """Build a data loader for HuggingFace datasets."""
+    dataset_name = job_config.training.dataset
+    dataset_path = job_config.training.dataset_path
+    batch_size = job_config.training.local_batch_size
+    seq_len = job_config.training.seq_len
+
+    sft_config = job_config.sft_config
+    # TODO: Improving the dataset loading, its easy to fix
+    dataset = load_dataset(
+        dataset_path,
+        sft_config.dataset_subset,
+        split=sft_config.split,
+        streaming=sft_config.stream_dataset,
+    )
+
+    message_builder = DATASET_MESSAGE_BUILDERS[dataset_name]
+    hf_ds = SFTDataset(
+        dataset=dataset,
+        message_builder=message_builder,
+        tokenizer=tokenizer,
+        seq_len=seq_len,
+        dp_rank=dp_rank,
+        dp_world_size=dp_world_size,
+        infinite=infinite,
+        sft_config=sft_config,
+    )
+
+    return ParallelAwareDataloader(
+        dataset=hf_ds,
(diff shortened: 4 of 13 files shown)


rakkit:
Just confirm it works on the SFT dataset in "multiturn" format, with/ and w/o the tool, and on reasoning data.

(Be aware that when it turns on `apply_chat_template`, we are supposed to provide the `chat_template.jinja` in tokenzier's folder. ~~There is no such file for "Meta-Llama-3.1-8B-Instruct". For test purposes, you can use, e.g,. tokenizer from [Olmo-3-7B-Instruct](https://huggingface.co/allenai/Olmo-3-7B-Instruct)~~.

llama-3.1 (pretrained) model dont have chat_template,  and you can find chat_template for Llama-3.1-8B-Instruct in `tokenizer_config.json`

---
### TO reproduce it on the `multiturn` format dataset [ Dolci-Instruct-SFT](https://huggingface.co/datasets/allenai/Dolci-Instruct-SFT), without `apply_chat_template`
```
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" NGPU=4 ./run_train.sh \
 --training.running_sft_training \
 --model.flavor=8B_flex \
 --model.hf_assets_path="$HOME//torchtitan_assets/Meta-Llama-3.1-8B-Instruct/" \
 --training.dataset_path=$HOME/sft/Dolci-Instruct-SFT/ \
 --sft_data_config.is_multiturn \
 --training.seq_len=2048 \
 --training.local_batch_size=2 \
 ```
image

---
### TO reproduce it on the `multiturn` format dataset [ Dolci-Instruct-SFT](https://huggingface.co/datasets/allenai/Dolci-Instruct-SFT), with `apply_chat_template`
```
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" NGPU=4 ./run_train.sh \
 --training.running_sft_training \
 --model.flavor=8B_flex \
 --model.hf_assets_path="$HOME//torchtitan_assets/Meta-Llama-3.1-8B-Instruct/" \
 --training.dataset_path=$HOME/sft/Dolci-Instruct-SFT/ \
 --sft_data_config.is_multiturn \
 --training.seq_len=2048 \
 --training.local_batch_size=2 \
--sft_data_config.apply_chat_template \
--sft_data_config.ignore_input_ids_mismatch \

 ```
image

---
### To reproduce it on tool datasets e.g, [ReTool-SFT-multi-turn](https://huggingface.co/datasets/swordfaith/ReTool-SFT-multi-turn) dataset
```bash
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" NGPU=4 ./run_train.sh \
 --training.running_sft_training \
 --model.flavor=8B_flex \
 --model.hf_assets_path="$HOME/Meta-Llama-3.1-8B-Instruct/" \
 --training.dataset_path=$HOME/ReTool-SFT-multi-turn/ \
 --sft_data_config.is_multiturn --sft_data_config.apply_chat_template \
 --training.seq_len=2048 \
 --training.local_batch_size=2 \
 --sft_data_config.ignore_input_ids_mismatch \
  --sft_data_config.tools_key=tools \
```
image

---
### To reproduce it on the reasoning dataset e.g., [nvidia/Puzzle-KD-Nemotron-Post-Training-Dataset-v2/](https://huggingface.co/datasets/nvidia/Puzzle-KD-Nemotron-Post-Training-Dataset-v2/)
```
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" NGPU=4 ./run_train.sh \
 --training.running_sft_training \
 --model.flavor=8B_flex \
 --model.hf_assets_path="$HOME/Meta-Llama-3.1-8B-Instruct/" \
 --training.dataset_path=$HOME/Puzzle-KD-Nemotron-Post-Training-Dataset-v2 \
 --sft_data_config.is_multiturn --sft_data_config.apply_chat_template \
 --sft_data_config.split=validation \
 --training.seq_len=2048 \
 --training.local_batch_size=2 \
 --sft_data_config.thinking_key=reasoning \
```
image

torchtitan/hf_datasets/sft_text_datasets.py
```diff
@@ -0,0 +1,559 @@
...
+    # get generate prompt tokens
+    token3 = tokenizer.apply_chat_template(
+        [{"role": "user", "content": ""}], add_generation_prompt=True, tokenize=True
+    )
+    generate_prompt = token3[len(token1) :]
+
+    return system_prompt, generate_prompt
+
+
+def preprocess_data(
```
tianyu-l: Is this function used?

torchtitan/components/tokenizer.py
```diff
@@ -409,6 +432,10 @@ def id_to_token(self, token_id: int) -> Optional[str]:
         """Convert ID to token."""
         return self.tokenizer.id_to_token(token_id)
 
+    def apply_chat_template(self, conversation, **kwargs):
+
+        return self.backup_hf_tokenizer.apply_chat_template(conversation, **kwargs)
```
tianyu-l: I learned what this doing from https://huggingface.co/docs/transformers/en/chat_templating

> (Be aware that when it turns on apply_chat_template, we are supposed to provide the chat_template.jinja in tokenzier's folder. There is no such file for "Meta-Llama-3.1-8B-Instruct". For test purposes, you can use, e.g,. tokenizer from [Olmo-3-7B-Instruct](https://huggingface.co/allenai/Olmo-3-7B-Instruct)

This sounds confusing. If it's supposed to function, why it's missing a file?

torchtitan/config/job_config.py
```diff
@@ -198,11 +198,36 @@ class LRScheduler:
     """
 
 
+@dataclass
+class SFTDataConfig:
```
tianyu-l: add comment for each field?

torchtitan/hf_datasets/sft_text_datasets.py
```diff
@@ -0,0 +1,559 @@
...
+        self.always_thinking = sft_data_config.always_thinking
+
+        self.is_multiturn = sft_data_config.is_multiturn
+        self.pad_mode = sft_data_config.pad_mode
+        self.greedy_packing = sft_data_config.greedy_packing
+        self.ignore_input_ids_mismatch = sft_data_config.ignore_input_ids_mismatch
+
+        if self.pad_mode == "no_padding":
+            self.collate_fn = partial(
+                collate_sft_batch, pad_token_id=self.tokenizer.pad_id
```
tianyu-l: seems this is doing padding either in dataloader, or in collate fn -- why do we need both? I saw that collate_fn always pad from right.

torchtitan/models/attention.py
```diff
@@ -340,3 +362,70 @@ def create_varlen_metadata_for_document(
         max_q=max_seqlen,
         max_k=max_seqlen,
     )
+
+
+def create_varlen_metadata_for_sft(
```
tianyu-l: Do we need both varlen and flex attention? Which one would you prefer? Is it true that for multimodal data where people want to do bidirectional attention within a picture, varlen cannot support it.

torchtitan/hf_datasets/sft_text_datasets.py
```diff
@@ -0,0 +1,559 @@
...
+from datasets import Dataset
+from datasets.distributed import split_dataset_by_node
+from torch.distributed.checkpoint.stateful import Stateful
+from torch.utils.data import IterableDataset
+
+from torchtitan.components.tokenizer import BaseTokenizer
+from torchtitan.config.job_config import SFTDataConfig
+from torchtitan.tools.logging import logger
+
+IGNORE_INDEX = -100
```
tianyu-l: This is also defined at https://github.com/pytorch/torchtitan/blob/c1f4e9491e224acadd69b8e59f8abd9b0aae7145/torchtitan/experiments/vlm/infra/loss.py#L21
Shall we put into a more central place?

torchtitan/components/tokenizer.py
```diff
@@ -59,6 +60,8 @@ def __init__(
         self.eos_id = None
         self.bos_token = None
         self.eos_token = None
+        self.pad_id = None  # only used for SFT
+        self.pad_token = None  # only used for SFT
```
tianyu-l: Let's have a separate (sub)class for SFT tokenizer, and not depending on `transformers`.

See related comments https://github.com/pytorch/torchtitan/pull/1802#discussion_r2423423920

torchtitan/models/attention.py
```diff
@@ -272,6 +272,28 @@ def sliding_window_mod(
     return sliding_window_mod
 
 
+def get_sft_padding_mask_mod(attention_masks: torch.Tensor) -> _mask_mod_signature:
```
tianyu-l: I don't think we want to materialize an attention_masks to general mask_mod. It defeats the purpose of why using mask_mod (to represent mask sparsely).

Instead, we should use e.g. pad_id to infer the mask_mod here, similar to how we generate document mask using eos_id.

Similar for varlen metadata generation.

torchtitan/config/job_config.py
tianyu-l: Due the limitation of current config system, we can't express that "we should either use a (default) Data when doing pretraining, or use the SFTData config when doing finetuning". Right now you'd have to register the SFT one even when doing pretraining.

We are thinking of redo the config system and leaning towards "python as the config". cc @shuhuayu

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -178,21 +178,42 @@ def build_text_dataloader(
...
+        hf_ds = HuggingFaceTextDataset(
+            dataset_name=dataset_name,
+            dataset_path=dataset_path,
+            tokenizer=tokenizer,
+            seq_len=seq_len,
+            dp_rank=dp_rank,
+            dp_world_size=dp_world_size,
+            infinite=infinite,
+        )
+    else:
```
tianyu-l: Let's create another `build_dataloader_fn` (so that one can choose during TrainSpec creation) instead of merging everything together with if-else. And yes, it means we may need to create another experiment folder, which I don't mind we start from. That way you could also define the SFTDataConfig in the folder.

Eventually we should revamp torchtitan config system. cc @shuhuayu

tianyu-l (commented):
Thanks, left some initial comments.

torchtitan/components/tokenizer.py
rakkit: i just aware there is a "chat_template" item in `Llama-3.1-8B-Instruct` HF ckpt. but for some reason, it does not work for me since it is forcing me to read "chat_template.jinja`.  i will double-check this and fix the documents

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: it used in SFTdataset [init](https://github.com/rakkit/torchtitan/blob/staging_sft/torchtitan/hf_datasets/sft_text_datasets.py#L156)  to infer the "system_prompt" and "generation_prompt", the length of these two prompts will be used to remove the repeated [SYSTEM] prompt in multiturn data.
e.g.  if will do `tokenzier(message_i) for message_i in row` we will get
`[SYSTEM …][USER …][SYSTEM …][ASSISTANT …][SYSTEM …][USER …]...`

So we want to remove the redundant `[SYSTEM …]` since the second message.
(condition on `index` of[ `_process_single_message` func](
https://github.com/rakkit/torchtitan/blob/staging_sft/torchtitan/hf_datasets/sft_text_datasets.py#L237))

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: The `collate_fn` pads all sequences in batches to `longest_sequence_in_this_batch`

So there is two options for users (`no_paddign` mode).

OPT-1,  `greedy_packing=False`,   at every step, we might get sequences with different sequence lengths,   (compile dont like this)   . the batch will be like
```
[
   [seq1],
   [seq2 + short padding],
   [seq3+ short padding],
]
```



OPT-2,  `greedy_packing=True`,  we pack multiple "sequence" into one "sequence" and padding the contacted sequence to `max_seq_len` to make sure compile happy,   the batch will be like
```
[
   [seq1,  seq2, seq3, padding],
   [seq4,  seq5, padding],
   [seq6 [truncated]]
]
```
We need `collate_fn` for OPT-1,  and no need it for OPT-2 (but ok to have)


-------
And by contract, we can just do right padding on all sequence to `max_seq_len`
```
[
   [seq1 + lots of padding],
   [seq2 + lots of padding],
   [seq3+ lots ofpadding],
]
```

torchtitan/models/attention.py
rakkit: Both varlen and flex work for SFT. From what I tested, flex is surprisingly faster than varlen (or maybe cuddn attention backedn craps). I think it's good if we can support both variants

rakkit:
thx a lot for the feedback, will fix them later today/tomorrow

torchtitan/models/attention.py
tianyu-l: from what I heard, right now varlen may be using flash v2 so known to be slower than flex. Flash v4 and cudnn backend are on the way (and should be faster)

torchtitan/hf_datasets/sft_text_datasets.py
tianyu-l: OPT-2 greedy packing sounds strictly better. Can we remove OPT-1?

torchtitan/hf_datasets/sft_text_datasets.py
```diff
@@ -0,0 +1,559 @@
...
+):
+    """Normalize dataset rows into prompt/response pairs that respect chat templates."""
+    if apply_chat_template:
+        if output_key:
+            prompt_message = data[input_key]
+            response_message = data[output_key]
+
+            if isinstance(prompt_message, str) and isinstance(response_message, str):
+                prompt_message = [{"role": "user", "content": prompt_message}]
+                response_message = [{"role": "assistant", "content": response_message}]
```
tianyu-l: Maybe noob question for sft:

Strings like "role" / "content" and "user" / "assistant" / "system" seem dataset specific.
1. when we start finetuning a model, do we assume pretrained model knows that they are somewhat special? or do we simply rely on `apply_chat_template` function to map them to special str / token_ids?
2. It seems these strings are specific to this particular data loader -- is this a contract with HF tokenizer's `apply_chat_template` function? If so can we document them clearly?
3. do we ever want to finetune a checkpoint that is already finetuned? say llama3_8b_instruct

torchtitan/hf_datasets/sft_text_datasets.py
tianyu-l: Is this done in `extract_system_prompt_and_generation`? My question is if `preprocess_data` is called anywhere in this PR.

> the length of these two prompts will be used to remove the repeated [SYSTEM] prompt in multiturn data.

why do we want to remove [SYSTEM] since the second message? another way to ask: why do we keep the first?

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: I am also not an experts on SFT these are based on my understandin:
> Strings like `"role"` / `"content"` and `"user"` / `"assistant"` / `"system"` seem dataset-specific.

SFT datasets don’t have a single standard schema. Many modern instruction/chat datasets are represented as a list of messages (`{role, content, ...}`; sometimes also tools / function calls). Other datasets use task-specific columns (e.g., GSM8K-style `question` / `answer`). (both of these two foramts are supported in this PR).

We could optionally provide a “row adapter” / registration function so users can define how to map each dataset row into messages (or prompt/response).

> Do we assume the pretrained model knows these are somewhat special?

In "tradition" nope. But in reality people already mixiing sft and reasoninng data in pretraining data, not surprised to see model knows some of these special tokens. 

> Do we simply rely on `apply_chat_template` to map them to special strings / token IDs?

i think so. and  `apply_chat_template` is an option that the user can turn on or off. 

> It seems these strings are specific to this particular data loader — is this a contract with HF tokenizer’s `apply_chat_template`? If so can we document them clearly?

Not HF' implementation, but these "role" thing depends on `chat_template` file.  What i see from HF's implementation is, one will get an error if these "role" things are not defined in the template.


> Do we ever want to fine-tune a checkpoint that is already fine-tuned (e.g., `llama3_8b_instruct`)?

Yes, depending on the goal. Multi-stage SFT (and sometimes model merging) is common: start from an instruct model and further specialize it on domain/style/tool-use data. Examples:

Really depends on what people want to do. Its common thought people do multi-stages SFT and do model mergining.
e.g., Ollmo3:
image

nemotorn nano 2:
image

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: oh mb, was accidently copied and did not delete it.  no we dont need this `preprocess_data`.

We want to remove [SYSTEM] because of the way we tokenize the data.
The actual conversation is 
`conversation=[USER …][ASSISTANT …][USER …][ASSISTANT …][USER …][ASSISTANT …]` 
but we  cannot do `apply_chat_template(conversation)`. 
instead we need to loop over each`[USER …][ASSISTANT …]` paries. So everytime we get lots of `[SYSTEM]` added unexpected in one conversatoin

[SYSTEM …][USER …][ASSISTANT …][SYSTEM …][USER …][ASSISTANT …][SYSTEM …][USER …][ASSISTANT …]

from second nessage

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: Sounds reasonable.   For whatever reason, people need opt-1; they should just try right-padding.

rakkit:
should we filter out `ids!= IGNORE_INDEX` to get the "correct" `ntokens_seen`?

torchtitan/components/tokenizer.py
rakkit: Confirming that i found `chat_template` for `Llama-3.1-8B-Instruct` its hidden in `tokenizer_config.json` and it works well with HF's tokenizer (`transformers             4.57.3`)

and its does not exist in `Llama-3.1-8B` as expected.  It raise error when we try to `apply_chat_template`

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: moved both to `torchtitan.components.loss`

torchtitan/models/attention.py
rakkit: made the refactor for flex/var len.  Now we manually compute the length of each document segment and indices of the inddics of first token of each segments.  We use them to reconstruct the varlen and flex attention data.

(We cannot use `eos_id` to infer the boundry of documents/sequences because `eos_id` in SFT maybe have differnt meanining, it might refer to the end of a sentence/query,  and it may appear many times espically in muliturn SFT.)

tianyu-l: > We cannot use eos_id to infer the boundry of documents/sequences because eos_id in SFT maybe have differnt meanining, it might refer to the end of a sentence/query, and it may appear many times espically in muliturn SFT.

Then in theory we can do `eod_id` (end of document) when encoding SFT dataset? I somehow don't see if SFT data loading is worth this extra complexity.

cc @lkhphuc if you have feedback.

lkhphuc: If I understand correctly we are talking about the extra `get_document_mask_mod_from_sequence_indices` function, which I agree is a bit overkill.
For both pretraining or SFT, document masking just need a special token to mark boundary of separate sequences. 
I think reusing the `get_document_mask_mod` is fine, since you/user can explicitly passes in whatever token is "end of sequence" token for your particular dataset.

> (We cannot use eos_id to infer the boundry of documents/sequences because eos_id in SFT maybe have differnt meanining, it might refer to the end of a sentence/query, and it may appear many times espically in muliturn SFT.)

I actually never seen this practice because usually people create another token like `<|end_of_turn|>` or `<|im_end|>` or something, but I might be wrong. 
In any case, since you are in charge of the sft dataloader here, nothing prevent us from enforcing `<|eos|>` as the actual document separator and uses different token for end of turn, thus reuse the same document masking function for flex attention.


Unrelated P/s: Sorry @tianyu-l for neglecting the vlm experiment for a while. End of year deadlines is a bit hectic but I will get back to continue it real soon.

lkhphuc:
> should we filter out ids!= IGNORE_INDEX to get the "correct" ntokens_seen?

No IMO. It's "token seens", not "token computed loss on" afterall. We just need to ensure  padding is removed from token seen.

tianyu-l:
> No IMO. It's "token seens", not "token computed loss on" afterall. We just need to ensure padding is removed from token seen.

I agree. "tokens seen" is used to measure pure infra. To track num of tokens trained with loss, it's probably worth its own metric, which doesn't sound urgent.

torchtitan/models/attention.py
rakkit: you mean `<|eos|>` or `eos_token`?  
`eos_token` will not work because some SFT/PT model `eos_token` does not means end of "documents"

`<|eos|>` can work, but we need to ask the user to explicitly provide a segment token, and we need to wrap around tokenizer and `get_document_mask_mod` to get the correct "segment token."


> I actually never seen this practice because usually people create another token like <|end_of_turn|> or <|im_end|> or something, but I might be wrong.

I think that's the argument, if we assume users are well known on SFT so can set it properly, or we wanna make sure the training can be done correctly, even for users who dont know anything.  (I am ok for both)

torchtitan/experiments/sft/test_mask.py
wwwjn, line 1: Should we move this file under `tests/unit_tests`?

torchtitan/models/attention.py
```diff
@@ -215,6 +218,92 @@ def document_mask(
...
+
+    if extra_inputs is None and eos_id is not None:
+        return get_document_mask_mod_using_eos_id(batch, eos_id)
+    elif extra_inputs is not None:
+        seq_lens = extra_inputs.pop("seq_lens", None)
+        seq_start_indices = extra_inputs.pop("seq_start_indices", None)
+        assert seq_lens is not None and seq_start_indices is not None
+        return get_document_mask_mod_using_sequence_indices(
+            batch, seq_lens, seq_start_indices
+        )
```
wwwjn: Error-out if both `eos_id` and `extra_inputs` are empty?

torchtitan/models/attention.py
```diff
@@ -215,6 +218,92 @@ def document_mask(
...
+        eos_id: End-of-sequence token ID that marks document boundaries
+        extra_inputs: Extra inputs to the mask modifier function
+
+    Returns:
+        A mask modifier function that implements document-level masking.
+    """
+    # batch is [b, s, h, d] shape
+
+    # this functioned can be called in two ways:
+    # 1. with eos_id: (we use this in the pre-training while eos_id is relible to get the document boundaries)
```
wwwjn:
```suggestion
    # 1. with eos_id: (we use this in the pre-training while eos_id is reliable to get the document boundaries)
```

torchtitan/models/attention.py
```diff
@@ -215,6 +218,92 @@ def document_mask(
...
+    # 2. with seq_lengths and seq_start_indices:
+    #    - seq_lengths: the length of each sequence
+    #    - seq_start_indices: the start index of each sequence
+    #    - we use this in the post-training while eos_id is not relible to get the document boundaries
```
wwwjn:
```suggestion
    #    - we use this in the post-training while eos_id is not reliable to get the document boundaries
```

torchtitan/models/attention.py
```diff
@@ -215,6 +218,92 @@ def document_mask(
...
+    batch: torch.Tensor,
+    eos_id: int,
+    extra_inputs: dict[str, torch.Tensor] | None = None,
+) -> _mask_mod_signature:
+    """Creates a document mask that prevents attention across document boundaries.
+
+    Args:
+        batch: Input batch tensor with shape [b, s, h, d]
+        eos_id: End-of-sequence token ID that marks document boundaries
+        extra_inputs: Extra inputs to the mask modifier function
```
wwwjn: I'm wondering if SFT needs to be compatible with PP. From the comments here: https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/train.py#L463 The `extra_inputs` are not passed through PP stages, which means for PP rank 1, the extra_inputs will not contain `seq_lens` and `seq_start_indices` , will this function still work? cc @fegin

torchtitan/components/tokenizer.py
wwwjn: I think this `backup_hf_tokenizer` will fragment and causing potential issue? We use torchtitan tokenizer for encode/decode, but relay on `backup_hf_tokenizer` to apply_chat_template

torchtitan/hf_datasets/sft_text_datasets.py
```diff
@@ -0,0 +1,562 @@
...
+
+        if not torch.equal(input_ids, inputs["input_ids"].squeeze(0)):
+            if self.ignore_input_ids_mismatch:
+                logger.warning_once(error_message)
+            else:
+                raise AssertionError(error_message)
+
+    def _process_one_row(self, row_dict: dict):
+        """Convert a dataset row into model-ready tensors with causal labels."""
+        messages = _build_messages(
```
wwwjn: `_build_message` seems to be dataset specific, depending on the dataset's fromat and key names. I would suggest we move current `_build_messages` function out of the `SFTDataset` class to make `SFTDataset` class more general. In torchtitan, we plug-in dataset-specific processor here, see `sample_processor`: https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/hf_datasets/text_datasets.py#L35-L51

wwwjn (commented):
Thanks for the work! Structurally, I would advocate if we start in an experiment folder, so we could highly increase readability for users. We could have something like 
- `experiement/sft`:
    - `tokenizer.py`: Inherit from torchtitan's `HuggingFaceToenizer` , or depend on AutoTokenizer 
    - `datasets.py`
    - `job_config.py`
    - `__init__.py `: use TrainSpec to plug in dataloader, tokenizer here

Currently branching in main trainer / job_config would seems confusing to me

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: yes thats smart. thx

rakkit:
Thanks again for the feedback ❤️.  Sft code now moved to `experiments` folder, also simplified lots of configs.
I will try to edit and improve documents. 

To run test on GSM8K:
```
CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh \
 --model.flavor=debugmodel_varlen_attn \
 --training.dataset question_answer \
 --training.dataset_path=openai/gsm8k \
 --sft_config.dataset_subset=main \
 --model.name=sft.llama3 --job.custom_config_module=torchtitan.experiments.sft.job_config \
```

and to run test on `Dolci-Instruct-SFT` with chat template, (llama3-8b-flex)
```

rm -rf ./outputs/*
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" NGPU=4 ./run_train.sh \
 --model.name=sft.llama3 --job.custom_config_module=torchtitan.experiments.sft.job_config \
 --model.flavor=8B_flex --training.dataset=multi_turn --sft_config.apply_chat_template \
 --training.dataset_path=$HOME/Dolci-Instruct-SFT/ \
  --model.hf_assets_path=$HOME//Meta-Llama-3.1-8B-Instruct/ \
 --training.local_batch_size=8 \
 --activation_checkpoint.mode=full \
 --metrics.log_freq=10 \
 --training.seq_len=2048 --training.steps=100 --lr_scheduler.warmup_steps=10 \
 --debug.seed 10 \
 --compile.enable \

```

torchtitan/config/job_config.py
rakkit: Now it has moved to the experiment folder

torchtitan/hf_datasets/text_datasets.py
rakkit: moved SFT dataset to experiment folder, and wrap the TrainSpec (now only llama3). So SFT will launch `build_sft_text_dataloader`

torchtitan/experiments/sft/test_mask.py
rakkit: thx. I moved all SFT stuff to the `experiment` folder now. (include `test_mask.py`) I will refactor the test function to make it as a real unit test if  we decide to have this new mask mod

torchtitan/hf_datasets/sft_text_datasets.py
rakkit: Now, for each "kind" of SFT dataset, we will pass a `message_builder` to process each row. Three commonly used formats are pre-registered

torchtitan/models/attention.py
acisseJZhong: would this be more clear? I think eos_id should never be none?
```
if extra_inputs is None:
    assert eos_id is not None
    get_document_mask_mod_using_eos_id()
else:
    assert seq_lens is not None and seq_start_indices is not None
    get_document_mask_mod_using_sequence_indices()
```

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,640 @@
...
+        `_build_messages` function. It must return a `messages` list in the same format.
+
+3) [Convert messages to tokens]
+   After we have `messages`, we tokenize them by calling `_process_one_row`.
+
+   3.1) If there is only ONE message, we tokenize it via `_process_single_message`:
+        - Either tokenize the message["content"] directly, OR
+        - Apply the tokenizer's chat template (requires the chat template to be configured).
+
+   3.2) If there are MULTIPLE messages, we call `_process_single_message` for each message.
```
acisseJZhong, line 120: wondering instead of removing system prompt later, would it be possible to add a flag to _process_single_message to control if we want to add the system prompt?

rakkit: We have a flag to control whether we wanna to apply the chat template or not.   
I checked the `apply_chattemple` function from HF, it does not support "render other prompts except system prompt"  if the system prompt was hard coded in the chat template. 

But it brings another question, if the chat template does not automatically plug the system prompt, do we wanna to manually add it or not?

(From what I understand,  "try to remove" system prompt seems to be a safer way to prevent these weird prompts from being plugged in the middle of conversations)

torchtitan/models/attention.py
rakkit: yep I think `eos_id` will never be none

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,46 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from torchtitan.components.tokenizer import BaseTokenizer
+from torchtitan.config import JobConfig
+from transformers import AutoTokenizer as HF_AutoTokenizer
+
+
+class AutoTokenizer(BaseTokenizer):
+    def __init__(self, tokenizer_path: str):
+        self.tokenizer = HF_AutoTokenizer.from_pretrained(tokenizer_path, use_fast=True)
```
tianyu-l: can we not reuse https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/tokenizer.py#L106?

torchtitan/models/attention.py
```diff
@@ -215,6 +218,92 @@ def document_mask(
     return document_mask
 
 
+def get_document_mask_mod_using_sequence_indices(
```
tianyu-l: I wonder if we could avoid the complexity here.

> you mean <|eos|> or eos_token?
eos_token will not work because some SFT/PT model eos_token does not means end of "documents"

> <|eos|> can work, but we need to ask the user to explicitly provide a segment token, and we need to wrap around tokenizer and get_document_mask_mod to get the correct "segment token."
> > I actually never seen this practice because usually people create another token like <|end_of_turn|> or <|im_end|> or something, but I might be wrong.

> I think that's the argument, if we assume users are well known on SFT so can set it properly, or we wanna make sure the training can be done correctly, even for users who dont know anything. (I am ok for both)

Since we/you are in charge of the dataloader (packing behavior) and tokenizer (eod_id), what would we expect user to know?

I'd like to take a step back and try to ask about the purpose of this experiment:
- If the purpose of this experiment is to show an example of SFT training in torchtitan, then we should make it simple enough so that people can easily learn and fork. Do you agree that "making an eod_id per document is nothing different from having eos_id to determine the mask"? Or do you think eos_id is a much wider accepted thing than something like eod_id? I personally think they are the same thing, conceptually; e.g., we don't need to have `get_document_mask_mod_using_eos_id`, we can just require pretraining data loader to also return `seq_lengths` and `seq_start_indices`.
- If the purpose is to provide out-of-box solution for SFT, then it also depends on what dataset people have. Users may not start from HF datasets (or do they?).

I personally think we should start with 1.

I'm also very curious how other libraries are dealing with the complexity in SFT attention mask creation. Would appreciate if you could educate.

torchtitan/models/attention.py
```diff
@@ -340,3 +448,41 @@ def create_varlen_metadata_for_document(
         max_q=max_seqlen,
         max_k=max_seqlen,
     )
+
+
+def create_varlen_metadata_for_document_using_sequence_indices(
```
tianyu-l: I don't see benefit of support varlen attention for SFT right now, so prefer we delay this work to later.

torchtitan/experiments/sft/job_config.py
```diff
@@ -0,0 +1,71 @@
...
+
+from dataclasses import dataclass, field
+from typing import Literal
+
+
+@dataclass
+class SFTConfig:
+    split: str = "train"
+    """Split to use"""
+    dataset_subset: str | None = None
```
tianyu-l, line 15: what is this field?

torchtitan/experiments/sft/job_config.py
```diff
@@ -0,0 +1,71 @@
...
+    """Subset to use"""
+    stream_dataset: bool = True
+    """Whether to stream the dataset"""
+    is_multiturn: bool = False
```
tianyu-l: is this field used?

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,640 @@
...
+    sft_config = job_config.sft_config
+    # TODO: Improving the dataset loading, its easy to fix
+    dataset = load_dataset(
+        dataset_path,
+        sft_config.dataset_subset,
+        split=sft_config.split,
+        streaming=sft_config.stream_dataset,
+    )
+
+    message_builder = DATASET_MESSAGE_BUILDERS[dataset_name]
```
tianyu-l, line 563: the key should be coming from a SFTConfig e.g. `task_type`, not using `dataset_name` which was originally used for dataset registration https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/text_datasets.py#L36

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,640 @@
...
+
+
+DATASET_MESSAGE_BUILDERS = {
+    "multi_turn": _build_multi_turn_messages_from_row_dict,
+    "prompt_response": _build_prompt_response_messages_from_row_dict,
+    "question_answer": partial(
+        _build_prompt_response_messages_from_row_dict,
+        prompt_key="question",
+        response_key="answer",
+    ),
```
tianyu-l, line 69: This separation sounds dataset specific. Without specifying dataset it's hard to tell if they are general.

Maybe consider registering a DATASETS field similar to https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/text_datasets.py#L36
if you have good candidate

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,640 @@
...
+    # return the message, tools, and enable_thinking
+    message = [
+        {"role": "user", "content": row_dict[prompt_key]},
+        {"role": "assistant", "content": row_dict[response_key]},
+    ]
+    return message, None, None
+
+
+DATASET_MESSAGE_BUILDERS = {
+    "multi_turn": _build_multi_turn_messages_from_row_dict,
```
tianyu-l, line 63: similar to this one -- too abstract without seeing an example and hard to verify if it's correct

torchtitan/experiments/sft/llama3/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+    return TrainSpec(
+        model_cls=Transformer,
+        model_args=llama3_args,
+        parallelize_fn=parallelize_llama,
+        pipelining_fn=pipeline_llm,
+        build_optimizers_fn=build_optimizers,
+        build_lr_schedulers_fn=build_lr_schedulers,
+        build_dataloader_fn=build_sft_text_dataloader,
+        build_tokenizer_fn=build_auto_tokenizer,
+        build_loss_fn=build_cross_entropy_loss,
```
tianyu-l: You may want to use https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/vlm/infra/loss.py#L100

otherwise with DP / CP it won't be correct.

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,640 @@
...
+                padded_loss_mask = torch.zeros(
+                    (target_length - sequence_length,), dtype=loss_mask.dtype
+                )
+
+                input_ids = torch.cat((input_ids, padded_input_ids))
+                loss_mask = torch.cat((loss_mask, padded_loss_mask))
+                position_ids = F.pad(
+                    position_ids, (0, target_length - sequence_length), value=0
+                )
+            elif sequence_length > target_length:
```
tianyu-l, line 447: do we have to support the three modes? Is there a single mainstream approach we should take?

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,640 @@
...
+                    loss_mask = loss_mask[:target_length]
+                    position_ids = position_ids[:target_length]
+                elif self.truncation == "error":
+                    raise ValueError(
+                        f"{sequence_length=} is larger than {target_length=}"
+                    )
+                else:
+                    raise ValueError(f"Unknown truncation method {self.truncation}")
+
+        elif self.pad_mode == "greedy_packing":
```
tianyu-l, line 453: similarly, do we have to support both packing and padding?

tianyu-l (changes requested):

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: emm from latest HF's `transformers` I dont see options to turn on/off system prompt in `apply_chattemple`.

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: It is mainly for using HF's AutoTokenizer to make `apply_chattemple` easier.   If we use torchtitan's tokenizer we might need to add `apply_chattemple` first?

torchtitan/experiments/sft/job_config.py
rakkit: Some datasets have multiple subsets. e.g. [openai/gsm8k](https://huggingface.co/datasets/openai/gsm8k) have `main` and `socratic`

torchtitan/experiments/sft/job_config.py
rakkit: removed `is_multiturn`

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: added example training configs  for gsm and allenai/Dolci-Think-SFT-7B dataset

torchtitan/experiments/sft/llama3/__init__.py
rakkit: thx. fixed.  is it for get the correct "token" for normlization?

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: deleted left and only keep right trunc now

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: padding (no greedy pack) makes it easier to understand how many rows (conversations) the training has processed. and easier to control steps/epochs for training.

rakkit:
Thanks a lot for the feedback and discussion.  Did rebase and some refactoring.

1. Major thing, I removed and reverted the varlen and flex attention (mask) related code.  
2. To make sure the current varlen/flex can find the correct document boundary, now we explicitly ask the user to provide the `pad token` and `eos token` (and `pad_id != eos_id`).  This might be the easiest way to do it and reduce code complexity   
3. For now, we compute attention on the full batch, including padding tokens. If we want to optimize this, we can do it in another PR.
4. add two configs, for `question-answer` data sft with gsm8k (no chat-template) and `multiturn` data sft with olmo's sft data (`allenai/Dolci-Think-SFT-7B`, suggest to use chat-template)

to run it  (expect to have Llama-3.1-8B-Instruct's tokenizer, not 3.1-base)
```
CONFIG_FILE="./torchtitan/experiments/sft/train_configs/llama3_8b_gsm8k.toml" NGPU=4 ./run_train.sh  --compile.enable 
```

and  
```
CONFIG_FILE="./torchtitan/experiments/sft/train_configs/llama3_8b_multiturn.toml" NGPU=4 ./run_train.sh --compile.enable 
```

---

And I manually checked around `system_prompt` and `generation_prompt`, here is a short summary of realted code logic.

W/o apply_chat_template,  the sequence is  `[user's question] - [assistant's answer] - [EOS]`.  (for multiturns it will be `[U] - [A] - [EOS] - [U] - [A] - [EOS]`.  Where `[EOS]` should be computed in loss.

When we apply the chat template.  HF's API allows us to encode one message at a time,  so the above message will become:   
`apply_chat_template([user]) -> [SYSTEM] - [Header of User] [ User] [End of User] ` and 
`apply_chat_template([assistant]) -> [SYSTEM] - [Header of assistant] [assistant] [End of assistant] `. Which turns the message into:

`[System] - {user} - [System] - {assistant} - [System] - {User}  - [System] - {assistant} `.   So every message will start with an `[System]`. (and the header/end of their role)

To fix that, we need

1. Remove all the `[System] ` since the second message. (keep the first one)
2. The actual response of assistant starts right after `[Header of assistant]` so we need to maksed out `[Header of assistant]` in loss computation 
3. To make varlen/flex attention work in torchtitan, we need to append an extra `[EOS]` token. From what I understand, we can ignore this one in loss because the assistant should learn to stop in ` [End of assistant].`


After these fixes, the conversation should be 

`[SYSTEM] - [Header of User] [ User] [End of User] -  [Header of assistant] [$assistant $ ] [End of assistant $ ]  - [EOS] `.

Here `$` means what we need to compute in loss, while others should be masked .

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torchtitan.components.tokenizer import BaseTokenizer
+from torchtitan.config import JobConfig
+from torchtitan.tools.logging import logger
+from transformers import AutoTokenizer as HF_AutoTokenizer
```
tianyu-l, line 10: what makes the existing tokenizer.py not enough -- is it because the apply_chat_template is only supported in `transformers`?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/tokenizer.py

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,622 @@
...
+
+   3.1) If there is only ONE message, we tokenize it via `_process_single_message`:
+        - Either tokenize the message["content"] directly, OR
+        - Apply the tokenizer's chat template (requires the chat template to be configured).
+
+   3.2) If there are MULTIPLE messages, we call `_process_single_message` for each message.
+        There is one special case:
+        When `apply_chat_template=True`, each call to `_process_single_message` may add a
+        [system prompt]. That can produce:
+        [System] [User] [System] [Assistant] [System] [User] [System] [Assistant] ...
```
tianyu-l, line 124: I have a dumb question on this

> [System] - {user} - [System] - {assistant} - [System] - {User}  - [System] - {assistant} . So every message will start with an [System]. (and the header/end of their role)

Why it's not good to have repeated `[System]` everywhere? Is the argument that
- Ultimately this trained model will be used for inference / generation.
- In real inference work, we only need to know when the assistant output ends -- we don't need to infer what user say (and should mask them in loss computation), and we don't need to infer the pattern about where `[System]` is showing up.

Is other popular finetune libraries also doing such rearrangement of multi-turn data manually?

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
...
+        assert (
+            pad_token_id != self.tokenizer.eos_token_id
+        ), "PAD token ID is the same as EOS token ID, this can cause problems with varlen/flex attention in dynamic packing."
+
+        self.eos_id = self.tokenizer.eos_token_id
+        self.bos_id = self.tokenizer.bos_token_id
+        self.bos_token = self.tokenizer.bos_token
+        self.eos_token = self.tokenizer.eos_token
+
+        self.pad_id = pad_token_id
```
tianyu-l, line 41: Are we assuming that the model has not been pre-trained with pad_id at all? Because if it has been, we are substitute it with something user given, the model will no longer understand padding -- I guess the argument is it doesn't need to, as padding should always be skipped by model.

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
...
+class HuggingFaceAutoTokenizer(BaseTokenizer):
+    def __init__(
+        self,
+        tokenizer_path: str,
+        eos_token: str,
+        pad_token_id: int,
+        pad_token: str | None = None,
+    ):
+        self.tokenizer = HF_AutoTokenizer.from_pretrained(
+            tokenizer_path, eos_token=eos_token, use_fast=True
```
tianyu-l, line 22: Related to this comment https://github.com/pytorch/torchtitan/pull/2148#discussion_r2626756799

You said that
> some SFT/PT model eos_token does not means end of "documents"

Is it always safe to override the `eos_token`? I had thought that we need a dedicated end-of-document token.

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
...
+
+        self.pad_id = pad_token_id
+
+        self.maybe_original_token_at_pad_id = self.tokenizer._convert_id_to_token(
+            pad_token_id
+        )
+        if pad_token is not None:
+            self.pad_token = pad_token
+        else:
+            self.pad_token = self.maybe_original_token_at_pad_id
```
tianyu-l, line 49: what if the `maybe_original_token_at_pad_id` token is already occupied and its purpose is not padding?

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
...
+        self.bos_token = self.tokenizer.bos_token
+        self.eos_token = self.tokenizer.eos_token
+
+        self.pad_id = pad_token_id
+
+        self.maybe_original_token_at_pad_id = self.tokenizer._convert_id_to_token(
+            pad_token_id
+        )
+        if pad_token is not None:
+            self.pad_token = pad_token
```
tianyu-l, line 47: where are we setting up the link between pad_token_id and pad_id?

torchtitan/experiments/sft/auto_tokenizer.py
acisseJZhong: what if user specify a `pad_token_id` but this id is occupied during pretrain?

torchtitan/experiments/sft/auto_tokenizer.py
acisseJZhong: I though pad_token_id here is equivalent to pad_id?

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
...
+    def apply_chat_template(self, messages: list[dict], **kwargs):
+        return self.tokenizer.apply_chat_template(messages, **kwargs)
+
+    def encode(self, text: str, *args, **kwargs):
+        return self.tokenizer.encode(text, *args, **kwargs)
+
+    def decode(self, tokens: list[int], *args, **kwargs):
+        decoded = self.tokenizer.decode(tokens, *args, **kwargs)
+        # this is an ad-hoc for debugging purpose
+        if self.pad_id in tokens:
```
acisseJZhong, line 65: can you add more comment about what is this for?

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: yep.

torchtitan/experiments/sft/auto_tokenizer.py
```diff
@@ -0,0 +1,90 @@
...
+
+class HuggingFaceAutoTokenizer(BaseTokenizer):
+    def __init__(
+        self,
+        tokenizer_path: str,
+        eos_token: str,
+        pad_token_id: int,
+        pad_token: str | None = None,
+    ):
+        self.tokenizer = HF_AutoTokenizer.from_pretrained(
```
acisseJZhong, line 21: curious why would we need this HuggingFaceAutoTokenizer wrapper? can we use transformers AutoTokenizer directly and pass in our own eos_token/pad_token?

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: you might be right.   we need to know "eos_token" and "eos_id" but not necessarily override HF's tokenizer. i will check the code logic again

torchtitan/experiments/sft/job_config.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    When apply chat template, we will pass these kwargs to the tokenizer.apply_chat_template function.
+    It has to be HF's apply_chat_template kwargs.
+    """
+    ignore_input_ids_mismatch: bool = False
+    """
+    Whether to ignore the input_ids mismatch when apply chat template.
+    """
+
+    eos_token: str | None = None
```
acisseJZhong, line 64: wouldn't the tokenizer own eos_token?

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,622 @@
...
+                logger.warning("Dataset is being re-looped")
+                # Ensures re-looping a dataset loaded from a checkpoint works correctly
+                if not isinstance(self._data, Dataset):
+                    if hasattr(self._data, "set_epoch") and hasattr(
+                        self._data, "epoch"
+                    ):
+                        self._data.set_epoch(self._data.epoch + 1)
+
+
+def build_sft_text_dataloader(
```
acisseJZhong, line 541: nit: can we unify build_sft_text_dataloader and build_sft_validation_dataloader?

torchtitan/experiments/sft/sft_text_datasets.py
```diff
@@ -0,0 +1,622 @@
...
+                    # Yield consistent dict structure immediately
+                    return_dict = {
+                        "input": input_ids,
+                        "positions": positions,
+                    }
+                    yield return_dict, labels
+                    self._sample_idx += 1
+                    continue
+
+                if self._buffer["current_len"] + new_len > self.buffer_max_length:
```
acisseJZhong, line 510: is this greedy packing? maybe put it inside else?

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: yes and it gonna bit of messy.  In short we should try to document everything and let user know what they are doing and we assume user should have some knowledge and can understand what will happen in this SFT receipt after read the document. (i can write documents).

For `pad`, we should "assume" it is not used in pre-training or even 'sft'.  it suppose to be a "meaningless" token and only used for packing.  And we completely drops pad after sft-training.
(if `pad` is occupied, as long as it is not "eos" it should not be problem because 1) its not count in loss 2) we pad the sequence by sequence length and it is independent to choice of `pad`)




For `eos`, this is really complicated and really depends on chat-template.

In general and common setup, we can consider to have conservation like <|start|>assistant xxxxxx <|end|>,  that <|end|> is end of sentence (count in loss).   And in this case, we append a "EOS" that only used for indicating split of document, to make it like:

[system] [user-1] [assisant-1] [user-2] <|start|>assistant xxxxxx <|end|>  "EOS"

where  is not count in loss.   
The option of "EOS" can be either the EOS in pre-training, or any of reserved token that have no meaning, but it cannot be <|end|> and cannot be token can appear in middle of sequence.     The <|end|> token should be configured in chat-template and we dont have to bother. After SFT,  user should set <|end|> as the actual "stop" token in generation, and optionally to have  (which should not affect).

There is always special cases (not sure we want to support them in first place), especially, gpt-oss and harmony format.  They explicitly want  as stop token, that above example should be
[system] [user-1] [assisant-1] [user-2] <|start|>assistant xxxxxx "EOS"    and here  "EOS" should be count in loss.

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: they are same.  `torchtitan` use style like "eos_id", "bos_id".   And HF uses style like "eos_token_id", "pad_token_id".   For now we accept "xxx_token_id" in configs but in actual code we always use "xxx_id".   i can unify all of them to "xxx_id" with torchtitan style

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: see comments above.

as long as pad != eos. and pad exists in vocab we are good

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: this tokenizer will not "overwrite" the actual vocab and actual tokenizer.

the "pad_id" is only a "place-holder" to pad the sequence to max-sequence-length.

when we set the "pad_token" we are only expected to use "pad_token" for debug-purpose.  and as i mentioned in above, ideally we want to uses a "new" "reserved" token for temporary padding purpose.  and once we decode it we will see the sequence be like:

`xxxxxx <RESERVED_1234><RESERVED_1234><RESERVED_1234>`

but for debug we probabely more want to see 

`xxxxxx <PAD> <PAD> <PAD>`

(i will write comment in code as well)

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: i think it is exactly opposite way.  the raw data usually looks like 

```json
[
{ 'role': 'user', 'content': 'xxx',},
{ 'role': 'assistant', 'content': 'xxx',},
{ 'role': 'user', 'content': 'xxx',},
{ 'role': 'assistant', 'content': 'xxx',},
]
```
So HF's tokenizer can actually read and render this entire lists of dictionary in one place and gives 
`[System] - {user} - {assistant} - {user} - {assistant}`.  This is exactly what we want.



For for whatever reason, in `verl` where i followed their conversation.  They decide to process the dict in list one by one. such that we are gonna apply chat-tempalte 4 times for above exmaple and each of them ends up like 

`[system] - {user}`
`[system] - {assistant}`
`[system] - {user}`
`[system] - {assistant}`

so when we ensemble them back to one conversation, we need to remove the extra `[system]` since second.

rakkit: > we don't need to infer the pattern about where [System] is showing up.

ideally yes.  from what `verl` tells fully rely on chat-template might be cause problem ([reference](https://verl.readthedocs.io/en/latest/sglang_multiturn/multiturn.html#special-cases))

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: this wrapper is mainly to avoid overwrite the actual tokenizer.   

and we need this wrapper anyhow as HF and torchtitan have differnt conversation of variable name.

for example, "eos_id" in torchtitan and it is "eos_token_id" in HF

torchtitan/experiments/sft/job_config.py
rakkit: you are right, i forget to fix job_configs.  

we should explicitly ask user to set "eos_token" and "pad_token_id"

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: sure

rakkit:
Thanks for the feedback @acisseJZhong  ! Will rebase and fix the bugs and documents

torchtitan/experiments/sft/auto_tokenizer.py
tianyu-l: If we set pad_id = 6 (something that has to be taken by original model, let's say it corresponds to "the"), what happens?

We would pad with "the" everywhere, and every "the" may be excluded from loss computation, if that's your plan?

torchtitan/experiments/sft/auto_tokenizer.py
tianyu-l: I don't mind migrating to _token_id everywhere, if that's what community convention is.

torchtitan/experiments/sft/auto_tokenizer.py
rakkit: yes we will pad "the" at the end of sequence, and dataloader will return original sequences + lots of "the".

But this in princple does not matter because all padded "the" will not be count in loss.  (the loss mask is computed by length, not by "pad" token

```
                padded_loss_mask = torch.zeros(
                    (target_length - sequence_length,), dtype=loss_mask.dtype
                )
 ```
 
 so it will not cause problem.
 
 
However, in the future if we want to extent flex attention to taking the "pad-id" and masked padding token to save compute then it cause the problem.   (in other words, we waste compute on padded token for now)

File: torchtitan/experiments/sft/sft_text_datasets.py [hunk 39]
```diff
@@ -0,0 +1,622 @@
...
+        """Tokenize one conversation turn while applying template overrides."""
+        apply_chat_template_kwargs = {**self.apply_chat_template_kwargs}
+        if enable_thinking is not None:
+            apply_chat_template_kwargs["enable_thinking"] = enable_thinking
+        if self.apply_chat_template:
+            inputs = self.tokenizer.apply_chat_template(
+                [message],
+                tools=tools,
+                add_generation_prompt=False,
+                tokenize=True,
```
acisseJZhong, line 312: instead of removing system prompt later, can you set this to False except for the first call? cc @tianyu-l

torchtitan/experiments/sft/sft_text_datasets.py
acisseJZhong: https://github.com/pytorch/torchtitan/pull/2148/changes#r2739193009 can you try this?

torchtitan/experiments/sft/auto_tokenizer.py
acisseJZhong: if it's only variable name, what can't we pass in HF's name to titan? 

apologies if I missed anything

rakkit: mainly about variable name.  in practice we can move "eos_token" "pad_token_id" "pad_token" required by this wrapper to datasets's init.     (if we dont overwrite HF tokenizer's vocab, decode cannot be used for debug the "reversed token")

torchtitan/experiments/sft/sft_text_datasets.py
rakkit: you mean if there is an flag in `apply_chat_template` to not add system prompt?  
For that answer is kind "no", the logic of what to do with "message" is mainly coded in the actual chat-template not this `apply_chat_template` functions.   Unless some chat-tempalte have this logic to allow "disable" system prompt.


it seems as long as we process message one by one it is unavoidable to remove system prompt later.

acisseJZhong: ic, makes sense. I somehow misread this `tokenizer` keyword to be whether if it will append system prompt or not.

torchtitan/experiments/sft/sft_text_datasets.py
tianyu-l: > For for whatever reason, in verl where i followed their conversation. They decide to process the dict in list one by one.

So it looks like that we followed verl impl which process the list elements one-by-one, so that we have to remove excess "system".

Are their alternatives than verl -- what's the down side of doing 

> So HF's tokenizer can actually read and render this entire lists of dictionary in one place and gives
[System] - {user} - {assistant} - {user} - {assistant}. This is exactly what we want.

Sorry if this has been answered elsewhere.

joecummings:
Hi @rakkit - I notice this PR has diverged from main and would require a fair amount of clean up to reconcile as well with the new chat template addition #2455.

Do you mind if I get the SFT work in Titan over the finish line? It would be heavily inspired by your work here and I aim to have something out later today or early tomorrow.

rakkit:
@joecummings sure, thx a lot ! looking forward to it !

i have a quick comment, not in this PR but we have tested a lot.

I am not sure if this is well documented or discussed by any one/model.  the thing is for MoE (w/ dpsk style Loss-free-load-balance) SFT,  what should we do about bias update on user's prompts.   We explicitly exclude loss on these tokens, so should we 

1) count all tokens in experts usage and for bias update
2) mask out user prompts (i.e. token with IGNORE_INDEX) in experts usage and only do bias-update on these chat-bot response tokens
3) free bias update.


this matters in the sense of with [3], the moe get extremely imbalanced and lots of experts are sort of died.  and from what we observed [2] can stablize the load balance and works better than [1].   


(another thing but not urgent, should we provide scripts allow user to re-init parts of embedding? especially for these new tokens added in SFT stages which never been trained in pre-training stages) 

CC @ofivite

and you might be interested in @garrett361

garrett361:
> and you might be interested in @garrett361

Thanks for the ping. And yep, just took a quick glance, but this all looks very familiar haha. (My SFT-in-titan impls were less mature I'm sure.)

There's a lot here. Is there anything you want particular comment on @rakkit ?

I know one thing that can cause many headaches is the loss reduction. If you want the same training dynamics across different training setups, you need to be very careful about how loss averages are computed. Very easy to implement this incorrectly so that your loss (assuming you want a mean-loss and not a sum-loss like OLMO uses, say) is not actually the mean loss over the world job, and is instead some incorrect mean-of-means.  

You might have covered this already. Didn't go through in detail.

rakkit:
@garrett361 i think current upstream code already solved the CE loss (with mask, gradient accumulation etc).

i am more refer to how to count "num_tokens_per_expert" in moe-sft,  (e.g. should we ignore masked tokens), which is used for bias update

garrett361:
> i think current upstream code already solved the CE loss (with mask, gradient accumulation etc).

Cool, thought it might, but haven't kept up w/ those changes closely. Figured I'd call it out in case. 

> i am more refer to how to count "num_tokens_per_expert" in moe-sft, (e.g. should we ignore masked tokens), which is used for bias update

Yeah, I have had similar debates (e.g. for z-loss). I suggest leaving it configurable; not sure there's a best universal answer.

tianyu-l:
SFT should be supported now so closing

vadimkantorov:
@tianyu-l Could you please advise, is there anywhere guide/docs for the new SFT support? Thanks!

tianyu-l:
Here's an example https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/config_registry.py#L349

vadimkantorov:
It does support CE masks, right? (to skip training on the prompt prefix)

PR #2149, opened 2025-12-12
[full DTensor] Use all DTensor for Qwen3 and llama4 at TP region

wwwjn:
## Command to run

NGPU=8 MODULE=llama3 CONFIG=llama3_debugmodel ./run_train.sh 

NGPU=8 MODULE=qwen3 CONFIG=qwen3_14b ./run_train.sh

NGPU=8 MODULE=qwen3 CONFIG=qwen3_moe_debug ./run_train.sh

If need to disable SP, add `--parallelism.no-enable-sequence-parallel`

## Test
1. FSDP + TP , no SP:  Run successfully
2. FSDP + TP + SP: Run successfully 
3. FSDP + TP (no SP) + PP: blocked by https://github.com/pytorch/pytorch/issues/172419
5. FSDP + TP + SP + PP: blocked by https://github.com/pytorch/pytorch/issues/172419
6. FSDP + TP + SP + EP: Run successfully
7. FSDP + TP + SP + CP: run successfully 

Numerics check: 
Screenshot 2026-04-22 at 1 48 17 PM


wwwjn:
If this change looks good, I will go ahead and change TP plan of deepseek_v3, llama3, llama4

tianyu-l (commented):
nice, it seems it failed with MoE. Could you try to adjust MoE so that only the boundaries (input and output) are DTensor? We can discuss if you hit issues.

wwwjn:
> nice, it seems it failed with MoE. Could you try to adjust MoE so that only the boundaries (input and output) are DTensor? We can discuss if you hit issues.

Updated! As qwen3 and llama4 share the `apply_moe_ep_tp` function, I updated llama4 TP plan together in this PR

tianyu-l:
test didn't pass

wwwjn:
> test didn't pass

Updated dsv3 as well - Because dsv3 reused llama4 `apply_moe_ep_tp`.

Run loss comparison on dsv3 and qwen3 as well, they both passed.

tianyu-l:
Could you also modify the test_generate script in torchtitan to use sp instead of no-sp?

Btw, I think FlexAttention doesn't work with DTensor yet. Is your PR only enforce DTensor in SP regions (norm) but not TP regions (embedding, attention, mlp)? Specifically, have you tried your change with FlexAttn?

wwwjn:
> Is your PR only enforce DTensor in SP regions (norm) but not TP regions (embedding, attention, mlp)? Specifically, have you tried your change with FlexAttn?

No, it's acutally enforce DTensor in all TP regions. I tested dsv3 with FlexAttention and it works because of the special `attention_kernel_plan`: https://github.com/pytorch/torchtitan/blob/64dc922fc39ed60f1f5997e7866026b6f67b4ec0/torchtitan/models/deepseek_v3/infra/parallelize.py#L220 , which turns DTensor to plain tensor before attention kernel

But I don't think qwen3 / llama4 will work with FlexAttn out-of-box, let me fix, Thanks for catching!

fegin:
Nice, but DSV3 integration test fails. I believe it is because `tok_embeddings` doesn't exist for some stages when PP is enabled. You will need to check if `tok_embeddings` before applying the plan.

torchtitan/models/llama4/parallelize.py
```diff
@@ -247,18 +242,22 @@ def apply_non_moe_tp(
...
+            "attention_norm": SequenceParallel(
+                use_local_output=False,
+            ),
             # NOTE: when the fourth argument (positions) is not None, its input layout
             # and desired input layout should be Replicate()
             "attention": prepare_module_input(
-                input_layouts=(Shard(1), None, None, None),
-                desired_input_layouts=(Replicate(), None, None, None),
+                input_layouts=(Shard(1), Replicate(), None, None),
+                desired_input_layouts=(Replicate(), Replicate(), None, None),
```
fegin, line 279: Nice, now we consistently make freqs_cis as a DTensor. The only one model that still uses plain tensor for freqs_cis is llama3.

wwwjn: Thank you so much! I plan to add change for llama3 as well

wwwjn:
> Nice, but DSV3 integration test fails. I believe it is because `tok_embeddings` doesn't exist for some stages when PP is enabled. You will need to check if `tok_embeddings` before applying the plan.

From the error stack, it's raising error from attention() part , which might not related to `tok_embeddings`? is it related to freqs_cis?  cc @H-Huang 

```
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1830, in inner
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/pytorch/torchtitan/torchtitan/models/deepseek_v3/model/model.py", line 497, in forward
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1882, in _call_impl
      return inner()
             ^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1830, in inner
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/pytorch/torchtitan/torchtitan/models/deepseek_v3/model/model.py", line 384, in forward
      x = x + self.attention(
              ^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1882, in _call_impl
      return inner()
             ^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1819, in inner
      args_result = hook(self, args)
                    ^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/parallel/style.py", line 580, in <lambda>
      lambda _, inputs: self._prepare_input_fn(inputs, device_mesh)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/parallel/style.py", line 552, in _prepare_input_fn
      self._prepare_input_arg(inp, device_mesh, input_layout, desired_layout)
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/parallel/style.py", line 529, in _prepare_input_arg
      dt_inp = dt_inp.redistribute(placements=(desired_layout,))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/_api.py", line 566, in redistribute
      return Redistribute.apply(
             ^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/autograd/function.py", line 583, in apply
      return super().apply(*args, **kwargs)  # type: ignore[misc]
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/_redistribute.py", line 949, in forward
      output = redistribute_local_tensor(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/_redistribute.py", line 840, in redistribute_local_tensor
      new_local_tensor = current_placement._to_replicate_tensor(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/tensor/placement_types.py", line 303, in _to_replicate_tensor
      result = funcol.all_gather_tensor(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/_functional_collectives.py", line 200, in all_gather_tensor
      group_size = c10d._get_group_size_by_name(group_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/torch/distributed/distributed_c10d.py", line 1118, in _get_group_size_by_name
      group = _resolve_process_group(group_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  RuntimeError: Could not resolve the process group registered under the name 18
  
```

fegin:
You are right, `tok_embeddings` is just a warning, not the actual error.

wwwjn:
The error is when TP + PP is applied, both dense model and MoE model has this issue.

When PP rank0 send activation to PP rank1, the activation will be DTensor with this PR. The dtensor's placement was created on PP rank 0, where the TP mesh on PP rank0 has device 0,1. However, when the dtensor was send to PP rank 1, the TP mesh on PP rank 1 has device 2,3. So the underlaying PG will be different. 

cc @H-Huang @wconstab  @sanketpurandare  , we could propose a plan to make PP work with DTensor?

wwwjn:
Blocked by https://github.com/pytorch/pytorch/issues/172419 , @sanketpurandare  is testing the implementation

wconstab:
@wwwjn would it make sense to consider workarounds to unblock this PR rather than waiting for the PP sending DTensor PR to land?

Most of what this PR changes is internals of the transformer block (relating to TP) thought there are some other changes too.  One workaround would be to unwrap/rewrap the DTensor at the boundaries of transformer block, so PP will continue to not see DTensor, but DTensor coverage inside the block is landed.  Question: would it be hard in torchtitan to know the expected dtensor placements at the boundaries of transformer-blocks given the configurable parallelism settings affect this?

after writing this i suspect my suggestion is not useful and maybe we should just block on the PP + Dtensor enablement PR.

wwwjn:
> and maybe we should just block on the PP + Dtensor enablement PR.

Thanks! I'd prefer just leave this PR here and wait PP change to land, as other workstream is not blocked by this PR

wwwjn:
The new improved RFC for DTensor for PP: https://github.com/pytorch/pytorch/issues/172419

Update:
[PR1](https://github.com/pytorch/pytorch/pull/177727), [PR2](https://github.com/pytorch/pytorch/pull/177728), [PR3](https://github.com/pytorch/pytorch/pull/177729)

wwwjn:
@tianyu-l  @fegin @acisseJZhong This PR is ready to review

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -10,6 +10,7 @@
 
 import torch
 import torch.nn as nn
+from torch.distributed._mesh_layout import _MeshLayout
```
acisseJZhong, line 13: do you need to rebase or is this new change?

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -273,9 +272,9 @@ def apply_non_moe_tp(
                         input_layouts=(sp_layout,),
                         desired_input_layouts=(Replicate(),),
                     ),
-                    "feed_forward.w1": colwise_parallel(),
+                    "feed_forward.w1": colwise_parallel(use_local_output=False),
                     "feed_forward.w2": rowwise_output_plan,
```
acisseJZhong, line 272: nit: can we move `rowwise_output_plan` inline? It's easier to reason in this way. Similar for norm_plan, they are both repeated twice so might be fine to define inline?

acisseJZhong:
you might also want to update qwen3 vl https://github.com/pytorch/torchtitan/blob/834c07d78d41fec1ffc760b7a49d0f0b20096af1/torchtitan/models/qwen3_vl/parallelize.py#L72

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -187,15 +184,17 @@ def apply_non_moe_tp(
...
+        use_local_output=False,
     )
 
     parallelize_module(
         model,
         tp_mesh,
         {
             "tok_embeddings": embed_plan,
-            "norm": SequenceParallel() if enable_sp else NoParallel(),
+            "norm": SequenceParallel(use_local_output=False)
```
joecummings, line 192: Isn't `use_local_output=False` by default? https://github.com/pytorch/pytorch/blob/70d99e998b4955e0049d13a98d77ae1b14db1f45/torch/distributed/tensor/parallel/style.py#L374

acisseJZhong: https://docs.pytorch.org/docs/stable/distributed.tensor.parallel.html#torch.distributed.tensor.parallel.RowwiseParallel it's default false for SP but not TP. Let's be explicit to avoid confusions.

torchtitan/distributed/pipeline_parallel.py
wwwjn: This is from @sanketpurandare 's change

torchtitan/models/deepseek_v3/parallelize.py
wwwjn: Yes I explicitly aligned SP with TP behavior using `use_local_output=False`

torchtitan/distributed/pipeline_parallel.py
tianyu-l: @sanketpurandare why do we need this? I'd imagine you could save spmd meshes on each PP rank, and not have to let user provide this callback

wwwjn: `_MeshLayout` is only used in the type annotation of `_build_get_mesh_callback` which is not strictly needed.

But the same question, why we don't put this callback function into PP stage, as PP can access the mesh information on each rank by inspecting DTensor

fegin: @tianyu-l @wwwjn A receiver never has a chance to save SPMD mesh unless the users specify -- the first DTensor a receiver receives is the activation from the remote rank. Unless PP check the model's parameters. But this is also not sound as different parameters can have different SPMD meshes.

torchtitan/models/qwen3_vl/parallelize.py
```diff
@@ -22,6 +22,7 @@
     ColwiseParallel,
     parallelize_module,
     PrepareModuleInput,
+    PrepareModuleInputOutput,
```
shuhuayu, line 25: Why is this necessary?

torchtitan/models/qwen3_vl/parallelize.py
```diff
@@ -55,17 +56,16 @@ def _apply_non_moe_tp_to_decoder(
...
+    connection (``x = x + self.attention(...)``) requires matching types, and
+    DeepStack uses boolean indexing which DTensor does not support. NoParallel
+    on norms wraps plain tensors as DTensor(Replicate) at each layer boundary.
     """
-    # Parallelize embedding, norm, and output — no SequenceParallel
     top_level_plan = {
         "tok_embeddings": RowwiseParallel(
             input_layouts=Replicate(),
             output_layouts=Replicate(),
+            use_local_output=True,  # needed for _scatter
```
tianyu-l: I'm puzzled by the comments above. I probably missed it during earlier review.

@shuhuayu could you review and follow up? Let's remove blockers for full DTensor.

torchtitan/models/llama4/parallelize.py
```diff
@@ -531,7 +530,23 @@ def apply_moe_ep_tp(
     comm_backend: str = "standard",
     enable_sp: bool = True,
     pad_multiple: int | None = None,
+    use_local_output: bool = False,
```
tianyu-l: why do we need to introduce this? Can we just set it to False below

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -44,6 +45,20 @@
 ]
 
 
+def _build_get_mesh_callback(
```
tianyu-l, line 48: add a docstring explaining why we need this function -- because PP requires an spmd mesh during module split so that during runtime the current PP rank could reconstruct a DTensor (non serializable because ProcessGroup is not) after receiving a plain tensor from previous PP rank

torchtitan/models/qwen3_vl/parallelize.py
shuhuayu: There is output tensor `x` from token embedding, and we have some vision embeddings from vision encoder.  we want to replace the placeholders in `x` by the actual vision token embeddings. using scatter operation. But this op is not supported by dtensor, so we used plain tensors here.

I will add dtensor support for this op. We can use plain tensors for now.

torchtitan/models/qwen3_vl/parallelize.py
wwwjn: Removed!

torchtitan/models/qwen3_vl/parallelize.py
wwwjn: Yes here I followed the previous Qwen3-vl behavior. So technically Qwen3-vl is not full DTensor in TP region

torchtitan/models/qwen3_vl/parallelize.py
```diff
@@ -317,6 +311,7 @@ def parallelize_qwen3_vl(
             ep_etp_mesh=parallel_dims.get_optional_mesh(["ep", "etp"]),
             enable_sp=False,
             pad_multiple=find_pad_multiple(model_converters.converters),
+            use_local_output=True,  # Plain tensors for DeepStack boolean indexing
```
wwwjn: Because plain tensor is needed for boolean indexing, so I add a parameter to the shared function `apply_moe_ep_tp()`. However, I updated design to be: Always let MoE module return a DTensor activation,  unwrap DTensor and rewrap DTensor around DeepStack boolean indexing (this might be replace to a local map)

torchtitan/models/llama4/parallelize.py
wwwjn: Replied here: https://github.com/pytorch/torchtitan/pull/2149/changes#r3126236834 

Because of Qwen3_vl model. I updated and remove this field

wwwjn: @shuhuayu  I updated Qwen3-VL's parallel plan:
The idea is there are 2 places we have to use plain tensor instead of full DTensor
1. _scatter_vision_embeds:  This is handled by set `tok_embedding` layer's output, and wrap each transformer layers
2. DeepStack: Use `PrepareModelnputOutput` to turn each transformer layers' output to be a plain tensor,  and it was changed back to DTensor at the beginning of next transformer layer's `PrepareModelnputOutput`
    - This introduce an extra clone, because we implicitly called `to_local()` on the output of Transformer layer. 

Can you help review this change makes sense for you?

torchtitan/models/qwen3/parallelize.py
```diff
@@ -270,9 +272,9 @@ def apply_non_moe_tp(
                         input_layouts=(sp_layout,),
                         desired_input_layouts=(Replicate(),),
                     ),
-                    "feed_forward.w1": colwise_parallel(),
+                    "feed_forward.w1": colwise_parallel(use_local_output=False),
                     "feed_forward.w2": rowwise_output_plan,
```
tianyu-l: Is reusing fine? Not sure if this stateful thing can be reused, please double check. Alternatively may choose to not share and always inline build a new object.

Either way, it will be refactored by fegin soon.

torchtitan/models/qwen3_vl/model.py
```diff
@@ -445,6 +445,16 @@ def _deepstack_process(
                 f"but {num_vision_embeds} vision embeddings."
             )
 
+        # DTensor does not support boolean indexing. Convert to plain
```
tianyu-l: add a TODO for @shuhuayu to use DTensor

tianyu-l (approved):

torchtitan/models/llama4/parallelize.py
shuhuayu: Sorry for the late reply. I do not get why we need `PrepareModelnputOutput` here. The `NoParallel` at the entry of the transformer block can handle a plain tensor as an input.

PR #2154, opened 2025-12-15
[transformers_modeling_backend] Upgrade transformers from 4.57.1 to 5.0.0rc0

3outeille:
This fixes: https://github.com/huggingface/torchtitan/issues/6

Thanks to do that we can do `torch.compile` + 4D-//ism on HF model (cf https://github.com/huggingface/torchtitan/pull/5)


torchtitan/experiments/transformers_modeling_backend/model/model.py
```diff
@@ -270,7 +271,10 @@ def _init_weights_patched(self, module):
                     module.weight.data.normal_(mean=0.0, std=std)
 
                 if module.padding_idx is not None:
-                    module.weight.data[module.padding_idx].zero_()
+                    if isinstance(module.weight.data, DTensor):
+                        module.weight.data.to_local()[module.padding_idx].zero_()
```
tianyu-l: I think `to_local()` will incur a copy. Maybe try
```suggestion
                        module.weight.data._local_tensor[module.padding_idx].zero_()
```

Also wonder if `module.weight.data[module.padding_idx].zero_()` doesn't work for DTensor?

3outeille:
Upgrading to transformers v5 fixes it as it no longer uses `kwargs` for `self.attn` https://github.com/pytorch/torchtitan/pull/2154

wwwjn (approved):
LGTM!

3outeille:
fixed linting

torchtitan/experiments/transformers_modeling_backend/model/model.py
```diff
@@ -270,7 +271,10 @@ def _init_weights_patched(self, module):
                     module.weight.data.normal_(mean=0.0, std=std)
 
                 if module.padding_idx is not None:
-                    module.weight.data[module.padding_idx].zero_()
+                    if isinstance(module.weight.data, DTensor):
+                        module.weight.data._local_tensor[module.padding_idx].zero_()
```
tianyu-l, line 275: sorry I probably didn't what you are doing here.
If the padding is on the "global tensor", we should just do the same thing `module.weight.data[module.padding_idx].zero_()`

The code here is doing local modification, which may or may not be correct depending on if the `padding_idx` is meant to be local or global.

PR #2155, opened 2025-12-15
Add local built pytorch path for pyrefly

fegin:
* #2156
* __->__ #2155

This assumes that the local built version has the same parent folder as torchtitan.

Also fixes some pyrefly errors for moe.py


fegin:
The failing test is not related. It is being fixed by https://github.com/pytorch/torchtitan/pull/2155.

tianyu-l (approved):

PR #2156, opened 2025-12-15
Let CUDA and ROCm read different loss result

fegin:
* __->__ #2156

1. CUDA and ROCm have different loss results. So we need to read from different loss result files.
2. The loss results of FSDP and HSDP start to diverge after 5th step when running with ROCm, we also need to adjust this. But this is more an unknown issue that AMD people may want to figure out the root cause or confirm that this is an expected behavior.

### DIFF
--- a/.github/workflows/integration_test_8gpu_features.yaml
+++ b/.github/workflows/integration_test_8gpu_features.yaml
@@ -76,9 +76,11 @@ jobs:
           LOSS_FILE="tests/assets/losses/llama3_cuda.txt"
           STEPS=10
         elif [[ "${{ matrix.gpu-arch-type }}" == "rocm" ]]; then
-          # The loss of ROCm runs start to diverge between FSDP and HSDP
-          # after 5 steps. Leave fore AMD people to fix this if this is
-          # something bother users.
+          # The loss results of FSDP and HSDP start to diverge after 5th
+          # step when running with ROCm, we also need to adjust this.
+          # But this is more an unknown issue that AMD people may want to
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -70,7 +70,23 @@ jobs:
...
+        # Set architecture-specific parameters
+        if [[ "${{ matrix.gpu-arch-type }}" == "cuda" ]]; then
+          LOSS_FILE="tests/assets/losses/llama3_cuda.txt"
+          STEPS=10
+        elif [[ "${{ matrix.gpu-arch-type }}" == "rocm" ]]; then
+          # The loss of ROCm runs start to diverge between FSDP and HSDP
+          # after 5 steps. Leave fore AMD people to fix this if this is
+          # something bother users.
+          LOSS_FILE="tests/assets/losses/llama3_rocm.txt"
+          STEPS=5
```
tianyu-l, line 85: why setting it to 5 if you are already diverging the loss file?

fegin: It's in the comment of the yaml and the summary of this PR.

The loss results of FSDP and HSDP start to diverge after 5th step when running with ROCm, we also need to adjust this. But this is more an unknown issue that AMD people may want to figure out the root cause or confirm that this is an expected behavior.

tianyu-l: oh I see, I missed that part.

PR #1975, opened 2025-10-31
Add deterministic RL training experiment with vLLM

bwasti:
This experiment provides a complete framework for bitwise-deterministic reinforcement learning training that combines vLLM for fast rollouts and TorchTitan for training with gradients.

Key features:
- Bitwise-deterministic forward and backward passes
- vLLM-compatible Qwen3 model with merged projections
- Custom Flash Attention with gradient support
- Gradient support for vLLM's batch-invariant operations
- Complete RL training loop with GRPO-style advantages
- Comprehensive test suite for determinism verification

Components:
- models/attention.py: VLLMCompatibleFlashAttention
- models/qwen3/model_vllm_compat.py: vLLM-compatible Qwen3 model
- batch_invariant_backward.py: Gradient support for vLLM operations
- simple_rl.py: End-to-end RL training loop
- tests/: Test suite for backward passes and determinism

PR #1660, opened 2025-08-29
Use new DeviceMesh unflatten to rewrite parallel_dims

fegin:
**Summary**
This PR utilizes the latest APIs provided by DeviceMesh to simplify the creation of all different meshes.

The design philosophy is as follow:

1. Create one world mesh with the shape as [world_size,]
2. Create all 1-D submeshes by using 1) unflattening from the world mesh, or 2) slicing and flatten from other derived meshes.
3. ParallelDims now provides an API, get_mesh() and get_optional_mesh(). which accepts str or list[str]. When the argument is str, the API directly return the corresponding 1-D submesh. If the argument is list[str], the dim names will be used to concatenate to form a n-D device mesh. The main difference between the two APIs is that the former one will raise an ValueError if the result mesh is None the later one will just return None.

PR #2158, opened 2025-12-16
Multiprocess simple RL loop

acisseJZhong:
Borrowed part of the implementations from deterministic RL https://github.com/pytorch/torchtitan/pull/1975 by @bwasti, set up simple RL running on multiple process. 

Features added:
- set up train and generator as Actors; potentially using an unified model definition from Torchtitan. 
- integrate with Monarch to run Torchtitan trainer on multiple processes using DDP, and vLLM generator using TP.

Added TODOs here: https://github.com/pytorch/torchtitan/blob/eb601cfc924b379b589af95bffee2a3a56f0a67f/torchtitan/experiments/rl/unified/README.md#todo

Command to run:
```
```

### DIFF
--- a/torchtitan/experiments/rl/unified/README.md
+++ b/torchtitan/experiments/rl/unified/README.md
@@ -82,7 +82,7 @@ Work on the RL loop:
 2. Remove hardcoded configs and dependency on Qwen3 Model. Use torchtitan's config/TrainSpec instead, to work with any model.
 3. Need to load the gsm8k dataset using TorchTitan dataset.
 4. Need to properly implement weight saving and loading using TorchTitan's checkpoint mechanism, or use TorchStore. Also need to
-    replace `vllm_to_torchtitan` and `torchtitan_to_vllm` calls to TorchTitan [state dict adaptor](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/qwen3/model/state_dict_adapter.py).torchtitan/experiments/rl/unified/
+    replace `vllm_to_torchtitan` and `torchtitan_to_vllm` calls to TorchTitan [state dict adaptor](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/qwen3/model/state_dict_adapter.py).
 5. Right now we only support trainer run on multiple processes using DDP, and generator using TP, need to onboard more parallelism.
 6. Right now we only support VLLM_COMPAT mode to achieve batch invariance and bitwise determinism, need to support UNIFIED mode.
 7. In the longer term, need to add trajectory queue to achieve async, right now trainer and generator are running synchronously.


torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -0,0 +1,447 @@
...
+        self.max_new_tokens = max_new_tokens
+        self.temperature = temperature
+        self.use_real_dataset = use_real_dataset
+        self.grpo_beta = grpo_beta
+        self.use_stable_grpo = use_stable_grpo
+        self.tp_size = tp_size
+
+        # Initialize vLLM engine
+        self.vllm_engine = VLLMRolloutEngine(model_path, tp_size=self.tp_size)
+        self.tokenizer = AutoTokenizer.from_pretrained(
```
tianyu-l, line 347: This field is not used in `trivial_reward_function`, pls remove. Also in this work we should try not depend on HF `transformers`

torchtitan/experiments/rl/unified/trainer.py
```diff
@@ -0,0 +1,148 @@
...
+        torch.nn.utils.clip_grad_norm_(self.model.parameters(), max_norm=1.0)
+        self.optimizer.step()
+
+        self.policy_version += 1
+
+        # TODO: save dcp checkpoint to file here instead of sending weight dicts
+
+        # Notify generator of updated weights
+        if self.generator:
+            titan_state = self.model.state_dict()
```
tianyu-l: since everything is titan, can we call it e.g. `trainer_model_states`?

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # Set up batch invariant
+    if batch_invariant:
+        print("Batch invariance detected - using vLLM-compatible model")
+        from torchtitan.experiments.rl.vllm_compat.batch_invariant_backward import (
+            enable_batch_invariant_backward_mode,
+        )
+
+        enable_batch_invariant_backward_mode()
+    else:
```
tianyu-l, line 78: let's error out, instead of silent fall back

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # Download and convert model
+    titan_checkpoint_path, model_path = download_and_convert_model(
+        model_name, cache_dir, output_dir
+    )
+
+    # Load dataset
+    if use_real_dataset:
+        print(f"Loading GSM8K dataset ({num_dataset_samples} samples)...")
+        prompt_texts, expected_answers = load_gsm8k_dataset(
```
tianyu-l, line 90: we should refactor this into a torchtitan dataset, maybe as a TODO

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -0,0 +1,180 @@
...
+        model = Qwen3Model(model_args)
+        # Load standard TorchTitan format directly
+        model.load_state_dict(state_dict, strict=False)
+
+    model.to(torch.bfloat16)
+
+    return model
+
+
+def create_parallel_dims_from_vllm_config(vllm_config: VllmConfig) -> ParallelDims:
```
tianyu-l: there seems already exist such a function in parallelism_utils.py

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -0,0 +1,180 @@
...
+from torchtitan.models.qwen3.model.args import Qwen3ModelArgs
+from transformers import AutoConfig
+from vllm.config import VllmConfig
+
+logger = logging.getLogger(__name__)
+
+
+class ModelMode(str, Enum):
+    UNIFIED = "unified"
+    BATCH_INVARIANT = "batch_invariant"
```
tianyu-l: what's the difference between the two?

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -0,0 +1,180 @@
...
+    torchtitan_to_vllm_compat,
+)
+from torchtitan.models.qwen3.model.args import Qwen3ModelArgs
+from transformers import AutoConfig
+from vllm.config import VllmConfig
+
+logger = logging.getLogger(__name__)
+
+
+class ModelMode(str, Enum):
```
tianyu-l, line 24: please add comments to the values

torchtitan/experiments/rl/unified/models/utils.py
wwwjn: I think this refer to Bram's approch - Creating a vllm-compatible torchtitan model and achieve bit-wise identity between train and inference. Can we rename it to be "vllm-compat" to align with current file structure? Ideally in the future we need to achieve batch-invariant in unified model path as well

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -0,0 +1,180 @@
...
+    UNIFIED = "unified"
```
wwwjn, line 36: I remember you mentioned this simple RL script works with unified model mode, but in the description of PR it says "Integrate with Unified Model, right now it's using batch invariant mode". Does it work with unified model definition?  If not, we should add comments here and clean up UNIFIED branch

torchtitan/experiments/rl/unified/trainer.py
```diff
@@ -0,0 +1,148 @@
...
+        self.tp_size = tp_size
+        self.parallel_dims = create_trainer_parallel_dims(self.ddp_size, self.tp_size)
+
+        # apply PT-D Tensor Parallel
+        # TODO: right now it only works for qwen3 model, need to formalize this to use parallize_fn from train_spec
+        from torchtitan.models.llama3.infra.parallelize import apply_ddp
+
+        apply_ddp(
+            self.model,
+            self.parallel_dims.world_mesh,
```
wwwjn:
```suggestion
            self.parallel_dims.get_mesh("dp_replicate")
```

This part was refactored on main branch recently #1660

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -0,0 +1,180 @@
...
+        # Create and load model that has bitwise determinism between training and inference
+        from torchtitan.experiments.rl.vllm_compat.models.qwen3 import (
+            Qwen3VLLMCompatModel,
+        )
+
+        model = Qwen3VLLMCompatModel(model_args)
+        # Convert to vLLM-compat format (merged gate_up_proj, down_proj)
+        vllm_compat_state = torchtitan_to_vllm_compat(state_dict)
+        model.load_state_dict(vllm_compat_state, strict=False)
+    else:
```
wwwjn, line 137: How does it differs from `model_mode == ModelMode.UNIFIED:`

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -0,0 +1,180 @@
...
+        checkpoint_path: Path to TorchTitan checkpoint
+        model_path: Path to HuggingFace model (for config)
+        model_mode: Indicates which model to use. Train inferece unified model, batch invariant Torchtitan model,
+            or plain Torchtitan model
+
+    Returns:
+        model: Loaded TorchTitan model
+    """
+    # Load HuggingFace config
+    hf_config = AutoConfig.from_pretrained(model_path, trust_remote_code=True)
```
wwwjn, line 91: (Thoughts about how we should refactor this part later, we could avoid using `transformers`)
We don't need the hf_config if we are loading "standard" qwen3 model. By "standard" I mean the model size that already defined in qwen3/__init__.py. 

An alternative way to do this is import qwen_args from [this file](https://github.com/pytorch/torchtitan/blob/1bd2548b14da014b1ec560830f8bdefb6ca568f4/torchtitan/models/qwen3/__init__.py#L32
), initialize a Qwen3ModelArgs() and Qwen3Model. If you'll need to load weights, use the same functions here is enough:  https://github.com/pytorch/torchtitan/blob/1bd2548b14da014b1ec560830f8bdefb6ca568f4/torchtitan/experiments/rl/unified/vllm_wrapper.py#L284-L329

torchtitan/experiments/rl/unified/README.md
```diff
@@ -48,17 +48,19 @@ pip install -e .
...
 4. Run inference:
 ```
-python torchtitan/experiments/rl/unified/infer.py --model torchtitan/experiments/deterministic_vllm_rl/example_checkpoint/qwen3-0.6B
+python torchtitan/experiments/rl/unified/infer.py --model torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B
 ```
 
 Run with TP: (work in progress)
 ```
-python torchtitan/experiments/rl/unified/infer.py --model torchtitan/experiments/deterministic_vllm_rl/example_checkpoint/qwen3-0.6B --tensor-parallel-size 2
+python torchtitan/experiments/rl/unified/infer.py --model torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B --tensor-parallel-size 2
```
wwwjn, line 64: Add commands in README to run simple_rl_multiprocess.py?

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn, line 1: I want to propose the following file structure under `rl` folder for easier understanding:
- Actors/
    - trainer.py
    - generator.py
- simple_rl_multiprocess.py: rl loop entrance for now
- Unified/: one approach to define model
    - infer.py will live here for now, as inference is not our main goal, but a test script
- vllm-compat/: another approach to define model
    - simple_rl.py: keep this under `vllm-compat` to make this folder self-contained for now. We could refactor/ remove it later

torchtitan/experiments/rl/unified/models/utils.py
acisseJZhong: `unified` is using Torchtitan's model definition but only replace with vllm attention(what @wwwjn has implemented in vllm wrapper)

`batch_invariant` is running Bram's vllm compatible model for training, and vllm model for inference, which could achieve bitwise determinism. 

Eventually we need to remove batch_invariant mode and achieve batch invariance under `unified` mode.

torchtitan/experiments/rl/unified/models/utils.py
acisseJZhong: ooops forgot to remove.

torchtitan/experiments/rl/unified/models/utils.py
tianyu-l: unified cannot do training, can it? If so and if it can achieve batch-invariance, we don't have to include the other version.

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: I don't think we should maintain the vllm-compat copy. Your proposal sounds like we should maintain it and make it work with monarch. Any benefit? If no obvious benefit, we should make unified self-contained and eventually remove vllm-compat.

torchtitan/experiments/rl/unified/models/utils.py
acisseJZhong: right now it can't, facing some forward context issue when using vllm.Attention. We should either switch to using [VLLMCompatibleFlashAttention](https://fburl.com/v9ogrzkh) in both trainer and generator, or add backward to vllm.Attention.

After that, we should remove the `batch_invariant` mode.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: sg, then let's make them totally separate and create `actors/` folder under `unified`

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -0,0 +1,447 @@
...
+        # We need to split our weights to match the original 2-shard structure
+        import glob
+        import json
+
+        shard_files = sorted(
+            glob.glob(os.path.join(self.temp_model_dir, "model-*.safetensors"))
+        )
+        index_file = os.path.join(self.temp_model_dir, "model.safetensors.index.json")
+
+        if len(shard_files) == 2 and os.path.exists(index_file):
```
wwwjn, line 144: This part is because HF weights are sharded into 2 safetensor files, but the trainer only write to one, we need to copy to matching the original hf format? We can leave a TODO here, all these copy thing can be removed once we use torchtitan checkpoint saving / loading.

nit: https://huggingface.co/Qwen/Qwen3-0.6B/tree/main this model only have 1 safetensor file, and not need all these complicated logic. And this code is too tight to 2 safetensor files

acisseJZhong: yup this part I directly copied from simple rl, we should also get rid of the `torchtitan_to_vllm` and `torchtitan_to_vllm_compat` which seems confusing to me.

torchtitan/experiments/rl/unified/models/utils.py
acisseJZhong: right now unified model is not working yet as we are using [VLLMAttention](https://github.com/pytorch/torchtitan/blob/ab2f8b7ca67265967f9bb13bb9e05e3876a09df1/torchtitan/experiments/rl/unified/models/attention.py#L11)

torchtitan/experiments/rl/unified/models/utils.py
acisseJZhong: sry i might have not updated the code, we are replacing vllm attention in unified mode.

torchtitan/experiments/rl/vllm_compat/simple_rl.py
```diff
@@ -892,7 +890,7 @@ def rl_update_step(
 
         # Compute rewards using provided reward function
         if reward_fn == trivial_reward_function:
-            rewards = reward_fn(completions, tokenizer, expected_answers, group_size)
+            rewards = reward_fn(completions, expected_answers, group_size)
```
tianyu-l: no need for if else?

torchtitan/experiments/rl/unified/README.md
```diff
@@ -44,25 +44,32 @@ rm -rf build dist *.egg-info
 uv pip uninstall -y vllm
 
 # Rebuild vLLM from source with CUDA 12.4
-pip install -e .
+uv pip install -e .
```
tianyu-l, line 47: are we recommending `uv` for everything now? Does pip alone not working?

acisseJZhong: I am using uv env and pip is not working. I guess pip will still work if you are using conda env. I updated all commands to uv for consistency.

PR #2162, opened 2025-12-18
Add docs to explain COMM_MODE

fegin:
* __->__ #2162

As title

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -115,7 +115,7 @@ NGPU=16 COMM_MODE="local_tensor" ./run_train.sh \
 ### Limitations
 
 - **Performance testing**: Neither mode provides accurate performance metrics; use actual distributed runs for benchmarking
-- **Memory requirement**: Local tensor runs require more memory on a single GPU than the actul distributed runs
+- **Memory requirement**: Local tensor runs require more memory on a single GPU than the actual distributed runs
 
 ## Troubleshooting jobs that timeout
 


docs/debugging.md
```diff
@@ -54,6 +54,69 @@ python -m torchtitan.config.manager --help
...
+
+```bash
+NGPU=32 COMM_MODE="local_tensor" ./run_train.sh
+```
+
+**What it does:**
+- Simulates multi-GPU behavior on a single shared GPU
+- Executes all collectives (all-reduce, all-gather, etc.) locally without network communication
+- Maintains the same code paths as distributed training for accurate debugging
+- Runs only one training step by default
```
tianyu-l, line 99: why having this as default?

docs/debugging.md
```diff
@@ -54,6 +54,69 @@ python -m torchtitan.config.manager --help
...
+
+**When to use it:**
+- Debugging distributed training logic (FSDP, TP, PP, CP, EP) with data dependencies without multi-GPU setup
```
tianyu-l: add a note that fsdp doesn't work today?

tianyu-l (approved):
`debugging.md` is becoming huge. We could either split it into multiple ones in a `docs/debugging/` folder, or create a table of contents in the single file.

docs/debugging.md
fegin: Because under fake or local tensor modes, running more than one step doesn't give any more information. We cannot use either modes to verify memory usages or performance as it is not accurate. So one step should give users enough information.

PR #2161, opened 2025-12-17
Fix pypa/gh-action-pypi-publish version to use SHA pinning

salmanmkc:
## Summary

Fix incorrect version reference for `pypa/gh-action-pypi-publish`.

## Problem

A previous PR incorrectly changed the action reference from `release/v1` (valid branch) to `v1` (non-existent tag). The `v1` tag doesn't exist in the pypa/gh-action-pypi-publish repository.

## Solution

Updated to use SHA pinning for release/v1.13:
```yaml
uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e  # release/v1.13
```

This follows [GitHub's security best practices](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) for third-party actions by pinning to an immutable SHA.

## Files Changed

- `.github/workflows/release.yml`

PR #2163, opened 2025-12-19
Add Dependabot for GitHub Actions updates

salmanmkc:
## Summary

Add Dependabot configuration to automatically keep GitHub Actions up to date.

Here's some more information about Dependabot: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

## Changes

- Added `.github/dependabot.yml` with weekly checks for GitHub Actions updates

## Context

As discussed in #2161 ([comment](https://github.com/pytorch/torchtitan/pull/2161#issuecomment-3667526716)), adding Dependabot to automatically manage GitHub Actions updates going forward.

## Why

Dependabot will automatically create PRs when new versions of GitHub Actions are available, helping to:
- Keep CI/CD workflows secure with the latest patches
- Get new features and improvements
- Maintain compatibility with GitHub's infrastructure

Each action update will be proposed as a separate PR for individual review and testing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"


tianyu-l (commented):
IIUC https://github.com/pytorch/torchtitan/pull/2161 is only updating pypi publish action.

This sounds more general than that.

It sounds cool, but is the PyPI change the only action that needs to be updated right now?

salmanmkc:
> IIUC #2161 is only updating pypi publish action.
> 
> This sounds more general than that.
> 
> It sounds cool, but is the PyPI change the only action that needs to be updated right now?

Just checked, here's one for node 24 actions: https://github.com/pytorch/torchtitan/pull/2164

These two were already merged in before:
https://github.com/pytorch/torchtitan/pull/2151
https://github.com/pytorch/torchtitan/pull/2152

Incorrectly I put the wrong version for one action in one of those PRs, that is fixed in this PR 

So the dependabot thing will do this all for you, but will do individual PRs for each action, but since it will do it every week it shouldn't be too many at once. There's a way to group them together with dependabot as well though.

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
tianyu-l, line 1: what does this version mean? who'd help update this version? lol

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
tianyu-l, line 4: Not sure what this means, can we restrict it to `.github/`?

salmanmkc:
Update: added Dependabot grouping so action updates can be combined into a single PR (group: github-actions, patterns: *).

tianyu-l:
@salmanmkc thanks! could you address my questions before we merge?

.github/dependabot.yml
salmanmkc: As far as I know it's the standard convention, don't know if it works without "/|

salmanmkc:
> @salmanmkc thanks! could you address my questions before we merge?

Yup sorry, went to sleep since it was quite late my time

tianyu-l:
@salmanmkc thanks, how about this one https://github.com/pytorch/torchtitan/pull/2163#discussion_r2633179034

.github/dependabot.yml
salmanmkc: This version I think has been version 2 for many years now, doubt it will change any time soon.

Docs say it must be version 2 https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#about-the-dependabotyml-file

salmanmkc:
> @salmanmkc thanks, how about this one [#2163 (comment)](https://github.com/pytorch/torchtitan/pull/2163#discussion_r2633179034)

yeah just replied

tianyu-l (approved):
sgtm

PR #2165, opened 2025-12-19
Support TP when using vLLM engine to run inference w/ torchtitan model definition

wwwjn:
As tiled.

To support TP, now we keep a seperate TP plan for Qwen3 model. There are 2 main difference with torchtitan core Qwen3 TP plan:
1. Use all DTensor in TP region
2. Add PrepareModuleInputOuput annotation for innner_attention (vllm.Attention())

TODO: Add numerics check

### DIFF
--- a/torchtitan/experiments/rl/unified/models/utils.py
+++ b/torchtitan/experiments/rl/unified/models/utils.py
@@ -12,6 +12,9 @@ import torch
 from safetensors.torch import load_file
 
 from torchtitan.experiments.rl.unified.models.attention import VLLMAttention
+from torchtitan.experiments.rl.vllm_compat.models.attention import (
+    VLLMCompatibleFlashAttention,
+)
 
 from torchtitan.experiments.rl.vllm_compat.weights_vllm_compat import (
     torchtitan_to_vllm_compat,
@@ -103,7 +106,7 @@ def load_model(
     checkpoint_path: str, model_path: str, model_mode: str = ModelMode.VLLM_COMPAT
 ):
     """
-    Load TorchTitan model from checkpoint.
+    Load TorchTitan model from checkpoint for trainer.
 
     Args:
         checkpoint_path: Path to TorchTitan checkpoint
@@ -112,7 +115,7 @@ def load_model(
             or plain Torchtitan model
 
     Returns:
-        model: Loaded TorchTitan model
+        model: Loaded TorchTitan model for trainer.
     """
     # Load HuggingFace config
     # TODO: do not depend on transformers.AutoConfig, use qwen_args directly
@@ -149,7 +152,8 @@ def load_model(
         # Set global default dtype to bfloat16. This is needed because vLLM's Attention
         # layer uses torch.get_default_dtype() and it doesn't support float32
         torch.set_default_dtype(torch.bfloat16)
-
+        # NOTE: Override attention to vllm compatible attention for backward capability.
+        # Only patch to vllm compatible attention for training.
         replace_with_vllm_compatible_flash_attention(model)
 
         # Load standard TorchTitan format directly
--- a/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
@@ -21,8 +21,12 @@ from torch.distributed.checkpoint.state_dict import (
     StateDictOptions,
 )
 
+from torchtitan.experiments.rl.unified.infra.parallelism_utils import (
+    create_job_config_from_vllm_config,
+    create_parallel_dims_from_vllm_config,
+)
+
 from torchtitan.experiments.rl.unified.models.utils import replace_with_vllm_attention
-from torch.distributed.tensor import DTensor, Replicate
 from torchtitan.models.qwen3.model.model import precompute_rope_cache
 from torchtitan.protocols.model import BaseModelArgs, ModelProtocol
 from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
@@ -30,8 +34,6 @@ from torchtitan.protocols.train_spec import ParallelizeFunction
 
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
-
-from .parallelism_utils import create_parallel_dims_from_vllm_config
 
 
 logger = init_logger(__name__)
@@ -87,36 +89,9 @@ class TorchTitanVLLMModelWrapper(nn.Module):
             base=self.config.rope_theta,
         )
 
-        # Replace attention with vLLM's attention
-        replace_with_vllm_attention(self.model)
-
-        # Create ParallelDims from vLLM config and apply parallelization
-        # NOTE: We need to apply parallelize within model.__init__ because w
-        parallel_dims = create_parallel_dims_from_vllm_config(vllm_config)
-        if parallel_dims.tp_enabled:
-            self.world_mesh = parallel_dims.world_mesh
-            tp_mesh = self.world_mesh["tp"]
-            parallelize_fn(
-                model=self.model,
-                tp_mesh=tp_mesh,
-                loss_parallel=False,
-                enable_float8_tensorwise_tp=False,
-                enable_async_tp=False,
-            )
-            logger.info(
-                f"Successfully initialized model with with TP={parallel_dims.tp}"
-            )
-        else:
-            logger.info("Single GPU mode - no parallelization needed")
-
         # Create ParallelDims and JobConfig from vLLM config at runtime
         # vLLM config contains the tensor_parallel_size from command-line args
         # and this will be consistent across all worker processes
-        from torchtitan.experiments.rl.unified.infra.parallelism_utils import (
-            create_job_config_from_vllm_config,
-            create_parallel_dims_from_vllm_config,
-        )
-
         self.parallel_dims = create_parallel_dims_from_vllm_config(vllm_config)
         self.parallel_config = create_job_config_from_vllm_config(
             vllm_config=vllm_config,
@@ -127,6 +102,7 @@ class TorchTitanVLLMModelWrapper(nn.Module):
             assert (
                 model_args.n_heads % tp_size == 0
             ), "Only support when n_heads can be divided by tp_size"
+
         replace_with_vllm_attention(self.model, tp_degree=tp_size)
 
         # NOTE: We need to apply parallelize within model.__init__ because vllm


torchtitan/experiments/rl/unified/utils.py
```diff
@@ -61,3 +62,55 @@ def create_parallel_dims_from_vllm_config(vllm_config: VllmConfig) -> ParallelDi
     )
 
     return parallel_dims
+
+
+def create_job_config_from_vllm_config(
```
tianyu-l, line 98: wait, it should be the other direction? we should construct a vllm config from torchtitan inference config

I guess the possible concern may be: to init a vLLM model one has to pass in a vLLM config -- but in RL you'd specify that from the RL orchestrator, in this case, also torchtitan. So it's even fine if we pass in a fake config and not use it (here you are doing translation back to torchtitan config, which you could've passed in from caller anyways)? What does the vLLM config do, actually?

wwwjn: >  to init a vLLM model one has to pass in a vLLM config

Yes this is the  problem for inference now - vLLM Engine spin up the model, and the only config vllm Engine could pass into the model is vLLMConfig. Currently the inference process is `vLLM Engine` -> `our Model` -> `vLLM Engine` -> `our Model` -> etc.

> So it's even fine if we pass in a fake config and not use it (here you are doing translation back to torchtitan config, which you could've passed in from caller anyways)? 

Do you mean during current inference, we pass a fake torchtitan config, but don't use it. During inference, we use vllm config fields directly to create mesh, apply parallel, etc?  Here I convert it back because the interface of `parallelize_fn` in torchtitan requires a JobConfig object.

> What does the vLLM config do, actually?
 vLLM Engine only pass in VLLMConfig for model initialization. vLLM config is mainly used by vLLM Engine / vLLM model runner . In our model code, we mainly need to know tp_size to apply parallel.

tianyu-l: > In our model code, we mainly need to know tp_size to apply parallel.

If you are the one who constructs the model, you don't have to translate into vllm config to know tp_size.

> vLLM Engine only pass in VLLMConfig for model initialization. vLLM config is mainly used by vLLM Engine / vLLM model runner

That's my question -- how vllm engine is using this config? As we already construct model and apply parallelism by ourselves, can we assume vllm config is useless and we can pass in random garbage and it will get ignored?

If it's for construct processes, then we only need to set the tp_size right.

tianyu-l: > Currently the inference process is vLLM Engine -> our Model -> vLLM Engine -> our Model -> etc.

this sounds unnecessary

wwwjn: > this sounds unnecessary

This interleaved control flow is because vllm used continuous batching, which will call our vllm_wrapper.forward() first, and then slicing and only get last token from each request to compute logprob. I think this behavior can not be fully avoided unless we want to unwrap the current vllm engine code and write our own. 

> how vllm engine is using this config?  As we already construct model and apply parallelism by ourselves, can we assume vllm config is useless and we can pass in random garbage and it will get ignored?

vllm engine internally use vLLMConfig a lot, eg [link](https://github.com/vllm-project/vllm/blob/83a317f650f210b86572b13b8198b7d38aaacb7e/vllm/v1/worker/gpu_model_runner.py#L278-L286). But I agree in model/parallelism part, the only important information from `vllm_config` is parallelism degrees, which we should also be able to get from RL orchestrator. With the RL orchestrator, we can safely assume the `vllm_config` in `TorchTitanVLLMWrapper.__init__(*, vllm_config)` (passed from vllm engine) can be ignored.

A side note is: With RL orchestrator in torchtitan, I think we only need to do one-directional conversion: torchtitan JobConfig -> vllmConfig.

torchtitan/experiments/rl/unified/vllm_wrapper.py
```diff
@@ -276,6 +277,16 @@ def compute_logits(
         sampling_metadata=None,
     ) -> torch.Tensor | None:
         """Compute logits from hidden states."""
+        if self.parallel_dims.tp_enabled:
+            # Turn hidden_states back to DTensor
```
acisseJZhong: is it possible to make this converting to DTensor happen in prepare_module_input for the norm layer? This avoids mixing parallelizing model with the model code itself.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -82,48 +80,49 @@ def __init__(
...
-        Assumes model has .layers dict with .attention.inner_attention structure.
-        Override in subclass if different structure.
+        Replace TorchTitan attention with vLLM paged attention. This function is parallel-aware,
+        initializing vLLM Attention with local num_heads and num_kv_heads.
         """
         assert hasattr(
             self.model, "layers"
         ), f"Model {type(self.model).__name__} must have .layers attribute"
 
+        tp_size = self.parallel_dims.tp
```
acisseJZhong: I feel like we should check this when creating the model, not here?

torchtitan/experiments/rl/unified/utils.py
tianyu-l: by any means we should let torchtitan config be single source of truth. We probably should do:
- torchtitan config -> vllm version of torchtitan model config -> calling titan parallelize functions
- torchtitan config -> vllm config

We should avoid
- torchtitan config -> vllm config -> vllm version of torchtitan model config -> calling titan parallelize functions

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: Accessing tp_size here is not for checking but we are passing it into VLLMAttention initialization because we need local number of heads / num kv heads

torchtitan/experiments/rl/unified/vllm_wrapper.py
wwwjn: That's a great question, technically we could do that using PrepareModuleInput to do this.  The reason I do this manually here is that I want to keep vllm relate changes within the wrapper. I want to keep the TP plan as the same as training TP plan. 

There are 2 places in the wrapper I do the manual conversion between  DTensor <> plain tensor:
1) At the end of forward(): The original output of transformers is Shard(1), and I manually all-gather it to be full tensor. This result will be passed to vllm engine. 
2) At the beginning of `compute_logits()`:  The input is a plain tensor passed from vllm engine, which is a slices from 1). As we did AG in 1), so the tensor here is Replicate()

Without these 2 conversion, the output of transformers will be Shard(1) and be passed into `self.norms` (which is expecting a Shard(1) DTensor as input). If we put 1) and 2) into TP plan, it will cause a uncessary AG if we apply the same plan in training.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
acisseJZhong: sry I meant the check below

torchtitan/experiments/rl/unified/vllm_wrapper.py
acisseJZhong: so the reason why we manually all_gather to get full tensor and later create DTensor is because vllm requires the function to return and takes in plain tensor?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: Oh sure! Let me remove it to __init__

torchtitan/experiments/rl/unified/vllm_wrapper.py
wwwjn: Yes, that's what I learned from llama code and vllm engine code

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -4,6 +4,134 @@
...
+    logger.info(
+        f"Created ParallelDims from vLLM config: "
+        f"DP={parallel_dims.dp_replicate}, TP={parallel_dims.tp}, "
+        f"CP={parallel_dims.cp}, PP={parallel_dims.pp}"
+    )
+
+    return parallel_dims
+
+
+def create_job_config_from_vllm_config(
```
tianyu-l: multiple functions with the same name

torchtitan/experiments/rl/unified/utils.py
```diff
@@ -0,0 +1,130 @@
...
+    logger.info(
+        f"Created ParallelDims from vLLM config: "
+        f"DP={parallel_dims.dp_replicate}, TP={parallel_dims.tp}, "
+        f"CP={parallel_dims.cp}, PP={parallel_dims.pp}"
+    )
+
+    return parallel_dims
+
+
+def create_job_config_from_vllm_config(
```
tianyu-l: one in utils.py

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -145,3 +149,47 @@ def load_model(
     model.to(torch.bfloat16)
 
     return model
+
+
+def create_job_config_from_vllm_config(
```
tianyu-l: the other in models/utils.py

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -18,7 +19,8 @@
 from torchtitan.models.qwen3.model.args import Qwen3ModelArgs
 from transformers import AutoConfig
 
-logger = logging.getLogger(__name__)
+
+logging_logger = logging.getLogger(__name__)
```
tianyu-l: why this name?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -230,6 +229,10 @@ def forward(
         for layer in self.model.layers.values():
             h = layer(h, rope_cache, attention_masks=None, positions=positions)
 
+        # When parallelism is applied, get full tensor before return to vLLM Engine
```
tianyu-l, line 241: comment on
- what is the placement of `h` before converging
- why vLLM needs (Replicate) full tensor

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -243,6 +246,16 @@ def compute_logits(
         sampling_metadata=None,
     ) -> torch.Tensor | None:
         """Compute logits from hidden states."""
+        if self.parallel_dims.tp_enabled:
+            # Turn hidden_states back to DTensor then `norm` can redistribute it into Shard(1)
+            hidden_states = DTensor.from_local(
+                hidden_states,
+                device_mesh=self.parallel_dims.get_mesh("tp"),
+                placements=[
+                    Replicate(),
```
tianyu-l, line 269: comment on what happens in vllm engine, and why it should be Replicate here

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -30,20 +24,19 @@
...
-
 
 logger = init_logger(__name__)
 
 
 class TorchTitanVLLMModelWrapper(nn.Module):
     """
-    Generic vLLM-compatible model wrapper for TorchTitan models.
+    Generic vLLM-compatible model wrapper for TorchTitan models. Implemented
+    required interface required by vLLM Engine.
```
tianyu-l, line 45: If the interface is required by vLLM, then we should inherit from some abstract class / interface?

We should at least link to a document where they specify such protocols.

wwwjn: I didn't find they have official documents / protocols other than: https://docs.vllm.ai/en/latest/contributing/model/basic/#computation-code In this doc, they recommend using vllm's llama implementation as a refers (which is not inherited from any protocol as well) https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/llama.py

tianyu-l: alright, let's link these

torchtitan/experiments/rl/unified/models/utils.py
wwwjn: This is a rebase issue

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
tianyu-l: seems you didn't answer the questions.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
tianyu-l: same -- didn't answer the questions

wwwjn: Sorry my bad, failed to save this file before commit

wwwjn:
CI failing seems unrelated - Nothing related to torchFT change, need to check why torchFT CI is triggered

PR #2166, opened 2025-12-19
Replace `logger.warn()` to `logger.warning()` , allow `log_validation` to log `extra_metrics` and expose common wandb args

EquationWalker:
1. Replace `logger.warn()` to `logger.warning()` 
2. allow `log_validation` to log `extra_metrics`
3. expose common wandb init args, it is userful when resume training.


tianyu-l:
lgtm, please fix linting

PR #2171, opened 2025-12-22
Add Flex flash backend to flex attention module

drisspg:
Stacked PRs:
 * #2728
 * __->__#2171

--- --- ---

# Summary 

### command
```
torchrun --nproc_per_node=1 -m torchtitan.train \
    --job.config-file torchtitan/models/llama3/train_configs/debug_model.toml \
    --model.flavor debugmodel_flex_flash \
    --profiling.enable_profiling \
    --profiling.save_traces_folder ./traces_flash \
    --profiling.profile_freq 5
```

#### Misc things

```Py
"""
Reproduction: 2D tensor indexing in mask_mod causes ~3.5x slowdown with FLASH backend

When using `tensor[b, idx]` inside a mask_mod, FLASH is ~3.5x slower than
equivalent 1D indexing `tensor_flat[b * S + idx]`. Both produce identical results.

To run: python issue_repro_2d_indexing.py
"""

import torch
from torch.nn.attention.flex_attention import flex_attention, create_block_mask

torch._dynamo.config.automatic_dynamic_shapes = False
flex_compiled = torch.compile(flex_attention)
create_block_mask_compiled = torch.compile(create_block_mask)


def benchmark(fn, warmup=3, iters=100):
    for _ in range(warmup):
        fn()
    torch.cuda.synchronize()

    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)
    start.record()
    for _ in range(iters):
        fn()
    end.record()
    torch.cuda.synchronize()
    return start.elapsed_time(end) / iters * 1000  # μs


def main():
    B, S, H, D = 8, 8192, 2, 128

    # Create per-batch document IDs (simulating document masking)
    torch.manual_seed(42)
    sequence_indices = torch.zeros(B, S, dtype=torch.int32, device="cuda")
    for b in range(B):
        num_docs = torch.randint(5, 12, (1,)).item()
        eos_positions = torch.sort(torch.randint(100, S - 100, (num_docs,)))[0]
        doc_id, prev = 0, 0
        for pos in eos_positions:
            sequence_indices[b, prev:pos+1] = doc_id
            doc_id += 1
            prev = pos + 1
        sequence_indices[b, prev:] = doc_id

    print(f"Config: B={B}, S={S}, H={H}, D={D}")
    print(f"Documents per sequence: {[(sequence_indices[b].max().item() + 1) for b in range(B)]}")

    def causal(b, h, q_idx, kv_idx):
        return q_idx >= kv_idx

    # 2D indexing (SLOW on FLASH)
    seq_idx_2d = sequence_indices
    def doc_mask_2d(b, h, q_idx, kv_idx):
        return seq_idx_2d[b, q_idx] == seq_idx_2d[b, kv_idx]
    def combined_2d(b, h, q_idx, kv_idx):
        return causal(b, h, q_idx, kv_idx) & doc_mask_2d(b, h, q_idx, kv_idx)

    # 1D flattened indexing (FAST on FLASH)
    seq_idx_flat = sequence_indices.flatten()
    def doc_mask_1d(b, h, q_idx, kv_idx):
        offset = b * S
        return seq_idx_flat[offset + q_idx] == seq_idx_flat[offset + kv_idx]
    def combined_1d(b, h, q_idx, kv_idx):
        return causal(b, h, q_idx, kv_idx) & doc_mask_1d(b, h, q_idx, kv_idx)

    block_mask_2d = create_block_mask_compiled(
        combined_2d, B=B, H=None, Q_LEN=S, KV_LEN=S, device="cuda", BLOCK_SIZE=(256, 128)
    )
    block_mask_1d = create_block_mask_compiled(
        combined_1d, B=B, H=None, Q_LEN=S, KV_LEN=S, device="cuda", BLOCK_SIZE=(256, 128)
    )

    # Correctness tests
    print(f"\nBlock mask sparsity: {block_mask_2d.sparsity():.2f}%")
    print(f"kv_num_blocks equal: {torch.equal(block_mask_2d.kv_num_blocks, block_mask_1d.kv_num_blocks)}")
    print(f"kv_indices equal: {torch.equal(block_mask_2d.kv_indices, block_mask_1d.kv_indices)}")
    assert torch.equal(block_mask_2d.kv_num_blocks, block_mask_1d.kv_num_blocks)
    assert torch.equal(block_mask_2d.kv_indices, block_mask_1d.kv_indices)

    q = torch.randn(B, S, H, D, device="cuda", dtype=torch.bfloat16).transpose(1, 2)
    k = torch.randn(B, S, H, D, device="cuda", dtype=torch.bfloat16).transpose(1, 2)
    v = torch.randn(B, S, H, D, device="cuda", dtype=torch.bfloat16).transpose(1, 2)

    out_2d = flex_compiled(q, k, v, block_mask=block_mask_2d, kernel_options={"BACKEND": "FLASH"})
    out_1d = flex_compiled(q, k, v, block_mask=block_mask_1d, kernel_options={"BACKEND": "FLASH"})
    print(f"Attention output max diff: {(out_2d - out_1d).abs().max().item():.2e}")
    assert torch.allclose(out_2d, out_1d, atol=1e-3, rtol=1e-3)

    # Benchmark
    print("\nBenchmark results:")
    results = {}
    for name, mask in [("2D [b, idx]", block_mask_2d), ("1D [b*S+idx]", block_mask_1d)]:
        for backend in ["FLASH", "TRITON"]:
            def run():
                return flex_compiled(q, k, v, block_mask=mask, kernel_options={"BACKEND": backend})
            time_us = benchmark(run)
            results[(name, backend)] = time_us
            print(f"  {name:15s} {backend:6s}: {time_us:8.2f} μs")

    flash_2d = results[("2D [b, idx]", "FLASH")]
    flash_1d = results[("1D [b*S+idx]", "FLASH")]
    print(f"\nFLASH slowdown from 2D indexing: {flash_2d / flash_1d:.2f}x")


if __name__ == "__main__":
    main()

```

for forward our 2d indexing code is surprisingly much slower than our 1d case, need to investigate, otherwise we are seein gmuch faster backward at 8192 ~2x for doc masking and equal for fwd. Having to bump the q_len doesnt help here, we are pretty much memory bound for these distributions since the block_mask is basically 80% sparse 

A quick look in ncu shows we are register splilling like crazy


torchtitan/models/attention.py
```diff
@@ -110,6 +118,10 @@ class FlexAttentionWrapper(torch.nn.Module):
         },
     )
 
+    def __init__(self, kernel_options: dict | None = None):
+        super().__init__()
```
drisspg: In theory this can/should be runtime args but for now Im just using for global defaults

torchtitan/models/llama3/__init__.py
```diff
@@ -39,6 +39,17 @@
         attn_type="flex",
         attn_mask_type="block_causal",
     ),
+    "debugmodel_flex_flash": TransformerModelArgs(
+        dim=256,
+        n_layers=6,
+        n_heads=2,  # head_dim=128, required for Flash backend
```
drisspg: 64 is also supported, but deepseek mla is currently not

torchtitan/models/llama3/__init__.py
```diff
@@ -39,6 +39,17 @@
...
+        vocab_size=2048,
+        rope_theta=500000,
+        attn_type="flex",
+        attn_mask_type="block_causal",
+        flex_attention_kernel_options={"BACKEND": "FLASH"},
```
tianyu-l: I have two questions:
1. what gpus can it run on? can it run on H100 / A100?
2. if so, is it the most efficient kernel on all hardware?

Basically, I'd like to understand: as a user, when should I use what.

drisspg: yeah here is my current world model:
On B200 it is the fastest when its supported. The primary things that aren't supported and being worked on 
-> Dynamic shapes (its on my list)
-> Learnable bias ( I dont think this applies for any titan models)
-> Head dim constraints (deepseek mla is not but can/should be)

For hopper it is generally faster but I have been mostly focusing on blackwell so I cant tell you definitively.  And the backward needs to be implemented - I will do this. There is an A100 kernel but I would expect triton to be fine and maybe better since there is much more robust benchmarking.

drisspg: The other main requirement is that users need to have fa4 installed for this to work. it is very easy to install but still.

So in the future I might roll this into `auto` which is the current behavior, we make not promises if we lower to a triton impl or fa4. But I am mostly using this PR for testing right now

Aneureka:
Hi @drisspg, it looks like the slowdown in 2D tensor indexing within mask_mod is primarily due to the fact that we are re-computing the 2D index in a column-major order to make CuTe happy, e.g., (b, s) -> (b + s * B), which can lead to additional significant integer calculation instructions and increase register pressure.

However, CuTe DSL by design accepts hierarchical coordinates (e.g., [`__getitem__`](https://github.com/NVIDIA/cutlass/blob/main/python/CuTeDSL/cutlass/cute/tensor.py#L188) of Tensor), so this re-computation is unnecessary. In fact, when passing the re-computed scalar index, CuTe DSL will need to convert it back to the hierarchical form to compute the offset (see the implementation of [crd2idx](https://github.com/NVIDIA/cutlass/blob/main/include/cute/stride.hpp#L102)).

In [this fix](https://github.com/Aneureka/pytorch/commit/38cd913348103eb31885375e8aff04142bf1c384), I changed `_fixed_indexer_cute` to return the index directly and tweaked `ssa_to_indexable` to accept a sequence of TensorSSA, after which I observed that 2D indexing performance roughly matched (~1.08x slower) 1D indexing. Could you apply the fix and try again? Note that this fix has not been thoroughly tested and may not be robust for all cases.

Thank you!

drisspg:
@Aneureka Thanks again for the tip, I put up: https://github.com/pytorch/pytorch/pull/171222 and good tip on [crd2idx](https://github.com/NVIDIA/cutlass/blob/main/include/cute/stride.hpp?rgh-link-date=2025-12-23T14%3A00%3A05Z#L102) My intuition was that under the hood things would need to be flattened, but I guess I was already thinking strided so this makes sense.

drisspg:
Other perf issue with update: https://github.com/pytorch/pytorch/issues/171267

drisspg:
Now that 4.3.5 is out and we updated it:

fwd:
image

BWD
image



Triton:
fwd
image

image


TODO:
Extra contiguous calls - this is expansion of blocksparse tensors since I needed this to work around a bug in the past, fwd perf is a lil low. Perhaps CLC will help here

torchtitan/models/llama3/__init__.py
```diff
@@ -71,14 +71,44 @@
...
+            ffn_norm=RMSNorm.Config(),
+            feed_forward=FeedForward.Config(
+                hidden_dim=compute_ffn_hidden_dim(256, multiple_of=256),
+            ),
+            attention=GQAttention.Config(
+                n_heads=2,
+                attn_backend="flex",
+                attn_mask_type="block_causal",
+                rope_backend="complex",
+                flex_attention_kernel_options={"BACKEND": "FLASH"},
```
tianyu-l, line 106: when should we use Flash backend?

torchtitan/models/llama3/__init__.py
```diff
@@ -71,14 +71,44 @@
...
+                flex_attention_block_size=(256, 128),
```
tianyu-l, line 107: should this be configured by user at all? what's the guidance?

torchtitan/models/llama3/__init__.py
tianyu-l: should user config anything other than `BACKEND`?

torchtitan/models/common/attention.py
```diff
@@ -543,6 +552,8 @@ class Config(BaseAttention.Config):
         attn_backend: str = "sdpa"
         attn_mask_type: str = "causal"
         rope_backend: str = "complex"  # "complex" or "cos_sin"
+        flex_attention_kernel_options: dict | None = None
+        flex_attention_block_size: int | tuple[int, int] | None = None
```
tianyu-l: This is violating encapsulation. Instead of hardcoding flex related fields in general `GQAttention` module, we should make `inner_attention` configurable.

tianyu-l:
@claude review this PR

torchtitan/models/llama3/__init__.py
drisspg: the currently supported backends,  Literal["AUTO", "TRITON", "FLASH", "TRITON_DECODE"]
For training workloads auto is default and should be used. IN the future we might Allow for Auto to dispatch to FLASH but not yet since it does require changes to block mask sizes on blackwell

torchtitan/models/llama3/__init__.py
drisspg: Yeah it should be, FWIW @liangel-02 will want something like this for the rl stuff. 

In blackwell to use FLASH you will need at least* (256, 128) due to how the kernel works.  But this in general is not a bitwise matching change. So for now making it very explicit

torchtitan/models/common/attention.py
drisspg: Sounds good, in general being able to impl specific options feels desirable

torchtitan/models/llama3/__init__.py
tianyu-l: @drisspg When should we use which? Depending on the answer we can provide the most reasonable UX in titan.

wwwjn:
@drisspg  Add an instruction for users , document when to use which

drisspg:
good question:                                                                                                                
                                                                                                                                                           
  flash attention wheel                                                                                                                                    
  - we dont require users to have flash_attn installed. if you try to use the FLASH backend without it, we error with a clear message telling you whats missing                                                                                                                                              
  - its currently an optional runtime dep. theres a possibility we make it a full runtime dep of pytorch in the future but we're not doing that yet
  - It is however a pure python package that can be installed here: https://pypi.org/project/flash-attn-4/. Currently releasing beta wheels ever wednesday

  when to use FLASH backend                                                                                                                                
  - if you're on Hopper or Blackwell, i'd recommend installing the wheel and using the FLASH backend. its basically always faster for training workloads, which are typically compute bound for almost every mask mod i've tested on both architectures                                                                     
  - that said, its opt-in right now bc the newer code requires newer pytorch versions and is still very actively developed
                                                                                                                                                           
  block sizes                                                                                                                                              
  - on Blackwell with BlockMask, we require larger block sizes (the (256, 128) you see in the config). this is bc the min tile size is dictated by what the flash attention kernel needs, so we're constrained there                                                                                                 
  - being able to configure block sizes is useful regardless of which backend you're on tbh. people often autotune offline to find optimal block sizes and then bake those in so they skip runtime autotuning and get deterministic behavior

tianyu-l:
Thanks @drisspg ! If you don't mind, I'll take over and prepare a PR for this.

PR #2172, opened 2025-12-22
Add rocm support for models, flux & torchft integration tests.

akashveramd:
In this PR, adding rocm support for models, flux & torchft integration tests.
Also, enabled model_only_hf_checkpoint features test for ROCm.


akashveramd:
> It seems model tests were not triggered?

It is getting triggered now and passing for both rocm & cuda.

tianyu-l (approved):
sgtm, torchft related failures are not relevant, and are being fixed in https://github.com/meta-pytorch/torchft/pull/305

PR #2174, opened 2025-12-23
[RL] Support Trainer and Generator Unified Model

acisseJZhong:
Support using `Qwen3TorchTitanForCausalLM` for both trainer and generator. For train, we replace attention with `VLLMCompatibleFlashAttention` since `VLLMAttention` doesn't have backward yet. For inference, we replace attention with `VLLMAttention`. When TP=1, train and inference has bitwise determinism. 

After we verify the speed for using unified model, we can delete the `VLLM_COMPAT` code path.

Command
```
```

### DIFF
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -143,7 +143,7 @@ class VLLMRolloutEngine:
 
         # TODO: need to replace this with Torchtitan's checkpoint save and load
         # right now we hardcoded to work with 2 safe tensor files which we only
-        # tested on Qwen3 1.7B model. In the longer term, need to use TorchStore
+        # tested on Qwen3 0.6B model. In the longer term, need to use TorchStore
         # to achieve the weight communication.
         # only generator rank 0 saves the weight
         if torch.distributed.get_rank() == 0:


torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```diff
@@ -156,11 +163,10 @@ async def main():
 
     for step in range(num_steps):
         # Fully sync RL loop
-        batch = await generator.generate.call_one()
-        metrics = await trainer.step.call(batch)
-        metrics = metrics.item(gpus=0)
-        weights = (await trainer.get_weights.call()).item(gpus=0)
-        await generator.update.call_one(metrics["policy_version"], weights)
+        batch = generator.generate.call().get().item(gpus=0)
```
wwwjn, line 166: What is the differences between `call().get().item(gpus=)` and `call_one()`?

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```diff
@@ -111,6 +111,20 @@ async def main():
...
 
+    # Set up distributed env vars so that titan actors are connected via c10d
+    await setup_env_for_distributed(
+        trainer_mesh,
+        master_addr="localhost",  # TODO: figure out what to set
+        master_port=29500,  # TODO: figure out what to set
+    )
+
+    # Set up distributed env vars so that titan actors are connected via c10d
+    await setup_env_for_distributed(
```
wwwjn, line 122: We have to initialized for each SPMD mesh, right?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -342,8 +356,13 @@ def __init__(
         self.use_stable_grpo = use_stable_grpo
         self.tp_size = tp_size
 
+        # Initialize distributed environment for SPMD generator
+        world_size = dist_utils.init_distributed(
+            Comm(),
+        )
         # Initialize vLLM engine
         self.vllm_engine = VLLMRolloutEngine(model_path, tp_size=self.tp_size)
+        # TODO: do not depend on transformers.AutoTokenizer, use Torchtitan's tokenizer
```
tianyu-l: Is this tokenizer used anywhere?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -141,59 +146,68 @@ def update_weights(self, vllm_compat_state: dict) -> None:
         # right now we hardcoded to work with 2 safe tensor files which we only
         # tested on Qwen3 1.7B model. In the longer term, need to use TorchStore
```
tianyu-l: are you changing to 0.6B?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -141,59 +146,68 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
-        else:
-            # Ensure weights stay in bfloat16
-            vllm_state = {
-                k: v.to(torch.bfloat16) if v.dtype == torch.float32 else v
-                for k, v in vllm_state.items()
-            }
-            # Fallback: save as single file
-            save_file(vllm_state, checkpoint_path)
+        # only generator rank 0 saves the weight
+        if torch.distributed.get_rank() == 0:
```
tianyu-l, line 149: Not sure if I understand this:

The state dict in the input is passed from trainer. If trainer uses any model parallel, e.g. tp, we do need both parts of state dict be passed from trainer?

IIRC our concern of not using TP is because we don't load from file, but here we are anyways loading from file, should we just support TP?

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l, line 1: If we use UNIFIED model, we should deprecate things like `torchtitan_to_vllm` and `torchtitan_to_vllm_compat`

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
acisseJZhong: `call_one()` is used for mesh of size 1, now we make generator SPMD, we need `gen_mesh` to be on 2 gpus for generator TP=2. So I changed it to `call()`

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: it's not used 🤣 let me delete this

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: yes, I am changing to 0.6B as we only tested 0.6B `Qwen3TorchTitanForCausalLM` for vllm

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
acisseJZhong: yes I have to do this otherwise in vllm process I cannot access RANK, will get this error 
```
 ValueError: Error initializing torch.distributed using env:// rendezvous: environment variable RANK expected, but not set
```

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: the `vllm_compat_state` we passed into the `update_weights` function should be full state dict. There's only 1 copy maintained.

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: yup I left a todo here https://fburl.com/ahxa8xjq

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > IIRC our concern of not using TP is because we don't load from file, but here we are anyways loading from file, should we just support TP?

Previously our offline sync to decide not go with TP first, because the trainer can not get full state dict [here](https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/experiments/rl/unified/actors/trainer.py#L84) if we don't refactor the save/load path.

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
wwwjn: > we need gen_mesh to be on 2 gpus for generator TP=2.

Should be DDP=2?

So `call()` will call the generate() function on both DP rank? And we are only getting result from rank 0 by `get().item(gpus=0)`

acisseJZhong: yes, call() will call the function for each process spawned by this mesh, and it will return a value mesh https://meta-pytorch.org/monarch/api/monarch.actor.html#monarch.actor.ValueMesh and I get the value from gpu 0 using `.item(gpus=0)`

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: Then we should change the comment here
```suggestion
        # tested on Qwen3 0.6B model. In the longer term, need to use TorchStore
```

PR #2176, opened 2025-12-24
[rl] Update callsite to init_batch_invariance to pass attention backend.

zhxchen17:
Summary:

In https://github.com/vllm-project/vllm/pull/26315 and https://github.com/vllm-project/vllm/pull/30704,
vllm has deprecated usage of VLLM_ATTENTION_BACKEND, and init_batch_invariance() now takes a
required argument for attention backend. So we're updating the callsites according to the
latest change from vllm.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchtitan/experiments/rl/unified/README.md
+++ b/torchtitan/experiments/rl/unified/README.md
@@ -66,7 +66,7 @@ python torchtitan/experiments/rl/unified/infer.py --model-ckpt-path <path_to_mod
 
 5. Run simple rl loop
 ```
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
 ```
 Right now we only support VLLM_COMPAT mode, which could achieve trainer and generator bitwise identical. We are working on support UNIFIED mode,
 which uses a unified model definition for trainer and generator.
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -23,6 +23,7 @@ from torchtitan.experiments.rl import unified  # noqa: F401
 from torchtitan.experiments.rl.vllm_compat.simple_rl import (
     compute_grpo_advantages,
     compute_grpo_advantages_stable,
+    get_vllm_flash_attention_backend,
     math_reward_function,
     trivial_reward_function,
 )
@@ -210,6 +211,7 @@ class VLLMRolloutEngine:
                 seed=42,  # Fixed seed for determinism
                 enforce_eager=True,
                 tensor_parallel_size=self.tp_size,  # Explicitly single GPU
+                attention_config={"backend": get_vllm_flash_attention_backend()},
             )
             logger.info("Created new vLLM engine")
         else:
--- a/torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+++ b/torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
@@ -14,10 +14,13 @@ This demonstrates:
 The architecture mirrors monarch's grpo_actor.py but adapted for vLLM rollouts + TorchTitan training.
 
 Command to run:
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
 """
 import asyncio
 import logging
+import os
+
+os.environ["VLLM_BATCH_INVARIANT"] = "1"
 
 import torch
 from monarch.actor import this_host
@@ -27,6 +30,7 @@ from torchtitan.experiments.rl.unified.actors.trainer import Trainer
 from torchtitan.experiments.rl.unified.models.utils import ModelMode
 from torchtitan.experiments.rl.vllm_compat.simple_rl import (
     download_and_convert_model,
+    get_vllm_flash_attention_backend,
     load_gsm8k_dataset,
 )
 from vllm.model_executor.layers.batch_invariant import (
@@ -63,7 +67,7 @@ async def main():
     trainer_tp_size = 1
     generator_tp_size = 1
 
-    init_batch_invariance()
+    init_batch_invariance(get_vllm_flash_attention_backend())
     batch_invariant = vllm_is_batch_invariant()
     mode = ModelMode.UNIFIED
 
--- a/torchtitan/experiments/rl/vllm_compat/simple_rl.py
+++ b/torchtitan/experiments/rl/vllm_compat/simple_rl.py
@@ -36,11 +36,17 @@ from torchtitan.experiments.rl.vllm_compat.weights_vllm_compat import (
 
 from torchtitan.models.qwen3.model.args import Qwen3ModelArgs
 from transformers import AutoConfig, AutoTokenizer
-
 from vllm import LLM, SamplingParams
+
+from vllm.attention.backends.registry import AttentionBackendEnum
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 
-init_batch_invariance()
+
+def get_vllm_flash_attention_backend() -> AttentionBackendEnum:
+    return AttentionBackendEnum.FLASH_ATTN
+
+
+init_batch_invariance(get_vllm_flash_attention_backend())
 
 
 class VLLMRolloutEngine:


torchtitan/experiments/rl/vllm_compat/simple_rl.py
```diff
@@ -37,10 +37,18 @@
...
+from vllm.attention.backends.registry import AttentionBackendEnum
 from vllm import LLM, SamplingParams
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 
-init_batch_invariance()
+
+def get_vllm_attention_backend() -> AttentionBackendEnum:
+    if os.getenv("VLLM_ATTENTION_BACKEND") is None:
+        raise RuntimeError("VLLM_ATTENTION_BACKEND is not set.")
+    return getattr(AttentionBackendEnum, os.getenv("VLLM_ATTENTION_BACKEND"))
```
tianyu-l: https://github.com/vllm-project/vllm/pull/30704 says
> The VLLM_ATTENTION_BACKEND environment variable has been deprecated by https://github.com/vllm-project/vllm/pull/26315. This PR updates the batch invariant initialization accordingly.

Why we still use this envvar?

Personally I think we should avoid envvar as much as possible, so I'd prefer if we don't have to call
```
```

acisseJZhong: wondering what error are you seeing if you don't set `VLLM_ATTENTION_BACKEND=FLASH_ATTN` and don't pass anything into `init_batch_invariance`?

zhxchen17: @acisseJZhong the error we saw with init_batch_invariance():
```
  File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/unified/simple_rl_multiprocess.py", line 25, in <module>
    from torchtitan.experiments.rl.unified.actors.generator import Generator
  File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/unified/actors/generator.py", line 18, in <module>
    from torchtitan.experiments.rl.vllm_compat.simple_rl import (
  File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/vllm_compat/simple_rl.py", line 43, in <module>
    init_batch_invariance()
TypeError: init_batch_invariance() missing 1 required positional argument: 'attention_backend'
```

and the error when init_batch_invariance(None):
```
Traceback (most recent call last):
  File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/unified/simple_rl_multiprocess.py", line 25, in <module>
    from torchtitan.experiments.rl.unified.actors.generator import Generator
  File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/unified/actors/generator.py", line 18, in <module>
    from torchtitan.experiments.rl.vllm_compat.simple_rl import (
  File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/vllm_compat/simple_rl.py", line 43, in <module>
    init_batch_invariance(None)
  File "/data/users/zhxchen17/vllm/vllm/model_executor/layers/batch_invariant.py", line 1057, in init_batch_invariance
    override_envs_for_invariance(attention_backend)
  File "/data/users/zhxchen17/vllm/vllm/model_executor/layers/batch_invariant.py", line 1025, in override_envs_for_invariance
    raise RuntimeError(error)
RuntimeError: VLLM batch_invariant mode requires an attention backend in ['FLASH_ATTN', 'FLASHINFER', 'FLASH_ATTN_MLA', 'TRITON_MLA'], but got 'None'. Please use --attention-backend or attention_config to set one of the supported backends before enabling batch_invariant.
```

zhxchen17: @tianyu-l Right now other than envvar there seems no better way inject configs to simple_rl_multiprocess.py. 
Other 3 options without using envvar:
1. Hardcode FLASH_ATTN backend for now in the main script and make it clear it's hardcoded.
2. Make simple_rl_multiprocess.py read some config files which contains the attention backend name. 
3. Make simple_rl_multiprocess.py accept cmd line arg, but this seems not consistent with the main trainer script, so not considering it atm.

Seems 1. is better for now if we stick with flash attention for a while.

torchtitan/experiments/rl/unified/README.md
```diff
@@ -66,7 +66,7 @@ python torchtitan/experiments/rl/unified/infer.py --model torchtitan/experiments
 
 5. Run simple rl loop
 ```
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```
tianyu-l, line 69: maybe also modify https://github.com/pytorch/torchtitan/blob/30ab580cd6922173becb5b53f7dbe30afe636162/torchtitan/experiments/rl/unified/simple_rl_multiprocess.py#L17

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```diff
@@ -14,10 +14,13 @@
...
 
 Command to run:
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
 """
 import asyncio
 import logging
+import os
+
+os.environ["VLLM_BATCH_INVARIANT"] = "1"
```
tianyu-l, line 23: Does this always take effect? If so I think it's ok to put it here.

IIUC this simultaneously takes effect for both trainer and generator attention module?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -197,6 +198,7 @@ def update_weights(self, vllm_compat_state: dict) -> None:
                 seed=42,  # Fixed seed for determinism
                 enforce_eager=True,
                 tensor_parallel_size=self.tp_size,  # Explicitly single GPU
+                attention_config={"backend": get_vllm_flash_attention_backend()},
```
tianyu-l, line 214: curious: if we already do `init_batch_invariance(get_vllm_flash_attention_backend())` in simple_rl, do we still have to separately config it for generator?

torchtitan/experiments/rl/vllm_compat/simple_rl.py
wwwjn: I personally want to persue path 2 but it needs more refactor. I think we can go env var for now and refactor the config system in another PR for better seperation. The config flow I think should be:
1. Initialize a torchtitan config using `ConfigManager()`, read config from toml file. (where we put the attention backend)
2. We could have `ConfigAdapter` to translate into VllmConfig object.

torchtitan/experiments/rl/unified/actors/generator.py
zhxchen17: I think currently the reason we're even using init_batch_invariance() is orthogonal: we need to call it explicitly because without calling it our op patching in `enable_batch_invariant_backward_mode()` will fail. 

Normally it will be called automatically inside LLM generator process, so it's not needed if we don't patch `vllm.model_executor.layers.batch_invariant._batch_invariant_LIB` inside `enable_batch_invariant_backward_mode`.

torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
zhxchen17: Yes, I think we're using Monarch which should propagate this to the generator process, so we have this envvar to be the same across trainer and generator.

torchtitan/experiments/rl/unified/actors/generator.py
zhxchen17: btw init_batch_invariance() doesn't control the backend in vllm directly, so we still need to set it separately here.

tianyu-l: Got it.

Actually I'm not sure why we need to switch mode in trainer. In this unified effort, we are using torchtitan's own model definition (no vllm definition other than attention). However, seems this `enable_batch_invariant_backward_mode` is simply trying to register backward for vllm ops such as matmul and linear.

```
    _batch_invariant_backward_lib.impl(
        "aten::matmul_backward", matmul_backward_impl, "CUDA"
    )
    _batch_invariant_backward_lib.impl(
        "aten::linear_backward", linear_backward_impl, "CUDA"
    )
```

Maybe we don't need to call `enable_batch_invariant_backward_mode` in this unified effort?

Moreover, I wonder if we don't have to call `init_batch_invariance()` whatsoever? Maybe for vllm attention in trainer we still need it?

wwwjn: Aha, nice catch! I think by calling `enable_batch_invariant_backward_mode()` here, it is not patching rms norm kernels and linear kernels to be batch-invarient. We achieved bit-wise identity between training and inference only because we used the unified model.  So ideally we could remove `enable_batch_invariant_backward_mode()` in simple_rl_multiprocess.py.

Attention between training and inference are not aligned yet (2 separate code path now):
1. Inference is using `vllm.Attention()`. We are not specifying  attention backend explicitly.
2. Training is using VLLMCompatibleFlashAttention() which is calling `flash_attn_varlen_func` from VLLM. 

So my understanding is : here we passed `attention_backend` to explicitly tell inference to use FlashAttention, although `vllm.Attention` use FlashAttention by default.  But there are 2 source-of-truth of attention backend: 1)get_vllm_flash_attention_backend and 2) `VLLMCompatibleFlashAttention`

tianyu-l: > here we passed attention_backend to explicitly tell inference to use FlashAttention, although vllm.Attention use FlashAttention by default. But there are 2 source-of-truth from 1)get_vllm_flash_attention_backend and 2) VLLMCompatibleFlashAttention

OK there are two separate questions:
1. what attention backend to use
2. batch invariant mode or not

For 1, I guess even if we don't do anything here (calling `init_batch_invariance`, specifying backend explicitly), we'll default to flash backend for both inference (`vllm.Attention`) and training (`VLLMCompatibleFlashAttention `). Since we explicitly control training anyways, making it explicit for inference here sounds not bad?

For 2, I'm a bit confused. Is it true that setting `os.environ["VLLM_BATCH_INVARIANT"] = "1"` is enough for both trainer and generator per this comment? https://github.com/pytorch/torchtitan/pull/2176#discussion_r2651565584
If so why do we still need to  call `init_batch_invariance` at all??

tianyu-l: any updates to question 2?

PR #2177, opened 2025-12-26
Fix bug in Qwen3 MOE expert ordering

daniellepintz:
When we sort the `experts.keys()` in order to stack the experts, we are currently sorting them lexicographically as they are strings. We need to pass `key=int` to `sorted()` to sort them as ints.

This causes a bug where the experts are stored in the wrong order when loading the model in TorchTitan.

### DIFF
--- a/torchtitan/models/utils.py
+++ b/torchtitan/models/utils.py
@@ -287,8 +287,8 @@ class MoEStateDictAdapter(StateDictAdapter):
             return None
 
         print("Printing type of experts.keys():")
-        print(type(experts.keys()))
-        sorted_expert_ids = sorted(experts.keys(), key=int)
+        print(type(list(experts.keys())[0]))
+        sorted_expert_ids: list[int] = sorted(experts.keys(), key=int)
         print("Applied sorting fix")
         sorted_experts = [experts[i] for i in sorted_expert_ids]
         # pyrefly: ignore [missing-attribute]


torchtitan/models/utils.py
```diff
@@ -286,7 +286,10 @@ def _concatenate_expert_weights_dtensor(
         if len(experts) < expected_n_experts:
             return None
 
-        sorted_expert_ids = sorted(experts.keys())
+        print("Printing type of experts.keys():")
```
daniellepintz, line 289: will remove the prints before merging

wwwjn:
Hi @daniellepintz Can we do something like  https://github.com/pytorch/torchtitan/blob/a95d20385e254c4905bed8c40bf11d1f0a69f145/torchtitan/models/deepseek_v3/model/state_dict_adapter.py#L178-L180 to make it more clear? and we could also add assert before the sort() at here https://github.com/pytorch/torchtitan/blob/a95d20385e254c4905bed8c40bf11d1f0a69f145/torchtitan/models/utils.py#L289 to double check

daniellepintz:
Closing this PR as this was fixed in https://github.com/pytorch/torchtitan/commit/fb549717a78362ad29ebc5e89bc0b4950cdfc04c#diff-fc1695b061ff6977615a8999542435d67833f16dddced3c7d078afa7888a4831

Issue #2170, opened 2025-12-22
Attention scaling is never used


francesco-bertolotti:
In the Qwen3 model implementation, attention scaling is defined but never actually applied.

Specifically, in the following code, the scaling factor is mentioned but not used:
https://github.com/pytorch/torchtitan/blob/1bd2548b14da014b1ec560830f8bdefb6ca568f4/torchtitan/models/qwen3/model/model.py#L176

I suspect this may be a byproduct of using the PyTorch 2.1+ Flash Attention backend, where scaling is handled internally.

Is this behavior expected? If so, are there any recommended workarounds or configuration options to explicitly control attention scaling?

NB: Deepseek implementation does not suffer from this but it does not support the flash attention backend

tianyu-l:
cc @wwwjn

wwwjn:
Thanks for asking and nice catch! The scaling was introduced in PR: https://github.com/pytorch/torchtitan/pull/1704/changes#diff-b325355aeabc98b656c761ee8faef22c0022c7cb6e11afaed482cf4231d9feccL212 and we were applying it to SDPA attention.  In later refactors, this field is dropped.

For sdpa, we need to fix and passing scaling during forward - I will fix this and create a PR

For varlen attention / Flash attention - cc @drisspg  Do you know how we handle scale?

For FlexAttention, we need to fix and passing scale here: https://github.com/pytorch/torchtitan/blob/1bd2548b14da014b1ec560830f8bdefb6ca568f4/torchtitan/models/attention.py#L135

drisspg:
@liangel-02 can you add softmax scaling to varlen_api in pt and then make the update here

wwwjn:
https://github.com/pytorch/torchtitan/pull/2173 Here's for Flex and sdpa attention

PR #2178, opened 2025-12-26
add attention scaling to varlen for qwen3

liangel-02:
as title, fixes [#2170](https://github.com/pytorch/torchtitan/issues/2170)

### DIFF
--- a/torchtitan/models/attention.py
+++ b/torchtitan/models/attention.py
@@ -60,6 +60,7 @@ class VarlenAttentionWrapper(torch.nn.Module):
         xv: torch.Tensor,
         head_dim: torch.Tensor,
         attention_masks: VarlenMetadata,
+        scale: float | None = None,
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
         cu_seq_q = attention_masks.cu_seq_q
         cu_seq_k = attention_masks.cu_seq_k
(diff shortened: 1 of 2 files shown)


wwwjn:
Do we support varlen attention for deepseek_v3 / llama4 model? If not , do we want to explicitly fail when they apply varlen attention? 

I was asking because deepseek_v3 model also need a scale here: https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/deepseek_v3/model/model.py#L308, but I see it haven't supported Varlen Attention yet

tianyu-l:
also llama3? https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/model/model.py#L284

liangel-02:
> Do we support varlen attention for deepseek_v3 / llama4 model? If not , do we want to explicitly fail when they apply varlen attention?
> 
> I was asking because deepseek_v3 model also need a scale here: https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/deepseek_v3/model/model.py#L308, but I see it haven't supported Varlen Attention yet

@wwwjn we don't support varlen attn for those models, i can add a check in a separate pr

update: @wwwjn see #2179

liangel-02:
> integration test (feature) failed, is it because it's not in pytorch yet?

yes, i have https://github.com/pytorch/pytorch/pull/171199 to add to pytorch, we can land that first

PR #2179, opened 2025-12-26
add safety checks for varlen 

liangel-02:
since varlen attn is not supported for deepseek v3 and llama4, we should raise an error if someone attempts to use it

### DIFF
--- a/torchtitan/models/deepseek_v3/model/model.py
+++ b/torchtitan/models/deepseek_v3/model/model.py
@@ -237,9 +237,13 @@ class Attention(nn.Module):
         match self.attn_type:
             case "flex":
                 self.inner_attention = FlexAttentionWrapper()
-            case _:
+            case "sdpa":
                 # pyrefly: ignore [bad-assignment]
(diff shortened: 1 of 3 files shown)


wwwjn:
Thank you! Can we also change llama3 to align with other models: https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/model/model.py#L228

jquesnelle:
@wwwjn what's the barrier to varlen support for DSv3? from a cursory glance using the existing `VarlenAttentionWrapper` as in Qwen3 ought to work?

liangel-02:
@jquesnelle the underlying flash attention kernel expects Q, K, V to have the same head dimensions 
https://github.com/Dao-AILab/flash-attention/blob/main/csrc/flash_attn/flash_api.cpp#L611-L620 but DSv3 uses different dims for Q/K and V https://github.com/pytorch/torchtitan/blob/d12ff2fc1a5ad12d2663fe069d50e9750aa19c29/torchtitan/models/deepseek_v3/model/model.py#L302-L304

cc @drisspg

PR #2182, opened 2025-12-28
Add epoch-based training support

### LINKED ISSUE #613: Ability to train based on epoch
At the moment, you can train using a number of steps, but it would be great if we could train on a number of epochs of the passed in dataset.

Additionally, some way to know how far into a dataset you are when training would be amazing.

yurekami:
## Summary
This PR adds the ability to train models for a specified number of epochs rather than a fixed number of steps, addressing issue #613.

**Key changes:**
- Add `epochs` config option to `Training` dataclass (default `-1` for step-based training)
- Add `epoch` tracking state to `Trainer` class with checkpoint support
- Modify training loop to handle epoch transitions via `DataloaderExhaustedError`
- Add `BaseDataLoader.reset()` method for epoch transitions
- Update logging to show epoch progress when epoch-based training is enabled
- Update checkpointing to properly handle final epoch completion

## Usage
To enable epoch-based training, set `training.epochs` to a positive integer in your config:
```toml
[training]
epochs = 3  # Train for 3 epochs (ignores 'steps' setting)
```

When `epochs` is set to `-1` (default), the existing step-based training behavior is preserved.

## Test plan
- [ ] Verify syntax check passes (done locally)
- [ ] Run existing unit tests
- [ ] Test epoch-based training with a small dataset
- [ ] Verify checkpoint save/load with epoch state
- [ ] Verify backward compatibility with existing step-based configs

Fixes #613

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/torchtitan/components/dataloader.py
+++ b/torchtitan/components/dataloader.py
@@ -41,6 +41,18 @@ class BaseDataLoader(Stateful, ABC):
     def __iter__(self):
         ...
 
+    def reset(self) -> None:
+        """Reset the dataloader state for a new epoch.
+
+        This method is called when epoch-based training transitions to a new epoch.
+        Subclasses should override this if they need to clear any internal state
+        that would prevent re-iteration from the beginning of the dataset.
+
+        The default implementation does nothing, as most dataloaders automatically
+        reset when __iter__ is called again.
+        """
+        pass
+
 
 # pyrefly: ignore [inconsistent-inheritance]
 class ParallelAwareDataloader(StatefulDataLoader, BaseDataLoader):
--- a/torchtitan/components/dataloader.py
+++ b/torchtitan/components/dataloader.py
@@ -151,3 +151,13 @@ class ParallelAwareDataloader(StatefulDataLoader, BaseDataLoader):
         # We don't have to use pickle as DCP will serialize the state_dict. However, we have to
         # keep this for backward compatibility.
         super().load_state_dict(pickle.loads(state_dict[self._rank_id]))
+
+    def reset(self) -> None:
+        """Reset the dataloader state for a new epoch.
+
+        Delegates to the underlying dataset's reset method if available,
+        allowing stateful datasets to clear their internal state and
+        restart iteration from the beginning.
+        """
+        if hasattr(self.dataset, "reset"):
+            self.dataset.reset()
--- a/torchtitan/config/job_config.py
+++ b/torchtitan/config/job_config.py
@@ -257,7 +257,14 @@ class Training:
     """Max norm for gradient clipping"""
 
     steps: int = 10000
-    """How many train steps to run"""
+    """How many train steps to run (ignored if epochs > 0)"""
+
+    epochs: int = -1
+    """
+    Number of epochs to train for. If set to a positive value, training will run
+    for the specified number of epochs and the 'steps' setting will be ignored.
+    If set to -1 (default), training will use step-based termination via 'steps'.
+    """
 
     enable_cpu_offload: bool = False
     """
--- a/torchtitan/hf_datasets/text_datasets.py
+++ b/torchtitan/hf_datasets/text_datasets.py
@@ -165,6 +165,15 @@ class HuggingFaceTextDataset(IterableDataset, Stateful):
 
         return _state_dict
 
+    def reset(self) -> None:
+        """Reset dataset state for a new epoch.
+
+        Clears the token buffer and resets the sample index to start
+        iteration from the beginning of the dataset.
+        """
+        self._sample_idx = 0
+        self._token_buffer = []
+
 
 def build_text_dataloader(
     dp_world_size: int,
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -68,6 +68,7 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
     # additional training states
     step: int
     ntokens_seen: int
+    epoch: int
 
     # Enable debug tracing on failure: https://pytorch.org/docs/stable/elastic/errors.html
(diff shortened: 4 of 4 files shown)


torchtitan/components/dataloader.py
```diff
@@ -41,6 +41,18 @@ class BaseDataLoader(Stateful, ABC):
     def __iter__(self):
         ...
 
+    def reset(self) -> None:
```
wwwjn, line 44: Thank you! Can you explain more why we need a reset method when we reached the end of the dataset?  And seems this method is not implemented for `text_dataset.py`

yurekami: Thank you for the review! Great questions.

**Why we need a reset method:**
When using epoch-based training, after completing an epoch (i.e., iterating through all data), we need to start fresh from the beginning of the dataset for the next epoch. The `reset()` method clears any accumulated internal state that would otherwise prevent clean re-iteration.

Specifically for `HuggingFaceTextDataset`:
- `_sample_idx` tracks the current position in the dataset
- `_token_buffer` accumulates tokens across samples for efficient sequence packing

Without resetting these, the next epoch would either:
1. Start from the wrong position (if `_sample_idx` isn't reset)
2. Have leftover tokens from the previous epoch (if `_token_buffer` isn't cleared)

**Implementation added:**
I've now implemented `reset()` for `text_datasets.py` in the latest commit:

```python
def reset(self) -> None:
    """Reset dataset state for a new epoch."""
    self._sample_idx = 0
    self._token_buffer = []
```

And `ParallelAwareDataloader.reset()` delegates to the underlying dataset:

```python
def reset(self) -> None:
    if hasattr(self.dataset, "reset"):
        self.dataset.reset()
```

This ensures proper epoch transitions in the training loop (train.py line 697).

tianyu-l (changes requested):
Thanks for contributing.

IIUC modern language model training is not epoch-based. There usually are "infinite" data which the model never see anything repeated.

Diffusion model may be an exception, which has its own trainer.

So, could you explain why you need this functionality?

yurekami:
Thanks for the thoughtful question, @tianyu-l — you're right that large-scale pretraining typically streams effectively infinite data without epochs. 

We're adding this because there are concrete scenarios that benefit from explicit epoch support:

- **Fine-tuning** on curated datasets (domain adaptation, instruction following)
- **RLHF/DPO** where preference data is limited and requires multiple passes
- **Instruction tuning** with finite datasets (Alpaca, FLAN, Dolly-style)
- **Research/academic** workflows with constrained compute and smaller corpora
- **Domain-specific training** (medical, legal, multilingual) with curated data
- **Reproducibility** — epochs provide stable units for reporting and baselines

The change keeps backward compatibility (`epochs = -1` remains the default for step-based training), and it addresses issue #613, which maintainers labeled a valid enhancement.

Happy to add documentation clarifying these use cases if that would help. Let me know if this addresses your concern!

tianyu-l:
@yurekami Thanks for the context!

Could you share some citations / references for the use cases you mentioned? They could be 
- high-influence papers / tech reports
- code pointers to other popular codebases
- etc.

yurekami:
@tianyu-l Here are citations and references for the epoch-based use cases:

## High-Influence Papers

| Paper | Training Details | Link |
|-------|------------------|------|
| **InstructGPT** (OpenAI) | SFT trained for **16 epochs** on 13K prompts | [arXiv:2203.02155](https://arxiv.org/abs/2203.02155) |
| **Stanford Alpaca** | **3 epochs** on 52K instruction examples | [Blog](https://crfm.stanford.edu/2023/03/13/alpaca.html) |
| **QLoRA/Guanaco** | Multiple epochs on OASST1 dataset | [arXiv:2305.14314](https://arxiv.org/abs/2305.14314) |
| **LoRA** | Epoch-based fine-tuning paradigm | [arXiv:2106.09685](https://arxiv.org/abs/2106.09685) |
| **FLAN** | Instruction tuning with epoch control | [arXiv:2109.01652](https://arxiv.org/abs/2109.01652) |
| **DPO** | Preference optimization (typically 1-2 epochs) | [arXiv:2305.18290](https://arxiv.org/abs/2305.18290) |

## Popular Codebases Using Epoch-Based Training

| Codebase | Epoch Parameter | Link |
|----------|-----------------|------|
| **HuggingFace Transformers** | `num_train_epochs` (default: 3.0) | [Trainer Docs](https://huggingface.co/docs/transformers/main_classes/trainer) |
| **axolotl** | `num_epochs` in YAML config | [GitHub](https://github.com/axolotl-ai-cloud/axolotl) |
| **LLaMA-Factory** (ACL 2024) | `num_train_epochs` | [GitHub](https://github.com/hiyouga/LLaMA-Factory) |
| **Stanford Alpaca** | `--num_train_epochs 3` | [GitHub](https://github.com/tatsu-lab/stanford_alpaca) |

These cover both foundational papers and widely-used fine-tuning frameworks that rely on epoch-based training for SFT, LoRA, RLHF, and DPO workflows. Let me know if you'd like additional references!

tianyu-l:
@yurekami thanks, makes sense. Let me think about it and review your PR.

I also would like to get your feedback: most of the repos you shared support much more models than torchtitan, some of which also having nD parallel scaling capability. I'd assume they also have many finetune / post training features which are absent in torchtitan. So what makes you / people adopting torchtitan, exclusively or inclusively?

yurekami:
Closing this one. It has fallen behind main, and the open design questions (whether epoch-based training fits torchtitan's infinite-stream data model, and how `reset()` should work for the text datasets) deserve an issue discussion before a patch. Thanks for the review and the questions, and sorry for the long silence.

PR #2185, opened 2025-12-29
make get tp mesh optional in llama4 parallelize

### LINKED ISSUE #2184: bug in parallelize llama4 with tensor_parallel_degree=1
## Command

```
CONFIG_FILE="/home/$USER/torchtitan/torchtitan/models/llama4/train_configs/llama4_17bx16e.toml" NGPU=4 ./run_train.sh \
--metrics.log_freq=10 \
--training.steps=200  \
--parallelism.expert_parallel_degree=4 \
--parallelism.tensor_parallel_degree=1 \
--parallelism.expert_tensor_parallel_degree=1 \
--profiling.enable_profiling \
--profiling.profile_freq=30 \
--compile.enable \
--training.seq_len=8192 \
--activation_checkpoint.mode=none \
--model.print_after_conversion
```

## Error

```
    File "/home/danvm/torchtitan/torchtitan/train.py", line 256, in __init__
      model = self.train_spec.parallelize_fn(model, parallel_dims, job_config)
    File "/home/danvm/torchtitan/torchtitan/models/llama4/infra/parallelize.py", line 138, in parallelize_llama
      tp_mesh = parallel_dims.get_mesh("tp")
    File "/home/danvm/torchtitan/torchtitan/distributed/parallel_dims.py", line 273, in get_mesh
      raise ValueError(
      ...<2 lines>...
      )
  ValueError: Mesh 'tp' is not available. Ensure the corresponding parallelism dimension is enabled (size > 1).
```

### Versions

latest torchtitan main branch

danielvegamyhre:
Fixes #2184 

get_mesh() throws an exception if the mesh dim is none, so the current code implicitly requires TP > 1.

we should use get_optional_mesh() so the user isn't required to use TP.

### DIFF
--- a/torchtitan/models/llama4/infra/parallelize.py
+++ b/torchtitan/models/llama4/infra/parallelize.py
@@ -133,12 +133,9 @@ def parallelize_llama(
     if parallel_dims.tp_enabled or parallel_dims.ep_enabled:
         dual_pipe_v = get_dual_pipe_v_flag(job_config, parallel_dims)
 
-        # tp_mesh might have been set above if tp_enabled, otherwise get it here
-        if tp_mesh is None:
-            tp_mesh = parallel_dims.get_optional_mesh("tp")
         apply_moe_ep_tp(
             model,
-            tp_mesh=tp_mesh,
+            tp_mesh=parallel_dims.get_optional_mesh("tp"),
             ep_mesh=parallel_dims.get_optional_mesh("ep"),
             etp_mesh=parallel_dims.get_optional_mesh("etp"),
             ep_etp_mesh=parallel_dims.get_optional_mesh(["ep", "etp"]),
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
cc @tianyu-l @fegin for review

tianyu-l (approved):
sg, do we need to modify other moe models as well? Qwen, gpt-oss, deepseek v3

torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -135,7 +135,7 @@ def parallelize_llama(
 
         # tp_mesh might have been set above if tp_enabled, otherwise get it here
         if tp_mesh is None:
```
fegin: You can also remove line 136 and line 137.

fegin (approved):
Thanks for catching the bug. @tianyu-l yes, other models with EP supported need the same change. I can do this or @danielvegamyhre if you are able to fix it before landing, that also be good.

torchtitan/models/llama4/infra/parallelize.py
danielvegamyhre: updated and applied fix to qwen3 as well. dsv3 and gpt-oss both seem to use get_optional_mesh() already

PR #2186, opened 2025-12-29
auto-chunk unembed & loss

shunting314:
Add the ability to compile the loss together with the unembed linear layer. The benefit is that we would be able to chunk the logits (which is usually quite large due to large vocab size) by the compiler and reduce peak memory usage.

Here are testing results on qwen3 1.7B. With batch-size=16, the baseline uses 115.85GiB peak memory and gets 54_450 tps. 
By applying the autochunker, we uses 84.15GiB peak memory with 54_244 tps. This is ***37.7%*** peak memory saving trading with ***0.38%*** perf loss. For larger model, the percentage of saving can be smaller since memory usage per layer and activations/optimizer state becomes larger. But saving peak memory usage by auto-chunking is still very nice if perf trade off is very small.

Command for baseline: ```NGPU=1 CONFIG_FILE=torchtitan/models/qwen3/train_configs/qwen3_1.7b.toml ./run_train.sh --compile.enable --training.local_batch_size=16 ```
Command enabling autochunking: ```NGPU=1 CONFIG_FILE=torchtitan/models/qwen3/train_configs/qwen3_1.7b.toml ./run_train.sh --compile.enable --training.local_batch_size=16 --compile.components=model,unembed_and_loss ```

To enable a model for auto-chunking, one tiny change is needed. The forward method need to have an 'unembed' boolean arguments. If it's false, the forward method should not do the unembed linear compuation so we can do that together with the loss compulation

cc @jansel , @eellison , @v0i0

### DIFF
--- a/torchtitan/models/qwen3/model/model.py
+++ b/torchtitan/models/qwen3/model/model.py
@@ -560,6 +560,7 @@ class Qwen3Model(nn.Module, ModelProtocol):
         tokens: torch.Tensor,
         attention_masks: AttentionMasksType | None = None,
         positions: torch.Tensor | None = None,
+        unembed: bool = True,
     ):
         """
         Perform a forward pass through the Transformer model.
@@ -586,5 +587,5 @@ class Qwen3Model(nn.Module, ModelProtocol):
         # pyrefly: ignore [not-callable]
         h = self.norm(h) if self.norm else h
         # pyrefly: ignore [not-callable]
-        output = self.output(h) if self.output else h
+        output = self.output(h) if self.output and unembed else h
         return output
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -26,7 +26,11 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
 def build_cross_entropy_loss(job_config: JobConfig, **kwargs):
     del kwargs  # delete any unused arguments
     loss_fn = cross_entropy_loss
-    if job_config.compile.enable and "loss" in job_config.compile.components:
+    if (
+        job_config.compile.enable
+        and "loss" in job_config.compile.components
+        and "unembed_and_loss" not in job_config.compile.components
+    ):
         logger.info("Compiling the loss function with torch.compile")
         loss_fn = torch.compile(loss_fn, backend=job_config.compile.backend)
     return loss_fn
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -263,6 +263,17 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
 
             self.model_parts = [model]
 
+            if "unembed_and_loss" in self.job_config.compile.components:
+                old_loss_fn = self.loss_fn
+
+                def unembed_and_loss(pred, loss):
+                    # pyrefly: ignore [missing-attribute]
+                    return old_loss_fn(model.output(pred), loss)
+
+                self.loss_fn = torch.compile(
+                    unembed_and_loss, options={"auto_chunker.enable": True}
+                )
+
         self.ft_manager.maybe_set_all_reduce_hook(self.model_parts)
 
         # initialize device memory monitor and get peak flops for MFU calculation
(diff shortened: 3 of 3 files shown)


tianyu-l (changes requested):
Agreed that chunked loss computation is an important feature. It's very nice that we can do it "automatically" with compile!

However, there are several worries:
1. Does it work with FSDP? IIUC, putting model.output in the compile region would cause graph break. Is it not the case?
2. Does it work with loss parallel on the TP mesh? This can be verified by setting `tensor_parallel_degree` > 1.
3. The change is intrusive and making things hard to reason. I guess one "proper" way of doing it might be introducing an output_processor / logits_processor in transformer model code (which also aligns with libraries like vllm) and apply compile to them in parallelize.py.
4. Besides, maybe it's worth adding the eager chunked loss as baseline.

shunting314:
Thanks for reviewing @tianyu-l .

> Besides, maybe it's worth adding the eager chunked loss as baseline.

Is there any specific eager chunked loss you are looking for? This one https://github.com/linkedin/Liger-Kernel/blob/6a383424208b1d79bca2462f7d93bcfb9d13da05/src/liger_kernel/ops/fused_linear_cross_entropy.py#L279 ?

tianyu-l:
oh, I don't think we have to define backward?
e.g. you can check https://fburl.com/code/mfx2xodx

torchtitan/models/qwen3/model/model.py
```diff
@@ -586,5 +587,5 @@ def forward(
         # pyrefly: ignore [not-callable]
         h = self.norm(h) if self.norm else h
```
wwwjn, line 588: Thank you! I have a n00b question - why we don't compile this norm together and let the `self.norm` get chunked as well? 

I was asking because I see vLLM / other model usual split the forward pass into 2 functions (as @tianyu-l  suggested):
1). Transfomers layers forward
2). post transformer layer process 

`self.norm` sometimes is put in 1) (Eg, vllm logit_processor), but sometimes it was put in 2).

tianyu-l:
Is it right that `loss.backward` call doesn't need to be in the compile region? I think we might need to put loss function into model code. cc @wwwjn

torchtitan/models/qwen3/model/model.py
shunting314: the `self.norm` is not a good candidate for chunking. A simplified answer is, ops operating on tensors with a V (vocabulary size) dimension get chunked. The tensors with a V dimension are quite large since V is large and chunking them would be beneficial to reduce peak memory. But even if we change the model implementation to compile `self.norm` together with linear+loss, Inductor would still be able to skip `self.norm` for chunking

wwwjn: Thank you, chunking on Vocab_size dimension makes sense!

wwwjn:
> Is it right that `loss.backward` call doesn't need to be in the compile region? I think we might need to put loss function into model code. cc @wwwjn

Let me try to understand this more: So if we put loss function into model code, we could compile self.output() + loss forward  together as a single region?

tianyu-l:
> Let me try to understand this more: So if we put loss function into model code, we could compile self.output() + loss forward together as a single region?

@wwwjn I think so?

torchtitan/models/qwen3/model/model.py
tianyu-l: @shunting314 sorry I still have some confusion

According to your description, the auto-chunk is similar to https://github.com/apple/ml-cross-entropy and chunks on vocab dimension.

In contrast, https://fburl.com/code/mfx2xodx chunks on the sequence dimension, whose loss computation is simpler because there's no aggregation across chunks.

May I ask what's the benefit of chunking on vocab dim compared with batch / sequence dim?

shunting314: > According to your description, the auto-chunk is similar to https://github.com/apple/ml-cross-entropy and chunks on vocab dimension.

AutoChunker still chunks on the 'flattened' batch+seqlen dimension. I mention vocab-size above since that's a big motivation to chunk the tensors/ops. If the tensor is small, chunking does not bring much benefit.

PR #2187, opened 2025-12-30
experiments: add nemotron3 model to experiments folder

aghilann:
### Summary

  Adds an end-to-end Nemotron3 (NemotronH) experiment under torchtitan/experiments/nemotron3 with:

  - Hybrid model implementation (Mamba2, Attention, MLP, MoE)
  - HF checkpoint conversion/loading support
  - FSDP path for multi-GPU training
  - Python config registry integration (no TOML train configs)

  This is based on NVIDIA’s public Nemotron-3 Nano 30B release and intended to make TorchTitan ready for larger Nemotron variants.

  ### Training Result

  Loss curve on C4 with HF checkpoint load-in (single B200 node, 8 GPUs):
  image

  ### Model Architecture

  hybrid_override_pattern controls layer layout:

  | Symbol | Layer Type | Description |
  |---|---|---|
  | M | Mamba2 | Selective state-space layer |
  | * | Attention | Multi-head attention (GQA) |
  | E / O | MoE | Mixture-of-Experts layer |

  ### Included Configurations

  | Config | Description |
  |---|---|
  | nemotron3_nano_30b | Full Nano-30B-style config |
  | nemotron3_debugmodel | Small debug config for bring-up/testing |

  ### Attention Backends

  - Supported in Nemotron3 experiment: sdpa, flex
  - varlen is explicitly rejected for Nemotron3 in this PR

  ### What’s Included

  - Full model implementation in a flat module layout (aligned with newer TorchTitan model scaffolding)
  - Config registry functions for runtime selection via --module/--config
  - State-dict adapter for HF <-> TorchTitan mapping
  - Nemotron-specific parallelize entrypoint
  - README updates with new run flow

  ### Files
```
  torchtitan/experiments/nemotron3/
  ├── __init__.py
  ├── config_registry.py
  ├── model.py
  ├── parallelize.py
  ├── state_dict_adapter.py
  └── README.md
```

  ### Usage

  # Example: Debug Run
```
  NGPU=1 MODULE=nemotron3 CONFIG=nemotron3_debugmodel /workspace/torchtitan/run_train.sh
```
  # Example: Real Training Run
```
  NGPU=8 MODULE=nemotron3 CONFIG=nemotron3_nano_30b /workspace/torchtitan/run_train.sh
```
  ### Notes

  - Implementation adapted from NVIDIA Nemotron HF release semantics.
  - Verified HF checkpoint flow through adapter path.
  - Mamba dt/A initialization aligned more closely with Mamba2-style initialization.

  ### Future Work

  - [ ] Tensor Parallelism
  - [ ] Context Parallelism
  - [ ] Expert Parallelism
  - [ ] Throughput/MFU tuning (kernel + memory optimizations)

  ### References

  - https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16
  - https://arxiv.org/abs/2312.00752


aghilann:
@tianyu-l Haha, there was actually a much longer commit history. I squashed a bunch of very ugly debug commits before pushing 😆

aghilann:
Hey @tianyu-l, the PR is ready for review. I’m still getting up to speed on distributed training and TorchTitan, so please feel free to correct anything and share as much feedback as you’d like. I’d really appreciate it.

On the question of feedback from people actively using the model, most of what I’ve found is from online threads and word of mouth, plus the benchmark image from the release post. Users have been reporting performance comparable to Qwen 30B, which makes sense given that both are roughly 30B MoE models. At a similar quality level, Nemotron’s use of Mamba layers should require fewer FLOPs, which helps explain the higher inference throughput.

I’ve also heard that some companies are cautious about adopting models produced outside of American labs and tend to prioritize experimenting with or deploying models released by US based organizations. From that perspective, Nemotron could be valuable as well. There does not seem to be a lot of community chatter yet, likely because the release landed right before the holidays.

Relevant thread for context:
[https://www.reddit.com/r/LocalLLaMA/comments/1pocsdy/nemotron_3_nano_30b_is_amazing_tldr/]
image

torchtitan/experiments/nemotron3/model/args.py
```diff
@@ -0,0 +1,174 @@
...
+        ):
+            raise NotImplementedError(
+                "CP support for FlexAttention is still in progress."
+            )
+
+    def get_nparams_and_flops(
+        self, model: nn.Module, seq_len: int
+    ) -> tuple[int, float]:
+        # TODO(aghilann): this number should accurately consider the {Mambda, Attention, MoE layers}
+        return get_dense_model_nparams_and_flops(
```
aghilann: I still need to work on this, marked PR as ready since I did not want to block the PR on the correctness of this.

tianyu-l:
thanks a lot, will review

torchtitan/experiments/nemotron3/model/args.py
```diff
@@ -0,0 +1,174 @@
...
+    mamba_dt_init_floor: float = 1e-4
+    mamba_conv_bias: bool = True
+    mamba_proj_bias: bool = False
+    mamba_chunk_size: int = 128
+
+    # Residual scaling
+    rescale_prenorm_residual: bool = True
+
+    # MoE (Mixture of Experts) configuration
+    n_routed_experts: int = 8
```
wwwjn: Can we reuse MoE args: https://github.com/pytorch/torchtitan/blob/ec246c97c3f53365a5ab8f42f0b586c1f9edfd7b/torchtitan/models/moe/moe.py#L20

torchtitan/experiments/nemotron3/model/args.py
```diff
@@ -0,0 +1,174 @@
...
+    # Special tokens
+    pad_id: int = 0
+    bos_id: int = 1
+    eos_id: int = 2
+
+    # Weight tying
+    enable_weight_tying: bool = False  # tie_word_embeddings
+
+    # Mamba2 configuration
+    use_mamba_kernels: bool = True
```
wwwjn: Similarly, we can make all mamba related config a config dataclass

torchtitan/experiments/nemotron3/model/args.py
wwwjn: You can add a logger.warning, and trying to avoid user to use wrong numerical results?

aghilann:
Hey @tianyu-l , any chance yall could take a look sometime soon!

torchtitan/experiments/nemotron3/model/args.py
```diff
@@ -0,0 +1,174 @@
...
+    residual_in_fp32: bool = False
+
+    # Cache
+    use_cache: bool = True
+    num_logits_to_keep: int = 1
+
+    # Special tokens
+    pad_id: int = 0
+    bos_id: int = 1
+    eos_id: int = 2
```
tianyu-l: It seems several args that I checked are not used. Do you mind doing a filtering of the args?

torchtitan/experiments/nemotron3/model/model.py
```diff
@@ -0,0 +1,1082 @@
...
+        hidden_states = hidden_states.to(torch.float32)
+        variance = hidden_states.pow(2).mean(-1, keepdim=True)
+        hidden_states = hidden_states * torch.rsqrt(variance + self.variance_epsilon)
+        return (self.weight.to(torch.float32) * hidden_states).to(input_dtype)
+
+    def reset_parameters(self) -> None:
+        nn.init.ones_(self.weight)
+
+
+class MambaRMSNormGated(nn.Module):
```
tianyu-l: can we just use `nn.RMSNorm`?

torchtitan/experiments/nemotron3/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+    "Nemotron3ModelArgs",
+    "Nemotron3Model",
+    "Nemotron3StateDictAdapter",
+    "parallelize_nemotron3",
+    "nemotron3_args",
+]
+
+
+# NemotronH model flavors
+# Pattern key: M=Mamba2, *=Attention, -=MLP, E=MoE
```
tianyu-l: It seems MLP is never used.

torchtitan/experiments/nemotron3/model/model.py
```diff
@@ -0,0 +1,1082 @@
...
+
+        topk_weights = topk_weights * self.routed_scaling_factor
+        return topk_indices, topk_weights
+
+    def init_weights(self) -> None:
+        """Initialize router weights."""
+        nn.init.trunc_normal_(self.weight, mean=0.0, std=0.02)
+
+
+class MoE(nn.Module):
```
tianyu-l: can we replace this by using torchtitan MoE module? https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/moe/moe.py

This will ease the maintenance and application of EP later.

torchtitan/experiments/nemotron3/state_dict_adapter.py
```diff
@@ -0,0 +1,280 @@
...
+import re
+from typing import Any
+
+from torchtitan.models.utils import MoEStateDictAdapter
+from torchtitan.tools.logging import logger
+
+from .args import Nemotron3ModelArgs
+
+
+class Nemotron3StateDictAdapter(MoEStateDictAdapter):
```
tianyu-l, line 16: could you verify model + state dict adapter correctness by following https://github.com/pytorch/torchtitan/tree/main/scripts/checkpoint_conversion

torchtitan/experiments/__init__.py
tianyu-l, line 1: Would you be helping maintain this folder? If so, please add yourself to https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/README.md#current-experiments

torchtitan/experiments/nemotron3/README.md
```diff
@@ -0,0 +1,107 @@
...
+- **debug_model**: Can run on a single GPU for development
+
+## Roadmap
+
+Currently using **FSDP** (Fully Sharded Data Parallel) to get training up and running. The following parallelism strategies are planned for future support:
+
+| Feature | Status | Notes |
+|---------|--------|-------|
+| FSDP | ✅ Supported | Currently used for distributed training |
+| Tensor Parallelism | 🚧 To be added | Likely needed for larger model variants |
```
tianyu-l: Do you have an idea of how this can be done?

torchtitan/experiments/nemotron3/model/model.py
```diff
@@ -0,0 +1,1082 @@
...
+        query_states = query_states.transpose(1, 2)
+        key_states = key_states.transpose(1, 2)
+        value_states = value_states.transpose(1, 2)
+
+        # Repeat K, V for GQA
+        key_states = repeat_kv(key_states, self.num_kv_groups)
+        value_states = repeat_kv(value_states, self.num_kv_groups)
+
+        # SDPA attention (causal)
+        attn_output = F.scaled_dot_product_attention(
```
tianyu-l: It seems here you only specify causal attention. In general, does Mamba-2 only support full causal sequence modeling, or does it also support other mask patterns?

aghilann:
Thanks for the review @tianyu-l, I do want to give a small heads up in on vacation/travelling for the next 1 month so I'll address these once I'm back.

torchtitan/experiments/nemotron3/README.md
aghilann: I haven’t dug into this deeply yet. My current plan is to push FSDP2 + EP (when added) as far as possible. I expect we’ll need to add EP for larger Nemotron models, but FSDP2 + EP alone may already be enough to run them.

torchtitan/experiments/nemotron3/model/args.py
aghilann: Sure, I just directly copied stuff from the HF `config.json` but you are right, better not to have useless fields.

torchtitan/experiments/nemotron3/model/model.py
aghilann: I copied this implementation from NVIDIA’s reference file: https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/blob/main/modeling_nemotron_h.py (the Nemotron3 model file here is
entirely taken from that source). I just made a small cleanup by having MambaRMSNormGated inherit from nn.RMSNorm right now.

torchtitan/experiments/nemotron3/state_dict_adapter.py
aghilann: Yep! I used the process of going from HF to DCP to HF all the time to help me debug when I was first implementing the state dict. Done!

torchtitan/experiments/__init__.py
aghilann: Yes I would be glad to, done!

torchtitan/experiments/nemotron3/model/model.py
aghilann: I couldn’t use the standard torchtitan MoE path as-is and still preserve exact Nemotron behavior, because Nemotron’s router and expert computations differ. I implemented a Nemotron-specific
  MoE variant in the shared MoE module and wired it through the existing config/build path so Nemotron keeps 1:1 behavior while staying aligned with the library pattern.

torchtitan/experiments/nemotron3/model/model.py
aghilann: I’m not entirely sure, but based on my AI conversations, this is my current understanding: in this implementation, we only support full causal modelling (`is_causal=True`). Mamba-2 seems inherently causal and doesn’t natively support arbitrary attention-style masks the same way SDPA does. A lot of the SSM details are still new to me, and I took this directly from the upstream modelling file, so apologies for the weak/incomplete response.

aghilann:
Hey @tianyu-l , ready for another review, Thanks for the great feedback on the last review!

torchtitan/experiments/nemotron3/model/args.py
```diff
@@ -142,11 +141,60 @@ def update_from_config(self, job_config: JobConfig, **kwargs) -> None:
...
+        nparams = nparams_dense + nparams_sparse
+
+        if self.n_routed_experts > 0:
+            nparams_sparse_active = (
+                nparams_moe_router
+                + nparams_shared_experts
+                + nparams_experts * self.num_experts_per_tok // self.n_routed_experts
+            )
+        else:
+            nparams_sparse_active = nparams_moe_router + nparams_shared_experts
```
aghilann: I relied on AI a bit to write this estimate; it seems roughly correct

torchtitan/experiments/nemotron3/model/model.py
tianyu-l: @garrett361 could you advise here? I feel we at least need to support sequence packing with block causal mask?

torchtitan/experiments/README.md
tianyu-l, line 1: please put this under nemotron3/ instead of common/

torchtitan/experiments/nemotron3/model/model.py
garrett361: The upstream mamba2 kernels have [varlen support (via cu_seqlens)](https://github.com/state-spaces/mamba/blob/e579860e30117ec5eb64b0ada9cd110f1d2e348b/mamba_ssm/ops/triton/ssd_combined.py?plain=1#L593-L621), so vanilla sequence packing is certainly possible. 

It's also necessary to match pair the above w/ varlen support for the causal convolution. [FLA has this.](https://github.com/fla-org/flash-linear-attention/blob/2f18f7d4cf08f43512c9ae931ab55000b9c93047/fla/modules/conv/causal_conv1d.py?plain=1#L12-L26)

Not aware of any pure pytorch impl of either of these, though (also haven't looked much). Any attention pattern more complicated than simple packing is very architecturally unnatural for mamba2/most linear attn algos.

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,915 @@
...
+
+        y = y.reshape(batch_size, seq_len, -1)
+
+        # Apply gated normalization
+        scan_output = self.norm(y, gate)
+
+        # Output projection
+        return self.out_proj(scan_output.to(dtype))
+
+    def init_weights(self, init_std: float) -> None:
```
garrett361, line 693: Just noting that it's super important to get the init right. This came up on X again recently: https://x.com/MayankMish98/status/2026769614022259079?s=20

And the init_weights here doesn't seem to match [the mamba_ssm version](https://github.com/state-spaces/mamba/blob/e579860e30117ec5eb64b0ada9cd110f1d2e348b/mamba_ssm/modules/mamba2.py?plain=1#L119-L140). I suggest adhering to their code as closely as possible.

torchtitan/experiments/nemotron3/model/model.py
garrett361: Also, the pure pytorch impl above is likely very slow. Think it's also quadratic `O(seqlen^2)` in compute and memory, IIRC.

tianyu-l: @aghilann  We have sdpa (causal), varlen (block causal), flex attn (arbitrary mask) in torchtitan. Can we try to utilize the existing attention module similar to how other models are doing?

aghilann: Used the existing attention module system to extend it for flex attention too, but was unable to get it to work for varlen attention.

torchtitan/experiments/nemotron3/model.py
aghilann: Made the requested changes so the Nemotron-3 modeling code now aligns with the Mamba SSM code you linked. That said, I find it surprising that NVIDIA pre-trained a model without noticing any issues, given that the modeling code I’m using comes directly from the Hugging Face implementation.

aghilann:
Hey @tianyu-l 

Made sone changes, sorry for the late response, have been swamped lately. Ready for re-review! 

1) Refactor to match the 2 format and codebase structure
2) Used the existing attention selecting utils and it works with sdpa and flex attention but not varlen
3) Added in the small fix @garrett361 mentioned into the init_weight functions for the Mamba block

FYI I'm pretty sure NVIDIA will be releasing the larger variants of the Nemotron3 family of models at NVIDIA GTC in one week

torchtitan/experiments/nemotron3/model.py
garrett361: I don't think NVIDIA used the HF implementation. 

Where is the A_log and D init code? Am I missing it?

aghilann: Sorry, I meant to say the implementation NVIDIA released onto HF, this file: 
https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/blob/main/configuration_nemotron_h.py

I had the A_log and D Init code in the constructor and not the init_weights method, I'll move it to make it easier 

Added it here:
https://github.com/pytorch/torchtitan/pull/2187/changes/5180998c697bdd4308c94ec7ea3106823b695701#diff-343261e47cb3a75dec2813ef516fabf88f320216107a491b66320dc58d35fbeeR704

garrett361: Thanks, it's not enough to only have it in the constructor because that's all ignored for meta-device init. We rely on the `init_weights` to properly initialize weights in this case. Hence me harping on this.

aghilann:
Hey @tianyu-l , following up on a PR review here!

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,1236 @@
...
+
+        if self.n_kv_heads is None:
+            self.n_kv_heads = self.n_heads
+
+        if self.attn_backend not in {"sdpa", "varlen", "flex"}:
+            raise ValueError(
+                f"attn_type must be one of ['sdpa', 'varlen', 'flex'], got {self.attn_type}"
+            )
+        if self.attn_backend == "varlen":
+            raise ValueError("Varlen attention is not supported with Nemotron3.")
```
tianyu-l, line 131: any reason?

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,1236 @@
...
+    ndim = x.ndim
+    assert ndim > 1
+    seqlen = x.shape[1]
+    freqs_cis = freqs_cis[0:seqlen]
+    assert freqs_cis.shape == (seqlen, x.shape[-1])
+    shape = [d if i == 1 or i == ndim - 1 else 1 for i, d in enumerate(x.shape)]
+    return freqs_cis.view(*shape)
+
+
+def apply_rotary_emb(
```
tianyu-l, line 380: could we reuse what's in https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/rope.py

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,1236 @@
...
+    hidden_states = hidden_states[:, :, None, :, :].expand(
+        batch, num_kv_heads, n_rep, slen, head_dim
+    )
+    return hidden_states.reshape(batch, num_kv_heads * n_rep, slen, head_dim)
+
+
+# =============================================================================
+# Normalization Layers
+# =============================================================================
+class RMSNorm(nn.Module):
```
tianyu-l, line 411: use pytorch rmsnorm

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,1236 @@
...
+        )
+        dt = torch.clamp(dt, min=dt_init_floor)
+        # Inverse of softplus.
+        return dt + torch.log(-torch.expm1(-dt))
+
+
+# =============================================================================
+# Attention Layer
+# =============================================================================
+class Attention(nn.Module):
```
tianyu-l, line 791: can we reuse https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,1236 @@
...
+# Activation Functions
+# =============================================================================
+def relu2(x: torch.Tensor) -> torch.Tensor:
+    return F.relu(x).square()
+
+
+ACT2FN: dict[str, Callable[[torch.Tensor], torch.Tensor]] = {
+    "relu": F.relu,
+    "relu2": relu2,
+    "gelu": F.gelu,
```
tianyu-l, line 336: several of them are not used, do we want to keep them?

torchtitan/experiments/nemotron3/parallelize.py
```diff
@@ -0,0 +1,169 @@
...
+    This shards model parameters across GPUs to reduce memory usage.
+    """
+    mp_policy = MixedPrecisionPolicy(param_dtype=param_dtype, reduce_dtype=reduce_dtype)
+    fsdp_config = {"mesh": dp_mesh, "mp_policy": mp_policy}
+
+    if cpu_offload:
+        fsdp_config["offload_policy"] = CPUOffloadPolicy()
+
+    # Determine reshard_after_forward behavior
+    match reshard_after_forward_policy:
```
tianyu-l, line 131: can reuse https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/fsdp.py#L8

torchtitan/experiments/nemotron3/parallelize.py
```diff
@@ -0,0 +1,169 @@
...
+    # Shard final norm and output together (don't reshard by default as they're used immediately)
+    if model.norm is not None and model.output is not None:
+        fully_shard(
+            [model.norm, model.output],
+            **fsdp_config,
+            reshard_after_forward=reshard_after_forward_policy == "always",
+        )
+
+    # Apply FSDP to the entire model
+    fully_shard(model, **fsdp_config)
```
tianyu-l, line 169: need to call https://github.com/pytorch/torchtitan/blob/d84e83dc4ef4615afefe32dc83c1369a50132ba3/torchtitan/models/llama3/parallelize.py#L343

torchtitan/experiments/README.md
```diff
@@ -26,6 +26,8 @@ We provide this `experiments/` folder to host experiments that add significant v
...
 | [autoparallel](./autoparallel/) | [![Auto Parallel 8 GPU Integration Tests](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_autoparallel.yaml/badge.svg?branch=main)](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_autoparallel.yaml?query=branch%3Amain) | [@wconstab](https://github.com/wconstab) [@xmfan](https://github.com/xmfan) |
+| [nemotron3](./nemotron3/) | TBA | [@aghilann](https://github.com/aghilann) |
+| [torchcomms](./torchcomms/) | [![TorchComms 8 GPU Integration Tests](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_torchcomms.yaml/badge.svg?branch=main)](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_torchcomms.yaml?query=branch%3Amain) | [@d4l3k](https://https://github.com/d4l3k) [@fduwjj](https://github.com/fduwjj) [@mori360 ](https://github.com/mori360) |
```
tianyu-l, line 30: remove

torchtitan/models/common/moe/moe_nemotron.py
```diff
@@ -0,0 +1,226 @@
...
+            topk_weights = topk_weights / denominator
+
+        topk_weights = topk_weights * self.routed_scaling_factor
+        return topk_indices, topk_weights
+
+    def init_weights(self, init_std: float = 0.02) -> None:
+        nn.init.trunc_normal_(self.weight, mean=0.0, std=init_std)
+
+
+class NemotronMoE(MoE):
```
tianyu-l, line 135: can we just reuse https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/moe/moe.py

torchtitan/experiments/nemotron3/model.py
```diff
@@ -0,0 +1,1236 @@
...
+    """
+    Nemotron3 (NemotronH) Hybrid Model.
+
+    A hybrid architecture combining Mamba2, Attention, and MLP/MoE layers.
+    The layer pattern is defined by the `hybrid_override_pattern` configuration.
+    """
+
+    Config = Nemotron3Config
+
+    def __init__(self, config: Config) -> None:
```
tianyu-l, line 1071: we are doing a big refactor, after which maybe you don't need to define some of the functions in this class

torchtitan/experiments/nemotron3/parallelize.py
```diff
@@ -0,0 +1,169 @@
...
+    if parallel_dims.cp_enabled:
+        raise NotImplementedError(
+            "Context Parallelism (CP) is not yet supported for Nemotron3. "
+            "Please set parallelism.context_parallel_degree = 1."
+        )
+    if parallel_dims.ep_enabled:
+        raise NotImplementedError(
+            "Expert Parallelism (EP) is not yet supported for Nemotron3. "
+            "Please set parallelism.expert_parallel_degree = 1."
+        )
```
tianyu-l, line 80: If you switch to existing MoE, I think the EP support will be straightforward.

torchtitan/experiments/nemotron3/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    "Nemotron3StateDictAdapter",
+    "parallelize_nemotron3",
+    "nemotron3_configs",
+    "model_registry",
+]
+
+
+# NemotronH model flavors
+# Pattern key: M=Mamba2, *=Attention, E=MoE
+nemotron3_configs = {
```
tianyu-l, line 26: could you also add the Nemotron 3 Super?

torchtitan/experiments/nemotron3/config_registry.py
```diff
@@ -0,0 +1,189 @@
...
+
+def _configure_local_hf_cache() -> None:
+    # Keep default behavior when users explicitly configure cache env vars.
+    if "HF_HOME" not in os.environ:
+        os.environ["HF_HOME"] = "/tmp/huggingface"
+    if "HF_DATASETS_CACHE" not in os.environ:
+        os.environ["HF_DATASETS_CACHE"] = os.path.join(
+            os.environ["HF_HOME"], "datasets"
+        )
+    os.makedirs(os.environ["HF_DATASETS_CACHE"], exist_ok=True)
```
tianyu-l, line 36: why do we need this?

torchtitan/experiments/nemotron3/state_dict_adapter.py
tianyu-l: could you show some evidence in PR summary / README?

torchtitan/experiments/nemotron3/model.py
garrett361: What's the refactor? Link?

tianyu-l: https://github.com/pytorch/torchtitan/pull/2633
still WIP

PR #2188, opened 2025-12-31
[rl] Using JobConfig as the centralized config system for inference and simple GRPO

wwwjn:
1. Add job_config.py to extend current JobConfig. Now an issue is `trainer`'s config and `generator`'s config are not symmetric, eg `Parallelism` and `Generation.parallelism`
2. Use job config system as the centralized / source-of-truth config, loading config from `run_configs/qwen3_0.6b.toml` file.
3. Refactor the generator to use EngineArgs() and LLMEngine(), instead of LLM()
4. Rename simple_rl_multiprocess -> simple_grpo to be more descriptive
5. Clean up unused code branch

Test: (trainer ddp = 2, n_generator =1)
Screenshot 2025-12-30 at 7 34 00 PM

Following-up refactors:
-  Refactor2: vllm model register - using setup.py and plugin instead of import 
- Refactor3: Weight updater,  by directly passing state_dict (DTensor) between trainer and generator
- Refactor4: Use torchtitan Trainer, modularize each component

### DIFF
--- a/torchtitan/experiments/rl/unified/job_config.py
+++ b/torchtitan/experiments/rl/unified/job_config.py
@@ -5,9 +5,9 @@
 # LICENSE file in the root directory of this source tree.
 
 """
-Extended JobConfig for RL/Inference workloads.
+Extended JobConfig for RL/Generation workloads.
 
-This module extends TorchTitan's base JobConfig with inference-specific
+This module extends TorchTitan's base JobConfig with generation-specific
 configurations needed for vLLM integration.
 """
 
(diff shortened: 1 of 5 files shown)


torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,102 @@
...
+    Top-p (nucleus) sampling threshold.
+    Only tokens with cumulative probability up to top_p are considered.
+    """
+
+    max_tokens: int = 100
+    """Maximum number of tokens to generate"""
+
+
+@dataclass
+class Generate:
```
wwwjn, line 46: We really need a better name for this field, as well as `RL`

wwwjn:
Close with #2191

PR #2190, opened 2025-12-31
use comms in compiler toolkit

dolpm:
* __->__ #2190

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -14,9 +14,8 @@ from typing import Any, Iterable
 
 import torch
 import torch.distributed.checkpoint.stateful
-from torch.distributed.elastic.multiprocessing.errors import record
-
 import torchtitan.protocols.train_spec as train_spec_module
+from torch.distributed.elastic.multiprocessing.errors import record
 from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.components.dataloader import DataloaderExhaustedError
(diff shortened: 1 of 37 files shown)


torchtitan/train.py
```diff
@@ -248,10 +248,10 @@ def __init__(self, job_config: JobConfig):
             # confirm that user will be able to view loss metrics on the console
             ensure_pp_loss_visible(parallel_dims, job_config, color)
         else:
-            # apply PT-D Tensor Parallel, activation checkpointing, torch.compile, Data Parallel
-            model = self.train_spec.parallelize_fn(model, parallel_dims, job_config)
-
+            # Materialize model from meta device to actual CUDA memory BEFORE parallelization
+            # This is required for torchcomms which needs real tensors for NCCL broadcasts
```
tianyu-l: curious
- when would NCCL broadcasts happen for torchcomms, in parallelization?
- is this inherent limitation from torchcomms? Without meta init -> parallelization -> move to cuda, large models will OOM in init phase.

dolpm: yeah, parallelization. this kind of applies to any collective being ran before you materialize the model...

although i think, in retrospect, it might be because i added some barriers during mesh initialization that might somehow leak into here. i will look -- i was just trying to get things to work for the sake of some offline tests.

dolpm: this ended up being an issue with how i was handling meta tensors on the comms side.

PR #2191, opened 2026-01-02
[rl] Use torchtitan config system for inference and simple GRPO

wwwjn:
* #2244
* #2221
* #2194
* __->__ #2191

1. Add job_config.py to extend current JobConfig. Now an issue is `trainer`'s config and `generator`'s config are not symmetric, eg `Parallelism` and `Generation.parallelism`
2. Use job config system as the centralized / source-of-truth config, loading config from `run_configs/qwen3_0.6b.toml` file.
3. Refactor the generator to use EngineArgs() and LLMEngine(), instead of LLM()
4. Rename simple_rl_multiprocess -> simple_grpo to be more descriptive
5. Clean up unused code branch

Test: (trainer ddp = 2, n_generator =1)
Screenshot 2025-12-30 at 7 34 00 PM

Following-up refactors:
-  Refactor2: vllm model register - using setup.py and plugin instead of import 
- Refactor3: Weight updater,  by directly passing state_dict (DTensor) between trainer and generator
- Refactor4: Use torchtitan Trainer, modularize each component

### DIFF
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -4,9 +4,10 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import asyncio
+import glob
 import logging
 import os
+import shutil
 
 from dataclasses import dataclass, field
 from typing import List
@@ -14,8 +15,7 @@ from typing import List
 import torch
 from monarch.actor import Actor, endpoint
 from safetensors.torch import save_file
-from torchtitan.config import CommConfig, Configurable
-from torchtitan.config.configs import ParallelismConfig
+from torchtitan.config import CommConfig, Configurable, ParallelismConfig
 from torchtitan.distributed import utils as dist_utils
 
 # Import unified module - this automatically registers TorchTitan models with vLLM
@@ -67,294 +67,6 @@ class TrajectoryData:
     advantages: torch.Tensor
 
 
-class VLLMEngine(Configurable):
-    """
-    vLLM engine for fast rollouts with weight updates.
-
-    Note: vLLM loads from model_config.model path, so we create a temporary
-    directory with updated weights and restart the engine. This is faster than
-    recreating temp dirs repeatedly and handles config/tokenizer files properly.
-
-    Constructed via ``VLLMEngine(config, model_path=..., dump_folder=...)``.
-    """
-
-    @dataclass(kw_only=True, slots=True)
-    class Config(Configurable.Config):
-        """vLLM engine configuration for rollout generation."""
-
-        dtype: str = "bfloat16"
-        """Data type for model weights (auto, float16, bfloat16, float32)."""
-
-        gpu_memory_limit: float = 0.5
-        """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
-
-        enforce_eager: bool = True
-        """Disable CUDA graphs in vLLM (use eager execution)."""
-
-        seed: int = 42
-        """Random seed for reproducible generation."""
-
-        parallelism: ParallelismConfig = field(default_factory=ParallelismConfig)
-        """Parallelism configuration for the vLLM engine."""
-
-        sampling: VLLMSamplingConfig = field(default_factory=VLLMSamplingConfig)
-        """Default sampling parameters for generation."""
-
-    def __init__(
-        self,
-        config: Config,
-        *,
-        model_path: str,
-        dump_folder: str,
-    ) -> None:
-        self.config = config
-        self.base_model_path = model_path
-
-        self.temp_model_dir = os.path.abspath(
-            os.path.join(dump_folder, "vllm_temp_model")
-        )
-        os.makedirs(self.temp_model_dir, exist_ok=True)
-
-        import glob
-
-        # Copy config/tokenizer files from base model to temp dir
-        import shutil
-
-        for file in [
-            "config.json",
-            "tokenizer.json",
-            "tokenizer_config.json",
-            "special_tokens_map.json",
-            "merges.txt",
-            "vocab.json",
-        ]:
-            src = os.path.join(model_path, file)
-            if os.path.exists(src):
-                shutil.copy2(src, self.temp_model_dir)
-
-        # Copy the original model shard files if they exist
-        # We'll overwrite these with our single model.safetensors later
-        for shard_file in glob.glob(os.path.join(model_path, "model-*.safetensors")):
-            dst = os.path.join(self.temp_model_dir, os.path.basename(shard_file))
-            shutil.copy2(shard_file, dst)
-
-        # Copy index file if it exists
-        index_file = os.path.join(model_path, "model.safetensors.index.json")
-        if os.path.exists(index_file):
-            shutil.copy2(index_file, self.temp_model_dir)
-
-        self.engine = None
-        logger.info("vLLM rollout engine initialized (will load on first use)")
-
-    def update_weights(self, vllm_state: dict) -> None:
-        """
-        Update vLLM model weights from vLLM-compat state dict.
-
-        This converts weights to vLLM format, saves them, and reloads using
-        vLLM's reload_weights() API after updating the model path config.
-
-        Args:
-            vllm_state: vLLM model state dict
-        """
-        # Save to temp model directory
-        import os
-
-        checkpoint_path = os.path.join(self.temp_model_dir, "model.safetensors")
-
-        # Update the shard files that vLLM will actually load
-        # We need to split our weights to match the original 2-shard structure
-        import glob
-
-        shard_files = sorted(
-            glob.glob(os.path.join(self.temp_model_dir, "model-*.safetensors"))
-        )
-        index_file = os.path.join(self.temp_model_dir, "model.safetensors.index.json")
-
-        # TODO: need to replace this with Torchtitan's checkpoint save and load
-        # right now we hardcoded to work with 1 safetensor files which we only
-        # tested on Qwen3 0.6B model. In the longer term, need to use TorchStore
-        # to achieve the weight communication.
-        # only generator rank 0 saves the weight
-        if torch.distributed.get_rank() == 0:
-            logger.info(f"Saving weights to {checkpoint_path}")
-
-            # Ensure weights stay in bfloat16
-            vllm_state = {
-                k: v.to(torch.bfloat16) if v.dtype == torch.float32 else v
-                for k, v in vllm_state.items()
-            }
-            save_file(vllm_state, checkpoint_path)
-
-        # Synchronize all ranks before reloading to ensure rank 0 finished writing
-        torch.distributed.barrier()
-        logger.info(
-            f"[Rank {torch.distributed.get_rank()}] Synchronized after weight save"
-        )
-
-        # First time: create the engine using LLMEngine and EngineArgs
-        if self.engine is None:
-            cfg = self.config
-
-            engine_args = EngineArgs(
-                # Model configuration
-                model=self.temp_model_dir,
-                trust_remote_code=True,
-                dtype=cfg.dtype,
-                # Parallelism configuration
-                tensor_parallel_size=cfg.parallelism.tensor_parallel_degree,
-                distributed_executor_backend="external_launcher",
-                # Memory and performance
-                gpu_memory_utilization=cfg.gpu_memory_limit,
-                enforce_eager=cfg.enforce_eager,
-                # Seed
-                seed=cfg.seed,
-                # HuggingFace overrides to use TorchTitan model.
-                # TODO: make this field configurable and align with model registration
-                hf_overrides={"architectures": ["Qwen3TorchTitanForCausalLM"]},
-                attention_config=AttentionConfig(
-                    backend=AttentionBackendEnum.FLASH_ATTN,
-                ),
-            )
-
-            logger.info("Initializing LLMEngine from EngineArgs...")
-            self.engine = LLMEngine.from_engine_args(engine_args)
-            logger.info("Created new vLLM LLMEngine")
-        else:
-            # Direct parameter copy into model tensors.
-            # This bypasses vLLM's reload_weights() which uses a layerwise
-            # reload mechanism that moves params to meta device
-            from torchtitan.experiments.rl.vllm_compat.weights import vllm_to_torchtitan
-
-            titan_state = vllm_to_torchtitan(vllm_state)
-            self._direct_weight_update(titan_state)
-
-    def _direct_weight_update(self, titan_state: dict) -> None:
-        """Update model weights by copying directly into GPU parameters.
-
-        Args:
-            titan_state: TorchTitan format state dict (w1/w2/w3, wq/wk/wv/wo, etc.)
-        """
-
-        # Access model from vLLM engine
-        model = self.engine.model_executor.driver_worker.get_model()
-        params = dict(model.named_parameters())
-
-        for name, new_weight in titan_state.items():
-            # TorchTitanVLLMModelWrapper stores the model as self.model,
-            # so parameters have a "model." prefix
-            param_name = f"model.{name}"
-            if param_name in params:
-                param = params[param_name]
-                new_w = new_weight.to(device=param.device, dtype=param.dtype)
-                param.data.copy_(new_w)
-
-    @torch.no_grad()
-    def generate(
-        self,
-        prompt_texts: list[str],
-        max_new_tokens: int = 20,
-        temperature: float = 1.0,
-        n_samples_per_prompt: int = 4,
-    ) -> tuple[
-        list[str], torch.Tensor, list[list[int]], list[list[float]], list[list[int]]
-    ]:
-        """
-        Generate samples using vLLM LLMEngine.
-
-        Args:
-            prompt_texts: List of prompt strings
-            max_new_tokens: Max tokens to generate
-            temperature: Sampling temperature
-            n_samples_per_prompt: Number of samples per prompt
-
-        Returns:
-            completions: List of completion strings
-            log_probs: [batch] - Sum of log probs for each completion
-            token_ids: List of token ID lists for each completion (generated tokens only)
-            token_log_probs: List of per-token log prob lists for each completion
-            prompt_token_ids: List of prompt token ID lists for each completion
-        """
-        logger.info(
-            f"Starting generation: {len(prompt_texts)} prompts, "
-            f"n_samples_per_prompt={n_samples_per_prompt}, "
-            f"max_tokens={max_new_tokens}, temp={temperature}"
-        )
-
-        sampling_params = SamplingParams(
-            temperature=temperature,
-            max_tokens=max_new_tokens,
-            n=n_samples_per_prompt,
-            seed=42,
-            logprobs=1,
-            prompt_logprobs=1,  # Also get prompt log probs to access prompt token IDs
-            output_kind=RequestOutputKind.FINAL_ONLY,  # Only return completed outputs
-        )
-
-        # Add one request per prompt; vLLM handles n_samples_per_prompt via n=
-        for request_id, prompt in enumerate(prompt_texts):
-            self.engine.add_request(str(request_id), prompt, sampling_params)
-
-        # Step through engine until all requests are finished
-        all_outputs = []
-        while self.engine.has_unfinished_requests():
-            request_outputs = self.engine.step()
-            all_outputs.extend(request_outputs)
-
-        # Extract completions and log probs
-        completions = []
-        log_probs_list = []
-        token_ids_list = []
-        token_log_probs_list = []
-        prompt_token_ids_list = []
-
-        for output in all_outputs:
-            prompt_token_ids = output.prompt_token_ids
-
-            for sample in output.outputs:
-                completions.append(sample.text)
-
-                # Store prompt tokens for this sample
-                prompt_token_ids_list.append(prompt_token_ids)
-
-                # Extract token IDs (generated tokens only)
-                token_ids = sample.token_ids
-                token_ids_list.append(token_ids)
-
-                # Extract per-token log probs
-                per_token_log_probs = [
-                    list(logprob_dict.values())[0].logprob
-                    for logprob_dict in sample.logprobs
-                ]
-                token_log_probs_list.append(per_token_log_probs)
-
-                # Sum log probs across generated tokens
-                total_log_prob = sum(per_token_log_probs)
-                log_probs_list.append(total_log_prob)
-
-        log_probs = torch.tensor(log_probs_list, dtype=torch.float32)
-
-        return (
-            completions,
-            log_probs,
-            token_ids_list,
-            token_log_probs_list,
-            prompt_token_ids_list,
-        )
-
-    def __del__(self):
-        """Cleanup vLLM engine."""
-        if hasattr(self, "engine"):
-            del self.engine
-            torch.cuda.empty_cache()
-
-
-class GeneratorState:
-    """States for the Generator's state machine."""
-
-    READY_TO_GENERATE = "READY_TO_GENERATE"
-    READY_TO_UPDATE = "READY_TO_UPDATE"
-
-
 class Generator(Actor, Configurable):
     """
     Generates rollouts using vLLM engine.
@@ -377,8 +89,23 @@ class Generator(Actor, Configurable):
     class Config(Configurable.Config):
         """Generator actor configuration."""
 
-        vllm_engine: VLLMEngine.Config = field(default_factory=VLLMEngine.Config)
-        """vLLM rollout engine configuration."""
+        dtype: str = "bfloat16"
+        """Data type for model weights, passed directly to vLLM (auto, float16, bfloat16, float32)."""
+
+        gpu_memory_limit: float = 0.5
+        """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
+
+        enforce_eager: bool = True
+        """Disable CUDA graphs in vLLM (use eager execution)."""
+
+        seed: int | None = None
+        """Random seed for reproducible generation. None means no fixed seed."""
+
+        parallelism: ParallelismConfig = field(default_factory=ParallelismConfig)
+        """Parallelism configuration for the vLLM engine."""
+
+        sampling: VLLMSamplingConfig = field(default_factory=VLLMSamplingConfig)
+        """Default sampling parameters for generation."""
 
         vllm_attention_backend: str = "FLASH_ATTN"
         """vLLM attention backend to use (e.g., FLASH_ATTN, XFORMERS)."""
@@ -410,8 +137,8 @@ class Generator(Actor, Configurable):
 
         # Extract needed fields from configs
         self.model_path = model_path
-        self.max_new_tokens = config.vllm_engine.sampling.max_tokens
-        self.temperature = config.vllm_engine.sampling.temperature
+        self.max_new_tokens = config.sampling.max_tokens
+        self.temperature = config.sampling.temperature
         self.group_size = policy_optimization.group_size
         self.grpo_beta = policy_optimization.beta
         self.use_stable_grpo = policy_optimization.use_stable_grpo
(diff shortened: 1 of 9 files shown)


torchtitan/experiments/rl/unified/README.md
```diff
@@ -66,10 +66,10 @@ python torchtitan/experiments/rl/unified/infer.py --model-ckpt-path <path_to_mod
 
 5. Run simple rl loop
 ```
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_grpo.py \
+    --job.config_file torchtitan/experiments/rl/unified/run_configs/qwen3_0.6b.toml
 ```
-Right now we only support VLLM_COMPAT mode, which could achieve trainer and generator bitwise identical. We are working on support UNIFIED mode,
-which uses a unified model definition for trainer and generator.
+We uses a unified model definition for trainer and generator, which could achieve trainer and generator bitwise identical.
```
allenwang28:
```suggestion
We use a unified model definition for the trainer and generator, ensuring bitwise-identical models to address a class of subtle correctness bugs in RL for LLMs.
```
nit

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -319,58 +318,46 @@ class Generator(Actor):
...
         self.state = GeneratorState.READY_TO_UPDATE
         self.cond = asyncio.Condition()
         self.policy_version = 0
 
-        # Reward function
-        self.reward_fn = (
-            math_reward_function if use_real_dataset else trivial_reward_function
-        )
+        # Reward function. TODO: Use a real reward function
+        self.reward_fn = trivial_reward_function
```
allenwang28: I guess the RL job definition would need to define a callable here, is this the idea for later?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -243,16 +217,36 @@ def generate(
...
+        logger.info(
+            f"Starting generation: {len(prompt_texts)} prompts, "
+            f"n_samples_per_prompt={n_samples_per_prompt}, "
+            f"max_tokens={max_new_tokens}, temp={temperature}"
+        )
+
         sampling_params = SamplingParams(
             temperature=temperature,
             max_tokens=max_new_tokens,
-            n=n_samples_per_prompt,
```
allenwang28, line 280: `n_samples_per_prompt` fulfills the same purpose as `group_size` I assume. I see below that we're preferring to submit a prompt multiple times instead of relying on vLLM. Is this due to batch invariance or something else? I'd assume that letting vLLM do it is better performance wise

torchtitan/experiments/rl/unified/infer.py
```diff
@@ -5,113 +5,104 @@
...
-        type=float,
-        default=0.8,
-        help="Sampling temperature",
-    )
-    parser.add_argument(
-        "--tensor-parallel-size",
-        type=int,
-        default=1,
-        help="Number of GPUs for tensor parallelism (default: 1 for single GPU)",
+def infer():
```
allenwang28:
```suggestion
def generate():
```

nit, but `infer()` makes me think of like getting the logits

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -78,37 +63,24 @@ async def main():
...
+    ]
+    prompt_texts = [p[0] for p in prompts_with_answers]
+    expected_answers = [p[1] for p in prompts_with_answers]
 
     logger.info(f"Loaded {len(prompt_texts)} prompts")
 
     # Create process meshes
-    trainer_mesh = this_host().spawn_procs(per_host={"gpus": 2})
+    trainer_mesh = this_host().spawn_procs(
+        per_host={"gpus": trainer_ddp_size * trainer_tp_size}
```
allenwang28, line 116: in the future we should deduce the total number of GPUs needed for a given trainer parallelism

allenwang28 (commented):
I like this direction, thanks! Mostly nits here

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: For this part, I will refactor this part into a Scorer or Grader later

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > n_samples_per_prompt fulfills the same purpose as group_size I assume 

You are right! Submitting multiple times is because I didn't find a good API call to send it to LLMEngine.add_request() 😅

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -134,88 +133,63 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
-            else:
-                # Ensure weights stay in bfloat16
-                vllm_state = {
-                    k: v.to(torch.bfloat16) if v.dtype == torch.float32 else v
-                    for k, v in vllm_state.items()
-                }
-                # Fallback: save as single file
-                save_file(vllm_state, checkpoint_path)
+
+            # Ensure weights stay in bfloat16
```
acisseJZhong: curious shall we remove this conversion? what if we want to keep some weights(for example router weights) in float32?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -134,88 +133,63 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
-        import json
 
         shard_files = sorted(
             glob.glob(os.path.join(self.temp_model_dir, "model-*.safetensors"))
         )
         index_file = os.path.join(self.temp_model_dir, "model.safetensors.index.json")
 
         # TODO: need to replace this with Torchtitan's checkpoint save and load
-        # right now we hardcoded to work with 2 safe tensor files which we only
+        # right now we hardcoded to work with 2 safetensor files which we only
```
acisseJZhong: update the comment to 1 safetensor file?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -65,19 +66,21 @@ class VLLMRolloutEngine:
     recreating temp dirs repeatedly and handles config/tokenizer files properly.
 
     Args:
+        job_config: JobConfig dataclass containing all configuration
```
acisseJZhong: can we take in generator_config here instead of job_config?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -134,88 +133,63 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
         # Synchronize all ranks before reloading to ensure rank 0 finished writing
         torch.distributed.barrier()
         logger.info(
             f"[Rank {torch.distributed.get_rank()}] Synchronized after weight save"
         )
 
-        # First time: create the engine
-        if self.llm is None:
-            self.llm = LLM(
+        # First time: create the engine using LLMEngine and EngineArgs
```
acisseJZhong, line 228: qq: what's the difference between using LLM() and LLMEngine()?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -243,16 +217,36 @@ def generate(
...
-            n=n_samples_per_prompt,
             seed=42,
             logprobs=1,
             prompt_logprobs=1,  # Also get prompt log probs to access prompt token IDs
+            output_kind=RequestOutputKind.FINAL_ONLY,  # Only return completed outputs
         )
 
-        outputs = self.llm.generate(prompt_texts, sampling_params)
+        # Add requests to the engine
+        # For n_samples_per_prompt > 1, submit each prompt multiple times with different request id
```
acisseJZhong: qq: why do we want to submit each requests multiple times with different request ids? Can we just loop through the n samples for each outputs, like what we did before?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -319,58 +318,46 @@ class Generator(Actor):
...
-        max_new_tokens: int = 20,
-        temperature: float = 1.0,
-        use_real_dataset: bool = False,
-        grpo_beta: float = 0.1,
-        use_stable_grpo: bool = False,
-        tp_size: int = 1,
     ):
-        self.model_path = model_path
+        # Store job_config for accessing configuration
+        self.job_config = job_config
```
acisseJZhong: nit: just remove this?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -78,37 +63,24 @@ async def main():
...
+    )
     gen_mesh = this_host().spawn_procs(per_host={"gpus": 1})
```
acisseJZhong: this should also consider generator dp size and tp size

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -163,6 +123,8 @@ async def main():
 
     for step in range(num_steps):
         # Fully sync RL loop
+        # NOTE: This is only getting Trajectory generated from trainer 0, and trainer 1's data is ignored.
```
acisseJZhong, line 169: you mean generator 0 and generator 1?

in fully sync RL loop, we should get all trajectory data from generator and pass it to trainer.

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+    parallelism: Parallelism = field(default_factory=Parallelism)
+    """Parallelism configuration for generation"""
+
+    sampling: Sampling = field(default_factory=Sampling)
+    """Sampling configuration for generation"""
+
+
+@dataclass
+class RL:
```
tianyu-l: RL sounds too general. Is there a narrowed term just for the policy optimization?

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+    gpu_memory_utilization: float = 0.5
+    """Fraction of GPU memory to use for generation engine (0.0 to 1.0)"""
+
+    distributed_executor_backend: str = "external_launcher"
+    """
+    Backend for distributed execution.
+    'external_launcher' means vLLM does not spawn processes (use torchrun/external launcher)
+    """
+    seed: int = 42
```
tianyu-l: shouldn't by default enable a fixed seed? I feel it is more for debug usage.

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,102 @@
...
+    dtype: str = "bfloat16"
+    """Data type for model weights (auto, float16, bfloat16, float32)"""
+
+    gpu_memory_utilization: float = 0.5
+    """Fraction of GPU memory to use for generation engine (0.0 to 1.0)"""
+
+    distributed_executor_backend: str = "external_launcher"
+    """
+    Backend for distributed execution.
+    'external_launcher' means vLLM does not spawn processes (use torchrun/external launcher)
```
tianyu-l: do we care other launcher? if not this shouldn't be configurable

torchtitan/experiments/rl/unified/README.md
```diff
@@ -66,10 +66,10 @@ python torchtitan/experiments/rl/unified/infer.py --model-ckpt-path <path_to_mod
 
 5. Run simple rl loop
 ```
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
+VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_grpo.py \
```
tianyu-l: we set some envvar outside in the command, and some others in code using os.environ -- can we unify them or have a guideline on when to use what?

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: Yes nice catch, add this refactored in a follow up PR

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: This is a constrain by vllm API, as  I didn't find a proper API to submit a request mutiple times. Previously the API has a n_groups pararmeter, which is equivalent to submitting the same request multiple times, so it genereates a "group" of completion for GRPO.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: I heard LLM() is for general users and LLMEngine() is more for developers. My understanding is LLMEngine has more fine-grainularity control we could use, while LLM() we can only pass configs

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Why we want to have a separate config class / system?

torchtitan/experiments/rl/unified/job_config.py
wwwjn: An alternative name would be "GRPO", since all the field now is related to GRPO. Or simply "PolicyOptimizatoin", it's more extensible for other algorithm like PPO

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Generation configuration for vLLM engine.
+
+    This dataclass contains essential vLLM-specific settings for generation.
+    """
+
+    dtype: str = "bfloat16"
+    """Data type for model weights (auto, float16, bfloat16, float32)"""
+
+    gpu_memory_utilization: float = 0.5
```
joecummings: Are you doing co-located Trainer / Generator?

torchtitan/experiments/rl/unified/job_config.py
joecummings: cc @JenniferWang for Monarch as potential backend

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@dataclass
+class RL:
+    """Reinforcement Learning configuration for GRPO training."""
+
+    grpo_beta: int = 0.1
+    """Beta parameter for GRPO (Group Relative Policy Optimization) algorithm"""
+
+    use_stable_grpo: bool = False
```
joecummings: To what is this referring?

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: This is from original line 169 and that's a good question, I didn't verified why we don't convert the weights.  This issue is further bypassed in later PR as I removed the "save to file path" eventually, so we can just ignore it for now?

torchtitan/experiments/rl/unified/job_config.py
wwwjn: This was inherited from `vlllm_compat/simple_rl.py`.  We can remove it

torchtitan/experiments/rl/unified/job_config.py
wwwjn: I think yes, they are on the same set of GPUs

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+    """Whether to use stable version of GRPO algorithm"""
+
+    grpo_group_size: int = 8
+    """Number of samples in each GRPO group for policy optimization"""
+
+    vllm_batch_invariant: bool = True
+    """Enable vLLM batch invariant mode for deterministic backward pass"""
+
+    vllm_attention_backend: str = "FLASH_ATTN"
+    """vLLM attention backend to use (e.g., FLASH_ATTN, XFORMERS)"""
```
acisseJZhong: should these two fields live under `Generate`?

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: ideally Rollout Engine should only take generation related config rather than the whole JobConfig? I am fine leaving this as is now, since it will change as part of @tianyu-l 's refactor anyways.

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+@dataclass
+class JobConfig(BaseJobConfig):
+    """
+    Extended JobConfig with generation support.
+
+    This extends TorchTitan's base JobConfig by adding `generation` field
+    for vLLM-specific generation configurations.
+    """
+
+    generation: Generate = field(default_factory=Generate)
```
acisseJZhong: nit: can we rename the class name to `Generation` as well?

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: wondering what's the reason to submit multiple requests, instead of using this `n_samples_per_prompt`? is it because we want to keep track of generations from the same prompt?

torchtitan/experiments/rl/unified/inference_example.py
```diff
@@ -5,120 +5,110 @@
...
+        # Seed
+        seed=job_config.debug.seed,
+        # HuggingFace overrides
+        hf_overrides={"architectures": ["Qwen3TorchTitanForCausalLM"]},
     )
 
-    logger.info("vLLM engine initialized successfully")
-    logger.info(f"Prompt: {args.prompt}")
+    logger.info("Initializing LLMEngine from EngineArgs...")
+    engine = LLMEngine.from_engine_args(engine_args)
```
acisseJZhong, line 69: just curious what's the difference between LLMEngine and LLM?

torchtitan/experiments/rl/unified/README.md
acisseJZhong: I think we can simply remove `VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN` as we set them in https://fburl.com/0zf82k46. We also explicitly set `os.environ["VLLM_WORKER_MULTIPROC_METHOD"]` to `spawn` in the same file.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -14,101 +14,55 @@
...
-        ]
-        prompt_texts = [p[0] for p in prompts_with_answers]
-        expected_answers = [p[1] for p in prompts_with_answers]
+    # Step 1: Load job config using config manager
+    config_manager = ConfigManager()
+    job_config = config_manager.parse_args()
+
+    # compute world size for trainer and generator
+    # TODO: refine the world size computation and check
+    trainer_ddp_size = job_config.parallelism.data_parallel_replicate_degree
```
acisseJZhong: should this be replicate degree * shard degree?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -243,16 +220,36 @@ def generate(
...
+        )
+
         sampling_params = SamplingParams(
             temperature=temperature,
             max_tokens=max_new_tokens,
-            n=n_samples_per_prompt,
             seed=42,
             logprobs=1,
             prompt_logprobs=1,  # Also get prompt log probs to access prompt token IDs
+            output_kind=RequestOutputKind.FINAL_ONLY,  # Only return completed outputs
```
tianyu-l: curious what's the default if we don't set explicitly?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -243,16 +220,36 @@ def generate(
...
         )
 
-        outputs = self.llm.generate(prompt_texts, sampling_params)
+        # Add requests to the engine
+        # For n_samples_per_prompt > 1, submit each prompt multiple times with different request id
+        request_id = 0
+        prompt_indices = []  # Track which prompt each request corresponds to
+        for prompt_idx, prompt in enumerate(prompt_texts):
+            for sample_idx in range(n_samples_per_prompt):
```
tianyu-l: why are we sending the same requests multiple times -- I assume it's for algorithms like GRPO?
But from @daniellepintz 's PR it seems this can be alternatively achieved by having multiple rollout engine? Then are these solutions biased in one way or another?

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+
+    parallelism: Parallelism = field(default_factory=Parallelism)
+    """Parallelism configuration for generation"""
+
+    sampling: Sampling = field(default_factory=Sampling)
+    """Sampling configuration for generation"""
+
+
+@dataclass
+class PolicyOptimization:
```
tianyu-l: This seems mixing GRPO and non algorithm-specific settings together, violating modularity and encapsulation.

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+
+    vllm_batch_invariant: bool = True
+    """Enable vLLM batch invariant mode for deterministic backward pass"""
+
+    vllm_attention_backend: str = "FLASH_ATTN"
+    """vLLM attention backend to use (e.g., FLASH_ATTN, XFORMERS)"""
+
+
+@dataclass
+class JobConfig(BaseJobConfig):
```
tianyu-l: Use proper name, this name is ambiguous

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Generation configuration for vLLM engine.
+
+    This dataclass contains essential vLLM-specific settings for generation.
+    """
+
+    dtype: str = "bfloat16"
+    """Data type for model weights (auto, float16, bfloat16, float32)"""
+
+    gpu_memory_utilization: float = 0.5
```
tianyu-l: bad name, sounds like MFU which is not something you can "set"

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+configurations needed for vLLM integration.
+"""
+
+from dataclasses import dataclass, field
+
+from torchtitan.config.job_config import JobConfig as BaseJobConfig, Parallelism
+
+
+@dataclass
+class Sampling:
```
tianyu-l: I wish there's better name. Sampling vs. Generate vs. Generation, sound interchangeable but actually you are using Generate to own Sampling.

tianyu-l: From how you use it, you can call it `VllmSamplingParams` as it's just copying the params to vllm args.

torchtitan/experiments/rl/unified/job_config.py
tianyu-l: The split of JobConfig.parallelism for trainer and JobConfig.Generate.parallelism for generator sounds not natural.

torchtitan/experiments/rl/unified/job_config.py
felipemello1: Maybe we can try to follow vllm standard? they call is [SamplingParams](https://github.com/vllm-project/vllm/blob/c683d11c94655655cd7bf95a27aef7e245325102/vllm/sampling_params.py#L118) and [vllmConfig](https://github.com/vllm-project/vllm/blob/c683d11c94655655cd7bf95a27aef7e245325102/vllm/config/vllm.py#L224)

That way we probably dont even have to think about var names and descriptions. It would be better if we could just import and use it directly if possible

tianyu-l: I think that depends how much trust and stickiness we would like to have with vllm naming. My stance is we don't want to be tied with how vllm name things.

torchtitan/experiments/rl/unified/job_config.py
```diff
@@ -0,0 +1,100 @@
...
+
+    sampling: Sampling = field(default_factory=Sampling)
+    """Sampling configuration for generation"""
+
+
+@dataclass
+class PolicyOptimization:
+    """Policy optimization configuration for GRPO training."""
+
+    grpo_beta: int = 0.1
```
felipemello1: nit: better as a loss arg (we can do in a follow up)

torchtitan/experiments/rl/unified/job_config.py
joecummings: Sglang also uses [SamplingParams](https://docs.sglang.io/basic_usage/sampling_params.html#id1) so I think it's fair to say that's sticking around.

torchtitan/experiments/rl/unified/actors/generator.py
joecummings: Multiple rollout engines are for routing even more requests to maximize throughput. 

In this case though, the recommended path would be to specify that each prompt returns `n` responses. This is possible through a param in the inference engine that I don't recall off the top of my head.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: To clarify, this is GRPO algorithm details:

- We generate `n` (group_size) independent completions for each prompt. Each completion is a full autoregressive rollout: the model runs forward, samples a token, appends it, and repeats. Because sampling is stochastic (temperature might > 0), each of the `n` rollouts produces a different sequence.
    -  I was manually implementing this by sending `n` separate requests.
    -  After taking a details look of vllm's `n_samples_per_prompt` implementation, I will switch back to using this parameter

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: `n_samples_per_prompt`? https://github.com/pytorch/torchtitan/pull/2191#discussion_r2658481750

torchtitan/experiments/rl/unified/job_config.py
wwwjn: I'm ok with either `SamplingParams` or `VllmSamplingParams`. Although sglang and vllm share the same fields in SamplingParams, but we are not possible to support sglang in short term

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: the default is RequestOutputKind.CUMULATIVE. In that mode, engine.step() returns partial outputs at every step as the sequence is being decoded

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: This is generate a group of `n` samples because of GRPO algorithm (G=Group). I changed to using n_samples_per_prompt

torchtitan/experiments/rl/unified/job_config.py
tianyu-l: SamplingParams is fine, if vllm's SamplingParams is complete for us, in the sense that we don't have to put fields not in vllm's SamplingParams into our SamplingParams

torchtitan/experiments/rl/unified/job_config.py
wwwjn: Sorry I correct my comments here - The trainer and generator are not collocated in this PR. But they will be collocated in later PR stack

torchtitan/experiments/rl/unified/job_config.py
wwwjn: I rename it to be `vllm_engine.gpu_memory_limit", to clear it's a limit user set, wdyt? This original name is from vllm

torchtitan/experiments/rl/unified/README.md
```diff
@@ -48,11 +48,10 @@ python torchtitan/experiments/rl/unified/infer.py --tensor-parallel-size 2
 ```
 
 5. Run simple rl loop
-```bash
-VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
 ```
-Right now we only support VLLM_COMPAT mode, which could achieve trainer and generator bitwise identical. We are working on support UNIFIED mode,
-which uses a unified model definition for trainer and generator.
+python3 torchtitan/experiments/rl/unified/simple_grpo.py --checkpoint.initial_load_path=<path_to_model_checkpoint>
```
wwwjn: TODO: Fix this command , should be --trainer.checkpoint.initial_load_path=<path_to_model_checkpoint>

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -58,28 +67,51 @@ class TrajectoryData:
...
+
     def __init__(
         self,
+        config: Config,
+        *,
         model_path: str,
-        temp_checkpoint_dir: str = "./converted",
-        tp_size: int = 1,
-    ):
+        dump_folder: str,
```
joecummings, line 119: It would appear this is a temporary directory to store a converted state dict in which case tmp_checkpoint_dir or something like that is much more intuitive than `dump_folder`. Regardless, a comment here would be great!

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -274,16 +274,31 @@ def generate(
             token_log_probs: List of per-token log prob lists for each completion
             prompt_token_ids: List of prompt token ID lists for each completion
         """
+        logger.info(
```
joecummings: nit: maybe keep this at debug level. It seems like @daniellepintz pointed out these can flood the terminal

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,69 +6,115 @@
...
     def __init__(
         self,
-        titan_checkpoint_path: str,
-        model_path: str,
-        learning_rate: float = 1e-5,
-        model_mode: str = ModelMode.VLLM_COMPAT,
-        ddp_size: int = 1,
-        tp_size: int = 1,
+        config: Config,
+        policy_optimization: PolicyOptimizationConfig,
```
joecummings: kwargs only after config position

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -0,0 +1,129 @@
...
+                initial_load_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
+                initial_load_model_only=True,
+                initial_load_in_hf=True,
+            ),
+            activation_checkpoint=ActivationCheckpointConfig(
+                mode="selective",
+                selective_ac_option="op",
+            ),
+        ),
+        batch_invariant_mode=True,
```
joecummings: This is only applied in generator therefore it should be part of the Generator config, not the top-level config

torchtitan/experiments/rl/unified/simple_grpo.py
joecummings, line 1: Following the changes in #2386, the overall format would look like:

rl/
```
actors/
-- generator.py
-- trainer.py
infra /
-- ...
models/
-- ...
grpo_trainer.py <-- defined training process, rollouts, etc
train.py <-- parses config, calls grpo_trainer.py
```

The issue here is that theres an aggressive overuse of the term "train/trainer" that can definitely cause confusion. 

Our "grpo_trainer.py" in this case is actually orchestrating the collecting of rollouts, passing data to the training model which then does backprop, and finally a weight sync at some time step. The most similar other framework to this design is [prime-rl](https://github.com/PrimeIntellect-ai/prime-rl/tree/main/src/prime_rl/orchestrator), who calls this process an "orchestrator". The conceptual separation here is specifically catered towards their goal of large-scale asynchronous reinforcement learning with disaggregated workers. 

Others like Verl leaned more heavily into the standard PPO regime of actors and critics under a synchronous context. In this context, the [RayPPOTrainer](https://github.com/verl-project/verl/blob/main/verl/trainer/ppo/ray_trainer.py) is the main entrypoint coordinating everything. The concept of trainer and generator are put into an `ActorRolloutRef` which wraps a `DataParallelPPOActor` (coordinates forward/backward), a `vLLMRollout` (inference), and an optional Reference Model. These are then called by the `RayPPOTrainer`.

Sigh, that's a lot and very confusing  I know. My preference would be to rename something similar to orchestrator.py and main.py that calls orchestrator which in turn uses trainer.py for forward/backward. This is most similar to prime-rl. However, this probably warrants a slightly longer conversation, so we can do a quick follow up. I don't want this sitting out for very long though b/c it is confusing for us to collaborate on and draw parallels.

torchtitan/experiments/rl/unified/actors/generator.py
joecummings: Bunch of these throughout generation

torchtitan/experiments/rl/unified/config_registry.py
wwwjn: My understanding is if we want to achieve true "on-policy" RL, both trainer / generator side would need to be replaced with batch-invariant kernels. (So that even batch size on trainer / generator are different, they will produce exactly the same logprob for each sample). So I made this a top-level config, this is a "mode" we want to switch on/off globally

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: Thanks for the discussion. When I was setting up the config, the name "trainer" is also confusing to me. The current design is following:
- RLTrainer: Now this is only a "config" class, works as top-level config for "RL training", not a real class like VeRL's [RayPPOTrainer](https://github.com/verl-project/verl/blob/main/verl/trainer/ppo/ray_trainer.py#L221C7-L221C20).  The actual orchestrating logic is in main() in simple_grpo.py. 
- PolicyTrainer: updates the policy model's weights (forward / backward). There's an actual class (OOD style) so the we can call `PolicyTrainer.Config.build()`. 

I'm a not into the "orchestrater" idea is that we only have simple sync RL for now, there's no "orchestrion" at this moment. But I don't have a strong opinion on the naming and config structure right now, as I'm still reading other frameworks.

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: would it make sense to place `dump_folder` and `model_path` in Config? 

I don't know which way is better, just wondering the rational one way vs the other

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -136,96 +164,68 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
-                vllm_state = {
-                    k: v.to(torch.bfloat16) if v.dtype == torch.float32 else v
-                    for k, v in vllm_state.items()
-                }
-                # Fallback: save as single file
-                save_file(vllm_state, checkpoint_path)
+
+            # Ensure weights stay in bfloat16
+            vllm_state = {
+                k: v.to(torch.bfloat16) if v.dtype == torch.float32 else v
```
acisseJZhong, line 217: can we leave a TODO in comment? to look into the case where some part of the weights (e.g. router) is fp32?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -136,96 +164,68 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
-                dtype="bfloat16",
-                gpu_memory_utilization=0.1,  # Reduced from 0.5
-                distributed_executor_backend="external_launcher",  # vllm do not spawn processes
-                seed=42,  # Fixed seed for determinism
-                enforce_eager=True,
-                tensor_parallel_size=self.tp_size,
+                dtype=cfg.dtype,
+                # Parallelism configuration
+                tensor_parallel_size=cfg.parallelism.tensor_parallel_degree,
+                distributed_executor_backend="external_launcher",
```
acisseJZhong, line 240: add comment?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -274,16 +274,31 @@ def generate(
...
             max_tokens=max_new_tokens,
             n=n_samples_per_prompt,
             seed=42,
             logprobs=1,
             prompt_logprobs=1,  # Also get prompt log probs to access prompt token IDs
+            output_kind=RequestOutputKind.FINAL_ONLY,  # Only return completed outputs
         )
 
-        outputs = self.llm.generate(prompt_texts, sampling_params)
+        # Add one request per prompt; vLLM handles n_samples_per_prompt via n=
```
acisseJZhong:
```suggestion
        # Add one request per prompt; vLLM handles n_samples_per_prompt via passing n=n_samples_per_prompt
```

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -350,58 +364,75 @@ class Generator(Actor):
...
+        """vLLM rollout engine configuration."""
+
+        vllm_attention_backend: str = "FLASH_ATTN"
+        """vLLM attention backend to use (e.g., FLASH_ATTN, XFORMERS)."""
+
     def __init__(
         self,
+        config: Config,
+        *,
+        model_spec: ModelSpec,
```
acisseJZhong: same, curious what is the Criteria for keep fields here vs in Config? Should we align with how the rest of torchtitan behave?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -0,0 +1,194 @@
...
+from torchtitan.config import Configurable
+from torchtitan.experiments.rl.unified.actors.generator import Generator
+from torchtitan.experiments.rl.unified.actors.trainer import PolicyTrainer
+from torchtitan.experiments.rl.unified.configs import PolicyOptimizationConfig
+from torchtitan.protocols.model_spec import ModelSpec
+
+logger = logging.getLogger(__name__)
+
+
+class RLTrainer(Configurable):
```
acisseJZhong, line 38: wondering shall we change this naming to not contain Trainer? as it's a bit confusing with PolicyTrainer, and since this config has everything: trainer, generator, etc

acisseJZhong:
thanks for making the changes and rebasing to main, overall it looks good to me! just have a few questions around naming and config structure.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -106,10 +102,16 @@ def __init__(
         # NOTE: We need to apply parallelize within model.__init__ because vllm
         # doesn't separate model creation and parallelism application and instead
         # requires parallelization to be done inside model constructor.
+        cfg = self.parallel_config
```
joecummings: This is the trainer config, not the parallel_config, no?

torchtitan/experiments/rl/unified/simple_grpo.py
joecummings: Thanks for the overview! I want to be super clear about my thinking here:

The pattern being set forward for *pre-training* in the config changes are that `train.py` is a launcher file that parses through a training config from some model-specific `config_registry.py` and uses that to launch a `Trainer` class from `trainer.py`. The model is then trained. 

One could make the argument that *post-training* within the same library should follow a similar pattern to present a cohesive library for *all* training. With this logic, things would look like the above folder structure I laid out and we'd need to address some of the shortcomings mentioned above as well as the fact that RL is a different beast from pretraining and sft. 

Let's get this stack landed along with @daniellepintz's PR, then I'll put up a PR later this week with these suggested changes so we can have a sandboxed discussion there.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -8,27 +8,36 @@
...
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import List
 
 import torch
 from monarch.actor import Actor, endpoint
 from safetensors.torch import save_file
-from torchtitan.config import CommConfig
+from torchtitan.config import CommConfig, Configurable
+from torchtitan.config.configs import ParallelismConfig
```
tianyu-l: nit: CommConfig and ParallelismConfig should come from same path

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -58,28 +67,51 @@ class TrajectoryData:
...
+        dtype: str = "bfloat16"
+        """Data type for model weights (auto, float16, bfloat16, float32)."""
+
+        gpu_memory_limit: float = 0.5
+        """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
+
+        enforce_eager: bool = True
+        """Disable CUDA graphs in vLLM (use eager execution)."""
+
+        seed: int = 42
```
tianyu-l: shouldn't have a default, unless user explicit set?

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l, line 217: In VLLMEngine.Config you have a dtype field. Is it not for this cast to use?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -136,96 +164,68 @@ def update_weights(self, vllm_compat_state: dict) -> None:
...
             # Direct parameter copy into model tensors.
             # This bypasses vLLM's reload_weights() which uses a layerwise
             # reload mechanism that moves params to meta device
-            from torchtitan.experiments.rl.vllm_compat.weights_vllm_compat import (
-                vllm_compat_to_torchtitan,
-            )
+            from torchtitan.experiments.rl.vllm_compat.weights import vllm_to_torchtitan
 
-            titan_state = vllm_compat_to_torchtitan(vllm_compat_state)
+            titan_state = vllm_to_torchtitan(vllm_state)
```
tianyu-l, line 263: maybe leave a TODO on renaming `vllm_to_torchtitan` to hint that it's for state dict conversion

But actually, I wonder why we don't use our state_dict adapter / checkpointer utils for this conversion.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -111,23 +143,19 @@ def __init__(
         if os.path.exists(index_file):
             shutil.copy2(index_file, self.temp_model_dir)
 
-        self.llm = None
-        self.tp_size = tp_size
+        self.engine = None
```
tianyu-l: why `VLLMEngine` would own another `engine`?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -111,23 +143,19 @@ def __init__(
...
         logger.info("vLLM rollout engine initialized (will load on first use)")
 
-    def update_weights(self, vllm_compat_state: dict) -> None:
+    def update_weights(self, vllm_state: dict) -> None:
```
tianyu-l: - what is a vllm_state?
- when is this used?
- why do we have both `update_weights` and `_direct_weight_update`?

needs to add more comments if they are justified

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -350,58 +364,75 @@ class Generator(Actor):
...
+        self.grpo_beta = policy_optimization.beta
+        self.use_stable_grpo = policy_optimization.use_stable_grpo
 
         # Initialize distributed environment for SPMD generator
-        world_size = dist_utils.init_distributed(
-            CommConfig(),
+        world_size = dist_utils.init_distributed(CommConfig())
+
+        # Build vLLM engine
+        self.vllm_engine = VLLMEngine(
```
tianyu-l: should use `config.vllm_engine.build()`

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -350,58 +364,75 @@ class Generator(Actor):
...
+        policy_optimization: GRPO hyperparameters.
+        prompt_texts: List of prompt strings.
+        expected_answers: List of expected answer strings.
     """
 
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        """Generator actor configuration."""
+
+        vllm_engine: VLLMEngine.Config = field(default_factory=VLLMEngine.Config)
```
tianyu-l: need to document the rationale of doing Generator -> VLLMEngine -> LLMEngine, especially the necessity of VLLMEngine layer.

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,69 +6,115 @@
...
     Updates policy based on collected trajectories.
 
     Run model forward on trajectories, computes loss, and run backward.
+    Receives the top-level ``RLTrainer.Config`` and reads policy trainer
+    settings (batch_invariant_mode, grpo) directly from it, plus model /
+    optimizer / parallelism settings from the nested ``config.trainer``.
+
+    TODO: Use torchtitan Trainer for model init and parallelisation.
+
+    TODO: Use torchtitan PolicyTrainer for model init and parallelism.
```
tianyu-l: why two similar TODOs?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,69 +6,115 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-class Trainer(Actor):
+class PolicyTrainer(Actor, Configurable):
     """
     Updates policy based on collected trajectories.
 
     Run model forward on trajectories, computes loss, and run backward.
```
tianyu-l: did we agree on removing "trajectories" everywhere?

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -112,86 +91,38 @@ def replace_with_vllm_compatible_flash_attention(model):
...
-        model = Qwen3Model(model_args)
-        # Load standard TorchTitan format directly
-        model.load_state_dict(state_dict, strict=False)
+    if model_args.enable_weight_tying and "output.weight" not in titan_state_dict:
+        titan_state_dict["output.weight"] = titan_state_dict["tok_embeddings.weight"]
+
+    model = model_args.build()
+    # Set global default dtype to bfloat16. This is needed because vLLM's Attention
+    # layer uses torch.get_default_dtype() and it doesn't support float32
+    torch.set_default_dtype(torch.bfloat16)
```
tianyu-l: hmm, trainer also can only use torch.bfloat16 master weight?

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -112,86 +91,38 @@ def replace_with_vllm_compatible_flash_attention(model):
     )
 
 
-def load_model(
-    checkpoint_path: str, model_path: str, model_mode: str = ModelMode.VLLM_COMPAT
-):
+def load_trainer_model(model_path: str, model_config: BaseModel.Config):
```
tianyu-l: maybe?
```suggestion
def load_trainer_model_from_vllm_ckpt(vllm_model_ckpt_path: str, model_config: BaseModel.Config):
```

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -112,86 +91,38 @@ def replace_with_vllm_compatible_flash_attention(model):
...
-        hidden_dim=hf_config.intermediate_size,
-        norm_eps=hf_config.rms_norm_eps,
-        rope_theta=hf_config.rope_theta,
-        max_seq_len=getattr(hf_config, "max_position_embeddings", 32768),
-        qk_norm=True,
-        depth_init=True,
-        eos_id=getattr(hf_config, "eos_token_id", 151645),
-        enable_weight_tying=getattr(hf_config, "tie_word_embeddings", False),
-    )
+    model_args = model_config
```
tianyu-l: unnecessary cast

torchtitan/experiments/rl/unified/inference_example.py
tianyu-l, line 1: maybe rename to inference_example.py because you fixed a debug model and the prompt

torchtitan/experiments/rl/unified/actors/trainer.py
wwwjn: Rebase issue, removed

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: Sure, I think the structure you proposed makes sense and it's more clear, I'm too lazy to implement full OOD for GRPOTrainer class in this PR but let's refactor later
```
grpo_trainer.py <-- defined training process, rollouts, etc
train.py <-- parses config, calls grpo_trainer.py
```

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > wonder why we don't use our state_dict adapter / checkpointer utils for this conversion.

This is achieved by the 3rd PR in this stack (which is focusing on removing fqn conversion & saving to file during weight transfer). In this PR I tried to focus on Config system changes and make each PR easy to read

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: The rational here is just `dump_folder` is canonical and set by uppert-level configs, so I passed it from upper-level class. We want to keep the `dump_folder` here aligned with upper-level class's config without setting them twice, or hardcode it to be the same wight upper-level config, which makes it not really configurable

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Sure, this is inherited from previous code, I will invest why we need to keep vllm's weight in bf16

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: I don't have a better name in my mind now, some alternative is like GRPOTrainer(). I used `Trainer` in the name as the main torchtitan pre-train entry point is also a `Trainer`

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Good point, cleaned up

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: I passed it here because it was set by upper-level class' config (here the upper-level class is RLTrainer)

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: Nice catch, fixed!

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: This dtype conversion is from original code, and `VLLMEngine.Config.dtype` field is used to set EngineArgs(dtype=...), which is also `bfloat16`.  This latter fields controls the weight dtype of vllm model.

My understanding is that, vLLM model's weights are usually set to bfloat16, because: 1) FA kernels only takes bfloat16 inputs, and 2) the kv_cache needs to be in bfloat16/float16 (kv_cache does not support float32 because it takes too much memory). 


Once we have control of the forward path, we can control weights / activation dtype in a finer-granularity way, but we will need to figure-out where to change it to bfloat16 so it's compatible with vllm's kv cache mechanism.  Now I just simply set generator model's weights to be bfloat16 (and trainer's weights to be bfloat16 when checking bit-wise identity) to bypass this issue, which need more careful investigation of inner vllm

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > what is a vllm_state

This is vllm fqn based state dict. I added a Args to show the contents of this dictionary

> when is this used?

This function is used to update weights passed from trainer. Updated the function docstring

> why do we have both update_weights and _direct_weight_update?

`_direct_weight_update` is a temporary solution the vllm's `reload_weights` function call is causing problem.  I removed it in the 3rd PR in this stack when refactoring the weight transfer. I can also remove it now in this PR

torchtitan/experiments/rl/unified/actors/trainer.py
wwwjn: Yes we agreed on the renaming on all components, but the renaming happens in 4st PR in this stack :(  #2244. Maybe it's a bad practice to separate into several PRs, but I want to keep each PR relatively easier to read for people who don't have full context like you

torchtitan/experiments/rl/unified/models/utils.py
wwwjn: This is more for bit-wise identity check (as vllm is using bloat16 weights because of the reason [here](https://github.com/pytorch/torchtitan/pull/2191#discussion_r2851674098) , I remove it when bit-wise identity mode is not enabled

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: I still don't see why they wouldn't agree.

When engine doesn't exist, you use `EngineArgs` to create one, with VLLMEngine.Config.dtype.
Here the state dict is from PolicyTrainer to update the existing engine, so has to align with what's set above.

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: > This is vllm fqn based state dict.

- who is calling this function and why it is sending a vllm fqn based state dict?
- In unified model, why we still have differentiation between vllm fqn and titan fqn? (The initial checkpoint load from HF model should be handled by trainer, not generator)

torchtitan/experiments/rl/unified/actors/trainer.py
tianyu-l: that's fine, I think the problem is we started from somewhere not clean

torchtitan/experiments/rl/unified/models/utils.py
tianyu-l: since RL involves complex calling stack, in the future I would recommend always adding the expected caller in docstring

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -112,86 +91,38 @@ def replace_with_vllm_compatible_flash_attention(model):
...
+    # layer uses torch.get_default_dtype() and it doesn't support float32
+    torch.set_default_dtype(torch.bfloat16)
+    # NOTE: Override attention to vllm compatible attention for backward capability.
+    # Only patch to vllm compatible attention during training.
+    replace_with_vllm_compatible_flash_attention(model)
+
+    # Load standard TorchTitan format directly
+    model.load_state_dict(titan_state_dict, strict=True)
 
     model.to(torch.bfloat16)
```
tianyu-l: The dtype conversion story seems not clear, maybe leave a TODO at those conversion sites.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: For the layered structure, the current rationale is:
- LLMEngine: vllm's class 
- VLLMEngine: Shallow wrapper of `LLMEngine`, handles engine creation, model weights updating (handels wieght conv, create sampling configs, 
- Generator: handle's the state machine in RL. <- This part is achieved by a Conditional variable (`self.cond` now.)

I think we can remove the "state machine" in Generator, and squash the 2-layer wrapper into 1 layer (Generator-> LLMEngine). The generator should not care about "orchestrating", and as we are using monarch's blocking calls (`.get()`) so we don't actually manage a conditional variable and call notify_all() manually.  cc @acisseJZhong @allenwang28

tianyu-l: Sounds good. Maybe we can call this condensed layer VLLMGenerator, because it's a generator coupled with vllm engine. (It doesn't have to, but you used VLLMSamplingParams, if you are OK renaming it to SamplingParams, we can call it Generator which could have platform-agnostic `self.engine`)

PR #2194, opened 2026-01-02
[rl] refactor torchtitan model registery in vllm

wwwjn:
* #2244
* #2221
* __->__ #2194
* #2191

### DIFF
--- a/torchtitan/experiments/rl/unified/__init__.py
+++ b/torchtitan/experiments/rl/unified/__init__.py
@@ -13,16 +13,16 @@ To manually load the plugin:
 """
 
 from torchtitan.experiments.rl.unified.models.vllm_wrapper import (
-    create_torchtitan_config_from_vllm,
     TorchTitanVLLMModelWrapper,
 )
 
 # Export plugin register function for manual use (no auto-registration)
-from torchtitan.experiments.rl.unified.plugin import register
+from torchtitan.experiments.rl.unified.plugin import (
+    register_model_to_vllm_model_registry,
+)
 
 
 __all__ = [
     "TorchTitanVLLMModelWrapper",
-    "create_torchtitan_config_from_vllm",
-    "register",  # Export register function for manual use
+    "register_model_to_vllm_model_registry",  # Export register function for manual use
 ]
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -123,9 +123,13 @@ class Generator(Actor, Configurable):
         self.model_spec = model_spec
 
         # Register TorchTitan model with vLLM before any engine creation
-        from torchtitan.experiments.rl.unified.plugin import register
+        from torchtitan.experiments.rl.unified.plugin import (
+            register_model_to_vllm_model_registry,
+            VLLM_MODEL_NAME,
+        )
 
-        register(model_spec)
+        register_model_to_vllm_model_registry(model_spec)
+        self._vllm_model_name = VLLM_MODEL_NAME
 
         # Set vLLM environment variables from config before any vLLM initialization
         if batch_invariant_mode:
@@ -245,9 +249,8 @@ class Generator(Actor, Configurable):
                 enforce_eager=cfg.enforce_eager,
                 # Seed
                 seed=cfg.seed,
-                # HuggingFace overrides to use TorchTitan model.
-                # TODO: make this field configurable and align with model registration
-                hf_overrides={"architectures": ["Qwen3TorchTitanForCausalLM"]},
+                # HuggingFace overrides to use registered TorchTitan model
+                hf_overrides={"architectures": [self._vllm_model_name]},
                 attention_config=AttentionConfig(
                     backend=AttentionBackendEnum.FLASH_ATTN,
                 ),
--- a/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
@@ -22,15 +22,12 @@ from torch.distributed.checkpoint.state_dict import (
     StateDictOptions,
 )
 
-from torchtitan.config import ParallelismConfig, TrainingConfig
+from torchtitan.config import ParallelismConfig
 from torchtitan.distributed.parallel_dims import ParallelDims
 from torchtitan.experiments.rl.unified.models.attention import (
     replace_with_vllm_attention,
 )
-from torchtitan.protocols.model import BaseModel
-from torchtitan.protocols.model_spec import ParallelizeFunction
-from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
-from torchtitan.trainer import Trainer
+from torchtitan.protocols.model_spec import ModelSpec
 
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
@@ -39,14 +36,14 @@ from vllm.logger import init_logger
 logger = init_logger(__name__)
 
 
-def create_torchtitan_config_from_vllm(
+def create_torchtitan_config_from_vllm_config(
     vllm_config: VllmConfig,
-) -> tuple[ParallelDims, Trainer.Config]:
+) -> tuple[ParallelDims, ParallelismConfig]:
     """
-    Create ParallelDims and Trainer.Config from vLLM configuration.
+    Create ParallelDims and ParallelismConfig from vLLM configuration.
 
-    Maps vLLM parallelism settings to TorchTitan's ParallelDims and Trainer.Config
-    so that TorchTitan's parallelize functions can be called with the correct kwargs.
+    Maps vLLM parallelism settings to TorchTitan's config objects so that
+    TorchTitan's parallelize functions can be called with the correct kwargs.
 
     This is needed because vLLM doesn't separate model creation and parallelism
     application — it requires parallelization inside the model constructor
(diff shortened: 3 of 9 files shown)


torchtitan/experiments/rl/unified/__init__.py
```diff
@@ -7,89 +7,27 @@
...
-
-    # Register with vLLM
-    ModelRegistry.register_model(model_name, TorchTitanVLLMModelFromSpec)
-
-    logger.info(
-        f"Successfully registered {model_name} with vLLM using TrainSpec "
-        f"(model_cls={train_spec.model_cls.__name__}, flavor={model_flavor})"
-    )
+The plugin is loaded at runtime when needed (in generator.py), following
+the pattern used in native_sampler.py.
```
tianyu-l: where is native_sampler.py?

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+
+
+def register():
+    """
+    Register TorchTitan models with vLLM.
+
+    This function registers TorchTitan models (e.g., Qwen3) with vLLM's ModelRegistry.
+    It's called automatically when the unified module is imported, or can be called
+    explicitly for manual registration.
```
tianyu-l: why do we need the explicit way -- explicitly calling it would require importing in the first place, which would have already register it?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -166,6 +166,12 @@ def update_weights(self, vllm_state: dict) -> None:
 
         # First time: create the engine using LLMEngine and EngineArgs
         if self.engine is None:
+            # Load TorchTitan plugin at runtime
+            from torchtitan.experiments.rl.unified.plugin import register
+
+            register()
```
acisseJZhong: nit: can we use plugin.register()? to follow the docstring in the next file, or update the docstring

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Create dynamic model class directly from TrainSpec components
+        class TorchTitanVLLMModelFromSpec(TorchTitanVLLMModelWrapper):
+            def __init__(self, *, vllm_config, prefix=""):
+                super().__init__(
+                    model_cls=train_spec.model_cls,
+                    model_args=model_args,
+                    state_dict_adapter=train_spec.state_dict_adapter,
+                    # NOTE: This should be replaced with qwen3 parallelization plan in torchtitan core
+                    parallelize_fn=parallelize_qwen3,
```
acisseJZhong: should the parallelize function from train_spec?

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,88 @@
...
+                    vllm_config=vllm_config,
+                    prefix=prefix,
+                )
+
+        # Set the class name
+        TorchTitanVLLMModelFromSpec.__name__ = model_name
```
acisseJZhong: will the model always be the same for different flavors? .e.g for qwen3  model some is dense vs MoE

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+    # Register TorchTitan models
+    # TODO: Make this configurable instead of hardcoded
+    register_torchtitan_model_from_train_spec(
+        train_spec=get_train_spec("qwen3"),
+        model_name="Qwen3TorchTitanForCausalLM",
+        # TODO: Remove the model_flavor args when registering model,
+        # allow passing model flavor option from config system. Now we have to specify
+        # model_flavor during registration because we can not pass torchtitan job_config from LLM() Api
```
acisseJZhong: why we cannot take job_config here?

torchtitan/experiments/rl/unified/__init__.py
```diff
@@ -7,89 +7,24 @@
...
-# Auto-register TorchTitan models with vLLM when this module is imported
-register_torchtitan_model_from_train_spec(
-    train_spec=get_train_spec("qwen3"),
-    model_name="Qwen3TorchTitanForCausalLM",
-    # TODO: Remove the model_flavor args when registering model,
-    # allow passing model flavor option from config system. Now we have to specify
-    # model_flavor during registration because we can not pass torchtitan job_config from LLM() Api
-    model_flavor="0.6B",
+from torchtitan.experiments.rl.unified.infra.parallelism_utils import (
+    create_parallel_dims_from_vllm_config,
```
tianyu-l: I hope we could put all torchtitan-vllm glue code in one file / folder, and carefully document why we need each class / method. This one sounds one of them.

torchtitan/experiments/rl/unified/infer.py
```diff
@@ -39,6 +35,12 @@ def generate():
     config_manager = ConfigManager()
     job_config = config_manager.parse_args()
 
+    # Load TorchTitan plugin at runtime (like native_sampler does)
+    from torchtitan.experiments.rl.unified.plugin import register
+
+    register()
```
tianyu-l: I might have lost context, but the split of files into infer.py and actors/generator.py, both needing to register vllm model is confusing.

wwwjn: `infer.py` is separate from RL loop, and it's only for inferencing using vllm + torchtitan model, so we will need to register separately

torchtitan/experiments/rl/unified/README.md
```diff
@@ -44,8 +44,7 @@ torchrun --nproc_per_node=<world_size> \
 
 5. Run simple rl loop
 ```
-python3 torchtitan/experiments/rl/unified/simple_grpo.py \
-    --trainer.checkpoint.initial_load_path=<path_to_model_checkpoint>
```
wwwjn, line 48: NOTE: This CLI parse haven't been supported in this PR. It will be supported in next PR.

torchtitan/experiments/rl/unified/plugin.py
wwwjn: Yes, the model will be registered to vllm during RL runtime. As we only run one model at a time, the name is not conflicting in vLLM

torchtitan/experiments/rl/unified/plugin.py
wwwjn: Rebased!

torchtitan/experiments/rl/unified/__init__.py
wwwjn: Nice catch, refactored this part

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,64 @@
...
+    model_config = model_spec.model
+
+    # Create dynamic model class capturing ModelSpec components in the closure
+    class TorchTitanVLLMModelFromSpec(TorchTitanVLLMModelWrapper):
+        def __init__(self, *, vllm_config, prefix=""):
+            super().__init__(
+                model_config=model_config,
+                state_dict_adapter=model_spec.state_dict_adapter,
+                # NOTE: This should be replaced with qwen3 parallelization plan in torchtitan core
+                parallelize_fn=parallelize_qwen3,
```
tianyu-l: this can come from model_spec

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,64 @@
...
+    from torchtitan.experiments.rl.unified.plugin import register
+    register(model_spec)
+"""
+
+from torchtitan.protocols.model_spec import ModelSpec
+
+
+def register(
+    model_spec: ModelSpec,
+    model_name: str = "Qwen3TorchTitanForCausalLM",
```
tianyu-l: If people don't need to customize this, it can be a constant model agnostic string, or come from a string transformation from model_spec name and flavor

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,64 @@
...
+
+Usage:
+    from torchtitan.experiments.rl.unified.plugin import register
+    register(model_spec)
+"""
+
+from torchtitan.protocols.model_spec import ModelSpec
+
+
+def register(
```
tianyu-l: let's rename this to show more what it's doing, e.g. `register_model_to_vllm_model_registry`

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -38,6 +39,67 @@
 logger = init_logger(__name__)
 
 
+def create_torchtitan_config_from_vllm(
```
tianyu-l: I don't mind it to be more explicit, given there are so many various interactions.
```suggestion
def create_torchtitan_config_from_vllm_config(
```

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -38,6 +39,67 @@
...
+        dp_shard=1,
+        cp=parallel_config.decode_context_parallel_size,
+        tp=parallel_config.tensor_parallel_size,
+        pp=parallel_config.pipeline_parallel_size,
+        ep=1,
+        etp=1,
+        world_size=world_size,
+    )
+
+    config = Trainer.Config()
```
tianyu-l: why do you need to return whole default Trainer config with only parallelism and training being replaced? Could you return parallelism and training?

wwwjn: Nice catch, this is old behavior from JobConfig, changed.

PR #2198, opened 2026-01-05
[rl] Use vllm.Attention for trainer.

zhxchen17:
* __->__ #2198
* #2200


torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -42,7 +51,198 @@ class ModelMode(str, Enum):
...
+                    v=v,
+                    cu_seqlens_q=attn_metadata.query_start_loc,
+                    cu_seqlens_k=attn_metadata.query_start_loc,
+                    max_seqlen_q=attn_metadata.max_query_len,
+                    max_seqlen_k=attn_metadata.max_seq_len,
+                    softmax_scale=scale,
+                    causal=attn_metadata.causal,
+                    alibi_slopes=alibi_slopes,
+                    window_size=window_size,
+                    block_table=None,  # TODO Revisit if PagedAttention is really needed.
```
wwwjn, line 98: In original vllm attention implementation, (eg, vllm.Attention()), how does they create, store and pass the block_table information? Is it using forward context for some other method?

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -42,7 +51,198 @@ class ModelMode(str, Enum):
...
+                alibi_slopes,
+                window_size,
+                softcap,
+                s_aux,
+            ):
+                assert attn_metadata.num_actual_tokens == q.size(0)
+                # TODO Check if this is different from the state returned from FA fwd.
+                rng_state = torch.cuda.get_rng_state(device=q.device)
+
+                output, softmax_lse = flash_attn_varlen_func(
```
wwwjn, line 86: Curious why we don't use vllm.Attention() directly (as the only code path), as we created dummy metadata / forward context in utils.py

PR #2202, opened 2026-01-05
Add ROCm support for H100 tests

akashveramd:
This PR adds ROCm support for H100 tests.

### DIFF
--- a/.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
+++ b/.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
@@ -7,6 +7,7 @@ on:
       - ciflow/8gpu/*
     paths:
       - 'torchtitan/experiments/transformers_modeling_backend/**'
+      - '.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml'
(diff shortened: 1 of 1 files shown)


tianyu-l:
existing tests are failing, can we solve them first?
https://github.com/pytorch/torchtitan/actions/runs/20724721636

akashveramd:
> existing tests are failing, can we solve them first? https://github.com/pytorch/torchtitan/actions/runs/20724721636

It seems ROCm runner ran out of disk space. I've shared this issue with our runner team.

akashveramd:
The H100 tests are working for ROCm and CI for ROCm is enabled. However, we also wanted to change the name of the H100 test to something generic that goes with ROCm as well. The reason the tests were named H100 because there are some features that's only supported on H100, including async TP with symmetric memory, Float8 quantization. Hence the test name is H100.

As of now ROCm supports FP8 quant but does not support sync TP with symmetric memory. We are working on supporting sync TP with symmetric memory for ROCm. Hence, moved the PR to draft. Once the support is added we'll open the PR again and think about changing the test name.
cc: @tianyu-l @huydhn

PR #2204, opened 2026-01-06
Disable dynamo LRU cache when AC is enabled 

soulitzer:
Previously this config is flipped whenever compile is enabled. This PR updates it so that the config is flipped whenever AC is applied, so that the only flex compiled case is also covered. 

Code comment:
```
# Disable dynamo LRU cache to workaround an interaction between SAC, PP, and Flex:
#
# When forward runs with a second PP microbatch, it triggers recompilation with dynamic
# shapes enabled. Now there are two valid compiled graphs. By default, dynamo selects
# the latest one (the dynamic shapes version), so the runtime wrapper expects an extra
# symint output. When SAC caches the inductor HOP output from the static graph for
# batch_idx=0, it would miss that symint and cause an assertion failure. The workaround
# here is to disable the LRU cache, and select graphs in insertion order instead.
#
# Also see: https://github.com/pytorch/pytorch/issues/166926
```

### DIFF
--- a/torchtitan/distributed/activation_checkpoint.py
+++ b/torchtitan/distributed/activation_checkpoint.py
@@ -210,6 +210,19 @@ def apply_ac(
     Returns:
         None
     """
+    # Disable dynamo LRU cache to workaround an interaction between SAC, PP, and Flex:
+    #
+    # When forward runs with a second PP microbatch, it triggers recompilation with dynamic
+    # shapes enabled. Now there are two valid compiled graphs. By default, dynamo selects
+    # the latest one (the dynamic shapes version), so the runtime wrapper expects an extra
+    # symint output. When SAC caches the inductor HOP output from the static graph for
+    # batch_idx=0, it would miss that symint and cause an assertion failure. The workaround
+    # here is to disable the LRU cache, and select graphs in insertion order instead.
+    #
+    # Also see: https://github.com/pytorch/pytorch/issues/166926
+    # pyrefly: ignore [missing-attribute]
+    torch._C._dynamo.eval_frame._set_lru_cache(False)
+
     if ac_config.mode == "memory_budget":
         assert model_compile_enabled, "Memory budget mode requires model to be compiled"
         if ac_config.visualize_memory_budget_pareto:
--- a/torchtitan/models/llama4/infra/parallelize.py
+++ b/torchtitan/models/llama4/infra/parallelize.py
@@ -592,9 +592,6 @@ def apply_compile(model: nn.Module, compile_config: CompileConfig, ep_enabled: b
     # NOTE: This flag is needed for torch.compile to avoid graph breaking on dynamic shapes in token-choice MoE
     # but it is experimental.
     torch._dynamo.config.capture_scalar_outputs = True
-    # Workaround for https://github.com/pytorch/pytorch/issues/166926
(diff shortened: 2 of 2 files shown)


torchtitan/models/deepseek_v3/infra/parallelize.py
```diff
@@ -132,6 +132,17 @@ def parallelize_deepseekv3(
...
+        #
+        # When forward runs with a second PP microbatch, it triggers recompilation with dynamic
+        # shapes enabled. Now there are two valid compiled graphs. By default, dynamo selects
+        # the latest one (the dynamic shapes version), so the runtime wrapper expects an extra
+        # symint output. When SAC caches the inductor HOP output from the static graph for
+        # batch_idx=0, it would miss that symint and cause an assertion failure. The workaround
+        # here is to disable the LRU cache, and select graphs in insertion order instead.
+        #
+        # Also see: https://github.com/pytorch/pytorch/issues/166926
+        torch._C._dynamo.eval_frame._set_lru_cache(False)
```
fegin: What's the downside of always setting this to False?

soulitzer: hmm from the original issue https://github.com/pytorch/pytorch/issues/166926 it is mentioned that it would increase cache look up time, but I don't understand the details completely
maybe @williamwen42 @xmfan have more context

xmfan: When there's recompiles, there are many valid graphs for one compiled function. Each time the function is called, dynamo needs to evaluate each graph's guards in order. If you evaluate them in LRU order, you might be more likely to get an early cache hit.

xmfan:
Why does the setting need to be applied when compiled is not turned on?

If compiled is enabled, then it should be covered by: https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama4/infra/parallelize.py#L597

fegin:
@xmfan Compile is not turned one. But FlexAttention is already compiled regardless the compile flag.

torchtitan/models/deepseek_v3/infra/parallelize.py
xmfan: In the original report, this issue affected MoEs only (as dense was static shaped). In the FlexAttention case, would it also only for MoE? If so, you probably want a centralized apply_ac function inside of torchtitan similar to apply_compile to group MoE logic.

Alternatively, we can consider making flex's compile explicit in [the config system: compile.components](https://github.com/pytorch/torchtitan/blob/795a7a027eccf282c51a5ae1cc0c1c3459120c9b/torchtitan/models/llama4/train_configs/debug_model.toml#L71)

soulitzer: Not sure it is MoEs only, the dynamism is coming from mask_mod closing over the PP batch idx so it should happen whenever we are using PP and Flex? https://github.com/pytorch/pytorch/pull/164111

xmfan: Sounds like it doesn't only affect DSv3, so you'll need to add it to the other models

fegin: TorchTitan already has a centralized `apply_ac`.

fegin (approved):

Issue #1842, opened 2025-10-09
More robust distributed loss


lkhphuc:
From the discussion here: https://github.com/pytorch/torchtitan/pull/1803#discussion_r2415190883

Summary: currently, the loss is computed locally per rank, then FSDP average gradient across all ranks. This works for simple case where all ranks have the same amount of valid tokens, but breakdown in more complex scenarios:
- VLM training with native resolution. #1803 
- Expert Parallelism https://github.com/pytorch/torchtitan/pull/1793/files#r2415207041

Outline solution:
> - don't let FSDP do implicit gradient division
> - always run cross entropy with reduction="sum"
> - let data loader / trainer count the number of tokens involving in loss computation, e.g. by explicitly doing num_tokens = (labels != IGNORE_INDEX).sum() on each rank. (I agree that without imbalance we don't need to do this and the followed communication.)

wwwjn:
Loss computation needs to be correct when:
1. Each rank sees unbalanced number of tokens (eg, in SFT, or VLM, where some `labels` are marked as `IGNORE_INDEX` and should be masked out during loss calculation 
2. Gradient accumulation is enabled: Each gradient accumulation step might seen different number of tokens for loss computation. Now we are just torch.sum() all the loss from each gradient accumulation step


torchtune solution: 
> In tune we solved it by rescaling the gradients later (image below). This works fine, the issue is that scaling the gradients is not trivial with compile, and we need this function to scale them and keep it compile friendly: https://github.com/meta-pytorch/torchtune/blob/44271b570af36cfda8ee20a4479d2652770378c0/torchtune/training/_grad_scaler.py#L82

tianyu-l:
@wwwjn For DDP it's not fixed. @anshul-si is working on it.

wwwjn:
Right, there's a new issue tracking this #2318 , let me close this one because the other one have better context

PR #2206, opened 2026-01-07
Fix loss computation by handling valid token imbalance in train loop

wwwjn:
Solve #1842

The previous implementation used per-rank/per-microbatch normalization, which caused incorrect gradient weighting when different ranks or microbatches had different numbers of valid tokens. This PR implements global token-based loss normalization to ensure each valid token contributes equally to the final loss, regardless of which rank or microbatches it appears in. 

## What changed
1. cross_entropy_loss always use "sum" reduction instead of "mean" reduction.
2. Pre-fetch all the microbatches when PP , sum all the valid tokens (global_valid_tokens) across DP ranks before gradient accumulation start. 
3. Use the global_valid_tokens to rescale loss in train loop , so each token has *equal* contribution to loss value.
4. When PP is applied: use sum cross entropy loss, rescale and handle imbalance between pp-microbatchs in train loop
5. Change `validation.py`: for validation, only need to handle imbalance tokens between ranks

NOTE: This change will change the loss value for integration tests, as we are back-propagating with loss = local_rank_loss_sum / global_valid_tokens (in contrast with previously, we were back-propagating using loss = local_rank_loss_sum / local_total_tokens). 

The convergence speed / gradient will be the same because we disabled the AVG of gradient in FSDP. 

Cons:
- This change would incur one all_reduce() to get # global valid tokens

## Testing
1. DP + gradient accumulation: loss is identical comparing `main` vs this branch
2. PP: NGPU=2, PP=2 compare `main` vs this branch, loss is identical. 
3. TP / loss parallel: comparing loss_parallel=true vs. loss_parallel=false, the loss are identical.

### DIFF
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -187,7 +187,7 @@ class Validator(BaseValidator):
             inputs, labels, extra_inputs, extra_kwargs = self.post_dataloading_process(
                 input_dict, labels, model_parts
             )
-            
+
             # Count valid tokens for this batch
             local_valid_tokens = torch.tensor(0, dtype=torch.int64, device=device_type)
             # pyrefly: ignore [missing-attribute]


torchtitan/train.py
```diff
@@ -559,12 +567,28 @@ def train_step(
...
-        # If data runs out during gradient accumulation, that
-        # entire step will not be executed.
+        # Collect all microbatches first to count total valid tokens
+        microbatches = []
+        local_valid_tokens = torch.tensor(0, dtype=torch.int64, device=self.device)
         for _microbatch in range(self.gradient_accumulation_steps):
             # pyrefly: ignore [no-matching-overload]
             input_dict, labels = next(data_iterator)
+            local_valid_tokens += (labels != IGNORE_INDEX).sum()
+            microbatches.append((input_dict, labels))
```
felipemello1: this is clever! A couple of questions:
1. Are these batches already on GPU? If so, is this wasteful?

2. This may be a bit problematic for RL, where we cannot overlap data_iterator (a generator) and training.

i.e. usually we can train on grad_step 1 while the batch of grad_step 2 is being generated. But in this setup, we need to first generate all grad_step batches before we start training step 1.

torchtitan/train.py
```diff
@@ -559,12 +567,28 @@ def train_step(
...
+
+        # All-reduce to get global token count across DP ranks
```
felipemello1, line 568: Noob question: Should we also do cp_mesh, or only dp_mesh?

torchtitan/train.py
```diff
@@ -590,8 +614,12 @@ def train_step(
             loss = loss.detach()
             ft_pg = self.ft_manager.loss_sync_pg
             loss_mesh = parallel_dims.get_optional_mesh("loss")
+
+            # loss = local_loss_sum / global_valid_tokens
```
felipemello1, line 617: Not saying you should do the same, but passing loss_scale to the loss fn was a common pattern i saw in other RL libraries too

This is what i did in my own PR:
```
def aggregate(
    per_token_loss: torch.Tensor,
    mask: torch.Tensor,
    agg_type: AggType = "token_mean",
    loss_scale: torch.Tensor | None = None,
) -> tuple[torch.Tensor, list[Metric]]:
    if agg_type == "token_mean":
        loss = masked_mean(per_token_loss, mask, loss_scale)

    elif agg_type == "fixed_horizon":
        # divide by (B * S)
        loss = (per_token_loss * mask).sum() / mask.numel()

    elif agg_type == "sequence_mean":
        seq_lengths = mask.sum(dim=-1).clamp(min=1.0)
        seq_means = (per_token_loss * mask).sum(dim=-1) / seq_lengths
        loss = seq_means.mean()

    else:
        raise ValueError(f"Unknown agg_type: {agg_type}")
    return loss, metrics
```

```
def masked_mean(
    values: torch.Tensor,
    mask: torch.Tensor,
    loss_scale: torch.Tensor | None = None,
) -> torch.Tensor:
    masked_sum = (values * mask).sum()
    if loss_scale is not None:
        divisor = loss_scale.clamp(min=1.0)
    else:
        divisor = mask.sum().clamp(min=1.0)
    return masked_sum / divisor
```

```
class GRPOLoss():
    def __call__(..., agg_type: "token_mean", loss_scale:int = None):
        ...
        loss, agg_m = aggregate(loss_per_token, loss_mask, agg_type, loss_scale)
        return loss
```

torchtitan/train.py
felipemello1: The other alternative seems to be to do as you are doing, i.e. loss computes the sum, not the mean --> compute gradients --> then later scale the **GRADIENTS** by 1/num_valid_tokens, NOT the loss.

wwwjn: > The other alternative seems to be to do as you are doing, i.e. loss computes the sum, not the mean --> compute gradients --> then later scale the GRADIENTS by 1/num_valid_tokens, NOT the loss.

A potential problem is loss can explosion if there are many tokens on a single rank, wdyt?

torchtitan/train.py
wwwjn: The number of valid tokens only comes from dataloader, cp mesh will not affect data loading behavior

torchtitan/train.py
felipemello1: I see. It makes sense. I mentioned it to Evan/Philip, they said 

"I guess.. honestly I have seen more gradient underflow problems than overflow ones tbh" 
"We actually discussed this at the time. If I recall, you could divide  the tokens by some initial estimate and then rescale at the end given the real token count relative to the estimate. But it was much more complicated and Evan tested it a bunch with bf16 and found this wasn’t a problem. For titan however they had to support more dtypes and context lengths so maybe it would be a problem".

Are we sure about the explosion? if so, maybe another option could be to allow the user to normalize per grad_step, instead of per total_sum for all grad_steps? Its not perfect, but the user can choose if they want approximate normalization or wait for all generations to be finished. I just dont know if it can be done cleanly

wwwjn: > honestly I have seen more gradient underflow problems than overflow ones tbh
> Are we sure about the explosion?

I'm not too sure about the explosion or underflow problem, as I haven't done enough experiments as Evan/ Philip. Maybe we could have a group to discuss more about their debugging experience. Thank you!

> you could divide the tokens by some initial estimate and then rescale at the end given the real token count relative to the estimate

This is mathematically doable but I feel like this will reduce a lot of code readability, but still need 1 all_reduce (the communication overhead is the same)

> allow the user to normalize per grad_step, instead of per total_sum for all grad_steps

I'm not objective about letting user normalize the loss per_step, but I think that's more for power users. We want to as least give a correct loss result for most of users as a starting point, and our standard  would be the numerics are correct and no underflow/ overflow issues.

torchtitan/models/llama3/train_configs/debug_model.toml
```diff
@@ -24,6 +24,10 @@ flavor = "debugmodel"
 hf_assets_path = "./tests/assets/tokenizer"
 # converters = ["float8"]
 
+[debug]
+deterministic = true
+seed = 0
```
tianyu-l: is this intentional -- to reproduce training loss every time?

torchtitan/train.py
```diff
@@ -210,9 +210,9 @@ def __init__(self, job_config: JobConfig):
             job_config.training.local_batch_size * batch_degree
         )
         assert self.gradient_accumulation_steps > 0
-        self.loss_fn = rescale_accumulated_loss(
-            self.loss_fn, self.gradient_accumulation_steps
-        )
+        # NOTE: We no longer wrap loss_fn with rescale_accumulated_loss
```
tianyu-l: we just need to explain what we are doing now, don't need to explain what we are no longer doing -- people who read the new code won't be able to see history

torchtitan/train.py
```diff
@@ -559,12 +567,28 @@ def train_step(
...
+        local_valid_tokens = torch.tensor(0, dtype=torch.int64, device=self.device)
         for _microbatch in range(self.gradient_accumulation_steps):
             # pyrefly: ignore [no-matching-overload]
             input_dict, labels = next(data_iterator)
+            local_valid_tokens += (labels != IGNORE_INDEX).sum()
+            microbatches.append((input_dict, labels))
+
+        # All-reduce to get global token count across DP ranks
+        if parallel_dims.dp_enabled:
+            dp_mesh = parallel_dims.get_optional_mesh("batch")
```
tianyu-l: `batch_mesh`, following https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/train.py#L95

torchtitan/train.py
```diff
@@ -559,12 +567,28 @@ def train_step(
...
+            ft_pg = self.ft_manager.loss_sync_pg
```
tianyu-l: I don't know the details here, what is `loss_sync_pg`? For non-ft, we should sync along DDP, FSDP, and CP for loss, but only DDP, FSDP for valid tokens.

torchtitan/train.py
```diff
@@ -528,10 +528,11 @@ def forward_backward_step(
             # accumulate losses across pipeline microbatches
             # TODO: PP+FSDP unexpectedly puts the loss back to the CPU
             loss = (
-                # using sum instead of mean because we already rescale the
-                # loss_fn down by a factor of n_microbatches in
-                # torchtitan/distributed/pipeline_parallel.py
-                torch.sum(torch.stack(losses)).to(self.device)
+                # Rescale PP loss to be "local loss sum / global valid tokens)
+                # because each microbathes could have different number of valid tokens
+                (torch.sum(torch.stack(losses)) / self.global_valid_tokens).to(
```
tianyu-l: let's do explicit arg passing, instead of implicitly via self. variables which sounds error-prone

torchtitan/train.py
```diff
@@ -541,11 +542,18 @@ def forward_backward_step(
...
+
+                    # Scale the loss by the inverse of the total weight denominator before backward
+                    # This ensures gradients are properly normalized across all microbatches
+                    loss = loss_sum / self.global_valid_tokens
+
                 # need to free pred before bwd to avoid peaking memory
                 del pred
                 loss.backward()
 
+        # The returned loss here is local_sum_loss / global_valid_tokens
```
tianyu-l: local_sum_loss not defined

torchtitan/train.py
tianyu-l: @felipemello1 
> The other alternative seems to be to do as you are doing, i.e. loss computes the sum, not the mean --> compute gradients --> then later scale the GRADIENTS by 1/num_valid_tokens, NOT the loss.

That means calling a (foreach) rescale on every gradient? My main worry is not sure how inefficient it would be.

> This may be a bit problematic for RL, where we cannot overlap data_iterator (a generator) and training.

IIRC you are handling microbatching by yourself and not depending on torchtitan -- as long as the loss computation is sum and accounts for IGNORE_INDEX, you are unblocked, right?

The main blocker we can't make local loss to use mean reduction is because PP api doesn't expose fine-grained control so that we can't rescale in the middle of PP microbatches. cc @H-Huang

torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -420,9 +429,18 @@ def apply_fsdp(
...
+        #       gradient division. We handle gradient scaling in the training
+        #       loop based on global token count.
+        model.norm.set_gradient_divide_factor(1.0)
+        model.output.set_gradient_divide_factor(1.0)
 
     fully_shard(model, **fsdp_config)
 
+    # Set gradient_divide_factor=1.0 to disable FSDP's automatic gradient division
+    # We handle gradient scaling ourselves in the training loop with global token count
+    model.set_gradient_divide_factor(1.0)
```
tianyu-l: Do we have to call `set_gradient_divide_factor(1.0)` every time we call `fully_shard`?
@weifengpy 

In the worst case, maybe we can have a separate loop to set it based on isinstance FSDPModule?

torchtitan/train.py
tianyu-l: @felipemello1 
> Are these batches already on GPU? If so, is this wasteful?

This is good point.

[1st] I think current approach is a regression in that
- per microbatch, it loads data onto GPU
- after all microbatches, it computes the total valid tokens, all on GPU
- start per-microbatch training loops

[2nd] Instead we could do
- load microbatches on CPU and compute total valid tokens
- per microbatch, send to GPU to train

[3rd] or like you said
- per microbatch train with sum reduction
- eventually scale down the gradients by total valid tokens

The 2nd approach has the benefit of not occupying GPU memory unnecessarily; the 3rd approach has the benefit that it doesn't require total valid tokens ahead of time thus don't need to load all data before any microbatch, but at the cost of doing a per-param gradient scaling down.

I would advocate approach 2. With ultra-long context and when CP is used, we first do full sequence loading and then chunking them in `post_dataloading_process` -- so if we load microbatch, it could add to a lot of memory pressure. cc @fegin 

For 3 the benefit is not obvious to me, other than underflow issue, which I don't know how severe it is.

torchtitan/models/llama4/infra/parallelize.py
wwwjn: From my testing, if we only call this on the outermost FSDPModule, it won't work and the gradients will still be averaged. Need @weifengpy  to double confirm.

torchtitan/models/llama3/train_configs/debug_model.toml
wwwjn: This is for my local test, and now we can reproduce the same training loss before vs. after this change.  I will remove this part

torchtitan/train.py
wwwjn: I follow the practice when we log the loss in train.py: https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/train.py#L589-L603. 

> For non-ft, we should sync along DDP, FSDP, and CP for loss, but only DDP, FSDP for valid tokens.

This is right, thank you

torchtitan/train.py
felipemello1: > For 3 the benefit is not obvious to me

@tianyu-l , in RL, the dataloader is actually a generator. Imagine that generating a batch takes 10s, and training takes 5s. Lets say we have grad_steps=3.

2nd setup we do:
Generation * 3 -> 30s
Training * 3 -> 15s
[wait training to end to start generating again]
Total: 45s

3rd setup we do:
Generation -> 10s
train + generation overlap ->10s
train + generation overlap ->10s
train -> 5s
[wait training to end to start generating again]
35s

Unfortunately, its not just a warm up penalty, because the model is updated on every train step, so the generator cant just keep running.

So, if this trainer is to be used as a backend for RL, preloading all the batches might be an issue. Thats why i suggested to either do grad scaling, or let the user have control over NOT preloading, at the cost of accuracy (i.e. the scaling is done per micro batch).

Here is how Tune did it: https://github.com/meta-pytorch/torchtune/blob/44271b570af36cfda8ee20a4479d2652770378c0/recipes/full_finetune_distributed.py#L361

The utility gets a bit complicated because we compile the scaling op
image

felipemello1: You said:

>IIRC you are handling microbatching by yourself and not depending on torchtitan -- as long as the loss computation is sum and accounts for IGNORE_INDEX, you are unblocked, right?

Thats true, If the user can rewrite this loop as they wish, then yes, users can normalize per microbatch

torchtitan/components/loss.py
```diff
@@ -4,22 +4,26 @@
...
-import functools
 from typing import Callable, TypeAlias
 
 import torch
 
 from torchtitan.config import JobConfig
 from torchtitan.tools.logging import logger
 
+# PyTorch's default ignore index for cross-entropy loss
+IGNORE_INDEX = -100
```
fegin, line 15: Do you think we should make this configurable? Or we are pretty sure that the dataset will always use the PyTorch default ignore index, in the job config?

torchtitan/components/validate.py
```diff
@@ -164,18 +165,31 @@ def validate(
...
+            accumulated_losses.append(loss_sum.detach())
 
             num_steps += 1
 
-        # Compute average loss
-        loss = torch.sum(torch.stack(accumulated_losses))
-        loss /= num_steps
+        # All-reduce token count across DP ranks to get global token count
+        if parallel_dims.dp_enabled:
+            loss_mesh = parallel_dims.get_optional_mesh("dp_replicate", "dp_shard")
```
fegin: Why CP is not included here? `parallel_dims.get_optional_mesh("loss")` is `("dp_replicate", "dp_shard", "cp")`.  We need to do `dist_sum` across all these dimensions to get global valid tokens? I may misunderstand something here.

torchtitan/distributed/parallel_dims.py
```diff
@@ -118,7 +117,7 @@ def unflatten_mesh(
             """
             backend_override = {}
             for name, degree in zip(dim_names, dim_degrees, strict=True):
-                if (not self._mesh_exist(name, degree)) or name == "batch":
+                if not self._mesh_exist(name, degree):
```
fegin, line 120: Why this changes? When do we need `batch` dim process group? If this is needed, we should update the docstring.

torchtitan/experiments/transformers_modeling_backend/infra/parallelize.py
```diff
@@ -304,7 +304,10 @@ def apply_fsdp(
 
     """
     mp_policy = MixedPrecisionPolicy(param_dtype=param_dtype, reduce_dtype=reduce_dtype)
-    fsdp_config = {"mesh": dp_mesh, "mp_policy": mp_policy}
+    fsdp_config = {
+        "mesh": dp_mesh,
+        "mp_policy": mp_policy,
+    }
```
fegin: Unnecessary new lines.

torchtitan/experiments/transformers_modeling_backend/infra/parallelize.py
```diff
@@ -328,6 +331,10 @@ def apply_fsdp(
             **fsdp_config,
             reshard_after_forward=reshard_after_forward,
         )
+        # NOTE: Set gradient_divide_factor to 1.0 to disable FSDP's automatic
+        #       gradient division. We handle gradient scaling in the training
+        #       loop based on global token count.
```
fegin: One comment is enough, this comment is repeat multiple times. We can just add this to the docstring of this function.

torchtitan/experiments/vlm/infra/parallelize.py
```diff
@@ -137,7 +137,10 @@ def apply_fsdp(
 
     """
     mp_policy = MixedPrecisionPolicy(param_dtype=param_dtype, reduce_dtype=reduce_dtype)
-    fsdp_config = {"mesh": dp_mesh, "mp_policy": mp_policy}
+    fsdp_config = {
+        "mesh": dp_mesh,
+        "mp_policy": mp_policy,
+    }
```
fegin: ditto. and other apply_fsdp as well

torchtitan/experiments/vlm/infra/parallelize.py
```diff
@@ -161,18 +164,30 @@ def apply_fsdp(
             **fsdp_config,
             reshard_after_forward=reshard_after_forward,
         )
+        # NOTE: Set gradient_divide_factor to 1.0 to disable FSDP's automatic
+        #       gradient division. We handle gradient scaling in the training
+        #       loop based on global token count.
+        model.tok_embeddings.set_gradient_divide_factor(1.0)
```
fegin: ditto. and other apply_fsdp as well

torchtitan/train.py
fegin: Should we start to think about dataloading pipeline in TorchTitan? I wanted to implement that but there is no obvious benefit before because dataloader has its own async processes. The only thing that cannot be overlapped with the current design is `to.(device)`. But there seems to be more and more "post-data-processing" that are computation hunger, e.g., BlockMask creation, and valid tokens computation. 

If we can do dataloading pipeline (1st fwd/bwd/optimier overlapping with 2nd dataloading and post-data-processing), the overlapping can be enabled even PP is not used.

Dataloading pipeline requires some design and discussion though. I would like to hear people's feedback about whether we should implement this approach.

torchtitan/train.py
fegin: This is a bit misleading though. After CP is applied, the input and label are sharded along sequence dimension. You can ignore cp_mesh and use `batch` mesh only because you are doing the computation before CP is applied. This causes the inconsistency between the logic here and the logic to compute the loss, where we need to consider CP dimension.

We should consider to put the logic  into `post_dataloading_process` and move `post_dataloading_process` out from the `forward_backward_step`.

torchtitan/components/loss.py
felipemello1: in tune we put it is a constant in some file. I dont see a need for it to be user configurable. Just need to make sure that everywhere uses the same constant, e.g. in collate or when using some mask, if necessary.

torchtitan/train.py
wwwjn: @felipemello1  I want to understand more details about gradient accumulation in RL

> in RL, the dataloader is actually a generator. Imagine that generating a batch takes 10s, and training takes 5s. Lets say we have grad_steps=3.

For 1 training step, it contains 3 gradient accumulate steps, and then the trainer model weights get updated and sync weight from trainer to generator. I made some notes on the following calculation, could you help check if this makes sense (in a synchronize setup)?

2nd setup we do:
Generation * 3 -> 30s **(get the total number of tokens here)**
Training * 3 -> 15s 
[wait training to end to start generating again] 
- **(rescale loss, loss.backward(), optimizer.step() to update trainer model weights )**

Total: 45s (= 30s + 15s) per train step. Then sync model weights from trainer to generator

3rd setup we do:
Generation -> 10s  
train + generation overlap ->10s
train + generation overlap ->10s
train -> 5s
[wait training to end to start generating again] 
- This would require SUM loss for each step
- **(loss.backward(), rescale gradients, optimizer.step() to update trainer model weights

35s per train step. Then sync model weights from trainer to generator

I can see the advantage of 3rd method (rescale grad)

> Unfortunately, its not just a warm up penalty, because the model is updated on every train step, so the generator cant just keep running.

In asynchronous RL, the overhead of 2nd setup will be less because we can overlaph the training * 3 (15s) with the generation * 3 = 30s from next training step?

> or let the user have control over NOT preloading, at the cost of accuracy (i.e. the scaling is done per micro batch)

If loss is re-scaled per microbatch , the loss will be wrong and it's not only accuracy.

felipemello1: Yes, the math looks correct!

> In asynchronous RL, the overhead of 2nd setup will be less because we can overlaph the training * 3 (15s) with the generation * 3 = 30s from next training step?

I got a bit confused about what you call 2nd setup, since we also call it 3rd method :P. But yes, for rescale grad we can overlap generator + training.

> If loss is re-scaled per microbatch , the loss will be wrong and it's not only accuracy.

Right, thats what i tried to say with "accuracy". My bad. It would be an approximation, normalizing it per grad_step, instead of normalizing per all valid tokens. It is not ideal, but an option if everything else is just as bad.

tianyu-l: @felipemello1 
I feel we are mixing up concepts.
1. For "async RL" I thought it means generator using outdated trainer weights (off by some steps)
2. Here it's about microbatching in trainer. It assumes we can't wait to get all the data in a full batch before a training iteration starts.

I would like to learn a bit more about real RL training scenarios. Could you show some examples of such use cases?
- E.g. if we have a replay buffer we can always generate rollouts to a buffer and make the batch available for next iteration.
- Do we not need to wait full batch to generate rewards, etc.

tianyu-l: @fegin 

> If we can do dataloading pipeline (1st fwd/bwd/optimier overlapping with 2nd dataloading and post-data-processing), the overlapping can be enabled even PP is not used.

Can we achieve this with `prefetch_factor`? https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/job_config.py#L227

wwwjn: Thank @tianyu-l  for explaining! Yes here's confusion,  I use the term "async RL" to refer to the asynchronous betwee generator and trainer, not "async between loading microbatches and training steps"

felipemello1: > E.g. if we have a replay buffer we can always generate rollouts to a buffer and make the batch available for next iteration.

This would be called off-policy, meaning that the data from the inference is from an older checkpoint. This is usually desirable for efficiency and many training runs use that, but it might hurt the training loss. I have seen all 3: very large off-policiness, e.g. checkpoints that are >10 iterations older, off-policy by 1, and on-policy. However, it might be worth checking with an RL researcher how frequently these happen. Let me get back to you on this.

fegin: @tianyu-l `prefetch_factor` doesn't reflect what I meant. I can quickly vibe code one to demonstrate. However for RL cases, dataloader pipeline requires at least off-policy by 1. I don't know how useful it is to fit the RL workflow though.

fegin: @tianyu-l  https://github.com/pytorch/torchtitan/pull/2232/

also cc., @wwwjn @felipemello1 to see if the idea works for RL use cases.

wwwjn: > > E.g. if we have a replay buffer we can always generate rollouts to a buffer and make the batch available for next iteration.
> 
> This would be called off-policy, meaning that the data from the inference is from an older checkpoint. This is usually desirable for efficiency and many training runs use that, but it might hurt the training loss. I have seen all 3: very large off-policiness, e.g. checkpoints that are >10 iterations older, off-policy by 1, and on-policy. However, it might be worth checking with an RL researcher how frequently these happen. Let me get back to you on this.

@felipemello1  @tianyu-l  Did we heard back from them yet? IIRC if in off-policy, the generator can be overlapped with next step's training, 2nd solution (defined [here](https://github.com/pytorch/torchtitan/pull/2206#discussion_r2677299020)) is good. If it is necessary to overlap between microbatchs, not only batches, then I will refactor this PR to follow 3rd solution

felipemello1: i asked 2 researchers, just heard from one. The one that talked to me said that there is always some degree of off-policy going on. It seems that it may or may not be a bottleneck depending on how off policy, how long generation takes, how long training takes, bsz, group size, buffer size, etc.

Lets proceed with it since you have it tested, later on we can get some numbers, and worst case we refactor.

How does that sound?

tianyu-l: Sounds good. We can always refactor.

torchtitan/components/loss.py
wwwjn: Thank you! If it's not configurable then I will leave it here as a constant.

torchtitan/distributed/parallel_dims.py
wwwjn: I will need to dist.sum() to get global_valid_tokens across the batch_dim in train_step()

torchtitan/models/llama3/infra/parallelize.py
```diff
@@ -322,14 +338,18 @@ def apply_fsdp(
     # since FSDP would prefetch them immediately after the forward pass
     if model.norm is not None and model.output is not None:
         # pyrefly: ignore [no-matching-overload]
-        fully_shard(
+        module = fully_shard(
```
tianyu-l: why this change?

torchtitan/models/llama3/train_configs/debug_model.toml
```diff
@@ -47,6 +47,7 @@ data_parallel_replicate_degree = 1
 data_parallel_shard_degree = -1
 fsdp_reshard_after_forward = "default" # default / never / always
 tensor_parallel_degree = 1
+disable_loss_parallel = false
```
tianyu-l: revert?

torchtitan/train.py
```diff
@@ -529,10 +528,9 @@ def forward_backward_step(
             # accumulate losses across pipeline microbatches
             # TODO: PP+FSDP unexpectedly puts the loss back to the CPU
```
tianyu-l, line 519: @H-Huang 
Is this expected, or fixed? IIUC putting loss on CPU would incur d2h sync which could've been avoided?

@wwwjn can we create an issue to track?

torchtitan/train.py
tianyu-l: My question is not resolved -- why do we sync across CP for num_valid_token computation?

torchtitan/components/validate.py
```diff
@@ -103,6 +103,9 @@ def validate(
         parallel_dims = self.parallel_dims
 
         accumulated_losses = []
+        accumulated_tokens = torch.tensor(
```
tianyu-l: let's use the same variable naming as in train.py?

torchtitan/components/validate.py
tianyu-l: this should be batch mesh, not loss mesh?

torchtitan/components/validate.py
```diff
@@ -164,18 +165,31 @@ def validate(
...
+            loss_mesh = parallel_dims.get_optional_mesh("dp_replicate", "dp_shard")
+            global_valid_tokens = dist_utils.dist_sum(
+                accumulated_tokens, loss_mesh, None
+            )
+        else:
+            global_valid_tokens = accumulated_tokens.float()
+
+        # Compute total loss and normalize by global token count
+        total_loss = torch.sum(torch.stack(accumulated_losses))
+        loss = total_loss / torch.clamp(global_valid_tokens, min=1.0)
```
tianyu-l: why not clamping in train.py?

torchtitan/train.py
```diff
@@ -474,7 +470,10 @@ def post_dataloading_process(
         return inputs, labels, extra_inputs, extra_kwargs
 
     def forward_backward_step(
-        self, input_dict: dict[str, torch.Tensor], labels: torch.Tensor
+        self,
+        input_dict: dict[str, torch.Tensor],
+        labels: torch.Tensor,
+        global_valid_tokens: torch.Tensor,
```
tianyu-l, line 486: Please use kwargs.
```suggestion
        self,
        *,
        input_dict: dict[str, torch.Tensor],
        labels: torch.Tensor,
        global_valid_tokens: torch.Tensor,
```

torchtitan/train.py
wwwjn: > My question is not resolved -- why do we sync across CP for num_valid_token computation?

We didn't sync cross CP for num_valid_token computation, here batch dimension is only "dp_replicate" and "dp_shard", not "CP". 


We have discussion with @fegin about this, CP is applied in forward_backward_step ([here](https://github.com/pytorch/torchtitan/blob/a25dd8f880e0759c1496d07ff0ca6973c6dc8b65/torchtitan/train.py#L498)),  which is after the num_valid_token calculation. But when we log the loss after `forward_backward_step` returned, cp are already applied. 
> You can ignore cp_mesh and use batch mesh only because you are doing the computation before CP is applied. This causes the inconsistency between the logic here and the logic to compute the loss, where we need to consider CP dimension.

> We should consider to put the logic into post_dataloading_process and move post_dataloading_process out from the forward_backward_step.

This part is more tricky because now we put input / labels on CPU and calculate the number of valid tokens before `post_dataloading_process`. Calculating attention mask needs the inputs to be on GPU. I think about this because of Flux: FLUX do the same `post_dataloading_process` to apply image encoder on the label image, which will change the shape of the label. So for FLUX model, we should only calculate the number of valid tokens after the `post_dataloading_process`, which has already make all the labels on GPU.

tianyu-l: to clarify: my question is on the usage of `ft_pg = self.ft_manager.loss_sync_pg`. Do you know what it does? In the long term, could you help move ft-related logic to experiments?

wwwjn: Oh I see where the "loss" comes from, thanks for clarification! I don't exactly know what this pg is doing, but I think we could remove here, and long-term move ft related to experiments

torchtitan/train.py
wwwjn: Do you mean we should put `global_valid_tokens` in the kwargs, instead of have a separate field here? Or do you want to enforce we specify the field name when calling the function, eg `labels=labels`

torchtitan/components/validate.py
```diff
@@ -151,10 +154,8 @@ def validate(
 
                 # accumulate losses across pipeline microbatches
                 # TODO: PP+FSDP unexpectedly puts the loss back to the CPU
-                loss = (
-                    # using sum instead of mean because we already rescale the
-                    # loss_fn down by a factor of n_microbatches in
-                    # torchtitan/distributed/pipeline_parallel.py
+                loss_sum = (
+                    # using sum because loss_fn already uses reduction='sum'
```
felipemello1:
```suggestion
                    # we compute the sum of losses and later normalize by accumulated_tokens
```

torchtitan/train.py
tianyu-l: > Or do you want to enforce we specify the field name when calling the function, eg labels=labels

latter

torchtitan/components/validate.py
wwwjn: My bad, for validation, I moved the global_valid_token calculation before the CP is applied, and right after dataloader

torchtitan/models/llama3/infra/parallelize.py
wwwjn: reverted!

torchtitan/components/validate.py
wwwjn: I remove the clamping all together with following 2 reason:
1. It's low chance there's no valid tokens within a batch
2. If it happens, user should notice this data issue, instead of we did the clamping and give wrong loss results

torchtitan/models/flux/train.py
```diff
@@ -91,8 +93,25 @@ def __init__(self, job_config: JobConfig):
             )
 
     def forward_backward_step(
-        self, input_dict: dict[str, torch.Tensor], labels: torch.Tensor
+        self,
+        *,
+        input_dict: dict[str, torch.Tensor],
+        labels: torch.Tensor,
+        global_valid_tokens: torch.Tensor | None,
```
tianyu-l: - set default to None so that you don't have to call it from train_step
- assert this to be None in the beginning of this function, so that we don't accidentally involve it in computation if someone sends it

tianyu-l (approved):
please address the last comment

wwwjn:
CI failing "DeepSeek V3 Flex+PP+FSDP+EP+SACOP" not related to this change

tianyu-l:
may need to rebase for the varlen error to disappear

PR #2210, opened 2026-01-08
feat: add differential learning rate and weight decay support

eous:
Add support for per-parameter-group learning rates and weight decay multipliers, enabling fine-grained control during fine-tuning.

Changes:
- Add LRMultipliers and WeightDecayMultipliers dataclasses in job_config.py
- Add classify_parameters_for_groups() function with explicit pattern ordering
- Add OptimizersContainerWithParamGroups class for differential settings
- Add PP incompatibility check (raises NotImplementedError)
- Add warning when differential settings used with early_step_in_backward
- Add differential LR logging in train.py

Pattern groups: bias, embeddings, output, attention, experts, routers, norms (bias matched first, takes precedence for all *.bias parameters)

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -620,8 +620,9 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
                 for i, (lr_val, param_group) in enumerate(zip(lr_list, optimizer.param_groups)):
                     group_name = param_group.get('group_name', f'group_{i}')
                     extra_metrics[f"lr/{group_name}"] = lr_val
-            except (AttributeError, IndexError):
-                # Fallback: just use numeric indices
+            except (AttributeError, IndexError) as e:
+                # Fallback: just use numeric indices (optimizer may not have group_name keys)
+                logger.debug(f"Using numeric indices for LR logging (no group_name): {e}")
                 for i, lr_val in enumerate(lr_list):
                     extra_metrics[f"lr/group_{i}"] = lr_val
 


eous:
> > Add support for per-parameter-group learning rates and weight decay multipliers, enabling fine-grained control during fine-tuning.
> 
> Is there any literature / experiments supporting such use cases?

https://huggingface.co/eousphoros/persona_eta_20b_131k I have been mainly using it to slow down parts of the model that seem to naturally absorb most of the gradient during the backprop like embed/output to encourage specialization in moe instead of just rerouting. This seems to reduce the amount of damage I do to the models base capabilities during SFT. Nothing formal here though.

Not exactly apples to apples since there were other differences but from my limited experimentation this sort of pattern tends to hold

weight_diff_analysis

Note the top down change instead of the bottom down change

weight_diff_analysis_s3

tianyu-l:
Interesting. My impression is that we need more study to learn this type of work and figure out the APIs in pytorch to support flexible optimizer / LR setup. cc @janeyx99 if you have more context.

Meanwhile, I wonder if you have heard about / tried things like muP (https://github.com/microsoft/mup) which looks at scalings of initialization, learning rate, etc. more systematically.

janeyx99:
I have not read through the code here, but from the PR description, "per-parameter-group learning rates and weight decay multipliers" are likely supportable through:
- param_groups already have their own weight decay/lr
- some external lightweight function to decay LR/weight decay. For LR, I'm wondering where this use case is not covered already by LRScheduler?

PR #2211, opened 2026-01-08
fix(gpt-oss): correct attention sink from sigmoid to LSE renormalization

eous:
The previous sigmoid-based attention sink implementation was mathematically incorrect. This fix uses proper LSE (Log-Sum-Exp) renormalization that is equivalent to HuggingFace's concat+softmax approach.

Mathematical equivalence:
- HF approach: concat sink logit to scores, softmax over K+1, drop sink position
- Our approach: compute combined_lse = logsumexp([lse, sink]), then renormalize output by exp(old_lse - new_lse)

Changes:
- Replace sigmoid(lse - sink) with proper LSE renormalization
- Add clamping [-20, 0] for numerical stability
- Add comprehensive test suite validating equivalence to HF reference

Reference: HuggingFace transformers/integrations/flex_attention.py lines 309-322


tianyu-l:
@wwwjn could you take a look?

PR #2212, opened 2026-01-08
feat(moe): add topk_before_score routing and use_router_bias support

eous:
Add support for GPT-OSS style MoE routing with TopK-then-Score strategy.

Changes to MoEArgs:
- Add topk_before_score: bool = False (preserve backward compat)
- Add use_router_bias: bool = False (for GPT-OSS learned router bias)
- Add use_expert_bias: bool = False (for GPT-OSS expert biases)

Changes to TokenChoiceTopKRouter:
- Support both routing strategies:
  - TopK-then-Score (GPT-OSS): topk on raw logits, then score_func
  - Score-then-TopK (DeepSeek-V3): score_func to all, then topk
- Add validation: node-limited routing not supported with topk_before_score
- Fix expert_bias handling: gather unbiased logits for score computation
- Add _debug_force_load_balance support for TopK-then-Score path
- Add explicit bias initialization in init_weights
- Change histc to bincount for token counting (more reliable)

GPT-OSS config updates:
- 20B/120B: Add topk_before_score=True, use_router_bias=True, use_expert_bias=True
- 120B: Fix top_k from 4 to 8 (correct value for 120B model)
- Add GptOssStateDictAdapter to __all__

### DIFF
--- a/torchtitan/models/gpt_oss/__init__.py
+++ b/torchtitan/models/gpt_oss/__init__.py
@@ -35,13 +35,14 @@ gptoss_configs = {
             num_experts=8,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,  # Normalize over selected experts (match production configs)
             route_scale=1.0,
             score_before_experts=False,
             top_k=4,
             use_grouped_mm=True,
             load_balance_coeff=1e-3,
         ),
+        use_router_bias=True,  # GPT-OSS models have learned router gate bias
         attn_mask_type="causal",
     ),
     "20b": GptOssModelArgs(
@@ -50,17 +51,15 @@ gptoss_configs = {
             num_experts=32,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,  # Normalize over selected experts (equivalent to softmax over top-k)
             route_scale=1.0,
             score_before_experts=False,
             top_k=4,
             use_grouped_mm=True,
             load_balance_coeff=None,  # Disabled for fine-tuning - not part of official GPT-OSS model
-            use_router_bias=True,  # GPT-OSS models have learned router bias
-            use_expert_bias=True,  # GPT-OSS models have learned expert biases
-            topk_before_score=True,  # GPT-OSS uses TopK-then-Score routing
         ),
-        attn_mask_type="causal",  # Preserve sample boundaries using EOS token
+        use_router_bias=True,  # GPT-OSS models have learned router gate bias
+        attn_mask_type="causal",
     ),
     "120b": GptOssModelArgs(
         n_layers=36,
@@ -68,17 +67,15 @@ gptoss_configs = {
             num_experts=128,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,  # Normalize over selected experts (equivalent to softmax over top-k)
             route_scale=1.0,
             score_before_experts=False,
             top_k=8,  # 120B uses top-8 routing (vs top-4 for 20B)
             use_grouped_mm=True,
             load_balance_coeff=None,  # Disabled for fine-tuning - not part of official GPT-OSS model
-            use_router_bias=True,  # GPT-OSS models have learned router bias
-            use_expert_bias=True,  # GPT-OSS models have learned expert biases
-            topk_before_score=True,  # GPT-OSS uses TopK-then-Score routing
         ),
-        attn_mask_type="causal",  # Preserve sample boundaries using EOS token
+        use_router_bias=True,  # GPT-OSS models have learned router gate bias
+        attn_mask_type="causal",
     ),
 }
 
--- a/torchtitan/models/gpt_oss/model/args.py
+++ b/torchtitan/models/gpt_oss/model/args.py
@@ -58,6 +58,7 @@ class GptOssModelArgs(BaseModelArgs):
     # MoE
     moe_args: MoEArgs = field(default_factory=MoEArgs)
     swiglu_limit: float = 7.0
+    use_router_bias: bool = False  # GPT-OSS models have learned router gate bias
     # Multi-Head Latent Attention (MLA)
     head_dim: int = 64
     n_heads: int = 64
--- a/torchtitan/models/gpt_oss/model/moe.py
+++ b/torchtitan/models/gpt_oss/model/moe.py
@@ -5,15 +5,63 @@
 # LICENSE file in the root directory of this source tree.
 
 
-from typing import Callable
+from typing import Callable, Literal
 
 import torch
 from torch import nn
 from torch.distributed.tensor import DTensor
-from torchtitan.models.moe.moe import MoE
+from torchtitan.models.moe.moe import MoE, TokenChoiceTopKRouter
 from torchtitan.models.moe.utils import _permute, _unpermute
 
 from .args import GptOssModelArgs
+
+
+class GptOssTokenChoiceTopKRouter(TokenChoiceTopKRouter):
+    """GPT-OSS router that supports learned router gate bias.
+
+    GPT-OSS models use a router with a learned bias term. This class extends
+    the base TokenChoiceTopKRouter to support bias in the gate linear layer.
+
+    Note: Since softmax is a monotonically increasing function, the topk selection
+    is the same whether applied before or after softmax. Therefore, we use the
+    standard Score-then-TopK approach with route_norm=True to normalize over
+    the selected experts (equivalent to softmax over just top-k).
+    """
+
+    def __init__(
+        self,
+        dim: int,
+        num_experts: int,
+        num_expert_groups: int | None,
+        num_limited_groups: int | None,
+        top_k: int,
+        score_func: Literal["softmax", "sigmoid"],
+        route_norm: bool,
+        route_scale: float,
+        use_router_bias: bool = True,
+        _debug_force_load_balance: bool = False,
+    ):
+        super().__init__(
+            dim=dim,
+            num_experts=num_experts,
+            num_expert_groups=num_expert_groups,
+            num_limited_groups=num_limited_groups,
+            top_k=top_k,
+            score_func=score_func,
+            route_norm=route_norm,
+            route_scale=route_scale,
+            _debug_force_load_balance=_debug_force_load_balance,
+        )
+        # Override the gate with bias support for GPT-OSS.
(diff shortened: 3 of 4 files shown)


torchtitan/models/moe/moe.py
tianyu-l, line 1: gpt-oss has its own MoE implementation, please don't modify the vanilla one.
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/gpt_oss/model/moe.py

eous: Ok, Ill take a look. Thank you for the pointer.

shuhuayu:
@eous  Thanks for the pr. Can you move the changes to `moe/moe.py` to `gpt_oss/model/moe.py`?

shuhuayu:
@eous I rethink about it. it seems that there is no need to do topk before scores. Now both score functions we support are monotonically increasing functions, so topk before or after does not change the experts we select. For gptoss models, the thing we need to do is set `route_norm = True`, then we do sum normalization over the selected experts softmax scores (over all), this reduces to the same result of softmax over topk experts. However, we do need to add `route_gate_bias` and `expert_bias` options in `MoEArgs` and set them to be `True` for gptoss models. What do you think?

eous:
> @eous I rethink about it. it seems that there is no need to do topk before scores. Now both score functions we support are monotonically increasing functions, so topk before or after does not change the experts we select. For gptoss models, the thing we need to do is set `route_norm = True`, then we do sum normalization over the selected experts softmax scores (over all), this reduces to the same result of softmax over topk experts. However, we do need to add `route_gate_bias` and `expert_bias` options in `MoEArgs` and set them to be `True` for gptoss models. What do you think?

@shuhuayu Thanks, the monotonicity insight simplifies things. I've reworked the PR in [33ec345](https://github.com/pytorch/torchtitan/pull/2212/commits/c9488775574ee5c61fb1af48529b1a06c35e5957) - reverted the vanilla moe.py changes and added a GptOssTokenChoiceTopKRouter subclass that overrides the gate with bias support, plus set route_norm=True as you suggested.                        
                                                                                                                                                    
  One limitation: the subclass has to recreate the gate after the parent already creates it with bias=False, which is a bit awkward and introduces an implicit ordering dependency. Might be cleaner to just plumb use_router_bias into the base TokenChoiceTopKRouter as a backward-compatible param (default False). Let me know if you'd prefer that approach.

---edit
[rank0]:[titan] 2026-01-23 00:33:15,951 - root - INFO - step:  1  loss:  2.2174  grad_norm: 111.5000  memory: 89.57GiB(94.31%)  tps: 292  tflops: 51.40  mfu: 16.47%
[rank0]:[titan] 2026-01-23 00:49:28,102 - root - INFO - step:  5  loss:  1.1087  grad_norm: 20.0000  memory: 89.57GiB(94.31%)  tps: 1,079  tflops: 190.13  mfu: 60.94%
Testing it out on my work branch on my theta train. So far at least just from loss/grad_norm ir seems to be fine.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -34,13 +35,14 @@
             num_experts=8,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,  # Normalize over selected experts (match production configs)
```
shuhuayu, line 38: We can simplify by removing inline comments.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -34,13 +35,14 @@
...
             route_scale=1.0,
             score_before_experts=False,
             top_k=4,
             use_grouped_mm=True,
             load_balance_coeff=1e-3,
         ),
+        use_router_bias=True,  # GPT-OSS models have learned router gate bias
```
shuhuayu, line 45: I think it makes better sense to add it as an attribute of the dataclass `MoEArgs`, and name it `use_gate_bias`, also remove the inline comments.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
             num_experts=32,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,  # Normalize over selected experts (equivalent to softmax over top-k)
```
shuhuayu, line 54: ditto.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
             route_scale=1.0,
             score_before_experts=False,
             top_k=4,
             use_grouped_mm=True,
-            load_balance_coeff=1e-3,
+            load_balance_coeff=None,  # Disabled for fine-tuning - not part of official GPT-OSS model
```
shuhuayu, line 59: revert.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
         ),
+        use_router_bias=True,  # GPT-OSS models have learned router gate bias
```
shuhuayu, line 61: ditto.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
+        attn_mask_type="causal",
```
shuhuayu, line 62: revert.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
     ),
     "120b": GptOssModelArgs(
         n_layers=36,
         moe_args=MoEArgs(
             num_experts=128,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,  # Normalize over selected experts (equivalent to softmax over top-k)
```
shuhuayu, line 70: ditto.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
             route_scale=1.0,
             score_before_experts=False,
-            top_k=4,
+            top_k=8,  # 120B uses top-8 routing (vs top-4 for 20B)
```
shuhuayu, line 73: revert. check https://arxiv.org/pdf/2508.10925 section 2.2 2nd paragraph.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
             use_grouped_mm=True,
-            load_balance_coeff=1e-3,
+            load_balance_coeff=None,  # Disabled for fine-tuning - not part of official GPT-OSS model
```
shuhuayu, line 75: revert.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
         ),
+        use_router_bias=True,  # GPT-OSS models have learned router gate bias
```
shuhuayu, line 77: ditto.

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,27 +51,31 @@
...
+        attn_mask_type="causal",
```
shuhuayu, line 78: revert.

torchtitan/models/gpt_oss/model/args.py
```diff
@@ -58,6 +58,7 @@ class GptOssModelArgs(BaseModelArgs):
     # MoE
     moe_args: MoEArgs = field(default_factory=MoEArgs)
     swiglu_limit: float = 7.0
+    use_router_bias: bool = False  # GPT-OSS models have learned router gate bias
```
shuhuayu, line 61: I think it makes better sense to move it inside `MoEArgs`. Renaming it as `use_gate_bias` since we use it in the `self.gate`. And set the default value as `False`. In `gpt_oss/__init__.py`, we set them as `True` within `MoEArgs`.

torchtitan/models/gpt_oss/model/moe.py
```diff
@@ -5,17 +5,65 @@
...
 
 
-from typing import Callable
+from typing import Callable, Literal
 
 import torch
 from torch import nn
 from torch.distributed.tensor import DTensor
-from torchtitan.models.moe.moe import MoE
+from torchtitan.models.moe.moe import MoE, TokenChoiceTopKRouter
```
shuhuayu, line 13: revert.

torchtitan/models/gpt_oss/model/moe.py
```diff
@@ -5,17 +5,65 @@
 # LICENSE file in the root directory of this source tree.
 
 
-from typing import Callable
+from typing import Callable, Literal
```
shuhuayu, line 8: revert.

torchtitan/models/gpt_oss/model/moe.py
```diff
@@ -5,17 +5,65 @@
...
+        self,
+        dim: int,
+        num_experts: int,
+        num_expert_groups: int | None,
+        num_limited_groups: int | None,
+        top_k: int,
+        score_func: Literal["softmax", "sigmoid"],
+        route_norm: bool,
+        route_scale: float,
+        use_router_bias: bool = True,
```
shuhuayu, line 41: I think in this case, to simplify, we can directly add `use_gate_bias` into the `init` in `TokenChoiceTopKRouter` in `moe/moe.py`. In the `MoE` module under `moe/moe.py`, one can pass `moe_args.use_gate_bias` to `self.router` similar to `route_norm`. In this way,  we no longer need to create this new class and share most code.

torchtitan/models/gpt_oss/model/moe.py
```diff
@@ -286,3 +340,19 @@ def __init__(self, model_args: GptOssModelArgs, dim: int, hidden_dim: int):
...
+                num_experts=moe_args.num_experts,
+                num_expert_groups=moe_args.num_expert_groups,
+                num_limited_groups=moe_args.num_limited_groups,
+                top_k=moe_args.top_k,
+                score_func=moe_args.score_func,
+                route_norm=moe_args.route_norm,
+                route_scale=moe_args.route_scale,
+                use_router_bias=True,
+                _debug_force_load_balance=moe_args._debug_force_load_balance,
+            )
```
shuhuayu, line 358: revert.

shuhuayu:
@eous Thanks for the rework. Left some comments.

eous:
This looks like its now a duplicate of https://github.com/pytorch/torchtitan/pull/2319, closing it out

Issue #1525, opened 2025-08-04
Transformer is running with float32 instead of bfloat16 !


githubsgi:
### Bug description

Modified the Llama3 modle.py to print dtype as follows and ran just 1 rank. The  

```
    def forward(
        self,
        tokens: torch.Tensor,
        eos_id: int | None = None,
        input_batch: torch.Tensor | None = None,
    ):
        """
        Perform a forward pass through the Transformer model.

        Args:
            tokens (torch.Tensor): Input token indices if pipeline parallelism is not enabled.
                If pipeline parallelism is enabled, this will be the input token indices
                for the ranks on the first pipeline stage. This will be the activation of the
                previous pipeline stage if the current rank is not on the first stage.
            input_batch (torch.Tensor): The input batch read from the dataloader.
                This will always be the input batch regardless of the pipeline stage.
                This field is required for non-first PP stages to perform document
                masking attention (to analyze the boundary of the document).

        Returns:
            torch.Tensor: Output logits after applying the Transformer model.

        """
        if self.model_args.use_flex_attn:
            init_attention_mask(
                input_batch if input_batch is not None else tokens, eos_id=eos_id
            )

        print (f"tokens.dtype {tokens.dtype}")
        # passthrough for nonexistent layers, allows easy configuration of pipeline parallel stages
        h = self.tok_embeddings(tokens) if self.tok_embeddings else tokens
        print (f"h.dtype {h.dtype}")

        for layer in self.layers.values():
            h = layer(h, self.freqs_cis)
            print (f"h.dtype {h.dtype}")

        h = self.norm(h) if self.norm else h
        print (f"h.dtype {h.dtype}")
        output = self.output(h) if self.output else h
        print (f"output.dtype {h.dtype}")
        return output

```
Seeing only float32 datatypes as follows.

```
tokens.dtype torch.int64
h.dtype torch.float32
h.dtype torch.float32
h.dtype torch.float32
h.dtype torch.float32
h.dtype torch.float32
h.dtype torch.float32
h.dtype torch.float32
h.dtype torch.float32
output.dtype torch.float32

```

The config is:

[...]

PR #2213, opened 2026-01-08
fix: enable torch.autocast for TP parallelism without FSDP

eous:
Remove overly conservative restriction that disabled mixed precision for TP-only configurations. torch.autocast operates at the operator level and is orthogonal to tensor parallelism.

Before: TP-only training would show warning and disable mixed precision
After: TP-only training uses torch.autocast for mixed precision

Note: PP-only training uses schedule-based execution and doesn't use maybe_enable_amp (unchanged by this PR).

Affected configurations:
- TP-only (now enabled)
- DDP-only (was already enabled)
- Single-device (was already enabled)
- FSDP/HSDP (unchanged - handled internally by fully_shard)

### DIFF
--- a/torchtitan/distributed/utils.py
+++ b/torchtitan/distributed/utils.py
@@ -250,24 +250,19 @@ def maybe_enable_amp(
     parallel_dims: ParallelDims, mixed_precision_param: str, device_type: str
 ) -> contextlib.AbstractContextManager[None]:
     if parallel_dims.fsdp_enabled:
-        # FSDP handles mixed precision internally
+        # FSDP handles mixed precision internally via MixedPrecisionPolicy
         logger.info("Mixed precision training is handled by fully_shard")
         return contextlib.nullcontext()
     else:
-        if parallel_dims.tp_enabled or parallel_dims.pp_enabled:
-            logger.warning(
-                "Mixed precision training with TP or PP is only supported when FSDP/HSDP/CP is enabled."
-            )
-            logger.info("Mixed precision training is disabled")
-            return contextlib.nullcontext()
-        else:
-            # the following code will only be executed for DDP or single-device training
-            logger.info("Mixed precision training is handled by AMP")
-            # pyrefly: ignore [bad-return]
-            return torch.autocast(
-                device_type,
-                dtype=TORCH_DTYPE_MAP[mixed_precision_param],
-            )
+        # Enable autocast for non-FSDP cases (DDP, TP-only, single-device)
+        # Note: PP uses its own schedule-based execution and doesn't use this context
+        # torch.autocast works at the operator level and is orthogonal to parallelism strategy
+        logger.info("Mixed precision training is handled by AMP")
+        # pyrefly: ignore [bad-return]
+        return torch.autocast(
+            device_type,
+            dtype=TORCH_DTYPE_MAP[mixed_precision_param],
+        )
 
 
 def init_fake_mode(world_size: int, comm_mode: str = "fake_backend"):


tianyu-l (changes requested):
Have you verified it works properly? Could you show evidence, in terms of param / activation / grad dtype, and throughput comparison with mixed precision off?

I vaguely remember that I've tried it before and it didn't work as expected.

eous:
> Have you verified it works properly? Could you show evidence, in terms of param / activation / grad dtype, and throughput comparison with mixed precision off?
> 
> I vaguely remember that I've tried it before and it didn't work as expected.
weight_diff_analysis_s3

https://huggingface.co/eousphoros/persona_eta_20b_131k This model was trained with TP=4 no fdsp. The output with autocast was inline with what I expected though I lack the depth of knowledge to formerly confirm this.

tianyu-l:
Thanks. It's hard to tell from the plots that it is working properly.
My impression is that `autocast` is not actively maintained. Is it true? @fegin 

Also curious why would you use amp with TP but without FSDP?

fegin:
> My impression is that autocast is not actively maintained. Is it true? @fegin

autocast is the core feature. But autocast with parallelisms is not actively maintained. We have seen performance gap between autocast + DDP and FSDP implementation with world_size being the same.  IMO, if we can use FSDP we should use FSDP.

slabhs-aws:
@fegin @tianyu-l imo we should still support autocast for TP-only (no FSDP) workloads. During model bringup, we often start small with single core workloads, scale to TP, and then scale to FSDP. Accuracy validation at each point requires FP32, AMP, MXFP8 etc precisions supported. Ideally all precisions should support all parallelisms..

tianyu-l:
@slabhs-aws I don't think people have plans working on autocast + TP.

We are baking some new idea, which I think would tackle the problem along the way. See https://github.com/pytorch/pytorch/compare/main...ailing/mst#diff-e326c855a215daccfd35d8953f8ce5f7c75cf2c446eb9880b22b512a8b1d7158

cc @ailzhang @weifengpy

weifengpy:
> @slabhs-aws I don't think people have plans working on autocast + TP.
> 
> We are baking some new idea, which I think would tackle the problem along the way. See [pytorch/pytorch@main...ailing/mst#diff-e326c855a215daccfd35d8953f8ce5f7c75cf2c446eb9880b22b512a8b1d7158](https://github.com/pytorch/pytorch/compare/main...ailing/mst#diff-e326c855a215daccfd35d8953f8ce5f7c75cf2c446eb9880b22b512a8b1d7158)
> 
> cc @ailzhang @weifengpy

@tianyu-l This autocast happens after "mst.to_compute_state()" so I didn't understand your comment yet

tianyu-l:
@weifengpy mst should handle mixed precision as FSDP2 does, so autocast is not needed.

I didn't know much on who's maintaining autocast. If you know, could you tag them here?

weifengpy:
> @weifengpy mst should handle mixed precision as FSDP2 does, so autocast is not needed.

mst takes care of replicated parameters. for DDP or TP-only, autocast is still relevant?

> I didn't know much on who's maintaining autocast. If you know, could you tag them here?

@janeyx99 on autocast

tianyu-l:
> mst takes care of replicated parameters. for DDP or TP-only, autocast is still relevant?

what is the difference between autocast and FSDP2 mixed precision? Since mst is not tied to DP and is about unifying storage, why mst wouldn't handle all params' dtype conversion? E.g. I imagine in TP, `mst.to_compute_state` would take fp32 shard param (e.g. colwise paralleled tensor), and returns a bf16 shard param for computation.

weifengpy:
> what is the difference between autocast and FSDP2 mixed precision?

autocast does dtype cast op-by-op. fsdp mixed precision does dtype cast layer-by-layer (including input/output tensors). there is perf advantage when fsdp takes care of mixed precision

when MST is ready, it does dtype cast for its managed parameters. But input/output (activations) tensors are not managed by MST. We need pre/post forward/backward hooks to take care of dtype casting for input/output

tianyu-l:
@weifengpy
fwiw:
- in LLM training, the input is integer. With FSDP we don't have to do this casting because the embedding layer is run in bf16 thanks to fsdp mixed precision handling.
- autocast doesn't support embedding, per https://github.com/pytorch/torchtitan/issues/1525

weifengpy:
> * in LLM training, the input is integer.

good point! I missed this

PR #2214, opened 2026-01-08
feat(moe): add use_expert_bias config for optional expert biases

eous:
Add support for optional expert biases (mlp1_bias, mlp2_bias) in GptOssGroupedExperts, required for loading GPT-OSS pretrained models.

Changes to GptOssGroupedExperts:
- Add use_expert_bias parameter (default=True for GPT-OSS)
- Add compute_dtype parameter for configurable compute precision
- Add caching (_cached_tp_degree, _is_dtensor) for performance
- Add _get_tp_degree() method with cache lookup
- Proper None checks in _run_experts_for_loop and _run_experts_grouped_mm
- ScaleBiasForward custom autograd for TP bias scaling

Changes to expert_parallel.py:
- Handle optional bias distribution in GptossTensorParallel
- Handle optional bias distribution in GptossExpertTensorParallel
- Invalidate caches after parallelization

Changes to MoEArgs:
- Add use_expert_bias: bool = False field

Config updates:
- Set use_expert_bias=True for GPT-OSS 20B/120B models

### DIFF
--- a/torchtitan/models/gpt_oss/model/moe.py
+++ b/torchtitan/models/gpt_oss/model/moe.py
@@ -268,10 +268,13 @@ class GptOssGroupedExperts(nn.Module):
         tp_degree = self._get_tp_degree()
 
         if self.use_grouped_mm:
-            if (
-                not self._is_dtensor
-                or "ep" not in self.mlp1_weight.device_mesh.mesh_dim_names
-            ):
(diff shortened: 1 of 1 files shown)


eous:
> existing implementation already supports expert bias, what is this PR trying to achieve?

Apologies, this appears to be an experiment I was running that got sweeped into its own change. Closing it out.

PR #2216, opened 2026-01-08
feat(gpt-oss): add YaRN RoPE extensions with mscale for extended context

eous:
Implement YaRN (Yet another RoPE extensioN) for context length extension:
- Add frequency correction functions for smooth RoPE interpolation
- Implement precompute_rope_cache with YaRN scaling (rope_factor, beta_fast/slow)
- Add mscale attention scaling factor for numerical stability at extended lengths
- Apply mscale to softmax_scale in Attention module

YaRN allows extending GPT-OSS from 4096 to 131072 tokens (32x) while maintaining perplexity. The mscale factor (0.1 * ln(rope_factor) + 1.0) compensates for the attention distribution changes at longer contexts.

### DIFF
--- a/torchtitan/models/gpt_oss/model/model.py
+++ b/torchtitan/models/gpt_oss/model/model.py
@@ -62,9 +62,19 @@ def precompute_rope_cache(
 
         # Compute correction range (NTK by parts)
         d_half = dim / 2
-        low = d_half * math.log(original_seq_len / (ntk_beta * 2 * math.pi)) / math.log(base)
-        high = d_half * math.log(original_seq_len / (ntk_alpha * 2 * math.pi)) / math.log(base)
-        assert 0 < low < high < d_half - 1, f"Invalid YaRN params: 0 < {low} < {high} < {d_half - 1}"
+        low = (
+            d_half
+            * math.log(original_seq_len / (ntk_beta * 2 * math.pi))
+            / math.log(base)
+        )
+        high = (
+            d_half
+            * math.log(original_seq_len / (ntk_alpha * 2 * math.pi))
+            / math.log(base)
+        )
+        assert (
+            0 < low < high < d_half - 1
+        ), f"Invalid YaRN params: 0 < {low} < {high} < {d_half - 1}"
 
         # Linear ramp for smooth interpolation between low and high
         ramp = (torch.arange(d_half, dtype=torch.float32) - low) / (high - low)


torchtitan/models/gpt_oss/model/model.py
```diff
@@ -22,10 +24,96 @@
...
+
+    # YaRN scaling for extended context (applied when seqlen > original_seq_len)
+    if max_seq_len > original_seq_len and rope_factor > 1.0:
+        low, high = _find_yarn_correction_range(
+            beta_fast, beta_slow, dim, base, original_seq_len
+        )
+        smooth = 1 - _linear_ramp_factor(low, high, dim // 2)
+        # Apply YaRN interpolation: blend scaled and unscaled frequencies
+        freqs = freqs / rope_factor * (1 - smooth) + freqs * smooth
+
```
shuhuayu, line 78: Can you put all logics into this one function of `precompute_rope_cache` . For now, I do not see the need to use four additional separate functions to handle this scaling logic. You may refer to openai's implementation here: https://github.com/openai/gpt-oss/blob/599476783c6f88508dab8577808b5ead5cbee8d2/gpt_oss/torch/model.py#L85. The output of this function would be similar to `_compute_cos_sin` in openai's implementation, and the `concentration` in their implementation can be hided in our case.

torchtitan/models/gpt_oss/model/model.py
```diff
@@ -22,10 +24,96 @@
...
+        original_seq_len: Original pretrained context length (GPT-OSS uses 4096)
+
+    Returns:
+        Precomputed RoPE cache: [max_seq_len, dim * 2] (cos and sin concatenated)
+    """
+    # Basic RoPE frequency calculation
     freqs = 1.0 / (base ** (torch.arange(0, dim, 2)[: (dim // 2)].float() / dim))
+
+    # YaRN scaling for extended context (applied when seqlen > original_seq_len)
+    if max_seq_len > original_seq_len and rope_factor > 1.0:
```
shuhuayu: why the first condition is needed?

shuhuayu:
@eous thanks for the fix! Left some comments.

shuhuayu:
We can use most parts of the yarn implementations in deepseek v3 except one difference. When computing pairs to rotate in hidden dimension, for example when it is `[1, 2, 3, 4, 5, 6]`, in deepseek v3, it uses pairs `[1, 2], [3, 4], [5, 6]`, see https://github.com/deepseek-ai/DeepSeek-V3/blob/9b4e9788e4a3a731f7567338ed15d3ec549ce03b/inference/model.py#L390; in gptoss it uses pairs `[1, 4], [2, 5], [3, 6]`, see https://github.com/openai/gpt-oss/blob/599476783c6f88508dab8577808b5ead5cbee8d2/gpt_oss/torch/model.py#L57. 

For Qwen3, I double checked that all our listed models do not enable RoPE scaling per HuggingFace configs. We can add it make optional (by default set it as off) since some Qwen 3 models do require `YaRN` type RoPE scaling like this one https://huggingface.co/Qwen/Qwen3-14B-AWQ/blob/refs%2Fpr%2F2/config.json.

tianyu-l:
@shuhuayu 
I see. It sounds that if we keep two rotary embedding impls, we need two corresponding yarn impls.

shuhuayu:
> @shuhuayu I see. It sounds that if we keep two rotary embedding impls, we need two corresponding yarn impls.

Yes, we have two ways to apply RoPE as we see above, and we can couple it with different RoPE scaling techniques along hidden dimension, since they are orthogonal. After this fix, as you suggested we may consider consolidating the rotary embeddings implementations into `models/utils.py`, where one can use configs to choose how to apply RoPE, and what types of scaling to use.

eous:
Thanks @shuhuayu for pointing me to the official GPT-OSS implementation. While verifying our implementation against it, I discovered our mscale (concentration) was being applied incorrectly — we had it on softmax_scale instead of on the cos/sin values in the RoPE cache. The official applies:

cos = freqs.cos() * concentration
sin = freqs.sin() * concentration

This gives mscale² scaling on attention logits (since both Q and K are rotated). Our previous approach only achieved mscale scaling — about 25% weaker than intended for rope_factor=32. I pushed a fix and a followup commit to address review feedback.

shuhuayu:
@eous Thanks for the update, LGTM! Please rebase to fix conflict and lint.

shuhuayu:
@eous There is still a lint issue, can you help fix?

eous:
Sorry about the delay, ran ufmt and it made a couple of corrections.

PR #2219, opened 2026-01-09
[Experimental][rl][vllm compat] Update simple_rl example to work with vLLM nightly

Lucaskabela:
See title; running with the latest version of vLLM nightly ran into a few issues as these APIs have changed and within torch titan some paths have changed

### Test plan
```
VLLM_BATCH_INVARIANT=1 VLLM_FLASH_ATTN_VERSION=3 python -m torchtitan.experiments.rl.vllm_compat.simple_rl
```

### Results
```

Step   2 | Loss: -0.0033 | Reward: +0.213 | Samples: 160
  Sample:  Natalia sold 48 clips in April. In May, she sold half as many as she sold in Ap...
Converted to 311 vLLM weights
```

### DIFF
--- a/torchtitan/experiments/rl/vllm_compat/README.md
+++ b/torchtitan/experiments/rl/vllm_compat/README.md
@@ -75,6 +75,7 @@ init_batch_invariance()
 ### Quick Start
 
 ```python
+from vllm.attention.backends.registry import AttentionBackendEnum
 import torch
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 from torchtitan.experiments.rl.vllm_compat import (
@@ -83,7 +84,12 @@ from torchtitan.experiments.rl.vllm_compat import (
 )
 
 # 1. Enable deterministic mode
-init_batch_invariance()
+try:
+    init_batch_invariance()
+except TypeError:
+    # Fallback to new version requiring positional arg for attention
+    init_batch_invariance(AttentionBackendEnum.FLASH_ATTN)
+
 enable_batch_invariant_backward_mode()
 
 # 2. Load model
@@ -95,7 +101,7 @@ model_args = Qwen3ModelArgs(
     n_kv_heads=2,
     vocab_size=151936,
 )
-model = Qwen3VLLMCompatModel(model_args)
+model = Qwen3VLLMCompatModel(model_args).to('cuda').to(torch.bfloat16)
 
 # 3. Forward pass (deterministic)
 input_ids = torch.randint(0, 151936, (2, 128), device='cuda')
@@ -104,6 +110,8 @@ logits = model(input_ids)
 # 4. Backward pass
 loss = logits.sum()
 loss.backward()
+
+print("Done running simple model")
 ```
 
 ### Full RL Training
--- a/torchtitan/experiments/rl/vllm_compat/README.md
+++ b/torchtitan/experiments/rl/vllm_compat/README.md
@@ -51,8 +51,12 @@ Note: Currently supports single-device training only.
 ### Prerequisites
 
 ```bash
-# Install vLLM with deterministic support
-pip install vllm
+# Install vLLM with deterministic support (from source)
+git clone https://github.com/vllm-project/vllm.git
+cd vllm
+python use_existing_torch.py
+uv pip install -r requirements/build.txt
+uv pip install --no-build-isolation -e .
 
 # Install TorchTitan (from the repository root)
 pip install -e .
@@ -75,21 +79,17 @@ init_batch_invariance()
 ### Quick Start
 
 ```python
-from vllm.attention.backends.registry import AttentionBackendEnum
-import torch
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
+from vllm.v1.attention.backends.registry import AttentionBackendEnum
+
+import torch
 from torchtitan.experiments.rl.vllm_compat import (
     enable_batch_invariant_backward_mode,
     Qwen3VLLMCompatModel,
 )
 
 # 1. Enable deterministic mode
-try:
-    init_batch_invariance()
-except TypeError:
-    # Fallback to new version requiring positional arg for attention
-    init_batch_invariance(AttentionBackendEnum.FLASH_ATTN)
-
+init_batch_invariance(AttentionBackendEnum.FLASH_ATTN)
 enable_batch_invariant_backward_mode()
 
 # 2. Load model
@@ -112,6 +112,7 @@ loss = logits.sum()
 loss.backward()
 
 print("Done running simple model")
+
 ```
 
 ### Full RL Training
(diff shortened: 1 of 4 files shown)


Lucaskabela:
cc @wwwjn  @acisseJZhong for review

Lucaskabela:
cc @bwasti for review on this actually

torchtitan/experiments/rl/vllm_compat/README.md
```diff
@@ -83,7 +84,12 @@ from torchtitan.experiments.rl.vllm_compat import (
 )
 
 # 1. Enable deterministic mode
-init_batch_invariance()
+try:
+    init_batch_invariance()
+except TypeError:
+    # Fallback to new version requiring positional arg for attention
```
tianyu-l: Can we converge to one single way instead of maintaining backward compatible code for vllm?

Is it OK if we always run with latest vllm nightly? @jessicazhongeee @zhxchen17

torchtitan/experiments/rl/vllm_compat/README.md
```diff
@@ -95,7 +101,7 @@ model_args = Qwen3ModelArgs(
     n_kv_heads=2,
     vocab_size=151936,
 )
-model = Qwen3VLLMCompatModel(model_args)
+model = Qwen3VLLMCompatModel(model_args).to('cuda').to(torch.bfloat16)
```
tianyu-l, line 104: curious
- why in the past we don't need this but now we need? Is it because previously it was AI-generated and it actually wouldn't work?
- why making it bfloat16? I'm assuming we need to align train and inference, but do we know that inference is everywhere bf16? e.g. how about RMSNorm in vllm.

torchtitan/experiments/rl/vllm_compat/simple_rl.py
```diff
@@ -38,9 +38,14 @@
 from transformers import AutoConfig, AutoTokenizer
 
 from vllm import LLM, SamplingParams
+from vllm.attention.backends.registry import AttentionBackendEnum
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 
-init_batch_invariance()
+try:
```
tianyu-l: similar here

torchtitan/experiments/rl/vllm_compat/README.md
Lucaskabela: Seconding here - if we want to only support nightly we can remove this try/except

torchtitan/experiments/rl/vllm_compat/README.md
Lucaskabela: * I am not sure why it didn't work, but when I tried running it it did not. These lines were necessary to make the example in the readme run correctly
* This is the dtype that the model is expected in (see https://huggingface.co/Qwen/Qwen3-0.6B model card tensortype).  Without it, the code does not run with this error:
```
  File "/home/lucaskabela/torchtitan/torchtitan/experiments/rl/vllm_compat/models/attention.py", line 180, in forward
    output_varlen = FlashAttnWithBackward.apply(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucaskabela/pytorch/torch/autograd/function.py", line 583, in apply
    return super().apply(*args, **kwargs)  # type: ignore[misc]
  File "/home/lucaskabela/torchtitan/torchtitan/experiments/rl/vllm_compat/models/attention.py", line 78, in forward
    output = flash_fn(
             ^^^^^^^^^
  File "/home/lucaskabela/vllm/vllm/vllm_flash_attn/flash_attn_interface.py", line 279, in flash_attn_varlen_func
    out, softmax_lse, _, _ = torch.ops._vllm_fa3_C.fwd(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucaskabela/pytorch/torch/_ops.py", line 1243, in __call__
    return self._op(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: FlashAttention only supports fp16, bf16, and fp8_e4m3 data type
```

Lucaskabela:
cc @tianyu-l  @wwwjn  for re-review :)

torchtitan/experiments/rl/vllm_compat/README.md
wwwjn: > This is the dtype that the model is expected in (see https://huggingface.co/Qwen/Qwen3-0.6B model card tensortype

This indicating the model weight in the safetensor file is stored as bf16. Looking at the error message, it's complaining the input (activation) of flash_attn kernel needs to be bf16. In torchtitan, we can keep the model weights float32 but make the activations to be bf16

So I also want to understand why we need to do this and explicitly convert model weights to bf16

torchtitan/experiments/rl/vllm_compat/batch_invariant_backward.py
```diff
@@ -62,10 +62,14 @@ def forward(ctx, x):
...
             output: silu(gate) * up, shape [..., hidden_dim]
         """
+        from vllm.config import set_current_vllm_config, VllmConfig
         from vllm.model_executor.layers.activation import SiluAndMul as VLLMSiluAndMul
 
         # Use vLLM's implementation for forward
-        vllm_silu_and_mul = VLLMSiluAndMul()
+        # vLLM custom ops require a config context to be set
+        # Since these are parameter free we instantiate default config
+        with set_current_vllm_config(VllmConfig()):
```
wwwjn, line 71: For this part I'm not sure , cc @bwasti for review

torchtitan/experiments/rl/vllm_compat/simple_rl.py
```diff
@@ -170,6 +172,7 @@ def update_weights(self, vllm_compat_state: dict) -> None:
                 gpu_memory_utilization=0.3,  # Reduced from 0.5
                 seed=42,  # Fixed seed for determinism
                 enforce_eager=True,
+                attention_config={"backend": AttentionBackendEnum.FLASH_ATTN},
```
wwwjn, line 175: Why we need this `AttentionBackendEnum` again  in vLLM Engine params?

wwwjn (approved):
I tested reinstall from source using vllm latest code, the AttentionBackend changes are necessary

torchtitan/experiments/rl/vllm_compat/README.md
Lucaskabela: From the error, the flash attention backend is not compatible with float32 (which to my understanding was the backend we are using)

torchtitan/experiments/rl/vllm_compat/simple_rl.py
Lucaskabela: Requirement of the new API

Lucaskabela (commented):
See response

PR #2221, opened 2026-01-13
[rl] refactor save and load model weights using DCP

wwwjn:
* #2244
* __->__ #2221
* #2194
* #2191

## What's new in this PR
- Directly passing weights as tensor (plain tensor) from trainer to generator. 
- Remove the burden of writing and reading from files.
- Supported test: Trainer supports DDP, generator only supports TP=1 (no DTensor in both sides yet)

### DIFF
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -118,9 +118,12 @@ class Generator(Actor, Configurable):
         self.model_spec = model_spec
 
         # Register TorchTitan model with vLLM before any engine creation
-        from torchtitan.experiments.rl.unified.plugin import register
+        from torchtitan.experiments.rl.unified.plugin import (
+            register_model_to_vllm_model_registry,
+            VLLM_MODEL_NAME,
+        )
 
-        register(model_spec)
+        register_model_to_vllm_model_registry(model_spec)
 
         # Set vLLM environment variables from config before any vLLM initialization
         if batch_invariant_mode:
@@ -150,8 +153,7 @@ class Generator(Actor, Configurable):
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=config.enforce_eager,
             seed=config.seed,
-            # TODO: make this field configurable and align with model registration
-            hf_overrides={"architectures": ["Qwen3TorchTitanForCausalLM"]},
+            hf_overrides={"architectures": [VLLM_MODEL_NAME]},
             attention_config=AttentionConfig(
                 backend=AttentionBackendEnum.FLASH_ATTN,
             ),
--- a/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
@@ -23,15 +23,12 @@ from torch.distributed.checkpoint.state_dict import (
     StateDictOptions,
 )
 
-from torchtitan.config import ParallelismConfig, TrainingConfig
+from torchtitan.config import ParallelismConfig
 from torchtitan.distributed.parallel_dims import ParallelDims
 from torchtitan.experiments.rl.unified.models.attention import (
     replace_with_vllm_attention,
 )
-from torchtitan.protocols.model import BaseModel
-from torchtitan.protocols.model_spec import ParallelizeFunction
-from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
-from torchtitan.trainer import Trainer
+from torchtitan.protocols.model_spec import ModelSpec
 
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
@@ -40,14 +37,14 @@ from vllm.logger import init_logger
 logger = init_logger(__name__)
 
 
-def create_torchtitan_config_from_vllm(
+def create_torchtitan_config_from_vllm_config(
     vllm_config: VllmConfig,
-) -> tuple[ParallelDims, Trainer.Config]:
+) -> tuple[ParallelDims, ParallelismConfig]:
     """
-    Create ParallelDims and Trainer.Config from vLLM configuration.
+    Create ParallelDims and ParallelismConfig from vLLM configuration.
 
-    Maps vLLM parallelism settings to TorchTitan's ParallelDims and Trainer.Config
-    so that TorchTitan's parallelize functions can be called with the correct kwargs.
+    Maps vLLM parallelism settings to TorchTitan's config objects so that
+    TorchTitan's parallelize functions can be called with the correct kwargs.
 
     This is needed because vLLM doesn't separate model creation and parallelism
     application — it requires parallelization inside the model constructor
@@ -57,7 +54,7 @@ def create_torchtitan_config_from_vllm(
         vllm_config: vLLM configuration object
 
     Returns:
-        Tuple of (ParallelDims, Trainer.Config) mapped from vLLM config
+        Tuple of (ParallelDims, ParallelismConfig) mapped from vLLM config
 
     Note:
         vLLM doesn't use FSDP sharding (dp_shard=1) or expert parallelism (ep=1, etp=1)
@@ -77,8 +74,7 @@ def create_torchtitan_config_from_vllm(
         world_size=world_size,
     )
 
-    config = Trainer.Config()
-    config.parallelism = ParallelismConfig(
+    parallelism = ParallelismConfig(
         data_parallel_replicate_degree=parallel_config.data_parallel_size,
         data_parallel_shard_degree=1,
         context_parallel_degree=parallel_config.decode_context_parallel_size,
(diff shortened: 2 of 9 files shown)


torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -79,124 +78,62 @@ def __init__(
         self.job_config = job_config
         self.base_model_path = model_path
 
-        self.temp_model_dir = os.path.abspath(
-            os.path.join(job_config.job.dump_folder, "vllm_temp_model")
+        # Load TorchTitan plugin at runtime
+        from torchtitan.experiments.rl.unified.plugin import register
```
acisseJZhong: use `from torchtitan.experiments.rl.unified import register`?
can we move the import statement to the header?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -314,10 +304,63 @@ def load_weights(self, weights_iter):
...
+        # Load HF state dict using DCP
+        hf_state_dict = adapter.to_hf(self.model.state_dict())
+        dcp.load(hf_state_dict, storage_reader=storage_reader)
+
+        # Convert HF state dict to TorchTitan format
+        torchtitan_state_dict = adapter.from_hf(hf_state_dict)
+
+        return self.load_weights_from_state_dict(torchtitan_state_dict)
+
+    def load_weights(self, weights_iter):
```
acisseJZhong, line 370: shall we load weights from `weights_iter`?

fegin:
I'm wondering that should we refactor TorchTitan checkpointer so that it can be directly used in this case. While the current PR work, if TorchTitan migrates to a new checkpoint library other use cases need the same updates as well. This is more future work, not blocking this PR.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -275,37 +283,19 @@ def compute_logits(
 
         return logits
 
-    def load_weights(self, weights_iter):
+    def load_weights_from_state_dict(self, titan_state_dict):
```
tianyu-l: `titan_state_dict` is ambiguous -- both sides should be titan models.
What other name could we use, e.g. `trainer_state_dict`?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -79,124 +78,62 @@ def __init__(
...
-        # Save to temp model directory
-        import os
-
-        checkpoint_path = os.path.join(self.temp_model_dir, "model.safetensors")
 
-        # Update the shard files that vLLM will actually load
-        # We need to split our weights to match the original 2-shard structure
-        import glob
+        # directly update model weights in place
+        load_weights = self._get_model().load_weights_from_state_dict(state_dict)
```
tianyu-l: - IIUC this only works when trainer and generator are on exactly the same global mesh. Is it right?
- Is it true that this has been the assumption before this PR? I.e. is our current monarch script only allow colocated trainer and generator?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: No I didn't find a way to do it iteratively

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: 1. I think so but this part need a more closer look and numerics test, and as we are passing DTensors directly, if trainer / generator on different mesh, these will causing issue.
2. Collocated is never an assumption. From my read, the trainer and generator are created on different mesh in `simple_grpo.py`. cc @allenwang28 

I will took a detailed look into current mechanism and numerics verification, not only e2e bit-wise identity check.

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn, line 126: @allenwang28  n00b question, can you help check if the trainer and generator are located on different set of GPUs?

How should we spawn collocated trainer / generator using monarch?

allenwang28: ah, so we are spinning up trainer/generator on different processes but `"gpus"`  is unfortunately not a literal term for Monarch

We need to set `CUDA_VISIBLE_DEVICES`

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -177,11 +181,13 @@ def _extend_rope_cache_if_needed(
         return rope_cache
 
     def embed_input_ids(self, input_ids: torch.Tensor) -> torch.Tensor:
-        """Convert input token IDs to embeddings."""
+        """vLLM required API.
+        Convert input token IDs to embeddings."""
         return self.model.tok_embeddings(input_ids)
 
     def get_input_embeddings(self, input_ids: torch.Tensor) -> torch.Tensor:
```
acisseJZhong, line 220: why do we need this function? seems duplicated with `embed_input_ids`?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -82,86 +81,60 @@ def __init__(
...
-        Update vLLM model weights from vLLM-compat state dict.
+        Update vLLM actor model weights from torchtitan state dict.
 
         This converts weights to vLLM format, saves them, and reloads using
-        vLLM's reload_weights() API after updating the model path config.
+        vLLM's reload_weights() API.
 
         Args:
-            vllm_state: vLLM model state dict
+            state_dict: vLLM model state dict
```
acisseJZhong: this is actually torchtitan state dict right?

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: I invest more, in this PR the state_dict is in plain tensor - I haven't add torchtitan as the trainer , so I only tested the RL loop with DDP on trainer-side , not parallelism on generator side. So here we are sending plain tensor, which is serializable.

In next PR, to avoid serialize DTensor, I made the trainer and generator colocated , and unwrap the DTensor before sending it.  With the strong assumption that trainer / generator supports same parallelism and collocated now, I wrap the Dtensor back on the receiver side

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -114,6 +115,9 @@ def __init__(
             job_config=self.parallel_config,
         )
 
+        # Initial load model weights from HuggingFace checkpoint path
```
tianyu-l, line 163: Do we really need this complication?
IIUC we can do
- trainer loads checkpoint (can be HF one thanks to torchtitan sd adapters)
- generator only get weigts from trainer, even the initial one
- generator never needs to know HF checkpoints or worry about the conversion

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -313,10 +304,63 @@ def load_weights(self, weights_iter):
...
+        # Load HF state dict using DCP
+        hf_state_dict = adapter.to_hf(self.model.state_dict())
+        dcp.load(hf_state_dict, storage_reader=storage_reader)
+
+        # Convert HF state dict to TorchTitan format
+        torchtitan_state_dict = adapter.from_hf(hf_state_dict)
+
+        return self.load_weights_from_state_dict(torchtitan_state_dict)
+
+    def load_weights(self, weights_iter):
```
tianyu-l, line 370: Do we still need this function? Can it be deleted or overridden with "NotImplementedError"?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: This is one required vllm's API for different version of vllm engine

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: This is to make sure when it maintains the same expectation whenever / wherever you initialize a `LLMEngine(model_path=<checkpointer_folder>, hf_override={"model": "TorchtitanVLLMModel"})`. My "expectation" to this call is the underlaying model is initialized with the weights from checkpoint folder. 
- vllm achieve this "expectation" by calling `TorchtitanVLLMWrapper.load_weights()` when initializing the LLMEngine.
- We achieve this "expectation" by: 1) make `TorchtitanVLLMWrapper.load_weights()` a no-op, and 2) Call `_initial_load_weights` in Wrapper's `__init__`. 

This "expectation" would help inference as well. 

The way you described would work perfectly in RL, but people might implicitly ignore the model weights are not initialized during inference. wdyt?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: This function will still be called internally by vLLM

PR #2222, opened 2026-01-13
[ROCm] Support mxfp8 on gfx950.

RuibinCheung:
* Support mxfp8 on gfx950.

It depends on TorchAO (https://github.com/pytorch/ao/pull/3620).

### DIFF
--- a/torchtitan/components/quantization/mx.py
+++ b/torchtitan/components/quantization/mx.py
@@ -41,7 +41,7 @@ class MXLinearConverter(QuantizationConverter):
         # Can be removed if we enable the emulated versions
         assert has_cuda_capability(10, 0) or has_rocm_capability(
             9, 5
-        ), "MXFP8 is only supported on SM100 or later or ROCm gfx950 or later"
+        ), "MXFP8 is only supported on CUDA SM100 or later, or ROCm gfx950 or later"
 
         # TP not yet supported with torch.compile
         model_compile_enabled = (
--- a/torchtitan/tools/utils.py
+++ b/torchtitan/tools/utils.py
@@ -25,12 +25,9 @@ def has_cuda_capability(major: int, minor: int) -> bool:
     )
 
 
-def is_rocm() -> bool:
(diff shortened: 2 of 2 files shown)


RuibinCheung:
Ready review because the depend PR is merged into TorchAO.

torchtitan/tools/utils.py
```diff
@@ -25,6 +25,17 @@ def has_cuda_capability(major: int, minor: int) -> bool:
     )
 
 
+def is_rocm() -> bool:
```
tianyu-l: If it's only used in the function below, can we merge the code into it, instead of making it standalone? We can revisit this later when things change.

torchtitan/components/quantization/mx.py
```diff
@@ -39,9 +39,9 @@ def __init__(self, job_config: JobConfig, parallel_dims: ParallelDims):
             )
 
         # Can be removed if we enable the emulated versions
-        assert has_cuda_capability(
-            10, 0
-        ), "MXFP8 is only supported on SM100 or architectures"
+        assert has_cuda_capability(10, 0) or has_rocm_capability(
+            9, 5
+        ), "MXFP8 is only supported on SM100 or later or ROCm gfx950 or later"
```
tianyu-l: please mention counterpart of ROCm, e.g.
```suggestion
        ), "MXFP8 is only supported on CUDA SM100 or later, or ROCm gfx950 or later"
```
or Blackwell

torchtitan/tools/utils.py
RuibinCheung: ACK. Thx.

torchtitan/components/quantization/mx.py
RuibinCheung: ACK. Thx.

RuibinCheung:
Seems the Pyrefly check failed but I don't think my change affect these code. Could you help me take a look? @tianyu-l

PR #2226, opened 2026-01-13
[Experimental][rl][unified] Update infer.py example to work with vLLM nightly

Lucaskabela:
NOTE: This PR should be landed on top of https://github.com/pytorch/torchtitan/pull/2219

Few changes seem to be needed to get working; namely:

1) Update to use `spawn` method for CUDA fork
2) Modify attention import to be correct after refactor
3) Modify attention dimensions to be correct after refactor

### Test Plan
Follow steps in the readme then run
```
python torchtitan/experiments/rl/unified/infer.py --model-ckpt-path torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B
```

### Output

```
[2026-01-13 15:20:58] INFO infer.py:113: Generation complete

Prompt: Hello, my name is
Generated text: " Josh and I'm in the middle of a project to develop a hybrid mobile app. I'm looking for guidance on how to go about using modular frameworks. I want to use React and Vue. I need to decide on the framework to use. Can you help me choose the right framework and suggest some best practices for using them?\n\nAdditionally, I want to know what are the best practices for using a modular framework in the context of a web application? Also, what are the best practices for using a"
```

### DIFF
--- a/torchtitan/experiments/rl/unified/infer.py
+++ b/torchtitan/experiments/rl/unified/infer.py
@@ -5,11 +5,16 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
+
+# Must set spawn method before any CUDA operations or vLLM imports
+# CUDA cannot be re-initialized in forked subprocesses
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
 import argparse
 
 # Import unified module - this automatically registers TorchTitan models with vLLM
 from torchtitan.experiments.rl import unified  # noqa: F401
-
 from vllm import LLM, SamplingParams
 from vllm.logger import init_logger
 
--- a/torchtitan/experiments/rl/unified/infer.py
+++ b/torchtitan/experiments/rl/unified/infer.py
@@ -9,6 +9,7 @@ import os
 
 # Must set spawn method before any CUDA operations or vLLM imports
 # CUDA cannot be re-initialized in forked subprocesses
+# See also https://docs.vllm.ai/en/v0.8.3/design/multiprocessing.html#python-multiprocessing
 os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
 
 import argparse
--- a/torchtitan/experiments/rl/unified/models/attention.py
+++ b/torchtitan/experiments/rl/unified/models/attention.py
@@ -75,7 +75,8 @@ class VLLMAttention(torch.nn.Module):
             output: [batch, num_heads, seq_len, head_dim]
         """
         # Input is (batch, num_heads, seq_len, head_dim)
-        # TODO: may be good to use einops in future
+        # TODO: may be good to use einops in future as we can explicitly reshape
+        # with dimension names - see https://github.com/arogozhnikov/einops
         batch_size, num_heads, seq_len, head_dim = q.shape
         _, num_kv_heads, _, _ = k.shape
 
@@ -85,6 +86,7 @@ class VLLMAttention(torch.nn.Module):
         k = k.transpose(1, 2)
         v = v.transpose(1, 2)
 
+        # TODO: reimplement as a 4d tensor once vLLM fix has landed
         # Then flatten batch and seq_len: (batch * seq_len, num_heads, head_dim)
         q = q.reshape(batch_size * seq_len, num_heads, head_dim)
         k = k.reshape(batch_size * seq_len, num_kv_heads, head_dim)
--- a/torchtitan/experiments/rl/unified/models/attention.py
+++ b/torchtitan/experiments/rl/unified/models/attention.py
@@ -75,17 +75,26 @@ class VLLMAttention(torch.nn.Module):
             output: [batch, num_heads, seq_len, head_dim]
         """
         # Input is (batch, num_heads, seq_len, head_dim)
+        # TODO: may be good to use einops in future
         batch_size, num_heads, seq_len, head_dim = q.shape
+        _, num_kv_heads, _, _ = k.shape
 
-        # Transpose to (batch, seq_len, num_heads, head_dim) for vLLM
+        # vLLM expects (num_tokens, num_heads, head_dim) where num_tokens = batch * seq_len
+        # First transpose to (batch, seq_len, num_heads, head_dim)
         q = q.transpose(1, 2)
         k = k.transpose(1, 2)
         v = v.transpose(1, 2)
 
-        output_varlen = self.vllm_attn(q, k, v)
+        # Then flatten batch and seq_len: (batch * seq_len, num_heads, head_dim)
+        q = q.reshape(batch_size * seq_len, num_heads, head_dim)
+        k = k.reshape(batch_size * seq_len, num_kv_heads, head_dim)
+        v = v.reshape(batch_size * seq_len, num_kv_heads, head_dim)
 
-        # Reshape back to batch format
-        output = output_varlen.view(batch_size, seq_len, num_heads, head_dim)
+        # vLLM attention returns (num_tokens, hidden_size) where hidden_size = num_heads * head_dim
+        output_flat = self.vllm_attn(q, k, v)
+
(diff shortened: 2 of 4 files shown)


Lucaskabela:
cc @wwwjn @acisseJZhong for review :)

acisseJZhong:
LGTM! thanks for the fix. Curious why vllm's Attention requires batch and seq_len dim flattened? Is attention input shape (batch, seq_len, num_heads, head_dim) no longer working?

Lucaskabela:
> LGTM! thanks for the fix. Curious why vllm's Attention requires batch and seq_len dim flattened? Is attention input shape (batch, seq_len, num_heads, head_dim) no longer working?

Yes I got this particular error: 
```
File "/home/lucaskabela/torchtitan/torchtitan/experiments/rl/unified/models/attention.py", line 88, in forward
(EngineCore_DP0 pid=3139209) ERROR 01-13 15:17:35 [core.py:936]     output = output_varlen.view(batch_size, seq_len, num_heads, head_dim)
(EngineCore_DP0 pid=3139209) ERROR 01-13 15:17:35 [core.py:936]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3139209) ERROR 01-13 15:17:35 [core.py:936] RuntimeError: shape '[1, 16384, 16, 128]' is invalid for input of size 2048
```

Let me see if I can link the actual pr doing the attention change on vLLM side

Lucaskabela:
Ah sorry - that actually seems like just an output issue, probably doesn't need the input reshape let me see if I can refactor it simpler

torchtitan/experiments/rl/unified/infer.py
```diff
@@ -5,11 +5,16 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
+
+# Must set spawn method before any CUDA operations or vLLM imports
+# CUDA cannot be re-initialized in forked subprocesses
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
```
tianyu-l, line 13: > Update to use spawn method for CUDA fork

what does this do? could you add links to vllm documentation?

Lucaskabela:
I took a look, and seems that this refactor (https://github.com/vllm-project/vllm/pull/31596) caused the 4d input to not be handled correctly - only 2d and 3d input is handled so this change seems necessary

cc @acisseJZhong

torchtitan/experiments/rl/vllm_compat/models/attention.py
```diff
@@ -8,7 +8,15 @@
 import math
 
 import torch
-from vllm.attention.utils.fa_utils import flash_attn_varlen_func
+
+try:
```
tianyu-l: what's the purpose of try/except -- can we always use the latest version?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -75,17 +75,26 @@ def forward(
             output: [batch, num_heads, seq_len, head_dim]
         """
         # Input is (batch, num_heads, seq_len, head_dim)
+        # TODO: may be good to use einops in future
```
tianyu-l: could you say why?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -75,17 +75,26 @@ def forward(
...
         batch_size, num_heads, seq_len, head_dim = q.shape
+        _, num_kv_heads, _, _ = k.shape
 
-        # Transpose to (batch, seq_len, num_heads, head_dim) for vLLM
+        # vLLM expects (num_tokens, num_heads, head_dim) where num_tokens = batch * seq_len
```
tianyu-l, line 83: could you add code pointer or documentation?

tianyu-l (changes requested):

acisseJZhong:
> I took a look, and seems that this refactor ([vllm-project/vllm#31596](https://github.com/vllm-project/vllm/pull/31596)) caused the 4d input to not be handled correctly - only 2d and 3d input is handled so this change seems necessary
> 
> cc @acisseJZhong

looks like there's an ongoing fix for 4d input https://github.com/vllm-project/vllm/pull/32274

torchtitan/experiments/rl/vllm_compat/models/attention.py
```diff
@@ -8,7 +8,15 @@
...
+    from vllm.v1.attention.backends.fa_utils import get_flash_attn_version
+    from vllm.vllm_flash_attn import flash_attn_varlen_func
+except ImportError:
+    from vllm.attention.utils.fa_utils import (
```
wwwjn: Under what situation we might hit this line of import?

torchtitan/experiments/rl/unified/models/attention.py
wwwjn: When I developed this part of code, I read the vllm code / doc and my impression is vllm supports both 3D(num_tokens, num_heads, head_dim) and 4D(bsz, seq_len, num_heads, head_dim) input. If the input is 4D, vllm will convert it before sending into attention kernel.I keep it as 4D to align with torchtitan's shape

As @acisseJZhong  suggested, there are ongoing fix to be landed. With this fix, can we still use 4D input?

torchtitan/experiments/rl/unified/infer.py
Lucaskabela: Sure thing - updating to link the doc in the code as well:

https://docs.vllm.ai/en/v0.8.3/design/multiprocessing.html#python-multiprocessing 

Tl;dr: This sets the python multiprocessing method; spawn will be the default in 3.14 python onwards, and without spawn my machine cannot run.  If this is not an issue other people are having I can just update the readme to mention one may need to prefix their command with this

torchtitan/experiments/rl/vllm_compat/models/attention.py
Lucaskabela: This is just for backwards compatibility - I don't know the standards of BC in torchtitan so we can have some discussion on this.  If we only want to use the latest version, then we can remove the except

torchtitan/experiments/rl/unified/models/attention.py
Lucaskabela: Sure :) Einops is just a more readable way to do reshaping - basically combining the comment and the reshape ops into code

torchtitan/experiments/rl/unified/models/attention.py
Lucaskabela: Presuming that fix lands we definitely can - that said I am not sure if we have a concrete timeline on when this will land so maybe we can leave a todo to refactor back to 4d once the PR lands

torchtitan/experiments/rl/vllm_compat/models/attention.py
Lucaskabela: This is for backwards compatibility - please advise on the projects approach for that.  There is a attention refactor currently in progress (see https://github.com/vllm-project/vllm/pull/31916 landed ~5 days ago)

torchtitan/experiments/rl/unified/infer.py
acisseJZhong: I faced the same "CUDA cannot be re-initialized" issue when running this file, and I fixed it by prefixing the command. I think we can set the env here as well.

torchtitan/experiments/rl/vllm_compat/models/attention.py
wwwjn: As we discussed offline, let's use vllm nightly / source code as the source-of-truth. In torchtitan, we don't have BC promise so it's ok to directly make BC changes

torchtitan/experiments/rl/unified/models/attention.py
wwwjn: I think it's OK to leave a todo and refactor it back, thank you!

Lucaskabela:
Thanks for patience - the lintrunner signal seems preexisting :) ready for re-review

tianyu-l (approved):
Thanks!

PR #2227, opened 2026-01-14
[MoE] Fix experts DTensor metadata bug for dcp

shuhuayu:
Previously, individual experts are marked as `Replicate` in EP dimension in global `global_device_mesh`. Local experts are first created on `global_device_mesh` and are turned into a 2d tensor using  `squeeze(0)`, which only removes the extra dimension, but the remaining metadata `Replicate` is still there. The wrong metadata results in bug when DCP saves `DTensor`. This PR fixes this bug by:
1. Use a sub-mesh that excludes expert dimension, i.e., dim 0. 
2. When sub-mesh is empty, use plain tensor instead of `DTensor`.

### DIFF
--- a/torchtitan/models/utils.py
+++ b/torchtitan/models/utils.py
@@ -257,7 +257,9 @@ class MoEStateDictAdapter(StateDictAdapter):
                 expert_weight = local_grouped_weights[local_expert_index, :, :]
             else:
                 # Use slicing and unsqueeze get a 3D tensor, then create DTensor and squeeze
-                expert_weight_3d = local_grouped_weights[local_expert_index, :, :].unsqueeze(0)
+                expert_weight_3d = local_grouped_weights[
+                    local_expert_index, :, :
+                ].unsqueeze(0)
                 expert_weight = DTensor.from_local(
                     expert_weight_3d,
                     sub_mesh,


torchtitan/models/utils.py
```diff
@@ -202,26 +202,44 @@ def _get_local_experts_weights(
         # Step 2: Store indices for potential future use in from_hf()
         self.local_experts_indices[titan_abstract_key] = (start_index, end_index)
 
-        # Step 3: Create new placements for individual expert weights
-        new_placements = []
+        # Step 3: Identify mesh dimensions that shard on dim-0 (expert dimension)
+        # remove expert dimension
+        # and build new sub-mesh/placements for individual expert weights
+        dim0_mesh_names = []
```
tianyu-l: why do we keep this?

torchtitan/models/utils.py
```diff
@@ -202,26 +202,44 @@ def _get_local_experts_weights(
...
         for i, name in enumerate(device_mesh.mesh_dim_names):
             placement = dtensor_placements[i]
-            if placement.dim == 0:
-                # Convert dim-0 sharding to replication for individual experts
-                new_placements.append(Replicate())
+            if isinstance(placement, Replicate):
+                # Replicate (hybrid) doesn't shard any dim, keep in sub-mesh
+                sub_mesh_names.append(name)
+                sub_placements.append(Replicate())
+            elif isinstance(placement, (Shard, _StridedShard)) and placement.dim == 0:
```
tianyu-l, line 215: this doesn't sound right -- 

Imagine we have (fsdp, ep) mesh, and placements are [_SS(0), S(0)], would this code make it a plain tensor? I think it should be (fsdp,) with S(0).

torchtitan/models/utils.py
shuhuayu: I removed it, it is not necessary.

torchtitan/models/utils.py
wwwjn: Do you mean (dp_shard_mod_ep , dp_shard_in_ep) mesh? The original code make  groupedExpert's weights from [_SS(0), S(0)] to [R, R]. Now with this change, it will make the groupedExpert's weight to a plain tensor.

Can you explain more why the targeted placement should be (fsdp,) with S(0)?

tianyu-l: hmm, my bad, I think you are right

shuhuayu: I think in the (fsdp, ep) case, both are splitting dim0, i.e., expert dim. As a complete expert now is stored in one rank, if we use (fsdp, ) with S(0), this is confusing since the expert tensor is now complete, not a `Shard` along fsdp ranks.

torchtitan/models/utils.py
```diff
@@ -240,15 +258,21 @@ def _get_local_experts_weights(
...
-            expert_dtensor = DTensor.from_local(
-                expert_weight, device_mesh, new_placements, run_check=False
-            ).squeeze(0)
-
-            local_expert_tensors[expert_key] = expert_dtensor
+            if use_plain_tensor:
+                local_expert_tensors[expert_key] = expert_weight
+            else:
+                # Add batch dimension temporarily for DTensor creation
+                expert_weight_3d = expert_weight.unsqueeze(0)
```
tianyu-l: could you remind me why we needed this unsqueeze + squeeze?

shuhuayu: Sure. We use `unsqueeze(0)` and `squeeze(0)` to automatically take care of `Shard(i)` for $i \ge 1$. Turning a 2d tensor to 3d using `unsqueeze(0)` enables us to use `from_local` in the previous global `device_mesh`,  without changing `i` in `Shard(i)`. Then, when we use `.squeeze(0)`, and `dim0` has size 1, `dim0` will be removed, and all `Shard(i)` will change to `Shard(i-1)`. But the `Replicate` in the `device_mesh` will not be removed, which is the problem.

tianyu-l: I see. In the DTensor case, is it more straightforward if we use `torch.index_select` instead of indexing + unsqueeze?

torchtitan/models/utils.py
```diff
@@ -240,15 +258,21 @@ def _get_local_experts_weights(
...
+            expert_weight = local_grouped_weights[local_expert_index, :, :]
 
-            # Create DTensor and remove batch dimension (experts dimension is removed)
-            expert_dtensor = DTensor.from_local(
-                expert_weight, device_mesh, new_placements, run_check=False
-            ).squeeze(0)
-
-            local_expert_tensors[expert_key] = expert_dtensor
+            if use_plain_tensor:
+                local_expert_tensors[expert_key] = expert_weight
```
tianyu-l: It seems this code can be shared.

shuhuayu: Fixed. Thanks for the suggestions.

aghilann:
@shuhuayu 

Were you able to test this end to end? I ran into an NCCL error once training started. This change did fix the earlier OOM issue when loading the model, but I think there may still be a bug remaining.

I have included the stack trace below in case it is helpful. When I disable `initial_load_in_hf`, the run begins training successfully. [logs.txt](https://github.com/user-attachments/files/24664690/logs.txt) having it enabled is important for my use case since I am fine tuning a model.

shuhuayu:
> @shuhuayu
> 
> Were you able to test this end to end? I ran into an NCCL error once training started. This change did fix the earlier OOM issue when loading the model, but I think there may still be a bug remaining.
> 
> I have included the stack trace below in case it is helpful. When I disable `initial_load_in_hf`, the run begins training successfully. [logs.txt](https://github.com/user-attachments/files/24664690/logs.txt) having it enabled is important for my use case since I am fine tuning a model.

Hi @aghilann, thanks for reporting this issue. Do you have a smaller setup where I can reproduce this bug? And just to confirm you are exactly on this commit right? We tested training locally loading from a Qwen3 30B-A3B hf ckpt.

shuhuayu:
Fixed a bug in `from_hf` when plain tensors going to the offline path. @aghilann

aghilann:
Hey @shuhuayu , I'm not on the same commit. I'm on a fork where I've modified some stuff to add LoRA's. I just directly copied the code over, with the changes you added last night I went back to OOM'ing at the `local_tensor = torch.stack(sorted_experts, dim=0)` line. Here's a stack trace if it's useful 
[logs.txt](https://github.com/user-attachments/files/24675877/logs.txt)

I did not OOM when running this on a single node with a debug version of DSV3. Will mention it does work fine when I don't load from HF.

shuhuayu:
> Hey @shuhuayu , I'm not on the same commit. I'm on a fork where I've modified some stuff to add LoRA's. I just directly copied the code over, with the changes you added last night I went back to OOM'ing at the `local_tensor = torch.stack(sorted_experts, dim=0)` line. Here's a stack trace if it's useful [logs.txt](https://github.com/user-attachments/files/24675877/logs.txt)
> 
> I did not OOM when running this on a single node with a debug version of DSV3. Will mention it does work fine when I don't load from HF.

It's kind of difficult to reproduce this particular issue. We will merge the current fix for now, and look into this later if we find it reproducible locally on a torchtitan branch. For now, I would suggest adding an ad hoc fix to replace `torch.stack()`, which creates a new tensor and is memory heavy. Like you can create a empty tensor. and assign values to this empty tensor one expert by one expert, and delete an expert right away once it is copied. Hope that helps.

shuhuayu:
Tested offline conversion, and memory usage before/after fix is similar in a test training loaded from Qwen3 30B-A3B hf ckpt.

PR #2228, opened 2026-01-14
Enable memory snapshot for generic devices

frost-intel:
Pytorch recently added `_record_memory_history` for Intel XPU. This PR makes the memory snapshot device-generic.

@pkourdis @githubsgi

### DIFF
--- a/torchtitan/distributed/dual_pipe_v.py
+++ b/torchtitan/distributed/dual_pipe_v.py
@@ -260,14 +260,13 @@ def overlap_callback(action: _Action, ctx: _PipelineContext):
     )
     # PP computation ========================================================
     _hook_coordinator.enable_coordination(num_layers=min_num_layers)
-    main_stream = torch.accelerator.current_stream(device_module)
+    main_stream = torch.accelerator.current_stream(device_type)
 
     # Shared container for exception from backward thread
     def run_backward():
         # pyrefly: ignore [missing-attribute]
         schedule._assert_unsharded(backward_stage)
         # Set the backward thread to use the same stream as forward
-        # pyrefly: ignore [missing-attribute]
         device_module.set_stream(main_stream)
         with record_function(
             f"backward_stage_{backward_stage_index}_mb_{backward_mb_index}"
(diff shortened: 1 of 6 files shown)


torchtitan/tools/profiling.py
```diff
@@ -104,7 +104,8 @@ def maybe_enable_memory_snapshot(
 
         class MemoryProfiler:
             def __init__(self, step_num: int, freq: int):
-                torch.cuda.memory._record_memory_history(
+                self.device_module = torch.get_device_module()
```
tianyu-l: could you import from https://github.com/pytorch/torchtitan/blob/main/torchtitan/tools/utils.py#L33

tianyu-l:
test failed

frost-intel:
I found device_module typing was incorrect, so I fixed that and the relevant typing overrides.

fegin (approved):

torchtitan/distributed/dual_pipe_v.py
```diff
@@ -260,14 +260,13 @@ def overlap_callback(action: _Action, ctx: _PipelineContext):
     )
     # PP computation ========================================================
     _hook_coordinator.enable_coordination(num_layers=min_num_layers)
-    main_stream = torch.accelerator.current_stream(device_module)
+    main_stream = torch.accelerator.current_stream(device_type)
```
tianyu-l, line 263: nit: this change is not absolutely necessary? https://docs.pytorch.org/docs/stable/generated/torch.accelerator.current_stream.html

OK to change though

frost-intel: After fixing the typing for `device_module`, using it here throws a pyre typing error.

`current_stream` accepts a `device` type, which can be `torch.device | str | int | None` (from [types.py](https://github.com/pytorch/pytorch/blob/1857f59573fa01c48f13b067191e7a2763b89226/torch/types.py#L74) ).

It looks like `torch.cuda` or `torch.xpu` can convert to one of these types, but it's definitely not the intended design, and we should use the correct `device_type` since we already have it.

Issue #2223, opened 2026-01-13
Incorrect transpose in Qwen3 varlen attention backend


francesco-bertolotti:
## incorrect transpose in Qwen3 varlen attention backend

While reviewing the Qwen3 implementation, I noticed that the attention output is always transposed at the end, regardless of which attention backend is used.

For the `flex` and `sdpa` backends, this final transpose appears correct. However, for the `varlen` backend, it seems to introduce an extra transpose that swaps the head dimension with the head size dimension. This is then masked by a subsequent `.view()` that reshapes the tensor back to the expected dimensions, effectively hiding the issue.

Relevant code:

* Qwen3 model output handling:
  [https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/models/qwen3/model/model.py#L271-L298](https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/models/qwen3/model/model.py#L271-L298)
* Attention backend wrappers:
  [https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/models/attention.py#L70-L88](https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/models/attention.py#L70-L88)

The root cause seems to be that the `varlen` backend already transposes the head and head_size dimensions before calling into the backend so that Q/K/V are in the expected layout. As a result, the output from `varlen` should *not* be transposed again. In contrast, the `flex` and `sdpa` wrappers do not perform this pre-transposition, so their outputs do require the final transpose.

A straightforward fix would be to move the final `output.transpose(...)` inside the `match` cases for `sdpa` and `flex`, but exclude it from the `varlen` case.
___

If you can confirm that my understanding is correct, I’m happy to submit a quick PR to address this.

francesco-bertolotti:
This issue is related to this https://github.com/pytorch/torchtitan/issues/2224

tianyu-l:
Thanks

> incorrect transpose in Qwen3 varlen attention backend

to confirm my understanding: is it efficiency issue or correctness issue?

francesco-bertolotti:
There are two separate issues. The first concerns efficiency (see [https://github.com/pytorch/torchtitan/issues/2224](https://github.com/pytorch/torchtitan/issues/2224)). The second concerns modeling correctness.

[...]

PR #2229, opened 2026-01-14
fix sdpa-varlen attention mismatch in qwen3

francesco-bertolotti:
Fix for https://github.com/pytorch/torchtitan/issues/2223


francesco-bertolotti:
@liangel-02 no problem at all. Thank you for your work on this repo. 

I should be able to commit something tomorrow.

francesco-bertolotti:
Hello @liangel-02,

I’ve applied the same fix to the LLaMA-3 model.

I also added a test to verify that the forward pass of LLaMA-3 attention behaves identically for the flex and varlen backends.

Additionally, I ran ablation experiments using sub-1B models for 300 steps. The results appear to be close to the expected behavior, as shown below:

Screenshot 2026-01-15 at 10 48 20 Screenshot 2026-01-15 at 10 47 45

tests/unit_tests/test_gqa.py
```diff
@@ -0,0 +1,102 @@
...
+
+    sdpa_gqa_attention = qwen3.model.model.Attention(
+        sdpa_args := qwen3.model.args.Qwen3ModelArgs(
+            dim = 64,
+            n_heads = 8,
+            n_kv_heads = 4,
+            hidden_dim = 512,
+            attn_type = "sdpa"
+        ),
+    ).to(device="cuda:0", dtype=dtype)
```
tianyu-l: btw: our CI cannot run unit test on GPU, it's CPU only right now.

francesco-bertolotti: Should I remove those tests? I can also do something like `if no gpu available: return`

francesco-bertolotti:
New runs were performed with the seed set to 42 and deterministic mode enabled. Small differences remain, but they are now significantly more contained.

Screenshot 2026-01-15 at 12 48 24
Screenshot 2026-01-15 at 12 51 00

liangel-02:
> New runs were performed with the seed set to 42 and deterministic mode enabled. Small differences remain, but they are now significantly more contained.
> 
> Screenshot 2026-01-15 at 12 48 24 

@francesco-bertolotti 
the qwen3 losses lgtm, but i think the llama ones are still a bit off esp after ~75 steps. for reference this is what the loss curve currently looks like. i think we should at least understand why the varlen loss performs better.

Screenshot 2026-01-15 at 10 57 09 AM

cc @drisspg

tests/unit_tests/test_gqa.py
tianyu-l: If it's not guarded by CI, it's better to remove them. In the future we should make torchtitan test more flexible, but for now showing results before vs. after the change in PR summary should be good.

francesco-bertolotti:
Hi @liangel-02,
I believe that, for some reason, smoothing was messing with the visualization. I have re-run the llama varlen vs fles baselines. Here's what we are at:

flex-varlen-llama

PR #2232, opened 2026-01-14
[DONT LAND] Implement PrefetchedDataloader for overlapped data loading

fegin:
* __->__ #2232

This is a demonstration of how we can overlap to(device) and post dataloading processing with forward/backward/optimizer.  

IGNORE the flux change.

Right now this PR assumes post dataloading processing uses GPU, which will still use another stream other than main stream to compute. While this can make the dataloading processing and forward/backward/optimizer computation run concurrently, the overlapping may slow down the main computation. This can be avoid by move the post dataloading processing to CPU.

This PR is Mostly written by Claude and only serves as a demonstration.

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -10,7 +10,7 @@ import json
 import os
 import time
 from datetime import timedelta
-from typing import Any, Iterable
+from typing import Any, Callable, Iterable
 
 import torch
 import torch.distributed.checkpoint.stateful
@@ -34,6 +34,72 @@ from torchtitan.tools.profiling import (
     maybe_enable_memory_snapshot,
     maybe_enable_profiling,
 )
+
+
+class PrefetchedDataloader:
+    """Prefetches batches using a separate CUDA stream to overlap with compute."""
+
+    def __init__(
+        self,
+        data_iterable: Iterable[tuple[dict[str, torch.Tensor], torch.Tensor]],
+        device_type: str,
+        post_dataloading_process_fn: Callable[
+            [dict[str, torch.Tensor], torch.Tensor],
+            tuple[torch.Tensor, torch.Tensor, dict[str, torch.Tensor], dict[str, Any]],
+        ],
+    ):
+        self.data_iterator = iter(data_iterable)
+        self.device_type = device_type
+        self.post_dataloading_process_fn = post_dataloading_process_fn
+        # pyrefly: ignore [missing-attribute]
+        self.stream = utils.device_module.Stream()
+        self.next_batch: (
+            tuple[torch.Tensor, torch.Tensor, dict[str, torch.Tensor], dict[str, Any]]
+            | None
+        ) = None
+
+    def preload(self) -> None:
+        """Prefetch next batch under self.stream."""
+        try:
+            batch = next(self.data_iterator)
+        except StopIteration:
+            self.next_batch = None
+            return
+
+        input_dict, labels = batch
+
+        # pyrefly: ignore [missing-attribute]
+        with utils.device_module.stream(self.stream):
+            for k, v in input_dict.items():
+                if isinstance(v, torch.Tensor):
+                    input_dict[k] = v.to(self.device_type, non_blocking=True)
+            labels = labels.to(self.device_type, non_blocking=True)
+
+            (
+                inputs,
+                labels,
+                extra_inputs,
+                extra_kwargs,
+            ) = self.post_dataloading_process_fn(input_dict, labels)
+
+        self.next_batch = (inputs, labels, extra_inputs, extra_kwargs)
+
+    def __iter__(self):
+        self.preload()
+        return self
+
+    def __next__(
+        self,
+    ) -> tuple[torch.Tensor, torch.Tensor, dict[str, torch.Tensor], dict[str, Any]]:
+        # pyrefly: ignore [missing-attribute]
+        utils.device_module.current_stream().wait_stream(self.stream)
+
+        if self.next_batch is None:
+            raise StopIteration
+
+        current_batch = self.next_batch
+        self.preload()
+        return current_batch
 
 
 class Trainer(torch.distributed.checkpoint.stateful.Stateful):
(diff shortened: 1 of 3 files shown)


torchtitan/train.py
```diff
@@ -36,6 +36,72 @@
...
+        self,
+    ) -> tuple[torch.Tensor, torch.Tensor, dict[str, torch.Tensor], dict[str, Any]]:
+        # pyrefly: ignore [missing-attribute]
+        utils.device_module.current_stream().wait_stream(self.stream)
+
+        if self.next_batch is None:
+            raise StopIteration
+
+        current_batch = self.next_batch
+        self.preload()
```
wconstab, line 101: this approach issues the kernels (and incurs cpu overhead) for the dataloading operations for next batch before returning the already prefetched batch.

while the dataloading operations are added to a side stream, they are issued to cuda driver before the subsequent forward op kernels on the main stream.  i'm curious how this performs.

did you check if the main stream ends up running its kernels eagerly without much delay under this implementation?

do we have a way to set a lower priority on the stream used by this module?

torchtitan/train.py
```diff
@@ -36,6 +36,72 @@
...
+            self.next_batch = None
+            return
+
+        input_dict, labels = batch
+
+        # pyrefly: ignore [missing-attribute]
+        with utils.device_module.stream(self.stream):
+            for k, v in input_dict.items():
+                if isinstance(v, torch.Tensor):
+                    input_dict[k] = v.to(self.device_type, non_blocking=True)
```
tianyu-l, line 75: Putting post processing aside for a bit: If there is no other explicit sync between device and host, e.g. in dense llama3 training, iteration i+1's dataloading is already overlapping with iteration i's training (because GPU runs longer than CPU). I think this change will help in the case where there are other syncs e.g. MoE/EP, logging, or when it's CPU bounded in general.

tianyu-l: As for https://github.com/pytorch/torchtitan/pull/2206, with grad accumulation of N steps I believe this preload needs to be called N times before any actual training, so the overlapping is mainly between data loading (CPU) and post processing (GPU) themselves.

Another way to do grad accumulation is to load a full batch from data loader in one time, and then split it into multiple microbatches, but this may go further against the async RL data generator idea. cc @wwwjn

tianyu-l: btw JAX training cookbook also mentions this trick https://docs.jax.dev/en/latest/the-training-cookbook.html#data-loading

tianyu-l: besides, JAX mentions in https://docs.jax.dev/en/latest/the-training-cookbook.html#efficiency-via-asynchronous-dispatch that
> What happens next is interesting. Because [jax.jit()](https://docs.jax.dev/en/latest/_autosummary/jax.jit.html#jax.jit)-decorated calls are non-blocking, the call to train_step returns to the Python interpreter immediately. While the computation is enqueued on the accelerator, no work is actually performed yet. The Python loop continues, advancing the step counter and calling next(batch) for the next iteration. Once the second call to train_step is made, its inputs are now the mutated reference to train_state from the previous JIT call and a fresh batch of data. The runtime is clever and sees that in order to execute the second call to train_step, we first need to realize the train_state result of step 0 to perform the mutation. And so it fires off the computation for the first step, and, crucially, while this happens, train_step, once again, returns immediately, and the loop skips over again. Python now runs ahead until it encounters the next(batch) function at step 3, which proceeds to execute in Python, loading data, while the first train step is executing (for real this time). And just like that, we can simultaneously load data and perform math on the accelerator, without any traditional multiprocessing.

I didn't completely get it, but it sounds partly related to `prefetch_factor`? cc @divyanshk if you have insights.

torchtitan/train.py
fegin: Well, this approach should NOT be used to overlap GPU computation (fwd/bwd/opt) with other GPU computation. 

This dataloader pipeline implementation is usually implemented for overlapping GPU computation with the following two operations 1) CPU->GPU sync (if the input batch is not small, e.g., video, 2) CPU ops.

The big picture:

When the **GPU** is running **N** iteration fwd/bwd/optimizer, **CPU** is running **N + 1** iteration 1) dataloading (done by the dataloader), 2) batch post-processing e.g., creating FlexAttention BlockMask, 3) move everything from CPU to GPU. These CPU tasks must run in order -- we need batch to be loaded before performing post-processing before moving data to GPU.

Ideally, the overlapping can be done without interfering the GPU tasks. I just want to demonstrate the idea and didn't completely implement the CPU ops overlapping (should be done in a separate thread).

Let me know if the explanation makes sense to you.  @wconstab

torchtitan/train.py
fegin: > Putting post processing aside for a bit: If there is no other explicit sync between device and host, e.g. in dense llama3 training, iteration i+1's dataloading is already overlapping with iteration i's training (because GPU runs longer than CPU).

Yes, **N+1** dataloading is already overlapping with **N** iteration. But 1) BlockMask creation and 2) CPU->GPU are not overlapping with **N** iteration. The two plus the new post-dataloading are the ones can be overlapped.

Please also check my response to @wconstab comment. @tianyu-l 

> As for https://github.com/pytorch/torchtitan/pull/2206, with grad accumulation of N steps I believe this preload needs to be called N times before any actual training, so the overlapping is mainly between data loading (CPU) and post processing (GPU) themselves.

This is not correct. Based on @wwwjn's implementation, only one preload is needed.

torchtitan/train.py
wconstab: that makes sense, i think. i don't think your approach is bad, i was more asking if you have learned anything about how it behaves (would be interested to see profile trace, etc). anyway, feel free to ignore this comment for now if you have other things to do first.

torchtitan/train.py
tianyu-l: > This is not correct. Based on @wwwjn's implementation, only one preload is needed.

@wwwjn could you confirm? I thought it would run num_microbatch steps of preload.

divyanshk (commented):
Very cool to see this being attempted. @ramanishsingh and I have been talking about eventually enabling async D2H in torch.utils.data - although it might not be trivial giving the need for wide coverage. I can give my high level thoughts on this topic (I still haven't gone through all the discussions on the PR already)

> Right now this PR assumes post dataloading processing uses GPU 

This is usually a big assumption. Typically folks don't want more work on GPU which isn't pure model inference. Right now the space of GPU processing is a bit niche with not every post-processing op being supported by all GPUs. Now, if you want to do a model inference as part of your post-processing, I'd say that would come under "model" work and not data loading. Just that you would have two model inferences (one for post-processing, and one main model run feeding the output of the former). Then the problem becomes pipeline-ing two kernels. Also, doing GPU work as part of dataloading can be tricky because CUDA context can often cause multiprocessing workers (i.e. data loading workers) to crash.

> but it sounds partly related to prefetch_factor

Yes. That is what the pytorch dataloader does. It runs its own background loop to fetch+preprocess+pinning-the-batch-to-page-locked-memory. But it doesn't transfer the data over to the GPU. Yes, pinning memory helps with that significantly. cc. @tianyu-l 

That said, what we can do is have a separate CUDA stream enable host to device data transfer as part of the data loading loop. I think that is what this PR is doing? This way we have the data for the next batch on the GPU before it asks for it. But there are caveats:
    * Here we are assuming the D2H is a bottleneck. It can be, but with pin_memory=True D2H should be fast. Worth noting the pinning of memory is happening in a background CPU loop. The dataloader is pinning N+1, N+2, N+3, ..... batch while the GPU is working on batch N - it depends on how further ahead the CPU can get from the GPU.
    * There can be costs to the new cuda stream creation, syncing etc. Some of theses are highlighted above in the thread. For eg, you need to sync the streams to ensure that batch N is on GPU before doing fwd-bwd on batch N. 
    * Also, we can't just keep throwing batching onto the accelerator. It will cost GPU memory to store batches. So the question becomes if the costs of stream creation + GPU memory + stream sync outweigh the slowdown of D2H.

torchtitan/train.py
fegin: It performs `num_microbatch` calculation of the validate tokens  but it can be done in just one prefetch and record the results in a dictionary.

fegin:
@divyanshk 

> This is usually a big assumption. Typically folks don't want more work on GPU which isn't pure model inference. Right now the space of GPU processing is a bit niche with not every post-processing op being supported by all GPUs. Now, if you want to do a model inference as part of your post-processing, I'd say that would come under "model" work and not data loading. Just that you would have two model inferences (one for post-processing, and one main model run feeding the output of the former). Then the problem becomes pipeline-ing two kernels. Also, doing GPU work as part of dataloading can be tricky because CUDA context can often cause multiprocessing workers (i.e. data loading workers) to crash.

Yes, I was just lazy. I use GPU to demonstrate the idea because it is easy to implement the pipelining. In reality, we should do post-processing op on CPU, or unless only doing this on GPU if we know that there are some exposed communication in the beginning (which not for LLM).

> That said, what we can do is have a separate CUDA stream enable host to device data transfer as part of the data loading loop. I think that is what this PR is doing? This way we have the data for the next batch on the GPU before it asks for it. But there are caveats:

Totally agree, there are some tradeoffs so I didn't want to try it before. However, given that we are doing more post-processing, I start to wonder that should we utilize the spare CPU resource for post-processing and H2D movement?

PR #2244, opened 2026-01-16
[rl] Generator enables TP, using torchtitan as Trainer, add grader for reward calculation

wwwjn:
* __->__ #2244
* #2221
* #2194
* #2191

Current status:
- Applied same parallelism on trainer / generator, eg, Trainer (TP=2) and Generator (TP=2) 
    - NOTE: now we have a strong assumption that the trainer and generator must have the same parallelism, and are collocated. This is because we unwrap the DTensor and assum we can always wrap it back with the same device mesh and placement.
    - We should be able to remove this constrain once we have more powerful weight sync. cc @daniellepintz 
- Weight transfer: Simply unwrap DTensor in trainer before send
- Collocated trainer and generator 

Next Step:
- Patch batch-invariant kernel, test if the trainer / generator is batch-invariant
- Investigate collective communication kernel when
- CI as a guard

Toy GRPO training:
```
INFO:__main__:[actor=<root>] 
Step   8 | Loss: -0.0078 | Reward: +1.352
INFO:__main__:[actor=<root>]   Sample:  Paris, the capital of Italy is Rome, and the capital of Spain is Madrid. The ca...
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] os.getpid()=3397374 Generating start generate (policy v9)...
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Starting generation: 5 prompts, n_samples_per_prompt=8, max_tokens=100, temp=0.8
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] os.getpid()=3397877 Generating start generate (policy v9)...
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Starting generation: 5 prompts, n_samples_per_prompt=8, max_tokens=100, temp=0.8
INFO 02-10 17:06:33 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 18.6 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%
INFO 02-10 17:06:33 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 18.6 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%
INFO 02-10 17:06:44 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Engine 000: Avg prompt throughput: 26.9 tokens/s, Avg generation throughput: 77.0 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%
INFO 02-10 17:06:44 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Engine 000: Avg prompt throughput: 26.9 tokens/s, Avg generation throughput: 76.7 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.1%, Prefix cache hit rate: 0.0%
INFO 02-10 17:06:54 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 88.2 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.3%, Prefix cache hit rate: 0.0%
INFO 02-10 17:06:54 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 88.4 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.3%, Prefix cache hit rate: 0.0%
INFO 02-10 17:07:05 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 88.5 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.4%, Prefix cache hit rate: 0.0%
INFO 02-10 17:07:05 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 88.4 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.4%, Prefix cache hit rate: 0.0%
INFO 02-10 17:07:15 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 88.3 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.5%, Prefix cache hit rate: 0.0%
INFO 02-10 17:07:15 [loggers.py:259] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 88.3 tokens/s, Running: 40 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.5%, Prefix cache hit rate: 0.0%
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Generated 40 completions for scoring
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] os.getpid()=3397877 Generating finish generate (policy v9)...
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Generated 40 completions for scoring
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] os.getpid()=3397374 Generating finish generate (policy v9)...
INFO:torchtitan.experiments.rl.unified.actors.grader:[actor=<root>.<torchtitan.experiments.rl.unified.actors.grader.Grader grader{'gpus': 1/2}>] Grader scoring trajectory (policy v9)...
INFO:torchtitan.experiments.rl.unified.actors.grader:[actor=<root>.<torchtitan.experiments.rl.unified.actors.grader.Grader grader{'gpus': 0/2}>] Grader scoring trajectory (policy v9)...
INFO:torchtitan.experiments.rl.unified.actors.grader:[actor=<root>.<torchtitan.experiments.rl.unified.actors.grader.Grader grader{'gpus': 1/2}>] Grader finished scoring: reward_mean=1.3548, reward_std=0.7215
INFO:torchtitan.experiments.rl.unified.actors.grader:[actor=<root>.<torchtitan.experiments.rl.unified.actors.grader.Grader grader{'gpus': 0/2}>] Grader finished scoring: reward_mean=1.3548, reward_std=0.7215
INFO:torchtitan.experiments.rl.unified.actors.trainer:[actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 0/2}>] os.getpid()=3397374 Trainer starts to train 9 on traj:
INFO:torchtitan.experiments.rl.unified.actors.trainer:[actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 1/2}>] os.getpid()=3397877 Trainer starts to train 9 on traj:
INFO:torchtitan.experiments.rl.unified.actors.trainer:[actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 0/2}>] Logprob verification: bitwise_identical=False, max_delta=1.729112e-01, avg_delta=2.079452e-02, tokens_checked=4000
INFO:torchtitan.experiments.rl.unified.actors.trainer:[actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 1/2}>] Logprob verification: bitwise_identical=False, max_delta=1.729112e-01, avg_delta=2.079452e-02, tokens_checked=4000
INFO:torchtitan.experiments.rl.unified.actors.trainer:[actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 0/2}>] os.getpid()=3397374 Trainer finish step 10
INFO:torchtitan.experiments.rl.unified.actors.trainer:[actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 1/2}>] os.getpid()=3397877 Trainer finish step 10
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Updated weights into vLLM engine actor model. Number of parameters: 311
INFO:torchtitan.experiments.rl.unified.models.vllm_wrapper:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Weights already loaded during model initialization.             Returning 311 loaded parameter names to satisfy vLLM safety check.
INFO 02-10 17:07:41 [gpu_model_runner.py:4381] [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] Reloading and processing weights took 0.01 seconds
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/2}>] os.getpid()=3397374 Generator updating weights to policy v10...
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Updated weights into vLLM engine actor model. Number of parameters: 311
INFO:torchtitan.experiments.rl.unified.models.vllm_wrapper:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] Weights already loaded during model initialization.             Returning 311 loaded parameter names to satisfy vLLM safety check.
INFO:torchtitan.experiments.rl.unified.actors.generator:[actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 1/2}>] os.getpid()=3397877 Generator updating weights to policy v10...
INFO:__main__:[actor=<root>] 
Step   9 | Loss: -0.0082 | Reward: +1.355
INFO:__main__:[actor=<root>]   Sample:  Paris, the capital of Italy is Rome, and the capital of Spain is Madrid. The ca...
INFO:__main__:[actor=<root>] 
================================================================================
INFO:__main__:[actor=<root>] RL Training complete
INFO:__main__:[actor=<root>] ================================================================================
/home/jianiw/.conda/envs/pytorch-3.12/lib/python3.12/multiprocessing/resource_tracker.py:255: UserWarning: resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '

```


torchtitan/experiments/rl/unified/actors/scorer.py
```diff
@@ -0,0 +1,114 @@
...
+import torch
+from monarch.actor import Actor, endpoint
+from torchtitan.experiments.rl.unified.job_config import JobConfig
+from torchtitan.experiments.rl.vllm_compat.simple_rl import trivial_reward_function
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class TrajectoryData:
```
tianyu-l: I thought we deprecated the name trajectory which is intrinsically ambiguous, but I don't know what we replace it by, Episode?

torchtitan/experiments/rl/unified/actors/scorer.py
```diff
@@ -0,0 +1,114 @@
...
+    policy_version: int
+    completions: List[str]
+    vllm_token_ids: List[List[int]]
+    vllm_token_log_probs: List[List[float]]
+    prompt_token_ids: List[List[int]]
+    expected_answers: List[str]
+    rewards: torch.Tensor
+
+
+class Scorer(Actor):
```
tianyu-l: I thought you chose to use Grader. Not sure what's the difference but but aligned.

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -38,40 +44,125 @@ def __init__(
...
+            f"Trainer initialized with TorchTitan trainer, "
+            f"model={job_config.model.name}, flavor={job_config.model.flavor}, "
+            f"group_size={self.group_size}, grpo_beta={self.grpo_beta}, "
+            f"use_stable_grpo={self.use_stable_grpo}"
+        )
+
+    def _load_initial_weights(self, model: torch.nn.Module, model_path: str) -> None:
+        """Load initial weights from HuggingFace checkpoint."""
+        from torchtitan.experiments.rl.vllm_compat.weights.converter import (
+            vllm_to_torchtitan,
```
tianyu-l: why using this function instead of our utils like `from_hf`?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -78,9 +78,10 @@ def forward(
         batch_size, num_heads, seq_len, head_dim = q.shape
 
         # Transpose to (batch, seq_len, num_heads, head_dim) for vLLM
-        q = q.transpose(1, 2)
-        k = k.transpose(1, 2)
-        v = v.transpose(1, 2)
+        # vLLM attention expects bfloat16 /  inputs
```
tianyu-l: I think this can't just happen for attention.
For torchtitan, by default dtype is fp32, and mixed precision is handled by FSDP so under pure TP forward dtype is fp32.
If vllm by default use overall bf16, we should match. O/w this is another place where torchtitan-native vllm forward would be slow.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -8,8 +8,9 @@
 Multiprocess RL training loop using Monarch Actors.
 
 This demonstrates:
-1. Distributed actor architecture with Generator (vLLM) and Trainer (TorchTitan) components
+1. Distributed actor architecture with Generator (vLLM), Scorer, and Trainer (TorchTitan) components
 2. File based weight synchronization between trainer and generator
```
tianyu-l: is this still true?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -104,6 +106,13 @@ async def main():
         job_config,  # Pass full job_config
     )
 
+    # Spawn scorer on trainer mesh (can share resources with trainer)
```
tianyu-l: Would like to learn more on how Scorer/Grader work with trainer / generator.
Naively I would think they should be put on generator mesh, not trainer_mesh, although they may be the same and you are only using gpus=0 right now.

torchtitan/experiments/rl/unified/models/attention.py
wwwjn: Yes dtype difference could be a reason that we are 40% slow when TP is not enabled

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: https://github.com/meta-pytorch/monarch/blob/main/docs/source/examples/grpo_actor.py#L505

I follow the practice here, the scorer is spawned on trainer mesh. My intuition is the main bottleneck is generator (generator takes longer time), so we want to put more work (eg, calculate rewards + advantages) on trainer side instead. 

If we only think about algorithm , we can put it on trainer or generator. If we put it on generator, the generated "episode" will be scored episode. If we put it on trainer, the generator can just pass "unscored" episode to trainer

torchtitan/experiments/rl/unified/models/attention.py
tianyu-l: so I don't think we should do it ad hoc for just attention layer. I imagine we should init a bf16 model and load state dict in bf16 precision.

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,114 @@
...
+    policy_version: int
+    completions: List[str]
+    vllm_token_ids: List[List[int]]
+    vllm_token_log_probs: List[List[float]]
+    prompt_token_ids: List[List[int]]
+    expected_answers: List[str]
+    rewards: torch.Tensor
+
+
+class Grader(Actor):
```
tianyu-l, line 18: Usage of score / grade / reward is a bit arbitrary right now.
claude tells me:
> For general RL work: Stick with "reward function" or "reward model"
For RLHF pipelines: "Reward model" is standard
If choosing between scorer/grader: "Scorer" is slightly more aligned with ML conventions, as it emphasizes the quantitative nature of the output

When comparing scorer & grader, it sounds to me scorer is better than grader because the latter (more or less) suggests discrete reward value only.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -122,11 +131,14 @@ async def main():
...
-        # Fully sync RL loop
-        # NOTE: This is only getting Trajectory generated from trainer 0, and trainer 1's data is ignored.
-        # .get() is a monarch synchronize API which makes the loop fully sync
-        batch = generator.generate.call().get().item(gpus=0)
-        metrics = trainer.step.call(batch).get().item(gpus=0)
+        # Fully sync RL loop with separate scoring step
+        # 1. Generator produces episode (without rewards)
+        episode = generator.generate.call().get().item(gpus=0)
+        # 2. Grader computes rewards
+        episode = grader.score.call(episode).get().item(gpus=0)
```
tianyu-l:
```suggestion
        scored_episode = grader.score.call(episode).get().item(gpus=0)
```

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -122,11 +131,14 @@ async def main():
...
 
     for step in range(num_steps):
-        # Fully sync RL loop
-        # NOTE: This is only getting Trajectory generated from trainer 0, and trainer 1's data is ignored.
-        # .get() is a monarch synchronize API which makes the loop fully sync
-        batch = generator.generate.call().get().item(gpus=0)
-        metrics = trainer.step.call(batch).get().item(gpus=0)
+        # Fully sync RL loop with separate scoring step
+        # 1. Generator produces episode (without rewards)
+        episode = generator.generate.call().get().item(gpus=0)
```
tianyu-l: There should be difference between episode vs. episodes. The generate call seems returning an Episodes object?

torchtitan/experiments/rl/unified/actors/trainer.py
tianyu-l: also, this comment seems not addressed

torchtitan/experiments/rl/unified/actors/utils.py
felipemello1, line 1: This looks like a loss file. We should probably put it under /rl/loss. Here is the work i did in Forge, maybe we can just copy/paste? https://github.com/meta-pytorch/torchforge/tree/main/src/forge/rl/loss

If the number of losses look overwhelming, we can just keep GRPO/DAPO

torchtitan/experiments/rl/unified/actors/utils.py
```diff
@@ -0,0 +1,184 @@
...
+        "ratio_clipped_frac": (torch.abs(ratio - clipped_ratio) > 1e-6)
+        .float()
+        .mean()
+        .item(),
+    }
+
+    return total_loss, metrics, batch_token_log_probs
+
+
+def verify_logprob_identity(
```
felipemello1, line 150: if we log ratio (i.e. logprob train / logprog generator), we can check if this is always equal to 1. This is present in all losses in forge.

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,115 @@
...
+import torch
+from monarch.actor import Actor, endpoint
+from torchtitan.experiments.rl.unified.job_config import JobConfig
+from torchtitan.experiments.rl.vllm_compat.simple_rl import trivial_reward_function
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Episodes:
```
felipemello1: I think that this should be in some types.py file

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -98,12 +70,12 @@ def __init__(
             dtype=generation_config.dtype,
             # Parallelism configuration
             tensor_parallel_size=generation_config.parallelism.tensor_parallel_degree,
-            distributed_executor_backend=generation_config.distributed_executor_backend,
+            distributed_executor_backend="external_launcher",
```
acisseJZhong: maybe add comment on why we hardcode this?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -98,12 +70,12 @@ def __init__(
...
             # Memory and performance
             gpu_memory_utilization=generation_config.gpu_memory_utilization,
             enforce_eager=generation_config.enforce_eager,
-            # Seed
-            seed=generation_config.seed,
+            # Seed (use debug seed if set, otherwise default to 42)
+            seed=job_config.debug.seed if job_config.debug.seed is not None else 42,
```
acisseJZhong: instead of defaulting here, can we set job_config.debug.seed default to 42?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -335,9 +271,14 @@ def __init__(
         self.use_stable_grpo = job_config.policy_optimization.use_stable_grpo
 
         # Initialize distributed environment for SPMD generator
-        world_size = dist_utils.init_distributed(
-            Comm(),
-        )
+        # When running under Monarch, setup_env_for_distributed already
+        # initializes the process group, so skip re-initialization.
+        if torch.distributed.is_initialized():
```
acisseJZhong: it's already handled here https://github.com/pytorch/torchtitan/blob/7a6e45112a20e3547c91dd25a2df16df5469594e/torchtitan/distributed/utils.py#L297 so I think you can switch back to the original version

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -8,24 +8,20 @@
 import logging
 import os
 
-from dataclasses import dataclass
 from typing import List
 
 import torch
 from monarch.actor import Actor, endpoint
+from torch.distributed._tensor import DTensor
```
```suggestion
from torch.distributed.tensor import DTensor
```

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: yeah we need better documentation everywhere, e.g. to have justifications on why certain things are done -- you can use claude

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: why do we need a default when user doesn't want deterministic behavior? Also what if there are multiple generators, should they use different seeds?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -121,57 +93,21 @@ def _get_model(self):
 
     def update_weights(self, state_dict: dict) -> None:
```
tianyu-l: The naming sounds ... arbitrary. Sometimes you use load_weights, sometimes update_weights, can we canonicalize?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -121,57 +93,21 @@ def _get_model(self):
...
-            self.engine = LLMEngine.from_engine_args(engine_args)
-            logger.info("Created new vLLM LLMEngine")
-        else:
-            # Use collective_rpc to call reload_weights on all workers
-            # This reloads weights from temp_model_dir without recreating the engine
-            self.engine.collective_rpc("reload_weights")
+        assert self.engine is not None
+        # Use collective_rpc to call reload_weights on all workers
+        # This reloads weights from temp_model_dir without recreating the engine
+        self.engine.collective_rpc("reload_weights")
```
tianyu-l: then there is also reload_weights...

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -335,9 +271,14 @@ def __init__(
...
+        # When running under Monarch, setup_env_for_distributed already
+        # initializes the process group, so skip re-initialization.
+        if torch.distributed.is_initialized():
+            world_size = torch.distributed.get_world_size()
+        else:
+            world_size = dist_utils.init_distributed(
+                Comm(),
+            )
         # Initialize vLLM engine with job_config
         self.vllm_engine = VLLMGenerator(job_config, self.model_path)
```
tianyu-l: `Generator` owns `VLLMGenerator`...

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -426,15 +342,41 @@ async def generate(self) -> None:
             return trajectory
 
     @endpoint
-    async def update(self, version: int, vllm_compat_state: dict) -> None:
-        """Update generate weights.
+    async def update(self, version: int, all_weights: dict) -> None:
```
tianyu-l: now another `update(... dict)`

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -83,8 +83,8 @@ def forward(
         batch_size, num_heads, seq_len, head_dim = q.shape
         _, num_kv_heads, _, _ = k.shape
 
-        # vLLM expects (num_tokens, num_heads, head_dim) where num_tokens = batch * seq_len
-        # First transpose to (batch, seq_len, num_heads, head_dim)
+        # Transpose to (batch, seq_len, num_heads, head_dim) for vLLM
+        # vLLM attention expects bfloat16 /  inputs
```
tianyu-l:
```suggestion
        # vLLM attention expects bfloat16 inputs
```

torchtitan/models/llama3/train_configs/debug_model.toml
```diff
@@ -46,7 +46,7 @@ dataset = "c4_test"  # supported datasets: c4_test (2K), c4 (177M)
 data_parallel_replicate_degree = 1
 data_parallel_shard_degree = -1
 fsdp_reshard_after_forward = "default" # default / never / always
-tensor_parallel_degree = 1
+tensor_parallel_degree = 2
```
tianyu-l: revert

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -51,36 +69,123 @@ def __init__(
...
-        local_rank = int(os.environ["LOCAL_RANK"])
-        device = torch.device(f"cuda:{local_rank}")
-        torch.cuda.set_device(local_rank)
+        # Device setup
+        device_module, device_type = utils.device_module, utils.device_type
+        self.device = torch.device(f"{device_type}:{int(os.environ['LOCAL_RANK'])}")
+        device_module.set_device(self.device)
+
+        # Initialize distributed
+        # When running under Monarch, setup_env_for_distributed already
```
joecummings: This is a Monarch Actor - under what circumstances would you not be "running under Monarch"?

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,115 @@
...
+
+        logger.info(
+            f"Grader finished scoring: "
+            f"reward_mean={reward_mean.item():.4f}, reward_std={reward_std.item():.4f}"
+        )
+
+        return trajectory
+
+    @endpoint
+    async def set_reward_fn(self, reward_fn: Callable) -> None:
```
joecummings: Curious when this would happen that the Grader gets a new reward function?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -95,48 +200,103 @@ async def get_weights(self) -> dict:
...
         """
         logger.info(
             f"{os.getpid()=} Trainer starts to train {self.policy_version} on traj:"
         )
+
+        # Compute advantages from rewards
+        advantages = self._compute_advantages(episode.rewards)
+
+        # Compute reference log probs using frozen ref_model
+        ref_token_log_probs = []
```
joecummings, line 297: Ideally reference model should be able to be separate from the trainer due to both memory and performance constraints. Can do that in a follow up PR if you'd like.

torchtitan/experiments/rl/unified/actors/utils.py
joecummings: +1 to not having a catchall utils here. For now, even just putting all this in the trainer.py file would be fine IMO while we figure out the best structure for losses.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: The vllm's `EngineArgs` has a default value : https://docs.vllm.ai/en/stable/configuration/engine_args/#-seed 

Their explaination is about then TP is applied, which doesn't really makes sense to me... But even if we don't set it, vllm's behavior will be deterministic from the doc

torchtitan/experiments/rl/unified/actors/grader.py
wwwjn: I don't have strong opinion here, in our current task, the reward value is simple discrete value, and actual reward can be continuous or complicated

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: yes, the class is named "Episodes", but we are not doing batching here. So each instance of "Episodes" only contains one training prompt + completion + etc. I would rename it to be "Episode"

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: This update() updates Generator's internal state (Version number + model weights)

PR #2246, opened 2026-01-16
[Typing] Improve ModelProtocol typing

fegin:
* #2260
* __->__ #2246
* #2245


torchtitan/protocols/model.py
```diff
@@ -41,15 +40,18 @@ def get_nparams_and_flops(self, model: nn.Module, seq_len: int) -> tuple[int, in
         pass
 
 
-class ModelProtocol(Protocol):
+class ModelProtocol(nn.Module):
```
tianyu-l, line 43: Is it because multi-inheritance is not friendly for typing?

fegin: Well, it is more about we use ModelProtocol as the type through all the code, which pyrefly cannot understand the relation between ModelProtocol and nn.Module. So the original multi-inheritance that downstream models inheriting both from nn.Module and ModelProtocol doesn't work. We can do:

1. ModelProtocol inherits from both nn.Module and Protocol, but then I don't see the benefit of using Protocol + nn.Module compared to inheriting from only nn.Module.

2. Make the ModelProtocol define all the nn.Module methods that we may use. This seems to be overkill.

tianyu-l: I see. I don't have strong objection, but it indeed sounds a bit weird that a "protocol" inherits a heavy class.

> which pyrefly cannot understand the relation between ModelProtocol and nn.Module

This does sound incompatibility between pyrefly and multi-inheritance? Google tells me
> Python: Dynamic typing relies on isinstance() for runtime checks, but static type checkers (like Mypy) use complex algorithms (C3 MRO) to determine the correct method resolution order for type inference in multiple inheritance.

I wonder what's your take @rchen152

rchen152: Do you have an example of a case in which pyrefly doesn't understand the relationship between ModelProtocol and nn.Module? Without a concrete example, it's hard for me to tell which situation is happening:
1. Classes are explicitly multiply inheriting from ModelProtocol and nn.Module, but pyrefly does not understand the multiple inheritance (a bug in pyrefly).
2. Classes are only explicitly inheriting from either nn.Module or ModelProtocol, but they are conceptually linked in a way that isn't encoded in the types (either a limitation of the type system or a bug in torchtitan).

fegin: > This does sound incompatibility between pyrefly and multi-inheritance? Google tells me

No, I don't think so. If we have an object that inherits from both ModelProtocol and nn.Module, pyrefly can figure it out. But the current issue is that we upcast the object to ModelProtocol in our code (e.g., our model_builder returns ModelProtocol), which loses the operation information of nn.Module (e.g., `state_dict`). pyrefly has to report the errors because ModelProtocol doesn't define `state_dict`.

tianyu-l: I see. It sounds that torchtitan doesn't enforce hygiene when using the protocol.

The alternative solution is to create another class which inherits `ModelProtocol` and `nn.Module`, which seems unnecessary. I think it's fine to land this -- I plan to do some global refactor anyways.

tianyu-l (approved):

PR #2248, opened 2026-01-17
Add ROCm CI support for Auto Parallel & Compiler Toolkit experiments

akashveramd:
In this PR, we are adding ROCm CI support for Auto Parallel & Compiler Toolkit experiments. 
Also, enabling test_generate features test for ROCm.

### DIFF
--- a/tests/integration_tests/run_tests.py
+++ b/tests/integration_tests/run_tests.py
@@ -107,9 +107,7 @@ def run_tests(args, test_list: list[OverrideDefinitions]):
             ran_any_test = True
 
     if not ran_any_test:
-        available_tests = [
-            t.test_name for t in test_list if not t.disabled if not t.skip_rocm_test
-        ]
+        available_tests = [t.test_name for t in test_list if not t.disabled]
         if hasattr(args, "test_suite"):
             logger.warning(
                 f"No tests were run for --test_name '{args.test_name}' in test suite '{args.test_suite}'.\n"


tests/integration_tests/run_tests.py
```diff
@@ -107,11 +107,19 @@ def run_tests(args, test_list: list[OverrideDefinitions]):
             ran_any_test = True
 
     if not ran_any_test:
-        available_tests = [t.test_name for t in test_list if not t.disabled]
-        logger.warning(
-            f"No tests were run for --test_name '{args.test_name}' in test suite '{args.test_suite}'.\n"
-            f"Available test names in '{args.test_suite}' suite: {available_tests}"
-        )
+        available_tests = [
+            t.test_name for t in test_list if not t.disabled if not t.skip_rocm_test
```
tianyu-l: This code seems general and not tied to rocm. Why do we tie the logic to rocm?

tests/integration_tests/run_tests.py
```diff
@@ -107,11 +107,19 @@ def run_tests(args, test_list: list[OverrideDefinitions]):
...
+        ]
+        if hasattr(args, "test_suite"):
```
tianyu-l, line 111: why having this if-else?

akashveramd: It's a safeguard for the tests that are not part of test_suite, like experiments. If there is only one test, and that too getting skipped (example autoparallel test has one test and getting skipped for ROCm) then ran_any_test will be False. Hence, in log, it tries to access args.test_suite which is empty and gives error. To avoid the error added this safeguard.

tests/integration_tests/run_tests.py
akashveramd: Rolled back.

tianyu-l:
compiler toolkit CI is failing in main branch. Could you take a look? @yiming0416 https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_compiler_toolkit.yaml?query=branch%3Amain

tianyu-l (approved):

yiming0416:
> compiler toolkit CI is failing in main branch. Could you take a look? @yiming0416 https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_compiler_toolkit.yaml?query=branch%3Amain

@tianyu-l 
Seems like a regression in dynamo. I was able to repro for simple-fsdp experiments as well with compile turned on.
Opened an issue here https://github.com/pytorch/pytorch/issues/174128

PR #2249, opened 2026-01-17
[mxfp8 moe training] support wgrad_with_hp recipe

danielvegamyhre:
Stacked PRs:
 * #2268
 * #2251
 * #2250
 * __->__#2249

--- --- ---

## Summary

torchao MXFP8 MoE training code now supports a new recipe: `wgrad_with_hp` (described below). This PR update the torchtitan integration to allow users to use. it

  Recipes:
      - "mxfp8": Use MXFP8 for all 3 grouped GEMMs in the forward and backward pass (output, dgrad, wgrad).
      - "mxfp8_wgrad_with_hp": Use MXFP8 for forward output and dgrad, but keep wgrad in high-precision.
           This can be used to trade-off some performance for improved accuracy.

Note: I plan to do some benchmarking to provide more concrete guidance to users on what expert shapes will result in better TPS using wgrad_with_hp

## Tests

mxfp8 recipe:

```
CONFIG_FILE=/home/dev/torchtitan/torchtitan/models/llama4/train_configs/llama4_17bx16e.toml ./run_train.sh --metrics.log_freq=10 \--training.steps=200  \--parallelism.data_parallel_shard_degree=8 \--parallelism.expert_parallel_degree=8 \--parallelism.tensor_parallel_degree=1 \                                                  
--parallelism.expert_tensor_parallel_degree=1 \
--profiling.enable_profiling --profiling.profile_freq=30 \
--training.seq_len=8192 \
--activation_checkpoint.mode=none \
--model.print_after_conversion \
--training.local_batch_size=12 \
--model.converters="quantize.grouped_mm.mx,quantize.linear.mx" \
--quantize.linear.mx.mxfp8_dim1_cast_kernel_choice="cuda" \
--quantize.linear.mx.filter_fqns="output,router.gate,wk,wv" \
--quantize.grouped_mm.mx.fqns="experts" --quantize.grouped_mm.mx.recipe_name="mxfp8_wgrad" \
--compile.enable --debug.moe_force_load_balance
```

mxfp8_wgrad_with_hp recipe:

```
CONFIG_FILE=/home/dev/torchtitan/torchtitan/models/llama4/train_configs/llama4_17bx16e.toml ./run_train.sh --metrics.log_freq=10 \--training.steps=200  \--parallelism.data_parallel_shard_degree=8 \--parallelism.expert_parallel_degree=8 \--parallelism.tensor_parallel_degree=1 \                                                  
--parallelism.expert_tensor_parallel_degree=1 \
[...]

Issue #1940, opened 2025-10-27
B200 Llama4 Compile dp2ep=4, pp=2 (interleave 1f1b), full AC fails when compiling token dispatch


xmfan:
### Bug description

Reported by @danielvegamyhre 

Repro:
- only on B200
- https://github.com/pytorch/torchtitan/pull/1895 but remove the `patch_experts = True` branch to compile token dispatch
- change to 2 layers in llama4/init.py

```
LOG_RANK=4 NGPU=8 CONFIG_FILE="/home/${USER}/core/torchtitan/torchtitan/models/llama4/train_configs/llama4_17bx16e.toml" wp tlp ./run_train.sh --metrics.log_freq=10 --training.steps=100  --parallelism.data_parallel_shard_degree=4 --parallelism.expert_parallel_degree=4 --parallelism.tensor_parallel_degree=1 --parallelism.expert_tensor_parallel_degree=1 --training.seq_len=8192 --training.local_batch_size=2 --model.print_after_conversion --training.debug_moe_force_load_balance --activation_checkpoint.mode="full" --parallelism.pipeline_parallel_degree 2 --parallelism.pipeline_parallel_schedule "Interleaved1F1B" --parallelism.pipeline_parallel_layers_per_stage 1
```

Error:
```python
    File "/home/danvm/.conda/envs/torch/lib/python3.13/site-packages/torch/_inductor/utils.py", line 3018, in run
      out = model(new_inputs)
    File "/tmp/torchinductor_danvm/uo/cuonnmrcfp4ekv4hymjuo2ycyeeujs7sjeadgtvus5z5fpm22to7.py", line 494, in call
      buf25 = torch.ops._c10d_functional.all_to_all_single.default(arg1_1, [u4, u5, u6, u7], [u0, u1, u2, u3], '22')
    File "/home/danvm/.conda/envs/torch/lib/python3.13/site-packages/torch/_ops.py", line 841, in __call__
      return self._op(*args, **kwargs)
             ~~~~~~~~^^^^^^^^^^^^^^^^^
  RuntimeError: Split sizes doesn't match total dim 0 size
```

Workaround is to not compile token dispatch. e.g. directly using https://github.com/pytorch/torchtitan/pull/1895

### Versions

torch: nightly
torchtitan: https://github.com/pytorch/torchtitan/pull/1895

laithsakka:
cc @bobrenjc93

danielvegamyhre:
[...]

PR #2250, opened 2026-01-17
[mxfp8 moe training] mxfp8 all to all

danielvegamyhre:
Stacked PRs:
 * #2268
 * #2251
 * __->__#2250
 * #2249

--- --- ---

[mxfp8 moe training] mxfp8 all to all

## High level design

- The goal is to reduce all-to-all communication volume in expert parallelism, to achieve a speedup.
- Key idea: instead of waiting until directly before the grouped GEMM to quantize the inputs, we quantize a couple steps earlier, before the all2all, and stay in MXFP8 through the token shuffle, so the MXFP8 grouped GEMM receives pre-quantized input activations.
 
Screenshot 2026-02-04 at 11 00 34 AM

Forward flow:
                1. a2a dispatch forward: high-precision input, MXFP8 output
                2. permute forward: MXFP8 input, MXFP8 output
                3. MXFP8 grouped GEMM: MXFP8 input, high-precision output
                4. unpermute forward: high-precision input and output
                5. a2a combine forward: high-precision input and output

Backward flow:
                1. a2a combine backward: high-precision input, MXFP8 output
                2. unpermute backward: MXFP8 input, MXFP8 output
                3. MXFP8 grouped GEMM: MXFP8 input, high-precision output
                4. permute backward: high-precision input and output
                5. a2a dispatch backward: high-precision input and output

## Benchmarks
- +10% TPS on single node 8xB200 training with EP over NVLink domain
- +24% TPS on multi-node B200 cluster with IB inter-node networking (more bandwidth constrained -> bigger speedup)

## Limitations
- Not compatible with DeepEP
    - Recipe `"mxfp8"` IS compatible with DeepEP, but `"mxfp8_wgrad_with_hp"` (which uses MXFP8 all2all, is not)

## Testing
 
(test depends on the full PR stack)

```bash
CONFIG_FILE=/home/dev/torchtitan/torchtitan/models/deepseek_v3/train_configs/deepseek_v3_16b.toml ./run_train.sh --metrics.log_freq=10 \
--training.steps=1500  \
--parallelism.expert_parallel_degree=4 \
--parallelism.expert_tensor_parallel_degree=1 \
--training.seq_len=8192 \
--activation_checkpoint.mode=full \
--model.print_after_conversion \
--training.local_batch_size=16 \
--quantize.linear.mx.mxfp8_dim0_cast_kernel_choice="triton" --quantize.linear.mx.mxfp8_dim1_cast_kernel_choice="cuda" \
--quantize.grouped_mm.mx.fqns="experts" --quantize.grouped_mm.mx.recipe_name="mxfp8_wgrad_with_hp" \
--compile.enable --compile.components="model,loss" --debug.moe_force_load_balance \
--model.converters="quantize.grouped_mm.mx"
```

### DIFF
--- a/torchtitan/config/job_config.py
+++ b/torchtitan/config/job_config.py
@@ -793,6 +793,36 @@ class MXGroupedMM:
         - "mxfp8_wgrad_with_hp": Use MXFP8 for forward output and dgrad, but keep wgrad in high-precision.
              This can be used to trade-off some performance for improved accuracy. For some smaller expert shapes,
              it is also better for performance.
+
+             This recipe also reduces communication volume by doing certain all-to-all operations in MXFP8.
+             Specifically, instead of doing MXFP8 quantization right before the MXFP8 grouped GEMMs for the
+             routed experts, we do the quantization earlier, before the preceding all-to-all dispatch, and
+             stay in MXFP8 through the token shuffle.
+
+             This speeds up the all-to-all by sending fewer bytes over the network. Similarly, in the backward
+             pass, we quantize before the all-to-all combine, stay in MXFP8 through token unpermutation, and then
+             do the MXFP8 grouped GEMM. The diagram below visualzes this:
+
+               Forward flow:
+                1. a2a dispatch forward: high-precision input, MXFP8 output
+                2. permute forward: MXFP8 input, MXFP8 output
+                3. MXFP8 grouped GEMM: MXFP8 input, high-precision output
+                4. unpermute forward: high-precision input and output
+                5. a2a combine forward: high-precision input and output
+
+               Backward flow:
+                1. a2a combine backward: high-precision input, MXFP8 output
+                2. unpermute backward: MXFP8 input, MXFP8 output
+                3. MXFP8 grouped GEMM: MXFP8 input, high-precision output
+                4. permute backward: high-precision input and output
+                5. a2a dispatch backward: high-precision input and output
+
+               Limitations:
+                - Not compatible with DeepEP
+
+               MXFP8 requires installation of torchao nightly build for CUDA 12.8+: https://github.com/pytorch/ao
+
+
     Example: --quantize.grouped_mm.mx.recipe_name="mxfp8"
     """
 
--- a/torchtitan/distributed/expert_parallel.py
+++ b/torchtitan/distributed/expert_parallel.py
@@ -107,32 +107,13 @@ class ExpertParallel(BaseExpertParallel):
         ep_degree = device_mesh.shape[0]
         num_local_experts = num_tokens_per_expert.shape[0] // ep_degree
 
-        # generate the input splits and output splits for all-to-all
-        with torch.no_grad():
-            num_tokens_per_expert_group = all_to_all_single(
-                num_tokens_per_expert,
-                None,
-                None,
-                group=device_mesh.get_group(),
-            )
-            # Need to wait explicitly because it is used by a triton kernel later
-            # which doesn't realize that AsyncCollectiveTensor needs unwrapping
-            num_tokens_per_expert_group = torch.ops._c10d_functional.wait_tensor(
-                num_tokens_per_expert_group
-            )
-            input_splits = (
-                num_tokens_per_expert.view(ep_degree, -1)
-                .sum(dim=1)
-                .to(torch.device("cpu"), non_blocking=True)
-            )
-            # NOTE: this would incur a device-to-host sync
-            output_splits = (
-                num_tokens_per_expert_group.view(ep_degree, -1)
-                .sum(dim=1)
-                .to(torch.device("cpu"), non_blocking=False)
-            )
-            self.input_splits = input_splits.tolist()
-            self.output_splits = output_splits.tolist()
+        # first all-to-all to calculate output splits from input splits.
+        # note: this will incur a d2h sync
+        (
+            self.input_splits,
+            self.output_splits,
+            num_tokens_per_expert_group,
+        ) = get_a2a_splits(num_tokens_per_expert, device_mesh, ep_degree)
 
         # perform all-to-all
         routed_input = all_to_all_single_autograd(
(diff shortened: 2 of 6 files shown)


danielvegamyhre:
@tianyu-l I will polish this up but any earlier thoughts on the high level design?

torchtitan/models/moe/utils.py
```diff
@@ -95,3 +97,49 @@ def wrapper(
         return out
 
     return wrapper
+
+
+def get_a2a_splits(
```
tianyu-l: Given the current code organization, this function is a distributed util, so probably should be in expert_parallel.py.

torchtitan/config/job_config.py
```diff
@@ -459,6 +459,54 @@ class Parallelism:
     https://github.com/deepseek-ai/DeepEP.
     """
 
+    expert_parallel_a2a_dispatch_fwd_precision: Literal["default", "mxfp8"] = "default"
```
tianyu-l: Is it not a net benefit to enable mxfp8 dispatch in fwd and mxfp8 combine in backward? If they are, they don't need to be configurable -- we can just enable them whenever EP and mxfp8 grouped mm is used.

danielvegamyhre: It is but there is an important limitation, which is that mxfp8 grouped mm recipe must be `mxfp8_wgrad_with_hp` (computes forward output and dgrad in mxfp8, computes wgrad in high precision). It is not compatible with `mxfp8` recipe yet (uses mxfp8 for all output/dgrad/wgrad). 

We could automatically enable it when the recipe is `mxfp8_wgrad_with_hp` perhaps?

Context on this requirement if you are curious:

This is because if inputs come in pre-quantized along dim0, in the backward pass `wgrad = grad_out_t @ input` where we need inputs quantized along dim1, we would need to dequant along dim0 then requant along dim1. This is definitely an option to do, especially if we have a fast fused kernel for this, however, there is some debate about the numerical implications of doing this (as it is not equivalent to doing dim1 quant on the original bf16 tensor). 

Therefore, for now we assert the recipe if `mxfp8_wgrad_with_recipe` pending (1) the fast dequant/requant kernel, and (2) numerics experiments.

tianyu-l: > It is but there is an important limitation, which is that mxfp8 grouped mm recipe must be mxfp8_wgrad_with_hp

Then we should couple it with `mxfp8_wgrad_with_hp`, unless you see reasons not to?

> This is because if inputs come in pre-quantized along dim0, in the backward pass wgrad = grad_out_t @ input where we need inputs quantized along dim1, we would need to dequant along dim0 then requant along dim1.

I get that wgrad_with_hp makes sense, but I didn't get what's its relationship to doing fwd combine / bwd dispatch of dgrad in high-precision.

danielvegamyhre: >Then we should couple it with mxfp8_wgrad_with_hp, unless you see reasons not to?

I agree this sounds good, the only blocker on that right now is that torch.compile is working in unit tests but not in torchtitan e2e training with these mxfp8 EP building blocks. cc @bdhirsh the nonstrict_trace dynamo decorator works in the torchao integration test w/ compile, but in torchtitan the same "tensor metadata mismatch" rears its head again. Will DM you about this.

bdhirsh: I made a diagram to describe why things are breaking:
image

torch.compile has a hard requirement today, where when you compile a given region of code, the forward outputs and their corresponding gradients (the backward inputs to that compiled region) must have the *same* subclass type.

You can see from this image that in the `mxfp8` codepath, both the `GroupedExperts.forward`, **and** the `MXFP8ExpertParallel._token_dispatch` violate this assumption.

That doesn't mean that we can't support this case. But it *does* mean that we are unable to compile these two blocks of code in isolation. Instead, our options are either:

(1) don't compile the token_dispatch or the `GroupedExperts.forward` at all
(2) compile a single graph, containing all of `_token_dispatch` + `GroupedExperts.forward` + `_token_combine`

I have a short diff for both 1 and 2, although I think there are a few things worth thinking more about.

diff for not compiling `GroupedExperts.forward`: https://gist.github.com/bdhirsh/970a671b84c35cc95a76f33657ca4d69

diff for compiling all 3 into a single graph: https://gist.github.com/bdhirsh/8b0ced2f6381b52eebd56fdff8e62093 This was a bit strange - it looks like when we try to compile `GroupedExperts`, dynamo ends up creating a separate graph for the `GroupedExperts.forward` and the module's forward pre-hooks / post-hooks. I "fixed" it by moving the `GroupedExperts` into an nn module wrapper and compiling that, but I imagine this isn't actually landable since it will e.g. muck up the state dict. I also saw some previous comments from @xmfan indicating that there were previous problems trying to compile `_token_dispatch/combine`: https://github.com/pytorch/torchtitan/issues/1940. I didn't understand the issue though.

bdhirsh: To me, the best option would be if we can tweak FSDP to handle multiple device meshes, so we can avoid having to FSDP-wrap the `experts` region (cc @weifengpy). Since I think that would solve this particular issue, and would also avoid a lot of the other issues that we've seen in titan around compiling MoE.

I'm not sure what the timeline for that is, so depending on @danielvegamyhre 's timeline, we might want to either just skip compiling the grouped_mm to unblock, or figure out how to land (2) above.

xmfan: Thanks for the clear diagram! Are there implementation changes we can do to make MXTensor <-> bf16 conversions to be moved into a different graph? Otherwise, I think the FSDP2 changes are necessary to compile a single graph with dispatch/experts/combine

bdhirsh: @danielvegamyhre can speak to that better than me, but I would imagine... probably not, unless you are willing to completely abandon the structure today of "_token_dispatch/combine" are module hooks that are logically separated from the grouped_mm forward call".

> Otherwise, I think the FSDP2 changes are necessary to compile a single graph with dispatch/experts/combine

I agree that this feels like an overall much cleaner state. One dumb question I have: I was able to get token_dispatch + grouped_mm + token_combine into a single graph with this change: https://gist.github.com/bdhirsh/8b0ced2f6381b52eebd56fdff8e62093. I agree it is ugly (and potentially not landable because it changes the state dict), but it seemed to work for me on Daniel's example. It sounds like there are other issues you ran into trying to compile the token_dispatch/combine - do you know what they are and what caused them?

xmfan: It seemed to be something with the unbacked shapes, I'm not sure if the error is still present. I believe the main consideration for module hook was to provide the same clean model definition for both ep and non-ep, so changing APIs seems okay as long as that we can ensure that

danielvegamyhre: > It is but there is an important limitation, which is that mxfp8 grouped mm recipe must be mxfp8_wgrad_with_hp

Then we should couple it with mxfp8_wgrad_with_hp, unless you see reasons not to?

@tianyu-l I went ahead made this change. One thing to note though is that this makes the `mxfp8_wgrad_with_hp` recipe incompatible with DeepEP. The standard `mxfp8` recipe (mxfp8 for all grouped gemms) is still compatible with DeepEP though, and we got great results with this combination recently (+41% TPS on DSV3 671B)

torchtitan/models/deepseek_v3/infra/parallelize.py
```diff
@@ -120,6 +120,10 @@ def parallelize_deepseekv3(
     else:
         use_deepep = False
 
+    use_mxfp8_a2a = (
```
tianyu-l, line 123: Since we have DeepEP backend now enabled with this config https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/job_config.py#L474
I think we should just add another backend mxfp8, and pass it to `apply_moe_ep_tp`, instead of using both `use_deepep` and `use_mxfp8_a2a`.

torchtitan/distributed/expert_parallel.py
```diff
@@ -382,3 +468,49 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
             input_fn=self._token_dispatch,  # pyrefly: ignore [bad-argument-type]
             output_fn=self._token_combine,  # pyrefly: ignore [bad-argument-type]
         )
+
+
+def get_a2a_splits(
```
tianyu-l, line 473: I now think it's better to make `_get_a2a_splits` a static method of ExpertParallel, so that subclasses of it can inherit.

torchtitan/distributed/expert_parallel.py
```diff
@@ -192,6 +173,111 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
         )
 
 
+class MXFP8ExpertParallel(ExpertParallel):
```
tianyu-l, line 176: Since it's related to mxfp8 only, let's put this in quantization/mx.py?

weifengpy:
per-layer FSDP2 wrapping is landed. no more fsdp2 preventng tracing at moe boundary

next gap for per-layer compile is `mark_dynamic` at `_run_experts_grouped_mm_dynamic`: https://github.com/pytorch/torchtitan/pull/2741

danielvegamyhre:
> per-layer FSDP2 wrapping is landed. no more fsdp2 preventng tracing at moe boundary
> 
> next gap for per-layer compile is `mark_dynamic` at `_run_experts_grouped_mm_dynamic`: #2741

amazing, thanks @weifengpy! will revisit this PR soon

weifengpy:
> > next gap for per-layer compile is `mark_dynamic` at `_run_experts_grouped_mm_dynamic`: #2741
> amazing, thanks @weifengpy! will revisit this PR soon

@danielvegamyhre per-layer compile with moe works with this PR (no more mark_dynamic graph break): https://github.com/pytorch/torchtitan/pull/2741

I am waiting for another pytorch-side fix before landing it

weifengpy:
> > per-layer FSDP2 wrapping is landed. no more fsdp2 preventng tracing at moe boundary
> > next gap for per-layer compile is `mark_dynamic` at `_run_experts_grouped_mm_dynamic`: #2741
> 
> amazing, thanks @weifengpy! will revisit this PR soon

Hi @danielvegamyhre , per-layer compile is landed https://github.com/pytorch/torchtitan/pull/2741

danielvegamyhre:
@weifengpy awesome, thanks!

Issue #2252, opened 2026-01-19
Weight tying not working


francesco-bertolotti:
### Bug description

It appears that weight tying is not functioning correctly for the Qwen3 model. The relevant code is here:

[https://github.com/pytorch/torchtitan/blob/09c6d74bf0535a283c5a2cf58dee5d5f701f39a0/torchtitan/models/qwen3/infra/parallelize.py#L176-L180](https://github.com/pytorch/torchtitan/blob/09c6d74bf0535a283c5a2cf58dee5d5f701f39a0/torchtitan/models/qwen3/infra/parallelize.py#L176-L180)

The weight tying logic would work as expected if the model were already materialized on a device. However, at this point the model is still on the `meta` device. When materialization later occurs, both `model.output` and `model.tok_embeddings` are independently materialized, which breaks the intended weight tying.

I verified this by printing `self.output.weight` and `self.tok_embeddings.weight` during the forward pass and confirmed that the two tensors are different:

[https://github.com/pytorch/torchtitan/blob/09c6d74bf0535a283c5a2cf58dee5d5f701f39a0/torchtitan/models/qwen3/model/model.py#L583-L592](https://github.com/pytorch/torchtitan/blob/09c6d74bf0535a283c5a2cf58dee5d5f701f39a0/torchtitan/models/qwen3/model/model.py#L583-L592)

As an experiment, I moved the application of parallelization to occur *after* model materialization. With this change, the expected behavior is restored, and `self.output.weight` is equal to `self.tok_embeddings.weight`:

[https://github.com/pytorch/torchtitan/blob/09c6d74bf0535a283c5a2cf58dee5d5f701f39a0/torchtitan/train.py#L256-L261](https://github.com/pytorch/torchtitan/blob/09c6d74bf0535a283c5a2cf58dee5d5f701f39a0/torchtitan/train.py#L256-L261)

### Versions

This issue can be reproduced on the latest pytorch.

toml
```toml
[job]
dump_folder = "./outputs"
description = "Qwen 3 0.6B training"

[profiling]
enable_profiling = false
save_traces_folder = "profile_trace"
profile_freq = 100

[metrics]
log_freq = 1
enable_tensorboard = true
save_tb_folder = "tb"

[model]
name = "qwen3"
flavor = "0.6B"
# hf_assets_path = "./assets/hf/Qwen3-0.6B"
# converters = ["float8"]

[optimizer]
name = "AdamW"
lr = 3e-4
eps = 1e-8

[...]

PR #2253, opened 2026-01-19
weight tying fix for qwen3

francesco-bertolotti:
PR for https://github.com/pytorch/torchtitan/issues/2252

The PR addresses the weight tying issue by using a lambda function. 
Whenever `enable_weight_tying` is enabled instead of using a `nn.Linear` layer, we can use the functional interface `F.linear` using `tok_embedding` as weights. 

I have run a small model to verify the implementation. 

`pre-varlen-note-qwen3-600M` refers to the model pre-PR without weight tying
`pre-varlen-te-qwen3-600M` refers to the model pre-PR with weight tying
`varlen-note-qwen3-600M` refers to the model post-PR without weight tying
`varlen-te-qwen3-600M` refers to the model post-PR with weight tying

Screenshot 2026-01-19 at 11 51 15

Let me know if something like this could work. If so I could check also the llama architecture for the same problem.

### DIFF
--- a/torchtitan/models/qwen3/model/model.py
+++ b/torchtitan/models/qwen3/model/model.py
@@ -498,7 +498,7 @@ class Qwen3Model(nn.Module, ModelProtocol):
         cutoff_factor = 3
 
         if self.enable_weight_tying:
-            # since when the model is intialized on meta device,
+            # since when the model is initialized on meta device,
             # the tying in the __init__ may not have worked correctly
             # we ensure the weights are tied here
             assert self.tok_embeddings is not None and self.output is not None


torchtitan/models/qwen3/model/model.py
```diff
@@ -451,6 +451,7 @@ def __init__(self, model_args: Qwen3ModelArgs):
         self.n_layers = model_args.n_layers
         self.eos_id = model_args.eos_id
```
tianyu-l: this is no longer used, please help remove

torchtitan/models/qwen3/model/model.py
```diff
@@ -463,7 +464,14 @@ def __init__(self, model_args: Qwen3ModelArgs):
             self.layers[str(layer_id)] = TransformerBlock(layer_id, model_args)
         self.norm = nn.RMSNorm(model_args.dim, eps=model_args.norm_eps)
 
-        self.output = nn.Linear(model_args.dim, model_args.vocab_size, bias=False)
+        if self.enable_weight_tying:
+            self.output = lambda x: F.linear(x, weight=self.tok_embeddings.weight)
```
tianyu-l: I'm not sure if it plays well with TP, as in https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/qwen3/infra/parallelize.py#L206

If not, I'd prefer we try https://github.com/pytorch/torchtitan/issues/2252#issuecomment-3767635636

torchtitan/models/qwen3/model/model.py
```diff
@@ -587,6 +594,5 @@ def forward(
 
         # pyrefly: ignore[not-callable, invalid-argument]
         h = self.norm(h) if self.norm else h
-        # pyrefly: ignore[not-callable, invalid-argument]
-        output = self.output(h) if self.output else h
+        output = self.output(h)
```
tianyu-l: This would break PP-compatibility so please revert the change, although PP is not enabled for Qwen3 today.

torchtitan/models/qwen3/model/model.py
francesco-bertolotti: Yep you are correct, it does not work even with fsdp. 

I tried 
```python
class TiedOutputHead(torch.nn.Module):
    def __init__(self, tok_embeddings):
        super().__init__()
        self.tok_embeddings = tok_embeddings

    def forward(self, x):
        return F.linear(x, self.tok_embeddings.weight)
```

and seems to work well. 

Regardless, your proposal seems better. I am going with your suggestion

torchtitan/models/qwen3/model/model.py
francesco-bertolotti: I see, good point.

francesco-bertolotti:
@tianyu-l , your suggestion seems to work well. Here's the new plots
Screenshot 2026-01-19 at 14 28 29

torchtitan/models/qwen3/model/model.py
```diff
@@ -494,6 +494,10 @@ def init_weights(
         final_out_std = self.model_args.dim**-0.5
         cutoff_factor = 3
 
+        if self.enable_weight_tying:
+            assert self.tok_embeddings is not None and self.output is not None
+            self.output.weight = self.tok_embeddings.weight
+
         # If weight tying is enabled, we don't need to initialize the output layer
```
tianyu-l: given the comment, should we put the code block into else branch of `if self.enable_weight_tying:`?

francesco-bertolotti:
Hi @tianyu-l,

I’m not very familiar with pyrefly, but I noticed that the linting action failed. From what I can tell, the failure doesn’t appear to be related to the file I modified.

Please let me know if there’s any action required on my side.

Thanks!

torchtitan/models/qwen3/model/model.py
francesco-bertolotti: yep, seems correct to me

torchtitan/models/qwen3/model/model.py
```diff
@@ -494,15 +494,19 @@ def init_weights(
...
-            nn.init.trunc_normal_(
-                self.output.weight,
-                mean=0.0,
-                std=final_out_std,
-                a=-cutoff_factor * final_out_std,
-                b=cutoff_factor * final_out_std,
-            )
+        if self.enable_weight_tying:
+            assert self.tok_embeddings is not None and self.output is not None
+            self.output.weight = self.tok_embeddings.weight
```
tianyu-l, line 505: Please also do this in `__init__` so that if people load weights without calling `init_weights`, weight-tying is still effective.

francesco-bertolotti: I’ll try that; however, it’s possible we’ll run into the original issue. If the model is initialized on the meta device, then self.output.weight = self.tok_embeddings.weight may not take effect once the final model is materialized.

Either way, I’ll give it a try tomorrow.

I can also do something like https://github.com/pytorch/torchtitan/pull/2253#discussion_r2704693272 if this does not work?

tianyu-l: I'm saying we should do them in BOTH places. As long as we don't delete it here, I don't think it'll cause regression?

francesco-bertolotti: My bad, I see what you mean.

francesco-bertolotti:
I’ve also added the weight tying in `__init__`. I think the only case where tying might fail is when the model is initialized on the meta device and `init_weights` is not called. However, that would already be problematic on its own, since the model still needs to be initialized after materialization on the target device.

tianyu-l:
please fix linting

torchtitan/models/qwen3/model/model.py
```diff
@@ -494,15 +497,22 @@ def init_weights(
...
-        if self.output is not None:
-            nn.init.trunc_normal_(
-                self.output.weight,
-                mean=0.0,
-                std=final_out_std,
-                a=-cutoff_factor * final_out_std,
-                b=cutoff_factor * final_out_std,
-            )
+        if self.enable_weight_tying:
+            # since when the model is intialized on meta device,
```
tianyu-l:
```suggestion
            # since when the model is initialized on meta device,
```

Issue #2225, opened 2026-01-13
Expert Parallel Fast Path and Router Dtype Restoration (from solar-open 102B)


rakkit:
Hi, I was reading [Solar Open-102B ](https://www.arxiv.org/abs/2601.07022), who use torchtitan on B200 to train a12w100B MoE recently.   They mentioned there were two tricks to improve MoE training 

<img width="2652" height="844" alt="Image" src="https://github.com/user-attachments/assets/e197f721-e108-4a05-a933-38442e3d9720" />

**Expert Parallel Fast Path** Basically removing token padding for non-EP path.

**Router Dtype Restoration** (did not really get this point, so I quote the original text)
>We identify a subtle implementation issue where the router’s sigmoid operation correctly casts to FP32 for numerical stability, but does not restore the original
dtype afterward, causing subsequent matrix operations to execute in FP32.

I wonder what's your thoughts are on this, and if we wanna adapt these them.

tianyu-l:
Thank you!

> Expert Parallel Fast Path Basically removing token padding for non-EP path.

bf16 grouped mm kernel used to require padding. It no longer requires so, and we can remove padding and shuffling, with or without EP.
However, fp8 / mxfp8 quantization still require padding, so there's some refactoring to do. It'd be great if you could help.
(cc @shuhuayu)

> We identify a subtle implementation issue where the router’s sigmoid operation correctly casts to FP32 for numerical stability, but does not restore the original dtype afterward, causing subsequent matrix operations to execute in FP32.

The score is in fp32, but grouped mm should be in bf16. The activations is cast to fp32 to multiply with score, then cast back to bf16.
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/moe/moe.py#L513

If that's not the case, we should fix it.

rakkit:
thx @tianyu-l, sure I will open a PR to fix.

> bf16 grouped mm kernel used to require padding. It no longer requires so, and we can remove padding and shuffling, with or without EP.

Can you remember from which PyTorch version we no longer need padding?   

I think the easiest fix is to detect the PyTorch version that sets `TOKEN_GROUP_ALIGN_SIZE_M` to 8 (old PyTorch version for Backward compatibility) or set to 1 if the user have new PyTorch. 

and in `indices_padding_wrapper` we skip permute if `TOKEN_GROUP_ALIGN_SIZE_M` is 1

tianyu-l:
@rakkit 
[...]

PR #2255, opened 2026-01-20
Remove unnecessary token padding for MoE in BF16 mode

rakkit:
fix of https://github.com/pytorch/torchtitan/issues/2225

Context: [Solar Open-102B ](https://www.arxiv.org/abs/2601.07022) points out in BF16 mode, Expert parallel did unnecessary token padding  (ps. also non-EP case).   
This PR set `TOKEN_GROUP_ALIGN_SIZE_M=1` by Default.

- For non-ep case,  we skip `indices_padding_wrapper`
- For ep case, `_permute` takes `TOKEN_GROUP_ALIGN_SIZE_M=1` and `padded_max_len = x.shape[0]` that can avoid any padding.  
- Now (extremely case) allows no-token experts.  For B100 its might need https://github.com/pytorch/pytorch/issues/172439

Test:

## Original implement (with `TOKEN_GROUP_ALIGN_SIZE_M=8`)
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --debug.seed 10`
image
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --debug.seed 10 --parallelism.expert_parallel_degree=2`
image

## And with this PR
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --debug.seed 10`
image
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --debug.seed 10 --parallelism.expert_parallel_degree=2`
image

## Another update to completely remove `TOKEN_GROUP_ALIGN_SIZE_M`
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --debug.seed 10`

image

`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --debug.seed 10 --parallelism.expert_parallel_degree=2`
image

### DIFF
--- a/torchtitan/models/moe/utils.py
+++ b/torchtitan/models/moe/utils.py
@@ -8,7 +8,6 @@
 import torch
 
 
-@torch.no_grad()
 def generate_permute_indices(
     tokens_per_expert_group: torch.Tensor,  # shape [num_ranks * experts_per_rank]
     experts_per_rank: int,
@@ -65,7 +64,10 @@ def _permute(x, num_tokens_per_expert, ep_degree, num_local_experts):
     max_len = x.shape[0]
 
     with torch.no_grad():
-        (permuted_indices, num_tokens_per_expert,) = generate_permute_indices(
+        (
+            permuted_indices,
+            num_tokens_per_expert,
+        ) = generate_permute_indices(
             num_tokens_per_expert,
             num_local_experts,
             ep_degree,


torchtitan/models/moe/utils.py
```diff
@@ -12,8 +12,8 @@
 
 from .kernels import generate_permute_indices
 
-TOKEN_GROUP_ALIGN_SIZE_M = 8
-ValidTokenGroupAlignmentSize = Literal[8, 16, 32]
+TOKEN_GROUP_ALIGN_SIZE_M = 1
```
tianyu-l: This fix is "soft", in the sense that the padding code path still exists for bf16.

I wonder whether it's viable to go one step further -- remove all padding logic for bf16 and move padding logic to quantized paths only. cc @danielvegamyhre

danielvegamyhre: Either way is fine, the 8 token alignment is needed if we want to use TMA in any kernels operating on each token group (8*2 bytes per elem = 16 byte alignment). However, if we are only doing that in the low precision code path, then there's no reason to pad.

Feel free to remove bf16 padding entirely.

danielvegamyhre:
Solar Open-102B technical report is very interesting @rakkit, thanks for sharing it!

torchtitan/models/moe/utils.py
ngimel: For bf16 8-token alignment is not needed anywhere, see 
```
import torch
x = torch.randn(2048, 4096, device="cuda", dtype=torch.bfloat16).requires_grad_(True)
w = torch.randn(2, 4096, 7168, device="cuda", dtype=torch.bfloat16).requires_grad_(True)
# odd offsets
offs = torch.tensor([1023, 2048], device="cuda", dtype=torch.int32)
out = F.grouped_mm(x, w, offs=offs)
gO = torch.rand_like(out)
out.backward(gO)
# check that gradients are computed
print(x.grad.sum(), w.grad.sum())
```

tianyu-l: @danielvegamyhre
Right now we are mixing padding and permutation into one kernel. Since bf16 doesn't require padding, I wonder if it makes sense to move padding to quantization kernel? The argument is that the kernel itself should be general and not require user to do padding from outside.

rakkit: sure if we agree move the padding logic to quant paths then i will refactor to remove `TOKEN_GROUP_ALIGN_SIZE_M` in torchtitan.

danielvegamyhre: @tianyu-l we have a version of the permutation and pad/fill kernel in torchao now, used in the MXFP8 EP primitives. It is not fused with quantization though. To clarify, are you asking if we can delete the permute+pad kernel from torchtitan and replace it with fused permute+pad+quantize kernel in torchao?

tianyu-l: @danielvegamyhre  My request is that we remove padding from torchtitan entirely, while keeping correctness.

In the past we have the permute+pad kernel to avoid d2h sync on the padding part. Now that if we no longer need padding for bf16, I'd hope we remove the kernel altogether, but that requires torchao to handle padding.

danielvegamyhre: That is doable

rakkit:
thanks @danielvegamyhre @tianyu-l 

made a refactor and this aligment things are completely removed now. (PR message also updated for test run).

i think we still need `generate_permute_indices` kernels for EP permute, but it runs on no-padding mode by default now

torchtitan/models/moe/utils.py
```diff
@@ -4,94 +4,32 @@
...
             ep_degree,
-            padded_max_len,
-            TOKEN_GROUP_ALIGN_SIZE_M,
+            max_len,
         )
 
-    x = torch.vstack((x, x.new_zeros((x.shape[-1]))))
     input_shape = x.shape
-    x = x[permuted_indices, :]
+    x = torch.index_select(x, 0, permuted_indices)
```
tianyu-l: curious, what's the reason for this change?

torchtitan/models/moe/utils.py
```diff
@@ -4,94 +4,32 @@
...
-
 
 def _permute(x, num_tokens_per_expert, ep_degree, num_local_experts):
-    global TOKEN_GROUP_ALIGN_SIZE_M
-    x_padded_per_expert = x.shape[0] + num_local_experts * TOKEN_GROUP_ALIGN_SIZE_M
-    padded_max_len = _round_up(x_padded_per_expert, TOKEN_GROUP_ALIGN_SIZE_M)
+    max_len = x.shape[0]
+
     with torch.no_grad():
         (permuted_indices, num_tokens_per_expert, _offsets,) = generate_permute_indices(
```
tianyu-l: > i think we still need `generate_permute_indices` kernels for EP permute, but it runs on no-padding mode by default now

Do we? IIUC we needed `generate_permute_indices` to avoid D2H sync (in getting pad sizes). Now that we don't do padding, can we just use pytorch code to do permutation? Please correct me if I'm wrong.

torchtitan/models/moe/utils.py
rakkit: i revert the changes, i though they end up at same kernel

torchtitan/models/moe/utils.py
rakkit: i addede a pytorch version `generate_permute_indices` [above](https://github.com/pytorch/torchtitan/pull/2255/changes#diff-98909a33198ce809df1ed0d20614172be08e34bf4f845287c0a08c8519b02840R11).

tianyu-l:
@danielvegamyhre let us know when the mxfp8 kernels supports padding.

danielvegamyhre:
@rakkit i have not closely reviewed the PR yet but as an intermediate solution can we just set TOKEN_GROUP_ALIGNMENT_SIZE=1 for bf16 until a performant solution for automatic padding in MXFP8? I'm prototyping some solutions to this, but it's harder than it sounds to not kill performance. The TLDR is it needs to be fused into the alltoall dispatch, which will take time, maybe ~weeks

rakkit:
hi @danielvegamyhre  i am not sure if i understand it correctly.

do you mean, set TOKEN_GROUP_ALIGNMENT_SIZE=1, and we let 
- BF16, without EP, no padding,
- BF16, with EP, using current padding kernels (with size=1)
- (MX)FP8, using current padding kernels

then its the init commits of this PR you can check it [here](https://github.com/pytorch/torchtitan/pull/2255/changes/97fae6b0d1c702dc90b88dd761ba1c4c9ba8da75)  so i can checkout to that commits?

danielvegamyhre:
> * BF16, with EP, using current padding kernels (with size=1)

on the bf16 path if alignment is 1, we can early exit. i am just asking to keep the padding path as it exists, and not delete it for now, so mxfp8 can continue to use it, until we land a proper solution, if is ok with everyone. 

additional extra context - at the moment, it looks like it will be either using HybridEP's native handling for this, or alternatively, if that doesn't work out, extending an experimental MXFP8 all-to-all implementation we have in torchao to also handle the padded.

once that solution is tested and landed, we can delete the current non-optimal padding logic entirely.

rakkit:
@danielvegamyhre 

> on the bf16 path if alignment is 1, we can early exit.

the kernel can do padding and perturb. 

what i mean is,  (BF16) we early exit for None-EP path, and using current padding kernels (with size=1) for perturb for EP path.   and thats init [commit](https://github.com/pytorch/torchtitan/pull/2255/changes/97fae6b0d1c702dc90b88dd761ba1c4c9ba8da75) of this PR

pianpwk:
@rakkit sorry, would you mind also looking into removing the for-loop path padding/unpadding as part of this PR? This would be an alternative fix for https://github.com/pytorch/torchtitan/pull/2399. Possibly conditional on `TOKEN_GROUP_ALIGNMENT_SIZE=1`, depending on how that settles: https://github.com/pytorch/torchtitan/blob/353c0643a02a7ad42f6323929b8365ac8b7a5382/torchtitan/models/common/moe/moe.py#L24

This patch should work, and on just one test, it gives me equivalent numerics before/after: 
```patch
diff --git a/torchtitan/models/moe/moe.py b/torchtitan/models/moe/moe.py
--- a/torchtitan/models/moe/moe.py
+++ b/torchtitan/models/moe/moe.py
@@ -81,12 +81,8 @@ def _run_experts_for_loop(
     # NOTE: this would incur a synchronization between device and host
     num_tokens_per_expert_list = num_tokens_per_expert.tolist()

-    # side-effect code due to the usage of generate_permute_indices
-    num_padding = x.shape[0] - sum(num_tokens_per_expert_list)
-
     # a tuple of tensors indexed by experts
     # each with shape (tokens_per_expert(varying), dim)
     x_splits = torch.split(
-        x[: sum(num_tokens_per_expert_list)],
+        x,
         split_size_or_sections=num_tokens_per_expert_list,
         dim=0,
     )
@@ -100,8 +96,6 @@ def _run_experts_for_loop(
         out_experts_splits.append(h)
     out = torch.cat(out_experts_splits, dim=0)

-    # side-effect code due to the usage of generate_permute_indices
-    out = torch.vstack((out, out.new_zeros((num_padding, out.shape[-1]))))
-
     return out
```

I think the forloop path only activates on < SM90 though.

rakkit:
thanks @pianpwk you are right we should remove potential padding for `_run_experts_for_loop` path as well

rakkit:
emm @danielvegamyhre  set alignment=1 might cause the trouble in BF16 + EP route.
when some experts get 0 tokens, alignment=1 will pad 1 tokens there

danielvegamyhre:
> emm @danielvegamyhre set alignment=1 might cause the trouble in BF16 + EP route. when some experts get 0 tokens, alignment=1 will pad 1 tokens there

ok if it's too much hassle feel free to just remove padding path for bf16

rakkit:
@danielvegamyhre thx, that will be pretty much the current version ( no condition on padding at all, + use pytorch for perturb w/o kernels), i will reabse and check again

tianyu-l:
> when some experts get 0 tokens, alignment=1 will pad 1 tokens there

I can confirm this is no longer an issue. With bf16, we don't need any padding.

Also the slight longer-term thinking is if it's possible to remove all padding even for mxfp8. E.g. the HybridEP kernel being landed can do padding during communication time.

danielvegamyhre:
padding/unpadding support in torchao has landed, feel free to resume this

rakkit:
thx a lot @danielvegamyhre 

I am not sure if @pianpwk is still working on https://github.com/pytorch/torchtitan/pull/2475; otherwise, I can close this PR and keep the conversation there.

pianpwk:
> thx a lot @danielvegamyhre
> 
> I am not sure if @pianpwk is still working on #2475; otherwise, I can close this PR and keep the conversation there.

ah yes, I'll resume soon, thanks @rakkit @danielvegamyhre

danielvegamyhre:
Created #2620 and referenced your torch native permute function @rakkit (thanks!). We can close this.

wwwjn:
Close because of #2620, @rakkit  feel free to reopen it if needed

PR #2263, opened 2026-01-20
[LoRA] Add LoRA converter for LoRA finetuning

mori360:
There's config component update at the end of Feb, we move to https://github.com/pytorch/torchtitan/pull/2484 to skip a lot of rebase, and apply ghstack to stack the further feature support on lora.

### Summary
To support LoRA (Low-Rank Adaptation) in torchtitan, this PR is the first step that adds LoRA adapters into the model and enables LoRA finetuning.

LoRALinear - A dynamically generated Linear layer subclass that adds trainable low-rank adapters (lora_a, lora_b) while keeping base weights frozen. Uses the decomposition output = base_output + (α/r) * B(A(x)) where r is rank and α is scaling factor. Compatible with any nn.Linear subclass (including Float8Linear).

LoRAConverter - A model converter that:
- Recursively replaces all nn.Linear layers with their LoRA equivalents
- Freezes base model weights (requires_grad=False)
- Overrides init_weights() to properly initialize LoRA adapters on the target device

### Details
1. Enable LoRA via ModelConverter
LoRA is integrated using the existing ModelConverter protocol, making it consistent with other model transformations (e.g., quantization). Users can enable LoRA by adding "lora" to the model.converters config:
```
[model]
converters = ["lora"]

[lora]
rank = 8
alpha = 16.0
dropout = 0.0
```

2. In-place Linear Replacement with LoRALinear Subclass
The implementation uses dynamic class creation via create_lora_linear() to wrap any nn.Linear subclass with LoRA adapters:

- Dynamic subclassing: `create_lora_linear(parent_cls) `creates a new class that inherits from the original Linear class, preserving custom forward logic (e.g., RowwiseParallel, ColwiseParallel)
- In-place modification: Instead of creating new modules, we create subclass preserving existing weights
- Frozen base weights: All base model parameters are frozen (requires_grad=False), only LoRA adapters are trainable
- Composability with quantization: we currently has float8 low-precision quantization as another choice in converters. Here we will conduct convert in order that convert to `Float8Linear` first, then in lora converter to convert from `Float8Linear` to `LoRAFloat8Linear` by `create_lora_linear` described above. # converters = ["quantize.linear.float8", "lora"]
    LoRA architecture is quite simple, that only adds lora_out to the base_out. So that as long as it has `base_out = super().forward(x)`, lora subclass won't touch the parent forward, so we could compose them together

### Test Plan
Tested with Llama3 debug model on 8 GPU
- lora has loess memory cost
- lora could work with float8

without lora converter
```
[rank0]:[titan] 2026-02-02 14:41:32,656 - root - INFO - step:  1  loss:  8.2000  grad_norm:  1.3955  memory:  1.01GiB(1.07%)  tps: 5,479  tflops: 0.39  mfu: 0.04%
[rank0]:[titan] 2026-02-02 14:41:33,139 - root - INFO - step:  5  loss:  5.3874  grad_norm:  2.3446  memory:  1.06GiB(1.11%)  tps: 343,441  tflops: 24.59  mfu: 2.49%
[rank0]:[titan] 2026-02-02 14:41:33,396 - root - INFO - step: 10  loss:  4.1315  grad_norm:  1.8140  memory:  1.06GiB(1.11%)  tps: 320,740  tflops: 22.96  mfu: 2.32%
```

with lora converter
```
[rank0]:[titan] 2026-02-02 14:42:44,379 - root - INFO - step:  1  loss:  8.1225  grad_norm:  0.3293  memory:  0.98GiB(1.03%)  tps: 6,222  tflops: 0.46  mfu: 0.05%
[rank0]:[titan] 2026-02-02 14:42:44,687 - root - INFO - step:  5  loss:  8.0054  grad_norm:  0.5108  memory:  1.02GiB(1.07%)  tps: 274,052  tflops: 20.05  mfu: 2.03%
[rank0]:[titan] 2026-02-02 14:42:44,988 - root - INFO - step: 10  loss:  7.8191  grad_norm:  0.7809  memory:  1.02GiB(1.07%)  tps: 266,976  tflops: 19.53  mfu: 1.98%
```

with lora converter + float8
```
[rank0]:[titan] 2026-02-02 14:43:38,608 - root - INFO - step:  1  loss:  8.1689  grad_norm:  0.2351  memory:  1.00GiB(1.05%)  tps: 5,067  tflops: 0.37  mfu: 0.04%
[rank0]:[titan] 2026-02-02 14:43:39,152 - root - INFO - step:  5  loss:  8.0590  grad_norm:  0.3869  memory:  1.04GiB(1.10%)  tps: 134,235  tflops: 9.82  mfu: 0.99%
[rank0]:[titan] 2026-02-02 14:43:39,792 - root - INFO - step: 10  loss:  7.9155  grad_norm:  0.6083  memory:  1.04GiB(1.10%)  tps: 122,904  tflops: 8.99  mfu: 0.91%
```

E2E memory result, running llama3_8b on 16 gpu, dp=16 with activation_checkpoint=full
without lora converter
Screenshot 2026-01-21 at 3 22 50 PM

with lora converter, and activation_checkpoint=full
Screenshot 2026-01-22 at 3 21 07 PM

### Future work
There are some todo works that not included in this PR and will be future work
1. Fix convert order, quantization first then lora
2. Better support QAT lora.
3. Composable with TP, the current plan does not work with TP, here are 2 plans we may try later and need further discussion:
- Keep lora_adapters as Replicate so that they won't be shardded by TP
- Generate layer_plan for lora adapters so that they will be processed by TP as well
4. Checkpoint options, needs to enable the choices:
- Save adapter params only
- Add adapter params to origin weight and save lora-processed weight
- Lora multi sources when origin model and lora adapters are saved differently
5. Support supervised finetuning

### DIFF
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -36,7 +36,16 @@ def apply_lora(linear: nn.Linear, rank: int, alpha: float) -> nn.Linear:
 
         class LoRALinear(parent_cls):  # type: ignore[valid-type, misc]
             def __init__(self, *args: Any, **kwargs: Any) -> None:
-                raise RuntimeError("LoRALinear should not be instantiated directly. ")
+                raise RuntimeError("LoRALinear should not be instantiated directly.")
+
+            @classmethod
+            def from_linear(
+                cls, linear: nn.Linear, rank: int, alpha: float
+            ) -> "LoRALinear":
+                """Convert an existing Linear to LoRALinear in-place."""
+                linear.__class__ = cls
+                linear._init_lora(rank, alpha)  # type: ignore[attr-defined]
+                return linear  # type: ignore[return-value]
 
             def _init_lora(
                 self,
@@ -77,11 +86,7 @@ def apply_lora(linear: nn.Linear, rank: int, alpha: float) -> nn.Linear:
         LoRALinear.__qualname__ = f"LoRA{parent_cls.__name__}"
         _lora_class_cache[parent_cls] = LoRALinear
 
-    # Change class in-place to avoid re-initializing the Linear
-    linear.__class__ = _lora_class_cache[parent_cls]
-    # Initialize LoRA adapters as post-init
-    linear._init_lora(rank, alpha)  # type: ignore[attr-defined]
-    return linear
+    return _lora_class_cache[parent_cls].from_linear(linear, rank, alpha)
 
 
 class LoRAConverter:


torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,304 @@
...
+import torch.nn.functional as F
+
+from torchtitan.config import JobConfig
+from torchtitan.distributed import ParallelDims
+from torchtitan.protocols.model_converter import register_model_converter
+from torchtitan.tools.logging import logger
+
+
+@dataclass
+class LoRAConfig:
```
wwwjn: Do we need this config even we have job config?

mori360: No we don't, remove this config and use job config directly.

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+
+    Forward: out = X @ W.T + bias + scale * (X @ A.T @ B.T)
+
+    Memory optimizations:
+    - Only saves X, A, B for backward
+    - Uses in-place addmm_ operations
+    """
+
+    @staticmethod
+    def forward(ctx, X, W, bias, A, B, scale):  # type: ignore[override]
```
fegin: Typing for the input please.

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+        if X.dim() == 3:
+            out = out.view(orig_shape[0], orig_shape[1], -1)
+
+        ctx.custom_saved_tensors = (W, scale)
+        ctx.save_for_backward(A, B, X)
+        ctx.has_bias = bias is not None
+        return out
+
+    @staticmethod
+    def backward(ctx, dY):  # type: ignore[override]
```
fegin: ditto

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+    - Only saves X, A, B for backward
+    - Uses in-place addmm_ operations
+    """
+
+    @staticmethod
+    def forward(ctx, X, W, bias, A, B, scale):  # type: ignore[override]
+        orig_shape = X.shape
+        X_2d = X.view(-1, X.shape[-1]) if X.dim() == 3 else X
+
+        out = torch.empty(X_2d.shape[0], W.shape[0], dtype=X.dtype, device=X.device)
```
fegin: Why create the output first? Can we just do `out=torch.mm(X_2d, W.t())`?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def __init__(
+        self,
+        in_dim: int,
+        out_dim: int,
+        rank: int,
+        alpha: float,
+        dropout: float = 0.0,
+        use_bias: bool = False,
+        weight: Optional[torch.Tensor] = None,
```
fegin: `torch.Tensor | None = None` and all other typings that use `Optional`.

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+            else:
+                self.register_parameter(
+                    "bias",
+                    nn.Parameter(
+                        torch.empty(out_dim, device="meta", dtype=dtype),
+                        requires_grad=False,
+                    ),
+                )
+        else:
+            self.register_parameter("bias", None)
```
fegin: What does this mean? `use_bias` but `bias` is None? `use_bias` seems to be not useful.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+                    ),
+                )
+        else:
+            self.register_parameter("bias", None)
+
+        self.dropout = nn.Dropout(p=dropout) if dropout > 0.0 else nn.Identity()
+
+        # LoRA layers on meta device
+        self.lora_a = nn.Linear(in_dim, rank, bias=False, device="meta", dtype=dtype)
+        self.lora_b = nn.Linear(rank, out_dim, bias=False, device="meta", dtype=dtype)
```
fegin: Model definition should not hard code device. The trainer will decide the device.

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def _lora_a_init_params(x: nn.Linear) -> None:
+    """Initialize LoRA A weight to Kaiming uniform."""
+    nn.init.kaiming_uniform_(x.weight, a=math.sqrt(5))
+
+
+def _lora_b_init_params(x: nn.Linear) -> None:
+    """Initialize LoRA B weight to zeros."""
+    nn.init.zeros_(x.weight)
```
fegin: Both are one-liner functions. We can merge them to `initialize_parameters`

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+
+        batch, seq_len, hd = X.shape
+        dY = dY.reshape(-1, dY.shape[-1])
+        X = X.reshape(-1, X.shape[-1])
+        A, B = A.t(), B.t()
+
+        d_A = torch.empty_like(A)
+        d_B = torch.empty_like(B)
+        d_A.addmm_(X.t(), dY @ B.t(), alpha=scale, beta=0)
+        d_B.addmm_(A.t() @ X.t(), dY, alpha=scale, beta=0)
```
fegin: Same here, why do we need to call `empty_like` first then inplace operations? Shouldn't we use functional APIs?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def to_empty(
+        self, *, device: Optional[Union[str, torch.device, int]], recurse: bool = True
```
fegin: `str | torch.device | int | None`

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+        self.lora_b.to_empty(device=device, recurse=recurse)
+        return self
+
+    def initialize_parameters(self):
+        _lora_a_init_params(self.lora_a)
+        _lora_b_init_params(self.lora_b)
+
+    def adapter_params(self) -> list[str]:
+        """Return names of LoRA adapter parameters."""
+        return ["lora_a.weight", "lora_b.weight"]
```
fegin: This method is not called anywhere in this PR. If it is needed in the future, you should add the method in the future PRs.

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+
+from torchtitan.config import JobConfig
+from torchtitan.distributed import ParallelDims
+from torchtitan.protocols.model_converter import register_model_converter
+from torchtitan.tools.logging import logger
+
+
+class _LoRALinearFunction(torch.autograd.Function):
```
fegin: Why do you need to use `torch.autograd.Function`? More specifically, why do you need to customize the backward? I may miss some context but I don't see the need of customizing the backward.

fegin (changes requested):
Can you update the summary to describe the design?

torchtitan/components/lora/lora.py
wwwjn: n00b q: would these autograd functions supports distributed tensors? eg, when TP is applied, W is sharded?

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,276 @@
...
+    """Memory-efficient LoRA linear computation.
+
+    Forward: out = X @ W.T + bias + scale * (X @ A.T @ B.T)
```
wwwjn: So in LoRA, A and B are trainable parameters,but W is the weight from pre-train and it's frozen during LoRA?

mori360: Yeah, linear.weight is frozen and won't be trained

mori360:
> Can you update the summary to describe the design?

_autograd is removed. With results updated in the description that full activation checkpoint could get the expected memory cost

torchtitan/components/lora/__init__.py
tianyu-l, line 1: It seems we don't need the lora folder and this file -- components/lora.py seems good enough.

torchtitan/__init__.py
```diff
@@ -6,6 +6,9 @@
 
 from importlib.metadata import version
 
+# Import to register lora module.
+import torchtitan.components.lora  # noqa: F401
```
tianyu-l, line 10: I think the order matters between lora and quantization, if both are used?

If you first do LoRA, then the original nn.Linear structure is gone so fp8 may be invalidated.
If you first do fp8, then LoRA probably doesn't work, because the LoRA implementation in this PR is replacing nn.Linear with F.linear calls.

cc @danielvegamyhre

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__()
+        self.in_dim = in_dim
+        self.out_dim = out_dim
+        self.rank = rank
+        self.alpha = alpha
+        self.scaling = alpha / rank
+        self.use_bias = use_bias
+        self.disabled = False
```
tianyu-l: why having this field?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,225 @@
...
+            f"LoRA training active with rank={self.rank}, alpha={self.alpha}, "
+            f"dropout={self.dropout}"
+        )
+
+    def convert(self, model: nn.Module) -> None:
+        """Inplace conversion of the model to use LoRA adapters."""
+        replacements = []
+        for module in model.modules():
+            for child_name, child in module.named_children():
+                if isinstance(child, nn.Linear) and not isinstance(child, LoRALinear):
```
tianyu-l: Do we eventually need to support MoE? If so, add a TODO for `GroupedExperts`.

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def convert(self, model: nn.Module) -> None:
+        """Inplace conversion of the model to use LoRA adapters."""
+        replacements = []
+        for module in model.modules():
+            for child_name, child in module.named_children():
+                if isinstance(child, nn.Linear) and not isinstance(child, LoRALinear):
+                    replacements.append((module, child_name, child))
+
+        for parent_module, child_name, child in replacements:
```
tianyu-l: why keeping a replacements list and do an extra loop, instead of converting in the first pass?

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+                in_dim=child.in_features,
+                out_dim=child.out_features,
+                rank=self.rank,
+                alpha=self.alpha,
+                dropout=self.dropout,
+                use_bias=has_bias,
+                weight=original_weight,
+                bias=original_bias,
+                dtype=child.weight.dtype
+                if child.weight is not None
```
tianyu-l: when would child.weight be None?

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+                else original_weight.dtype,
+            )
+            setattr(parent_module, child_name, lora_linear)
+
+        self._set_lora_requires_grad(model)
+        self._converted_model = model
```
tianyu-l: why having this field?

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+            if has_bias:
+                child.bias = None  # type: ignore[assignment]
+
+            lora_linear = LoRALinear(
+                in_dim=child.in_features,
+                out_dim=child.out_features,
+                rank=self.rank,
+                alpha=self.alpha,
+                dropout=self.dropout,
+                use_bias=has_bias,
```
tianyu-l: don't need this field if we are passing bias below as optional

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+                weight=original_weight,
+                bias=original_bias,
+                dtype=child.weight.dtype
+                if child.weight is not None
+                else original_weight.dtype,
+            )
+            setattr(parent_module, child_name, lora_linear)
+
+        self._set_lora_requires_grad(model)
```
tianyu-l: we only need to set `requires_grad` in two places
1. LoRALinear constructor
2. LoRALinear `init_weights`

torchtitan/components/lora.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # Wrap init_weights to also initialize LoRA parameters
+        original_init_weights = model.init_weights
+
+        def init_weights_with_lora(*args, **kwargs):
+            if callable(original_init_weights):
+                original_init_weights(*args, **kwargs)
+            self._init_lora_params(model)
+
+        object.__setattr__(model, "init_weights", init_weights_with_lora)
```
tianyu-l: we should do this during the original for loop over all modules, instead of doing an additional for-loop in `_init_lora_params`

torchtitan/components/lora/lora.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # LoRA path: out += scale * (x @ A.T @ B.T)
+        x = self.dropout(x)
+        lora_out = self.lora_b(self.lora_a(x))
+        out = out + self.scaling * lora_out
+
+        return out
+
+
+def _lora_a_init_params(x: nn.Linear) -> None:
```
tianyu-l: no need to have these one-line functions

torchtitan/components/lora.py
```diff
@@ -0,0 +1,225 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+
+from torchtitan.config import JobConfig
+from torchtitan.distributed import ParallelDims
+from torchtitan.protocols.model_converter import register_model_converter
+from torchtitan.tools.logging import logger
+
+
+class LoRALinear(nn.Module):
```
tianyu-l: There are two ways to do this class:
1. having original nn.Linear as a submodule
2. getting the weight and bias in nn.Linear and apply nn.Linear on them this module

What are the trade-offs?

Speaking of parallelisms, I think
- FSDP would work and just shard the lora adapters, although we discussed that maybe they can just stay replicated?
- TP would break because dtensor TP API is looking for nn.Linear and assuming only weight and bias are there https://github.com/pytorch/pytorch/blob/b731ffe3467840e1b6da911487310944be4f1687/torch/distributed/tensor/parallel/style.py#L149. @fegin and I are considering rewriting TP without the above dtensor APIs so we can delay the integration with TP for now.
- quantization likely won't work, see comment above. We should re-design and make sure they work together, if this is a use case @joecummings

torchtitan/__init__.py
danielvegamyhre: Yeah, a linear subclass for LoRA will be incompatible with FP8 linear subclass, seems like the user must use one or the other. For now, I'd suggest we add validation enforcing this.

torchtitan/components/lora.py
mori360: Thanks for the comments, I planed lora+quantization as the prs later.
Will have some try on lora+quantization, redesign and update in this pr.

torchtitan/__init__.py
fegin: @danielvegamyhre @mori360 

Will this style work? https://gist.github.com/fegin/ebc28fc9542300e97a5e88893d7e87dd

danielvegamyhre: Hmm I'm not sure. I'd say just try a minimal script/prototype using the torchao Float8Linear layer and the Lora layer that uses this pattern and test it out. I can help test / consult on this as well if you want

mori360: > Hmm I'm not sure. I'd say just try a minimal script/prototype using the torchao Float8Linear layer and the Lora layer that uses this pattern and test it out. I can help test / consult on this as well if you want

Are there any suggested tests on LoraFloat8Linear that we care when using Float8Linear?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            **kwargs: Any,
+        ) -> None:
+            super().__init__(*args, **kwargs)
+            self._init_lora(rank, alpha, dropout)
+
+        def _init_lora(self, rank: int, alpha: float, dropout: float) -> None:
+            self._lora_scaling = alpha / rank
+            self._lora_dropout = nn.Dropout(dropout) if dropout > 0 else nn.Identity()
+            self.lora_a = nn.Linear(
+                self.in_features,
```
tianyu-l: Are you assuming parent_cls is a subclass of nn.Linear? If not, we need to assert such things like self.in-features, self.weight exist.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            alpha: float = 16.0,
+            dropout: float = 0.0,
+            **kwargs: Any,
+        ) -> None:
+            super().__init__(*args, **kwargs)
+            self._init_lora(rank, alpha, dropout)
+
+        def _init_lora(self, rank: int, alpha: float, dropout: float) -> None:
+            self._lora_scaling = alpha / rank
+            self._lora_dropout = nn.Dropout(dropout) if dropout > 0 else nn.Identity()
```
tianyu-l: how common is dropout used in LoRA implementations?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _apply_lora(self, module: nn.Module) -> None:
+        """Recursively freeze params and wrap Linear layers with LoRA."""
+        for name, child in list(module._modules.items()):
+            if name in ("lora_a", "lora_b") or child is None:
+                continue
+            # Freeze direct parameters of this module
+            for param in child.parameters(recurse=False):
+                param.requires_grad_(False)
+            if isinstance(child, nn.Linear):
```
tianyu-l, line 107: do we only care about Linear? What if it's mxfp8 linear?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            device = args[0] if args else kwargs.get("buffer_device", "cuda")
+            for module in self._lora_modules:
+                module.weight.requires_grad_(False)
+                if module.bias is not None:
+                    module.bias.requires_grad_(False)
+                module.lora_a.to_empty(device=device, recurse=True)  # type: ignore[union-attr]
+                module.lora_b.to_empty(device=device, recurse=True)  # type: ignore[union-attr]
+                self._init_lora_weights(module)
+
+        object.__setattr__(model, "init_weights", new_init_weights)
```
tianyu-l: does this only init weights for lora modules but not others?

Like I mentioned before, I think we should modify `init_weights` only for leaf lora modules, not the parent model.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+            self.lora_b = nn.Linear(
+                rank,
+                self.out_features,
+                bias=False,
+                device=self.weight.device,
+                dtype=self.weight.dtype,
+            )
+            nn.init.kaiming_uniform_(self.lora_a.weight, a=math.sqrt(5))
+            nn.init.zeros_(self.lora_b.weight)
```
tianyu-l: do we need to call init here?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+        linear.lora_b = nn.Linear(  # type: ignore[attr-defined]
+            self.rank,
+            linear.out_features,
+            bias=False,
+            device=linear.weight.device,
+            dtype=linear.weight.dtype,
+        )
+
+        if linear.weight.device.type != "meta":
+            self._init_lora_weights(linear)
```
tianyu-l: we need to override init_weights here, not doing the actual init

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            device=linear.weight.device,
+            dtype=linear.weight.dtype,
+        )
+        linear.lora_b = nn.Linear(  # type: ignore[attr-defined]
+            self.rank,
+            linear.out_features,
+            bias=False,
+            device=linear.weight.device,
+            dtype=linear.weight.dtype,
+        )
```
tianyu-l: why doing this again?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+        self.rank = job_config.lora.rank
+        self.alpha = job_config.lora.alpha
+        self.dropout = job_config.lora.dropout
+        self._lora_modules: list[nn.Module] = []
+
+    def convert(self, model: nn.Module) -> None:
+        """Apply LoRA to all Linear layers, freezing base model weights."""
+        self._apply_lora(model)
+
+        if self._is_meta_device(model):
```
tianyu-l: the override should happen regardless where the model is put.

torchtitan/components/lora.py
fegin: Does mxfp8 subclass from nn.Linear? cc., @danielvegamyhre

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            self._hook_init_weights(model)
+
+    def _apply_lora(self, module: nn.Module) -> None:
+        """Recursively freeze params and wrap Linear layers with LoRA."""
+        for name, child in list(module._modules.items()):
```
fegin: Why use `_modules` and do you need to make this _apply_lora a recursive function? Since `instance(child, nn.Linear)` is the terminate condition, I think `for name, child in module.named_modules` is enough?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            self._init_lora_weights(linear)
+
+        return linear
+
+    def _init_lora_weights(self, module: nn.Module) -> None:
+        """Initialize LoRA weights and enable gradients."""
+        nn.init.kaiming_uniform_(module.lora_a.weight, a=math.sqrt(5))  # type: ignore[union-attr]
+        nn.init.zeros_(module.lora_b.weight)  # type: ignore[union-attr]
+        module.lora_a.weight.requires_grad_(True)  # type: ignore[union-attr]
+        module.lora_b.weight.requires_grad_(True)  # type: ignore[union-attr]
```
fegin: This function should be `_init_weight()` and it should call `super()._init_weight()` and should be in `Lora` class.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _is_meta_device(self, model: nn.Module) -> bool:
+        return any(p.device.type == "meta" for p in model.parameters())
```
fegin: Model shouldn't care about meta device or not.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _hook_init_weights(self, model: nn.Module) -> None:
```
fegin: This function is not needed at all.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+    """Create a LoRA-enabled subclass of a Linear layer.
+
+    Args:
+        parent_cls: A nn.Linear subclass to extend with LoRA.
+
+    Returns:
+        A new class with LoRA adapters that inherits from parent_cls.
+    """
+    if parent_cls in _lora_class_cache:
+        return _lora_class_cache[parent_cls]
```
fegin: We should assert that parent_clss is an `nn.Linear` provided that MXFP8 inherits from `nn.Linear`. cc., @danielvegamyhre please correct me if MXFP8 doesn't inherit from `nn.Linear`.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+        def __init__(
+            self,
+            *args: Any,
+            rank: int = 8,
+            alpha: float = 16.0,
+            dropout: float = 0.0,
+            **kwargs: Any,
+        ) -> None:
+            super().__init__(*args, **kwargs)
+            self._init_lora(rank, alpha, dropout)
```
fegin: We never actually call this __init__(). We should assert if this is called.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,167 @@
...
+            if isinstance(child, nn.Linear):
+                setattr(module, name, self._wrap_linear(child))
+                self._lora_modules.append(getattr(module, name))
+            else:
+                self._apply_lora(child)
+
+    def _wrap_linear(self, linear: nn.Linear) -> nn.Module:
+        """Wrap a Linear layer with LoRA adapters."""
+        lora_cls = create_lora_linear(linear.__class__)
+        linear.__class__ = lora_cls
```
fegin: Can't we do `linear._init_lora()` to initialize? The rest should go there. In fact, the rest just duplicate what `_init_lora()` does.

fegin (changes requested):
There are refactors needed for Lora factory function.

mori360:
> There are refactors needed for this PR.

I reference https://github.com/pytorch/ao/blob/0a5638afebf0e1c627b54ba133aa406bdadf6eb3/torchao/prototype/mx_formats/mx_linear.py#L188 to take MXFP8Linear as nn.Linear

fegin:
> I reference [pytorch/ao@0a5638a/torchao/prototype/mx_formats/mx_linear.py#L188](https://github.com/pytorch/ao/blob/0a5638afebf0e1c627b54ba133aa406bdadf6eb3/torchao/prototype/mx_formats/mx_linear.py#L188) to take MXFP8Linear as nn.Linear

The high level idea is correct but there are some method overwrites that should happen just in Lora class, such as `_init_weight`

torchtitan/components/lora.py
mori360: It's used in Tune and unsloth. Removed now, we can add when it's needed.

torchtitan/components/lora.py
tianyu-l: cc @andrewor14 who has some QATLinear work to be integrated with LoRA, too

andrewor14: Yes we have `FakeQuantizedLinear` for QAT that we wish to compose with LoRA. This is a subclass of `nn.Linear` though:

https://github.com/pytorch/ao/blob/1b4b6d998bf988f059e97a10181cbc4aec269b69/torchao/quantization/qat/linear.py#L42

torchtitan/components/lora.py
mori360: In llama3, Attenion calls 'init_weights', and init its wq, wk, wv, wo by `nn.init.trunc_normal_`, thus there's no 'init_weights' call at nn.Linear level(for wq, wk, wv, wo). The current design calls at nn.Linear's parents, which is the most downstream module that still defines an init_weights method.
https://github.com/pytorch/torchtitan/blob/41753c67b092518d5f797c9efefa7293fe7134f6/torchtitan/models/llama3/model/model.py#L223-L226

torchtitan/components/lora.py
```diff
@@ -0,0 +1,173 @@
...
+        def _init_weight(self) -> None:
+            _super_init_weight = getattr(super(), "_init_weight", None)
+            if callable(_super_init_weight):
+                _super_init_weight()
+            nn.init.kaiming_uniform_(self.lora_a.weight, a=math.sqrt(5))
+            nn.init.zeros_(self.lora_b.weight)
+
+        def forward(self, input: torch.Tensor) -> torch.Tensor:
+            base_out = super().forward(input)
+            # Compute LoRA in weight dtype, cast output to match base_out
```
fegin: This is a bit confusing. Where is the casting?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,173 @@
...
+    Returns:
+        The same linear object, now with LoRA adapters attached.
+    """
+    lora_cls = create_lora_linear(type(linear))
+    # Change class in-place to avoid re-initializing the Linear
+    linear.__class__ = lora_cls
+    # Initialize LoRA adapters as post-init
+    _init_lora = getattr(linear, "_init_lora", None)
+    if callable(_init_lora):
+        _init_lora(rank, alpha)
```
fegin: Why do we need this if? Since we inherit from LoRALinear, we should always have `_init_lora`?

torchtitan/components/lora.py
fegin: This is not changed? Question is still the same, why do you need to access `_module`?

torchtitan/components/lora.py
mori360: Yeah, this is for Pyrefly `Expected a callable, got `Tensor`Pyrefly[not-callable]`

torchtitan/components/lora.py
mori360: module._modules here could provide the parent module that has child in `nn.Linear`, they are the most leaf module to overwrite init_weight, module._modules could find them much easier. (those `nn.Linear` are not inited by init_weight).

Let me have some try to use module.named_modules and set init_weights for the leafs.

torchtitan/components/lora.py
fegin: Then a better way is to do `assert _init_lora is not None and callable(_init_lora)`.

torchtitan/components/lora.py
mori360: It means that lora adapters are cast to bfloat16 by fully_shard, so that base_out and lora_out are both in bf16. Yeah, it's confusing, will remove it.

torchtitan/components/lora.py
mori360: Now store the fqn for the parent module who has child in nn.Linear and apply lora.
Later we find the parents by fqn to set init_weights

torchtitan/components/lora.py
```diff
@@ -0,0 +1,175 @@
...
+                # set `init_weights`` later
+                if "." in fqn:
+                    parent_fqn = fqn.rsplit(".", 1)[0]
+                    parents_with_lora.add(parent_fqn)
+
+        for parent_fqn in parents_with_lora:
+            parent = module.get_submodule(parent_fqn)
+            self._override_parent_init_weights(parent)
+
+    def _override_parent_init_weights(self, module: nn.Module) -> None:
```
fegin: I don't understand why we need to override _init_weights. We already define `_init_weight` in LoRALinear and that `_init_weight` calls super()._init_weight(). What are the items that we cannot do everything there?

fegin (changes requested):

torchtitan/components/lora.py
mori360: The current PR tested on Llama3. In llama3, Attenion calls 'init_weights', and init its wq, wk, wv, wo by nn.init.trunc_normal_, thus there's no 'init_weights' call at nn.Linear level(for wq, wk, wv, wo). Thus super()._init_weight() is not called. So that we set init_weight to wq, wk, wv, wo's parent, at attention.

https://github.com/pytorch/torchtitan/blob/41753c67b092518d5f797c9efefa7293fe7134f6/torchtitan/models/llama3/model/model.py#L223-L226

torchtitan/components/lora.py
```diff
@@ -0,0 +1,122 @@
...
+class LoRAConverter:
+    """Converts a model to use LoRA adapters on all Linear layers."""
+
+    def __init__(self, job_config: JobConfig, parallel_dims: ParallelDims):
+        self.rank = job_config.lora.rank
+        self.alpha = job_config.lora.alpha
+
+    def convert(self, model: nn.Module) -> None:
+        """Freeze base model weights and apply LoRA to all Linear layers."""
+        model.requires_grad_(False)
```
mori360, line 101: We need to freeze all params(not only nn.Linear), so that freeze cannot be included in LoraLinear

torchtitan/components/lora.py
```diff
@@ -0,0 +1,124 @@
...
+                return base_out + self._lora_scaling * lora_out
+
+        LoRALinear.__name__ = f"LoRA{parent_cls.__name__}"
+        LoRALinear.__qualname__ = f"LoRA{parent_cls.__name__}"
+        _lora_class_cache[parent_cls] = LoRALinear
+
+    # Change class in-place to avoid re-initializing the Linear
+    linear.__class__ = _lora_class_cache[parent_cls]
+    # Initialize LoRA adapters as post-init
+    linear._init_lora(rank, alpha)  # type: ignore[attr-defined]
```
tianyu-l: Curious why we do it here, instead of in the constructor of `LoRALinear`?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,124 @@
...
+                self,
+                rank: int,
+                alpha: float,
+                device: torch.device | None = None,
+                dtype: torch.dtype | None = None,
+            ) -> None:
+                self._lora_scaling = alpha / rank
+                device = device if device is not None else self.weight.device
+                dtype = dtype if dtype is not None else self.weight.dtype
+                self.lora_a = nn.Linear(
```
tianyu-l, line 60: @andrewor14 told me that in the case of QAT, the adapters should be of QAT subclass, not nn.Linear.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,124 @@
...
+                apply_lora(child, self.rank, self.alpha)
+
+        # Override the model's init_weights to also reinitialize LoRA weights
+        original_init_weights = getattr(module, "init_weights", None)
+
+        def new_model_init_weights(*args: Any, **kwargs: Any) -> None:
+            if original_init_weights is not None and callable(original_init_weights):
+                original_init_weights(*args, **kwargs)
+            # Reinitialize LoRA weights for all LoRALinear modules
+            for sub_module in module.modules():
```
tianyu-l, line 117: This is fine, but I think I'm leaning towards having our own nn.Linear subclass (in some sense, TorchtitanLinear) which has `init_weights` definition, to be consistent with all other modules. I'm not sure how well it plays with quantization linear subclasses.
cc @andrewor14 @danielvegamyhre

torchtitan/components/lora.py
mori360: In the usecase of QAT here https://github.com/meta-pytorch/torchtune/blob/b3c964f360299da49f3a519aed18752ceedeffcf/torchtune/training/quantization.py#L203-L242
It has `fake_quantize(W_frozen) @ fake_quantize(x) + BAx` with quantization to weight and input x, lora adapters are kept as nn.Linear

tianyu-l: @andrewor14 could you help confirm? thanks!

torchtitan/components/lora.py
mori360: Here we set `init_weight` at model level (because we may not call init_weight at nn.Linear), so the new init weight here adds `init adapter weight` after model's origin `init weight`.

torchtitan/components/lora.py
andrewor14: > in the case of QAT, the adapters should be of QAT subclass, not nn.Linear.

That's right, they should be `FakeQuantizedLinear` instead. @mori360 sorry the torchtune implementation is a bit oudated. Here's the more updated QAT + LoRA impmlementation in [unsloth](https://github.com/unslothai/unsloth/blob/9e75e1b7e3c5ebb28787aa8ae0bbdf40afdde8d8/unsloth/kernels/utils.py#L280).

But I think the current code is fine since `FakeQuantizedLinear` doesn't exist in torchtitan yet. After this PR lands, the way I can see us integrating QAT + LoRA is as follows:
1. First do torchao QAT module swap, so all `nn.Linear`s -> `FakeQuantizedLinear`
2. Call `apply_lora` here with `parent_cls` == `FakeQuantizedLinear`
3. In `apply_lora`, detect this case and initialize the LoRA adapters to `FakeQuantizedLinear` with the same configs instead of a simple `nn.Linear`

torchtitan/components/lora.py
andrewor14: The main torchao QAT API expects `nn.Linear`s in the model. Even if `TorchtitanLinear` inherits from `nn.Linear` we will just swap this away, so whatever functionality is in `TorchtitanLinear` will be lost.

Alternatively, we can integrate at a finer-grained level where we don't use torchao's `FakeQuantizedLinear` at all (some of our internal customers did this), where we just apply `FakeQuantizer(x)` to transform the input activations / weights / adapters directly.

torchtitan/components/lora.py
fegin: Because we don't know the exact arguments of the parent linear.

torchtitan/components/lora.py
mori360: > so whatever functionality is in TorchtitanLinear will be lost

If we convert by `nn.Linear` -> `FakeQuantizedLinear` -> `TorchtitanLinear(FakeQuantizedLinear)`, could we keep functionalities?

fegin: We can change the converters, including QAT and Lora, in TorchTitan to convert the linear back to TorchTitanLinear. @mori360 is this what you suggest?

torchtitan/components/lora.py
mori360: The current design uses `from_linear` instead if `__init__`, so that we still do not need exact arguments as well, only rank and alpha for lora adapters.

torchtitan/components/lora.py
andrewor14: > If we convert by nn.Linear -> FakeQuantizedLinear -> TorchtitanLinear(FakeQuantizedLinear), could we keep functionalities?

Yep that works! Assuming `TorchtitanLinear` calls `super().forward()`

mori360:
There's config component update at the end of Feb, we move to https://github.com/pytorch/torchtitan/pull/2484 to skip a lot of rebase, and apply ghstack to stack the further feature support on lora.

PR #2265, opened 2026-01-21
fix sdpa-varlen attention mismatch in qwen3 (#2229)

tf170898:
Fix for https://github.com/pytorch/torchtitan/issues/2223


tianyu-l:
won't land

PR #2267, opened 2026-01-21
[varlen_attn] change is_causal to window_size

liangel-02:
`is_causal` flag has been deprecated in `varlen_attn`, use `window_size = [-1, 0]` instead, see [this PR](https://github.com/pytorch/pytorch/pull/172245)

*Test*
Screenshot 2026-01-21 at 11 55 59 AM

### DIFF
--- a/torchtitan/models/attention.py
+++ b/torchtitan/models/attention.py
@@ -84,8 +84,18 @@ class VarlenAttentionWrapper(torch.nn.Module):
             cu_seq_k,
             max_q,
             max_k,
-            is_causal=True,
             scale=scale,
+            # window_size=(left, right) controls the attention window relative to each
+            # query position. 'left' is how many tokens before the query to attend to,
+            # and 'right' is how many tokens after. A value of -1 means unlimited.
+            #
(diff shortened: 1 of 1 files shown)


torchtitan/models/attention.py
```diff
@@ -84,8 +84,8 @@ def forward(
             cu_seq_k,
             max_q,
             max_k,
-            is_causal=True,
             scale=scale,
+            window_size=[-1, 0],
```
fegin: Can we add descriptions to describe what -1 means and what 0 means? I checked the API's docstring, but it is still not very clear to me.

torchtitan/models/attention.py
```diff
@@ -84,8 +84,10 @@ def forward(
             cu_seq_k,
             max_q,
             max_k,
-            is_causal=True,
             scale=scale,
+            # causal attention, token looks at all tokens before it
+            # and itself but no tokens after it
+            window_size=(-1, 0),
```
tianyu-l, line 98: It's still not clear what -1 and 0 means. Maybe explain it here or at least add a link to documents.

PR #2270, opened 2026-01-22
Fix grad norm clipping for AutoP and dsv3 model init

sanketpurandare:
Stacked PRs:
 * #2271
 * __->__#2270

--- --- ---

### DIFF
--- a/torchtitan/experiments/autoparallel/local_map_deepseek_v3/model.py
+++ b/torchtitan/experiments/autoparallel/local_map_deepseek_v3/model.py
@@ -15,4 +15,9 @@ from .args import DeepSeekV3ModelArgs
 # Need to share same base class with torchtitan models
 class DeepSeekV3Model(_DeepSeekV3Model, ModelProtocol):
     def __init__(self, model_args: DeepSeekV3ModelArgs):
-        super().__init__(model_args)
+        # Call _DeepSeekV3Model.__init__ which calls nn.Module.__init__
+        # Note: We don't call ModelProtocol.__init__ separately because:
+        # 1. nn.Module.__init__() is already called by _DeepSeekV3Model.__init__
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/autoparallel/local_map_deepseek_v3/model.py
```diff
@@ -15,4 +15,9 @@
 # Need to share same base class with torchtitan models
 class DeepSeekV3Model(_DeepSeekV3Model, ModelProtocol):
     def __init__(self, model_args: DeepSeekV3ModelArgs):
-        super().__init__(model_args)
+        # Call _DeepSeekV3Model.__init__ which calls nn.Module.__init__
+        # Note: We don't call ModelProtocol.__init__ separately because:
+        # 1. nn.Module.__init__() is already called by _DeepSeekV3Model.__init__
+        # 2. Calling ModelProtocol.__init__ after would reset all module state
+        #    (nn.Module.__init__ clears _modules, _parameters, etc.)
+        _DeepSeekV3Model.__init__(self, model_args)
```
xmfan, line 23: will break when ModelProtocol.__init__ does subclass init things

sanketpurandare: Do you have a better solution in mind?

xmfan: we could use torchtitan as the model repository, instead of autoparallel

sanketpurandare: That's a good idea, we'll do it after we have tested CP as well in auto parallel since that would alos require local map and is faster to iterate when we have model definition in AutoP

PR #2259, opened 2026-01-20
GQA without kv repeats

francesco-bertolotti:
PR for https://github.com/pytorch/torchtitan/issues/2224

This is a first tentative commit for the GQA attention without extra transpose and repeats. @tianyu-l, please let me know what you think. Since I changed the attention wrappers I can also check if other architecture are affected.

PR #2277, opened 2026-01-27
[rl] GQA attention fix

wwwjn:
Depends on #2259 

Context: Fix RL attention implementation, add numerics check scripts comparing torchtitan Qwen3 model (with vllm wrapper) and vllm native Qwen3 model

NOTE: I tried enabling CI for this work, but need more efforts install vllm from source code (need to build vllm from scratch). Will enable CI in later PRs


fegin:
Can you split the test to another PR? It is not related to the GQA fix.

wwwjn:
Closed because of #2291

PR #2282, opened 2026-01-28
[Typing] Fix pyrefly-ignore in train.py

fegin:
The main goal is to fix train.py. Since train.py uses many components, this PR also fixes some other components as well.

Trainer.model_parts should be `list[ModelProtocol]` but this change will affect many files. Will do this in another PR.

Some ignores are already useless, not sure why those ignores are not removed previously using `pyrefly check --remove-unused-ignores`.

Total pyrefly ignores removed: 19

### DIFF
--- a/torchtitan/components/dataloader.py
+++ b/torchtitan/components/dataloader.py
@@ -9,7 +9,9 @@
 import inspect
 import pickle
 from abc import ABC, abstractmethod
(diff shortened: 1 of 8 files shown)


fegin:
* __->__ #2282

tianyu-l (approved):

PR #2283, opened 2026-01-28
[Typing] Fix pyrefly ignores in checkpoint.py

fegin:
* #2286
* #2285
* #2284
* __->__ #2283

No logic change.

This PR also changes optimizer.py and state_dict_adapter.py because checkpointer depends on Optimizer and StateDictAdapter.

### DIFF
--- a/tests/unit_tests/test_checkpoint.py
+++ b/tests/unit_tests/test_checkpoint.py
@@ -14,6 +14,7 @@ from unittest import mock
 
 import torch
 import torch.nn as nn
+from torch.distributed.checkpoint.state_dict_saver import AsyncSaveResponse
 from torch.utils.data import DataLoader
 from torchtitan.components.checkpoint import CheckpointManager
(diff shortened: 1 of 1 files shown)


torchtitan/components/checkpoint.py
```diff
@@ -526,9 +523,8 @@ def save(self, curr_step: int, last_step: bool = False) -> None:
                     checkpoint_id=checkpoint_id,
                     async_mode=self.async_mode,
                 )
-                # pyrefly: ignore [missing-attribute]
+                assert isinstance(result, AsyncSaveResponse)
```
wwwjn, line 526: This line causing CPU unit test error?

joecummings:
> CI needs fixing

Trying here https://github.com/pytorch/torchtitan/pull/2287, saw it earlier

torchtitan/components/checkpoint.py
fegin: Thanks!

PR #2289, opened 2026-01-28
[BE week] Disable CPU wheel builds in nightly CI

joecummings:
TorchTitan requires triton which isn't available on CPU-only builds. Since there's no use case for CPU-only torchtitan wheels, disable them rather than adding complexity to support CPU imports.

Testing: All green here: https://github.com/pytorch/torchtitan/actions/runs/21458236045?pr=2289 and on CI

### DIFF
--- a/.github/workflows/build_whl_and_publish.yaml
+++ b/.github/workflows/build_whl_and_publish.yaml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - nightly
-      - fix-nightly-cpu  # TODO: Remove after testing
-  pull_request:  # TODO: Remove after testing
   workflow_dispatch:
 
 permissions:


wconstab:
Were we actually testing some part of torchtitan on CPU previously and now we're only testing on gpu? Or what is the nuance?

joecummings:
> Were we actually testing some part of torchtitan on CPU previously and now we're only testing on gpu? Or what is the nuance?

This exact workflow builds a nightly version of torchtitan without any accelerator support and publishes it to the PyTorch Index. Torchtitan doesn't actually have a use for a nightly build that does NOT include any accelerator support. However, the CPU build comes as default with the workflow. This wasn't a problem until a direct triton import was added. This caused the CPU nightly build smoketest to fail. Rather than gating the Trition import, it makes sense to just remove the CPU nightly build.

Note: This does not affect the "CPU Tests" in CI b/c those don't use the nightly CPU build.

PR #2292, opened 2026-01-29
[rl] Add numerics test against vllm native inference

wwwjn:
* __->__ #2292
* #2291


tianyu-l (commented):
It seems this test is not being run by CI so we can't guard against it?

wwwjn:
> It seems this test is not being run by CI so we can't guard against it?

Yes, I tried setup a CI, but the main blocker now is building vllm from source would take ~20mins (we need to build vllm from source because torchtitan need pytorch-nightly). I don't have a good solution for this right now. So this test is used for local reproduce mainly - I think we can also leave it here and merge it.

wwwjn:
Close because of #2300

PR #2293, opened 2026-01-29
Add Transformer-Engine Fused_Adam Optimizer Support

vivekgoe:
### Summary
For some models such as DeepSeek-V3, optimizer states (moments) can be stored in lower precision (torch.bfloat16) without loosing training accuracy. Storing optimizer states in lower precisions translates into significant memory saving for large models such as DeepSeek-V3-671B and allows to increase device utilization by increasing Batch-Size. 
This PR adds Transformer-Engine(TE) Fused_Adam optimizer to list of optimizers supported by TorchTitan. This optimizer is a drop-in replacement for torch.optim.AdamW with additional features such as lower precision optimizer states storage support.

#### Accuracy
Tested on DeepSeek-V4-671B (EP=64, PP=4, GBS=1024).
Code_Generated_Image

#### Memory
Tested on DeepSeek-V4-671B (EP=64, PP=4, GBS=1024). 
Code_Generated_Image (1)

### DIFF
--- a/torchtitan/components/optimizer.py
+++ b/torchtitan/components/optimizer.py
@@ -314,7 +314,7 @@ def build_optimizers(
 
     if name == 'TE_FusedAdamW':
         try:
-            from transformer_engine.pytorch.optimizers.fused_adam import TE_FusedAdam
+            from transformer_engine.pytorch.optimizers.fused_adam import FusedAdam
         except (ImportError, ModuleNotFoundError) as e:
             raise ImportError(
                 "FusedAdam optimizer could not be imported from transformer_engine. "
@@ -323,7 +323,7 @@ def build_optimizers(
                 "git+https://github.com/NVIDIA/TransformerEngine.git@stable "
                 "or use another optimizer"
             ) from e
-        optimizer_classes["TE_FusedAdamW"] = TE_FusedAdam
+        optimizer_classes["TE_FusedAdamW"] = FusedAdam
         del optimizer_kwargs['fused'], optimizer_kwargs['foreach']
         optimizer_kwargs['exp_avg_dtype'] = torch.bfloat16
         optimizer_kwargs['exp_avg_sq_dtype'] = torch.bfloat16


vivekgoe:
@tianyu-l Please help review this PR. I have kept it as draft for now because this needs a bug fix in Transformer-Engine (TE) related to DTensor handling which will be available in next TE release (fix already available in a pre-release branch). If it looks good to you I can take it out of draft after next TE release. Thanks.

torchtitan/components/optimizer.py
```diff
@@ -311,6 +311,23 @@ def build_optimizers(
...
+                "FusedAdam optimizer could not be imported from transformer_engine. "
+                "Install transformer_engine with: "
+                "NVTE_FRAMEWORK=pytorch pip install --no-build-isolation "
+                "git+https://github.com/NVIDIA/TransformerEngine.git@stable "
+                "or use another optimizer"
+            ) from e
+        optimizer_classes["TE_FusedAdamW"] = FusedAdam
+        del optimizer_kwargs['fused'], optimizer_kwargs['foreach']
+        optimizer_kwargs['exp_avg_dtype'] = torch.bfloat16
+        optimizer_kwargs['exp_avg_sq_dtype'] = torch.bfloat16
```
tianyu-l, line 329: Thanks @vivekgoe 

My impression is that low-precision optimizer states and TE dependency are two separate topics, each worth its own discussions.

For low-precision optimizer states
- @vivekgoe Could you show evidences that this is becoming the default for training?
- @janeyx99 Does pytorch optimizer plan to support this feature?

In terms of dependency, I know that TE is an important and popular optimization library. However, torchtitan has been about pytorch native, prioritizing simplicity & maintainability, and platform neutral.
- In the past, we have asked other hardware partners to create their own [fork](https://github.com/pytorch/torchtitan/blob/main/README.md?plain=1#L23) to host special optimizations. We try to organize the codebase to support extensibility (while not losing too much readbiility) so that modification on top is simple.
- On the other hand, we recommend shipping critical features to pytorch before they are integrated in torchtitan.
 
Would love to hear your thoughts on this.

vivekgoe: @tianyu-l Thanks for detailed review comments. I agree It makes sense to separate out adopting low-precision optimizer states from creating TE dependency. 

Regarding your question "Could you show evidences that this is becoming the default for training?", as far as I know DeepSeek pioneered this approach and I have not seen this used elsewhere, perhaps because support for this is not available in off the shelf optimizers. 
I did a quick check on Qwen3-32B and LLama4-17Bx16e to confirm if models other than DeepSeek maintain quality (along with smaller memory footprint) with this feature, results look promising (see plots below). I can create a issue in pytorch repo to check if there is interest in adding this feature to torch.optim.adamw directly.      
Code_Generated_Image (2)
Code_Generated_Image (3)

Regarding adding TE dependency, you make very valid arguments. As a short to medium term solution, will it be ok to move TE dependency to a new folder within experiment area so that users who wish to benefit from this feature can use it?  Longer term we can add feature to native pytorch optimizer (assuming maintainers agree) and remove it from experiment area.

tianyu-l: > As a short to medium term solution, will it be ok to move TE dependency to a new folder within experiment area so that users who wish to benefit from this feature can use it? Longer term we can add feature to native pytorch optimizer (assuming maintainers agree) and remove it from experiment area.

The long-term plan sounds good to me.

For the short-term plan, the feedback we heard is that the current experiments folder is too messy, and we plan to clean things up (by deleting as much as possible). That doesn't mean we won't consider TE as an experiment, but we would like to see a clear vision and integration/maintenance plan, so that we can discuss and review together if it makes sense to the community. If it's just a feature like this fused adam optimizer, maybe showcasing it in a PR is good enough.

Hope it is not nonsense to you.

janeyx99:
I had looked into adding a low prec AdamW into pytorch/pytorch last year, see https://github.com/pytorch/pytorch/issues/146542#issuecomment-2836583889 but ended up giving up because of lack of interest/no discussion for what a good frontend would look like. I'm noticing that in this PR the frontend is to set kwargs specifying the dtype for each state; I would want a more encompassing/scalable design that'd let us unlock low precision for more optimizers. I'm happy to kickstart this discussion again if we have users/people interested in co-designing a reasonable frontend and bring the PR to completion!

vivekgoe:
> I had looked into adding a low prec AdamW into pytorch/pytorch last year, see [pytorch/pytorch#146542 (comment)](https://github.com/pytorch/pytorch/issues/146542#issuecomment-2836583889) but ended up giving up because of lack of interest/no discussion for what a good frontend would look like. I'm noticing that in this PR the frontend is to set kwargs specifying the dtype for each state; I would want a more encompassing/scalable design that'd let us unlock low precision for more optimizers. I'm happy to kickstart this discussion again if we have users/people interested in co-designing a reasonable frontend and bring the PR to completion!

@janeyx99 Thanks for your detailed response. I am interested in co-designing a reasonable frontend and bring the PR you started to completion. My suggestion is I start with the changes you already have and test these with a medium size TorchTitan model like Qwen3-32B to check for convergence (I have proven it with a different implementation https://github.com/pytorch/torchtitan/pull/2293#discussion_r2762733494 so I have a reference to compare against). If these changes work then we can work together on defining a good frontend, cleaning-up code and adding other bells and whistles. 
Please let me know if this sounds good to you, I can start once you confirm. Shall I add a comment here as well https://github.com/pytorch/pytorch/issues/146542.

janeyx99:
@vivekgoe yes sounds great! We may also want to test on larger models, but the 32B is a good starting point. and yes, feel free to comment on the issue!

Issue #2273, opened 2026-01-24
[DeepEP] How is _handle_cache handled during inference?


goldhuang:
When I search `_handle_cache` in the code base, I only found the lines below. Seems the cache is not properly released if I do inference in between training steps, like inference to get validation loss etc., because `_handle_cache.pop(ctx.cache_id_int, None)` is only in dispatch_bwd.

```
_handle_cache: dict = {}
_cache_counter: int = 0
 
 
 
_handle_cache[cache_id.item()] = handle
 
handle = _handle_cache.get(cache_id.item())
 
handle = _handle_cache.get(ctx.cache_id_int)
 
_handle_cache.pop(ctx.cache_id_int, None)
 
ctx.saved_handle = _handle_cache.get(ctx.cache_id_int)

```
@elfiegg

tianyu-l:
cc @elfiegg @shuhuayu

elfiegg:
This is by design - cache_id is to save the `handle` returned by forward dispatch API call, and this `handle` will be used for the following combine and both processes' backward. So the life cycle would be:
```
dispatch generates handle and cache_id -> saves id to ctx.id -> combine retrieves handle from cache -> saves handle to ctx.handle.
dispatch backward retrieves handle via cache.get(ctx.id) -> release cache
combine backward retrieves handle via ctx.handle
```
Since we need to access handle across ops (dispatch and combine), as well as in both ops' backwards - we can't clean up the cache immediately after dispatch, but it's possible to save handle in dispatch_setup_context and clean it up after combine_setup_context. So the flow could potentially be:
```
dispatch generates handle and cache_id -> saves id to ctx.id, also saves handle to ctx.handle -> combine retrieves handle from cache -> saves handle to ctx.handle, and release the cache.
dispatch backward retrieves handle via ctx.handle
combine backward retrieves handle via ctx.handle
```
Do you run into any issue with the current implementation?

References:
1. DeepEP implementation: https://github.com/deepseek-ai/DeepEP?tab=readme-ov-file#example-use-in-model-training-or-inference-prefilling
2. HybridEP deisgn doc, "Handle Cache Strategy": https://docs.google.com/document/d/1i8zlu-3S2psDztKc9hHuyRAu2tsskDRBmMl3egDo7Vo/edit?tab=t.0#heading=h.t9wg1x5hf31w

goldhuang:
[...]

PR #2296, opened 2026-01-29
[MoE] DeepEP refactor and fix memory leak during training and inference

shuhuayu:
1. Simplified the token permutation logic.
2. Updated the handle management so there will be no memory leak during training and inference. Related issue: https://github.com/pytorch/torchtitan/issues/2273

On training on 16b deepseek v3 model, before the fix there was a growing memory usage. 
image
After the fix, the memory usage stabilizes. 
image

### DIFF
--- a/torchtitan/distributed/deepep/deepep.py
+++ b/torchtitan/distributed/deepep/deepep.py
@@ -331,6 +331,7 @@ def _unpermute_tokens(
 @dataclass
 class DispatchState:
     """State from dispatch needed for combine."""
+
     handle_id: torch.Tensor  # CPU tensor used to retrieve cached handle
     permuted_indices: torch.Tensor
     num_recv_tokens: int


tianyu-l:
@elfiegg please take a look

elfiegg:
Looks good, is my understanding correct that this PR mainly implements:
1. use `torch.argsort` for sorting the tokens for performance reasons
2. save `handle` in `dispatch`'s `setup_context`, and clear handle cache in `combine` forward if inference mode else clear it in setup_context

elfiegg:
FYI @goldhuang to unblock your work

shuhuayu:
> Looks good, is my understanding correct that this PR mainly implements:
> 
> 1. use `torch.argsort` for sorting the tokens for performance reasons
> 2. save `handle` in `dispatch`'s `setup_context`, and clear handle cache in `combine` forward if inference mode else clear it in setup_context

Yes. During training, handles are saved in both `dispatch_ctx` and `combine_ctx`, and the handle in `_handle_cache` is cleared in `combine_setup_context`. During inference, no handles saved in `op_ctx`, and `_handle_cache` is cleared during `combine_forward`.

elfiegg:
Sounds good, logic makes totally sense to me. Good to know inference mode `setup_context` won't be called

goldhuang:
@shuhuayu With the changes in the PR, can I train without selective activation checkpointing on deepep? The extra memory cost from SAC is huge. It may not be a good idea to enable SAC in some cases.

elfiegg:
@goldhuang You can configure `--activation_checkpoint.mode=full` and ` --parallelism.expert_parallel_comm_backend=deepep` to turn on full checkpointing with deepep being all-to-all backend

goldhuang:
@shuhuayu @elfiegg My point is that deepep integration in current main branch also has memory leak when you use it without SAC (meaning it's running in the pattern of `fwd-fwd-bwd`). You may want to make sure your changes also cover the `fwd-fwd-bwd` case.
The current main branch can only do fwd-bwd without a leak. Both `fwd` and `fwd-fwd-bwd` will cause a leak. I only reported the `fwd` case earlier.

shuhuayu:
> @shuhuayu With the changes in the PR, can I train without selective activation checkpointing on deepep? The extra memory cost from SAC is huge. It may not be a good idea to enable SAC in some cases.

@goldhuang Thanks for the question.

1. You can use full AC to save memory. 
2. Currently,  during training, `dispatch_op.ctx` and `combine_op.ctx` save layout medata, i.e., the `handle` regardless AC configuration, which should be not big. 
3. If we use `selective` and `op` in AC, deepep ops are included in the `op_sac_save_list`, so their activations get saved. We can remove them from `op_sac_save_list` by commenting out these two lines: https://github.com/pytorch/torchtitan/blob/cee9482989a754cb00ead1427a7d156d9c1b9441/torchtitan/models/llama4/infra/parallelize.py#L132-L133

I think 1 saves most memory and 3 saves memory specifically from deepep communications. 

In my test of a deepseek 16b model on 16 h100s (seqlen=4096, bsz=4, fsdp=ep=8, pp=2, attention=sdpa, compile=loss, moe_communication=deepep, no moe force load balance), the results are: 
1. selective ac with op: mfu 11.46%,  memory 27.5%.
2. selective ac with op, but recompute deepep ops (bullet 3 above): mfu 11.5%, memory 27.38%
3. full ac: mfu 12.8%, memory 24.5%.

So in these tests the savings from excluding deepep ops from sac save list are not significant in small scale.

shuhuayu:
> @shuhuayu @elfiegg My point is that deepep integration in current main branch also has memory leak when you use it without SAC (meaning it's running in the pattern of `fwd-fwd-bwd`). You may want to make sure your changes also cover the `fwd-fwd-bwd` case. The current main branch can only do fwd-bwd without a leak. Both `fwd` and `fwd-fwd-bwd` will cause a leak. I only reported the `fwd` case earlier.

@goldhuang Thanks for pointing this out. 
1. `fwd` case should be fixed if you use `torch.inference_mode()`.
2. `fwd-fwd-bwd` case like in full ac is now covered in this pr and tested by running full ac test.

PR #2299, opened 2026-01-29
[rl] GQA attention enablement in torchtitan vllm wrapper

wwwjn:
* #2300
* __->__ #2299

Following https://github.com/pytorch/torchtitan/pull/2259

Context: Fix RL attention implementation, add numerics check scripts comparing torchtitan Qwen3 model (with vllm wrapper) and vllm native Qwen3 model

NOTE: I tried enabling CI for this work, but need more efforts install vllm from source code (need to build vllm from scratch). Will enable CI in later PRs

### DIFF
--- a/torchtitan/experiments/rl/unified/models/utils.py
+++ b/torchtitan/experiments/rl/unified/models/utils.py
@@ -62,11 +62,10 @@ def replace_with_vllm_attention(model, tp_degree=1):
         # Number of KV heads is greater than TP size, so we partition
         # the KV heads across multiple tensor parallel GPUs.
         assert total_num_kv_heads % tp_degree == 0
+        num_kv_heads = total_num_kv_heads // tp_degree
     else:
-        # Number of KV heads is less than TP size, so we replicate
-        # the KV heads across multiple tensor parallel GPUs.
-        assert tp_degree % total_num_kv_heads == 0
-    num_kv_heads = max(1, total_num_kv_heads // tp_degree)
+        # TODO: Handle this branch correctly
+        raise ValueError("num_kv_heads are smaller than tp_degree")
 
     for layer_name, layer in model.layers.items():
         if not hasattr(layer, "attention"):


wwwjn:
retry merge #2291, the code is the same as #2291

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -54,15 +54,29 @@ def replace_with_vllm_attention(model, tp_degree=1):
...
+
+    # Reference: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/qwen3.py#L80
+    # Calculate num_kv_heads based on TP size
+    total_num_kv_heads = model_args.n_kv_heads
+    if total_num_kv_heads >= tp_degree:
+        # Number of KV heads is greater than TP size, so we partition
+        # the KV heads across multiple tensor parallel GPUs.
+        assert total_num_kv_heads % tp_degree == 0
+    else:
+        # Number of KV heads is less than TP size, so we replicate
```
acisseJZhong: how did we replicate kv heads across multiple TP ranks? curious why would we passed in num_kv_heads=1 to VLLMAttnention when total_num_kv_heads < tp_degree?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -238,9 +238,10 @@ def forward(
         for layer in self.model.layers.values():
             h = layer(h, rope_cache, attention_masks=None, positions=positions)
 
+        h = self.model.norm(h)
```
acisseJZhong, line 241: so norm is not included in `self.model` and we need to manually apply outside?

wwwjn: We are manually calling forward for each module within self.model. Previously the norm layer is in `compute_logits`, but now I moved it to `forward` function similiar to vllm's native qwen3 implementation. From the numerics test, I don't see these change affecting final results

torchtitan/experiments/rl/unified/models/utils.py
wwwjn: Nice catch, I borrowed this part from vllm qwen3 implementation, and this "replicate" is not accurate for our case. if the number of kv heads are smaller than tp_degree (which is rare case), we haven't tested out, I will just error out for now

tianyu-l (approved):

PR #2300, opened 2026-01-29
[rl] Add CI for RL integration tests

wwwjn:
* __->__ #2300

Test cases:
1. Integration tests: 
   - multiple GPU (with TP),  no compile + cudagraph
   - multiple GPU, with compile + cudagraph
   - This test runs on A10G (default CI GPU type),  and **uses FA2 for attention part as FA3 doesn't support A10G machine.**
2. ~Numerics parity test: vLLM native model vs vLLM + TorchTitan wrapper.~
    - test_weights_match:          max_diff <= 1e-5 (exact weight loading)
    - test_attention_module:       atol=1e-5 (TP=1)
    - test_end_to_end_logits:      atol=1e-3 (TP=1)
    - ~We would need to run numerics test for only TP=1. This is because we are assuming both torchtitan and vllm will make sure their multi-GPU implementation is on par with single GPU. And we can add more numerics test under parallelism if needed.~
    - ~This test runs on H100, and runs FA3 kernel for attention.~

Both tests are downloading from huggingface and run test on qwen3 0.6B.

### DIFF
--- a/torchtitan/experiments/rl/models/attention.py
+++ b/torchtitan/experiments/rl/models/attention.py
@@ -7,7 +7,10 @@
 import logging
 
 import torch
-from torch.nn.attention import activate_flash_attention_impl
+from torch.nn.attention import (
+    activate_flash_attention_impl,
+    current_flash_attention_impl,
+)
 from torchtitan.models.common.attention import LocalMapAttention
 from torchtitan.tools.utils import has_cuda_capability
 
@@ -59,7 +62,8 @@ class PyTorchFlashAttentionImpl(FlashAttentionImpl):
         # activate_flash_attention_impl("FA3") succeeds even on SM80.
         # Fall back to FA2 which requires page_size to be a multiple of 256.
         if has_cuda_capability(9, 0):
-            activate_flash_attention_impl("FA3")
+            if current_flash_attention_impl() != "FA3":
+                activate_flash_attention_impl("FA3")
             self._use_fa3 = True
         else:
             logger.warning(


torchtitan/experiments/rl/tests/test_numerics.py
tianyu-l, line 1: If we are not guarding on it, I don't mind us keeping it in a PR / branch / internal so people could refer to easily.

wwwjn: Sure, use this PR as reference and close the PR

.github/workflows/integration_test_8gpu_rl.yaml
```diff
@@ -0,0 +1,75 @@
...
+        # Install vllm with nightly torch
+        uv pip install torch vllm xformers --pre \
+          --extra-index-url https://download.pytorch.org/whl/nightly/cu128 \
+          --index-strategy unsafe-best-match
+
+        sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
+        sudo chown -R $(id -u):$(id -g) "$RUNNER_TEMP/artifacts-to-be-uploaded"
+
+        # Run the numerics tests for RL unified (test with TP=1 for exact numerics match)
+        torchrun --nproc-per-node=1 -m pytest torchtitan/experiments/rl/unified/tests/test_numerics.py -v
```
joecummings: It looks like you only use 1 GPU - why is this called an 8 GPU test?

torchtitan/experiments/rl/tests/test_numerics.py
joecummings, line 1: Maybe I'm missing something, but I see a lot of checks for distributed but I don't see where the model is actually parallelized or distributed in any way?

torchtitan/experiments/rl/unified/run_configs/qwen3_0.6b.toml
joecummings, line 1: Where is this used?

torchtitan/experiments/rl/tests/test_numerics.py
wwwjn: This is not a comprehensive test suite now, it only contains one test case now (which is checking if vllm native qwen3 inference has the same numerics with vllm+torchtitan model). I would hope to adding more CI later and we can continue adding tests in this file

torchtitan/experiments/rl/unified/run_configs/qwen3_0.6b.toml
wwwjn: Thanks, will remove once I rebase the PR

torchtitan/experiments/rl/unified/tests/test_numerics.py
```diff
@@ -0,0 +1,605 @@
...
+
+
+def get_distributed_info():
+    world_size = int(os.environ.get("WORLD_SIZE", 1))
+    rank = int(os.environ.get("RANK", 0))
+    local_rank = int(os.environ.get("LOCAL_RANK", 0))
+    return world_size, rank, local_rank
+
+
+def ensure_distributed_initialized():
```
acisseJZhong: I think you might no longer need this just use init_distributed_environment?

.github/workflows/integration_test_4gpu_rl.yaml
```diff
@@ -0,0 +1,79 @@
...
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.set-matrix.outputs.matrix) }}
+    with:
+      runner: ${{ matrix.runner }}
+      gpu-arch-type: ${{ matrix.gpu-arch-type }}
+      gpu-arch-version: ${{ matrix.gpu-arch-version }}
+      docker-image: ${{ matrix.docker-image }}
+      repository: pytorch/torchtitan
+      upload-artifact: outputs
+      timeout: 45
```
felipemello1, line 48: noob q: is this to 45s to run the script?

torchtitan/experiments/rl/tests/integration_tests.py
```diff
@@ -0,0 +1,161 @@
...
+            raise Exception(
+                f"RL integration test failed: {test_flavor.test_descr}, command: {cmd}"
+            )
+
+
+def run_tests(args, test_list: list[OverrideDefinitions]) -> None:
+    ran_any = False
+    for test_flavor in test_list:
+        if args.test_name != "all" and test_flavor.test_name != args.test_name:
+            continue
```
felipemello1, line 108: maybe add a warning or raise? why would someone select a test_flavor that is not in a test_list?

torchtitan/experiments/rl/unified/tests/test_numerics.py
```diff
@@ -0,0 +1,605 @@
...
+    global _distributed_initialized
+    world_size, rank, local_rank = get_distributed_info()
+
+    if model_parallel_is_initialized() or _distributed_initialized:
+        return
```
felipemello1: nit: 
1. this sounds like a bit of overkill. Why do we need 2 ways of checking it? 
2. Why do we need to check in the first place? do we try to initialize in multiple places?

torchtitan/experiments/rl/unified/tests/test_numerics.py
```diff
@@ -0,0 +1,605 @@
...
+    """Extract model flavor from checkpoint path (e.g., 'Qwen/Qwen3-0.6B' → '0.6B').
+
+    Falls back to '0.6B' if the flavor cannot be determined.
+    """
+    basename = os.path.basename(checkpoint_path.rstrip("/"))
+    if "Qwen3" in basename:
+        parts = basename.split("-", 1)
+        if len(parts) == 2:
+            return parts[1]
+    return "0.6B"
```
felipemello1: if i pass anything that is not "qwen-XX', it will return 0.6b, e.g. 'llama_3b' would return '0.6b'

torchtitan/experiments/rl/tests/test_numerics.py
```diff
@@ -0,0 +1,605 @@
...
+        from huggingface_hub import snapshot_download
+
+        model_checkpoint_path = snapshot_download(model_checkpoint_path)
+
+    # Create vLLM config first (no distributed initialization needed)
+    engine_args = EngineArgs(
+        model=model_checkpoint_path,
+        skip_tokenizer_init=True,
+        enforce_eager=True,
+        gpu_memory_utilization=0.1,
```
felipemello1: this seems rather low

torchtitan/experiments/rl/tests/test_numerics.py
```diff
@@ -0,0 +1,605 @@
...
+
+    # Create ModelSpec from registry, matching simple_grpo.py
+    model_flavor = get_model_flavor(model_checkpoint_path)
+    model_spec = model_registry(model_flavor)
+    # Patch to use RL-specific parallelize function (same as simple_grpo.py)
+    model_spec.parallelize_fn = parallelize_qwen3
+
+    # vLLM config must be set before distributed init (initialize_model_parallel
+    # calls get_current_vllm_config), model loading, and wrapper creation.
+    with set_current_vllm_config(vllm_config):
```
felipemello1: i understand that this is an imported function, but i have no clue why this needs to be a ctx manager

torchtitan/experiments/rl/tests/test_numerics.py
felipemello1, line 1: i found it a bit hard to understand this file. It feel like its doing much more than testing. There are many opinionated functions changing internals, almost like it should live in the library.

There is also a mix of functions that generalize tests + hardcode for qwen.

I wonder if it is worth trying to simplify and rewrite this. Maybe try claude `/simplify`?

wwwjn: This file guards vllm inference torchtitan's qwen3 model has the same numerics ( < 1e-3 difference) with vllm native qwen3 model. Yes this is thighed to Qwen3 model, and we are not able to generalize this numerics guard to other model. I want to discuss if we need to guard the numerics parity between our model and vllm's

Working on simplify it

wwwjn: > the model is actually parallelized or distributed in any way

Update: ~We setup distritbuted to enable this test case for TP>1 as well. The model is parallelized with in `TorchtitanVLLMWrapper.__init__`. If we passed vllm_config (with tensor_parallelisim_degree=2 in vllm-config) to TorchtitanVLLMWrapper, it will apply parallelization and DTensor-ify the model with in `TorchtitanVLLMWrapper.__init__`~

I removed TP=2 tests

wwwjn: I updated the main frame of test_numerics.py. The whole purpose for this file is to guard we have numerics parity with vllm native Qwen3 model, in case any model changes, or refactor in torchtitan breaks this gurantee. 

Currently it's tighted to Qwen3 in following ways (by claude):
- _get_model_flavor (line 67-74) — parses "Qwen3-0.6B" to extract "0.6B". It falls back to "0.6B" for unrecognized names, so a non-Qwen3 model would silently use the wrong flavor.
- model_registry(_get_model_flavor(checkpoint)) (line 199) — uses torchtitan's Qwen3 model registry. For a different model family (e.g., Llama3), you'd need to use that model's registry instead.
- model_spec.parallelize_fn = parallelize_qwen3 (line 200) — hardcoded to Qwen3's parallelization function.

But they are not fundamentally tied to Qwen3, it's much cleaner than the previous version, please take another look, thanks

.github/workflows/integration_test_4gpu_rl.yaml
wwwjn: This is 45 mins time out for this test

torchtitan/experiments/rl/tests/integration_tests.py
wwwjn: Here we keeps align with other integration test: we ignore if the name is wrong. The user can notice the test is not running because no log will be printed in this case

torchtitan/experiments/rl/tests/test_numerics.py
wwwjn: This  set_current_vllm_config() function is from vllm, my understanding is that they want to share some metadata / configs across a lot of places

torchtitan/experiments/rl/tests/test_numerics.py
```diff
@@ -0,0 +1,370 @@
...
+        "vllm_config": vllm_config,
+        "tokens": tokens,
+        "positions": positions,
+        "attn_meta": attn_meta,
+        "dim": hf_cfg.hidden_size,
+        "num_kv_heads": num_kv_heads,
+        "head_dim": head_dim,
+        "seq_len": seq_len,
+        "device": device,
+    }
```
felipemello1: nit (you can ignore): i think that some names could be a bit more obvious, like 'native', 'tt' and 'attn_meta'. They are not intuitive unless i check what they do

torchtitan/experiments/rl/tests/integration_tests.py
```diff
@@ -0,0 +1,161 @@
...
+        logger.info(
+            f"===== {time.strftime('%Y-%m-%d %H:%M:%S')} "
+            f"RL integration test: {test_flavor.test_descr}, command: {cmd} ====="
+        )
+
+        result = subprocess.run(cmd, text=True, shell=True)
+        if result.returncode != 0:
+            raise Exception(
+                f"RL integration test failed: {test_flavor.test_descr}, command: {cmd}"
+            )
```
felipemello1, line 101: what is being tested? just if the recipe runs? Should we add any numerics of checks if we get the expected number of steps or if rewards go up?

.github/workflows/integration_test_4gpu_rl.yaml
```diff
@@ -0,0 +1,83 @@
...
+        sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
+        sudo chown -R $(id -u):$(id -g) "$RUNNER_TEMP/artifacts-to-be-uploaded"
+
+        # Run numerics unit tests (TP=1)
+        torchrun --nproc-per-node=1 -m pytest torchtitan/experiments/rl/unified/tests/test_numerics.py -v
+        rm -rf $RUNNER_TEMP/artifacts-to-be-uploaded/*/checkpoint
+
+        # Run E2E RL integration tests (TP=1 on 2 GPUs, TP=2 on 4 GPUs)
+        python -m torchtitan.experiments.rl.unified.tests.integration_tests \
+          $RUNNER_TEMP/artifacts-to-be-uploaded --ngpu 4
```
felipemello1: noob q: should we find a way to parallelize this instead of running sequentially? I am unsure about best practices here, i.e. do we have a bunch of gpus idle while one of the tests is running?

torchtitan/experiments/rl/tests/integration_tests.py
wwwjn: Yes, this is just testing if the recipe subprocess finishes successfully without crash. The numerics check / expected behavior change usually sit within each test, we can do this by adding `assert`

torchtitan/experiments/rl/tests/test_numerics.py
wwwjn: Will update this with more descriptive names

.github/workflows/integration_test_4gpu_rl.yaml
wwwjn: > should we find a way to parallelize this instead of running sequentially 

I guess we can do this by adding a separate yaml file, but within 1 file, I would assume it is executed sequentially

wwwjn:
Will rebase this PR after https://github.com/pytorch/torchtitan/pull/2364#pullrequestreview-3977651149 lands

tianyu-l:
Can we use the H100 CI resource to test FAv3? We can consider creating a separate yaml for it, if piggybacking on existing one doesn't make sense.

wwwjn:
> Can we use the H100 CI resource to test FAv3? We can consider creating a separate yaml for it, if piggybacking on existing one doesn't make sense.

If we baked FA3 into our RL model, we can not run all the integration tests on H100 because of long wait time and queue.

I would propose the following:
1. For integration test, we fallback to FA2 in PyTorchFlashAttentionImpl to make CI runnable. I talked with @liangel-02 and the FA2 should work because the parameters are the same as FA3, and we just need to set vllm's page size to be 256. I'm trying this path.
2. For unit test, we set a separate unit test for attention part on H100 (all other unit test can run on A20), checking our PyTorchFlashAttentionImpl forward pass numerics agains vllm's Flash attention numerics.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -186,9 +186,14 @@ def __init__(
         self.top_p = config.sampling.top_p
         self.num_samples_per_prompt = config.num_samples_per_prompt
 
+        # FA2 requires block_size divisible by 256; FA3 (SM 9.0+) supports any
+        _sm_90 = torch.cuda.get_device_capability()[0] >= 9
+        block_size = 32 if _sm_90 else 256
```
acisseJZhong: qq: why it's 32 instead of vllm default 16?

torchtitan/experiments/rl/README.md
```diff
@@ -15,6 +15,7 @@ The integration consists of the following components:
 
 0. Create and activate environment with uv:
 ```bash
+pip install uv
```
acisseJZhong, line 18: do you need this? running anything start with pip gives proxy error on my devgpu 😅

torchtitan/experiments/rl/actors/generator.py
wwwjn: oh nice catch, let me reset it to 16

torchtitan/experiments/rl/README.md
wwwjn: > start with pip gives proxy error on my devgpu 

This is a specific restriction for our devgpu, but we would assume the README is for general users, which I assume they already under proper working enviroment. I add this line because next line directly use uv which might not be installed yet

acisseJZhong: sounds good, somehow on devgpu i don't have to install uv. But good to have it for general users.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -35,6 +36,20 @@
 logger = logging.getLogger(__name__)
 
 
+def _log_gpu_memory(label: str) -> None:
```
tianyu-l: do we need this?

torchtitan/experiments/rl/tests/test_bitwise_identity.py
tianyu-l, line 1: is this tested?

.github/workflows/integration_test_1gpu_h100_rl.yaml
tianyu-l, line 1: why having two tests? can we run 2 gpu h100 for everything?

torchtitan/experiments/rl/tests/test_bitwise_identity.py
wwwjn: This is from PR #2364, and I rename it and disabled it in this PR. We discussed with @zhxchen17  that the bit-wise identity mode is not supported yet, and this testing is always failing. I disable this test in this PR. We could remove this file entirely

torchtitan/experiments/rl/actors/trainer.py
wwwjn: Should be removed

.github/workflows/integration_test_1gpu_h100_rl.yaml
wwwjn: H100 usually have longer wait time because of capacity, and I'm expecting to add more models and test case cover in RL CI later. So I followed this design here: https://github.com/pytorch/torchtitan/pull/2300#issuecomment-4112576755 , only put necessary tests on H100 while getting quick signal on A10G.

wwwjn:
Update: remove numerics test in this PR. 

Will adding more rigorous numerics test along with deterministic mode in next PRs. Discussed with @tianyu-l  offline, we should aim for adding the numerics using the same pattern with pre-training side CI:
1. Locally check our model numerics matches vllm.
2. Enable deterministic mode. Run with specific task and getting training metrics , eg reward , loss. Record these metrics in a file.
3. Run CI to compare against the loss in the file.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -57,12 +54,18 @@ class PyTorchFlashAttentionImpl(FlashAttentionImpl):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
-        # Activate PyTorch's FA3 flash attention implementation. This is required for
-        # the varlen attention API to work with paged attention (KV cache in block
-        # layout) because vLLM uses default page size 16 but FA2 only supports
-        # page_size=256. Error out if FA3 is not available in the current environment.
-        if current_flash_attention_impl() != "FA3":
```
tianyu-l, line 64: why deleting this line? I'm not sure if there's a cost associated with this line. cc @liangel-02 @zhxchen17

torchtitan/experiments/rl/models/attention.py
```diff
@@ -57,12 +54,18 @@ class PyTorchFlashAttentionImpl(FlashAttentionImpl):
...
+        # FA3 requires SM 9.0+ (e.g. H100); check capability explicitly because
+        # activate_flash_attention_impl("FA3") succeeds even on SM80.
+        # Fall back to FA2 which requires page_size to be a multiple of 256.
+        if torch.cuda.get_device_capability()[0] >= 9:
```
tianyu-l: use https://github.com/pytorch/torchtitan/blob/main/torchtitan/tools/utils.py#L21

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -186,9 +192,14 @@ def __init__(
         self.top_p = config.sampling.top_p
         self.num_samples_per_prompt = config.num_samples_per_prompt
 
+        # FA2 requires block_size divisible by 256; FA3 (SM 9.0+) supports any
+        _sm_90 = torch.cuda.get_device_capability()[0] >= 9
+        block_size = 16 if _sm_90 else 256
```
tianyu-l: Why default to 16 if _sm_90? This can change previous behavior. Can we not set it until necessary?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -186,9 +192,14 @@ def __init__(
         self.top_p = config.sampling.top_p
         self.num_samples_per_prompt = config.num_samples_per_prompt
 
+        # FA2 requires block_size divisible by 256; FA3 (SM 9.0+) supports any
+        _sm_90 = torch.cuda.get_device_capability()[0] >= 9
```
tianyu-l: same https://github.com/pytorch/torchtitan/blob/main/torchtitan/tools/utils.py#L21

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -145,6 +145,12 @@ class Config(Configurable.Config):
         num_samples_per_prompt: int = 8
         """Number of completions to generate per prompt."""
 
+        max_model_len: int | None = None
```
tianyu-l: why do we need it, in this PR?

torchtitan/experiments/rl/actors/generator.py
tianyu-l: have we removed `"FLASH_ATTN"`? If so, can we remove this field in general, until flex is supported?

torchtitan/experiments/rl/models/attention.py
wwwjn: Previously the check is saying if current attention is not FA3, then we should activate FA3. However on A20G machine, this should not be true as we want to use FA2 instead of FA3. 

So I changed the check to be if the machine has SM 9.0+, then activate FA3

torchtitan/experiments/rl/actors/generator.py
wwwjn: This field is a parameter directly sending to vllmEngine
"FLASH_ATTENTION" is vllm's native flash attention
"CUSTOM" is pytorch's varlen attention

Now I changed to hardcode it as "CUSTOM"

torchtitan/experiments/rl/models/attention.py
tianyu-l: Oh that's not what I was asking. I'm asking why don't we do
```
if has_cuda_capability(9, 0):
  if current_flash_attention_impl() != "FA3":
    activate_flash_attention_impl("FA3")
```
why do we omit `if current_flash_attention_impl() != "FA3":` now? If it is not helping, why we have it before?

tianyu-l (approved):

torchtitan/experiments/rl/models/attention.py
wwwjn: You are right, it will just activate FA3 once, not every layer's attention. Changed

PR #2301, opened 2026-01-30
[BE][NFC] Add integration test for simplefsdp + CP deepseek_v3

aditvenk:
This combination is currently not tested. CP is only tested with LLama3

Testing: Manually tested that the combination works.


aditvenk:
CI may be red  till https://github.com/pytorch/pytorch/pull/173790 is merged.

fegin (approved):

PR #2305, opened 2026-01-30
[Typing] Fix pyrefly ignores in deepseek model.py

fegin:
* __->__ #2305
* #2304
* #2303
* #2302

As title

### DIFF
--- a/torchtitan/models/deepseek_v3/model/model.py
+++ b/torchtitan/models/deepseek_v3/model/model.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import math
+from typing import cast
 
 import torch
 from torch import nn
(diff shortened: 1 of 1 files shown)


wwwjn (approved):

torchtitan/models/deepseek_v3/model/model.py
```diff
@@ -398,8 +399,7 @@ def init_weights(self, buffer_device: torch.device):
             norm.reset_parameters()
         self.attention.init_weights(self.weight_init_std)
         if self.moe_enabled:
-            # pyrefly: ignore [not-callable, missing-attribute]
-            self.moe.init_weights(self.weight_init_std, buffer_device)
+            cast(MoE, self.moe).init_weights(self.weight_init_std, buffer_device)
```
Skylion007, line 402: Do we have to cast or can we do an assert isinstance check utility?

Skylion007 (approved):

torchtitan/models/deepseek_v3/model/model.py
fegin: yes, that's a good suggestion. We should not only cast but also assert and make this a util.

PR #2306, opened 2026-01-31
[docs] Fix type mismatch in model layers comments

DamonFool:
Please review this tiny fix to improve the code comments.
Thanks.


DamonFool:
Thanks @tianyu-l for your review.

The two failing checks shouldn't be caused by this change, right?

PR #2307, opened 2026-01-31
Add peak flops for NVIDIA H20 GPUs

DamonFool:
This PR adds the peak BF16/FP16 performance metrics for the NVIDIA H20 GPU to the `get_peak_flops` utility.
This ensures that Model FLOPs Utilization (MFU) can be accurately calculated when training on H20-based clusters.
Thanks.


DamonFool:
Thanks @tianyu-l for the review.

I pushed one more change to fix the lint check failure due to the long website link.
Thanks.

PR #2309, opened 2026-02-02
Fix FLUX attention by exposing is_causal in SDPA

### LINKED ISSUE #2297: Flux attention is_causal need be False + Flex Attention support with is_causal=False
https://github.com/pytorch/torchtitan/blob/374ea0f37621669918e924771c8498602c6d9235/torchtitan/models/flux/model/layers.py#L142
ScaledDotProductAttentionWrapper always set is_causal=True。 Doesn't Flux should be False？

wwwjn:
fix #2297 

the official implementation from black-forest-lab (https://github.com/black-forest-labs/flux/blob/main/src/flux/math.py#L6) they are using `torch.nn.functional.scaled_dot_product_attention` with default is_casual = False

### DIFF
--- a/torchtitan/models/attention.py
+++ b/torchtitan/models/attention.py
@@ -185,10 +185,11 @@ class ScaledDotProductAttentionWrapper(torch.nn.Module):
         *,
         scale: float | None = None,
         enable_gqa: bool = False,
+        is_casual: bool = True,
     ) -> torch.Tensor:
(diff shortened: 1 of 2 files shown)


wwwjn:
8 GPU Feature test failed because of loss mismatch
8 GPU model test failed because of OOM issue of "DeepSeek V3 FSDP+EP+compile", both failure are not related

PR #2311, opened 2026-02-02
separate out training for fault tolerance

tushar00jain:
Summary:
- extract logic in train.py that was different for ft in separate functions
- override these functions in ft's train.py

Test Plan:
Run lighthouse and two replicas

```bash
RUST_BACKTRACE=1 torchft_lighthouse --min_replicas 2 --quorum_tick_ms 100 --join_timeout_ms 10000

TRAIN_FILE=torchtitan.experiments.ft.train CONFIG_FILE=./torchtitan/models/llama3_ft/train_configs/debug_model.toml CUDA_VISIBLE_DEVICES=0 NGPU=1 ./run_train.sh --parallelism.data_parallel_shard_degree=1 --fault_tolerance.enable --fault_tolerance.group_size=2 --fault_tolerance.replica_id=0


TRAIN_FILE=torchtitan.experiments.ft.train CONFIG_FILE=./torchtitan/models/llama3_ft/train_configs/debug_model.toml CUDA_VISIBLE_DEVICES=1 NGPU=1 ./run_train.sh --parallelism.data_parallel_shard_degree=1 --fault_tolerance.enable --fault_tolerance.group_size=2 --fault_tolerance.replica_id=1
```

### DIFF
--- a/torchtitan/experiments/ft/torchft.md
+++ b/torchtitan/experiments/ft/torchft.md
@@ -36,12 +36,12 @@ RUST_BACKTRACE=1 torchft_lighthouse --min_replicas 1 --quorum_tick_ms 100 --join
 2. Launch the first TorchTitan instance:
 
 ```bash
-NGPU=4 CUDA_VISIBLE_DEVICES=0,1,2,3 CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" ./run_train.sh --fault_tolerance.enable --fault_tolerance.replica_id=0 --fault_tolerance.group_size=2 --parallelism.data_parallel_shard_degree=4
+NGPU=4 CUDA_VISIBLE_DEVICES=0,1,2,3 TRAIN_FILE=torchtitan.experiments.ft.train CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" ./run_train.sh --fault_tolerance.enable --fault_tolerance.replica_id=0 --fault_tolerance.group_size=2 --parallelism.data_parallel_shard_degree=4
 ```
 3. Launch the second TorchTitan instance:
 
 ```bash
-NGPU=4 CUDA_VISIBLE_DEVICES=4,5,6,7 CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" ./run_train.sh --fault_tolerance.enable --fault_tolerance.replica_id=1 --fault_tolerance.group_size=2 --parallelism.data_parallel_shard_degree=4
+NGPU=4 CUDA_VISIBLE_DEVICES=4,5,6,7 TRAIN_FILE=torchtitan.experiments.ft.train CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" ./run_train.sh --fault_tolerance.enable --fault_tolerance.replica_id=1 --fault_tolerance.group_size=2 --parallelism.data_parallel_shard_degree=4
 ```
 
 ### Explanation
@@ -68,12 +68,12 @@ The `--training.global_batch_size` parameter refers to global batch size that wi
 
 #### Replica Group 0
 ```bash
-CONFIG_FILE=./torchtitan/models/llama3_ft/train_configs/debug_model.toml CUDA_VISIBLE_DEVICES=0,1,2,3 NGPU=4 ./run_train.sh --parallelism.data_parallel_shard_degree=4 --fault_tolerance.enable --fault_tolerance.group_size=2 --fault_tolerance.replica_id=0
+TRAIN_FILE=torchtitan.experiments.ft.train CONFIG_FILE=./torchtitan/models/llama3_ft/train_configs/debug_model.toml CUDA_VISIBLE_DEVICES=0,1,2,3 NGPU=4 ./run_train.sh --parallelism.data_parallel_shard_degree=4 --fault_tolerance.enable --fault_tolerance.group_size=2 --fault_tolerance.replica_id=0
 ```
 
 #### Replica Group 1
 ```bash
-CONFIG_FILE=./torchtitan/models/llama3_ft/train_configs/debug_model.toml CUDA_VISIBLE_DEVICES=4,5,6,7 NGPU=4 ./run_train.sh --parallelism.data_parallel_shard_degree=4 --fault_tolerance.enable --fault_tolerance.group_size=2 --fault_tolerance.replica_id=1
+TRAIN_FILE=torchtitan.experiments.ft.train CONFIG_FILE=./torchtitan/models/llama3_ft/train_configs/debug_model.toml CUDA_VISIBLE_DEVICES=4,5,6,7 NGPU=4 ./run_train.sh --parallelism.data_parallel_shard_degree=4 --fault_tolerance.enable --fault_tolerance.group_size=2 --fault_tolerance.replica_id=1
 ```
 
 ## Fault Tolerance Configuration Options
--- a/torchtitan/experiments/ft/train.py
+++ b/torchtitan/experiments/ft/train.py
@@ -4,13 +4,335 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import dataclasses
+import importlib
+import json
 import os
+import time
+from datetime import timedelta
+from typing import cast, Iterator
 
+import torch
+import torch.distributed.checkpoint.stateful
+
+import torchtitan.protocols.train_spec as train_spec_module
+from torch.distributed.elastic.multiprocessing.errors import record
+from torchtitan.components.checkpoint import CheckpointManager
+from torchtitan.components.dataloader import DataloaderExhaustedError
+from torchtitan.components.ft import FTManager, maybe_semi_sync_training
+from torchtitan.components.loss import IGNORE_INDEX
+from torchtitan.components.metrics import (
+    build_metrics_processor,
+    ensure_pp_loss_visible,
+)
+from torchtitan.config import JobConfig, TORCH_DTYPE_MAP
 from torchtitan.distributed import ParallelDims, utils as dist_utils
+from torchtitan.protocols import ModelProtocol
+from torchtitan.protocols.model_converter import build_model_converters
+from torchtitan.tools import utils
+from torchtitan.tools.logging import logger
+from torchtitan.tools.profiling import (
+    maybe_enable_memory_snapshot,
+    maybe_enable_profiling,
+)
 from torchtitan.train import main, Trainer
 
 
 class FTTrainer(Trainer):
+    ft_manager: FTManager | None = None
+
+    # Enable debug tracing on failure: https://pytorch.org/docs/stable/elastic/errors.html
+    @record
+    def __init__(self, job_config: JobConfig):
+        torch._C._log_api_usage_once("torchtitan.train")
+
+        self.job_config = job_config
+
+        logger.info(f"Starting job: {job_config.job.description}")
+
+        if job_config.experimental.custom_import:
+            importlib.import_module(job_config.experimental.custom_import)
+
+        device_module, device_type = utils.device_module, utils.device_type
+        # pyrefly: ignore [read-only]
+        self.device = torch.device(f"{device_type}:{int(os.environ['LOCAL_RANK'])}")
+        # Device has to be set before creating TorchFT manager.
+        device_module.set_device(self.device)
+
+        # init distributed and build meshes
+        self.parallel_dims = parallel_dims = self.init_distributed()
+
+        # Logging needs to happen after distributed initialized
+        job_config.maybe_log()
+
+        if parallel_dims.dp_enabled:
+            batch_mesh = parallel_dims.get_mesh("batch")
+            batch_degree, batch_rank = batch_mesh.size(), batch_mesh.get_local_rank()
+        else:
+            batch_degree, batch_rank = 1, 0
+
+        self.ft_manager = FTManager(job_config.fault_tolerance)
+        batch_degree, batch_rank = self.ft_manager.get_dp_info(batch_degree, batch_rank)
+
+        # take control of garbage collection to avoid stragglers
+        self.gc_handler = utils.GarbageCollection(
+            gc_freq=job_config.training.gc_freq, debug=job_config.training.gc_debug
+        )
+
+        # Set random seed, and maybe enable deterministic mode
+        # (mainly for debugging, expect perf loss).
+        dist_utils.set_determinism(
+            parallel_dims,
+            self.device,
+            job_config.debug,
+            distinct_seed_mesh_dims=["pp"],
+        )
+        self.train_spec = train_spec_module.get_train_spec(job_config.model.name)
+
+        # build tokenizer and dataloader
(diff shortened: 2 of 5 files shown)


torchtitan/experiments/ft/torchft.md
wwwjn, line 1: Move this doc to `experiment/ft/README.py`?

torchtitan/experiments/ft/train.py
```diff
@@ -46,6 +63,74 @@ def init_distributed(self) -> ParallelDims:
...
+            ft_manager=self.setup_ft_manager(),
+        )
+
+    def build_optimizers(self):
+        return self.train_spec.build_optimizers_fn(
+            self.model_parts,
+            self.job_config.optimizer,
+            self.parallel_dims,
+            self.setup_ft_manager(),
+        )
```
fegin: These two methods make trainer layering become more complicated and verbose. We already have `build_optimizers_fn` and that's provided by the individual train_spec, including FT. So we shouldn't need another wrapper to wrap these two methods.

Please remove these two methods and change TorchFT train_spec to have the same signature as the original build_optimizers_fn. One way to do this is to have TorchFT specific train_spec per model, similar to other experiments, like simple_fsdp, https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/simple_fsdp/deepseek_v3/__init__.py.

You can have a specific build_optimizers_fn for FT and use that for each train_spec. You can make FTManager a singleton class so that FT's build_optimizers_fn and build_loss_fn and FTTrainer can all that class and share the instance.

torchtitan/experiments/ft/train.py
```diff
@@ -46,6 +63,74 @@ def init_distributed(self) -> ParallelDims:
...
+            sd_adapter=(
+                self.train_spec.state_dict_adapter(
+                    model_args, job_config.model.hf_assets_path
+                )
+                if self.train_spec.state_dict_adapter
+                else None
+            ),
+            base_folder=job_config.job.dump_folder,
+            ft_manager=self.setup_ft_manager(),
+        )
```
fegin: I think we should also consider to make this checkpointer as a builder in train_spec so that each experiment can overwrite. But for this PR, you don't need builder, you simply need to create another checkpointer and overwrite the original instance. So this method is also not needed.

torchtitan/experiments/ft/train.py
```diff
@@ -46,6 +63,74 @@ def init_distributed(self) -> ParallelDims:
...
+
+    def get_loss_sync_pg(self):
+        """Return the process group for loss synchronization."""
+        return self.setup_ft_manager().loss_sync_pg
```
fegin: Exposing this method doesn't make sense as this makes the main trainer less readable (worse than without migration). What does `get_loss_sync_pg` mean in the original trainer? However, I think we can make the `compute_global_losses` as a method and move the logic of computing losses here. This makes more sense in that `compute_global_losses` has enough meaning and has been changed sometimes. So IMO, making `compute_global_losses` can improve the readability and also can help this PR migrations.

torchtitan/experiments/ft/train.py
```diff
@@ -46,6 +63,74 @@ def init_distributed(self) -> ParallelDims:
...
+            n_layers=(
+                self.model_args.n_layers if hasattr(self.model_args, "n_layers") else 0
+            ),
+            optimizer=self.optimizers,
+            fragment_fn=(
+                self.train_spec.fragment_fn
+                if hasattr(self.train_spec, "fragment_fn")
+                else None
+            ),
+        )
```
fegin: Same as above. These two methods make no sense in the original trainer. Since the `FTTrainer.train()` is largely different from `Trainer.train()`, you should just overwrite `train()` similar to FluxTrainer.

torchtitan/train.py
```diff
@@ -189,9 +187,7 @@ def __init__(self, job_config: JobConfig):
             init_device = device_type
             buffer_device = None
 
-        self.loss_fn = self.train_spec.build_loss_fn(
-            job_config, parallel_dims=parallel_dims, ft_manager=self.ft_manager
-        )
+        self.loss_fn = self.build_loss_fn()
```
fegin: This should just be 
```
self.loss_fn = self.train_spec.build_loss_fn(job_config, parallel_dims=parallel_dims)
```

See the comment above.

torchtitan/train.py
```diff
@@ -275,9 +269,7 @@ def __init__(self, job_config: JobConfig):
         )
 
         # build optimizer after applying parallelisms to the model
-        self.optimizers = self.train_spec.build_optimizers_fn(
-            self.model_parts, job_config.optimizer, parallel_dims, self.ft_manager
-        )
```
fegin, line 280:
```
self.optimizers = self.train_spec.build_optimizers_fn(self.model_parts, job_config.optimizer, parallel_dims)
```
See the above comment.

torchtitan/train.py
```diff
@@ -297,23 +289,7 @@ def __init__(self, job_config: JobConfig):
...
-                self.train_spec.state_dict_adapter(
-                    model_args, job_config.model.hf_assets_path
-                )
-                if self.train_spec.state_dict_adapter
-                else None
-            ),
-            base_folder=job_config.job.dump_folder,
-            ft_manager=self.ft_manager,
-        )
+        self.checkpointer = self.build_checkpointer(model_args)
```
fegin: Please keep this and remove `ft_manager` and overwrite self.checkpointer in `FTTrainer.__init__()`.

torchtitan/train.py
```diff
@@ -385,6 +361,56 @@ def init_distributed(self) -> ParallelDims:
...
+            checkpoint_config=job_config.checkpoint,
+            sd_adapter=(
+                self.train_spec.state_dict_adapter(
+                    model_args, job_config.model.hf_assets_path
+                )
+                if self.train_spec.state_dict_adapter
+                else None
+            ),
+            base_folder=job_config.job.dump_folder,
+        )
```
fegin: Remove these three methods.

torchtitan/train.py
```diff
@@ -385,6 +361,56 @@ def init_distributed(self) -> ParallelDims:
...
+
+    def get_profiler_leaf_folder(self) -> str:
+        """Return the leaf folder path for profiler outputs."""
+        return ""
+
+    def get_training_context(self) -> AbstractContextManager:
+        """Return additional context manager for training loop."""
+        from contextlib import nullcontext
+
+        return nullcontext()
```
fegin: Remove these two methods.

torchtitan/train.py
```diff
@@ -385,6 +361,56 @@ def init_distributed(self) -> ParallelDims:
...
+                )
+                if self.train_spec.state_dict_adapter
+                else None
+            ),
+            base_folder=job_config.job.dump_folder,
+        )
+
+    def get_loss_sync_pg(self):
+        """Return the process group for loss synchronization."""
+        return None
```
fegin: Change this to `compute_global_losses()` and put it right above `train_step()`.

torchtitan/train.py
fegin, line 632: This block can be replaced with `self.compute_global_losses` and FTTrainer can overwrite the method.

torchtitan/train.py
```diff
@@ -653,11 +679,7 @@ def train(self):
         self.checkpointer.load(step=job_config.checkpoint.load_step)
         logger.info(f"Training starts at step {self.step + 1}")
 
-        leaf_folder = (
-            ""
-            if not self.ft_manager.enabled
-            else f"replica_{self.ft_manager.replica_id}"
-        )
+        leaf_folder = self.get_profiler_leaf_folder()
```
fegin: Remove this completely.

torchtitan/train.py
```diff
@@ -671,22 +693,7 @@ def train(self):
...
-                    else 0
-                ),
-                optimizer=self.optimizers,
-                fragment_fn=(
-                    self.train_spec.fragment_fn
-                    if hasattr(self.train_spec, "fragment_fn")
-                    else None
-                ),
-            ),
+            self.get_training_context(),
```
fegin: Remove this too.

torchtitan/train.py
fegin, line 647: Please just overwrite this method in FTTrainer.

torchtitan/train.py
tushar00jain: Doesn't `CheckpointManager.__init__()` performed side effects on the file system? if so, initializing it two times could have conflicts

torchtitan/experiments/ft/train.py
tushar00jain: Done, but had to create a separate method in FT manager call setup because FT manager expects distributed to be initialized but it only gets initialized in train.py

torchtitan/train.py
```diff
@@ -385,6 +364,30 @@ def init_distributed(self) -> ParallelDims:
             world_size=world_size,
         )
 
+    def get_dp_info(self, batch_degree: int, batch_rank: int) -> tuple[int, int]:
```
tianyu-l: I feel the whole point is for ft trainer not to influence how main trainer API looks like. These functions are there purely because ft trainer can override, which sounds the opposite.

The point is that, ft and main cannot share all the code, so they probably shouldn't. E.g. the Flux trainer has its own logic. I feel the ft trainer can just rewrite, instead of trying to share as much as possible.

tushar00jain: `batch_degree` is being used a lot which will require pretty much just copy pasting all off __init__.py. lmk if you'd still prefer that

torchtitan/train.py
fegin: No, I don't think `CheckpointManager.__init__()` actually create any files.

torchtitan/train.py
```diff
@@ -611,47 +614,48 @@ def train_step(
...
+                dist_utils.dist_sum(loss, loss_mesh, None),
+                dist_utils.dist_max(loss, loss_mesh, None),
                 dist_utils.dist_sum(
                     torch.tensor(
                         self.ntokens_seen, dtype=torch.int64, device=self.device
                     ),
                     loss_mesh,
-                    ft_pg,
+                    None,
                 ),
```
fegin: I believe the default is Non. So you can just remove all None.

torchtitan/train.py
```diff
@@ -660,11 +664,7 @@ def train(self):
         self.checkpointer.load(step=job_config.checkpoint.load_step)
         logger.info(f"Training starts at step {self.step + 1}")
 
-        leaf_folder = (
-            ""
-            if not self.ft_manager.enabled
-            else f"replica_{self.ft_manager.replica_id}"
-        )
+        leaf_folder = ""
```
fegin: The default is "", so this is not needed.

torchtitan/train.py
```diff
@@ -678,22 +678,6 @@ def train(self):
                 base_folder=job_config.job.dump_folder,
                 leaf_folder=leaf_folder,
```
fegin: Remove this line.

torchtitan/train.py
fegin: IMO, whether a change makes sense or not depending on whether the function is too intrusive.  This doesn't make sense because this function is not needed for the regular trainer.  I'm okay if this method contains the original logic in __init__():
```
        def get_batch_dim_info(parallel_dims):
            if parallel_dims.dp_enabled:
                batch_mesh = parallel_dims.get_mesh("batch")
                batch_degree, batch_rank = batch_mesh.size(), batch_mesh.get_local_rank()
            else:
                batch_degree, batch_rank = 1, 0
            return batch_degree, batch_rank
```

This is just my thought.

torchtitan/train.py
```diff
@@ -385,6 +364,30 @@ def init_distributed(self) -> ParallelDims:
...
+            checkpoint_config=job_config.checkpoint,
+            sd_adapter=(
+                self.train_spec.state_dict_adapter(
+                    model_args, job_config.model.hf_assets_path
+                )
+                if self.train_spec.state_dict_adapter
+                else None
+            ),
+            base_folder=job_config.job.dump_folder,
+        )
```
fegin: I still think we can remove this extra function.

torchtitan/models/flux/train.py
```diff
@@ -246,19 +246,18 @@ def train_step(
...
+                dist_utils.dist_sum(loss, loss_mesh, None),
+                dist_utils.dist_max(loss, loss_mesh, None),
                 dist_utils.dist_sum(
                     torch.tensor(
                         self.ntokens_seen, dtype=torch.int64, device=self.device
                     ),
                     loss_mesh,
-                    ft_pg,
+                    None,
                 ),
```
fegin, line 260: Remove all Nones.

torchtitan/train.py
```diff
@@ -385,6 +373,16 @@ def init_distributed(self) -> ParallelDims:
             world_size=world_size,
         )
 
+    def get_dp_info(self) -> tuple[int, int]:
+        """"""
```
fegin: Forgot to delete?

torchtitan/train.py
```diff
@@ -97,14 +95,7 @@ def __init__(self, job_config: JobConfig):
...
 
-        if parallel_dims.dp_enabled:
-            batch_mesh = parallel_dims.get_mesh("batch")
-            batch_degree, batch_rank = batch_mesh.size(), batch_mesh.get_local_rank()
-        else:
-            batch_degree, batch_rank = 1, 0
-
-        self.ft_manager = FTManager(job_config.fault_tolerance)
-        batch_degree, batch_rank = self.ft_manager.get_dp_info(batch_degree, batch_rank)
+        batch_degree, batch_rank = self.get_dp_info()
```
wwwjn: Can you override batch_degree and batch_rank in ft/trainer.py, and override following `self.dataloader ` and batch_size verify logic to avoid adding extra function in main train.py? Adding these 2 function is not useful for core train.py but makes the structure more confusing

torchtitan/train.py
```diff
@@ -611,9 +609,36 @@ def train_step(
...
+        }
+        self.metrics_processor.log(
+            self.step,
+            global_avg_loss,
+            global_max_loss,
+            grad_norm.item(),
+            extra_metrics=extra_metrics,
+        )
+
+    def compute_global_losses(
```
wwwjn: We can override the whole train_step() function in ft/train.py to avoid having a `compute_global_losses` function.

torchtitan/train.py
tushar00jain: No just followed your suggestion on having the entire logic contained in this method

torchtitan/train.py
tushar00jain: The variables are used inside `init` and so overwriting it after `init` will not have any use

torchtitan/train.py
tushar00jain: This was suggested by @fegin so we also don't have to copy over all the logic. Happy to change it though, lmk what you all prefer @fegin @wwwjn @tianyu-l

fegin: I personally am okay with this function. But it seems that people have different opinions, if so, we can overwrite this method.

tushar00jain: so rewrite both `__init__` and `train_step`? I guess @fegin is ok with the current approach. so @wwwjn and @tianyu-l please lmk which of these methods should I rewrite

PR #2313, opened 2026-02-02
Register `_ScaledPartial` placement

Aidyn-A:
In the PR https://github.com/pytorch/pytorch/pull/171482, the `PlacementClassVariable` and `PlacementVariable` were removed from TorchDynamo, so the explicit registration for `_ScaledPartial` is now required, otherwise it will fail with:
```
torch._dynamo.exc.InternalTorchDynamoError: AsPythonConstantNotImplementedError: 
UserDefinedObjectVariable(_ScaledPartial) is not a constant
```

This PR adds the registration for `_ScaledPartial` placement to the simple FSDP.

### DIFF
--- a/torchtitan/experiments/simple_fsdp/simple_fsdp.py
+++ b/torchtitan/experiments/simple_fsdp/simple_fsdp.py
@@ -350,16 +350,17 @@ def data_parallel(
 
 
 def _maybe_register_placement_as_opaque():
-    # The PyTorch PR https://github.com/pytorch/pytorch/pull/171482
-    # has removed PlacementVariable, so now TorchDynamo requires an opaque type
-    # registration for placements and derived objects like `_ScaledPartial`.
-    # According to https://github.com/pytorch/pytorch/blob/main/torch/_library/opaque_object.py
+    # The `PlacementVariable` was removed from TorchDynamo in
+    # https://github.com/pytorch/pytorch/pull/171482, so now TorchDynamo
+    # requires an opaque type, registration for placements. That includes
+    # derived classes like `_ScaledPartial`. According to the notes in:
+    # https://github.com/pytorch/pytorch/blob/main/torch/_library/opaque_object.py
     # Opaque objects are the way TorchDynamo allows custom operators to accept
-    # a user-defined "black box" object as an input. Users can register a custom class as:
-    #    register_opaque_type(MyClass, typ=..., members=...)
-    # where `typ` is either "reference" or "value", and `members` is a dictionary mapping
-    # member names (attributes, properties, or methods) to their MemberType, which controls
-    # how they are handled during torch.compile tracing
+    # a user-defined "black box" object as an input. Users can register their
+    # custom classes as `register_opaque_type(MyClass, typ=..., members=...)`,
+    # where `typ` is either "reference" or "value", and `members` is a dictionary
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/simple_fsdp/simple_fsdp.py
```diff
@@ -347,3 +347,22 @@ def data_parallel(
...
+    from torch._library.opaque_object import MemberType, register_opaque_type
+
+    allowed_members = {
+        "reduce_op": MemberType.USE_REAL,
+        "is_shard": MemberType.USE_REAL,
+        "is_partial": MemberType.USE_REAL,
+        "is_replicate": MemberType.USE_REAL,
+        "__eq__": MemberType.USE_REAL,
+    }
+    register_opaque_type(
```
tianyu-l: please comment
- what this function is doing
- why we need it
- what is allowed_members

Thanks!

Aidyn-A: Thanks for the review. I have added a comment. However, the PR https://github.com/pytorch/pytorch/pull/171482 got reverted. If it lands, this PR should land as well.

torchtitan/experiments/simple_fsdp/simple_fsdp.py
```diff
@@ -347,3 +347,38 @@ def data_parallel(
...
+
+        allowed_members = {
+            "reduce_op": MemberType.USE_REAL,
+            "is_shard": MemberType.USE_REAL,
+            "is_partial": MemberType.USE_REAL,
+            "is_replicate": MemberType.USE_REAL,
+            "__eq__": MemberType.USE_REAL,
+        }
+        register_opaque_type(
+            _ScaledPartial,
```
tianyu-l, line 378: @wwwjn We introduced `_ScaledPartial` to mimic FSDP2's `set_gradient_divide_factor`. Now that we handles gradient scale by ourselves, I feel we can deprecate this field. WDYT?

wwwjn: As long as simpleFSDP don't rescale gradients, we should be good. Does `_ScaledPartial` means the gradients on each rank are scaled (divided by fsdp degree) when the information in the tensor are still partial / unreduced?

tianyu-l: yeah good point. It does scaling here https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/simple_fsdp/simple_fsdp.py#L204 by using Partial(avg) instead of P(sum).

@Aidyn-A could you help change it to P(sum) and deprecate `_ScaledPartial`? We also need to compare loss curve with FSDP2 to make sure the change is correct.

If it sounds too involved, we can do it.

Aidyn-A: Sure, do you want me to add a warning that `_ScaledPartial` is deprecated or immediately remove `_ScaledPartial`. In any case, what should I do with `reduction_divide_factor`? Where will that go?

tianyu-l: Don't need to worry about warning. For `reduction_divide_factor`, definition and usage should all go away. Thanks!

Aidyn-A:
`_ScaledPartial` was removed in #2337. Closing.

PR #2314, opened 2026-02-02
Fixing global_max_loss computation

Shagun-G:
Summary:
As per the fix to handle token imbalance D91432940 , the global_max_loss computed is below the global_avg_loss as shown in the test plan. This is because of the pre division of the loss to the global tokens. While the avg loss computation is correct, the global max loss in this case becomes the max of the loss contribution of each rank, no the max of the local loss of each rank.

One fix is to recale the local loss to the average loss for each rank to get the local average loss and then perform the max operation. The result can be seen in the test plan.

Differential Revision: D92079302

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -616,12 +616,19 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
             ft_pg = self.ft_manager.loss_sync_pg
             loss_mesh = parallel_dims.get_optional_mesh("loss")
 
+            # For global_avg_loss, we want the average loss across all ranks:
             # loss = local_loss_sum / global_valid_tokens
-            # summing across ranks will have result:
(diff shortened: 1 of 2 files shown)


torchtitan/train.py
```diff
@@ -619,9 +619,14 @@ def train_step(
             # loss = local_loss_sum / global_valid_tokens
```
wwwjn, line 620:
```suggestion
            # loss is define as local_loss_sum / global_valid_tokens
```

wwwjn: Thanks! Let's make the calculation of `global_avg_loss` and `global_avg_loss` comment more clear.  Can you help modify the comment about how to calculate `global_avg_loss` as well?

And please change the definition here as well: https://github.com/pytorch/torchtitan/blob/fdd3a37a0f3e35dbf3953c30b6262d4428ecfb9b/torchtitan/components/metrics.py#L411-L412

PR #2315, opened 2026-02-03
Fix the CI loss issue

fegin:
* __->__ #2315

https://github.com/pytorch/pytorch/pull/173569 changes how NormPartial works and potentially change the loss.

This PR changes
1. Only compare FSDP and HSDP for one step.
2. Have a seperate run to compare the loss of FSDP after 100 steps with previous result (also FSDP).


tests/assets/losses/llama3_cuda.txt
```diff
@@ -1,10 +1,100 @@
...
-7 4.5606
-8 4.3724
-9 4.347
-10 4.2004
+5 5.4271
+6 4.7476
+7 4.3239
+8 4.0675
+9 3.9687
+10 3.7249
```
tianyu-l: hmm, why the change is so drastic -- does it mean previously there are bugs?
cc @pianpwk

fegin: Not sure if it is unrealistic but the first couple steps we have big learning rate, so some minor difference in the gradient can result in big change?

tianyu-l: No if we use the same seed, it's absolutely not nondeterminism -- there must be something going on.

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -77,21 +77,24 @@ jobs:
         # Set architecture-specific parameters
         if [[ "${{ matrix.gpu-arch-type }}" == "cuda" ]]; then
           LOSS_FILE="tests/assets/losses/llama3_cuda.txt"
-          STEPS=10
+          STEPS=1
```
tianyu-l, line 80: are they (steps=1) intentional?

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -77,21 +77,24 @@ jobs:
...
         else
           echo "Error: Unknown GPU architecture type: ${{ matrix.gpu-arch-type }}"
           exit 1
         fi
 
-        python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --test-options="${test_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --assert-equal --steps=${STEPS} --import-result ${LOSS_FILE}
+        python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --test-options="${test_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --assert-equal --steps=1
+        rm -rf $RUNNER_TEMP/artifacts-to-be-uploaded/*
+
+        python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --import-result="${LOSS_FILE}" --assert-equal --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs/loss.txt" --steps=100
```
wwwjn: Why we only compare step=1 with the LOSS_FILE in previous comparison, but run again with 100 steps, why don't we merge these comparision?

tests/assets/losses/llama3_cuda.txt
shuhuayu: Can i ask why the first 4 steps remain unchanged?

.github/workflows/integration_test_8gpu_features.yaml
fegin: Yes, see the below discussion.

.github/workflows/integration_test_8gpu_features.yaml
fegin: We cannot compare FSDP and HSDP for 100 steps anymore. These are essentially two different comparisons. For one step, that is to ensure that FSDP and HSDP first forward loss is on par. After the first step, we couldn't guarantee the same loss anymore.

tests/assets/losses/llama3_cuda.txt
fegin: > No if we use the same seed, it's absolutely not nondeterminism -- there must be something going on.

I don't understand the `nondeterminism` statement. The loss change is from https://github.com/pytorch/pytorch/pull/173569 removing allreduce, according @pianpwk. So the loss change is expected. The loss is deterministic after https://github.com/pytorch/pytorch/pull/173569. But the loss is different from what we have previously, https://github.com/pytorch/pytorch/pull/173569.

I thought your concern is the degree of the change, why `nondeterminism`?

> Can i ask why the first 4 steps remain unchanged?
Only the first step doesn't change, but we only compare the 4 digits. We can print out more digits if this confuses people.

pianpwk: Ah also, that PR is reverted for now... the same issue showed up internally (doesn't match numerics before)

.github/workflows/integration_test_8gpu_features.yaml
wwwjn: Thanks, so here's my understanding of the code:
1. `llama3_cuda/rocm.txt` has loss from FSDP (dp_shard=8) experiments
2. This line we are comparing a new FSDP run (baseline_options = "dp_replicate=1", so FSDP degree = 8), with the result in `llama3_cuda/rocm.txt`

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -77,21 +77,24 @@ jobs:
...
           LOSS_FILE="tests/assets/losses/llama3_rocm.txt"
-          STEPS=5
+          STEPS=1
         else
           echo "Error: Unknown GPU architecture type: ${{ matrix.gpu-arch-type }}"
           exit 1
         fi
 
-        python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --test-options="${test_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --assert-equal --steps=${STEPS} --import-result ${LOSS_FILE}
+        python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --test-options="${test_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --assert-equal --steps=1
```
wwwjn, line 97:
```suggestion
        echo "Checking FSDP8 loss from a new run v.s. FSDP8 loss from text file parity"
        python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --test-options="${test_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --assert-equal --steps=1
```

tests/assets/losses/llama3_cuda.txt
tianyu-l: ok do I understand correctly that, with the PR reverted, the current PR should fail?

Also, if the change is caused by that PR, why the change is so drastic?

fegin: > ok do I understand correctly that, with the PR reverted, the current PR should fail?
Yes

And since the PR cause more than one training framework numeric changes, I would suggest that @pianpwk  verifies both frameworks before relanding the PR.

torchtitan/components/metrics.py
```diff
@@ -447,7 +447,7 @@ def log(
         color = self.color
         logger.info(
             f"{color.red}step: {step:2}  "
-            f"{color.green}loss: {global_avg_loss:7.4f}  "
+            f"{color.green}loss: {global_avg_loss:8.5f}  "
```
tianyu-l, line 463: not sure if this format change would break anything in your internal loss check script

fegin: We used space to separate the metrics, so this should be okay. Will fix it if this does break something.

PR #2319, opened 2026-02-03
Fix gpt-oss implementation (MoE router gate bias + top‑k renorm)

### LINKED ISSUE #2258: GPT-OSS MoE's implementation should use `route_norm=True`
### Bug description

The current implementation of gpt-oss sets `route_norm=False`, meaning softmax is applied over all expert logits instead of only the selected topk. However, in the paper, it is stated that "weight the output of each expert by the softmax of the router projection over only the selected experts." 

So we should set `route_norm=True` to apply softmax only over the selected experts. 

### Versions

The code is latest commit a25dd8f880e0759c1496d07ff0ca6973c6dc8b65

linyuhongg:
Fixes #2258.

The official gpt-oss MoE router uses a biased gate projection and computes routing weights as a softmax over the selected top‑k logits (so the selected weights sum to 1). Torchtitan’s gpt_oss configs previously used a bias-less router gate and did not renormalize the selected top‑k scores, which diverges from the reference implementation and makes `mlp.router.bias` checkpoint mapping ineffective.

Changes:
- Add `MoEArgs.route_gate_bias` and plumb it into `TokenChoiceTopKRouter` (`nn.Linear(..., bias=route_gate_bias)`).
- Enable `route_gate_bias=True` and `route_norm=True` in all gpt_oss model configs to match the reference behavior.
  - With `score_func=softmax`, `route_norm=True` is equivalent to doing softmax over the selected top‑k logits.


shuhuayu:
Hi @linyuhongg, thank you for the pr. Could you sign the CLA so we can proceed with the merge?

torchtitan/models/moe/moe.py
wwwjn, line 206: thanks, please add an definition of the new parameter in docstring

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -49,8 +50,9 @@
             num_experts=32,
             num_shared_experts=0,
             score_func="softmax",
-            route_norm=False,
+            route_norm=True,
             route_scale=1.0,
+            route_gate_bias=True,
```
wwwjn: nit: probably just name it `gate_bias`, seems `route_*` are related to operation on scores

wwwjn (approved):
Thanks for the fix, LGTM

PR #2321, opened 2026-02-04
Fixed autoparallel integration tests on ROCm.

wenchenvincent:
This PR fixes the autoparallel integration tests on ROCm.

Background:
HIP runtime relies on the file /opt/amdgpu/share/libdrm/amdgpu.ids to look up the product name of AMDGPU. But this file is missing in the CI docker container. Only `/usr/share/libdrm/amdgpu.ids` exists in the docker container and it is out of date - it does not include newer products like MI300. One way to address this is to use `-v /opt/amdgpu:/opt/amdgpu:ro`  when launching the docker container to map the /opt/amdgpu/share/libdrm/amdgpu.ids on the host to the docker container. This, however, requires changes of the workflows files in pytorch/test_infra and pytorch/pytorch. This quick fix works around the issue by updating the amdgpu.ids file within the CI docker container.

### DIFF
--- a/torchtitan/experiments/autoparallel/tests/integration_tests.py
+++ b/torchtitan/experiments/autoparallel/tests/integration_tests.py
@@ -30,7 +30,6 @@ def build_autoparallel_test_list() -> list[OverrideDefinitions]:
             "llama3 AutoParallel FSDP+TP",
             "llama3_autoparallel_fsdp_tp",
             ngpu=4,
-            skip_rocm_test=True,
         ),
         # TODO: Re-enable this once we fix the test
         # deepseek_v3 tests


wenchenvincent:
> > HIP runtime relies on the file /opt/amdgpu/share/libdrm/amdgpu.ids to look up the product name of AMDGPU. But this file is missing in the CI docker container.
> 
> I don't think we have the urgency to enable autoparallel CI on rocm. I don't mind waiting until a proper fix (in docker, pytorch, or test_infra) is available, compared with landing this workaround.

Hi @tianyu-l . I talked to our pytorch rocm team about this issue. Using `-v /opt/amdgpu:/opt/amdgpu:ro` when launching the docker container isn't a good option either as proprietary and open-source amdgpu driver uses different paths. Pytorch ROCm heavyweight wheels actually packaged its own copy of libdrm and amdgpu.ids file to address this issue. However, it was broken silently due to an update in libdrm. There was a recent PR in pytorch to fix this: https://github.com/pytorch/pytorch/pull/174811 and I verified it with the latest nightly wheel (https://download.pytorch.org/whl/nightly/rocm7.1/torch-2.11.0.dev20260215%2Brocm7.1-cp312-cp312-manylinux_2_28_x86_64.whl). I updated this PR to just enable the autoparallel test and the test should be passing when CI is re-triggered.

PR #1895, opened 2025-10-16
[MoE][compile][full ac] weave torch.compile around the FSDP(GroupedExperts) graph break

xmfan:
Stacked PRs:
 * __->__#1895

--- --- ---

This PR changes how we compile MoE layers to work around Compile + AC limitations. When you `AC(Compile(block))` or `Compile(AC(block))` and there is a graph break in `block`, we fall back the entire block to eager. For llama3, we've worked around this problem by addressing all graph breaks. With MoE models particularly dp2ep, we need to wrap`FSDP(block.moe.experts)`,  meaning that we will have graph breaks when tracing `block.moe.experts.__call__`, meaning that whenever AC was enabled, the entire block for MoE would fallback to eager: https://gist.github.com/xmfan/50f4de1e89d789cd63a21aca9e600132 (Note in the tlparse, graph 0/1 is empty and it corresponds to the block containing the MoE). 

The workaround in this PR is to avoid tracing `block.moe.experts.__call__`. This is done by individually wrapping torch.compile on submodules of TransformerBlock. Note that we are leaving some perf on the table as this might exclude some ops in TransformerBlock.forward and MoE.forward. This is an API limitation, as we have no way to acquire those ops while decoupling the wrapper from model code. This workaround will no longer be necessity when either:
- We can do Compile + AC with graph breaks
- We remove the FSDP graph break

This change introduces a small regression to the non-AC configuration. You can see a small perf dip from [before this PR](https://gist.github.com/xmfan/0b32e95980d263cf3f62869fa4d85921) and [after this PR](https://gist.github.com/xmfan/11561b5406b3f92ecd08da94bc5ee4e3). Given that AC is a necessity to run non-toy configurations of these models, I chose to stick to this implementation to make comparisons easier.

Validated on DSv3 debug model:
- dp2ep, no AC, no compile: https://gist.github.com/xmfan/927f354158ad36f4c5c1ffedde4e4ebe
- dp2ep, no AC, compile: https://gist.github.com/xmfan/11561b5406b3f92ecd08da94bc5ee4e3
  - before this PR (compile w/ nested graph break): https://gist.github.com/xmfan/0b32e95980d263cf3f62869fa4d85921
- dp2ep, full AC, compile: https://gist.github.com/xmfan/6ed5b48aa51ce0ac2b6bfceb86a0c482
  - before this PR (whole moe block in eager): https://gist.github.com/xmfan/50f4de1e89d789cd63a21aca9e600132
[...]

PR #2281, opened 2026-01-28
[FSDP2] enable per-param mesh FSDP2 for MoE

weifengpy:
command: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_16b ./run_train.sh --training.steps 20 --parallelism.expert_parallel_degree 4`

fsdp2 support per-param mesh: https://github.com/pytorch/pytorch/pull/173509

this PR applies fully_shard on transformer_block, sharding experts on `edp_mesh`, and other params on `dp_mesh`. rebased on latest main and handles ep size == 1 carefully 

Screenshot 2026-03-26 at 15 29 31

loss and grad_norm are identical using 10 decimal
Screenshot 2026-03-26 at 17 11 17

Each layer issues 2 AGs and 2 RS, back-to-back. AG and RS overlapps if one is on dp mesh and another on efsdp mesh
Screenshot 2026-03-26 at 17 13 42

active memory are exactly the same. reserved memory are tiny tiny higher
Screenshot 2026-03-26 at 17 16 40

PR #2322, opened 2026-02-04
Apply #1895 only when really necessary

ericschreiber:
We observed a significant performance regression for MoE models using pure FSDP, with MFU dropping from [24.5%](https://gist.github.com/ericschreiber/4c260ff1d40954cc979fbb35fc3c3b2a) to [16.3%](https://gist.github.com/ericschreiber/5e5ddd0f3fa6bd4aacff596abd16f1fc) when running [Qwen3 30A3B](https://gist.github.com/ericschreiber/2d5e795c60b07bd759c2a8e98b946e18) on 2×8 B200s. This regression was traced back to commit 2a7a148, associated with issue #1895. 

While 2a7a148 correctly fixes Dynamo graph breaks for combined EP + FSDP setups, the change is unnecessary for pure FSDP configurations and introduces avoidable overhead in that case.

This pull request refactors and extends the `apply_compile` logic used for MoE model parallelization (DeepSeekV3, Llama4, Qwen3) to better distinguish between EP/FSDP combinations. The fix from #1895 is now applied only when it is actually required. We verified that the experiments from #1895 continue to work as expected.

Because FSDP is applied after `apply_compile`, we introduce an additional `fsdp_enabled` flag, which results in corresponding updates to files that inherit from this logic.

Finally, as was the case prior to commit 2a7a148, enabling SAC with Qwen3 30A3B triggers the warning:
```
Detected that context_fn is passed to torch.utils.checkpoint under torch.compile.
Please make sure the checkpointed region does not contain in-place ops (e.g. torch.relu_).
```

This appears to be related to how SAC augments `context_fn`. For Qwen3, we verified that this warning can be safely ignored. Still, we would appreciate reviewer input on how this warning should be handled going forward.

### DIFF
--- a/torchtitan/models/deepseek_v3/infra/parallelize.py
+++ b/torchtitan/models/deepseek_v3/infra/parallelize.py
@@ -156,7 +156,12 @@ def parallelize_deepseekv3(
         )
 
     if model_compile_enabled:
-        apply_compile(model, job_config.compile, parallel_dims.ep_enabled)
+        apply_compile(
+            model,
+            job_config.compile,
+            parallel_dims.ep_enabled,
+            parallel_dims.fsdp_enabled,
+        )
 
     dp_mesh: DeviceMesh | None = None
     if parallel_dims.fsdp_enabled or parallel_dims.ep_enabled:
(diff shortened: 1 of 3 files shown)


wconstab:
@xmfan can you take a look at this?

weifengpy:
> I believe @weifengpy has an ongoing work (block-level compile even with EP) to fix FSDP / EP + FSDP altogether.
> 
> I'd prefer we waiting for that, instead of landing this temp workaround.

I am verifying profiler traces of per-param mesh FSDP2, that will unblock per-layer torch.compile for MoE. Should be able to publish soon: https://github.com/pytorch/torchtitan/pull/2281/

PR #1714, opened 2025-09-15
[torchtitan][replicate] experimenting new replicate integration with torchtitan

anshul-si:
**Summary:** During this experiment to integrate the new replicate function into torchtitan, I used https://github.com/pytorch/pytorch/pull/162021, which has not been landed. However, since this is more about making replicate more efficient rather than changing replicate's core code, https://github.com/pytorch/pytorch/pull/160135, which has landed, should be fine. https://github.com/pytorch/pytorch/pull/160133 is the last time replicate_with_fsdp.py and its replicate api were touched. 

In order to enable the new replicate, which uses a 2D device mesh (since it is a specialized version of HSDP), I changed the parallelism code to include dp_shard dim = 1 only if dp_replicate > 1, and created device mesh that I pass down in apply_ddp. 

The numeric tests for tp + replicate and pp + replicate can be seen below. In order to ensure that they worked, I also compared them with HSDP (n, 1) (replicate, shard).

image

https://fburl.com/mlhub/5k9v43w3

**Test Case**
1. CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml" ./run_train.sh (set replicate to 8)

Expected output of this experiment should be something like:
[rank0]:[titan] 2025-09-15 17:38:26,676 - root - INFO - Starting job: Llama 3 debug training
[rank0]:[titan] 2025-09-15 17:38:29,094 - root - WARNING - ENV[TORCH_NCCL_ASYNC_ERROR_HANDLING] = 1 will be overridden to 3 based on job config
**[rank0]:[titan] 2025-09-15 17:38:29,097 - root - INFO - Building 2-D device mesh with ['dp_replicate', 'dp_shard'], [8, 1]**
[rank0]:[titan] 2025-09-15 17:38:29,104 - root - INFO - [GC] Initial GC collection 0.00 seconds
[rank0]:NCCL version 2.27.5+cuda12.6
[rank0]:[titan] 2025-09-15 17:38:35,439 - root - INFO - Loading tokenizer from tokenizer.json
[rank0]:[titan] 2025-09-15 17:38:35,441 - root - INFO - Preparing c4_test dataset from tests/assets/c4_test
[...]

PR #2323, opened 2026-02-04
Disable DDP averaging to avoid repeated gradient averaging

Shagun-G:
Summary:
From the change to averaging over  microbatches to averaging over the global tokens, the averaging for FSDP was disabled in D91432940 but not for DDP. This adds an additional scaling to the gradients diving them twice by the number for DP ranks when using pure DDP. This error does not emerge in change in loss for short runs due to the scale invariance property of the AdamW optimizer but can be seen clearly in the grad norm measuerment and difference in the measurement from FSDP.

To fix this, as DDP does not have a property set_gradient_divide_factor() as FSDP, the solution employed is to put in a comm hook that replaces the default all reduce average operation with an all reduce sum operation

Differential Revision: D92301896

### DIFF
--- a/torchtitan/models/llama3/infra/parallelize.py
+++ b/torchtitan/models/llama3/infra/parallelize.py
@@ -8,6 +8,7 @@
 # training techniques (e.g. activation checkpointing and compile) to the Llama model.
 
 import torch
+import torch.distributed as dist
 import torch.nn as nn
 from torch.distributed._composable.fsdp import FSDPModule
 from torch.distributed._composable.replicate import replicate
@@ -284,6 +285,38 @@ def disable_fsdp_gradient_division(model: nn.Module) -> None:
             module.set_gradient_divide_factor(1.0)
 
 
+def _ddp_sum_allreduce_hook(
+    process_group: dist.ProcessGroup, bucket: dist.GradBucket
+) -> torch.futures.Future[torch.Tensor]:
+    """
+    DDP communication hook that performs all-reduce with SUM (no averaging).
+
+    Unlike the default DDP hook which divides by world size before all-reduce,
+    this hook performs a pure sum reduction. This is used when gradient scaling
+    is handled manually in the training loop (e.g., dividing by global token count).
+    """
+    return (
+        dist.all_reduce(bucket.buffer(), group=process_group, async_op=True)
+        .get_future()
+        .then(lambda fut: fut.value()[0])
+    )
+
+
+def disable_ddp_gradient_averaging(model: nn.Module, dp_mesh: DeviceMesh) -> None:
+    """
+    Disable DDP's automatic gradient averaging by registering a custom comm hook.
+
+    By default, DDP divides gradients by world size before all-reduce.
+    This function registers a custom hook that uses SUM reduction instead,
+    allowing manual gradient scaling in the training loop (e.g., by global token count).
+
+    Args:
+        model: The model wrapped with DDP (via replicate())
+        dp_mesh: The device mesh used for data parallelism
+    """
+    model.register_comm_hook(dp_mesh.get_group(), _ddp_sum_allreduce_hook)
+
+
 def apply_fsdp(
     model: nn.Module,
     dp_mesh: DeviceMesh,
@@ -370,5 +403,8 @@ def apply_ddp(
 
     # pyrefly: ignore [invalid-param-spec]
     replicate(model, device_mesh=dp_mesh, bucket_cap_mb=100)
+
+    # Disable DDP's automatic gradient averaging for all DDP modules
+    disable_ddp_gradient_averaging(model, dp_mesh)
 
     logger.info("Applied DDP to the model")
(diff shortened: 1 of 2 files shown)


torchtitan/models/llama3/infra/parallelize.py
```diff
@@ -284,6 +285,38 @@ def disable_fsdp_gradient_division(model: nn.Module) -> None:
...
+) -> torch.futures.Future[torch.Tensor]:
+    """
+    DDP communication hook that performs all-reduce with SUM (no averaging).
+
+    Unlike the default DDP hook which divides by world size before all-reduce,
+    this hook performs a pure sum reduction. This is used when gradient scaling
+    is handled manually in the training loop (e.g., dividing by global token count).
+    """
+    return (
+        dist.all_reduce(bucket.buffer(), group=process_group, async_op=True)
```
wwwjn, line 299: Thanks, have you tried using https://github.com/pytorch/torchtitan/blob/1a36996b8dffb464340c280d3f0a8139ca4c36b6/torchtitan/distributed/utils.py#L30 instead of all_reduce directly? The difference is this function handles when the tensor is DTensor (eg, when TP is applied and the gradient bucket thensor might be sharded)

Can you attach more test results, eg:
1. Does HSDP + TP  work?
2. DDP only vs FSDP vs HSDP grad norm curve?

Shagun-G: I have updated the test plan showing all the cases. Simultaneously, as DDP does not use DTensor within torch titan and cannot be used with TP or any other parallelism as restricted by design, I think using all_reduce directly is better here to keep things simple.

Shagun-G:
One major update in the subsequent version of this pull request is the addition of no_sync context management in the training loop for DDP. I am not sure what other way is there to manage this but as updated in the diff summary, this is necessary to ensure correct computations when performing gradient accumulation.

fegin (changes requested):
This has been done in PP, https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/stage.py#L639. Instead of adding the logic to TorchTitan, we should investigate why PP's logic doesn't work. My best guess is we are using replicate so the module type is not DistributedDataParallel, but replicate (or some other module class) cc., @H-Huang

tianyu-l (changes requested):
@fegin 
> This has been done in PP

Wait why does it have anything to do with PP?

The proper fix is to support `set_gradient_divide_factor` in `replicate()`, no?

cc @anshul-si @wwwjn

anshul-si:
@fegin @tianyu-l to be clear, we still haven't merged changes integrating replicate_fsdp version with torchtitan. I think the old replicate is based on DDP right?

fegin:
@tianyu-l  My comment was not clear. The `set_gradient_divide_factor` is just one change, which this PR uses `register_comm_hook` to achieve. But the modification to PP determining whether set_requires_gradient_sync to True or False should already be handled by PP, https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/stage.py#L639. That's what I meant.

Shagun-G:
Hi everyone, I just want to summarize the diff to help clear confusion. The main source of change is the change from global averaging over tokens from individual local averaging over microbatches from a previous PR. While the implicit default averaging over the individual DP ranks was disabled for FSDP, the DDP averaging in the all reduce operation was not. As the DDP settings used here are from the old primitive, not using DTensor backend (according to what I see, please correct if wrong), no option for set_gradient_divide_factor is available and the all reduce operation has been explicitly changed from averaging to sum.

This change brings up another difficulty as the old DDP primitive performed the all reduce comm operation every backward pass. This is was fine previously as an average was performed every backward pass but now is incorrect as this causes double summation over multiple gradients when using gradient accumulation. To counteract this, all reduce has been disabled for all backward passes other than the last one. This is not necessary under FSDP as is already implemented there. This is also completely independent of PP as it is necessary to obtain the correct gradients even when PP = 1.

fegin:
I misunderstood the code change. I had the impression that the code change was in forward_backward_step, but it is in train_step. So, my comment was incorrect, this was independent from PP.

wwwjn:
> @fegin @tianyu-l to be clear, we still haven't merged changes integrating replicate_fsdp version with torchtitan. I think the old replicate is based on DDP right?

Yes, the ideal fix would be integrating `replicate_fsdp` , and then call `set_gradient_divide_factor()` as well when pure data parallel is applied.  Can you remind me how much work are need to be done before we can merge the replicate_fsdp PR?

Shagun-G:
Hi everyone, I just wanted to ask if it is possible to land this change or if there is other parallel work going on to update the codebase to fix the error as there is an error in the current version of the codebase for DDP.

wwwjn:
> Hi everyone, I just wanted to ask if it is possible to land this change or if there is other parallel work going on to update the codebase to fix the error as there is an error in the current version of the codebase for DDP.

As #1714 solved this issue, we would pursue the recplicate() method in #1714

PR #2324, opened 2026-02-04
[rl][ez] Fix import errors.

zhxchen17:
* #2325
* __->__ #2324

Summary:

I'm seeing some import errors while rebasing on the latest vllm and torchtitan. Fixing these.

Test Plan:
python torchtitan/experiments/rl/unified/infer.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchtitan/experiments/rl/unified/models/attention.py
+++ b/torchtitan/experiments/rl/unified/models/attention.py
@@ -5,7 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
-from vllm.attention.layer import Attention
+
+try:
+    from vllm.attention.layer import Attention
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -5,7 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
-from vllm.attention.layer import Attention
+
+try:
+    from vllm.attention.layer import Attention
+except ModuleNotFoundError:
+    from vllm.model_executor.layers.attention import Attention
```
tianyu-l, line 12: why both? let's use the latest one

PR #2329, opened 2026-02-05
[SimpleFSDP] Fix HSDP placement mismatch in _distribute_dtensor

SongyuanZhao:
The current_spec and target_spec in _distribute_dtensor were only providing 
one placement for HSDP's 2D outer_mesh, causing a dimension mismatch.

Fixed by using (Replicate(),) * len(dp_placements) to automatically generate 
the correct number of placements matching the mesh dimensionality:
- 1 placement for FSDP/DDP (1D mesh)
- 2 placements for HSDP (2D mesh)

### DIFF
--- a/torchtitan/experiments/simple_fsdp/simple_fsdp.py
+++ b/torchtitan/experiments/simple_fsdp/simple_fsdp.py
@@ -130,7 +130,7 @@ def _distribute_dtensor(
     # HSDP case needs 2 placements for 2D outer_mesh
     current_placements = (Replicate(),) * len(dp_placements)
     target_placements = tuple(dp_placements)
-    
+
     current_spec = DTensorSpec(
         mesh=outer_mesh,
         placements=current_placements,


torchtitan/experiments/simple_fsdp/simple_fsdp.py
```diff
@@ -127,14 +127,18 @@ def _distribute_dtensor(
             f"Unsupported placement {dp_placements} for distributing DTensor {tensor}"
         )
 
+    # HSDP case needs 2 placements for 2D outer_mesh
+    current_placements = (Replicate(),) * len(dp_placements)
```
tianyu-l, line 131: This makes sense to me because the `outer_mesh` used below should be 2d mesh in HSDP case.

I wonder if why this test HSDP(2,2) TP2 didn't captures it?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/simple_fsdp/tests/integration_tests.py#L183

sorenzzio: I believe this is primarily because the sizes of dp_replica and dp_shard are identical.

First, a 1D placement is passed in while the device_mesh is 2D. In this case, sharding is performed along mesh_dim 0 (dp_replica), so calling _distribute_dtensor does not trigger an error.

Second, although sharding is performed on the incorrect dimension, the resulting shape remains correct because dp_replica == dp_shard. Consequently, the subsequent execution proceeds without errors.

tianyu-l: that's very unfortunate -- I think we should add assertion in `DTensorSpec` creation that the number of placement elements should match the mesh ndim. cc @wconstab @pianpwk

tianyu-l (approved):

torchtitan/experiments/simple_fsdp/simple_fsdp.py
fegin: Does it even make sense for a DTensorSpec to accept the case where mesh and placements lengths are different? 

cc., @tianyu-l @wconstab

tianyu-l: It doesn't.

PR #1284, opened 2025-06-12
Redefine lr_scheduler behavior

wwwjn:
## Context

The current warmup-stable-decay lr_scheduler behavior is not intuitive. For example, in `debug_model.toml`, the configurattion is: 
```
[lr_scheduler]
warmup_steps = 2  # lr scheduler warm up, normally 20% of the train steps
decay_ratio = 0.8  # lr scheduler decay ratio, 80% of the train steps
decay_type = "linear"
lr_min = 0.0
```
So we are expected warmup_steps =2, decay steps=8, total steps=10.

And we got the curret learning rate (blue line) There are 2 issues:
1. The max learning rate is not reaching 1 (expected max lr = 1 since we are calculating ratio here). 
2. Intuitively, the user would expect to see learning rate increase by 0.5 (suppose max_lr = 1) each step during warm up stage, and decrease by 1/8 each step during decay stage. But in blue line, the lr is increasing by 1/3, and decreasing by 1/9, which is counter-intuitive. 

Thus we propose a standard lr_scheduler behavior, which aligns with user's intuitive and the meaning of the parameter names. (the red line)
learning_rate_schedule

## Standard lr_scheduler behavior
- Warm up stage:  LR increase by 1/{warmup_steps}
- Stable stage: Length of  stable stage = total_train_step + 1 - warmup_stage - decay_stage. We manually add one step to stable stage (so we have a fake step 11), which is preventing if decay is enabled the step 10 learning rate drops to 0. 
- Decay stage:   LR decrease by 1/{decay_steps}

PR #2333, opened 2026-02-05
Torchtitan changes to integrate into Verl

acisseJZhong:
**Goal:** This PR makes the changes so that we can integrate Torchtitan as a trainer to Verl: https://github.com/verl-project/verl/pull/5051

**Major changes:** 
1. ~~Change LR schedule to be 0 indexed instead of 1 indexed; to align with Verl's [fsdp util ](https://github.com/verl-project/verl/blob/d987199906f09ba53139df13e4528b2d575ec4ce/verl/utils/torch_functional.py#L745)
See more analysis in https://docs.google.com/document/d/1YiFUvIa_JqTYpBd2Xj7ReH3Bw6wS07nKldycBX--uVE/edit?usp=sharing~~ ==> We decide not to change Titan's LR Scheduler behavior. 
image

2. ~~add `position_block_causal` attn mask type, which creates block causal mask based on `position_id` for both varlen and flex attention: [transformers reference](https://github.com/huggingface/transformers/blob/0c89522f2af2f85cf997193645a1e727d6b8c1d7/src/transformers/masking_utils.py#L708)~~ ==> this is added in Verl's Torchtitan Engine code instead 

**Todos:**
1. Enable PP, right now [`pp_schedule.eval()` ](https://github.com/pytorch/pytorch/blob/03406903616077227734f772d682fc6027513ecf/torch/distributed/pipelining/schedules.py#L402)does the microbatch split for us, as it takes in the whole batch. However, in verl we split batch into microbatches before pp, and we'd love to pass in a list of pre-split microbatches to pp schedule. (thanks for @H-Huang's help)

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -699,6 +699,7 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful):
                 ),
             ),
         ):
+            assert self.dataloader is not None, "Dataloader is not initialized"
             data_iterator = self.batch_generator(self.dataloader)
             while self.should_continue_training():
                 self.step += 1


torchtitan/distributed/utils.py
```diff
@@ -293,6 +293,10 @@ def init_distributed(
     base_folder: str = "",
     ranks: list[int] | None = None,
 ) -> int:
+    # Skip initialization if already initialized
+    if torch.distributed.is_initialized():
+        return torch.distributed.get_world_size()
```
tianyu-l, line 302: If going this path, it means a lot of the setting in this function / config won't take effect. Shall we add a warning to users?

torchtitan/components/lr_scheduler.py
```diff
@@ -155,20 +155,17 @@ def linear_warmup_stable_decay(
         """
         warmup_stable_steps = warmup_steps + stable_steps
         if current_step < warmup_steps:
-            # linear warmup
-            # 0-indexed step, hence + 1 adjustments
-            current_step += 1
+            # linear warmup (0-indexed to match FSDP/HuggingFace)
```
tianyu-l: we spent a lot of time converging to the current behavior.
this change will likely break unit test and user code (silent change)
let's be careful about this change.
cc @wwwjn

torchtitan/distributed/utils.py
fegin: Can we add a comment describing what will change, like in what RL setting? This also means that users cannot use fake backend mode anymore.

acisseJZhong: makes sense, will add a warning. But I think this also means user initialize distributed env somewhere else with their own settings.

torchtitan/components/lr_scheduler.py
wwwjn: https://github.com/pytorch/torchtitan/pull/1284 Here's more context why we converge to current behavior (red line) in this PR description. Can you explain in similar graph removing this `current_step += 1` would affect the shape of learning rate scheduler

acisseJZhong: thanks! ~~turns out i don't need this change and can still achieve numeric parity. This was added during my debugging journey.~~

sorry I take that back, the change seems necessary, only with this change lr schedule exactly matches with verl's [fsdp impl](https://github.com/verl-project/verl/blob/d987199906f09ba53139df13e4528b2d575ec4ce/verl/utils/torch_functional.py#L745), and therefore loss exactly matches. Otherwise torchtitan lr is always one step ahead of FSDP lr(FSDP uses 0 index but titan use 1 index). 

Image

wondering what's the reason for this +1 adjustment in the beginning?

acisseJZhong: sync with @wwwjn offline, likely moving from 1 indexed to 0 indexed will not change the shape of lr schedule, it only shift by 1 step. But I will do more thorough testing to confirm the new lr schedule.

tianyu-l: I understand that lr schedule mismatch would cause loss curve mismatch, and aligning with verl fsdp would show stronger numerical alignment.

What I don't understand why you think verl is the golden standard. Do you think we could either
- not changing either side and bear with this difference
- change verl side LR scheduling

If not, could you give me an argument that it is us who needs to change? Thanks!

torchtitan/train.py
```diff
@@ -548,6 +568,33 @@ def forward_backward_step(
         # The returned loss here is local SUM loss / global_valid_tokens
         return loss
 
+    def forward_step(
```
tianyu-l: we might eventually need this as we build RL by ourselves

but for now, can we put this in verl engine?

torchtitan/models/attention.py
```diff
@@ -234,6 +243,77 @@ def document_mask(
...
+    Returns:
+        A tensor of shape [batch, seq] where each unique integer indicates
+        tokens belonging to the same document/sequence.
+    """
+    first_dummy_value = positions[:, :1] - 1
+    position_diff = torch.diff(positions, prepend=first_dummy_value, dim=-1)
+    return (position_diff != 1).cumsum(-1)  # [batch, seq]
+
+
+def create_sdpa_document_causal_mask(positions: torch.Tensor) -> torch.Tensor:
```
tianyu-l: I think in general torchtitan wouldn't advocate using sdpa with is_causal=False. The main reason is that this mask is too memory-heavy to work with. cc @drisspg @fegin to confirm.

The general recommendation is to work with varlen attention and flex, if one needs block causal capability.

torchtitan/models/llama3/model/model.py
```diff
@@ -531,6 +532,9 @@ def get_attention_masks(
         extra_inputs: dict[str, torch.Tensor] | None = None,
     ) -> AttentionMasksType:
         match self.model_args.attn_type:
+            case "sdpa":
+                assert extra_inputs is not None and "positions" in extra_inputs
```
tianyu-l: `positions` won't be always available -- wouldn't this break torchtitan in general?

but anyways, we probably don't want to do sdpa with positions, see the other comment

torchtitan/models/attention.py
```diff
@@ -234,6 +243,77 @@ def document_mask(
...
+    # Combine: document mask AND causal mask
+    return doc_mask & causal_mask.unsqueeze(0).unsqueeze(0)
+
+
+def get_document_mask_mod_from_positions(
+    positions: torch.Tensor,
+) -> _mask_mod_signature:
+    """Creates a document mask from position_ids for flex attention.
+
+    Detects document boundaries where position_ids reset (diff != 1).
```
tianyu-l: Let's be careful here.

Previously, in torchtitan, `position_ids` is used to indexing rope cache (for CP to work with sharded sequence and replicate rope cache), in a sense that it's decoupled from attention block mask decision (which is determined by `eos_id` in inputs today).

According to what you are trying to do, it sounds like verl / HF is coupling `position_ids` with mask creation? IIUC it only works for block-causal assuming the block boundary is given by `positions_ids`. In particular, if masking is more complicated than block causal, e.g. in multimodal training, where the attention is bidirectional within an image, it can't be expressed using `position_ids`?

So, now there will be two ways to create block_causal masking, one by `eos_id` and the other by `positions_id`, is that correct?

cc @fegin do you think it's fine to create another `model_args.attn_mask_type` for this?

torchtitan/models/attention.py
acisseJZhong: yes @fegin suggested supporting document mask only for flex and varlen attention, I will update the PR

drisspg: +1

torchtitan/models/attention.py
acisseJZhong: Verl's fsdp engine uses transformers model, whose forward takes in `position_ids` and generate block causal mask based on that. Therefore, apart from `get_document_mask_mod` which generate `block_mask` from `eos_id`, I add another `get_document_mask_mod_from_positions`

drisspg: should we make this generic to `<separator>` and this separator can be eos_id or position_ids?

torchtitan/components/lr_scheduler.py
acisseJZhong: yes this is a valid question, apologies for not doing enough research on this. I think we should aim for exact same lr schedule and loss(although loss difference is not large; still same decreasing trend). 

Image
the pink one is with titan's original lr schedule.

I asked agent to do some research https://docs.google.com/document/d/1YiFUvIa_JqTYpBd2Xj7ReH3Bw6wS07nKldycBX--uVE/edit?usp=sharing and most frameworks use 0 based index(current_step starts from 0). It would be great if titan also switch to 0 based if it doesn't cause other issues? But I am also fine if we want to preserve the difference.

torchtitan/models/attention.py
tianyu-l: @drisspg position_ids is not a separator. You need a different function to translate it to a mask mod.

torchtitan/components/lr_scheduler.py
tianyu-l: Having learning rate 0 at the first update sounds ... pointless?

tests/unit_tests/test_lr_scheduler.py
```diff
@@ -88,16 +87,16 @@ def test_linear_warmup_decay(self):
...
-            1.0,  # Step 1: 100% of max LR (warmup complete)
-            1.0,  # Step 2: We maunally added step of stable phase, to prevent LR from dropping to 0 at last step
-            7.0 / 8.0,  # Step 3: 7/8 of max LR
-            6.0 / 8.0,  # Step 4: 3/4 of max LR
-            5.0 / 8.0,  # Step 5: 5/8 of max LR
-            4.0 / 8.0,  # Step 6: 1/2 of max LR
-            3.0 / 8.0,  # Step 7: 3/8 of max LR
-            2.0 / 8.0,  # Step 8: 1/4 of max LR
-            1.0 / 8.0,  # Step 9: 1/8 of max LR
+            0.0,  # Step 0: 0% of max LR (warmup start)
```
wwwjn: +1, this step is not updating weights at all ,the forward and backward computation are wasted.

torchtitan/components/lr_scheduler.py
wwwjn: We don't need to align with other frameworks if we don't understand their rationale. +1 that we don't want to have any step with lr = 0 (basically waste foward and backward computation). Would it be possible to make "offset 1" operation in verl?

torchtitan/models/deepseek_v3/model/model.py
```diff
@@ -465,7 +465,16 @@ def get_attention_masks(
             case "block_causal":
                 B = input_batch.shape[0]
                 assert tokenizer.eos_id is not None
-                mask_mods.append(get_document_mask_mod(input_batch, tokenizer.eos_id))
+                mask_mods.append(
+                    get_document_mask_mod(
+                        input_ids=input_batch, eos_id=tokenizer.eos_id
+                    )
+                )
+            case "position_block_causal":
```
wwwjn: So in "position block causal", the EOS id is not used to create block mask, and the user might accidentally use both, and don't know which one is actually taking effect?

If they are both referring to "block_causal", one possible way to do this is let get_document_mask_mod takes both EOS id and positions, and specifically add warning when both is not None and let user know which field we are using to create the "block_causal" mask

torchtitan/models/attention.py
```diff
@@ -209,22 +214,44 @@ def _causal_mask(
...
+    Raises:
+        ValueError: If neither or both separator methods are provided.
     """
-    # batch is [b, s, h, d] shape
-    eos_mask = batch == eos_id
-    eos_mask[:, -1] = True
-    cumulative_mask = torch.cumsum(torch.where(eos_mask, 1, 0), dim=1)
-    sequence_indices = torch.zeros_like(cumulative_mask, dtype=torch.int32)
-    sequence_indices[:, 1:] = cumulative_mask[:, :-1]
+    if positions is not None:
```
wwwjn: Would this algorithm work for inference time continuous batching, with mixed prefix and decode request? eg, the position could be "[4,5,6,0,1,2,3,7,8,9,10]"

torchtitan/models/deepseek_v3/model/model.py
wwwjn: I'm curious about knowing more about the actual data format (the Trajectory) that the generator passed to trainer. For the following field:  
- Prompt + Completion: Does this field has EOS in it? If so, can we use EOS id instead of position_id?

And for trajectory, do they do padding or packing? If padding, would `position_id` field also be padded? If yes, does the mask_mod algorithm handled padded field correctly?

torchtitan/models/attention.py
wwwjn: And does it handled padding correctly?

torchtitan/components/lr_scheduler.py
acisseJZhong: verl have this [`init_lr_ratio`](https://github.com/verl-project/verl/blob/3df21a68edcfc08ab1f94adabf32ff9db68c1561/verl/utils/torch_functional.py#L742) so it's lr schedule is 0-indexed with configurable init. If `init_lr_ratio` is not set(which is the default case), then first step lr will be 0. 

[Megatron](https://github.com/NVIDIA/Megatron-LM/blob/c072f894da962453bc91a6df62a36073a1422e87/megatron/core/optimizer_param_scheduler.py#L208), [deepspeed](https://github.com/deepspeedai/DeepSpeed/blob/a44fb58f134afc5399b29c154a5502e14272774c/deepspeed/runtime/lr_schedules.py#L693) also adopts a similar 0 index with configurable min_lr(default to 0)

We can also keep  the lr schedule different, as offset by 1 step likely will not affect model quality? wdyt?

torchtitan/models/deepseek_v3/model/model.py
acisseJZhong: the user would have to specify using `block_causal`(with eos id) or `position_block_causal`(using positions); they can't use both at the same time

acisseJZhong: > Prompt + Completion: Does this field has EOS in it? If so, can we use EOS id instead of position_id?

I checked the input prompt + completion does has EOS in it, but it's not corresponding to positions. Looks like there is one more new line token after EOS for each sample.

torchtitan/models/attention.py
acisseJZhong: yes it detects sequence boundary through `position_diff!=1` so it will give `[0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2]`. 

it will not handle padding tho(each padding marked as separate sequence), as we don't expect any padding, and use packing.

torchtitan/models/attention.py
drisspg: Yeah fair one could also probably just have a base `doc_mask(, segments)` and it does the check for equal segments but I dont think it adds to much value since you still need the find_packed_sequence_indices

torchtitan/train.py
```diff
@@ -128,11 +128,15 @@ def __init__(self, job_config: JobConfig):
...
-            tokenizer=self.tokenizer,
-            job_config=job_config,
+        self.dataloader = (
+            self.train_spec.build_dataloader_fn(
+                dp_world_size=batch_degree,
+                dp_rank=batch_rank,
+                tokenizer=self.tokenizer,
+                job_config=job_config,
+            )
+            if self.train_spec.build_dataloader_fn is not None
```
tianyu-l, line 138: according to https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/train_spec.py#L51, it can't be None

I'm OK with type change, but you'll need to assert not None in torchtitan before it's used?

acisseJZhong: Since we are using verl's dataloader, i don't want to initialize titan's dataloader(otherwise I will encounter loading c4 dataset error). I did a hack here https://github.com/verl-project/verl/pull/5051/changes#diff-f658afe18d14b480f4067f7544fbdb0ef6962a20ef3b5f5d0c709ae31e91809dR101

torchtitan/train.py
```diff
@@ -128,11 +128,15 @@ def __init__(self, job_config: JobConfig):
             else None
         )
 
-        self.dataloader = self.train_spec.build_dataloader_fn(
-            dp_world_size=batch_degree,
-            dp_rank=batch_rank,
-            tokenizer=self.tokenizer,
-            job_config=job_config,
+        self.dataloader = (
```
tianyu-l, line 131: please assert on the dataloader before feeding into batch_generator.

tianyu-l (approved):

tianyu-l:
@acisseJZhong sorry https://github.com/pytorch/torchtitan/pull/2386 reverts your change because of explicit typing requirement, but I think I have an alternative solution. Let's chat.

PR #2339, opened 2026-02-06
Add run-to-run determinism testing to H100 CI

xmfan:
Add run-to-run determinism testing to H100 CI

This adds automatic run-to-run determinism verification for H100 integration tests. Tests marked with `determinism_test=True` will run twice with identical configuration and deterministic flags, then compare losses to ensure they match exactly.

The core loss extraction logic is factored into `torchtitan/tools/loss_utils.py` and shared between the integration test runner and the existing `loss_compare.py` script. The scripts directory is now a package to enable clean imports via `python -m scripts.loss_compare`.

The Float8 and HSDP+CP+compile+Float8 tests in the H100 suite are enabled for determinism testing (CUDA only).

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,10 @@
+# Commit messages
+
+Don't commit unless the user explicitly asks you to.
+
+When writing a commit message, don't make a bullet list of the individual
+changes. Instead, if the PR is large, explain the order to review changes
+(e.g., the logical progression), or if it's short just omit the bullet list
+entirely.
+
+Disclose that the PR was authored with Claude.
--- a/tests/integration_tests/__init__.py
+++ b/tests/integration_tests/__init__.py
@@ -24,6 +24,7 @@ class OverrideDefinitions:
     ngpu: int = 4
     disabled: bool = False
     skip_rocm_test: bool = False
+    determinism_test: bool = False  # Run twice and verify losses are identical
 
     def __repr__(self):
         return self.test_descr
--- a/torchtitan/tools/loss_utils.py
+++ b/torchtitan/tools/loss_utils.py
@@ -1,0 +1,90 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Shared utilities for loss extraction and comparison.
+
+This module provides common functionality used by both:
+- scripts/loss_compare.py (CLI tool for comparing losses across commits)
+- tests/integration_tests/run_tests.py (integration test runner)
+"""
+
+import re
+
+
+def extract_losses_from_log(log_file: str) -> dict[int, float]:
+    """Extract step and loss pairs from a training log file.
+
+    Parses log lines matching the pattern: "step: N loss: X.XXXX"
+    Handles ANSI escape codes that may be present in colored terminal output.
+
+    Args:
+        log_file: Path to the training log file
+
+    Returns:
+        Dictionary mapping step numbers to loss values
+    """
+    losses = {}
+    step_loss_pattern = re.compile(r"step:\s*(\d+)\s*loss:\s*(\d+\.\d+)")
+    ansi_escape = re.compile(r"\x1b\[[0-9;]*m")
+
+    with open(log_file, "r") as f:
+        for line in f:
+            # Strip ANSI codes before matching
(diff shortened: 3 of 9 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,10 @@
+# Commit messages
+
+Don't commit unless the user explicitly asks you to.
+
+When writing a commit message, don't make a bullet list of the individual
+changes. Instead, if the PR is large, explain the order to review changes
+(e.g., the logical progression), or if it's short just omit the bullet list
+entirely.
+
+Disclose that the PR was authored with Claude.
```
xmfan, line 10: copied over from pytorch's claude md

tests/integration_tests/__init__.py
```diff
@@ -24,6 +24,7 @@ class OverrideDefinitions:
     ngpu: int = 4
     disabled: bool = False
     skip_rocm_test: bool = False
+    determinism_test: bool = False  # Run twice and verify losses are identical
```
tianyu-l, line 27: The point is not only about being deterministic, but also not changing before vs. after
- pytorch nightly updates
- user commits

Is it correct that this PR doesn't address such issues?

xmfan: This pr just makes sure that when you run the same command twice, it produces the same outputs. by adding this to PR time CI, you would run H100 CI twice on each PR, both against the same pytorch nightly.

wwwjn: So this test only guards the deterministic is setup correctly and working correctly, right? I think if we make sure the loss doesn't change before vs. after (pytorch nightly, and user commits), it already covers the deterministic check:
- If it's not run-to-run deterministic, it's impossible to achieve identical loss before and after

CLAUDE.md
wwwjn: With this file, we can directly ask Claude code to create a PR for us?

torchtitan/tools/loss_utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+This module provides common functionality used by both:
+- scripts/loss_compare.py (CLI tool for comparing losses across commits)
+- tests/integration_tests/run_tests.py (integration test runner)
+"""
+
+import re
+
+
+def extract_losses_from_log(log_file: str) -> dict[int, float]:
```
wwwjn, line 18: What are other use cases for this function to put it in an utils file?

tests/integration_tests/__init__.py
xmfan: Can you expand more on the setup for "loss doesn't change before vs. after"?

The existing tests that I see only cover if the first run on a process always matches the expected loss, not whether the same process will keep producing the same loss (which is what you need when you develop locally).

torchtitan/tools/loss_utils.py
xmfan: wdym? this pr uses it in 2 places. one of them is in tests

CLAUDE.md
xmfan: you can always do that. this just makes claude disclose that the PR was co-authored with Claude in the description.

tests/integration_tests/__init__.py
wwwjn: > if the first run on a process always matches the expected loss 

My thought is that: the expected loss comes from a deterministic run (we see deterministic seed, and the use deterministic algorithm). The "first run" here you are referring to, is also a deterministic run. If the deterministic algorithm does not work , or any missing deterministic setting, these 2 runs is impossible to have identical loss, as randomness will make the loss difference.

In this sense, the current test "first run on a process always matches the expected loss" already covers 1) deterministic and 2) any potential changes from user commit and pytorch nightly. It's a combined effect of both.

xmfan: some components persist cache on the machine, like compile, and maybe some other modules. the second run has a warm cache and runs different code paths than the first run. for the compile case, you can always run with TORCHINDUCTOR_FORCE_DISABLE_CACHES, and idk what would be the solution for others. we don't need to land this if CI is too constrained.

wwwjn: > some components persist cache on the machine, like compile, and maybe some other modules. 

That's a valid reason to run 2 consecutive runs, but the current loss compare script is testing on non-compile model. And I feel like the correctness of cached path should also be guaranteed by finer-granularity unit test

PR #2340, opened 2026-02-07
Maintain same LR schedule for early stop debug runs 

acisseJZhong:
addressed user request  #2326 to debug training runs with early stop steps but maintain the original lr schedule. 

**Test:**
```
[lr_scheduler]
warmup_steps = 2 
# Users need to explicitly add `schedule_steps` to overwrite; default `job_config.training.steps`.
total_steps = 1000
```
```
CONFIG_FILE="./torchtitan/models/qwen3/train_configs/qwen3_0.6b.toml" ./run_train.sh
```

### DIFF
--- a/torchtitan/components/lr_scheduler.py
+++ b/torchtitan/components/lr_scheduler.py
@@ -105,7 +105,7 @@ def build_lr_schedulers(
         training_steps (int): The total number of training steps.
     """
     # Use total_steps from config if set, otherwise fall back to training_steps
-    schedule_steps = (
+    total_steps = (
         lr_scheduler_config.total_steps
         if lr_scheduler_config.total_steps is not None
         else training_steps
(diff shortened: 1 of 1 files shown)


torchtitan/config/job_config.py
```diff
@@ -172,6 +172,14 @@ class LRScheduler:
     Steps for lr scheduler warmup, normally 1/5 of --training.steps
     """
 
+    schedule_steps: int | None = None
```
tianyu-l: `schedule_steps` in `Scheduler` sounds redundant, suggesting
```suggestion
    total_steps: int | None = None
```

torchtitan/train.py
```diff
@@ -278,8 +278,13 @@ def __init__(self, job_config: JobConfig):
         self.optimizers = self.train_spec.build_optimizers_fn(
             self.model_parts, job_config.optimizer, parallel_dims, self.ft_manager
         )
+        lr_schedule_steps = (
```
tianyu-l: maybe put this in `build_lr_schedulers_fn` instead of train.py.

torchtitan/components/lr_scheduler.py
```diff
@@ -104,28 +104,35 @@ def build_lr_schedulers(
         lr_scheduler_config (LRSchedulerConfig): The lr scheduler config.
         training_steps (int): The total number of training steps.
     """
+    # Use total_steps from config if set, otherwise fall back to training_steps
+    schedule_steps = (
```
tianyu-l: similarly, schedule steps in lr_scheduler.py sounds redundant, can we use total_steps? I think it's unambiguous in this file.

tianyu-l (approved):
one more nit comment

torchtitan/components/lr_scheduler.py
acisseJZhong: sure, I personally think `schedule_steps` is more intuitive. But yeah it's better we unify to what's in the config.

Issue #2269, opened 2026-01-22
Numerical instability in truncated normal initialization with bfloat16


francesco-bertolotti:
### Bug description

I am experiencing numerical instability issues in training with qwen3. I tracked down the issue to the initialization of the model. In particular, to the usage of `torch.nn.init.trunc_normal_`. The issue is illustrated in the following code snippet. It is apparent that this function does not produce the intended behavior and it is skewed heavily toward the argument of `a`, in this case `-2`. I see that this function has an history of being problematic (see https://github.com/pytorch/pytorch/issues/155588 and https://github.com/pytorch/pytorch/issues/145498)

```python
In [20]: import torch
    ...: x = torch.zeros(10,500, dtype=torch.bfloat16)
    ...: torch.nn.init.trunc_normal_(x, mean=0, std=0.02, a=-2, b=2)
    ...: print("min", x.min(-1)[0])
    ...: print("max", x.max(-1)[0])
    ...: print("mean", x.mean(-1)[0])
    ...: print("sum", x.sum(-1)[0])
```
```
min tensor([-2.0000, -2.0000, -2.0000, -2.0000, -2.0000, -0.0532, -2.0000, -2.0000, -2.0000, -2.0000], dtype=torch.bfloat16)
max tensor([0.0532, 0.0532, 0.0532, 0.0532, 0.0532, 0.0532, 0.0532, 0.0532, 0.0532, 0.0483], dtype=torch.bfloat16)
mean tensor(-0.0084, dtype=torch.bfloat16)
sum tensor(-4.1875, dtype=torch.bfloat16)
```

My current workaround is to cast to float32, use truncated_init_, and cast back to bfloat16. This is the related code-snippet

```python
@torch.no_grad()
def trunc_normal_(tensor, mean=0.0, std=1.0, a=-2.0, b=2.0):
    orig_dtype = tensor.dtype

    if isinstance(tensor, torch.distributed.tensor.DTensor):
        # Convert DTensor -> float32 DTensor
        tensor_fp32 = tensor.to(dtype=torch.float32)

        # Local init (safe because trunc_normal_ is elementwise)
        torch.nn.init.trunc_normal_(tensor_fp32._local_tensor, mean=mean, std=std, a=a, b=b)

        # Cast back to original dtype and copy into original tensor
        tensor.copy_(tensor_fp32.to(dtype=orig_dtype))
    else:
        # Regular Tensor
        tensor_fp32 = tensor.to(dtype=torch.float32)
[...]

Issue #1879, opened 2025-10-15
[qwen3] Loss is extremely hight when initialized with seed checkpoint


wwwjn:
### Bug description

The seed checkpoint was initialized on CPU with following command, and the high loss is reproducible with the seed checkpoint created in following command:
```
NGPU=1 CONFIG_FILE="./torchtitan/models/qwen3/train_configs/qwen3_0.6b.toml" ./run_train.sh --checkpoint.enable --checkpoint.create_seed_checkpoint --parallelism.data_parallel_replicate_degree 1 --parallelism.data_parallel_shard_degree 1 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1
```

With the initialized checkpoint, run job with FSDP=2 and TP=2, the loss are extremely high

<img width="3094" height="670" alt="Image" src="https://github.com/user-attachments/assets/503768fa-cc3a-4ee4-be67-d89b96359a6e" /> 

### Versions

torchtitan main branch [6bccdb6](https://github.com/pytorch/torchtitan/commit/6bccdb652847e6fdcc259092d4b057146cd343ef)

rakkit:
emm. I just noticed there is no `model.safetensors.index.json` file in 0.6b's [repo](https://huggingface.co/Qwen/Qwen3-0.6B), so torchtitan will not read the HF checkpoint at all.   

but the problem also appears for 1.7b model

tianyu-l:
@rakkit 
> so torchtitan will not read the HF checkpoint at all.

Are you sure? @ankitageorge told me if there's no `.index.json`, everything (the only file) will be read.

ankitageorge:
Yes, at least from the DCP side, we can read checkpoints without index.json file without any issues

> [@rakkit](https://github.com/rakkit)
> 
> > so torchtitan will not read the HF checkpoint at all.
> 
> Are you sure? [@ankitageorge](https://github.com/ankitageorge) told me if there's no `.index.json`, everything (the only file) will be read.

rakkit:
[...]

PR #2342, opened 2026-02-08
added custom trunc_normal

francesco-bertolotti:
This PR addresses https://github.com/pytorch/torchtitan/issues/2269

Briefly, there is a numerical instability in `torch.nn.init.trunc_normal_` that causes an abnormal number of left bounds to appear in the weights.

I consistently swapped all the usage of `torch.nn.init.trunc_normal_` with a custom implementation of `trunc_normal_` located in `torchtitan/models/utils.py`

___

I have done two other fixes that would require some attention that do not have much to do with the `trunc_normal_` but they felt right.

1. Qwen3 was not initializing the output weights when `enable_weight_tying=True`. This would mean that embedding initialization would have been used for the output weights, which would have caused the loss to skyrocket past 500.

2. GPTOSS was calling `init_weights` in its `__init__` method, which does cause some errors.

3. I did not swap the initialization in the `experiments` folder, but I can easily add a commit changing those too.

---
Here I have some debug runs with associated losses:

deepseek
llama3
llama4
qwen3_moe
gptoss
---
Since using debug models do not picture the difference that using the new `trunc_normal_` does, I have a longer run (1000 steps) on the [synth](https://huggingface.co/datasets/PleIAs/SYNTH) dataset using Qwen 4B. It appears that this instability has a compounding factor wrt model size.

qwen3-4b

### DIFF
--- a/torchtitan/models/qwen3/model/model.py
+++ b/torchtitan/models/qwen3/model/model.py
@@ -24,8 +24,8 @@ from torchtitan.models.attention import (
     VarlenAttentionWrapper,
     VarlenMetadata,
 )
-from torchtitan.models.utils import trunc_normal_
 from torchtitan.models.moe import MoE
+from torchtitan.models.utils import trunc_normal_
 from torchtitan.protocols.model import AttentionMasksType
 from torchtitan.protocols.train_spec import ModelProtocol
 
--- a/torchtitan/experiments/rl/vllm_compat/models/qwen3/model_vllm_compat.py
+++ b/torchtitan/experiments/rl/vllm_compat/models/qwen3/model_vllm_compat.py
@@ -20,9 +20,9 @@ from torchtitan.experiments.rl.vllm_compat.batch_invariant_backward import (
 
 # Import from main torchtitan
 from torchtitan.models.qwen3.model.args import Qwen3ModelArgs
+from torchtitan.models.utils import trunc_normal_
 from torchtitan.protocols.model import AttentionMasksType
 from torchtitan.protocols.train_spec import ModelProtocol
-from torchtitan.models.utils import trunc_normal_
 
 # Import from local experiment's models
 from ..attention import VLLMCompatibleFlashAttention
--- a/torchtitan/experiments/transformers_modeling_backend/model/model.py
+++ b/torchtitan/experiments/transformers_modeling_backend/model/model.py
@@ -10,10 +10,10 @@ import math
 import torch
 from torch import nn
 from torch.nn import init
+from torchtitan.models.utils import trunc_normal_
 from torchtitan.tools.logging import logger
 from transformers.configuration_utils import PretrainedConfig
 from transformers.modeling_utils import PreTrainedModel
-from torchtitan.models.utils import trunc_normal_
 
 from .args import HFTransformerModelArgs
 
--- a/torchtitan/experiments/vlm/model/siglip2.py
+++ b/torchtitan/experiments/vlm/model/siglip2.py
@@ -17,8 +17,9 @@ from torchtitan.models.attention import (
     get_causal_mask_mod,
     get_document_mask_mod,
 )
-from torchtitan.protocols.model import AttentionMasksType
 from torchtitan.models.utils import trunc_normal_
+from torchtitan.protocols.model import AttentionMasksType
+
 from .args import Siglip2ModelArgs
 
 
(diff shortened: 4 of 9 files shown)


torchtitan/models/qwen3/model/model.py
```diff
@@ -498,16 +499,16 @@ def init_weights(
...
-            if self.output is not None:
-                nn.init.trunc_normal_(
-                    self.output.weight,
-                    mean=0.0,
-                    std=final_out_std,
-                    a=-cutoff_factor * final_out_std,
-                    b=cutoff_factor * final_out_std,
-                )
+
+        # If weight tying is enabled, we don't need to initialize the output layer
```
tianyu-l: > Qwen3 was not initializing the output weights when enable_weight_tying=True. This would mean that embedding initialization would have been used for the output weights, which would have caused the loss to skyrocket past 500.

To clarify, do you mean we should override embedding init with this output weight init? Why one direction of override is better than the other direction?

francesco-bertolotti: The problem is that the output initialization should have a small std, otherwise we will get high logits and a high loss consequently. If the weight are tied we should prioritize the output layer initialization. At the momement runnning 
```bash
NGPU=1 CONFIG_FILE="./torchtitan/models/qwen3/train_configs/qwen3_0.6b.toml" ./run_train.sh --training.steps 100 --training.seq_len 256 --compile.no-enable --training.dtype bfloat16 --metrics.enable-tensorboard --job.dump-folder "./outputs/qwen3"
```

yield

```text
[rank0]:[titan] 2026-02-09 09:03:01,409 - root - INFO - step:  1  loss: 127.88081  grad_norm: 42.2500  memory:  4.48GiB(28.75%)  tps: 267  tflops: 0.75  mfu: 0.24%
[rank0]:[titan] 2026-02-09 09:03:01,696 - root - INFO - step:  2  loss: 124.81187  grad_norm: 66.5000  memory:  6.21GiB(39.92%)  tps: 3,576  tflops: 10.08  mfu: 3.23%
[rank0]:[titan] 2026-02-09 09:03:01,905 - root - INFO - step:  3  loss: 119.61317  grad_norm: 43.0000  memory:  6.21GiB(39.92%)  tps: 4,909  tflops: 13.84  mfu: 4.44%
[rank0]:[titan] 2026-02-09 09:03:02,110 - root - INFO - step:  4  loss: 119.98587  grad_norm: 238.0000  memory:  6.21GiB(39.92%)  tps: 5,006  tflops: 14.11  mfu: 4.52%
[rank0]:[titan] 2026-02-09 09:03:02,314 - root - INFO - step:  5  loss: 115.53140  grad_norm: 29.8750  memory:  6.21GiB(39.92%)  tps: 5,035  tflops: 14.19  mfu: 4.55%
```

francesco-bertolotti:
I didn’t swap the initialization in the `experiments` folder yet, but I can add a commit for that as well.

> any reason we don't? If not, I think we should.

No objection on my end — I just wanted to confirm before doing it.

---

> why it cause errors? we can remove those if so

In `gpt_oss`, `init_weights` is called inside `__init__`, which causes issues when using the custom `trunc_normal_`. Specifically, it triggers
`RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.`

This only happens for `gpt_oss`. Since no other model calls `init_weights` in its own `__init__`, I assumed weight initialization is expected to happen when the model is materialized in `train.py`.

If keeping `init_weights` inside `__init__` is required, an alternative would be to add `@torch.no_grad()` to the custom `trunc_normal_`.

torchtitan/models/qwen3/model/model.py
tianyu-l: I see. I guess this is the reason for https://github.com/pytorch/torchtitan/issues/1879? @wwwjn 
@francesco-bertolotti could you add a comment why we overriding?

tianyu-l:
@francesco-bertolotti 
> If keeping init_weights inside __init__ is required, an alternative would be to add @torch.no_grad() to the custom trunc_normal_.

No it's not required / expected. Let's just remove it.

francesco-bertolotti:
In the last commits I have:
- changed a bit the documentation for `trunc_normal_`.
- introduced a comment for the re-initialization of the output embeddings in qwen3
- runned `pre-commit` 
- changed init function in the experiments folder with the custom one.

tianyu-l:
@francesco-bertolotti could you help remove this line as well?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/simple_fsdp/deepseek_v3/model.py#L15

PR #2343, opened 2026-02-08
[DRAFT] Optimize MoE Routing via `torch.sort` Indices DType Injection

voidbag:
## [DRAFT] Optimize MoE Routing via `torch.sort` Indices DType Injection

> [!IMPORTANT]
> This PR is a **Draft** and is blocked by the merging of upstream **[pytorch/pytorch#170978](https://github.com/pytorch/pytorch/pull/170978)**.

### Description
This PR optimizes the **Mixture of Experts (MoE)** routing hot-paths within `deepep.py` and `moe.py`. By utilizing the new `out-variant` index `dtype` injection introduced in the upstream PyTorch PR, we can significantly reduce the memory overhead and latency of token-to-expert assignment.

HBM and DDR memory are premium resources in modern training clusters. Traditionally, `torch.argsort` defaults to `int64` indices, which is overkill for most expert routing scales. This patch allows us to use `int32` (and eventually smaller types), directly addressing the "memory wall" bottleneck in high-performance compute.

If PyTorch indexing supports every integer type indexing, MoE can enjoy full features of indices_dtype.

📊 **[Detailed Performance Benchmark Results & Latency Analysis](https://github.com/pytorch/pytorch/pull/170978#issuecomment-3865412916)**

---

###  Key Improvements
* **Memory Efficiency:** Reduces the memory footprint of routing sort indices by **50% (2x improvement)** by switching from `int64` to `int32`.
* **Index Compatibility:** Uses `int32` to ensure the resulting indices are immediately usable in advanced indexing operations (e.g., token shuffling) without requiring a costly `.to(torch.long)` cast.
* **Throughput:** Up to **2.24x speedup** and **3.33x memory footprint saving** by minimizing data movement across the memory bus.
* **Zero-Overhead Scaling:** Enables "free" performance gains by matching index precision to the actual requirements of the sort dimension.

---

### ✅ Status & Todo
- [x] Integrate `_indices_dtype_by_sort_size` logic into `torchtitan.distributed.deepep`.
- [x] Update `TokenReorderer` in `torchtitan.models.moe`.
- [ ] **Blocked by:** [pytorch/pytorch#170978](https://github.com/pytorch/pytorch/pull/170978)
- [ ] **optional** When sort is merged, topk can be modified to support indices_dtype in out-variant op.

---

### 💡 Implementation Note
By using the `out=` variant with a pre-allocated `int32` tensor, we avoid the implicit allocation of a 64-bit index buffer. This is particularly beneficial for the large-scale token shuffles used in `torchtitan` to ensure peak memory is kept to a minimum.

### DIFF
--- a/torchtitan/models/moe/moe.py
+++ b/torchtitan/models/moe/moe.py
@@ -356,6 +356,24 @@ class TokenChoiceTopKRouter(nn.Module):
         nn.init.trunc_normal_(self.gate.weight, mean=0.0, std=init_std)
 
 
+def _indices_dtype_by_sort_size(data: Tensor, sort_dim=-1) -> torch.dtype:
+    sort_size = data.size(sort_dim)
+    indices_dtype = torch.long
+    if sort_size - 1 <= torch.iinfo(torch.int8).max:
+        indices_dtype = torch.int8
+    elif sort_size - 1 <= torch.iinfo(torch.uint8).max:
+        indices_dtype = torch.uint8
+    elif sort_size - 1 <= torch.iinfo(torch.int16).max:
+        indices_dtype = torch.int16
+    elif sort_size - 1 <= torch.iinfo(torch.uint16).max:
+        indices_dtype = torch.uint16
+    elif sort_size - 1 <= torch.iinfo(torch.int32).max:
+        indices_dtype = torch.int32
+    else:
+        indices_dtype = torch.long
+    return indices_dtype
+
+
 # NOTE: the reason we make this a stateless module is to support
 #       expert_tensor_parallel_degree=1 with consistent TP/EP APIs.
 class TokenReorderer(nn.Module):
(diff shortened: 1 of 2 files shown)


torchtitan/models/moe/moe.py
```diff
@@ -356,6 +356,24 @@ def init_weights(self, init_std: float):
         nn.init.trunc_normal_(self.gate.weight, mean=0.0, std=init_std)
 
 
+def _indices_dtype_by_sort_size(data: Tensor, sort_dim=-1) -> torch.dtype:
```
tianyu-l, line 359: can this type of util also put into pytorch core, or embedded into argsort?

voidbag: Thanks for your comments! I actually explored integrating this into the standard argsort utility during the initial design of the aten sort optimizations. However, maintainer feedback at the time emphasized that modifying the signatures of fundamental operators like sort/argsort posed backward-compatibility risks.

To align with that upstream preference for API stability, I decided to use the out variant instead. It allows us to support dynamic indices effectively without the friction of a core signature change. I'm open to moving this into a shared utility if we find a way to do so!

[Updated]
There is the code snippet doing the function in case of down up cast
https://github.com/voidbag/pytorch/blob/76357cc5d9e02e5ac830712bbe348444224df937/aten/src/ATen/native/Sorting.cpp#L966-L975

PR #2346, opened 2026-02-09
[simple_fsdp] Use schedule_overlap_bucketing_from_inductor_configs for overlap_passes

IvanKobzarev:
Use schedule_overlap_bucketing_from_inductor_configs that picks default config values.

Previous impl uses default argument values which are not equal to config (E.g. bucketing is off by default).

We want bucketing by default for perf.


eellison (approved):

IvanKobzarev:
> Some CI failed, I don't think this PR breaks the CI, but I would recommend to resubmit the PR and trigger again later and ensure the CI is green.

Yes, I think this is connected with todays github infra failures. I will rebase to verify.

tianyu-l (approved):

PR #2347, opened 2026-02-09
[ci] Add DSv3 SimpleFSDP auto_bucketing to h100 ci jobs

IvanKobzarev:
Adding DSv3 SimpleFSDP + inductor auto_bucketing passes to H100 CI.

We need H100 to go through grouped_mm path.

Testing:
```
rm -rf /tmp/test_output && python -m tests.integration_tests.run_tests /tmp/test_output --test_suite h100 --test_name simplefsdp_deepseekv3_auto_bucketing --ngpu 8
```

Originally I wanted to add this CI to AutoParallel. But we can not do it as it runs on A100 and goes through non-groupedmm moe part.

### DIFF
--- a/tests/integration_tests/h100.py
+++ b/tests/integration_tests/h100.py
@@ -78,5 +78,24 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             "hsdp+cp+compile+float8",
             ngpu=8,
         ),
+        # Experimental, non-blocking: PRs can land if only this test fails
+        OverrideDefinitions(
+            [
+                [
+                    "--job.config_file ./torchtitan/models/deepseek_v3/train_configs/debug_model.toml",
+                    "--model.name simple_fsdp.deepseek_v3",
+                    "--parallelism.tensor_parallel_degree 1",
+                    "--parallelism.expert_parallel_degree 8",
+                    "--job.custom_config_module=torchtitan.experiments.simple_fsdp.job_config",
+                    "--compile.graph_passes auto_bucketing",
+                    "--activation_checkpoint.mode none",
+                    "--compile.backend inductor",
+                    "--compile.enable",
+                ]
+            ],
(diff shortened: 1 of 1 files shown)


tests/integration_tests/__init__.py
```diff
@@ -24,6 +24,7 @@ class OverrideDefinitions:
     ngpu: int = 4
     disabled: bool = False
     skip_rocm_test: bool = False
+    config_file: str | None = None
```
tianyu-l: do you have to add this change?

IvanKobzarev: yeah, I tried to avoid it, but by default it using different toml.

tianyu-l: could you take a look at https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/simple_fsdp/tests/integration_tests.py?

wwwjn: When we designed the last version, we explicitly avoid having a `config_file` field, but choose to override on top of the `base_config.toml`, as they share most of the fields between the base_config and other debug model configs.

IvanKobzarev: @tianyu-l  @wwwjn Yes, we can avoid it. Overrides work fine - I just put config file as argument in the list with other args, the latest override will have precedence. (updated the PR)

tianyu-l: SG. It seems CI failed for a different reason?

IvanKobzarev: yes, smth with nvlink sharp that I have seen without this change.

tianyu-l (approved):

IvanKobzarev:
> My only concern is that we said experiments like SimpleFSDP shouldn't block main development, but landing this will violate that.
> 
> Maybe good for now if you could add a TODO.

Yeah, agree - good point. I think we should mark CI jobs for experiments as "non blocking" or smth like this, that it is rather "yellow" signal than "red" for landing.

PR #2357, opened 2026-02-09
[SAC] Centralize selective AC policy and remove per-model op save lists

mori360:
### Summary

  - Remove layer-frequency selective activation checkpointing (`selective_ac_option` and `_layer_sac_count`) — per-op SAC is now the only selective mode
  - Centralize the op save list into `default_activation_checkpoint_policy()` in `activation_checkpoint.py`, removing duplicated `_op_sac_save_list` sets from per-model `parallelize.py` files (llama3, llama4, deepseek_v3, qwen3, gpt_oss, graph_trainer)
  - Remove the `op_sac_save_list` parameter from `apply_ac` — models no longer need to pass their own op sets
  - Build the centralized policy from `get_default_op_list()` (upstream PyTorch) plus explicit compute ops (SDPA, FlexAttention, inductor, varlen_attn) and communication ops (reduce_scatter, all_to_all, deepep, hybridep), with conditional resolution for optional dependencies
  - Use `@lru_cache` with `cache_hash` on the policy factory for dynamo recompilation avoidance and AOTAutograd cache compatibility
  - Add `--activation_checkpoint.mode full` to PP integration tests (`InterleavedZeroBubble`, `ZBVZeroBubble`, `PipelineScheduleMulti`) since they relied on layer_sac
  - Clean deepep imports, now we import from `torchtitan.distirbuted.deepep.deepep` or  `torchtitan.distirbuted.deepep.hybridep`, to keep them symmetrical.

  ### Test

Added `test_force_recompute_mm_fqns`: verifies that `per_op_sac_force_recompute_mm_shapes_by_fqns` controls exactly which matmuls are recomputed vs stored during backward. Uses a TorchDispatchMode tracker to count aten.mm calls per weight tensor


mori360:
## Discussion 
  
  ### 1. Shape-based matching in `per_op_sac_force_recompute_mm_shapes_by_fqns`                                              
                                                                                                                             
  The current implementation collects weight shapes from FQN-matched `nn.Linear` modules, then at policy time force-recomputes ANY `aten.mm` whose operand matches that shape. This is by design (documented in the config docstring), but it means unrelated linears that happen to share the same `(in_features, out_features)` will also be force-recomputed.

  For example, matching `"moe.router.gate"` (a `Linear(4096, 4096)`) will also affect any other `Linear(4096, 4096)` in the same transformer block (e.g. `wq`, `wk` if they share dimensions).
  
  ### 2. CUDA→CPU `_to_copy` MUST_SAVE for activation offloading

  The wrapped policy currently includes a special case that forces MUST_SAVE for `aten._to_copy` when copying from CUDA to CPU, protecting activation offloading scenarios where recomputing a device transfer is wasteful.
  
  Alternative approach is to ensure recomputed tensors stay on the same device as the original, avoiding the need for the MUST_SAVE special case entirely.

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,134 @@
...
     checkpoint_wrapper as ptd_checkpoint_wrapper,
 )
+from torch.utils.checkpoint import CheckpointPolicy
 
 from torchtitan.config import ActivationCheckpointConfig as ACConfig
 from torchtitan.tools.logging import logger
 
 
-_layer_sac_count = 0
+_PolicyFn = Callable[..., CheckpointPolicy]
```
tianyu-l: we don't need this alias?

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,134 @@
...
+
+    communication_intensive_ops = _resolve_ops(_COMM_OPS)
+
+    policy_fn = partial(
+        _sac_policy_fn,
+        compute_intensive_ops=compute_intensive_ops,
+        communication_intensive_ops=communication_intensive_ops,
+    )
+    # pyrefly: ignore [missing-attribute]
+    policy_fn.cache_hash = "default_activation_checkpoint_policy"
```
tianyu-l: > Use `@lru_cache` with `cache_hash` on the policy factory for dynamo recompilation avoidance and AOTAutograd cache compatibility

What's the reason of recompilation if we don't use `lru_cache`?

Why don't we need it before?

cc @xmfan to review this

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,134 @@
...
-
-    Returns:
-        nn.Module: The module with selective activation checkpointing applied.
-    """
-    from torch.utils.checkpoint import (
-        CheckpointPolicy,
-        create_selective_checkpoint_contexts,
-    )
+    """Apply per-op selective activation checkpointing to the module."""
+    from torch.utils.checkpoint import create_selective_checkpoint_contexts
```
tianyu-l: we can move this to file top

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -94,16 +144,17 @@ def _apply_op_sac(
...
 
+    base_policy = default_activation_checkpoint_policy()
     # Some backends (e.g. PrivateUse1) register aten.linear as a leaf op
     # instead of decomposing it into aten.mm, so we must handle both.
     mm_ops = (torch.ops.aten.mm.default, torch.ops.aten.linear.default)
 
-    def _get_custom_policy(meta):
-        def _custom_policy(ctx, func, *args, **kwargs):
+    def _create_wrapped_policy():
+        meta = {"forward_mm_count": 0, "recompute_mm_count": 0}
```
tianyu-l, line 123: Previously `meta` is defined within `selective_checkpointing_context_fn` outside `_get_custom_policy` ; now it is defined within `_create_wrapped_policy`. I feel they have the same effect, but would be good to confirm that there is no behavior change.

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,134 @@
...
-    )
+    """Apply per-op selective activation checkpointing to the module."""
+    from torch.utils.checkpoint import create_selective_checkpoint_contexts
 
-    # Collect weight shapes to force-recompute, stored as mm RHS shape
-    # (in_f, out_f). For aten.linear we transpose args[1].shape at lookup
-    # time to match, since linear's weight is (out_f, in_f).
     mm_recompute_shapes = set()
-    if len(ac_config.per_op_sac_force_recompute_mm_shapes_by_fqns) > 0:
+    recompute_fqns = ac_config.per_op_sac_force_recompute_mm_shapes_by_fqns
```
tianyu-l: to be consistent
```suggestion
    mm_recompute_fqns = ac_config.per_op_sac_force_recompute_mm_shapes_by_fqns
```

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -114,33 +165,31 @@ def _custom_policy(ctx, func, *args, **kwargs):
...
+                return CheckpointPolicy.PREFER_RECOMPUTE
+            return base_decision
 
-        return _custom_policy
+        return wrapped_policy
 
-    def selective_checkpointing_context_fn():
-        meta = defaultdict(int)
-        return create_selective_checkpoint_contexts(_get_custom_policy(meta))
+    def context_fn():
```
tianyu-l: don't need this extra function? you can set 
```
context_fn = create_selective_checkpoint_contexts(_create_wrapped_policy())
```
below

torchtitan/models/llama4/parallelize.py
```diff
@@ -149,14 +130,9 @@ def parallelize_llama(
         if comm_backend == "hybridep":
```
tianyu-l: remove

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -94,16 +144,17 @@ def _apply_op_sac(
...
-        )
 
+    base_policy = default_activation_checkpoint_policy()
     # Some backends (e.g. PrivateUse1) register aten.linear as a leaf op
     # instead of decomposing it into aten.mm, so we must handle both.
     mm_ops = (torch.ops.aten.mm.default, torch.ops.aten.linear.default)
 
-    def _get_custom_policy(meta):
-        def _custom_policy(ctx, func, *args, **kwargs):
+    def _create_wrapped_policy():
```
tianyu-l: why changing the name?

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -94,16 +144,17 @@ def _apply_op_sac(
...
+        meta = {"forward_mm_count": 0, "recompute_mm_count": 0}
+
+        def wrapped_policy(ctx, func, *args, **kwargs) -> CheckpointPolicy:
+            # Always save CUDA→CPU copies (activation offloading).
```
tianyu-l: comment is not correct -- this is indeed saving CUDA->CPU result to avoid recomputing, but not for activation offloading; instead, it is from MoE D2H sync to get all-to-all metadata.

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,134 @@
...
 from torchtitan.tools.logging import logger
 
 
-_layer_sac_count = 0
+_PolicyFn = Callable[..., CheckpointPolicy]
 
 
-def _apply_layer_sac(module: nn.Module, ac_config: ACConfig) -> nn.Module:
-    """Apply layer selective activation checkpointing to the module.
+def _sac_policy_fn(
```
tianyu-l: don't need to create a separate function for this simple logic

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -114,33 +165,31 @@ def _custom_policy(ctx, func, *args, **kwargs):
 
             mode = "recompute" if ctx.is_recompute else "forward"
             mm_count_key = f"{mode}_mm_count"
+
             if func in mm_ops:
                 weight_shape = args[1].shape
-                # linear weight is (out, in); normalize to (in, out) to match mm
```
tianyu-l, line 119: why removing this?

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,134 @@
...
+    (torch.ops, "deepep.dispatch.default"),
+    (torch.ops, "deepep.combine.default"),
+    # HybridEP (available when hybridep is installed)
+    (torch.ops, "hybridep.dispatch.default"),
+    (torch.ops, "hybridep.combine.default"),
+]
+
+
+@lru_cache()
+def default_activation_checkpoint_policy() -> _PolicyFn:
```
tianyu-l: It doesn't sounds more readable to have both `default_activation_checkpoint_policy` and use it in `_create_wrapped_policy`. Let's try to have just need one sac policy function.

You can have everything in one function and try to let claude clean up the logic.

torchtitan/models/llama4/parallelize.py
mori360: Shall we remove the hybridep import as well, take import deepep as default?

tianyu-l: you don't need any import here because you removed the code which requires import

mori360: We import deepep first, then apply_ac with the 
```
_COMM_OPS = [
    ...,
    (torch.ops, "deepep.dispatch.default"),
    (torch.ops, "deepep.combine.default"),
    (torch.ops, "hybridep.dispatch.default"),
    (torch.ops, "hybridep.combine.default"),
    ...]
```
`_COMM_OPS` list is declarative, it saves these if they exist and the import in parallelize.py is what makes them exist.

torchtitan/distributed/activation_checkpoint.py
mori360: In both cases, the factory is invoked once per forward by checkpoint_wrapper, so meta is created once per forward. No behavior change.

torchtitan/distributed/activation_checkpoint.py
mori360: Removed lru_cache, we can add it once needed.

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,105 @@
...
-        nn.Module: The module with selective activation checkpointing applied.
-    """
-    from torch.utils.checkpoint import (
-        CheckpointPolicy,
-        create_selective_checkpoint_contexts,
-    )
-
-    # Collect weight shapes to force-recompute, stored as mm RHS shape
-    # (in_f, out_f). For aten.linear we transpose args[1].shape at lookup
-    # time to match, since linear's weight is (out_f, in_f).
```
tianyu-l, line 78: don't delete this comment

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -94,16 +115,16 @@ def _apply_op_sac(
                 )
             out_f, in_f = submod.weight.shape
             mm_recompute_shapes.add((in_f, out_f))
-        logger.debug(
-            f"Selective op AC force recomputing mms with rhs shapes {mm_recompute_shapes}"
-        )
 
-    # Some backends (e.g. PrivateUse1) register aten.linear as a leaf op
-    # instead of decomposing it into aten.mm, so we must handle both.
```
tianyu-l, line 102: same

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -8,84 +8,105 @@
...
 from torchtitan.config import ActivationCheckpointConfig as ACConfig
 from torchtitan.tools.logging import logger
 
 
-_layer_sac_count = 0
-
-
-def _apply_layer_sac(module: nn.Module, ac_config: ACConfig) -> nn.Module:
-    """Apply layer selective activation checkpointing to the module.
+def _resolve_ops(op_specs: list) -> dict:
```
tianyu-l: make this an inline function in `_get_save_ops`

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -94,16 +115,16 @@ def _apply_op_sac(
                 )
             out_f, in_f = submod.weight.shape
             mm_recompute_shapes.add((in_f, out_f))
-        logger.debug(
-            f"Selective op AC force recomputing mms with rhs shapes {mm_recompute_shapes}"
-        )
 
-    # Some backends (e.g. PrivateUse1) register aten.linear as a leaf op
-    # instead of decomposing it into aten.mm, so we must handle both.
+    save_ops = _get_save_ops()
```
tianyu-l: nit: call this at the beginning in this function `_apply_op_sac`.

torchtitan/distributed/deepep/__init__.py
tianyu-l, line 1: can we remove this file?

PR #2358, opened 2026-02-09
[Bugfix] Fix bitwise determinism after vLLM SiluAndMul change

Lucaskabela:
## Summary
https://github.com/vllm-project/vllm/pull/32806 Changed the behavior of `SiluAndMul` to use `torch.compile` inside the custom op.  This causes a divergence in the vllm definition and torchtitan definition, causing the RL script to fail

We fix this by changing the implementation to call through to the kernel used in vLLM for equivalence

## Test Plan

```bash
VLLM_BATCH_INVARIANT=1 VLLM_FLASH_ATTN_VERSION=3 python -m torchtitan.experiments.rl.vllm_compat.simple_rl
```

### Before (Main)
```bash
  ⚠ vLLM-TorchTitan logprobs differ: 59/100 tokens
    Max delta: 3.650573e-01, Avg delta: 1.829216e-02
    vLLM logprobs:     ['-0.0642131940', '-0.1617958397', '-0.0011243457', '-0.0051660384', '-0.2546415329']
    TorchTitan logprobs: ['-0.0609663166', '-0.1456209123', '-0.0010027625', '-0.0048254938', '-0.2543271184']
```
### After
```bash
INFO 02-09 15:18:20 [llm.py:343] Supported tasks: ['generate']
✓ Created new vLLM engine
Adding requests: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 402.86it/s]
Processed prompts: 100%|████████████████████████████████████████████████████████| 80/80 [00:03<00:00, 22.59it/s, est. speed input: 1473.08 toks/s, output: 2259.32 toks/s]
Adding requests: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 495.83it/s]
Processed prompts: 100%|████████████████████████████████████████████████████████| 80/80 [00:03<00:00, 24.28it/s, est. speed input: 1583.37 toks/s, output: 2428.49 toks/s]
```

### DIFF
--- a/torchtitan/experiments/rl/vllm_compat/batch_invariant_backward.py
+++ b/torchtitan/experiments/rl/vllm_compat/batch_invariant_backward.py
@@ -69,7 +69,12 @@ class SiluAndMulFunction(Function):
         # vLLM custom ops require a config context to be set
         # Since these are parameter free we instantiate default config
         with set_current_vllm_config(VllmConfig()):
-            vllm_silu_and_mul = VLLMSiluAndMul()
+            # compile_native=False to avoid compiling the inner details of
+            # the CustomOp for now - NOTE: if we enable compile in vLLM in
+            # the future, we need to compile this op as well.
+            # see vllm-project/vllm#32806  for more details
+            vllm_silu_and_mul = VLLMSiluAndMul(compile_native=False)
+
         output = vllm_silu_and_mul(x)
 
         # Save for backward


Lucaskabela:
cc @wwwjn @tianyu-l @PaulZhang12

torchtitan/experiments/rl/vllm_compat/batch_invariant_backward.py
```diff
@@ -69,7 +69,7 @@ def forward(ctx, x):
         # vLLM custom ops require a config context to be set
         # Since these are parameter free we instantiate default config
         with set_current_vllm_config(VllmConfig()):
-            vllm_silu_and_mul = VLLMSiluAndMul()
+            vllm_silu_and_mul = VLLMSiluAndMul(compile_native=False)
```
wwwjn, line 76: Does `compile_native=False` means we don't use torch.compile inside the custom op? Can you add a line of comment to explain why this field is neede? Thanks!

Lucaskabela: Updated :)

acisseJZhong (approved):

torchtitan/experiments/rl/vllm_compat/batch_invariant_backward.py
wwwjn: n00b q: Is `vllm_compat` path using vllm's compile mechanism, or it applies compile  manually? Or compile is not yet enabled?

> NOTE: if we enable compile in vLLM in the future, we need to compile this op as well.

I'm a little bit confused here, and want to high-levely understand how we should enable compile properly. IIUC there should be 2 ways of applying compile: 
1. apply compile by ourself, like `apply_compile()` function, and send the compiled model to vllm (set vllm.compilation_config.level =0); 
2. Let vllm engine enable compile (vllm.compilation_config.level = 3), and decorate our model with decorator @support_torch_compile

Lucaskabela: Currently there is no compile enabled yet - however, we can compile either the vLLM model or (inclusive) the torch.Titan model. 

1. To compile the vLLM model, we would need `enforce_eager=False` then setting the vllm.compilation_config to our desired level of opimizations.  This won't affect the numerics of torch titan definition

2. To compile the torchtitan model, we would need to apply torch compile (there are some options for how we can do this, but we should in general start with aot_eager backend for numerics).  This will not affect the numerics of the vLLM definition

Hopefully that clarifies - we have two levers for compile which do not interact so we can (and should) see how we can leverage both

PR #2359, opened 2026-02-10
[Bugfix] Fix `simple_rl_multiprocess.py` to be runnable with recent vLLM version

Lucaskabela:
## Summary

We make a handful of QOL changes to make `simple_rl_multiprocess.py` work with recent vLLM. These are:

1) Update to `README.md` to add monarch requirement

2) `AttentionBackendEnum` forwarded in locations needed (similar to https://github.com/pytorch/torchtitan/pull/2219)

3) Wiring through `enable_gqa` support in vllm_compat attention - this requires reshaping when `q.shape[1]` != `k.shape[1]` and using an output tensor explicitly with the proper shape

With these changes, we are able to run the simple_rl_multiprocess script again

## Test Plan

```bash
```

### Before (Main)
```bash
  File "/home/lucaskabela/.conda/envs/vllm/lib/python3.12/asyncio/base_events.py", line 691, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/lucaskabela/torchtitan/torchtitan/experiments/rl/unified/simple_rl_multiprocess.py", line 66, in main
    init_batch_invariance()
TypeError: init_batch_invariance() missing 1 required positional argument: 'attention_backend'
```
And after patching with the enum:
```
  File "/home/lucaskabela/pytorch/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
  File "/home/lucaskabela/pytorch/torch/nn/modules/module.py", line 1790, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: VLLMCompatibleFlashAttention.forward() got an unexpected keyword argument 'enable_gqa'
```
### After 
```bash
Adding requests: 100%|█████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 2935.95it/s]
Processed prompts: 100%|███████████████| 40/40 [00:01<00:00, 27.61it/s, est. speed input: 193.29 toks/s, output: 552.25 toks/s]
[2026-02-09 16:11:59] INFO generator.py:446: [actor=<root>.<torchtitan.experiments.rl.unified.actors.generator.Generator generator{'gpus': 0/1}>] os.getpid()=300001 Generating finish generate (policy v0)...
[2026-02-09 16:11:59] INFO trainer.py:101: [actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 1/2}>] os.getpid()=300426 Trainer starts to train 0 on traj:
[2026-02-09 16:11:59] INFO trainer.py:101: [actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 0/2}>] os.getpid()=299553 Trainer starts to train 0 on traj:
NCCL version 2.28.9+cuda12.9
  ✓ vLLM-TorchTitan bitwise determinism verified: 20 tokens match exactly
  ✓ vLLM-TorchTitan bitwise determinism verified: 20 tokens match exactly
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,7 @@ Sessionx.vim
 
 # Vibe coding
 .claude
+
+# Experiment rl
+converted/
+models/
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,3 @@ Sessionx.vim
 
 # Vibe coding
 .claude
-
-# Experiments/rl artifacts
-/converted/
-/models/
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,6 @@ Sessionx.vim
 # Vibe coding
 .claude
 
-# Experiment rl
-converted/
-models/
+# Experiments/rl artifacts
+/converted/
+/models/
--- a/torchtitan/distributed/parallel_dims.py
+++ b/torchtitan/distributed/parallel_dims.py
@@ -121,7 +121,11 @@ class ParallelDims:
                     backend_override[name] = "fake"
 
             return world_mesh._unflatten(
-                0, dim_degrees, dim_names, backend_override=backend_override
+                0,
+                dim_degrees,
+                dim_names,
+                # pyrefly: ignore [bad-argument-type]
+                backend_override=backend_override,
             )
 
         logger.info(
--- a/torchtitan/distributed/parallel_dims.py
+++ b/torchtitan/distributed/parallel_dims.py
@@ -124,7 +124,6 @@ class ParallelDims:
                 0,
                 dim_degrees,
                 dim_names,
-                # pyrefly: ignore [bad-argument-type]
                 backend_override=backend_override,
             )
 
--- a/torchtitan/experiments/rl/unified/README.md
+++ b/torchtitan/experiments/rl/unified/README.md
@@ -14,9 +14,10 @@ The integration consists of two main components:
 ## Quick Start
 ### Prerequisites
 
-1. Install PyTorch nightly for torchtitan:
+1. Install PyTorch nightly & Monarch for torchtitan:
 ```
 pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall
+pip3 install torchmonarch
 ```
 
 
--- a/torchtitan/experiments/rl/unified/README.md
+++ b/torchtitan/experiments/rl/unified/README.md
@@ -15,9 +15,9 @@ The integration consists of two main components:
 ### Prerequisites
 
 1. Install PyTorch nightly & Monarch for torchtitan:
-```
-pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall
-pip3 install torchmonarch
+```bash
+uv pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall
+uv pip install torchmonarch
 ```
 
 
@@ -34,7 +34,7 @@ uv pip install --no-build-isolation -e .
 
 NOTE: If `flash_attn_varlen_func` hits error "torch.AcceleratorError: CUDA error: the provided PTX was compiled with an unsupported toolchain" during forward path, this is due to GPU driver version is not compatible with vLLM/PyTorch compiled version. Use the following command to recompile vLLM.
 
-```
+```bash
 # Set CUDA version environment variable
 export CUDA_HOME=/usr/local/cuda-12.4
 export PATH=/usr/local/cuda-12.4/bin:$PATH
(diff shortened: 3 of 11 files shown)


torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -99,7 +98,8 @@ def forward(
         output_flat = self.vllm_attn(q, k, v)
 
         # Output is (batch * seq_len, num_heads * head_dim), reshape to (batch, seq_len, num_heads, head_dim)
-        output = output_flat.view(batch_size, seq_len, num_heads, head_dim)
+        # Use self.num_heads and self.head_dim since vLLM Attention outputs based on its configured dimensions
+        output = output_flat.view(batch_size, seq_len, self.num_heads, self.head_dim)
```
acisseJZhong: are you suggesting `self.num_heads` might be different from `num_heads` here?

torchtitan/experiments/rl/unified/README.md
```diff
@@ -14,9 +14,10 @@ The integration consists of two main components:
 ## Quick Start
 ### Prerequisites
 
-1. Install PyTorch nightly for torchtitan:
+1. Install PyTorch nightly & Monarch for torchtitan:
 ```
 pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu126 --force-reinstall
+pip3 install torchmonarch
```
acisseJZhong: nit: can we update to `uv pip install` as the rest of this doc uses uv?

torchtitan/experiments/rl/vllm_compat/models/attention.py
```diff
@@ -122,6 +134,13 @@ def backward(ctx, grad_output):
                 out_t = out_batch.transpose(1, 2)
                 grad_out_t = grad_out_batch.transpose(1, 2)
 
+                # For GQA, we need to expand K/V to match Q's num_heads
```
acisseJZhong, line 151: Won't vllm handle GQA internally?

wwwjn: I think it's for backward function, and it's manually calculate the attention, instead of using vllm kernel here.

torchtitan/experiments/rl/unified/models/attention.py
wwwjn: +1, what's the difference between configured `self.num_heads` and `num_heads`

Lucaskabela: These are the same so this change is not necessary - however, given `self.num_heads` and `self.head_dim` are what is used to instantiate `vllm_attn`, I made this change for clarity.  It's not necessary though so I will remove it

torchtitan/experiments/rl/vllm_compat/models/attention.py
```diff
@@ -122,6 +148,13 @@ def backward(ctx, grad_output):
...
+                # Each KV head serves (num_heads // num_kv_heads) Q heads
+                if num_kv_heads != num_heads:
```
acisseJZhong: maybe use `num_heads > num_kv_heads` (similar [here](https://github.com/pytorch/torchtitan/pull/2259/changes#diff-89327c30617bfd44cb09fda573b5c508dd2aaddd9ec3291296fc3ec859eb4e5bR199)) and assert it can divide?

acisseJZhong (approved):
thanks for the fix!

torchtitan/distributed/parallel_dims.py
```diff
@@ -121,7 +121,11 @@ def unflatten_mesh(
                     backend_override[name] = "fake"
 
             return world_mesh._unflatten(
-                0, dim_degrees, dim_names, backend_override=backend_override
+                0,
```
wwwjn, line 124: In our current CI, even missing these pyrefly ignore, the CI still passed. Do we still need to add these pyrefly ignore?

.gitignore
```diff
@@ -45,3 +45,7 @@ Sessionx.vim
 
 # Vibe coding
 .claude
+
+# Experiments/rl artifacts
```
wwwjn: Should be added to .gitignore under experiments/rl folder

torchtitan/distributed/parallel_dims.py
Lucaskabela: There could be some mismatch between the configuration in CI and local - when I install pre-commit though these are blocking for me, resulting in extra friction. 

I am using the pyrefly version recommended (v0.45.1), so imagine I can't be the only person that experiences this friction

IMO adding these suppressions is harmless and a better dev experience so we should do it

.gitignore
Lucaskabela: Moving those to the subfolder won't work since following the README.md, the `models` and `converted` directories are populated at the top level (not in the `experiments/rl` subfolder)

These need to be top level if following the README.md

wwwjn: We can solely don't add these experiment artifacts in main .gitignore. Two rationale: 
1. the name seems confusing for user that don't run RL in torchtitan
2. Most of the code change will under torchtitan/torchtitan/, which is not automatically included when `git add`, these seems unnecessary to me

torchtitan/distributed/parallel_dims.py
wwwjn: I would recommend remove the unnecessary lines if the CI is not complaining. The CI might be flaky but the local environment could also be the issue.

For local, you could use `SKIP=pyrefly-check git commit` to avoid it preventing you from commit the change.

wwwjn (approved):

PR #2363, opened 2026-02-10
Add FP16 mixed precision training support with GradScaler

### LINKED ISSUE #2356: Float16 support with loss / gradient scaling
-- From user feedback

> While might sound counter intuitive, but fp16 is back! https://arxiv.org/abs/2510.26788 
> Supporting fp16 with proper scaling is essential for a more stable RL run. At the moment titan has no support for such a regime from our understanding.

zpcore:
FP16 has higher mantissa precision (10 bits vs BF16's 7 bits) which reduces rounding errors during RL fine-tuning, but requires loss scaling to prevent gradient underflow due to its limited exponent range (5 bits).

This commit integrates torch.amp.GradScaler into the training loop when FP16 is configured via training.dtype or training.mixed_precision_param, with proper support for FSDP, Pipeline Parallel, gradient clipping, checkpointing, and FluxTrainer. The implementation of composing `grad scalar` with `grad clipping` follows the example in https://github.com/pytorch/pytorch/blob/8f5c518dd325a7779d194c8961f670befb28f5d9/docs/source/notes/amp_examples.rst#gradient-clipping.

Resolves #2356

Authorized by Claude

### DIFF
--- a/torchtitan/distributed/utils.py
+++ b/torchtitan/distributed/utils.py
@@ -21,7 +21,12 @@ from torch import distributed as dist
 from torch.distributed.device_mesh import DeviceMesh
 from torch.distributed.tensor import DTensor
 
-from torchtitan.config import Comm as CommConfig, Debug as DebugConfig, TORCH_DTYPE_MAP
+from torchtitan.config import (
+    Comm as CommConfig,
+    Debug as DebugConfig,
+    JobConfig,
+    TORCH_DTYPE_MAP,
+)
 from torchtitan.distributed.parallel_dims import ParallelDims
 from torchtitan.tools.logging import logger
 from torchtitan.tools.utils import device_module, device_type
@@ -261,6 +266,27 @@ def maybe_enable_amp(
                 device_type,
                 dtype=TORCH_DTYPE_MAP[mixed_precision_param],
             )
+
+
+def build_grad_scaler(
+    job_config: JobConfig,
+    device_type: str,
+) -> torch.amp.GradScaler | None:
+    """Build a GradScaler if FP16 mixed precision is configured.
+
+    FP16 has only 5 exponent bits, so gradients can underflow to zero.
+    GradScaler applies dynamic loss scaling to prevent this.
+    Returns None if FP16 is not in use (i.e., BF16 or FP32).
+    """
+    fp16_in_use = (
+        job_config.training.mixed_precision_param == "float16"
+        or job_config.training.dtype == "float16"
+    )
+    if not fp16_in_use:
+        return None
+
+    logger.info("FP16 training enabled - creating GradScaler for loss scaling")
+    return torch.amp.GradScaler(device=device_type)
 
 
 def init_fake_mode(world_size: int, comm_mode: str = "fake_backend"):
--- a/torchtitan/models/flux/train.py
+++ b/torchtitan/models/flux/train.py
@@ -206,7 +206,11 @@ class FluxTrainer(Trainer):
             # need to free to before bwd to avoid peaking memory
             # pyrefly: ignore[unsupported-delete]
             del (latent_noise_pred, noise, target)
-            loss.backward()
+
+            if self.grad_scaler is not None:
+                self.grad_scaler.scale(loss).backward()
+            else:
+                loss.backward()
 
         return loss
 
@@ -229,6 +233,9 @@ class FluxTrainer(Trainer):
 
         loss = self.forward_backward_step(input_dict=input_dict, labels=labels)
 
+        if self.grad_scaler is not None:
+            self.grad_scaler.unscale_(self.optimizers)
+
         grad_norm = dist_utils.clip_grad_norm_(
             [p for m in self.model_parts for p in m.parameters()],
             self.job_config.training.max_norm,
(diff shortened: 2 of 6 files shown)


torchtitan/distributed/utils.py
```diff
@@ -263,6 +268,27 @@ def maybe_enable_amp(
...
+    """
+    fp16_in_use = (
+        job_config.training.mixed_precision_param == "float16"
+        or job_config.training.dtype == "float16"
+    )
+    if not fp16_in_use:
+        return None
+
+    logger.info("FP16 training enabled - creating GradScaler for loss scaling")
+    return torch.amp.GradScaler(device=device_type)
```
wconstab, line 289: does this work with torch.compile and with SAC / AC? can you also add some pointers to the best AMP docs to look at if you want to understand what this does?

torchtitan/models/flux/train.py
```diff
@@ -237,7 +244,13 @@ def train_step(
             ep_enabled=parallel_dims.ep_enabled,
         )
         self.checkpointer.maybe_wait_for_staging()
-        self.optimizers.step()
+
+        if self.grad_scaler is not None:
+            self.grad_scaler.step(self.optimizers)
```
wconstab, line 249: what does the grad_scaler step do?  we already unscale_ above, so i was expecting we don't need to do anything further?

zpcore: - Why call `unscale_` above:
This is the composability issue between `grad_scale` and `grad_clipping`. We don't want to clip on a scaled gradient.

- What does grad_scaler step do:
https://github.com/pytorch/pytorch/blob/449b1768410104d3ed79d3bcfe4ba1d65c7f22c0/torch/amp/grad_scaler.py#L89-L105. It may call optimizer.step() or just skip depending on whether there is NaNs in the gradient or not.

torchtitan/distributed/utils.py
zpcore: Just added the integration test for compile/sac/ac in the PR. 

This is the example code on how grad scaler should work with grad clipping:  https://github.com/pytorch/pytorch/blob/8f5c518dd325a7779d194c8961f670befb28f5d9/docs/source/notes/amp_examples.rst#gradient-clipping. Let me paste the link in the PR description.

torchtitan/train.py
```diff
@@ -215,6 +228,18 @@ def __init__(self, job_config: JobConfig):
         )
         assert self.gradient_accumulation_steps > 0
 
+        self.grad_scaler = dist_utils.build_grad_scaler(job_config, device_type)
```
wconstab, line 231: does the grad_scaler have state Tensors? if so, what device are they on? it probably isn't a big deal either way, i was just wondering if we need to move them to the device the user specified for training, or just leaving them as default is always fine.

wconstab (commented):
overall lgtm, @tianyu-l do you want to do a review or else i can stamp

tianyu-l (changes requested):
I'm doing a big refactor right now, so prefer to land this moderately large PR later, if it's not urgent.

In general, I feel the footprint of this change is quite scattered, e.g. a lot of `if self.grad_scaler is not None`, increasing the challenge to understand and maintain. I would rethink if the abstraction is right.

In the past, I've seen it modeled as an object and does no op if not enabled. Not saying it makes the most sense though.

wconstab:
well, you could replace all the 'if not none' checks by having a grad_scaler wrapper that offered a bunch of 'maybe_' apis.  that would reduce a few lines of code in each spot, but it doesn't make the footprint less scattered. i think its hard to do this without at least 4 touch points in the trainer code.

zpcore:
This is not urgent, I can land after the refactor. `grad scaler` need to touch loss computation and optimizer step, I am not sure if there is a cleaner way to introduce it.

torchtitan/train.py
```diff
@@ -742,11 +788,16 @@ def should_continue_training(self) -> bool:
...
+        sd: dict[str, Any] = {"step": self.step, "ntokens_seen": self.ntokens_seen}
+        if self.grad_scaler is not None:
+            sd["grad_scaler"] = self.grad_scaler.state_dict()
+        return sd
 
     def load_state_dict(self, state_dict: dict[str, Any]):
         self.step = state_dict["step"]
         self.ntokens_seen = state_dict["ntokens_seen"]
+        if self.grad_scaler is not None and "grad_scaler" in state_dict:
+            self.grad_scaler.load_state_dict(state_dict["grad_scaler"])
```
fegin, line 800: We should treat grad_scaler as a state and directly put into Checkpointer, not modifying state_dict and load_state_dict.

fegin:
I feel some information is leaked into Trainer init where it can be avoided. Like the raise exception can be done in the build_grad_scaler. Creating a loss function with grad_scaler can also be done in a util function. I think you can create a grad_scaler.py and have these utility functions. Then at least the init part should only have one or two lines changes.

tianyu-l:
I don't think this is an urgent feature and would love us to prioritize other things until this is absolutely necessary. If I'm convinced otherwise, I would take a closer look on the necessity of the scattering part.

PR #2364, opened 2026-02-10
[rl][poc] Native PyTorch Varlen Attention for inference.

zhxchen17:
Summary:

### what this PR does
This PR adds the option to use PyTorch's native varlen Flash Attention kernel (instead of vllm's custom FA kernel) for vLLM Generator.

### why doing this

Traditionally different training and inference model definitions are used for training and rollout, this causes instability issues. To address this problem, we want to make sure training and inference model definition are matching.
In order to match FA kernel on both sides. There're 3 ways to do this:
1. Use trainer's existing FA kernel on generator side.
2. Use generator's existing FA kernel on trainer side.
3. Have a common dependency and patch it to be used on both sides.

This PR essentially takes a step towards option 3. Since PyTorch is a common dependency to every repo (titan and vllm), and more importantly it recently adds support for pagedd attention, which makes it possible to be used for inference case like vLLM.

### how to enable
When user add `attention_backend='CUSTOM'` to Generator's config, it will automatically switch to PyTorch's native FA kernel for inference.

Test Plan:

```
torchrun --nproc_per_node=2     torchtitan/experiments/rl/unified/validation/test_attn_numerics.py 
```
output result:

```
[2026-03-05 08:57:23] INFO test_attn_numerics.py:118: vLLM LLMEngine ready.
[2026-03-05 08:57:23] INFO test_attn_numerics.py:77: vLLM attention module: VLLMAttention
WARNING 03-05 08:57:23 [input_processor.py:254] Passing raw prompts to InputProcessor is deprecated and will be removed in v0.18. You should instead pass the outputs of Renderer.render_cmpl() or Renderer.render_chat().
[2026-03-05 08:57:31] INFO test_attn_numerics.py:150: vLLM generated 30 tokens (prompt length=5)
[2026-03-05 08:57:31] INFO test_attn_numerics.py:291: Prompt token IDs: [9707, 11, 847, 829, 374]
[2026-03-05 08:57:31] INFO test_attn_numerics.py:292: Generated token IDs (30): [444, 2210, 13, 358, 2776, 264, 220, 17, 17, 4666, 6284, 5458, 504, 5616, 13, 358, 2776, 8014, 304, 20956, 304, 279, 2274, 13, 358, 2776, 3330, 369, 264, 2618]
[2026-03-05 08:57:31] INFO attention.py:368: Successfully replaced TorchTitan attention with VLLMCompatibleFlashAttention (28 layers)
[2026-03-05 08:57:31] INFO parallel_dims.py:131: Building device mesh with parallelism: pp=1, dp_replicate=1, dp_shard=1, cp=1, tp=2, ep=1, etp=1
[2026-03-05 08:57:31] INFO attention.py:368: Successfully replaced TorchTitan attention with VLLMCompatibleFlashAttention (28 layers)
[2026-03-05 08:57:31] INFO parallel_dims.py:131: Building device mesh with parallelism: pp=1, dp_replicate=1, dp_shard=1, cp=1, tp=2, ep=1, etp=1
[2026-03-05 08:57:31] INFO parallel_dims.py:184: Successfully created meshes with active dimensions: ['tp', 'efsdp']
[2026-03-05 08:57:31] INFO parallel_dims.py:184: Successfully created meshes with active dimensions: ['tp', 'efsdp']
[2026-03-05 08:57:32] WARNING state_dict_adapter.py:96: model.safetensors.index.json not found at hf_assets_path: torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B/model.safetensors.index.json.                     Defaulting to saving a single safetensors file if checkpoint is saved in HF format
[2026-03-05 08:57:32] WARNING state_dict_adapter.py:96: model.safetensors.index.json not found at hf_assets_path: torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B/model.safetensors.index.json.                     Defaulting to saving a single safetensors file if checkpoint is saved in HF format
[2026-03-05 08:57:32] INFO test_attn_numerics.py:224: Loaded HF weights from torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B (311 params)
[2026-03-05 08:57:32] INFO test_attn_numerics.py:77: Trainer attention module: VLLMCompatibleFlashAttention
[2026-03-05 08:57:32] INFO test_attn_numerics.py:310: Trainer computed 30 token log-probs
[2026-03-05 08:57:32] INFO test_attn_numerics.py:314:   vLLM   log-probs[:5]: [-4.064868450164795, -1.785841941833496, -0.708375871181488, -0.2134280651807785, -0.9917267560958862]
[2026-03-05 08:57:32] INFO test_attn_numerics.py:315:   Trainer log-probs[:5]: [-4.064868450164795, -1.785841941833496, -0.708375871181488, -0.2134280651807785, -0.9917267560958862]
[2026-03-05 08:57:32] INFO test_attn_numerics.py:327: ============================================================
[2026-03-05 08:57:32] INFO test_attn_numerics.py:328: LOGPROB COMPARISON RESULTS
[2026-03-05 08:57:32] INFO test_attn_numerics.py:329: ============================================================
[2026-03-05 08:57:32] INFO test_attn_numerics.py:330:   Bitwise identical : True
[2026-03-05 08:57:32] INFO test_attn_numerics.py:331:   Tokens checked    : 30
[2026-03-05 08:57:32] INFO test_attn_numerics.py:332:   Tokens different  : 0
[2026-03-05 08:57:32] INFO test_attn_numerics.py:333:   Max delta         : 0.000000e+00
[2026-03-05 08:57:32] INFO test_attn_numerics.py:334:   Avg delta         : 0.000000e+00
[2026-03-05 08:57:32] INFO test_attn_numerics.py:335:   Diff mean         : 0.000000e+00
[2026-03-05 08:57:32] INFO test_attn_numerics.py:336:   Diff max          : 0.000000e+00
[2026-03-05 08:57:32] INFO test_attn_numerics.py:337: ============================================================
[2026-03-05 08:57:32] INFO test_attn_numerics.py:344: PASS: vLLM and trainer log-probs are bitwise identical.
```

perf benchmark:
```
torchrun --nproc_per_node=2     torchtitan/experiments/rl/unified/validation/bench_attn_perf.py
```

```
================================================================
  Attention Backend Performance Comparison
  Model: Qwen3-0.6B | TP: 2 | Prompts: 8 | Prompt len: 512
  Decode tokens: 128 | Warmup: 1 | Timed runs: 3
================================================================

Backend        Phase        Tokens     Time (s)      Tok/s
----------------------------------------------------------
FLASH_ATTN     Prefill        4096       0.2325      17615
FLASH_ATTN     Decode         1024      28.6333         36
CUSTOM         Prefill        4096       0.2300      17809
CUSTOM         Decode         1024      29.1132         35

Speedup (CUSTOM / FLASH_ATTN):
  Prefill: 1.01x
  Decode:  0.98x
================================================================
```
Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchtitan/experiments/rl/models/attention.py
+++ b/torchtitan/experiments/rl/models/attention.py
@@ -8,12 +8,160 @@ import logging
 
 import torch
 from torch.distributed.tensor import DTensor
-from torchtitan.experiments.rl.models.vllm_compat_attention import (
-    VLLMCompatibleFlashAttention,
+from torch.nn.attention import (
+    activate_flash_attention_impl,
+    current_flash_attention_impl,
 )
 from torchtitan.protocols.module import Module
 
 from vllm.model_executor.layers.attention import Attention
+from vllm.v1.attention.backend import AttentionType
+from vllm.v1.attention.backends.flash_attn import (
+    FlashAttentionBackend,
+    FlashAttentionImpl,
+    FlashAttentionMetadata,
+)
+from vllm.v1.attention.backends.registry import AttentionBackendEnum, register_backend
+
+
+@register_backend(AttentionBackendEnum.CUSTOM)
+class PyTorchFlashAttentionBackend(FlashAttentionBackend):
+    """Custom vLLM attention backend using PyTorch's native FlashAttention kernel.
+
+    This class is not directly referenced in user code. It is registered into
+    vLLM's attention backend registry via the ``@register_backend`` decorator
+    and selected at runtime when the vLLM engine is configured to use a CUSTOM
+    attention backend.
+
+    Inheriting from ``FlashAttentionBackend`` is not strictly required for all
+    backends, but it is convenient here to reuse metadata construction logic.
+    """
+
+    @staticmethod
+    def get_name():
+        # vLLM requires any custom attention backend to return "CUSTOM" as its
+        # name so the backend registry can look it up correctly.
+        return "CUSTOM"
+
+    @staticmethod
+    def get_impl_cls():
+        return PyTorchFlashAttentionImpl
+
+
+class PyTorchFlashAttentionImpl(FlashAttentionImpl):
+    """
+    Custom vLLM attention backend impl using PyTorch's native FlashAttention varlen API.
+    Instead of using vLLM's FlashAttention kernel, this implementation takes the kernel
+    dependency from torch directly while supporting the same interface.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        # Activate PyTorch's FA3 flash attention implementation. This is required for
+        # the varlen attention API to work with paged attention (KV cache in block
+        # layout) because vLLM uses default page size 16 but FA2 only supports
+        # page_size=256. Error out if FA3 is not available in the current environment.
+        if current_flash_attention_impl() != "FA3":
+            activate_flash_attention_impl("FA3")
+
+    # Based on vLLM's FlashAttentionImpl.forward():
+    # https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/flash_attn.py
+    def forward(
+        self,
+        layer: torch.nn.Module,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        kv_cache: torch.Tensor,
+        attn_metadata: FlashAttentionMetadata,
+        output: torch.Tensor | None = None,
+        output_scale: torch.Tensor | None = None,
+        output_block_scale: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        """Forward pass with FlashAttention.
+
+        Args:
+            query: shape = [num_tokens, num_heads, head_size]
+            key: shape = [num_tokens, num_kv_heads, head_size]
+            value: shape = [num_tokens, num_kv_heads, head_size]
+            kv_cache: shape =
+                [2, num_blocks, block_size, num_kv_heads, head_size]
+            attn_metadata: Metadata for attention.
+        Returns:
+            shape = [num_tokens, num_heads * head_size]
+        """
+        assert output is not None, "Output tensor must be provided."
+        assert (
+            self.vllm_flash_attn_version is not None
+        ), "FlashAttention version not detected."
+
+        if output_scale is not None or output_block_scale is not None:
+            raise NotImplementedError(
+                "fused output quantization is not yet supported for FlashAttentionImpl"
+            )
+
+        if attn_metadata is None:
+            # Profiling run.
+            return output.fill_(0)
+
+        attn_type = self.attn_type
+
+        # IMPORTANT!
+        # NOTE(woosuk): With piece-wise CUDA graphs, this method is executed in
+        # eager-mode PyTorch. Thus, we need to be careful about any CPU overhead
+        # in this method. For example, `view` and `slice` (or `[:n]`) operations
+        # are surprisingly slow even in the case they do not invoke any GPU ops.
+        # Minimize the PyTorch ops in this method as much as possible.
+        # Whenever making a change in this method, please benchmark the
+        # performance to make sure it does not introduce any overhead.
+
+        num_actual_tokens = attn_metadata.num_actual_tokens
+
+        assert attn_type not in (
+            AttentionType.ENCODER_ONLY,
+            AttentionType.ENCODER,
+        ), "Encoder-only attention not supported yet."
+
+        # For decoder and cross-attention, use KV cache as before
+        key_cache, value_cache = kv_cache.unbind(0)
+
+        assert not self.kv_cache_dtype.startswith(
+            "fp8"
+        ), "FP8 KV cache not supported yet."
+
+        assert not attn_metadata.use_cascade, "Cascade not supported yet."
+
+        cu_seqlens_q = attn_metadata.query_start_loc
+        seqused_k = attn_metadata.seq_lens
+        max_seqlen_q = attn_metadata.max_query_len
+        max_seqlen_k = attn_metadata.max_seq_len
+        block_table = attn_metadata.block_table
+
+        assert self.dcp_world_size == 1, "DCP not supported yet."
+
+        if attn_metadata.causal:
+            sliding_window_size = (-1, 0)
+        else:
+            raise RuntimeError("Non-causal attention not supported yet.")
+
+        assert self.alibi_slopes is None, "Alibi slopes not supported yet."
+
+        return torch.nn.attention.varlen.varlen_attn_out(
+            output[:num_actual_tokens],
+            query[:num_actual_tokens],
+            key_cache,
+            value_cache,
+            cu_seqlens_q,
+            None,
+            max_seqlen_q,
+            max_seqlen_k,
+            scale=self.scale,
+            window_size=sliding_window_size,
+            block_table=block_table,
+            seqused_k=seqused_k,
+        )
+
 
 logger = logging.getLogger(__name__)
 
--- a/torchtitan/experiments/rl/tests/test_attn_numerics.py
+++ b/torchtitan/experiments/rl/tests/test_attn_numerics.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Test that vLLM engine and TorchTitan Module forward produce bitwise identical
+log-probs for the same prompt + generated token sequence.
+
+Flow:
+  1. Generate tokens from a single prompt via the vLLM engine (greedy decoding).
+  2. Feed [prompt + generated tokens] through a standalone TorchTitan model
+     forward and compute per-token log-probs.
+  3. Compare the two sets of log-probs bitwise.
+
+Run:
+    torchrun --nproc_per_node=2 \
+        torchtitan/experiments/rl/tests/test_attn_numerics.py
+"""
+
+import logging
+import os
+
+# Must set spawn method before any CUDA operations or vLLM imports
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+import torch
+import torch.distributed as dist
+import torch.distributed.checkpoint as dcp
+from torch.distributed.checkpoint.state_dict import (
+    set_model_state_dict,
+    StateDictOptions,
+)
+
+from vllm import EngineArgs, LLMEngine, SamplingParams
+from vllm.config.compilation import CompilationConfig, CUDAGraphMode
+from vllm.model_executor.layers.batch_invariant import init_batch_invariance
+from vllm.sampling_params import RequestOutputKind
+from vllm.v1.attention.backends.registry import AttentionBackendEnum
+
+from torchtitan.config import CommConfig, TORCH_DTYPE_MAP
+from torchtitan.config.configs import ParallelismConfig
+from torchtitan.distributed import ParallelDims, utils as dist_utils
+from torchtitan.experiments.rl.actors.generator import SamplingConfig, VLLMGenerator
+from torchtitan.experiments.rl.actors.trainer import PolicyTrainer
+from torchtitan.experiments.rl.actors.utils import (
+    compute_token_log_probs,
+    verify_logprob_identity,
+)
+from torchtitan.experiments.rl.models.parallelize import parallelize_qwen3
+from torchtitan.experiments.rl.plugin import (
+    register_model_to_vllm_model_registry,
+    VLLM_MODEL_NAME,
+)
+from torchtitan.experiments.rl.simple_grpo_sum_digits import RLTrainer
+from torchtitan.models.qwen3 import model_registry
+from torchtitan.tools import utils
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def _log_attention_module(model, label):
+    """Log the inner_attention class name from layer 0 (rank 0 only)."""
+    if dist.get_rank() != 0:
+        return
+    layer_0 = next(iter(model.layers.values()))
+    attn_cls = type(layer_0.attention.inner_attention).__name__
+    logger.info(f"{label} attention module: {attn_cls}")
+
+
+# ---------------------------------------------------------------------------
+# Step 1: vLLM engine generation
+# ---------------------------------------------------------------------------
+
+
+def create_vllm_engine(config):
+    """Create a vLLM LLMEngine from the RL config."""
+    gen_config = config.generator
+    model_path = config.hf_assets_path
+
+    # Enable batch-invariant mode so vLLM uses the deterministic flash-attn path.
+    init_batch_invariance(AttentionBackendEnum[gen_config.attention_backend])
+
+    assert gen_config.attention_backend == "CUSTOM"
+    engine_kwargs = dict(
+        model=model_path,
+        trust_remote_code=True,
+        dtype=gen_config.model_dtype,
+        tensor_parallel_size=gen_config.parallelism.tensor_parallel_degree,
+        distributed_executor_backend="external_launcher",
+        gpu_memory_utilization=gen_config.gpu_memory_limit,
+        hf_overrides={"architectures": [VLLM_MODEL_NAME]},
+        attention_backend=gen_config.attention_backend,
+        compilation_config=CompilationConfig(
+            backend="eager",
+            cudagraph_mode=CUDAGraphMode.FULL_AND_PIECEWISE,
+        ),
+    )
+    if gen_config.seed is not None:
+        engine_kwargs["seed"] = gen_config.seed
+    engine_args = EngineArgs(**engine_kwargs)
+
+    if dist.get_rank() == 0:
+        logger.info("Creating vLLM LLMEngine ...")
+    engine = LLMEngine.from_engine_args(engine_args)
+    if dist.get_rank() == 0:
+        logger.info("vLLM LLMEngine ready.")
+    return engine
+
+
+def generate_with_vllm(engine, prompt, gen_config):
+    """Generate tokens from *prompt*, return IDs + log-probs."""
+    sampling = gen_config.sampling
+    sampling_params = SamplingParams(
+        temperature=sampling.temperature,
+        top_p=sampling.top_p,
+        max_tokens=sampling.max_tokens,
+        logprobs=1,
+        prompt_logprobs=1,
+        output_kind=RequestOutputKind.FINAL_ONLY,
+    )
+
+    engine.add_request("0", prompt, sampling_params)
+
+    outputs = []
+    while engine.has_unfinished_requests():
+        step_outputs = engine.step()
+        outputs.extend(step_outputs)
+
+    assert len(outputs) == 1, f"Expected 1 output, got {len(outputs)}"
+    output = outputs[0]
+
+    prompt_token_ids = list(output.prompt_token_ids)
+    sample = output.outputs[0]
+    generated_token_ids = list(sample.token_ids)
+    vllm_log_probs = [list(lp_dict.values())[0].logprob for lp_dict in sample.logprobs]
+
+    if dist.get_rank() == 0:
+        logger.info(
+            f"vLLM generated {len(generated_token_ids)} tokens "
+            f"(prompt length={len(prompt_token_ids)})"
+        )
+    return prompt_token_ids, generated_token_ids, vllm_log_probs
+
+
+# ---------------------------------------------------------------------------
+# Step 2: Build trainer-side TorchTitan model
+# ---------------------------------------------------------------------------
+
+
+def build_trainer_model(config):
+    """Build, parallelize, and load weights for the trainer model.
+
+    Mirrors PolicyTrainer._build_model() but without the Monarch actor
+    framework.
+
+    TODO(zhxchen17) Switch trainer model to use varlen backend with batch-invariant mode.
+    """
+    model_spec = config.model_spec
+    model_config = model_spec.model
+    hf_assets_path = config.hf_assets_path
+
+    device_module, device_type = utils.device_module, utils.device_type
+    local_rank = int(os.environ.get("LOCAL_RANK", 0))
+    device = torch.device(f"{device_type}:{local_rank}")
+    device_module.set_device(device)
+
+    world_size = dist.get_world_size()
+    parallelism = config.trainer.parallelism
+
+    parallel_dims = ParallelDims(
+        dp_shard=parallelism.data_parallel_shard_degree,
+        dp_replicate=parallelism.data_parallel_replicate_degree,
+        cp=parallelism.context_parallel_degree,
+        tp=parallelism.tensor_parallel_degree,
+        pp=parallelism.pipeline_parallel_degree,
+        ep=parallelism.expert_parallel_degree,
+        etp=parallelism.expert_tensor_parallel_degree,
+        world_size=world_size,
+    )
+
+    # Build model on meta device
+    with torch.device("meta"):
+        with utils.set_default_dtype(TORCH_DTYPE_MAP["bfloat16"]):
+            model = model_config.build()
+
+    # Parallelize (trainer path: has_position_id=False)
+    model = parallelize_qwen3(
+        model,
+        parallel_dims=parallel_dims,
+        parallelism=parallelism,
+    )
+
+    # Materialize on device
+    model.to_empty(device=device_type)
+    with torch.no_grad():
+        model.init_weights(buffer_device=None)
+
+    # Load HF checkpoint (same logic as PolicyTrainer._load_initial_hf_weights)
+    if model_spec.state_dict_adapter is not None:
+        sd_adapter = model_spec.state_dict_adapter(model_config, hf_assets_path)
+        storage_reader = sd_adapter.get_hf_storage_reader(hf_assets_path)
+        hf_state_dict = sd_adapter.to_hf(model.state_dict())
+        dcp.load(hf_state_dict, storage_reader=storage_reader)
+        torchtitan_state_dict = sd_adapter.from_hf(hf_state_dict)
+        set_model_state_dict(
+            model=model,
+            model_state_dict=torchtitan_state_dict,
+            options=StateDictOptions(strict=False),
+        )
+        if dist.get_rank() == 0:
+            logger.info(
+                f"Loaded HF weights from {hf_assets_path} "
+                f"({len(torchtitan_state_dict)} params)"
+            )
+
+    model.eval()
+    return model, device
+
+
+# ---------------------------------------------------------------------------
+# Main
+# ---------------------------------------------------------------------------
+
+
+def _test_config() -> RLTrainer.Config:
+    """Test-specific config: greedy sampling, fewer tokens, single sample."""
+    return RLTrainer.Config(
+        model_spec=model_registry("0.6B"),
+        hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
+        batch_invariant_mode=True,
+        trainer=PolicyTrainer.Config(
+            parallelism=ParallelismConfig(
+                tensor_parallel_degree=2,
+                data_parallel_replicate_degree=1,
+            ),
+        ),
+        generator=VLLMGenerator.Config(
+            model_dtype="bfloat16",
+            gpu_memory_limit=0.5,
+            parallelism=ParallelismConfig(
+                tensor_parallel_degree=2,
+            ),
+            num_samples_per_prompt=1,
+            sampling=SamplingConfig(
+                temperature=0.0,
+                top_p=1.0,
+                max_tokens=30,
+            ),
+            attention_backend="CUSTOM",
+        ),
+    )
+
+
+def main():
+    # ---- Config setup ----
+    config = _test_config()
+    config.model_spec.parallelize_fn = parallelize_qwen3
+
+    # Register model with vLLM
+    register_model_to_vllm_model_registry(config.model_spec)
+
+    # Initialize distributed (needed for both vLLM and trainer model)
+    dist_utils.init_distributed(CommConfig())
+
+    # ---- Step 1: Generate via vLLM ----
+    prompt = "Hello, my name is"
+
+    engine = create_vllm_engine(config)
+    vllm_model = engine.model_executor.driver_worker.get_model().model
+    _log_attention_module(vllm_model, "vLLM")
+    prompt_token_ids, generated_token_ids, vllm_log_probs = generate_with_vllm(
+        engine, prompt, config.generator
+    )
+
+    if dist.get_rank() == 0:
+        logger.info(f"Prompt token IDs: {prompt_token_ids}")
+        logger.info(
+            f"Generated token IDs ({len(generated_token_ids)}): {generated_token_ids}"
+        )
+
+    # Free vLLM engine GPU memory before building the trainer model
+    del vllm_model, engine
+    torch.cuda.empty_cache()
+
+    # ---- Step 2: Build trainer model and compute log-probs ----
+    trainer_model, device = build_trainer_model(config)
+    _log_attention_module(trainer_model, "Trainer")
+
+    with torch.no_grad():
+        trainer_token_log_probs = compute_token_log_probs(
+            trainer_model, prompt_token_ids, generated_token_ids, device
+        )
+
+    if dist.get_rank() == 0:
+        logger.info(
+            f"Trainer computed {trainer_token_log_probs.shape[0]} token log-probs"
+        )
+        n_preview = min(5, len(vllm_log_probs))
+        logger.info(f"  vLLM   log-probs[:{n_preview}]: {vllm_log_probs[:n_preview]}")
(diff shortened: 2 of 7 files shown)


torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -5,7 +5,119 @@
...
+        cu_seqlens_q = attn_metadata.query_start_loc
+        seqused_k = attn_metadata.seq_lens
+        max_seqlen_q = attn_metadata.max_query_len
+        max_seqlen_k = attn_metadata.max_seq_len
+        block_table = attn_metadata.block_table
+        scheduler_metadata = attn_metadata.scheduler_metadata
+
+        descale_shape = (cu_seqlens_q.shape[0] - 1, self.num_kv_heads)
+
+        q_descale = layer._q_scale.expand(descale_shape)
```
wwwjn: What is this descale doing?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -5,7 +5,119 @@
...
+
 from vllm.model_executor.layers.attention import Attention
+from vllm.v1.attention.backend import AttentionType
+from vllm.v1.attention.backends.flash_attn import (
+    FlashAttentionImpl,
+    FlashAttentionMetadata,
+)
+
+
+class NativeFlashAttentionImpl(FlashAttentionImpl):
```
wwwjn: So the current implementation is inherit from FlashAttention class and patch to varlen attention? The current implementation seems working to change vllm's attention to run PyTorch's varlen attention. 

I think we would need to verify the numerics are the bit-wise identitical for following runs, given the same input:
1. vllm.Attention() forward (vllm-native) v.s. varlen attention forward
2. End-2-End Qwen3 model forward

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: This seems to be a feature for flash attention and vllm to fuse descaling into the kernel. Currently pytorch's FA kernel doesn't support it but I already let @liangel-02 know so hopefully we can support it in the next iteration.

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: Sounds good. I will run some scripts to verify the numerics.

torchtitan/experiments/rl/unified/models/attention.py
liangel-02: is descaling for FP8 quantized attention smth we need? if so i can definitely add

zhxchen17: we can leave descale out of scope for now.

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: Added to validation/test_attn_numerics.py

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -8,10 +8,128 @@
...
+from vllm.v1.attention.backends.flash_attn import (
+    FlashAttentionBackend,
+    FlashAttentionImpl,
+    FlashAttentionMetadata,
+)
+from vllm.v1.attention.backends.registry import AttentionBackendEnum, register_backend
+
+
+@register_backend(AttentionBackendEnum.CUSTOM)
+class UnifiedFlashAttentionBackend(FlashAttentionBackend):
```
tianyu-l: Is this used anywhere?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -8,10 +8,128 @@
...
 from torch.distributed.tensor import DTensor
+from torch.nn.attention import activate_flash_attention_impl
 from torchtitan.experiments.rl.vllm_compat.models.attention import (
     VLLMCompatibleFlashAttention,
 )
+
+try:
+    activate_flash_attention_impl("FA3")
+except Exception:
+    pass
```
tianyu-l: pls document what this is doing

torchtitan/experiments/rl/unified/models/attention.py
tianyu-l: Does other backend not need to inherit `FlashAttentionBackend`?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -8,10 +8,128 @@
 
 import torch
 from torch.distributed.tensor import DTensor
+from torch.nn.attention import activate_flash_attention_impl
 from torchtitan.experiments.rl.vllm_compat.models.attention import (
     VLLMCompatibleFlashAttention,
 )
+
+try:
+    activate_flash_attention_impl("FA3")
```
wwwjn: Why we need to "activate" it first?

tianyu-l (changes requested):
It it not clear what this PR is trying to do. Would you please add documentation, before people can review? Thanks!

torchtitan/experiments/rl/unified/models/attention.py
wwwjn: I would not call it "unified", maybe PyTorchFlashAttentionBackend

torchtitan/experiments/rl/models/attention.py
```diff
@@ -8,10 +8,128 @@
...
+from vllm.v1.attention.backends.registry import AttentionBackendEnum, register_backend
+
+
+@register_backend(AttentionBackendEnum.CUSTOM)
+class UnifiedFlashAttentionBackend(FlashAttentionBackend):
+    """Unified FlashAttention backend for RL."""
+
+    @staticmethod
+    def get_name():
+        return "CUSTOM"
```
wwwjn, line 44: Does it has to be "CUSTOM" ? Should we have a descriptive field?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -8,10 +8,128 @@
...
+            query: shape = [num_tokens, num_heads, head_size]
+            key: shape = [num_tokens, num_kv_heads, head_size]
+            value: shape = [num_tokens, num_kv_heads, head_size]
+            kv_cache: shape =
+                [2, num_blocks, block_size, num_kv_heads, head_size]
+            attn_metadata: Metadata for attention.
+        Returns:
+            shape = [num_tokens, num_heads * head_size]
+        """
+        assert output is not None, "Output tensor must be provided."
```
wwwjn, line 94: These part seems from vllm. Can we add a reference link here so it's easier to keep updated later

wwwjn:
> It it not clear what this PR is trying to do. Would you please add documentation, before people can review? Thanks!

I think I have a little bit more context here: We want to use PyTorch's FA kernel in vllm Attention. Now this PR use vllm's standard extension way to register PyTorch's FA kernel to be a option of vllm's Attention Backend. 

Can you remind me why we don't just use something like `torch.ops._vllm_fa3_C.fwd = <PyTorch's varlen attention>`? No matter which way we decide to go with, I think the following statement is true:
- During inference, We still have to replace torchtitan model's inner attention to be `vllm.Attention()` to let vLLM Engine passing `attention_metadata` to attention kernel

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: This will be registered to vllm's attention registry to be used later, so it's not directly referenced in the code.

zhxchen17: Inheriting from FlashAttentionBackend is not a strict requirement for every backend.

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: This is the requirement to use varlen API with paged attention. cc @liangel-02

torchtitan/experiments/rl/models/attention.py
zhxchen17: @wwwjn I think vllm requires any custom backend to always have the name of "CUSTOM".

zhxchen17:
> Can you remind me why we don't just use something like `torch.ops._vllm_fa3_C.fwd = <PyTorch's varlen attention>`? 

@wwwjn I think we can do that as well, it's just less stable as we patch a private op and there's no guarantee will always be the right op to patch as vllm evolve.

zhxchen17:
@tianyu-l @wwwjn sorry my bad being ignorant on the PR summary and documentation. 

Just added more docs to the code to explain what we're doing, and also updated PR summary/message. Feel free to have a look again and lemme know if there's any questions.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -4,14 +4,159 @@
...
+                "fused output quantization is not yet supported for FlashAttentionImpl"
+            )
+
+        if attn_metadata is None:
+            # Profiling run.
+            return output.fill_(0)
+
+        attn_type = self.attn_type
+
+        # IMPORTANT!
```
Lucaskabela, line 110: Yeah this is my chief concern - I see numerics benchmarking here but no latency benchmarking.  Do we have any handle on that?

torchtitan/experiments/rl/unified/models/attention.py
liangel-02: vLLM uses default page size 16 but FA2 only supports page_size= 256

zhxchen17:
@Lucaskabela as I mentioned in the PR summary, this is part of the unified trainer/generator RL experiment where we're using the same kernel to match the numerics. As I discussed offline with @wwwjn, we want to have this work as an established numerics baseline that we can iterate on.

vLLM's attention kernel is opaque to torch.compile so I'm not changing anything related to compile as far as I know (but happy to verify based on your PR). Some context behind this PR: https://docs.google.com/document/d/11IwAR-8hg2WH52pYC4eBqHJS4LCzBFoOArwv0GleC5A/edit?tab=t.0#heading=h.vz4r7rher0f7

Could you be more specific about what type of change you're requesting here? E.g. if you meant to have some perf benchmarking here with compile, I can definitely include that in the same PR, but note that this is not the default path now, so it's not affecting our iteration now and we already have the work in parallel improve it with @liangel-02 which should be non blocking.

Lucaskabela:
@zhxchen17 yes, the request is for some perf investigation to be added to the summary and to verify this is compilable :)

liangel-02:
@Lucaskabela working on perf investigation :)

zhxchen17:
@Lucaskabela I've tested against your PR and with some minor fixes it works with compile :)

See https://github.com/pytorch/torchtitan/pull/2502

wwwjn:
> Can you check how this integrates with #2486 (compile support)?
> 
> My concern would be switching to this custom backend may cause some issues with the piecewise cudagraphs. I am also curious what the hit to perf is - my understanding is Pytorch's attention is much less performant than vLLMs so if numerics can match using vLLM's definition, I don't understand why we would want to switch towards a slower impl

@Lucaskabela  I can provide some context here: We want to use pytorch as the single source of truth for the MODEL definition. And we use vllm as our inference ENGINE. 

We want to use truth canonical model definition (including the same attention kernel) on trainer and generator, for bitwise identity / true on-policy RL training. We care about numerics parity on trainer and generator side. Right now on trainer side, we are manually adding backward to vllm's FA kernel to make it achieve bit-wise identity with generator. This is not idea because we let the MODEL definition depend on vllm ENGINE. 

> may cause some issues with the piecewise cudagraphs 

Why is that? Will directly patching the ops `torch.ops._vllm_fa3_C.fwd = <PyTorch's varlen attention>` affect cudagraph?

Lucaskabela:
@zhxchen17  thanks for testing :) glad it works!

> Why is that? Will directly patching the ops torch.ops._vllm_fa3_C.fwd = <PyTorch's varlen attention> affect cudagraph?

@wwwjn part of the piecewise cudagraphs is splitting around the attention operations.  I was not sure if there was some limitation/issue with this for custom attention backends, but since @zhxchen17 verified it works I don't think that is a concern anymore

Lucaskabela:
I would like to see some perf impact before approving, but dismissed by request changes :)

torchtitan/experiments/rl/tests/test_attn_numerics.py
tianyu-l, line 1: Please put under `tests/` to be consistent with rest of repo.

Also, should we add it to CI? o/w the test can break any time.

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -42,28 +42,31 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
                 tensor_parallel_degree=2,
-                data_parallel_replicate_degree=1,
             ),
             num_samples_per_prompt=8,
             sampling=SamplingConfig(
                 temperature=0.8,
                 top_p=0.95,
                 max_tokens=100,
             ),
+            # TODO (zhxchen17) Switch to custom attention backend once upstream is ready.
```
tianyu-l: If it doesn't work today, we shouldn't merge this PR.
If it works after upstream is ready, we should remove this TODO.

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -67,6 +215,8 @@ def __init__(
             quant_config=None,
             prefix=f"model.layers.{layer_name}.attention.inner_attention",
         )
+        # assert attention.attn_backend is PyTorchFlashAttentionBackend
```
tianyu-l: what is this comment for?

torchtitan/experiments/rl/models/attention.py
tianyu-l: What can we do if we want to support both flex attention (in the future) and varlen attention as custom backend -- is the only option to share the "CUSTOM" name and dynamically register either one of them during construct time?

torchtitan/experiments/rl/tests/test_attn_numerics.py
```diff
@@ -0,0 +1,346 @@
...
+# ---------------------------------------------------------------------------
+
+
+def create_vllm_engine(config):
+    """Create a vLLM LLMEngine from the RL config."""
+    gen_config = config.generator
+    model_path = config.trainer.hf_assets_path
+
+    # Enable batch-invariant mode so vLLM uses the deterministic flash-attn path.
+    init_batch_invariance(AttentionBackendEnum[gen_config.attention_backend])
```
tianyu-l, line 85: Does CUSTOM varlen backend support batch invariance?

torchtitan/experiments/rl/unified/validation/test_attn_numerics.py
```diff
@@ -0,0 +1,346 @@
...
+        world_size=world_size,
+    )
+
+    # Build model on meta device
+    with torch.device("meta"):
+        with utils.set_default_dtype(TORCH_DTYPE_MAP["bfloat16"]):
+            model = model_config.build()
+
+    # Patch attention to use the same flash-attn kernel as vLLM
+    replace_with_vllm_compatible_flash_attention(model, tp_size=parallel_dims.tp)
```
tianyu-l: This is making trainer using vllm's flash attention kernel.

Meanwhile in this test it seems generator is using pytorch varlen kernel. Is this test saying there produce bitwise identical result? That sounds surprising. Is it because under the hood they are the same thing?

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -42,28 +42,31 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
             attention_backend="FLASH_ATTN",
```
tianyu-l: Is `"FLASH_ATTN"` still providing any value to us, in terms of perf, batch invariance, compile/cudagraph compatibility etc.? If not we should just remove this backend.

liangel-02:
@Lucaskabela i was able to verify that creating some unnecessary tensors (ie debug_attn_mask, philox_offset) were causing some performance gaps so put up this PR https://github.com/pytorch/pytorch/pull/176723 to fix. with these changes, i am able to match vllm's performance both with and without compile.

zhxchen17:
attached perf benchmark to PR summary:
```
torchrun --nproc_per_node=2     torchtitan/experiments/rl/unified/validation/bench_attn_perf.py
```

```
================================================================
  Attention Backend Performance Comparison
  Model: Qwen3-0.6B | TP: 2 | Prompts: 8 | Prompt len: 512
  Decode tokens: 128 | Warmup: 1 | Timed runs: 3
================================================================

Backend        Phase        Tokens     Time (s)      Tok/s
----------------------------------------------------------
FLASH_ATTN     Prefill        4096       0.2325      17615
FLASH_ATTN     Decode         1024      28.6333         36
CUSTOM         Prefill        4096       0.2300      17809
CUSTOM         Decode         1024      29.1132         35

Speedup (CUSTOM / FLASH_ATTN):
  Prefill: 1.01x
  Decode:  0.98x
================================================================
```

cc @Lucaskabela

torchtitan/experiments/rl/tests/test_attn_numerics.py
zhxchen17: ok I will add these to CI

torchtitan/experiments/rl/unified/validation/test_attn_numerics.py
zhxchen17: Yeah I think today vllm and varlen FA kernels produce bitwise identical result by accident. This is possible because vllm also forks FA kernel and only add page attention as an extension on top so likely the numerics are the same accidentally.

torchtitan/experiments/rl/unified/config_registry.py
zhxchen17: Yeah agree, I will make the switch in this PR then.

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: oops forgot to remove this.

torchtitan/experiments/rl/tests/test_attn_numerics.py
zhxchen17: Haven't looked closely into this, will verify with @liangel-02

zhxchen17:
> What can we do if we want to support both flex attention (in the future) and varlen attention as custom backend -- is the only option to share the "CUSTOM" name and dynamically register either one of them during construct time?

@tianyu-l Good point. For now vllm only supports 1 custom attention backend as far as I know. I don't think it will be hard to ask for a second slot in the upstream. 
Otherwise we can also patch the existing FlexAttentionBackend in vllm as a short term workaround if needed, also possible to only have 1 custom backend with dispatch to our own 2 backends.

torchtitan/experiments/rl/unified/config_registry.py
zhxchen17: the only thing not sure is on batch invariance. other feature should all work.

torchtitan/experiments/rl/unified/validation/test_attn_numerics.py
zhxchen17: removed this in favor of the default sdpa impl now.

wwwjn:
> > What can we do if we want to support both flex attention (in the future) and varlen attention as custom backend -- is the only option to share the "CUSTOM" name and dynamically register either one of them during construct time?
> 
> 
> 
> @tianyu-l Good point. For now vllm only supports 1 custom attention backend as far as I know. I don't think it will be hard to ask for a second slot in the upstream. 
> 
> Otherwise we can also patch the existing FlexAttentionBackend in vllm as a short term workaround if needed, also possible to only have 1 custom backend with dispatch to our own 2 backends.

Register the attention backend at construction time and share the same name should works fine, as long as the model doesn't use both varlen and flex attention at the same time(in the same model)

.github/workflows/integration_test_2gpu_rl_unified.yaml
```diff
@@ -0,0 +1,86 @@
...
+            --index-url ${{ matrix.index-url }}
+        else
+          python -m pip install --force-reinstall --pre \
+            torch --index-url ${{ matrix.index-url }}
+        fi
+
+        # Install RL dependencies: vLLM, monarch, flash-attn-3
+        pip install --pre vllm xformers --extra-index-url ${{ matrix.index-url }}
+        pip install torchmonarch
+        pip install flash-attn-3 --extra-index=https://download.pytorch.org/whl/test/cu126
```
wwwjn: Why Cuba 126 here, while other places in titan has been updated to cuda128

.github/workflows/integration_test_2gpu_rl_unified.yaml
```diff
@@ -0,0 +1,86 @@
...
+          python -m pip install --force-reinstall --pre \
+            "torch==${{ matrix.torch-version }}" \
+            --index-url ${{ matrix.index-url }}
+        else
+          python -m pip install --force-reinstall --pre \
+            torch --index-url ${{ matrix.index-url }}
+        fi
+
+        # Install RL dependencies: vLLM, monarch, flash-attn-3
+        pip install --pre vllm xformers --extra-index-url ${{ matrix.index-url }}
```
wwwjn: Can you double check and make sure this is installing vllm nightly , and the vllm nightly has correct dependency on PyTorch nightly?

torchtitan/experiments/rl/unified/README.md
```diff
@@ -25,6 +25,13 @@ source titan-rl/bin/activate
 uv pip install torchmonarch
 ```
 
+Install Flash Attention v3 kernels:
+```
+# CUDA 12
+pip install flash-attn-3 --extra-index=https://download.pytorch.org/whl/test/cu126
```
wwwjn: Can we use 128

torchtitan/experiments/rl/unified/tests/bench_attn_perf.py
wwwjn, line 1: Do we want to always guard attention performance in CI? If not, please remove this file

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -4,14 +4,162 @@
...
 from torch.distributed.tensor import DTensor
+from torch.nn.attention import activate_flash_attention_impl
 from torchtitan.experiments.rl.vllm_compat.models.attention import (
     VLLMCompatibleFlashAttention,
 )
+
+# Activate PyTorch's FA3 flash attention implementation. This is required for
+# the varlen attention API to work with paged attention (KV cache in block
+# layout) because vLLM uses default page size 16 but FA2 only supports
+# page_size=256. If FA3 is not available in the current environment, we silently
```
tianyu-l: Does FA2 supports KV cache? If not we shouldn't silently fall back?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -4,14 +4,162 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+"""Custom vLLM attention backend using PyTorch's native varlen flash attention.
```
tianyu-l: It's not proper to put this docstring at file head, because this file it NOT ONLY about this custom backend.

torchtitan/experiments/rl/unified/validation/test_attn_numerics.py
tianyu-l: > removed this in favor of the default sdpa impl now.

wait, are you saying that sdpa, vllm FA, varlen FA are all bitwise identical??

Without packing it is theoretically possible. With packing it is impossible because titan sdpa is doing all causal masking, instead of block causal masking used in varlen / vllm.

I think here we should change trainer model code to force attention config use "varlen" backend as well.

In the short term, we can force generator to use varlen. Let's error out if trainer model attention backend is not "varlen". This should happen in the future PR (if not this PR) that deprecates FLASH_ATTN patching.
In the longer term, with flex supported in generator, we should automatically replace generator model code with corresponding vllm attention backend.
cc @wwwjn

.github/workflows/integration_test_2gpu_rl_unified.yaml
```diff
@@ -0,0 +1,86 @@
...
+    tags:
+      - ciflow/8gpu/*
+    paths:
+      - 'torchtitan/experiments/rl/unified/**'
+      - '.github/workflows/integration_test_2gpu_rl_unified.yaml'
+  pull_request:
+    paths:
+      - 'torchtitan/experiments/rl/unified/**'
+      - '.github/workflows/integration_test_2gpu_rl_unified.yaml'
+  schedule:
```
wwwjn: I don't think we need regularly runs for RL now, removing this could relief the CI pressure a little bit

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -4,14 +4,162 @@
...
+
+This module registers a custom attention backend into vLLM's attention registry
+so that vLLM's inference engine uses PyTorch's native ``varlen_attn_out`` kernel
+(with paged KV cache support) instead of vLLM's built-in flash attention.
+
+This is based on PyTorch side PR https://github.com/pytorch/pytorch/pull/173212
```
wwwjn: Remove this line as this PR has been closed for other stack. 

And I think we'd better land this torchtitan PR after pytorch change to make sure the CI runs correctly

torchtitan/experiments/rl/tests/test_attn_numerics.py
wwwjn, line 1: I have a old PR to test numerics: https://github.com/pytorch/torchtitan/pull/2300/changes . I formatted it slightly different into several unit tests, not an end-to-end bit-wise identity check between (And why are you running bitwise identity check between trainer(titan) and generator (vllm)? This won't give you direct information which part is failing, right?)

Please give a look at that PR if that cover's your numerics test

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -4,14 +4,162 @@
...
+        block_table = attn_metadata.block_table
+
+        assert self.dcp_world_size == 1
+
+        if attn_metadata.causal:
+            sliding_window_size = (-1, 0)
+        else:
+            raise RuntimeError("Non-causal attention not supported yet.")
+
+        assert self.alibi_slopes is None
```
acisseJZhong: can we add error msgs for each of these assertions?

torchtitan/experiments/rl/models/attention.py
acisseJZhong: Image is this what you want?

torchtitan/experiments/rl/models/attention.py
acisseJZhong: I think get_name() can return any string as it's only used for logging purpose; as long as we have @register_backend(AttentionBackendEnum.CUSTOM), the registry lookup should work.

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -67,6 +215,7 @@ def __init__(
             quant_config=None,
             prefix=f"model.layers.{layer_name}.attention.inner_attention",
         )
+        self.vllm_attn = attention
```
acisseJZhong: nit: why this change?

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: FA2 doesn't support KV cache. Moving forward if we are going to make FA3 a hard requirement for RL branch, I can remove this.

torchtitan/experiments/rl/tests/test_attn_numerics.py
zhxchen17: Yeah I used your unit tests locally. I usually prefer e2e test but I can definitely incorporate your tests into this PR as well.

torchtitan/experiments/rl/models/attention.py
zhxchen17: I remember running into some issue changing this to anything other than "CUSTOM", but I can try again.

torchtitan/experiments/rl/unified/models/attention.py
zhxchen17: previously I was debugging the code and want to inspect what's inside the returned value.

torchtitan/experiments/rl/unified/models/attention.py
tianyu-l: Yeah I think without KV cache it's not useful, so let's have hard dependency on FA3.

torchtitan/experiments/rl/models/attention.py
zhxchen17: @acisseJZhong This is the error after I changed the name:
```
[rank0]:   File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/unified/tests/test_attn_numerics.py", line 273, in main
[rank0]:     engine = create_vllm_engine(config)
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/data/users/zhxchen17/torchtitan/torchtitan/experiments/rl/unified/tests/test_attn_numerics.py", line 88, in create_vllm_engine
[rank0]:     init_batch_invariance(AttentionBackendEnum[gen_config.attention_backend])
[rank0]:                           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/data/users/zhxchen17/vllm/vllm/v1/attention/backends/registry.py", line 28, in __getitem__
[rank0]:     raise ValueError(
[rank0]: ValueError: Unknown attention backend: 'CUSTOM_PYTORCH_VARLEN'. Valid options are: FLASH_ATTN, FLASH_ATTN_DIFFKV, TRITON_ATTN, ROCM_ATTN, ROCM_AITER_MLA, ROCM_AITER_TRITON_MLA, ROCM_AITER_FA, ROCM_AITER_MLA_SPARSE, TORCH_SDPA, FLASHINFER, FLASHINFER_MLA, FLASHINFER_MLA_SPARSE, TRITON_MLA, CUTLASS_MLA, FLASHMLA, FLASHMLA_SPARSE, FLASH_ATTN_MLA, NO_ATTENTION, FLEX_ATTENTION, TREE_ATTN, ROCM_AITER_UNIFIED_ATTN, CPU_ATTN, CUSTOM
```

zhxchen17:
benchmark script linked: https://gist.github.com/zhxchen17/a468937070cb808a307caf559acdd618

zhxchen17:
Update 03/15/26:
I've addressed minor comments locally. 
Currently working on switching to varlen API on the trainer side as well. (ETA 1-2 weeks)
Will also add batch variance support based on @liangel-02 's landed PR https://github.com/pytorch/pytorch/pull/176905 .
cc @wwwjn @tianyu-l

torchtitan/experiments/rl/unified/models/attention.py
tianyu-l: doesn't seem to be addressed

torchtitan/experiments/rl/tests/test_attn_numerics.py
wwwjn: let's remove this file for now, as it's testing numerics parity between trainer and generator, but we haven't support bit-wise identity yet. Would prefer adding this test back later

zhxchen17:
Discussed offline with @wwwjn , so we want to land the work piece by piece, meaning the following work will be landed separately:
- trainer switch to varlen backend
  - @tianyu-l to clarify, I'm aware we have support for this, but currently I saw some dtensor related error in unittests. I'm debugging this and will send in a separate PR.
- batch invariance mode support (and corresponding tests)

I just updated the PR addressing all the code level comments. Feel free to have another scan and see if it's landable. Thanks :)

torchtitan/experiments/rl/tests/test_attn_numerics.py
wwwjn: It might better to use a unit test to test only Attention part:
1. Attention Functionality test: It will be covered by e2e test / integration tests in #2300 
2.Attention numerics: Can you use a unit test to test "using vllm engine and PyTorchFlashAttentionImpl backend" has the same numerics with "vllm engine inference with vllm's FlashAttention"

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -4,15 +4,160 @@
...
 from torchtitan.experiments.rl.vllm_compat.models.attention import (
     VLLMCompatibleFlashAttention,
 )
 from torchtitan.protocols.module import Module
+
+# Activate PyTorch's FA3 flash attention implementation. This is required for
+# the varlen attention API to work with paged attention (KV cache in block
+# layout) because vLLM uses default page size 16 but FA2 only supports
+# page_size=256. Error out if FA3 is not available in the current environment.
+activate_flash_attention_impl("FA3")
```
tianyu-l: can we activate in the pytorch varlen attention module, instead of globally here?

torchtitan/experiments/rl/unified/config_registry.py
tianyu-l: should we switch this?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -4,15 +4,160 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+"""This module provides customized attention functions used for generators:
+  - A custom Attention Backend implemented with PyTorch's native varlen API.
+  - Adapter torch.nn.Module types for replacing the original attention modules.
+  - Transformations to apply attention module given root modules.
+"""
```
tianyu-l: This doesn't read well. Maybe delete and use docstring on each individual functions/modules.

wwwjn (approved):

PR #2366, opened 2026-02-10
Fixing extra averaging performed in validation error

Shagun-G:
Summary: The change to averaging over  microbatches to averaging over the global tokens in D91432940, the train loss computation was fixed to account of the change from mean to sum but not the validation loss computation, thus incurring an extra division computation in the validation loss. The operations here has been changed from a mean to a sum.

Differential Revision: D92888963


wwwjn (approved):
thanks for the fix

Issue #2312, opened 2026-02-02
torch.compile fails with DeepSeekV3 + SimpleFSDP


axeber01:
### Bug description

# Description
I am trying to run DeepSeekV3 with SimpleFSDP and torch.compile enabled, but the compilation fails. The error seems to be related to dynamic tensor shapes (the number of tokens per expert) in the MoE model.

# Error log

```
    File "/home/axel/ctml/torchtitan/torchtitan/train.py", line 696, in train      
      self.train_step(data_iterator)                                                                                                                                                 
    File "/home/axel/ctml/torchtitan/torchtitan/train.py", line 588, in train_step        
      loss = self.forward_backward_step(                                                  
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                             
    File "/home/axel/ctml/torchtitan/torchtitan/train.py", line 536, in forward_backward_step    
      pred = model_parts[0](inputs, **extra_inputs, **extra_kwargs)                                                                                                                  
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                  
    File "/home/axel/condaenv/envs/torchtitan/lib/python3.12/site-packages/torch/_dynamo/eval_frame.py", line 469, in __call__
      return super().__call__(*args, **kwargs)                                                                                                                                       
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                     
    File "/home/axel/condaenv/envs/torchtitan/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
      return self._call_impl(*args, **kwargs)                                                                                                                                        
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                        
    File "/home/axel/condaenv/envs/torchtitan/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1790, in _call_impl
      return forward_call(*args, **kwargs)                                                                                                                                           
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                           
    File "/home/axel/condaenv/envs/torchtitan/lib/python3.12/site-packages/torch/_dynamo/eval_frame.py", line 1006, in compile_wrapper
      raise e.remove_dynamo_frames() from None  # see TORCHDYNAMO_VERBOSE=1                                                                                                          
[...]

PR #2369, opened 2026-02-11
[Bugfix] Fix torch.compile with MoE for-loop path (#2312)

zpcore:
Replace the per-expert for-loop in `_run_experts_for_loop` with a padded batched matmul approach so that all intermediate shapes are static. The original implementation used `.tolist()` and Python `sum()` which create unbacked symbolic integers that `torch.compile(fullgraph=True)` cannot handle, breaking the SimpleFSDP + DeepSeek V3 path on pre-SM90 hardware.

The new implementation uses `torch.cumsum`, advanced indexing, `torch.bmm`, and `scatter_add_` — all tensor ops with no device-to-host sync — so the function is fully compile-compatible.

Authorized by Claude

### DIFF
--- a/torchtitan/models/moe/moe.py
+++ b/torchtitan/models/moe/moe.py
@@ -73,8 +73,10 @@ class FeedForward(nn.Module):
             trunc_normal_(linear.weight, mean=0.0, std=init_std)
 
 
-# NOTE: keeping this for-loop implementation for comparison
-#       and readability, may remove later
+# Compile-friendly padded batched matmul implementation (replaces the original
+# for-loop which used .tolist() / sum() that create unbacked symbolic integers).
+# All intermediate shapes are static: (E, T, D) and (E, T, H).
+# Data-dependent routing is expressed through index *values*, not shapes.
 def _run_experts_for_loop(
     w1: torch.Tensor,
     w2: torch.Tensor,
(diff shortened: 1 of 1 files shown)


zpcore:
Here's a full trace with `num_experts=3`, `total_tokens=7`, `dim=2`, and `num_tokens_per_expert=[2, 3, 2]`. The input `x` is already sorted by expert assignment:

```
x (7×2):                          num_tokens_per_expert = [2, 3, 2]
  row 0: [1,  2]  ← expert 0
  row 1: [3,  4]  ← expert 0
  row 2: [5,  6]  ← expert 1
  row 3: [7,  8]  ← expert 1
  row 4: [9, 10]  ← expert 1
  row 5: [11,12]  ← expert 2
  row 6: [13,14]  ← expert 2
```

---

**Step 1 — Offsets** (`cumsum`)

```
offsets = [0, 2, 5, 7]
              ↑  ↑  ↑
         expert starts: expert 0 starts at 0, expert 1 at 2, expert 2 at 5
```

---

**Step 2 — Counts, starts, token_range**

```
token_range = [0, 1, 2, 3, 4, 5, 6]     # (7,)
counts      = [[2], [3], [2]]            # (3, 1)
starts      = [[0], [2], [5]]            # (3, 1)
```

---

**Step 3 — `valid` mask** (`token_range < counts`, broadcasts to 3×7)

```
                  col:  0     1     2     3     4     5     6
expert 0 (2 tok):     [True  True  False False False False False]
expert 1 (3 tok):     [True  True  True  False False False False]
expert 2 (2 tok):     [True  True  False False False False False]
                       ← real tokens →   ← padding (zeroed out) →
```

---

**Step 4 — `idx` gather indices** (`(starts + token_range).clamp(max=6)`)

```
                  col:   0  1  2  3  4  5  6
expert 0 (start=0):     [0, 1, 2, 3, 4, 5, 6]   no clamping needed
expert 1 (start=2):     [2, 3, 4, 5, 6, 6, 6]   cols 5,6 clamped from 7,8 → 6
expert 2 (start=5):     [5, 6, 6, 6, 6, 6, 6]   cols 2–6 clamped from 7–11 → 6
                         ↑real↑  ↑ padding (clamped to last valid row) ↑
```

Clamping keeps indices in-bounds. The gathered values at padded positions don't matter because they'll be zeroed out.

---

**Step 5 — `x_batched`** (`x[idx] * valid`, shape 3×7×2)

```
expert 0:  [[ 1, 2], [ 3, 4], [0,0], [0,0], [0,0], [0,0], [0,0]]
expert 1:  [[ 5, 6], [ 7, 8], [9,10],[0,0], [0,0], [0,0], [0,0]]
expert 2:  [[11,12], [13,14], [0,0], [0,0], [0,0], [0,0], [0,0]]
             ← real data →          ← padded zeros →
```

`x[idx]` gathers real tokens into each expert's first few columns. The `* valid` mask zeros out everything past each expert's token count. Padded positions are all-zero, so they produce zero through the matmuls too (`0 @ W = 0`, `silu(0) = 0`).

---

**Step 6 — Batched FFN** (all experts in parallel, shape stays 3×7×2)

```python
h          = silu(bmm(x_batched, W1ᵀ))        # (3, 7, hidden)
h          = h * bmm(x_batched, W3ᵀ)           # (3, 7, hidden)  — SwiGLU
out_batched = bmm(h, W2ᵀ)                      # (3, 7, 2)
out_batched = out_batched * valid               # zero padding again for safety
```

Suppose the FFN produces (showing only real positions, padding is zero):

```
expert 0:  [[a, b], [c, d],  0,  0,  0,  0,  0]
expert 1:  [[e, f], [g, h], [i, j], 0,  0,  0,  0]
expert 2:  [[k, l], [m, n],  0,  0,  0,  0,  0]
```

---

**Step 7 — Scatter back** (`scatter_add_` using `idx`, output shape 7×2)

Each expert's output is scattered back to the original row positions using `idx`:

```
out = zeros(7, 2)

Expert 0 scatters to idx [0,1, ...]:  out[0] += [a,b],  out[1] += [c,d]
Expert 1 scatters to idx [2,3,4,...]: out[2] += [e,f],  out[3] += [g,h],  out[4] += [i,j]
Expert 2 scatters to idx [5,6, ...]:  out[5] += [k,l],  out[6] += [m,n]

Padded positions scatter 0 to clamped indices → no effect via scatter_add_

Final out (7×2):
  row 0: [a, b]    ← expert 0
  row 1: [c, d]    ← expert 0
  row 2: [e, f]    ← expert 1
  row 3: [g, h]    ← expert 1
  row 4: [i, j]    ← expert 1
  row 5: [k, l]    ← expert 2
  row 6: [m, n]    ← expert 2
```

---

**The padding trade-off**: Each expert's "batch" is padded to `total_tokens` (7), so expert 0 (2 real tokens) wastes 5 positions of compute. The entire `x_batched` tensor is `(3, 7, 2)` = 42 elements instead of the minimal 14. This is the memory/compute cost of keeping all shapes static for `torch.compile`.

zpcore:
So each expert's "batch" is padded to total_tokens to do the matmul. I am not sure if this trade off is acceptable...

zpcore:
Discussed offline with @tianyu-l , `_run_experts_for_loop` was intentionally written with for loop for readability instead of focusing on torch compile compatibility.

PR #2371, opened 2026-02-11
validate attn_type fix

francesco-bertolotti:
The `post_dataloading_process` function does not verify the model's attention type, whereas the trainer’s `post_dataloading_process` does.

Validating a model with `sdpa` triggers the following assertion error:

```python
File "/leonardo_scratch/fast/iGen_train/fbertolo/torchtitan/torchtitan/models/llama4/model/model.py", line 285, in forward
    assert attention_masks is None and self.attn_type == "sdpa"
AssertionError
```

The issue is resolved by adding a check for the attention type before generating the attention mask.

### DIFF
--- a/torchtitan/models/llama4/model/model.py
+++ b/torchtitan/models/llama4/model/model.py
@@ -577,9 +577,7 @@ class Transformer(ModelProtocol):
             case "sdpa":
                 raise TypeError("SDPA attention does not require attention masks")
             case _:
-                raise ValueError(
-                    f"Unknown attention type: {self.model_args.attn_type}"
-                )
+                raise ValueError(f"Unknown attention type: {self.model_args.attn_type}")
 
     def forward(
         self,


francesco-bertolotti:
Hi @fegin ,

just to confirm. Running the following should not trigger the assertion error?
```
NGPU=1 \
CONFIG_FILE="./torchtitan/models/llama4/train_configs/debug_model.toml" \
./run_train.sh \
    --model.flavor debugmodel \
    --training.local-batch-size 8 \
    --training.steps 20 \
    --training.seq-len 256 \
    --validation.seq-len 256 \
    --compile.no-enable \
    --training.dtype bfloat16 \
    --validation.freq 10 \
    --validation.enable
```

fegin:
Yes. I verified it and I saw the errors. The reason is that the `get_attention_masks` of llama4 doesn't correctly mimics what llama3 does. Do you want to change this PR to do this fix instead?

francesco-bertolotti:
I see. I will give it a shot.

francesco-bertolotti:
Hi @fegin, this should do it.

The Llama4 model's `get_attention_masks` method now:
1. Only generates masks for "flex" attention type
2. Raises TypeError for "sdpa"

PR #2373, opened 2026-02-12
[RL] Enable vLLM cudagraphs and compile in expriment/rl for efficiency

Lucaskabela:
## Summary

We utilize vLLM's compile and cudagraph integration in order to greatly speedup iteration time.

To do so, we use a compilation config with `full_and_piecewise` cudagraphs, and `eager` compilation (of note, `full_and_piecewise` is the most efficient for speed, but comes with extra memory overhead; `eager` is needed for compilation, and `inductor` impacts numerics so can't be used here)

Running vLLM inference for generating 20 tokens ~100 times, we see the following average difference:
```bash
SUMMARY

  Eager mode:    199.104 ms ± 4.669 ms
  Compiled mode: 51.841 ms ± 1.415 ms
  Speedup:       3.84x

  ✓ Generated tokens are identical
  ✓ Logprobs are nearly identical (max diff: 0.00e+00)
```

Similarly, in our testplan, we see ~4x speedup in iteration speed and tok/s with no impact on numerics

## Test plan

```bash
```

### Result (main)
```bash
✓ Created new vLLM engine
Adding requests: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 2411.35it/s]
Processed prompts: 100%|████████████████████████████████████████████████████████| 80/80 [00:03<00:00, 24.69it/s, est. speed input: 1609.70 toks/s, output: 2468.86 toks/s]
Adding requests: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 3464.36it/s]
Processed prompts: 100%|████████████████████████████████████████████████████████| 80/80 [00:02<00:00, 27.21it/s, est. speed input: 1774.20 toks/s, output: 2721.16 toks/s]

Step   0 | Loss: -0.0036 | Reward: +0.075 | Samples: 160
  Sample:  48 + 24 = 72.
Let me check the reasoning.
Okay, so Natalia sold clips to 48 of ...
Converted to 311 vLLM weights
```
### Result (these changes)
<TODO: Fill in>
```bash
```

### DIFF
--- a/torchtitan/experiments/rl/unified/infra/parallelism_utils.py
+++ b/torchtitan/experiments/rl/unified/infra/parallelism_utils.py
@@ -17,7 +17,7 @@ from torchtitan.config.job_config import Comm, JobConfig, Model, Parallelism, Tr
 from torchtitan.distributed import utils as dist_utils
 
 from torchtitan.distributed.parallel_dims import ParallelDims
-from vllm.config import VllmConfig
+from vllm.config import CompilationConfig, VllmConfig
 from vllm.logger import init_logger
 
 
@@ -137,3 +137,30 @@ def create_job_config_from_vllm_config(
     )
 
     return job_config
+
+
+def build_compilation_config(
+    vllm_compile_and_cudagraph: bool,
+    tp_size: int = 1,
+) -> CompilationConfig | None:
+    """Build a vLLM CompilationConfig for torch.compile + piecewise CUDA graphs.
+
+    Full CUDA-graph mode cannot capture DTensor operations (CPU metadata),
+    so we fall back to piecewise-only when tensor parallelism is active.
+    See https://docs.vllm.ai/en/stable/design/cuda_graphs/#cudagraphmodes
+
+    Args:
+        vllm_compile_and_cudagraph: Whether to enable compilation.
+            When False, returns None (enforce_eager should be True).
+        tp_size: Tensor parallel size. Controls CUDA-graph capture mode.
+
+    Returns:
+        CompilationConfig or None.
+    """
+    if not vllm_compile_and_cudagraph:
+        return None
+    cudagraph_mode = "piecewise" if tp_size > 1 else "full_and_piecewise"
+    return CompilationConfig(
+        cudagraph_mode=cudagraph_mode,
+        backend="eager",
+    )
--- a/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
@@ -11,8 +11,6 @@ This module provides TorchTitanVLLMModel: Core model class that adapts
 TorchTitan models for vLLM.
 """
 
-from functools import partial
-
 import torch
 import torch.nn as nn
 from torch.distributed._tensor import DTensor, Replicate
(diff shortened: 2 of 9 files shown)


torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -115,18 +112,28 @@ def __init__(
         )
 
     def _extend_rope_cache_if_needed(
```
Lucaskabela: This is the only real change needed for cudagraphs - we can't change the memory with cudagraphs so can no longer dynamically increase

The rest of this code is just wiring though  minor changes

torchtitan/experiments/rl/vllm_compat/simple_rl.py
```diff
@@ -38,13 +38,30 @@
...
+def build_compilation_config(
+    vllm_compile_and_cudagraph: bool,
+) -> CompilationConfig | None:
+    """Build a CompilationConfig when vllm_compile_and_cudagraph is enabled."""
+    if not vllm_compile_and_cudagraph:
+        return None
+    # NOTE: these options optimize speed while retaining numerics, but
+    # greatly increase memory usage
+    # see https://docs.vllm.ai/en/stable/design/cuda_graphs/#cudagraphmodes
+    return CompilationConfig(
```
wwwjn: Is this only enabling cuda graph, not even compile (frontend)?

Lucaskabela: The `backend="eager"` portion enables compile

As to why compile is needed here, it is an integral part of piecewise cudagraphs in vLLM.  This is because we use the graph captured by Dynamo and split around ops to get graphs for portions that otherwise would have been eager only 

See https://docs.vllm.ai/en/stable/design/cuda_graphs/#detailed-design for more discussion of this

wwwjn: Thanks, the name `backend=eager` means enable only enable Dynamo to capture the graph. 

But I'm curious our model is not decorated by `@support_torch_compile`, so will the model be able to captured by Dynamo? Or `support_torch_compile` is only used for backend like inductor?

Lucaskabela: @wwwjn This is a good question - `@support_torch_compile` is indeed needed on the lllm backbone for it to be compiled - however, it is not typically applied to the the top level nn.Module but instead a submodule (typically the llm backbone)

That said, since the underlying `self.model` we are putting in the TorchTitanVLLMWrapper does have @support_torch_compile enabled, it will be compiled (see Qwen3Model vs the Qwen3CausalModel definitions here https://github.com/vllm-project/vllm/blob/4c078fa546016eacab87f833ff625463421f7d29/vllm/model_executor/models/qwen3.py#L244) 

https://docs.vllm.ai/en/latest/design/torch_compile/#torchcompile-integration may also be helpful to read here for more info (but doesn't explicitly detail this)

wwwjn: Thanks, the current `self.model` is an instance of torchtitan Qwen3 instance https://github.com/pytorch/torchtitan/blob/27930cb612df4d6ebafdb21909749a9694bc167e/torchtitan/models/qwen3/model/model.py#L425 , not a vllm Qwen3 instance. So this model is not decorated by `support_torch_compile`

Lucaskabela: I believe that is the torchtitan (training) definiton, not the vLLM (RolloutEngine) definition. I see that vLLM is instantiated here: 
https://github.com/pytorch/torchtitan/blob/27930cb612df4d6ebafdb21909749a9694bc167e/torchtitan/experiments/rl/vllm_compat/simple_rl.py#L167

And when I check the call stack, I clearly see Qwen3 is being compiled

Here are logs and I can view the cached/traced artifact:
```bash
[0;36m(EngineCore_DP0 pid=740070)[0;0m WARNING 02-13 13:01:00 [decorators.py:547] Detected eager backend, disabling AOT compile.
[0;36m(EngineCore_DP0 pid=740070)[0;0m INFO 02-13 13:01:04 [backends.py:860] Using cache directory: /home/lucaskabela/.cache/vllm/torch_compile_cache/802ce93b23/rank_0_0/backbone for vLLM's torch.compile
[0;36m(EngineCore_DP0 pid=740070)[0;0m INFO 02-13 13:01:04 [backends.py:920] Dynamo bytecode transform time: 3.07 s
[0;36m(EngineCore_DP0 pid=740070)[0;0m INFO 02-13 13:01:05 [backends.py:317] Compiling a graph for compile range (1, 16384) takes 0.49 s
[0;36m(EngineCore_DP0 pid=740070)[0;0m INFO 02-13 13:01:05 [monitor.py:34] torch.compile takes 3.56 s in total
[0;36m(EngineCore_DP0 pid=740070)[0;0m INFO 02-13 13:01:07 [gpu_worker.py:359] Available KV cache memory: 19.59 GiB
```

Lucaskabela: Clarified: This is the case for the vllm_compat path which is not being actively invested in anymore - the unified path gives an error that the model is not being compiled, but full cudagraphs can still be applied

I will work on support_torch_compile to the unified version so that we can use full_and_piecewise there!

torchtitan/experiments/rl/unified/infra/parallelism_utils.py
```diff
@@ -137,3 +137,30 @@ def create_job_config_from_vllm_config(
...
+
+    Returns:
+        CompilationConfig or None.
+    """
+    if not vllm_compile_and_cudagraph:
+        return None
+    cudagraph_mode = "piecewise" if tp_size > 1 else "full_and_piecewise"
+    return CompilationConfig(
+        cudagraph_mode=cudagraph_mode,
+        backend="eager",
```
wwwjn: Also a n00b q: Does this "eager" backend means Dynamo captures the graph, but runs it via eager dispatch (no compilation)

PR #2376, opened 2026-02-12
initial skeleton

daniellepintz:


experiments/reinforcement_learning/training.py
```diff
@@ -0,0 +1,522 @@
...
+
+    stats.wall_time = time.perf_counter() - t0
+    return stats
+
+
+# ---------------------------------------------------------------------------
+# Async training loop
+# ---------------------------------------------------------------------------
+
+def run_async_loop(
```
joecummings, line 249: Let's remove async loop for now

experiments/reinforcement_learning/training.py
```diff
@@ -0,0 +1,522 @@
...
+    stop_flag.set()
+
+    for t in gen_threads:
+        t.join(timeout=2.0)
+
+    stats.wall_time = time.perf_counter() - t0
+    return stats
+
+
+# ---------------------------------------------------------------------------
```
joecummings, line 376: Remove all reporting code

experiments/reinforcement_learning/run.py
joecummings, line 1: Combine this file and the training.py file into one main.py

experiments/reinforcement_learning/environment.yml
```diff
@@ -0,0 +1,19 @@
+name: torchtitan-rl
+channels:
+  - defaults
+dependencies:
+  - python=3.12
+  - pip
+  - pip:
+      - --extra-index-url https://download.pytorch.org/whl/nightly/cu126
+      - torch
+      - marimo>=0.19.9
```
joecummings, line 10: don't need marimo here

experiments/reinforcement_learning/environment.yml
```diff
@@ -0,0 +1,19 @@
...
+      - transformers>=4.40.0
+      - torchmonarch==0.3.0
+      - accelerate
+      - html5lib
+      - matplotlib>=3.10.8
+      - scipy>=1.15.3
+      - torchshow>=0.5.2
```
joecummings, line 17: lol wth is torchshow? @allenwang28

experiments/reinforcement_learning/training.py
```diff
@@ -0,0 +1,522 @@
...
+    simultaneously, waits for ALL generators to finish, then trains.
+    """
+    print("\n" + "=" * 60)
+    print("SYNC MODE: Broadcast Generate -> Train")
+    print("=" * 60)
+
+    trainer = actors["trainer"]
+    generators = actors["generators"]
+
+    stats = TimingStats(
```
joecummings, line 174: Rip out all this TimingStats stuff

experiments/reinforcement_learning/environment.yml
joecummings, line 1: IIUC this is just for the experiment, not the environment?

experiments/reinforcement_learning/src/monarch_utils/__init__.py
joecummings, line 1: Keep everything flat, no need to nest files yet

joecummings, line 1: Put all of this under torchtitan/experiments/rl/FOLDER_NAME

Feeling uncreative so idk about folder name but this should live amongst the rest of the RL stuff for now

experiments/reinforcement_learning/actors.py
```diff
@@ -0,0 +1,565 @@
...
+            "size": len(self.buffer),
+            "total_added": self.total_added,
+            "avg_reward": sum(rewards) / len(rewards),
+            "correct_rate": sum(1 for t in self.buffer if t.is_correct) / len(self.buffer),
+            "format_rate": sum(1 for t in self.buffer if t.has_answer_tag) / len(self.buffer),
+            "failure_modes": failure_modes,
+        }
+
+
+class TrainerActor(Actor):
```
joecummings, line 139: Just call this trainer

experiments/reinforcement_learning/actors.py
```diff
@@ -0,0 +1,565 @@
...
+            loss=avg_loss.item() if torch.is_tensor(avg_loss) else avg_loss,
+            batch_size=len(trajectories),
+            avg_reward=avg_reward,
+            policy_version=self.policy_version,
+            correct_rate=correct_rate,
+            format_rate=format_rate,
+        )
+
+    @endpoint
+    def evaluate_zorplex(self, num_samples: int = 10, seed: int = 42) -> dict:
```
joecummings, line 337: Put this in the controller code IMO

experiments/reinforcement_learning/src/zorplex_rl/task_specs.py
```diff
@@ -0,0 +1,608 @@
...
+
+
+# =============================================================================
+# REGISTRY
+# =============================================================================
+
+TASK_SPECS = {
+    "simple": SimpleLookupSpec,
+    "compositional": CompositionalSpec,
+    "multi_step": MultiStepSpec,
```
joecummings, line 582: Delete all reference to multi_step and recursive

experiments/reinforcement_learning/src/zorplex_rl/task_specs.py
```diff
@@ -0,0 +1,608 @@
...
+
+
+def get_spec(name: str, **kwargs) -> TaskSpec:
+    """Get a task spec by name."""
+    if name not in TASK_SPECS:
+        raise ValueError(f"Unknown task spec: {name}. Available: {list(TASK_SPECS.keys())}")
+    return TASK_SPECS[name](**kwargs)
+
+
+if __name__ == "__main__":
```
joecummings, line 594: no

experiments/reinforcement_learning/src/zorplex_rl/evaluate.py
joecummings, line 1: combine all this task specific stuff into one zorplex file

joecummings (changes requested):
Stopping on comments here - essentially just make everything even simpler. Flat structure, no complex environments, minimal printing.

experiments/reinforcement_learning/src/monarch_utils/services.py
joecummings, line 1: Should we consider no services in this initial skeleton? I think it's the right abstractions for large scale training but in the spirit of no holds barred minimal skeleton, I don't think we need them.

experiments/reinforcement_learning/src/zorplex_rl/task_specs.py
```diff
@@ -0,0 +1,608 @@
...
+class EvalResult:
+    """Result of evaluating a single task."""
+    task: Task
+    model_output: str
+    tool_calls: list[ToolCall]
+    extracted_answer: int | str | None
+    is_correct: bool
+
+
+class TaskSpec(ABC):
```
joecummings, line 45: No need to generalize here - we only have on taskspec

experiments/reinforcement_learning/src/zorplex_rl/evaluate.py
felipemello1: i am tempted to say that we should delete all the zorplex and just make it gsm8k, single turn, no tool call or env or anything like that.

experiments/reinforcement_learning/src/rl_primitives.py
```diff
@@ -0,0 +1,42 @@
...
+Provides common dataclasses used across RL training:
+- Trajectory: A single rollout from generation
+- TrainMetrics: Metrics from a training step
+"""
+
+from dataclasses import dataclass, field
+
+
+@dataclass
+class Trajectory:
```
felipemello1, line 13: if we agree to do single_turn + gsm8k, this can all be much simpler, maybe bring from forge?

experiments/reinforcement_learning/src/monarch_utils/services.py
```diff
@@ -0,0 +1,192 @@
...
+    _get_registry().register.call_one(name, service).get()
+
+
+def list_services() -> list[str]:
+    """List all registered service names."""
+    return _get_registry().list_services.call_one().get()
+
+
+class Service(Actor):
+    """Manages a pool of worker replicas with routing and health tracking.
```
tianyu-l, line 60: What is a worker replica and why do we need them, is it for fault tolerance?

Do we need fault tolerance in very first version?

experiments/reinforcement_learning/src/monarch_utils/services.py
```diff
@@ -0,0 +1,192 @@
...
+Provides:
+- ServiceRegistry: Singleton for service discovery
+- Service: Manages worker replicas with health tracking and routing
+- get_service/register_service: Helper functions for discovery
+"""
+
+from monarch.actor import Actor, endpoint, get_or_spawn_controller
+
+
+class ServiceRegistry(Actor):
```
tianyu-l, line 13: Is this absolutely needed? If so, why Monarch doesn't provide this API?

experiments/reinforcement_learning/src/zorplex_rl/__init__.py
tianyu-l, line 1: Please justify the importance of this folder. If possible, replace with something absolutely minimal.

experiments/reinforcement_learning/src/rl_primitives.py
tianyu-l: Would like to see an absolute minimal version.

Also, @allenwang28 told us there shouldn't be Trajectory, there should only be Episodes. Please educate.

experiments/reinforcement_learning/actors.py
```diff
@@ -0,0 +1,565 @@
...
+- GeneratorWorker: Generates trajectories with tool-augmented inference
+"""
+
+from collections import deque
+import random
+
+import torch
+import torch.nn.functional as F
+
+from transformers import AutoModelForCausalLM, AutoTokenizer
```
tianyu-l, line 17: Can we not depend on transformers?

experiments/reinforcement_learning/training.py
```diff
@@ -0,0 +1,522 @@
...
+)
+from monarch_utils.services import Service, register_service
+
+
+# ---------------------------------------------------------------------------
+# Data classes
+# ---------------------------------------------------------------------------
+
+@dataclass
+class TimingEvent:
```
tianyu-l, line 33: can these be added later?

experiments/reinforcement_learning/training.py
```diff
@@ -0,0 +1,522 @@
...
+    buffer_max_size: int = 500,
+) -> dict:
+    """Spawn and initialize all actors.
+
+    Returns a dict with keys: trainer, buffer, generators, zorplex_svc, _proc_meshes.
+    """
+    host = this_host()
+
+    # 1. ZorplexWorkers -- wrapped in a Service for health tracking and round-robin
+    zorplex_worker_procs = host.spawn_procs(per_host={"procs": num_zorplex})
```
tianyu-l, line 81: I don't think we should hard code zorplex into train.py code.

experiments/reinforcement_learning/actors.py
tianyu-l, line 1: This file needs full documentation.

allenwang28:
Thanks @daniellepintz for taking a first stab at this! I want to share more thoughts on scoping - not really about code quality, but on what the right size is for landing this quickly, then iterating from there.

The goal of this skeleton, as I see it:
- Something we're totally fine with throwing away in 2 months
- Useful enough that we can A/B test changes against it (e.g. what are the learning dynamic differences with and without unified model)? 
- Simple enough that someone unfamiliar with Forge/Monarch or the intricacies of async RL can review it in one sitting
- Not so toy that everything works by accident, but not so complex that it takes days to run to validate.

But before that, I would *start with evaluation*, not training. The first thing to get right isn't the RL loop — it's confirming a concrete hypothesis about what's learnable.

My hypothesis: a model like Qwen 2.5 0.5B can already do lookups consistently — it knows how to call LOOKUP[word] and use the result. But it doesn't emit [ANSWER] tags reliably. That gap — "can use tools but can't format a final answer" — should be learnable through RL without needing SFT first, because the behavior is simple and the reward signal is unambiguous.

So the first milestone is: set up an evaluation framework, run Zorplex tasks at the simplest difficulty, and confirm that the model actually has this gap. If it does, that's our starting point — RL to teach answer formatting on a task where the model already has the underlying capability. Then we incrementally up the difficulty. (Another note - this naturally opens a door for curriculum learning as well down the line)

We ran into exactly this problem with GSM8K before: the model was already trained on it, so it was impossible to tell whether RL was actually teaching anything or just reformatting answers the model already knew. Zorplex avoids this because the values are random and arbitrary — the model literally cannot know the answers without using the tool.

A note on task selection more broadly: this is partially a research question. There's real work (DeepSeek R1, etc.) showing that you often need SFT first to teach format compliance, then RL to improve reasoning within that format. We probably won't find the perfect task immediately, and we should keep iterating on difficulty and task design as we go. But we don't need the perfect task to land the skeleton — we need one task where we have a testable hypothesis and a clean eval. Zorplex SimpleLookup + answer formatting is that task.

Then the skeleton becomes a single file:
1. Zorplex SimpleLookup spec
2. Single-turn
3. REINFORCE training step, the simplest math that works
4. Sync loop only (generate, train, sync weights, repeat)
5. Weight sync is only done via load_state_dict through DCP. It's slow but it's correct, no extra zero infra
6. Console logging of loss + reward + accuracy per step
7. Held-out evaluation every N steps with a fixed seed, printing example trajectories so a human can see what the model is doing. 

**What we add in future PRs**
These are valuable, but I'm proposing we land them in isolated PRs where the review questions are focused and answerable against the skeleton baseline:
|Feature|Review questions it answers|
|-|-|
|Multi-turn agentic generation|Can the model learn to chain multiple tool calls and combine results?|
|Service / ServiceRegistry / everything in monarch_utils|How do we handle Monarch actor pools?|
|RDMA circular buffer weight sync, building up to torchstore|Does RDMA give the same results as state_dict / what is the speed up?|
|Async loop + ReplayBuffer|Does async converge to the same accuracy as sync? What is the throughput gain?|
|Additional task specs / difficulty tiers|How does difficulty scaling affect learning dynamics?|
|Timeline plotting|Lands naturally with the async PR|

Thoughts?

felipemello1:
My only comments would be about (1) and (3)

(1) IMO we should use something that is in other frameworks, so we can compare. Do you know if any other framework uses zorplex.

I like that we can see the curves from prime-rl here: https://github.com/PrimeIntellect-ai/prime-rl/tree/main/examples/reverse_text

But, like you said, they do SFT before, which makes the example less interesting for a skeleton.

I also think that the tool_call, env abstractions, etc, can add some necessary overhead atm. But i can be convinced.

(3) I looked at the losses a lot, and i think that simple reinforce will take much longer to converge and debug. Being able to check entropy, clip ratio, all other metrics is also very helpful.

IMO we should just copy from froge the DAPO/GRPO loss. Its a low hanging fruit and shouldnt add complexity to the skeleton.

(7) extra: Eval might add extra complexity that we dont need atm. Looking at rewards may be good enough.

allenwang28:
@felipemello1 
> (1) IMO we should use something that is in other frameworks, so we can compare. Do you know if any other framework uses zorplex.

Fair point, no AFAIK no one uses Zorplex 😁 you're right that it's useful to have a comparison point that we can sanity check against. Maybe something like Zorplex already exists and we should consider using that. 

But for the broader point of sanity checking correctness - yes it will be required, but perhaps we can plug it in later after the skeleton implementation.

 > IMO we should just copy from forge the DAPO/GRPO loss. Its a low hanging fruit and shouldnt add complexity to the skeleton

that's fair, but a few thoughts..
1) Ideally the task we start with should be easy enough that REINFORCE can apply. If it's not, probably need to consider a different task
2) GRPO builds on REINFORCE and DAPO builds on GRPO and the upgrade route should be clear

in other words, we should do these things but IMO as fast follows after the basic skeleton

felipemello1:
> Maybe something like Zorplex already exists and we should consider using that.

I think its worth checking and, if we can get away with not having to think about env/tools for basic skeleton, i think it can make the PR simpler and discuss tool/env abstractions in a follow up, IMO.

> in other words, we should do these things but IMO as fast follows after the basic skeleton

agreed

daniellepintz:
Honestly I really like the Zorplex task, I think it is very clear and easily understandable, and I like that there is no way the model can have memorized it beforehand. Also I like defining it right in the source code itself rather than using something external, because again I think it makes it more understandable for users

daniellepintz:
@allenwang28 thanks for your comments! Agree with everything, and it makes sense to separate out the skeleton and the subsequent PRs as you mentioned

experiments/reinforcement_learning/src/monarch_utils/services.py
daniellepintz: Removed this. Yea this is for fault tolerance, so we can have multiple replicas of an actor, for example multiple Generators running generations in parallel

experiments/reinforcement_learning/src/monarch_utils/services.py
daniellepintz: Removing

tianyu-l:
@allenwang28 thanks a lot for helping understand the selection of tasks!

After studying the new PR https://github.com/pytorch/torchtitan/pull/2381, I realized that even the in-house zorplex tasks assume HF transformers model protocols (such as `model.generate`, `tokenizer.apply_chat_template`), which seems breaking the unified model assumption today. cc @wwwjn

experiments/reinforcement_learning/actors.py
felipemello1: we could need to depend only on tokenizers, which titan also uses, and have the generator be the vllm as in the titan

daniellepintz:
Closing in favor of https://github.com/pytorch/torchtitan/pull/2381

Issue #1935, opened 2025-10-25
deepseek compile fail to run


BoyuanFeng:
### Bug description

Command:
`CONFIG_FILE="/data/users/boyuan/torchtitan/torchtitan/models/deepseek_v3/train_configs/deepseek_v3_16b.toml" ./run_train.sh --model.name simple_fsdp.deepseek_v3 --compile.enable`

Log:
```
traceback : Traceback (most recent call last):
    File "/data/users/boyuan/pytorch/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 362, in wrapper
      return f(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^
    File "/data/users/boyuan/torchtitan/torchtitan/train.py", line 611, in train
      self.train_step(data_iterator)
    File "/data/users/boyuan/torchtitan/torchtitan/train.py", line 511, in train_step
      loss = self.forward_backward_step(input_dict, labels)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/data/users/boyuan/torchtitan/torchtitan/train.py", line 487, in forward_backward_step
      pred = model_parts[0](inputs, **extra_inputs, **extra_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/data/users/boyuan/pytorch/torch/_dynamo/eval_frame.py", line 433, in __call__
      return super().__call__(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/data/users/boyuan/pytorch/torch/nn/modules/module.py", line 1780, in _wrapped_call_impl
      return self._call_impl(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/data/users/boyuan/pytorch/torch/nn/modules/module.py", line 1791, in _call_impl
      return forward_call(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/data/users/boyuan/pytorch/torch/_dynamo/eval_frame.py", line 922, in compile_wrapper
      raise e.with_traceback(None) from e.__cause__  # User compiler error
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2377, opened 2026-02-13
[compiler toolkit] Add graph based SAC

yiming0416:
Add graph-based SAC as a joint graph pass in alternative to eager-mode AC. Previously eager AC and fullmodel trace don't compose in DSv3 (See https://github.com/pytorch/torchtitan/issues/1935). With graph based AC, it's no longer an issue.

Set `--compile.joint_passes apply_sac` to turn it on

Llama3-8B (TP=2, FSDP=2)
no-AC: memory: 67.19GiB(37.67%)  tps: 9,569  tflops: 554.17  mfu: 24.63%
eager SAC: memory: 45.88GiB(25.73%)  tps: 8,627  tflops: 499.62  mfu: 22.21%
graph SAC: memory: 44.20GiB(24.78%)  tps: 8,007  tflops: 463.74  mfu: 20.61%

DeepSeek-v3 Debug model(TP=2, FSDP=2, EP=2)
no-AC: memory: memory: 12.26GiB(6.88%)  tps: 59,927  tflops: 32.54  mfu: 1.45%
eager SAC: doesn't work (due to AC HOP mutation issue https://github.com/pytorch/torchtitan/issues/1935)
graph SAC:  memory: 11.01GiB(6.17%)  tps: 53,939  tflops: 29.29  mfu: 1.30%

### DIFF
--- a/torchtitan/experiments/compiler_toolkit/passes.py
+++ b/torchtitan/experiments/compiler_toolkit/passes.py
@@ -14,7 +14,7 @@ Pass Types:
 - Joint custom passes: Applied to the joint forward-backward graph before partitioning
 - Compiler passes: Applied to the partitioned forward/backward graphs
 """
-
+import operator
 from collections.abc import Sequence
 from typing import Any
 
@@ -28,6 +28,7 @@ from torch._inductor.fx_passes.bucketing import (
 from torch._inductor.fx_passes.overlap_manual_scheduling import manual_overlap_bucketing
 from torch._inductor.fx_passes.overlap_scheduling import schedule_overlap_bucketing
 from torch.fx.passes.regional_inductor import regional_inductor
+from torch.utils.checkpoint import CheckpointPolicy
 from torchtitan.experiments.compiler_toolkit.cudagraph import (
     CUDAGraphWrapper,
     get_static_input_indices,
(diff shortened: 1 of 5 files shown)


torchtitan/experiments/compiler_toolkit/passes.py
```diff
@@ -103,6 +105,91 @@ def validate_flex_attn_annotation_pass(
...
+    torch.ops.aten._scaled_dot_product_fused_attention_overrideable.default,
+    torch.ops._c10d_functional.reduce_scatter_tensor.default,
+    torch.ops.aten.max.default,
+    torch._higher_order_ops.flex_attention,
+    torch.ops.torch_attn._varlen_attn,
+    torch._higher_order_ops.inductor_compiled_code,
+}
+
+
+def apply_sac_pass(
```
aditvenk, line 126: How will user pass the op_list_to_save from command line?

Can we re-use the existing AC config system to configure `apply_sac_pass`?

yiming0416: Currently the ops to be saved in SAC are in a hardcoded list for each model. 
Not sure if it's a good UX for user to pass from command line though, as the list itself could be pretty long and not cli-friendly

aditvenk (approved):

PR #2378, opened 2026-02-13
[WIP] FlexShard

weifengpy:
# Eager reshard=True (1.63 GiB)                                                                          
  NGPU=4 MODULE=graph_trainer.flex_shard_llama3 CONFIG=graph_trainer_flex_shard_llama3_debugmodel \          
    ./run_train.sh --training.local_batch_size 8 --training.seq_len 6144 --training.steps 5 \                
    --compile.mode None                                                                                      
                                                                                                             
  # Eager reshard=False (4.33 GiB)                                                                           
  NGPU=4 MODULE=graph_trainer.flex_shard_llama3 CONFIG=graph_trainer_flex_shard_llama3_debugmodel \        
    ./run_train.sh --training.local_batch_size 8 --training.seq_len 6144 --training.steps 5 \                
    --compile.mode None --parallelism.fsdp_reshard_after_forward never
                                                                                                             
  # JIT reshard=True (1.14 GiB)                                                                              
  NGPU=4 MODULE=graph_trainer.flex_shard_llama3 CONFIG=graph_trainer_flex_shard_llama3_debugmodel \          
    ./run_train.sh --training.local_batch_size 8 --training.seq_len 6144 --training.steps 5 \                
    --compile.mode jit --compile.backend inductor                                                            
   
  # JIT reshard=False (3.00 GiB)                                                                             
  NGPU=4 MODULE=graph_trainer.flex_shard_llama3 CONFIG=graph_trainer_flex_shard_llama3_debugmodel \        
    ./run_train.sh --training.local_batch_size 8 --training.seq_len 6144 --training.steps 5 \                
    --compile.mode jit --compile.backend inductor --parallelism.fsdp_reshard_after_forward never             
                                                                                                             
  # AOT reshard=True (2.53 GiB)                                                                              
  NGPU=4 MODULE=graph_trainer.flex_shard_llama3 CONFIG=graph_trainer_flex_shard_llama3_debugmodel \        
    ./run_train.sh --training.local_batch_size 8 --training.seq_len 6144 --training.steps 5 \                
    --compile.mode aot                                                                                       
   
  # AOT reshard=False (2.53 GiB)                                                                             
  NGPU=4 MODULE=graph_trainer.flex_shard_llama3 CONFIG=graph_trainer_flex_shard_llama3_debugmodel \        
    ./run_train.sh --training.local_batch_size 8 --training.seq_len 6144 --training.steps 5 \                
    --compile.mode aot --parallelism.fsdp_reshard_after_forward never

<img width="645" height="134" alt="Screenshot 2026-04-09 at 16 23 32" src="https://github.com/user-attachments/assets/58abad9d-1e7b-43be-b3a1-d7a44c8f1fc2" />

    
                                                                                                                                                                                                                          
  Loss comparison (numerics, not traces):                                                                  
  python scripts/loss_compare.py . . \                                                                       
    --baseline-module=graph_trainer.flex_shard_llama3 \                                                    
    --test-module=graph_trainer.flex_shard_llama3 \                                                          
    --baseline-options="--compile.mode=none" \                                                               
    --test-options="--compile.mode jit --compile.backend aot_eager"     


AIling's original PR: https://github.com/pytorch/pytorch/compare/main...ailing/mst#diff-e326c855a215daccfd35d8953f8ce5f7c75cf2c446eb9880b22b512a8b1d7158


weifengpy:
I am polishing MST to be landable to experiment folder. DTensor RNG, custom placement resitration function has to be living in pytorch. I am patching DTensor for now

torchtitan/experiments/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,1674 @@
...
+    def state(self) -> ShardedState:
+        """Current state (SHARDED or UNSHARDED)."""
+        return self._state
+
+    @property
+    def world_size(self) -> int:
+        """World size of the mesh."""
+        return self._mesh.size()
+
+    def get_local_view(self, fqn: str) -> torch.Tensor:
```
wconstab, line 2739: are you always allowed to call this? or do you have to be in the 'sharded' state first?

wconstab (commented):
I think I understand some key points though I skipped over a lot of the code.
1) the DStorage is created after first creating a byte-storage, then slicing views out of it per-parameter, and wrapping those views in DTensor
2) the actual sharding seems handled by DTensor placements.  Owned is a special case, but Shard() is the other case, handling even split.  I didn't notice examples of uneven or unusual splits.
3) I didn't see an example of how this composes with tensor parallelism.  If we do TP 'first' and then apply fully_shard second, I think this proposal, like current fully_shard, would need to use StridedShard placement to deal with the ordering. or did I misunderstand?
4) since the module fqn points to a DTensor which is a view into a storage, i think checkpointing stuff should all 'just work' since it piggyback's onto DTensors infra. otoh, this also means it requires DTensor to continue knowing about any fsdp sharding types.  Maybe same for RNG, though I am not sure if Owned needs additional help to work correctly.

torchtitan/experiments/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,1674 @@
...
+        # Register post-backward callback (must be done during backward)
+        self._register_post_backward_callback()
+        # Params are already unsharded from forward, no need to unshard again
+        return grad
+
+    def _register_post_backward_callback(self) -> None:
+        """Register callback to run after backward completes."""
+        if self._post_backward_called:
+            return
+        Variable._execution_engine.queue_callback(self._post_backward)
```
zpcore: Oh, IIUC, the code conducts ReduceScatter for all parameters after the full gradient graph finishes the computation. It doesn't overlap the gradient reduce scatter of layer N with grad computation of layer N-1 (also the reduce_scatter_tensor triggered is not async). Does this show a higher peak memory than FSDP2?

torchtitan/experiments/flex_shard/__init__.py
tianyu-l, line 1: Maybe just call it dstorage instead of mst?

Also we should discuss the oss strategy for it. In torchtitan we don't want core folders to depend on experiments. It would be easier it the API can be drop-in placement to FSDP2; if not, I'm OK with swapping FSDP2 with this new version and keep iterating.

torchtitan/experiments/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,1674 @@
...
+# Module attribute name for storing DStorage
+_DSTORAGE_ATTR = "_dstorage"
+
+
+def get_dstorage(module: nn.Module) -> DStorage | None:
+    """Get the DStorage associated with a module, if any."""
+    return getattr(module, _DSTORAGE_ATTR, None)
+
+
+class Owned(Placement):
```
tianyu-l, line 1791: What's the scope of this implementation?

I think eventually we should have clear separation at several abstraction levels
- DStorage should be highly customizable. The compute side must be a DTensor, but the storage side doesn't have to be a DTensor -- it should just have DCP-like utils to assemble what it receives before apply sharding. Maybe `Placement` is the right abstraction, maybe it's not. `Owned` here seems a special example, rather than a generic framework.
- DStorage should be decoupled from actual optimized implementation like FSDP2. E.g. there should be at least two files, one for FSDP2-like optimization (prefetching, stream management, memory management, etc.), the other for general DStorage which can be used by SimpleFSDP or any compile friendly code.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,357 @@
...
+#### Parameters
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `module` | `nn.Module` | required | Module to shard |
+| `mesh` | `DeviceMesh` | required | 1D device mesh for sharding |
+| `placements` | `tuple[Placement, ...]` | `(Shard(0),)` | Default sharding placement |
+| `reshard_after_forward` | `bool` | `True` | Reshard after forward for memory savings |
+| `register_hooks` | `bool` | `True` | Auto-register forward/backward hooks |
+| `shard_strategy` | `str` | `"per_param"` | `"per_param"` or `"param_boundary"` |
```
tianyu-l: related to the comment above, I think this API is mixing multiple level of abstractions together. `placements` and `shard_placement_fn` should be DStorage-level concept, and `shard_strategy` sounds high-level syntactic sugar. All of them sound overlapping. I would encourage us to reinvent the API if necessary.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,357 @@
...
+
+| Method | Description |
+|--------|-------------|
+| `unshard()` | All-gather byte buffer, swap to unsharded params |
+| `reshard()` | Reduce-scatter gradients, restore sharded DTensors |
+| `get_local_view(fqn)` | Get typed tensor view from sharded storage |
+| `get_unsharded_view(fqn)` | Get typed tensor view from unsharded storage |
+| `all_gather()` | Perform all-gather, return unsharded byte buffer |
+
+## Sharding Strategies
```
tianyu-l: We should have at least two prototypical implementations to verify it works with various parallelisms and DCP
1. flat-parameter sharding (for extreme efficiency?)
2. Muon optimizer which put full param on DP ranks.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,357 @@
...
+chunk_cat(grads, num_chunks=world_size)
+reduce_scatter_tensor(output, input, op=AVG)
+```
+
+**For Owned params:** Point-to-point reduce to owner:
+```python
+reduce(grad, dst=owner_rank, op=AVG)
+```
+
+## Nested Wrapping
```
tianyu-l: For Muon, broadcast may be more efficient than all-gather-v? Also we can't do average splitting on each TransformerBlock layer, because the number of parameters in a Transformer can be very limited. So the UX of bucketing and UX of DStorage must be able to decouple.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,357 @@
...
+
+| Feature | DStorage | FSDP2 |
+|---------|----------|-------|
+| Storage | Unified byte buffer | Per-FSDPParam flat storage |
+| Collectives | Batched per-DStorage | Batched per-FSDPState |
+| Mixed dtypes | Single buffer with alignment | Separate buffers |
+| Placement | Shard + Owned | Shard only |
+| Complexity | Simpler, single abstraction | More features (prefetch, etc.) |
+
+## Future Work
```
tianyu-l: Is DCP supported in this version?

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,357 @@
...
+
+## Comparison with FSDP2
+
+| Feature | DStorage | FSDP2 |
+|---------|----------|-------|
+| Storage | Unified byte buffer | Per-FSDPParam flat storage |
+| Collectives | Batched per-DStorage | Batched per-FSDPState |
+| Mixed dtypes | Single buffer with alignment | Separate buffers |
+| Placement | Shard + Owned | Shard only |
+| Complexity | Simpler, single abstraction | More features (prefetch, etc.) |
```
tianyu-l: Does this version supports prefetch?

torchtitan/experiments/flex_shard/flex_shard.md
rakkit: to make muon efficient, ideally we want fine-grained split. i.e. per weights matrix.  for transformer its about (QKVO + W123) * layers.

split by Transformer blocks -> lots of redundant compute

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+## API
+
+### `flex_shard(module, mesh, **options) → FlexShardModule`
+
+```python
+from torchtitan.experiments.flex_shard import flex_shard
+
+# Basic usage
+mesh = init_device_mesh("cuda", (world_size,))
+flex_shard(model, mesh)
```
ailzhang: nit: would be nice to explain what flex_shard did to the `model` on the `mesh` in one line comment here

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+
+# model is now a FlexShardModule
+model.unshard()   # all-gather (for prefetching / manual control)
+model.shard()     # release unsharded parameter, restore sharded state
+
+# Nested wrapping (inner-first)
+for layer in model.layers:
```
ailzhang: nit: whether `flex_shard` handles nested wrapping or users need to do this manually need to be listed in the docstring for `flex_shard`.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+    flex_shard(layer, mesh)
+flex_shard(model, mesh)  # Only wraps root params
+```
+
+## Why not DTensor
+
+DStorage manages the full parameter lifecycle: hooks handle unshard (all-gather) before compute and reshard (reduce-scatter) after backward. The actual forward/backward runs on plain unsharded tensors. DTensor's dispatch machinery is never used, making it pure overhead:
+
+| | DTensor | Tensor + hidden attrs |
```
ailzhang: note to self: check the impl, tensor + hidden attrs is not user friendly. Users at least to know whether they should expect the hidden attrs to exist or not. 
iiuc this should be a plain tensor subclass to make it explicit to users.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+|---|---|---|
+| `__torch_dispatch__` | Active — intercepts every op | None — zero dispatch overhead |
+| `torch.compile` | Graph breaks from dispatch + `.data` + `resize_` | Clean tracing — plain tensor, no subclass |
+| Placement info | `param.placements`, `param._spec` | `get_placements(param)`, `get_global_shape(param)` |
+| DCP integration | Native `DTensor.__create_chunk_list__` | Custom handler using hidden attributes |
```
ailzhang: for the new impl this is still a `__create_chunk_list__` like method right?

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+## Memory Layout
+
+Buffer is organized by region for mixed Shard/Owned/RaggedShard placements:
+
+```
+┌─────────────────────────────────────────────────────────────────────────────────────┐
+│                               Byte Storage                                          │
+├─────────────────────────┬──────────────────────────┬───────────────────────────────┤
+│     Shard Region        │   RaggedShard Region     │      Owned Region             │
+│  (symmetric sharding)   │  (asymmetric sharding)   │  (full param on one rank)     │
```
ailzhang: qq: does symmetric/asymmetric mean even/uneven? 
is it for perf reason that we separate these 3 scenarios? e.g. symmetric and owned sharding are both special cases of ragged sharding here. I was curious if we are able to express the layout in general without depending on their shard/ragged/owned property.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+  comm:   all_gather(output_list, packed_buffer)         # variable-size
+  unpack: gathered buffers → per-param unsharded tensors (offset table lookup)
+
+reshard (reduce-scatter):
+  pack:   split unsharded grads → per-rank contiguous send buffers
+  comm:   reduce_scatter(recv_buf, send_bufs)
+  unpack: recv buffer → sharded grad views
+```
+
+The Shard region avoids this entirely — uniform sizes across ranks allow a single `all_gather_into_tensor` / `reduce_scatter_tensor` directly on the contiguous region. The pack/unpack overhead exists only for RaggedShard and Owned regions.
```
ailzhang: qq: which data structure owns the lifecycle of the contiguous region?

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+    """Base class for FlexShard placement strategies (separate from DTensor's Placement)."""
+
+    def compute_local_numel(self, global_shape, rank, world_size) -> int:
+        """How many elements this rank holds for a param with global_shape."""
+        ...
+
+    # Unshard: local shards → full param
+    def pack_unshard(self, local_tensor, buffer, offset) -> None: ...
+    def comm_unshard(self, send_buf, recv_buf, mesh) -> None: ...
+    def unpack_unshard(self, buffer, offset, global_shape) -> Tensor: ...
```
ailzhang: nit: probably need to rename these

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+
+Parameter is split evenly along `dim` across all ranks. Supports uneven sharding where `dim % world_size != 0` using ceiling-based chunk sizes.
+
+```
+World size = 4, Parameter shape = (100, 256)
+Shard(0) → Each rank gets (25, 256) local shard
+```
+
+```python
+flex_shard(model, mesh)  # default: Shard(0) for all params
```
ailzhang: nit: probably good to be explicit instead of setting Shard(0) as default.

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+```python
+flex_shard(model, mesh, shard_strategy="param_boundary")
+```
+
+### `RaggedShard(dims, local_units)` — Asymmetric Sharding
+
+Parameter is distributed across all ranks with variable chunk sizes. `local_units` specifies relative allocation ratios.
+
+```
+World size = 4, Parameter shape = (100, 256)
```
ailzhang, line 244: nit: can we add a similar dataclass example for RaggedShard? the FlatShard one is very helpful.

ailzhang (commented):
Thanks a ton! Overall this becomes much clear now, I have a few questions around the boundary of these abstractions 🙏

torchtitan/experiments/flex_shard/flex_shard.md
weifengpy: > does symmetric/asymmetric mean even/uneven?

strictly speaking, it's rank 0 having x elements, rank 1 having y elements, x != y. need all-gather-v 

I usually associate even/uneven with padding with the ultimate goal of achieving rank 0...n having same x elements after padding. need all-gather with equal length

>  was curious if we are able to express the layout in general without depending on their shard/ragged/owned property

good suggestion! ideally "Byte Storage" should just call placement.unshard_pack without if-else base on placement

I will try more alone your suggested direction

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+unshard():
+  model.weight = nn.Parameter(unsharded_tensor)  # plain tensor, full shape
+    .shape = (100, 256)
+
+reshard():
+  model.weight = nn.Parameter(local_view)     # back to sharded view
+    ._placements = (Shard(0),)
+```
+
+During forward/backward compute, parameters are plain unsharded tensors — no subclass overhead on the compute path.
```
wconstab: we should probably clarify here: we're gonna support both paths for SPMD-types and DTensor right? so in the DTensor case, the parameters are gonna be DTensors after unshard?

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+
+All placements inherit from `Placement`, which defines the pack/comm/unpack contract for DStorage. Users can register custom placements by subclassing `Placement` and implementing the 7 methods.
+
+```python
+from torchtitan.experiments.flex_shard import Placement
+
+class Placement:
+    """Base class for FlexShard placement strategies (separate from DTensor's Placement)."""
+
+    def compute_local_numel(self, global_shape, rank, world_size) -> int:
```
wconstab: is this API actually going to be flexible enough? let me see if you have usage examples for dcp.  i am wondering if we have strided patterns, or blocked patterns, if a single int return value is enough information

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,280 @@
...
+reshard():
+  model.weight = nn.Parameter(local_view)     # back to sharded view
+    ._placements = (Shard(0),)
+```
+
+During forward/backward compute, parameters are plain unsharded tensors — no subclass overhead on the compute path.
+
+## Memory Layout
+
+Buffer is organized by region for mixed Shard/Owned/RaggedShard placements:
```
wconstab: can you explain why one buffer has multiple types of sharding mixed together? if they are going to need different collectives why not use separate buffers?

weifengpy: good question! I just did a refactor of using one single collective for the buffer

torchtitan/experiments/flex_shard/flex_shard.md
```diff
@@ -0,0 +1,287 @@
+# FlexShard
+
+Implementation: https://github.com/pytorch/pytorch/pull/174267
```
tianyu-l: do we still need this link?

torchtitan/experiments/flex_shard/flex_shard_design.md
```diff
@@ -0,0 +1,299 @@
...
+    shard_placement_fn: PlacementFn | dict[str, Placement] | None = None,
+    *,
+    buckets: list[list[str] | BucketSpec] | None = None,
+)
+```
+
+One call shards a module. The same function works for eager, JIT, and AOT — no mode-specific API.
+
+```python
+mesh = init_device_mesh("cuda", (world_size,))
```
tianyu-l, line 27: Right now it's only covering the case where computation mesh is the same as flex_shard mesh. Next we also need to align API when they are different, e.g. (dp_replicate, fsdp_cp) for flex_shard but (dp, cp) for computation.

torchtitan/experiments/flex_shard/flex_shard_design.md
```diff
@@ -0,0 +1,299 @@
...
+#       total = sum(p.numel() for _, p in named_params)
+#       result, offset = {}, 0
+#       for fqn, p in named_params:
+#           result[fqn] = (FlatShard(offset, p.numel(), total),)
+#           offset += p.numel()
+#       return result
+#
+# Example for a bucket with weight (256x768) + bias (768):
+#   weight → FlatShard(flat_offset=0,      numel=196608, total=197376)
+#   bias   → FlatShard(flat_offset=196608, numel=768,    total=197376)
```
tianyu-l, line 59: It seems `FlatShard` itself doesn't have reference to the start of the buffer. I think it may work with DCP (who has access to a tensor and the placement), but just to confirm.

torchtitan/experiments/flex_shard/flex_shard_design.md
```diff
@@ -0,0 +1,299 @@
...
+  └──────────────────────────────────────────────────────────────┘
+```
+
+Autograd sees the all-gather as a differentiable op. The graph pass marks it `MUST_RECOMPUTE` so the compiler frees `full_W` after forward and re-all-gathers in backward.
+
+## Parametrization: The Shared Front End
+
+### How it works
+
+`_register_parametrization()` creates a dynamic subclass of each leaf module's class with property descriptors. When `module.weight` is accessed, the property getter checks for a batched all-gather result first, falling through to the per-param parametrization if none:
```
tianyu-l, line 158: how about optimizer updates, when we may not want parametrization to be triggered?

torchtitan/experiments/flex_shard/flex_shard_design.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Reshard-After-Forward
+
+Goal: **free unsharded parameter memory after each layer's forward, recompute in backward**.
+
+### Eager: Checkpoint + Batched Collectives
+
+Three mechanisms work together:
+
+**1. Per-layer checkpoint** (`_apply_reshard_checkpoint`): Each layer is wrapped in `checkpoint_wrapper` with `_flex_shard_reshard_policy`. The policy marks collective ops (`all_gather_into_tensor`, `wait_tensor`, `broadcast`) as `MUST_RECOMPUTE` and everything else as `PREFER_RECOMPUTE`. After each layer's forward, checkpoint discards all intermediates — including the detached-leaf unsharded params (which autograd treats as ordinary intermediates since it doesn't know they came from an all-gather). In backward, checkpoint re-runs the layer's forward: the pre_forward_hook fires again, doing another batched all-gather to recreate the unsharded params. This achieves per-layer memory release without any special reshard logic — checkpoint's standard discard-and-recompute mechanism handles it naturally because the all-gather is outside the autograd graph.
```
tianyu-l: Is this for compile-only or both? I think there are two challenges:
1. cannot differentiate `all_gather_into_tensor` in FlexShard vs. `all_gather_into_tensor` in user code -- this can be avoided by registering FlexShard specific ops.
2. Haven't read where you are going to apply this per-layer checkpoint, but current AC doesn't support double / nested wrapping -- what if user would like to wrap AC for their code, would it conflict with FlexShard AC wrapping? This is more of an eager problem.

torchtitan/experiments/flex_shard/flex_shard_design.md
```diff
@@ -0,0 +1,299 @@
...
+
+**2. Batched all-gather** (`_install_batched_allgather_hooks`): Pre-forward hooks on each layer call `Placement.unshard()` — one batched `dist.all_gather` per bucket (instead of N per-param `_c10d_functional.all_gather_into_tensor` calls). The property getter returns the pre-gathered result as a detached leaf.
```
tianyu-l: How is this related to the parametrization mentioned above?

torchtitan/experiments/flex_shard/flex_shard.md
weifengpy: removed the link. I am fully focused on this titan PR now

torchtitan/experiments/flex_shard/flex_shard_design.md
weifengpy: good pointed. added this as TODO in flex_shard_design.md

torchtitan/experiments/flex_shard/flex_shard_design.md
weifengpy: that's correct. added `test_dcp_save_load_flat_shard` test to showcase flex_shard placement + DCP

torchtitan/experiments/flex_shard/flex_shard_design.md
weifengpy: We use model.parameters() to construct the optimizer, which reads _parameters directly, so it won't trigger parametrization

If people do want explicit control, we have a context manager `disable_active_parametrization` to disable parametrization and let people read model.weight without all-gather.

torchtitan/experiments/flex_shard/flex_shard_design.md
weifengpy: > cannot differentiate all_gather_into_tensor in FlexShard vs. all_gather_into_tensor in user code -- this can be avoided by registering FlexShard specific ops.

sounds like a universal problem across flex_shard and simplefsdp? A native solution would be registering FlexShard-specific custom ops (e.g., torch.ops.flex_shard.all_gather_into_tensor)

weifengpy: > Haven't read where you are going to apply this per-layer checkpoint, but current AC doesn't support double / nested wrapping -- what if user would like to wrap AC for their code, would it conflict with FlexShard AC wrapping? This is more of an eager problem.

this is indeed a bigger risk! I added `test_ac_composition.py` to showcase merge policy_fn between flex_shard and original AC, to avoid nesting AC(AC), but have a single AC with merged_policy

```
def merged_policy(
    sctx, func, *args, _orig=original_policy, **kwargs
):
    if func in _FLEX_SHARD_COLLECTIVE_OPS:
        return CheckpointPolicy.MUST_RECOMPUTE
    return _orig(sctx, func, *args, **kwargs)

ctx.policy_fn = merged_policy
```

I will explore more on how prefetch can be decoupled from compute

torchtitan/experiments/flex_shard/flex_shard_design.md
weifengpy: I updated the .md to make this clearer

FlexShard intercepts parameter access so that `module.weight` triggers an all-gather
* *Compiled modes (JIT/AOT)**: falls through to `parametrization.forward()`, which emits `_c10d_functional` ops (per-param, async, FX-traceable). `torch.compile` traces these into the FX graph for compiler passes to optimize
* **Eager mode**: short-circuits via `_pre_gathered` — batched hooks run `dist.*` collectives (per-bucket, synchronous) before the property getter fires, so `parametrization.forward()` never runs. Per-layer checkpoint and Batched all-gather all belong to this

ailzhang:
Great work, thanks a ton! I took a quick look and here're two minor suggestions
- would be nice to clarify where padding is allowed (this can be future work for sure). 
- it's possible that I missed in the code, does our dcp test support resharding? I was curious what data structure looks like to support this feature, given a placement and rank, how its local elements maps to "global tensor" (sharding info). And DCP can define the mapping from local elements to the per rank ckpt file.

weifengpy:
landed in https://github.com/meta-pytorch/flex_shard

Issue #1323, opened 2025-06-20
Why `preserve_rng_state=False` in activation checkpointing


awgu:
Why does torchtitan set `preserve_rng_state=False` for activation checkpointing? E.g.:
https://github.com/pytorch/torchtitan/blob/f4048f8e1b36827156c4dc861c9680333a8542f9/torchtitan/models/llama3/infra/parallelize.py#L238

tianyu-l:
@soulitzer for any context / suggestions

soulitzer:
It is surprising to see `preserve_rng_state=False`. Although you remove the overhead of stashing/restoring RNG state, if any operations use randomness, then you can end up with incorrect gradients.

tianyu-l:
@soulitzer I see.

I believe Llama is not using randomness -- for any model whose model code doesn't use randomness, it should be fine either way? If a model is using randomness then it's recommended to use `preserve_rng_state=True`. Is this the right conclusion?

Then I wonder why we have this option at all -- why don't we only support the `True` case for correctness. Is it because some model doesn't care about incorrect gradients from recomputation, so we can save the overhead?

soulitzer:
> If a model is using randomness then it's recommended to use preserve_rng_state=True. Is this the right conclusion?

Yeah

> I believe Llama is not using randomness

Are we using dropout in sdpa btw? (This reminds me that for SAC I might be able to stop treating it as a random op if dropout_p is set to 0)

> so we can save the overhead?

We might've checked this at some point to see if the overhead was minor enough to just not care about, I don't think that was the case.

awgu:
The extra overhead is mainly some extra CPU overhead?

tianyu-l:
@soulitzer 
Llama is not using dropout in sdpa afaik.

> We might've checked this at some point to see if the overhead was minor enough to just not care about, I don't think that was the case.

Then why don't we always set it to True?

soulitzer:
> Then why don't we always set it to True?

@tianyu-l Sorry I worded that kind of badly lol, what I mean to say is the overhead didn't seem minor enough to ignore :P

> The extra overhead is mainly some extra CPU overhead?

@awgu Not sure, we just looked at the runtime and didn't try to stare too hard at it. Maybe worth checking again 🤔

sippycoder:
[...]

PR #2380, opened 2026-02-13
[AC] Set preserve_rng_state=True as default for activation checkpointing

soulitzer:
Fixes https://github.com/pytorch/torchtitan/issues/1323

Set preserve_rng_state to True for all checkpoint_wrapper calls to ensure deterministic output compared to non-checkpointed passes. This changes the default in the ActivationCheckpoint config and updates the hardcoded values in the Flux model's activation checkpointing.

Measured overhead is 30-50us on 8gpu node. For comparison, AC already has a 350us constant overhead. Overhead scales linearly when inputs are on more distinct devices, but that is rare . This shouldn't matter anyway because AC regions are fairly large (transformer blocks). It is possible the constant overhead will matter more when only several ops are being AC'd.

```python
  """Measure the overhead of preserve_rng_state in torch.utils.checkpoint."""

  import torch
  import torch.nn as nn
  import torch.utils.checkpoint as checkpoint
  import time


  class Block(nn.Module):
      def __init__(self, dim=1024):
          super().__init__()
          self.linear1 = nn.Linear(dim, dim)
          self.linear2 = nn.Linear(dim, dim)

      def forward(self, x):
          return self.linear2(torch.relu(self.linear1(x)))


  def bench(block, x, preserve_rng_state, warmup=10, iters=100):
      for _ in range(warmup):
          out = checkpoint.checkpoint(block, x, preserve_rng_state=preserve_rng_state, use_reentrant=False)
          out.sum().backward()
          torch.cuda.synchronize()

      torch.cuda.synchronize()
      t0 = time.perf_counter()
      for _ in range(iters):
          out = checkpoint.checkpoint(block, x, preserve_rng_state=preserve_rng_state, use_reentrant=False)
          out.sum().backward()
          torch.cuda.synchronize()

      return (time.perf_counter() - t0) / iters * 1000  # ms per iter


  block = Block(dim=1024).cuda()
  x = torch.randn(64, 128, 1024, device="cuda", requires_grad=True)

  ms_false = bench(block, x, preserve_rng_state=False)
  ms_true = bench(block, x, preserve_rng_state=True)

  print(f"preserve_rng_state=False: {ms_false:.3f} ms/iter")
  print(f"preserve_rng_state=True:  {ms_true:.3f} ms/iter")
  print(f"Overhead: {ms_true - ms_false:.3f} ms ({(ms_true / ms_false - 1) * 100:.2f}%)")
```

### DIFF
--- a/torchtitan/models/flux/infra/parallelize.py
+++ b/torchtitan/models/flux/infra/parallelize.py
@@ -119,13 +119,13 @@ def apply_ac(model: nn.Module, ac_config):
 
     # pyrefly: ignore [missing-attribute]
     for layer_id, block in model.double_blocks.named_children():
-        block = ptd_checkpoint_wrapper(block, preserve_rng_state=False)
+        block = ptd_checkpoint_wrapper(block, preserve_rng_state=True)
         # pyrefly: ignore [missing-attribute]
         model.double_blocks.register_module(layer_id, block)
 
     # pyrefly: ignore [missing-attribute]
     for layer_id, block in model.single_blocks.named_children():
-        block = ptd_checkpoint_wrapper(block, preserve_rng_state=False)
+        block = ptd_checkpoint_wrapper(block, preserve_rng_state=True)
         # pyrefly: ignore [missing-attribute]
         model.single_blocks.register_module(layer_id, block)
 
--- a/torchtitan/config/job_config.py
+++ b/torchtitan/config/job_config.py
@@ -696,7 +696,7 @@ class ActivationCheckpoint:
     https://github.com/pytorch/pytorch/pull/126320#discussion_r1625104015
     """
 
-    preserve_rng_state: bool = False
+    preserve_rng_state: bool = True
     """
     If deterministic output compared to non-checkpointed passes is required, set
     to true. Results in stashing and restoring the RNG state during each checkpoint,


torchtitan/models/flux/infra/parallelize.py
```diff
@@ -119,13 +119,13 @@ def apply_ac(model: nn.Module, ac_config):
 
     # pyrefly: ignore [missing-attribute]
     for layer_id, block in model.double_blocks.named_children():
-        block = ptd_checkpoint_wrapper(block, preserve_rng_state=False)
+        block = ptd_checkpoint_wrapper(block, preserve_rng_state=True)
```
tianyu-l, line 122: @wwwjn do you still remember why we didn't apply llm-like per-layer AC wrapping to flux? Over there we support SAC and some other features.

tianyu-l (approved):

torchtitan/models/flux/infra/parallelize.py
wwwjn: I don't remember any blocker prevent us using SAC on flux model, but I don't recall we tested SAC on FLUX. The only reason I could think of is apply_ac() function in [LLM](https://github.com/pytorch/torchtitan/blob/fde830de29c34c55b4cdc0209ac51f5b8084244e/torchtitan/distributed/activation_checkpoint.py#L194) evolved a lot after FLUX is developed. Since FLUX has `single_blocks` and `double_blocks`, FLUX keeps a separate apply_ac function which is not changed yet. We should test SAC + new AC features on FLUX

PR #2381, opened 2026-02-13
Add RL skeleton

daniellepintz:
Add basic RL skeleton code which runs RL on a simple task and learns how to make use of tool calling results and follow formatting instructions. See README for more info on the Zorplex task.

### DIFF
--- a/experiments/reinforcement_learning/generator.py
+++ b/experiments/reinforcement_learning/generator.py
@@ -7,7 +7,6 @@
 """
 Generator actor for RL training.
 
-Generates trajectories for the digit sum task.
 Weight sync uses load_state_dict from the trainer.
 """
 


experiments/reinforcement_learning/generator.py
```diff
@@ -0,0 +1,166 @@
...
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        self.policy_version = 0
+        self.generations = 0
+
+        print(
+            f"[Generator:{self.rank}] Loading model {self.model_name} on GPU {gpu_id}..."
+        )
+
+        self.tokenizer = AutoTokenizer.from_pretrained(self.model_name)
+        self.model = AutoModelForCausalLM.from_pretrained(
```
wwwjn, line 66: This part should be easily switch to torchtitan model (to unblock unified model definition development)

experiments/reinforcement_learning/generator.py
```diff
@@ -0,0 +1,166 @@
...
+
+        full_ids = self.tokenizer(
+            prompt_text + model_only_text,
+            return_tensors="pt",
+            add_special_tokens=False,
+            truncation=True,
+            max_length=1024,
+        )["input_ids"]
+
+        # Reward shaping:
```
wwwjn: Love this simple reward desgin!

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+
+        self.model.train()
+        self.optimizer.zero_grad()
+
+        losses = []
+        valid_count = 0
+
+        for traj in trajectories:
+            if not traj.input_ids or traj.prompt_length == 0:
+                continue
```
tianyu-l: why would we need this code?

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+
+            # Load pre-tokenized sequence from the generator
+            full_ids = torch.tensor(traj.input_ids, device=self.device).unsqueeze(0)
```
tianyu-l, line 145: why do we need to unsqueeze?

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+            prompt_len = traj.prompt_length
+
+            if full_ids.shape[1] <= prompt_len + 1:
+                continue
```
tianyu-l: why would we need this code?

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+
+            # REINFORCE loss = -log_prob * advantage
+            advantage = traj.reward - baseline
+            losses.append(-token_log_probs.sum() * advantage)
+            valid_count += 1
+
+        # Optimizer step
+        if valid_count > 0:
+            avg_loss = torch.stack(losses).sum() / valid_count
```
tianyu-l: IIUC `valid_count` is the amount of valid trajectories, and we take unweighted (each valid trajectory contributes the same * its own advantage) average loss over it.

When would a trajectory be invalid -- this is related to my two comments before.

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+            # logits[i] predicts token[i+1], so start at prompt_len - 1
+            shift_logits = logits[:, prompt_len - 1 : -1, :]
+            shift_labels = full_ids[:, prompt_len:]
+            log_probs = F.log_softmax(shift_logits, dim=-1)
+            token_log_probs = log_probs.gather(2, shift_labels.unsqueeze(-1)).squeeze(
+                -1
+            )
+
+            # REINFORCE loss = -log_prob * advantage
+            advantage = traj.reward - baseline
```
tianyu-l: shall we rename `baseline` to `advantage_baseline`?

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+            "success": 0,
+            "wrong_format": 0,
+            "tool_spam": 0,
+            "wrong_answer": 0,
+        }
+        for _ in range(num_samples):
+            task = spec.generate_task()
+            result = generate_with_tools(
+                self.model,
+                self.tokenizer,
```
tianyu-l: This seems trainer tokenizer, set up only for tool calling. What if this tokenizer is different from generator tokenizer?

experiments/reinforcement_learning/zorplex.py
```diff
@@ -0,0 +1,321 @@
...
+                **inputs,
+                max_new_tokens=max_tokens_per_turn,
+                pad_token_id=tokenizer.eos_token_id,
+                eos_token_id=tokenizer.eos_token_id,
+                stopping_criteria=stop_criteria,
+                do_sample=do_sample,
+            )
+            if do_sample:
+                gen_kwargs["temperature"] = temperature
+            outputs = model.generate(**gen_kwargs)
```
tianyu-l: This seems relying on HF transformers protocol?

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+"""
+
+from dataclasses import dataclass, field
+
+import torch
+import torch.nn.functional as F
+from monarch.actor import Actor, current_rank, endpoint
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from zorplex import generate_with_tools, ZorplexSpec
```
tianyu-l:
```suggestion
from .zorplex import generate_with_tools, ZorplexSpec
```

experiments/reinforcement_learning/README.md
```diff
@@ -0,0 +1,73 @@
...
+### Options
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--num-steps` | 20 | Number of training steps |
+| `--num-generators` | 2 | Number of generator workers |
+| `--eval-samples` | 16 | Number of evaluation samples |
+| `--verbose` | off | Print sample generations on each step |
+
+## Zorplex Task
```
tianyu-l: Can I get a reference link to Zorplex task introduction in general? I tried search but nothing is available.

experiments/reinforcement_learning/main.py
```diff
@@ -0,0 +1,262 @@
...
+    Returns a dict with keys: trainer, generators, _proc_meshes.
+    """
+    from generator import Generator
+    from monarch.actor import this_host
+    from trainer import Trainer
+
+    host = this_host()
+
+    # 1. Generators -- plain ActorMesh
+    gen_procs = host.spawn_procs(per_host={"procs": num_generators})
```
tianyu-l, line 41: IIUC multiple `Generator`s are created independently (with different random seed), to get different zorplex tasks. In particular, there is no parallelism.

tianyu-l (commented):
Thanks, this version is very readable. Left some comments.

experiments/reinforcement_learning/trainer.py
daniellepintz: it's just a check to make sure, but will remove as these shouldn't be true! thanks

experiments/reinforcement_learning/trainer.py
daniellepintz: to add batch dimension. adding a comment here.

the model's forward expects input like [batch_size, seq_len], so we need to add batch dimension which will be [1, seq_len]

experiments/reinforcement_learning/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+
+            # Load pre-tokenized sequence from the generator
+            full_ids = torch.tensor(traj.input_ids, device=self.device).unsqueeze(0)
+            prompt_len = traj.prompt_length
+
+            if full_ids.shape[1] <= prompt_len + 1:
+                continue
+
+            # Forward pass, then slice at prompt_length for response-only log-probs
+            with torch.amp.autocast("cuda", enabled=self.device == "cuda"):
```
joecummings: The way this is currently written, there would be no need to autocast.

experiments/reinforcement_learning/README.md
```diff
@@ -0,0 +1,73 @@
+# Reinforcement Learning
+
+## Setup
+
+```bash
+conda create -n torchtitan-rl python=3.12 -y
```
allenwang28: prefer uv?

experiments/reinforcement_learning/README.md
allenwang28: Zorplex is a fully made up task! But the premise is related to the thesis listed in the abstract here: https://arxiv.org/pdf/2509.24210

experiments/reinforcement_learning/main.py
allenwang28: > there is no parallelism

Clarifying my understanding, you mean that there isn't any form of parallelism like TP/PP/EP etc.? If so, that's correct.

We'll want to move towards representing each generator as an independent actor using vLLM

experiments/reinforcement_learning/zorplex.py
allenwang28: yeah this example was pretty hardcoded for the sake of the GPU mode lecture, but we should push towards zorplex not being tied to HuggingFace

I.e. we should replace this with vLLM using the unified model implementation. I think vLLM may express the stopping criteria etc. differently

experiments/reinforcement_learning/trainer.py
daniellepintz: good call

experiments/reinforcement_learning/trainer.py
daniellepintz: it's just a check in case there are no response tokens, but I agree this is unnecessary and will remove!

experiments/reinforcement_learning/trainer.py
daniellepintz: yes, since i removed the early returns from your two comments before, i can remove this too, thanks!!

experiments/reinforcement_learning/main.py
daniellepintz: > IIUC multiple Generators are created independently (with different random seed), to get different zorplex tasks.

yes exactly

> We'll want to move towards representing each generator as an independent actor using vLLM

yes, can do this as a follow up

experiments/reinforcement_learning/zorplex.py
daniellepintz: yes agreed we should move to vLLM and unified model impl., can this be done as a follow up?

wwwjn: As we discussed offline - Let's evaluate how difficult the following paths are:
1. integrating zorplex into current simple_rl_multiprocessing.py
2. Adding vllm as rollout engine for this skeleton

I think we should avoid landing a temporary design and refactor later, as we are not in a hurry to unblock ppl. Refactoring into vllm rollout is burdensome, and might ends up redesign the skeleton again. Eg, you might need to use https://github.com/pytorch/torchtitan/blob/7a6e45112a20e3547c91dd25a2df16df5469594e/torchtitan/experiments/rl/unified/actors/generator.py#L61C7-L61C24 which needs to leave in generator, not zorplex.py

experiments/reinforcement_learning/main.py
```diff
@@ -0,0 +1,262 @@
...
+                print()
+
+        gen_time = time.perf_counter() - gen_start
+        total_generations += num_generators
+
+        # Train directly on the batch (no buffer in sync mode)
+        train_start = time.perf_counter()
+        if batch:
+            metrics = trainer.train_step.call_one(batch, advantage_baseline).get()
+            advantage_baseline = 0.9 * advantage_baseline + 0.1 * metrics.avg_reward
```
wwwjn, line 195: Why we update the advantage outside of trainer? And in terms of mathmatical formula, is this standard REINFORCEMENT?

daniellepintz:
Closing in favor of https://github.com/pytorch/torchtitan/pull/2464

Issue #560, opened 2024-08-23
Adjust MFU to account for FP8 


lessw2020:
From internal discussions, logging an issue around updating our MFU calculations so that if FP8 is used, we can generate an accurate MFU number. 

Atm - FP8 replaces wq/wk/wv/wo in Attention, and w1/w2/w3 in the MLP. 

Thus, need an adjusted calculation.

In addition, would like to correctly pull the proper MFU (fp8 or bf16) based on the training config being run so this is handled automatically for the user.

raghukiran1224:
yes, this needs to be updated. The MFU computations for `fp8` are too good to be true :) 
CC: @lchu-ibm

tianyu-l:
Hmm I don't see a proper way to calculate MFU when some computations (linear) use FP8 and some others use BF16 (everything else, e.g. sdpa). What should be the right way?

msaroufim:
I chatted with @stas00 briefly offline and he suggested 2 approaches
1. Assume peak is all computations are done in FP8, as this is no different than how MFU computations are done for BF16
2. Assume peak is a weighted average of BF16 and FP8

1 feels simpler with the main con being that numbers we share might be "low" compared to other vendors if they use 2

carmocca:
My personal thoughts:

I was surprised when I saw that torchtitan uses the simple and overoptimistic "academic" flops formula (https://github.com/pytorch/torchtitan/blob/main/torchtitan/utils.py#L231) considering that `torch.utils.flop_counter.FlopCounterMode` already exists (and in my experience, works quite well).

I feel like PyTorch could offer a very similar utility that measures peak flops per operator. Imagine that you prepare your model first (parallelize, fp8, etc) and then measure the peak flops. From my understanding, such utility could understand the differences between the theoretical performance of the operators based on runtime information like their dtypes.

But if you want to go the easy route, option 1 seems the better choice to me. For the big models most flops will go to the linears anyways and forkers can always customize the calculation.

But please do not perpetuate the abuse of MFU as a marketing number 🙏. You can report tokens per second (if not padding) or sequences per second. MFU measuring contests are pointless unless the flops calculation is proved to be exactly the same

tianyu-l:
thanks @carmocca
1. To me there is no right way of doing mixed MFU with BF16 and FP8, the approaches @msaroufim mentioned sound "less wrong" but not correct.
[...]

PR #2382, opened 2026-02-14
Add peak BF16 TFLOPS for AWS Trainium/Inferentia devices

rthekini-aws:
Add TensorEngine-only BF16 peak TFLOPS for Neuron devices to enable accurate MFU calculation on AWS Trainium and Inferentia instances.

- NeuronCore-v2 (trn1, trn1n, inf2): 90 TFLOPS/core, LNC=1
- NeuronCore-v3 (trn2, trn2n, trn2u): 79 TFLOPS/core, LNC=2
- NeuronCore-v4 (trn3, trn3u): 79 TFLOPS/core, LNC=2

### DIFF
--- a/torchtitan/tools/utils.py
+++ b/torchtitan/tools/utils.py
@@ -73,7 +73,8 @@ class GarbageCollection:
         logger.info("[GC] %s took %.2f seconds", reason, time.monotonic() - begin)
 
 
-# hardcoded BF16 type peak flops for NVIDIA A100, H20, H100, H200, B200 GPU and AMD MI250, MI300X, MI325X, MI355X and Intel PVC
+# hardcoded BF16 type peak flops for NVIDIA A100, H20, H100, H200, B200 GPU,
+# AMD MI250, MI300X, MI325X, MI355X, Intel PVC, and AWS Trainium/Inferentia
 def get_peak_flops(device_name: str) -> float:
     try:
         # Run the lspci command and capture the output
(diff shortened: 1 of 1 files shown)


tianyu-l (approved):

wwwjn:
Can you fix lint error before merging it? @rthekini-aws

rthekini-aws:
The CLA should be signed. I'll follow up with `cla@meta.com`.

torchtitan/tools/utils.py
```diff
@@ -138,6 +139,27 @@ def get_peak_flops(device_name: str) -> float:
...
+            # https://awsdocs-neuron.readthedocs-hosted.com/en/latest/about-neuron/arch/neuron-hardware/neuron-core-v2.html
+            return 90e12 * 1
+        elif neuron_device_name in ("trn2", "trn2n", "trn2u"):
+            # NeuronCore-v3 TensorEngine: 79 BF16 TFLOPS/core, LNC=2
+            # https://awsdocs-neuron.readthedocs-hosted.com/en/latest/about-neuron/arch/neuron-hardware/neuron-core-v3.html
+            return 79e12 * 2
+        elif neuron_device_name in ("trn3", "trn3u"):
+            # NeuronCore-v4 TensorEngine: 79 BF16 TFLOPS/core, LNC=2
+            # https://awsdocs-neuron.readthedocs-hosted.com/en/latest/about-neuron/arch/neuron-hardware/neuron-core-v4.html
+            return 79e12 * 2
```
tianyu-l: nit: why having two entries for the same thing?

rthekini-aws: They are different hardware generations (with different links), so I've kept them on different lines. It's also possible that things change (e.g., LNC) or we may want to consider MXFP8 FLOPs (which do differ).

If you feel this justification isn't sufficient, let me know, and I can collapse this into one entry.

tianyu-l: Sorry what is LNC?

Also, there is no way to use non-bf16 peak flops today, see https://github.com/pytorch/torchtitan/issues/560

Even if things would change, the entry here is hard-coded so anyway you need to come back and update. Let's collapse (you can keep both comments). Thanks!

rthekini-aws: Sounds good; I'll make the change. There are also some internal refactors I've been asked to make, so I'll update the branch when both are ready.

For details on LNC see: https://awsdocs-neuron.readthedocs-hosted.com/en/latest/nki/get-started/about/lnc.html. In practice, LNC=2 means one torch device (or process) corresponds to 2 Neuron Cores (i.e., 1/4 of a chip for Trn2/Trn3).

PR #2385, opened 2026-02-16
Add possibility of creating a dataset from multiple sources

slimfrkha:
# Problem:
Current code doesnt give the possibility to combine multiple data sources. As a consequence user needs to create a big dataset in one file in order to launch training.

# Solution:
Give possibility to specify multiple data sources and correspondant weights.

# Key Changes
Needed to change the design of data section in job config to cater for multi source data specs.
Design of data dataclass can be discussed further. I am not 100% satisfied with the current design.

# Tests
Added a unit test for multi source dataset beside the legacy dataset test.

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -1,5 +1,5 @@
 # Custom Datasets in torchtitan
-
+TODO update this file
 `torchtitan` is designed to work seamlessly with most HuggingFace datasets. While we provide the C4 dataset for numerics and convergence testing, you can easily add support for your own datasets. Here's how to do it using Wikipedia as an example.
 
 ## Quick Start
@@ -57,7 +57,8 @@ DATASETS = {
 In your training configuration file (`.toml`), set your dataset:
 
 ```toml
-dataset = "wikipedia"
+[training.data]
+name = "wikipedia"
 ```
 
 That's it! Your custom dataset is now ready to use with `torchtitan`.
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -1,5 +1,4 @@
 # Custom Datasets in torchtitan
-TODO update this file
 `torchtitan` is designed to work seamlessly with most HuggingFace datasets. While we provide the C4 dataset for numerics and convergence testing, you can easily add support for your own datasets. Here's how to do it using Wikipedia as an example.
 
 ## Quick Start
@@ -73,3 +72,22 @@ That's it! Your custom dataset is now ready to use with `torchtitan`.
 - Use `streaming=True` for large datasets to manage memory efficiently
 
 Now you can start training with your custom dataset!
+
+## Going further
+You can create a dataset from multiple sources. The logic is the same as above with minor changes:
+1. `MultiDatasetConfig` instead of `DatasetConfig` introducing:
+   1. `paths`: list of datasets
+   1. `weights`: list of sampling weights
+2. config file
+    ```toml
+    [training.data]
+    name = "wikipedia"
+    paths = [
+        "source 1 path",
+        "source 2 path",
+    ]
+    weights = [
+        w1,
+        w2
+    ]
+    ```
--- a/docs/evaluation.md
+++ b/docs/evaluation.md
@@ -10,9 +10,10 @@ Below is an example validation config:
 ```toml
 [validation]
 enable = true
-dataset = "c4_validation"
 freq = 500
 steps = -1 # consumes the entire validation set
+[validation.data]
+name = "c4_validation"
 ```
 
 ## Third-Party Evaluation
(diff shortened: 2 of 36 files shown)


rakkit:
i remember there were some "problem" of using HF's interleave dataset.

The token buffer of this PR (pack of interleaved-datasets) will pack sequence from differnt data source that the break the expected sample ratio / weights. 

we should expect more or less like interleaved of packed datasets to preserve sample ratio/weights.

FYI our [implementation](https://github.com/rakkit/torchtitan/blob/llm-0.3.1/torchtitan/hf_datasets/text_datasets.py)

slimfrkha:
humm, the implementation supposes shuffled, packed datasets.
Maybe we can add a pre processing script under `scripts/data/shuffle_pack.py` that does that ? as long the context length of the training is dividable by context len of pre processing script, weights ratio should hold.

rakkit:
>  the implementation supposes shuffled, packed datasets.

what is the context of "packing" here? do you mean that pack each dataset to have a $SEQ-LEN then send packed dataset(s) to wrapped by hf's interleaved dataset?    so logic gonna be pack -> interleaved dataset -> pack?.  If so thats exactly what i mean  "interleaved of packed datasets" and we don't need second pack.

(shuffle anyhow is not mandotory, one can pre-shuffle their dataset of us HF's shuffle(for json) / window shuffle (for parquet).

we can for sure write a scripts to do "pre" & "offline"-pack. 
Pros: actual dataloader's state is super light. 
Cons: given 100B token ~ 200GB storage, do we really want user to run a scripts to do this?  and this way does not work for streaming dataset.

Or we can do "online" packing, more like our version (we have used it to train over 50 datasets mix for over 5T tokens). Pros: user dont have to anything, works for streaming and larger scale dataset.  
Cons: it is important to aware this way leads to each dataloader need to maintain approx $NUM_DATASET * $SEQ-LEN buffer.     (so state sync between dataloader snapshots and dataloder gonna be slow, thats why we need to infer dataloder steps between checkpoint steps to reduce such sync [here](https://github.com/rakkit/torchtitan/blob/llm-0.3.1/torchtitan/hf_datasets/text_datasets.py#716). However this still gonna slowdown training every time we  save checkpoint

tianyu-l (commented):
cc @divyanshk for feedback on multi-source data loading.

Also I have a big refactor pending, which I prefer to land first so that new dataloader work could rebase on it.

slimfrkha:
> > the implementation supposes shuffled, packed datasets.
> 
> what is the context of "packing" here? do you mean that pack each dataset to have a $SEQ-LEN then send packed dataset(s) to wrapped by hf's interleaved dataset? so logic gonna be pack -> interleaved dataset -> pack?. If so thats exactly what i mean "interleaved of packed datasets" and we don't need second pack.
> 
> (shuffle anyhow is not mandotory, one can pre-shuffle their dataset of us HF's shuffle(for json) / window shuffle (for parquet).
> 
> we can for sure write a scripts to do "pre" & "offline"-pack. Pros: actual dataloader's state is super light. Cons: given 100B token ~ 200GB storage, do we really want user to run a scripts to do this? and this way does not work for streaming dataset.
> 
> Or we can do "online" packing, more like our version (we have used it to train over 50 datasets mix for over 5T tokens). Pros: user dont have to anything, works for streaming and larger scale dataset. Cons: it is important to aware this way leads to each dataloader need to maintain approx $NUM_DATASET * $SEQ-LEN buffer. (so state sync between dataloader snapshots and dataloder gonna be slow, thats why we need to infer dataloder steps between checkpoint steps to reduce such sync [here](https://github.com/rakkit/torchtitan/blob/llm-0.3.1/torchtitan/hf_datasets/text_datasets.py#716). However this still gonna slowdown training every time we save checkpoint

The philosophy of the current implementation is to have something simple and light that user can extend. my personal opinion is that the alternative solution has 2 caveats:
1. too complex and long
2. slower to iterate over in training

if we go with a preprocessing script:
1. we pack datasets sample. in training we can either choose to pack or not. why ? maybe you dont want to pre-preprocess dataset for each new context len value (example: pre processing packing at 2k, training packing at 4k). In the current code, weights hold ratio if all samples from all sources use a fixed len (independenly if it is = training context len or not). Simplest case scenario is still packing at preprocessing + no packing in training (same context len value)
2. shuffle is needed anyways. so the user preprocess datasets each time.
3. shuffled packed datasets wouldnt be saved in tokens but rather decoded back to text for more efficient storage usage

To be clear this is what i have in mind for `scripts/data/shuffle_pack.py`: loop over data sources (or in parallel with num workers). for each data source:
1. read the whole dataset in memory (if it fits), shuffle. 
2. loop over shuffled, tokenize, pack, decode back to text
3. save shuffled packed text dataset on disk

Issue #1415, opened 2025-07-18
[Feature request] Use omegaconf or hydra for the config system


yzhao30:
Is there a plan to use Omegaconf or Hydra for the configuration system?

The current .toml-based configuration system is simple but verbose: it does not support configuration inheritance or composition, which prevents config reuse. 

If this is needed, I am interested in contributing an alternative configuration solution based on Omegaconf.

tianyu-l:
Hi, could you explain

What the E2E use cases look like
> support configuration inheritance or composition, which prevents config reuse.

How to define new configs in Omegaconf / Hydra

What is the relationship between Omegaconf / Hydra

yzhao30:
1. E2E use case: 
Say we have a base_train.toml, and there are downstream tasks or experiments, like `train_tp.toml` or `train_smaller_lr.toml`. The changes are only a few settings, such as:
```toml
[parallelism]
tensor_parallel_degree = 8
```
or 
```toml
[optimizer]
lr = 1e-5
```
With TOML, we have to copy all the contents and just modify a few lines. There is redundancy and, more importantly, we lose the relationship between these experiments. This is not ideal for experiment tracking.

OmegaConfi uses yaml format. Define a config is very similar to toml. It supports config merge and override. The Hydra built on it support inheritance like:
```yaml
defaults:
  - base_train

optimizer:
  lr: 0.0001
```
2. OmegaConf is a config engine. Hydra builds on OmegaConf. It is a framework that supports more complex management and powerful features like runtime overrides. Roughly speaking, compared to the current TorchTitan solution: OmegaConf serves a similar role to a TOML parser, while Hydra functions as a configuration manager, but more flexible and powerful. 

I think adopting OmegaConf is lightweight and can be easily integrated with the current config manager as a secondary solution. Hydra is a relatively heavy lift.

soumith:
i dont think we're looking to incorporate Omegaconf or Hydra. Those are too heavy-hitting and too opinionated on the config system for a fork-friendly project like TorchTitan; and honestly they're not to my taste at all.
The feeling I get when using Hydra is that I'm effectively programming in another language that I need to learn, keep track of side-effects and the codebase indeed becomes like a Hydra.

PR #2386, opened 2026-02-17
[BC Breaking] Config System Refactor: TOML to Python Dataclass Registry

tianyu-l:
**NOTE**: This PR is a large refactor of the codebase. https://github.com/pytorch/torchtitan/releases/tag/v0.2.2 contains a latest release right before this PR is merged.

# author's note

This refactor is mainly trying to address two issues:
- bad encapsulation: previously a monolithic `JobConfig` is leaked everywhere
- not easy to iterate and experiment on model architecture and training components

The main changes are:
- Strict encapsulation, even at the cost of (hopefully temporary) bloated interface when calling subcomponents (e.g. validator). We should try to find the right abstraction on cross-components visibility.
- Each `Configurable` component owns its own `Config`, which builds the owner component. It achieves modularization via polymorphism and inheritance, both classic concepts in OOP.
  - This is partly inspired by repos like [AXLearn](https://github.com/apple/axlearn) (in particular, @ruomingp's [ML API Styles](https://github.com/apple/axlearn/blob/main/docs/ml_api_style.md)), github issues (e.g. https://github.com/pytorch/torchtitan/issues/1055), and offline discussions (with @Chillee, @ailzhang etc.).
  - Similar functionality can be alternatively achieved by other ways, e.g. `_target_` in [Hydra](https://hydra.cc/docs/advanced/instantiate_objects/overview/), but there are opinions not to couple with Hydra's other offerings. See https://github.com/pytorch/torchtitan/issues/1415
- Main entry point switches from TOML files to Python functions (a.k.a. `config_registry.py` in each model).
  - TOML has the constraint that everything needs to be registered explicitly before it can be used, e.g. our quantization components need to be registered with string names. Python's language level implicit registration is what we believe to be more minimal, and should be fairly easy to extended/modified to support TOML/YAML when users builds upon / fork torchtitan.
  - That said, Python config provides much more power, e.g. one can use arbitrary logic to create (the config of) a component, which is hard to express with TOML/YAML, thus creating extra difficulty when users want to migrate to their own favorite config system. The only thing we can do is to stay conservative on the usage of such power.
- We still uses [tyro](https://github.com/brentyi/tyro) to convert config dataclass to CLI, still with the limitation that users need to construct customized config classes, all the way from root level (`Trainer.Config` now, `JobConfig` in the past).
  - If CLI is not needed, new trainer (or any high-level) config is not required.
  - To support "polymorphic construction" from CLI without the hassle, check out [chz](https://github.com/openai/chz/blob/main/docs/04_command_line.md#polymorphic-construction).

This PR also
- updates the docs -- there might be remaining outdated docs, please raise issues or help fix
- moves ft to experiments, continuing the effort in https://github.com/pytorch/torchtitan/pull/2311

Remaining work
- [AutoParallel CI failure](https://github.com/pytorch/torchtitan/actions/runs/22165425254/job/64091572780?pr=2386) seems caused by the way RoPE is authored, and needs change in autoparallel. (cc @xmfan)
  - being fixed in https://github.com/meta-pytorch/autoparallel/pull/321
- [CompilerToolkit CI failure](https://github.com/pytorch/torchtitan/actions/runs/22168015737/job/64099486707?pr=2386) `TypeError: forward() missing 1 required positional argument: 'fwd_rng_state_2'` cc @yiming0416 please help take a look
- [SimpleFSDP CI failure](https://github.com/pytorch/torchtitan/actions/runs/22168015749/job/64099486149?pr=2386) is the same as https://github.com/pytorch/torchtitan/issues/2312 around dynamic shape for for-loop MoE experts computation. (cc @pianpwk)
  - being fixed in https://github.com/pytorch/torchtitan/pull/2399
- Fix integration scripts for MAST, Zoomer, etc.
- organize docs from `docs/` to subfolders, as we are having more contents to cover in general
- generate and store serialized configs (maybe not in the repo)
- continue SAC refactor in https://github.com/pytorch/torchtitan/pull/2357, but somehow keep the every-other-mm policy (cc @mori360)
- refactor RoPE in general, at least resolving the following TODOs in code (cc @shuhuayu)
  - having to set / no validation on rope dim == decoder dim // attention n_heads
  - consolidate `apply_rotary_emb_complex` and `apply_rotary_emb_single_complex`
  - address https://github.com/pytorch/torchtitan/issues/2417

Longer-term issues
- More careful design about what to put config vs. runtime build kwargs. (thanks @ailzhang)
- ModelSpec not serializable. There may be multiple solutions, but we can potentially consolidate `model.py` and `parallelize.py` by
  - sharing AC, compile, DP application across all Decoder models
  - putting per-module TP/CP/EP sharding plan inside model itself
- Right now `BaseModel.update_from_config` violates encapsulation by passing the Trainer config into Model config. This could be avoided by python logic either in config construction time, or in trainer.
- Refactor `init_weights` into `Module.Config` instead of staying in `Module`
  - The benefit is that param init can be configurable; o/w we are coupling module implementation and its weight init.
  - This may require refactor of current TransformerBlock and its config. E.g. `weight_init_std` may need to be put in config, with `__post_init__` determining its value. (See related complaints / discussions on `__post_init__` by [chz](https://github.com/openai/chz/blob/main/docs/21_post_init.md))

Note to reviewer:
Although I believe the changes in this PR come naturally in a bundle, you may (or may not) find the stack of 16 commits easier to review, as I tried to split the changes in some logic manner. I apologize for the giant PR.

# claude-generated summary

## Summary

This PR refactors torchtitan's configuration and training infrastructure in 15 incremental, backwards-incompatible commits. The central change replaces TOML config files and a monolithic `JobConfig` parser with **typed Python dataclass configs**, a **`Configurable` base class** pattern, and a **`config_registry`** module per model.

**270 files changed, 10,025 insertions, 11,418 deletions.**

---

## Motivation

The previous system used TOML files parsed by a custom `ConfigManager` that layered CLI overrides on top. While simple, this had several friction points:

1. **No type safety at config boundaries.** TOML values are strings/ints/floats parsed at runtime. A typo in a key name (e.g., `training.stpes`) silently becomes a default value.
4. **Flat namespace.** All config sections (`[model]`, `[training]`, `[optimizer]`, `[checkpoint]`, ...) lived in a single `JobConfig` class. Every component received the full `JobConfig` even when it only needed a few fields.
5. **Experiment extension was ad-hoc.** Experiments that needed custom config fields (e.g., SimpleFSDP's `compile.graph_passes` or FaultTolerant's `fault_tolerance.*`) required a `custom_config_module` TOML key and a runtime `_merge_configs` call to graft new fields onto `JobConfig`.
6. **Model args were disconnected from model code.** A `ModelArgs` dataclass in `args.py` defined hyperparameters, but the `TrainSpec` that bundled model + parallelization + loss was registered separately, with no type-level link between them.

---

### 1. `Configurable` Base Class

A new `Configurable` base class (`torchtitan/config/configurable.py`) establishes a universal pattern:

```python
class Configurable:
    class Config:
        def build(self, **kwargs):
            return self._owner(config=self, **kwargs)

    def __init_subclass__(cls, **kwargs):
        # Auto-wires Config.build() -> cls(config=..., **kwargs)
        # Enforces @dataclass(kw_only=True, slots=True) on every Config
```

Every configurable component (Trainer, model, optimizer, tokenizer, dataloader, checkpoint manager, metrics, validators, quantization converters, ...) follows this pattern. Calling `config.build()` constructs the owning class.

### 2. `Trainer.Config` Replaces `JobConfig`

The monolithic `JobConfig` is replaced by `Trainer.Config`, a nested dataclass that aggregates typed sub-configs:

```python
class Trainer(Stateful, Configurable):
        model_spec: ModelSpec | None = None    # set by config_registry, suppressed from CLI
        job: JobConfig = ...
        training: TrainingConfig = ...
        parallelism: ParallelismConfig = ...
        optimizer: OptimizersContainer.Config = ...
        lr_scheduler: LRSchedulersContainer.Config = ...
        checkpoint: CheckpointManager.Config = ...
        dataloader: BaseDataLoader.Config = ...
        metrics: MetricsProcessor.Config = ...
        # ... etc.
```

Each sub-config is the `Config` class of the component that consumes it (e.g., `CheckpointManager.Config` is defined inside `CheckpointManager`). Components receive only their own config, not the entire training config.

### 3. `config_registry.py` Replaces TOML Files

Each model defines a `config_registry.py` with functions that return complete `Trainer.Config` instances:

```python
# torchtitan/models/llama3/config_registry.py

def llama3_debugmodel() -> Trainer.Config:
    return Trainer.Config(
        job=JobConfig(description="Llama 3 debug training", ...),
        model_spec=model_registry("debugmodel"),
        optimizer=OptimizersContainer.Config(lr=8e-4),
        training=TrainingConfig(local_batch_size=8, seq_len=2048, steps=10),
        dataloader=HuggingFaceTextDataLoader.Config(dataset="c4_test"),
        # ...
    )

def llama3_debugmodel_float8() -> Trainer.Config:
    config = llama3_debugmodel()
    config.model_converters = ModelConvertersContainer.Config(
        converters=[Float8LinearConverter.Config(enable_fsdp_float8_all_gather=True)]
    )
    return config
```

### 4. `TrainSpec` -> `ModelSpec`

`TrainSpec` is renamed to `ModelSpec` with a narrower scope: it holds only model-specific concerns (model config, parallelization function, loss function, state dict adapter). All training-level concerns (optimizer, LR scheduler, checkpointing, etc.) live in `Trainer.Config`.

### 5. Model Configs: Flat `ModelArgs` -> Nested Dataclass Hierarchy

Model hyperparameters move from a flat `ModelArgs` dataclass into a nested `Config` hierarchy that mirrors the module tree:

```python
# Before (main): flat args.py
@dataclass
class ModelArgs:
    dim: int = 4096
    n_layers: int = 32
    n_heads: int = 32
    # ... 20+ flat fields

# After (this PR): nested Config in model class
class Llama3Model(Decoder):
    class Config(Decoder.Config):
        layer: Llama3TransformerBlock.Config  # contains attention + FFN configs
        rope: RoPE.Config                    # contains RoPE-specific params
```

### 6. `train.py` Split

The monolithic `train.py` (~800 lines) is split into:
- `train.py` (~60 lines): thin entry point that calls `ConfigManager.parse_args()` and `config.build()`
- `trainer.py` (~850 lines): the `Trainer` class with training loop logic

### 7. Experiment Extension via Inheritance

Experiments extend the config system through dataclass subclassing instead of runtime config merging:

```python
# torchtitan/experiments/simple_fsdp/configs.py
class SimpleFSDPConfig(Trainer.Config):
    compile: SimpleFSDPCompileConfig = field(default_factory=SimpleFSDPCompileConfig)
```

Their `config_registry.py` returns the subclassed config type, and `tyro` auto-generates CLI parsing for the extended fields.

---

### Launching Training

```bash
# Before (main)
CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.py" ./run_train.sh

# After (this PR)
MODEL=llama3 CONFIG=llama3_8b ./run_train.sh
```

### CLI Overrides

```bash
# Before (main)
CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml" ./run_train.sh \
  --training.steps 100 --parallelism.tensor_parallel_degree 2

# After (this PR)
./run_train.sh --training.steps 100 --parallelism.tensor_parallel_degree 2
# (defaults to MODEL=llama3, CONFIG=llama3_debugmodel via run_train.sh)
```

CLI override syntax is unchanged (`--section.field value`), but `tyro` now provides typed `--help` output generated from the dataclass tree.

### Defining a New Model Config

```bash
# Before: create a new TOML file, copy-paste sections, edit values
cp train_configs/debug_model.toml train_configs/my_experiment.toml
vim train_configs/my_experiment.toml

# After: write a Python function that mutates an existing config
def my_experiment() -> Trainer.Config:
    config = llama3_debugmodel()
    config.training.steps = 100
    config.optimizer.lr = 1e-4
    return config
```

### Adding Experiment-Specific Config Fields

```python
# Before (main): custom_config_module in TOML + runtime _merge_configs
# Requires: TOML key pointing to a Python module, dynamic dataclass creation

# After (this PR): dataclass inheritance
class MyExperimentConfig(Trainer.Config):
    my_custom_field: str = "default"
```

### Float8 / Quantization Configuration

```python
# Before (main): TOML section
# [quantize.linear.float8]
# enable_fsdp_float8_all_gather = true
# precompute_float8_dynamic_scale_for_fsdp = true

# After (this PR): typed config object
model_converters=ModelConvertersContainer.Config(
    converters=[
        Float8LinearConverter.Config(
            enable_fsdp_float8_all_gather=True,
            precompute_float8_dynamic_scale_for_fsdp=True,
        ),
    ],
),
```

---

### 1. Configs are no longer declarative text files

TOML files were readable by anyone without Python knowledge. The new config_registry functions are Python code, which requires understanding imports, function calls, and dataclass construction. For users who only need to tweak hyperparameters, the CLI override syntax (`--training.steps 100`) works the same, but understanding the full config requires reading Python.

### 2. Steeper learning curve for contributors

Adding a new model now requires understanding the `Configurable` protocol, nested `Config` dataclass hierarchy, and the `config_registry` pattern. The old approach of copying a TOML file and editing values had a lower barrier to entry.

### 3. Config serialization is more complex

TOML files were trivially serializable and diffable. The new system supports `to_dict()` + JSON serialization, but configs containing callables (e.g., `ModelSpec.parallelize_fn`) cannot be fully round-tripped. The `model_spec` field is excluded from serialization and suppressed from CLI parsing.

### 4. tyro dependency

The CLI parsing now depends on `tyro`, a third-party library. While `tyro` is well-maintained and provides typed CLI generation from dataclasses, it is an additional dependency that must be kept compatible with the dataclass patterns used here.

### 5. `@dataclass(slots=True)` constraints

The `Configurable` base class enforces `@dataclass(kw_only=True, slots=True)` on all Config classes. While this provides memory efficiency and prevents accidental attribute assignment, `slots=True` prevents dynamic attribute addition and makes multiple inheritance with other slotted classes more constrained. Each Config subclass in a deep hierarchy must repeat the `@dataclass(kw_only=True, slots=True)` decorator.

### 6. Two-level indirection for model selection

The old system required one identifier: `--job.config_file path/to/file.toml`. The new system requires two: `--module llama3 --config llama3_8b`. While this separates model identity from training recipe, it adds an extra argument.

---

## Numerics Verification

All model configs were verified for numerical equivalence against the main branch (commit `10d8a306`):

NOTE
- only models that can fit on 8 GPUs are tested
- only subset of parallelism combination are tested

| Model | Status | Notes |
|-------|--------|-------|
| llama3 (debugmodel, 8B) | Bitwise match | |
| llama3 (debugmodel_flex_attn) | Bitwise match | |
| qwen3 (0.6B, 1.7B, 32B, MoE debugmodel) | Bitwise match | |
| deepseek_v3 (debugmodel, 16B) | Close (max diff 0.00014) | Pre-existing main branch bug: missing `eps` in final RMSNorm |
| llama4 debugmodel | Bitwise match | _irope variants don't work on main (FlexAttn `'dict' object has no attribute 'BLOCK_SIZE'`) but now work after this PR |
| **gpt_oss** debugmodel | --debug.deterministic causes loss to be NaN; o/w first step loss match, minor difference after (likely caused by flex?) | |
| flux | Bitwise match | |

---

## Migration Guide

| Old (main) | New (this PR) |
|---|---|
| `CONFIG_FILE="path/to/config.toml" ./run_train.sh` | `MODEL=llama3 CONFIG=llama3_8b ./run_train.sh` |
| `--job.config_file path.toml` | `--module llama3 --config llama3_8b` |
| `train_configs/*.toml` | `config_registry.py` functions |
| `TrainSpec` | `ModelSpec` |
| `ModelArgs` / `args.py` | Nested `Model.Config` dataclass |
| `custom_config_module` + `_merge_configs()` | Subclass `Trainer.Config` |
| `build_model_converters()` free function | `ModelConvertersContainer.Config.build()` |
| `build_metrics_processor()` free function | `MetricsProcessor.Config.build()` |


wconstab:
Have you tried asking claude to help split this into a stack of PRs that can be reviewed/landed independently? I did see your comment/apology to reviewers, but i still think honestly nobody is going to review this PR in its entirety so are you asking for an uncareful scan and a stamp, or do you want to break out important pieces of the code that you want careful review on?

tianyu-l:
@wconstab While I understand how intimidating it could be for reviewing a huge PR, I would like to initially deliver the package as whole instead of letting people only see incremental changes (if it's possible at all). Maybe I would like to achieve
- [alignment] most reviewers would pick a single model / config registry and play with it and convince themselves the change looks OK in general. The correctness would be guaranteed by my numerics test and CI (will fix at least the core ones)
- [more careful check] If reviewers get aligned and would like to help review line-by-line, I'm more than happy to split into a stack of PRs.

scripts/generate/test_generate.py
```diff
@@ -101,21 +102,25 @@ def test_generate(
...
     # Tokenizer setup
-    # pyrefly: ignore [not-callable]
-    tokenizer = train_spec.build_tokenizer_fn(config)
+    from torchtitan.components.tokenizer import HuggingFaceTokenizer
+
+    tokenizer = HuggingFaceTokenizer(config.job.hf_assets_path)
 
-    model_args = train_spec.model_args[config.model.flavor]
-    model_args.update_from_config(config)
+    model_config = train_spec.model
```
acisseJZhong: in the previous test, we get `model_config` from `model_config = config.model_spec.model`, what's the difference between these two?

tests/integration_tests/features.py
```diff
@@ -443,35 +443,24 @@ def build_features_test_list() -> list[OverrideDefinitions]:
...
                     "--parallelism.data_parallel_replicate_degree=2",
                     "--training.enable_cpu_offload",
-                    "--optimizer.early_step_in_backward",
                 ],
             ],
             "Enable CPU Offload, Optimizer in backward with TP, DP, CP",
             "cpu_offload+opt_in_bwd+TP+DP+CP",
             ngpu=8,
         ),
-        OverrideDefinitions(
```
acisseJZhong, line 464: why do we delete this?

tests/unit_tests/test_dataset_checkpointing.py
```diff
@@ -60,21 +59,16 @@ def test_c4_resumption(self):
...
-                dataset_name,
-                "--training.local_batch_size",
-                str(batch_size),
-                "--training.seq_len",
-                str(seq_len),
-            ]
+
+        dl_config = HuggingFaceTextDataLoader.Config(
+            dataset=dataset_name,
+            local_batch_size=batch_size,
```
acisseJZhong: isn't `seq_len` and `local_batch_size` passed in through [init](https://github.com/pytorch/torchtitan/blob/991507604f9bf420d218ddb8f3f2747b03702359/torchtitan/hf_datasets/text_datasets.py#L193)?

tests/unit_tests/test_dataset_checkpointing.py
```diff
@@ -60,21 +59,16 @@ def test_c4_resumption(self):
...
+            seq_len=seq_len,
         )
 
-        return build_text_dataloader(
-            tokenizer=tokenizer,
+        return HuggingFaceTextDataLoader(
```
acisseJZhong: curious when shall we use config.build() vs direct instantiation?

tests/integration_tests/features.py
tianyu-l: IIUC memory estimation is not maintained any more. You can see the status has been set to disabled for while. cc @sanketpurandare

scripts/generate/test_generate.py
tianyu-l: This is claude-generated and seems a bit unclear. I'll refactor.

tests/unit_tests/test_dataset_checkpointing.py
tianyu-l: thanks, seems this test was in a limbo state -- will fix

torchtitan/components/optimizer.py
```diff
@@ -366,7 +257,7 @@ def _is_recomputation_enabled(module):
 
     def _update_expert_bias(
         model_parts: list[nn.Module],
-        parallel_dims: ParallelDims,
+        parallel_dims: "ParallelDims",
```
acisseJZhong: wondering what does the quote mean?

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -43,29 +50,33 @@
 def pipeline_llm(
     model: nn.Module,
     parallel_dims: ParallelDims,
-    job_config: JobConfig,
+    *,
```
acisseJZhong, line 50: what's the rule for differentiating the positional args and kw args?

torchtitan/distributed/dual_pipe_v.py
```diff
@@ -29,7 +30,12 @@
 """
 
 
-def get_dual_pipe_v_flag(job_config, parallel_dims) -> bool:
+def get_dual_pipe_v_flag(
+    *,
+    parallelism: ParallelismConfig,
+    ac_config: ActivationCheckpointConfig,
+    parallel_dims,
```
acisseJZhong: nit: missing type

torchtitan/experiments/rl/unified/__init__.py
```diff
@@ -22,42 +23,28 @@
 logger = init_logger(__name__)
 
 
-def register_torchtitan_model_from_train_spec(
-    train_spec: TrainSpec,
+def register_torchtitan_model_from_model_spec(
+    model_spec: ModelSpec,
     model_name: str,
```
acisseJZhong, line 28: `model_name` should be part of `model_spec`

torchtitan/models/common/attention.py
```diff
@@ -185,16 +195,11 @@ def forward(
         *,
         scale: float | None = None,
         enable_gqa: bool = False,
-        is_causal: bool = True,
+        is_casual: bool = True,
```
acisseJZhong: nit: typo

torchtitan/trainer.py
```diff
@@ -0,0 +1,851 @@
...
+        job: JobConfig = field(default_factory=JobConfig)
+        profiling: ProfilingConfig = field(default_factory=ProfilingConfig)
+        metrics: MetricsProcessor.Config = field(
+            default_factory=MetricsProcessor.Config
+        )
+        # TODO: remove the optional flag once Flux tokenizer is modeled properly
+        tokenizer: BaseTokenizer.Config | None = field(
+            default_factory=HuggingFaceTokenizer.Config
+        )
+        dataloader: BaseDataLoader.Config = field(default_factory=BaseDataLoader.Config)
```
acisseJZhong, line 84: do you know where's the code we skip loading?

run_train.sh
```diff
@@ -27,21 +27,21 @@ set -ex
...
-    NGPU="${NGPU}" LOCAL_RANK=0 python3 -m "${TRAIN_FILE}" --job.config_file "${CONFIG_FILE}" "$@" --comm.mode=${COMM_MODE} --training.steps=1
+    NGPU="${NGPU}" LOCAL_RANK=0 python3 -m torchtitan.train --model ${MODEL} --config ${CONFIG} "$@" --comm.mode=${COMM_MODE} --training.steps 1
 else
     # Normal training with torchrun
     PYTORCH_ALLOC_CONF="expandable_segments:True" \
     TORCHFT_LIGHTHOUSE=${TORCHFT_LIGHTHOUSE} \
     torchrun --nproc_per_node=${NGPU} --rdzv_backend c10d --rdzv_endpoint="localhost:0" \
     --local-ranks-filter ${LOG_RANK} --role rank --tee 3 \
-    -m ${TRAIN_FILE} --job.config_file ${CONFIG_FILE} "$@"
+    -m torchtitan.train --model ${MODEL} --config ${CONFIG} "$@"
```
wwwjn: What is this `model` config here doing? The `trainer.model_spec.model` is the correct source of "which model is this experiment running". Is it trying to override the model field?

torchtitan/models/qwen3/config_registry.py
```diff
@@ -0,0 +1,174 @@
...
+    ParallelismConfig,
+    TrainingConfig,
+)
+from torchtitan.hf_datasets.text_datasets import HuggingFaceTextDataLoader
+from torchtitan.trainer import Trainer
+
+from . import model_registry
+
+
+def qwen3_debugmodel() -> Trainer.Config:
```
wwwjn, line 22: I think the name of this function is confusing -- It seems referring to a model, but it actually refers to a specific experiment , or a specific run of experiment. 

I think we should rename it to something like `qwen3_debugmodel_pretrain` (later we could have rl/lora/sft, etc

wwwjn: Ref: https://fburl.com/code/eed0f0zp, they name each config instance with their each single experiment parameters

torchtitan/experiments/rl/unified/__init__.py
wwwjn: Model name refers to another thing: `model_name: Name to register in vLLM (e.g., "Qwen3TorchTitanForCausalLM")`. Maybe need a more descriptive name here, but that could be done in a separate PR

torchtitan/models/flux/flux_datasets.py
```diff
@@ -407,51 +362,95 @@ def __iter__(self):
...
+        local_batch_size: int,
+        **kwargs,
+    ):
+
+        t5_tokenizer, clip_tokenizer = build_flux_tokenizer(
+            encoder_config=config.encoder,
+            hf_assets_path=config.hf_assets_path,
+        )
+
+        if config.generate_timesteps:
```
wwwjn, line 417: This PR added `classifier_free_guidance_prob` to FluxDataset API, which makes the `FluxValidationDataset` has to take this classifier_free_guidance_prob as well.  However, in validation, the classifier_free_guidance_prob should always be 0.

Now `generate_timesteps` is used to distinguish if the dataloader is validation dataloader or training dataloader. Let's  hard code the `classifier_free_guidance_prob = 0` and add a warning logging when validation

torchtitan/models/flux/README.md
```diff
@@ -25,7 +25,7 @@ Run the following command to train the model on a single GPU:
 
 If you want to train with other model args, run the following command:
 ```bash
-CONFIG_FILE="./torchtitan/models/flux/train_configs/flux_schnell_model.toml" ./torchtitan/models/flux/run_train.sh
+MODEL=flux CONFIG=flux_schnell ./torchtitan/models/flux/run_train.sh
```
wwwjn: `flux/run_train.sh` is removed, I guess we want to get the run_train.sh back?

torchtitan/config/configs.py
```diff
@@ -0,0 +1,440 @@
...
+Configs without a clear single owner (or with circular-import constraints)
+live here.
+"""
+
+from dataclasses import dataclass, field
+from typing import Literal
+
+
+@dataclass(kw_only=True, slots=True)
+class DataLoaderConfig:
```
wwwjn: Why this doesn't lives in `ParallelAwareDataLoader.Config`? These field are only used as `kwargs` there

torchtitan/config/configs.py
```diff
@@ -0,0 +1,440 @@
...
+    Note that you may want to lower the training steps to avoid generating too
+    many temporary files.
+    """
+
+    dataloader: DataLoaderConfig = field(default_factory=DataLoaderConfig)
+    """DataLoader configuration"""
+
+
+@dataclass(kw_only=True, slots=True)
+class JobConfig:
```
wwwjn: This name sounds to broad to me. Previously we use `job_config` to refer to all the configs related to a single job / a run / an experiment. Now this class is super light-weight , and I think a better way to add these field:
- `hf_assets_path` / `dump_folder` -> TrainingConfig,
- Remove `description` 
- `print_config` / `save_config_file` -> DebugConfig

torchtitan/experiments/transformers_modeling_backend/README.md
```diff
@@ -6,20 +6,15 @@ This enables HF transformers models to be trained with `4D parallelism + torch.c
...
-
-+[hf_transformers]
-+model = "Qwen/Qwen3-4B-Instruct-2507"
+- --model llama3
++ --model transformers_modeling_backend
+--config transformers_modeling_backend_debugmodel
 ...
 ```
-- Train: `LOG_RANK=7 CONFIG_FILE=<YOUR_PATH>/torchtitan/experiments/transformers_modeling_backend/configs/qwen3.toml ./run_train.sh --job.custom_config_module=torchtitan.experiments.transformers_modeling_backend.job_config --compile.enable`
+- Train: `LOG_RANK=7 MODEL=transformers_modeling_backend CONFIG=transformers_modeling_backend_debugmodel ./run_train.sh --compile.enable`
```
wwwjn, line 17: `MODEL=transformers_modeling_backend ` is a confusing cause this experiment folder is not actually split by each model

run_train.sh
wwwjn: Oh I see, in `config/manager.py`. I think one main reason we organize the whole repo by models, and assume that `--model` would refer to folder which is named after a model's name. Also it's because we stored `config_registery` in a de-centralized way.

wwwjn (commented):
Mainly took a look on `components`, `config`, `experiments/rl`, `models`, `train.py` and `trainer.py`

torchtitan/trainer.py
```diff
@@ -0,0 +1,851 @@
...
+        def __post_init__(self):
+            if isinstance(self.optimizer, OptimizersInBackwardContainer.Config):
+                if self.parallelism.expert_parallel_degree > 1:
+                    raise NotImplementedError(
+                        "Optimizers in backward is not supported with Expert Parallel."
+                    )
+                if self.parallelism.pipeline_parallel_degree > 1:
+                    raise NotImplementedError(
+                        "Optimizers in backward is not supported with Pipeline Parallel."
+                    )
```
fegin, line 116: Why do we do the check here? Is it because this is the only place where we have enough information to do the check?

torchtitan/models/flux/README.md
tianyu-l: why? now people will always use the main run_train.sh, because the Trainer class is from config_registry

torchtitan/config/configs.py
tianyu-l: Removing description sounds fine. I kept it there because I wasn't sure if ModelSpec need to keep `name` and `flavor` -- I somehow still think they are not absolutely necessary, if you think about it from experiment perspective.

I can put `print_config` and `save_config_file` to DebugConfig.

torchtitan/models/flux/README.md
wwwjn: Oh I see, if we can launch flux model using main run_trian.sh, we need to modify the README this line - It still points to flux/run_train.sh

torchtitan/components/quantization/float8.py
```diff
@@ -235,5 +278,11 @@ def post_optimizer_hook(self, model: nn.Module | list[nn.Module]):
         pass
 
 
-register_model_converter(Float8LinearConverter, "quantize.linear.float8")
-register_model_converter(Float8GroupedMMConverter, "quantize.grouped_mm.float8")
```
danielvegamyhre, line 239: i see this removes the converter names (`quantize.grouped_mm.float8` etc) - what does the command line API for this look like now?

torchtitan/components/quantization/float8.py
```diff
@@ -235,5 +278,11 @@ def post_optimizer_hook(self, model: nn.Module | list[nn.Module]):
...
+def find_float8_linear_config(
```
danielvegamyhre, line 281: where is this called? i tried searching but couldn't find it, maybe it's in a minimized file somewhere

torchtitan/components/metrics.py
```diff
@@ -361,20 +402,31 @@ class MetricsProcessor:
...
-        job_config: JobConfig,
+        config: Config,
+        *,
         parallel_dims: ParallelDims,
+        dump_folder: str = "./outputs",
+        pp_schedule: str = "1F1B",
+        ft_enable: bool = False,
+        ft_replica_id: int = 0,
+        config_dict: dict[str, Any] | None = None,
         tag: str | None = None,
```
fegin, line 329: I also am worried about the kwargs being a loophole where people passing configurations around.

fegin: An alternative approach is to require each component define these shared configurations and resolve the shared configurations when constructing the root configuration (Trainer.Config).

torchtitan/models/qwen3/config_registry.py
tianyu-l: This is not introduced by this PR. In the past we call it debugmodel.toml. Feel free to improve it in a separate PR, although `pretrain` also doesn't sound accurate.

torchtitan/trainer.py
tianyu-l: Yes.

torchtitan/components/metrics.py
tianyu-l: Yeah, this is the top issue I put in "Longer-term Issues" in PR summary, which I couldn't handle entirely in this initial PR.

First, we need to figure out the boundary between "shared config" and "runtime kwargs". We can use more shared config, but that is "utilizing" (a.k.a. "abusing") the python config power in a way that makes it harder to transform to pure yaml solution, which may be OK.

More importantly, we need to reconsider if the current function calling structure makes sense at all. Current metrics logging is limited and hard to customize -- e.g. in MoE how to log the number of tokens each expert processes? In this sense, such problems are reflecting the design flaws we have in torchtitan -- in the past, these are omitted due to the usage of JobConfig everywhere. I think this is one of the good things about this refactor.

torchtitan/components/quantization/float8.py
tianyu-l: llama3/parallelize.py

torchtitan/components/quantization/float8.py
tianyu-l: We lose CLI capability for adjusting this, because there is no string attached to each converter anymore.

torchtitan/experiments/transformers_modeling_backend/README.md
tianyu-l: I can change to MODULE, but this translating field sounds not necessary.

tests/unit_tests/test_dataset_checkpointing.py
tianyu-l: Should always call build, for polymorphism. But in test it doesn't matter that much.

torchtitan/components/optimizer.py
tianyu-l: removed

torchtitan/trainer.py
tianyu-l: There's no code for that. You need to create / replace the dataloader to be a NoDataLoader whose constructor does nothing and returns `None` (or a no-op dataloader).

torchtitan/config/configs.py
tianyu-l: putting hf_assets_path / dump_folder to root level for now.

torchtitan/models/flux/flux_datasets.py
tianyu-l: leaving a TODO for future work

torchtitan/models/llama3/model.py
```diff
@@ -0,0 +1,122 @@
...
+        dim (int): Model dimension.
+        n_layers (int): Total number of layers.
+        config (Llama3TransformerBlock.Config): Block configuration.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(TransformerBlock.Config):
+        depth_init: bool = True
+
+    def __init__(self, config: Config, *, layer_id: int, dim: int, n_layers: int):
```
wconstab, line 35: unintuitive to me: why is dim not part of Config, for example? does this mean that dim is not easily overridable from configs?

tianyu-l: @wconstab Great point.

Short answer is that `dim` should be coming from the parent config decoder. I guess the "normal" way of doing this is https://github.com/apple/axlearn/blob/main/docs/ml_api_style.md#configuration-of-composite-modules:
- dim being of the transformer block config
- in build time, override the existing block_config.dim by decoder.dim

Indeed, people should be able to config child module entirely, without assuming parent behavior.

I personally find that to be less readable, because now you can set a block_config when authoring the model, without actually use it during runtime. Of course, you can get the source of truth after config build is finished and logged.

I feel this level of details and opinions should be discussed in followup PRs than in this PR. In this PR, I would like to get a minimal working version.

torchtitan/distributed/pipeline_parallel.py
tianyu-l: I'm inspired by https://github.com/apple/axlearn/blob/main/docs/ml_api_style.md#avoid-multiple-positional-arguments. Here I'm moving `parallel_dims` to kwarg as well.

acisseJZhong: > limit the number of positional arguments to <= 1 and use keyword arguments for the rest

what's the reason not making them all kwargs?

torchtitan/models/llama3/model.py
wconstab: I see- this tradeoff makes sense. I don't know which version I prefer.

run_train.sh
tianyu-l: changing to --module

torchtitan/distributed/pipeline_parallel.py
tianyu-l: I don't know for sure. Likely because for some functions, there would always be a "main" arg that is always there and doesn't introduce ambiguity / error-proneness. E.g. if a function only takes one arg, like parallelize(model), maybe it's fine? You can imagine later on when people adds more and more optional kwargs to the function, the model part doesn't need to be changed.

docs/checkpoint.md
```diff
@@ -60,15 +61,15 @@ A seed checkpoint does initialization of the model on a single CPU, and can be l
 To create a seed checkpoint, use the same model config as you use for training.
 e.g.
 ```bash
-NGPU=1 CONFIG_FILE=<path_to_model_config> ./run_train.sh --checkpoint.enable --checkpoint.create_seed_checkpoint --parallelism.data_parallel_replicate_degree 1 --parallelism.data_parallel_shard_degree 1 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1
+NGPU=1 ./run_train.sh --checkpoint.create_seed_checkpoint --parallelism.data_parallel_replicate_degree 1 --parallelism.data_parallel_shard_degree 1 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1
```
wwwjn: Nit: Need to add `--module <module_name>` and `--config <config_name>` to this command as the seed checkpoint is not sharable across different models?

README.md
```diff
@@ -142,7 +142,7 @@ python scripts/download_hf_assets.py --repo_id meta-llama/Llama-3.1-8B --assets
 Llama 3 8B model locally on 8 GPUs
 
 ```bash
-CONFIG_FILE="./torchtitan/models/llama3/train_configs/llama3_8b.toml" ./run_train.sh
+MODEL=llama3 CONFIG=llama3_8b ./run_train.sh
```
wwwjn:
```suggestion
MODULE=llama3 CONFIG=llama3_8b ./run_train.sh
```

docs/debugging.md
```diff
@@ -81,7 +83,7 @@ NGPU=32 COMM_MODE="fake_backend" ./run_train.sh
 **Example use case:**
 ```bash
 # Validate a 128-GPU configuration on a single GPU
-NGPU=128 COMM_MODE="fake_backend" CONFIG_FILE="./train_configs/llama3_70b.toml" ./run_train.sh
+NGPU=128 COMM_MODE="fake_backend" MODEL=llama3 CONFIG=llama3_70b ./run_train.sh
```
wwwjn: M
```suggestion
NGPU=128 COMM_MODE="fake_backend" MODULE=llama3 CONFIG=llama3_70b ./run_train.sh
```

docs/debugging.md
```diff
@@ -177,15 +179,15 @@ For multiple experimental runs with different parallelism configs, we need to us
 #### Creating a Seed Checkpoint
 
 ```bash
-NGPU=1 CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml" ./run_train.sh --checkpoint.enable --checkpoint.create_seed_checkpoint --parallelism.data_parallel_replicate_degree 1 --parallelism.data_parallel_shard_degree 1 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1
+NGPU=1 ./run_train.sh --checkpoint.enable --checkpoint.create_seed_checkpoint --parallelism.data_parallel_replicate_degree 1 --parallelism.data_parallel_shard_degree 1 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1
```
wwwjn:
```suggestion
NGPU=1 MODLUE=llama3 CONFIG=llama3_debugmodel ./run_train.sh --checkpoint.enable --checkpoint.create_seed_checkpoint --parallelism.data_parallel_replicate_degree 1 --parallelism.data_parallel_shard_degree 1 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1
```

wwwjn: Explicit list the module and config here, in case of misleading users to use the implicit default values

docs/debugging.md
```diff
@@ -177,15 +179,15 @@ For multiple experimental runs with different parallelism configs, we need to us
...
 ```
 
 #### Loading Seed Checkpoints for Debugging
 
 When using seed checkpoints for debugging or validation purposes, you can enable the `load_only` configuration to load checkpoints without saving any new ones during training. This is particularly useful when you only want to verify model correctness or compare different configurations without cluttering your disk:
 
 ```bash
-CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml" ./run_train.sh --checkpoint.enable --checkpoint.load_only
+./run_train.sh --checkpoint.enable --checkpoint.load_only
```
wwwjn:
```suggestion
MODLUE=llama3 CONFIG=llama3_debugmodel  ./run_train.sh --checkpoint.enable --checkpoint.load_only
```

docs/extension.md
```diff
@@ -34,53 +31,46 @@ To perform various tasks, from adding a new model (possibly with a new modality)
...
 ```
 
-Or specify it in your `.toml` config:
+Then run with:
 
-```toml
-[job]
-custom_config_module = "torchtitan.experiments.your_folder.job_config"
+```bash
+MODEL=your_folder CONFIG=my_experiment_debugmodel ./run_train.sh
```
wwwjn:
```suggestion
MODULE=your_folder CONFIG=my_experiment_debugmodel ./run_train.sh
```

Need to fix "MODEL" -> "MODULE" globally

docs/debugging.md
```diff
@@ -1,8 +1,8 @@
 ## Enable Memory Profiling
 
-Launch training job with the following command (or alternatively set configs in toml files)
+Launch training job with the following command (or alternatively set configs in your config_registry function)
 ```
-CONFIG_FILE="./train_configs/debug_model.toml" ./run_train.sh --profiling.enable_memory_snapshot --profiling.save_memory_snapshot_folder memory_snapshot
+./run_train.sh --profiling.enable_memory_snapshot --profiling.save_memory_snapshot_folder memory_snapshot
```
wwwjn:
```suggestion
MODLUE=llama3 CONFIG=llama3_debugmodel  ./run_train.sh --profiling.enable_memory_snapshot --profiling.save_memory_snapshot_folder memory_snapshot
```

benchmarks/llama3-8b_h200_202506_trainy-whitefiber.md
```diff
@@ -26,7 +25,7 @@ To reproduce the results using the original torchtitan commit, change all instan
     --rdzv_backend c10d \
     --rdzv_endpoint "$MASTER_ADDR:29500" \
     torchtitan/train.py \
-    --job.config-file torchtitan/models/llama3/train_configs/llama3_8b.toml \
+    --job.config-file torchtitan/models/llama3/train_configs/llama3_8b.py \
```
fegin: We should keep the toml since this is the original command.

tests/unit_tests/test_checkpoint.py
```diff
@@ -93,9 +92,9 @@ def fake_async_save(*args, **kwargs):
 
 
 class DummyJobConfig:
```
wwwjn: Can we fully get rid of the name job_config in this file? There's a lot of apearence of `job_config` in this file. Maybe name it "trainer_config"

tests/unit_tests/test_dataset_flux.py
```diff
@@ -53,33 +51,35 @@ def test_load_dataset(self):
                 # in the dataset, then the test will fail, due to huggingface's
                 # non-resumption when checkpointing after the first epoch
 
-                path = "torchtitan.models.flux.job_config"
+                # Load flux config via --model/--config
```
wwwjn:
```suggestion
                # Load flux config via --module/--config
```

scripts/generate/run_llama_generate.sh
```diff
@@ -12,7 +12,8 @@ set -e
 # LOG_RANK=0,1 NGPU=4 ./run_llama_generate.sh
 NGPU=${NGPU:-"1"}
 LOG_RANK=${LOG_RANK:-0}
-CONFIG_FILE=${CONFIG_FILE:-"./torchtitan/models/llama3/train_configs/debug_model.toml"}
+MODEL=${MODEL:-"llama3"}
```
fegin: To be consistent with `run_train.sh`, we should use `MODULE`

scripts/generate/README.md
```diff
@@ -17,15 +17,15 @@ This script is ideal for users who need to:
 #### Run on a single GPU.
 
 ```bash
-NGPU=1 CONFIG_FILE=./torchtitan/models/llama3/train_configs/llama3_8b.toml CHECKPOINT_DIR=./outputs/checkpoint/ \
+NGPU=1 MODEL=llama3 CONFIG=llama3_8b CHECKPOINT_DIR=./outputs/checkpoint/ \
```
fegin:
```suggestion
NGPU=1 MODULE=llama3 CONFIG=llama3_8b CHECKPOINT_DIR=./outputs/checkpoint/ \
```

scripts/generate/README.md
```diff
@@ -17,15 +17,15 @@ This script is ideal for users who need to:
...
 PROMPT="What is the meaning of life?" \
 ./scripts/generate/run_llama_generate.sh --max_new_tokens=32 --temperature=0.8 --seed=3
 ```
 
 #### Run on 4 GPUs and pipe results to a json file.
 
 ```bash
-NGPU=4 CONFIG_FILE=./torchtitan/models/llama3/train_configs/llama3_8b.toml CHECKPOINT_DIR=./outputs/checkpoint/ \
+NGPU=4 MODEL=llama3 CONFIG=llama3_8b CHECKPOINT_DIR=./outputs/checkpoint/ \
```
fegin:
```suggestion
NGPU=4 MODULE=llama3 CONFIG=llama3_8b CHECKPOINT_DIR=./outputs/checkpoint/ \
```

torchtitan/protocols/model_spec.py
```diff
@@ -0,0 +1,51 @@
...
+    # signatures elsewhere in the codebase.
+    build_loss_fn: Callable
+    parallelize_fn: Callable
+    pipelining_fn: Callable | None
+    post_optimizer_build_fn: Callable | None
+    state_dict_adapter: type[BaseStateDictAdapter] | None
+
+
+@dataclass
+class FaultTolerantModelSpec(ModelSpec):
```
wwwjn, line 50: What is the fragment_fn and FaultTolerantModelSpec do? Sorry the diff is so huge to search

tests/unit_tests/test_job_config.py
```diff
@@ -4,211 +4,137 @@
...
-from dataclasses import dataclass
 
 import pytest
-import tomli_w
-from torchtitan.config import ConfigManager, JobConfig
+from torchtitan.config import ConfigManager
+from torchtitan.trainer import Trainer
 
 
 class TestJobConfig(unittest.TestCase):
```
wwwjn: Again should we still call it "job config" in this test, and this file name? This file can be renamed to "test_config_manager" as it mainly test parsing different CLI and other config manager functionalities

tests/unit_tests/test_activation_checkpoint.py
```diff
@@ -28,7 +29,7 @@
     # used to compute the scaling factor for quantization.
     torch.ops.aten.max.default,
     torch._higher_order_ops.flex_attention,
-    torch.ops.torch_attn._varlen_attn,
+    torch.ops.torch_attn._varlen_attn.default,
```
fegin, line 32: Is this change correct? I saw some don't have ".default" but some have. But this is a custom op, so I would expect no ".default". cc., @liangel-02

tests/README.md
```diff
@@ -31,13 +30,12 @@ pip install -r requirements.txt
 To run the integration tests:
 
 ```bash
-python -m tests.integration_tests.run_tests <output_dir> [--config_path CONFIG_PATH] [--test_suite TEST_SUITE] [--test_name TEST_NAME] [--ngpu NGPU]
```
wwwjn, line 34: Previously we have this `config_path` is to quickly let user run the same test suites on different model configs locally, eg, running multiple parallelism tests on deepseek-v3. Do you think this is still a valid use case? If so, we should be able to pass --module and --config to run_tests.py

torchtitan/trainer.py
```diff
@@ -0,0 +1,858 @@
...
+            lr_schedulers=self.lr_schedulers,
+            states={"train_state": self},
+            checkpoint_config=config.checkpoint,
+            sd_adapter=(
+                model_spec.state_dict_adapter(model_config, config.hf_assets_path)
+                if model_spec.state_dict_adapter
+                else None
+            ),
+            base_folder=config.dump_folder,
+        )
```
fegin: Curious why Checkpointer doesn't follow build() style?

torchtitan/components/lr_scheduler.py
```diff
@@ -83,117 +230,27 @@ def load_state_dict(self, state_dict: dict[str, Any]) -> None:
 
 def build_lr_schedulers(
```
wwwjn: When do we need these "build_xxx" function? In our current config system, trainer.config will automatically call the LRScheduler class's config.build() IIUC?

Similar for dataloader, etc

torchtitan/components/validate.py
```diff
@@ -15,44 +16,90 @@
...
 
 
 class Validator(BaseValidator):
     """
     Simple validator focused on correctness and integration.
 
     Args:
-        job_config: Job configuration
-        validation_dataloader: The validation dataloader
+        config: Validator.Config configuration
```
wwwjn, line 55: We need to pass all these fields to validator because these information are from trainer.config() and we are not passing the whole job_config anymore, right? It sounds hard to simplify this part

torchtitan/experiments/autoparallel/README.md
```diff
@@ -12,14 +12,14 @@ Requires installing [git@github.com:meta-pytorch/autoparallel.git](https://githu
 
 **Llama3**
 
-`CONFIG_FILE=./torchtitan/models/llama3/train_configs/debug_model.toml ./run_train.sh --model.name autoparallel.llama3 --parallelism.tensor_parallel_degree 4 --job.custom_config_module=torchtitan.experiments.autoparallel.job_config`
+`MODEL=autoparallel.llama3 CONFIG=autoparallel_llama3_debugmodel ./run_train.sh --parallelism.tensor_parallel_degree 4`
```
wwwjn:
```suggestion
`MODULE=autoparallel.llama3 CONFIG=autoparallel_llama3_debugmodel ./run_train.sh --parallelism.tensor_parallel_degree 4`
```

torchtitan/experiments/autoparallel/README.md
```diff
@@ -12,14 +12,14 @@ Requires installing [git@github.com:meta-pytorch/autoparallel.git](https://githu
...
 
 **DeepSeekv3**
 
-`NGPU=2 CONFIG_FILE=./torchtitan/models/deepseek_v3/train_configs/debug_model.toml ./run_train.sh --model.name autoparallel.deepseek_v3 --job.custom_config_module=torchtitan.experiments.autoparallel.job_config`
+`NGPU=2 MODEL=autoparallel.deepseek_v3 CONFIG=autoparallel_deepseek_v3_debugmodel ./run_train.sh`
```
wwwjn:
```suggestion
`NGPU=2 MODULE=autoparallel.deepseek_v3 CONFIG=autoparallel_deepseek_v3_debugmodel ./run_train.sh`
```

torchtitan/experiments/autoparallel/README.md
```diff
@@ -12,14 +12,14 @@ Requires installing [git@github.com:meta-pytorch/autoparallel.git](https://githu
...
 
 **DeepSeekv3 local_map**
 
 This is a variant of titan's DSv3, which uses a local_map for the expert parallel region. This only supports 2D mesh right now. NOTE: the mesh provided are just to reuse torchtitan's trainer mesh setup code. Autoparallel is not bound to use dp2ep.
 
-`NGPU=2 CONFIG_FILE=./torchtitan/models/deepseek_v3/train_configs/debug_model.toml tlp ./run_train.sh --model.name autoparallel.local_map_deepseek_v3 --job.custom_config_module=torchtitan.experiments.autoparallel.job_config --parallelism.data_parallel_shard_degree 2 --parallelism.expert_parallel_degree 2`
+`NGPU=2 MODEL=autoparallel.local_map_deepseek_v3 CONFIG=autoparallel_local_map_deepseek_v3_debugmodel ./run_train.sh --parallelism.data_parallel_shard_degree 2 --parallelism.expert_parallel_degree 2`
```
wwwjn:
```suggestion
`NGPU=2 MODULE=autoparallel.local_map_deepseek_v3 CONFIG=autoparallel_local_map_deepseek_v3_debugmodel ./run_train.sh --parallelism.data_parallel_shard_degree 2 --parallelism.expert_parallel_degree 2`
```

torchtitan/experiments/forge/job_config.py
```diff
@@ -7,46 +7,47 @@
...
+    ParallelismConfig,
+    TrainingConfig,
 )
+from torchtitan.protocols.model_converter import ModelConvertersContainer
+from torchtitan.protocols.model_spec import ModelSpec
 
 
 # Parity w/ TorchTitan commit: 8ec37d2bca7ac9d3f7517ba70ac10e75e22a7bcb
 @dataclass
 class ForgeJobConfig:
```
wwwjn: Probably rename it to `ForgeTrainerConfig`?

Also why we don't put this into ForgeEngine.Config?

torchtitan/experiments/rl/unified/infra/parallelism_utils.py
```diff
@@ -114,14 +117,10 @@ def create_job_config_from_vllm_config(
     # Create JobConfig with defaults
     job_config = JobConfig()
```
wwwjn, line 118: You can leave the config system in RL for now, I will rebase my PRs on top of this config system change and refactor the config in RL at the same time

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -47,17 +51,196 @@ def __len__(self):
...
+    @dataclass(kw_only=True, slots=True)
+    class Config(BaseModel.Config, PretrainedConfig):
+        """Configuration that bridges TorchTitan and HuggingFace Transformers.
+
+        Uses properties to provide TorchTitan-style access while maintaining
+        HuggingFace compatibility.
+        """
+
+        # Define all possible mappings organized by argument type
+        _TT_TO_HF_MAPPINGS = {
```
wwwjn: This field should be a property of HFTransformerModel class, or hardcoded. They should not be configurable?

torchtitan/models/common/rope.py
wwwjn, line 1: I would suggest running `test_loss.py` locally to compare if model loss changes (before and after this PR, for other models like deepseek-v3, qwen3, etc,  that we didn't guarded in CI), I feel like extracting these common components are dramatical changes, and need to verify correctness.

torchtitan/models/common/rope.py
```diff
@@ -0,0 +1,347 @@
...
+        cfg = self.config
+        dim = cfg.dim
+        max_seq_len = cfg.max_seq_len
+        base = cfg.theta
+
+        freq = base ** (torch.arange(0, dim, 2)[: (dim // 2)].float() / dim)
+        mscale = 1.0
+
+        if cfg.scaling == "llama":
+            raise NotImplementedError("Cos/sin RoPE does not support Llama scaling.")
```
wwwjn, line 151: What is llama scaling?

torchtitan/models/common/rope.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
+    xk_ = torch.view_as_complex(xk.float().reshape(*xk.shape[:-1], -1, 2))
+    freqs_cis = _reshape_for_broadcast_complex(freqs_cis, xq_, positions)
+    xq_out = torch.view_as_real(xq_ * freqs_cis).flatten(3)
+    xk_out = torch.view_as_real(xk_ * freqs_cis).flatten(3)
+    return xq_out.type_as(xq), xk_out.type_as(xk)
+
+
+def apply_rotary_emb_single_complex(
```
wwwjn, line 309: No need to have 2 separate `apply_rotary_emb_single_complex` and `apply_rotary_emb_complex`, the only difference for these 2 function is taking xq / xk together or separately? We should be able to align deepseek-v3 and llama3 here

torchtitan/models/common/rope.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def _precompute_complex(self) -> torch.Tensor:
+        cfg = self.config
+        dim = cfg.dim
+        end = cfg.max_seq_len
+        theta = cfg.theta
+
+        freqs = 1.0 / (theta ** (torch.arange(0, dim, 2)[: (dim // 2)].float() / dim))
+
+        if cfg.scaling == "llama":
```
wwwjn, line 71: We name it "llama3" because llama first applies the complex style RoPE ?

torchtitan/models/llama3/__init__.py
```diff
@@ -5,112 +5,217 @@
...
+                hidden_dim=compute_ffn_hidden_dim(
+                    16384, multiple_of=4096, ffn_dim_multiplier=1.2
+                )
+            ),
+            attention=GQAttention.Config(
+                n_heads=128, n_kv_heads=8, attn_backend="sdpa", rope_backend="complex"
+            ),
+        ),
+        rope=RoPE.Config(
+            dim=16384 // 128,
```
wwwjn, line 202: This is so magic - Should it be  dim // n_heads?

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -5,168 +5,237 @@
...
+                qk_rope_head_dim=64,
+                v_head_dim=128,
+                mscale=0.70,
+                attn_backend="flex",
+                attn_mask_type="block_causal",
+            ),
+            feed_forward=FeedForward.Config(hidden_dim=10944),
+        ),
+        rope=RoPE.Config(
+            dim=64,
```
wwwjn, line 133: This dim should not be randomly set / experimented with IIUC, it should be `self.model_args.dim // self.model_args.n_heads`. Previously we are passing this dim from  Transformer. Can we enforce this constrain?

torchtitan/models/common/decoder.py
wwwjn, line 1: @shuhuayu  Need your help on reviewing the `common/` folder, this change would affect most of the model initialization / forward

torchtitan/models/common/rope.py
tianyu-l: I indeed verified, see PR summary.

torchtitan/models/common/rope.py
tianyu-l: It is a rope scaling which llama3 and llama4 used. Check old llama code.

torchtitan/models/common/rope.py
tianyu-l: Not sure if I understand the question: yes only llama applied the style, but it's not complex. Complex is another dimension.

torchtitan/models/llama3/__init__.py
tianyu-l: I don't think I can make rope perfect in this PR, feel free to change after this first version.

torchtitan/models/deepseek_v3/__init__.py
tianyu-l: feel free to change this in next PR

torchtitan/models/llama3/__init__.py
shuhuayu: Let me handle rope across models after this pr. But I will have a first look here.

wwwjn: Cool, let's create an issue to track all the short-term TODOs after this PR being landed, and assign to folks?

torchtitan/models/common/rope.py
wwwjn: Thanks, what's the kown issue on main for  llama4, gpt_oss, flux? The link to `report_v2.md` is missing

tianyu-l: llama4 and flux can't run on main before, under the parallelism I used, so I can't compare. After this PR they can run.
gpt_oss on main and this PR produce the same NaN result. I'll link report_v2.md somewhere.

torchtitan/models/llama3/__init__.py
tianyu-l: sounds good, I'll do another pass and mark all the things I can't fix as TODO or in the PR summary.

torchtitan/models/common/attention.py
```diff
@@ -359,3 +364,176 @@ def create_varlen_metadata_for_document(
...
+    class Config(BaseAttention.Config):
+        n_heads: int
+        n_kv_heads: int | None = None
+        head_dim: int | None = None
+        qk_norm: bool = False
+        norm_eps: float = 1e-5
+        bias: bool = False
+        use_rope: bool = True
+        attn_backend: str = "sdpa"
+        attn_mask_type: str = "causal"
```
fegin, line 414: We had the validation for attn_mask_type + attn_backend combinations. This seem to be missing. We can add a TODO if it is too much for this PR.

torchtitan/models/common/rope.py
wwwjn: It's superising, I check the latest model test CI and the gpt-oss debug model is not NaN: http://github.com/pytorch/torchtitan/actions/runs/22205981792/job/64229909903 

Which gpt-oss model are you running?

> llama4 and flux can't run on main before

Do you mean larger model? We have CI guard on debug model. Could it because of pytorch-nightly version?

scripts/checkpoint_conversion/numerical_tests_example.py
```diff
@@ -119,10 +120,14 @@ def forward_tt(config_path, checkpoint_path, test_set):
...
-    # pyrefly: ignore [not-callable]
-    tokenizer = train_spec.build_tokenizer_fn(config)
+    config = config_manager.parse_args(
+        ["--module", model_name, "--config", config_name]
+    )
+
+    from torchtitan.components.tokenizer import HuggingFaceTokenizer
+
+    # pyrefly: ignore [missing-argument, missing-attribute]
+    tokenizer = HuggingFaceTokenizer(config.hf_assets_path)
```
acisseJZhong, line 130: why not using `tokenizer = HuggingFaceTokenizer.Config().build()` here?

torchtitan/components/metrics.py
```diff
@@ -331,21 +342,51 @@ def _build_metric_logger(
...
+        only stage that computes loss metrics.
+        """
+
+        enable_wandb: bool = False
+        """Whether to log metrics to Weights & Biases"""
+
     logger: BaseLogger
     parallel_dims: ParallelDims
-    job_config: JobConfig
+    config: "MetricsProcessor.Config"
```
acisseJZhong:
```suggestion
    config: MetricsProcessor.Config
```

torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
```diff
@@ -68,26 +77,26 @@ def parallelize_deepseekv3(
...
             ),
         )
 
-    should_compile = job_config.compile.enable
+    should_compile = compile_config.enable
     if should_compile:
         # TODO: support more options in AP API
-        assert job_config.compile.components == ["model"]
-        assert job_config.compile.backend == "inductor"
+        assert compile_config.components == ["model"]
```
wconstab, line 99: one thing i wonder, which may be a longer term change, is whether we want a way from the config system to override the compile settings per layer or per region.

benchmarks/llama3-8b_h200_202506_trainy-whitefiber.md
wconstab: what do you mean?

torchtitan/config/configurable.py
```diff
@@ -0,0 +1,67 @@
...
+    - Gets build() auto-wired via __init_subclass__ (no manual override needed)
+    - Accepts __init__(self, config: Config, **runtime_kwargs)
+
+    Enforcement: Configurable.__init_subclass__ checks that every Config uses
+    @dataclass(kw_only=True, slots=True). This check runs on the OUTER class
+    (not Config.__init_subclass__) because @dataclass(slots=True) replaces the
+    class, so Config.__init_subclass__ sees the pre-decorator version.
+    """
+
+    @dataclass(kw_only=True, slots=True)
```
wconstab, line 26: i assume the kwonly design choice is to ensure config files are readable and not error prone, any other reason?

scripts/checkpoint_conversion/numerical_tests_example.py
tianyu-l: Config build provides flexibility, not constraints. You don't have to use it everywhere, especially in this hardcoded script.

torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
tianyu-l: So an explicit place where this refactor is different from internal is that I don't feel necessary to let each module to own compile spec and ac spec, given all LLMs share a similar structure Decoder -> TransformerLayers -> ... and we always FSDP / compile / AC on the same places (a.k.a TransformerLayers). The only other mode would be full graph compile, which can also be conveniently done outside.

Sharding is something @fegin and I would like to move to per-module based, along with the full DTensor change.

torchtitan/config/configurable.py
tianyu-l: right, that's the reason -- a more aggressive config system would also require frozen=True.

aditvenk:
cc @yiming0416  -- this PR is running into a failure in compiler toolkit. We should fix the experiment once this is merged.

torchtitan/experiments/autoparallel/local_map_deepseek_v3/parallelize_deepseekv3.py
wconstab: i agree, it seems more important to customize how compile is applied within a block, rather than how it's used across blocks.  (Unless the blocks themselves are structurally different).

xmfan:
@tianyu-l Autoparallel failure fix upstream: https://github.com/meta-pytorch/autoparallel/pull/321. Once landed, CI should pass when you restart the job, because AP experiment clones github repo.

torchtitan/protocols/model_spec.py
tianyu-l: Please check FaultTolerantTrainSpec in main and tushar's past PRs.

benchmarks/llama3-8b_h200_202506_trainy-whitefiber.md
tianyu-l: @fegin means that this command was used to generate the benchmark result, in case anyone wants to reproduce.
Whenever they try to repro, they need to roll back to past version, where the repo accepts .toml, not .py config.

tests/unit_tests/test_activation_checkpoint.py
tianyu-l: confirmed that we should have .default everywhere

torchtitan/trainer.py
tianyu-l: modified

torchtitan/components/validate.py
tianyu-l: Depending on how flexible you want validator to be, we can have different abstractions. E.g. if validator only runs forward on a different dataset, maybe we can just reuse trainer's forward capability to do validation.

torchtitan/components/lr_scheduler.py
tianyu-l: Modified lr_scheduler. I think I've modified dataloader, validator, etc. to deprecate build_xxx function. Let me know if I missed anything.

torchtitan/models/common/attention.py
tianyu-l: added post_init check, including
```
            if self.attn_backend == "sdpa" and self.attn_mask_type == "block_causal":
                raise ValueError(
                    "attn_mask_type 'block_causal' is not supported with attn_backend 'sdpa'"
                )
```

torchtitan/models/common/rope.py
tianyu-l: added a TODO

tests/README.md
tianyu-l: good point, added

tianyu-l:
I believe I've addressed all comments from reviewers, except the following.

All numerics match except for gpt-oss debugmodel
- turning on deterministic causes loss to be NaN
- turning off deterministic, loss match on first step and become slightly different afterwards. I think it's caused by flex (in the torchtitan setup) doesn't have deterministic guarantee on backward.

CI failures are mostly also happening on main. The ones caused by this PR are only
- autoparallel, being fixed by @xmfan's https://github.com/meta-pytorch/autoparallel/pull/321.
- compiler toolkit, need @yiming0416 to take a look

I left a TODO to refactor RoPE. I can work with @shuhuayu on this after this PR is merged.

For other remaining work, check the "Remaining work" section of the PR summary.

cc @wwwjn @fegin @wconstab

torchtitan/models/flux/README.md
```diff
@@ -17,15 +17,14 @@ python scripts/download_hf_assets.py --repo_id black-forest-labs/FLUX.1-dev --ad
...
-./torchtitan/models/flux/run_train.sh
-
+MODEL=flux CONFIG=flux_debugmodel .run_train.sh
 ```
 
-If you want to train with other model args, run the following command:
+If you want to train with other configs, run the following command:
 ```bash
-CONFIG_FILE="./torchtitan/models/flux/train_configs/flux_schnell_model.toml" ./torchtitan/models/flux/run_train.sh
+MODEL=flux CONFIG=flux_schnell ./run_train.sh
```
wwwjn:
```suggestion
MODULE=flux CONFIG=flux_schnell ./run_train.sh
```

torchtitan/models/gpt_oss/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Quick Start
 ```bash
-CONFIG_FILE="./torchtitan/experiments/gpt_oss/train_configs/debug_model.toml" ./run_train.sh
+MODEL=gpt_oss CONFIG=gpt_oss_debugmodel ./run_train.sh
```
wwwjn:
```suggestion
MODULE=gpt_oss CONFIG=gpt_oss_debugmodel ./run_train.sh
```

PR #2388, opened 2026-02-17
fix(MoE): Apply grad_placements=Partial() to router to_local for TP

### LINKED ISSUE #2387: MoE router replication is broken w/ TP
### Bug description

The router replication is broken whenever TP is enabled in a MoE network, the weight diverges right after the first step among different TP ranks.
The issue is that the gradient coming from downstream the router is not being reduced. Changing outputs.to_local()
to outputs.to_local(grad_placements=(Partial(),)) in NoParallel fixes it. It's a similar type of bug as the one fixed here https://github.com/pytorch/torchtitan/commit/3ff755178cb65bd03d9c2766dba56ca45a04c8ef .

I can’t think of a worse class of bugs for MoEs to be honest, considering the router drives the majority of the weights. I’m curious if the recent proposal on DTensor erasure https://blog.ezyang.com/2026/02/dtensor-erasure/ would have caught it, or not really since this is mixed local/global SPMD scenario.

### Versions

Tested with torchtitan commit cf86edc977fb629fd8d93d1e0d1e2e8cb21a17d4 from a few days ago and looking at recent changes I don't see anything that could have fixed it.

fatih-uzlmz:
## Description
Fixes #2387 (MoE router replication broken w/ TP).

## The Issue
When Tensor Parallelism (TP) is enabled, the MoE router weights were diverging immediately after the first step. This was caused by `to_local()` being called on the router weights without specifying `grad_placements`. As a result, the backward pass was treating the local gradients as final gradients rather than partial sums, skipping the necessary AllReduce synchronization across ranks.

## The Fix
Added `grad_placements=(Partial(),)` to the `to_local()` conversion in `torchtitan/models/moe/moe.py`. This explicitly signals to the autograd engine that the resulting local tensor gradients must be reduced (summed) across the TP mesh before the optimizer step.

## Verification
- Validated that `Partial` is correctly imported from `torch.distributed._tensor`.
- Ran `tests/unit_tests/test_compile_moe.py` locally; tests passed successfully.
moefix

### DIFF
--- a/tests/unit_tests/test_moe_tp.py
+++ b/tests/unit_tests/test_moe_tp.py
@@ -3,59 +3,52 @@ import torch
 import torch.distributed as dist
 import torch.nn as nn
 from torch.distributed.device_mesh import init_device_mesh
-from torch.distributed._tensor import DTensor, Replicate
-from torchtitan.models.moe.moe import GroupedExperts
+from torch.distributed._tensor import DTensor, Replicate, Partial
 
-def test_moe_experts_grad_reduction():
-    # 1. Setup a fake distributed world (Simulated on CPU)
+from torchtitan.distributed.__init__ import NoParallel
+
+def test_noparallel_router_grad_reduction():
+    # 1. Setup Fake Distributed World
     os.environ["MASTER_ADDR"] = "localhost"
     os.environ["MASTER_PORT"] = "29500"
     
     if not dist.is_initialized():
         dist.init_process_group(backend="gloo", rank=0, world_size=1)
     
-    # 2. Define a dummy Mesh (TP=1 simulation)
     mesh = init_device_mesh("cpu", (1,))
     
-    # 3. Initialize GroupedExperts (This is where your fix lives!)
-    # We use use_grouped_mm=False to force the for-loop path or standard MM
-    # but the key is that we want to trigger the .to_local() call.
-    experts = GroupedExperts(
-        dim=16,
-        hidden_dim=32,
-        num_experts=4,
-        use_grouped_mm=False 
-    )
+    # 2. Initialize a dummy module (simulating the Router Gate)
+    class DummyRouterGate(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.gate = nn.Linear(16, 4, bias=False)
+            
+        def forward(self, x):
+            return self.gate(x)
+            
+    module = DummyRouterGate()
     
-    # 4. Mock the weights as DTensors
-    # This forces the code to hit the "if isinstance(self.w1, DTensor)" block in your fix
-    w1_local = experts.w1.detach()
-    w2_local = experts.w2.detach()
-    w3_local = experts.w3.detach()
+    # 3. Apply NoParallel wrapper (This triggers our new fix!)
+    parallel_style = NoParallel()
+    distributed_module = parallel_style._apply(module, mesh)
     
-    experts.w1 = nn.Parameter(DTensor.from_local(w1_local, mesh, [Replicate()]))
-    experts.w2 = nn.Parameter(DTensor.from_local(w2_local, mesh, [Replicate()]))
-    experts.w3 = nn.Parameter(DTensor.from_local(w3_local, mesh, [Replicate()]))
-
-    # 5. Run Forward Pass
-    # We need dummy inputs: x and num_tokens_per_expert
-    x = torch.randn(10, 16) # 10 tokens, dim 16
-    # Create a dummy distribution of tokens (e.g., 2 tokens for exp0, 3 for exp1, etc.)
-    num_tokens_per_expert = torch.tensor([2, 3, 2, 3], dtype=torch.long)
+    # 4. Run Forward Pass
+    x_local = torch.randn(10, 16)
+    x_dtensor = DTensor.from_local(x_local, mesh, [Replicate()])
     
-    # Run forward
-    output = experts(x, num_tokens_per_expert)
+    # The output will be a local tensor because use_local_output=True by default
+    output_local = distributed_module(x_dtensor)
     
-    # 6. Run Backward Pass
-    loss = output.sum()
+    # 5. Run Backward Pass
+    loss = output_local.sum()
     loss.backward()
     
-    # 7. Verification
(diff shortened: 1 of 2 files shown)


torchtitan/train.py
wconstab: accidental change?

wconstab (commented):
can you add a small test case showing the numerics issue / validating the fix?

torchtitan/train.py
fatih-uzlmz: Thanks for the catch :)

I accidentally branched off my PR history, just reverted train.py so it should be clean now,

And I am also currently working on the unit test to verify the gradients, will update you soon

fatih-uzlmz:
Alright, have gone ahead and added tests/unit_tests/test_moe_tp.py 

Basically it mocks a distributed environment and validates that gradients correctly flow through GroupedExperts weights when they are wrapped as DTensors. I then verified locally and it passes, here is a screenshot from my terminal


MoE_test_case

volcacius:
I opened the original issue here https://github.com/pytorch/torchtitan/issues/2387 this does not look like the right fix. It might work for TP but I don't think is correct for EP.

fatih-uzlmz:
Thanks @volcacius for the input,

You're right, I think my fix enforces `Partial` reduction, which is correct for replicated case (like the TP divergence we saw in #2387), but it'd probably be incorrect for EP where weights are Sharded

I think the robust fix is to make it conditional, so something like

`# Only enforce Partial grad placement if the weight is actually replicated`
`is_replicated = isinstance(self.w1.placements[0], Replicate)`
`grad_placement = (Partial(),) if is_replicated else None`


`w1 = self.w1.to_local(grad_placements=grad_placement)`
`# Repeat for w2, w3...`


Does that align with what you were thinking?

fatih-uzlmz:
cc: @volcacius @wconstab 

I went ahead and pushed a new commit with this conditional logic, 

the code now explicitly checks `isinstance(self.w1.placements[0], Replicate)` before applying `Partial()`, so this fixes the TP divergence issue while remaining completely safe for EP/sharded scenarios.

I also re ran our unit test with the new logic to ensure everything is still functioning properly and everything passes 


Screenshot 2026-02-19 at 5 12 09 PM

Let me know if there's anything else needed here 👍

torchtitan/models/moe/moe.py
```diff
@@ -153,11 +154,17 @@ def forward(
         if isinstance(self.w1, DTensor):
             # Convert parameters from DTensors to plain Tensors, to work with
             # dynamic-shape inputs in EP which cannot be easily expressed as DTensors.
-            w1 = self.w1.to_local()
+            
+            # Only enforce Partial grad placement if the weight is actually replicated (TP).
+            # If it's sharded (EP), we shouldn't reduce gradients across different experts.
+            is_replicated = isinstance(self.w1.placements[0], Replicate)
+            grad_placement = (Partial(),) if is_replicated else None
```
tianyu-l: I'm not sure if it fixes anything. In EP / TP, the weights are always sharded.
For activations, https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py#L57 fixes the issue.
I think there's one more issue here https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py#L285

tianyu-l (changes requested):

volcacius:
I don't think this works. It's not the expert weights that are replicated, it's the gate weights through NoParallel. If you look at my original issue I suggested a different fix.

fatih-uzlmz:
Okay I think I understand whats going on now, thank you guys both for the guidance.

First off my apologies, I accidentally applied the `Partial()` fix to the GroupedExperts weights (which @tianyu-l rightly pointed out are always sharded, making my fix useless).

@volcacius is right about the divergence reported in #2387 , its happening because the Router Gate weights are replicated, and they are losing their backward reduction when passing through the `NoParallel` wrapper during DTensor erasure.

I will revert my changes to moe.py and move the `grad_placements=(Partial(),) `fix directly into the `NoParallel` class where the gate outputs are localized.

@tianyu-l for the manual slice in expert_parallel.py L285—since that drops the DTensor backward communication, would you like me to include a fix for that in this PR as well, or should we isolate this PR strictly to the NoParallel router fix?

fatih-uzlmz:
I reverted the `moe.py` changes and moved the fix directly to the `NoParallel` wrapper in `torchtitan/distributed/__init__.py`

now the `_prepare_output_fn` conditionally applies `grad_placements=(Partial(),)` if the output layout is Replicate right before `to_local()` erases the DTensor.

I also rewrote the unit test to specifically target the `NoParallel` wrapper with a dummy router gate, and it successfully verifies that gradients now flow correctly through the erasure during the backward pass, here is a ss.


Screenshot 2026-02-19 at 10 24 11 PM

Let me know if this looks good!

tianyu-l:
@fatih-uzlmz 
> @volcacius is right about the divergence reported in https://github.com/pytorch/torchtitan/issues/2387 , its happening because the Router Gate weights are replicated, and they are losing their backward reduction when passing through the NoParallel wrapper during DTensor erasure.

Sorry it's not clear to me what the root cause is, and why it's not the one that I mentioned.
Having replicate weights with grad_placement=Replicate is fine, as long as what interacts with it is also Replicate (the TP activations). So I don't think the fix should be on router weights.

Could you include under what parallelism do you see the problem? The slicing one I mentioned would only occur during ETP=1 TP > 1 EP > 1. I would be surprised if it happens with TP = ETP.

fatih-uzlmz:
Yeah fair point,
`Replicate` x `Replicate` should yield identical gradients across ranks naturally, so the router weights shouldn't strictly need a Partial reduction to avoid divergence.

Since I picked this up from the issue tracker, I don't have the original reproduction config. 

@volcacius could you share your exact TP, EP, and ETP degrees?
If you were running ETP=1, TP>1, EP>1, then @tianyu-l  is probably right about that manual slice corrupting the downstream gradients.

fatih-uzlmz:
Looks like #2416 addresses the root cause via the activations.

Closing this out, thanks for the great discussion everyone :)

PR #2393, opened 2026-02-18
[RL] Prepare vllm definition for support_torch_compile compatibility

Lucaskabela:
## Summary

This PR makes the necessary model changes to support `support_torch_compile` for the vLLM model definition.  These are:

**For Graph Capture**
- Handle DTensor-to-local conversion in VLLMAttention with narrow() - this is a workaround for prim_to_local symbolic shape propogation causing issues under compile
- Remove PrepareModuleInputOutput from TP plan as VLLMAttention now handles DTensor conversion internally ( RowwiseParallel uses from_local(Shard(-1))) (related to the DTensor change above)
- Simplify forward() to be compile-friendly: remove conditionals, use _tp_enabled bool flag, and  reshape with -1
to avoid datadendent errors

**For CudaGraph Capture**
- Add weak_ref_tensor DTensor patches (FakeTensor kernel, sharding strategy, Python-level guard) for piecewise CUDA-graph capture in vLLM
- Pre-extend RoPE cache in __init__ instead of dynamic extension (to keep memory static for cudagraph)

See https://github.com/Lucaskabela/torchtitan/pull/4 for next step enablement (will resubmit against main after this Pr lands)

### Test Plan

```bash
```

### Baseline (losses/rewards)
| Step | Loss |
| --- | --- |
| 0 | -0.0196 |
| 1 | 7.4252 |
| 2 | 3.4577 |
| 3 | 1.2154 |
| 4 | 0.4386 |
| 5 | 0.2471 |
| 6 | 0.2507 |
| 7 | 0.2722 |
| 8 | 0.2140 |
| 9 | 0.1330 |

### This PR (losses/rewards)
| Step | Loss |
| --- | --- |
| 0 | -0.0196 |
| 1 | 7.4252 |
| 2 | 3.4577 |
| 3 | 1.2154 |
| 4 | 0.4386 |
| 5 | 0.2471 |
| 6 | 0.2507 |
| 7 | 0.2722 |
| 8 | 0.2140 |
| 9 | 0.1330 |

These match exactly showing we preserve training stability

### Metric comparisson
On top of https://github.com/pytorch/torchtitan/pull/2398 we get the following metrics: 

| Metric                | Baseline   | %     | After Changes   | %     |
|-----------------------|------------|-------|-----------------|-------|
| Total wall-clock      | 135.15s    |       | 133.62s         |       |
| Cumul. rollout        | 9.45s      | 7.0%  | 8.27s           | 6.2%  |
| Cumul. train          | 56.23s     | 41.6% | 49.94s          | 37.4% |
| Cumul. optimizer      | 0.31s      | 0.2%  | 0.30s           | 0.2%  |
| Cumul. weight_sync    | 69.47s     | 51.4% | 75.40s          | 56.4% |
| Peak mem (rollout)    | 4.09 GiB   |       | 4.09 GiB        |       |
| Peak mem (train)      | 9.58 GiB   |       | 9.58 GiB        |       |
| Peak mem (optimizer)  | 8.44 GiB   |       | 8.44 GiB        |       

of note - there is no significant change in overall runtime, nor memory usage

Authored with claude


torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -67,38 +69,58 @@ def forward(
...
+        v = v.reshape(-1, num_kv_heads, head_dim)
 
         # vLLM attention returns (num_tokens, hidden_size) where hidden_size = num_heads * head_dim
         output_flat = self.vllm_attn(q, k, v)
 
-        # Output is (batch * seq_len, num_heads * head_dim), reshape to (batch, seq_len, num_heads, head_dim)
-        output = output_flat.view(batch_size, seq_len, num_heads, head_dim)
+        # Reshape back: (batch, seq_len, num_heads, head_dim)
+        # Use view(1, -1, ...) since batch is always 1 in vLLM varlen format
+        output = output_flat.view(batch_size, -1, num_heads, head_dim)
```
felipemello1: if we know "seq_len", isnt it safer to use seq_len instead of -1?

torchtitan/experiments/rl/unified/infra/parallelize.py
```diff
@@ -120,16 +119,14 @@ def apply_non_moe_tp(
...
-                output_layouts=(Shard(1)),  # output
-                desired_output_layouts=(Shard(1)),
-                use_local_output=False,
-            ),
+            # VLLMAttention handles DTensor-to-local conversion internally.
+            # It captures the correct symbolic seq_len from the DTensor global
+            # shape before calling to_local(), then uses narrow() to fix the
+            # corrupted symbolic dimension that prim_to_local produces under
+            # torch.compile.  Its output is a plain tensor; RowwiseParallel on
+            # attention.wo wraps it back into a DTensor via from_local(Shard(-1)).
```
wwwjn, line 127: So the only change to make qwen3 model compile-able is change the attention part (let vllm.Attention to handle TP instead of apply TP ourselves)? What's the difference between using `PrepareModuleInputOutput` which inserts hooks behind the scene, and convert to DTensor manually?

torchtitan/experiments/rl/unified/models/utils.py
```diff
@@ -77,6 +77,7 @@ def replace_with_vllm_attention(model, tp_degree=1):
             head_dim=model_args.head_dim,
             layer_name=layer_name,
             scale=model_args.head_dim**-0.5,
+            tp_enabled=(tp_degree > 1),
```
wwwjn, line 80: This patch is only applied in Generator model. Do you only apply compile in generator, not trainer model now?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -11,33 +11,67 @@
...
 
 
 logger = init_logger(__name__)
 
+import vllm.utils.torch_utils as _vllm_torch_utils
+
+# ---------------------------------------------------------------------------
+# vLLM weak_ref_tensor + DTensor compatibility patches
+#
+# Piecewise CUDA-graph capture calls weak_ref_tensor() on every subgraph
```
wwwjn, line 41: What is weak_ref_tensor()? Can you point me to some docs?

torchtitan/experiments/rl/unified/models/utils.py
wwwjn: And if we only apply compile in generator side, not trainer side, would this make the log probability distribution in both side further different, once we further apply inductor?

Lucaskabela: Yes to clarify - this PR is strictly aimed at compiling the generator/rollout/vllm definition. The trainer is orthogonal since it won't use the same vllm compilation mechanism.

Inductor will impact numerics regardless of if we apply on one or both models (due to capturing different graphs), so we don't turn it on for now (that is a later question around where/how to apply inductor - a prerequisite is graph capturability which this PR stack will handle)

From summary, see https://github.com/Lucaskabela/torchtitan/pull/4 for the next step turning on vLLM compilation - this does not affect logits since we only use eager compile and cudagraphs

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: Sure - following the code pointer in this comment we lead to: 

https://github.com/vllm-project/vllm/blob/a0c70816956298f7dd1d0cf47cfa1a169a413692/vllm/compilation/cuda_graph.py#L21

This weak_ref_tensor is a utility used to record the tensor memory which is required to be kept alive for cudagraphs.  

https://github.com/vllm-project/vllm/blob/a0c70816956298f7dd1d0cf47cfa1a169a413692/vllm/utils/torch_utils.py#L649 is the definition

torchtitan/experiments/rl/unified/infra/parallelize.py
Lucaskabela: For compilation enablement, we also need to modify some of the forward code in vllm_wrapper to avoid data dependence where possible (this is mostly achieved through the unsqueeze changes)  

In regards to the TP change, we change dtensor conversion to be handled internally to vllm attention to fix an incorrect shape propogation causing runtime crashes (potentially a bug in dtensor + compile; I am working on getting a minimal repro)

For the other changes, these are needed for cudagraph enablement, and these are: 
1) Monkeypatching `weak_ref_tensors` to extract the memory from the DTensor in vllm_wrapper (needed for cudagraphs)
2) Statically allocating the rope_cache (since cudagraph requires memory be statically recorded)

torchtitan/experiments/rl/unified/models/utils.py
wwwjn: > The trainer is orthogonal since it won't use the same vllm compilation mechanism.

But since we use the same model definition, enabling compile should be the same on both side? Can you explain more what's the differences on how to enabling compile in vllm vs. non-vllm?

torchtitan/experiments/rl/unified/infra/parallelize.py
Lucaskabela: Re the difference: See [175690](https://github.com/pytorch/pytorch/issues/175690)

This is the repro, which I will have a potential PR fix for shortly.  In the event we don't get that landed though, the edits in this PR will also fix the symbolic shape failure

Lucaskabela: https://github.com/pytorch/pytorch/pull/175692 for potential fix (need feedback on this ofc)

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: Thanks, seems this change is needed because the model is DTensor-ified, and the weak_ref_tensor is assuming plain tensor 

Also curious, without weak_ref_tensor, does CUDAGraph in vllm still work? Or it's totally broken?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -11,33 +11,67 @@
...
+#    ("The specified pointer resides on host memory").
+# ---------------------------------------------------------------------------
+from torch.distributed.tensor._ops._pointwise_ops import pointwise_strategy
+from torch.distributed.tensor._ops.utils import register_op_strategy
+
+torch.library.register_fake(
+    "_C::weak_ref_tensor",
+    lambda tensor: torch.empty_like(tensor),
+)
+register_op_strategy(torch.ops._C.weak_ref_tensor.default)(pointwise_strategy)
```
wwwjn, line 62: Registering this is telling DTensor sharding propgation system that "weak_ref_tensor" is pointwise ops?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -67,38 +69,58 @@ def forward(
...
         k = k.transpose(1, 2)
         v = v.transpose(1, 2)
 
-        # TODO: reimplement as a 4d tensor once vLLM fix has landed
-        # Then flatten batch and seq_len: (batch * seq_len, num_heads, head_dim)
-        q = q.reshape(batch_size * seq_len, num_heads, head_dim)
-        k = k.reshape(batch_size * seq_len, num_kv_heads, head_dim)
-        v = v.reshape(batch_size * seq_len, num_kv_heads, head_dim)
+        # Flatten batch and seq_len: (batch * seq_len, num_heads, head_dim)
+        # Use -1 to avoid relying on symbolic shape from prim_to_local
```
wwwjn, line 113: Please educate me what is `prim_to_local`

wwwjn (commented):
Also from the PR description , seems enabling CUDAGraph + Compile front-end (Dynamo) didn't bring too much runtime / memory gain or burden. This might because of the benchmarking is not representative, eg the prompt is too short. Can you try how much speed up we can get from the a larger scale benchmark (eg, https://fburl.com/gdoc/mr91ct3v cudagraph path here)

I was thinking would user only compile the generator model, or the user would strictly prefer compile both trainer and generator model together. Theoritically this PR also helps enable trainer side compile as well, right?

torchtitan/experiments/rl/unified/models/utils.py
Lucaskabela: They are not the same definition since trainer also needs to have graph capturable backwards functions - these are the orthogonal components in reference. When we enable compile for the trainer model most changes will be related to that.

Outside of the fact the graphs we are capturing are different,  vLLM compile integration is quite different: the biggest difference is that it drops guards for inference speed at the cost of safety (causing issues such as the DTensor one we observe - if the guards were kept, when the symbolic dimension changes, it would trigger a recompilation)

See https://docs.vllm.ai/en/latest/design/torch_compile/ for more info on this.

Lucaskabela:
> Also from the PR description , seems enabling CUDAGraph + Compile front-end (Dynamo) didn't bring too much runtime / memory gain or burden. This might because of the benchmarking is not representative, eg the prompt is too short. Can you try how much speed up we can get from the a larger scale benchmark (eg, https://fburl.com/gdoc/mr91ct3v cudagraph path here)

I tried to make this clear from the summary, but I can reiterate here; there is significant speedup to the rollout engine (vLLM) going from 10s for the 10 steps to 2s.  It is listed in the followup PR which actually enables compile - this PR is the prep work to get us ready to turn on compile

Still, I will test the larger benchmark too

> I was thinking would user only compile the generator model, or the user would strictly prefer compile both trainer and generator model together. Theoritically this PR also helps enable trainer side compile as well, right?

Yes, but trainer has more work to be done (capture backward); this will also impact the cudagraph strategy but that is outside the scope of this PR; it is best to keep this PR singular in focus to the vLLM rollout engine

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: We were seeing crashes, but when I patched the fix for DTensor I no longer need this code so can clean it up probably

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: Yes that is correct - mostly for fake tensor prop which is needed for tracing

torchtitan/experiments/rl/unified/models/attention.py
Lucaskabela: prim_to_local is the function that backs DTensor.to_local() by extracting a local tensor shard from a distributed tensor (DTensor)

Lucaskabela:
Running vLLM TorchTitan without cudagraphs vs with cudagraphs using the benchmark script from above::

```bash
vLLM TorchTitan
--------------------------------------------------------------------------------
  Throughput (tokens/sec):
    Mean:       276.64 ± 5.97
    Median:     276.40
    Range:      268.52 - 283.93

  Latency (ms/token):
    Mean:         3.62 ± 0.08
    Median:       3.62
    Range:        3.52 - 3.72

  First Token Latency (ms):
    Mean:         0.00 ± 0.00

  Peak Memory (GB):
    Mean:         0.00 ± 0.00

  Total tokens: 10,240
  Runs: 5
```
vs 
```bash
vLLM TorchTitan
--------------------------------------------------------------------------------
  Throughput (tokens/sec):
    Mean:       955.85 ± 1.08
    Median:     955.49
    Range:      954.88 - 957.94

  Latency (ms/token):
    Mean:         1.05 ± 0.00
    Median:       1.05
    Range:        1.04 - 1.05

  First Token Latency (ms):
    Mean:         0.00 ± 0.00

  Peak Memory (GB):
    Mean:         0.00 ± 0.00

  Total tokens: 10,240
  Runs: 5
```

So nearly 4-5x speedup in the generator (consistent with previous findings).  Of course this won't be enabled directly until the next PR

torchtitan/experiments/rl/unified/infra/parallelize.py
acisseJZhong: @Lucaskabela curious to learn more about what's the bug if we keep the DTensor-to-local logic here, i.e. module forward pre hook, and why keeping it inside the attention module can fix? 

sorry for being dumb, read your fix PR but didn't have enough background to understand 🤔

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -67,38 +69,58 @@ def forward(
...
-        # with dimension names - see https://github.com/arogozhnikov/einops
-        batch_size, num_heads, seq_len, head_dim = q.shape
-        _, num_kv_heads, _, _ = k.shape
+        batch_size = q.shape[0]
+        num_heads = self.num_heads
+        num_kv_heads = self.num_kv_heads
+        head_dim = self.head_dim
+
+        if self._tp_enabled:
+            # Capture correct seq_len from DTensor global shape BEFORE
```
acisseJZhong, line 87: ahhh now I can understand more about the bug: so you are saying the seq_len was wrong if we capture it after to_local? if your PR https://github.com/pytorch/pytorch/pull/175692 fixed this issue, we no longer need this change?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -67,38 +69,58 @@ def forward(
         Forward pass using vLLM's Attention layer for inference.
 
         Args:
-            q: Query tensor [batch, num_heads, seq_len, head_dim]
-            k: Key tensor [batch, num_kv_heads, seq_len, head_dim]
-            v: Value tensor [batch, num_kv_heads, seq_len, head_dim]
+            q: Query tensor [batch, num_heads, seq_len, head_dim] (DTensor when TP enabled)
```
acisseJZhong, line 72: maybe add comment about what `tp_enabled` mean in this context, when this is enabled/why it is introduced? 

IIUC, this should be always enabled if we want to capture graph correctly?

torchtitan/experiments/rl/unified/models/attention.py
acisseJZhong, line 113: will this also get fixed once the symbolic shape is correct?

torchtitan/experiments/rl/unified/models/utils.py
acisseJZhong: for my own learnings sake, do you know where can I read more about the guards? and what guards we dropped for vllm inference? 

thanks for sharing the doc! but I am lacking some knowledge around guards, backed, unbacked, etc.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -11,33 +11,67 @@
...
+
+# ---------------------------------------------------------------------------
+# vLLM weak_ref_tensor + DTensor compatibility patches
+#
+# Piecewise CUDA-graph capture calls weak_ref_tensor() on every subgraph
+# output (see vllm/compilation/cuda_graph.py). When TP is active some of
+# those outputs are DTensors with Shard(1) placement.  Three things are
+# needed to make that work:
+#
+# 1. A FakeTensor ("Meta") kernel so torch.compile tracing can infer the
```
acisseJZhong, line 46: after reading this docstring, I am still unclear why we need to introduce weak_ref_tensor here, do you mind explain more?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -112,67 +139,37 @@ def __init__(
...
-        if required_len <= rope_cache.shape[0]:
-            return rope_cache
-
-        # If no extension function provided, return original cache
-        if self.rope_cache_extension_fn is None:
-            logger.warning(
-                f"RoPE cache extension needed (required_len={required_len}, "
-                f"current_len={rope_cache.shape[0]}) but no rope_cache_extension_fn provided. "
-                "Returning original cache."
+        # Store compile-friendly TP flag (bool attribute — no graph break)
```
acisseJZhong, line 142: can we still put that inside a private helper function and call the function here?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -259,7 +242,7 @@ def compute_logits(
         # When TP is applied, we return the full tensor (plain tensor) to vLLM engine
         # at the end of TorchTitanVLLMModelWrapper.forward().
         # We need to wrap the input from vLLM engine back to DTensor with Replicate() placement.
-        if self.parallel_dims.tp_enabled:
+        if self._tp_enabled:
```
acisseJZhong, line 245: actually why don't we just use `self.parallel_dims.tp_enabled`? without introducing a new param

torchtitan/experiments/rl/unified/models/utils.py
Lucaskabela: @acisseJZhong  vLLM drops most guards for inference (virtually all).

Guards in this context, are conditions we trigger recompilation on in Dynamo/PT2.  Since we have dropped them, this may lead to not recompiling and running an incorrect graph, which could cause a failure

See https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/torch.compiler_dynamo_overview.html#what-is-a-guard for more info on guards

Lucaskabela:
Closing in favor of #2486 post rebase

PR #2396, opened 2026-02-19
fix varlen op saving for simple fsdp

liangel-02:


rakkit:
btw it still broken here in unit test https://github.com/pytorch/torchtitan/blob/56999086d01c19c900ff86ef9a366f0ff7288ad7/tests/unit_tests/test_activation_checkpoint.py#L31

PR #2399, opened 2026-02-19
moe unbacked symint

pianpwk:
is this fine as a fix for https://github.com/pytorch/torchtitan/issues/2312

### DIFF
--- a/torchtitan/models/moe/moe.py
+++ b/torchtitan/models/moe/moe.py
@@ -83,6 +83,8 @@ def _run_experts_for_loop(
     num_tokens_per_expert_list = num_tokens_per_expert.tolist()
 
     # side-effect code due to the usage of generate_permute_indices
+    # torch._check informs the compiler that the slicing is in-bounds,
+    # and the output shape is exactly total_tokens.
     total_tokens = sum(num_tokens_per_expert_list)
     torch._check(total_tokens >= 0)
     torch._check(total_tokens <= x.shape[0])


torchtitan/models/moe/moe.py
```diff
@@ -86,12 +83,15 @@ def _run_experts_for_loop(
     num_tokens_per_expert_list = num_tokens_per_expert.tolist()
 
     # side-effect code due to the usage of generate_permute_indices
-    num_padding = x.shape[0] - sum(num_tokens_per_expert_list)
+    total_tokens = sum(num_tokens_per_expert_list)
+    torch._check(total_tokens >= 0)
+    torch._check(total_tokens <= x.shape[0])
```
tianyu-l, line 90: Could you share a bit more on
- what do they do?
- should we expect such checks everywhere in a program when working with torch.compile?
- it wasn't needed in the past, what regressed recently?

aditvenk: These are runtime checks; but during compile they constrain the symbolic value evaluation system which allows it to reason about expressions involving the unbacked symint. If the system cannot reason and there is some logic that requires it (e.g., a conditional) it will throw a data-dependent error (DDE). 

I don't think this is a regression -- this particular combination (SimpleFSD + DSv3 + compile + fallback loop-based MoE) most likely didn't work before. The loop-based implementation creates unbackend symint. 

cc @laithsakka  -- we should share reference to documentation that explains more about this and why/when user should do it. 

I find this on Google https://docs.pytorch.org/tutorials/intermediate/torch_export_tutorial.html#guards-torch-check

wconstab: the checks look ok to me, we could add a comment to the code explaining why the help compiler.

i don't actually know where the compiler hits a DDE today.  is it erroring on x[:total_tokens] when total_tokens is not known to be a valid range compared to x's shape? or somewhere deeper?

pianpwk: Ah, it seems FSDP2 compilation for MoE is skipped: https://github.com/pytorch/torchtitan/blob/9810191bbb0385d8ac91cd31abbc7f4156a0d462/torchtitan/models/llama4/parallelize.py#L645-L648

> i don't actually know where the compiler hits a DDE today

The slicing `x[:total_tokens]` doesn't know how exactly to compute the output size (is it in-bounds slicing or not?), and falls back to creating a new unknown (unbacked) output size for it, which sometimes has issues with inductor. If we inform it's inbounds, compilation is much easier.

wconstab: @tianyu-l LGTM, what do you think?

tianyu-l: The padding is added as a side-effect of the grouped gemm needing padding. Discussed with @pianpwk offline, and he will help follow up with https://github.com/pytorch/torchtitan/pull/2255 to remove padding for for-loop. With that I think we don't need this PR anymore.

laithsakka: @aditvenk  
https://docs.pytorch.org/docs/main/user_guide/torch_compiler/compile/dynamic_shapes_troubleshooting_guardon_errors.html#torch-check-cond-msg-fn

probably should update export documentation to refer to this.

torchtitan/models/moe/moe.py
```diff
@@ -86,12 +83,17 @@ def _run_experts_for_loop(
...
+    num_padding = x.shape[0] - total_tokens
 
     # a tuple of tensors indexed by experts
     # each with shape (tokens_per_expert(varying), dim)
     x_splits = torch.split(
-        x[: sum(num_tokens_per_expert_list)],
+        x[:total_tokens],
```
laithsakka, line 96: @ColinPeppler  is proposing 
https://github.com/pytorch/pytorch/pull/175819/changes#r2875182777
@pianpwk  or is it different issue.

PR #2405, opened 2026-02-20
[rl/unified] Update default `model-ckpt-path` in infer.py to the one from README

daniellepintz:
* Update default `model-ckpt-path` in infer.py to the one from README (`torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B/`)
* Update README to remove passing `--model-ckpt-path` to make life a little easier for users.

### DIFF
--- a/torchtitan/experiments/rl/unified/README.md
+++ b/torchtitan/experiments/rl/unified/README.md
@@ -31,19 +31,19 @@ uv pip install torch vllm xformers  --pre \
 **NOTE:** The pre-built vLLM wheels are only compatible with CUDA 12.8, though they should work with most older CUDA versions. Alternatively, you can install the corresponding vLLM pre-built wheels directly from https://download.pytorch.org/whl/nightly/cu128, for example: `uv pip install vllm-1.0.0.dev20260219+cu130-<suffix>.whl`. Ensure the build version number (e.g., `dev20260219`) matches your PyTorch nightly installation.
 
 
-3. Download Qwen/Qwen3-0.6B checkpoint from HuggingFace and put into `torchtitan/experiments/rl/example_checkpoint` folder.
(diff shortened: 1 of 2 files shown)


Lucaskabela (approved):
Definitely makes things easier when setting up :)

torchtitan/experiments/rl/unified/README.md
```diff
@@ -31,19 +31,19 @@ uv pip install torch vllm xformers  --pre \
...
-3. Download Qwen/Qwen3-0.6B checkpoint from HuggingFace and put into `torchtitan/experiments/rl/example_checkpoint` folder.
+3. Download `Qwen/Qwen3-0.6B` checkpoint from HuggingFace to `torchtitan/experiments/rl/example_checkpoint` folder.
 ```bash
 python scripts/download_hf_assets.py --repo_id Qwen/Qwen3-0.6B --local_dir torchtitan/experiments/rl/example_checkpoint --all --hf_token=...
 ```
 
 4. Run inference:
 ```bash
-python torchtitan/experiments/rl/unified/infer.py --model-ckpt-path <path_to_model_checkpoint>
+python torchtitan/experiments/rl/unified/infer.py
```
wwwjn, line 41: I explicit put the model-ckpt-path here to make the user define which model checkpoint he wants to use. Otherwise infer.py are too tighted to 0.6B model (although the model registry is specific to 0.6B model, which will be alleviate soon int the PR stack), and user might not realize which checkpoint path they are using. 

I would prefer leave `--model-ckpt-path` here. Other changes looks good

daniellepintz: Sure, added back in https://github.com/pytorch/torchtitan/pull/2424

PR #2406, opened 2026-02-20
[simple_fsdp] Support SimpleFSDP DSv3 + EP full/selective activation checkpointing.

IvanKobzarev:
Before that we required SimpleFSDP DSv3 to not use activation_checkpointing.
It resulted in illegal memory acceses because of differences in recomputed num tokens per expert in fw and bw.

1. Enabling them via must_save for all_to_all (for full we set it via graph_pass) 
For selective it is added explicitly.

3. Setting `torch._dynamo.config.skip_fwd_side_effects_in_bwd_under_checkpoint = True` to bypass mutation checks from dynamo (currently in MoE impl there is a field mutation)


aditvenk:
cc @yiming0416

torchtitan/experiments/simple_fsdp/reshard_after_forward.py
```diff
@@ -88,3 +88,19 @@ def force_recompute_node(node):
...
+    """Mark EP collectives as MUST_SAVE in the AC graph.
+
+    Full AC recomputes MoE routing, which can produce different token
+    assignments across ranks. Saving the all_to_all outputs prevents
+    cross-rank mismatches that cause device-side asserts in grouped_mm.
+    """
+    for node in gm.graph.nodes:
+        if (
+            node.op == "call_function"
+            and node.target == torch.ops._c10d_functional.all_to_all_single.default
```
wconstab: it feels bad to me to assume that "all_to_all" == EP routing, even though this is true in current models.  I feel like a better solution would apply some tag to the particular a2a in the model code saying "this is one that should be saved".  I think we can do this more easily if we have @soulitzer's proposal for named activation tensors and use those names for AC policy.

torchtitan/experiments/simple_fsdp/deepseek_v3/parallelize.py
```diff
@@ -183,6 +203,7 @@ def parallelize_deepseekv3(
     if compile_config.enable:
         torch._inductor.config.reorder_for_peak_memory = False
         torch._dynamo.config.capture_scalar_outputs = True
+        torch._dynamo.config.skip_fwd_side_effects_in_bwd_under_checkpoint = True
```
wconstab, line 213: what's the reason for this? can you add a comment above the config saying why its needed?

wconstab (commented):
seems ok to me, @tianyu-l i'll let you stamp

torchtitan/experiments/simple_fsdp/reshard_after_forward.py
IvanKobzarev: Yes, it is not generic and we it will be much better to put it in the model code. Will be a good usecase for new API.

torchtitan/experiments/simple_fsdp/deepseek_v3/parallelize.py
IvanKobzarev: in expert_parallel.py we mutate  `self.input_splits =  input_splits.tolist()` https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py#L134

By default this side-effect fails in dynamo:

```
  torch._dynamo.exc.Unsupported: HOP: Unsafe side effect
    Higher Order Operator: torch.utils.checkpoint.checkpoint
    Explanation: Mutating a variable from outside the scope of this HOP is not supported.
    Hint: If the HOP is activation checkpointing (torch.utils.checkpoint.checkpoint), this points to a side effect in forward method. Eager activation checkpointing replays that side-effect while recomputing the forward in the backward. If you are ok with side-effect not replayed in the backward, try setting `torch._dynamo.config.skip_fwd_side_effects_in_bwd_under_checkpoint = True`

    Developer debug context: Attempted to mutate UserDefinedObjectVariable(ExpertParallel)

   For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0067.html

  from user code:
     File "/data/users/ivankobzarev/g/torchtitan/torchtitan/models/deepseek_v3/model/model.py", line 502, in forward
    File "/data/users/ivankobzarev/g/pytorch/torch/distributed/algorithms/_checkpoint/checkpoint_wrapper.py", line 169, in forward
      return self.checkpoint_fn(  # type: ignore[misc]
    File "/data/users/ivankobzarev/g/torchtitan/torchtitan/models/deepseek_v3/model/model.py", line 393, in forward
      x = x + self.moe(self.ffn_norm(x))
    File "/data/users/ivankobzarev/g/torchtitan/torchtitan/models/moe/moe.py", line 524, in forward
      routed_output = self.experts(routed_input, num_tokens_per_expert)
    File "/data/users/ivankobzarev/g/pytorch/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
    File "/data/users/ivankobzarev/g/pytorch/torch/nn/modules/module.py", line 1882, in _call_impl
      return inner()
    File "/data/users/ivankobzarev/g/pytorch/torch/nn/modules/module.py", line 1822, in inner
      args_result = hook(self, args)
    File "/data/users/ivankobzarev/g/pytorch/torch/distributed/tensor/_api.py", line 1104, in <lambda>
      lambda mod, inputs: input_fn(mod, inputs, device_mesh)
    File "/data/users/ivankobzarev/g/torchtitan/torchtitan/distributed/expert_parallel.py", line 134, in _token_dispatch
      self.input_splits = input_splits.tolist()
```

torchtitan/experiments/simple_fsdp/deepseek_v3/parallelize.py
```diff
@@ -28,6 +28,21 @@
 
 from ..simple_fsdp import data_parallel, MixedPrecisionPolicy
 
+# Selective op AC: save compute-expensive ops (matmuls, attention kernels,
+# collectives). Cheap ops like MoE routing (softmax, topk, argsort) are
+# recomputed. See also ac_config.per_op_sac_force_recompute_mm_shapes_by_fqns
+# which defaults to ["moe.router.gate"] to force-recompute the gate mm.
+_op_sac_save_list = {
```
tianyu-l: can we reuse the list in torchtitan/models/deepseek_v3/parallelize.py?

torchtitan/experiments/simple_fsdp/deepseek_v3/parallelize.py
tianyu-l: +1 on adding a comment

torchtitan/experiments/simple_fsdp/README.md
```diff
@@ -41,7 +41,7 @@ Some of the features require the updates from PyTorch, with which we are working
 |Distributed Checkpointing| ✅ |
 |Float8 Training| 🚧 |
 |Expert Parallelism | ✅ |
-|Expert Parallelism + Activation Checkpointing| 🚧 |
+|Expert Parallelism + Activation Checkpointing| ✅ |
```
tianyu-l, line 44: please add op-based SAC to CI default config

torchtitan/experiments/simple_fsdp/reshard_after_forward.py
tianyu-l: Why do we need this pass, if SAC policy already marks all-to-all to be must save? Is it because it doesn't work with full AC? Then we should raise exception in full AC, instead of having this pass.

IvanKobzarev: Yes, this is only for non selective_ac_option="op" case. For "op" it works fine. Do you mean just to raise Error if "full" is specified for deepseek+moe allowing to use none or selective only? ( Applied ti this and removed pass for full)

torchtitan/experiments/simple_fsdp/deepseek_v3/parallelize.py
```diff
@@ -115,7 +119,25 @@ def parallelize_deepseekv3(
         )
 
     if ac_config.mode != "none":
-        apply_ac(model, ac_config)
+        if parallel_dims.ep_enabled:
+            # EP collectives (all_to_all) must be saved, not recomputed, because
+            # recomputing MoE routing can produce different token assignments
```
tianyu-l, line 124: hmm I'm a bit puzzled.

Do we know the root cause of this difference -- is it because gate / top-k / etc. computation is non-deterministic?

Why don't we have this issue for eager?

torchtitan/experiments/simple_fsdp/tests/integration_tests.py
```diff
@@ -304,6 +310,7 @@ def build_simple_fsdp_test_list() -> list[OverrideDefinitions]:
                     "--config simple_fsdp_deepseek_v3_debugmodel",
                     "--parallelism.data_parallel_shard_degree 2",
                     "--parallelism.context_parallel_degree 2",
+                    "--activation_checkpoint.mode selective",
```
tianyu-l, line 313: here it's missing "--activation_checkpoint.selective_ac_option op"? I would recommend editting the base config `simple_fsdp_deepseek_v3_debugmodel` so that you don't have to include the change on every single test

PR #2407, opened 2026-02-20
Add RL artifacts directories to `.gitignore`

daniellepintz:
These two directories (converted/ and models/) are created when running:
```
python torchtitan/experiments/rl/unified/simple_rl_multiprocess.py
```

We don't want to commit these so adding them to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,9 +43,9 @@ Sessionx.vim
 # env files
 .env
 
-# Local checkpoints/models
-converted/
-models/
+# Downloaded HF models and converted checkpoints (RL experiments)
+/models/
+/converted/
 
 # Vibe coding
 .claude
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -43,5 +43,9 @@ Sessionx.vim
 # env files
 .env
 
+# Local checkpoints/models
```
wwwjn: These are temporary folders which should be removed in later version. In short, we don't want to experiment folders to affect core torchtitan. `models` might accidentally ignore some other folders. I prefer delay this change

tianyu-l: https://stackoverflow.com/questions/6794717/git-ignore-certain-files-in-sub-directories-but-not-all

PR #2411, opened 2026-02-20
[BE] Apply pep585 and turn on flake8 rule so no regressions

Lucaskabela:
Since python>=3.10 is defined in the pyproject.toml, we can enable [pep585](https://peps.python.org/pep-0585/) (in python >=3.9)

Why? 

1) Fewer imports making code simpler
2) Standardization - prior to this change, we could have `List` and `list` intermixed in the codebase
3) Future proof for direction of python type system

### Test 
Ran 
```
 python -m flake8 --pep585-activation=always .
```
Explicitly

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,7 @@ repos:
           - flake8-bugbear == 22.4.25
           - pep8-naming == 0.12.1
           - torchfix
+          - flake8-pep585
         args: ['--config=.flake8']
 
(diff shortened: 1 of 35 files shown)


fegin:
Nice BE task, but you have to rebase after https://github.com/pytorch/torchtitan/pull/2386 is landed.

Lucaskabela:
Hi @fegin  - should be ready for review :)

tianyu-l (approved):

fegin (approved):

PR #2415, opened 2026-02-21
[BE] Apply pep604 and turn on flake8 rule so no regressions

Lucaskabela:
Since python>=3.10 is defined in the pyproject.toml, we can enable [pep604](https://peps.python.org/pep-0604/) (in python >=3.10)

Why? 

1) Fewer imports making code simpler
2) Standardization - prior to this change, we could have `Optional` or the  `| None` intermixed in the codebase
3) Future proof for direction of python type system

### Test 
Ran 
```
 python -m flake8 --pep604-activation=always .
```
Explicitly

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 # Suggested config from pytorch that we can adapt
-select = B,C,E,F,N,P,T4,W,B9,TOR0,TOR1,TOR2
+select = B,C,E,F,N,NU,P,T4,W,B9,TOR0,TOR1,TOR2
 max-line-length = 120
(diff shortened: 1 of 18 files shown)


tianyu-l (commented):
Thanks, this looks quite useful! Would prefer landing this after https://github.com/pytorch/torchtitan/pull/2386

Lucaskabela:
Review https://github.com/pytorch/torchtitan/pull/2411 first then will rebase and fix merge conflicts

tianyu-l (approved):

fegin (approved):

PR #1878, opened 2025-10-15
fix MoE TP backward

tianyu-l:
When TP is used for the matmul / grouped mm between input activations and routed experts, the same inputs are seeing different parameter values across TP ranks (inputs Replicate, param Shard) and thus the gradient flowing back should have `Partial` placement and need to be reduced. However, since we are using plain Tensor, instead of DTensor which can properly handle the gradient flow, we have to "manually" handle it. This PR fixes the issue by specifying the `grad_placements` field of `to_local`.

One can see that one more all-reduce is inserted in the backward every time it gets gradient from routed experts matmul.
before
image
after
image

PR #2416, opened 2026-02-21
[Bugfix] Force all reduce in bwd for router at MoE boundary 

acisseJZhong:
solve #2387 where user reported a bug that MoE router weights diff on each rank due to a missing all reduce under certain senarios:
- `self.score_before_experts=False` for both ETP=1 and ETP=TP
- `self.score_before_experts=True` and ETP=1 

similar fix was proposed #2388 by @fatih-uzlmz 
## Bug analysis 
**Case1:** 
When ETP=1 < TP, MoE is using TP2EP(`etp_mesh` is None, there is no ETP), `gate` is Replicated(via NoParallel), experts weight is Sharded(via EP mesh), MoE input is Replicated. 

_Root Cause:_ `ReordererSequenceParallell` is applied to split `top_scores` across TP ranks. Since each TP rank process its local slice of tokens through the experts, `d_routed_input` is Partial and there is no forced all-reduce. 

**Case2:** 
When ETP = TP, MoE is using DP2EP, `gate` is Replicated(via NoParallel), experts weight is Sharded(via ETP which force all reduce in bwd #1878), MoE input is Replicated on TP mesh. 

_Root Cause:_ The gradient flowing back to `gate.weight` is Partial when `self.score_before_experts=False`, which we incorrectly marked as Replicated and therefore results in wrong numerics. 

_Explanation:_ 
When `self.score_before_experts=True`, `d_routed_input` is already all reduced because of the fix #1878 we added to ETP. 
```
routed_input = routed_input * top_scores_experts_sorted
routed_output = self.experts(routed_input, ...) 
out_experts = routed_output_unsorted.sum(dim=1)    
```

When `self.score_before_experts=False`, gradient of `top_scores` from `bmm` is Partial. The previous all reduce fix only force all reduce on `d_routed_input`, but not `d_top_scores` which is wrong. 
```
routed_output = self.experts(routed_input, ...)
out_experts = bmm(top_scores, routed_output_unsorted) 
```

## The fix we propose that could fix both Case1 and 2
- We revert the fix in #1878 to keep Partial dx flowing back, and delay the all reduce later at the boundary of MoE. 
- We choose to not use [`ColWiseParallel/RowwiseParallel`](https://github.com/pytorch/pytorch/blob/e81980ea58c65a283cf01a7132e0f420834ecf10/torch/distributed/tensor/parallel/style.py#L181) for shared expert as they force all reduce dx; but we want to keep Partial dx. So we introduced `MoEColWiseParallel/MoERowwiseParallel` which removed the input and output hooks. See the [reasoning here](https://gist.github.com/acisseJZhong/b27ecd7eca9ca1f0c050282516de2e21) for more details. 
- We annotate gradient of router to be partial in bwd through adding a field to NoParallel(), so that DTensor will handle the all reduce 

image
image

**Caveat:** The graph is plotted from [Claude generated computational graph ](https://gist.github.com/acisseJZhong/7101c2542530365b7b06bdf1f78068cb), please take it with a grain of salt. 

cc @volcacius @fatih-uzlmz

### DIFF
--- a/torchtitan/distributed/tensor_parallel.py
+++ b/torchtitan/distributed/tensor_parallel.py
@@ -146,7 +146,7 @@ class ColwiseParallelWithGradPlacement(ColwiseParallel):
         if not isinstance(input_tensor, DTensor):
             assert local_input_grad_placements is not None, (
                 "local_input_grad_placements must be specified when input is a "
-                "plain tensor. Please think about what your want the from_local(Replicate) backward behavior like."
+                "plain tensor. Please think about what you want the from_local(Replicate) backward behavior like."
             )
             input_tensor = DTensor.from_local(
                 input_tensor,


torchtitan/distributed/expert_parallel.py
```diff
@@ -286,6 +286,14 @@ def _split_along_first_dim(x: torch.Tensor) -> torch.Tensor:
...
             return output
 
+        # NOTE: top_scores is Replicate across EP ranks, but the split below
+        #       gives each rank only its local slice. In backward, the gradient
+        #       of the slice is nonzero only for the local tokens (Partial).
+        #       We declare grad_placements=(Partial(),) to trigger an all-reduce
+        #       so the router gate receives the correct full gradient.
+        top_scores = DTensor.from_local(
+            top_scores, device_mesh, (Replicate(),)
+        ).to_local(grad_placements=(Partial(),))
```
wconstab: is doing 2 DTensor operations (to, from) really the best way here? why not just insert one autograd.function where the backwards is an all_reduce?

acisseJZhong: yeah I guess we could also insert a customized function where fwd=no_op and bwd does all_reduce? I was following a similar fix https://github.com/pytorch/torchtitan/pull/1878 
cc @tianyu-l if you know why we might prefer this over customized function

wconstab: ok. well short term it seems harmless to land this, other than it would have more overhead.  (worth confirming, but I assume it does not do any extra collectives or operations?)  anyway, we could fix them all to use a dedicated function from spmd_types once that's available.

acisseJZhong: it should not have any extra collectives other than the all reduce in bwd. yes I think we could directly use the function from spmd_types once it's ready!

fatih-uzlmz:
@acisseJZhong 
Thanks for the ping and for putting together the great deep dive, very fascinating to see how the bug split into two distinct cases based on the ETP/TP config

I'm glad @volcacius and I were on the right track with the NoParallel wrapper being the correct conceptual boundary for the reduction, but your fix handling the activations directly is a really clean solution.

Really happy to have helped kick off this investigation and grateful for the opportunity to learn from the team here :)

Thanks everyone once again

acisseJZhong:
@fatih-uzlmz thanks for the kind words! Turns out your fix was definitely on the right direction! 

Please feel free to help review the PR, and also test on your end to verify the bug goes away, and let us know if you have doubts. I used Claude extensively(e.g. [claude thinking](https://gist.github.com/acisseJZhong/b27ecd7eca9ca1f0c050282516de2e21)) for this bug fix so might be missing something lol

Although now each rank router weights are the same, there's actually no strong guarantee that MoE numerics are correct? Going forward, we plan to introduce [spmd_types ](https://github.com/ezyang/spmd_types/blob/main/sixlib/spmd_types/DESIGN.md) as a type checker for DTensor to guarantee correctness, and hopefully this kind of bugs could surface quicker.

fatih-uzlmz:
@acisseJZhong haha yeah, it's honestly super validating to hear that! it's also pretty cool to know even the team relies on Claude's thinking mode for this kind of heavy DTensor math

I unfortunately don't have a multi-GPU hardware setup on my end to run the physical reproduction tests right now but the logic in the PR looks very solid to me.

and introducing `spmd_types` as a type checker to catch these boundary issues sounds like a massive quality-of-life improvement for the whole ecosystem. Excited for that!

torchtitan/models/common/moe/moe.py
```diff
@@ -478,6 +488,12 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         Returns:
             out (torch.Tensor): Output tensor with shape ``(bs, slen, dim)``.
         """
+        # Convert DTensor to local tensor for MoE-internal computation.
+        # grad_placements=(Partial(),) ensures x.grad is Partial in backward,
+        # so gradient reduction happens once at the MoE boundary (reduce-scatter)
+        # rather than being duplicated inside the MoE.
+        if isinstance(x, DTensor):
+            x = x.to_local(grad_placements=(Partial(),))
```
tianyu-l, line 478: You are assuming
- 1D mesh
- The mesh is TP

At least do assertions on them.

torchtitan/models/moe/moe.py
```diff
@@ -65,6 +65,16 @@ def __init__(
         self.w3 = nn.Linear(dim, hidden_dim, bias=False)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if getattr(self.w1, "_moe_no_hooks", False):
```
tianyu-l: hmm what's going on here

torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -497,6 +497,61 @@ def apply_fsdp(
...
+    def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
+        if isinstance(module, nn.Linear):
+            partition_fn = self._partition_linear_fn
+        elif isinstance(module, nn.Embedding):
+            partition_fn = self._partition_embedding_fn
+        else:
+            raise NotImplementedError(
+                "MoERowwiseParallel currently only supports nn.Linear and nn.Embedding!"
+            )
+        module._moe_no_hooks = True
```
tianyu-l: why do you need this field?

torchtitan/models/llama4/infra/parallelize.py
```diff
@@ -497,6 +497,61 @@ def apply_fsdp(
             transformer_block.set_modules_to_backward_prefetch([model.tok_embeddings])
 
 
+class MoEColwiseParallel(ColwiseParallel):
```
tianyu-l: Name is bad. Does it have anything to do with MoE?

tianyu-l: Put them in distributed/tensor_parallel.py instead of here

torchtitan/distributed/__init__.py
```diff
@@ -34,12 +40,14 @@ def __init__(
         input_layout: Placement | None = None,
         output_layout: Placement | None = None,
         use_local_output: bool = True,
+        output_grad_as_partial: bool = False,
```
tianyu-l: This sounds too restrictive. Maybe merge use_local_output and this field as `local_output_grad_placements`. If None, then don't cast to local.

In some sense, we would require user to think about grad placement when casting to local, as there's no guarantee on correctness automatically.

tianyu-l: also need to audit all callsites
https://github.com/search?q=repo%3Apytorch%2Ftorchtitan%20NoParallel&type=code

The qwen3 one seems correct as it's not using local outputs.

tianyu-l (changes requested):

torchtitan/models/moe/moe.py
acisseJZhong: this part is where I am not very certain of 🤣 

Seems like this changes inside `FeedFoward` forward is necessary: because we remove the input and output hooks for Col and RowwiseParallel, and MoE input is Partial plain tensor because L496, so we need to convert the shared expert weight to plain tensor to avoid DTensor wrapping x as DTensor(Replicate). 

However, `FeedForward` is used by both shared expert and dense FFN layers, and I only want this fix to apply for the first case. Therefore I added the check: to decide if I should do `F.linear` or `self.w1`

torchtitan/models/llama4/infra/parallelize.py
acisseJZhong: answered above

torchtitan/models/llama4/infra/parallelize.py
acisseJZhong: good point, it's parallel to MoE. Maybe rename it to `ColwisePartitionOnly` which indicates that it only partition the weights/no hooks?

torchtitan/distributed/__init__.py
```diff
@@ -68,15 +72,17 @@ def _prepare_input_fn(
     @staticmethod
     def _prepare_output_fn(
         output_layout: Placement,
-        use_local_output: bool,
+        local_output_grad_placements: Sequence[Placement] | None,
```
tianyu-l:
```suggestion
        local_output_grad_placements: tuple[Placement] | None,
```

torchtitan/models/llama4/parallelize.py
```diff
@@ -555,15 +558,22 @@ def apply_moe_ep_tp(
...
+                # Use grad_placements to keep d_x as Partial in backward
+                # (avoids the all-reduce that from_local(Replicate) would
+                # otherwise trigger). The reduction happens once at the MoE
+                # output boundary (PrepareModuleInputOutput).
                 # pyrefly: ignore [no-matching-overload]
                 moe_layer_plan.update(
                     {
-                        "moe.shared_experts.w1": ColwiseParallel(),
+                        "moe.shared_experts.w1": ColwiseParallel(
+                            grad_placements=(Partial(),),
```
tianyu-l: wait did you add this support somewhere?
I didn't see it in https://github.com/pytorch/torchtitan/pull/2416

I don't mind us creating ad hoc ParallelStyle in torchtitan/distributed/tensor_parallel.py

acisseJZhong: oops I literally was tying git push command and we went for lunch 😆  

I added to ColwiseParallel https://github.com/pytorch/pytorch/pull/175867/changes#diff-b4e400af23b02b381f3f4ab90eb0bc0ae2691b63d2b587257c76d22848dd33bbL108, if you think it's too much change in pytorch, I am okay with creating a new Style in Torchtitan as well.

torchtitan/distributed/__init__.py
acisseJZhong: I was following https://github.com/acisseJZhong/pytorch/blob/3b592439e657e777709b8074cb156aa22a5f71e7/torch/distributed/tensor/_api.py#L415?

torchtitan/models/llama4/parallelize.py
tianyu-l: I think we can temporarily put it in torchtitan, given that we are going to migrate to spmd anyway.
The necessity to change ParallelStyle in core is not as much as that for changing from_local.

torchtitan/distributed/__init__.py
tianyu-l, line 1: Is it fair to say `NoParallel` is only used on TP mesh? If so, let's move it to tensor_parallel.py.

torchtitan/models/common/moe/moe.py
```diff
@@ -453,6 +453,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         Returns:
             out (torch.Tensor): Output tensor with shape ``(bs, slen, dim)``.
         """
+        # Convert DTensor to local tensor for MoE-internal computation.
```
tianyu-l, line 456: mention that in all cases [TP only, TP + EP and ETP=1, TP + EP and ETP=TP] x [router.gate, routed experts, shared experts] x [score before vs. after experts], the gradient that's flowing back is always Partial. That's why we are converting it to Partial here.

torchtitan/models/common/moe/moe.py
tianyu-l, line 1: should also update moe file, including gpt-oss/moe.py and moe_deepep.py

torchtitan/distributed/__init__.py
acisseJZhong: yeah I think so, since the default desired input and output layout are all Replicate(1 dimensional). It's fair to claim this. All the current usages of NoParallel also always apply on TP mesh.

torchtitan/models/common/moe/moe.py
acisseJZhong: GptOssMoE inherits MoE here. Will update moe_deepep

torchtitan/models/common/moe/moe.py
acisseJZhong: updated, please lmk if i misunderstand anything.

torchtitan/models/common/moe/moe_deepep.py
```diff
@@ -51,6 +52,29 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        # Why grad(x) is Partial on the tp_mesh across all parallelism:
+        # - TP only / TP+EP with ETP=TP: TP-sharded expert weights (Colwise on
+        #   w1/w3, Rowwise on w2) produce Partial output gradients.
+        # - TP+EP with ETP=1: each TP rank processes a disjoint token subset
+        #   (via ReordererSequenceParallel), so grad(x) is non-zero only at
+        #   each rank's token positions(Partial).
+        #
+        # This holds for all MoE components (router.gate, routed experts, shared
+        # experts) and regardless of score_before_experts.
+        if isinstance(x, DTensor):
```
acisseJZhong, line 70: this should be correct for deepep moe as it also has Reorderer when ETP=1

torchtitan/distributed/__init__.py
```diff
@@ -4,97 +4,10 @@
...
-from torch.distributed.tensor import DeviceMesh, distribute_module, DTensor, Replicate
-from torch.distributed.tensor.parallel import ParallelStyle
-from torch.distributed.tensor.placement_types import Placement
-
 from torchtitan.distributed.parallel_dims import ParallelDims
+from torchtitan.distributed.tensor_parallel import NoParallel
 
 __all__ = [
     "ParallelDims",
     "NoParallel",
```
tianyu-l: NoParallel is not very special. Let's not add this shortcut. Instead, imports need to be changed to `from ...tensor_parallel`

torchtitan/distributed/tensor_parallel.py
```diff
@@ -5,14 +5,185 @@
...
+            ),
+            partial(
+                self._prepare_output_fn,  # pyrefly: ignore [bad-argument-type]
+                self.output_layout,
+                self.local_output_grad_placements,
+            ),
+        )
+
+
+class ColwiseParallelWithGradPlacement(ColwiseParallel):
```
tianyu-l, line 109: Maybe should change all torchtitan to use these two, as they are strictly safer. Don't need to do in this PR.

torchtitan/distributed/tensor_parallel.py
```diff
@@ -5,14 +5,185 @@
...
+        device_mesh,
+    ):
+        input_tensor = inputs[0]
+        if not isinstance(input_tensor, DTensor):
+            input_tensor = DTensor.from_local(
+                input_tensor,
+                device_mesh,
+                input_layouts,
+                run_check=False,
+                grad_placements=local_input_grad_placements,
```
tianyu-l, line 156: I think we should enforce this field to be not None if input is plain tensor.

acisseJZhong: sure, it's not enforced in pytorch from_local, but doesn't hurt to add a check in titan to force user think.

PR #2418, opened 2026-02-21
feat(profiling): Add compute-communication overlap diagnostic for FSDP2

### LINKED ISSUE #2408: [RFC] Feature: Automated Compute-Communication Overlap Diagnostic in Profiler
## [RFC] Feature: Automated Compute-Communication Overlap Diagnostic in Profiler

**Status:** Proposed **Target Repo:** `pytorch/torchtitan` **Suggested Tags:** `feature-request`, `performance`, `experimental`

### 1\. Motivation

In large-scale training (e.g., Llama 3.1 70B/405B), achieving high Model FLOPs Utilization (MFU) requires a perfect orchestration of compute and communication overlap. In FSDP2/3D parallel setups, identifying "Communication Bubbles"---where the main compute stream stalls while waiting for `all_gather` or `reduce_scatter` kernels---is currently a manual, labor-intensive process.

Engineers must download multi-gigabyte `.json` trace files and visually inspect CUDA streams in the Chrome Trace Viewer to see if NCCL operations are effectively "hidden" under GEMM kernels.

### 2\. Proposed Feature

I propose adding a lightweight **`OverlapAnalyzer`** utility that automatically quantifies this overlap. This tool hooks into the `on_trace_ready` callback of the PyTorch profiler and outputs a high-level diagnostic summary directly to the logs.

**Key Metrics:**

-   **Total Compute Time:** Sum of CUDA time for math-heavy kernels (GEMM, cuBLAS, aten).

-   **Total NCCL Time:** Sum of CUDA time for communication kernels (`all_gather`, `reduce_scatter`).

-   **Overlap Efficiency %:** A heuristic calculating how much of the network latency was successfully overlapped with computation.

### 3\. Design & API Contract

To maintain `torchtitan`'s "clean-room" philosophy and minimize overhead, I suggest the following implementation details:

-   **Location:** House the core logic in `torchtitan/experiments/overlap_analysis/` or as a utility in `torchtitan/components/profiler_utils.py`.

-   **Config Toggle:** Add an experimental flag to the `[profiling]` section of the `.toml` config:

    Ini, TOML

    ```
    [profiling]
    enable_profiling = true
    enable_overlap_analysis = true  # New experimental flag

    ```

-   **No "Hot-Path" Impact:** The analysis only runs inside the `on_trace_ready` callback, meaning it has zero impact on actual training performance steps.

**Alternative (Very Low Risk)** I can house this in `torchtitan/experiments/overlap_analysis` to keep it decoupled from the core trainer

### 4\. Expected Impact

This provides an immediate "Network Health" signal. For example, if a user sees an **Efficiency of < 50%**, it's a clear signal to investigate their interconnect (NVLink vs. PCIe) or adjust their sharding strategy, without needing to open a single trace file.

pbpatre:
## Summary

Adds an automated compute-communication overlap diagnostic to help identify communication bottlenecks in FSDP2 training. Implements #2408

## Motivation

Achieving high MFU in large-scale training requires effective overlap of compute and NCCL communication. Currently, identifying "communication bubbles" requires manually downloading multi-GB trace files and inspecting them in Chrome Trace Viewer. This feature provides an immediate diagnostic signal directly in the logs.

### New Feature: `CommsComputeOverlapAnalyzer`

- Analyzes profiler traces to compute overlap efficiency between compute kernels and NCCL operations
- Reports Total Compute Time, Total NCCL Time, Total Active Time, Overlap Efficiency, and GPU time unutilized due to NCCL comms
- Enabled via `--profiler.enable_overlap_analysis` flag (off by default)

### Implementation Details

- Uses `prof.events()` to collect raw CUDA kernel events and their wall-clock time intervals
- Classifies kernels by name: `nccl` for communication; `gemm`, `aten`, `cublas`, `cudnn`, `cutlass`, `triton`, `flash` for compute
- Computes the union of time intervals per category using interval merging, correctly deduplicating concurrent kernels across CUDA streams
- Overlap is derived as `compute_us + comm_us - active_us`, where `active_us` is the union across both categories

## Example Output

**Llama 3 8B (8B params) on 2x H200:**
```
[CommsComputeOverlapAnalyzer] Compute-Communication Overlap Report
  Total Compute Time               : 165.12 ms
  Total NCCL Time                  : 255.94 ms
  Total Active Time                : 386.52 ms
  Overlap Efficiency               : 13.5 %
  GPU Unutilized due to NCCL comms : 86.5 %
```

**Debug model (6M params) on 2x H200:**
```
[CommsComputeOverlapAnalyzer] Compute-Communication Overlap Report
  Total Compute Time               : 4.80 ms
  Total NCCL Time                  : 19.97 ms
  Total Active Time                : 23.83 ms
  Overlap Efficiency               : 4.7 %
  GPU Unutilized due to NCCL comms : 95.3 %
```

Both results are expected communication-bound baselines with no overlap optimizations enabled. The debug model is more severely bottlenecked because its compute-to-communication ratio (0.24x) is far lower than the 8B model (0.65x) — there is almost no compute to overlap with.

## Testing

Tested on 2x NVIDIA H200 with:

- Llama 3 8B:
  ```
  NGPU=2 ./run_train.sh --module llama3 --config llama3_8b \
    --hf_assets_path=./tests/assets/tokenizer \
    --profiler.enable_profiling --profiler.profile_freq=10 \
    --profiler.enable_overlap_analysis --training.steps=20 \
    2>&1 | tee 8B_output.log
  ```
- Debug model:
  ```
  NGPU=2 ./run_train.sh --module llama3 --config llama3_debugmodel \
    --profiler.enable_profiling --profiler.profile_freq=10 \
    --profiler.enable_overlap_analysis --training.steps=20 \
    2>&1 | tee 6M_output.log
  ```

## Limitations

- Only CUDA kernels are considered; CPU-side work is excluded
- Kernels not matching the classify patterns (e.g. custom ops) are excluded from both categories and do not affect the reported overlap percentage


torchtitan/tools/profiling.py
```diff
@@ -18,6 +18,101 @@
...
+        trace_duration_us = self._get_trace_duration_us()
+        if trace_duration_us == 0.0:
+            trace_duration_us = compute_us + comm_us
+
+        trace_duration_us = max(trace_duration_us, compute_us, comm_us)
+
+        raw_overlap = compute_us + comm_us - trace_duration_us
+        overlap_pct = max(0.0, min(raw_overlap / comm_us * 100.0, 100.0))
+
+        status = "OPTIMAL" if overlap_pct >= 50.0 else "COMMUNICATION BOUND"
```
wconstab: nit: not a very useful 'status' message, why is >=50 "optimal"? why not just report the efficiency and leave it at that

pbpatre: @wconstab  Agreed. The status does not add much value after the overlap percentage. Instead I added a GPU unutilized metric (100-overlap_pct) which is what we want the users to notice and then they can decide the further course of action.

wconstab:
wonder what others think about adding this to torchtitan. seems nice, but also, its specific to FSDP2 and not sure if it belongs in torchtitan. cc @fegin @tianyu-l @wwwjn

tianyu-l:
@wconstab 
> its specific to FSDP2

What part of the code is specific to FSDP2? It looks pretty generic to me.

wconstab:
yes, you're right. it is actually specific to nccl kernel names, but was tested for FSDP2, maybe works in general.

tianyu-l (commented):
I feel this could be useful, and the code complexity doesn't seem to be crazy.

Meanwhile I would love to see refactor of
- Profiling file into "config build style" following recent refactor, so `ProfilingConfig` becomes `Profiler.Config`
- provide a generic (list of) profile analyzer which this `OverlapAnalyzer` is the first example

Let me know if this makes sense or not.

pbpatre:
@tianyu-l I agree that we can do the quick refractoring for config build style as part of this change. I have updated the PR with this change and added test cases. Let me know if it looks good 

> Meanwhile I would love to see refactor of
> 
> * Profiling file into "config build style" following recent refactor, so `ProfilingConfig` becomes `Profiler.Config`
> * provide a generic (list of) profile analyzer which this `OverlapAnalyzer` is the first example
>

torchtitan/trainer.py
```diff
@@ -74,7 +70,7 @@ class Config(Configurable.Config):
         dump_folder: str = "./outputs"
         """Folder to dump job outputs"""
 
-        profiling: ProfilingConfig = field(default_factory=ProfilingConfig)
+        profiling: Profiler.Config = field(default_factory=Profiler.Config)
```
tianyu-l: renaming to be consistent
```suggestion
        profiler: Profiler.Config = field(default_factory=Profiler.Config)
```

torchtitan/trainer.py
```diff
@@ -789,14 +785,13 @@ def train(self):
         self.checkpointer.load(step=config.checkpoint.load_step)
         logger.info(f"Training starts at step {self.step + 1}")
 
+        profiler = Profiler(config.profiling)
```
tianyu-l: 1. Let's put this in Trainer instance variable list, and build it during constructing.
2. Use config.profiler.build() instead of calling constructor directly

torchtitan/tools/profiling.py
```diff
@@ -8,69 +8,236 @@
...
+        profiling is enabled and the process group is initialized (distributed run).
+        """
 
-    if enable_profiling:
-        trace_dir = os.path.join(base_folder, profiling_config.save_traces_folder)
+    def __init__(
+        self,
+        config: "Profiler.Config",
+        *,
+        analyzers: list[ProfileAnalyzer] | None = None,
```
tianyu-l: Let's not put this as part of constructor.

The consistent way to do things is to make OverlapAnalyzer configurable, and have a list of `ProfileAnalyzer.Config` in `Profiler.Config`.

We don't have to do it today -- we can just have a singleton self.analyzers list for now.

torchtitan/tools/profiling.py
```diff
@@ -8,69 +8,236 @@
...
+
+        enable_memory_snapshot: bool = False
+        """Whether to dump memory snapshot."""
 
-    save_memory_snapshot_folder: str = "memory_snapshot"
-    """Memory snapshot files location"""
+        save_memory_snapshot_folder: str = "memory_snapshot"
+        """Memory snapshot files location."""
 
+        experimental_diagnostics: bool = False
```
tianyu-l:
```suggestion
        enable_overlap_analysis: bool = False
```

torchtitan/tools/profiling.py
```diff
@@ -8,69 +8,236 @@
...
+    ) -> None:
+        self._config = config
+        self._analyzers: list[ProfileAnalyzer] = []
+        if config.experimental_diagnostics:
+            self._analyzers.append(OverlapAnalyzer())
+        if analyzers:
+            self._analyzers.extend(analyzers)
+
+    @contextlib.contextmanager
+    def enable_profiling(
```
tianyu-l: maybe keep "maybe"? I don't have strong opinion here.
```suggestion
    def maybe_enable_profiling(
```

torchtitan/trainer.py
```diff
@@ -789,14 +785,13 @@ def train(self):
...
         with (
-            maybe_enable_profiling(
-                config.profiling,
+            profiler.enable_profiling(
                 global_step=self.step,
                 base_folder=config.dump_folder,
             ) as torch_profiler,
-            maybe_enable_memory_snapshot(
-                config.profiling,
+            profiler.enable_memory_snapshot(
```
tianyu-l: Instead of calling profiler.enable_profiling and profiler.enable_memory_snapshot, I think we should build the profiler as a container, similar to the `LoggerContainer` here https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/metrics.py#L178

If this is too much, feel free to leave a TODO here.

tianyu-l (commented):
Thanks, left some comments.

torchtitan/trainer.py
pbpatre: Fixed here and few other usages

torchtitan/trainer.py
pbpatre: Makes sense. Fixed this and other usages

torchtitan/tools/profiling.py
pbpatre: I have created a list for now and controlled this via a boolean flag. I have kept the support list[ProfileAnalyzer.Config] in Profiler.Config as TODO for now.
Creating profile analyzers via config leads to cleaner code. In addition to this, having a way for users to control this experimental analysis during execution is helpful too, say via say a CLI flag.

What do you think? We can do this in a followup PR but the current configurable list should be good for now.

torchtitan/tools/profiling.py
pbpatre: Renamed it since we have only one optional analysis for now

torchtitan/trainer.py
pbpatre: Fixed this so that trainers are simplified

pbpatre:
@tianyu-l Resolved comments with one TODO

torchtitan/tools/profiling.py
```diff
@@ -8,69 +8,286 @@
...
         )
 
         rank = torch.distributed.get_rank()
+        run_diagnostics = (
+            bool(self._analyzers)
+            and torch.distributed.is_initialized()
+            and rank == 0
+        )
+
+        analyzers = self._analyzers
```
tianyu-l: remove this alias, since it's only used once.

torchtitan/tools/profiling.py
```diff
@@ -8,69 +8,286 @@
...
+    def analyze(self, prof: torch.profiler.profile) -> None:
+        """Analyze a completed profiler trace.
+
+        Args:
+            prof: A ``torch.profiler.profile`` object with collected trace data.
+        """
+        ...
+
+
+class OverlapAnalyzer(ProfileAnalyzer):
```
tianyu-l: Overlap is a bit too generic. Maybe `CommOverlapAnalyzer` because your percentage is computed as (overlap / comm time)

torchtitan/tools/profiling.py
```diff
@@ -8,69 +8,286 @@
...
+        if self._mem is not None:
+            self._mem.step()
+
+
+class Profiler(Configurable):
+    """Owns profiling and memory snapshot lifecycle for a training run.
+
+    Instantiate via ``Profiler.Config.build()`` and use the context-manager
+    methods ``maybe_enable_profiling`` and ``maybe_enable_memory_snapshot``
+    in the training loop.
```
tianyu-l: I think this solution is over complicated then necessary, as you are trying to reuse the existing context managers, which themselves are already nested.

We should try to flatten everything:
- Profiler itself is a context manager, which defines `__enter__`,` __exit__`, and `step()` methods.
- torch profiler and memory profiler should be direct children, aka `self.torch_profiler` and `self.memory_profiler`.
- Profiler has `build_torch_profiler` and `build_memory_profiler` functions which return started profiler which we can do `step` on. (Here we don't have to use config build style, as they shouldn't change much.)

WDYT? Let me know if you hit difficulties.

tianyu-l (changes requested):

torchtitan/tools/profiling.py
pbpatre: Renamed to CommsComputeOverlapAnalyzer

torchtitan/tools/profiling.py
pbpatre: Makes sense. Flattened the structure.

torchtitan/tools/profiling.py
```diff
@@ -4,73 +4,341 @@
...
+        export to report compute-communication overlap efficiency. This is only
+        active when profiling is enabled and the process group is initialized
+        (distributed run).
+        """
 
-    save_memory_snapshot_folder: str = "memory_snapshot"
-    """Memory snapshot files location"""
+    def __init__(self, config: "Profiler.Config") -> None:
+        self._config = config
+        self._analyzers: list[ProfileAnalyzer] = []
```
acisseJZhong: wondering why this is a list? what other kind of ProfileAnalyzer do we potentially have?

acisseJZhong:
Thanks for adding this! 

Regarding limitations:
> When multiple kernels run concurrently on different CUDA streams, actual overlap may be higher. 

Should actual overlap be lower? as we count the compute stream overlap as compute comm overlap. One way to solve this is to union different streams before the overlap calculation? 

if we use this formula: `Computes overlap: (compute + nccl - trace_duration) / nccl * 100`, another limitation might be with gpu idle time, trace_duration > compute + comm, so the estimated overlap might be lower.

torchtitan/tools/profiling.py
pbpatre: The list is to provide extensibility in adding more analyzers in the future like GPU utilization etc

pbpatre:
> Thanks for adding this!
> 
> Regarding limitations:
> 
> > When multiple kernels run concurrently on different CUDA streams, actual overlap may be higher.
> 
> Should actual overlap be lower? as we count the compute stream overlap as compute comm overlap. One way to solve this is to union different streams before the overlap calculation?
> 
> if we use this formula: `Computes overlap: (compute + nccl - trace_duration) / nccl * 100`, another limitation might be with gpu idle time, trace_duration > compute + comm, so the estimated overlap might be lower.

@acisseJZhong Thanks for the suggestion. I indeed missed the overstimation part.

To fix both the limitations, I have updated the algorithm to now work on raw events which union the compute and communication time instead of using estimates. This should be accurate and a lot more useful for users. Let me know if this looks good now. I have updated the profiling numbers on 2x H200 using the updated algo.

@tianyu-l I have addressed all the comments. Please take a look.

(Note: I'll be away from my dev machine for the next 4 weeks, so responses may be delayed.)

torchtitan/tools/profiling.py
```diff
@@ -4,73 +4,341 @@
...
+        """
+        Enable compute-communication overlap analysis after each profiler trace.
+
+        When set to true, a CommsComputeOverlapAnalyzer is run after each trace
+        export to report compute-communication overlap efficiency. This is only
+        active when profiling is enabled and the process group is initialized
+        (distributed run).
+        """
+
+    def __init__(self, config: "Profiler.Config") -> None:
```
tianyu-l:
```suggestion
    def __init__(self, config: Config) -> None:
```

torchtitan/tools/profiling.py
```diff
@@ -4,73 +4,341 @@
...
+        # runtime args stored by active(); defaults used if active() not called
+        self._global_step: int = 0
+        self._base_folder: str = ""
+        self._leaf_folder: str = ""
+        self.torch_profiler = None
+        self.memory_profiler = None
 
-    save_memory_snapshot_folder: str = "memory_snapshot"
-    """Memory snapshot files location"""
+    def active(
```
tianyu-l, line 165: The existence of `active` seems unnecessary. Can we merge this into constructor? i.e. constructor would take Config and also the three kwargs here (later we'll do refactor to move them in config as well, don't need to worry about them right now)

torchtitan/trainer.py
```diff
@@ -789,18 +787,10 @@ def train(self):
...
-                base_folder=config.dump_folder,
-            ) as torch_profiler,
-            maybe_enable_memory_snapshot(
-                config.profiling,
-                global_step=self.step,
-                base_folder=config.dump_folder,
-            ) as memory_profiler,
-        ):
+        with self.profiler.active(
+            global_step=self.step,
```
tianyu-l, line 830: I don't know why we need global_step, but here it seems a bug that we are sending constant 0.

torchtitan/trainer.py
```diff
@@ -789,18 +787,10 @@ def train(self):
...
-                global_step=self.step,
-                base_folder=config.dump_folder,
-            ) as torch_profiler,
-            maybe_enable_memory_snapshot(
-                config.profiling,
-                global_step=self.step,
-                base_folder=config.dump_folder,
-            ) as memory_profiler,
-        ):
+        with self.profiler.active(
```
tianyu-l: continue my last comment, So we can call it
`with self.profiler as profiler:`

torchtitan/tools/profiler.py
tianyu-l, line 1: can we rename to profiler.py, to be consistent with other components.

tianyu-l (commented):
Thanks, a few more nit comments.

torchtitan/tools/profiler.py
```diff
@@ -4,73 +4,341 @@
...
+            cfg.profile_freq,
+            cfg.profiler_warmup,
+            cfg.profiler_active,
         )
 
         rank = torch.distributed.get_rank()
+        run_diagnostics = (
+            bool(self._analyzers)
+            and torch.distributed.is_initialized()
+            and rank == 0
```
felipemello1: noob question: is it a good decision to only do it for rank 0? e.g. for the MoE case, is it possible that rank_0 overlapping may be very different from some other rank?

If its not overly complex, i wonder if getting histograms/median would be desirable

torchtitan/tools/profiler.py
```diff
@@ -4,73 +4,341 @@
...
+from torch.autograd import DeviceType
+from torchtitan.config import Configurable
 from torchtitan.tools.logging import logger
 from torchtitan.tools.utils import device_module
 
 # how much memory allocation/free ops to record in memory snapshots
 MEMORY_SNAPSHOT_MAX_ENTRIES = 100000
 
+_COMM_KEYWORDS: tuple[str, ...] = ("nccl",)
+_COMPUTE_KEYWORDS: tuple[str, ...] = (
```
felipemello1: sorry, i haven't checked the checked the code deeply, but is it possible that we may be missing (or wrongly adding) some overlap because of the keyword mismatch?

torchtitan/tools/profiling.py
pbpatre: Fixed

torchtitan/tools/profiling.py
pbpatre: Agreed. Merged active() kwargs into the constructor and removed active().

torchtitan/trainer.py
pbpatre: Good catch. The construction is now moved after checkpoint load, so self.step reflects the resumed step correctly.

torchtitan/tools/profiler.py
pbpatre: Fixed along with the test file

torchtitan/tools/profiler.py
pbpatre: Great point. For FSDP2, all ranks do symmetric work so rank 0 is representative. For MoE, per-rank differences would matter. I'll leave a TODO noting this as a future enhancement and keep rank-0-only for now to avoid adding a collective inside the trace handler. Filed as a follow-up.

torchtitan/tools/profiler.py
pbpatre: Kernels that don't match either keyword list are excluded from both categories, so they can't produce a wrong overlap -- they just aren't counted. The main risk is underreporting if a novel kernel type appears. The keyword lists cover the standard PyTorch/CUDA ecosystem. We can extend them as needed. This is called out in the PR "Limitations" section.

githubsgi:
Has anyone verified  the following.  

1. The implementation  really  measures comp-comm overlap ?
2. Works with non-cuda GPU's - at the least does not break them ?

torchtitan/trainer.py
tianyu-l: curious did you not do https://github.com/pytorch/torchtitan/pull/2418#discussion_r2865863809 because you wanted to still pass `global_step=self.step` which is only available at runtime?

I feel it is a previously existing bug to pass `global_step=self.step` -- what is `global_step` doing?

tianyu-l:
> The implementation really measures comp-comm overlap ?

I think it's a proxy.

> Works with non-cuda GPU's - at the least does not break them ?

If it doesn't work, maybe you could help `raise NotImplementedError` if this flag is on and it's not CUDA.

acisseJZhong:
@pbpatre maybe worthwhile to checkout https://github.com/facebookresearch/HolisticTraceAnalysis

tianyu-l (changes requested):
People have found https://github.com/facebookresearch/HolisticTraceAnalysis to be rich and battle tested.

I think the Profiler refactor part is still great work. If you'd be able to ship that before your leave, that would be appreciated.

Thanks for understanding!

torchtitan/trainer.py
pbpatre: global_step is needed so that when resuming from a checkpoint, trace directories are named correctly (e.g. iteration_100 instead of iteration_0) and
in memoryProfiler the step() method increments step_num and checks step_num % freq != 0 to decide when to dump. Without the correct starting value, snapshot frequency alignment would be wrong after resume
In the current code, self.step is passed after self.checkpointer.load(), so it reflects the loaded checkpoint step correctly.
Does this make sense?

torchtitan/trainer.py
pbpatre: fixed this

pbpatre:
> People have found https://github.com/facebookresearch/HolisticTraceAnalysis to be rich and battle tested.
> 
> I think the Profiler refactor part is still great work. If you'd be able to ship that before your leave, that would be appreciated.
> 
> Thanks for understanding!

@tianyu-l Makes sense. I've removed CommsComputeOverlapAnalyzer and will take a deeper look at HTA. The Profiler refactor should be in good shape now.

That said, I do think having a lightweight inline signal (e.g. GPU time unutilized) directly in the default profiler output would be useful — it prompts users to start an efficiency investigation without requiring them to download and analyze traces manually. HTA is great for deep offline analysis (Claude code would be fine too), but there's value in a zero-config first signal in the logs.

One direction worth exploring: a lightweight HTA integration that runs inline after each trace export and surfaces a summary — users get the immediate signal, and can fall back to full HTA analysis when needed. Happy to file a follow-up issue to track this if that sounds reasonable.

torchtitan/trainer.py
tianyu-l: I see. That makes sense. I didn't know why we need global_step before. Thanks for explaining!

torchtitan/trainer.py
```diff
@@ -789,18 +785,10 @@ def train(self):
...
-            maybe_enable_memory_snapshot(
-                config.profiling,
-                global_step=self.step,
-                base_folder=config.dump_folder,
-            ) as memory_profiler,
-        ):
+        with config.profiler.build(
+            global_step=self.step,
+            base_folder=config.dump_folder,
+        ) as prof_session:
```
tianyu-l, line 832: nit on naming
```suggestion
        ) as profiler:
```

tianyu-l (commented):
lgtm, one nit comment.

> One direction worth exploring: a lightweight HTA integration that runs inline after each trace export and surfaces a summary — users get the immediate signal, and can fall back to full HTA analysis when needed. Happy to file a follow-up issue to track this if that sounds reasonable.

This sounds reasonable. cc @felipemello1 if would like to follow up.

tianyu-l (approved):

felipemello1:
hey @pbpatre 

> "One direction worth exploring: a lightweight HTA integration that runs inline after each trace export and surfaces a summary — users get the immediate signal"

I agree that this could be nice as a follow up PR! Should we try to get this one merged first?

I think you were going on pto, if you havent yet, can we fix the unit tests + make sure that profiler runs and generates a trace for memory and torch profile? We can merge with those fixes and then chat about HTA. wdyt?

pbpatre:
@felipemello1 Makes sense. Can you please help with fixing the failing checks and get this merged? I am on PTO and do not have access to my dev machine so can't fix these and don't want it to get blocked. Let me know if there are any questions.
cc: @tianyu-l

pbpatre:
@tianyu-l I have fixed conflicts with the latest upstream. Can you approve my workflow execution?

torchtitan/trainer.py
tianyu-l: nit: could you address this?

torchtitan/trainer.py
```diff
@@ -863,10 +851,7 @@ def train(self):
                     self.validator.validate(self.model_parts, self.step)
 
                 # signal the profiler that the next profiling step has started
-                if torch_profiler:
-                    torch_profiler.step()
-                if memory_profiler:
-                    memory_profiler.step()
+                prof_session.step()
```
tianyu-l, line 854: also here

run_train.sh
tianyu-l, line 1: why do we need this change?

tianyu-l (changes requested):
@pbpatre 
Hey sorry I forgot to merge this PR earlier.

Not sure if you have come back from PTO, but if so could you address the remaining comments so that we can merge?

tianyu-l:
since no response for a while and not directly landable, we moved to #3020 to unblock

PR #2427, opened 2026-02-23
[RL] Fix rl on main

Lucaskabela:
Following the config changes in https://github.com/pytorch/torchtitan/pull/2386 the scripts for RL fail on main.  

This PR modifies the code to restore them

### Simple RL

```bash
VLLM_BATCH_INVARIANT=1 VLLM_ATTENTION_BACKEND=FLASH_ATTN python3 torchtitan/experiments/rl/vllm_compat/simple_rl.py
```

output:
```bash
Step   0 | Loss: -0.0036 | Reward: +0.075 | Samples: 160
  Sample:  48 + 24 = 72.
Let me check the reasoning.
Okay, so Natalia sold clips to 48 of ...
...
```

### Simple Multiprocess RL
```bash
```

```
[2026-02-23 15:03:25] INFO trainer.py:114: [actor=<root>.<torchtitan.experiments.rl.unified.actors.trainer.Trainer trainer{'gpus': 0/2}>] os.getpid()=3548191 Trainer starts to train 1 on traj:
  ✓ vLLM-TorchTitan bitwise determinism verified: 20 tokens match exactly
  ✓ vLLM-TorchTitan bitwise determinism verified: 20 tokens match exactly
```

### DIFF
--- a/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
@@ -11,7 +11,7 @@ This module provides TorchTitanVLLMModel: Core model class that adapts
 TorchTitan models for vLLM.
 """
 
-from functools import partial
+from dataclasses import replace as dataclass_replace
 
 import torch
 import torch.nn as nn
@@ -27,7 +27,7 @@ from torchtitan.experiments.rl.unified.infra.parallelism_utils import (
 )
 
 from torchtitan.experiments.rl.unified.models.utils import replace_with_vllm_attention
-from torchtitan.models.qwen3.model import precompute_rope_cache
+from torchtitan.models.common import RoPE
 from torchtitan.protocols.model import BaseModel
 from torchtitan.protocols.model_spec import ParallelizeFunction
 from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -27,7 +27,9 @@
 )
 
 from torchtitan.experiments.rl.unified.models.utils import replace_with_vllm_attention
-from torchtitan.models.qwen3.model import precompute_rope_cache
+from torchtitan.experiments.rl.vllm_compat.models.qwen3.model_vllm_compat import (
```
wwwjn: This import path doesn't make sense. The implementation in model_vllm_compat is still the old version or RoPE, but we should align to `common/rope.py` implementation.

wwwjn (changes requested):
I updated this PR to use new Config System, you can use it as a reference for the RoPE part : https://github.com/pytorch/torchtitan/pull/2191 

But this PR contains more - It redesigns the RL side config system so it might go through several rounds of review. If you want to land your fix first, I'm also ok with it

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -216,8 +220,8 @@ def forward(
 
         # Get RoPE cache (handle model-specific attribute names)
         # Use hasattr to avoid ambiguous boolean value error with tensors
-        if hasattr(self.model, "rope_cache"):
-            rope_attr = self.model.rope_cache
+        if hasattr(self.model, "rope") and hasattr(self.model.rope, "cache"):
+            rope_attr = self.model.rope.cache
```
wwwjn: After the config system change, this branch would never be hit

Lucaskabela: Good catch - updated PR

PR #2428, opened 2026-02-24
[Module] Add configurable Embedding with init_weights support

fegin:
* __->__ #2428

**Why the Change**
This is the first step to make all submodules configurable and inherit from Module. We will further extend Module to support per-module sharding configurations.

**Summary**

- Introduce Embedding(nn.Embedding, Module) in torchtitan/models/common/embedding.py — a configurable embedding that uses diamond inheritance to reuse nn.Embedding logic while satisfying the Module protocol (init_weights + Configurable.Config.build()).

- Replace raw nn.Embedding(vocab_size, dim) construction in Decoder with config.tok_embeddings.build(), making the token embedding configurable (e.g., custom init_mean/init_std) without changing any model subclass.

- Harden Module.init_weights with raise NotImplementedError for runtime enforcement, since abstractmethod alone is insufficient (nn.Module uses type, not ABCMeta).

- Add unit tests for the Module diamond inheritance pattern and Embedding config/build lifecycle.


torchtitan/models/common/embedding.py
```diff
@@ -0,0 +1,51 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass, field
+
+import torch.nn as nn
+
+from torchtitan.protocols.module import Module
+
+
+class NNEmbedding(nn.Embedding, Module):
```
tianyu-l: I'd prefer just calling it `Embedding`.

torchtitan/models/common/embedding.py
tianyu-l, line 1: also at least need to have `RMSNorm` and `Linear`

torchtitan/protocols/module.py
```diff
@@ -20,5 +20,15 @@ class Module(nn.Module, Configurable):
 
     @abstractmethod
     def init_weights(self, **kwargs) -> None:
```
tianyu-l, line 22: I think we should put this in config, instead of in module, because what I mentioned in the refactor PR summary https://github.com/pytorch/torchtitan/pull/2386:
> Refactor init_weights into Module.Config instead of staying in Module
The benefit is that param init can be configurable; o/w we are coupling module implementation and its weight init.
This may require refactor of current TransformerBlock and its config. E.g. weight_init_std may need to be put in config, with __post_init__ determining its value. (See related complaints / discussions on __post_init__ by [chz](https://github.com/openai/chz/blob/main/docs/21_post_init.md))

tianyu-l: also would love to have a better name.
Previous this was called init_weights to be different from `reset_parameters`, but `weights` is a too narrow term. FSDP1 called `param_init_fn` (see https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md); internal called it `param_init`.

fegin: I prefer `init_states` as we also initialize buffers such as freqs_cis. As for moving into config, I prefer to do that after I modify all the three modules (Embedding, RMSNorm and Linear) to Module.

torchtitan/models/common/decoder.py
```diff
@@ -64,6 +65,7 @@ class Config(BaseModel.Config):
         n_layers: int
         vocab_size: int
         norm_eps: float = 1e-5
+        tok_embeddings: Embedding.Config = field(default_factory=Embedding.Config)
```
tianyu-l: I think we don't need a default here, unless it has to be CLI-configurable.
```suggestion
        tok_embeddings: Embedding.Config
```

torchtitan/models/common/decoder.py
```diff
@@ -72,11 +74,17 @@ class Config(BaseModel.Config):
         rope: RoPE.Config
         layer: TransformerBlock.Config  # required, no default
 
+        def __post_init__(self):
+            # NOTE: If subclasses override __post_init__, they MUST call
+            # super().__post_init__() to ensure tok_embeddings is configured.
+            self.tok_embeddings.num_embeddings = self.vocab_size
+            self.tok_embeddings.embedding_dim = self.dim
```
tianyu-l: This could be tricky, e.g. https://github.com/openai/chz/blob/main/docs/21_post_init.md

I'd like to further discuss how fields in a submodule should be set by parent module. cc @shuhuayu

fegin: I decided to use kwargs as I found some issues in a later PR.

shuhuayu: > This could be tricky, e.g. https://github.com/openai/chz/blob/main/docs/21_post_init.md
> 
> I'd like to further discuss how fields in a submodule should be set by parent module. cc @shuhuayu

Nice reference. I think the current style is lean in that parent modules use `build` function to pass shared fields to the submodule, and no need to duplicate them on the submodules. I see this is consistent across several modules.

Do you think it is useful to add an utility to print all configs so people can inspect the config trees conveniently? Sometimes people may get some undesired default values silently.

tianyu-l: @shuhuayu is this something you are looking for? https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L390

shuhuayu: I think it only prints training configs. I was thinking model configs, which is printed by default. In this pr, token embedding has configurable init parameters so they will be printed, but for other modules, they do not.

tianyu-l: trainer config contains model config
all modules should have init params in config, that's a refactor to be done. Actually I'm not sure how serializable it can be.

torchtitan/protocols/module.py
wconstab: would it be cleaner to make a new object 'ModuleInitializer' and then use config to choose which ModuleInitializer to use for a Module?  It feels a little weird to me to put actual model init logic into a Config class directly.

tianyu-l: I guess a question would be -- is it possible to create infinite ways of initialization with a finite set of ModuleInitializer configs.

torchtitan/config/configurable.py
```diff
@@ -15,7 +15,19 @@ class Configurable:
     - Inherits from Configurable (or Module for nn.Module components)
     - Defines a nested Config(Configurable.Config) with @dataclass(kw_only=True, slots=True)
     - Gets build() auto-wired via __init_subclass__ (no manual override needed)
-    - Accepts __init__(self, config: Config, **runtime_kwargs)
+    - Accepts __init__(self, config: Config) or __init__(self, config: Config, **runtime_kwargs)
```
tianyu-l, line 21: @shuhuayu CC and I had some discussion on this. I can share with you tmr.

torchtitan/config/configurable.py
```diff
@@ -34,7 +46,34 @@ def build(self, **kwargs):
                     f"{type(self).__name__} has no owner class. "
                     "Define Config inside a Configurable subclass."
                 )
-            return self._owner(config=self, **kwargs)
+            if not kwargs:
+                return self._owner(config=replace(self))
```
tianyu-l: oh nice, does `replace` creates a clone? shallow or deep?

torchtitan/models/common/decoder.py
```diff
@@ -19,6 +19,7 @@
     get_causal_mask_mod,
     get_document_mask_mod,
 )
+from torchtitan.models.common.embedding import Embedding
 from torchtitan.models.common.feed_forward import FeedForward
 from torchtitan.models.common.moe.moe import MoE
 from torchtitan.models.common.rope import RoPE
```
tianyu-l, line 25: This is more of a comment to my earlier PR -- I think we should be able to import all of them just from `common`?

torchtitan/models/common/embedding.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        # ``num_embeddings`` and ``embedding_dim`` are usually passed by the
+        # parent modules through build(). So the default values are None to
+        # fit Configurable.Config convention.
+        num_embeddings: int | None = None
+        embedding_dim: int | None = None
+        init_mean: float = 0.0
+        init_std: float = 1.0
```
tianyu-l, line 38: Maybe leaving a TODO to move them to states_init config later.

torchtitan/models/common/embedding.py
```diff
@@ -0,0 +1,50 @@
...
+    the parent model and passed into a cloned config.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        # ``num_embeddings`` and ``embedding_dim`` are usually passed by the
+        # parent modules through build(). So the default values are None to
+        # fit Configurable.Config convention.
+        num_embeddings: int | None = None
+        embedding_dim: int | None = None
```
tianyu-l: Did you tell me default None is unsafe?

What if we don't set default, e.g.
```suggestion
        num_embeddings: int
        embedding_dim: int
```

torchtitan/config/configurable.py
fegin: Yes, a new dataclass. Shallow.

torchtitan/models/common/decoder.py
fegin: Good point. Didn't think about this. Let me do this.

torchtitan/models/common/embedding.py
fegin: I haven't thought about the best place/best way to do this. TODO is not needed, I already have an init_state locally. I'm not very satisfied that yet.

torchtitan/models/common/embedding.py
fegin: uh, yes, this is the part I feel it is not perfect but also there is no perfect way to do. So If we don't have the default, then users always have to specify these values even though `num_embeddings` and `embedding_dim` will also be passed from `build()`. 

None is not safe 1) Python docs mention None is not safe because None may be a legit value for some dataclass field so Python introduces MISSING but this is minor for our case. 2) None is not safe in that all future Module developers need to remember to set the default values of shared fields as None.

Having default will give users the flexibility to NOT specify the values but also at some points they can do it. For example, a Module can have some fields that may or may not always share values with other components. Embedding is not the case so we can remove `None`. However if we decide to offer the flexibility, we have to keep the rule strict and simple, all fields that share values with others must set the default value as None. So Embedding should also define None.

An alternative is that users always need to set the values explicitly even though they are going to be passed from `build()` as well. IMO, this is also okay as our config is a python script, it is easy to share some variables in the config. 

Either way works. We just need to have a consensus which one to do.

torchtitan/models/common/decoder.py
fegin: Hmm, I changed `common/__init__.py` to import everything instead of explicitly specifying in two places. But if you are talking about decoder importing everything just from common, for decoder.py, then no. That would cause circular dependency because decoder.py is also in common.

fegin:
Now the fields that share values among components will use `field(init=False)`.

torchtitan/config/configurable.py
```diff
@@ -48,7 +110,7 @@ def __init_subclass__(cls, **kwargs):
                         "@dataclass(kw_only=True, slots=True)"
                     )
                 for f in fields(config_cls):
-                    if not f.kw_only:
+                    if f.init and not f.kw_only:
```
tianyu-l, line 152: why do we need this change?

torchtitan/config/configurable.py
```diff
@@ -27,14 +41,62 @@ class Configurable:
     class Config:
         _owner: ClassVar[type | None] = None
 
+        def _replace(self, **overrides):
+            """Copy this config via ``replace()``, apply *overrides*, and
+            validate that every ``field(init=False)`` slot has been set.
+
+            Raises ``TypeError`` if any ``init=False`` field is neither
```
tianyu-l, line 79: Do we really need this safety check? What if we don't -- I imagine the runtime code will error out, which is fine? I'm from a keeping things minimal perspective.

torchtitan/config/configurable.py
```diff
@@ -15,7 +15,21 @@ class Configurable:
...
+    build() auto-detects kwargs mode:
+    - All kwargs are config fields → absorbed into a cloned config, __init__ gets only config.
+    - All kwargs are NOT config fields → forwarded to __init__ as keyword arguments.
+    - Mixed → raises TypeError.
+
+    Fields that are supplied at ``build()`` time should use
+    ``field(init=False)`` so they are excluded from ``Config.__init__``.
+    Pre-set values (via attribute assignment or inheritance) that conflict
+    with a ``build()`` kwarg raise ``ValueError``; matching values are
+    accepted.  The legacy ``None``-sentinel pattern (``field: type | None =
```
tianyu-l: why do we need the legacy support? BC shouldn't be a concern, especially when this refactor is still new.

torchtitan/config/configurable.py
fegin: `field(init=False)` fields are init=False and kw_only=False (because they are not in __init__(), kw_only or not does matter).

torchtitan/config/configurable.py
fegin: The runtime code will error out. But erroring out here can give a clear message that things should be provided through `build()`. If not, then we are going to get a less clear error. People can still figure it out after checking the code.  

My perspective is that this `_replace` is coupled with `build()`, so we should erroring out with a clear message that  users do not provide the correct arguments to `build()`.

torchtitan/config/configurable.py
fegin: ye, I realized this was added by Claude. I deleted in the new version.

torchtitan/config/configurable.py
tianyu-l: if we do `@dataclass(kw_only=True)` but some field has `field(init=False)`, does it work? sounds counter-intuitive, lol

torchtitan/models/common/decoder.py
```diff
@@ -76,7 +80,9 @@ def __init__(self, config: Config):
         super().__init__()
         self.config = config
 
-        self.tok_embeddings = nn.Embedding(config.vocab_size, config.dim)
+        self.tok_embeddings = config.tok_embeddings.build(
```
wwwjn, line 83: This way of passing parent class' field to child config class' __init__ is super cool

torchtitan/models/common/__init__.py
```diff
@@ -4,50 +4,10 @@
...
-    "compute_ffn_hidden_dim",
-    "trunc_normal_",
-]
+from .attention import *  # noqa: F401,F403
+from .decoder import *  # noqa: F401,F403
+from .embedding import *  # noqa: F401,F403
+from .feed_forward import *  # noqa: F401,F403
+from .moe import *  # noqa: F401,F403
+from .rope import *  # noqa: F401,F403
+from .utils import *  # noqa: F401,F403
```
tianyu-l: I don't know how much value it brings vs. the hole it creates. E.g. if a submodule doesn't implement `__all__` then any private helpers would be exposed as well. I personally prefer revert to what we had before.

torchtitan/models/common/decoder.py
tianyu-l: oh, I forgot the circular dependency part.

torchtitan/config/configurable.py
tianyu-l: I checked and seems the annotation on individual field will override class level setup.

fegin: Yes, it will work.

torchtitan/models/common/__init__.py
fegin: sure, this is fine for me too.

tests/unit_tests/test_configurable.py
```diff
@@ -5,12 +5,45 @@
 # LICENSE file in the root directory of this source tree.
 
 import unittest
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from torchtitan.config.configurable import Configurable
 
 
 class TestConfigurable(unittest.TestCase):
```
shuhuayu, line 13: `to_dict` is an important method, should we also add a unit test for it? maybe test the two layer case?

tests/unit_tests/test_embedding.py
```diff
@@ -0,0 +1,86 @@
...
+        config = Embedding.Config()
+        emb = config.build(num_embeddings=50, embedding_dim=16)
+
+        nn.init.zeros_(emb.weight)
+        self.assertTrue(torch.all(emb.weight == 0))
+        emb.init_weights()
+        # After init_weights, weights should no longer be all zero
+        self.assertFalse(torch.all(emb.weight == 0))
+
+    def test_custom_init_std(self):
```
shuhuayu, line 44: This test seems to be weak. A un-customized init often falls into this category and pass. Maybe use a customized mean such as `0.1` and also test the sample mean using `assertAlmostEqual`.

torchtitan/config/configurable.py
shuhuayu: Thanks. The design looks elegant to me. We use `init=True` and `init=False` to deal with config inheritance, and overrides in run-time. IIUC, now the config tree can define the whole model and build it within via auto-wire.

fegin:
Rebase to get the loss CI signals

PR #2429, opened 2026-02-24
[ForgeEngine] fix broken ForgeEngine:

hann-wang:
* support launching custom trainer;
* init trainer components through .build() (#2386);
* move data to GPU by micro-batch;
* remove rescale_accumulated_loss (#2206).

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -13,7 +13,7 @@ from torchtitan.tools.logging import init_logger, logger
 from torchtitan.trainer import Trainer
 
 
-def main(custom_trainer_class: type[Trainer] | None = None) -> None:
+def main() -> None:
     """Main entry point for training."""
     init_logger()
 
@@ -38,10 +38,7 @@ def main(custom_trainer_class: type[Trainer] | None = None) -> None:
             return
 
         # pyrefly: ignore [missing-attribute]
-        if custom_trainer_class is not None:
-            trainer = custom_trainer_class(config)
-        else:
-            trainer = config.build()
+        trainer = config.build()
 
         # pyrefly: ignore [missing-attribute]
         if config.checkpoint.create_seed_checkpoint:
--- a/torchtitan/experiments/forge/example_train.py
+++ b/torchtitan/experiments/forge/example_train.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import time
 from collections.abc import Iterable
 from datetime import timedelta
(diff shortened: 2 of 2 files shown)


torchtitan/train.py
```diff
@@ -38,7 +38,10 @@ def main() -> None:
             return
 
         # pyrefly: ignore [missing-attribute]
-        trainer = config.build()
+        if custom_trainer_class is not None:
+            trainer = custom_trainer_class(config)
```
tianyu-l: shouldn't need to modify this file?

hann-wang: The example trainer is initialized with `main(CustomTrainer)`:
https://github.com/hann-wang/torchtitan/blob/8c6fee127b689ca9a8bb52d2045be597542e8ebe/torchtitan/experiments/forge/example_train.py#L408

`config_registry` is returning a `Trainer.Config` object and `config.build()` is not the custom trainer we want. Is there a better way to init a custom trainer?

joecummings: This raises a good point about the extensibility of these new Titan changes with other forms of training. I see a few options here:

1) If modifying this file, change the main function to take in any class that satisfies a "Trainer" protocol (I think technically the only requirement for main is that it can be built from a config and includes a `train`, `close` method), defaulting to the Trainer from `torchtitan.trainer`. This logic is easier to follow than `custom_trainer_class` that is deployed if provided, otherwise ignored for a _possibly_ unknown class. 
```def main(trainer: type[Trainer] = torchtitan.trainer.Trainer) ...```

2) If we're not ready to tackle the extensibility of torchtitan for other forms of training, copy the relevant logic of main.py to L408 of the forge example to make it work with the example. This should be straightforward.

Unless this *needs* to be unblocked immediately, I would have a slight preference for something like 1. 

@tianyu-l ?

tianyu-l: > config_registry is returning a Trainer.Config object and config.build() is not the custom trainer we want. Is there a better way to init a custom trainer?

To use a custom trainer, you need to assemble a config whose root node is custom trainer config instance.

> If we're not ready to tackle the extensibility of torchtitan for other forms of training, copy the relevant logic of main.py to L408 of the forge example to make it work with the example. This should be straightforward.

Yeah I think we are not ready to tackle the extensibility challenge yet, so I'm OK with option 2 to favor separation of issues for now. Sorry that I broke the forge example, but curious why you'd use it anyway?

tianyu-l (approved):

torchtitan/train.py
hann-wang: Going for option 2.

> Sorry that I broke the forge example, but curious why you'd use it anyway?

I have a custom trainer that integrates Quantization/Sparsification-Aware Training.

Issue #2430, opened 2026-02-24
Signed Integer Overflow in MoE _unpermute with Inductor Compilation


hirschsn:
### Bug description

tl;dr: There is an integer overflow in the Torch Inductor Triton implementation of `aten::index_put` in the MoE grouped_experts unpermute.

# Details

If you enable Torch Inductor compilation for the MoE grouped GEMMs (needs a code change, not only `enable=True`, c.f. https://github.com/pytorch/torchtitan/pull/2322) and run a large model, I get:

- Silent memory corruption (e.g. we observed loss and grad norm NaNs)
- Illegal memory access (e.g. reported by `cudaGetLastError`), training crashes

Backtrace:

- Triton kernel for `aten::index_put`
- `_unpermute` (torchtitan/models/moe/utils.py:65)
- `indices_padding_wrapper` (torchtitan/models/moe/utils.py:93)
- `GroupedExperts.forward`(torchtitan/models/moe/moe.py:175)
- `MoE.forward` (torchtitan/models/moe/moe.py:524)

The calling code in `_unpermute`:

```
    out_unpermuted[permuted_indices, :] = out
```

Triton implementation emitted by the Torch Inductor:

```
def triton_poi_fused_index_put_10(permuted_indices, out, out_unpermuted, ...):
    # ... omitted for brevity ...
    # tmp6 is a value from "out" to be copied to "out_unpermuted"
    # tmp4 is an int32 index into the "out_unpermuted" tensor (comes from permuted_indices)
    tl.device_assert((0 <= tmp4) & (tmp4 < 1048577), "index out of bounds: 0 <= tmp4 < 1048577")
    tl.store(out_unpermuted + (x0 + 2048*tmp4), tmp6, None) # <--- possible overflow 2048*tmp4
```

These indices are all 32-bit due to `permuted_indices` being a Tensor with 32-bit indices.
 
- The 32-bit indices are fine here, since they are in `{0, ..., 1048576}`
- But to index into the output tensor, this value gets multiplied with "out_unpermute"'s stride (=2048), which causes an overflow.

The int32 Tensor comes from:

- `fill_index_wrapper` (torchtitan/models/moe/kernels.py:80)
- `generate_permute_indices` (torchtitan/models/moe/kernels.py:207)
- `_permute` (torchtitan/models/moe/utils.py:47)
- `indices_padding_wrapper` (torchtitan/models/moe/utils.py:87)
- ... see backtrace above ...

[...]

PR #2431, opened 2026-02-24
Make tensor of indices a LongTensor

hirschsn:
Indices get multiplied with strides in operations like aten::index_put. A 32-bit index can lead to a silent overflow, since the Torch Inductor does not do any overflow checking on its own.

This commit fixes GitHub issue #2430

### DIFF
--- a/torchtitan/models/common/moe/kernels.py
+++ b/torchtitan/models/common/moe/kernels.py
@@ -135,7 +135,7 @@ def fill_indices_cpu(
                 permuted_indices[write_start:end_idx] = torch.arange(
                     start_index,
                     start_index + (end_idx - write_start),
-                    dtype=torch.int32,
+                    dtype=torch.int64,
                     # device=device,
                 )
             write_start += length


hirschsn:
The kernel that gets launched by `index_put` in eager mode is (according to a Torch profile) `void at::native::index_elementwise_kernel</* ... */, at::native::gpu_index_kernel<at::native::index_put_kernel_impl //...` .
The kernel that does the indexing is `gpu_index_kernel` (aten/src/ATen/native/cuda/IndexKernel.cu:101ff) and it receives the indices as 64-bit integers.

I think they are converted in `index_put_with_sort_kernel` (which seems to be the dispatch target for CUDA), there the indices go through `makeLinearIndex` (aten/src/ATen/native/cuda/Indexing.cu:615), which in turn does:
```
  for (auto & i : indices) {
    if (i.defined() && i.dtype() == at::kInt) {
      i = i.to(at::kLong);
    }
  }
```

hirschsn:
We measured on a Qwen3 MoE model with moderate local batch size (16), performance impact was negligible (<1% TPS).

PR #2434, opened 2026-02-24
[Module] Add configurable RMSNorm that inherits from Module

fegin:
* __->__ #2434

**BC Breaking**
This is a breaking change for any downstream code that passes `norm_eps` directly to model configs.

**Why the Change**
Same as https://github.com/pytorch/torchtitan/pull/2428. This PR changes RMSNorm to inherit from Module.

**Summary**
Introduces RMSNorm, a configurable wrapper around nn.RMSNorm that inherits from the Module. This enables RMSNorm to participate in the Config.build() pattern and provides a standardized init_weights() method.

All model families (Llama3, Llama4, Qwen3, DeepSeekV3, GptOss, Flux) are updated to:
- Use RMSNorm.Config fields in their dataclass configs instead of bare norm_eps: float
- Build norms via config.build(normalized_shape=dim) instead of nn.RMSNorm(dim, eps=...)
- Call norm.init_weights() instead of norm.reset_parameters()

The norm_eps field is removed from TransformerBlock.Config and Decoder.Config since the eps value is now encapsulated in RMSNorm.Config.

### DIFF
--- a/torchtitan/experiments/transformers_modeling_backend/model.py
+++ b/torchtitan/experiments/transformers_modeling_backend/model.py
@@ -4,9 +4,10 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import copy
 import importlib
 import math
-from dataclasses import dataclass
+from dataclasses import dataclass, fields
 
 import torch
 from torch import nn
@@ -105,6 +106,29 @@ class HFTransformerModel(BaseModel):
             self._configure_hf_attention(attn_implementation)
 
             self._initialize_dense_attributes(titan_dense_config)
+
+        def _replace(self, **overrides):
+            """Override to use ``copy.copy()`` instead of ``dataclasses.replace()``.
+
+            ``dataclasses.replace()`` re-invokes ``__init__``, which is
+            incompatible with the custom ``__init__`` here (it expects
+            ``titan_dense_config`` and calls ``PretrainedConfig.__init__``).
+            A shallow copy preserves all dynamically-set HF attributes.
+            """
+            clone = copy.copy(self)
+            for f in fields(self):
+                if f.init:
+                    continue
+                if f.name in overrides:
+                    setattr(clone, f.name, overrides[f.name])
+                elif hasattr(self, f.name):
+                    setattr(clone, f.name, getattr(self, f.name))
+                else:
+                    raise TypeError(
+                        f"{type(self).__name__} field '{f.name}' "
+                        f"(init=False) was not provided via build()"
+                    )
+            return clone
 
         def _initialize_dense_attributes(self, titan_dense_config):
             """Initialize all dense model attributes."""


torchtitan/models/qwen3/__init__.py
```diff
@@ -23,21 +24,29 @@
 
 # Adding different variants of the model
 
+# Shared config object: safe because RMSNorm.Config is an immutable-style
+# dataclass (slots=True, no mutable fields).  If mutable fields are ever
+# added, each model variant should get its own instance instead.
+norm_config = RMSNorm.Config(eps=1e-6)
```
tianyu-l: question 1: why only qwen3 is doing this shared norm config, not other models?
question 2: if it's shared by multiple nodes in the config tree including the root node, why don't we let root node pass the config to children, just like hidden dimension / vocab_size / etc.

torchtitan/models/common/decoder.py
```diff
@@ -46,10 +47,11 @@ class TransformerBlock(Module):
 
     @dataclass(kw_only=True, slots=True)
     class Config(Module.Config):
-        norm_eps: float = 1e-5
         attention: BaseAttention.Config  # required, no default
         feed_forward: FeedForward.Config | None = None
         moe: MoE.Config | None = None
+        attention_norm: RMSNorm.Config = field(default_factory=RMSNorm.Config)
+        ffn_norm: RMSNorm.Config = field(default_factory=RMSNorm.Config)
```
tianyu-l: don't need `default_factory` here?

torchtitan/models/qwen3/__init__.py
fegin: Question1: other models don't change the default eps. Qwen3 does. But we can also create _default_norm_config for each model. I don't have preferences. I lean toward to have explicitly config but Claude keeps telling me that config doesn't change the default, so it is better to just create RMSNorm.Config() each time, lol. 

Question2: then your decode loses the ability to do customized RMSNorm config for different norms. I understand that no one is doing this. But isn't the new config system aiming to make research different modeling possible? dimension and vocab_size are different cases as you could have two different vocab_size at least for the decoder, though the new implementation now allows users to pass vocab_size into Embedding.Config, it is just going to error out if it mismatches with what is passed from build().

 IMO sharing config style is better as it gives users more flexibility.

torchtitan/experiments/rl/vllm_compat/models/qwen3/model_vllm_compat.py
```diff
@@ -303,7 +303,7 @@ def __init__(self, model_args: Qwen3Model.Config):
         for layer_id in range(model_args.n_layers):
             self.layers[str(layer_id)] = TransformerBlock(layer_id, model_args)
 
-        self.norm = VLLMRMSNorm(model_args.dim, eps=model_args.norm_eps)
+        self.norm = VLLMRMSNorm(model_args.dim, eps=model_args.norm.eps)
```
acisseJZhong: nit: curious why we use `norm.eps` for model_args, and `hf_config.rms_norm_eps` below, can we align them?

tianyu-l: btw I think this file it to be deprecated.

torchtitan/models/common/attention.py
```diff
@@ -407,7 +408,8 @@ class Config(BaseAttention.Config):
         n_kv_heads: int | None = None
         head_dim: int | None = None
         qk_norm: bool = False
```
tianyu-l: can we delete this field?

torchtitan/models/flux/model/layers.py
```diff
@@ -118,12 +119,13 @@ def forward(self, x: Tensor) -> Tensor:
 class QKNorm(torch.nn.Module):
```
tianyu-l, line 119: maybe need to refactor this later to also use config build style, not urgent

torchtitan/models/qwen3/__init__.py
```diff
@@ -23,22 +24,30 @@
 
 # Adding different variants of the model
 
+# Shared config object: safe because RMSNorm.Config is an immutable-style
+# dataclass (slots=True, no mutable fields).  If mutable fields are ever
```
tianyu-l: This is NOT true? `frozen=False` so it's still mutable.

tianyu-l: To be consistent with other models, maybe we should use `get_qwen3_norm_config()` function to init a new config object each time. It also avoids instantiating a config object during import time.

torchtitan/models/common/rmsnorm.py
```diff
@@ -0,0 +1,52 @@
...
+    It is typically supplied via ``build()`` kwargs from the parent model
+    and passed into a cloned config.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        # ``normalized_shape`` are usually passed by the parent modules
+        # through build(). So the default values are None to fit
+        # ``Configurable.Config`` convention.
+        normalized_shape: int | None = None
```
wwwjn: Should we use (init=False) for `normalized_shape` field?

wwwjn: Why not set `init=False` for `normalized_shape` this field? We we want allow users to set this field from config as well?

fegin:
This PR is not up-to-date. I realized that it is painful to do two PRs at the same time while https://github.com/pytorch/torchtitan/pull/2428 keeps changing. So I didn't update this PR even though https://github.com/pytorch/torchtitan/pull/2428 is updated. I'll update this PR after https://github.com/pytorch/torchtitan/pull/2428 is landed.

fegin:
@tianyu-l @acisseJZhong @wwwjn  Is it safe to remove torchtitan/experiments/rl/vllm_compat/models/qwen3/model_vllm_compat.py, I don't know how to deal with this file though.

tianyu-l:
> Is it safe to remove torchtitan/experiments/rl/vllm_compat/models/qwen3/model_vllm_compat.py, I don't know how to deal with this file though.

What's the latest status -- If we can run inference benchmark to compare with vllm + we can achieve batch invariance with unified, then we can remove it. O/w we need to keep it for longer.

@fegin it doesn't have to use latest config system, it will be deprecated.

fegin:
@tianyu-l I didn't make `torchtitan/experiments/rl/vllm_compat/models/qwen3/model_vllm_compat.py` to be the new config. But we need to do some changes to ensure that it can still get the eps configuration as we remove `norm_eps` from the model config.

fegin:
Anyway, this PR is ready.  cc., @acisseJZhong

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -22,14 +22,21 @@
 ]
 
 
+# Safe to share: build() clones the config, so each module gets its own instance.
```
tianyu-l: I think it's fine. Two comments that we don't have to address. 
1. It is safe in the sense that build() clones the config; it is not safe in the sense that if they are modified elsewhere before the build() we have no control. Not saying that creating one config for each entry in the config registry will solve the problem, but at least there won't be cross-entry contamination. (This unsafety can be alleviated if store eventual config for each model variant somewhere.)
2. Despite calling them default config, eventually they won't be the same because each config in the registry will update the missing field e.g. `vocab_size` at module construction time. So I was thinking if we should name them e.g. `norm_config_template` but that's not super accurate name, either.

fegin: For 1) the best way to address the safety is through frozen. But `update_from_config()` may change some modules' config, which makes it impossible to set frozen for module.config in general. 

I also think it is fine to create one config per entry, which is actually what I originally did. But I did see some comments that people think it is very verbose. Forget which comment. I may revert it back one config per entry if this change cause more doubts than it solves.

fegin:
Cherry Pick https://github.com/pytorch/torchtitan/pull/2454 to this PR as we accidentally merge https://github.com/pytorch/torchtitan/pull/2454 to the wrong branch.

PR #2435, opened 2026-02-24
Fix compiler toolkit CI by removing duplicated buffer registration

yiming0416:
`self.rope.cache` and `self.freqs_cis` are the same tensor object registered as buffers on two different modules. Tracing would see them as two distinct graph inputs for the same underlying data.

This PR removes the `register_buffer` from RoPE and just store cache as a plain tensor attribute there, keeping only the Decoder-level `register_buffer("freqs_cis", ...)`.

This fixes the compiler toolkit CI:

```
NGPU=4 TRAIN_FILE=torchtitan.experiments.compiler_toolkit.train MODULE=compiler_toolkit.llama3 CONFIG=compiler_toolkit_llama3_debugmodel ./run_train.sh --parallelism.data_parallel_shard_degree=2 --parallelism.tensor_parallel_degree=2
```

### DIFF
--- a/torchtitan/models/common/rope.py
+++ b/torchtitan/models/common/rope.py
@@ -50,8 +50,7 @@ class RoPE(Module):
     def __init__(self, config: Config):
         super().__init__()
         self.config = config
-        # Buffer registered later in init_weights
-        self.register_buffer("cache", self._precompute(), persistent=False)
+        self.cache: torch.Tensor = self._precompute()
 
     def _precompute(self) -> torch.Tensor:
         cfg = self.config


yiming0416:
@tianyu-l  This fixes the compiler toolkit CI failure after your config system change. 

let me know if you are okay with changing the model code. Otherwise I can fix it from the compiler toolkit experiment.

tianyu-l (commented):
sounds OK to me, @fegin please take a look

wconstab:
> This PR removes the register_buffer from RoPE and just store cache as a plain tensor attribute there
this doesn't have any impact to checkpointing does it?

yiming0416:
> > This PR removes the register_buffer from RoPE and just store cache as a plain tensor attribute there

>  this doesn't have any impact to checkpointing does it?

@wconstab  Originally the cache was registered as a non-persistent buffer, which won't appear in state_dict, so I assume it shouldn't affect checkpointing?

yiming0416:
@fegin could you take a look? thanks!

torchtitan/models/common/rope.py
```diff
@@ -50,8 +50,7 @@ class Config(Module.Config):
     def __init__(self, config: Config):
         super().__init__()
         self.config = config
-        # Buffer registered later in init_weights
-        self.register_buffer("cache", self._precompute(), persistent=False)
+        self.cache: torch.Tensor = self._precompute()
```
wconstab, line 53: one other consideration, iiuc when registered as a non-persisten buffer, 'cache' will at least be 'moved' to device when the module is moved .to(device).  Having it as a plain tensor will not do this.  Will this break our initialization flow (starting as 'meta' and moving to 'cuda' for example?)

yiming0416: Good point! But I think the change is still safe because:
1. `rope.cache` is only consumed by `decoder.freqs_cis` and `freqs_cis` itself is a registered buffer. 
2. `decoder.init_weights()` is called with `buffer_device`. This calls `rope.init_weights()` with `buffer_device` which recomputes `self.cache` with `torch.device(buffer_device)`.

wconstab: i see. i think this is ok then

PR #2437, opened 2026-02-24
Deduplicate experiment configs by deriving from base model configs

yiming0416:
- Add `to_compiler_toolkit_config()` and `to_simple_fsdp_config()` converter functions that create experiment configs from base model configs, eliminating duplicated field definitions
 -  Add config variants for larger model sizes (8b, 70b, 405b for llama3; 16b, 671b for deepseek_v3) in both experiments
 - Fix `MODEL` → `MODULE` env var name in READMEs

### DIFF
--- a/torchtitan/experiments/compiler_toolkit/llama3/config_registry.py
+++ b/torchtitan/experiments/compiler_toolkit/llama3/config_registry.py
@@ -4,52 +4,47 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from torchtitan.components.checkpoint import CheckpointManager
-from torchtitan.components.lr_scheduler import LRSchedulersContainer
-from torchtitan.components.metrics import MetricsProcessor
-from torchtitan.components.optimizer import OptimizersContainer
-from torchtitan.config import (
-    ActivationCheckpointConfig,
-    ParallelismConfig,
-    TrainingConfig,
+from torchtitan.experiments.compiler_toolkit.configs import (
+    CompilerToolkitCompileConfig,
+    to_compiler_toolkit_config,
 )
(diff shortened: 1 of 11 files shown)


torchtitan/experiments/compiler_toolkit/trainer.py
```diff
@@ -5,28 +5,13 @@
...
-    """Compiler pass names to apply to the partitioned forward/backward graphs."""
-
-
 class CompilerToolkitTrainer(Trainer):
-    @dataclass(kw_only=True, slots=True)
-    class Config(Trainer.Config):
-        compile: CompilerToolkitCompileConfig = field(
-            default_factory=CompilerToolkitCompileConfig
-        )
+    Config = CompilerToolkitConfig
```
tianyu-l: Don't need this alias? Can just define `CompilerToolkitTrainer.Config` here (replacing `CompilerToolkitConfig`)

yiming0416: Done. But l left `CompilerToolkitCompileConfig` and the newly added `to_compiler_toolkit_config` in the newly created `configs.py`

torchtitan/experiments/compiler_toolkit/llama3/config_registry.py
```diff
@@ -4,52 +4,47 @@
...
+    llama3_70b,
+    llama3_8b,
+    llama3_debugmodel,
+    llama3_debugmodel_flex_attn,
+)
 
 from . import model_registry
 
 
 def compiler_toolkit_llama3_debugmodel() -> CompilerToolkitTrainer.Config:
```
tianyu-l, line 23: hmmm, I noticed that using strings as config names can benefit such use case more

tianyu-l (approved):
LGTM.

PR #2438, opened 2026-02-24
[Claude] Add /torch_bisect skill for bisecting PyTorch regressions

fegin:
* __->__ #2438

Try it with /torch_bisect for bisecting a problem when PyTorch daily build break TorchTitan.

**Summary**

Add a new Claude Code skill (**/torch_bisect**) that automates git bisect against the PyTorch repo to identify commits that introduce regressions breaking TorchTitan.

The skill guides the user through four phases: gathering inputs (PyTorch path, build/test commands,
known-good commit), setting up the bisect session with connectivity checks, running the build-test-record loop, and producing a summary report with the offending commit and associated PR details.

Also updates .gitignore to selectively ignore Claude Code local files. The .gitignore follows `torch/.gitignore`.

### DIFF
--- a/.claude/skills/torch_bisect/SKILL.md
+++ b/.claude/skills/torch_bisect/SKILL.md
@@ -28,15 +28,26 @@ After collecting, offer to save answers to the cache file.
 
 ## Phase 2: Setup
 
-1. `cd <pytorch> && git fetch origin` (timeout 300000ms). On failure, ask user to fix.
-2. Check `git bisect log 2>&1` — if a bisect is active, ask user to reset or abort.
-3. Resolve the good commit:
+1. **Check GitHub connectivity**: run `curl -s -o /dev/null --connect-timeout 20 -w "%{http_code}" https://github.com`.
+   If it fails or returns non-200, tell the user GitHub is not reachable and
+   suggest checking internet connectivity or proxy settings. On Meta internal
+   servers, suggest setting `https_proxy=http://fwdproxy:8080`. Do not proceed
+   until connectivity is confirmed.
+2. `cd <pytorch> && git fetch origin` (timeout 300000ms). On failure, ask user
+   to fix (remind about proxy/fwdproxy if it looks like a network error).
+3. Check `git bisect log 2>&1` — if a bisect is active, ask user to reset or abort.
+4. Resolve the good commit:
    - Hash: verify with `git log -1 <hash>`
    - Date: `git log origin/main --before="<DATE>T23:59:59" --format="%H %s" -1`
      — show result and confirm with user.
-4. Bad commit is `origin/main`. Show it to user.
-5. Show `git rev-list --count <good>..origin/main` and estimated steps (~log2).
-6. `git bisect start && git bisect bad origin/main && git bisect good <good>`
+5. Bad commit is `origin/main`. Show it to user.
+6. Show `git rev-list --count <good>..origin/main` and estimated steps (~log2).
+7. Run each bisect command individually so failures can be reported cleanly:
+   - `git bisect start`
+   - `git bisect bad origin/main`
+   - `git bisect good <good>`
+   If any step fails, show the error and run `git bisect reset` before asking
+   the user how to proceed.
 
 ## Phase 3: Bisect Loop
 


.claude/skills/torch_bisect/SKILL.md
```diff
@@ -0,0 +1,261 @@
...
+   (e.g., `/data/users/chienchin/mywork/pytorch`).
+
+2. **PyTorch build command** — The shell command to recompile PyTorch after
+   each checkout (e.g., `python setup.py develop`).
+
+3. **TorchTitan test command** — The command to run from the TorchTitan
+   directory that reproduces the issue. Tell the user:
+   > "This command must **exit 0 on a good PyTorch commit** and **exit non-zero
+   > on a bad commit**. I will use only the exit code to judge good vs bad — I
+   > will NOT analyze log output to determine the result."
```
wconstab: wondering if you're being too rigid here.  i would probably trust claude to analyze log output for me.  that could help if the issue is a loss degradation or something that requires a bit of judgement.  but this is also OK as a starting point.

.claude/skills/torch_bisect/SKILL.md
```diff
@@ -0,0 +1,261 @@
...
+# PyTorch Git Bisect for TorchTitan
+
+You are performing a git bisect on the PyTorch repository to find the exact
+commit that introduced a regression affecting TorchTitan. Follow each phase
+below precisely.
+
+## Phase 1: Gather Information
+
+First, record the TorchTitan directory by running `pwd`. Store this absolute
+path — you will need it throughout the bisect to switch back.
```
xmfan: This is surprising, did you have cases where claude got lost? The latest claude code should be smart enough to spin up subtasks (separated context) to do the build process, and that should keep the main claude's context very available

wconstab (commented):
looks useful! but also tbh, looks more detailed than i would have thought it needs to be.  I think if i just tell claude very vaguely- bisect pytorch for me using this torchtitan test command and this pytorch build command, it will probably do OK.  does it need so much exact specification?  (just curious, really)

.claude/skills/torch_bisect/SKILL.md
```diff
@@ -0,0 +1,261 @@
...
+After collecting all answers, ask the user if they want to save them to
+`~/.claude/torchtitan_torch_bisect_cache.json` for future runs. If yes, write
+all 5 answers as JSON to that file.
+
+## Phase 2: Setup
+
+### Step 1: Fetch latest PyTorch
+
+```bash
+cd <pytorch_path> && git fetch origin
```
xmfan: do you have fwdproxy enabled on your environment at all times? maybe we can ask claude to enable it or remind the user to enable it

.claude/skills/torch_bisect/SKILL.md
```diff
@@ -0,0 +1,261 @@
...
+Use a timeout of 600000ms (10 minutes).
+
+**If the command completes**: use its exit code:
+- Exit 0 → good commit
+- Non-zero → bad commit
+
+**If the command times out**: follow the user's chosen timeout policy from
+Phase 1 — either automatically mark as bad, or ask the user what to do.
+
+Do NOT analyze the test output to determine pass/fail. Use ONLY the exit code.
```
xmfan: i'm not sure this is desired, there might be transient failures causing the bisect to abort early. we could ask the subagent to match the error exactly

.gitignore
```diff
@@ -43,5 +43,9 @@ Sessionx.vim
 # env files
 .env
 
-# Vibe coding
-.claude
+# Claude Code local configuration
+CLAUDE.local.md
+/test_*.py
+/debug_*.py
```
xmfan, line 49: 🤔

tianyu-l: I also feel they are somewhat random and seemingly tied to this skill.

fegin: No, I copied from PyTorch,  lol. 

I can explain some files, when you ask Claude to try some features, it will create some test_*.py to understand the behavior. I think the rules in PyTorch just want to ensure that these files are not accidentally committed. I do see the 4.6 model always deleted the files but I sometimes ask Claude to preserve so that I can read as well.

I don't mind to remove these.

.claude/skills/torch_bisect/SKILL.md
fegin: No, I didn't ask Claude to do this. Claude added the absolute path requirement but I think it is okay. So I didn't delete it but we can remove this. I don't think it will go wrong.

.claude/skills/torch_bisect/SKILL.md
fegin: The answer is yes. I'm being rigid. The reason is that I saw Claude once tried to fix PyTorch installation issue when I asked it to run some tests. Claude "fix" PyTorch and the test can be run. But in reality, Claude just used some incorrect way to fix PyTorch and the test signal I got was completely incorrect. 

For bisect, we have a clear signal either pass or not pass. I bet Claude can help to parse the log but I'm also afraid that Claude may incorrectly choose the direction and accidentally mark the commit to be good or bad.

My opinion is that the test should have good signal, either pass or not pass. Even for loss degradation, we have command that we can ask Claude to use.

.claude/skills/torch_bisect/SKILL.md
fegin: Yes, I do have fwdproxy enabled by default. It's okay to ask Claude to do so.

.claude/skills/torch_bisect/SKILL.md
fegin: See my answer to @wconstab  above. I have seen Claude incorrectly fix a PyTorch installation issue. My opinion is that we should have clear test signal if we want to do bisect. Otherwise, it is possible Claude may steer to an incorrect direction. I trust it can find many cases most of the time but not always.

fegin:
> but also tbh, looks more detailed than i would have thought it needs to be. I think if i just tell claude very vaguely- bisect pytorch for me using this torchtitan test command and this pytorch build command, it will probably do OK. does it need so much exact specification? (just curious, really)

I asked Claude to make it more concise about the writing which reduces the file to less than 100 LoC. I'll update it later. 

But I think you are asking that if we need to give so many instructions. The short answer is we don't need that many instructions, less detailed instructions will work most of time too (I didn't test it but I believe it would work).

The reason why I want to be rigid and give Claude more instructions is that I treat this skill as a tool rather than some skills that I'll use for development. So I want a deterministic result and a lower chance to get false positive results. I don't need it to be creative.  We have clear test signals of failure or success, at least for now. For example, I added the output format because I found I may need to have 1-2 more conversations as it sometime only show me PR number. Those "not analyze the log" is also by design. I currently don't see any benefit of letting Claude analyze the log, at least from my previous bisect experience using Claude. This may change in the future. 

This is just my own opinion. We can change the instructions to fit most people's working styles.

PR #2439, opened 2026-02-25
Update ROCm docker image to ubuntu-24.04 and clang-15 version

akashveramd:
In this PR, Updated ROCm docker image to ubuntu-24.04 and clang-15 version. It is to address the issue seen in https://github.com/pytorch/torchtitan/actions/runs/22235826367/job/64327070658

### DIFF
--- a/.github/workflows/integration_test_8gpu_features.yaml
+++ b/.github/workflows/integration_test_8gpu_features.yaml
@@ -45,7 +45,7 @@ jobs:
       docker-image: ${{ matrix.docker-image }}
       repository: pytorch/torchtitan
       upload-artifact: outputs
-      timeout: 45
+      timeout: 60
       script: |
         set -eux
 


.github/workflows/integration_test_8gpu_compiler_toolkit.yaml
```diff
@@ -68,14 +68,7 @@ jobs:
...
 
-        if [ "${{ matrix.gpu-arch-type }}" = "rocm" ]; then
-          python -m pip install --force-reinstall --pre \
-            "torch==${{ matrix.torch-version }}" \
-            --index-url ${{ matrix.index-url }}
-        else
-          python -m pip install --force-reinstall --pre \
-            torch --index-url ${{ matrix.index-url }}
-        fi
+        python -m pip install --force-reinstall --pre torch --index-url ${{ matrix.index-url }}
```
jagadish-amd, line 71: If this PR is going to be merged (depends on the testing), lets keep matrix.torch-version for both rocm and cuda(remove if else).  In set-matrix.yaml, pass the "torch-version" as blank for both rocm and cuda so that in future we can use "torch-version" parameter if required.

akashveramd:
Closing this PR since ubuntu-24.04, clang-15 is not helping fix it.

PR #2441, opened 2026-02-25
[Checkpointer] Remove the dependencies on PyTorch distributed state_dict APIs

fegin:
* __->__ #2441

**Why this PR**
TorchTitan checkpointer relies on PyTorch distributed state_dict APIs to ensure optimizer and model state_dicts are as expected (e.g., FQNs are correct, flattened optimizer state_dict). However distributed state_dict APIs provide way too many features and support many legacy APIs (FSDP1, DDP).

This PR aims to clean up the dependency of PyTorch distributed state_dict APIs and extract the necessary logic to TorchTitan. We can upstream some utilities back to PyTorch in the future.

**Summary**
- Removes the dependency on torch.distributed.checkpoint.state_dict APIs (get_model_state_dict, set_model_state_dict, get_optimizer_state_dict, set_optimizer_state_dict, StateDictOptions), which carry excessive complexity from legacy FSDP1/DDP support.
- Introduces `torchtitan/tools/checkpoint_utils.py` with focused utilities: _orig_mod prefix stripping for model state dicts, FQN-keyed optimizer state dict conversion using PyTorch's param_names feature, and flatten/unflatten logic for DCP-compatible optimizer checkpointing.
- Constructs optimizers with (name, param) tuples so that FQN mappings are natively available without relying on distributed state_dict introspection.
- **Adds an unit test (test_state_dict_keys.py) that validates the exact model and optimizer state dict key contract against hard-coded expected values for the llama3 debugmodel, catching regressions from refactoring or upstream PyTorch changes.**
- **Adds a tool like loss_compare.py but is used to verify if a checkpoint generated by "old PR" can be correctly loaded by a "new PR".**


torchtitan/components/checkpoint_utils.py
tianyu-l, line 1: for proximity, can we put this in components/checkpointer/utils.py?

torchtitan/components/optimizer.py
```diff
@@ -138,23 +143,26 @@ def zero_grad(self, *args, **kwargs) -> None:
...
+
+        Note: This method has a side effect. If the optimizer states have not
+        been initialized yet (i.e., no training step has been taken), it
+        initializes them by performing a zero-grad step via ``init_optim_state``.
+        This is required to produce a valid state dict before the first optimizer
+        step for DCP.
+        """
+        result: dict[str, Any] = {}
+        for optim in self.optimizers:
+            init_optim_state(optim)
```
tianyu-l: Is `get_optimizer_state_dict` doing this internally already?

torchtitan/components/checkpoint.py
```diff
@@ -42,6 +36,10 @@
 from torchtitan.components.optimizer import OptimizersContainer
 from torchtitan.config import Configurable, TORCH_DTYPE_MAP
 from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
+from torchtitan.tools.checkpoint_utils import (
+    clean_model_load_state_dict,
+    clean_model_state_dict,
```
tianyu-l: `clean_` sounds corresponding to `dirty` -- why would we have dirty things in the first place?

Can we have a better name, maybe `get_canonical_model_state_dict`

fegin: `clean_name`/`clean_path` are quite common to represent that processed paths, especially for state_dict/checkpointer in internal code base and even torch logger also use `clean_name`. We can probably use `get_clean_fqn` to be more specific.

As for `state_dict` APIs, we can probably use `canonical`.

torchtitan/components/optimizer.py
fegin: yes

torchtitan/components/checkpoint.py
tianyu-l: sure, I don't want to fight with convention, lol

torchtitan/components/checkpoint.py
```diff
@@ -68,23 +66,18 @@ def __init__(self, model: nn.Module | list[nn.Module]) -> None:
...
-        func = functools.partial(
-            set_model_state_dict,
-            model_state_dict=state_dict,
-            options=StateDictOptions(strict=False),
-        )
-        list(map(func, self.model))
-        # `set_model_state_dict()` does change the keys of the input state_dict,
-        # we will need to reinitialize the cache_state_dict.
+        for model in self.model:
+            clean_model_load_state_dict(model, state_dict, strict=False)
```
wwwjn: The only function I used in RL is set_model_state_dict(), and in RL there's no fqn name prefix issue so I guess I could directly call model.load_state_dict()

fegin: @wwwjn Is torch.compile being used? If so, then you have the fqn prefix issue. If not, then you can just use `model.load_state_dict()` as you don't need optimizer state_dict.

torchtitan/components/checkpoint_utils.py
fegin: This is a quite tricky as there may be checkpointer->optimizer->checkpointer/utils circular dependency if the order doesn't set correct in checkpointer. Let me see what I can do.

torchtitan/components/checkpoint_utils.py
```diff
@@ -0,0 +1,269 @@
...
+                else 0.0
+            )
+    optim.step(closure=None)
+    for param_group in optim.param_groups:
+        if "lr" in param_group:
+            param_group["lr"] = lrs.pop(0)
+    optim.zero_grad(set_to_none=True)
+
+
+def optim_state_dict_to_fqn_keys(optim_sd: dict[str, Any]) -> dict[str, Any]:
```
mori360, line 123: why it's clean in model state dict, but not in optim?

.github/workflows/integration_test_8gpu_models.yaml
```diff
@@ -76,5 +76,6 @@ jobs:
 
         python -m tests.integration_tests.run_tests --gpu_arch_type ${{ matrix.gpu-arch-type }} --test_suite models $RUNNER_TEMP/artifacts-to-be-uploaded --ngpu 8
         python -m tests.integration_tests.flux $RUNNER_TEMP/artifacts-to-be-uploaded/flux --ngpu 8
+        torchrun --nproc_per_node=8 tests/integration_tests/test_state_dict_keys.py
```
tianyu-l: This sounds a GPU unit test rather than integration test.

torchtitan/components/optimizer.py
```diff
@@ -13,16 +12,18 @@
...
 from torch.optim import Optimizer
 from torchtitan.config import Configurable
 from torchtitan.distributed import ParallelDims
+from torchtitan.tools.checkpoint_utils import (
+    clean_name,
+    flatten_optim_state_dict,
+    init_optim_state,
+    optim_state_dict_to_fqn_keys,
+    unflatten_and_load_optim_state_dict,
+)
```
tianyu-l: The API surface seems lacking structure.

tianyu-l (changes requested):

torchtitan/components/checkpoint_utils.py
fegin: put it as component/checkpoint_utils.py

scripts/checkpoint_compat_test.py
tianyu-l, line 1: do you plan to remove this script before merge?

torchtitan/components/checkpoint_utils.py
```diff
@@ -0,0 +1,298 @@
...
```
tianyu-l, line 123: do we need this function to stand alone? May be fine to put the content directly in `get_flat_optim_state_dict`

torchtitan/components/checkpoint_utils.py
```diff
@@ -0,0 +1,298 @@
...
+            current = result
+            for part in parts[:-1]:
+                if part not in current:
+                    current[part] = {}
+                current = current[part]
+            current[parts[-1]] = val
+    return result
+
+
+def unflatten_and_load_optim_state_dict(
```
tianyu-l, line 216: similar nit, maybe don't need another level of function calling

torchtitan/components/checkpoint_utils.py
```diff
@@ -0,0 +1,298 @@
...
+    for pg in sd["param_groups"]:
+        for fqn in pg["params"]:
+            for k, v in pg.items():
+                if k != "params":
+                    ret[f"param_groups.{fqn}.{k}"] = v
+
+    return ret
+
+
+def _reconstruct_nested(flat_sd: dict[str, Any], prefix: str) -> dict[str, Any]:
```
tianyu-l, line 199: maybe make this inline function

torchtitan/components/checkpoint_utils.py
```diff
@@ -0,0 +1,298 @@
...
+            raise KeyError(
+                f"Optimizer state contains param ID {int_id} that is not in any "
+                f"param_group. Known param IDs: {sorted(id_to_fqn.keys())}"
+            )
+        new_state[id_to_fqn[int_id]] = state_val
+
+    return {"state": new_state, "param_groups": new_param_groups}
+
+
+def _flatten_state_nested(d: dict[str, Any], prefix: str, out: dict[str, Any]) -> None:
```
tianyu-l, line 162: similar nit: maybe make inline

torchtitan/components/checkpoint_utils.py
tianyu-l: Other than `canonical_fqn` which can be put somewhere neutral (maybe torchtitan/tools/utils.py), the other functions are either for model, or for optimizer
- for the ones for model, we can just put them in checkpoint.py because that's where we define `ModelWrapper`
- for the ones for optimizer, we can put them in optimizer.py

The argument is that each component should be responsible for its own state dict logic.

scripts/checkpoint_compat_test.py
fegin: No. I actually repeated this verification multiple times when changing state_dict and checkpointer. This script will be used again in the future.

torchtitan/components/checkpoint_utils.py
fegin: That was the original design after the distributed state_dict received one major feedback from users, there are too many features in two APIs and many of these are not always needed and are proved to by composable. The goal of this PR is not just removing the dependencies but also create utilities that we can eventually upstream back to PyTorch.
The example here, init_optim and optim_state_dict_to_fqn_keys  are the two common ones. Checkpointing team is already asking these utilities.

torchtitan/components/checkpoint_utils.py
fegin: I think these APIs should be put together. These APIs are not used to process `ModelWrapper` nor `OptimizerWrapper`, they are used to process nn.Module and optim.Optimizer state_dict logic. So logically speaking, they belong to the same place -- one place where it is for processing model and optimizer state_dict for `distributed checkpointing`. Without `distributed checkpointing`, these APIs are not needed.

torchtitan/components/checkpoint_utils.py
fegin: This is a recursive function, that's why it is a standalone one.

torchtitan/components/checkpoint_utils.py
fegin: I can make this a private function. But I don't see the reason why we need to inline this function, though I do want to move `optim.load_state_dict` out from this function to make this function purely unflattening purpose.

torchtitan/components/checkpoint_utils.py
fegin: I intentionally do this because the original distributed state_dict code is hard to understand because every function is complicated. I don't see the reason to inline this one.

fegin:
I'll do another round to make some functions private, but I don't want to inline many of them. Also, IMO, these APIs are logically together -- they are just a simpler version of distributed state_dict APIs. The best place to put them "separately' are where nn.Module and Optimizer are defined, not where our ModuleWrapper and OptimizerWrapper are defined.

torchtitan/components/checkpoint_utils.py
tianyu-l: I'm OK with not putting it in the function where it's used. But being recursive doesn't mean it cannot be an inline, embedded function?

torchtitan/components/checkpoint_utils.py
tianyu-l: > But I don't see the reason why we need to inline this function

The reason I'm asking these is that this file looks like a collection of many function, some of them are related to each other, some of them are not. Putting small util functions inline help the reader to understand this is a private function only used by another user-facing function, thus establishing some structure (o/w user can only search callsites).

I don't have strong opinions tho. Upstreaming to pytorch sounds good.

tianyu-l (approved):

PR #2423, opened 2026-02-23
[RL] Add sum digits task and pre and post training evaluation

daniellepintz:
* Add sum digits task which model can learn and improve on in 12 steps
* Add `evaluate` function and runs evaluation before and after training
* Convert the INFO logging which runs on every step to DEBUG logging so we can visualize loss and rewards
* Add per-token normalization of log probs to avoid long completions causing loss explosion

Final result:
```
[2026-02-23 06:44:14] INFO simple_rl_multiprocess.py:205: [actor=<root>] Evaluating pre-training baseline...
[2026-02-23 06:44:24] INFO simple_rl_multiprocess.py:87: [actor=<root>] Eval: Accuracy=20% (4/20) Format=30% (6/20)
[2026-02-23 06:44:24] INFO simple_rl_multiprocess.py:209: [actor=<root>] ================================================================================
[2026-02-23 06:44:24] INFO simple_rl_multiprocess.py:210: [actor=<root>] Starting RL training for 12 steps
[2026-02-23 06:44:24] INFO simple_rl_multiprocess.py:211: [actor=<root>] ================================================================================
NCCL version 2.28.9+cuda12.9
/home/daniellepintz/torchtitan/titan-rl-env/lib/python3.12/site-packages/torch/distributed/c10d_logger.py:83: UserWarning: barrier(): using the device under current context. You can specify `device_id` in `init_process_group` to mute this warning.
  return func(*args, **kwargs)
[2026-02-23 06:45:06] INFO simple_rl_multiprocess.py:222: [actor=<root>] Step   0 | Loss: -0.0057 | Reward: -0.450 | Correct: 11/40
[2026-02-23 06:45:41] INFO simple_rl_multiprocess.py:222: [actor=<root>] Step   1 | Loss: -0.0051 | Reward: -0.600 | Correct: 8/40
[2026-02-23 06:46:19] INFO simple_rl_multiprocess.py:222: [actor=<root>] Step   2 | Loss: -0.0049 | Reward: -0.100 | Correct: 18/40
[2026-02-23 06:46:55] INFO simple_rl_multiprocess.py:222: [actor=<root>] Step   3 | Loss: -0.0047 | Reward: +0.205 | Correct: 24/40
[2026-02-23 06:47:31] INFO simple_rl_multiprocess.py:222: [actor=<root>] Step   4 | Loss: -0.0051 | Reward: +0.100 | Correct: 22/40
[...]

PR #2443, opened 2026-02-26
[rl][combo] Refactor simple RL loop with torchtitan components

wwwjn:
This PR is a combination of all the 4 different PRs: 
https://github.com/pytorch/torchtitan/pull/2244
https://github.com/pytorch/torchtitan/pull/2221
https://github.com/pytorch/torchtitan/pull/2194
https://github.com/pytorch/torchtitan/pull/2191

## Current status:

- Applied **the same parallelism** on trainer / generator, eg, Trainer (TP=2) and Generator (TP=2)
    - NOTE: now we have a strong assumption that the trainer and generator must have the same parallelism, and are collocated. This is because we unwrap the DTensor and assume we can always wrap it back with the same device mesh and placement.
- We should be able to remove this constrain once we have more powerful weight sync. cc @daniellepintz
- Weight transfer: Simply unwrap DTensor in trainer before sending
- Collocated trainer and generator

## Next Step:
- Patch batch-invariant kernel, test if the trainer / generator is batch-invariant
- Supports different parallelism on trainer / generator side
- CI as a guard


torchtitan/experiments/rl/unified/configs.py
```diff
@@ -0,0 +1,42 @@
...
+
+    top_p: float = 0.95
+    """Nucleus sampling threshold."""
+
+    max_tokens: int = 100
+    """Maximum number of tokens to generate per completion."""
+
+
+@dataclass(kw_only=True, slots=True)
+class PolicyOptimizationConfig:
```
tianyu-l: If it's GRPO specific, the name should reflect that. Also maybe better place to put this is simple_grpo.py

torchtitan/experiments/rl/unified/configs.py
```diff
@@ -0,0 +1,42 @@
...
+
+These are plain dataclasses (not ``Configurable``) since they are just data,
+not buildable components.
+"""
+
+from dataclasses import dataclass
+
+
+@dataclass(kw_only=True, slots=True)
+class VLLMSamplingConfig:
```
tianyu-l: This could be put into generator.py if that's the only place using it.

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,60 @@
...
+Usage:
+    from torchtitan.experiments.rl.unified.plugin import register
+    register(model_spec)
+"""
+
+from torchtitan.protocols.model_spec import ModelSpec
+
+# Model-agnostic name used for vLLM model registration.
+# Must match the hf_overrides["architectures"] value passed to EngineArgs.
+VLLM_MODEL_NAME = "TorchTitanForCausalLM"
```
tianyu-l: what's the "For" mean?
```suggestion
VLLM_MODEL_NAME = "TorchTitanCausalLM"
```

torchtitan/experiments/rl/unified/plugin.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+vLLM plugin for TorchTitan models.
+
+Usage:
+    from torchtitan.experiments.rl.unified.plugin import register
+    register(model_spec)
```
tianyu-l: need to update

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -0,0 +1,227 @@
...
+        generator: Generator.Config = field(default_factory=Generator.Config)
+        """Generator actor configuration (vLLM engine, sampling)."""
+
+
+async def main():
+    """Run the distributed RL training loop using Monarch."""
+
+    config = ConfigManager().parse_args()
+
+    # Patch model_spec to use the RL-specific parallelize function.
```
tianyu-l: most things below should be put into RLTrainer, after config.build() gives you an instance of RLTrainer.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -0,0 +1,227 @@
...
+    # but RL training needs the version with inner_attention hooks
+    # (PrepareModuleInputOutput) to convert DTensors to local tensors for
+    # vLLM's flash attention kernels.
+    from torchtitan.experiments.rl.unified.models.parallelize import parallelize_qwen3
+    config.model_spec.parallelize_fn = parallelize_qwen3
+
+    if config.batch_invariant_mode:
+        # Validate that trainer and generator have the same parallel plan
+        # since they are collocated on the same mesh
+        assert config.trainer.parallelism == config.generator.parallelism, (
```
tianyu-l, line 95: > NOTE: now we have a strong assumption that the trainer and generator must have the same parallelism, and are collocated. This is because we unwrap the DTensor and assume we can always wrap it back with the same device mesh and placement.

If you have this limitation, the assertion should happen regardless of batch_invariant_mode or not.

torchtitan/experiments/rl/unified/actors/utils.py
tianyu-l, line 1: if it's loss related, should we call it loss.py?
it's only used by trainer.py, should we put it in actors/trainer/loss.py?

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,114 @@
...
+import torch
+from monarch.actor import Actor, endpoint
+from torchtitan.experiments.rl.unified.configs import PolicyOptimizationConfig
+from torchtitan.experiments.rl.vllm_compat.simple_rl import trivial_reward_function
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Episode:
```
tianyu-l: this is used by all of trainer, generator, and grader, can we put it in a separate file?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,353 @@
...
+        """
+        with torch.device("meta"):
+            with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
+                model = model_spec.model.build()
+
+        # Replace attention with vLLM compatible attention for RL training.
+        # NOTE: Long-term this will be replaced by pytorch attention
+        # supporting paged attention / kv cache.
+        if batch_invariant_mode:
+            replace_with_vllm_compatible_flash_attention(model)
```
tianyu-l: why keeping `replace_with_vllm_attention`, is it faster?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
-    completions: list[str]
-    vllm_token_ids: list[list[int]]
-    vllm_token_log_probs: list[list[float]]
-    prompt_token_ids: list[list[int]]
-    rewards: torch.Tensor
-    advantages: torch.Tensor
-
-
-class VLLMRolloutEngine:
+class Generator(Actor, Configurable):
```
tianyu-l: call it VLLMGenerator if it's being coupled with vllm tightly.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
-            )
-            logger.info("Created new vLLM engine")
-        else:
-            # Direct parameter copy into model tensors.
-            # This bypasses vLLM's reload_weights() which uses a layerwise
-            # reload mechanism that moves params to meta device
-            from torchtitan.experiments.rl.vllm_compat.weights_vllm_compat import (
-                vllm_compat_to_torchtitan,
-            )
+        vllm_attention_backend: str = "FLASH_ATTN"
```
tianyu-l: If you call it VLLMGenerator.Config, you no longer need to have "vllm_" prefix any more.

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -61,39 +115,33 @@ class TorchTitanVLLMModelWrapper(nn.Module):
...
-
-        # Setup RoPE cache extension function if provided
-        self.rope_cache_extension_fn = partial(
-            precompute_rope_cache,
-            dim=self.config.head_dim,
-            base=self.config.rope_theta,
-        )
+        self.config = model_spec.model
+        logger.debug(f"Creating model with config: {self.config}")
+        self.model = self.config.build()
```
tianyu-l, line 135: you may need to do meta init when model is large, OK to put a TODO

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -214,22 +256,13 @@ def forward(
...
-            rope_attr = None
-
         # Extend RoPE cache if needed (vLLM profiling may use 2x max_seq_len)
         if positions is not None:
             max_position = positions.max().item()
         else:
             max_position = 0
 
-        rope_cache = self._extend_rope_cache_if_needed(rope_attr, max_position)
+        rope_cache = self._extend_rope_cache_if_needed(self.model.freqs_cis, max_position)
```
tianyu-l: I think we should build this capability into torchtitan models. cc @fegin @shuhuayu

tianyu-l: @wwwjn maybe make an issue

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -308,13 +360,32 @@ def load_weights(self, weights_iter):
...
-            model_state_dict=torchtitan_state_dict,
-            options=StateDictOptions(strict=False),
-        )
+        return self.load_weights_from_state_dict(torchtitan_state_dict)
 
-        loaded_params = {f"model.{name}" for name in torchtitan_state_dict.keys()}
+    def load_weights(self, weights_iter):
+        """
+        vLLM required API.
+        Load weights from HF checkpoint using the provided state dict adapter.
```
tianyu-l: is this comment meaningful?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
tianyu-l: Ideally we should do this on the config, rather than on the model. The vllm attention module needs to support build from config.

torchtitan/experiments/rl/unified/actors/generator.py
joecummings: nit -> VllmGenerator

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+        """vLLM attention backend to use (e.g., FLASH_ATTN, XFORMERS)."""
+        
+        vllm_model_dtype: str = "bfloat16"
+        """Data type for model weights, passed directly to vLLM (auto, float16, bfloat16, float32)."""
 
-            titan_state = vllm_compat_to_torchtitan(vllm_compat_state)
-            self._direct_weight_update(titan_state)
+        vllm_gpu_memory_limit: float = 0.5
```
joecummings: I think in another comment, you said the trainer and generator were *not* co-located, in which case, why are we limiting this to 0.5? The default is at least 0.9 IIRC.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+        self.model_spec = model_spec
 
-        return (
-            completions,
-            log_probs,
-            token_ids_list,
-            token_log_probs_list,
-            prompt_token_ids_list,
+        # Register TorchTitan model with vLLM before any engine creation
+        from torchtitan.experiments.rl.unified.plugin import (
```
joecummings: why is this a nested import?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+        """Access the model from the vLLM engine.
+        Returns a TorchTitanVLLMModelWrapper instance.
+        """
+        return self._engine.model_executor.driver_worker.get_model()
+
     @endpoint
-    async def generate(self) -> None:
-        """Generate trajectories and compute rewards/advantages."""
-        logger.info(
+    async def generate(self) -> Episode:
```
joecummings: Just a note that this will change in #2423 to allow for a proper passing of prompt to the generate function so no one calls this out

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+        logger.debug(
             f"{os.getpid()=} Generating start generate (policy v{self.policy_version})..."
         )
-        async with self.cond:
-            # Wait until ready to generate (weights have been updated)
-            await self.cond.wait_for(
-                lambda: self.state == GeneratorState.READY_TO_GENERATE
-            )
 
+        with torch.no_grad():
```
joecummings, line 168: Is it actually necessary to do this under `torch.no_grad`? I don't see that pattern in other vLLM use cases

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,114 @@
...
+
+        logger.info(
+            f"Grader finished scoring: "
+            f"reward_mean={reward_mean.item():.4f}, reward_std={reward_std.item():.4f}"
+        )
+
+        return episode
+
+    @endpoint
+    async def set_reward_fn(self, reward_fn: Callable) -> None:
```
joecummings: Apologies for the repeat comment, but when might we want to use this function?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,353 @@
...
-        learning_rate: float = 1e-5,
-        model_mode: str = ModelMode.VLLM_COMPAT,
-        ddp_size: int = 1,
-        tp_size: int = 1,
+        config: Config,
+        *,
+        model_spec: ModelSpec,
+        policy_optimization: PolicyOptimizationConfig,
+        batch_invariant_mode: bool,
+        hf_assets_path: str = "./tests/assets/tokenizer",
```
joecummings: Is there a way to default to None or to a common location rather than a test asset?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,353 @@
...
+                "No state_dict_adapter available, skipping initial weight load"
+            )
+            return
+
+        storage_reader = self.sd_adapter.get_hf_storage_reader(checkpoint_path)
+        hf_state_dict = self.sd_adapter.to_hf(model.state_dict())
+        dcp.load(hf_state_dict, storage_reader=storage_reader)
+        torchtitan_state_dict = self.sd_adapter.from_hf(hf_state_dict)
+
+        from torch.distributed.checkpoint.state_dict import (
```
joecummings: Why is this a nested import?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,353 @@
...
+                episode.prompt_token_ids, episode.vllm_token_ids
+            ):
+                token_lps = compute_token_log_probs(
+                    self.ref_model, prompt_toks, gen_toks, device
+                )
+                ref_token_log_probs.append(token_lps)
+
         # Compute loss
-        loss, loss_metrics = compute_policy_gradient_loss_vllm(
+        loss, loss_metrics, batch_token_log_probs = compute_policy_gradient_loss(
```
joecummings, line 283: Can we compute the forward_backward first and then pass this to the loss to keep the loss function *only* computing the loss itself? This will match the eventual pattern we want with Trainer.

torchtitan/experiments/rl/unified/actors/utils.py
joecummings: We can even put it directly in trainer.py for now until we have a more generic loss abstraction

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: In current PR, the trainer and generator are collocated

torchtitan/experiments/rl/unified/actors/utils.py
wwwjn: This file is not only loss related now, and it's more like a "catch-all" `utils.py`, including: extract log probability,  calculate loss, verify trainer/generator log prob. Agree that the loss part haven't been carefully designed, and need to refactor later. Now it should not be called loss.py

torchtitan/experiments/rl/unified/actors/trainer.py
wwwjn: UPDATE: We will need replace_with_vllm_attention to patch **generator** model's attention to be vLLM's attention (`vllm.Attention()`) . This is because `vllm.Attention()` layer handles kv cache management.

I will update the docstring to clearily state why we need 2 different patch now

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > VLLMGenerator

vLLM is their correct name :(, so I will go with VLLMGenerator()

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+    via weight sync. Generates completions for given prompts and
+    computes rewards/advantages.
 
     Args:
-        model_path: Path to HuggingFace model (for config/tokenizer)
-        temp_checkpoint_dir: Directory to save temporary weight checkpoints
+        config: Generator-specific configuration.
+        model_path: Path to the HF model checkpoint.
+        batch_invariant_mode: Enable batch-invariant mode for deterministic ops.
+        policy_optimization: GRPO hyperparameters.
```
felipemello1: "policy_optimization: GRPO hyperparameters."

its very hard for me, by just reading this, to understand what is "policy_optimization". Policy makes me think of model, what are we optimizing? I would think that number of rollouts, for example, is just another "sampling param", for example.

I think i would also remove the specific "GRPO" term, since there are variations of RL that would also benefit from a generator.

It would be helpful too if there was a type hint here, maybe
`policy_optimization (type): description`

but i can just go ahead and read the actual inputs

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+        prompt_texts: List of prompt strings.
```
felipemello1, line 61: this is tricky. Usually we want to do tokens in and tokens out, i.e. we pass tokens to the generator. But lets say we pass texts. Ideally we should pass a dict of messages, e.g.
{
 Message(role="system", text="you are a helpful assistant"),
 Message("role"="user", "text"="how much is 2+2+"),
}

it doesnt need to be addressed in this PR, since its already huge, but i think its an important fix

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+        expected_answers: List of expected answer strings.
```
felipemello1: Its unclear to me why a generator would need to know expected answer. This seems to be crossing a boundary from the environment. A generator generates, an env or reward_fn judges the outcome

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: Added a TODO. Now we directly import from the `models/qwen3/__init__.py` to get existing titan model registry, need replace the dataclass' Attention config to be VLLMAttention()

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
wwwjn: I quickly check the meta init possibility, and I feel like we have to return a non-meta init mode before TorchtitanVLLMWrapper.__init__() returns. This is because vllm engine will run a dummy run right after the engine initialization, before calling vllm_engine.generate(), and FA kernels are not be able to run with meta init tensors

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
-            CommConfig(),
+
+        # Extract needed fields from configs
+        self.model_path = model_path
+        self.max_new_tokens = config.sampling.max_tokens
+        self.temperature = config.sampling.temperature
+        self.group_size = policy_optimization.group_size
+
+        # Build vLLM engine
+        engine_kwargs = dict(
```
felipemello1, line 126: this is ok for this PR, but my intuition is that maybe it would be healthier if the user could just pass their own EngineConfig, and we just did 

```
engine_kwargs = {**EngineConfig}
```

or maybe something like:
```
def some_fn(arg1, arg, **custom_engine_kwargs)
	engine_kwargs = {arg1, arg2, **custom_engine_kwargs}
```

otherwise, 
1. every time the user wants to have access to some extra arg, they would have to change the API of this class.
2. A lof ot cognitive load for the user. They have to learn our codebase, instead of just looking at vllm

Same idea for SamplingParams.

Letting the user pass what they want would be the most flexible/future proof.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
-            rewards = self.reward_fn(
-                completions, self.expected_answers, self.group_size
+            sampling_params = SamplingParams(
+                temperature=self.temperature,
+                max_tokens=self.max_new_tokens,
+                n=self.group_size,
+                seed=self.config.vllm_seed,
+                logprobs=1,
+                prompt_logprobs=1,  # Also get prompt log probs to access prompt token IDs
+                output_kind=RequestOutputKind.FINAL_ONLY,  # Only return completed outputs
```
felipemello1, line 177: what happens when its a truncated sample? do we resample or handle it well?

felipemello1: I think that this shouldnt be hidden here. I dont think that the user expects that this is the behavior. Should probably be configurable.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Oh that's a good point, now in VLLMGenerator.Config, there's several fields are only used and directly used in LLMEngine. 

I think we could have a EngineConfig class within VLLMGenerator.Config. We could have a follow up refactor and I will leave a  TODO

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+                    prompt_token_ids_list.append(prompt_token_ids)
+                    token_ids_list.append(sample.token_ids)
+                    per_token_log_probs = [
+                        list(logprob_dict.values())[0].logprob
+                        for logprob_dict in sample.logprobs
+                    ]
+                    token_log_probs_list.append(per_token_log_probs)
+
+        # Create episode with zero rewards (Grader will fill them in)
+        episode = Episode(
```
felipemello1: hmm, i would think that this should return a list[Episode], so if i want 8 outputs, i would get 8 objects.

but it seems that we return:
Episode(arg1=List[...], arg2=List[]), and then later the user has to parse it again to reconstruct their episodes. This doesnt seem intuitive to me.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,254 @@
...
+            policy_version=self.policy_version,
+            completions=completions,
+            vllm_token_ids=token_ids_list,
+            vllm_token_log_probs=token_log_probs_list,
+            prompt_token_ids=prompt_token_ids_list,
+            expected_answers=self.expected_answers,
```
felipemello1: this was a zero-op. IMO, expected answers shouldnt be passed to the generator. Whoever is calling the generator should update the episode with the expected answer

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > I think i would also remove the specific "GRPO" term, since there are variations of RL that would also benefit from a generator.

I change it back to "GRPOConfig" per @tianyu-l 's comment. "PO" standards for "Policy Optimization"  in "GRPO" , that's why I tried to use "Policy Optimization" to represent all different kinds of RL algorithm at the beginner.

Now I've more lean into keep this Config field narrow to GRPO, and let's brainstorm a better name

felipemello1: Edit: it seems that it only provides n_groups, which is a sampling param. I think it should be defined with config.sampling, and we can remove it from here and simplify

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: I wrote this to make sure TorchTitanVLLMWrapper.model is under inference mode, for other vllm native model, they might do this within vllm (I;ll need to double check)

torchtitan/experiments/rl/unified/actors/trainer.py
wwwjn: Thanks, I apologies that loss part need a redesign, we can probably do it in https://github.com/pytorch/torchtitan/pull/2423? I want to address this part in next PR and land this first


Added a TODO here

torchtitan/experiments/rl/unified/actors/utils.py
wwwjn: > We can even put it directly in trainer.py for now until we have a more generic loss abstraction

I was a little bit concerned because compute_token_log_probs + compute_policy_gradient_loss is too long

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: "complete" output means: don't generate intermediate results during autoregressive generation.

> what happens when its a truncated sample

If vllm generates a truncated sample, then vllm is not reliable. Truncated sample can be figoured out within vllm

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: That's right, we can follow up in later PR. Left a TODO here

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: nice catch, removed

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Aha, nice catch! Refactored this part

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Thanks, I simplified and removed the GRPOConfig as no necessary field it would need. I don't think "n_group" is purely a sampling param, it's part of the GRPO algorithm. And I put it to VLLMGenerator.Config().

`VLLMGenerator.Config.SamplingConfig` is mainly used for decoding strategy now

torchtitan/experiments/rl/unified/README.md
```diff
@@ -36,23 +36,17 @@ uv pip install torch vllm xformers  --pre \
 python scripts/download_hf_assets.py --repo_id Qwen/Qwen3-0.6B --local_dir torchtitan/experiments/rl/example_checkpoint --all --hf_token=...
 ```
 
-4. Run inference:
+4. Run inference with unified model definition:
 ```bash
-python torchtitan/experiments/rl/unified/infer.py
+torchrun --nproc_per_node=<world_size> \
+      torchtitan/experiments/rl/unified/infer.py
```
tianyu-l: need to update

torchtitan/experiments/rl/unified/types.py
```diff
@@ -0,0 +1,38 @@
...
+
+        text: str
+        token_ids: list[int]
+        token_log_probs: list[float]
+
+    policy_version: int
+    prompt_token_ids: list[int]
+    expected_answer: str = ""
+    completions: list[Completion] = field(default_factory=list)
+    rewards: torch.Tensor = field(default_factory=lambda: torch.zeros(0))
```
tianyu-l: why it's not one reward in each Completion?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,256 @@
...
+        # Register TorchTitan model with vLLM before any engine creation
+        register_model_to_vllm_model_registry(model_spec)
 
-class Generator(Actor):
-    """
-    Generates rollouts using vLLM engine.
+        # Set vLLM environment variables from config before any vLLM initialization
+        if batch_invariant_mode:
+            os.environ["VLLM_BATCH_INVARIANT"] = "1"
+            init_batch_invariance(AttentionBackendEnum.FLASH_ATTN)
```
acisseJZhong: use `config.attention_backend`?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,256 @@
...
+        Returns a TorchTitanVLLMModelWrapper instance.
+        """
+        return self._engine.model_executor.driver_worker.get_model()
+
     @endpoint
-    async def generate(self) -> None:
-        """Generate trajectories and compute rewards/advantages."""
-        logger.info(
+    async def generate(self) -> list[Episode]:
+        """Generate episodes and return list of Episode (one per prompt) with zero rewards.
```
acisseJZhong: wdym by zero rewards?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,256 @@
...
+                    )
+
+                episodes.append(
+                    Episode(
+                        policy_version=self.policy_version,
+                        prompt_token_ids=prompt_token_ids,
+                        completions=completions,
+                        rewards=torch.zeros(len(completions)),
+                    )
                 )
```
felipemello1, line 214: 1. We are adding to Episode a list[Completions]. I think i dont follow why there would be multiple samples in one output. Isnt it a single generation/completion?

2. its ok to keep this for now, but IMO we should add a todo to remove "Episode" from here. Episode is an RL abstraction. Different users may have different definitions.

the generator should only output the completion, i.e. tokens, logprobs, and stop reason. Everything downstream is the controller problem.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -4,489 +4,256 @@
...
+                if isinstance(tensor, DTensor):
+                    continue
+                target_dtensor = model_state_dict[name]
+                local_state_dict[name] = DTensor.from_local(
+                    tensor.to(target_dtensor.device_mesh.device_type),
+                    device_mesh=target_dtensor.device_mesh,
+                    placements=target_dtensor.placements,
+                )
+
+        load_weights = self._get_model().load_weights_from_state_dict(local_state_dict)
```
felipemello1, line 250: this looks right, just a sanity check, have we ever confirmed that weights actually get updated?

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Evaluates completions and assigns rewards to episode data.
+
+    The Grader receives episode data from the Generator
+    and computes rewards using a reward function. Advantage computation
+    is done by the Trainer.
+
+    Args:
+        reward_fn: Optional custom reward function. If not provided,
+                   uses trivial_reward_function from simple_rl.
```
felipemello1: i think its generally a good thing to have defaults, but here it would probably be safer to error if the user forgets to define a reward_fn

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Score episodes by computing rewards.
+
+        Args:
+            episodes: List of Episode data (one per prompt, with completions)
+
+        Returns:
+            Episodes with computed rewards
+        """
+        policy_version = episodes[0].policy_version if episodes else -1
```
felipemello1: this sounds defensive, we can probably remove it and let it error

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,359 @@
...
+    Uses ModelSpec for model construction, parallelization, and weight loading.
 
     Args:
-        titan_checkpoint_path: Path to TorchTitan checkpoint
-        model_path: Path to HuggingFace model
-        learning_rate: Learning rate for optimizer
-        model_mode: Indicates which model to use. Train inferece unified model, batch invariant Torchtitan model,
-            or plain Torchtitan model
+        config: PolicyTrainer.Config for model/optimizer/parallelism settings.
+        num_samples_per_prompt: Number of completions per prompt (from sampling config).
```
felipemello1: this is only used to calculate the advantage, and only necessary to compute batch//num_samples_per_prompt.

A better option, that avoid this division, would be to keep the advantage calculation close to where the group is created. e.g.

```
episodes = generate()
episodes = compute_rewards(episodes)
episodes = compute_advantages(episodes)
```

now the trainer can just do episode.advantage.

the trainer should look like a regular SFT/pretraining. It doesnt need to know about RL

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,359 @@
...
         )
+        model.train()
+        self.model = model
+        self.model_parts = [model]
 
-        self.model = self.model.to(device)
-        self.model.train()
+        # Create reference model for KL divergence (frozen copy of initial policy)
+        ref_model = self._build_model(
+            model_spec, config, device_type, batch_invariant_mode, hf_assets_path
```
felipemello1, line 119: Maybe a TODO to add here for follow up, but ref_model should be a different actor, IMO. In the controller, you can do:

```
episodes = ...
episodes.ref_logprob = RefModelActor(episodes)
loss = ModelActor(episodes)
```

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -6,131 +6,359 @@
...
+        reward_mean = rewards.mean()
+        reward_std = rewards.std()
+        if reward_std > 1e-8:
+            rewards_normalized = (rewards - reward_mean) / reward_std
+        else:
+            rewards_normalized = rewards - reward_mean
+
+        return compute_grpo_advantages_stable(
+            rewards_normalized, self.num_samples_per_prompt
+        )
```
felipemello1: it looks like we are computing advantage twice.

torchtitan/experiments/rl/unified/actors/grader.py
```diff
@@ -0,0 +1,72 @@
...
+        logger.info(
+            f"Grader scoring {len(episodes)} episodes (policy v{policy_version})..."
+        )
+
+        all_rewards = []
+        for episode in episodes:
+            completion_texts = [c.text for c in episode.completions]
+            rewards = self.reward_fn(completion_texts, episode.expected_answer)
+            episode.rewards = rewards
```
felipemello1: this is a bit error prone. Its safer to have a episode.reward, no plural. If we have multiple rewards that we want to log, we can add to metadata or .raw_rewards

torchtitan/experiments/rl/vllm_compat/simple_rl.py
felipemello1, line 599: can we delete this and just keep "compute_grpo_advantages", or vice versa?

felipemello1, line 633: we are using the loss in util, can we delete it, or vice versa?

felipemello1, line 657: is this repeated from utils/trainer?

felipemello1, line 919: this seems repeated from the trainer, i guess i dont understand the difference between simple_rl and trainer

torchtitan/experiments/rl/unified/README.md
```diff
@@ -62,10 +56,10 @@ Work on batch invariance:
...
-2. Remove hardcoded configs and dependency on Qwen3 Model. Use torchtitan's config/ModelSpec instead, to work with any model.
+2. Remove hardcoded configs and dependency on Qwen3 Model. Use torchtitan's config/TrainSpec instead, to work with any model.
 3. Need to load the gsm8k dataset using TorchTitan dataset.
 4. Need to properly implement weight saving and loading using TorchTitan's checkpoint mechanism, or use TorchStore. Also need to
     replace `vllm_to_torchtitan` and `torchtitan_to_vllm` calls to TorchTitan [state dict adaptor](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/qwen3/model/state_dict_adapter.py).
 5. Right now we only support trainer run on multiple processes using DDP, and generator using TP, need to onboard more parallelism.
 6. Right now we only support VLLM_COMPAT mode to achieve batch invariance and bitwise determinism, need to support UNIFIED mode.
-7. In the longer term, need to add trajectory queue to achieve async, right now trainer and generator are running synchronously.
+7. In the longer term, need to add episode queue to achieve async, right now trainer and generator are running synchronously.
```
acisseJZhong, line 65: nit: don't need to change. I might missing the discussion on using `Episode` as the name, do you know why we want to call it `Episode`?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -0,0 +1,227 @@
...
+            metrics = self.trainer_actor.step.call(scored_episodes).get().item(gpus=0)
+            # 4. Sync weights back to generator (all TP ranks)
+            weight_mesh = self.trainer_actor.get_weights.call().get()
+            weights = {
+                gpu: weight_mesh.item(gpus=gpu)
+                for gpu in range(self.trainer_world_size)
+            }
+            self.generator.update.call(metrics["policy_version"], weights).get()
+
+            logger.info(
```
acisseJZhong, line 201: nit: can we also log the trainer generator log probs diff?

acisseJZhong (approved):
yayyy!

torchtitan/experiments/rl/unified/types.py
wwwjn: Now the reward is flattened in as a list, I agree it's more clear if we put one reward in each Completion. Updated

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > I think i dont follow why there would be multiple samples in one output. Isnt it a single generation/completion?

This is part of GRPO algorithm, for each prompt, will generate `n_samples_per_prompt` completions and use this group to calculate reward for each completion.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Yes, I have checked this through bit-wise identity check (trainer 1 GPU + generator 1GPU). This is not a direct check, but I have fixed several weight update problem which will causing bit-wise identity check failure. We can think about CI plan - At least we need unit test for each component, eg weight update

torchtitan/experiments/rl/unified/actors/grader.py
wwwjn: Nice catch, now changed to error out when there's no explicit reward function.

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: > IMO we should add a todo to remove "Episode" from here. Episode is an RL abstraction. Different users may have different definitions

Do you mean we should let orchestrator to assemble the Episode?

torchtitan/experiments/rl/unified/actors/trainer.py
wwwjn: Simplified this part and remove this unnecessary import / helper function

wwwjn:
> hey, thanks for all of the work here! Its a lot
> 
> I left a bunch of comments, but i understand that we want to land and then refactor. Its up to you if you want to address them or mark as todo.
> 
> Great job!

Thanks, comments under `rl/unified` are addressed. But i didn't address comments on `rl/vllm_compat`

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: The `reward` field is not fulfilled yet, because now we put "calculating reward" part of `Grader`'s job

torchtitan/experiments/rl/unified/README.md
wwwjn: IIRC the name Episode is aligned with original forge name

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: Nice catch, modified

torchtitan/experiments/rl/unified/actors/generator.py
felipemello1: >Do you mean we should let orchestrator to assemble the Episode?

yes, think of the multiturn case, we do:
```
while True:
     completion = generation(msg_history) # currently this returns an episode
     msg_history.append(completion)
     env_output = env(msg_history)
     msg_history.append(env_output)

episode = make_episode(msg_history)
```

a generators job is only to generate. The controller will put things together.

felipemello1: >This is part of GRPO algorithm, for each prompt, will generate n_samples_per_prompt

i agree, i guess i dont understand the nested for loop. Why would 1 episode have N completions?

```
# first loop of n_samples_per_prompt
for output in all_outputs:
      prompt_token_ids = output.prompt_token_ids
      completions = []

	  # second loop that adds N completions to 1 Episode
      for sample in output.outputs:
			completions.append(
                        Episode.Completion(...
```

daniellepintz (approved):

PR #2444, opened 2026-02-26
[simple_fsdp] Add CUDAGraph support for SimpleFSDP via --compile.use_cudagraph

IvanKobzarev:
Reuse compiler_toolkit's CUDAGraphWrapper (warmup -> record -> replay lifecycle with shared memory pool) to capture NCCL collectives inside CUDA graphs. Inductor's built-in cudagraph integration cannot handle NCCL ops, so we disable it and wrap fw/bw compiled graphs externally.

For inductor + CG: use aot_autograd with custom compilers that apply bucketing pass -> compile_fx_inner -> CUDAGraphWrapper. This gives us Triton-compiled kernels with CUDA graph capture around the full graph.

Only supported with backend='inductor' (aot_eager has CUDA graph memory pool conflicts between forward and backward recording).

Usage:
  NCCL_GRAPH_REGISTER=0 torchrun ... --compile.use_cudagraph

### DIFF
--- a/torchtitan/experiments/simple_fsdp/backend.py
+++ b/torchtitan/experiments/simple_fsdp/backend.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from collections.abc import Callable
 from typing import Any
 
 import torch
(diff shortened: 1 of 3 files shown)


aditvenk:
@IvanKobzarev  Why is NCCL_GRAPH_REGISTER=0  needed in the usage?

torchtitan/experiments/simple_fsdp/backend.py
```diff
@@ -15,11 +16,91 @@
...
+
+def _build_aot_eager_cg_backend(bucketing_pass_fn: Callable) -> Callable:
+    """Build an aot_autograd backend for aot_eager + CUDAGraph.
+
+    Applies: bucketing_pass -> CUDAGraphWrapper wrapping the GraphModule.
+    Returns CUDAGraphWrapper directly (not gm with replaced forward) to avoid
+    recursion through LazyGraphModule's self-call in _lazy_forward.
+    """
+    from torch._dynamo.backends.common import aot_autograd as aot_autograd_backend
+    from torchtitan.experiments.compiler_toolkit.cudagraph import (
```
yiming0416, line 76: In general, we have `compiler_toolkit` experiment depends on `simple_fsdp` experiment. So it's not a good idea to import functions in the other direction.

I'd strongly recommend adding CUDAGraph support to the JIT case after we merge the two experiments.

yiming0416 (changes requested):

torchtitan/experiments/simple_fsdp/configs.py
```diff
@@ -21,24 +19,16 @@ class SimpleFSDPCompileConfig(CompileConfig):
         auto_bucketing, transformer_block_bucketing
     """
 
+    use_cudagraph: bool = False
+    """
+    Enable CUDA Graphs via compiler_toolkit.cudagraph.CUDAGraphWrapper.
+    Captures compiled fw/bw graphs into CUDA graphs (warmup -> record -> replay)
+    with a shared memory pool, reducing kernel launch overhead.
+    Supported with backend='inductor' and 'aot_eager'.
+    Requires NCCL_GRAPH_REGISTER=0 env var when NCCL collectives are present.
```
fegin, line 28: Is it possible to automatically set the env?

torchtitan/experiments/simple_fsdp/backend.py
IvanKobzarev: It will be good to have only one entrypoint. If we switch torch.compile and  compiler_toolkit - than we can have only one path. 

But until we have many paths to use compile - it makes sense to show how to use CudaGraphs in all of them.

I think torch.compile path is more natural for users, user  just want to call it and potentially set minimal configs if default behavior is not good enough. 

We can think how to refactor not to have deps (we can just inline that wrapper).
I think with inline dep it should not be a problem.

IvanKobzarev:
> @IvanKobzarev Why is NCCL_GRAPH_REGISTER=0 needed in the usage?

Torchtitan by default uses `PYTORCH_CUDA_ALLOC_CONF="expandable_segments:True"` which is not playing well with NCCL buffer registration in case of cuda graphs(pinning memory). 
expandable_segments can give physically non-contiguous region as virtual-memory allocation; pinning will not work with non-contiguous physical addresses.

In specifically our usecase - without setting `NCCL_GRAPH_REGISTER=0` with expandable_segments I saw IllegalMemoryAccess in cudagraphs.

The alternative is to turn off expandable_segments for runs with cudagraphs. (without expandable_segments we increase probability of OOM; with NCCL_GRAPH_REGISTER=0 we loss ~1% of perf not pinning the buffers).

torchtitan/experiments/simple_fsdp/configs.py
IvanKobzarev: I think yes, I do not know when nccl reads the env, but I think it should be possible.

torchtitan/experiments/simple_fsdp/backend.py
yiming0416: I am merging the two experiments in https://github.com/pytorch/torchtitan/pull/2457.

Would appreciate if you can wait that to be landed first and rebase on it. Then we won't have the circular deps issue.

PR #2389, opened 2026-02-17
Execute moe.gate in float32

chelsea0x3b:
Original discussion #2225.

Per comments this PR now changes the gate to happen in f32.

Run on 8xb200.

| | AC=none | AC=full |
| --- | --- | --- |
| gate in bfloat16 (main branch) | memory: 175.60GiB(98.45%)  tps: 12,437  tflops: 389.45  mfu: 17.31% | memory: 66.91GiB(37.51%)  tps: 10,050  tflops: 314.69  mfu: 13.99% |
| gate in float32 (this PR) | memory: 171.76GiB(96.30%)  tps: 12,393  tflops: 388.07  mfu: 17.25% | memory: 66.91GiB(37.51%)  tps: 9,990  tflops: 312.83  mfu: 13.90% |

## output from runs

<details>
<summary> Full AC with float32 gate (this PR)</summary>

```
[titan] 2026-02-24 19:48:29,513 - root - INFO - step:  1  loss: 12.69021  grad_norm: 10.0625  memory: 66.91GiB(37.51%)  tps: 2,075  tflops: 64.97  mfu: 2.89%
[titan] 2026-02-24 19:48:29,513 - root - INFO - step:  1  loss: 12.69021  grad_norm: 10.0625  memory: 66.91GiB(37.51%)  tps: 2,072  tflops: 64.87  mfu: 2.88%
[titan] 2026-02-24 19:48:29,513 - root - INFO - step:  1  loss: 12.69021  grad_norm: 10.0625  memory: 66.91GiB(37.51%)  tps: 2,106  tflops: 65.95  mfu: 2.93%
[titan] 2026-02-24 19:48:29,513 - root - INFO - step:  1  loss: 12.69021  grad_norm: 10.0625  memory: 66.91GiB(37.51%)  tps: 2,062  tflops: 64.58  mfu: 2.87%
[titan] 2026-02-24 19:48:29,513 - root - INFO - step:  1  loss: 12.69021  grad_norm: 10.0625  memory: 66.91GiB(37.51%)  tps: 2,075  tflops: 64.96  mfu: 2.89%
[titan] 2026-02-24 19:48:29,513 - root - INFO - step:  1  loss: 12.69021  grad_norm: 10.0625  memory: 66.91GiB(37.51%)  tps: 2,075  tflops: 64.97  mfu: 2.89%
[...]

PR #2448, opened 2026-02-27
MoE gate in fp32 

acisseJZhong:
#2389

created PR for testing

### DIFF
--- a/torchtitan/experiments/forge/engine.py
+++ b/torchtitan/experiments/forge/engine.py
@@ -174,7 +174,8 @@ class ForgeEngine(torch.distributed.checkpoint.stateful.Stateful, Configurable):
             buffer_device = None
 
         self.loss_fn = self.train_spec.build_loss_fn(
-            config.compile, parallel_dims=parallel_dims)
+            config.compile, parallel_dims=parallel_dims
+        )
 
         # verify batch sizes
         global_batch_size = config.training.global_batch_size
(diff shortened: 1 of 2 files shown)


garrett361:
@acisseJZhong I see a few failing tests, can you tell if they're due to this change?

garrett361:
Hmm, some of these also seem to be failing on the latest commit, so maybe not related https://github.com/pytorch/torchtitan/actions/runs/22447324713/job/65041514394?pr=2432

acisseJZhong:
@garrett361 yeah I think the failing test is due to other reason. Running MoE locally with router TP2 I don't see any issues(i.e. the previous error is mixed DTensor and Tensor)
```
MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel NGPU=4 ./run_train.sh   --parallelism.pipeline_parallel_degree 1   --parallelism.data_parallel_shard_degree 2   --parallelism.tensor_parallel_degree 2   --parallelism.expert_parallel_degree 2   --parallelism.expert_tensor_parallel_degree 2  
````

garrett361:
What was failing in the `Linear` wrapper approach? Something with compile?

acisseJZhong:
you mean the current approach in #2389? i think this might be too complicated if using torch.cast alone is enough. But I just kicked off the CI test for #2389.

garrett361:
Ok, think I misunderstood you here https://github.com/pytorch/torchtitan/pull/2389#issuecomment-3971443317 , thought you were saying the Linear wrapper impl had some inductor failures.

acisseJZhong:
np, I meant the CI for this approach has failure on inductor, but locally running MOE TP2 is fine.

PR #2451, opened 2026-02-27
[MFU] Skip MFU display when FP8/MX quantization is active and add GB200/GB3…

shuhuayu:
Hide MFU when quantization is active and add peak tflops for GB200 and GB300. This is a short-term answer to https://github.com/pytorch/torchtitan/issues/560. 

Before this pr, when fp8 is enabled. MFU is displayed but over bf16 peak flops.
1-26 2311505,123 - го61 -

After this pr, when fp8/mx quantization is active, MFU is shown to be N/A. 
6 71148 gred 10rk 2 445

### DIFF
--- a/torchtitan/protocols/model_converter.py
+++ b/torchtitan/protocols/model_converter.py
@@ -4,10 +4,11 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 from dataclasses import dataclass, field
-from typing import ClassVar, Protocol
+from typing import Protocol
 
 import torch.nn as nn
 
+from torchtitan.components.quantization import QuantizationConverter
 from torchtitan.config import Configurable
 from torchtitan.distributed import ParallelDims
 from torchtitan.tools.logging import logger
@@ -29,18 +30,6 @@ class ModelConverter(Protocol):
     def post_optimizer_hook(self, model: nn.Module | list[nn.Module]):
         """Post-optimizer (optional) hook (e.g. compute weights statistics)."""
         ...
-
-
-class QuantizationConverter(Configurable):
-    """Base class for quantization converters (FP8, MX, etc.).
-
-    All quantization converter classes should inherit from this so they can be
-    identified via isinstance checks.
-    """
-
-    @dataclass(kw_only=True, slots=True)
-    class Config(Configurable.Config):
-        _quantization_type: ClassVar[str]
 
 
 class ModelConvertersContainer(Configurable, ModelConverter):
--- a/torchtitan/tools/utils.py
+++ b/torchtitan/tools/utils.py
@@ -113,15 +113,11 @@ def get_peak_flops(device_name: str) -> float:
         # Ref: https://www.tomshardware.com/news/
         # nvidias-latest-regulation-compliant-gpu-for-china-has-been-delayed-to-early-next-year
         return 148e12
-    elif "GB300" in device_name:
-        # Grace Blackwell Ultra Superchip (Grace CPU + Blackwell Ultra GPU)
+    elif "GB200" in device_name or "GB300" in device_name:
+        # Grace Blackwell Superchips (Grace CPU + Blackwell GPU)
         # BF16 dense per GPU: 2,500 TFLOPS (half of 5,000 TFLOPS with sparsity)
-        # data from https://www.nvidia.com/en-us/data-center/dgx-gb300
-        return 2.5e15
-    elif "GB200" in device_name:
-        # Grace Blackwell Superchip (Grace CPU + Blackwell GPU)
-        # BF16 dense per GPU: 2,500 TFLOPS (10 PFLOPS for 2 GPUs with sparsity / 2 / 2)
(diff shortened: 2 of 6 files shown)


torchtitan/components/metrics.py
```diff
@@ -479,8 +481,13 @@ def log(
         # model FLOPS utilization
         # For its definition and calculation, please refer to the PaLM paper:
         # https://arxiv.org/abs/2204.02311
-        mfu = 100 * self.num_flops_per_token * tps / self.gpu_peak_flops
+        # MFU is based on BF16 peak FLOPS which is misleading when quantization
+        # (FP8/MX) is active, so we skip it in that case.
         tflops = self.num_flops_per_token * tps / 1e12
+        if self.has_quantization:
+            mfu = None
```
tianyu-l, line 490: can we set it to `"N/A"` here to avoid the
`self.logger.log({k: v for k, v in metrics.items() if v is not None}, step)` below? Not sure if it can be "logged"

torchtitan/trainer.py
```diff
@@ -272,6 +272,12 @@ def __init__(self, config: Config):
         )
         model_converters.convert(model)
 
+        # Check if any converter uses quantization (FP8, MX, etc.)
+        has_quantization = any(
+            getattr(cc, "_quantization_type", None) is not None
+            for cc in config.model_converters.converters
+        )
```
tianyu-l, line 280: This is too hacky. A less hacky way is to let all quantization converters inherit QuantizationConverter parent class, and use `isinstance(..., QuantizationConverter.Config)` to test here.

torchtitan/trainer.py
```diff
@@ -286,6 +292,7 @@ def __init__(self, config: Config):
             model_param_count,
             self.metrics_processor.num_flops_per_token,
         ) = model_config.get_nparams_and_flops(model, config.training.seq_len)
+        self.metrics_processor.has_quantization = has_quantization
```
tianyu-l: let's put it in build arg, instead of doing this direct setter. Things can become hard to read / debug if we use too many direct setter.

torchtitan/components/metrics.py
shuhuayu: `"N/A"` cannot be handled by TensorBoard so I propose to put mfu into metrics dict only when it is not `None`.

torchtitan/trainer.py
shuhuayu: Sounds great. Built a wrapper config `QuantizationConverter(Configurable)`.

torchtitan/tools/utils.py
```diff
@@ -113,6 +113,16 @@ def get_peak_flops(device_name: str) -> float:
         # Ref: https://www.tomshardware.com/news/
         # nvidias-latest-regulation-compliant-gpu-for-china-has-been-delayed-to-early-next-year
         return 148e12
+    elif "GB300" in device_name:
+        # Grace Blackwell Ultra Superchip (Grace CPU + Blackwell Ultra GPU)
+        # BF16 dense per GPU: 2,500 TFLOPS (half of 5,000 TFLOPS with sparsity)
+        # data from https://www.nvidia.com/en-us/data-center/dgx-gb300
+        return 2.5e15
+    elif "GB200" in device_name:
```
tianyu-l: let's consolidate the two entries

torchtitan/protocols/model_converter.py
```diff
@@ -31,6 +31,18 @@ def post_optimizer_hook(self, model: nn.Module | list[nn.Module]):
         ...
 
 
+class QuantizationConverter(Configurable):
```
tianyu-l: Because of proximity, let's put this in quantization folder, maybe quantization/__init__.py

torchtitan/protocols/model_converter.py
```diff
@@ -83,13 +95,14 @@ def post_optimizer_hook(self, model: nn.Module | list[nn.Module]):
...
+    Each quantization converter Config inherits from QuantizationConverter.Config
+    and defines a `_quantization_type` ClassVar (e.g. "float8" or "mx").
+    This function asserts they are all the same.
     """
     existing_type: str | None = None
     for config in converters:
-        qt = getattr(config, "_quantization_type", None)
-        if qt is not None:
+        if isinstance(config, QuantizationConverter.Config):
+            qt = config._quantization_type
```
tianyu-l, line 94: Can we just use the type of the config, instead of this extra string? If so, let's deprecate `_quantization_type`.

shuhuayu: Good suggestion. This is a little bit tricky. For example `Float8GroupedMMConverter` and `Float8LinearConverter` have the same quantization type `float8`. If we only compare the config types, these legitimate combination would be rejected. What do you think? There are ways to check this by using config's module, but i guess this is hacky.

tianyu-l: Oh good point. One way to do this is to let Float8 converters share an intermediate level of parent, e.g. `Float8Converter`. That seems not necessary for now, so let's keep what you are doing.

PR #2455, opened 2026-02-27
Add chat template support to BaseTokenizer

joecummings:
Post-training workflows (SFT, RL) need to format multi-turn conversations into the model's expected token format before tokenization. Chat templates define how roles, turn boundaries, and special tokens are arranged.

This PR adds `set_chat_template()` and `apply_chat_template()` to `BaseTokenizer`, with automatic loading in `HuggingFaceTokenizer`.

### Template loading

Jinja is the established standard for chat templates. See [HF chat template docs](https://huggingface.co/docs/transformers/main/en/chat_templating).

Following the HF transformers convention (`CHAT_TEMPLATE_FILE` in `transformers/utils/hub.py`), we check two sources with the same priority order:

1. **Standalone `chat_template.jinja` file** (highest priority) -- used by models like GPT-OSS that store complex templates separately
2. **Inline `chat_template` field in `tokenizer_config.json`** (fallback) -- used by most models (Llama3, Qwen3, DeepSeek V3, etc.)

This covers both third-party instruct models (Qwen3-Instruct, GPT-OSS) that already ship with a template, and models pretrained from scratch in TorchTitan where `set_chat_template()` can be called explicitly to add one for post-training.

Note: `HuggingFaceTokenizer` now warns if `tokenizer_config.json` is missing

### TODO

- Wire this into a SFT dataloader under `experiments/sft/` for an e2e training. Build off #2148 

### Test plan

Unit tests covering template rendering, auto-loading from both sources, priority ordering, and error cases.

### DIFF
--- a/torchtitan/components/tokenizer.py
+++ b/torchtitan/components/tokenizer.py
@@ -72,8 +72,12 @@ class BaseTokenizer(ABC, Configurable):
             lstrip_blocks=True,
             extensions=[jinja2.ext.loopcontrols],
         )
-        env.globals["raise_exception"] = raise_exception  # pyrefly: ignore [unsupported-operation]
-        env.globals["strftime_now"] = strftime_now  # pyrefly: ignore [unsupported-operation]
+        env.globals[
+            "raise_exception"
+        ] = raise_exception  # pyrefly: ignore [unsupported-operation]
+        env.globals[
+            "strftime_now"
+        ] = strftime_now  # pyrefly: ignore [unsupported-operation]
         env.filters["tojson"] = tojson
         self._chat_template = env.from_string(template)
 


joecummings:
> 1. Could you please try it with some family of models? For example, i took a quick look at gptoss, and chat_template.jinja is a different file that wouldnt be in tokenizer.cfg
> 
> however, the code does:
> 
> ```
> self._hf_config = self._load_config(
>             os.path.join(tokenizer_path, "tokenizer_config.json")
>         )
> if self._hf_config and "chat_template" in self._hf_config:
>             self.set_chat_template(self._hf_config["chat_template"])
> ```
> 
> so it would silently never trigger `if self._hf_config and "chat_template" in self._hf_config`, because chat_template is NOT in the tokenizer config, its a different file.
> 
> I think that the user experience should be to NOT have to think about it. They just do: "load_tokenizer(model_path), and it takes care of loading jinja
> 
> 2. I see that you did in some places ""{% for msg in messages %}<|im_start|>{{ msg.role }}\n{{ msg.content }}<|im_end|>\n{% endfor %}","
> 
> but some other models may have other fields, e.g. gpt-oss is very different. It stores think and tool call separately. Could you take a look and see if this would work for gpt-oss?

Good call - GPT-OSS is the *only* model that uses a separate Jinja file but I added support for this on the Hugging Face model. In terms of tools, etc the chat template doesn't really care about that. It just parses everything into a format. GPT-OSS doesn't do anything differently here.

felipemello1:
> In terms of tools, etc the chat template doesn't really care about that. It just parses everything into a format. GPT-OSS doesn't do anything differently here.

i am not sure if this is true. If you have something like:
"`<think>...</think>` answer is 42", gptoss will save:
{
	role: assistant
	think: `<think>...</think> `
    content: "answer is 42"
}

if there is a tool call field, i think that it errors if its not well defined in the message when rendering it

gptoss is a different beast, hence why i think its worth testing with a real example

joecummings:
> > In terms of tools, etc the chat template doesn't really care about that. It just parses everything into a format. GPT-OSS doesn't do anything differently here.
> 
> i am not sure if this is true. If you have something like: "`<think>...</think>` answer is 42", gptoss will save: { role: assistant think: `<think>...</think> ` content: "answer is 42" }
> 
> if there is a tool call field, i think that it errors if its not well defined in the message when rendering it
> 
> gptoss is a different beast, hence why i think its worth testing with a real example

`apply_chat_template` renders a string. If there's a `think` field passed in and the Jinja template defines a `think `field then it will render it correctly. Anything here that would not work would be an underlying Jinja issue or a user error. 

Where things get trickier is how to construct the actual tokenized prompts as they are passed to a SFT trainer or a RL trainer. I would like to do the e2e tests with a specific model in those PRs which are downstream of this one.

tests/unit_tests/test_tokenizer.py
```diff
@@ -16,7 +17,38 @@
     instantiate_parametrized_tests,
     parametrize,
 )
-from torchtitan.components.tokenizer import HuggingFaceTokenizer
+from torchtitan.components.tokenizer import BaseTokenizer, HuggingFaceTokenizer
+
+
+CHATML_TEMPLATE = (
```
tianyu-l, line 25: curious what is `CHATML` short for?

torchtitan/components/tokenizer.py
```diff
@@ -40,14 +46,44 @@ def decode(self, *args, **kwargs) -> str:
     def get_vocab_size(self) -> int:
         ...
 
+    def set_chat_template(self, template: str) -> None:
+        """Compile and store a Jinja chat template."""
+        # usort:skip -- usort rewrites this to `import jinja2`, which doesn't load the sandbox submodule
+        import jinja2.sandbox
```
tianyu-l, line 49: do we not need to put this as dependency? I see that CPU test passed.

tests/assets/tokenizer/tokenizer_config.json
```diff
@@ -60,6 +60,7 @@
     "input_ids",
     "attention_mask"
   ],
+  "chat_template": "{% for msg in messages %}<|im_start|>{{ msg.role }}\n{{ msg.content }}<|im_end|>\n{% endfor %}",
```
tianyu-l, line 63: curious what is im_start "im" short for?

torchtitan/components/tokenizer.py
```diff
@@ -40,14 +46,44 @@ def decode(self, *args, **kwargs) -> str:
...
+        def raise_exception(msg):
+            raise jinja2.exceptions.TemplateError(msg)
+
+        env = jinja2.sandbox.SandboxedEnvironment()
+        env.globals[
+            "raise_exception"
+        ] = raise_exception  # pyrefly: ignore [unsupported-operation]
+        self._chat_template = env.from_string(template)
+
+    def apply_chat_template(self, messages: list[dict[str, str]], **kwargs) -> str:
```
tianyu-l, line 84: Do we eventually want to support `tokenize=True` kwarg? Actually if not, why do we put this into tokenizer, instead of MessageParser or something else.

tianyu-l, line 84: why would you put them in BaseTokenizer? The functionality sounds not needed for all tokenizer, e.g. one for pretraining. The `apply_chat_template` name sounds HF.

torchtitan/components/tokenizer.py
```diff
@@ -40,14 +46,44 @@ def decode(self, *args, **kwargs) -> str:
...
+        """Render messages through the Jinja chat template. Returns formatted text.
+
+        Messages should be a list of dicts with "role" and "content" keys, e.g.
+        [{"role": "user", "content": "Hello"}, {"role": "assistant", "content": "Hi"}].
```
tianyu-l, line 88: What is the contract here? If an instruct model is provided, should we always assume its HF checkpoint contains a template that can recognize
- role / content -- if I specify "contents" would it fail?
- user / assistant / system / etc.

What if I provide wild string that a template / tokenizer cannot handle -- is there safety guarantee?

One of the discomfort (limitation in understanding) I have with SFT is that everything seems to follow convention, rathe than clear contract, but there are many "conventions" having subtle differences among them.

tianyu-l (commented):
wow thanks for bringing in structure, I didn't know it can be this elegant.

still having some questions to push for clarity.

tests/assets/tokenizer/tokenizer_config.json
rakkit: i guess `im_start` = `i am start`

tests/unit_tests/test_tokenizer.py
joecummings: Chat Markup Language - https://github.com/openai/openai-python/blob/release-v0.28.0/chatml.md

torchtitan/components/tokenizer.py
joecummings: I believe it's a dependency of PyTorch: https://github.com/pytorch/pytorch/blob/bbf2942a200bdc4f6de3c36bc0f71844b0e99c31/requirements.txt#L12 so we get it for free

tests/assets/tokenizer/tokenizer_config.json
joecummings: Probably `input message start`? https://github.com/openai/openai-python/discussions/2453

But there's no official ruling

torchtitan/components/tokenizer.py
joecummings: Will be covered in follow-up PRs, but yes will either do `tokenize=True` or a separate thing like [`tokenize_messages`](https://github.com/meta-pytorch/torchtune/blob/d0f63bb33d00b8bd3905a010b71d8c6324c2e980/torchtune/modules/transforms/tokenizers/_utils.py#L62). It's definitely helpful if the tokenizer is aware of a message format and can just be passed around for constructing multi-turn or creatively masked inputs in a dataloader.

joecummings: Here was my thinking. The name itself is not HF-specific, but there's a few scenarios for how someone might incorporate post training in Titan:

1) Part of the whole lifecycle. Start with pretrain then sft, rl wtvr. In this case, the tokenizer is likely not an off-the-shelf tokenizer, but they would still need to apply the chat template at the later stages of training. In this case, users could could incorporate their tokenizer onto the `BaseTokenizer` class easily.
2) Post training only. Load from the Hugging Face tokenizer and chat template and go for it. Supported automatically in this PR.

I like the flexibility this separation allows for option 1, but it looks like some of the open model companies who train their own tokenizer [do it via the Hugging Face `tokenizers` library](https://github.com/arcee-ai/trinity-large-tech-report/blob/main/Arcee%20Trinity%20Large.pdf) so I could also see the argument for attaching it directly to the `HuggingFaceTokenizer` class. 

To be clear, I don't feel strongly either way.

torchtitan/components/tokenizer.py
joecummings: * If loading from a Hugging Face post-trained model, we assume that its HF checkpoint contains a chat template. This is technically *not* a contract for uploading a model to Hugging Face Hub, but every major model release since Llama 3 does include one as the easiest way to correctly use the model weights. Any of the models that Titan supports will have chat templates specified for the foreseeable future.
* Providing "contents" would error out. Hugging Face does have a very solid contract for what message format is allowed: https://huggingface.co/docs/transformers/en/chat_content_patterns. For the most part, it follows the OpenAI message format. As part of the data processing step, users would be expected to transform their data into this message format since the tokenizer from Hugging Face would expected this format.
* If you provide a string that the template / tokenizer cannot handle, Jinja will error and point to the offending syntax. I didn't add a test here b/c that's a guarantee covered by Jinja rendering itself, but can do so if you'd prefer. 

> One of the discomfort (limitation in understanding) I have with SFT is that everything seems to follow convention, rathe than clear contract, but there are many "conventions" having subtle differences among them.

Fair enough - things are constantly evolving. The advantage Titan has is that it doesn't claim to support every model or feature OOTB at release time so we can take this case by case. Open discussions about all these nuances will also help I imagine.

tianyu-l: > Providing "contents" would error out.

Who'd error out? I think in this PR we are just transforming the message, not applying tokenization, so we are not using any HF / transformers API at the step of `apply_chat_template`.

If the template is `"chat_template": "{% for msg in messages %}<|im_start|>{{ msg.role }}\n{{ msg.content }}<|im_end|>\n{% endfor %}"`, I guess Jinja can error out if there is no "content". But if I provide both "content" and "contents", would Jinja be able to error out on "contents"? If not, I don't think tokenizer could, either?

tianyu-l: A separate question, from this link https://huggingface.co/docs/transformers/en/chat_templating they have different examples of decoded / pre-tokenizing messages
```
<|system|>
You are a friendly chatbot who always responds in the style of a pirate</s>
<|user|>
How many helicopters can a human eat in one sitting?</s>
<|assistant|>
```

```
<|im_start|>user
Hi there!<|im_end|>
<|im_start|>assistant
Nice to meet you!<|im_end|>
<|im_start|>user
Can I ask a question?<|im_end|>
<|im_start|>assistant
```
The latter seems very close to the example template in this PR. I guess the first can also be achieved with a different chat template, and `encode` / `decode` is an invertible pair of functions.

tianyu-l (approved):
LGTM, one more inline question.

Maybe wait until @felipemello1 also approves.

tests/unit_tests/test_tokenizer.py
wwwjn: Can you add a comment for that?  This abbr is not straightforward to me

joecummings:
> If the template is "chat_template": "{% for msg in messages %}<|im_start|>{{ msg.role }}\n{{ msg.content }}<|im_end|>\n{% endfor %}", I guess Jinja can error out if there is no "content". But if I provide both "content" and "contents", would Jinja be able to error out on "contents"? If not, I don't think tokenizer could, either?

In the latter case, the Jinja would *not* error out on "contents", it would just skip that value in the message. We could do some very simple schema validation but I'd keep it out of `apply_chat_template` since that puts unnecessary restrictions on what kind of formatting people may want to do. Instead we can provide an OOTB `validate_message_schema` method on the `HuggingFaceTokenizer` that directly maps to a valid Hugging Face message schema as defined here: https://huggingface.co/docs/transformers/en/chat_content_patterns.

torchtitan/components/tokenizer.py
```diff
@@ -40,20 +40,71 @@ def decode(self, *args, **kwargs) -> str:
...
+    # Valid message schema following HF chat content patterns:
+    # https://huggingface.co/docs/transformers/en/chat_content_patterns
+    VALID_ROLES = {"system", "user", "assistant", "tool"}
+    VALID_CONTENT_TYPES = {"text", "image", "video", "audio"}
+    ALLOWED_KEYS: dict[str, set[str]] = {
+        "system": {"role", "content"},
+        "user": {"role", "content"},
+        "assistant": {"role", "content", "tool_calls"},
+        "tool": {"role", "content", "name", "tool_call_id"},
+    }
```
felipemello1: i am still a bit skeptical that everything here will work for gpt-oss. Their jinja template has a message.thinking, for example, which is not an allowed key.

I would feel more comfortable if you did the following:
1. Load gptoss with vllm
2. Generate some output
3. Take the output tokens and decode->encode.
4. See if the output matches with vllm output.
5. If it doesnt match, it means that the decode(encode(tokens)) is lossy, e.g. removing important tokens like <think_start>, and secretly ruining training
6. Repeat this for all models supported. If all pass, lets merge!

Image

joecummings: So before the validation, everything passed into the string would be rendered. After the validation, things are less permissive. To fit this mold, I removed the strict validation (like the original PR). 

It's a tradeoff and one in which I'm in favor of following the current example set forth which is very permissive. A future consideration we could do is print out the first rendered sample at the beginning of SFT training which would allow the user to see if something was wrong. I created a jist that proves this render->encode->decode no tokens loss thing works for GPT-OSS.

https://gist.github.com/joecummings/2ff4c75acfe88eb912447328a303c6d1

felipemello1:
```
from transformers import AutoTokenizer

MODEL = "openai/gpt-oss-120b"

tokenizer = AutoTokenizer.from_pretrained(MODEL, trust_remote_code=True)

correct = tokenizer.apply_chat_template(
    [
        {"role": "user", "content": "What is 2+2?"},
        {"role": "assistant", "content": "4.", "thinking": "Simple arithmetic: 2+2=4."},
    ],
    tokenize=False,
    add_generation_prompt=False,
)

wrong = tokenizer.apply_chat_template(
    [
        {"role": "user", "content": "What is 2+2?"},
        {"role": "assistant", "content": "Simple arithmetic: 2+2=4.\n\n4."},
    ],
    tokenize=False,
    add_generation_prompt=False,
)

print("=== WITH thinking FIELD ===")
print(correct)
print("\n=== WITH thinking IN content ===")
print(wrong)
print(f"\n=== SAME? {correct == wrong} ===")
```

```
=== WITH thinking FIELD ===
<|start|>user<|message|>What is 2+2?<|end|>
<|start|>assistant<|channel|>analysis<|message|>Simple arithmetic: 2+2=4.<|end|><|start|>assistant<|channel|>final<|message|>4.<|return|>

=== WITH thinking IN content ===
<|start|>user<|message|>What is 2+2?<|end|>
<|start|>assistant<|channel|>final<|message|>Simple arithmetic: 2+2=4.

4.<|return|>

=== SAME? False ===
```

If you dont put the thinking field, gpt-oss will **NOT** produce the >analysis< portion, which is the portion for thinking. You will be training it in a way differently than it was supposed to. Its a silent footgun. In my opinion, a bug.

cc: @joecummings @tianyu-l

felipemello1: if you ask vllm to generate "what is 2+2", the "analysis" is there. So thats the correct behavior. With the tokenizer above, we would be teaching something different, no?

```
from transformers import AutoTokenizer

MODEL = "openai/gpt-oss-120b"

llm = LLM(model=MODEL, tensor_parallel_size=8, trust_remote_code=True)
tokenizer = AutoTokenizer.from_pretrained(MODEL, trust_remote_code=True)

result = llm.chat([{"role": "user", "content": "What is 2+2?"}], SamplingParams(temperature=0.6, max_tokens=256))[0]
print(tokenizer.decode(result.outputs[0].token_ids, skip_special_tokens=False))
```

```
<|channel|>analysis<|message|>User asks a simple math question: "What is 2+2?" Answer: 4. Possibly add brief explanation.<|end|><|start|>assistant<|channel|>final<|message|>2 + 2 = 4.<|return|>
```

tianyu-l: @felipemello1 thanks for verifying

Would it be correct now if we remove allowed key and don't do validation any more? It feels strange to me because I would have expected that if we specify allowed key which doesn't have "thinking" then if the actual message has "thinking" we will error out, but it seems it runs with silently wrong behavior?

joecummings: This is a really great discussion - I'd like to take a step back b/c I believe I understand where your confusion is coming from @felipemello1. 

Let's break out the different paradigms and build a shared intuition.  

**Pretraining**: No chat template, super easy. I love it.

**SFT**: Likely a chat template involved. Data is structured, the target is clear and dense, and the loss is a simple CE. In this case, the designation of "thinking" being absent from the chat template as wrong is incorrect. Some data includes thinking traces. Some does not. I think everyone would agree that we shouldn't force every sample SFT'd through GPT-OSS to include a thinking tag. 

**RL**: I believe this is where the point you're trying to make is b/c this is where we have generations going back in to train the model. The absolute easiest way to handle this is to pass in raw text to vLLM and have them handle the construction of a chat template and subsequent tokens. This approach can cause issues when you move to multi-step training b/c e.g. you may want to hold on to the entire token history while certain models (ahem, Qwen) automatically removes some history. This would result in lossy tokenization. The solution is to handle the tokenization yourself but here we run into the very issue your script identifies which is that vLLM **may** tokenize differently from what our tokenizer sets up. 

***

1. How does this discussion affect this PR? It doesn't. This PR is very simply adding the ability to utilize a chat template. IMO it is very straightforward and simply a building block needed to do any proper post-training.
2. If we opt to do our own tokenization in RL, what do we do about this potential mismatch between generation tokenization and our own? How do we avoid silent training errors without becoming overly rigid in what we allow users to do? This is an excellent question and, to be a bit hand-wavy, one that we should look to others who have tackled this problem before. I'm happy to take the lead here, but I like to move it off of Github comments.

felipemello1: To be clear, I am **not** worried about RL. I guess what sparked this convo was that in your PR you didn't have "thinking" in the allowed fields. You removed all validation, so i think thats ok now. I missed it. The remaining point of confusion:

>  I think everyone would agree that we shouldn't force every sample SFT'd through GPT-OSS to include a thinking tag.

Steps:

0. User wants to do SFT
1. user loads dataset with `message={role: "assistant", "content": "<think> simple calculation </think> 2+2 is 4"}`
2. They do `tokenizer.set_template(gpt_oss_jinja)`
3. They do `tokenizer.apply_chat_template(message)`
4. If they are **not** aware of GPT-OSS behavior requiring extra 'thinking' field, it **will** render it wrongly.
5. I think what you are saying is that its up to the user to know that they should put the thinking field there.

I could agree with that, although i think that we should detect and warn to save hours of debugging -- its too easy to get that wrong. Do you think that this is the correct behavior?

But, going back to my initial point: if the user does put 'thinking' in the message, previously it would error. Its fixed now, thanks mate!
image

joecummings: The validation fields had already been reverted to the original state of the PR before this particular discussion. 

> I think what you are saying is that its up to the user to know that they should put the thinking field there.

If they want to use GPT-OSS in that particular way, yes. I think adding model-specific comments calling out any interesting chat templating behavior is common practice and doing something like printing out the first few samples when starting could be a good way to alert people to any strange formatting.

PR #2457, opened 2026-02-27
[RFC] Merge simple_fsdp and compiler_toolkit experiments

yiming0416:
## Summary

image

This PR merges the `simple_fsdp` and `compiler_toolkit` experiments into a single unified `graph_trainer` experiment to demonstrate graph-based distributed training in torchtitan through toolkit-style usage of PyTorch's compiler technologies. The goal is to give users more explicit control over the compiler stack in terms of performance, numerics, and debuggability during large-scale distributed training. Two compilation modes are currently supported:
  - **AOT mode** (`--compile.mode aot`): Joint graph export + custom graph passes (former `compiler_toolkit` experiment)
  - **JIT mode** (`--compile.mode jit`): `torch.compile()` with custom backend (former `simple_fsdp` experiment)

## Test plan
  - `integration_test_8gpu_fullmodel.yaml` CI (JIT + AOT integration tests)
  - `test_numerics.py` unit tests (fullmodel matches FSDP2 eager)
  - `test_passes.py` unit tests

## File changes
  
**New files (6)**
  - `graph_trainer/README.md` - Unified docs with JIT/AOT usage examples
  - `graph_trainer/compilation.py` - JIT/AOT compilation dispatcher
  - `graph_trainer/configs.py` - `FullmodelCompileConfig` with mode (jit/aot), backend, passes, joint_passes
  - `graph_trainer/deepseek_v3/config_registry.py` - Merged config registry for DeepSeek V3
  - `graph_trainer/llama3/config_registry.py` - Merged config registry for Llama3
  - `graph_trainer/tests/integration_tests.py` - Unified integration tests
  - `graph_trainer/tests/test_numerics.py` - Numerics tests comparing fullmodel (JIT & AOT) vs FSDP2 eager baseline

**Moved/Renamed files**

 No change at all:
  - `compiler_toolkit/cudagraph.py` -> `graph_trainer/cudagraph.py`
  - `simple_fsdp/deepseek_v3/model.py` -> `graph_trainer/deepseek_v3/model.py`
  - `simple_fsdp/llama3/model.py` -> `graph_trainer/llama3/model.py`
  - `simple_fsdp/simple_fsdp.py` -> `graph_trainer/simple_fsdp.py`

Import and name change only:
  - `compiler_toolkit/trainer.py` -> `graph_trainer/trainer.py`
  - `compiler_toolkit/passes.py` -> `graph_trainer/passes.py`
  - `simple_fsdp/llama3/__init__.py` -> `graph_trainer/llama3/__init__.py`
  - `simple_fsdp/deepseek_v3/__init__.py` -> `graph_trainer/deepseek_v3/__init__.py`
  - `compiler_toolkit/tests/test_passes.py` -> `graph_trainer/tests/test_passes.py`

Refactor and clean-up only
  - `compiler_toolkit/common_utils.py` -> `graph_trainer/common_utils.py`
  - `compiler_toolkit/graph_utils.py` -> `graph_trainer/graph_utils.py`
  - `compiler_toolkit/trainer.py` -> `graph_trainer/trainer.py`
  - `simple_fsdp/backend.py` -> `graph_trainer/jit_backend.py`
  - `simple_fsdp/deepseek_v3/parallelize.py` -> `graph_trainer/deepseek_v3/parallelize.py`
  - `simple_fsdp/llama3/parallelize.py` -> `graph_trainer/llama3/parallelize.py`
  - `simple_fsdp/reshard_after_forward.py` -> `graph_trainer/reshard_after_forward.py`


xmfan (commented):
Looks good. We can converge towards sharing passes progressively as we improve them.  For the naming, `Fullmodel` without the mention of compile is pretty confusing, should be at least fullmodel_compile i'd say.

tianyu-l:
> For the naming, Fullmodel without the mention of compile is pretty confusing, should be at least fullmodel_compile i'd say.

- we should mention compile / graph
- fullmodel sounds limiting, e.g. what if we support compiling the optimizer, etc.

I think we should build this folder to a full-graph-based trainer, as opposed to the eager-based trainer. I forgot the argument of why we shouldn't call it `fullgraph`, but maybe we can call it `graph_trainer` (not training graphs, but training based on graphs, lol).

Let's have discussions, and maybe have a poll.

.github/workflows/integration_test_8gpu_graph_trainer.yaml
```diff
@@ -34,6 +32,8 @@ jobs:
   # Step 1: Dynamically compute the matrix based on conditions
   set-matrix:
     uses: ./.github/workflows/set-matrix.yaml
+    with:
+      runner-cuda: linux.aws.h100.8
```
tianyu-l: h100 CI has long wait time, maybe not good idea to run all tests with h100. Maybe OK if we split them into two sets.

tests/integration_tests/h100.py
```diff
@@ -77,14 +77,16 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
-                    "--module simple_fsdp.deepseek_v3 --config simple_fsdp_deepseek_v3_debugmodel",
+                    "--module fullmodel.deepseek_v3 --config fullmodel_deepseek_v3_debugmodel",
```
tianyu-l: if graph trainer folder has its own h100 test, this entry should be moved over there

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
```diff
@@ -20,37 +20,38 @@
...
 
-def get_transformer_block_buckets(model) -> list[list[str] | str]:
-    module_list = [
-        model.tok_embeddings,
-        [model.norm, model.output],
-    ]
-    for layer_id, transformer_block in model.layers.items():
-        # [TODO](ruisizhang123) add EP support for transformer block bucketing
-        module_list.append(transformer_block)
+def annotate_deepseekv3() -> None:
```
tianyu-l, line 34: need better docstr for this function -- why we need it, why only need it for aot, etc.

torchtitan/experiments/fullmodel/deepseek_v3/parallelize.py
```diff
@@ -81,6 +82,13 @@ def parallelize_deepseekv3(
     ):
         raise NotImplementedError("CP support is only supported for SDPA.")
 
+    mode = getattr(compile_config, "mode", None)
+
+    # AOT mode: annotate and disable eager AC if using fullmodel SAC
+    if mode == "aot":
+        annotate_deepseekv3()
+        maybe_disable_eager_ac(compile_config, ac_config)
```
tianyu-l: explain why disable eager AC if using SAC

torchtitan/experiments/graph_trainer/tests/test_numerics.py
```diff
@@ -3,15 +3,184 @@
...
+    cmd = [
+        sys.executable,
+        "scripts/loss_compare.py",
+        ".",
+        ".",
+        f"--baseline-module={baseline_module}",
+        f"--baseline-config={baseline_config}",
+        f"--test-module={test_module}",
+        f"--test-config={test_config}",
+        "--assert-equal",
```
tianyu-l, line 54: is this bitwise equal?

torchtitan/experiments/graph_trainer/tests/test_numerics.py
```diff
@@ -3,15 +3,184 @@
...
+    def test_dsv3_manual_bucketing_aot_vs_eager(self):
+        self.assertTrue(
+            _run_deepseek_v3_loss_compare(
+                test_options_extra="--compile.mode aot --compile.passes transformer_block_bucketing"
+            ),
+        )
+
+    def test_llama3_jit_vs_eager(self):
+        self.assertTrue(
+            _run_llama3_loss_compare(test_options_extra="--compile.mode jit"),
```
tianyu-l, line 155: are these tests always with aot_eager backend? I vaguely remember that simple fsdp folder still uses inductor, which shouldn't give us identical results right now?

torchtitan/experiments/graph_trainer/tests/test_numerics.py
```diff
@@ -3,15 +3,184 @@
...
+        )
+
+    def test_llama3_auto_bucketing_jit_vs_eager(self):
```
tianyu-l, line 158: wait, what's the eager baseline -- transformer block bucketing? why can we achieve bitwise identity with auto_bucketing?

torchtitan/experiments/graph_trainer/__init__.py
tianyu-l, line 1: do we need this file?

torchtitan/experiments/graph_trainer/compile.py
tianyu-l, line 1: compile vs. compilation, let's try to unify the name at least within this folder. Maybe compile as it's used more universally.

torchtitan/experiments/graph_trainer/jit_backend.py
```diff
@@ -116,9 +122,12 @@ def inductor_transformer_block_bucketing_reordering_pass(
                 gm: torch.fx.Graph,
             ) -> torch.fx.GraphModule:
                 return manual_overlap_bucketing(
-                    gm.owning_module, insert_overlap_deps=True
+                    gm.owning_module,
+                    module_bucket_plans=fsdp_manual_buckets,
+                    insert_overlap_deps=True,
                 )
 
+            torch._inductor.config.allow_buffer_reuse = False
```
tianyu-l, line 132: the position of such flag changed, which I'm not able to verify correctness

torchtitan/experiments/README.md
```diff
@@ -24,8 +24,7 @@ We provide this `experiments/` folder to host experiments that add significant v
...
-| [compiler_toolkit](./compiler_toolkit/) | [![Compiler Toolkit 8 GPU Integration Tests](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_compiler_toolkit.yaml/badge.svg?branch=main)](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_compiler_toolkit.yaml?query=branch%3Amain) | [@SherlockNoMad](https://github.com/SherlockNoMad) [@yiming0416](https://github.com/yiming0416) |
+| [fullmodel](./fullmodel/) | [![Fullmodel 8 GPU Integration Tests](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_fullmodel.yaml/badge.svg?branch=main)](https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_fullmodel.yaml?query=branch%3Amain) | [@ruisizhang123](https://github.com/ruisizhang123) [@tianyu-l](https://github.com/tianyu-l) [@SherlockNoMad](https://github.com/SherlockNoMad) [@yiming0416](https://github.com/yiming0416) |
```
tianyu-l: you can remove my name

torchtitan/experiments/graph_trainer/README.md
tianyu-l, line 1: Seems you removed the "supported features" matrix. IIRC there are things not supported
- EP + AC may already be supported?
- how about EP + PP, shall we explicit add a note on what's not supported and why

torchtitan/experiments/graph_trainer/jit_backend.py
tianyu-l, line 1: I feel it'd helpful if we still have two folders `jit` and `aot` as some structure.
E.g. simple_fsdp.py and backend.py should clearly go to `jit`.
How about cudagraph? Does jit path support it? Let me know if it doesn't make sense.

tianyu-l (commented):
Very nice! Left some comments.

torchtitan/experiments/graph_trainer/jit_backend.py
xmfan: why don't we allow buffer reuse for simplefsdp?

.github/workflows/integration_test_8gpu_graph_trainer.yaml
yiming0416: good point. we only need to run deepseek moe related tests on H100 machines.

torchtitan/experiments/graph_trainer/tests/test_numerics.py
yiming0416: Currently I set the default backend in `FullmodelCompileConfig` to be `aot_eager`. So unless users specifically set `--compile.backend inductor`, it will always use `aot_eager` backend.

torchtitan/experiments/graph_trainer/tests/test_numerics.py
yiming0416: good question. The numerics tests here use `scripts/loss_compare.py` from @fegin 
I just realized that `loss_compare.py` compares the loss by parsing the log with format `{:8.5f}` so it only checks loss equivalence up to 5 decimals. So it DOES NOT check bitwise equivalence. 

The easiest way to check bitwise equivalence would be removing `:8.5f` during metrics logging.

torchtitan/experiments/graph_trainer/__init__.py
yiming0416: Actually we need this. Otherwise I will run into `ModuleNotFoundError: No module named 'torchtitan'` when running unittests

torchtitan/experiments/graph_trainer/tests/test_numerics.py
yiming0416: The eager baseline is FSDP2 with `compile.enable=False`. In terms of bucketing decisions, it should be equivalent to transformer block bucketing.

However, auto bucketing does achieve bitwise identity with the eager baseline (I checked the loss equivalence without any truncation). I recall that on a single node, the bucketing decision doesn't affect numerics - no bucketing, auto bucketing, and manual bucketing all match.

torchtitan/experiments/fullmodel/deepseek_v3/parallelize.py
yiming0416: Oh it meant we should disable eager AC when using graph-based SAC (not using SAC in general)

- Eager AC wraps models with `torch.utils.checkpoint` (maybe compatible with compile)
- Graph-based SAC is a graph pass that annotates node with recompute policies (should always work with compile)

They are essentially doing the same thing via different approaches so they shouldn't be applied at the same time. 

And ideally in the graph-based (compiler-first) world, we should prioritize using graph-based AC.

torchtitan/experiments/graph_trainer/jit_backend.py
yiming0416: > the position of such flag changed, which I'm not able to verify correctness

Any flag start with `torch._inductor.config` should only affect inductor backend, that's why I moved it under the `compile_config.backend == "inductor"` branch. 

Previously the config flags were scattered around at multiple places so I did some clean up.

torchtitan/experiments/graph_trainer/jit_backend.py
yiming0416: > simple_fsdp.py and backend.py should clearly go to jit

I don't think `simple_fsdp.py` is tied to jit, as it's a compiler-friendly FSDP that could be used in both jit and aot paths.

> How about cudagraph? Does jit path support it? 

cudagraph could be supported in jit but we don't have it yet in torchtitan. I know there is a PR to add this support in https://github.com/pytorch/torchtitan/pull/2444

> have two folders jit and aot as some structure

I'm a bit hesitant on this. I think `jit` and `aot` should really just be a graph capture API difference, while other components (like simple_fsdp and graph passes) should be shared.

torchtitan/experiments/fullmodel/deepseek_v3/parallelize.py
tianyu-l: SAC has two things to config
- the region to do recompute (which we need to save all the inputs of), e.g. transformer block
- the ops / tensors to save / recompute

In graph-based SAC, are you still relying on eager AC wrapped regions? Or does graph-based AC control the regions by itself? I feel it's the former which makes me think it's not fair to call it graph-based.

torchtitan/experiments/graph_trainer/jit_backend.py
tianyu-l: makes sense

torchtitan/experiments/fullmodel/deepseek_v3/parallelize.py
yiming0416: The current graph-based SAC doesn't rely on regions from eager AC or control regions by itself. 

It simply traverses all nodes in the joint graph and annotate them with recompute policies based on the `DEFAULT_SAC_SAVE_OPS` list.

tianyu-l: Correct me if I'm wrong:
Without regions, in order to recompute `layer_32.mlp.h`, you need to start all the way from raw inputs. The "checkpoint" in AC means that we store checkpoints in the middle of the model so that in backward we can resume from somewhere.

yiming0416: In graph-based SAC, nodes annotated with `MUST_SAVE` are the "checkpoints". 

The min-cut partitioner constructs the backward graph to start from those saved tensors based on the annotation, so we don't need to start all the way from raw inputs for recompute.

torchtitan/experiments/graph_trainer/compile.py
yiming0416: changed from `compilation.py` to `compile.py`

.github/workflows/integration_test_8gpu_graph_trainer.yaml
yiming0416: now we have two sets of CI tests
- integration_test_8gpu_graph_trainer.yaml
- integration_test_8gpu_graph_trainer_h100.yaml

torchtitan/experiments/graph_trainer/README.md
yiming0416: Added back the "Composabilty" section. Also added more rows for features to be supported in the future

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
yiming0416: Added docstring. 
Also, it’s not necessarily only for AOT (so I removed the if branch). For debugging-purpose annotations, they should also work with JIT. For regional inductor purpose annotations, it’s just a no-op for JIT.

yiming0416:
Changed the experiment folder name to `graph_trainer`

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
```diff
@@ -20,37 +20,45 @@
...
 
-from ..simple_fsdp import data_parallel, MixedPrecisionPolicy
+def annotate_deepseekv3() -> None:
+    """Attach annotations to FX graph nodes with ``torch.fx.traceback.annotate_fn``
 
+    - Expert Parallel (EP) annotations: Tags "dispatch", "combine", and "compute"
+      regions in MoE for debugging purposes.
+    - Flex attention annotation: Tags FlexAttentionWrapper.forward with
+      {"compile_with_inductor": "flex_attention"} so the compiler can apply
+      regional inductor pass based on the annotation.
```
tianyu-l: > For regional inductor purpose annotations, it’s just a no-op for JIT.

Can you add this to docstring as well?

Also curious why only AOT needs it?

torchtitan/experiments/graph_trainer/llama3/parallelize.py
```diff
@@ -164,32 +166,14 @@ def parallelize_llama(
...
-            fullgraph=True,
-        )
+    # Apply compilation based on mode
+    model = apply_compile(
+        model,
+        compile_config=compile_config,
+        parallelism=parallelism,
+        parallel_dims=parallel_dims,
+        dump_folder=dump_folder,
+        transformer_block_buckets=get_transformer_block_buckets(model),
```
tianyu-l: can this be put into `_apply_jit_compile` instead of passing in from `apply_compile`?

torchtitan/experiments/graph_trainer/compile.py
```diff
@@ -0,0 +1,177 @@
...
+    joint_graph_builder,
+    make_compiler_with_passes,
+)
+from torchtitan.experiments.graph_trainer.jit_backend import (
+    get_compile_backend_with_passes,
+)
+from torchtitan.tools.logging import logger
+
+
+def _get_reshard_policy(
```
tianyu-l: if this is shared by every parallelize.py, maybe we should put it in torchtitan/distributed/fsdp.py

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -0,0 +1,106 @@
...
+|Activation Offloading| 🚧 |
+|Mixed Precision Training| ✅ |
+|Tensor Parallelism| ✅ |
+|Context Parallelism| ✅ |
+|Pipeline Parallelism| ✅ |
+|Distributed Checkpointing| ✅ |
+|CUDA Graphs| ✅ |
+|Float8 Training| 🚧 |
+|Expert Parallelism| ✅ |
+|Expert Parallelism + Activation Checkpointing| 🚧 |
```
tianyu-l, line 89: should be ready after https://github.com/pytorch/torchtitan/pull/2406 so need to update depends on which lands first @IvanKobzarev

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -0,0 +1,106 @@
...
+|Meta Initialization| ✅ |
+|Activation Checkpointing| ✅ |
+|Activation Offloading| 🚧 |
+|Mixed Precision Training| ✅ |
+|Tensor Parallelism| ✅ |
+|Context Parallelism| ✅ |
+|Pipeline Parallelism| ✅ |
+|Distributed Checkpointing| ✅ |
+|CUDA Graphs| ✅ |
+|Float8 Training| 🚧 |
```
tianyu-l: this should be done thanks to @pianpwk who did verification

we can add a test on H100

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
yiming0416: Oh because now only AOT path supports regional inductor compilation.

torchtitan/experiments/graph_trainer/compile.py
yiming0416: There is no `torchtitan/distributed/fsdp.py`, are you thinking about `torchtitan/distributed/utils.py`?

Also `_get_reshard_policy` is not being used outside this experiment. So I think fine to leave it here for now.

torchtitan/experiments/graph_trainer/README.md
yiming0416: yep I am aware of that PR. 
hoping that we can prioritize landing this refactor PR.

torchtitan/experiments/graph_trainer/compile.py
tianyu-l: > Also _get_reshard_policy is not being used outside this experiment.

It is not but it should? as the same code is used everywhere
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/parallelize.py#L335-L347
We can call it `get_fsdp_reshard_after_forward_policy`

> There is no torchtitan/distributed/fsdp.py

We should create one.

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
yiming0416: I mean the annotation still happens under JIT (so the annotation itself is not a no-op).

We don't support regional inductor in JIT so the annotation won't be used anywhere. But there is no need to guard the annotation under AOT mode

torchtitan/experiments/graph_trainer/compile.py
tianyu-l: Doesn't have to be done in this PR.

PR #2458, opened 2026-02-27
[torchtitan][resubmit] experimenting new replicate integration with torchtitan

anshul-si:
**Summary**
                                                                                                                                           
  - Replaces the old apply_ddp (DDP-based replication) with a new apply_replicate function that uses
  torch.distributed._composable.replicate with per-module wrapping and MixedPrecisionPolicy support — mirroring the same wrapping pattern
  as FSDP (tok_embeddings, transformer blocks, norm+output, full model)
  - Updates all model parallelization files (llama3, llama4, qwen3, deepseek_v3, gpt_oss, VLM, transformers_modeling_backend, RL trainer)
  to use apply_replicate instead of apply_ddp
  - Removes the old "DDP has not supported > 1D parallelism" restrictions, since replicate integrates with the composable parallelism stack
  - Disables automatic gradient division for replicate modules (same as FSDP) so gradient scaling is handled by the training loop
  - Updates maybe_enable_amp in distributed utils to recognize dp_replicate_enabled for mixed precision handling

 **This is a resubmit PR of (https://github.com/pytorch/torchtitan/pull/1714)**

* __->__ #2458


fegin (approved):
Did a very light review since this is a resubmit PR. Please explicitly mention that this is a resubmit PR to the summary and title. 

linter failure is real, please fix it. 

We should also wait for the integration tests are green though our CI has some issues now. 

Also, do we even have CI for replicate case? Maybe it worths to add one for llama3. Can be done in another PR.

PR #2459, opened 2026-02-27
Update cuda version for installing PyTorch pre-built wheels 

wwwjn:
Attempt to upgrade cuda version for CI

### DIFF
--- a/.github/workflows/set-matrix.yaml
+++ b/.github/workflows/set-matrix.yaml
@@ -65,9 +65,9 @@ jobs:
             "name": "cuda",
             "runner": "${CUDA_RUNNER}",
             "gpu-arch-type": "cuda",
-            "gpu-arch-version": "12.6",
+            "gpu-arch-version": "12.8",
             "docker-image": "torchtitan-ubuntu-20.04-clang12",
-            "index-url": "https://download.pytorch.org/whl/nightly/cu126"
+            "index-url": "https://download.pytorch.org/whl/nightly/cu128"
           }
           EOF
           )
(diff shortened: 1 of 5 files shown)


.github/workflows/set-matrix.yaml
```diff
@@ -65,9 +65,9 @@ jobs:
             "name": "cuda",
             "runner": "${CUDA_RUNNER}",
             "gpu-arch-type": "cuda",
-            "gpu-arch-version": "12.6",
```
wwwjn, line 68: Not 100% sure if we should update this. Is this field related to which CUDA version does our CI machine installed?

wwwjn: UDPATE: From the CI log, when we run `nividia-smi`, it shows CUDA version 13.0 so it's fine to update this field

wconstab: can you just ask claude to figure out what exactly this field "gpu-arch-version" is used for?

I am guessing the update is fine, but it's a misleading name for a field becuase 12.6 / 12.8 are not gpu-arch versions, those would be like SM 90 or something.

wconstab (approved):

.github/workflows/set-matrix.yaml
wwwjn: From claude: It represents the
  CUDA toolkit version (or ROCm version) and is used by the PyTorch test-infra to:

  - Select the correct pip index URL for installing PyTorch nightly (e.g., cu126 →
  https://download.pytorch.org/whl/nightly/cu126)
  - Set up the correct CUDA/ROCm runtime environment in the Docker container

tianyu-l (approved):

PR #2464, opened 2026-03-02
[RL] Use sum digits task in simple_grpo

daniellepintz:
Use a task that looks like this in `simple_grpo.py`:
```
User: What is the total digit sum of [12, 345, 67]?
Assistant: Break each number into digits:
12 → 1, 2
345 → 3, 4, 5
67 → 6, 7
Sum all digits: 1 + 2 + 3 + 4 + 5 + 6 + 7 = 28
[ANSWER] 28
```

This demonstrates the model can really learn and improve accuracy and formatting in 10 steps.

Results from 10 steps of GRPO:
```
Pre-training:  Accuracy=40% (8/20) Format=55% (11/20)
Post-training: Accuracy=55% (11/20) Format=90% (18/20)
```

Also changed some logging to debug mode to avoid a lot of noise on every step and so we can clearly see step level metrics like loss, avg. reward, etc.

### DIFF
--- a/torchtitan/experiments/rl/unified/simple_grpo.py
+++ b/torchtitan/experiments/rl/unified/simple_grpo.py
@@ -65,7 +65,7 @@ class RLTrainer(Configurable):
         operations and bitwise-reproducible forward/backward passes."""
 
         num_episodes_per_step: int = 5
-        """Number of prompts to generate per training step."""
+        """Number of episodes to create before every training step."""
 
         log_samples: bool = False
         """Log first completion per episode during training and eval."""


torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -133,6 +126,7 @@ def __init__(
             distributed_executor_backend="external_launcher",
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=config.enforce_eager,
+            disable_log_stats=True,
```
daniellepintz, line 140: needed to avoid noisy vllm logging on each step

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -31,7 +31,7 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
         num_steps=10,
         batch_invariant_mode=True,
         trainer=PolicyTrainer.Config(
-            optimizer=OptimizersContainer.Config(lr=1e-6),
+            optimizer=OptimizersContainer.Config(lr=2e-6),
```
daniellepintz, line 34: needed for better learning

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -48,7 +48,7 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
-            gpu_memory_limit=0.5,
+            gpu_memory_limit=0.4,
```
daniellepintz: needed to avoid oom

joecummings: Why would the memory be higher in this PR versus main?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
-                f"max={metrics['logprob_diff_max']:.4e}"
+                f"Correct: {correct_count}/{total_count} | "
+                f"Avg tokens: {avg_len:.0f} | "
+                f"Time: generate={t_generate:.1f}s train={t_train:.1f}s sync_weights={t_sync_weights:.1f}s"
             )
-            logger.debug(f"  Sample: {metrics['sample_completion']}...")
+
+            if self.config.log_samples:
+                for ep, answer in zip(scored_episodes, self.expected_answers):
+                    idx = 0  # Log first completion per prompt
```
joecummings: nit: no need to specify this separately

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
+            all_token_lens = [
+                len(c.token_ids) for ep in scored_episodes for c in ep.completions
+            ]
+            avg_len = sum(all_token_lens) / len(all_token_lens)
 
             logger.info(
-                f"\nStep {step:3d} | Loss: {metrics['loss']:.4f} | "
+                f"Step {step:2d} | Loss: {metrics['loss']:.4f} | "
                 f"Reward: {metrics['reward_mean']:+.3f} | "
-                f"Logprob diff: mean={metrics['logprob_diff_mean']:.4e}, "
```
wwwjn, line 204: Can you keep the log prob diff logging between trainer and generator

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
+
             # 4. Sync weights back to generator (all TP ranks)
+            t0 = time.perf_counter()
             weight_mesh = self.trainer_actor.get_weights.call().get()
             weights = {
                 gpu: weight_mesh.item(gpus=gpu)
                 for gpu in range(self.trainer_world_size)
             }
             self.generator.update.call(metrics["policy_version"], weights).get()
+            t_sync_weights = time.perf_counter() - t0
```
wwwjn: Let's remove the naive logging for now, and keep the main loop clean until we have better observability solution.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
                 f"Reward: {metrics['reward_mean']:+.3f} | "
-                f"Logprob diff: mean={metrics['logprob_diff_mean']:.4e}, "
-                f"max={metrics['logprob_diff_max']:.4e}"
+                f"Correct: {correct_count}/{total_count} | "
+                f"Avg tokens: {avg_len:.0f} | "
+                f"Time: generate={t_generate:.1f}s train={t_train:.1f}s sync_weights={t_sync_weights:.1f}s"
             )
-            logger.debug(f"  Sample: {metrics['sample_completion']}...")
+
+            if self.config.log_samples:
```
wwwjn: Again, this part is tight for this task and also only for logging purpose. Can we move it into reward function? And this i

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
+                logger.info("!" * 80)
                 break
 
-        logger.info("\n" + "=" * 80)
+        # Post-training evaluation
         logger.info("RL Training complete")
+        logger.info("Evaluating post-training performance...")
+        post_eval = await self.evaluate()
+
+        logger.info("=" * 80)
```
wwwjn, line 315: These loggings can be moved into evaluate()

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -64,6 +69,12 @@ class Config(Configurable.Config):
         """Enable batch-invariant mode for deterministic NCCL collective
         operations and bitwise-reproducible forward/backward passes."""
 
+        num_prompts_per_step: int = 5
+        """Number of prompts to generate per training step."""
```
wwwjn: Does this field mean "Number of prompts to generate per training step", similar idea to `batch_size` in pre-training? When parallelism is applied, is this field referring to local batch size on a single GPU, or a global batch size on all GPUs

Let's have a more indicated name

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -103,19 +114,9 @@ async def setup(self):
...
-            ("The first president of the United States was", "washington"),
-            ("The chemical symbol for water is", "h2o"),
-            ("The largest planet in our solar system is", "jupiter"),
-        ]
-        self.prompt_texts = [p[0] for p in prompts_with_answers]
-        self.expected_answers = [p[1] for p in prompts_with_answers]
-
-        logger.debug(f"Loaded {len(self.prompt_texts)} prompts")
+        # Task specification for generating prompts
+        self.task_spec = SumDigitsSpec(seed=42)
```
wwwjn: Now the SumDigitsSpec is hard coded in the main RL loop, can we make it pluggable / configurable as all other components?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
             model_spec=config.model_spec,
             model_path=config.hf_assets_path,
             batch_invariant_mode=config.batch_invariant_mode,
-            prompt_texts=self.prompt_texts,
         )
 
         # Initialize generator with trainer weights.
         self.generator.update.call(0, initial_weights).get()
 
+    def _generate_prompts(self) -> tuple[list[str], list[str]]:
```
wwwjn: This seems more like a dataloader function, can we move dataloader related functionality and configs to the `datasets.py`

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
+
+        Generates on eval prompts, scores them, and reports accuracy.
+
+        Args:
+            num_samples: Number of eval prompts to generate
+
+        Returns:
+            Dict with accuracy, correct, total, format_rate
+        """
+        eval_spec = SumDigitsSpec(seed=99)  # Different seed from training
```
wwwjn: We can have a general evaluate() with pluggable tasks?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
+        # Score: check first completion per episode
+        correct = 0
+        format_ok = 0
+        for episode, task in zip(episodes, eval_tasks):
+            text = episode.completions[0].text
+            extracted = extract_answer(text)
+            is_correct = extracted == task.correct_answer
+            has_tag = bool(re.search(r"\[ANSWER\]", text))
+            correct += int(is_correct)
+            format_ok += int(has_tag)
```
wwwjn, line 198: The format check and correct check should be the job of Grader?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
             weights = {
                 gpu: weight_mesh.item(gpus=gpu)
                 for gpu in range(self.trainer_world_size)
             }
             self.generator.update.call(metrics["policy_version"], weights).get()
+            t_sync_weights = time.perf_counter() - t0
+
+            # Count correct rewards from scored episodes
+            all_rewards = [c.reward for ep in scored_episodes for c in ep.completions]
+            correct_count = sum(1 for r in all_rewards if r > 0)
```
wwwjn: This part of calculation is tight to sum digits task, and it's purely for logging purpose now. It should not exist in main loop

torchtitan/experiments/rl/sum_digits.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class SumDigitsSpec:
+    """Generates sum-of-digits tasks with sequences of 2-4 integers (10-99)."""
+
+    def __init__(self, seed: int = 42):
+        self.seed = seed
+        self._rng = random.Random(seed)
+
+    def generate_task(self) -> Task:
```
wwwjn: This is a dataloader for RL

wwwjn (commented):
Thanks, from what I read, there are 2 parts to define a Task:
1. dataset / dataloder
2. Reward function to be plugged into Grader. 
We can make these 2 parts configurable and pluggable 

And dataloader needs to be a separate actor , and spawned on trainer_mesh IIUC? This part might need some design cause right now the data loading part seems not compatible with parallelism. Then DP / TP is applied on both trainer / generator, multiple generator will consume the same data now?

torchtitan/experiments/rl/unified/simple_grpo.py
joecummings: The functionality should be very self contained for now. I'd love to take a stab at a more comprehensive dataloading refactor in a follow up to this PR.

torchtitan/experiments/rl/sum_digits.py
joecummings: For the purposes of this PR, yes it's operating very much like a dataloader. I'd imagine we'll move to a more generic dataloading that then be be constructed from many different tasks and data sources.

torchtitan/experiments/rl/unified/simple_grpo.py
tianyu-l: Agreed, especially considering @felipemello1 is working on this.

torchtitan/experiments/rl/sum_digits.py
```diff
@@ -0,0 +1,112 @@
...
+User: What is the total digit sum of [12, 345, 67]?
+Assistant: Break each number into digits:
+12 → 1, 2
+345 → 3, 4, 5
+67 → 6, 7
+Sum all digits: 1 + 2 + 3 + 4 + 5 + 6 + 7 = 28
+[ANSWER] 28"""
+
+
+class SumDigitsSpec:
```
tianyu-l: Spec is a bit general, do we consider rename e.g. to `SumDigitsTaskSpec` / `SumDigitsTaskGenerator` or whatever makes more sense?

torchtitan/experiments/rl/sum_digits.py
```diff
@@ -0,0 +1,112 @@
...
+        numbers = [self._rng.randint(10, 99) for _ in range(n)]
+        answer = sum(int(d) for num in numbers for d in str(num))
+        question = f"What is the total digit sum of {numbers}?"
+        return Task(question=question, correct_answer=answer)
+
+    def get_system_prompt(self) -> str:
+        return _SYSTEM_PROMPT
+
+
+def sum_digits_reward_function(
```
tianyu-l: Does it make sense to couple this into the TaskSpec, in the sense that a task specification should specify how the results should be evaluated? Is there a case which we shouldn't?

torchtitan/experiments/rl/sum_digits.py
tianyu-l, line 1: I'd suggest putting it under `rl/unified` as it's not used by the other to-be-deprecated folder.

torchtitan/experiments/rl/unified/config_registry.py
daniellepintz: good point, this was needed in a previous iteration of this PR; reverting to 0.5

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Done!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Yea I think it would be ideal if we leave dataloading abstractions to a later PR

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Done!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: I moved this code to the Task definition, in a method called `evaluate_completion`, since it's a little different than the reward fn which the Grader handles. but pls lmk if you see a better way!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Hmm I dont see a good way to do this, since we want it to print something like this at the end of training for comparison:
```
Pre-training:  Accuracy=40% (8/20) Format=55% (11/20)
Post-training: Accuracy=55% (11/20) Format=90% (18/20)
```
Please lmk if you see a better way!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: this refers to the number of prompts sent to generator per step globally. the global generator batch size is: `this number * num_samples_per_prompt`

currently we aren't using data parallelism so this is fine

this is similar to what we do on current main:

https://github.com/pytorch/torchtitan/blob/de329200bd1fd05a67d7db1de0ebb46797130586/torchtitan/experiments/rl/unified/simple_grpo.py#L108-L114

but yes we need a follow up (I think @joecummings mentioned he is looking into this) to create a dataloader that handles data parallelism.

meanwhile I will rename this to `num_episodes_per_step` but pls lmk if anyone has a better name!

torchtitan/experiments/rl/sum_digits.py
daniellepintz: Renamed to `SumDigitsTaskSpec`!

torchtitan/experiments/rl/sum_digits.py
daniellepintz: yes, definitely makes sense, done!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Makes sense, moved to a fn `compute_step_metrics` on TaskSpec!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Sure, done!

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -144,7 +145,7 @@ async def setup(self):
         self.grader = trainer_mesh.spawn(
             "grader",
             Grader,
-            trivial_reward_function,
+            sum_digits_reward_function,
```
acisseJZhong: +1 to the comment above, let's make this configurable?

torchtitan/experiments/rl/unified/simple_grpo.py
acisseJZhong: this logic seems general?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,60 +162,183 @@ async def setup(self):
...
+                    logger.info(f"       {comp.text[:200].replace(chr(10), ' ')}")
 
             # Check for divergence
             if not torch.isfinite(torch.tensor(metrics["loss"])):
-                logger.debug("\n" + "!" * 80)
-                logger.debug("ERROR: Loss is NaN/Inf! Training diverged.")
-                logger.debug("!" * 80)
+                logger.info("!" * 80)
+                logger.info("ERROR: Loss is NaN/Inf! Training diverged.")
+                logger.info("!" * 80)
```
acisseJZhong, line 307: now that i see logger debug and info is mixed; can we have a rule?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -213,7 +210,7 @@ async def generate(self) -> list[Episode]:
                     )
                 )
```
felipemello1, line 222: do you understand what this code is doing? i.e.

1. all_outputs = N_prompts
but then we do "for sample in output.outputs:"

2 why do we have N samples per output?

3. and do we add N completions per episode?

torchtitan/experiments/rl/unified/simple_grpo.py
felipemello1: hmm, i think we have room here to be more precise with names.

we have:
episode
step
prompt
generate

for example, we don't "generate prompts", right?

Maybe take a look at the arg names we use in forge, and try to keep it similar

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,35 +157,101 @@ async def setup(self):
...
 
         # Initialize generator with trainer weights.
         self.generator.update.call(0, initial_weights).get()
 
+    def _generate_prompts(self) -> tuple[list[str], list[str]]:
+        """Generate a batch of prompts and expected answers from the task spec."""
+        prompt_texts = []
+        expected_answers = []
+        for _ in range(self.config.num_episodes_per_step):
+            task: Task = self.task_spec.generate_task()
```
wwwjn: The name "generate_task" is really cofusing

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: The name "_generate_prompts" is really confusing, it's basically a data loading function. Let's have a better name

torchtitan/experiments/rl/unified/sum_digits.py
```diff
@@ -0,0 +1,147 @@
...
+User: What is the total digit sum of [12, 345, 67]?
+Assistant: Break each number into digits:
+12 → 1, 2
+345 → 3, 4, 5
+67 → 6, 7
+Sum all digits: 1 + 2 + 3 + 4 + 5 + 6 + 7 = 28
+[ANSWER] 28"""
+
+
+class SumDigitsTaskSpec:
```
wwwjn: This function doesn't use the TaskSpec portocal

torchtitan/experiments/rl/unified/task_spec.py
```diff
@@ -0,0 +1,64 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Protocol
+
+import torch
+
+
+@dataclass
+class Task:
```
wwwjn: This is a single sample from dataset / or generated, why call it "Task"?

torchtitan/experiments/rl/unified/task_spec.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def generate_task(self) -> Task: ...
+
+    def get_system_prompt(self) -> str: ...
+
+    def create_eval_instance(self) -> TaskSpec:
+        """Return a new instance for evaluation with a different seed."""
+        ...
+
+    def reward_function(
```
wwwjn: I think we would need to move reward function out of TaskSpec. For example, for the sum digit task, different reward function can be plugged in. We can reward only correct answer but not format, or we can reward both correctness and format

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Can you put a comment here to indicate we explicitly disabled the log from vllm? It would help if someone want to debug vllm

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: In GRPO algorithm, for each prompt, we generate `n_samples_per_prompt` completions which is used for averaging reward and calculate relative rewards. 

So, if we have 5 prompts per RL step, and `n_samples_per_prompt=4`, we will generate 5 `Episode` using VLLMGenerator. Within each `Episode`, the length of `Episode.completions` will be 4 (because we put the 4 completion for the single prompt into a list).

felipemello1: oohhh! I see. Ok, i think that this is not how this is usually defined. My understanding is that 4 outputs = 4 Episodes. If we want to call that a Group, then we do GroupedEpisode = List[Episode]

so if you have 5 prompts, and 4 generations per prompt, we have 5x4 = 20 episodes.

Does this make sense? It doesnt need to be addressed in this PR, but should be part of some refactor, so nomenclature matches what libraries usually do

felipemello1: external reference: https://github.com/thinking-machines-lab/tinker-cookbook/blob/2b76deade6e80f1fbaaf0778e584e8d663eb461f/tinker_cookbook/rl/rollouts.py#L97

wwwjn: > i think that this is not how this is usually defined

I don't know what's usually defined, but this version is my first attempt to define this episode, we should align with how other ppl use it. Let's refactor and put a TODO (I can fix this later)

torchtitan/experiments/rl/unified/task_spec.py
felipemello1, line 1: do we need an abstraction for tasks? I think that abstractions should be introduced when we have >1 example and we are confident about what we want to do. This `task_spec` is something that i didnt see in other frameworks. Since its early days, i would say just create fucntions in `simple_grpo` specific to `sum_digits`, like how we hardcoded `gsm8k` to `forge`. Once we know what we want to do, we can create new abstractions that fit multiple cases.

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: yeah good point, totally agree! fixing!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Do you mean a rule for when to use debug and when to use info? In this particular case, I thought that if the loss is NaN it is important enough that we should always log it even when not running with debug mode

torchtitan/experiments/rl/unified/task_spec.py
daniellepintz: Agreed, removing the abstractions :)

torchtitan/experiments/rl/unified/task_spec.py
daniellepintz: removed the dataclass and simplified :)

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Fixed!

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,35 +157,133 @@ async def setup(self):
             model_spec=config.model_spec,
             model_path=config.hf_assets_path,
             batch_invariant_mode=config.batch_invariant_mode,
-            prompt_texts=self.prompt_texts,
         )
 
         # Initialize generator with trainer weights.
         self.generator.update.call(0, initial_weights).get()
 
+    def _load_prompts(self) -> tuple[list[str], list[str], list[str]]:
```
tianyu-l: prefer to not have this function, given how short it is

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -161,35 +157,133 @@ async def setup(self):
...
+                self.expected_answers,
+                self.questions,
+            ) = self._load_prompts()
+
             # Fully sync RL loop with separate scoring step
             # 1. VLLMGenerator produces episodes (one per prompt, without rewards)
             # TODO: Create a queue to use all episode from all GPUs
-            episodes = self.generator.generate.call().get().item(gpus=0)
+            episodes = (
+                self.generator.generate.call(self.prompt_texts).get().item(gpus=0)
```
tianyu-l: what's the semantics here?

- We are sending the same `self.prompt_texts` to every single generator rank.
- We expect TP to be used, so that's expected.
- We expect DP to be not used, o/w there will be redundant generation, not because of `num_samples_per_prompt` but because of DP degree.

If so, can we add an assertion in Generator parallelism config that data parallel degrees are 1?

Also, since we only have one generator, why do we need this seed field at all? Can we delete for now?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/unified/actors/generator.py#L92
cc @wwwjn

wwwjn: > https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/unified/actors/generator.py#L92
cc @wwwjn 

The `seed` field would be useful to test bit-wise identity mode (in my next PR)

tianyu-l: @wwwjn just to confirm, where do we have randomness in generator today -- is it from vllm engine having random rollout?

wwwjn: We don't have randomness in VLLMGenerator() wrapper, vllm engine could have randomness in it. I'm not sure about all details in the vllm engine's randomness, but I can think of examples, eg, when not using greedy decoding,

tianyu-l: got it, let's keep the seed in vllmGenerator

daniellepintz: > We are sending the same self.prompt_texts to every single generator rank.
We expect TP to be used, so that's expected.
We expect DP to be not used, o/w there will be redundant generation, not because of num_samples_per_prompt but because of DP degree.

yes, that's correct! Added an assertion, good call

wwwjn: I have a followup comment here:
- `episodes = (self.generator.generate.call(self.prompt_texts).get().item(gpus=0)` 

This line is still getting the Generator-0's episodes only and drop Generator-1's episode. Even if no DP is applied between generator-0 and generator-1, generator-1's generation is also wasted.

This is a known issue from original script, @daniellepintz can you help fix this issue here? A simple fix is:
- Given we have strong assumption trainer and generator doesn't have DP now, we can just call .get(gpus=0) and .get(gpus=1) separately, concat them together and send to trainers

daniellepintz: My understanding is the current implementation is correct, when we are not doing DP -- my understanding is vLLM will use both generator rank 0 and generator rank 1 for generation since the model is sharded across them, but will produce the same results on both ranks; so we only need to collect the generations from one rank.

We can see this when running inference_example.py, we also use generator tp=2 and the exact same output is printed twice

PR #2466, opened 2026-03-02
Performance Improvement: Implement custom op for MoE token combination

hirschsn:
In Qwen3-30B-A3B, we get a large Magma SGEMM or Cutlass SIMT kernel on B200 for the token combination bmm backward from the Autograd.
This kernel seems to be very inefficient, taking ~6-9ms for local_batch_size=16, where a memory-bound kernel should take something like 1ms. The Torch Inductor is not able to remove this kernel by default.

This commit reformulates the problem into an explicit outer product, which seems to lower to a much better kernel for B200s than the one produced by the Autograd. Together with torch.compile enabled for MoE (which is not the default), this change saves about 4.5 ms per backward layer.

Before:
grafik

After:
grafik

Average duration of the backwards CompiledFxGraph (this is not the whole backward, FSDP communication is missing):
Before: 62.6875 ms
After: 56.167 ms

Averages generated with:
```
select avg(dur/1000/1000) as avg_ms from slices
 where category = 'gpu_user_annotation'
   and name = '## Call CompiledFxGraph None ##'
   and ts >= <timestamp-where-backward-pass-begins>;
```

*This commit does not change the performance of Qwen3-30B-A3B in eager mode or in the current default Torch Titan setting, where the grouped_gemms in MoE are exempt from compilation (the called kernels are different, but the runtime is practically the same).*

### DIFF
--- a/tests/unit_tests/test_moe_token_combine.py
+++ b/tests/unit_tests/test_moe_token_combine.py
@@ -1,0 +1,48 @@
+import pytest
+import torch
+from torchtitan.models.common.moe.moe import ExpertCombineFunction
+
+NUM_TOKENS = 42
+# Use some Qwen3-30B-A3B dimensions for testing.
+DIM = 2048
+NUM_EXPERTS = 128
+TOP_K = 8
+
+def original_token_combine(s: torch.Tensor, R: torch.Tensor) -> torch.Tensor:
+    """Original token combine implementation using bmm, for testing against the custom autograd function."""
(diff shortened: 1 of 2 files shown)


tianyu-l (changes requested):
We should investigate instead of landing this solution.

Is this the same issue as https://github.com/pytorch/torchtitan/issues/2225#issuecomment-3807002038?

hirschsn:
Indeed, yes it is the same issue. Thanks for pointing me to this discussion!

hirschsn:
Superseded by this PR: https://github.com/pytorch/pytorch/pull/176552

PR #2467, opened 2026-03-02
Small fixes to `inference_example.py` and README

daniellepintz:
inference_example.py
  - Made seed conditional — only passes seed to EngineArgs when not None, this avoids an error when seed is 
  None (which it currently is in the config)
  -  Overrides model_spec.parallelize_fn with `parallelize_qwen3` as is also done in simple_grpo.py (this avoids an error about missing has_position_id)

README
  - Hardcoded `--nproc_per_node=2` instead of <world_size> placeholder, because the current code only works with `--nproc_per_node=2`; this because tensor_parallel_degree in the VLLMGenerator config is set to 2, and they must match
  - Added a NOTE that --nproc_per_node must match the tensor_parallel_degree in the VLLMGenerator config

### DIFF
--- a/torchtitan/experiments/rl/unified/inference_example.py
+++ b/torchtitan/experiments/rl/unified/inference_example.py
@@ -36,7 +36,8 @@ def generate():
     gen_config = config.generator
     model_path = config.trainer.hf_assets_path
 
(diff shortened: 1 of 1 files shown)


Lucaskabela (approved):
Was just looking at this - thanks for the change :)

PR #2469, opened 2026-03-02
[ez] [rl] fix deprecation warning and add more descriptive error for checkpoint error

joecummings:
As title

SPMD module will contain all monarch/spmd stuffs after 0.0.3

And Monarch would give a super unhelpful error saying the Actor method was invalid even though the TRUE error was that the checkpoint path was empty.

### DIFF
--- a/torchtitan/experiments/rl/unified/simple_grpo.py
+++ b/torchtitan/experiments/rl/unified/simple_grpo.py
@@ -32,7 +32,7 @@ from dataclasses import dataclass, field
 
 import torch
 from monarch.actor import this_host
-from monarch.utils import setup_env_for_distributed
+from monarch.spmd import setup_torch_elastic_env_async
 from torchtitan.config import Configurable
 from torchtitan.config.manager import ConfigManager
 from torchtitan.experiments.rl.unified.actors.generator import VLLMGenerator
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -28,7 +28,7 @@
 import torch
 
 from monarch.actor import this_host
-from monarch.utils import setup_env_for_distributed
+from monarch.spmd import setup_torch_elastic_env_async
```
acisseJZhong, line 35: what's the difference between these two functions? FWIW @allenwang28 points me to this util.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -124,7 +124,7 @@ async def setup(self):
         )
 
         # Set up distributed env vars so that actors are connected via c10d
-        await setup_env_for_distributed(
+        await setup_torch_elastic_env_async(
             trainer_mesh,
             master_addr="localhost",  # TODO: figure out what to set
             master_port=29501,  # TODO: figure out what to set
```
acisseJZhong: @allenwang28 do you know what's a better way of handling this?

torchtitan/experiments/rl/unified/simple_grpo.py
joecummings: It's the same function - they renamed it in everything > 0.3 and if you look at the logs they give you a deprecation warning telling you to switch to the spmd module.

joecummings: I was just annoyed by the deprecation warning lol

wwwjn:
My experience is Monarch will hang there and finally gives a "Actor method invalid" error, when the True error is very straightforward. Eg, a syntax error, a missing path error. Is there a way to let Monarch produce the True error directly?

PR #2470, opened 2026-03-02
[mxfp8 training] refactor to use single MXFP8 training converter

danielvegamyhre:
## Summary
- torchao is unifying the mxfp8 linear and grouped_mm training abstractions in https://github.com/pytorch/ao/pull/3968
- This PR updates torchtitan's integration to be current with this refactor

TL;DR of torchao refactor

- Previous:
    - Separate configs for linear and grouped_mm
    - linear uses MXLinear module swap, while grouped_mm uses tensor subclass weight wrapper with grouped_mm op override
- New:
    - Single unified MXFP8TrainingOpConfig for both linear and grouped_mm
    - Single unified MXFP8TrainingWeightWrapperTensor subclass with op overrides for both linear and grouped_mm ops. Consistent approach for both, no more module swaps. 

## Tests
 
Updated deepseek_v3_16b config with different fqn options and validated only the intended layers were using MXFP8 via logging.

Example:

```python
        model_converters=ModelConvertersContainer.Config(
            converters=[
                MXFP8Converter.Config(fqns=["moe.experts"]),
            ],
            print_after_conversion=True,
        ),
```

### DIFF
--- a/torchtitan/components/quantization/mx.py
+++ b/torchtitan/components/quantization/mx.py
@@ -5,9 +5,8 @@
 # LICENSE file in the root directory of this source tree.
 
 from dataclasses import dataclass, field
-from functools import partial
 from importlib.util import find_spec
-from typing import Any, ClassVar, Literal
+from typing import ClassVar, Literal
 
 import torch.nn as nn
 from torchtitan.components.quantization import (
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
@tianyu-l for review

torchtitan/components/quantization/mx.py
```diff
@@ -5,137 +5,36 @@
...
+    """
+    Wraps the weight tensors of target nn.Linears or 3D nn.Parameters with a tensor subclass
+    that overrides grouped_mm and linear ops, dispatching to autograd functions that implement
+    dynamic quantization and MXFP8 grouped_m/linear ops, based on the given config.
+    """
 
     @dataclass(kw_only=True, slots=True)
-    class Config(QuantizationConverter.Config):
-        _quantization_type: ClassVar[str] = "mx"
+    class Config(Configurable.Config):
```
tianyu-l: revert this change?

torchtitan/components/quantization/mx.py
```diff
@@ -5,137 +5,36 @@
...
+        _quantization_type: ClassVar[str] = "mxfp8"
 
-        recipe_name: Literal["mxfp8"] = "mxfp8"
+        recipe_name: Literal["mxfp8_rceil"] = "mxfp8_rceil"
         """
-        Quantization recipe name for grouped GEMMs. Options: ["mxfp8"]
+        Quantization recipe name for grouped GEMMs. Options: ["mxfp8_rceil"]
```
tianyu-l, line 37: please explain what `rceil` means here

torchtitan/components/quantization/mx.py
```diff
@@ -169,15 +68,25 @@ def __init__(
...
-        self.moe_fqns = config.fqns
-        if self.moe_fqns:
+        # If using mxfp8 on grouped_mm, we need to pad the token group sizes to a multiple of 32
+        # TODO: remove once (1) hybridEP landed and we require it to use mxfp8 on experts, or (2) torchao
+        # has an efficient token group padding implementation for mxfp8 grouped_mm
+        def _has_experts_fqn(fqns: list[str]) -> bool:
+            for fqn in fqns:
+                if "shared_experts" in fqn:
+                    continue
+                if "experts" in fqn or "moe" in fqn:
```
tianyu-l: why this `or`? this would return True on router.gate?

danielvegamyhre: true, updated.

torchtitan/components/quantization/mx.py
danielvegamyhre: Updated docstring with explanation. it is the rounding mode used when calculating the e8m0 scale factors, which involves casting fp32 -> e8m0 and you have to decide how to round.

torchtitan/components/quantization/mx.py
danielvegamyhre: that's weird, must have happened when resolving the merge conflict. fixed.

PR #2471, opened 2026-03-03
[ROCm] Use nightly ROCm Pt 7.1 whl

jagadish-amd:
Unify CI torch install logic across CUDA/ROCm by using matrix.torch-version with a safe fallback when empty.
Use env variable HIPBLASLT_TENSILE_LIBPATH to pick the right search path for tensile libs.

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -40,7 +40,6 @@ def main() -> None:
         # pyrefly: ignore [missing-attribute]
         trainer = config.build()
 
-
         # pyrefly: ignore [missing-attribute]
         if config.checkpoint.create_seed_checkpoint:
             assert (


torchtitan/train.py
```diff
@@ -40,6 +41,18 @@ def main() -> None:
...
+        if torch.version.hip is not None and torch.cuda.is_available():
+            try:
+                warm_x = torch.randn(1, 1, 8, device="cuda", dtype=torch.bfloat16)
+                warm_w = torch.randn(8, 8, device="cuda", dtype=torch.bfloat16)
+                _ = F.linear(warm_x, warm_w, None)
+                torch.cuda.synchronize()
+                logger.info("ROCm warmup linear completed")
+            except RuntimeError as error:
+                logger.warning("ROCm warmup linear failed: %s", error)
+
```
tianyu-l: sorry but this change is not acceptable to train.py

acisseJZhong: we could also keep the temp fix in pytorch if it's too ugly in titan https://github.com/pytorch/pytorch/pull/175834/changes. But I would like to know if we can still figure out the root cause for this bug, instead of using warmup @jagadish-amd

jagadish-amd:
cc @jeffdaily @lizamd

jagadish-amd:
I am able to run the tests with HIPBLASLT_TENSILE_LIBPATH env set, without any workaround. There seems to be issue with tensile files discovery (confirmed though hipblaslt logs and strace) leading to invalid status even though the returned algo is able to proceed further. Setting the right path with env variable HIPBLASLT_TENSILE_LIBPATH fixes the issue. 
I can try the same here, need to add the env variable in workflow files.
cc @jeffdaily

.github/workflows/integration_test_8gpu_autoparallel.yaml
```diff
@@ -66,13 +66,21 @@ jobs:
...
         pip config --user set global.progress_bar off
 
-        if [ "${{ matrix.gpu-arch-type }}" = "rocm" ]; then
-          python -m pip install --force-reinstall --pre \
-            "torch==${{ matrix.torch-version }}" \
-            --index-url ${{ matrix.index-url }}
-        else
-          python -m pip install --force-reinstall --pre \
-            torch --index-url ${{ matrix.index-url }}
+        TORCH_SPEC="torch"
```
tianyu-l, line 69: maybe worth to refactor so that multiple tests reuse the setup code here, can be in its own PR

PR #2472, opened 2026-03-03
[rl] non-colocated trainer / generator

joecummings:
As title - this is the schema we want to target

**What does this PR do?**
* Creates a ToyProvisioner that splits trainer and generator on different GPUs (within the same node)
* Updates the weight sync to gather full state dict
* Spawns all three actors on different meshes

**Why non-colocated?** While it's true that colocated trainer/generator is the fastest because you can drop and reload weights in the same memory, it requires smaller models and usually necessitates that the sharding strategy between trainer and generator is the same. These are significant limitations for training RL at scale, so we should pursue the harder and more realistic option: non-colocated. 

**Why do we need a ToyProvisioner?** This is truly as simple as possible. You can see a more complex provisioner system in forge but this is all we need for now. Monarch doesn't automatically stop you from allocating as many things as you want starting from rank 0, so this plugs into the mechanism to block that from happening.

**Why do you gather full state dict now?** This allows us to have different sharding patterns. The resharding is very naive but this is where torchstore will plug in soon.

### DIFF
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass, field
 
 import torch
 from monarch.actor import Actor, endpoint
-from torch.distributed.tensor import DTensor
+from torch.distributed.tensor import distribute_tensor, DTensor
 from torchtitan.config import Configurable
 from torchtitan.config.configs import ParallelismConfig
 from torchtitan.experiments.rl.unified.plugin import (
@@ -76,7 +76,7 @@ class VLLMGenerator(Actor, Configurable):
         model_dtype: str = "bfloat16"
         """Data type for model weights, passed directly to vLLM (auto, float16, bfloat16, float32)."""
 
-        gpu_memory_limit: float = 0.5
+        gpu_memory_limit: float = 0.9
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
         enforce_eager: bool = True
@@ -233,29 +233,22 @@ class VLLMGenerator(Actor, Configurable):
 
         Args:
             version: New policy version number
-            state_dict: Per-rank state dicts keyed by GPU index,
-                e.g. {0: state_dict_gpu0, 1: state_dict_gpu1}
+            state_dict: Full (unsharded) state dict with plain tensors.
         """
-        # Extract this rank's state dict from the per-rank dict
-        rank = int(os.environ.get("LOCAL_RANK", 0))
-        local_state_dict = state_dict[rank]
-
-        # Convert plain local tensors (TP shards from trainer) to DTensors
-        # matching the vLLM model's sharding layout. The trainer exports
-        # weights via to_local() which strips DTensor metadata.
+        # Reshard full tensors to match this generator's DTensor layout
         model_state_dict = dict(self._get_model().model.state_dict())
-        for name, tensor in local_state_dict.items():
+        for name, tensor in state_dict.items():
             if name in model_state_dict and isinstance(model_state_dict[name], DTensor):
                 if isinstance(tensor, DTensor):
                     continue
                 target_dtensor = model_state_dict[name]
-                local_state_dict[name] = DTensor.from_local(
+                state_dict[name] = distribute_tensor(
                     tensor.to(target_dtensor.device_mesh.device_type),
                     device_mesh=target_dtensor.device_mesh,
                     placements=target_dtensor.placements,
                 )
 
-        load_weights = self._get_model().load_weights_from_state_dict(local_state_dict)
+        load_weights = self._get_model().load_weights_from_state_dict(state_dict)
         self.policy_version = version
         logger.debug(
             f"Updated weights into vLLM engine model. "
--- a/torchtitan/experiments/rl/unified/actors/trainer.py
+++ b/torchtitan/experiments/rl/unified/actors/trainer.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass, field
 from typing import Any
 
 import torch
+import torch.distributed as dist
 import torch.distributed.checkpoint as dcp
 from monarch.actor import Actor, endpoint
 from torch.distributed._tensor import DTensor
@@ -131,7 +132,14 @@ class PolicyTrainer(Actor, Configurable):
         self.policy_version = 0
         self.generator: Any | None = None
 
-        logger.debug("PolicyTrainer initialized")
+        # Data parallelism: determine this rank's shard of the batch.
+        self.dp_size = self.parallel_dims.dp_replicate * self.parallel_dims.dp_shard
+        self.dp_rank = dist.get_rank() // self.parallel_dims.non_data_parallel_size
+        self.dp_enabled = self.parallel_dims.dp_enabled
+
+        logger.debug(
+            f"PolicyTrainer initialized (dp_rank={self.dp_rank}, dp_size={self.dp_size})"
+        )
 
     def _load_initial_hf_weights(self, model, checkpoint_path: str) -> None:
         """Load model weights from HF checkpoint using DCP and state_dict_adapter.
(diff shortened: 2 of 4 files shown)


torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -80,7 +80,7 @@ class Config(Configurable.Config):
         model_dtype: str = "bfloat16"
         """Data type for model weights, passed directly to vLLM (auto, float16, bfloat16, float32)."""
 
-        gpu_memory_limit: float = 0.5
+        gpu_memory_limit: float = 0.9
```
joecummings, line 79: This is the default in vLLM - we should not change their defaults unless we have a really good reason.

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -117,44 +154,34 @@ async def setup(self):
...
-        # Set up distributed env vars so that actors are connected via c10d
-        await setup_env_for_distributed(
-            trainer_mesh,
-            master_addr="localhost",  # TODO: figure out what to set
-            master_port=29501,  # TODO: figure out what to set
+        generator_mesh = this_host().spawn_procs(
+            per_host={"gpus": self.generator_world_size},
+            bootstrap=provisioner.allocate(self.generator_world_size),
         )
+        grader_mesh = this_host().spawn_procs()
```
joecummings, line 175: Grader mesh just colocated on rank 0 b/c it's a very simple proc

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -117,44 +154,34 @@ async def setup(self):
...
         trainer_mesh = this_host().spawn_procs(
-            per_host={"gpus": self.trainer_world_size}
+            per_host={"gpus": self.trainer_world_size},
+            bootstrap=provisioner.allocate(self.trainer_world_size),
         )
-
-        # Set up distributed env vars so that actors are connected via c10d
-        await setup_env_for_distributed(
-            trainer_mesh,
-            master_addr="localhost",  # TODO: figure out what to set
```
joecummings, line 125: Can remove this b/c it'll be set automatically and processes don't need to be on separate ports

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: Didn't understand the argument -- vLLM alone shouldn't care about training, but RL should. Are you saying in non-colocated it should be set to 0.9? Then next question is why not 1.0, or 0.99?

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -48,10 +48,10 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
-            gpu_memory_limit=0.5,
             enforce_eager=True,
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
```
tianyu-l, line 51: can we set the trainer to have FSDP 2?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -40,6 +43,37 @@
 logger = logging.getLogger(__name__)
 
 
+class ToyProvisioner:
```
tianyu-l: Despite being toy, it doesn't have to appear in name. It can be in docstring.
```suggestion
class Provisioner:
```

torchtitan/experiments/rl/unified/actors/generator.py
joecummings: This is the dataclass for instantiating a vLLM instance. It's largely set up to mirror the relevant params from the vLLM `LLM` class instantiation. As such, we should keep the default the same so users can expect the same default for vLLM and our "wrapper" of vLLM. 

In colocated, we couldn't do that b/c we needed enough memory for both models, but now we can. If need be users, can easily override this but the default is the same one they would expect for vLLM>

Screenshot 2026-03-04 at 10 56 28 AM

^ From the docs

As for why it's `0.9`, the best I could find is this discussion: https://discuss.vllm.ai/t/what-does-gpu-memory-utilisation-include/1651

tianyu-l: > (e.g., CUDA graphs, kernels, fragmentation, and temporary allocations)

makes sense to me

tianyu-l (approved):

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -41,16 +41,16 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
                 seq_len=4096,
             ),
             parallelism=ParallelismConfig(
-                tensor_parallel_degree=2,
-                data_parallel_replicate_degree=1,
+                tensor_parallel_degree=1,
+                data_parallel_shard_degree=2,
```
daniellepintz: Can we do this, because we are passing the same data to every rank of trainer right?

https://github.com/pytorch/torchtitan/blob/14c33ef40ff6255a463bc4467c1e8706c329b80d/torchtitan/experiments/rl/unified/simple_grpo.py#L309

daniellepintz: also dont we need to include `data_parallel_shard_degree` in world size calculation?

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -40,6 +43,37 @@
 logger = logging.getLogger(__name__)
 
 
+class Provisioner:
+    """Allocates non-overlapping GPU ranges from a fixed pool.
```
daniellepintz: nit but can we make the docstring a bit more clear, I didn't immediately understand this : )

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -40,6 +43,37 @@
...
+
+    Each call to `allocate` reserves the next contiguous block and
```
daniellepintz: contiguous block of GPUs right?

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: so if we don't have this Provisioner, Monarch will just allocate everything on ranks 0 - N? If so, can we document this here

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: maybe the word contiguous throws me off a bit, maybe consecutive is better? super nit haha

joecummings: In this case, I think contiguous is the right word

torchtitan/experiments/rl/unified/config_registry.py
joecummings: Yeah I updated this!

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: should we call it GPUProvisioner? totally up to you

torchtitan/experiments/rl/unified/config_registry.py
joecummings: Okay upon some further work, I think the changes needed to properly support anything besides TP are out of scope for a PR just purports to be doing non-colocated trainer / generator. 

Therefore, I'm going to do that as a quick follow up.

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -265,31 +270,38 @@ async def step(self, episodes: list[Episode]) -> dict:
         advantages = torch.cat(all_advantages)
         all_rewards_tensor = torch.tensor(all_rewards)
 
-        # Compute reference log probs using frozen ref_model
+        # Shard flattened completions across DP ranks so each rank processes
```
wwwjn: nit: I'll have to refactor `Episodes` and don't use nested  completions

PR #2477, opened 2026-03-04
Update loss_compare script

yiming0416:
`loss_compare.py` seems to be outdated after the config system refactor.

We no longer use `TRAIN_FILE` and `CONFIG_FILE`. Should be switched to `MODULE` and `CONFIG`

### DIFF
--- a/scripts/loss_compare.py
+++ b/scripts/loss_compare.py
@@ -25,23 +25,22 @@ Example usages:
 
 2. Compare losses between two commits with custom config and options:
    loss_compare.py main my_branch \
-       --baseline-config='./custom.py' \
+       --baseline-config='llama3_8b' \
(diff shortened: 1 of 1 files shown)


scripts/loss_compare.py
```diff
@@ -25,7 +25,7 @@
 
 2. Compare losses between two commits with custom config and options:
    loss_compare.py main my_branch \
-       --baseline-config='./custom.py' \
+       --baseline-config='llama3_8b' \
```
tianyu-l, line 28: also showcase how to use --baseline-module added in this PR?

yiming0416:
> the CI doesn't catch the regression is because our CI use the default value, which doesn't require the changes here.

Yes exactly.

Issue #2446, opened 2026-02-27
qwen3 TP + CP bug


tianyu-l:
### Bug description

I tried but hit the following error. It's likely because CP's assumption doesn't work with how TP interacts with DTensor.

```
  traceback : Traceback (most recent call last):
    File "/home/lty/local/pytorch/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 362, in wrapper
      return f(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/torchtitan/torchtitan/trainer.py", line 809, in train
      self.train_step(data_iterator)
    File "/home/lty/local/torchtitan/torchtitan/trainer.py", line 719, in train_step
      loss = self.forward_backward_step(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/torchtitan/torchtitan/trainer.py", line 667, in forward_backward_step
      pred = model_parts[0](inputs, **extra_inputs, **extra_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/pytorch/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
      return self._call_impl(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/pytorch/torch/nn/modules/module.py", line 1885, in _call_impl
      return inner()
             ^^^^^^^
    File "/home/lty/local/pytorch/torch/nn/modules/module.py", line 1833, in inner
      result = forward_call(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/torchtitan/torchtitan/models/common/decoder.py", line 135, in forward
      h = layer(h, self.freqs_cis, attention_masks, positions)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/pytorch/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
      return self._call_impl(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/lty/local/pytorch/torch/nn/modules/module.py", line 1885, in _call_impl
      return inner()
[...]

PR #2480, opened 2026-03-04
Qwen3 TP + CP with local_map region

pianpwk:
* __->__ #2480

Fixes https://github.com/pytorch/torchtitan/issues/2446 by using local_map region for CP


torchtitan/models/qwen3/parallelize.py
```diff
@@ -40,9 +40,30 @@
...
+    kwargs: dict,
+) -> tuple[tuple, dict]:
+    """Convert DTensor args to local tensors before the CP attention hook.
+
+    When TP uses use_local_output=False, q/k/v arrive at inner_attention as
+    DTensors on the TP mesh. The CP hook expects plain tensors, so we convert
+    them here. This hook must be registered before the CP hook.
+    """
+    new_args = tuple(
+        arg.to_local() if isinstance(arg, DTensor) else arg for arg in args
```
tianyu-l: No we shouldn't do this. `to_local()` should always come with `grad_placement`, which is hard to make generic.

Maybe you can use `local_map` explicitly around CP region inside Attention.

torchtitan/models/qwen3/parallelize.py
```diff
@@ -40,9 +41,39 @@
...
+        super().__init__()
+        self._inner = inner_attn
+        # q, k, v are sharded on heads dim (dim=1 after transpose in GQAttention)
+        # Each PlacementType is a Sequence[Placement], e.g. (Shard(1),)
+        shard_heads = (Shard(1),)
+        self._local_map_fn = local_map(
+            inner_attn,
+            # Single output: wrap in tuple so local_map sees Tuple[PlacementType]
+            out_placements=(shard_heads,),
+            in_placements=(shard_heads, shard_heads, shard_heads),
```
acisseJZhong, line 69: what about placements for other args passed into inner attention? do we need to consider them? https://github.com/pytorch/torchtitan/blob/c7378f666c6ca0465287505e786346c87a16d996/torchtitan/models/common/attention.py#L511

torchtitan/models/qwen3/parallelize.py
```diff
@@ -139,6 +170,20 @@ def parallelize_qwen3(
             attn_backend,
         )
 
+    # When both TP and CP are enabled, wrap inner_attention with local_map
+    # to properly handle DTensor <-> local tensor conversion at the TP/CP
+    # boundary. TP's use_local_output=False produces DTensor q/k/v, but CP
+    # hooks expect plain tensors. local_map handles this with correct
+    # grad_placements for backward pass.
+    if parallel_dims.tp_enabled and parallel_dims.cp_enabled:
```
acisseJZhong, line 180: do we need to check if `use_local_output=False` for TP?

pianpwk:
sorry, please ignore the PR in its current state, everything's from Claude. will put more thought into this

torchtitan/models/qwen3/parallelize.py
```diff
@@ -40,9 +41,39 @@
 )
 from torchtitan.models.qwen3.model import Qwen3Model
 from torchtitan.protocols.model_converter import ModelConvertersContainer
+
 from torchtitan.tools.logging import logger
 
 
+class _LocalMapInnerAttention(nn.Module):
```
wwwjn, line 49: This is also useful for torchtitan + vllm as we have to manually patch vllm's attention kernel now which also requires a plain tensor

fatih-uzlmz:
Hey @pianpwk thanks for getting this started! I was actually looking into this exact issue earlier today and trying to solve it locally. If you don't mind, I’d love to try taking a crack at implementing the local_map fix that @tianyu-l requested in a new PR. Lmk if that works for you :)

wwwjn:
@pianpwk  Do you wanna refine this PR and we can land it as a workaround before full DTensor mode?

torchtitan/models/qwen3/parallelize.py
pianpwk: looks like they're non-Tensor args that local_map will flatten away, while BlockMask/VarlenMetadata isn't supported yet; qwen CP only supports SDPA backend. Adding a note there

torchtitan/models/qwen3/parallelize.py
pianpwk: hmm I can't find a clean way to check this

torchtitan/models/qwen3/parallelize.py
acisseJZhong: I believe CP with Flex attention will be enabled by @fegin soon. Not sure about what's the gap for CP + varlen attention.

torchtitan/models/qwen3/parallelize.py
acisseJZhong: can we move this function outside qwen3 as it is useful in general?

torchtitan/models/qwen3/parallelize.py
acisseJZhong: ic, maybe we can just assume we always use full DTensor for TP, then we need to land https://github.com/pytorch/torchtitan/pull/2149 first?

torchtitan/models/qwen3/parallelize.py
```diff
@@ -140,6 +172,30 @@ def parallelize_qwen3(
...
+    # boundary. TP's use_local_output=False produces DTensor q/k/v, but CP
+    # hooks expect plain tensors.
+    # TODO(pianpwk): remove this once full DTensor lands
+    if parallel_dims.tp_enabled and parallel_dims.cp_enabled:
+        tp_mesh = parallel_dims.get_mesh("tp")
+        # pyrefly: ignore [missing-attribute, not-callable]
+        for block in model.layers.values():
+            attn = block.attention  # pyrefly: ignore [missing-attribute]
+            # Workaround: cuDNN SDPA backward has a stride mismatch bug with CP.
+            # Exclude cuDNN until PyTorch fix lands.
```
acisseJZhong, line 186: do you know what PR is the fix? can you link it in comment?

torchtitan/models/qwen3/parallelize.py
```diff
@@ -140,6 +172,30 @@ def parallelize_qwen3(
...
+            attn.inner_attention.sdpa_backends = (
+                [  # pyrefly: ignore [missing-attribute]
+                    SDPBackend.FLASH_ATTENTION,
+                    SDPBackend.MATH,
+                ]
+            )
+            attn.inner_attention = (
```
acisseJZhong, line 193: wont this override L187?

torchtitan/models/qwen3/parallelize.py
pianpwk: Still an open issue: https://github.com/pytorch/pytorch/issues/176915#issuecomment-4026752935

torchtitan/models/qwen3/parallelize.py
pianpwk: I think this should be fine? L187 was meant to override this class' backend attribute, which is wrapped over: https://github.com/pytorch/torchtitan/blob/b0a7d322cd04c7272f3901a09315deeb60d2dd38/torchtitan/models/common/attention.py#L169-L190

acisseJZhong: ic, my bad. It was a little bit confusing since we are overriding `attn.inner_attention`. But indeed the backend attribute for the real `attn.inner_attention`(L195) should be updated.

torchtitan/models/qwen3/parallelize.py
```diff
@@ -40,9 +42,39 @@
 )
 from torchtitan.models.qwen3.model import Qwen3Model
 from torchtitan.protocols.model_converter import ModelConvertersContainer
+
 from torchtitan.tools.logging import logger
 
 
+class _LocalMapInnerAttention(nn.Module):
+    """Wraps inner_attention with local_map for TP + CP composition.
```
tianyu-l, line 50: Instead of making it only supports TP+CP, can we make it generic and support arbitrary tensor / dtensor inputs -- e.g. every inner_attention wrapper (flex, varlen, sdpa) should inherit an InnerAttentionWrapper which does this local_map treatment. cc @acisseJZhong this aligns with our idea of using local_map for attention in full DTensor.

acisseJZhong: just synced with @fegin we should do this to make local_map work with every attention.

acisseJZhong: > every inner_attention wrapper (flex, varlen, sdpa) 

Does SDPA support Dtensor input already?

tianyu-l: > Does SDPA support Dtensor input already?

Yes, so it's not absolutely necessary. But unifying the interface with local_map sounds cleaner in torchtitan.

PR #2482, opened 2026-03-04
[RL] Some cleanups around `hf_assets_path`

daniellepintz:
* The `hf_assets_path` arg on the PolicyTrainer config was not used; removing
* In the config_registry `hf_assets_path` on the RLTrainer config was set to `/data/users/jianiw/model/qwen3-0.6b`; changing to set it to the path in the README (`torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B`)
* In the README remove `--hf_assets_path=<path_to_model_checkpoint>` from the command in step 6 to reduce some friction for the user having to find the correct path; in step 4 we tell the user in which path to store the checkpoint. Include a note in case someone wants to store the checkpoint in a different path

### DIFF
--- a/torchtitan/experiments/rl/unified/README.md
+++ b/torchtitan/experiments/rl/unified/README.md
@@ -56,8 +56,11 @@ torchrun --nproc_per_node=2 torchtitan/experiments/rl/unified/inference_example.
 
 6. Run simple GRPO RL loop
 ```bash
-python torchtitan/experiments/rl/unified/simple_grpo.py --module rl.unified --config rl_grpo_qwen3_0_6b --hf_assets_path=<path_to_model_checkpoint>
+python torchtitan/experiments/rl/unified/simple_grpo.py --module rl.unified --config rl_grpo_qwen3_0_6b
 ```
+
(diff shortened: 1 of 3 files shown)


torchtitan/experiments/rl/unified/README.md
```diff
@@ -56,8 +56,11 @@ torchrun --nproc_per_node=2 torchtitan/experiments/rl/unified/inference_example.
 
 6. Run simple GRPO RL loop
 ```bash
-python torchtitan/experiments/rl/unified/simple_grpo.py --module rl.unified --config rl_grpo_qwen3_0_6b --hf_assets_path=<path_to_model_checkpoint>
+python torchtitan/experiments/rl/unified/simple_grpo.py --module rl.unified --config rl_grpo_qwen3_0_6b
 ```
+
+**NOTE:** If you downloaded your HF model to a different path than the one in step 4, specify it in your command with `--hf_assets_path=<path_to_model_checkpoint>`.
```
wwwjn, line 62: Thanks, this is good

PR #2484, opened 2026-03-04
[1/N] Add lora adapters for finetune

mori360:
### Summary
  - Add LoRAConverter model converter that applies low-rank adaptation (LoRA) to all nn.Linear layers in a model
  - LoRA uses dynamic subclass creation (__class__ swap) to wrap existing Linear layers, preserving compatibility with any nn.Linear subclass (e.g.,
  FakeQuantizedLinear)
  - Base model weights are frozen (requires_grad=False), only LoRA adapter weights (lora_a, lora_b) are trainable
  - Patches init_weights on the model to reinitialize LoRA adapters during the meta-device → real-device init flow
  - Add converter ordering validation: quantization must come before LoRA
  - Add llama3_debugmodel_lora debug config and unit tests for freeze/trainability and training correctness
https://github.com/pytorch/torchtitan/issues/2515 

### Test Plan
 - test_lora_freeze_and_trainability: check LoRA params have requires_grad=True, base params (original weights/biases) have requires_grad=False.
 - test_lora_trains_base_frozen, after training, asserts every base param is bitwise identical to its snapshot. Asserts at least one LoRA param has changed.
 - some other tests with results below to show the memory performance and loss convergence.

### Details
1. Enable LoRA via ModelConverter
LoRA is integrated using the existing ModelConverter protocol, making it consistent with other model transformations (e.g., quantization). Users can enable LoRA by adding "lora" to the model.converters config:
```
[model]
converters = ["lora"]
```

2. In-place Linear Replacement with LoRALinear Subclass
The implementation uses dynamic class creation via create_lora_linear() to wrap any nn.Linear subclass with LoRA adapters:

- Dynamic subclassing: `create_lora_linear(parent_cls) `creates a new class that inherits from the original Linear class, preserving custom forward logic (e.g., RowwiseParallel, ColwiseParallel)
- In-place modification: Instead of creating new modules, we create subclass preserving existing weights
- Frozen base weights: All base model parameters are frozen (requires_grad=False), only LoRA adapters are trainable
- Composability with quantization: we currently has float8 low-precision quantization as another choice in converters. Here we will conduct convert in order that convert to `Float8Linear` first, then in lora converter to convert from `Float8Linear` to `LoRAFloat8Linear` by `create_lora_linear` described above. # converters = ["quantize.linear.float8", "lora"]
    LoRA architecture is quite simple, that only adds lora_out to the base_out. So that as long as it has `base_out = super().forward(x)`, lora subclass won't touch the parent forward, so we could compose them together

### Parallelism Compatibility
FSDP (Data Parallel) — Works out of the box. LoRA adapters (lora_a, lora_b) are registered as child nn.Modules of the wrapped linear, so fully_shard() treats them like any other submodule. Base weights are frozen (requires_grad=False) and LoRA weights are trainable — FSDP only all-gathers/reduces gradients for trainable parameters, so no wasted communication.

Pipeline Parallel (PP) — Works. LoRA doesn't change the model's module structure or the number of layers, so pipeline stage splitting is unaffected. Each stage's linear layers independently get LoRA adapters during convert().

Context Parallel (CP) — Works. CP operates on the attention module's sequence dimension, not on linear layer internals. LoRA adapters participate in the same forward pass as the base linear.

Tensor Parallel (TP) — Not supported yet. PyTorch's ColwiseParallel._partition_linear_fn (in
torch/distributed/tensor/parallel/style.py) uses named_parameters(recurse=True), which yields dotted names like
lora_a.weight from child modules. register_parameter() rejects names containing ".", causing a KeyError. 

### Test Plan
Tested with Llama3 debug model on 8 GPU
- lora has loess memory cost
- lora could work with float8

without lora converter
```
[rank0]:[titan] 2026-02-02 14:41:32,656 - root - INFO - step:  1  loss:  8.2000  grad_norm:  1.3955  memory:  1.01GiB(1.07%)  tps: 5,479  tflops: 0.39  mfu: 0.04%
[rank0]:[titan] 2026-02-02 14:41:33,139 - root - INFO - step:  5  loss:  5.3874  grad_norm:  2.3446  memory:  1.06GiB(1.11%)  tps: 343,441  tflops: 24.59  mfu: 2.49%
[rank0]:[titan] 2026-02-02 14:41:33,396 - root - INFO - step: 10  loss:  4.1315  grad_norm:  1.8140  memory:  1.06GiB(1.11%)  tps: 320,740  tflops: 22.96  mfu: 2.32%
```

with lora converter
```
[rank0]:[titan] 2026-02-02 14:42:44,379 - root - INFO - step:  1  loss:  8.1225  grad_norm:  0.3293  memory:  0.98GiB(1.03%)  tps: 6,222  tflops: 0.46  mfu: 0.05%
[rank0]:[titan] 2026-02-02 14:42:44,687 - root - INFO - step:  5  loss:  8.0054  grad_norm:  0.5108  memory:  1.02GiB(1.07%)  tps: 274,052  tflops: 20.05  mfu: 2.03%
[rank0]:[titan] 2026-02-02 14:42:44,988 - root - INFO - step: 10  loss:  7.8191  grad_norm:  0.7809  memory:  1.02GiB(1.07%)  tps: 266,976  tflops: 19.53  mfu: 1.98%
```

with lora converter + float8
```
[rank0]:[titan] 2026-02-02 14:43:38,608 - root - INFO - step:  1  loss:  8.1689  grad_norm:  0.2351  memory:  1.00GiB(1.05%)  tps: 5,067  tflops: 0.37  mfu: 0.04%
[rank0]:[titan] 2026-02-02 14:43:39,152 - root - INFO - step:  5  loss:  8.0590  grad_norm:  0.3869  memory:  1.04GiB(1.10%)  tps: 134,235  tflops: 9.82  mfu: 0.99%
[rank0]:[titan] 2026-02-02 14:43:39,792 - root - INFO - step: 10  loss:  7.9155  grad_norm:  0.6083  memory:  1.04GiB(1.10%)  tps: 122,904  tflops: 8.99  mfu: 0.91%
```

Loss tests, running full finetune, lora finetune(rank=8), lora finetune(rank=64) on llama3 debugmodel
loss_curves

### Future work
There are some todo works that not included in this PR and will be future work
1. Composable with TP, the current plan does not work with TP, here are 2 plans we may try later and need further discussion:
- Keep lora_adapters as Replicate so that they won't be shardded by TP
- Generate layer_plan for lora adapters so that they will be processed by TP as well
2. Support supervised finetuning

* #2569
* #2488
* #2485
* __->__ #2484


torchtitan/components/lora.py
```diff
@@ -0,0 +1,123 @@
...
+                apply_lora(child, self.rank, self.alpha)
+
+        # Patch init_weights to also reinitialize LoRA adapters
+        original_init_weights = getattr(module, "init_weights", None)
+
+        def new_model_init_weights(*args: Any, **kwargs: Any) -> None:
+            if original_init_weights is not None and callable(original_init_weights):
+                original_init_weights(*args, **kwargs)
+            for sub_module in module.modules():
+                if type(sub_module) in _lora_class_cache.values():
```
tianyu-l: It seems we need this because nn.Linear doesn't have `init_weights` and instead in torchtitan we directly operating on the weight and bias in any nn.Linear submodule.

I believe @fegin will refactor this to create our own Linear module with init_weights capability. So let's leave a TODO here to remove this `new_model_init_weights`.

torchtitan/models/llama3/config_registry.py
```diff
@@ -108,6 +109,19 @@ def llama3_debugmodel_float8_emulate() -> Trainer.Config:
     return config
 
 
+def llama3_debugmodel_lora() -> Trainer.Config:
+    config = llama3_debugmodel()
+    config.model_converters = ModelConvertersContainer.Config(
```
tianyu-l, line 119: Among all the converters (quantization, lora, etc.), order matters. For now let's create a `__post_init__` in `ModelConvertersContainer.Config` to validate the order.

Note that it can't capture the wrong order, if we first init the main config and then modify a field.

@fegin Validation is becoming an issue. If we don't go with chz, maybe we should at least freeze the config, and require users to use dataclass.replace to trigger post init checks.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,107 @@
...
+                    .to(device=device, dtype=dtype)
+                )
+                self.lora_b = (
+                    Linear.Config(bias=False)
+                    .build(in_features=rank, out_features=self.out_features)
+                    .to(device=device, dtype=dtype)
+                )
+
+            def init_weights(self, **kwargs) -> None:
+                super().init_weights(**kwargs)  # pyrefly: ignore [not-callable]
```
mori360, line 71: Remove new_model_init_weights after refactoring Linear @fegin

mori360:
We plan to hold on this pr until the pr to refactor parallelism next week, and fix composability issue to tp that time. (will support ep in the pr later)

Andrei-Aksionov:
@tianyu-l 
Since LoRA was added in #3264 shouldn't this be closed?

PR #2485, opened 2026-03-04
[2/N] Support lora checkpoint on partial save and multi-source load

mori360:
### Summary
Consolidate converter-checkpoint integration from 5 scattered dynamic attributes into a single `ConverterCheckpointHooks` dataclass, simplifying ModelWrapper and eliminating closure factories in LoRA.
 
#### Design: ConverterCheckpointHooks
```
  class ConverterCheckpointHooks:
      key_filter: Callable[[str], bool] | None = None
      save_last_fn: Callable | None = None
      load_additional_fn: Callable | None = None
      finalize_fn: Callable | None = None
```

  - `key_filter` — Distinguishes converter-added keys (e.g. LoRA's .lora_a., .lora_b.) from base model keys. Used by
  ModelWrapper.state_dict() to save only adapter keys in periodic checkpoints (small, fast saves), and by base_state_dict() to exclude them when loading base weights from HF. Without this, every periodic checkpoint would save the full model.
  - `save_last_fn` — Converter-specific save format for the final checkpoint. LoRA uses this to write PEFT-compatible adapter_model.safetensors + adapter_config.json so the result can be loaded directly by HuggingFace PEFT/vLLM. Only called when last_save_in_hf=True. Without this, the final save would use DCP format which external tools can't read.
  - `load_additional_fn` — Loads converter-specific checkpoint formats during resume. LoRA uses this to load PEFT safetensors and remap keys back to torchtitan naming. Called for secondary checkpoint paths (the adapter checkpoint, separate from the base model HF checkpoint). Without this, resuming from a PEFT save would fail.
  - `finalize_fn` — End-of-training model transformation before the last save. Runs ModelConvertersContainer.finalize() in reverse order: LoRA merge (fold adapters into base weights) then QAT CONVERT (replace fake-quantized modules with real quantized ones). This is set by ModelConvertersContainer.convert(), not by individual converters. Without this, the final checkpoint would contain training-time artifacts (LoRA adapters, fake quantization) instead of a clean deployable model.
Screenshot 2026-03-20 at 3 45 01 PM

#### Changes

  - model_converter.py: Added ConverterCheckpointHooks dataclass. ModelConvertersContainer.convert() merges into existing hooks.
  - lora.py: Replaced closure factories (_make_peft_save_fn, _make_peft_load_fn, _make_merge_fn) with regular methods (_save_peft, _load_peft). convert() builds hooks directly. finalize() cleans up hooks after merge.
  - checkpoint.py: ModelWrapper simplified from 7 methods to 5. Added _get_hooks() as single access point. Removed scattered getattr(part, "converter_*") reads.
  - qat.py: Extracted apply_qat_prepare() shared by QATConverter.convert() and LoRAConverter._apply_adapter_qat().

  ### Test plan

  - [ ] `pytest tests/unit_tests/test_checkpoint.py -x` — new `TestModelWrapperConverterKeys` tests (strict vs
   partial planner)
  - [ ] `pytest tests/unit_tests/test_model_converter.py -x` — new `test_lora_key_remap_roundtrip`
  - [ ] `torchrun --nproc_per_node=4 -m torchtitan.train --module llama3 --config llama3_debugmodel_lora` —
  verify LoRA training runs end-to-end
  - [ ] `torchrun --nproc_per_node=4 -m torchtitan.train --module llama3 --config llama3_8b_lora` — verify 8B
  LoRA config with PEFT save

* #2569
* #2488
* __->__ #2485
* #2484
* #2484

### DIFF
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -58,13 +58,21 @@ def apply_lora(linear: nn.Linear, rank: int, alpha: float) -> nn.Linear:
                 device = self.weight.device
                 dtype = self.weight.dtype
                 self.lora_a = (
-                    Linear.Config(bias=False)
-                    .build(in_features=self.in_features, out_features=rank)
+                    Linear.Config(
+                        in_features=self.in_features,
+                        out_features=rank,
+                        bias=False,
+                    )
+                    .build()
                     .to(device=device, dtype=dtype)
                 )
                 self.lora_b = (
-                    Linear.Config(bias=False)
-                    .build(in_features=rank, out_features=self.out_features)
+                    Linear.Config(
+                        in_features=rank,
+                        out_features=self.out_features,
+                        bias=False,
+                    )
+                    .build()
                     .to(device=device, dtype=dtype)
                 )
 


fegin:
> With small scale, loading main weights in HF format, and loading lora weight in HF / DCP is good enough.

Do we have the number for this? iirc, the conversion is not cheap though it may be acceptable on a single machine. 

And this doesn't simplify the logic as we will need to save trainer/optimizer states for fault tolerance purpose. The original design is coupled that with the model weights, which are frozen now.

If we are okay with always loading with HF format, even for the fault tolerance ones, we should treat the HF folder as the `additional folder`. lora adapter + trainer + optimizer are in the `original checkpoint folder`. cc., @mori360

PR #2486, opened 2026-03-05
[RL] Turn on vllm `support_torch_compile` in RL speeding up training

Lucaskabela:
## Summary

Implements the changes necessary for vllm enablement.  These are:
* Support weak_ref capture of DTensors in vLLM required for cudagraph
* Rewrite vllm_wrapper to avoid graph break (extending rope cache, .max().item())
* Rewrite attention for simplicity (fold reshape operations)
* Narrow output of vLLM Attention for correct symbolic shape propogation (can be resolved with https://github.com/pytorch/pytorch/issues/175690 fix but not clear the scope/necessity due to dropping guards in vLLM)

For more context, this is a reimplementation of https://github.com/pytorch/torchtitan/pull/2393 post rebase

We then test with **eager backend (dynamo/torch compile) and piecewise cudagraphs** and verify there is no numeric difference.  We observe ~1.85x speedup end to end, and 15x speedup in generator time.  That said, there is additional overhead (~50s) of compiling and capturing cudagraphs, but considering we save ~200s in runtime, the net benefit is ~150s or 3 minutes

## Test 
Running
```bash
python torchtitan/experiments/rl/unified/simple_grpo.py --module rl.unified --config rl_grpo_qwen3_0_6b --hf_assets_path=torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B
```
with simple instrumentation for timing (just wrapping call() in grpo with perf counters) 
                                         
  ### Startup                                                                                                                                                                
                                                                                                                                                                             
  |                          | Baseline (eager) | Compile+CUDAGraph | Delta    |                                                                                             
  |--------------------------|------------------|-------------------|----------|                                                                                             
  | Generator init + sync    | 10.21s           | 60.92s            | +50.71s  |

  Compile overhead: torch.compile 16.35s + CUDA graph capture 33s

  ### Per-Step Comparison

  | Step | Generate (eager) | Generate (compile eager + piecwise) | Speedup | Logprob diff match? |
  |------|------------------|--------------------|---------|---------------------|
  | 0    | 24.09s           | 2.95s              | **8.2x** | Yes (mean=7.30e-06, max=2.19e-01) |
  | 1    | 23.08s           | 1.39s              | **16.6x** | Yes (mean=-1.41e-03, max=2.43e-01) |
  | 2    | 24.03s           | 1.44s              | **16.7x** | Yes (mean=-2.31e-03, max=6.65e-01) |
  | 3    | 23.96s           | 1.48s              | **16.2x** | Yes (mean=-1.51e-03, max=9.60e-01) |
  | 4    | 25.58s           | 1.40s              | **18.3x** | Yes (mean=-3.98e-03, max=1.17e+00) |
  | 5    | 24.16s           | 1.35s              | **17.9x** | Yes (mean=-8.51e-03, max=2.32e+00) |
  | 6    | 24.69s           | 1.25s              | **19.8x** | Yes (mean=-4.18e-03, max=2.08e+00) |
  | 7    | 22.53s           | 1.39s              | **16.2x** | Yes (mean=-7.42e-03, max=1.74e+00) |
  | 8    | 23.51s           | 1.54s              | **15.3x** | Yes (mean=-5.23e-03, max=2.25e+00) |
  | 9    | 24.39s           | 1.67s              | **14.6x** | Yes (mean=-8.91e-03, max=2.10e+00) |

  ### Totals

  |              | Baseline | Compile+CUDAGraph | Speedup |
  |--------------|----------|--------------------|---------|
  | Generate     | 240.03s  | 15.87s             | **15.1x** |
  | Train        | 151.02s  | 152.50s            | ~1.0x   |
  | Sync         | 88.66s   | 93.26s             | ~1.0x   |
  | **Total**    | **479.87s** | **261.71s**     | **1.83x** |

  Loss, reward, and logprob diffs are **identical** across all steps -- compile+cudagraph is numerically equivalent

### DIFF
--- a/torchtitan/experiments/rl/unified/actors/generator.py
+++ b/torchtitan/experiments/rl/unified/actors/generator.py
@@ -7,6 +7,7 @@
 import logging
 import os
 from dataclasses import dataclass, field
+from typing import Literal
 
 import torch
 from monarch.actor import Actor, endpoint
@@ -20,12 +21,71 @@ from torchtitan.experiments.rl.unified.plugin import (
 from torchtitan.experiments.rl.unified.types import Episode
 from torchtitan.protocols.model_spec import ModelSpec
 from vllm import EngineArgs, LLMEngine, SamplingParams
-from vllm.config import AttentionConfig
+from vllm.config import AttentionConfig, CompilationConfig
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass(kw_only=True, slots=True)
+class GeneratorCompileConfig:
+    """Compilation and CUDA graph settings for the vLLM generator."""
+
+    backend: Literal["none", "eager", "inductor"] = "eager"
+    """torch.compile backend for vLLM
+    When set to a value other than "none", enables compilation with the specified backend.
+    See https://docs.vllm.ai/en/stable/api/vllm/config/#vllm.config.CompilationConfig.backend
+    NOTE: "eager" means compile with dynamo backend (like torch.compile(backend="eager"))
+    NOTE: inductor will offer the best performance, but will impact numerics - use eager for
+    bitwise identical results."""
+
+    cudagraph_mode: Literal[
+        "none", "piecewise", "full", "full_and_piecewise"
+    ] = "piecewise"
+    """CUDA graph capture mode for vLLM.
+    Piecewise capture supports dynamic sizes and splits cudagraphs around non capturable
+      ops like attention
+    Full capture captures one graph at the expense of less dynamism and requires full
+      capturability
+    full_and_piecewise does both and selects which to use based on dynamism
+    NOTE: Piecewise graph capture requires torch.compile for graph capture and splitting
+    See https://docs.vllm.ai/en/latest/design/cuda_graphs/#cudagraphmodes for more details."""
+
+    def __post_init__(self) -> None:
+        if self.backend == "none" and self.cudagraph_mode in (
+            "piecewise",
+            "full_and_piecewise",
+        ):
+            raise ValueError(
+                f"cudagraph_mode='{self.cudagraph_mode}' requires piecewise graph "
+                "capture which depends on torch.compile. Set backend "
+                "to 'eager' or 'inductor'."
+            )
+
+    @property
+    def is_eager(self) -> bool:
+        """Inferred from backend and cudagraph_mode."""
+        return self.backend == "none" and self.cudagraph_mode == "none"
+
+    def get_vllm_compilation_config(self) -> CompilationConfig | None:
+        """Build a vLLM ``CompilationConfig``, or return ``None`` when both
+        compilation and CUDA graphs are disabled.
+        """
+        if self.is_eager:
+            return None
+
+        kwargs: dict = dict(cudagraph_mode=self.cudagraph_mode)
+        if self.backend == "none":
+            # Disable torch.compile but keep CUDA graphs (e.g. full mode).
+            # mode=0 (CompilationMode.NONE) prevents vLLM from inferring
+            # VLLM_COMPILE based on the default optimization level.
+            kwargs["mode"] = 0
+        else:
+            kwargs["backend"] = self.backend
+
+        return CompilationConfig(**kwargs)
 
 
 @dataclass(kw_only=True, slots=True)
@@ -79,8 +139,8 @@ class VLLMGenerator(Actor, Configurable):
         gpu_memory_limit: float = 0.9
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        enforce_eager: bool = True
-        """Disable CUDA graphs in vLLM (use eager execution)."""
+        compile: GeneratorCompileConfig = field(default_factory=GeneratorCompileConfig)
+        """Compilation and CUDA graph settings for the vLLM engine."""
 
         num_samples_per_prompt: int = 8
         """Number of completions to generate per prompt."""
@@ -135,14 +195,16 @@ class VLLMGenerator(Actor, Configurable):
             # tells vLLM to run one worker per process (no subprocess spawning)
             distributed_executor_backend="external_launcher",
             gpu_memory_utilization=config.gpu_memory_limit,
-            enforce_eager=config.enforce_eager,
-            # Disable vLLM logging to avoid noisy logs on every step
-            disable_log_stats=True,
+            enforce_eager=config.compile.is_eager,
             hf_overrides={"architectures": [VLLM_MODEL_NAME]},
             attention_config=AttentionConfig(
                 backend=AttentionBackendEnum[config.attention_backend],
             ),
+            disable_log_stats=True,
         )
+        vllm_compilation_config = config.compile.get_vllm_compilation_config()
+        if vllm_compilation_config is not None:
+            engine_kwargs["compilation_config"] = vllm_compilation_config
         if config.seed is not None:
             engine_kwargs["seed"] = config.seed
         engine_args = EngineArgs(**engine_kwargs)
--- a/torchtitan/experiments/rl/unified/config_registry.py
+++ b/torchtitan/experiments/rl/unified/config_registry.py
@@ -15,6 +15,7 @@ from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
 from torchtitan.config.configs import ParallelismConfig, TrainingConfig
 from torchtitan.experiments.rl.unified.actors.generator import (
+    GeneratorCompileConfig,
     SamplingConfig,
     VLLMGenerator,
 )
@@ -43,7 +44,10 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
-            enforce_eager=True,
+            compile=GeneratorCompileConfig(
+                backend="eager",
+                cudagraph_mode="piecewise",
+            ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
                 data_parallel_replicate_degree=1,
@@ -78,7 +82,10 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
             ),
         ),
         generator=VLLMGenerator.Config(
-            enforce_eager=True,
+            compile=GeneratorCompileConfig(
+                backend="eager",
+                cudagraph_mode="piecewise",
+            ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=1,
                 data_parallel_replicate_degree=1,
--- a/torchtitan/experiments/rl/unified/models/attention.py
+++ b/torchtitan/experiments/rl/unified/models/attention.py
@@ -89,6 +89,11 @@ class VLLMAttention(torch.nn.Module):
         Returns:
             ``(batch, num_heads, seq_len, head_dim)``
         """
+        # Capture the original symbolic seq_len from the input BEFORE
+        # to_local() so that the symbol is the same one GQAttention uses
+        # in its .view(bs, seqlen, -1) call.
+        batch_size, _, seq_len, head_dim = q.shape
+
         # Unwrap DTensor inputs to local tensors for attention computation
         device_mesh = None
         placements = None
@@ -99,33 +104,28 @@ class VLLMAttention(torch.nn.Module):
             k = k.to_local()
             v = v.to_local()
 
-        # Input is (batch, num_heads, seq_len, head_dim)
         # TODO: may be good to use einops in future as we can explicitly reshape
         # with dimension names - see https://github.com/arogozhnikov/einops
-        batch_size, num_heads, seq_len, head_dim = q.shape
-        _, num_kv_heads, _, _ = k.shape
+        # Convert from (batch, num_heads, seq_len, head_dim)
+        #   to (batch*seq_len, num_heads (or num_kv_heads), head_dim) for vLLM Attn
+        q = q.transpose(1, 2).reshape(batch_size * seq_len, -1, head_dim)
+        k = k.transpose(1, 2).reshape(batch_size * seq_len, -1, head_dim)
+        v = v.transpose(1, 2).reshape(batch_size * seq_len, -1, head_dim)
 
-        # Transpose to (batch, seq_len, num_heads, head_dim) for vLLM
-        q = q.transpose(1, 2)
-        k = k.transpose(1, 2)
-        v = v.transpose(1, 2)
-
-        # TODO: reimplement as a 4d tensor once vLLM fix has landed
-        # Then flatten batch and seq_len: (batch * seq_len, num_heads, head_dim)
-        q = q.reshape(batch_size * seq_len, num_heads, head_dim)
-        k = k.reshape(batch_size * seq_len, num_kv_heads, head_dim)
-        v = v.reshape(batch_size * seq_len, num_kv_heads, head_dim)
-
-        # vLLM attention returns (num_tokens, hidden_size) where hidden_size = num_heads * head_dim
+        # vLLM attention returns (num_tokens, num_heads/num_kv_heads * head_dim)
         output_flat = self.vllm_attn(q, k, v)
 
-        # Output is (batch * seq_len, num_heads * head_dim), reshape to (batch, seq_len, num_heads, head_dim)
-        output = output_flat.view(batch_size, seq_len, num_heads, head_dim)
+        # vLLM's flash attention backend may pad the token count (e.g.
+        # round up to an even number), which introduces a new symbolic
+        # shape under torch.compile.  Narrow to trim this padding
+        # NOTE: this error only happens when batch_size and seq_len are 1
+        # which happens with cudagraph capture for dummy input
+        output_flat = output_flat.narrow(0, 0, batch_size * seq_len)
 
-        # Transpose back to TorchTitan format: (batch, num_heads, seq_len, head_dim)
+        # Reshape back to titan: (batch, num_heads_local, seq_len, head_dim)
+        output = output_flat.view(batch_size, seq_len, -1, head_dim)
         output = output.transpose(1, 2)
 
-        # Wrap output back as DTensor if inputs were DTensors
         if device_mesh is not None:
             output = DTensor.from_local(
                 output, device_mesh=device_mesh, placements=placements
--- a/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/unified/models/vllm_wrapper.py
@@ -29,12 +29,31 @@ from torchtitan.experiments.rl.unified.models.attention import (
     replace_with_vllm_attention,
 )
 from torchtitan.protocols.model_spec import ModelSpec
-
+from vllm.compilation.decorators import support_torch_compile
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
+from vllm.utils import torch_utils as _torch_utils
 
 
 logger = init_logger(__name__)
+
+# NOTE: Monkeypatch vLLM's weak_ref_tensor to handle DTensor
+# This is because piecewise CUDA-graph capture calls weak_ref_tensor()
+# on every subgraphoutput (see vllm/compilation/cuda_graph.py).
+# When TP is active some of those outputs are DTensors which fail with
+# ("The specified pointer resides on host memory").  to_local
+# converts the DTensor to a plain tensor. which succeeds with this
+# cudagraph implementation.
+_original_weak_ref_tensor = _torch_utils.weak_ref_tensor
+
+
+def _dtensor_safe_weak_ref_tensor(tensor):
+    if isinstance(tensor, DTensor):
+        tensor = tensor._local_tensor
+    return _original_weak_ref_tensor(tensor)
+
+
+_torch_utils.weak_ref_tensor = _dtensor_safe_weak_ref_tensor
 
 
 def create_torchtitan_config_from_vllm_config(
@@ -93,6 +112,12 @@ def create_torchtitan_config_from_vllm_config(
     return parallel_dims, parallelism
 
 
+@support_torch_compile(
+    dynamic_arg_dims={
+        "input_ids": 0,
+        "positions": 0,
+    }
+)
 class TorchTitanVLLMModelWrapper(nn.Module):
     """
     Generic vLLM-compatible model wrapper for TorchTitan models. Implemented
@@ -158,28 +183,32 @@ class TorchTitanVLLMModelWrapper(nn.Module):
             has_position_id=True,  # vLLM always passes positions explicitly
         )
 
+        # Pre-extend RoPE cache to cover vLLM's max model length (profiling
+        # may use up to 2x max_seq_len, so use max_model_len which already
+        # accounts for this).  This avoids data-dependent control flow in
+        # forward() which is incompatible with torch.compile.
+        max_model_len = vllm_config.model_config.max_model_len
+        if self.model.freqs_cis.shape[0] < max_model_len:
+            self.model.freqs_cis = self._extend_rope_cache(
+                self.model.freqs_cis, max_model_len
+            )
+
         # Initial load model weights from HuggingFace checkpoint path
         self._initial_load_weights(checkpoint_path=vllm_config.model_config.model)
 
-    def _extend_rope_cache_if_needed(
-        self, rope_cache: torch.Tensor, max_position: int
+    def _extend_rope_cache(
+        self, rope_cache: torch.Tensor, required_len: int
     ) -> torch.Tensor:
         """
-        Extend RoPE cache if needed during vLLM profiling stage.
+        Build an extended RoPE cache of at least ``required_len`` positions.
 
         Args:
             rope_cache: Current RoPE cache tensor
-            max_position: Maximum position index needed
+            required_len: Minimum number of positions the cache must cover
 
         Returns:
-            Extended RoPE cache if needed, otherwise original cache
+            Extended RoPE cache tensor
         """
-        required_len = max_position + 1
-
-        # No extension needed
-        if required_len <= rope_cache.shape[0]:
-            return rope_cache
-
         # Handle DTensor case
         is_dtensor = isinstance(rope_cache, DTensor)
         if is_dtensor:
@@ -253,15 +282,7 @@ class TorchTitanVLLMModelWrapper(nn.Module):
         # Get embeddings
         h = self.model.tok_embeddings(tokens_2d)
 
-        # Extend RoPE cache if needed (vLLM profiling may use 2x max_seq_len)
-        if positions is not None:
-            max_position = positions.max().item()
-        else:
-            max_position = 0
-
-        rope_cache = self._extend_rope_cache_if_needed(
-            self.model.freqs_cis, max_position
-        )
+        rope_cache = self.model.freqs_cis
         positions = positions.unsqueeze(0)
 
         # Pass through transformer layers
@@ -275,9 +296,8 @@ class TorchTitanVLLMModelWrapper(nn.Module):
 
         # Convert to vLLM format: [total_tokens, hidden_size]
         if h.dim() == 3:
-            batch_size, seq_len, hidden_size = h.shape
-            h = h.view(batch_size * seq_len, hidden_size)
-
+            hidden_size = h.size(-1)
+            h = h.view(-1, hidden_size)
         return h
 
     def compute_logits(
--- a/torchtitan/experiments/rl/unified/inference_example.py
+++ b/torchtitan/experiments/rl/unified/inference_example.py
@@ -11,7 +11,7 @@ Example inference script using TorchTitan models with vLLM LLMEngine.
 This script uses the RL unified config_registry to configure both
 the vLLM engine and sampling parameters.
 
-Run: torchrun --nproc_per_node=<world_size> \
+Run: torchrun --nproc_per_node=2 \
       torchtitan/experiments/rl/unified/infer.py
 """
 import os
@@ -70,10 +70,13 @@ def generate():
         distributed_executor_backend=("external_launcher"),
         # Memory and performance
         gpu_memory_utilization=gen_config.gpu_memory_limit,
-        enforce_eager=gen_config.enforce_eager,
+        enforce_eager=gen_config.compile.is_eager,
         # HuggingFace overrides
         hf_overrides={"architectures": [VLLM_MODEL_NAME]},
     )
+    vllm_compilation_config = gen_config.compile.get_vllm_compilation_config()
+    if vllm_compilation_config is not None:
+        engine_kwargs["compilation_config"] = vllm_compilation_config
     if gen_config.seed is not None:
         engine_kwargs["seed"] = gen_config.seed
     engine_args = EngineArgs(**engine_kwargs)


Lucaskabela:
For folks wondering why this timing data is different than the previous iteration of this PR - post rebase generator has tp=2.  This is a significant change in the code

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -89,44 +91,44 @@ def forward(
...
 
-        # Output is (batch * seq_len, num_heads * head_dim), reshape to (batch, seq_len, num_heads, head_dim)
-        output = output_flat.view(batch_size, seq_len, num_heads, head_dim)
+        # vLLM's flash attention backend may pad the token count (e.g.
+        # round up to an even number), which introduces a new symbolic
+        # shape under torch.compile.  Narrow back to the original
+        # batch_size * seq_len symbols so that downstream DTensor ops
+        # (like the wo projection) see shapes consistent with the
+        # weight sharding.
+        output_flat = output_flat.narrow(0, 0, batch_size * seq_len)
```
Lucaskabela, line 123: This is the only actual change here in this function - the rest is cosmetic rewrites for simplification.  I will make an iteration of the PR with the ordering closer to the original to make that more obvious

Lucaskabela: **Update** We tracked this down to an issue where batch_size and seq_len are both 1 for dummy cudagraph tracing - in this case, the vllm_attn actually pads the input so we do need to narrow.  For other times this will be a no-op, so I think it is safe to do this unconditionally here

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -83,8 +101,18 @@ class Config(Configurable.Config):
         gpu_memory_limit: float = 0.5
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        enforce_eager: bool = True
-        """Disable CUDA graphs in vLLM (use eager execution)."""
+        enforce_eager: bool = False
```
tianyu-l: do you still need this flag? There doesn't seem that many degrees of freedom.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -83,8 +101,18 @@ class Config(Configurable.Config):
...
+        """When True, disables compile/cudagraph optimization.  Useful for debugging but harms performance."""
+
+        compilation_backend: Literal["eager", "inductor"] | None = None
+        """torch.compile backend for vLLM.
+        When set, enables compilation. Requires ``enforce_eager=False``."""
+
+        cudagraph_mode: Literal[
```
tianyu-l: can we turn on cudagraph with compile off?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -83,8 +101,18 @@ class Config(Configurable.Config):
         gpu_memory_limit: float = 0.5
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        enforce_eager: bool = True
-        """Disable CUDA graphs in vLLM (use eager execution)."""
+        enforce_eager: bool = False
+        """When True, disables compile/cudagraph optimization.  Useful for debugging but harms performance."""
+
+        compilation_backend: Literal["eager", "inductor"] | None = None
```
```suggestion
        compilation_backend: Literal["none", "inductor"] = "none"
```
should also at least add "aot_eager"?

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: can trainer side not use cudagraph?

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: vLLM doesn't have aot_eager - only "eager" or "inductor".

Let me modify the code to link the compilation backend (see also https://docs.vllm.ai/en/stable/api/vllm/config/#vllm.config.CompilationConfig.backend)

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: There are not, there are just two.  Since false is equivalent of compilation_backend=None and cudagraph_mode=None, I think it is fine to just infer from those two flags and avoid the muck of needed this flag here

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: > can we turn on cudagraph with compile off?

Piecewise cudagraph requires compile to produce a partitioned graph around non cudagraphable operations (like communication ops); so we cannot turn off compile to capture piecewise cudagraph.  I will modify the code to make this a bit more obvious

> can trainer side not use cudagraph?

Not yet - this code is orchestrating cudagraph through `@support_torch_compile` which works through vLLM/ the vLLM model registry.  I will submit another PR working on compile + cudgraph support for the trainer, and most of the changes for that will focus on traceability of the backwards ops since this PR gets forward in a fairly good state

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -101,6 +129,11 @@ def __init__(
         batch_invariant_mode: bool,
         prompt_texts: list[str],
     ):
+        if config.compilation_backend is not None and config.enforce_eager:
```
acisseJZhong: what's the difference between compilation backend being `eager` or `None`, can we deduplicate these two fields?

torchtitan/experiments/rl/unified/actors/generator.py
acisseJZhong: > Piecewise cudagraph requires compile to produce a partitioned graph around non cudagraphable operations (like communication ops); so we cannot turn off compile to capture piecewise cudagraph.

maybe briefly add this to docstring?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -89,44 +91,44 @@ def forward(
...
+        q = q.transpose(1, 2).flatten(0, 1)
+        k = k.transpose(1, 2).flatten(0, 1)
+        v = v.transpose(1, 2).flatten(0, 1)
 
-        # Transpose to (batch, seq_len, num_heads, head_dim) for vLLM
-        q = q.transpose(1, 2)
-        k = k.transpose(1, 2)
-        v = v.transpose(1, 2)
-
-        # TODO: reimplement as a 4d tensor once vLLM fix has landed
```
acisseJZhong, line 113: can we keep this?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -89,44 +91,44 @@ def forward(
...
-        if isinstance(q, DTensor):
+        # Capture the original symbolic seq_len from the input BEFORE
+        # to_local() so that the symbol is the same one GQAttention uses
+        # in its .view(bs, seqlen, -1) call.
+        batch_size, _, seq_len, head_dim = q.shape
+
+        # Unwrap DTensor inputs to local tensors — vLLM's Attention uses
+        # local head counts in views/reshapes that are incompatible with
+        # DTensor's global shape semantics under torch.compile.
+        if self.tp_enabled:
```
acisseJZhong: can we keep the original isinstance(q, DTensor) as we might use other parallelism in vllm? Maybe only guard ` output_flat.narrow` using `tp_enabled`?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -89,44 +91,44 @@ def forward(
...
+        output_flat = output_flat.narrow(0, 0, batch_size * seq_len)
 
-        # Transpose back to TorchTitan format: (batch, num_heads, seq_len, head_dim)
+        # Reshape back to titan: (batch, num_heads_local, seq_len, head_dim)
+        output = output_flat.view(batch_size, seq_len, -1, head_dim)
         output = output.transpose(1, 2)
 
-        # Wrap output back as DTensor if inputs were DTensors
-        if device_mesh is not None:
+        if self.tp_enabled:
```
acisseJZhong: same above

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -29,13 +29,32 @@
...
+# When TP is active some of those outputs are DTensors which fail with
+# ("The specified pointer resides on host memory").  to_local
+# converts the DTensor to a plain tensor. which succeeds with this
+# cudagraph implementation.
+_original_weak_ref_tensor = _torch_utils.weak_ref_tensor
+
+
+def _dtensor_safe_weak_ref_tensor(tensor):
+    if isinstance(tensor, DTensor):
+        tensor = tensor.to_local()
```
acisseJZhong: can we explicitly pass in grad_placements to to_local?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -29,13 +29,32 @@
...
 from vllm.logger import init_logger
+from vllm.utils import torch_utils as _torch_utils
 
 
 logger = init_logger(__name__)
 
+# NOTE: Monkeypatch vLLM's weak_ref_tensor to handle DTensor
+# This is because piecewise CUDA-graph capture calls weak_ref_tensor()
+# on every subgraphoutput (see vllm/compilation/cuda_graph.py).
+# When TP is active some of those outputs are DTensors which fail with
```
acisseJZhong, line 43: is this saying DTensor output doesn't work well with cuda graph capture and you have to convert to local tensor? 

I think this differs from our initial effort to use full DTensor?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -253,31 +282,20 @@ def forward(
...
 
         # Pass through transformer layers
         for layer in self.model.layers.values():
             h = layer(h, rope_cache, attention_masks=None, positions=positions)
 
         h = self.model.norm(h)
-        # When parallelism is applied, get full tensor before return to vLLM Engine
-        if isinstance(h, DTensor):
+        # When TP is applied, gather the full tensor before returning to vLLM.
+        if self.parallel_dims.tp_enabled:
```
acisseJZhong: same above

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -253,31 +282,20 @@ def forward(
...
             h = h.full_tensor()
 
         # Convert to vLLM format: [total_tokens, hidden_size]
-        if h.dim() == 3:
-            batch_size, seq_len, hidden_size = h.shape
-            h = h.view(batch_size * seq_len, hidden_size)
-
+        h = h.view(-1, h.size(-1))
```
acisseJZhong: I personally prefer the original approach which express dimension more clear/easier to understand d

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -86,7 +89,9 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
         ),
         generator=VLLMGenerator.Config(
             gpu_memory_limit=0.3,
-            enforce_eager=True,
+            enforce_eager=False,
```
acisseJZhong: does it make sense to wrap the three fields under some Config class to group them together?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -24,14 +25,31 @@
...
-from vllm.config import AttentionConfig
+from vllm.config import AttentionConfig, CompilationConfig
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
 
 logger = logging.getLogger(__name__)
 
 
+def _resolve_cudagraph_mode(cudagraph_mode: str, tp_degree: int) -> str:
```
acisseJZhong: move to util?

torchtitan/experiments/rl/unified/models/attention.py
Lucaskabela: Yeah we can - I just preferred the boolean check over the isinstance check but shouldn't be a blocker

torchtitan/experiments/rl/unified/models/attention.py
Lucaskabela: Sure :) - I had collapsed this to transpose/flatten to make it shorter but I think that obscures what we are actually doing - let me fold this

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: Yes, specifically for vLLM's cudagraph capture, see https://github.com/vllm-project/vllm/blob/86483ca7749b3d7a2ae16283a7896c203983f1ef/vllm/compilation/cuda_graph.py#L289

This is done for memory management to avoid losing the memroy. we get a weak_ref to the output tensor.  That weak_ref calls `torch.ops._C.weak_ref_tensor` which errors for DTensor, hence the necessity to convert to_local here

For reference, I tested removing these lines of code, and they pass for this model (i.e we can run without the weak_ref capture); but we can't remove this code from vLLM as it is needed for some models there

So monkeypatching like this should be fine/safe

Lucaskabela: Also this monkeypatch is strictly used in cudagraph capture for memory management - it doesn't affect the DTensor usage in the model at all to my understanding

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: @acisseJZhong  The vLLM_wrapper is inference only - no grad so to my understanding this would be a nop

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: We can keep this then - my philosophy was rewriting to remove conditionals and make reshapes as close to one line as possible, but the minimal perf improvements probably aren't worth it over readability especially for experimental code

wwwjn:
> For folks wondering why this timing data is different than the previous iteration of this PR - post rebase generator has tp=2. This is a significant change in the code

To clarify, the time reported in the description is 1) TP=2 on both trainer and generator side 2) With compile (which backend? aot_eager? w/ or w/o inductor?) and CUDAGraph enabled at the same time (piece-wise cuda graph) 3) forward only on vllmWrapper generator?

What about numerics? Is there numerics change?

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -99,33 +106,28 @@ def forward(
             k = k.to_local()
             v = v.to_local()
 
-        # Input is (batch, num_heads, seq_len, head_dim)
         # TODO: may be good to use einops in future as we can explicitly reshape
         # with dimension names - see https://github.com/arogozhnikov/einops
-        batch_size, num_heads, seq_len, head_dim = q.shape
```
Lucaskabela, line 105: This needs to move above to_local for correct symbolic capture

Lucaskabela:
Great questions - I will add the clarity to the summary.  To answer here for completeness

> To clarify, the time reported in the description is 1) TP=2 on both trainer and generator side 2) With compile (which backend? aot_eager? w/ or w/o inductor?) and CUDAGraph enabled at the same time (piecewise cuda graph) 3) forward only on vllmWrapper generator?

1) Yes TP=2 on both (just using the stock config) with vllm compile settings: eager + piecewise.  Yes forward only on vLLMWrapper Generator

> What about numerics? Is there numerics change?

No change in the numerics - this is the `Logprob diff match?` column showing that the numerics were identical across runs

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: 'eager' actually corresponds to `dynamo` in stock torch.compile - this is just unfortunate naming in vLLM :/

torchtitan/experiments/rl/unified/config_registry.py
Lucaskabela: Done!

wwwjn:
> Yes (mean=7.30e-06, max=2.19e-01)

I'm confuse with the numerics result in the table: What does logprob match = Yes, but with a mean / max value here?

In my previous refactor PR, the bit-wise parity between trainer and generator is not achieved yet when TP=2.  So I want s to learn if the log prob output between 1. compiled generator model and 2. eager generator model are the same.

Lucaskabela:
> I'm confuse with the numerics result in the table: What does logprob match = Yes, but with a mean / max value here?

@wwwjn This is the logprob diff reported from each step of training extracted from this log https://github.com/pytorch/torchtitan/blob/32f9c58f6368b24296d4b2d1c322b3b72a951397/torchtitan/experiments/rl/unified/simple_grpo.py#L334

More generally, all the values in this log match between baseline and compile + cudagraph on which I believe means the numerics are identical.

Lucaskabela:
We had a question from @joecummings around the vllm Native definition roofline - I did a quick experiment to test and here is the comparison (both compilations use eager + piecewise cudagraphs for Qwen3/0.6B)

                                                                                                                                                                          
  |                          | Baseline (eager) | vllm Wrapper compile| vLLM Native compile |                                                                                             
  |--------------------------|------------------|-------------------|----------|                                                                                             
  | Generator init + sync    | 10.21s           | 60.92s            | 14.78s |

  Compile Wrapper Overhead: torch.compile 16.35s + CUDA graph capture 33s
  vLLM Native Compile Overhead: torch.compile 4.62 + Cudagraph capture 1s

  ### Totals

  |              | Baseline | vllm Wrapper compile | vLLM Native compile |
  |--------------|----------|--------------------|---------|
  | Generate     | 240.03s  | 15.87s             |  9.35s|
  | Train        | 151.02s  | 152.50s            | 147.45s  |
  | Sync         | 88.66s   | 93.26s             | 84.73s  |
  | **Total**    | **479.87s** | **261.71s**     | 241.61s |

So there is definitely some more improvement to make here - the biggest takeways I see are as follows:

1) The difference in how these models handle parallelism (DTensor for vLLM Wrapper, raw communicaiton collectives with optimizations in vLLM Native) is the largest difference and probably has the biggest impact on startup time as well as runtime 

2) There is some room for us to investigate the vLLM Wrapper startup time and improve this - I will take an action item on this following this PR

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
 
+@dataclass(kw_only=True, slots=True)
+class GeneratorCompilationConfig:
+    """Compilation and CUDA graph settings for the vLLM generator."""
+
+    compilation_backend: Literal["none", "eager", "inductor"] = "eager"
+    """torch.compile backend for vLLM.
+    When set to a value other than "none", enables compilation with the specified backend.
+    See https://docs.vllm.ai/en/stable/api/vllm/config/#vllm.config.CompilationConfig.backend
+    NOTE: inductor will offer the best performance, but will impact numerics - use eager for
```
wwwjn, line 41: Thanks for the detailed note here

wwwjn: And a n00b q: In compile, "eager" means using Dynamo to capture the graph, and use Eager mode when dispatch (no compile), right? 

If so, the graph capturing increased the setup time, but it's required for cudagraph. Cudagraph contributes to the speed up. Is this a correct understanding?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
+    bitwise identical results."""
+
+    cudagraph_mode: Literal[
+        "none", "piecewise", "full", "full_and_piecewise"
+    ] = "piecewise"
+    """CUDA graph capture mode for vLLM.
+    NOTE: Full graph is not compatible with TP > 1 due to DTensor collectives.
```
wwwjn: Full graph only works with single GPU for trainer / generator?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
+from vllm.config import AttentionConfig, CompilationConfig
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass(kw_only=True, slots=True)
+class GeneratorCompilationConfig:
```
tianyu-l: To be consistent with https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L329
```suggestion
class GeneratorCompileConfig:
```

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
+    """Compilation and CUDA graph settings for the vLLM generator."""
+
+    compilation_backend: Literal["none", "eager", "inductor"] = "eager"
```
tianyu-l: what's the difference between eager and none?
specifically, why eager is a compile backend?

```suggestion
    backend: Literal["none", "eager", "inductor"] = "eager"
```

torchtitan/experiments/rl/unified/actors/generator.py
tianyu-l: I'm surprised by this claim. Do we know why?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
+    cudagraph_mode: Literal[
+        "none", "piecewise", "full", "full_and_piecewise"
+    ] = "piecewise"
+    """CUDA graph capture mode for vLLM.
+    NOTE: Full graph is not compatible with TP > 1 due to DTensor collectives.
+    NOTE: Piecewise graph capture requires torch.compile for graph capture and splitting
+    See https://docs.vllm.ai/en/latest/design/v1/torch_compile.html#cuda-graph for more details."""
+
+    @property
+    def enforce_eager(self) -> bool:
```
```suggestion
    def is_eager(self) -> bool:
```

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
+        self, tp_degree: int
+    ) -> CompilationConfig | None:
+        """Build a vLLM ``CompilationConfig``, or return ``None`` when both
+        compilation and CUDA graphs are disabled.
+
+        Resolves cudagraph mode for TP compatibility (full capture hangs with
+        TP > 1 due to DTensor collectives) and validates that piecewise capture
+        is not requested without torch.compile.
+        """
+        if self.compilation_backend == "none" and self.cudagraph_mode == "none":
```
tianyu-l: use shortcut since you created one

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,28 +6,96 @@
...
+    NOTE: Full graph is not compatible with TP > 1 due to DTensor collectives.
+    NOTE: Piecewise graph capture requires torch.compile for graph capture and splitting
+    See https://docs.vllm.ai/en/latest/design/v1/torch_compile.html#cuda-graph for more details."""
+
+    @property
+    def enforce_eager(self) -> bool:
+        """Inferred from compilation_backend and cudagraph_mode."""
+        return self.compilation_backend == "none" and self.cudagraph_mode == "none"
+
+    def resolve_vllm_compilation_config(
```
tianyu-l: Don't resolve, just validate -- throw exceptions explicitly

Maybe put in `__post_init__` in the config class.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -76,35 +144,28 @@ class Config(Configurable.Config):
         model_dtype: str = "bfloat16"
         """Data type for model weights, passed directly to vLLM (auto, float16, bfloat16, float32)."""
 
-        gpu_memory_limit: float = 0.9
+        gpu_memory_limit: float = 0.5
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        enforce_eager: bool = True
-        """Disable CUDA graphs in vLLM (use eager execution)."""
+        compilation: GeneratorCompilationConfig = field(
```
```suggestion
        compile: GeneratorCompilationConfig = field(
```

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -135,14 +198,17 @@ def __init__(
             # tells vLLM to run one worker per process (no subprocess spawning)
             distributed_executor_backend="external_launcher",
             gpu_memory_utilization=config.gpu_memory_limit,
-            enforce_eager=config.enforce_eager,
-            # Disable vLLM logging to avoid noisy logs on every step
-            disable_log_stats=True,
+            enforce_eager=config.compilation.enforce_eager,
```
tianyu-l: this is redundant info in addition to `vllm_compilation_config` -- what's the point of having it in vllm? Does it change anything if we don't send it?

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
```diff
@@ -93,6 +112,12 @@ def create_torchtitan_config_from_vllm_config(
     return parallel_dims, parallelism
 
 
+@support_torch_compile(
+    dynamic_arg_dims={
```
wwwjn, line 116: What is dynamic_arg_dims do?

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: I think @acisseJZhong asked this too and i marked resolved - I will comment in code. 

Eager here means `dynamo` graph capture.  (so like torch.compile(backend='eager')).  It is unfortunately overloaded in naming

torchtitan/experiments/rl/unified/models/attention.py
```diff
@@ -99,33 +104,28 @@ def forward(
...
 
-        # vLLM attention returns (num_tokens, hidden_size) where hidden_size = num_heads * head_dim
+        # vLLM attention returns (num_tokens, num_heads/num_kv_heads * head_dim)
         output_flat = self.vllm_attn(q, k, v)
 
-        # Output is (batch * seq_len, num_heads * head_dim), reshape to (batch, seq_len, num_heads, head_dim)
-        output = output_flat.view(batch_size, seq_len, num_heads, head_dim)
+        # vLLM's flash attention backend may pad the token count (e.g.
+        # round up to an even number), which introduces a new symbolic
+        # shape under torch.compile.  Narrow to trim this
```
wwwjn:
```suggestion
        # shape under torch.compile.  Narrow to trim padding 
```

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: Partially correct - "eager" mode does execute the captured (compiled) graph 

See https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/torch.compiler_faq.html#why-is-my-code-crashingfor a bit more info on the difference in the backends

And yes, we use that graph with cudagraphs afterwards to contribute to the speedup

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: I disentangled this - we now do the validation in __post__init__, and retunr the config in this function and change the name to better reflect that

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: We are required to send it - enforce_eager is a separate flag from the compilation config for vLLM and inside the vLLM engine is not inferred but just read, so we need to infer and pass it ourselves

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: This just compiles with dynamic shapes for that dimension.  This means we will not recompile if we get a different batch size passed in which may happen from vLLM scheduler

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: Need to update this comment - this was a relic of the pre DTensor changes, but I believe this is working now with the DTensor monkey patch

Lucaskabela: This was because the issue with the DTensor weak_ref registration - just retried on top of my changes and full_and_piecewise is now working

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -43,7 +44,11 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
-            enforce_eager=True,
+            gpu_memory_limit=0.5,
```
Lucaskabela: NOTE: should remove these (this was from pre rebase on Joe's change)

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
tianyu-l: I forgot the difference, but using `to_local()` (a differentiable, functional op) here sounds not as "correct" as `tensor._local_tensor` (hackily getting the actual local plain tensor).

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -20,14 +21,62 @@
...
+    NOTE: "eager" means compile with dynamo backend (like torch.compile(backend="eager"))
+    NOTE: inductor will offer the best performance, but will impact numerics - use eager for
+    bitwise identical results."""
+
+    cudagraph_mode: Literal[
+        "none", "piecewise", "full", "full_and_piecewise"
+    ] = "piecewise"
+    """CUDA graph capture mode for vLLM.
+    NOTE: Piecewise graph capture requires torch.compile for graph capture and splitting
+    See https://docs.vllm.ai/en/latest/design/v1/torch_compile.html#cuda-graph for more details."""
```
tianyu-l: I didn't see `"none", "piecewise", "full", "full_and_piecewise"` in the link. I think it's worth adding brief explanation for each option here. E.g.
- is `piecewise` similar to fullgraph=False in torch.compile?
- `full_and_piecewise` is a very confusing name

tianyu-l (commented):
sgtm, will let @wwwjn review/stamp

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -20,14 +21,62 @@
...
+
+    def __post_init__(self) -> None:
+        if self.backend == "none" and self.cudagraph_mode in (
```
acisseJZhong, line 57: qq: so cuda graph mode `full` doesn't require torch compile to be turned on? I thought cuda graph capture is a stage in torch compile.

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: Good catch - I updated the link so it is clearer and added these descriptions

Lucaskabela: > is piecewise similar to fullgraph=False in torch.compile?

Somewhat, it does not require all operations to be cudagraphable (mainly for Flash Attention/custom kernel support). But to be clear, torch.compile always runs with fullgraph=True in vLLM

> full_and_piecewise is a very confusing name

It just runs with both modes and selects which graphs to use based on memory/dynamism constraints

torchtitan/experiments/rl/unified/actors/generator.py
Lucaskabela: It does if relying on native torch.compile; vLLM has a different cudagraph integration which does not require torch.compile for full, but for best performance torch.compile + piecewise is recommended (default settings in vLLM when enforce_eager=False)

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -79,8 +139,8 @@ class Config(Configurable.Config):
         gpu_memory_limit: float = 0.9
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        enforce_eager: bool = True
-        """Disable CUDA graphs in vLLM (use eager execution)."""
+        compile: GeneratorCompileConfig = field(default_factory=GeneratorCompileConfig)
```
wwwjn: Technically `Sampling config` is not equivalent to engine_config. Do you think this can be fit int VLLMGenrator.Config

torchtitan/experiments/rl/unified/models/attention.py
pianpwk: just curious, is this for perf, numerical errors, to get around some data-dependent error, or something else?

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: So we can support all 4 mode ("none", "piecewise", "full", "full_and_piecewise") now

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
pianpwk: maybe a dumb question, what's the interaction b/w cudagraph & DTensors? I thought cudagraph would be done per-rank, on local tensors, so it wouldn't see DTensors?

wwwjn (approved):
Thanks for the awesome work, lgtm

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
pianpwk: not sure how this affects things, but for inference we've been wanting to mark sequence length as unbacked

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
miladm: Would be interesting to learn the intersection of this work with spmd_types

cc @ezyang @zou3519 for viz

torchtitan/experiments/rl/unified/models/attention.py
Lucaskabela: It is an numerical error caused by vllm_attn kernel padding when batch_size and seq_len are 1

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: > Would be interesting to learn the intersection of this work with spmd_types

My understanding is that If SPMD type are gone during runtime it shouldn't be an issue as this issue came from when cudagraphs are captured during warmup running

> maybe a dumb question, what's the interaction b/w cudagraph & DTensors? I thought cudagraph would be done per-rank, on local tensors, so it wouldn't see DTensors?

It is - the issue is with the vLLM `weak_ref_tensor` being used to save output for replay; when that method is called on DTensor, it leads to a runtime failure as it is not implemented correctly (tries to call DataPtr).  So the simple fix is to monkeypatch the weak_ref_tensor to support DTensor by doing what cudagraph does - get the local tensor

torchtitan/experiments/rl/unified/models/vllm_wrapper.py
Lucaskabela: I believe @laithsakka is planning on making this default for vLLM compile soon :)

PR #2487, opened 2026-03-05
[3/N] Support qlora by nf4tensor

mori360:
### Summary

  - Add QLoRA (quantized LoRA) support via NF4 quantization of base weights using torchao's to_nf4
  - Base weights are quantized to NF4 via a one-shot forward pre-hook that fires before the first forward pass and self-removes, while LoRA adapters remain in
   full precision (float32)
  - Add quantize_base and nf4_scaler_block_size config fields to LoRAConverter.Config
  - Handle DTensor for distributed training compatibility
  - Proper error handling: ImportError with chained exception for missing torchao, ValueError for invalid quantize_base values and incompatible tensor
  dimensions
  - Add llama3_debugmodel_qlora debug config (derives from llama3_debugmodel_lora to inherit checkpoint settings)

### Test Plan

- test_qlora_base_weights_quantized_adapters_full_precision
For QLoRA with NF4Tensor, before first forward: asserts fc1.weight is NOT NF4Tensor (quantization hasn't fired yet). Calls model(torch.randn(2, 64)) — this triggers the one-shot register_forward_pre_hook that calls to_nf4() on all base weights. After first forward: for both fc1 and fc2, asserts weight.data IS an NF4Tensor, while lora_a.weight.dtype and lora_b.weight.dtype remain torch.float32. This validates the core QLoRA property: compressed base weights + full-precision adapters.

* #2569
* #2488
* __->__ #2487
* #2485
* #2484

### DIFF
--- a/tests/unit_tests/test_model_converter.py
+++ b/tests/unit_tests/test_model_converter.py
@@ -86,11 +86,15 @@ def test_lora_before_quantization_raises():
 
 def test_lora_freeze_and_trainability():
     """After convert: base params frozen, LoRA adapters present and trainable."""
-    model = nn.Sequential(OrderedDict([
-        ("fc1", nn.Linear(64, 64)),
-        ("relu", nn.ReLU()),
-        ("fc2", nn.Linear(64, 64)),
-    ]))
+    model = nn.Sequential(
+        OrderedDict(
+            [
+                ("fc1", nn.Linear(64, 64)),
+                ("relu", nn.ReLU()),
+                ("fc2", nn.Linear(64, 64)),
+            ]
+        )
+    )
     converter = LoRAConverter(LoRAConverter.Config(rank=4, alpha=8.0))
     converter.convert(model)
 
@@ -118,11 +122,15 @@ def test_lora_freeze_and_trainability():
 def test_lora_trains_base_frozen():
     """Train for several steps: LoRA params should change, base params should not."""
     torch.manual_seed(42)
-    model = nn.Sequential(OrderedDict([
-        ("fc1", nn.Linear(64, 64)),
-        ("relu", nn.ReLU()),
-        ("fc2", nn.Linear(64, 64)),
-    ]))
+    model = nn.Sequential(
+        OrderedDict(
+            [
+                ("fc1", nn.Linear(64, 64)),
+                ("relu", nn.ReLU()),
+                ("fc2", nn.Linear(64, 64)),
+            ]
+        )
+    )
     converter = LoRAConverter(LoRAConverter.Config(rank=4, alpha=8.0))
     converter.convert(model)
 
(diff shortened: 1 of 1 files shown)


torchtitan/components/lora.py
```diff
@@ -133,5 +151,64 @@ def converter_key_filter(key: str) -> bool:
...
+                ) from err
+
+            lora_classes = tuple(_lora_class_cache.values())
+            nf4_scaler_block_size = self.nf4_scaler_block_size
+
+            def _to_nf4_tensor(weight: torch.Tensor) -> torch.Tensor:
+                """Convert weight to NF4, handling both regular tensors and DTensors."""
+                nf4_block_size = 64  # NF4 default block size
+                is_dtensor = isinstance(weight, DTensor)
+                local_weight = weight.to_local() if is_dtensor else weight
```
tianyu-l, line 160: dtensor.to_local() should never be used without `grad_placements`, o/w we could lose necessary reductions on gradients.

torchtitan/components/lora.py
```diff
@@ -133,5 +151,64 @@ def converter_key_filter(key: str) -> bool:
...
+                        nf4_local, weight.device_mesh, weight.placements
+                    )
+                return nf4_local
+
+            def _quantize_hook(
+                mod: nn.Module, args: Any, handle: torch.utils.hooks.RemovableHandle
+            ) -> None:
+                for sub in mod.modules():
+                    if isinstance(sub, lora_classes):
+                        sub.weight = nn.Parameter(
```
tianyu-l, line 188: Do we have to do this conversion every single time the forward is called?

tianyu-l: what's the result of a nf4 @ nf4 matmul
- if it's bf16, then you have to convert the activations to a LoRA linear
- if it's nf4, then other modules like rmsnorm and attention also needs to be able to work with nf4

tianyu-l (changes requested):

torchtitan/components/lora.py
mori360: Yeah, it's a prototype to proof that we could run with NF4. We only convert origin weights to nf4, which are freezed(requires_grad=False), so that losing gradients here should be fine. If we decide to support qlora later, We will later corporate with torchao to support better on dtensor

torchtitan/components/lora.py
mori360: activation: bf16, weight: nf4
during forward, we would first dequantizes nf4 to bf16 first, so it's computation at bf16
nf4 @ bf16 -> bf16
we only have the quantization on origin weights, and the outputs are all in bf16, so that later computation are not affected, and we only save the memory for the origin weights (bf16 -> nf4)

PR #2488, opened 2026-03-05
[4/N] Add QAT (quantization-aware training) model converter

mori360:
### Summary
 #### What is QAT

  QAT trains a model to be robust to quantization by inserting fake quantization during training. Fake quantization simulates the precision loss of quantized inference (e.g. int4, float8) while keeping weights in their original dtype (float32/bfloat16) so gradients flow normally. At the end of training, a CONVERT step replaces fake-quantized modules with real quantized ones (e.g. FakeQuantizedLinear → WeightOnlyInt4Linear with packed int4 weights).

  Training time:  nn.Linear → FakeQuantizedLinear (full-precision weights, quantize/dequantize in forward)
  Deploy time:    FakeQuantizedLinear → WeightOnlyInt4Linear (real int4 packed weights)

  The implementation wraps torchao's quantize_(model, QATConfig(base_config, step=PREPARE)). The scheme config selects which quantization format to simulate (int4, intx, float8, nvfp4, mx, etc.).

####  QAT + LoRA

When QAT and LoRA are composed, QATConverter runs first so that nn.Linear is replaced with FakeQuantizedLinear. LoRA then dynamically inherits from FakeQuantizedLinear, creating LoRAFakeQuantizedLinear. LoRA also applies the same QAT fake quantization to the newly created adapter linears (lora_a, lora_b), so the entire forward pass trains under simulated quantization:

  - Base weights: fake-quantized
  - LoRA adapters: fake-quantized (same scheme, group_size clamped to LoRA rank, caution: low rank lora_b could not apply high group_size for some QAT configs)

  This ensures the model learns to produce good outputs under the precision constraints that will apply at deployment, including through the adapter path.

 At the end of training, finalization runs in reverse converter order:
  1. LoRA merge: fold adapters into base weights (W ← W + α/r * B @ A)
  2. QAT CONVERT: replace FakeQuantizedLinear with real quantized modules
Screenshot 2026-03-20 at 3 52 45 PM

  - **`QATConverter`** (`torchtitan/components/quantization/qat.py`): New model converter that applies QAT
  fake quantization via torchao's `quantize_` with `QATStep.PREPARE`. Supports 7 schemes including
  `int4_weight_only` (default), `int8_weight_only`, `intx_weight_only`, and float8 variants. Config fields:
  `scheme`, `group_size`, `convert_at_end`. The `finalize()` method runs `QATStep.CONVERT` to replace
  fake-quantized modules with real quantized ones (skipped if LoRA adapters are still present).

  - **LoRA + QAT integration** (`lora.py`): When `QATConverter` is applied before `LoRAConverter`, the LoRA
  `convert()` detects `_qat_scheme` on the model and calls `_apply_adapter_qat()` to apply the same fake
  quantization to adapter linears (lora_a, lora_b). Group size is clamped to LoRA rank for compatibility.
  `init_weights` is made QAT-aware: uses `normal_(std=0.02)` instead of `zeros_` for lora_b when it's a
  `FakeQuantizedLinear` (avoids NaN from dynamic scaling of all-zeros).

  - **Validation** (`model_converter.py`): `_validate_converter_ordering` extended to check QAT ordering (must
   come before LoRA). New `_validate_qat_lora_finalize` warns if `convert_at_end=True` but LoRA
  `merge_adapter=False` (CONVERT would be skipped).

  - **Configs**: Add `llama3_debugmodel_qat`, `llama3_debugmodel_qat_lora`,
  `llama3_debugmodel_qat_lora_merged` debug configs.

  ### Test plan

  - [ ] `pytest tests/unit_tests/test_model_converter.py -x` — new tests: `test_qat_all_schemes`
  (parametrized, 7 schemes), `test_qat_unknown_scheme_raises`,
  `test_qat_group_size_warning_for_unsupported_scheme`, `test_qat_lora_adapter_qat`
  - [ ] `torchrun --nproc_per_node=4 -m torchtitan.train --module llama3 --config llama3_debugmodel_qat` —
  verify QAT-only training
  - [ ] `torchrun --nproc_per_node=4 -m torchtitan.train --module llama3 --config llama3_debugmodel_qat_lora`
  — verify QAT + LoRA
  - [ ] `torchrun --nproc_per_node=4 -m torchtitan.train --module llama3 --config
  llama3_debugmodel_qat_lora_merged` — verify QAT + LoRA with merge + CONVERT

* #2569
* __->__ #2488
* #2485
* #2484

### DIFF
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -58,13 +58,21 @@ def apply_lora(linear: nn.Linear, rank: int, alpha: float) -> nn.Linear:
                 device = self.weight.device
                 dtype = self.weight.dtype
                 self.lora_a = (
-                    Linear.Config(bias=False)
-                    .build(in_features=self.in_features, out_features=rank)
+                    Linear.Config(
+                        in_features=self.in_features,
+                        out_features=rank,
+                        bias=False,
+                    )
+                    .build()
                     .to(device=device, dtype=dtype)
                 )
                 self.lora_b = (
-                    Linear.Config(bias=False)
-                    .build(in_features=rank, out_features=self.out_features)
+                    Linear.Config(
+                        in_features=rank,
+                        out_features=self.out_features,
+                        bias=False,
+                    )
+                    .build()
                     .to(device=device, dtype=dtype)
                 )
 


torchtitan/models/llama3/config_registry.py
```diff
@@ -144,6 +145,32 @@ def llama3_debugmodel_qlora() -> Trainer.Config:
     return config
 
 
+def llama3_debugmodel_qat() -> Trainer.Config:
+    config = llama3_debugmodel()
+    config.model_converters = ModelConvertersContainer.Config(
+        converters=[
+            QATConverter.Config(),
```
tianyu-l, line 138: similar to the comment in [1/N]: order matters

torchtitan/components/quantization/qat.py
```diff
@@ -0,0 +1,77 @@
...
+        logger.info(
+            f"QAT training active (dtype={self.dtype}, group_size={self.group_size})"
+        )
+
+    def convert(self, model: nn.Module) -> None:
+        from torchao.quantization.qat import FakeQuantizedLinear, IntxFakeQuantizeConfig
+        from torchao.quantization.quant_primitives import TorchAODType
+
+        dtype_map = {
+            "int4": TorchAODType.INT4,
```
andrewor14: This was only for torch 2.5 and before (which didn't have torch.int4), I think now you can just use torch.int4 directly here

torchtitan/components/quantization/qat.py
```diff
@@ -0,0 +1,77 @@
...
+    learns to compensate for quantization error.
+
+    When composed with LoRA (QATConverter listed before LoRAConverter in converters),
+    LoRA will inherit from FakeQuantizedLinear so base weights are fake-quantized
+    while LoRA adapters stay full-precision.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        dtype: Literal["int4", "int8"] = "int4"
```
andrewor14: Should we support activation quantization too? There are pre-set QAT configs in torchao that do this, e.g. `Int8DynamicActivationIntxWeightConfig` (used primarily for edge) can be passed to `QATConfig` as a base config ([docs here](https://docs.pytorch.org/ao/main/workflows/qat.html#quantize-api-recommended), a bit outdated).

Another example that I think will be important is `NVFP4DynamicActivationNVFP4WeightConfig`. I don't think a single `dtype` field will be sufficient in capturing that

torchtitan/components/quantization/qat.py
```diff
@@ -0,0 +1,77 @@
...
+        weight_config = IntxFakeQuantizeConfig(
+            dtype=torch_dtype,
+            group_size=self.group_size,
+            is_symmetric=True,
+        )
+
+        def _replace_recursive(parent: nn.Module) -> None:
+            for name, child in list(parent.named_children()):
+                if isinstance(child, nn.Linear):
+                    fq = FakeQuantizedLinear.from_linear(
```
andrewor14: Just for future reference but there's also [`NVFP4FakeQuantizedLinear`](https://github.com/pytorch/ao/blob/ab4a3366669733d3c3d861c317f29d4da81f7611/torchao/prototype/qat/nvfp4.py#L103) that's not a subclass of this. We can add support for this later but just wanted to bring it up early so we're not baked into this specific linear class

torchtitan/components/quantization/qat.py
```diff
@@ -0,0 +1,77 @@
...
+    LoRA will inherit from FakeQuantizedLinear so base weights are fake-quantized
+    while LoRA adapters stay full-precision.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        dtype: Literal["int4", "int8"] = "int4"
+        """Data type for fake quantization. Supported: 'int4', 'int8'."""
+
+        group_size: int = 256
```
andrewor14: By the way not all quantization configs have `group_size`, you can see the list of supported QAT configs [here](https://github.com/pytorch/ao/blob/03ca8e22a9db0da76cd4999dbc307b53991fd202/torchao/quantization/qat/fake_quantize_config.py#L347)

torchtitan/protocols/model_converter.py
```diff
@@ -87,20 +87,21 @@ def post_optimizer_hook(self, model: nn.Module | list[nn.Module]):
 def _validate_converter_ordering(converters: list[Configurable.Config]):
     """Validates that converters are in the correct order.
 
-    LoRA must come after quantization because quantization replaces nn.Linear
-    with specialized subclasses (e.g. Float8Linear), and LoRA dynamically
-    inherits from whatever linear class it wraps.
+    LoRA must come after quantization and QAT because both replace nn.Linear
```
andrewor14, line 128: Does the `LoRAConverter` already handle the case when the inner linear is `FakeQuantizedLinear`? I think additionally we will need to make the lora adapters also fake quantized (according to the same configs, e.g. int4). Does this part exist yet?

mori360: Currently we follow: Linear -> QAT(Linear) -> LoRA(QAT) that only adds adapters to the QATLinear, so the currently design has lora adapters in dtype same as the origin linear.

PR #2490, opened 2026-03-05
Temporarily disable CPU offload CI tests to unblock

wwwjn:
- "Enable CPU Offload, Optimizer in backward with TP, DP, CP" test takes ~19mins. Even if I change the timeout time to be 60min, the CI timeout on the same test.
image


fegin (approved):
Can we also add a issue to track this so that we won't forget to turn it back on?

PR #2497, opened 2026-03-05
[RL] Fix vLLM wrapper crash with configurable Embedding

daniellepintz:
After https://github.com/pytorch/torchtitan/pull/2428/ RL training is hanging on main; with a try-except the error is revealed to be `AttributeError: 'Config' object has no attribute 'num_embeddings'`

The root cause is the logger.debug line in `vllm_wrapper.py` calls repr() on the model config, which fails because Embedding.Config has `field(init=False)` slots (num_embeddings, embedding_dim) that aren't set until build() time. Calling `config.to_dict()` instead of `repr()` safely skips unset fields, and resolves the hang.


wwwjn:
Thanks for the fix!

PR #2498, opened 2026-03-05
[RL] Load weights in trainer with `strict=True`

daniellepintz:
Pass `strict=True` to `set_model_state_dict` since this doesn't error and is safer for the future


wwwjn (approved):
It's safe to change this field given our current supported parallelism. And we also plan to remove `set_model_state_dict` API as @fegin  is refactoring checkpoint

PR #2512, opened 2026-03-06
[Config] Overwrite _replace for HFTransformerModel.Config

fegin:
* __->__ #2512

```
torchrun --nproc_per_node=8 --rdzv_backend c10d --rdzv_endpoint=localhost:0   --tee 3 -m torchtitan.train --module llama3 --config llama3_debugmodel --module transformers_modeling_backend --config transformers_modeling_backend_debugmodel --hf_model Qwen/Qwen2.5-7B --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.pipeline_parallel_degree 2 --parallelism.pipeline_parallel_schedule 1F1B
```

This command is broken. This PR fixes it. Basically, `HFTransformerModel` does two things 1) titan_dense_config is not defined in the dataclass but in __init__. 2) update_from_config will dynamically create some fields.

### DIFF
--- a/torchtitan/config/configurable.py
+++ b/torchtitan/config/configurable.py
@@ -4,9 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import copy
 import dataclasses
-from dataclasses import dataclass, fields
+from dataclasses import dataclass, fields, replace
 from typing import ClassVar
 
 import torch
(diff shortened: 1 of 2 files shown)


fegin:
```
torchrun --nproc_per_node=8 --rdzv_backend c10d --rdzv_endpoint=localhost:0   --tee 3 -m torchtitan.train --module llama3 --config llama3_debugmodel --module transformers_modeling_backend --config transformers_modeling_backend_debugmodel --hf_model Qwen/Qwen2.5-7B --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.pipeline_parallel_degree 2 --parallelism.pipeline_parallel_schedule 1F1B
```

This command is broken. This PR fixes it. Basically, `HFTransformerModel` does two things 1) titan_dense_config is not defined in the dataclass but in __init__. 2) update_from_config will dynamically create some fields.

tianyu-l:
@fegin got it. If it's just for `HFTransformerModel`, can we fix it therein, e.g. by putting `titan_dense_config` and any others in its config? I don't feel it's necessary to adapt to the pattern when core titan doesn't need it. The HF folder is more-or-less a hack anyway.

fegin:
@tianyu-l That's my original fix. However, I felt that we don't actually have a strict contract for `update_from_config()`. So I change `_replace` to do copy. One nice thing for keeping `_replace` to use `replace` is that we enforce implicit contract that `update_from_config()` cannot add extra fields. If that's the desired contract, I can limit the fix to HFTransformerModel only.

tianyu-l (approved):

PR #2521, opened 2026-03-07
reverted trunc_normal_ implementation

francesco-bertolotti:
I retested the issue ([https://github.com/pytorch/torchtitan/issues/2269](https://github.com/pytorch/torchtitan/issues/2269)) using **torch 2.12.0.dev20260306+cu126**. The updated `trunc_normal_` now behaves very similarly to performing the initialization in `float32` and then casting to `bfloat16`. Since there’s no longer a clear reason to keep the custom implementation, I’m removing it.


francesco-bertolotti:
For reference, here's two training run with the two initialization methods
vsinit

Issue #2495, opened 2026-03-05
Should fully_shard policy change for tied embedding?


francesco-bertolotti:
Consider the PR https://github.com/pytorch/pytorch/pull/1762 stemming from issue https://github.com/pytorch/pytorch/issues/176076

If it lands, it seems that in the future FSDP2 on tied weight with separate calls won't be allowed:
```python
model.output.weight = model.tok_embeddings.weight
...
fully_shard(model.tok_embeddings.weight)
...
fully_shard(model.output.weight)
...
```

Instead, when embedding are tied, as suggest https://github.com/pytorch/pytorch/pull/176225#discussion_r2874905548,  one should:
```python
model.output.weight = model.tok_embeddings.weight
...
fully_shard([model.tok_embeddings.weight,model.output.weight])
```

What do you think?

tianyu-l:
that's fine, please submit proper fixes

PR #2522, opened 2026-03-07
tied parameters in the fsdp group

francesco-bertolotti:
This PR addresses https://github.com/pytorch/torchtitan/issues/2495

Briefly, since tied parameters are supposed to be put in the same FSDP group (see https://github.com/pytorch/pytorch/pull/176225#discussion_r2874905548), I changed `torchtitan/models/llama4/parallelize.py` to group `molde.output, model.norm`, and `model.tok_embeddings` in the case `model.enable_weight_tying` is set to true

### DIFF
--- a/torchtitan/models/llama4/parallelize.py
+++ b/torchtitan/models/llama4/parallelize.py
@@ -393,7 +393,7 @@ def apply_fsdp(
                 f"Invalid reshard_after_forward_policy: {reshard_after_forward_policy}."
             )
 
-    if getattr(model.enable_weight_tying, "enable_weight_tying", False):
+    if getattr(model, "enable_weight_tying", False):
         modules = [
             m for m in (model.tok_embeddings, model.norm, model.output) if m is not None
         ]


tianyu-l:
sgtm, @weifengpy please help review

tianyu-l:
errors seem real

francesco-bertolotti:
There was a bug, but it was unrelated to the pyrefly errors. I wasn’t sharding `model.norm` with `model.output` when the embeddings were not tied.

The issue is that pyrefly reports that `model.tok_embedding`, `model.output`, and `model.norm` in `parallelize.py` can be either a `Tensor` or an `nn.Module`. I didn’t investigate it too deeply, but we can work around it by using `pyrefly: ignore [no-matching-overload]` as it was set before.

torchtitan/models/llama4/parallelize.py
```diff
@@ -391,13 +393,33 @@ def apply_fsdp(
                 f"Invalid reshard_after_forward_policy: {reshard_after_forward_policy}."
             )
 
-    if model.tok_embeddings is not None:
+    if model.enable_weight_tying:
```
tianyu-l: Not all models have this field, e.g. DeepSeek V3
You'll have to workaround by doing https://github.com/pytorch/torchtitan/blob/e1847f4392fc8f91dc577fdcc59a4c91bc027465/torchtitan/models/utils.py#L432

tianyu-l (changes requested):

torchtitan/models/llama4/parallelize.py
francesco-bertolotti: I am using `getattr(model.enable_weight_tying, "enable_weight_tying", False)`, what do you think?

tianyu-l:
It seems the CI test failed again https://github.com/pytorch/torchtitan/actions/runs/22917480220/job/66508486101?pr=2522#step:16:1010

Could you try the trick I provided? Thanks!

francesco-bertolotti:
@tianyu-l 

It should be fixed. I have tested it with deepseek. Before I was using a single GPU which was not triggering  fully_shard.

PR #2525, opened 2026-03-09
[RL] Use gpu direct weight sync from torchstore

daniellepintz:
In https://github.com/meta-pytorch/torchstore/pull/138 we add GPU direct weight sync with RDMA to TorchStore. This allows the Generator GPUs to directly access the weights in memory on the Trainer GPUs, with no extra CPU buffer.

### No resharding
* Current main of `torchtitan/experiments/rl`: 21.250s
* Using MonarchRDMATransportBuffer: 4.894s
* Using GPU direct RDMA (this PR): 0.067s
* speedup = 317x

### With resharding
* Current main of `torchtitan/experiments/rl`: 22.944s
* Using MonarchRDMATransportBuffer: 5.339s
* Using GPU direct RDMA (this PR): 0.076s
* speedup = 301x

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -8,12 +8,29 @@ from collections.abc import Callable
 
 import torch
 import torch.distributed as dist
+import torch.nn as nn
 from torch.distributed.tensor import DTensor, Replicate
+from torch.fx.traceback import annotate_fn
 from torch.utils._pytree import register_pytree_node, tree_map
 
 from torchtitan.config import CompileConfig
 from torchtitan.distributed import ParallelDims
 from torchtitan.tools.logging import logger
+
+_AC_REGION_ID = "ac_region_id"
+
+
+def annotate_ac_regions(model: nn.Module) -> None:
+    """Annotate each transformer block with a unique AC region ID.
+
+    This enables apply_sac_pass to assign different ac_graph_id values
+    per block, creating AC region boundaries between transformer blocks.
+    """
+    layers = model.get_submodule("layers")
+    for layer_id, transformer_block in layers.named_children():
+        transformer_block.forward = annotate_fn({_AC_REGION_ID: int(layer_id)})(
+            transformer_block.forward
+        )
 
 
 def parallelize_inputs(parallel_dims, args, kwargs):
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -11,6 +11,7 @@ This module provides a cudagraph pass that can be applied to graph modules
 during compilation.
 """
 
+import logging
 import warnings
 from collections.abc import Callable, Sequence
 from typing import Any
@@ -19,45 +20,103 @@ import torch
 from torch._inductor.cudagraph_trees import _use_cuda_memory_pool_manager
 from torch.utils._ordered_set import OrderedSet
 
-
-def init_global_graph_pool() -> tuple[
-    torch.cuda.CUDAGraph, torch.cuda._POOL_HANDLE, torch.cuda.Stream
-]:
-    dummy_graph = torch.cuda.CUDAGraph()
-
-    # create a global cudagraph memory pool to allow memory reuse across cudagraphs.
-    graph_pool = torch.cuda.graph_pool_handle()
-
-    # create a global cuda stream for graph capture. we need to use a single stream
-    # for all allocations to the memory pool, otherwise the allocations to separate streams
-    # will not be used.
-    graph_capture_stream = torch.cuda.Stream()
-
-    # use a dummy graph to keep the global graph pool alive
-    with (
-        # suppress an empty cudagraph warning, since we intentionally create
-        # an empty cudagraph here
-        warnings.catch_warnings(record=True),
-        torch.cuda.graph(
-            dummy_graph,
-            pool=graph_pool,
-            stream=graph_capture_stream,
-            capture_error_mode="thread_local",
-        ),
-    ):
-        pass
-
-    return dummy_graph, graph_pool, graph_capture_stream
+logger = logging.getLogger(__name__)
 
 
-(
-    _global_dummy_graph,
-    _global_graph_pool,
-    _global_graph_capture_stream,
-) = init_global_graph_pool()
+class _CUDAGraphManager:
+    """A manager to hold a shared graph pool, stream, and wrapper registry."""
+
+    def __init__(self) -> None:
+        self._initialized = False
+        self._cudagraph_wrappers: list["CUDAGraphWrapper"] = []
+        self._teardown_called = False
+
+    def maybe_initialize(self) -> None:
+        if self._initialized:
+            return
+
+        self._initialized = True
+
+        # create a global cudagraph memory pool to allow memory reuse across cudagraphs.
+        self.graph_pool = torch.cuda.graph_pool_handle()
+
+        # create a global cuda stream for graph capture. we need to use a single stream
+        # for all allocations to the memory pool, otherwise the allocations to separate
+        # streams will not be used.
+        self.stream = torch.cuda.Stream()
+
+        # use a dummy graph to keep the global graph pool alive
+        self._dummy_graph = torch.cuda.CUDAGraph()
+        with (
+            # suppress an empty cudagraph warning, since we intentionally create
+            # an empty cudagraph here
+            warnings.catch_warnings(record=True),
+            torch.cuda.graph(
+                self._dummy_graph,
+                pool=self.graph_pool,
+                stream=self.stream,
+                capture_error_mode="thread_local",
+            ),
+        ):
+            pass
+
+    def register_wrapper(self, wrapper: "CUDAGraphWrapper") -> None:
+        assert not self._teardown_called, "Cannot register new cudagraph after teardown"
+        self._cudagraph_wrappers.append(wrapper)
+
+    def teardown(self) -> None:
+        """Destroy all cudagraphs and release the cudagraph memory pool.
+
+        Note [explicit cudagraph teardown]
+        cudagraph holds reference to nccl which prevents destroy process
+        group. so we need to explicitly delete cudagraph which is held
+        in _CUDAGraphManager and CUDAGraphWrapper. If cudagraph is not
+        used, this is a no-op.
+        """
+        if not self._initialized:
+            return
+        if self._teardown_called:
+            logger.warning("cudagraph manager teardown called twice")
+            return
+
+        for wrapper in self._cudagraph_wrappers:
+            wrapper.teardown()
+        self._cudagraph_wrappers.clear()
+
+        self._dummy_graph = None
+        self.stream = None
+        self.graph_pool = None
+        self._teardown_called = True
+
+
+_cg_manager = _CUDAGraphManager()
+
+
+def cudagraph_teardown() -> None:
+    """Destroy all cudagraphs and release the cudagraph memory pool.
+    See Note [explicit cudagraph teardown] for more details.
+    """
+    _cg_manager.teardown()
 
 
 class CUDAGraphWrapper:
+    """Wraps a callable with cudagraph. It warms up the callable, records cudagraph,
+    and replays cudagraph during runtime. It also handles static input tensors, which
+    are tensors whose tensor addresses do not change across runs.
+
+    Args:
+        runnable: The callable to wrap with CUDA graph. This can be a
+            torch.fx.GraphModule when used in an FX graph pass, or any
+            callable when used in PyTorch eager mode.
+        example_inputs: A list of example inputs to the callable.
+        static_input_indices: A tuple of indices identifying static input
+            tensors. Static inputs are tensors whose memory addresses remain
+            constant across invocations. Common examples include model weights,
+            buffers, and outputs from previously wrapped CUDA graph functions.
+        should_check_address: Whether to verify static input tensor addresses
+            at runtime. This should only be enabled for debugging purposes.
+    """
+
     def __init__(
         self,
         runnable: Callable,
@@ -65,31 +124,32 @@ class CUDAGraphWrapper:
         static_input_indices: tuple[int] | None = None,
         should_check_address: bool = False,
     ):
-        self.runnable = runnable
-        self.graph_pool = _global_graph_pool
-        self.stream = _global_graph_capture_stream
-        self.static_input_indices = OrderedSet(
+        _cg_manager.maybe_initialize()
+        _cg_manager.register_wrapper(self)
+
+        self._runnable = runnable
+        self._static_input_indices = OrderedSet(
             static_input_indices if static_input_indices is not None else []
         )
-        self.input_indices_to_copy = [
+        self._input_indices_to_copy = [
             i
             for i, inp in enumerate(example_inputs)
-            if isinstance(inp, torch.Tensor) and i not in self.static_input_indices
+            if isinstance(inp, torch.Tensor) and i not in self._static_input_indices
         ]
-        self.cudagraph: torch.cuda.CUDAGraph | None = None
-        self.has_warmup = False
+        self._cudagraph: torch.cuda.CUDAGraph | None = None
+        self._has_warmup = False
 
-        self.args = None
-        self.output = None
+        self._args = None
+        self._output = None
 
         # (debug only) whether check static input tensor addresses during runtime
-        self.should_check_address = should_check_address
+        self._should_check_address = should_check_address
 
-    def copy_non_static_inputs(self, *args):
-        for i in self.input_indices_to_copy:
-            self.args[i].copy_(args[i])
+    def _copy_non_static_inputs(self, *args):
+        for i in self._input_indices_to_copy:
+            self._args[i].copy_(args[i])
 
-    def check_input_types(self, inputs) -> None:
+    def _check_input_types(self, inputs) -> None:
         for inp in inputs:
             assert isinstance(inp, (torch.Tensor, int, torch._C.Generator)), (
                 "args must be tensor, integer (for dynamic shapes), "
@@ -97,47 +157,59 @@ class CUDAGraphWrapper:
                 f"but found {type(inp)}"
             )
 
-    def check_static_inputs_address(self) -> None:
-        for i in self.static_input_indices:
-            actual = self.args[i].data_ptr()
-            expected = self.input_addresses[i]
+    def _check_static_inputs_address(self) -> None:
+        for i in self._static_input_indices:
+            actual = self._args[i].data_ptr()
+            expected = self._input_addresses[i]
             assert expected == actual, (
                 "Expected the same static tensor address but found "
                 f"{expected} != {actual}"
             )
 
     def __call__(self, *args):
-        if not self.has_warmup:
-            self.has_warmup = True
+        if not self._has_warmup:
+            self._has_warmup = True
             device = torch.cuda.current_device()
 
             # warmup in cudagraph memory pool to avoid fragmentation
             # across eager memory pool and cudagraph memory pool.
-            with _use_cuda_memory_pool_manager(device, self.graph_pool, self.stream):
-                out = self.runnable(*args)
+            with _use_cuda_memory_pool_manager(
+                device, _cg_manager.graph_pool, _cg_manager.stream
+            ):
+                out = self._runnable(*args)
             return out
 
-        if self.cudagraph is None:
-            self.check_input_types(args)
-            self.args = args
-            self.input_addresses = [
+        if self._cudagraph is None:
+            self._check_input_types(args)
+            self._args = args
+            self._input_addresses = [
                 x.data_ptr() if isinstance(x, torch.Tensor) else None for x in args
             ]
 
-            self.cudagraph = torch.cuda.CUDAGraph()
+            self._cudagraph = torch.cuda.CUDAGraph()
 
             with torch.cuda.graph(
-                self.cudagraph, pool=self.graph_pool, stream=self.stream
+                self._cudagraph,
+                pool=_cg_manager.graph_pool,
+                stream=_cg_manager.stream,
             ):
                 # `output` is managed by pytorch's cudagraph pool
-                self.output = self.runnable(*args)
+                self._output = self._runnable(*args)
 
-        if self.should_check_address:
-            self.check_static_inputs_address()
+        if self._should_check_address:
+            self._check_static_inputs_address()
 
-        self.copy_non_static_inputs(*args)
-        self.cudagraph.replay()
-        return self.output
+        self._copy_non_static_inputs(*args)
+        self._cudagraph.replay()
+        return self._output
+
+    def teardown(self) -> None:
+        """Destroy cudagraph and release references.
+        See Note [explicit cudagraph teardown] for more details.
+        """
+        self._cudagraph = None
+        self._args = None
+        self._output = None
 
 
 def get_static_input_indices(gm: torch.fx.GraphModule, is_forward: bool) -> list[int]:
--- a/torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
+++ b/torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import torch.nn as nn
 from torch.distributed.device_mesh import DeviceMesh
 from torch.fx.traceback import annotate_fn
 
@@ -20,8 +19,14 @@ from torchtitan.distributed import ParallelDims
 
 from torchtitan.distributed.activation_checkpoint import apply_ac
 from torchtitan.distributed.tensor_parallel import maybe_enable_async_tp
-from torchtitan.experiments.graph_trainer.common_utils import maybe_disable_eager_ac
+from torchtitan.experiments.graph_trainer.common_utils import (
+    annotate_ac_regions,
+    maybe_disable_eager_ac,
+)
 from torchtitan.experiments.graph_trainer.compile import apply_compile
+from torchtitan.experiments.graph_trainer.deepseek_v3.model import (
+    GraphTrainerDeepSeekV3Model,
+)
 from torchtitan.experiments.graph_trainer.simple_fsdp import (
     data_parallel,
     MixedPrecisionPolicy,
@@ -31,7 +36,7 @@ from torchtitan.protocols.model_converter import ModelConvertersContainer
 from torchtitan.tools.logging import logger
 
 
-def annotate_deepseekv3() -> None:
+def annotate_deepseekv3(model: GraphTrainerDeepSeekV3Model) -> None:
     """Attach annotations to FX graph nodes with ``torch.fx.traceback.annotate_fn``
 
     - Expert Parallel (EP) annotations: Tags "dispatch", "combine", and "compute"
@@ -40,6 +45,9 @@ def annotate_deepseekv3() -> None:
       {"compile_with_inductor": "flex_attention"} so the compiler can apply
       regional inductor pass based on the annotation. Regional inductor is now only
       supported in AOT mode.
+    - AC region annotation: Tags each transformer block's forward with a unique
+      ac_region_id so that apply_sac_pass can assign per-block ac_graph_id
+      boundaries for the min-cut partitioner.
 
     """
     from torchtitan.distributed.expert_parallel import ExpertParallel
@@ -58,10 +66,12 @@ def annotate_deepseekv3() -> None:
         {"compile_with_inductor": "flex_attention"}
     )(FlexAttentionWrapper.forward)
 
+    annotate_ac_regions(model)
+
 
 # Adapted from llama4/infra/parallelize.py
 def parallelize_deepseekv3(
-    model: nn.Module,
+    model: GraphTrainerDeepSeekV3Model,
     *,
     parallel_dims: ParallelDims,
     training: TrainingConfig,
@@ -87,7 +97,7 @@ def parallelize_deepseekv3(
     ):
         raise NotImplementedError("CP support is only supported for SDPA.")
 
-    annotate_deepseekv3()
+    annotate_deepseekv3(model)
 
     maybe_disable_eager_ac(compile_config, ac_config)
 
--- a/torchtitan/experiments/graph_trainer/llama3/parallelize.py
+++ b/torchtitan/experiments/graph_trainer/llama3/parallelize.py
@@ -18,17 +18,19 @@ from torchtitan.config import (
 from torchtitan.distributed import ParallelDims
 from torchtitan.distributed.activation_checkpoint import apply_ac
 from torchtitan.distributed.tensor_parallel import maybe_enable_async_tp
-from torchtitan.experiments.graph_trainer.common_utils import maybe_disable_eager_ac
+from torchtitan.experiments.graph_trainer.common_utils import (
+    annotate_ac_regions,
+    maybe_disable_eager_ac,
+)
 from torchtitan.experiments.graph_trainer.compile import apply_compile
+from torchtitan.experiments.graph_trainer.llama3.model import GraphTrainerLlama3Model
 from torchtitan.experiments.graph_trainer.simple_fsdp import (
     data_parallel,
     MixedPrecisionPolicy,
 )
-from torchtitan.models.llama3.model import Llama3Model
 from torchtitan.models.llama3.parallelize import apply_tp
 from torchtitan.protocols.model_converter import ModelConvertersContainer
 from torchtitan.tools.logging import logger
-
 
 # for selective op activation checkpointing
 _op_sac_save_list = {
@@ -50,7 +52,7 @@ _op_sac_save_list = {
 }
 
 
-def annotate_llama() -> None:
+def annotate_llama(model: GraphTrainerLlama3Model) -> None:
     """Attach annotations to FX graph nodes with ``torch.fx.traceback.annotate_fn``
 
     - Flex attention annotation: Tags FlexAttentionWrapper.forward with
@@ -58,6 +60,9 @@ def annotate_llama() -> None:
       regional inductor pass based on the annotation. Regional inductor is now only
       supported in AOT mode.
 
+    - AC region annotation: Tags each transformer block's forward with a unique
+      ac_region_id so that apply_sac_pass can assign per-block ac_graph_id
+      boundaries for the min-cut partitioner.
     """
     from torchtitan.models.common.attention import FlexAttentionWrapper
 
@@ -65,9 +70,11 @@ def annotate_llama() -> None:
         {"compile_with_inductor": "flex_attention"}
     )(FlexAttentionWrapper.forward)
 
+    annotate_ac_regions(model)
+
 
 def parallelize_llama(
-    model: Llama3Model,
+    model: GraphTrainerLlama3Model,
     *,
     parallel_dims: ParallelDims,
     training: TrainingConfig,
@@ -94,7 +101,7 @@ def parallelize_llama(
         ({parallel_dims.tp}) and 2 * CP degree ({parallel_dims.cp}).
         """
 
-    annotate_llama()
+    annotate_llama(model)
 
     maybe_disable_eager_ac(compile_config, ac_config)
 
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -1,0 +1,363 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import itertools
+from collections.abc import Generator
+from contextlib import contextmanager
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+import torch.nn as nn
+import torch.utils._pytree as pytree
+from torch._functorch._aot_autograd.logging_utils import (
+    setup_stacktrace_preservation_hooks,
+)
+from torch._subclasses import FakeTensorMode
+from torch.fx.experimental.proxy_tensor import make_fx
+from torch.fx.traceback import preserve_node_meta
+from torch.nn.utils import stateless
+from torch.utils._python_dispatch import is_traceable_wrapper_subclass
+
+
+@dataclass
+class SubclassMeta:
+    cls: type
+    attrs: list[str]
+    ctx: Any
+    inner_metas: dict[str, tuple[int, Any]]
+    outer_size: torch.Size
+    outer_stride: tuple[int, ...]
+
+
+@dataclass
+class SubclassLayout:
+    num_tensors: int
+    meta: SubclassMeta | None
+
+
+@dataclass
+class TracedResult:
+    """Holds the traced graph and metadata needed to run it."""
+
+    gm: torch.fx.GraphModule
+    params_len: int
+    params_spec: pytree.TreeSpec
+    input_subclass_layouts: list[SubclassLayout]
+    output_subclass_layouts: list[SubclassLayout]
+
+
+def _unwrap_subclass(t: torch.Tensor) -> tuple[list[torch.Tensor], SubclassMeta | None]:
+    if not is_traceable_wrapper_subclass(t):
+        return [t], None
+    attrs, ctx = t.__tensor_flatten__()
+    all_inner = []
+    inner_metas = {}
+    for attr in attrs:
+        inner_t = getattr(t, attr)
+        tensors, meta = _unwrap_subclass(inner_t)
+        all_inner.extend(tensors)
+        inner_metas[attr] = (len(tensors), meta)
+    meta = SubclassMeta(
+        cls=type(t),
+        attrs=attrs,
+        ctx=ctx,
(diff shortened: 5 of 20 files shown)


torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -37,19 +36,16 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
                 warmup_steps=2,
                 decay_type="linear",
             ),
-            training=TrainingConfig(),
+            training=TrainingConfig(dtype="bfloat16"),
```
daniellepintz: dtype of trainer model must match dtype of generator model

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -37,19 +36,16 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
-            compile=GeneratorCompileConfig(
-                backend="eager",
-                cudagraph_mode="piecewise",
-            ),
+            enforce_eager=True,
             parallelism=ParallelismConfig(
-                tensor_parallel_degree=2,
+                tensor_parallel_degree=4,
```
daniellepintz, line 91: testing resharding works

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -289,33 +231,51 @@ async def generate(self, prompt_texts: list[str]) -> list[Episode]:
...
+            if isinstance(param, DTensor):
+                t = param.full_tensor()
+            else:
+                t = param
+            checksums[name] = t.to(torch.float64).sum().item()
+        return checksums
+
+    async def _fetch_rdma_handles(self) -> dict[str, list[RDMAWeightHandle]]:
+        """Fetch RDMA handles from all trainer ranks via TorchStore."""
+        num_ranks = await ts.get("policy_rdma/num_ranks")
```
allenwang28: do we want this as a defined constant?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -289,33 +231,51 @@ async def generate(self, prompt_texts: list[str]) -> list[Episode]:
...
+        all_handles: dict[str, list[RDMAWeightHandle]] = {}
+        for r in range(num_ranks):
+            rank_handles = await ts.get(f"policy_rdma/rank_{r}")
+            for name, handle in rank_handles.items():
+                all_handles.setdefault(name, []).append(handle)
```
allenwang28: what is this `setdefault()` for?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -218,19 +221,40 @@ def _build_model(
...
+        await ts.put(f"policy_rdma/rank_{rank}", handles)
+        if rank == 0:
+            await ts.put("policy_rdma/num_ranks", dist.get_world_size())
 
-        return {
-            k: v.full_tensor() if isinstance(v, DTensor) else v
-            for k, v in titan_state.items()
-        }
+    @endpoint
+    async def refresh_weights(self) -> None:
```
allenwang28: why does this need to be a separate endpoint? and actually, would be really helpful to better understand what this operation does. Reading between the lines it's like a "_maybe_copy_to_staging" but it isn't clear why this is needed at all

torchtitan/experiments/rl/unified/actors/generator.py
daniellepintz: this code is creating a dict mapping each param name to a list of RDMA handles, corresponding to the individual trainer shards for that param

the `setdefault()` creates an empty list for each param name if the param name is not in the dict yet

daniellepintz: Let me change this to use defaultdict instead..

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: > why does this need to be a separate endpoint?

Good point, this can live in `trainer.step`, doesn't need to be a separate endpoint

> would be really helpful to better understand what this operation does

For contiguous params, the RDMA buffers point to the params directly, so the generator just has to read from those tensors directly. But for non contiguous params, like column sharded params, we can't have an RDMA buffer point to them because RDMA buffers require contiguous tensors. so instead we create a contiguous "staging buffer" for these params of the right size, and then we need to copy the updated params (updated after optimizer.step) to the staging buffer after each update, so that generator can read updated params. 

pls lmk if that makes sense or if you see a better way of doing this

code pointer to `refresh` in torchstore:
https://github.com/meta-pytorch/torchstore/blob/87872e63fb3975827a4cb325419b001bd4b900a1/torchstore/direct_weight_sync.py#L149-L159

torchtitan/experiments/rl/unified/simple_grpo.py
```diff
@@ -318,9 +321,41 @@ async def train(self):
...
             # 3. Trainer computes advantages and updates policy
             metrics = self.trainer.step.call(scored_episodes).get().item(gpus=0)
-            # 4. Sync full weights to generator (each rank reshards locally)
-            weights = self.trainer.get_weights.call().get().item(gpus=0)
-            self.generator.update.call(metrics["policy_version"], weights).get()
+            # 4. Sync weights via direct RDMA
+            t0 = time.perf_counter()
+            self.trainer.refresh_weights.call().get()
+            t_refresh = time.perf_counter() - t0
+            self.generator.pull_weights.call(metrics["policy_version"]).get()
```
tianyu-l: Not sure how viable my suggestion is:

Right now the UX design is actor-centric, in the sense that trainer and generator each owns its own rdma_handles and is responsible for owning weight-sync related functions.

This has the issue that actors need to know the implementation details of weight sync, e.g. whether it's rdma-based or TCP-based.

What I would propose
- create a `WeightSyncManager` which owns the creation, destroy, and communications including `trainer.refresh_weights` and `generator.pull_weights`
- let orchestrator directly control this `weight_sync_manager`
- trainer and generator stay intact, except for exposing necessary fields like state_dict getter/setter

torchtitan/experiments/rl/unified/actors/generator.py
daniellepintz: yes, done!

daniellepintz:
> noob questions
> What's the difference between the two?

MonarchRDMATransportBuffer, the existing TorchStore implementation, is built with Async RL in mind. It writes the tensors to a separate Monarch actor called a `StorageVolume` every time we need to update weights (after every step). By "write" I mean the StorageVolume uses RDMA to read the tensors from the trainer, and then stores them in CPU memory.  Then when the generator is ready to pull the weights, it does an RDMA read from the StorageVolume into the generator tensors.

Essentially the StorageVolume is serving as a "CPU buffer" here, to hold the weights until the generator is ready to read them. The main benefit of this is timing; the generator can choose when to read the weights, allowing for async RL.

With GPUDirect RDMA (this PR) since we are doing sync RL, we can stop and have the generator read the weights directly from the trainer's RDMA-registered weights (meanwhile training and generation are both paused). Since there is no extra hop of going to the StorageVolume, it is significantly faster.

This is explained very well by @allenwang28 in Section 5 of https://allenwang28.github.io/monarch-gpu-mode/07_rdma_weight_sync.html

> How should user pick which option to use?

Generally for async RL we want to use the CPU buffer option, and for sync RL we want to use GPUDirect RDMA. As we transition to async RL later, we should definitely make sure we are continuously timing weight sync as we go, and ensuring things are optimal.

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: Thanks for the suggestion! a few concerns that I can think of:
1. the RDMA buffer registration and reading must happen in the process that owns the GPU memory - so we have to call `refresh_weights` on the trainer, and `pull_weights` on the generator, etc. - just to clarify are you suggesting to move this code out of the actors (trainer and generator)?
2. I am not sure if creating an abstraction is too early right now, especially since I don't think we need to support any other type of weight sync right now (like TCP-based)

also curious if @allenwang28 has thoughts here

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
```diff
@@ -318,9 +321,41 @@ async def train(self):
...
-            self.generator.update.call(metrics["policy_version"], weights).get()
+            # 4. Sync weights via direct RDMA
+            t0 = time.perf_counter()
+            self.trainer.refresh_weights.call().get()
+            t_refresh = time.perf_counter() - t0
+            self.generator.pull_weights.call(metrics["policy_version"]).get()
+            t_total = time.perf_counter() - t0
+            logger.info(f"Weight sync: refresh={t_refresh:.3f}s, total={t_total:.3f}s")
+
+            # Verify weight sync after first training step
```
tianyu-l: should it only happen after first step?

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
```diff
@@ -318,9 +321,41 @@ async def train(self):
...
+                        mismatches.append(
+                            f"  {name}: trainer={trainer_checksums[name]:.6f}, "
+                            f"gen={gen_checksums[name]:.6f}, "
+                            f"diff={abs(trainer_checksums[name] - gen_checksums[name]):.6e}"
+                        )
+                if mismatches:
+                    logger.warning(
+                        f"Post-step weight sync mismatch ({len(mismatches)} params):\n"
+                        + "\n".join(mismatches[:20])
+                    )
```
tianyu-l: should we error out instead of just warn -- what's the incentive to do sync training without weight synced?

tianyu-l: In general I wonder why we need this code. Is it testing if torchstore is set up correctly? That should go in test, not actual trainer.

Or is it by nature that sync loss happens often?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -218,19 +221,40 @@ def _build_model(
...
+        await ts.put(f"{RDMA_KEY_PREFIX}/rank_{rank}", handles)
+        if rank == 0:
+            await ts.put(f"{RDMA_KEY_PREFIX}/num_ranks", dist.get_world_size())
 
-        return {
-            k: v.full_tensor() if isinstance(v, DTensor) else v
-            for k, v in titan_state.items()
-        }
+    @endpoint
+    async def refresh_weights(self) -> None:
```
tianyu-l: name is a bit vague -- is it more accurate to call it, e.g. `refresh_model_state_dict_in_rdma_buffer` or simpler `rdma_refresh_model_states` (following `rdma_` prefix)

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -218,19 +221,40 @@ def _build_model(
         return model
 
     @endpoint
-    async def get_weights(self) -> dict:
-        """Get model weights for generator.
+    async def register_rdma_handles(self) -> None:
```
tianyu-l: similarly, since this function is put under trainer, without other abstraction layer, we should make the name more self-explanatory, e.g. `rdma_register_model_states_handle`. The `rdma_` prefix is to group weight-sync related method by name.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -289,33 +232,51 @@ async def generate(self, prompt_texts: list[str]) -> list[Episode]:
...
+            for name, handle in rank_handles.items():
+                all_handles[name].append(handle)
+        logger.debug(
+            f"Fetched RDMA handles from {num_ranks} trainer ranks "
+            f"({len(all_handles)} params)"
+        )
+        return all_handles
+
+    @endpoint
+    async def pull_weights(self, version: int) -> None:
```
tianyu-l: `rdma_pull_model_states`

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -289,33 +232,51 @@ async def generate(self, prompt_texts: list[str]) -> list[Episode]:
...
+        checksums = {}
+        for name, param in self._get_model().model.state_dict().items():
+            if isinstance(param, DTensor):
+                t = param.full_tensor()
+            else:
+                t = param
+            checksums[name] = t.to(torch.float64).sum().item()
+        return checksums
+
+    async def _fetch_rdma_handles(self) -> dict[str, list[RDMAWeightHandle]]:
```
tianyu-l: `_rdma_fetch_model_states_handle`

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -289,33 +232,51 @@ async def generate(self, prompt_texts: list[str]) -> list[Episode]:
         return episodes
 
     @endpoint
-    async def update(self, version: int, state_dict: dict) -> None:
-        """Update generator weights.
-        Called by the orchestrator (simple_grpo.py).
+    async def get_weight_checksums(self) -> dict[str, float]:
```
tianyu-l: let's avoid using "weight" in formal definition, which is ambiguous in nature (e.g. is `bias` a `weight`?)
```suggestion
    async def get_model_states_checksums(self) -> dict[str, float]:
```
I'm also OK with `model_state_dict` spelled out

torchtitan/experiments/rl/unified/config_registry.py
tianyu-l: No, we can't change this dtype. Training has to be able to have master weights in higher precision.

I think it's OK if we downcast before sending to torchstore (rdma refresh weights in a lower precision copy) in async mode, or before GPU direct communication in sync mode.

When you say "must", is it required only in checksum comparison?

torchtitan/experiments/rl/unified/simple_grpo.py
tianyu-l: Makes sense. I think it's fine to not have "abstraction", but the code organization can be improved without making further abstraction. E.g. maybe we can name all weight sync related fields with `_rdma_` or `rdma_` (you are already doing this for variables)

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -6,12 +6,14 @@
...
+from collections import defaultdict
 from dataclasses import dataclass, field
-from typing import Literal
 
 import torch
+import torchstore as ts
 from monarch.actor import Actor, endpoint
-from torch.distributed.tensor import distribute_tensor, DTensor
+from torch.distributed.tensor import DTensor
+from torchstore.direct_weight_sync import DirectWeightSyncDest, RDMA_KEY_PREFIX, RDMAWeightHandle
```
wwwjn: A n00b q: If we want to test out RDMA (or run the current RL code ), can we directly run using our devGPU? Are they RDMA enabled?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -214,6 +155,8 @@ def __init__(
         logger.info("vLLM rollout engine initialized")
 
         self.policy_version = 0
+        self._rdma_dest = DirectWeightSyncDest()
```
wwwjn: And not every user has RDMA support to run this? Beside lightning speed provided by RDMA, do we have other options now for weight transfer

torchtitan/experiments/rl/unified/simple_grpo.py
wwwjn: About "abstraction": my 2c is I might tend to over-abstract things, but this should not preventing us of creating class to make the code cleaner. My understanding is:  
- Abstraction is for generalizability , when creating an abstraction, we would need to make sure the interface works for all the cases when creating an abstraction
- Create a class / a function is fine, it creates separation between modules and increase readabiliyt.

What about we have a separate `TrainerRDMAWeightSyncManager` in `Trainer` and `GeneratorRDMAWeightSyncManager` for `Generator`, and the function names can be specific tied to RDMA

My worry is Trainer is overloaded with a lot of functions, and user would need to read one by one to figure out which part is related to training/ weight update.

cc @felipemello1

torchtitan/experiments/rl/unified/actors/generator.py
wwwjn: Does it fallback to any TCP/IP based communication ?

torchtitan/experiments/rl/unified/simple_grpo.py
felipemello1: Yeah, as i was reading it, i had the same thoughts and i think that all the RDMA logic should have its own class.

Its common to have a CheckpointManager, right? Tune has it. [Titan](https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/checkpoint.py#L125) has it.

I think its the same thing here. 

1. If later we have to branch and support `if 'RMDA' or if 'RPC`, the trainer and generator endpoints will stay the same
2. A separate class is much easier to be tested. We can create a dummy state dict and test it.
3. Its less cognitive load for users that only care about 'generator'/'trainer' and not internals of weight sync.

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
daniellepintz: I can remove this - this was just testing that weight sync is working correctly. Yes, this is covered by tests in TorchStore

torchtitan/experiments/rl/unified/simple_grpo.py
daniellepintz: I cleaned up the API in https://github.com/meta-pytorch/torchstore/pull/138, so now we use the TorchStore `put_state_dict` / `get_state_dict` APIs :)

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: I cleaned up the API so this name is gone, now we use TorchStore `put_state_dict` / `get_state_dict`

torchtitan/experiments/rl/unified/actors/generator.py
daniellepintz: It should work on most devGPUs!

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: API has been updated so now we use TorchStore `put_state_dict` / `get_state_dict`

torchtitan/experiments/rl/unified/actors/generator.py
daniellepintz: Yes good call, I am adding a fallback to use TorchStore StorageVolume, which will use Monarch RPC if RDMA is not available!

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -216,6 +216,14 @@ def __init__(
 
         self.policy_version = 0
 
+        try:
+            from monarch.rdma import RDMABuffer  # noqa: F401
```
allenwang28: do you want to import `is_rdma_available` and use that check here?

also iirc, I thought store had a similar check to this lol

torchtitan/experiments/rl/unified/simple_grpo.py
allenwang28: I think this is resolved, but re-opening this because I think the conversation is interesting!

re: @tianyu-l  comment:
> - create a WeightSyncManager which owns the creation, destroy, and communications including trainer.refresh_weights and generator.pull_weights
> - let orchestrator directly control this weight_sync_manager
> - trainer and generator stay intact, except for exposing necessary fields like state_dict getter/setter

I want to make sure I understand correctly, I think there's two possible interpretations:
(a) The transport details are a leaky abstraction. Actors should not need to detect RDMA availability or be concerned with those details. The WeightSyncManager is a proposal that is a symptom of torchstore not doing its job correctly in this case
(b) Actors should not own push/pull at all. A separate manager should control the full weight sync flow, and actors only expose state_dict getters/setters.

If (a), I think that's right and the fix belongs in TorchStore. I thought it was already doing that though, not sure about the details here @daniellepintz @amirafzali cc @LucasLLC 

If it's (b), IMO I would want to be more careful here. This split-sided pattern where each actor independently owns its push/pull preserves flexibility for async RL - the trainer would publish on its schedule and the generator pulls on its schedule - that way the orchestrator can evolve from sync to async without changing the actor APIs. If we had like a `manager.sync_weights()`, then this encodes a global sync point into the API that we'd need to break apart later, which basically ends up looking like this split side sync that we have now.

For context, vLLM is landing native weight sync APIs (https://github.com/vllm-project/vllm/issues/31848) that also converge on a split-side pattern, the inference engine owns a pluggable `WeightTransferEngine` (consumer-side only) and the trainer pushes via static utilities, where the transport backend is handled internally. TorchStore would play a similar role here. (also not saying we need to use the native weight sync APIs!)

torchtitan/experiments/rl/unified/config_registry.py
daniellepintz: Ok, reverted this change! Now we do downcasting in https://github.com/meta-pytorch/torchstore/pull/140

> When you say "must", is it required only in checksum comparison?

No, it is not about the checksum comparison, when the generator is reading from the trainer's tensors, it expects the sizes in bytes to be the same. downcasting fixes this though

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +231,25 @@ def _build_model(
...
-            for k, v in titan_state.items()
-        }
+        if self._use_direct_rdma:
+            state_dict = self.model.state_dict() if not self._weights_pushed else None
+        else:
+            state_dict = self.model.state_dict()
+        await ts.put_state_dict(
+            state_dict,
+            "policy_weights",
+            direct_rdma=self._use_direct_rdma,
```
amirafzali: I think you would want to base the usage of direct_rdma on some type of titan level RL config rather than just the availability of rdma on the host? I.e, if you're doing async RL (not direct) you would still require rdma availability. Ideally there is no `is_rdma_available` leakage on this side

daniellepintz: Hmm not sure I fully follow - currently for just sync RL, if rdma is available on the host we want to use it in all cases I believe - if not we have to fallback

allenwang28: I think Amir is saying like, `is_rdma_available` is being used to determine whether we should use `_use_direct_rdma` but if we were in the async case, we still need `is_rdma_available` to be true.

amirafzali: yes, also the fact that this pattern requires the client to be hardware-aware for transport strategy selection, which IMO should be TS territory

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +231,25 @@ def _build_model(
         return model
 
     @endpoint
-    async def get_weights(self) -> dict:
-        """Get model weights for generator.
+    async def push_weights(self) -> None:
```
tianyu-l:
```suggestion
    async def push_model_state_dict(self) -> None:
```

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +231,25 @@ def _build_model(
...
-            k: v.full_tensor() if isinstance(v, DTensor) else v
-            for k, v in titan_state.items()
-        }
+        if self._use_direct_rdma:
+            state_dict = self.model.state_dict() if not self._weights_pushed else None
+        else:
+            state_dict = self.model.state_dict()
+        await ts.put_state_dict(
+            state_dict,
+            "policy_weights",
```
tianyu-l:
```suggestion
            "model_state_dict",
```
or `policy_model_state_dict`

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +231,25 @@ def _build_model(
...
+    async def push_weights(self) -> None:
+        """Publish model weights for generator consumption via direct RDMA.
 
-        Returns:
-            model state dict with plain local tensors (DTensors unwrapped
-            to avoid cross-mesh issues when transferring through Monarch).
+        On the first call, registers RDMA handles for all model parameters
+        and stores them in TorchStore. On subsequent calls, refreshes
+        staging buffers for non-contiguous params (contiguous params are
+        updated in-place by the optimizer and need no refresh).
```
tianyu-l: Could you say more about the contiguous vs. non-contiguous? What does "refreshing staging buffers" mean and why torchstore requires a refresh?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +231,25 @@ def _build_model(
...
+            state_dict = self.model.state_dict() if not self._weights_pushed else None
+        else:
+            state_dict = self.model.state_dict()
+        await ts.put_state_dict(
+            state_dict,
+            "policy_weights",
+            direct_rdma=self._use_direct_rdma,
+            transfer_dtype=torch.bfloat16,
+        )
+        self._weights_pushed = True
```
tianyu-l: The naming is confusing. Do you mean "weights pushed for the first time"? Once it's set to true, it's never set to false again.

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -305,33 +311,26 @@ async def generate(
         return episodes
 
     @endpoint
-    async def update(self, version: int, state_dict: dict) -> None:
-        """Update generator weights.
-        Called by the orchestrator.
+    async def pull_weights(self, version: int) -> None:
```
```suggestion
    async def pull_model_state_dict(self, version: int) -> None:
```

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +231,25 @@ def _build_model(
...
-        return {
-            k: v.full_tensor() if isinstance(v, DTensor) else v
-            for k, v in titan_state.items()
-        }
+        if self._use_direct_rdma:
+            state_dict = self.model.state_dict() if not self._weights_pushed else None
+        else:
+            state_dict = self.model.state_dict()
+        await ts.put_state_dict(
+            state_dict,
```
tianyu-l: what happens if it's None?

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
```diff
@@ -222,12 +223,13 @@ async def setup(self):
             self.task.reward_function,
         )
 
-        # Trainer gathers full (unsharded) weights on every rank. We only
-        # need to collect from one rank since they're all identical.
-        initial_weights = self.trainer.get_weights.call().get().item(gpus=0)
+        # Initialize TorchStore
+        await ts.initialize(num_storage_volumes=2, strategy=ts.LocalRankStrategy())
```
tianyu-l: please add comment on the API as it's new
- what is `num_storage_volumes` and why it is set to 2
- what strategy means and what is `LocalRankStrategy`

Adding links to tutorial / doc would be helpful

torchtitan/experiments/rl/unified/config_registry.py
```diff
@@ -49,7 +49,7 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
                 cudagraph_mode="piecewise",
             ),
             parallelism=ParallelismConfig(
-                tensor_parallel_degree=2,
+                tensor_parallel_degree=4,
```
tianyu-l, line 52: does it only work with TP?

daniellepintz:
> What scale / what test setup to get the perf number? By reading the config, is it all intra-node, 4GPU for trainer and 4GPUs for generator?

@wwwjn yes I tested all intra-node, two settings:
- 2 GPUs for trainer, 4 GPUs for generator
- 2 GPUs for trainer, 2 GPUs for generator

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: ah sorry yes forgot to address this one

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: when using direct RDMA, we only need to pass the state_dict the first time so we can register RDMA handles for all the tensors. all times after that we will just re-copy the existing tensors of the state_dict into staging buffers ("refresh"). so we save ourselves having to call `model.state_dict()` each time when it's not needed.

alternatively we could simplify the API and just pass it each time, and ignore it all times after the first

tianyu-l: I would prefer not letting users (torchtitan) handle this first vs. subsequent calls. Ideally the `ts` should be able to figure it out. Also we can simplify the code here a bit, including `self._weights_pushed = True` which could be hidden in `ts`.

> Or is it just GPUDirect RDMA used because we are in sync mode? I don't see a config controlling this. Let's explicit document the current choice / status, and add TODO for followups.

btw this earlier question seems not addressed.

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: sorry, these are TS implementation details, I removed the comment as this should live in TS. But yes let me explain: when tensors are sharded, if they are sharded row-wise the tensors remain contiguous, each shard is a contiguous block in memory. but if they are sharded col-wise, the tensors are no longer contiguous. registering a region with RDMA **requires** contiguous memory. so for non contiguous tensors we create "staging buffers" in contiguous memory and copy the tensors over to the staging buffers each time they are updated by the optimizer

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: yes exactly that's what I mean. I can change this to "_rdma_registered"?

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: Ok, makes sense I will simplify the torchtitan side and always pass model_state_dict. this will add some small latency

yes sorry forgot to respond to that, will do in separate comment

daniellepintz:
> In this PR, are both supported? Or is it just GPUDirect RDMA used because we are in sync mode? I don't see a config controlling this. Let's explicit document the current choice / status, and add TODO for followups.

@tianyu-l With `ts.put_state_dict / ts.get_state_dict` (this PR) both are supported yeah. Currently the way it's controlled is if rdma is available we use GPU direct RDMA (by passing `direct_rdma=self._use_direct_rdma`). If not we fall back to using whatever transport TorchStore chooses based on what is available.

Do you think we should also have a config controlling this? 

Will add documentation on this - there aren't really any TODO follow ups in Titan atm - there are some in TorchStore

torchtitan/experiments/rl/unified/config_registry.py
acisseJZhong: +1, curious does DP resharding also work?

torchtitan/experiments/rl/unified/actors/generator.py
```diff
@@ -305,33 +314,22 @@ async def generate(
...
+            user_state_dict=model_sd,
+            strict=False,
+            direct_rdma=self._use_direct_rdma,
+        )
         self.policy_version = version
         logger.debug(
-            f"Updated weights into vLLM engine model. "
-            f"Number of parameters: {len(load_weights)}. "
-            f"{os.getpid()=} Generator updating weights to policy v{version}..."
+            f"{os.getpid()=} Generator pulled weights via RDMA for policy v{version}"
```
acisseJZhong: nit: might also fall back to RPC based weight sync? can we update the comment to reflect that?

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
```diff
@@ -222,12 +223,20 @@ async def setup(self):
...
-        # need to collect from one rank since they're all identical.
-        initial_weights = self.trainer.get_weights.call().get().item(gpus=0)
-
-        # Initialize generator with trainer weights
-        self.generator.update.call(0, initial_weights).get()
+        # Initialize TorchStore for weight sync between trainer and generator.
+        # num_storage_volumes: intermediate storage nodes that hold tensor data
+        #   in the non-RDMA (RPC) path. 2 volumes distributes the load so no
+        #   single volume is a bottleneck.
+        # LocalRankStrategy: routes each process to a storage volume based on
```
tianyu-l, line 230: From https://github.com/meta-pytorch/torchstore/blob/main/torchstore/strategy.py#L164:
> This strategy requires the LOCAL_RANK environment variable to be set and assumes
    one storage volume per local rank.

We only created 2 volumes, but there could be e.g. 8 local ranks on a node. What happens?

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
```diff
@@ -222,12 +223,20 @@ async def setup(self):
...
 
-        # Trainer gathers full (unsharded) weights on every rank. We only
-        # need to collect from one rank since they're all identical.
-        initial_weights = self.trainer.get_weights.call().get().item(gpus=0)
-
-        # Initialize generator with trainer weights
-        self.generator.update.call(0, initial_weights).get()
+        # Initialize TorchStore for weight sync between trainer and generator.
+        # num_storage_volumes: intermediate storage nodes that hold tensor data
+        #   in the non-RDMA (RPC) path. 2 volumes distributes the load so no
```
tianyu-l: If we are using GPU direct, why would we create 2 volumes?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +233,14 @@ def _build_model(
...
-        titan_state = self.model.state_dict()
-
-        return {
-            k: v.full_tensor() if isinstance(v, DTensor) else v
-            for k, v in titan_state.items()
-        }
+    async def push_model_state_dict(self) -> None:
+        """Publish model weights for generator consumption via TorchStore."""
+        await ts.put_state_dict(
+            self.model.state_dict(),
```
tianyu-l, line 250: also curious what happens if we send a different state dict on step 2

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -132,6 +132,15 @@ def __init__(
         self.policy_version = 0
         self.generator: Any | None = None
 
+        # Weight sync transport: if RDMA is available, use GPUDirect RDMA for
+        # one-hop GPU-to-GPU transfer. Otherwise, fall back to TorchStore's
+        # default RPC-based transport (two-hop via StorageVolumes).
+        from monarch.rdma import is_rdma_available
+
+        self._use_direct_rdma = is_rdma_available()
```
tianyu-l: Does `is_rdma_available()` comes with a cost? If not we don't need to store this field.

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -132,6 +132,15 @@ def __init__(
         self.policy_version = 0
         self.generator: Any | None = None
 
+        # Weight sync transport: if RDMA is available, use GPUDirect RDMA for
+        # one-hop GPU-to-GPU transfer. Otherwise, fall back to TorchStore's
+        # default RPC-based transport (two-hop via StorageVolumes).
```
tianyu-l: I think in the future sync or async should be user configurable (in `RLTrainer.Config`). Right now it's OK that we only support sync mode which optionally use GPU direct depending on availability.

In multi-node, does GPU direct still work?

> With ts.put_state_dict / ts.get_state_dict (this PR) both are supported yeah. Currently the way it's controlled is if rdma is available we use GPU direct RDMA (by passing direct_rdma=self._use_direct_rdma). If not we fall back to using whatever transport TorchStore chooses based on what is available.

I'm confused in that: are you saying "both" GPU direct and MonarchRDMATransportBuffer are supported?
- what happens if no RDMA is available at all. What does RPC-based transport in torchstore ultimately use? I had expected the answer to be TCP or whatever non-RDMA channel, but I'm not sure if it makes sense.

torchtitan/experiments/rl/unified/actors/trainer.py
tianyu-l: with the recent change that we always feed `model_state_dict`, do we still need this field?

torchtitan/experiments/rl/unified/actors/trainer.py
tianyu-l: are the buffers in GPU or CPU? If in GPU that's considerable amount of overhead; if in CPU then we "lose" GPU direct right?

allenwang28: cc @amirafzali

IIUC from https://github.com/meta-pytorch/torchstore/pull/138, these are indeed GPU staging buffers colocated with the source and adds memory overhead

other options we could explore are 1) multiple RDMA reads, or 2) [scatter gather lists (SGL)](https://stackoverflow.com/questions/52334677/rdma-scatter-gather-in-verbs-api), where you can specify multiple segments (in this case we would write one per non-contiguous segment) into a single operation. 

1) avoids the staging buffer, but each RDMA operation has a fixed overhead, so many small reads sequentially will be slower than a big one

2) is more complex to implement - not sure if torchcomms has support for this, but IIUC the final form of [Monarch's RDMAAction](https://github.com/meta-pytorch/monarch/blob/c359c55cad68a04ab446b2e09080bad9f94fbb6f/python/monarch/_src/rdma/rdma.py#L434) intends to drop down into SGLs.

allenwang28: cc @samlurye and @zdevito as well on RDMAAction for weight sync

daniellepintz: They are on GPU - same device as the source tensors. Yes, you are right about the tradeoff. For GPU, when using `transfer_dtype=bfloat16` we have to allocate an extra half the model size on GPU. But when using the same dtype, we only have to allocate staging buffers for the col-wise sharded params

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: we don't need it, removed it in latest version :)

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
daniellepintz: you are right, my bad - updated this to pass in the `trainer_mesh` to `ts.initialize`. That way one storage volume will be spawned on each trainer rank

torchtitan/experiments/rl/unified/simple_grpo_sum_digits.py
daniellepintz: good question - so when using GPU direct, the Storage Volumes are not used. This is only for the fallback case...

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: Great question! For GPU direct, if we send a different state dict, it will basically ignore it, because for GPU direct case we assume we are using the same regions in memory the entire time, after the first registration. This works because the optimizer updates weights in-place, so the registered memory regions stay valid across training steps. 

This is an assumption we make when using the unified `put_state_dict` API, compared to how it was earlier with explicit handle registration and then `refresh` calls. We can definitely consider refining this in the future, open to ideas!

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: > I think in the future sync or async should be user configurable

Agreed! sg

> In multi-node, does GPU direct still work?

Haven't tested yet, will do as a next step after landing this; but it should.

> are you saying "both" GPU direct and MonarchRDMATransportBuffer are supported?

Yes technically you can use both with the `put/get_state_dict` APIs. But currently our logic in Titan is: if RDMA is available, use GPU direct. If not, let TorchStore decide.

The way TorchStore decides is like this based on what's available:

https://github.com/meta-pytorch/torchstore/blob/ce506b0dfc763902d6d7326bf8a9eeba546ce558/torchstore/transport/__init__.py#L39-L57

(First will try Shared Memory, then Monarch RDMA, etc. and finally will use Monarch RPC)

> What does RPC-based transport in torchstore ultimately use? 

The data is sent via a Monarch RPC call

torchtitan/experiments/rl/unified/actors/trainer.py
daniellepintz: no real cost, so will remove

torchtitan/experiments/rl/unified/config_registry.py
daniellepintz: yes, it works with DP as well - The generator passes its own state dict (with sharding metadata) to `get_state_dict`, and TorchStore computes the overlap between source and destination tensor slices to determine which regions to read from each trainer rank

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -224,19 +224,21 @@ def _build_model(
...
 
-        return {
-            k: v.full_tensor() if isinstance(v, DTensor) else v
-            for k, v in titan_state.items()
-        }
+        await ts.put_state_dict(
+            self.model.state_dict(),
+            "model_state_dict",
+            direct_rdma=is_rdma_available(),
+            transfer_dtype=torch.bfloat16,
```
tianyu-l: fix to make sure it aligns with generator dtype

tianyu-l (approved):

PR #2527, opened 2026-03-09
[Module] Add configurable Linear that inherits from Module

fegin:
* __->__ #2527

**Why the Change**
Same as https://github.com/pytorch/torchtitan/pull/2428 and https://github.com/pytorch/torchtitan/pull/2434. This is the third PR that changes Linear to inherit from Module.

But unlike previous two PRs, we also need to consider converter cases that swap the linear modules with fp8/mx linear modules.

**Summary**

- Introduces `torchtitan.models.common.linear.Linear`, a configurable nn.Linear subclass that also inherits from Module, following the same Config/build() pattern as RMSNorm and Embedding.
- Updates all shared components (GQAttention, FeedForward, Decoder) and all model configs (Llama3, Llama4, Qwen3, DeepSeek V3, GPT-OSS, Flux) to use Linear.Config instead of raw nn.Linear.
- Adds capture_module_configs() and inject_module_protocol() utilities in protocols/module.py to preserve the Module protocol through Float8/MX quantization conversions. Float8 creates new instances (losing attributes), so configs are captured beforehand and re-attached afterwards. MX uses class swaps (attributes survive), so re-attachment acts as a safety net.
- Replaces the bias: bool field on GQAttention.Config with per-projection Linear.Config fields (wq, wk, wv, wo), enabling per-linear configuration of bias, init_std, and init_mean.
- Adds comprehensive unit tests for Linear, config injection, and the capture/re-attach round-trip.


torchtitan/models/common/attention.py
```diff
@@ -546,12 +558,8 @@ def forward(
 
     def init_weights(self, init_std: float = 0.02, **kwargs) -> None:
         for linear in (self.wq, self.wk, self.wv):
```
mori360, line 565: So after this pr, we will call `init_weights` to all weights in linear `init_weights`

torchtitan/models/common/linear.py
```diff
@@ -0,0 +1,59 @@
...
+            dtype=config.dtype,
+        )
+        self._init_mean = config.init_mean
+        self._init_std = config.init_std
+
+    def init_weights(self, **kwargs) -> None:
+        init_std: float | None = kwargs.pop("init_std", None)
+        std = init_std if init_std is not None else self._init_std
+        nn.init.trunc_normal_(self.weight, mean=self._init_mean, std=std)
+        if self.bias is not None:
```
mori360, line 55: So init_weights here does not consider dtype?

torchtitan/models/common/linear.py
```diff
@@ -0,0 +1,59 @@
...
+            config.in_features,
+            config.out_features,
+            bias=config.bias,
+            dtype=config.dtype,
+        )
+        self._init_mean = config.init_mean
+        self._init_std = config.init_std
+
+    def init_weights(self, **kwargs) -> None:
+        init_std: float | None = kwargs.pop("init_std", None)
```
acisseJZhong, line 52: can we use the init_std from Config? or can we do the same override for init_mean as well?

torchtitan/models/common/linear.py
acisseJZhong: I think it's from 

Image

torchtitan/models/common/moe/moe.py
```diff
@@ -166,7 +167,9 @@ def __init__(
         _debug_force_load_balance: bool = False,
     ):
         super().__init__()
-        self.gate = nn.Linear(dim, num_experts, bias=gate_bias)
+        self.gate = Linear.Config(bias=gate_bias).build(
```
acisseJZhong, line 170: can we also pass in bias here? It's default false, but this will be wrong if it's not None?

torchtitan/models/common/moe/moe.py
```diff
@@ -420,9 +421,12 @@ def __init__(self, config: Config, *, dim: int):
         )
         self.reorderer = TokenReorderer(num_experts=num_experts, top_k=config.top_k)
         self.shared_experts = (
-            FeedForward.Config(hidden_dim=hidden_dim * config.num_shared_experts).build(
-                dim=dim
-            )
+            FeedForward.Config(
+                hidden_dim=hidden_dim * config.num_shared_experts,
+                w1=Linear.Config(),
```
acisseJZhong: can this be default?

torchtitan/models/common/feed_forward.py
```diff
@@ -42,22 +42,27 @@ class FeedForward(Module):
...
+        w3: Linear.Config
+        dim: int = field(init=False)
 
-    def __init__(self, config: Config, *, dim: int):
+    def __init__(self, config: Config):
         super().__init__()
-        self.w1 = nn.Linear(dim, config.hidden_dim, bias=config.bias)
-        self.w2 = nn.Linear(config.hidden_dim, dim, bias=config.bias)
-        self.w3 = nn.Linear(dim, config.hidden_dim, bias=config.bias)
+        self.w1 = config.w1.build(
```
acisseJZhong: same: pass in config.bias?

tianyu-l:
@claude review this PR

torchtitan/models/common/moe/moe.py
fegin: `Linear.Config(bias=gate_bias)` 
We pass the bias?

torchtitan/models/common/moe/moe.py
fegin: I think this has been discussed in the previous PRs. There are pros and cons of different implementations:
1. Making a default in linear.py. I think we aim to not have a default config for Module Config and make Module config expclit.
2. Making a default in the user code (e.g., moe.py, llama3/__init__.py). So users create a default default_config = Linear.Config() and use this throughout the code. One concern is that unless the config is frozen, one change will contaminate all the components.
3. One entry per config. It's annoying but very explicit. 

cc., @tianyu-l

torchtitan/models/common/feed_forward.py
fegin: Same reply. `config.bias` has been set when config is created. Users should only pass arguments that need to be computed during runtime or shared across modules to `build()`.

torchtitan/models/common/linear.py
fegin: Good question, you may want to see the next PR, https://github.com/pytorch/torchtitan/pull/2540. You can just check the summary.

torchtitan/models/common/linear.py
fegin: If you are talking about `dtype`, that is used when creating the linear layer for GPT-OSS. After that, the dtype is already fix. And all the operations here are in_place which do not need dtype explicitly.

torchtitan/models/gpt_oss/model.py
```diff
@@ -275,14 +263,15 @@ def get_nparams_and_flops(
...
-            config.vocab_size,
-            dtype=torch.get_default_dtype(),
-            bias=False,
+        # Theoretically, we can put the dtype assignment in the real config.
+        # However, this is risky in the sense that we may not set_device
+        # before importing configurations or someone may `set_default_dtype()`
+        # in the trainer.__init__().
+        config = dataclasses.replace(
+            config,
+            output=dataclasses.replace(config.output, dtype=torch.get_default_dtype()),
```
tianyu-l: hmm, I don't think we need this dtype setting. For most models we are doing post-output layer upcast in loss computation https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/loss.py#L24

In gpt-oss repo, they are using bf16 same as our default https://github.com/openai/gpt-oss/blob/main/gpt_oss/torch/model.py#L379

torchtitan/models/common/feed_forward.py
```diff
@@ -42,22 +42,27 @@ class FeedForward(Module):
     @dataclass(kw_only=True, slots=True)
     class Config(Module.Config):
         hidden_dim: int
-        bias: bool = False
+        w1: Linear.Config
+        w2: Linear.Config
+        w3: Linear.Config
```
tianyu-l: I would consider these as "SwiGLUFeedForward" implementation details that doesn't need to be swapped / configurable, until we receive requests to configure them. It's kind verbose to have them in every model config registry without specifying any real arg.

You may argue that people still want to experiment with different init -- that's the reason why we have next PR.

Same for wq/wk/wv/wo.

torchtitan/models/common/moe/moe.py
tianyu-l, line 170: It seems some linears are put into config, whereas some else are creating config in constructor.

My perspective is that
- If something is relatively stable, e.g. doesn't change across qwen / gpt-oss / llama4, it's probably fine to create config ad hoc like here. This would include most feedforward w1/2/3 and attention wq/k/v/o.
- We could still expose handles that indeed would change, like with bias or not.

torchtitan/components/quantization/mx.py
```diff
@@ -107,11 +114,24 @@ def convert(self, model: nn.Module):
         from torchao.quantization import quantize_
 
         assert isinstance(self.torchao_config, TorchAOMXLinearConfig)
+
+        # Capture Module attrs before conversion (MX may swap classes, losing them).
+        # We need to first verify if all nn.Linear have been converted to Linear.
+        verify_module_protocol(model, nn.Linear, Linear)
+        saved_attrs = capture_module_attrs(
+            model, ["_init_mean", "_init_std"], nn_module_cls=nn.Linear
```
tianyu-l: Are these the only thing we need to save? I thought
- these fields could be moved to config, not in the Linear class itself.
- but then we still need to keep the inheritance from Float8Linear to torchtitan.Linear.

After all, if we have to keep such code, they should exist in torchao, not torchtitan.

@danielvegamyhre what's your latest thought on torchao APIs? In this PR, we are having torchtitan-specialized nn.Linear and I think torchao should still work without we do anything special here.

torchtitan/models/common/moe/moe.py
acisseJZhong: got it, I think 3 works fine!

torchtitan/models/common/moe/moe.py
acisseJZhong: oops somehow I misread.

torchtitan/components/quantization/mx.py
fegin: > these fields could be moved to config, not in the Linear class itself.
These fields are moved to StateInitializer in the next PR, https://github.com/pytorch/torchtitan/pull/2540. The only thing we need to save is `_state_initializer`.

> After all, if we have to keep such code, they should exist in torchao, not torchtitan.

I'm happy to remove this logic if torchao handle this. But I don't want to be blocked by another refactor from another repo. I would consider to move forward with this logic living inside this repo and refactor until both repo are happy.

torchtitan/models/gpt_oss/model.py
fegin: We can remove it. I just don't want to change the original logic. It may be our logic is from a legacy code? The comment specifically mentions the usage of `torch.get_default_dtype()`. I added this logic when reviewing the PR.

torchtitan/components/quantization/mx.py
tianyu-l: That's fine. If so, let's
- move these utils inside quantization, not in general `protocols` folder.
- mark them as temporary and leave TODOs

torchtitan/models/common/feed_forward.py
tianyu-l: Chatted offline
- we think it's not necessary to use build style at all for these internal details of a module.
- the init method can be owned by root module, as a dict from fqn to init function mapping, which can be changed by users who modify model code
- the sharding may still need to be defined per each module (unclear right now), but probably should still be owned by parent module

Issue #2217, opened 2026-01-09
Understanding RMSNorm Gradient Synchronization in Tensor-Parallel LLaMA


alpemreacar:
I am testing tensor parallelism (TP) using a simple FSDP-based implementation of LLaMA-3. I am on commit 9f211ec199bc887901b874edd6af5a20527a4175, using Python 3.10 and a nightly version of PyTorch, as noted in the README.

My understanding is that under TP, model weights are sharded across ranks so that each rank performs a fraction of the total computation. I am able to capture the execution graphs using TorchInductor by running:

`TORCH_COMPILE_DEBUG=1 CUDA_VISIBLE_DEVICES=0,1 torchrun --nproc_per_node=2 -m torchtitan.train --job.config_file=tp.oml --model.name simple_fsdp.llama3 --job.custom_config_module=torchtitan.experiments.simple_fsdp.job_config`

I have attached my toml configuration file along with a reference implementation.

As expected, I observe both all_gather and reduce_scatter operations in the forward and backward passes, since each rank is computing only a partial result.

However, I am confused about how the RMSNorm parameters are updated. In the LLaMA implementation, each transformer block processes tensors of shape (B, SeqLen / TP, HiddenDim). As a result, each rank only observes gradients corresponding to B × SeqLen / TP tokens for the RMSNorm parameters. These gradients should therefore be reduced across ranks.

However, when inspecting the generated graphs, I do not see any explicit all_reduce corresponding to the RMSNorm parameter gradients.

Could someone clarify where and how this gradient reduction is performed during execution? Is it fused, implicit, or handled outside of the graph I am inspecting?

Any insight would be appreciated.

[tptoml.txt](https://github.com/user-attachments/files/24511609/tptoml.txt)

sanketpurandare:
# Understanding RMSNorm Gradient Reduction in Tensor Parallelism 

**TL;DR**: RMSNorm parameter gradients ARE reduced, but through DTensor's `Partial` placement semantics rather than an explicit `all_reduce` operation in the compiled graph.


## How It Works

### 1. Sequence Parallel Setup for RMSNorm

RMSNorm is parallelized using `SequenceParallel` in the layer plan:

```python
# torchtitan/models/llama3/infra/parallelize.py
layer_plan = {
    "attention_norm": SequenceParallel(),
    "ffn_norm": SequenceParallel(),
    # ...
}
```

[...]

PR #2532, opened 2026-03-09
[DTensor Bugfix] Explicitly specify grad_placements in to_local to ensure necessary all reduce takes place 

acisseJZhong:
Solved #2217 by @alpemreacar
 ## Problem: 
`x.to_local()` was called on a multi-dimensional DTensor (e.g., on a `(dp, tp)` mesh with placements like `(Shard(0), Replicate())`). The bare `to_local()` strips all mesh dimensions and loses gradient placement info for the non-DP (TP) dimensions. For a `Replicate()` RMSNorm weight on the TP mesh, the backward should produce `Partial()` gradients (requiring all-reduce), but this information was lost.

 ## Fix: 
  1. Extract the non-DP placements from `x`
  2. Compute the corresponding grad placements — Replicate → Partial(), others stay as-is
  3. Pass the full grad placements (DP dims + non-DP dims) to to_local()

This ensures that in the backward pass, gradients flowing back through `to_local()` are properly wrapped as a DTensor with `Partial()` on the TP mesh dimension, which will trigger the necessary all-reduce across TP ranks.

**Note that this changes the numerics when using simple FSDP so I updated the test loss value for the GraphTrainer tests.** 

Authored with Claude.

### DIFF
--- a/torchtitan/experiments/graph_trainer/simple_fsdp.py
+++ b/torchtitan/experiments/graph_trainer/simple_fsdp.py
@@ -187,8 +187,12 @@ class ReplicateComputation(torch.nn.Module):
             dp_mesh = self.device_mesh
 
             # Compute grad placements for non-DP mesh dims so that
-            # backward properly reduces gradients (e.g., Replicate -> Partial
-            # for RMSNorm weights on the TP mesh that see Shard inputs).
+            # backward properly reduces gradients.
+            # TODO: We assume Replicate -> Partial here, which is correct
+            # when using Sequence Parallel (Shard inputs produce partial
+            # gradients). Without SP, the grad placement depends on
+            # downstream operations and may not be Partial. This could be
+            # fixed by enforcing spmd_types.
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/graph_trainer/simple_fsdp.py
```diff
@@ -185,8 +185,24 @@ def replicate_compute(self, x: DTensor) -> torch.Tensor:
                     "full_dtensor not implemented for nD parallelisms"
                 )
             dp_mesh = self.device_mesh
+
+            # Compute grad placements for non-DP mesh dims so that
+            # backward properly reduces gradients.
+            # TODO: We assume Replicate -> Partial here, which is correct
+            # when using Sequence Parallel (Shard inputs produce partial
+            # gradients). Without SP, the grad placement depends on
```
acisseJZhong, line 193: @pianpwk this could potentially be wrong if you disable SP in https://github.com/pytorch/torchtitan/pull/2535, I will add an assertion here to make sure SP turned on after your PR.

Issue #1867, opened 2025-10-14
Qwen3 model parallel including unnecessary all_gather


yanboliang:
### Bug description

Root cause:
https://github.com/pytorch/torchtitan/blob/cd304c7b99111bd1364ad2f28fdeb259df5e2c96/torchtitan/models/qwen3/infra/parallelize.py#L248-L249

Should be changed to:
```
"attention.q_norm": SequenceParallel(sequence_dim=2),
"attention.k_norm": SequenceParallel(sequence_dim=2),
```

As we only want to do norm across the last dim, we don't need to `redistribute/all_
gather` input. However, the original `NoParallel` actually check `input_layout != desired_input_layout` and then trigger a redistribute. Because `NoParallel`'s default `desired_input_layout` is `Replicate()` and the input layout at here is `Sharding`. After changing to `SequenceParallel`, there is no all_gather before `q_norm` and `k_norm`.

### Versions

main branch

PR #2535, opened 2026-03-10
[TP] Enable/disable SP for TP

pianpwk:
* __->__ #2535

decouples SP from TP with `enable_sp=True` for all models, MoE and non-MoE


torchtitan/experiments/rl/unified/models/parallelize.py
```diff
@@ -78,85 +81,150 @@ def apply_non_moe_tp(
...
+            },
+        )
+    else:
+        parallelize_module(
+            model,
+            tp_mesh,
+            {
+                "tok_embeddings": RowwiseParallel(
+                    input_layouts=Replicate(),
+                    output_layouts=Replicate(),
```
tianyu-l: Do you think we can make things a bit succinct by writing
```
output_layouts=Shard(1) if sequence_parallel else Replicate(),
```
similar to when loss parallel is enabled/disabled? It also gives more readability IMO.

tianyu-l (commented):
Let's do all models, also after https://github.com/pytorch/pytorch/pull/172422 we should be able to use all DTensor (`use_local_output=False`) for everything.

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -96,13 +96,19 @@ def parallelize_deepseekv3(
                 "Currently, float8 tensorwise TP is not tested for deepseekv3"
             )
 
+        enable_sp = parallelism.enable_sequence_parallel
+        loss_parallel = not parallelism.disable_loss_parallel
+        if not enable_sp:
+            loss_parallel = False
```
tianyu-l: why coupling them together?

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -229,23 +235,31 @@ def apply_non_moe_tp(
     loss_parallel: bool,
     enable_float8_tensorwise_tp: bool,
     cp_enabled: bool,
+    enable_sp: bool = True,
```
tianyu-l: extra request: could you help unify the interface? The four bools are named very differently.

torchtitan/experiments/rl/unified/models/parallelize.py
```diff
@@ -103,56 +109,64 @@ def apply_non_moe_tp(
...
+
+        layer_plan.update(
+            {
+                "attention_norm": norm_plan,
+                "attention": PrepareModuleInput(
+                    input_layouts=(
+                        Shard(1) if enable_sp else Replicate(),
+                        Replicate(),
+                        None,
+                        positions_layout,
```
acisseJZhong: super nit: can we do sth like `Replicate() if has_position_id else None`? similar to what you did L136

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -78,20 +81,23 @@ def apply_non_moe_tp(
...
             "tok_embeddings": RowwiseParallel(
                 input_layouts=Replicate(),
-                output_layouts=Shard(1),
-                use_local_output=False,
-            ),
-            "norm": SequenceParallel(
+                output_layouts=Shard(1) if enable_sp else Replicate(),
                 use_local_output=False,
             ),
+            "norm": norm_plan,
```
acisseJZhong, line 96: same nit: `SequenceParallel(use_local_output=False) if enable_sp else NoParallel()`?

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -229,23 +235,31 @@ def apply_non_moe_tp(
...
 ):
     """Apply tensor parallelism."""
     # 1. Parallelize the embedding and shard its outputs (which are the first
     # transformer block's inputs)
     # 2. Parallelize the root norm layer over the sequence dim
     # 3. Parallelize the final linear output layer
+    sp_layout = Shard(1) if enable_sp else Replicate()
```
acisseJZhong, line 244: nit: maybe you can use sp_layout below? 
```
embed_plan = RowwiseParallel(input_layouts=Replicate(), output_layouts=sp_layout)
```

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -54,9 +55,10 @@ def parallelize_qwen3(
         apply_non_moe_tp(
             model,
             tp_mesh,
-            loss_parallel=not parallelism.disable_loss_parallel,
+            enable_loss_parallel=not parallelism.disable_loss_parallel,
             enable_float8_tensorwise_tp=False,
```
tianyu-l, line 59: Actually let's remove the optimization related to `float8_tensorwise` in parallelize.py. AFAIK, tensorwise fp8 is not something people use today. Removing it will also help full DTensor rewrite.

Doesn't have to be in this PR.

tianyu-l (approved):

torchtitan/experiments/rl/unified/models/parallelize.py
```diff
@@ -102,57 +106,55 @@ def apply_non_moe_tp(
...
-            desired_input_layouts=(Replicate(), Replicate(), None, None),
-        )
+    positions_layout = Replicate() if has_position_id else None
 
     # pyrefly: ignore [not-callable]
     for transformer_block in model.layers.values():
+        qk_norm_plan = (
+            SequenceParallel(sequence_dim=2, use_local_output=False)
+            if enable_sp
+            else NoParallel()
```
wwwjn: Can you verify TP only plan works for rl? 
- if qk_norm is NoParallel(), it will mark input_layout to be Replicate() https://github.com/pytorch/torchtitan/blob/1a7a3d32bde858890c79fbfabb422da8d3a9693a/torchtitan/distributed/tensor_parallel.py#L45C14-L45C26. But the input here is sharded on dim-2. 
- And in rl/unified/model/attention.py, we unwrap DTensor and then wrap back with the same placement , which is Replicate(). This is also wrong placement here.

I guess the qk_norm plan be the same as previous plan: keep `SequenceParallel(sequence_dim=2, use_local_output=False)`

Here's some context: https://github.com/pytorch/torchtitan/issues/1867, we apply sequence parallel here is not the same as SP between TP region, it's because NoParallel() hardcoded desired input to be Replicate()

tianyu-l:
any progress on this? @wwwjn @pianpwk

pianpwk:
> any progress on this? @wwwjn @pianpwk

sorry, just updated, thanks @wwwjn for catching it, we don't see the allgathers anymore:
```
  BEFORE fix — q_norm (layer 0):

  FSDPQwen3Model.layers.0.attention.q_norm
  *module type: class 'torchtitan.models.common.rmsnorm.RMSNorm'
  *Parameter List
   *weight: (Replicate(),)
    FORWARD PASS
      *c10d_functional.all_gather_into_tensor: 1
      **aten.view.default
        shape: [torch.Size([8, 2048, 16, 128])]
        sharding: [(Shard(dim=2),)]
      **aten._fused_rms_norm.default
        shape: [torch.Size([8, 2048, 16, 128]), torch.Size([128])]
        sharding: [(Shard(dim=2),), (Replicate(),)]
      **_c10d_functional.all_gather_into_tensor.default
      **_c10d_functional._wrap_tensor_autograd.default
      **_c10d_functional.wait_tensor.default
      **aten.split.Tensor
      **aten.cat.default

  AFTER fix — q_norm (layer 0):

  FSDPQwen3Model.layers.0.attention.q_norm
  *module type: class 'torchtitan.models.common.rmsnorm.RMSNorm'
  *Parameter List
   *weight: (Replicate(),)
    FORWARD PASS
      **aten.view.default
        shape: [torch.Size([8, 2048, 16, 128])]
        sharding: [(Shard(dim=2),)]
      **aten._fused_rms_norm.default
        shape: [torch.Size([8, 2048, 16, 128]), torch.Size([128])]
        sharding: [(Shard(dim=2),), (Replicate(),)]
```

tianyu-l:
cc @shuhuayu you can refer to this PR about how to disable SP

shuhuayu:
> cc @shuhuayu you can refer to this PR about how to disable SP

OK. I currently have `apply_non_moe_tp` and `apply_moe_ep_tp` under local `parallelize.py`, and will rebase once this one is landed.

PR #2539, opened 2026-03-10
Add missing top_p for debug mode (fixes #2537, follow-up for #2538)

### LINKED ISSUE #2537: Unused variable top_p in torchtitan/experiments/rl/unified/actors/generator.py
### Bug description

In `torchtitan/experiments/rl/unified/actors/generator.py` we initialize `top_p` in line 98

```python
class SamplingConfig:
    """Sampling parameters passed to vLLM's SamplingParams."""
    temperature: float = 0.8
    """Sampling temperature. 0.0 = greedy, higher = more random."""
    top_p: float = 0.95
    """Nucleus sampling threshold."""
    max_tokens: int = 100
    """Maximum number of tokens to generate per completion."""
```
and it should probably be used in
```python
                temperature=self.temperature,
                top_p=self.top_p,
                max_tokens=self.max_new_tokens,
                n=self.num_samples_per_prompt,
                seed=self.config.seed,
                logprobs=1,
                prompt_logprobs=1,  # Also get prompt log probs to access prompt token IDs
                output_kind=RequestOutputKind.FINAL_ONLY,  # Only return completed outputs
            )
```

(I added it here) like temperature and max_tokens, but it's missing in the original code. (And it also needs to be added here:
```python
self.top_p = config.samling.top_p
```
(IMO)).

### Versions

Latest

benediktjohannes:
Add missing top_p for debug mode (fixes #2537, follow-up for #2538)

### DIFF
--- a/torchtitan/experiments/rl/unified/config_registry.py
+++ b/torchtitan/experiments/rl/unified/config_registry.py
@@ -96,5 +96,6 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
                 top_p=0.95,
                 max_tokens=50,
             ),
+            attention_backend="FLASH_ATTN",
         ),
     )


benediktjohannes:
cc @joecummings

benediktjohannes:
Also not quite sure whether `attention_backend="FLASH_ATTN",` is missing in debug mode as well or it's intended to be missing here (but probably it's not intended to be missing, so it's likely that we should add this as well here).

wwwjn:
> Also not quite sure whether `attention_backend="FLASH_ATTN",` is missing in debug mode as well or it's intended to be missing here (but probably it's not intended to be missing, so it's likely that we should add this as well here).

Thanks, we should add `attention_backend="FLASH_ATTN"` in debug mode

PR #2540, opened 2026-03-10
[RFC][Module] Replace init_weights with StateInitializer pattern

fegin:
* __->__ #2540
* #2527

Key changes:
1. New StateInitializer(Configurable) base class.
2. Module.Config gains state_initializer field, it is not optional.
3. Remove *kwargs from init_states/init_weights. All changes must
   go through StateInitializer.Config before constructing a module.
   So the parent module is responsible to pass through the state
   initialization config before constructing children.

**Item 3 is the key change that needs to discuss. Whether people agree this direction or people simply want to use kwargs to pass the config when calling init_states.**

### DIFF
--- a/tests/unit_tests/test_linear.py
+++ b/tests/unit_tests/test_linear.py
@@ -56,19 +56,6 @@ class TestLinear(unittest.TestCase):
             # After init_states, weights should no longer be all zero
             self.assertFalse(torch.all(linear.weight == 0))
 
-    def test_init_states_with_override(self):
-        """Linear.init_states respects init_std set via replace_state_init_field."""
-        config = Linear.Config(
-            state_initializer=LinearStateInitializer.Config(init_std=0.01)
-        ).replace_state_init_field(init_std=0.005)
-        linear = config.build(in_features=1000, out_features=500)
-
-        torch.manual_seed(42)
-        with torch.no_grad():
-            linear.init_states()
-        # std should be close to the config init_std
-        self.assertLess(linear.weight.std().item(), 0.005 * 2)
-
     def test_custom_init_std(self):
         """Linear respects custom init_mean and init_std."""
         config = Linear.Config(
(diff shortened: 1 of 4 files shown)


acisseJZhong:
@claude review this pr

torchtitan/models/common/moe/moe.py
```diff
@@ -125,10 +131,10 @@ def forward(
         else:
             return _run_experts_for_loop(w1, w2, w3, x, num_tokens_per_expert)
 
-    def init_weights(self, init_std: float):
+    def init_states(self):
         nn.init.trunc_normal_(self.w1, mean=0.0, std=0.02)
-        nn.init.trunc_normal_(self.w2, mean=0.0, std=init_std)
-        nn.init.trunc_normal_(self.w3, mean=0.0, std=init_std)
+        nn.init.trunc_normal_(self.w2, mean=0.0, std=self._init_std)
```
acisseJZhong, line 136: can we also use self._init_mean?

torchtitan/models/common/moe/moe.py
```diff
@@ -369,6 +378,32 @@ def forward(
         )
 
 
+class MoEStateInitializer(StateInitializer):
+    @dataclass(kw_only=True, slots=True)
+    class Config(StateInitializer.Config):
+        init_std: float = 0.02
```
acisseJZhong, line 384: same to last comment, why not also introducing mean here?

torchtitan/models/common/moe/moe.py
```diff
@@ -165,10 +171,13 @@ def __init__(
...
         gate_bias: bool,
         _debug_force_load_balance: bool = False,
+        init_std: float = 0.02,
     ):
         super().__init__()
-        self.gate = Linear.Config(bias=gate_bias).build(
-            in_features=dim, out_features=num_experts
+        self.gate = (
+            Linear.Config(bias=gate_bias)
+            .replace_state_init_field(init_std=init_std)
```
acisseJZhong, line 179: I found a little bit unintuitive to have `replace_state_init_field` here, can we modify `LinearStateInitializer` directly?

PR #2546, opened 2026-03-11
[graph_trainer] Fix inductor_decomposition_pass input mismatch for models with side effects

xmfan:
Stacked PRs:
 * #2548
 * #2549
 * __->__#2546

--- --- ---

### [graph_trainer] Fix inductor_decomposition_pass input mismatch for models with side effects

The pass was reconstructing joint graph inputs from model.parameters(),
buffers, and forward_inputs, but missed effect tokens that AOT Autograd
prepends as placeholders for side-effectful ops (e.g. MoE load balancing
copy_ mutations). This caused "Input count mismatch: 432 inputs vs 440
placeholders" on DeepSeek-V3.

Fix by reading fake tensor inputs directly from placeholder metadata
instead of reconstructing them, which is robust regardless of what
extra inputs AOT Autograd adds.

Tested with DSv3 16B, 8xH100, EP=8, batch=1, seq=4096, dataset=c4_test:
  step: 20  loss: 7.305  tps: 7,835  tflops: 141.86  mfu: 14.34%  memory: 48.99GiB


xmfan:
@claude review this

yiming0416:
Can the mismatch error be reproduced by `deepseek_v3_debugmodel`? If so, can we add a test in `graph_trainer/tests/integration_tests.py` to guard it with your fix?

aditvenk (approved):
nice! This is so much nicer :)

aditvenk:
> Can the mismatch error be reproduced by `deepseek_v3_debugmodel`? If so, can we add a test in `graph_trainer/tests/integration_tests.py` to guard it with your fix?

I remember testing the previous code with DSv3 and LLama3, but looks like we have an integration test for llama3 only. +1 on adding an integration test for DSv3

PR #2550, opened 2026-03-11
fix(flux): handle corrupted EXIF images in FluxDataset to prevent training crashes

LIUYellowBlack:
## Problem

Training on CC12M crashes at a random step with:

```
File "torchtitan/models/flux/flux_datasets.py", line 243, in __iter__
    sample = next(dataset_iterator)
  ...
File "PIL/ImageOps.py", line 711, in exif_transpose
File "PIL/TiffImagePlugin.py", line 297, in _limit_rational
ZeroDivisionError: division by zero
```

A CC12M image has a corrupted EXIF tag (rational value with denominator=0). PIL raises `ZeroDivisionError` inside `exif_transpose`, which is called by HuggingFace datasets during internal image decoding — before we receive a sample dict — so there is no opportunity to catch it in `_data_processor`.

We first observed this crash at training step 732 (loss=0.67, converging normally). The initial fix (iterator recreation) allowed training to run past 50,000 steps without a crash, and was submitted as this PR. After @wwwjn pointed out that streaming iterator recreation causes data repetition, we added the PIL patch on top. A subsequent 6,000-step run confirmed that the `except Exception` block was **never triggered** — corrupted images are now silently skipped at the PIL level with zero iterator restarts.

## Root Cause

HuggingFace datasets eagerly decode images during `next(dataset_iterator)`. If a CC12M image contains a corrupted EXIF field (rational denominator=0), PIL raises `ZeroDivisionError` inside `exif_transpose` **before** we receive a sample dict. The existing `except StopIteration` does not catch this, so the exception propagates and kills training.

## Fix

Add `except Exception` after `except StopIteration` in `FluxDataset.__iter__`.

On exception, we **recreate the iterator** via `_get_data_iter()` rather than continuing with the old one. This directly addresses the concern in the removed TODO comment: the old iterator'''s internal state may be broken after an exception, but a fresh iterator is clean.

A `consecutive_bad` counter (limit: 200) detects a genuinely broken dataset and raises promptly instead of looping forever.

The `consecutive_bad` counter is reset to 0 on each successfully yielded sample, so transient corrupted images do not accumulate toward the limit.

## Why the PIL patch approach

As @wwwjn noted, recreating the streaming iterator resets it to sample 0, causing data repetition. The PIL patch prevents EXIF errors from ever propagating to the iterator, so the `except Exception` block is effectively never triggered for EXIF corruption. Evidence: in a 6,000-step CC12M training run, the log shows one `UserWarning: Corrupt EXIF data` (caught silently by PIL) and **zero** iterator recreations.

## Why this matters

Corrupted EXIF tags are common in real-world web-crawled datasets (not just CC12M). This fix makes Flux training robust to such samples without requiring the dataset to be pre-filtered.

### DIFF
--- a/torchtitan/models/flux/flux_datasets.py
+++ b/torchtitan/models/flux/flux_datasets.py
@@ -261,10 +261,10 @@ class FluxDataset(IterableDataset, Stateful):
                     continue
             except Exception as e:
                 # HuggingFace datasets decode images internally during iteration;
-                # a corrupted EXIF tag (e.g. rational value with denominator=0
-                # in CC12M) raises ZeroDivisionError inside PIL before we receive
-                # a sample dict. We recreate the iterator rather than continuing
-                # with the old one, because its internal state may be broken
+                # errors (e.g. from truncated image data) are raised before we
+                # receive a sample dict. PIL EXIF errors are suppressed at startup
+                # by exif_safe_patch.apply(), so this path is rarely triggered.
+                # We recreate the iterator because its internal state may be broken
                 # after an exception.
                 consecutive_bad += 1
                 logger.warning(
@@ -277,6 +277,10 @@ class FluxDataset(IterableDataset, Stateful):
                         f"FluxDataset: aborting after {consecutive_bad} consecutive "
                         f"dataset errors. Last: {type(e).__name__}: {e}"
                     ) from e
+                # For non-streaming (map-style) datasets, advance the index so
+                # that _get_data_iter() skips the bad sample on the next attempt.
+                if isinstance(self._data, Dataset):
+                    self._sample_idx += 1
                 dataset_iterator = self._get_data_iter()
                 continue
 


LIUYellowBlack:
@meta-cla recheck

torchtitan/models/flux/flux_datasets.py
```diff
@@ -261,6 +258,26 @@ def __iter__(self):
                         ):
                             self._data.set_epoch(self._data.epoch + 1)
                     continue
+            except Exception as e:
```
wwwjn, line 262: It is worth to add a note here as `StopIteration` is a subclass of Exception, so we would need to keep `except Exception` always after the `except StopIteration:` branch

torchtitan/models/flux/flux_datasets.py
```diff
@@ -261,6 +258,26 @@ def __iter__(self):
...
+                    f"FluxDataset: skipping dataset error "
+                    f"({consecutive_bad}/{MAX_CONSECUTIVE_ERRORS}): "
+                    f"{type(e).__name__}: {e}"
+                )
+                if consecutive_bad >= MAX_CONSECUTIVE_ERRORS:
+                    raise RuntimeError(
+                        f"FluxDataset: aborting after {consecutive_bad} consecutive "
+                        f"dataset errors. Last: {type(e).__name__}: {e}"
+                    ) from e
+                dataset_iterator = self._get_data_iter()
```
wwwjn, line 284: For streaming datasets:  This line would return a  new data iterator without skipping already looped part and resume? I'm worrying this will restart looping from the beginning of the dataset. See _get_data_iter() implementation [here](https://github.com/pytorch/torchtitan/blob/156db2eadada6ad25324bbf9f4ee396a79faa914/torchtitan/models/flux/flux_datasets.py#L233)
- Can you  add a test , or show your local run result to check it also works for streaming dataset?

For non-streaming datasets:  Should we add 1 to sample_idx here to skip the known corrupted sample?

LIUYellowBlack:
> Does it work with checkpoint save and load to resume at where last checkpoint stopped? From CPU test passing it seems so.

Yes — for non-streaming Dataset, state is saved/restored via sample_idx; for streaming IterableDataset, it delegates to self._data.state_dict(). Both paths are covered by the existing CPU tests.

torchtitan/models/flux/flux_datasets.py
LIUYellowBlack: Good catch — I'll add a comment explaining that except StopIteration must come before except Exception since StopIteration is a subclass of Exception.

torchtitan/models/flux/flux_datasets.py
LIUYellowBlack: For streaming datasets, _get_data_iter() restarts from the beginning of the assigned shards — it does not skip past the corrupted sample. In a 30k-step training run on CC12M, 6 iterator restarts were logged by rank 0 (OSError: image file is truncated), while other ranks’ output was filtered. Consequently, the full picture across all ranks remains unknown. Training loss decreased steadily from 2.04 to 0.47, and the model continued to converge, but some data repetition from restarts cannot be ruled out. A proper fix would save/restore the streaming iterator’s state_dict() to resume precisely — this is noted as a known limitation.

For non-streaming datasets, sample_idx should indeed be incremented to skip the bad sample. Since this code path was not exercised in the reported training, this is also noted as a known limitation.

wwwjn: >  but some data repetition from restarts cannot be ruled out

Thanks for confirming, this would not matching user's expected behavior, and I think the training would be keep looping between data sample 0 to the first images. I believe I met this issue before and had a hard time realizing I'm only training with partial dataset. 

I think we'd explicitly error out for streaming dataset, otherwise it would be very error prone and hard to realize. cc @tianyu-l

wwwjn: @LIUYellowBlack  
> but some data repetition from restarts cannot be ruled out. 
This should be fixed at huggingface dataset level package, not torchtitan. And streaming dataset using repeated data is too error prone now.

LIUYellowBlack: Hi  thanks for the detailed feedback — you’re right that streaming iterator recreation is too error-prone.

I’ve addressed this by adding exif_safe_patch.py, which monkey-patches the two PIL functions that HuggingFace datasets calls internally (PIL.ImageOps.exif_transpose and PIL.Image.Image.getexif). Since the EXIF error is now caught inside PIL before it ever reaches our iterator, the except Exception block is never triggered for this case — no iterator recreation, no data repetition.

I validated this with a 6,000-step CC12M training run: the log shows one UserWarning: Corrupt EXIF data (handled silently by PIL) and zero except Exception triggers or iterator restarts. The except Exception block is kept as a safety net for other unexpected errors (e.g. network failures), and I also fixed the non-streaming dataset case by incrementing self._sample_idx before recreating the iterator, as you suggested.

wwwjn: Thanks, monkey patch does provide a solution for streaming dataset, but the cons are also obvious -- The user will silently ignore the dataset image corruption issue. In actual large-scale training, the user might need to know the data is corrupted and take action (eg, re-collect data if needed). 

And also the PIL change would be better hosted by upper-level lib like huggingface  dataset, and provide an option (eg, skip_corrupted_image=True/False) for training framework

PR #2556, opened 2026-03-11
Add ChatDataset and ChatDataLoader for SFT training

joecummings:
## Summary

Adds `ChatDataset` and `ChatDataLoader` to `torchtitan/hf_datasets/text_datasets.py` for supervised fine-tuning on instruction/conversation datasets.

- `ChatDataset`: single-turn [user, assistant] tokenization with prompt masking (IGNORE_INDEX), greedy sequence packing with per-document RoPE positions, checkpointing via Stateful. Supports both map-style and iterable/streaming datasets.
- `ChatDataLoader`: thin `ParallelAwareDataloader` subclass (same pattern as `HuggingFaceTextDataLoader`)
- SFT config functions for llama3 (`sft_debugmodel`) and qwen3 (`sft_qwen3_8b_math`)
- Removes `torchtitan/experiments/sft/` (promoted to core)
- Fixes `HuggingFaceTextDataset` to shuffle map-style datasets on re-loop (previously iterated in the same order every epoch)

## Training results

| Step | Loss | Grad Norm | MFU |
|------|------|-----------|-----|
| 1 | 1.906 | 24.89 | 0.59% |
| 10 | 0.846 | 3.37 | 21.89% |
| 50 | 0.853 | 2.42 | 21.86% |
| 100 | 0.753 | 1.95 | 21.85% |
| 180 | 0.664 | 1.96 | 21.85% |

WandB run: https://wandb.ai/jcummings/torchtitan/runs/wnk5kdbe

## Correctness verification

**Tokenization comparison** ([script](https://gist.github.com/joecummings/2fe7f59a5f9e087a033e457b694329c8), [output](https://gist.github.com/joecummings/1163488a7b4d52fd2cd1c7d36701609d)): Verified against HuggingFace `AutoTokenizer` on GSM8K with Qwen3-8B tokenizer. All 3 phases pass (chat template rendering, full tokenization, prompt/label split).

**Single-sample loss comparison** ([script](https://gist.github.com/joecummings/2730f3ef351e0fc79b6c20fea6ec28c5), [output](https://gist.github.com/joecummings/224070ffc3deb816c5fc96e07d66f160)): Forward pass on the same sample with the same Qwen3-0.6B model. Delta: 0.020.

### TRL loss comparison (Qwen3-0.6B, 20 steps, no packing, sequential data)

([script](https://gist.github.com/joecummings/1924a78bccdcb3e084f8b2fc5218f337), [output](https://gist.github.com/joecummings/6c045ca370e56ec822f227583831098f))

| Step | TorchTitan | TRL |
|------|-----------|-----|
| 1 | 1.279 | 1.656 |
| 2 | 0.912 | 0.996 |
| 3 | 0.680 | 0.744 |
| 4 | 0.812 | 0.814 |
| 5 | 0.843 | 0.799 |
| 10 | 0.963 | 0.993 |
| 14 | 0.415 | 0.415 |
| 20 | 0.417 | 0.527 |

TorchTitan losses are consistently lower. The gap is explained by TRL appending a duplicate EOS token to every sample. TRL's `add_eos` checks `completion.endswith(eos_token)`, but chat templates produce `<|im_end|>\n` (with a trailing newline), so the check fails and TRL appends a redundant `<|im_end|>`:

When we intentionally replicate TRL's behavior (add duplicate EOS + use HF internal shift), the losses match to within 0.0002:

## Test plan

- [x] 14 unit tests in `test_chat_dataset.py` (label masking, token shifting, packing, per-document positions, overflow drop, message validation, checkpointing round-trip, infinite looping)
- [x] `test_dataloader.py` and `test_dataset_checkpointing.py` still pass (existing completion path unchanged)
- [x] Full 180-step Qwen3-8B SFT run on GSM8K converges (loss 1.91 -> 0.66)
- [x] Tokenization matches HuggingFace `AutoTokenizer` exactly (Qwen3-0.6B and 8B)
- [x] Single-sample forward loss matches HF to within 0.02
- [x] TRL loss gap fully explained by duplicate EOS (matches to 0.0002 when replicated)
- [x] `pre-commit run --all-files` passes (flake8, ufmt, pydoclint, codespell)

Also included: [script to migrate old checkpoints](https://gist.github.com/joecummings/d94c21bf0618dd1bdc61a78eb818dabf)

### DIFF
--- a/torchtitan/models/qwen3/config_registry.py
+++ b/torchtitan/models/qwen3/config_registry.py
@@ -236,11 +236,7 @@ def sft_qwen3_8b_math() -> Trainer.Config:
             },
         ]
 
-    model_spec = model_registry("8B")
-    # pyrefly: ignore [missing-attribute]
-    model_spec.model.layer.attention.attn_backend = "varlen"
-    # pyrefly: ignore [missing-attribute]
-    model_spec.model.layer.attention.attn_mask_type = "block_causal"
+    model_spec = model_registry("8B", attn_backend_override="varlen")
     return Trainer.Config(
         hf_assets_path="./assets/hf/Qwen3-8B",
         model_spec=model_spec,


torchtitan/experiments/sft/dataset.py
tianyu-l, line 1: does it make sense to put in torchtitan/hf_datasets

torchtitan/experiments/sft/configs.py
```diff
@@ -0,0 +1,39 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass, field
+
+from torchtitan.experiments.sft.dataset import SFTDataLoader
+from torchtitan.trainer import Trainer
+
+
+@dataclass(kw_only=True, slots=True)
+class SFTTrainerConfig(Trainer.Config):
```
tianyu-l: I'm interested in exploring the feasibility to land this in core (outside `experiments/`), possibly by consolidating the trainer / dataloader and their configs.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+        # Variables for checkpointing
+        self._sample_idx = 0
+        self._epoch: int = 0
+        self._pack_buffer_input: list[int] = []
+        self._pack_buffer_label: list[int] = []
+        self._pack_buffer_positions: list[int] = []
+
+        self._logged_first_sample = False
+
+    def _get_data_iter(self):
```
tianyu-l, line 293: the impl is different from existing pretraining dataloader -- does this support iterable dataset?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+            logger.warning(
+                "ChatDataLoader: pack_sequences=True requires block_causal "
+                "attention (flex or varlen backend) to prevent cross-document "
+                "attention leakage. Ensure your model config sets "
+                "attn_backend='flex' or 'varlen' and attn_mask_type='block_causal'."
+            )
+
+        dataset = load_dataset(config.dataset_path, **config.load_dataset_kwargs)
+
+        chat_ds = ChatDataset(
```
tianyu-l, line 511: Is cls the only actual difference that necessitates `ChatDataLoader`? Maybe we can do if/else factory here in `TextDataLoader` based on the dataset type. (If we can't "tell" from the dataset, we can add a field in https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/__init__.py#L15)

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+                yield self._flush_pack_buffer()
+
+            if not self.infinite:
+                logger.warning(f"Chat dataset '{self._dataset_id}' has run out of data")
+                break
+            else:
+                self._sample_idx = 0
+                self._epoch += 1
+                # pyrefly: ignore [bad-assignment]
+                self._data = self._data.shuffle(seed=42 + self._epoch)
```
tianyu-l: why only do this to ChatDataset not TextDataset?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+                logger.warning(
+                    f"Chat dataset '{self._dataset_id}' is being re-looped "
+                    f"(epoch {self._epoch})"
+                )
+
+    def state_dict(self):
```
tianyu-l, line 444: similar to https://github.com/pytorch/torchtitan/pull/2556/changes#r2986065093 I think streaming=True/False may need different impl

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+        mask_end = min(prompt_len - 1, len(label_ids))
+        label_ids[:mask_end] = [IGNORE_INDEX] * mask_end
+
+        return input_ids, label_ids
+
+    def __iter__(self):
+        if self.pack_sequences:
+            yield from self._iter_greedy_packed()
+        else:
+            yield from self._iter_unpacked()
```
tianyu-l: how valuable is this in reality? Would user ever want to use it over packing?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+                result = self._tokenize_sample(sample)
+                self._sample_idx += 1
+                if result is None:
+                    continue
+
+                input_ids, label_ids = result
+
+                pad_len = self.seq_len - len(input_ids)
+                if pad_len > 0:
+                    input_ids = input_ids + [self._eos_id] * pad_len
```
tianyu-l: Padding by eos_id instead of pad_id sounds unintuitive to me. In current torchtitan, I wonder if the flex attention masking can handle this correctly: e.g. there should be 0 computation in attention, instead of eos_id attending to itself.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +217,319 @@ def __init__(
...
+        # with add_generation_prompt=True.
+        prompt_text = self._tokenizer.apply_chat_template(
+            messages[:1], add_generation_prompt=True
+        )
+        prompt_tokens = self._tokenizer.encode(prompt_text, add_bos=True, add_eos=False)
+        prompt_len = len(prompt_tokens)
+
+        # Mask prompt tokens in labels (shifted by one for next-token prediction)
+        mask_end = min(prompt_len - 1, len(label_ids))
+        label_ids[:mask_end] = [IGNORE_INDEX] * mask_end
```
tianyu-l, line 366: oh this is interesting, TIL

torchtitan/models/qwen3/config_registry.py
```diff
@@ -212,6 +221,61 @@ def qwen3_moe_debug() -> Trainer.Config:
...
+        ),
+        training=TrainingConfig(
+            local_batch_size=1,
+            seq_len=2048,
+            steps=180,
+        ),
+        dataloader=ChatDataLoader.Config(
+            dataset_path="openai/gsm8k",
+            load_dataset_kwargs={"name": "main", "split": "train"},
+            sample_processor=process_sample,
```
tianyu-l, line 258: I wonder if it makes sense to have a fixed registration table at https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/text_datasets.py#L35

It certainly makes writing test easier inside torchtitan (instead of ad hoc in config_registry like here), but makes it harder for people outside. Either way works for me, but let's make pretrain / sft datasets consistent.

torchtitan/hf_datasets/text_datasets.py
joecummings: Updated - I had it working first with just map-style datasets

torchtitan/hf_datasets/text_datasets.py
joecummings: I use this exclusively for testing. Since I don't imagine we will check in the script that directly compares loss curves against TRL, we can remove after proving parity.

torchtitan/hf_datasets/text_datasets.py
joecummings: See above comment about expanding support to iterable

torchtitan/hf_datasets/text_datasets.py
joecummings: You're correct that the dataloaders are just wrappers around the datasets which they construct. However, any change to unify the dataloaders (which I agree should be done) will require BC breaking changes to at least the params. 

Seeing as I'd like to propose some larger changes to how Titan does dataloading, I'd rather get this SFT work in with these clean boundaries.

torchtitan/models/qwen3/config_registry.py
joecummings: As we evolve more into posttraining, I'd like Titan to adopt more of this "recipe"-style config. Each one of these is so model/data/algo specific that it quickly becomes cumbersome to specify all as a fixed registration table and even more cumbersome to ensure that each piece is used with the correct combo of model/data/algo.

torchtitan/hf_datasets/text_datasets.py
joecummings: TextDataset is wrong here for map datasets (which aren't actually used in any of the configs, so not a big deal) - I can fix in that PR as well

torchtitan/hf_datasets/text_datasets.py
joecummings: This is the most common approach for packing (axolotl, trl). In practice, there is very little padding so this extra computation is negligible. 

But you're right, we could explore adding a padding mask to the existing document_mask via flex_attention's mask_mod. Let me add that as a TODO.

tests/integration_tests/features.py
```diff
@@ -562,30 +562,12 @@ def build_features_test_list() -> list[OverrideDefinitions]:
...
-                    "--compile.enable",
-                ],
-            ],
-            "FSDP+TP+PP+compile with torchcomms",
-            "torchcomms_3d_dp+tp+pp+compile",
-            ngpu=8,
-            skip_rocm_test=True,
+            "SFT ChatDataset integration test",
+            "sft",
+            ngpu=1,
```
tianyu-l: at least 2 / 4 so we get signal from distributed

torchtitan/hf_datasets/text_datasets.py
tianyu-l, line 72: For consistency, maybe we should rename this to 
```suggestion
class TextDataset(IterableDataset, Stateful):
```
and also similar for dataloader. Doesn't have to be in this PR tho.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -240,3 +222,345 @@ def __init__(
...
+        self._sample_processor = sample_processor
+
+        self._dataset_id = f"{dataset.info.dataset_name}/{dataset.split}"
+
+        # Variables for checkpointing
+        self._sample_idx = 0
+        self._epoch: int = 0
+        self._pack_buffer_input: list[int] = []
+        self._pack_buffer_label: list[int] = []
+        self._pack_buffer_positions: list[int] = []
```
tianyu-l: should also name txt dataset fields consistently

torchtitan/hf_datasets/text_datasets.py
joecummings: Agree, will do in a follow-up.

torchtitan/models/qwen3/config_registry.py
```diff
@@ -216,3 +219,56 @@ def qwen3_moe_debug() -> Trainer.Config:
...
+            {
+                "role": "assistant",
+                "reasoning_content": reasoning.strip(),
+                "content": final_answer.strip(),
+            },
+        ]
+
+    model_spec = model_registry("8B")
+    # pyrefly: ignore [missing-attribute]
+    model_spec.model.layer.attention.attn_backend = "varlen"
```
wwwjn: This override can be removed if #2649  landed. cc @zhxchen17  please update these part if this PR (#2556) lands first.

PR #2557, opened 2026-03-11
Adopt Local Map Wrapper for Inner Attention

acisseJZhong:
This is a continuation of work in #2480 by @pianpwk 

  **Summary**
                                                                                
  - Add _InnerAttentionBase base class to attention.py — overrides `__call__`
  to wrap `nn.Module.__call__` with local_map, converting TP DTensor inputs       
  to local tensors before any `forward_pre_hook` fires, and wrapping
  outputs back to DTensor after all `forward_hooks` complete. Placements
  and device mesh are inferred from the input DTensors at runtime.
 - Enable TP+CP support for Qwen3 by having `LocalMapModule` handle the
  TP/CP boundary. Qwen3 requires this because it uses use_local_output=False
  on wq/wk/wv (needed for QK norms with SequenceParallel), producing
  DTensors that CP hooks cannot directly consume.
  - Add integration test for Qwen3 FSDP+TP+CP on 8 GPUs.

**Test**
```
NCCL_NVLS_ENABLE=0 MODULE=qwen3 CONFIG=qwen3_debugmodel ./run_train.sh --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.context_parallel_degree 2 --compile.enable  --compile.backend eager 
```

### DIFF
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -114,7 +114,8 @@ class LocalMapAttention(Module):
                     f"LocalMapAttention requires Shard(1) placements "
                     f"(n_heads dim), but got {p} at position {i}"
                 )
-            # return_lse=True produces two outputs
+            # return_lse=True (e.g. gpt_oss attention sinks) produces
+            # 2 outputs instead of 1, requiring different out_placements.
             return_lse = kwargs.get("return_lse", False)
             out_placements = (
                 (q.placements, q.placements) if return_lse else (q.placements,)
@@ -127,9 +128,8 @@ class LocalMapAttention(Module):
                     in_grad_placements=(q.placements, k.placements, v.placements),
                     device_mesh=q.device_mesh,
                 )
-            return self._local_map_fn(
-                q, k, v, **kwargs
-            )  # pyrefly: ignore [bad-argument-count]
+            # pyrefly: ignore [bad-argument-count]
+            return self._local_map_fn(q, k, v, **kwargs)
         return super().__call__(q, k, v, **kwargs)
 
     def forward(
--- a/torchtitan/models/gpt_oss/parallelize.py
+++ b/torchtitan/models/gpt_oss/parallelize.py
@@ -242,14 +242,6 @@ def apply_non_moe_tp(
             "attention.wq": ColwiseParallel(use_local_output=False),
             "attention.wk": ColwiseParallel(use_local_output=False),
             "attention.wv": ColwiseParallel(use_local_output=False),
-            "attention.inner_attention": PrepareModuleInputOutput(
-                input_layouts=(Shard(1), Shard(1), Shard(1)),
-                desired_input_layouts=(Shard(1), Shard(1), Shard(1)),
-                use_local_input=True,
-                output_layouts=(Shard(1), Shard(1)),
-                desired_output_layouts=(Shard(1), Shard(1)),
-                use_local_output=False,
-            ),
             "attention.wo": RowwiseParallel(output_layouts=Shard(1)),
             "ffn_norm": SequenceParallel(),
         }


torchtitan/models/common/attention.py
```diff
@@ -60,16 +62,80 @@ class VarlenMetadata(NamedTuple):
...
+        k: torch.Tensor,
+        v: torch.Tensor,
+        **kwargs,
+    ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+        if isinstance(q, DTensor):
+            assert isinstance(k, DTensor) and isinstance(
+                v, DTensor
+            ), "q, k, v should all be DTensors"
+            if self._local_map_fn is None:
+                self._local_map_fn = local_map(
```
acisseJZhong, line 124: @fegin i had this local_map wrapping around CP pre-forward hook and post-forward hook, so that the local map will handle conversion of Dtensor to plain tensor before the CP pre-forward hook fires. If it's the other way around, i will hit assertion error https://github.com/pytorch/pytorch/blame/main/torch/distributed/tensor/experimental/_context_parallel/_attention.py#L1394 

not sure if this is the best way; and maybe you already have some plan on rewriting the CP API for full DTensor

fegin: yes, this makes sense. Some comments

1. Instead of `_InnerAttentionLocalMap`, we should just implement this logic into `Module`. This is going to be something we need anyway for MoE. The implementation can be less generic at this point and leave some TODOs that once we integrate the config system with sharding spec, the implementation should be more generic.

2. We will have to ensure `torch.compile` doesn't break. So try with `--compile.enable`.

fegin: I realized that these inner wrappers are not Module yet. You can just wrap them with Module with Config to be empty at this moment.

tianyu-l: Or can we inherit Module and call them `LocalMapModule`?

acisseJZhong: Compile is not working with this super().__call__ approach, Claude tells me that 

>   Why it breaks compile: When dynamo encounters self.inner_attention(xq, xk, xv, ...), it detects that
>   LocalMapModule has a custom __call__ (not the standard nn.Module.__call__). Dynamo then tries to trace through  
>   the custom __call__ as a regular Python function. Inside __call__, the super().__call__(q, k, v, **kwargs) call
>   bypasses dynamo's special nn.Module call handling (which normally inlines forward() and properly handles hooks).
>    This causes FX nodes to be created without proper meta["val"] metadata, leading to the inductor KeyError: 'val'
>    error.

Not sure how six handles compile with this __call__ override.

tianyu-l: @xmfan knows more details

xmfan: This implementation is different. nn.Module inheritance with compile is kinda brittle. If the goal here is to local_map a nn.Module forward, the more idiomatic way is to use nn.Module. register_forward_pre_hook (fires before module, you'll need prepend) and nn.Module.register_forward_hook (fires after module): https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html

You can verify on a standalone example too, IIRC compile does support it. 

There might be a weird interaction with FSDP hook registration order. IIRC, it did some prepending too.

acisseJZhong: how should the hook work with local_map? I want to have local_map wrapping around the pre-forward and post-forward hook of `super().__call__`? I can imagine to use `to_local` and `from_local` as hooks for nn.Module but I want to have `local_map` here.

tianyu-l: I think to_local and from_local on module boundary is fine, because we can still force a pair of them within clear boundary. Would be good to know if it works with spmd_types, which is a context manager.

xmfan: for context managers, we can manually call with `__enter__()` and `__exit__()`

tianyu-l: @xmfan there's some limitation due to torch_function, so we can't use enter and exit. This is the current entry point https://fburl.com/code/ypffryqp

acisseJZhong: @xmfan i changed to hook based approach here https://github.com/pytorch/torchtitan/pull/2557/changes/26cdceb2231806643c0d35330ce22dbd37f199b8 but compile still fails. P2234099693

torchtitan/models/common/attention.py
```diff
@@ -60,7 +63,69 @@ class VarlenMetadata(NamedTuple):
...
+            return args, kwargs
+        module._placements = args[0].placements
+        module._device_mesh = args[0].device_mesh
+        return tree_map(_to_local, args), kwargs
+
+    @staticmethod
+    def _post_hook(module, _input, output):
+        """Restore local tensors back to DTensors."""
+        placements = module._placements
+        mesh = module._device_mesh
```
tianyu-l: may not exist if args[0] is not DTensor?

torchtitan/models/common/attention.py
```diff
@@ -60,7 +63,69 @@ class VarlenMetadata(NamedTuple):
 AttentionMasksType = dict[str, BlockMask] | BlockMask | VarlenMetadata
 
 
-class VarlenAttentionWrapper(torch.nn.Module):
+def _to_local(x: Any) -> Any:
```
tianyu-l: can define inline? doesn't need to be at root level.

torchtitan/models/common/attention.py
```diff
@@ -60,7 +63,69 @@ class VarlenMetadata(NamedTuple):
...
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        pass
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.register_forward_pre_hook(
+            LocalMapModule._pre_hook, with_kwargs=True, prepend=True
```
tianyu-l: any reason why `prepend=True`?

torchtitan/models/common/attention.py
```diff
@@ -60,7 +63,69 @@ class VarlenMetadata(NamedTuple):
...
+def _from_local(x: Any, *, placements: Any, mesh: Any) -> Any:
+    """Convert a local tensor back to a DTensor."""
+    if isinstance(x, DTensor):
+        return x
+    if isinstance(x, torch.Tensor):
+        return DTensor.from_local(x, mesh, placements)
+    return x
+
+
+class LocalMapModule(Module):
```
tianyu-l:
```suggestion
class LocalMapAttention(Module):
```
The implementation is very restricted, in the sense that
- input placements and output placements must be the same
- grad placements of inputs must stay the same. From typing perspective, this is only true for `Shard`.

Taking Flex all-gather based CP as an example
- in TP mesh, things are Shard on head dim. This is fine.
- on CP mesh, q is Shard and kv are Replicate. The reason you can assume kv grad are also Replicate is because CP API internally does this reduce-scatter for us https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/experimental/_context_parallel/_cp_custom_ops.py#L37, which is fine for now

Let's restrict the scope to Attention, since this is not a general local map module.

Please add detailed docstring. (Ask me / claude if anything is not clear)

torchtitan/models/common/attention.py
```diff
@@ -60,7 +63,69 @@ class VarlenMetadata(NamedTuple):
...
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.register_forward_pre_hook(
+            LocalMapModule._pre_hook, with_kwargs=True, prepend=True
+        )
+        self.register_forward_hook(LocalMapModule._post_hook)
+
+    @staticmethod
+    def _pre_hook(module, args, kwargs):
```
tianyu-l: give it more meaningful name, e.g. `_inputs_to_local`, `_outputs_from_local`

torchtitan/models/qwen3/parallelize.py
```diff
@@ -125,19 +127,26 @@ def parallelize_qwen3(
...
         apply_cp_to_attention_module(
             # pyrefly: ignore [missing-attribute, not-callable]
             [block.attention.inner_attention for block in model.layers.values()],
             parallel_dims.get_mesh("cp"),
             attn_backend,
         )
 
+        # TODO: cuDNN SDPA backward has a stride mismatch bug with CP.
+        # Exclude cuDNN until PyTorch fix lands. See https://github.com/pytorch/pytorch/issues/176915.
+        if attn_backend == "sdpa":
```
tianyu-l: I don't know how I feel about this workaround.

If we have to do it, don't do per model. Instead, do it in SDPA module, e.g. maybe by override the pre forward hook and subtract cudnn from `self.sdpa_backends`

torchtitan/models/common/attention.py
fegin: The scope of this LocalMapModule is restricted to Attention because right now we don't have a way to configure local map, which is why this module is put here and everything  is fixed. 

However, I think this is a good start though to demonstrate how to do module level local map and extend it to a general implementation once our config includes local map.

torchtitan/models/common/attention.py
fegin: @acisseJZhong Is there a way to use `__call__` with `torch.compile`? I am worried that hook will cause the order issues, which happened many times before. We didn't have a good way to do this because we could not change nn.Module. But since now we have our own Module, I would prefer using __call__ if possible. `torch.compile` is definitely the key blocker we need to fix.

cc., @xmfan

torchtitan/models/common/attention.py
acisseJZhong: I mimic six's way D83451817, but moving inline sounds better.

torchtitan/models/common/attention.py
acisseJZhong: @tianyu-l in fact this is the reason I also prefer to have local map wrapping super.__call__, and solve any compile issues on that. Otherwise for this forward pre hook, and for the CP post forward hook, I need to have `prepend=True`; to make sure that the hooks I added can wrap around CP hooks.

I would prefer rolling back to my last commit and fix compile problem. https://github.com/pytorch/torchtitan/pull/2557/changes/26cdceb2231806643c0d35330ce22dbd37f199b8 @xmfan

torchtitan/models/common/attention.py
acisseJZhong: > on CP mesh, q is Shard and kv are Replicate

When I specify local_map/to/from_local, q, k, v should all be Shard on TP mesh. 
And later on CP mesh, q, k, v should all be Shard and this is because of CP pre hook and post hook? https://github.com/pytorch/pytorch/blob/4d01cdb5b2a633c45471bdaf8d8d544c4bb2572a/torch/distributed/tensor/experimental/_context_parallel/_attention.py#L1396

tianyu-l: oh, that's right -- if that's the case, can we assert every placement to be Shard?

acisseJZhong: yes, this is what I am doing in the local map approach. Let me roll back and add explicitly assertion to Shard for now

torchtitan/models/common/attention.py
fegin: Directly replacing forward with wrapped forward work, https://github.com/pytorch/torchtitan/pull/2621.

torchtitan/models/llama4/parallelize.py
```diff
@@ -638,6 +638,12 @@ def apply_compile(model: nn.Module, compile_config: CompileConfig, ep_enabled: b
     # NOTE: This flag is needed for torch.compile to avoid graph breaking on dynamic shapes in token-choice MoE
     # but it is experimental.
     torch._dynamo.config.capture_scalar_outputs = True
+    # Skip replaying forward side effects (e.g. RoPE cache updates) during
+    # the AC recompute in backward. Eager AC replays the forward python
+    # side-effects in backward, but torch.compile has no easy way to reapply
+    # python mutations in the backward. Setting this flag accepts this eager
+    # and compile divergence by skipping reapplication of side effects.
+    torch._dynamo.config.skip_fwd_side_effects_in_bwd_under_checkpoint = True
```
fegin: Can we centralize this option? Otherwise people may forget this when creating a new model parallelize.py. We should at least do this for the core module. We can create a util in distributed/compiler.py

fegin: Let's create an util for setting common compiler flags in `distributed/compiler.py`.

tianyu-l: Also, I think this entire function can be put in distributed/compile.py.
We can have two, one for dense one for sparse. If @weifengpy 's fix is ready, one is enough.

torchtitan/models/common/attention.py
```diff
@@ -60,16 +63,90 @@ class VarlenMetadata(NamedTuple):
 AttentionMasksType = dict[str, BlockMask] | BlockMask | VarlenMetadata
 
 
-class VarlenAttentionWrapper(Module):
+class LocalMapModule(Module):
```
tianyu-l:
```suggestion
class LocalMapAttention(Module):
```

acisseJZhong:
thanks for the command! the loss should match exactly with baseline 
```
[LOSS_COMPARE] Summary statistics:
[LOSS_COMPARE] Average baseline loss:  3.1069710278511047
[LOSS_COMPARE] Average test loss: 3.1069710278511047
[LOSS_COMPARE] Average difference:     0.000000
```

torchtitan/models/common/attention.py
```diff
@@ -60,16 +63,90 @@ class VarlenMetadata(NamedTuple):
...
+            for tensor, name in ((q, "q"), (k, "k"), (v, "v")):
+                for p in tensor.placements:
+                    assert isinstance(p, Shard), (
+                        f"LocalMapModule requires Shard placements, "
+                        f"but {name} has placement {p}"
+                    )
+            if self._local_map_fn is None:
+                self._local_map_fn = local_map(
+                    super().__call__,
+                    in_placements=(q.placements, k.placements, v.placements),
```
tianyu-l, line 126: is it better if we assert they all have the same placements

torchtitan/models/common/attention.py
```diff
@@ -60,16 +63,90 @@ class VarlenMetadata(NamedTuple):
...
+        return super().__call__(q, k, v, **kwargs)
+
+    def forward(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        **kwargs,
+    ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+        return self._forward_local(q, k, v, **kwargs)
```
tianyu-l: why having this indirection?

torchtitan/models/common/attention.py
```diff
@@ -60,16 +63,90 @@ class VarlenMetadata(NamedTuple):
...
+            # as qkv are sharded on head dim. CP is handled
+            # independently by _ContextParallel hooks inside
+            # nn.Module.__call__.
+            for tensor, name in ((q, "q"), (k, "k"), (v, "v")):
+                for p in tensor.placements:
+                    assert isinstance(p, Shard), (
+                        f"LocalMapModule requires Shard placements, "
+                        f"but {name} has placement {p}"
+                    )
+            if self._local_map_fn is None:
```
tianyu-l, line 123: IIUC the purpose of having this check is for 2nd, 3rd, ... call to this attention to reuse the `_local_map_fn`. I guess that's safe in torchtitan setup, where we never change the inputs placements from iteration to iteration, but we could've added a guard before this line.

torchtitan/distributed/compile.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    # Needed for torch.compile to avoid graph breaking on dynamic shapes in
+    # token-choice MoE, but it is experimental.
+    torch._dynamo.config.capture_scalar_outputs = True
+    # Skip replaying forward side effects (e.g. RoPE cache updates) during
+    # the AC recompute in backward. Eager AC replays the forward python
+    # side-effects in backward, but torch.compile has no easy way to reapply
+    # python mutations in the backward. Setting this flag accepts this eager
+    # and compile divergence by skipping reapplication of side effects.
+    torch._dynamo.config.skip_fwd_side_effects_in_bwd_under_checkpoint = True
```
tianyu-l: do you only need to set this for MoE?

torchtitan/distributed/compiler.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+
+def set_common_compiler_flags() -> None:
```
tianyu-l: the naming of this function is vague -- please check the other comments on how to organize this file -- also nit: we should call it compile.py following torchtitan convention.

torchtitan/models/common/attention.py
acisseJZhong: We already check placements at L105 for each call?

torchtitan/models/common/attention.py
acisseJZhong: I thought L103 does it? we checked q/k/v are Shard.

torchtitan/models/common/attention.py
tianyu-l: you didn't check Shard(0) / Shard(1) / Shard(2) / ...

torchtitan/models/common/attention.py
tianyu-l: same comment

torchtitan/models/common/attention.py
acisseJZhong: oops Claude indeed complicates this, removed!

torchtitan/distributed/compile.py
acisseJZhong: `capture_scalar_outputs` is MoE related but `skip_fwd_side_effects_in_bwd_under_checkpoint` is not MoE related. Both of them are used in llama4 parallelize plan and graph trainer compile.

torchtitan/distributed/compiler.py
acisseJZhong: I renamed this file to distributed/compile.py. 

> We can have two, one for dense one for sparse.

didn't get this part, does Wei has a PR?

torchtitan/models/common/attention.py
acisseJZhong: added! thanks for the catch

torchtitan/distributed/compiler.py
tianyu-l: yes https://github.com/pytorch/torchtitan/pull/2281

torchtitan/models/common/attention.py
```diff
@@ -60,7 +63,76 @@ class VarlenMetadata(NamedTuple):
...
+            # independently by _ContextParallel hooks inside
+            # nn.Module.__call__.
+            assert q.placements == k.placements == v.placements, (
+                f"q, k, v must have the same placements, "
+                f"but got q={q.placements}, k={k.placements}, v={v.placements}"
+            )
+            # qkv are (bs, n_heads, seqlen, head_dim) and must be sharded
+            # on the n_heads dim (dim 1)
+            for i, p in enumerate(q.placements):
+                assert p == Shard(1), (
```
tianyu-l, line 113: after full dtensor rewrite, DP mesh will be included, where qkv will be Shard(0) on DP mesh -- please leave a TODO

torchtitan/distributed/compile.py
tianyu-l: Does other model (e.g. llama3, gpt-oss) not need this to work with TP + CP? (Not sure if it's related to `use_local_output=False` in Qwen3). Btw, I think gpt-oss has an attention which should also be refactored.

In addition, my suggestion is that
- we have `apply_compile_dense` and `apply_compile_sparase` in this file. After Wei's https://github.com/pytorch/torchtitan/pull/2281 we may only need one.
- refactor each model to call into either one in this function, instead of writing their own or importing from llama3/llama4
- the dense compile function sets `skip_fwd_side_effects_in_bwd_under_checkpoint`
- the sparse compile function sets both

torchtitan/distributed/compile.py
```diff
@@ -0,0 +1,23 @@
...
+
+def set_dynamo_flags_for_compile() -> None:
+    """Set common torch._dynamo config flags for distributed compilation.
+
+    This should be called before torch.compile() in any compilation code path.
+    """
+    # Needed for torch.compile to avoid graph breaking on dynamic shapes in
+    # token-choice MoE, but it is experimental.
+    torch._dynamo.config.capture_scalar_outputs = True
+    # Skip replaying forward side effects (e.g. RoPE cache updates) during
```
tianyu-l, line 58: Where do we have this side effects exactly? Is it not a concern?

Does this change numerics between eager and compile(aot_eager backend), when AC is enabled?

acisseJZhong: yes this will make eager and compile converges, as noted in the comment. 

from @anijain2305 

> Eager AC replays the forward python side-effects in backward. Many users do not like or need it. They just need the fwd torch ops to be replayed in backward.

> Compile AC does not have a way to even replay python side-effects in backward. So it graph breaks. But if a user does not need to replay the python side effect in the backward (which is the case in torchtitan), they can use that flag.

> The default is false because we would like to have compile AC and eager AC not diverge w/o user buy-in.

tianyu-l: > But if a user does not need to replay the python side effect in the backward (which is the case in torchtitan), they can use that flag.

This is saying "does not need", my question is "do we have" any side-effect in eager. I mean if we don't have any side effect, why do we need to toggle this field in the first place?

acisseJZhong:
> Does other model (e.g. llama3, gpt-oss) not need this to work with TP + CP? (Not sure if it's related to use_local_output=False in Qwen3).

for llama3 which use `use_local_output=False`, we don't have this problem/no need to set any dynamo flags. I tried to set `use_local_output=True` calling `apply_compile_dense` and compile works, `skip_fwd_side_effects_in_bwd_under_checkpoint=True` is necessary for dense models.

torchtitan/distributed/compile.py
acisseJZhong: @anijain2305 might answer better here. I believe we should have side effects in eager(but idk what side effects exactly), and we need to skip when compile is turned on. 

If we don't have this flag, we would face compile error 
```
    File "/home/jessicazhong/torchtitan/torchtitan/models/common/decoder.py", line 151, in forward
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/jessicazhong/pytorch/torch/_dynamo/eval_frame.py", line 468, in __call__
      return super().__call__(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/jessicazhong/pytorch/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
    File "/home/jessicazhong/pytorch/torch/nn/modules/module.py", line 1884, in _call_impl
      return inner()
             ^^^^^^^
    File "/home/jessicazhong/pytorch/torch/nn/modules/module.py", line 1832, in inner
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/jessicazhong/pytorch/torch/_dynamo/eval_frame.py", line 1046, in compile_wrapper
      raise e.with_traceback(None) from e.__cause__  # User compiler error
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  torch._dynamo.exc.Unsupported: HOP: Unsafe side effect
    Higher Order Operator: torch.utils.checkpoint.checkpoint
    Explanation: Mutating a variable from outside the scope of this HOP is not supported.
    Hint: If the HOP is activation checkpointing (torch.utils.checkpoint.checkpoint), this points to a side effect in forward method. Eager activation checkpointing replays that side-effect while recomputing the forward in the backward. If you are ok with side-effect not replayed in the backward, try setting `torch._dynamo.config.skip_fwd_side_effects_in_bwd_under_checkpoint = True`
  
    Developer debug context: Attempted to mutate FSDPManagedNNModuleVariable(ScaledDotProductAttentionWrapper)
```

fegin: If the  Python mutation is safe to ignore (most of the time), then it is okay to set this flag as in the backward we don't need to replay the mutation, we just need to replay the forward ops. It looks like the mutation is from ScaledDotProductAttentionWrapper, which inherits LocalMapWrapper. And LocalMapWrapper __call__ mutate `self._local_map_fns`. I think that's where the mutation. It is safe to ignore as long as we replay the correct forward ops.

https://github.com/pytorch/torchtitan/pull/2621 works without setting this the flag because that PR swaps the forward BEFORE forward is called.

cc., @xmfan, @anijain2305 can confirm.

acisseJZhong: Claude explains to me that the mutation comes from `self._local_map_fn`. 
> In torch.compile, the compiler traces the forward graph.
   It sees the self._local_map_fn = ... mutation as a Python
   side effect. During AC's backward recompute, compile
  needs to replay forward ops but cannot replay arbitrary
  Python attribute mutations. The
  skip_fwd_side_effects_in_bwd_under_checkpoint flag tells
  compile to skip these mutations during backward replay.

  > Why it's safe to skip: The mutation is a one-time lazy
  cache. After the first forward, _local_map_fn is already
  set. Skipping it during AC's backward recompute doesn't
  affect the computation — the cached function is still
  there and used correctly.

However I am wondering if https://github.com/pytorch/torchtitan/pull/2621 is simpler, whether we should adopt this approach instead of inheriting from `LocalMapWrapper`.

torchtitan/models/gpt_oss/parallelize.py
```diff
@@ -242,14 +242,6 @@ def apply_non_moe_tp(
             "attention.wq": ColwiseParallel(use_local_output=False),
             "attention.wk": ColwiseParallel(use_local_output=False),
             "attention.wv": ColwiseParallel(use_local_output=False),
-            "attention.inner_attention": PrepareModuleInputOutput(
```
acisseJZhong, line 245: deleting this as now LocalMapAttention.__call__ handles the conversion from DTensor to local tensor and wrapping back local tensor to DTensor after. 

I also ran the loss comparison script to verify losses match. 
```
python scripts/loss_compare.py . main --baseline-module=gpt_oss --bas
eline-config=gpt_oss_debugmodel --baseline-options="--compile.enable" 
--test-module=gpt_oss --test-config=gpt_oss_debugmodel --test-options="--compile.enable" --assert-equal --no-seed-checkpoint
```

tianyu-l:
Could you follow up on whether eager and compile (with aot_eager backend) produce bitwise identical results, **before and after this PR**, **with and without FlexAttention**, always with SAC on.

acisseJZhong:
verified that both before and after this PR, eager and compile (with aot_eager backend) produce bitwise identical results. 

Also before and after this PR,  `qwen3_debugmodel_flex` with CP=2 and aot_eager, both training work.

PR #2560, opened 2026-03-12
Yield per-document RoPE position ids from dataset

### LINKED ISSUE #2559: RoPE positions are never set
I am looking into how RoPE positions are handled. I noticed that the models support a `positions` field:

[https://github.com/pytorch/torchtitan/blob/5732118a65c4bd4c86ee752b0e20c4411064568f/torchtitan/models/qwen3/model.py#L57-L63](https://github.com/pytorch/torchtitan/blob/5732118a65c4bd4c86ee752b0e20c4411064568f/torchtitan/models/qwen3/model.py#L57-L63)

However, when examining the dataloader and dataset implementation, it appears that `positions` is always `None`.

If document packing is used, this could lead to misaligned RoPE embeddings across documents. Specifically, the first token of a packed document might not receive the first positional embedding, because the position index continues from the previous document in the packed sequence. As a result, the RoPE embedding applied to the first token of a document may not correspond to position 0 of that document.

Is this expected behavior, or should positions be reset per document when packing is enabled?

joecummings:
Fixes #2559

`HuggingFaceTextDataset` now tracks a `_position_buffer` alongside the existing `_token_buffer.` Each document's tokens get positions [0, 1, ..., doc_len-1], resetting at every document boundary. Positions are yielded as {"input": input, "positions": positions} and flow through the trainer's `extra_inputs` into `Decoder.forward(positions=...)` automatically.

Checkpoint state_dict/load_state_dict updated to persist the position buffer (BC via .get()).

#### Longer-term consideration

Right now there are two considerations for packed datasets: attention masks and position IDs. Attention masks are computed in the post_dataloading_process  and, in this PR, position IDs are built in the dataset. Constructing masks purely based on EOS token id is fragile, especially with post-training multi-turn sequences where models could co-opts that token for end of sequence versus end of document.

The right long-term approach for torchtitan is that datasets yield `seq_lens` metadata alongside tokens (rather than `position_ids` directely), and both positions and attention masks are derived from that single source of truth in post-processing. This would retire the EOS-based 	`get_document_mask_mod` path entirely and co-locate both computations in one place.

```
# In dataloader
def _iter_greedy_packed(self):
      for sample in self._get_data_iter():
          input_ids = self._tokenize(sample)
          self._pack_buffer_input.extend(input_ids)
          self._pack_seq_lens.append(len(input_ids))  # just track the length

# In post dataloading process
 if "seq_lens" in extra_inputs:
          seq_lens = extra_inputs.pop("seq_lens")
          extra_inputs["positions"] = positions_from_seq_lens(seq_lens)
          extra_kwargs["attention_masks"] = mask_from_seq_lens(seq_lens)
```

Doesn't change how Decoder works.

Resources: https://github.com/NVIDIA/NeMo/blob/v2.7.0/nemo/collections/llm/gpt/data/core.py, https://github.com/pytorch/torchtune/blob/d0f63bb33d00b8bd3905a010b71d8c6324c2e980/torchtune/datasets/_packed.py#L108-L143, 

Or could possible switch everything to `varlen` but that would ensure that FA4 is working and the additional lift when integrating with context parallel.

 #### Test plan

Unit tests pass

Also for fun, comparison between WITH position ids and WITHOUT. Definitely different in the loss, but not by a ton:
Screenshot 2026-03-12 at 5 13 11 PM

## Next steps
First is #2610: Switch llama3 and qwen3 model configs from sdpa/causal to flex/block_causal, Regenerate expected loss files (`tests/assets/losses/llama3_cuda.txt`, `llama3_rocm.txt`) with new defaults, Remove now-redundant config variants (`debugmodel_flex_attn`, `8B_flex`, `debugmodel_flex`)

Later stuff like:
- Validate loss convergence on C4 with block_causal + per-document positions vs baseline
  - Propagate per-document positions through context parallel (currently `prepare_context_parallel_input` overwrites positions with sequential `torch.arange`)
  - Consider deriving attention masks from position IDs rather than EOS scanning, removing the tokenizer dependency from `get_attention_masks`

### DIFF
--- a/torchtitan/models/common/rope.py
+++ b/torchtitan/models/common/rope.py
@@ -9,7 +9,7 @@ from dataclasses import dataclass
 from typing import Literal
 
 import torch
-from torch.distributed.tensor import DTensor
+from torch.distributed.tensor import DTensor, Replicate, Shard
 
 from torchtitan.protocols.module import Module
 
@@ -290,31 +290,41 @@ def _rotate_half(x: torch.Tensor) -> torch.Tensor:
     return torch.cat((-x2, x1), dim=-1)
 
 
-def _maybe_to_dtensor(
-    tensor: torch.Tensor | None,
-    like: torch.Tensor,
+def _maybe_wrap_positions(
+    positions: torch.Tensor | None,
+    x: torch.Tensor,
 ) -> torch.Tensor | None:
-    """Convert tensor to a DTensor matching like's mesh and placements.
+    """Wrap positions as a DTensor deriving mesh and placements from x (xq/xk).
+
+    TODO: In a full DTensor rewrite, positions should be made a DTensor
+    in/right after dataloading, together with inputs and labels.
 
     When TP uses use_local_output=False (DeepSeek V3, Qwen3, GPT-OSS),
-    like (xq/xk) is a DTensor but tensor (positions) is a plain tensor.
-    The downstream torch.gather requires both operands to be the same type.
-    Positions should share placements with the input activations rather than
-    freqs_cis, because both are per-token and would be sharded the same way
-    under CP.
+    x is a DTensor but positions is a plain tensor. The downstream
+    torch.gather requires both operands to be the same type.
+
+    Positions (bsz, seqlen) has fewer dimensions than x (bsz, seqlen,
+    n_heads, head_dim), so we only preserve Shard placements for shared
+    dimensions. Shard dims beyond positions' rank (e.g. Shard(2) for TP
+    on heads) become Replicate.
     """
     if (
-        tensor is not None
-        and isinstance(like, DTensor)
-        and not isinstance(tensor, DTensor)
+        positions is not None
+        and isinstance(x, DTensor)
+        and not isinstance(positions, DTensor)
     ):
-        tensor = DTensor.from_local(
-            tensor,
-            like.device_mesh,
-            like.placements,
+        ndim = positions.ndim
+        placements = tuple(
+            p if not isinstance(p, Shard) or p.dim < ndim else Replicate()
+            for p in x.placements
+        )
+        positions = DTensor.from_local(
+            positions,
+            x.device_mesh,
+            placements,
             run_check=False,
         )
-    return tensor
+    return positions
 
 
 # TODO: consolidate apply_rotary_emb_complex and apply_rotary_emb_single_complex
@@ -332,7 +342,7 @@ def apply_rotary_emb_complex(
         freqs_cis: (max_seqlen, head_dim // 2) complex
         positions: optional position indices
     """
-    positions = _maybe_to_dtensor(positions, xq)
+    positions = _maybe_wrap_positions(positions, xq)
     xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
     xk_ = torch.view_as_complex(xk.float().reshape(*xk.shape[:-1], -1, 2))
     freqs_cis = _reshape_for_broadcast_complex(freqs_cis, xq_, positions)
@@ -353,7 +363,7 @@ def apply_rotary_emb_single_complex(
         freqs_cis: (max_seqlen, head_dim // 2) complex
         positions: optional position indices
     """
-    positions = _maybe_to_dtensor(positions, x)
+    positions = _maybe_wrap_positions(positions, x)
     dtype = x.dtype
     x = torch.view_as_complex(x.float().view(*x.shape[:-1], -1, 2))
     freqs_cis = _reshape_for_broadcast_complex(freqs_cis, x, positions)
@@ -375,7 +385,7 @@ def apply_rotary_emb_cos_sin(
         rope_cache: (max_seqlen, head_dim * 2) with cos and sin concatenated
         positions: optional position indices
     """
-    positions = _maybe_to_dtensor(positions, xq)
+    positions = _maybe_wrap_positions(positions, xq)
     head_dim = xq.shape[-1]
     rope_cache = _reshape_for_broadcast_cos_sin(rope_cache, xq, positions)
     cos = rope_cache[..., :head_dim].to(device=xq.device)


joecummings:
cc @tianyu-l @francesco-bertolotti : I did the fix that was discussed in #2559, but the "longer term fix" is also pretty simple. I might suggest we just do that in this PR, unless you have objections b/c that would technically be changing the behavior of the attention mask construction. Could be a follow up.

tianyu-l:
@joecummings 
The long term fix sounds reasonable. It can also replace varlen metadata creation https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L322

> co-locate both computations in one place

So you are suggesting putting it in dataloading. But then for more complicated, model-specific mask generation (e.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama4/model.py#L209), there still need to be this post_dataloading_processing https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L608, right?

rakkit:
> Also for fun, comparison between WITH position ids and WITHOUT. Definitely different in the loss, but not by a ton:


i think this is expected for rope

joecummings:
After realizing that the default testing is also incorrect b/c they rely on spda backend even with packing, opting to just get this fix in first and consider a proper refactor after.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -119,15 +120,21 @@ def __iter__(self):
                     sample_text, add_bos=True, add_eos=True
                 )
                 self._token_buffer.extend(sample_tokens)
+                self._position_buffer.extend(
+                    i % self.seq_len for i in range(len(sample_tokens))
```
felipemello1, line 130: why 'i % self.seq_len'?

if `seq_len` is 5, and `sample_tokens` is 10, you would expect: [1,2,3,4,5,1,2,3,4,5] for the same sentence?

Should we instead error if len(sample_tokens) > self.seq_elen?

torchtitan/models/llama3/__init__.py
```diff
@@ -44,7 +44,8 @@
             ),
             attention=GQAttention.Config(
                 n_heads=16,
-                attn_backend="sdpa",
+                attn_backend="flex",
+                attn_mask_type="block_causal",
```
felipemello1: q: should we use varlen as default instead? flex is more flexible, i just dont understand the backend differences too well

joecummings:
> this looks reasonable to me, i just wonder why 'flex' and not 'varlen'. I don't know the backends well enough and implications. Any thoughts?

Moved this to a different PR since I just want to get this fix in first before tackling the larger problem with packing in torchtitan. Let's have the discussion there when ready

torchtitan/trainer.py
```diff
@@ -591,9 +596,16 @@ def post_dataloading_process(
         # extra_kwargs are.
         extra_kwargs: dict[str, Any] = {}
 
-        # TODO: improve the logic on obtaining attention masks
+        # TODO: remove this guard once RoPE handles DTensor+positions.
+        # The positions!=None path in RoPE uses torch.gather which fails
+        # with DTensor+FSDP. For now, only pass positions through when
+        # using flex/varlen + block_causal (where it's needed and works).
```
tianyu-l: What error did you get when `causal` + `positions!=None` happen?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -153,7 +159,10 @@ def load_state_dict(self, state_dict):
             self._data.load_state_dict(state_dict["data"])
 
     def state_dict(self):
-        _state_dict: dict[str, Any] = {"token_buffer": self._token_buffer}
+        _state_dict: dict[str, Any] = {
+            "token_buffer": self._token_buffer,
+            "position_buffer": self._position_buffer,
```
tianyu-l, line 180: I think this will break existing checkpoints with dataloader state dict saved. But it's fine to break.

torchtitan/trainer.py
joecummings: It was an OOB error due to a reshape - I updated the comment to be more accurate.

torchtitan/hf_datasets/text_datasets.py
joecummings: IIUC it shouldn't. The issue would be trying to load a key that's not there and we allow that in the `load_state_dict` with a default from `.get()`. The behaviour would be different but the checkpoint should load.

tianyu-l: How about adding a `logger.warning` if the key is missing in state dic? It could be hard for users to debug, given that we change the behavior without failing their program.

torchtitan/models/common/rope.py
```diff
@@ -304,6 +334,7 @@ def apply_rotary_emb_complex(
         freqs_cis: (max_seqlen, head_dim // 2) complex
         positions: optional position indices
     """
+    positions = _maybe_wrap_positions(positions, freqs_cis)
```
joecummings: Public API boundary seemed like the proper place to do this wrapping, but lmk if you'd prefer this somewhere else @tianyu-l

torchtitan/models/common/rope.py
```diff
@@ -289,6 +290,35 @@ def _rotate_half(x: torch.Tensor) -> torch.Tensor:
...
+        and isinstance(freqs_cis, DTensor)
+        and not isinstance(positions, DTensor)
+    ):
+        assert all(
+            isinstance(p, Replicate) for p in freqs_cis.placements
+        ), f"Expected Replicate placements on freqs_cis, got {freqs_cis.placements}"
+        positions = DTensor.from_local(
+            positions,
+            freqs_cis.device_mesh,
+            freqs_cis.placements,
```
tianyu-l: Usually `positions` should have the same placements as `x`, rather than `freqs_cis`. We are not wrapping tensors on CP dimension, but if we do, `x` and `positions` will be sharded on sequence dim on CP, whereas `freqs_cis` will be Replicate on CP.

torchtitan/models/common/rope.py
```diff
@@ -289,6 +290,35 @@ def _rotate_half(x: torch.Tensor) -> torch.Tensor:
     return torch.cat((-x2, x1), dim=-1)
 
 
+def _maybe_wrap_positions(
```
tianyu-l, line 293: maybe 
```suggestion
def _maybe_to_dtensor(
```

torchtitan/models/common/rope.py
joecummings: I will change to borrow placements from `x`

tianyu-l: sorry, let me clarify:
`x` has more dimensions than `positions`. The placement should match on the dimensions they share (batch, sequence). But if `x` is sharded on extra dimensions (e.g. in TP `x` would be sharded on `head_dim` namely with placement `Shard(3)`) then the corresponding placement on positions should be `Replicate`.

joecummings: I see - in that case, I'll leave the function name more specific since it is truly tied to both x's state and whether or not we have positions

torchtitan/models/common/rope.py
```diff
@@ -289,6 +290,40 @@ def _rotate_half(x: torch.Tensor) -> torch.Tensor:
...
```
tianyu-l, line 293: sg, please leave a TODO: in full DTensor rewrite, we should make `positions` a DTensor in/right after dataloading, together with inputs and labels.

cc @fegin

PR #2561, opened 2026-03-12
[graph_trainer] Trace forward loss backward step with make_fx

yiming0416:
This PR adds a new `aot_fx_trace` compilation mode to the graph_trainer experiment that uses `make_fx_tracer` to trace the entire forward pass, loss computation, and backward pass as a single unified graph.

It introduces `FwdBwdStepModule` that composes `model(inputs)` → `loss_fn(pred, labels)` → `autograd.grad(loss, params)` into a single traceable forward. Overrides `forward_backward_step` in `GraphTrainer` to use this path when `compile.mode == "aot_fx_trace"`, with lazy tracing on the first step.

### DIFF
--- a/torchtitan/experiments/graph_trainer/compile.py
+++ b/torchtitan/experiments/graph_trainer/compile.py
@@ -317,5 +317,14 @@ def apply_compile(
             fsdp_reshard_after_forward,
             joint_passes=[],
         )
+    elif mode == "aot_fx_trace":
+        # aot_fx_trace traces fwd+loss+bwd together inside forward_backward_step,
+        # so no model-level wrapping is needed here.
+        logger.info(
+            "aot_fx_trace compile mode: graph capture will happen at training time"
+        )
+        return model
     else:
-        raise ValueError(f"Unknown compile mode: {mode}. Must be 'jit' or 'aot'.")
+        raise ValueError(
+            f"Unknown compile mode: {mode}. Must be 'jit', 'aot', or 'aot_fx_trace'."
+        )
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -262,6 +262,10 @@ def trace_module(
     graph is a pure function.  Tensor subclasses (e.g. DTensor) are recursively
     unwrapped into plain tensors for tracing, and the layouts needed to rewrap
     them are recorded in the returned :class:`TracedResult`.
+
+    Args:
+        mod: The module to trace.
+        args: The user arguments to trace with.
     """
     named_parameters = dict(mod.named_parameters(remove_duplicate=False))
     named_buffers = dict(mod.named_buffers(remove_duplicate=False))
--- a/torchtitan/experiments/graph_trainer/trainer.py
+++ b/torchtitan/experiments/graph_trainer/trainer.py
@@ -5,10 +5,38 @@
 # LICENSE file in the root directory of this source tree.
 
 from dataclasses import dataclass, field
+from typing import Any
+
+import torch
+import torch.nn as nn
 
 from torchtitan.experiments.graph_trainer.configs import GraphTrainerCompileConfig
 from torchtitan.experiments.graph_trainer.cudagraph import cudagraph_teardown
+from torchtitan.experiments.graph_trainer.make_fx_tracer import (
+    run_traced_module,
+    trace_module,
+    TracedResult,
+)
 from torchtitan.trainer import Trainer
+
+
+class FwdBwdStepModule(nn.Module):
+    """Wraps model + loss_fn + autograd.grad into a single traceable forward.
+
+    This allows make_fx to trace through the entire fwd+loss+bwd as one graph.
+    """
+
+    def __init__(self, model, loss_fn):
+        super().__init__()
+        self.model = model
+        self.loss_fn = loss_fn
+
+    def forward(self, inputs, labels, global_valid_tokens, extra_inputs, extra_kwargs):
+        pred = self.model(inputs, **extra_inputs, **extra_kwargs)
+        loss = self.loss_fn(pred, labels) / global_valid_tokens
+        params = [p for p in self.model.parameters() if p.requires_grad]
+        grads = torch.autograd.grad(loss, params)
+        return [loss] + list(grads)
 
 
 class GraphTrainer(Trainer):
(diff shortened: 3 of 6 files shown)


torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
             default_factory=GraphTrainerCompileConfig
         )
 
+    def __init__(self, config):
+        super().__init__(config)
+        # Lazy state for make_fx mode
+        self._train_step_module: TrainStepModule | None = None
+        self._traced_step: TracedResult | None = None
+
+    def forward_backward_step(
```
SherlockNoMad, line 61: Does this trainer work for other models?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
+            loss_fn = self.loss_fn
+            orig = getattr(loss_fn, "_torchdynamo_orig_callable", None)
+            if orig is not None:
+                loss_fn = orig
+            self._train_step_module = TrainStepModule(model, loss_fn)
+
+            self._traced_step = trace_module(
+                self._train_step_module,
+                (inputs, labels, global_valid_tokens, extra_inputs, extra_kwargs),
+                train_context=self.train_context,
```
SherlockNoMad: why is this needed?

torchtitan/experiments/graph_trainer/compile.py
```diff
@@ -155,5 +155,12 @@ def apply_compile(
             fsdp_reshard_after_forward,
             joint_passes=[],
         )
+    elif mode == "make_fx":
```
SherlockNoMad: can you think of better name?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -160,13 +161,21 @@ def _remove_cpu_shadow_chains(gm: torch.fx.GraphModule) -> None:
 def trace_module(
     mod: nn.Module,
     args: tuple,
+    *,
+    train_context: contextlib.AbstractContextManager | None = None,
```
SherlockNoMad: I would prefer we don't pass this into trace_module. 

1. can we do it at callsite? 
2. eventually we should upstream trace_module to pytorch, I would like to keep this API simply and general

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
 from torchtitan.experiments.graph_trainer.cudagraph import cudagraph_teardown
+from torchtitan.experiments.graph_trainer.make_fx_tracer import (
+    run_traced_module,
+    trace_module,
+    TracedResult,
+)
 from torchtitan.trainer import Trainer
 
 
+class TrainStepModule(nn.Module):
```
SherlockNoMad: train_step usually mean fw+loss+bw+optimizer.step()

let's call this FwdBwdStepModule

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
+
+    This allows make_fx to trace through the entire fwd+loss+bwd as one graph.
+    """
+
+    def __init__(self, model, loss_fn):
+        super().__init__()
+        self.model = model
+        self.loss_fn = loss_fn
+
+    def forward(self, inputs, labels, global_valid_tokens, extra_inputs, extra_kwargs):
```
SherlockNoMad, line 34: this forward func's signature is not general.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
+        pred = self.model(inputs, **extra_inputs, **extra_kwargs)
+        loss = self.loss_fn(pred, labels) / global_valid_tokens
```
SherlockNoMad, line 36: Does this loss pattern apply to all the models?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
+        self._traced_step: TracedResult | None = None
+
+    def forward_backward_step(
+        self,
+        *,
+        input_dict: dict[str, torch.Tensor],
+        labels: torch.Tensor,
+        global_valid_tokens: torch.Tensor,
+    ) -> torch.Tensor:
+        if self.config.compile.mode != "make_fx":
```
SherlockNoMad: it should raise here. 
non-make-fx code shouldn't reach here

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
+                loss_fn = orig
+            self._train_step_module = TrainStepModule(model, loss_fn)
+
+            self._traced_step = trace_module(
+                self._train_step_module,
+                (inputs, labels, global_valid_tokens, extra_inputs, extra_kwargs),
+                train_context=self.train_context,
+            )
+
+        params_and_buffers = {
```
SherlockNoMad, line 112: is params_and_buffers available from tracedresult, I feel it should? @ydwu4

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,133 @@
...
+            **dict(self._train_step_module.named_parameters(remove_duplicate=False)),
+            **dict(self._train_step_module.named_buffers(remove_duplicate=False)),
+        }
+        wrapped = run_traced_module(
```
SherlockNoMad: bad name

SherlockNoMad:
@claude review this. 

Think about generality, this path should be applicable for all models in the repo. 
Look for opportunity for better API and abstraction.

torchtitan/experiments/graph_trainer/trainer.py
ydwu4: Right now the separation of parameters/buffers and execution (the gm and metadata) feels OK. But if we want to follow something like exported program, packaging them together in traced result (e.g. one single object that gets serded) seems reasonable. So whatever that's easy to understand.

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: In `graph_trainer` experiment, we'll always use `GraphTrainer` which inherits the original `Trainer`. 

The `forward_backward_step` is overridden so that when we are in non-make-fx mode, we just fallback to `super().forward_backward_step`

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: We need `torch.distributed.tensor.parallel.loss_parallel()` context manager, which enables TP-aware loss parallel during tracing.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
yiming0416: yes, we can do it at callsite. Removed the `train_context` argument in `trace_module`

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: Yes. It follows the same pattern in `torchtitan/trainer.py`, which is used for all models in torchtitan

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: Moved the `self.train_context` out of `trace_module`

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: Well, all LLM models follow this pattern. The only exception is FLUX model which has its own customized trainer.

torchtitan/experiments/graph_trainer/compile.py
yiming0416: I think we might still want to mention `aot` in this mode.

What about `aot_step` or `aot_fx_trace`

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: It matches the caller-side contract used by `Trainer.forward_backward_step` (in `torchtitan/trainer.py`). 

All Decoder-based models (Llama3, Llama4, Qwen3, DeepSeekV3, GptOss) and the VLM model accept this pattern.

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: It works on existing llama3 and dsv3 models in `graph_trainer`. 
Added integration tests and numerics tests

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -5,19 +5,131 @@
...
+        with self.train_context(), self.maybe_enable_amp:
+            outputs = run_traced_module(
+                self._traced_step,
+                params_and_buffers,
+                (inputs, labels, global_valid_tokens, extra_inputs, extra_kwargs),
+            )
+        loss = outputs[0]
+        grads = outputs[1:]
+
+        for param, grad in zip(params, grads):
```
ydwu4, line 125: It's not always the case that we can directly += the grad i think? There is an optimizer step.  

Since we're not tracing into optimizer at the moment, the other options is to return the loss and grads and do the gradient accumulation outside. Don't have to do in this pr though

tugsbayasgalan: I think this is already doing outside the non-strict region?

PR #2565, opened 2026-03-13
[Module] Convert remaining nn.Module classes to Module protocol

fegin:
* __->__ #2565

**Why**
It is hard to do the remaining changes, 1) state initialization and 2) sharding spec/local map spec change with some modules being nn.Module. The logic will need several fall back plan.

This PR introduce a minimal change to conver all the nn.Module to Module.

1. Don't define Config for most of the Module.
2. Make init_weight() a materialized method with empty logic.

While 2 makes it not possible to detect if a Module accidentally forget to define, it is fine beause we are going to change init_weight() in the next PR.

**Summary**

Convert 35 classes from plain nn.Module to the torchtitan Module protocol
across core models (19) and experiments (16).

Key design decisions:

1. Module without Config for non-configurable classes: If all constructor
   args come from the parent module (its Config or runtime), the class
   inherits Module without defining Config -- just a direct constructor.
   Config + build() is reserved for classes with independently
   user-configurable fields.

2. init_weights is a default no-op in base Module: Changed from
   abstractmethod + raise NotImplementedError to a default pass
   implementation. Subclasses with learnable parameters override it;
   all others inherit the no-op. This eliminates boilerplate empty
   init_weights methods.

3. ModuleContainer for namespace grouping: Added ModuleContainer(Module)
   in protocols/module.py to replace bare nn.Module() instances used as
   attribute namespace containers (e.g., self.mid = ModuleContainer()
   in Flux autoencoder).

4. Container types not converted: nn.ModuleDict/nn.ModuleList subclasses
   (e.g., SliceableModuleDict) are left as-is. Diamond inheritance with
   these container types adds complexity for no benefit.

Core models: GroupedExperts, TokenChoiceTopKRouter, TokenReorderer,
GptOssGroupedExperts, VarlenAttentionWrapper, FlexAttentionWrapper,
ScaledDotProductAttentionWrapper, QKNorm, SelfAttention, Modulation,
AttnBlock, ResnetBlock, Downsample, Upsample, Encoder, Decoder,
DiagonalGaussian, AutoEncoder, FluxEmbedder.

Experiments: VLM siglip2 (5), Projector, RL/vLLM attention wrappers (3),
RL/vLLM Qwen3 components (4), graph trainer (2), vLLM model wrapper.

### DIFF
--- a/torchtitan/experiments/graph_trainer/graph_utils.py
+++ b/torchtitan/experiments/graph_trainer/graph_utils.py
@@ -231,7 +231,9 @@ class CompiledModule(Module):
     def init_weights(self, **kwargs) -> None:
         # Explicitly delegate to inner model. Without this override,
         # Module.init_weights (a no-op) would be found via MRO before
-        # __getattr__ is triggered, silently skipping weight initialization.
+        # the overwritten __getattr__ is triggered, silently skipping
+        # weight initialization.
+        # This is similar to state_dict, load_state_dict, ...
         self.inner.init_weights(**kwargs)
 
     def state_dict(self, *args, **kwargs) -> Any:


torchtitan/models/flux/model/autoencoder.py
```diff
@@ -164,7 +166,7 @@ def __init__(
             for _ in range(self.num_res_blocks):
                 block.append(ResnetBlock(in_channels=block_in, out_channels=block_out))
                 block_in = block_out
-            down = nn.Module()
+            down = ModuleContainer()
```
tianyu-l: It seems ModuleContainer is only used in flux autoencoder. My feeling is that the construction is not proper; since it has clear structure, we should just define an Module with the known fields. cc @wwwjn 

If you don't want to do it in this PR and have to use `ModuleContainer`, please define it within this file, not in the the general protocols.py.

fegin: Sounds good, I'll move it to Flux. I'm a little worried about major rewrite of Flux model as somehow my devegpu cannot run Flux (packages issue, which both Claude and I fail to fix the issue). So it is not something I want to tackle it as this moment.

PR #2568, opened 2026-03-13
[RL] Changes to enable compilation for trainer

### LINKED ISSUE #2508: [RL][Feature Request] Tun on torch.compile + cudagraphs for trainer definition
With Generator sped up by vllm's support_torch_compile, the new bottleneck is trainer.

Let's enable `torch.compile` and `cudagraph` there to get similar speedups

Lucaskabela:
## Summary

In this PR, we enable naive, JIT style torch.compile for the RL policy trainer.  This is the first step towards speeding up the trainer model. Changes are:

1. Wiring through compilation config:
  - Added TrainerCompileConfig dataclass with enable (bool) and backend (str, default "eager") fields
  - Added compile field to PolicyTrainer.Config with compile and aot_eager backend
  - Added apply_compile() method that calls .compile(backend=..., fullgraph=True, dynamic=True) on each transformer
   layer -> This is crticial, as `torch.compile()` results in weight name change which breaks the weight transfer and dynamism needed for seqlen

  2. config_registry.py — Enable compile by default in configs
  - Both rl_grpo_qwen3_0_6b and rl_grpo_qwen3_debug configs now set
  compile=TrainerCompileConfig(enable=True, backend="aot_eager")

  3. vllm_compat_attention.py — Deleted to consolidate on new attention backend
 
## Test Plan
```bash
```

Results in the same losses w/ and w/o compile, but different on main due to changing backend - the timing is now like: 
### Main
```bash
[actor=<root>] Step  9 | Loss: +0.0041 | Reward: +0.450 | Correct: 29/40 | Avg tokens: 100 | Logprob diff: mean=-1.1657e-01, max=1.1326e+01 | Time: 21.1s
[actor=<root>]   Timing: generate=2.98s  grade=0.01s  advantage=0.0003s  train=15.02s  sync=3.14s (push=0.13s)
[actor=<root>] --------------------------------------------------------------------------------
[actor=<root>] Training timing (10 steps in 219.2s, avg 21.92s/step):
[actor=<root>]   generate: 25.1s total, 2.51s avg, 11%
[actor=<root>]   grade:    0.1s total, 0.01s avg, 0%
[actor=<root>]   train:    155.9s total, 15.59s avg, 71%
[actor=<root>]   sync:     38.1s total, 3.81s avg, 17%
[actor=<root>] --------------------------------------------------------------------------------
[actor=<root>] Eval: Accuracy=55% (11/20) Format=100% (20/20)
[actor=<root>] Post-training: Accuracy=55% (11/20) Format=100% (20/20)
```
### Changes
```bash
[actor=<root>] Step  9 | Loss: -0.0072 | Reward: +0.300 | Correct: 26/40 | Avg tokens: 100 | Logprob diff: mean=-1.0181e-01, max=1.2256e+01 | Time: 12.9s
[actor=<root>]   Timing: generate=2.53s  grade=0.01s  advantage=0.0002s  train=7.12s  sync=3.24s (push=0.12s)
[actor=<root>] --------------------------------------------------------------------------------
[actor=<root>] Training timing (10 steps in 170.0s, avg 17.00s/step):
[actor=<root>]   generate: 24.0s total, 2.40s avg, 14%
[actor=<root>]   grade:    0.1s total, 0.01s avg, 0%
[actor=<root>]   train:    113.3s total, 11.33s avg, 67%
[actor=<root>]   sync:     32.7s total, 3.27s avg, 19%
[actor=<root>] --------------------------------------------------------------------------------
[actor=<root>] Eval: Accuracy=60% (12/20) Format=90% (18/20)
[actor=<root>] Post-training: Accuracy=60% (12/20) Format=90% (18/20)=
(vllm) [lucaskabela@devgpu007.eag6 ~/torchtitan (lucaskabela/enable_trainer_compile_03_10)]$ pytho
```

So we save ~40s of runtime e2e via compilation in this manner without affecting our logits/accuracy


torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -38,6 +38,16 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass(kw_only=True, slots=True)
+class TrainerCompileConfig:
```
tianyu-l: can we just use the existing one in configs.py?

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -109,6 +120,8 @@ def __init__(
             model_spec, config, device_type, batch_invariant_mode, hf_assets_path
         )
         model.train()
+        if config.compile.enable:
+            model = self._compile_model(model, config.compile.backend)
```
tianyu-l: do it at the same place where we do parallelism, i.e. model wrapper constructor

torchtitan/experiments/rl/models/vllm_compat_attention.py
tianyu-l, line 1: if we go with pytorch varlen, do we still need to worry about this file? cc @wwwjn

torchtitan/experiments/rl/unified/actors/trainer.py
Lucaskabela: Ah I wasn't aware of this - let me refactor to use this config

torchtitan/experiments/rl/unified/actors/trainer.py
```diff
@@ -221,6 +222,17 @@ def _build_model(
         # Load initial weights from HF
         self._load_initial_hf_weights(model, hf_assets_path)
 
+        # Apply per-layer torch.compile after parallelism and weight loading
```
tianyu-l: > do it at the same place where we do parallelism, i.e. model wrapper constructor

sorry ignore this comment, I was thinking about generator.

For trainer, let's follow trainer.py and call https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L394
Compile will be applied inside.

torchtitan/experiments/rl/models/vllm_compat_attention.py
Lucaskabela: This is post rebase - yes we do need these changes for the following reasons:

1) Moving `torch.autograd.Function` out of the inner context - this is so compile can trace it
2) `_flash_attn_varlen_fwd` custom op - this is so torch can trace it; without custom op, we don't know how shape will propogate through this

liangel-02: if ur going with pytorch varlen i think u can directly call [varlen_attn](https://github.com/pytorch/pytorch/blob/main/torch/nn/attention/varlen.py) (which is already a custom op) pytorch's varlen calls the upstream flash attention impl instead of vllm's flash attention

Lucaskabela: Sorry I misunderstood this question (I thought this was related to the deletion of vllm_attention directory).

We will need to move the autograd.Function out but yes we won't need the custom op in that case

torchtitan/experiments/rl/models/vllm_compat_attention.py
```diff
@@ -7,26 +7,196 @@
...
-from collections.abc import Callable
 
 import torch
 from torch.distributed._tensor import DTensor
 
 from torchtitan.protocols.module import Module
-from vllm.v1.attention.backends.fa_utils import flash_attn_varlen_func
+
+
+@torch.library.custom_op("rl::flash_attn_varlen_fwd", mutates_args=())
```
tianyu-l: let's not merge this, let use pytorch varlen attention in https://github.com/pytorch/torchtitan/pull/2364

cc @zhxchen17 to unblock

torchtitan/experiments/rl/models/parallelize.py
tianyu-l, line 34: @wwwjn could you work with @sanketpurandare to land PP + DTensor so that we can remove this ad hoc function?

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -60,9 +64,31 @@ def parallelize_qwen3(
...
+        compile_config is not None
+        and compile_config.enable
+        and "model" in compile_config.components
+    ):
+        apply_compile(model, compile_config)
+
     return model
 
 
+def apply_compile(model: nn.Module, compile_config: CompileConfig):
```
tianyu-l: we switched to a central impl in https://github.com/pytorch/torchtitan/pull/2615, but I understand that this is doing `model.compile()` not `torch.compile(model)`

@fegin can we universally switch to model.compile style (and it should still work with DCP)?

torchtitan/experiments/rl/models/parallelize.py
wwwjn: Sure, let me follow up with @sanketpurandare

Lucaskabela:
Converting to draft to rebase on the attention changes

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -206,18 +204,11 @@ def _build_model(
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
 
-        # Replace attention with vLLM compatible attention for RL training.
```
Lucaskabela, line 209: If we need to keep this definition around then we will need the custom op changes there to support the compilation

Additionally, deleting keeps us grounded on one source of truth to help avoid confusions so I think it best to just remove this code if we are switching to varlen

torchtitan/experiments/rl/models/parallelize.py
Lucaskabela: Thanks! Do we have an estimated timeline on this? In the meanwhile I don't think this should be blocking here but it is good to be aware of to coordinate :)

torchtitan/experiments/rl/models/parallelize.py
Lucaskabela: Similar comment here - we can switch to that definition once ready/as a follow up :) shouldn't be blocking (imo)

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -206,18 +204,11 @@ def _build_model(
...
-        # NOTE: Long-term this will be replaced by pytorch attention
-        # supporting paged attention / kv cache.
-        if batch_invariant_mode:
-            replace_with_vllm_compatible_flash_attention(
```
acisseJZhong, line 213: you might want to replace this with `replace_with_vllm_attention`. I also edited in https://github.com/pytorch/torchtitan/pull/2638/changes

torchtitan/experiments/rl/models/parallelize.py
acisseJZhong: can you put this function into the centralized compile.py introduced in https://github.com/pytorch/torchtitan/pull/2557/changes#diff-b369dd9c89f872a2eba30aff3ea51cabdcdcc84c2dfc2638080acaceec2f2954R1?

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -60,9 +64,36 @@ def parallelize_qwen3(
...
+    """Apply torch.compile JIT style to each TransformerBlock.
+
+    Follows the same pattern as torchtitan/models/llama3/parallelize.py.
+    """
+    # NOTE: we MUST use `.compile()` instead of `torch.compile()` here, because
+    # compatibility with weight naming between this model and vLLM definition.
```
acisseJZhong: can you put more detail on why torch.compile() wouldn't work here? 

also, can we all switch to model.compile() in this case?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -119,6 +121,7 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
                 tensor_parallel_degree=1,
                 data_parallel_replicate_degree=1,
             ),
+            compile=CompileConfig(enable=True, backend="aot_eager"),
```
acisseJZhong, line 124: nit: can we also introduce the `enable` field to `GeneratorCompileConfig` just for symmetry? Or is that we by default turn on generator compile, but no default on for trainer?

acisseJZhong (commented):
> Results in the same losses as on main - the timing is now like:

loss seems to differ based on your paste?

torchtitan/experiments/rl/models/parallelize.py
acisseJZhong: also wondering what is the timeline for the PP full dtensor PR? would also like to have that to unblock Verl integration. cc @sanketpurandare

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -31,8 +31,12 @@ def compute_token_log_probs(
         full_sequence, dtype=torch.long, device=device
     ).unsqueeze(0)
 
-    # Forward pass — trainer uses is_position_id=False so positions=None is fine
-    logits = model(full_tensor, attention_masks=None)
+    # Explicit positions avoid dynamic rope_cache[0:seqlen] slice in RoPE,
+    # which breaks torch.compile with symbolic shapes.
+    seq_len = full_tensor.shape[1]
+    positions = torch.arange(seq_len, device=device).unsqueeze(0)
```
tianyu-l, line 39: A separate question: It seems the full sequence here is always per episode, with batch size 1, and we have a for loop over all the episodes. Wouldn't this be very inefficient?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/trainer.py#L296

cc @joecummings @felipemello1

torchtitan/experiments/rl/models/attention.py
```diff
@@ -340,38 +337,3 @@ def replace_with_vllm_attention(model, tp_degree=1):
...
-    if not hasattr(model, "layers"):
-        raise AttributeError(
-            f"Model {type(model).__name__} must have .layers attribute"
-        )
-
-    for layer_name, layer in model.layers.items():
-        if not hasattr(layer, "attention"):
-            raise ValueError(f"Layer {layer_name} must have .attention attribute")
-
-        vllm_attn = VLLMCompatibleFlashAttention()
```
tianyu-l, line 370: can we also remove this class and the `vllm_compat_attention.py` file?

torchtitan/experiments/rl/models/parallelize.py
tianyu-l: It's OK to not change all pretrain compile in this PR. But please create an issue to track.

torchtitan/experiments/rl/actors/utils.py
tianyu-l: @Lucaskabela please create an issue to track

tianyu-l (approved):
please address remaining comments

torchtitan/experiments/rl/actors/utils.py
Lucaskabela: Yes absolutely - we can make a significant improvement by batching in the future but need to be careful about our logits

torchtitan/experiments/rl/models/attention.py
Lucaskabela: Already done :)

torchtitan/experiments/rl/config_registry.py
Lucaskabela: It is a bit more complicated than that unfortunately - the compile integration in generator is quite different than trainer, so there is not a simple "enable" flag - we need both a backend turned on and a cudagraph option.

There is an "enforce_eager" option that we can wire through to compile, but on the PR for enabling the GeneratorCompile I was discouraged from wiring this through as it adds complexity

Lucaskabela (commented):
Re logits changing: That is expected now, since this PR was rebased to change the backend (i.e removing the code in trainer).  There is no impact on the loss between compile=False and True, but between main and this PR there will be

torchtitan/experiments/rl/models/parallelize.py
Lucaskabela: .compile() is the general guidance due to subtle interactions with nn.Module - I will investigate this more though but IIRC this is a known issue with compile

acisseJZhong (approved):

torchtitan/experiments/rl/actors/utils.py
wwwjn: We are missing a "batcher" concept in the main RL loop

torchtitan/experiments/rl/models/parallelize.py
wwwjn: > Thanks! Do we have an estimated timeline on this? In the meanwhile I don't think this should be blocking here but it is good to be aware of to coordinate :)

DTensors for PP Stack is ready ([RFC](https://github.com/pytorch/pytorch/issues/172419)):[PR1](https://github.com/pytorch/pytorch/pull/177727), [PR2](https://github.com/pytorch/pytorch/pull/177728), [PR3](https://github.com/pytorch/pytorch/pull/177729)
I think the main blocker now is these PR needs review, and might not have a good ETA for this to be landed now. Removing this function here should be a separate concern.

PR #2570, opened 2026-03-13
[graph_trainer] Graph-based SAC with default policy

yiming0416:
- Replace hardcoded SAC save-ops list with a dynamic default policy that pulls compute-intensive ops from PyTorch's upstream get_default_op_list() partitioner.
- Propagate recompute tags to getitem and wait_tensor nodes so that tuple-unpacking nodes (e.g. from rmsnorm, sdpa) and async collective wait nodes inherit their parent's save/recompute decision, preventing them from being unnecessarily saved as graph outputs.
- Add integration tests for graph SAC with FSDP+TP on Llama3 and FSDP+TP+EP on DeepSeek V3.

Llama3-8B FSDP=4, TP=2
no AC: `memory: 49.65GiB(52.27%)  tps: 4,878  tflops: 282.50  mfu: 28.56%`
eager SAC: `memory: 28.54GiB(30.04%)  tps: 4,463  tflops: 258.44  mfu: 26.13%`
graph SAC: `memory: 28.09GiB(29.57%)  tps: 4,523  tflops: 261.95  mfu: 26.49%`

Deepseek-v3-16B FSDP=4, TP=2, EP=2, and local_batch_size=2 
no AC: `memory: 87.71GiB(92.33%)  tps: 2,530  tflops: 45.80  mfu: 4.63%`
eager SAC: doesn't work with full graph capture (due to AC HOP mutation issue #1935)
graph SAC: `memory: 52.03GiB(54.77%)  tps: 1,908  tflops: 34.55  mfu: 3.49%`

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -108,23 +108,43 @@ def validate_flex_attn_annotation_pass(
     return gm
 
 
-# Default set of ops whose outputs should be saved (not recomputed) during
-# activation checkpointing. These are compute-intensive or communication ops
-# where recomputation is expensive.
-DEFAULT_SAC_SAVE_OPS = {
-    torch.ops.aten.mm.default,
-    torch.ops.aten.linear.default,
-    torch.ops.aten._scaled_dot_product_efficient_attention.default,
-    torch.ops.aten._scaled_dot_product_flash_attention.default,
-    torch.ops.aten._scaled_dot_product_cudnn_attention.default,
-    torch.ops.aten._scaled_dot_product_attention_math.default,
-    torch.ops.aten._scaled_dot_product_fused_attention_overrideable.default,
-    torch.ops._c10d_functional.reduce_scatter_tensor.default,
-    torch.ops._c10d_functional.all_to_all_single.default,
-    torch.ops.aten.max.default,
-    torch._higher_order_ops.flex_attention,
-    torch._higher_order_ops.inductor_compiled_code,
-}
+def _get_default_sac_save_ops() -> set:
+    """Build the default set of ops whose outputs should be saved (not recomputed)
+    during activation checkpointing.
+
+    Compute-intensive ops are obtained dynamically from PyTorch's partitioner
+    (``get_default_op_list``) so the list stays in sync with upstream changes.
+    """
+    from torch._functorch.partitioners import get_default_op_list
+
+    # Compute-intensive ops from PyTorch's partitioner
+    compute_intensive_ops = {
+        op.default for op in get_default_op_list().compute_intensive_ops
+    }
+
+    # attention variants
+    scaled_dot_product_attention_ops = {
+        torch.ops.aten._scaled_dot_product_cudnn_attention.default,
+        torch.ops.aten._scaled_dot_product_attention_math.default,
+        torch.ops.aten._scaled_dot_product_fused_attention_overrideable.default,
+    }
+
+    higher_order_ops = {
+        torch._higher_order_ops.flex_attention,
+        torch._higher_order_ops.inductor_compiled_code,
+    }
+
+    communication_intensive_ops = {
+        torch.ops._c10d_functional.reduce_scatter_tensor.default,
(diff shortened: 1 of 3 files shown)


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -108,23 +108,43 @@ def validate_flex_attn_annotation_pass(
...
+    # attention variants
+    scaled_dot_product_attention_ops = {
+        torch.ops.aten._scaled_dot_product_cudnn_attention.default,
+        torch.ops.aten._scaled_dot_product_attention_math.default,
+        torch.ops.aten._scaled_dot_product_fused_attention_overrideable.default,
+    }
+
+    higher_order_ops = {
+        torch._higher_order_ops.flex_attention,
+        torch._higher_order_ops.inductor_compiled_code,
```
SherlockNoMad, line 134: In general, inductor compiled code shouldn't be here. 

It's here coz we are currently using regional inductor on flex_attn.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -108,23 +108,43 @@ def validate_flex_attn_annotation_pass(
...
+    during activation checkpointing.
+
+    Compute-intensive ops are obtained dynamically from PyTorch's partitioner
+    (``get_default_op_list``) so the list stays in sync with upstream changes.
+    """
+    from torch._functorch.partitioners import get_default_op_list
+
+    # Compute-intensive ops from PyTorch's partitioner
+    compute_intensive_ops = {
+        op.default for op in get_default_op_list().compute_intensive_ops
```
SherlockNoMad, line 122: FYI, you might wanna sync with https://github.com/pytorch/torchtitan/pull/2357

SherlockNoMad:
Does titan have an option to generate memory snapshot? 
e.g. somewhere in the log, it should print the memory snapshot folder
and we should have a one line command to visualize it like we we did for "zoomer llm ws <link>"

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -153,18 +173,36 @@ def apply_sac_pass(
...
+            # - wait_tensor: Semantically tied to its parent async collective
+            #   (e.g., reduce_scatter_tensor, all_gather_into_tensor) and must
+            #   share the same save/recompute decision.
+            parent = node.args[0]
+            if isinstance(parent, torch.fx.Node) and "recompute" in parent.meta:
+                node.meta["recompute"] = parent.meta["recompute"]
+                node.meta["ac_graph_id"] = parent.meta.get("ac_graph_id", 0)
             continue
 
         node.meta["ac_graph_id"] = 0
```
SherlockNoMad, line 208: we also need the incremental graph id thing. 
Every ac wrapped region should have its own graph id, this is needed for Tugsuu's pass to work correctly on joint graph

SherlockNoMad: This diff looks good to me, you can do it in the next diff.

torchtitan/experiments/graph_trainer/passes.py
yiming0416: I copied this list from eager AC's save op list.

It should be related to flex attention. (not regional inductor though). In torchtitan main, they have this use case where only the flex attention part is compiled, while other parts run in eager.

torchtitan/experiments/graph_trainer/passes.py
yiming0416: Good point. I will update graph_trainer to use utils from that PR when it's landed.

yiming0416:
> Does titan have an option to generate memory snapshot? e.g. somewhere in the log, it should print the memory snapshot folder and we should have a one line command to visualize it like we we did for "zoomer llm ws "

Yes, See https://github.com/pytorch/torchtitan/blob/main/docs/debugging.md#enable-memory-profiling

passing in `--profiling.enable_memory_snapshot` and the memory snapshot will be dumped to `./outputs/memory_snapshot/`. Then we can use https://docs.pytorch.org/memory_viz to visualize it.

PR #2571, opened 2026-03-13
[Module] Modularize MoE components 

acisseJZhong:
* __->__ #2571

# Motivation

  Continues the modularization effort from #2428 (Embedding), #2434 (RMSNorm), and
  #2527 (Linear). This PR makes MoE components (GroupedExperts,
  TokenChoiceTopKRouter, TokenReorderer) inherit from Module and use the
  Config/build() pattern, enabling per-component configurability and consistent
  weight initialization.

 # Summary

  - GroupedExperts inherits from Module with a nested Config dataclass. dim,
  hidden_dim, and num_experts use field(init=False) so they are set at build() time
   by the parent MoE.
  - TokenChoiceTopKRouter inherits from Module with a nested Config dataclass.
  Replaces the flat gate_bias: bool field with gate: Linear.Config, enabling full
  gate configuration (e.g., Linear.Config(bias=True)). dim and num_experts use
  field(init=False).
  - TokenReorderer inherits from Module with a no-op init_weights (stateless
  module). num_experts and top_k use field(init=False).
  - MoE.Config replaces flat router/expert fields (top_k, score_func, route_norm,
  route_scale, gate_bias, use_grouped_mm, num_expert_groups, num_limited_groups,
  _debug_force_load_balance) with nested experts: GroupedExperts.Config and router:
   TokenChoiceTopKRouter.Config.
  - Updates all model configs (DeepSeek V3, Qwen3, GPT-OSS, Llama4) to use nested
  router/expert configs.
  - Adds 34 unit tests covering Config/build(), init_weights, forward passes, and
  nested config propagation for all modularized components.

cc @fegin #2540 may needs to rebase on this.

### DIFF
--- a/torchtitan/models/utils.py
+++ b/torchtitan/models/utils.py
@@ -488,7 +488,7 @@ def get_moe_model_nparams_and_flops(
         nparams_sparse_active = (
             nparams_moe_router
             + nparams_shared_experts
-            + nparams_experts * moe_config.top_k // moe_config.num_experts
+            + nparams_experts * moe_config.router.top_k // moe_config.num_experts
         )
     else:
         nparams_sparse_active = 0


torchtitan/models/common/moe/moe.py
```diff
@@ -125,59 +139,71 @@ def forward(
...
+
+        dim: int = field(init=False)
+        num_experts: int = field(init=False)
+        num_expert_groups: int | None = None  # must be a divisor of num_experts
+        num_limited_groups: int | None = None
+        top_k: int = 1
+        score_func: Literal["softmax", "sigmoid"] = "sigmoid"
+        route_norm: bool = False
+        route_scale: float = 1.0
+        gate: Linear.Config = field(default_factory=Linear.Config)
```
tianyu-l: Maybe better if it doesn't come with a default, so that model config constructor have to specify.

torchtitan/models/common/moe/moe.py
```diff
@@ -125,59 +139,71 @@ def forward(
...
-        num_limited_groups: int | None,
-        top_k: int,
-        score_func: Literal["softmax", "sigmoid"],
-        route_norm: bool,
-        route_scale: float,
-        gate_bias: bool,
-        _debug_force_load_balance: bool = False,
-    ):
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
```
tianyu-l, line 164: makes sense. I wonder if we should put the following to router as well
- `tokens_per_expert` and `expert_bias` buffer to router module
- `load_balance_coeff` to router config

PR #2575, opened 2026-03-14
fix: prevent inductor from fusing away bf16→fp32 cast in RoPE

dean-mccoppin:
With compile.enable=true, Qwen3 produces different outputs from eager (max diff ~1.56 in bfloat16). Inductor traces the whole transformer block as one graph and legally eliminates the .to(dtype=xq.dtype) downcast between q_norm/k_norm and RoPE, keeping the multiply-add in fp32. Valid algebra, wrong dtype boundary relative to eager.

Fix borrowed from apply_rotary_emb_complex: upcast xq/xk to float32 before the multiply-add instead of downcasting cos/sin to match. The fp32 compute is now unconditional in the graph so Inductor has nothing to fuse away. Cast back with type_as at the end as before.

Fixes Qwen3 and GPT-OSS (the only callers of apply_rotary_emb_cos_sin). RoPE now always computes in fp32 in eager too: slightly more accurate, matches HF Qwen3 behavior, no checkpoint impact.

### DIFF
--- a/tests/unit_tests/test_rope.py
+++ b/tests/unit_tests/test_rope.py
@@ -1,0 +1,68 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+
+import torch
+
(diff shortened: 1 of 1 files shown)


tianyu-l (approved):
nice catch, thanks for fixing!

Issue #2523, opened 2026-03-07
Adding Z-Loss


francesco-bertolotti:
The **z-loss** is a commonly used technique to prevent logit explosion, and it is currently used in both OLMo3 and Trinity. Would you be interested in adding support for it in `torchtitan`, perhaps in [loss.py](https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/loss.py), with a configurable weight?

It could also be useful to log the z-loss alongside the NTP loss.

To support this and similar extensions, we could modify the loss interface so that it returns either a dictionary or a custom loss object that tracks both the main loss and any auxiliary losses.

---

If this sounds useful to you as well, I’d be happy to start working on a PR implementing this feature.

tianyu-l:
Thanks. Is it fair to say that z-loss has not become the de facto standard in llm training?

If so, I think we can start with adding z-loss to logging (in no-grad mode), but not using it as a default regularizer.

Whether we can make it configurable to turn on depends on how the implementation looks, which we can discuss after you have one. WDYT?

cc @felipemello1 who's looking at improving logging in general.

francesco-bertolotti:
@tianyu-l 

I should be able to submit a PR this week.

felipemello1:
@francesco-bertolotti , regarding the loss, for now you will have to do the manual all_reduce and logging like titan currently does it. After we make some changes to make titan for dtensor native, you should be able to do something like: `record_tensor_metric(sum=loss.sum(), count=(mask==1).sum())`, and we will handle the all reduce and logging for you.

That said, I am about to push a PR changing MetricsProcessor greatly. So perhaps just add the z-loss and print or log it naively, but dont make changes to the MetricsProcessor. You will have many conflicts. Then you can rebase on my PR.

PR #2576, opened 2026-03-14
Configurable Z Loss

francesco-bertolotti:
### Overview

Following [https://github.com/pytorch/torchtitan/issues/2523](https://github.com/pytorch/torchtitan/issues/2523), this PR introduces an initial implementation of **z-loss** and includes some refactoring to make the loss configuration more flexible.

### Z-Loss

I added z-loss support to the cross-entropy loss. The implementation is inspired by the one used in OLMo:
[https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/nn/cross_entropy_loss.py](https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/nn/cross_entropy_loss.py).

If `z_loss_weight` is set to a value different from `0`, the z-loss is computed and added to the cross-entropy loss, scaled by `z_loss_weight`.

### Refactoring

I also refactored the loss configuration so it can be defined via the CLI or through TOML configuration files. This is just a proposal and can be adapted if it does not align with the Torchtitan design.

The current setup allows configuring multiple loss types (currently **MSE** and **CrossEntropy**). Each loss is defined as a configurable object with the following fields:

* **enable**: whether the loss is active (exactly one loss must be enabled)
* **compile**: if `true`, the loss module is compiled with `torch.compile`

Additional options are available for specific losses:

* **CrossEntropyLoss**

  * `z_loss_weight` (default: `0.0`)
  * `ignore_index` (default: `-100`)

Both **CrossEntropy** and **MSE** return a `LossOutput` object containing:

* `main`: the loss used for gradient computation (the one `.backward()` should be called on)
* `aux`: a dictionary containing auxiliary values intended only for logging

For example, the CrossEntropy loss populates `LossOutput.aux` with both the **unscaled z-loss** and the **raw cross-entropy loss**.

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -4,32 +4,34 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from dataclasses import dataclass, field
 from collections.abc import Callable
+from dataclasses import dataclass, field
 from typing import TypeAlias
 
 import torch
 
 from torchtitan.config import Configurable
 from torchtitan.tools.logging import logger
-from torchtitan.distributed import ParallelDims
-from functools import partial
 
 # PyTorch's default ignore index for cross-entropy loss
 IGNORE_INDEX = -100
 
+
 @dataclass
 class LossOutput:
     """
-    A wrapper for the loss output, 
-    contains a main loss from which gradients will be computed, 
-    and auxiliary losses that can be used for logging 
+    A wrapper for the loss output,
+    contains a main loss from which gradients will be computed,
+    and auxiliary losses that can be used for logging
     but will not contribute to the gradient computation.
     """
+
     main: torch.Tensor
     aux: dict[str, torch.Tensor] = field(default_factory=dict)
 
+
 LossFunction: TypeAlias = Callable[..., LossOutput]
+
 
 class CrossEntropyLoss(Configurable):
     @dataclass(kw_only=True, slots=True)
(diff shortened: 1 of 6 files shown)


torchtitan/protocols/model_spec.py
```diff
@@ -39,7 +39,6 @@ class ModelSpec:
     # resolver does not handle ``Callable[..., X]`` (Ellipsis as param spec).
     # The detailed TypeAliases above are still available for use in function
     # signatures elsewhere in the codebase.
-    build_loss_fn: Callable
```
tianyu-l, line 42: Changing loss configuration from "build function" to config-based sounds good. However,
- why do we move it from `ModelSpec` to `Trainer`?
- by default only CrossEntropyLoss is used, which changes the default for FLUX model

torchtitan/components/loss.py
```diff
@@ -5,49 +5,132 @@
...
+    def mse_loss(pred: torch.Tensor, labels: torch.Tensor) -> LossOutput:
+        """Common MSE loss function with sum reduction for Transformer models training."""
+        return LossOutput(
+            main=torch.nn.functional.mse_loss(
+                pred.float(), labels.float().detach(), reduction="sum"
+            )
+        )
+
+
+class Loss(Configurable):
```
tianyu-l, line 110: We should have one `BaseLoss` class which crossentropy and mse inherit, rather than one monolithic Loss class responsible for everything.

torchtitan/components/loss.py
```diff
@@ -5,49 +5,132 @@
...
-def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
-    """Cross-entropy loss with sum reduction for token-based normalization."""
-    return torch.nn.functional.cross_entropy(
-        pred.flatten(0, 1).float(),
-        labels.flatten(0, 1),
-        reduction="sum",
-        ignore_index=IGNORE_INDEX,
-    )
+    main: torch.Tensor
+    aux: dict[str, torch.Tensor] = field(default_factory=dict)
```
tianyu-l, line 30: It seems `aux` is only for logging. I think it'd be cleaner to wait until https://github.com/pytorch/torchtitan/pull/2607 lands.

francesco-bertolotti: Yep aux is meant only for logging. No problem in waiting for https://github.com/pytorch/torchtitan/pull/2607

torchtitan/components/loss.py
francesco-bertolotti: Right now, cross-entropy and MSE losses are implemented as separate classes (though they could definitely share a common `BaseLoss` parent). The current `Loss` object simply routes to the appropriate implementation based on the configuration.

For example, the CLI for configuring the loss looks like this:

```bash
torchtitan/train.py \
  --config qwen3_debugmodel \
  --module qwen3 \
  --loss.cross-entropy.enable \
  --loss.cross-entropy.compile \
  --loss.cross-entropy.ignore-index -100 \
  --loss.cross-entropy.z-loss-weight 1e-5
```

torchtitan/protocols/model_spec.py
francesco-bertolotti: I was under the impression that this was no longer necessary, since the loss can now be configured directly from the config file. So I simplified it. However, it may make more sense to keep it, I'm happy to revert this change if needed.

torchtitan/components/loss.py
tianyu-l: the current thinking around CLI is that
- python config is strictly closer to the actual object than CLI, thus more "minimal" to support
- the design should focus on this minimal nature, not biased (too much) by CLI concern
- users can always build CLI oriented config on top of existing config

PR #2577, opened 2026-03-14
validate n_kv_heads divisibility by TP degree at config time

dean-mccoppin:
Qwen3.5-MoE (`n_kv_heads=2`) and similar GQA models crash with a cryptic reshape error when `tensor_parallel_degree > n_kv_heads` because `ColwiseParallel` on `wk`/`wv` ends up with a fractional number of KV heads per rank.

Added a `ValueError` in `update_from_config` for llama3, llama4, qwen3, and gpt_oss so you get a readable message at config time instead. DeepSeek V3 uses MLA so it's unaffected

Also added unit tests in `tests/unit_tests/test_tp_kv_heads_validation.py`

### DIFF
--- a/torchtitan/models/gpt_oss/model.py
+++ b/torchtitan/models/gpt_oss/model.py
@@ -250,9 +250,8 @@ class GptOssModel(Decoder):
             tp = parallelism.tensor_parallel_degree
             if tp > 1:
                 n_heads = self.layer.attention.n_heads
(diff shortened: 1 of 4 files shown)


dean-mccoppin:
I'll fix the linting, thanks for checking it out tianyu

tianyu-l:
please fix linting

PR #2578, opened 2026-03-14
fix: support torch.autograd.set_detect_anomaly with DTensor via check_nan=False

### LINKED ISSUE #1544: Nan detection using  torch.autograd.set_detect_anomaly(True) runs into error.
Using   torch.autograd.set_detect_anomaly(True) to track NaN issues. However using it  results in the following error.   

 NotImplementedError: Operator aten._is_any_true.default does not have a sharding strategy registered.

dean-mccoppin:
Fixes #1544  

`set_detect_anomaly(True)` does two things: records forward-pass creation stack
traces, and checks gradients for NaN/Inf via `aten._is_any_true`. DTensor has no
sharding strategy for `_is_any_true`, so distributed training crashes immediately.
Passing `check_nan=False` skips the NaN check while keeping the stack traces 
which is the useful part anyway, since it tells you where the bad tensor came from.

Also added a `detect_anomaly` field to `DebugConfig` so users can enable it via
`--debug.detect_anomaly=true` instead of calling the API manually and hitting the crash

Tested with `pytest tests/unit_tests/test_set_determinism.py`

### DIFF
--- a/tests/unit_tests/test_set_determinism.py
+++ b/tests/unit_tests/test_set_determinism.py
@@ -277,7 +277,6 @@ class TestSetDeterminismWithFakeMesh(unittest.TestCase):
             distinct_seed_mesh_dims=["pp"],
         )
 
-
     @patch("torch.distributed.distributed_c10d.get_world_size")
     @patch("torch.distributed.distributed_c10d.get_rank")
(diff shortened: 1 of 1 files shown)


tianyu-l:
sounds OK, please fix linting

tianyu-l (approved):

PR #1590, opened 2025-08-18
[Qwen3] Switch to verified RoPE implementation + Add weight tying support

wwwjn:
## Context
1. Current Qwen3 model RoPE used some trick to make numerical parity with HF. This trick is from un-official source and hard to reasoning mathematically. Switch to [torchtune based implementation](https://github.com/pytorch/torchtune/blob/main/torchtune/models/qwen2/_positional_embeddings.py#L14), which was directly contributed from Qwen team.  Thanks @ebsmothers for point us to this implementation!
    - For RoPE embedding, I change it to the same way as complex representation based RoPE in llama3: We initialize and precompute the RoPE embedding cos/sin value only once, and pass it into Attention module during forward. In this way, TP can be applied seamlessly.
    - In contrast, torchtune passed the RoPE class into initialize function for each layers' attention module.
2. Add weight tying support for Qwen3, verified with FSDP + TP

## Numerical verification for RoPE
Run end-to-end forward pass of Qwen3 model, the output and 
Screenshot 2025-08-18 at 2 48 48 PM

## Weight tying Verification:

1. With vs. without weight tying on torchtitan model: (FSDP=4, loss are exactly the same)
Screenshot 2025-08-18 at 6 19 13 PM

2. torchtitan with weight tying vs. HF  
Screenshot 2025-08-18 at 9 37 50 PM

3. Weight tying memory address / id check: (in train.py) - passed
```
assert id(model.tok_embeddings.weight) == id(model.output.weight), "id check 2"
assertEqual(model.tok_embeddings.weight, model.output.weight)

# model.forward()

assert id(model.tok_embeddings.weight.grad) == id(model.output.weight.grad), "id check 2"
assertEqual(model.tok_embeddings.weight.grad, model.output.weight.grad)
```

PR #2580, opened 2026-03-15
Add weight tying support for Llama3

### LINKED ISSUE #1524: Weight tying between embedding and LM head layer
Hi TorchTitan team,

I am wondering can simpleFSDP support weight tying which is also a thing in Llama-3? I did some initial exploration but it seems not supported yet?

Thank you very much!

dean-mccoppin:
Implements enable_weight_tying for Llama3, sharing tok_embeddings.weight with output.weight. It mirrors the Qwen3 implementation from #1590 (thanks!)

Changes cover model.py (config field, tying in __init__/init_weights, PP guard), parallelize.py (grouped FSDP unit for tied params), state_dict_adapter.py (skip/reconstruct output.weight for HF conversion), and a new unit test file

Closes #1524

### DIFF
--- a/tests/unit_tests/test_weight_tying.py
+++ b/tests/unit_tests/test_weight_tying.py
@@ -85,6 +85,7 @@ class TestLlama3WeightTying(unittest.TestCase):
         trainer_config.training.seq_len = 512
         trainer_config.parallelism.pipeline_parallel_degree = 2
         trainer_config.parallelism.context_parallel_degree = 1
+        trainer_config.parallelism.tensor_parallel_degree = 1
 
         with self.assertRaises(NotImplementedError):
             config.update_from_config(trainer_config=trainer_config)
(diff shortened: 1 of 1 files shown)


tests/unit_tests/test_weight_tying.py
```diff
@@ -0,0 +1,119 @@
...
+        FeedForward,
+    )
+    from torchtitan.models.common.linear import Linear
+    from torchtitan.models.common.rmsnorm import RMSNorm
+    from torchtitan.models.common.rope import RoPE
+    from torchtitan.models.llama3.model import Llama3Model, Llama3TransformerBlock
+
+    HAS_TORCHTITAN_MODELS = True
+except Exception:
+    HAS_TORCHTITAN_MODELS = False
```
tianyu-l: what's this for?

dean-mccoppin: my reasoning was because torchtitan/models/common/__init__.py re-exports from moe/, which imports triton at module level, and if triton isnt installed then the import chain fails. but i did just notice that many existing unit tests import from the same torchtitan.models.common.* submodules with no guard and they pass in CI, so I'll be removing this

tianyu-l:
please fix tests

Issue #2379, opened 2026-02-13
Feature Request: Async activation offloading to CPU with prefetch


niyunsheng:
### Motivation

torchtitan currently supports **selective activation checkpointing** (selective AC), which saves
the outputs of compute-intensive ops like `mm` and flash attention to avoid recomputing them in
backward. This effectively eliminates the major FLOP overhead of checkpointing.

However, selective AC still operates within **block-wise checkpointing** — the entire transformer
block is checkpointed, and all ops not in the save list are recomputed. This means memory-bound
ops like layernorm, activation functions, dropout, and other elementwise ops are still redundantly
re-executed in backward. While these ops contribute minimal FLOPs, they are **bandwidth-bound** and
consume real GPU execution time, reducing overall training throughput.

The most performant approach is to **eliminate block-wise checkpointing entirely** and instead use
**async activation offloading**: offload all activations to CPU during forward and prefetch them
back before backward. This avoids all redundant recomputation — including the memory-bound ops
that selective AC cannot eliminate — while the D2H/H2D transfers can be fully overlapped with
computation via separate CUDA streams, achieving near-zero overhead.

### Current state

- torchtitan has activation checkpointing (`torchtitan/distributed/activation_checkpoint.py`)
  with modes: `full`, `selective`, `memory_budget`.
- **No activation offloading support exists.**
- I have filed a feature request in PyTorch core (
  https://github.com/pytorch/pytorch/issues/174960) proposing to
  integrate activation offloading into FSDP's `CPUOffloadPolicy`. However, PyTorch core
  features have long design and review cycles, and this may take a while.
- In the meantime, torchtitan can implement this at the application level using
  `torch.autograd.graph.saved_tensors_hooks`, without waiting for PyTorch core support.

### Proposed usage

```python
for layer in model.layers:
    with activation_offload():
        x = layer(x)
```

Activation offloading is orthogonal to checkpointing — they can be used independently
or composed together.

### Implementation approach

I suggest following [Megatron-Kwai's `OffloadManager`](https://github.com/kwai/Megatron-Kwai/blob/sc25slimpipe/megatron/core/pipeline_parallel/offload.py)
(BSD 3-Clause licensed, compatible with torchtitan's BSD license). 

Key design points:
1. **`saved_tensors_hooks`**: Use `pack_hook` to collect activation tensors during forward,
then batch offload to CPU; use `unpack_hook` to return prefetched tensors in backward.
[...]

PR #2581, opened 2026-03-15
add async activation offloading to CPU with pinned memory pool

dean-mccoppin:
Implements async activation offloading to CPU during forward with prefetch back before backward (issue #2379). Activations are copied D2H on a dedicated CUDA stream, then prefetched H2D on a second stream, so transfers overlap with compute at near-zero overhead

Uses a pre-allocated pinned CPU slab to avoid per-tensor cudaMallocHost calls, which trigger implicit CUDA sync and kill the overlap benefit. enable_activation_offload: bool added to TrainingConfig. Not supported with pipeline parallel

Found a small bug during testing inside `with torch.cuda.stream(s)`, `torch.cuda.current_stream()` returns s, not the compute stream. The wait_stream call was a no-op, so d2h_stream could overwrite pool memory that h2d_stream was still reading from the previous step. Shows up as silent gradient corruption on pool reuse across steps, but only with async streams enabled. Fix is to capture the compute stream before the context switch.

Here's whats tested:
15 unit tests covering offload eligibility, gradient correctness across multiple steps, pool alignment for mixed dtypes (bf16 + f32), pool exhaustion fallback, and composability with checkpoint_wrapper and FSDP2. Numerical smoke tests confirm float32 and bfloat16 gradients match baseline within tolerance across 3 steps with pool reuse.

Here's my limitations:
Multi-GPU FSDP integration test was not run since only one GPU is available to me. The stream ordering is correct in theory but unverified empirically

### DIFF
--- a/torchtitan/distributed/activation_offloading.py
+++ b/torchtitan/distributed/activation_offloading.py
@@ -7,20 +7,21 @@
 """
 Async activation offloading to CPU with prefetch for torchtitan.
 
-Activations are copied D2H during forward on a dedicated CUDA stream, then
-prefetched H2D during backward on a second stream, so transfers overlap with
-compute at near-zero overhead.
+Activations are copied D2H during forward on a dedicated CUDA stream.  During
+backward, H2D transfers are prefetched one layer ahead using
+``register_full_backward_pre_hook``, so transfers overlap with backward compute.
 
-Key design decisions vs. torchtune's OffloadActivations:
-- Pre-allocated pinned CPU memory pool avoids per-tensor
-  ``torch.empty(..., pin_memory=True)`` calls, which trigger implicit CUDA
-  synchronisation and negate the overlap benefit.
-- Explicit event-based synchronisation keeps the compute stream from consuming
-  data before H2D transfers finish.
+Key design decisions:
+- GPU-side event ordering (``stream.wait_event``) instead of host-blocking
+  ``event.synchronize()`` keeps the CPU thread free to issue work.
+- Backward pre-hooks on transformer layers trigger H2D copies one layer
+  ahead of where backward currently is, overlapping transfers with compute.
+- Pinned CPU buffers are allocated with ``torch.empty(..., pin_memory=True)``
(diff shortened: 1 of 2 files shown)


dean-mccoppin:
Appreciate the review. Fixed the sync issues, replaced the host-blocking event.synchronize() with
_h2d_stream.wait_event() for GPU side ordering, and added _h2d_stream.synchronize() in __enter__ before resetting the
pool so we don't pull the rug on inflight H2D copies. For prefetching, added register_full_backward_pre_hook on each
transformer layer so we kick off H2D copies one layer ahead of where backward is (get_activation_offloading_ctx wires
this up automatically via model.layers). As for the pool benchmarked it against CachingHostAllocator and theres no
meaningful difference (22.24ms vs 21.79ms median on an RTX 4070 Super, 12-layer model) so I'll just remove it. Thanks

torchtitan/distributed/activation_offloading.py
```diff
@@ -0,0 +1,370 @@
...
+            self._d2h_stream = None
+            self._h2d_stream = None
+
+        # _fwd_stash[tensor_id] = (cpu_buf, gpu_tensor, d2h_event)
+        # Keeps the GPU tensor alive until the D2H copy has finished.
+        self._fwd_stash: dict[int, tuple[torch.Tensor, torch.Tensor, Any]] = {}
+
+        # _bwd_stash[tensor_id] = (gpu_buf, h2d_event, cpu_buf)
+        # Keeps both the prefetched GPU tensor AND the source CPU buffer alive
+        # until the next __enter__.  The cpu_buf ref is critical because
```
soulitzer, line 80: Do you have any measurement of peak memory? If next __enter__ happens at the next iteration then you're keeping all reloaded activations alive at once?

soulitzer:
> As for the pool benchmarked it against CachingHostAllocator and theres no
meaningful difference (22.24ms vs 21.79ms median on an RTX 4070 Super, 12-layer model) 

Thanks for benchmarking, although the point is not necessarily speed right? The CachingHostAllocator is on cpp. Couldn't it be more about control. Megatron implementations of offloading have this.

Issue #2599, opened 2026-03-16
No Weight Decay Parameters


francesco-bertolotti:
Often, certain model parameters—such as normalization layers and input/output embeddings—are configured to **exclude weight decay** during optimization. This is particularly important for **input embeddings**, since not all tokens appear in every batch. When a token is relatively rare, its embedding may be pushed toward zero by the optimizer due to weight decay, even though it has not received a gradient update.

Recent LLM architectures such as Trinity and OLMo3 follow this practice, so it could be a useful feature to support.

From an implementation perspective, this could be done by introducing a configuration parameter (e.g., `no_weight_decay_keywords` or `no_weight_decay_params`) in the optimizer component:

[https://github.com/pytorch/torchtitan/blob/fa8e6ccc973a4a6d32a6e236156193f9eacdfa18/torchtitan/components/optimizer.py#L64-L88](https://github.com/pytorch/torchtitan/blob/fa8e6ccc973a4a6d32a6e236156193f9eacdfa18/torchtitan/components/optimizer.py#L64-L88)

One possible approach would be to exclude parameters based on their names. This could be implemented in different ways, for example:

* **Hard match**: exclude a parameter only if its name exactly matches one of the entries.
* **Soft match**: exclude a parameter if its name contains one of the provided keywords (e.g., `any(kw in param_name for kw in no_weight_decay_keywords)`).

If this feature would be useful, I’d be happy to submit a PR. Is there a preferred approach between these matching strategies?

tianyu-l:
Makes sense to me. Could you share details to how Trinity and OLMo3 work with this? I'm interested in learning more.

In terms of implementations
- Should weight decay be part of LR Scheduler's config, instead of Optimizer's?
- If it's just input embedding, maybe we should do hard match (depending on what do you mean by hard which we can discuss in a PR).

francesco-bertolotti:
For Trinity, I only remember a Twitter discussion. For OLMo3, the training recipe uses the following setup:

https://github.com/allenai/OLMo-core/blob/002958a8f15afe5c729affdb12f4c4b12c2f26ad/src/scripts/official/OLMo3/OLMo-3-1025-32B-pretrain.py#L75-L82

which is implemented here:

https://github.com/allenai/OLMo-core/blob/002958a8f15afe5c729affdb12f4c4b12c2f26ad/src/olmo_core/optim/config.py#L120-L154

On second thought, globbing probably makes more sense, since it allows specifying parameters that span across layers more easily.

[...]

PR #2600, opened 2026-03-16
No Weight Decay Keywords

francesco-bertolotti:
Regarding the issue [https://github.com/pytorch/torchtitan/issues/2599](https://github.com/pytorch/torchtitan/issues/2599):

I’ve added a parameter group that excludes weight decay, which is populated based on the `no_weight_decay_keywords` configuration option. Currently, `no_weight_decay_keywords` is a comma-separated list of glob patterns.

For example, setting it to `*emb*,*norm*,*out*` will include in the no-weight-decay parameter group the token embeddings, all normalization layer weights, and the output layer weights.

### DIFF
--- a/torchtitan/components/optimizer.py
+++ b/torchtitan/components/optimizer.py
@@ -5,10 +5,9 @@
 # LICENSE file in the root directory of this source tree.
 
 import functools
-import dataclasses
 from collections.abc import Iterator
 from dataclasses import dataclass, field, replace
-from typing import Any, Generic, Literal, TypeVar
+from typing import Any, Generic, TypeVar
 
 import torch
 import torch.distributed.tensor
@@ -35,10 +34,11 @@ __all__ = [
 
 T = TypeVar("T", bound=Optimizer)
 
-class BaseOptimizer(Configurable):
 
+class BaseOptimizer(Configurable):
     @dataclass(kw_only=True, slots=True)
-    class Config(Configurable.Config): pass
+    class Config(Configurable.Config):
+        pass
 
     optimizer_cls: type[Optimizer]
     config: Config
@@ -61,9 +61,13 @@ class BaseOptimizer(Configurable):
         return {k: v for k, v in self.config.to_dict().items() if v is not None}
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, BaseOptimizer): return False
+        if not isinstance(other, BaseOptimizer):
+            return False
         # Must also check optimizer_cls, not just hyperparams
-        return self.optimizer_cls == other.optimizer_cls and self.config.to_dict() == other.config.to_dict()
+        return (
+            self.optimizer_cls == other.optimizer_cls
+            and self.config.to_dict() == other.config.to_dict()
+        )
 
     def __hash__(self) -> int:
         return hash(tuple(sorted(self.config.to_dict().items())))
@@ -75,8 +79,8 @@ class BaseOptimizer(Configurable):
     def __repr__(self) -> str:
         return self.__str__()
 
-class AdamW(BaseOptimizer):
 
+class AdamW(BaseOptimizer):
     @dataclass(kw_only=True, slots=True)
     class Config(BaseOptimizer.Config):
         lr: float = 8e-4
@@ -88,25 +92,29 @@ class AdamW(BaseOptimizer):
         foreach: bool = False
 
     optimizer_cls: type[Optimizer] = torch.optim.AdamW
+    # pyrefly: ignore [bad-override]
     config: Config
 
     def get_optimizer_kwargs(self) -> dict[str, Any]:
         return {
             "lr": self.config.lr,
             "eps": self.config.eps,
-            "betas": (self.config.beta1, self.config.beta2) if self.config.beta1 is not None and self.config.beta2 is not None else None,
+            "betas": (self.config.beta1, self.config.beta2)
+            if self.config.beta1 is not None and self.config.beta2 is not None
+            else None,
             "fused": self.config.fused,
             "foreach": self.config.foreach,
             "weight_decay": self.config.weight_decay,
         }
 
-class Adam(BaseOptimizer):
 
+class Adam(BaseOptimizer):
     @dataclass(kw_only=True, slots=True)
     class Config(AdamW.Config):
         weight_decay: float = 0.0
 
     optimizer_cls: type[Optimizer] = torch.optim.Adam
+    # pyrefly: ignore [bad-override]
     config: Config
 
     def get_optimizer_kwargs(self) -> dict[str, Any]:
@@ -119,6 +127,7 @@ class Adam(BaseOptimizer):
             "weight_decay": self.config.weight_decay,
         }
 
+
 class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
     """A container for multiple optimizers."""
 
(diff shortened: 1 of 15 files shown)


torchtitan/components/optimizer.py
```diff
@@ -78,6 +80,10 @@ class Config(Configurable.Config):
         weight_decay: float = 0.1
         """Weight decay to use"""
 
+        no_weight_decay_keywords: str = ""
```
tianyu-l: Do you think it makes sense to make this general?
E.g. we can make this a `dict[str, int]` from fqn re to their `weight_decay_multiplier`. The number of keys would correspond to the number of parameter groups.

torchtitan/components/optimizer.py
```diff
@@ -4,9 +4,10 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import fnmatch
```
tianyu-l: should we use `re` which is more powerful and more commonly used in torchtitan?

torchtitan/components/optimizer.py
tianyu-l, line 281: @fegin if we have multiple parameter group, what should we do here?

torchtitan/components/optimizer.py
tianyu-l: should the default be not applying any special multiplier, e.g. to embedding / norm / lm_head?

francesco-bertolotti: I’ve mostly seen weight decay applied selectively (i.e., excluded for certain parameters) rather than using different scaling multipliers, but your idea could be an interesting way to generalize this behavior.

That said, we should design it in a way that integrates cleanly with the CLI. One option could be to allow repeated flags, for example:

```bash
--weight_decay_scale *norm* 0
--weight_decay_scale *emb* 0
--weight_decay_scale *out* 0
```

What do you think?

torchtitan/components/optimizer.py
francesco-bertolotti: I understand your concern. However, if our goal is to select parameters using bash-style globbing, I think `fnmatch` is the better fit, it already provides the matching behavior we need without relying on custom regular expressions, which can be more error-prone.

If we decide to support a different syntax instead of bash-style globbing, then using another approach would make sense. Also, both `re` and `fnmatch` are part of the standard library, so adopting `fnmatch` wouldn’t introduce any additional dependencies.

That said, if using `fnmatch` isn’t an option, I’m happy to switch to `re` without any issues.

torchtitan/components/optimizer.py
tianyu-l: Yeah I noticed that idea might break CLI, too. Let me think about it.

tianyu-l: @francesco-bertolotti I discussed with @fegin and here are the latest thoughts

> I’ve mostly seen weight decay applied selectively (i.e., excluded for certain parameters) rather than using different scaling multipliers

We've definitely seen people using more than no-weight-decay, i.e. different parameters having different scaling multipliers, so that's real use case.

> That said, we should design it in a way that integrates cleanly with the CLI.

It is somewhat similar to parameter init which is being refactored in https://github.com/pytorch/torchtitan/pull/2633. In general it feels to me that CLI is not the right place to config them.

If users really want to use CLI to toggle on/off weight decay, they can introduce a translate layer (e.g. by subclassing `OptimizersContainer`) to support CLI on top of torchtitan.

WDYT?

torchtitan/components/optimizer.py
fegin: This should work if `all_params` is the parameter group list like what is done in this PR.

torchtitan/components/optimizer.py
```diff
@@ -111,16 +117,45 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
+                        for pattern in config.no_weight_decay_keywords.split(",")
+                    )
+                ]
+                logger.info(
+                    f"Not applying weight decay to parameters {[n for n,p in nowd_params]} in model {model._get_name()}"
+                )
+                optimizer.add_param_group(
+                    {"params": [p for n, p in nowd_params], **optimizer_nowd_kwargs}
+                )
+                all_params.extend(p for n, p in nowd_params)
```
fegin: We should make this a generic parameter group logic instead of hard-coding the weight decay here. People may use parameter groups for optimization choices. So as @tianyu-l mentions above, it seems to us that the generic way is to specify the parameter optimization options in the config similar to parameter init then you should be able to use a loop to categorize the parameters into different groups.

torchtitan/components/optimizer.py
francesco-bertolotti: I have looked at https://github.com/pytorch/torchtitan/pull/2633 and it seems like a more general way to approach this. I will try to adopt the same solution.

francesco-bertolotti:
Hi @tianyu-l and @fegin,

This PR refactors the optimizer configuration system to support per-parameter-group overrides, enabling fine-grained control over optimizer hyperparameters (e.g., assigning different learning rates or weight decay to specific parameter groups such as embeddings).

While working on this, the scope expanded slightly beyond hyperparameter overrides: the new system also supports using different optimizer types within the same run. This allows assigning distinct optimizers to different parameter groups, which can be particularly useful for cases like using `SparseAdam` for embedding layers.

---

### **Key Changes**

#### **1. New `BaseOptimizer` abstraction** (`torchtitan/components/optimizer.py`)

* Introduces a `BaseOptimizer` class with a nested `Config` dataclass.
* Adds concrete subclasses (`AdamW`, `Adam`) that replace the previous string-based optimizer resolution (`_resolve_optimizer_cls`).
* Each optimizer subclass:

  * Defines its own config schema (`lr`, `betas`, `eps`, `weight_decay`, `fused`, `foreach`)
  * Implements `get_optimizer_kwargs()` to construct optimizer arguments
* Supports config composition via `BaseOptimizer.__or__`, enabling clean layering of overrides on top of defaults.

---

#### **2. `OptimizersContainer` config overhaul**

* Replaces the previous flat config (`name`, `lr`, `beta1`, `beta2`, etc.) with a structured schema:

```python
default: BaseOptimizer.Config
overrides: dict[str, BaseOptimizer.Config]
```

* `default` defines the base optimizer configuration applied to all parameters.

* `overrides` enables per-parameter configuration using parameter name matching.

* Parameters are grouped based on override keys, and separate optimizer parameter groups are constructed with the appropriate configs.

* `OptimizersInBackwardContainer` has also been updated to follow this new structure.

---

#### **3. Config registry migrations (16 files)**

* Updated all model and experiment configs (e.g., `llama3`, `llama4`, `deepseek_v3`, `flux`, `gpt_oss`, `qwen3`, as well as `ft`, `rl`, `vlm`) from the flat syntax:

```python
OptimizersContainer.Config(lr=8e-4)
```

to the new nested format:

```python
OptimizersContainer.Config(
    default=AdamW.Config(lr=8e-4)
)
```

---

#### **4. Tests**

* Updated unit tests (`tests/unit_tests/test_train_spec.py`) to:

  * Use the new `Adam.Config` structure
  * Reflect the updated access pattern (`optimizers[0][0]`)

---

### **Example Usage**

```python
OptimizersContainer.Config(
    default=AdamW.Config(lr=3e-4, weight_decay=0.1),
    overrides={
        "model.tok_embeddings.weight": AdamW.Config(
            lr=1e-4,
            weight_decay=0.0,
        ),
    },
)
```

---

### **Results**

Loss curves before and after these changes:

vsopts

tianyu-l:
I'm giving a different try in https://github.com/pytorch/torchtitan/pull/2859. Could you help review?

PR #2602, opened 2026-03-16
[Observability 2/7] Structured system logging and gantt chart generation

felipemello1:
* #2607
* #2606
* #2605
* #2604
* #2603
* __->__ #2602
* #2601

## How to review

Prioritize the files under /observability. Check how they are used in the toy example.
No need to nitpick the toy example. Real implementation in trainer are in PRs 6 and 7.

## Summary

The system logging layer. Every process writes structured JSONL to disk via Python's logging module.

Two APIs:
- `record_span(description, event_type)` — context manager that writes START/END events with wall-clock duration. Used for timing training phases (forward/backward, optimizer, data loading, etc).
- `record_event({"key": value, ...})` — point-in-time scalar snapshots for per-rank diagnostics.

## Example

```python
from torchtitan.observability import (
    init_observability, set_step, add_step_tag,
    record_span, record_event, EventType,
)
from torchtitan.tools.logging import init_logger

# Console logging (stdout with [titan] format)
init_logger()

# JSONL file handlers for structured logging
init_observability(source="trainer", output_dir="./outputs")

for step in range(steps):
    # Stamp all subsequent JSONL entries with step=step
    set_step(step)

    if should_garbage_collect:
        add_step_tag("gc")
        with record_span("trainer_time/gc_s", EventType.GC_COLLECT):
            run_gc()

    with record_span("trainer_time/forward_backward_s", EventType.FWD_BWD):
        reduced_loss = model.fwd_bwd(batch)
        record_event({"loss": reduced_loss.item()})
```

```json
// dump_dir/system_logs/trainer_rank_0_system.jsonl
{"int": {"step": 42, "rank": 0, "time_us": 1708200121724000},
 "normal": {"log_type_name": "fwd_bwd_end", "source": "trainer",
            "message": "[step 42] trainer_time/forward_backward_s fwd_bwd_end took 123.45 ms",
            "caller": "torchtitan/trainer.py:730:train_step"},
 "double": {"value": 123.45},
 "normvector": {"step_tags": ["gc"]}}
```

### Output folder (from toy_spmd on this PR)

```
outputs/toy_spmd/
├── analysis/
│   └── system_metrics_gantt.json      ← generate_gantt_trace output
└── system_logs/
    ├── trainer_rank_0_system.jsonl
    ├── trainer_rank_1_system.jsonl
    ├── trainer_rank_2_system.jsonl
    └── trainer_rank_3_system.jsonl
```

### Gantt chart

**toy_rl**
gantt_rl

**toy_spmd**
ganttspmd

## Others

`init_observability(source, output_dir)` sets up the JSONL file handlers. Each rank gets its own file: `{output_dir}/system_logs/{source}_rank_{rank}_system.jsonl`.

The JSONL format uses four typed columns (`int`, `normal`, `double`, `normvector`) for easy ingestion into Grafana, DuckDB, etc.

`generate_gantt_trace(log_dir, output_path)` reads all system JSONL files and produces a Chrome Trace JSON. Open in Perfetto to see a gantt chart of every `record_span` across all ranks.

Also includes:
 - `EventType` enum for categorizing spans,
 - `set_step`/`add_step_tag` for step context,
 - `InflightEventTrackingHandler` for crash forensics

## Test plan

Run toy_spmd: `python -m torch.distributed.run --nproc_per_node=4 -m torchtitan.experiments.observability.toy_spmd`
Run toy_rl: `python -m torchtitan.experiments.observability.toy_rl`

- [x] 44 new unit tests for record_span, record_event, init_observability, gantt generation, step state
- [x] Integration: toy_spmd produces system JSONL with record_span start/end events
- [x] Integration: toy_rl produces system JSONL from 4 actor types
- [x] Integration: generate_gantt_trace produces Chrome Trace JSON from system JSONL

### Console output (toy_spmd)
```
step: 1  loss: 3.65555  grad_norm: 0.49419
step: 5  loss: 3.05283  grad_norm: 0.40876
step: 10  loss: 2.53639  grad_norm: 0.31771
  val loss: 2.4611
step: 20  loss: 2.02626  grad_norm: 0.22093
  val loss: 1.9949
Chrome Trace: outputs/toy_spmd/analysis/system_metrics_gantt.json
  576 events from 4 sources
```

### Console output (toy_rl)
```
step: 1  loss: 3.65156  grad_norm: 0.49804
step: 1  reward_mean: 1.00000
step: 3  loss: 3.32657  grad_norm: 0.44708
step: 3  reward_mean: 1.00000
step: 6  loss: 2.93530  grad_norm: 0.36988
step: 6  reward_mean: 1.00000
Chrome Trace: outputs/toy_rl/analysis/system_metrics_gantt.json
Done in 13.8s.
```

Sample of generated logs
```json
{"int": {"tid": 1056731, "thread_time_ns": 3325664547, "rank": 0, "pid": 1056731, "time": 1773610642, "time_ms": 1773610642167, "time_us": 1773610642167603, "seq_id": 0}, "normal": {"source": "trainer", "host_name": "devgpu011.ldc3.facebook.com", "log_type": "event", "log_type_name": "build_model_start", "event_name": "setup/model_build", "caller": "torchtitan/experiments/observability/toy_spmd.py:168:__init__", "log_file": "toy_spmd.py", "log_function": "__init__", "log_level": "INFO", "logger_name": "torchtitan.observability.system", "message": "[step N/A] setup/model_build build_model_start"}, "double": {"delta_ms": 0.00051083043217659}, "normvector": {}}
{"int": {"tid": 1056731, "thread_time_ns": 3998473694, "rank": 0, "pid": 1056731, "time": 1773610642, "time_ms": 1773610642858, "time_us": 1773610642858849, "seq_id": 1}, "normal": {"source": "trainer", "host_name": "devgpu011.ldc3.facebook.com", "log_type": "event", "log_type_name": "build_model_end", "event_name": "setup/model_build", "caller": "torchtitan/experiments/observability/toy_spmd.py:168:__init__", "log_file": "toy_spmd.py", "log_function": "__init__", "log_level": "INFO", "logger_name": "torchtitan.observability.system", "message": "[step N/A] setup/model_build build_model_end took 691.24 ms"}, "double": {"delta_ms": 691.2433146499097, "value": 691.238438244909}, "normvector": {}}
```

### DIFF
--- a/torchtitan/experiments/observability/toy_spmd.py
+++ b/torchtitan/experiments/observability/toy_spmd.py
@@ -34,6 +34,16 @@ from torch.distributed.tensor.parallel import (
 
 from torchtitan.distributed import ParallelDims
 from torchtitan.distributed.utils import clip_grad_norm_
+from torchtitan.observability import (
+    add_step_tag,
+    EventType,
+    init_observability,
+    record_event,
+    record_span,
+)
+from torchtitan.observability.analysis import generate_gantt_trace
+from torchtitan.observability.metrics_processor import MetricsProcessor
+from torchtitan.tools.logging import init_logger
 
 # ---- Config ----
 NUM_STEPS = 20
@@ -155,11 +165,14 @@ class ToyTrainer:
         self.output_dir = output_dir
         self.step = 0
 
+        self.metrics_processor = MetricsProcessor()
+
         torch.manual_seed(0)
-        model = TinyModel().to(device)
-        self._apply_tp(model, tp_mesh)
-        self._apply_compile(model)
-        self._apply_fsdp(model, dp_mesh)
+        with record_span("setup/model_build", EventType.BUILD_MODEL):
+            model = TinyModel().to(device)
+            self._apply_tp(model, tp_mesh)
+            self._apply_compile(model)
+            self._apply_fsdp(model, dp_mesh)
         self.model = model
         self.optimizer = torch.optim.AdamW(model.parameters(), lr=LR)
 
@@ -220,7 +233,9 @@ class ToyTrainer:
         """Wraps a dataloader into an iterator."""
         data_iterator = iter(data_iterable)
         while True:
-            yield next(data_iterator)
+            with record_span("trainer_time/data_loading_s", EventType.FETCHING_BATCH):
+                batch = next(data_iterator)
+            yield batch
 
     def compute_loss(self, logits, labels, loss_mask):
         """Compute cross-entropy loss with masking under loss_parallel.
@@ -240,25 +255,30 @@ class ToyTrainer:
 
     def train_step(self, tokens, labels, loss_mask):
         """One training step."""
-        with loss_parallel():
-            logits = self.model(tokens)
-            loss_sum, valid_tokens = self.compute_loss(logits, labels, loss_mask)
-            # Globally-normalized loss: each token contributes equally to
-            # gradients regardless of which DP rank it's on (matches titan).
-            global_valid_tokens = valid_tokens.detach().clone().float()
-            dist.all_reduce(global_valid_tokens, group=self.dp_mesh.get_group())
+        with record_span("trainer_time/forward_backward_s", EventType.FWD_BWD):
+            with loss_parallel():
+                logits = self.model(tokens)
+                loss_sum, valid_tokens = self.compute_loss(logits, labels, loss_mask)
+                # Globally-normalized loss: each token contributes equally to
+                # gradients regardless of which DP rank it's on (matches titan).
+                global_valid_tokens = valid_tokens.detach().clone().float()
+                dist.all_reduce(global_valid_tokens, group=self.dp_mesh.get_group())
+                loss = loss_sum / global_valid_tokens
+                self.optimizer.zero_grad()
+                loss.backward()
 
-            loss = loss_sum / global_valid_tokens
-            self.optimizer.zero_grad()
-            loss.backward()
-        grad_norm = clip_grad_norm_(self.model.parameters(), max_norm=1.0)
-        self.optimizer.step()
+        with record_span("trainer_time/optimizer_s", EventType.OPTIM):
+            grad_norm = clip_grad_norm_(self.model.parameters(), max_norm=1.0)
+            self.optimizer.step()
 
         # Report globally-reduced loss. Each DP rank has
         # local_loss_sum / global_valid_tokens; SUM gives the global loss.
         loss_scalar = loss.detach().full_tensor().clone()
         dist.all_reduce(
             loss_scalar, op=dist.ReduceOp.SUM, group=self.dp_mesh.get_group()
+        )
+        record_event(
+            {"train.loss": loss_scalar.item(), "train.grad_norm": grad_norm.item()}
         )
         if self.rank == 0:
             print(
@@ -289,15 +309,24 @@ class ToyTrainer:
 
         for step in range(1, num_steps + 1):
             self.step = step
-            tokens, labels, loss_mask = next(data_iterator)
-            self.train_step(tokens, labels, loss_mask)
+            self.metrics_processor.set_step(step)
+
+            # Simulate GC on every 5th step (mirrors gc_handler.run)
+            if step % 5 == 0:
+                add_step_tag("gc")
+
+            with record_span("trainer_time/step_s", EventType.STEP):
+                tokens, labels, loss_mask = next(data_iterator)
+                self.train_step(tokens, labels, loss_mask)
 
             if step % EVAL_FREQ == 0:
-                self.validate(tokens, labels, loss_mask)
+                add_step_tag("eval")
+                with record_span("trainer_time/validation_s", EventType.EVAL):
+                    self.validate(tokens, labels, loss_mask)
 
     def close(self):
-        """Cleanup. Subclasses override to close writers, profilers, etc."""
-        pass
+        """Cleanup."""
+        self.metrics_processor.close()
 
 
 def main():
@@ -312,6 +341,9 @@ def main():
         shutil.rmtree(OUTPUT_DIR)
     dist.barrier()
     os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+    init_logger()
+    init_observability(source="trainer", output_dir=OUTPUT_DIR, rank=rank)
 
     parallel_dims = ParallelDims(
         dp_replicate=1,
@@ -333,7 +365,10 @@ def main():
     trainer.close()
 
     if rank == 0:
-        print(f"Done. Output: {OUTPUT_DIR}")
+        sys_logs = os.path.join(OUTPUT_DIR, "system_logs")
+        trace_path = os.path.join(OUTPUT_DIR, "analysis", "system_metrics_gantt.json")
+        generate_gantt_trace(sys_logs, trace_path)
+        print(f"\nDone. Output: {OUTPUT_DIR}")
     dist.destroy_process_group()
 
 
--- a/torchtitan/observability/metrics_processor.py
+++ b/torchtitan/observability/metrics_processor.py
@@ -1,0 +1,31 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Toy MetricsProcessor for the observability experiment."""
+
+from torchtitan.observability import record_event
+from torchtitan.observability.step_state import set_step
+
+
+class MetricsProcessor:
+    """Step context manager for the toy trainer.
+
+    Mirrors the method order of components/metrics.py MetricsProcessor
+    so the toy and production versions are easy to compare.
+    """
+
+    def __init__(self):
+        self._step: int = 0
+
+    def set_step(self, step: int) -> None:
+        """Set the current training step."""
+        self._step = step
+        set_step(step)
+        record_event({"train.step": step})
+
+    def close(self) -> None:
+        """Shutdown. Extended with subprocess cleanup when logging is added."""
+        pass
--- a/torchtitan/observability/structured_logging.py
+++ b/torchtitan/observability/structured_logging.py
@@ -1,0 +1,459 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""System JSONL logging: init_observability, record_span, record_event."""
+
+import enum
+import itertools
+import json
+import logging
+import os
+import socket
+import threading
+import time
+from contextlib import ContextDecorator
+from timeit import default_timer as timer
+from typing import Any
+
+from torchtitan.observability._constants import SYSTEM_LOGGER_NAME
+from torchtitan.observability.step_state import get_step, get_step_tags
+
+MAX_MESSAGE_SIZE: int = 1000
+
+# ---------------------------------------------------------------------------
+# Enums
+# ---------------------------------------------------------------------------
+
+
+class StrEnum(enum.Enum):
+    """String-valued enum. str(member) returns the value, not 'Class.name'."""
+
+    def __str__(self) -> str:
+        return self.value
+
+
+class EventType(StrEnum):
+    """Standardized phase identifiers for structured JSONL events.
+    Used for categorization and analysis.
+    """
+
+    TORCH_DISTRIBUTED_INIT = "torch_distributed_init"
+    TORCH_DISTRIBUTED_TEARDOWN = "torch_distributed_teardown"
+    MODEL_PARALLELISM_INIT = "model_parallelism_init"
+    TOKENIZER_INIT = "tokenizer_init"
+    DATA_ITERATOR_INIT = "data_iterator_init"
+    RELOAD_DATA_LOADER_STATE = "reload_data_loader_state"
+    BUILD_MODEL = "build_model"
+    BUILD_LEARNER = "build_learner"
+    OPTIMIZER_INIT = "optimizer_init"
+    TRAINING_START = "training_start"
+    STEP = "step"
+    FETCHING_BATCH = "fetching_batch"
+    FWD_BWD = "fwd_bwd"
+    OPTIM = "optim"
+    CHECKPOINT = "checkpoint"
+    CHECKPOINT_INIT = "checkpoint_init"
+    CHECKPOINT_STAGE = "checkpoint_stage"
+    CHECKPOINT_LOAD = "checkpoint_load"
+    EVAL_LAUNCH = "eval_launch"
+    EVAL = "eval"
+    SUMMARY_WRITER = "summary_writer"
+    TORCH_MEMORY_BREAKDOWN = "torch_memory_breakdown"
+    GC_COLLECT = "gc_collect"
+    METRIC_VALUE = "metric_value"
+    STATE_DICT_INIT = "state_dict_init"
+    STATE_DICT_LOAD = "state_dict_load"
+
+    # RL
+    RL_GRADING = "rl_grading"
+    RL_ROLLOUT = "rl_rollout"
+    RL_GENERATE = "rl_generate"
+    RL_ENV = "rl_env"
+
+
+class LogType(StrEnum):
+    """Distinguishes event records from free-text log records in JSONL."""
+
+    EVENT = "event"
+    TEXT = "text"
+
+
+class ExtraFields(StrEnum):
+    """Keys for the `extra` dict passed to logging calls."""
+
+    LOG_TYPE = "log_type"
+    LOG_TYPE_NAME = "log_type_name"
+    EVENT_NAME = "event_name"
+    STEP = "step"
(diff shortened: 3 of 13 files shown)


torchtitan/observability/metrics_processor.py
```diff
@@ -0,0 +1,31 @@
...
+from torchtitan.observability import record_event
+from torchtitan.observability.step_state import set_step
+
+
+class MetricsProcessor:
+    """Step context manager for the toy trainer.
+
+    Mirrors the method order of components/metrics.py MetricsProcessor
+    so the toy and production versions are easy to compare.
+    """
```
felipemello1, line 18: this class is incrementally built over the prs 1-5 (toy examples), so its ready for prs 6-7 (integration with trainer)

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Initialization
+# ---------------------------------------------------------------------------
+
+_system_logger = logging.getLogger(SYSTEM_LOGGER_NAME)
+
+
+def init_observability(source: str, output_dir: str, rank: int | None = None) -> None:
```
felipemello1, line 328: this function (and maybe a couple others) should probably not be here, since its NOT specific to system metrics (its also shared by experiment metrics). I can move in a follow up.

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -289,15 +309,24 @@ def train(self, num_steps):
 
         for step in range(1, num_steps + 1):
             self.step = step
-            tokens, labels, loss_mask = next(data_iterator)
-            self.train_step(tokens, labels, loss_mask)
+            self.metrics_processor.set_step(step)
```
tianyu-l, line 312: When should we call `set_step` directly, vs. call it from `MetricsProcessor`?

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -240,26 +255,31 @@ def compute_loss(self, logits, labels, loss_mask):
...
+            grad_norm = clip_grad_norm_(self.model.parameters(), max_norm=1.0)
+            self.optimizer.step()
 
         # Report globally-reduced loss. Each DP rank has
         # local_loss_sum / global_valid_tokens; SUM gives the global loss.
         loss_scalar = loss.detach().full_tensor().clone()
         dist.all_reduce(
             loss_scalar, op=dist.ReduceOp.SUM, group=self.dp_mesh.get_group()
         )
+        record_event(
```
tianyu-l, line 280: shouldn't it be using `record_metrics`, although I haven't internalized the difference between `record_metrics` and `record_event`

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+class StrEnum(enum.Enum):
+    """String-valued enum. str(member) returns the value, not 'Class.name'."""
+
+    def __str__(self) -> str:
+        return self.value
+
+
+class EventType(StrEnum):
```
tianyu-l, line 38: How necessary is this? My worry is that we'd constantly need to add things to this entry. In the extreme case, it's hard to check if something I want is in the existing enum, so that I would blindly add a new one.

Why don't we just stick with whatever use give in str format? If users want, they can define Enums over the string values they have.

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+
+# ---------------------------------------------------------------------------
+# JSONL formatting
+# ---------------------------------------------------------------------------
+
+
+def to_structured_json(log_dict: dict[str, Any]) -> str:
+    """Convert a log dict to 4-column JSON (int/normal/double/normvector)."""
```
tianyu-l, line 121: could you share a specification on the naming? seems there's some convention going on

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+        self._thread_local.last_event_time = timer()
+        return event_delta
+
+
+# ---------------------------------------------------------------------------
+# Filters and handlers
+# ---------------------------------------------------------------------------
+
+
+class EventsOnlyFilter(logging.Filter):
```
tianyu-l, line 269: why do we need this? if everything is filtered, we could have required `LOG_TYPE_NAME` to be given when calling `record_span`

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+    Args:
+        source: Component name (e.g., "trainer", "generator").
+        output_dir: Root output directory.
+        rank: Global rank. If None, reads from RANK env var (default 0).
+    """
+    if rank is None:
+        rank = int(os.environ.get("RANK", 0))
+
+    # --- System handler ---
+    sys_logger = logging.getLogger(SYSTEM_LOGGER_NAME)
```
tianyu-l, line 354: would be helpful if you could also help improve logging module naming across all torchtitan components.

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+
+_system_logger = logging.getLogger(SYSTEM_LOGGER_NAME)
+
+
+def init_observability(source: str, output_dir: str, rank: int | None = None) -> None:
+    """Initialize observability JSONL file handlers.
+
+    Adds per-rank system and experiment JSONL handlers for structured
+    logging. Does NOT set up console logging — call ``init_logger()``
+    from ``torchtitan.tools.logging`` for that.
```
tianyu-l, line 333: do you think we can put both of them into torchtitan/components/observability?

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+        _system_logger.info(
+            f"[step {step if step is not None else 'N/A'}] {name}={value}",
+            extra=event_extra(
+                EventType.METRIC_VALUE, event_name=name, value=value, step=step
+            ),
+            stacklevel=2,
+        )
+
+
+class record_span(ContextDecorator):  # noqa: N801
```
tianyu-l, line 398: noob question: what's the difference between context decorator and context manager?

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+
+    # --- System handler ---
+    sys_logger = logging.getLogger(SYSTEM_LOGGER_NAME)
+    if not any(isinstance(h, StructuredLoggingHandler) for h in sys_logger.handlers):
+        sys_path = os.path.join(
+            output_dir, "system_logs", f"{source}_rank_{rank}_system.jsonl"
+        )
+        handler = StructuredLoggingHandler(filepath=sys_path)
+        handler.setFormatter(StructuredJSONFormatter(rank=rank, source=source))
+        sys_logger.addHandler(handler)
```
tianyu-l, line 361: oh IIUC this is using registration mechanism provided by python's `logging.getLogger`, sounds neat

torchtitan/experiments/observability/toy_spmd.py
felipemello1: record_metric goes to metric logs (something you want to put in wandb). Record event goes to system logs (something that goes to scuba). In your gantt, they are the arrows bellow. Imagine also that they can be something like "counters", for example to check if all ranks have reached that spot, or anything that you would like to query later

image

felipemello1: > 'shouldn't it be using record_metrics' 

added in the next PR. This PR is about record_spans and record_event

torchtitan/observability/structured_logging.py
felipemello1: I was also debating this, and my resolution was to make it optional in `record_span`. So, if the user does:

`record_span("my_string")`, it works.

The reason i decided to keep the Event, was that if we want to compare runs or experiments, and there are N engineers working on different branches, there are no guarantees that they will use the same EventType. So, imagine that you build some postprocessing code to check that "FWD_BWD" meets some specification. If we don't have the EventType, the algorithm is not reliable.

Do you think that this is a strong enough argument to keep it? or would you say to remove it entirely?

torchtitan/observability/structured_logging.py
felipemello1: yes, i wanted to minimize the amount of changes that werent necessary, since the PRs were already so long.

torchtitan/experiments/observability/toy_spmd.py
felipemello1: MetricsProcessor calls set_step (and i guess a few other things, like clear_tags). Its just a convenience.

Ideally, you call set_step at the start of every step. Take a look at the toyrl example.

torchtitan/observability/structured_logging.py
felipemello1: do you mean 'int/normal/double/normvector'? That's for scuba

torchtitan/observability/structured_logging.py
felipemello1: what do you have in mind? You want the logger to show the callsite info?

torchtitan/observability/structured_logging.py
felipemello1: TIL

A ContextDecorator is a class-based utility that enables a ContextManager to also be used as a function decorator

image

torchtitan/observability/structured_logging.py
felipemello1: It protects against user error. If someone, for some reason, does:

```
logger = logging.getLogger("torchtitan.observability.system") # logger dedicated to structured logger
logger.info("hello")
```

because of this filter, it would correctly **NOT** be added to the jsonl.

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+        log_dict["log_function"] = record.funcName
+        log_dict["log_level"] = record.levelname
+        log_dict["logger_name"] = record.name
+        log_dict["stack_info"] = record.stack_info
+
+        log_dict["seq_id"] = next(self._seq_counter)
+
+        message = record.getMessage()
+        if message is not None:
+            if len(message) <= MAX_MESSAGE_SIZE:
```
joecummings, line 248: Why is there a MAX_MESSAGE_SIZE?

torchtitan/observability/structured_logging.py
joecummings: If I'm a user in the above scenario, I would be very confused. Should include a warning saying my message X will not be logged and how to properly log something to the system.

torchtitan/observability/structured_logging.py
joecummings: > That's for scuba

Can you say more? Is this *just* a scuba thing or is it the standard for other visualization/logging tools as well?

torchtitan/observability/structured_logging.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+class LogType(StrEnum):
+    """Distinguishes event records from free-text log records in JSONL."""
+
+    EVENT = "event"
+    TEXT = "text"
+
+
+class ExtraFields(StrEnum):
```
joecummings, line 84: Sorry, I just don't understand this one. Any more you can share?

torchtitan/observability/structured_logging.py
joecummings: If anything, I would strip this down to the bare minimum for shared Enums (fwd_bwd, step, optim). The rest could be project specific. So when I go to setup my RL workflow, I can *choose* to include some common RL event types I want to track. Or I can ignore them entirely. The point is the granularity is up to me.

torchtitan/experiments/observability/toy_spmd.py
joecummings: IIUC, you actually *wouldn't* be logging the metrics here - you're just doing it to show it's possible but it would actually be handled by `record_metric`? 

Otherwise, the conflation of the two is a little confusing.

torchtitan/experiments/observability/toy_spmd.py
joecummings: It would be clearer if this functionality was pushed outside of Metrics Processor - otherwise, like @tianyu-l, I'm not sure when I would use `set_step` directly or use `MetricProcessor.set_step`.

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -289,15 +309,24 @@ def train(self, num_steps):
 
         for step in range(1, num_steps + 1):
             self.step = step
-            tokens, labels, loss_mask = next(data_iterator)
-            self.train_step(tokens, labels, loss_mask)
+            self.metrics_processor.set_step(step)
+
+            # Simulate GC on every 5th step (mirrors gc_handler.run)
+            if step % 5 == 0:
+                add_step_tag("gc")
```
joecummings, line 316: q: Why wouldn't I use `record_span` to get a timing on how long gc takes?

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -289,15 +309,24 @@ def train(self, num_steps):
...
+
+            with record_span("trainer_time/step_s", EventType.STEP):
+                tokens, labels, loss_mask = next(data_iterator)
+                self.train_step(tokens, labels, loss_mask)
 
             if step % EVAL_FREQ == 0:
-                self.validate(tokens, labels, loss_mask)
+                add_step_tag("eval")
```
joecummings, line 323: q: Why does "eval" get a step tag *and* a record_span? Do these go to different things? If I record span, shouldn't that also tell me where I am in the step?

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -333,7 +365,10 @@ def main():
     trainer.close()
 
     if rank == 0:
-        print(f"Done. Output: {OUTPUT_DIR}")
+        sys_logs = os.path.join(OUTPUT_DIR, "system_logs")
```
joecummings, line 368: These would be nice as convenience methods - not sure if added later in the stack

torchtitan/observability/structured_logging.py
felipemello1: yeah, scuba expects these 4 columns. I do NOT know if its a standard for other logging tools. I think that we would have to understand that with time/user requests, or have a follow up trying to integrate DuckDB or something like that. Also, I think that i will probably need to enable handles using ENV args. E.g.\

```torchrun .... --enable_handler_scuba```

So we can have meta specific handlers.

torchtitan/observability/structured_logging.py
felipemello1: > Should include a warning saying my message X will not be logged and how to properly log something to the system.

You do NOT want to use this `logging.getLogger("torchtitan.observability.system")` for normal logger.info. This is something that is internal to `record_span`. This would only ever happen if an user really tried to make a mistake, and if that happened, it would be safe. The logger would be a normal logger, printing to console, and it would not damage the output jsonl. To log to jsonl, user has to pass the `"extras"` field, which `record_span` does.

To initialize regular logger in titan, we call `init_logger()`.

torchtitan/experiments/observability/toy_spmd.py
felipemello1: we should. We dot it in pr7. I dont know if i forgot to do it in the toy example.

`add_step_tag("gc")` is just a tag that will be added to very other log moving forward. Its a way to say: "hey, this step may be slower than normal. So, if you want to, you can filter out all steps with "gc" tag".

torchtitan/experiments/observability/toy_spmd.py
felipemello1: yes, `record_span` saves the time. `add_step_tag` will add a tag to any subsequent log, so we can know that that log was generated in a step that also had eval. Does this make sense?

torchtitan/experiments/observability/toy_spmd.py
felipemello1: yeah, i agree its confusing. In later PRs Tianyu also questioned it and i gave an example on how sixlib does it. I think we can iron it out a bit.

torchtitan/observability/structured_logging.py
felipemello1: Avoid spamming the logs

torchtitan/observability/structured_logging.py
tianyu-l: so this is from our offline sync: as long as we give user a way to do it, that's good enough, we don't have to do it ourselves

e.g. users could simply define strs as Literal / Enum so that they always use consistent str across framework.

torchtitan/observability/structured_logging.py
tianyu-l: is this too internal (less general, and leaking)?

torchtitan/observability/structured_logging.py
tianyu-l: we are importing https://github.com/pytorch/torchtitan/blob/main/torchtitan/tools/logging.py#L12 in every component, where we could have use `logging.getLogger(specific module)`

torchtitan/observability/structured_logging.py
tianyu-l: from offline: optimize for whatever makes the most sense, not least amount of changes.

PR #2603, opened 2026-03-16
[Observability 3/7] Experiment metrics with background aggregation subprocess

felipemello1:
* #2607
* #2606
* #2605
* #2604
* __->__ #2603
* #2602
* #2601
## How to review

Prioritize the files under /observability. Check how they are used in the toy example.
No need to nitpick the toy example. Real implementation in trainer are in PRs 7 and 8.

## Summary

The experiment metrics layer. Each rank calls `record_metric(key, MetricValue)` which serializes to per-rank experiment JSONL. Five metric types: `MeanMetric`, `MaxMetric`, `MinMetric`, `SumMetric`, `NoOpMetric`.

The logging subprocess (`logging_worker` in `aggregation.py`) runs as a `multiprocessing.Process`. The training process signals it via `queue.put((step, is_validation))` after each log step.

## Example

```python
record_metric("trainer_throughput/tps_mean", MeanMetric(sum=tps))
```

```json
// dump_dir/experiment_logs/trainer_rank_0_experiment.jsonl
{"key": "trainer_throughput/tps_mean", "reduce": "MeanMetric",
 "sum": 1234.5, "weight": 1.0,
 "step": 42, "rank": 0, "source": "trainer",
 "caller": "torchtitan/trainer.py:542:train_step", "timestamp": 1708200121.724}

// dump_dir/experiment_logs/trainer_rank_1_experiment.jsonl
{"key": "trainer_throughput/tps_mean", "reduce": "MeanMetric",
 "sum": 1180.2, "weight": 1.0,
 "step": 42, "rank": 1, "source": "trainer",
 "caller": "torchtitan/trainer.py:542:train_step", "timestamp": 1708200121.725}

// After aggregation: (1234.5 + 1180.2) / (1.0 + 1.0) = 1207.35
```

## Aggregation (Logging Subprocess)

The training process never reads JSONL or writes to backends directly.
A background subprocess on rank 0 handles all aggregation and output:

```
                  Training Process (all ranks)
                  ┌─────────────────────────────────┐
                  │  record_metric("loss", ...)     │──→ experiment.jsonl (per rank)
                  │                                 │
                  │                                 │
                  │  # on log steps:                │
                  │  barrier()                      │
                  │  log_queue.put(step) ───────────│──┐  (~0.1ms, non-blocking)
                  └─────────────────────────────────┘  │
                                                       │
                  Logging Subprocess (rank 0 only)     │
                  ┌─────────────────────────────────┐  │
                  │  step = queue.get() ◄───────────│──┘
                  │  read all experiment.jsonl files│
                  │  aggregate by key (reduce)      │
                  │  write to WandB / TensorBoard   │
                  │  print to console               │
                  └─────────────────────────────────┘
```

The subprocess is spawned via `logging_worker` from `aggregation.py`. Here's
the RL pattern where the controller owns the subprocess directly:

```python
import multiprocessing
from torchtitan.observability import logging_worker

# Spawn the logging subprocess
log_queue = multiprocessing.Queue()
log_process = multiprocessing.Process(
    target=logging_worker,
    args=(log_queue, OUTPUT_DIR),
    kwargs={
        "enable_wandb": True,
        "enable_tensorboard": False,
        "console_log_metric_keys": [
            "training/loss_mean",
            "training/grad_norm_max",
            "rl/reward_mean",
            "trainer_throughput/tps_mean",
        ],
    },
    daemon=True,
)
log_process.start()

# Each actor calls record_metric locally — writes go to JSONL
# ...

# After each step, signal the subprocess to read + aggregate + flush
log_queue.put((step, False))  # (step, is_validation)

# Shutdown
log_queue.put(None)
log_process.join()
```

`console_log_metric_keys` controls which metrics appear in the console output.
The subprocess prints one line per step with only the keys listed.

## Perf

**Timing (local filesystem, 100 metrics/rank):**

Aggregation benchmarks are measured on local filesystem without multithreading. NFS will be slower for the read, but shouldn't be an issue since it is non-blocking + multithreading.

| Scale | Read | Aggregate | Total |
|-------|------|-----------|-------|
| 10 files (1K entries) | 9ms | 0.5ms | 10ms |
| 100 files (10K entries) | 88ms | 6ms | 94ms |
| 500 files (50K entries) | 333ms | 35ms | 368ms |

None of this blocks training. Training pays only the signal cost (~0.1ms).

## Test plan

Run toy_spmd: `python -m torch.distributed.run --nproc_per_node=4 -m torchtitan.experiments.observability.toy_spmd`
Run toy_rl: `python -m torchtitan.experiments.observability.toy_rl`

- [x] 23 new unit tests for aggregation + metrics (67 cumulative)
- [x] Integration: toy_spmd produces experiment JSONL with aggregated loss/grad_norm/lr
- [x] Integration: toy_rl produces experiment JSONL from 4 actor types (controller, rollouter, reward, trainer)
- [x] Integration: logging subprocess writes to WandB and prints console output

**Sample**

```json
{"reduce": "MeanMetric", "sum": 5.5534299463033676e-05, "weight": 1.0, "key": "trainer_time/data_loading_s", "step": 1, "rank": 0, "source": "trainer", "caller": "torchtitan/experiments/observability/toy_spmd.py:240:batch_generator", "timestamp": 1773611398.1511042}
{"reduce": "MeanMetric", "sum": 3.14233530825004, "weight": 1.0, "key": "trainer_time/forward_backward_s", "step": 1, "rank": 0, "source": "trainer", "caller": "torchtitan/experiments/observability/toy_spmd.py:262:train_step", "timestamp": 1773611401.293699}
{"reduce": "MeanMetric", "sum": 0.15071861585602164, "weight": 1.0, "key": "trainer_time/optimizer_s", "step": 1, "rank": 0, "source": "trainer", "caller": "torchtitan/experiments/observability/toy_spmd.py:274:train_step", "timestamp": 1773611401.444696}
```

**Console output (toy_spmd 4 GPUs, 20 steps)**

```
[titan] 2026-03-15 14:50:01,477 - torchtitan.observability.aggregation - INFO - [31mstep:  1  [31mtraining/loss_mean: 3.66  [32mtraining/grad_norm_max: 0.49  [33mtraining/lr: 0.0010  [39m
[titan] 2026-03-15 14:50:01,602 - torchtitan.observability.aggregation - INFO - [31mstep:  5  [31mtraining/loss_mean: 3.053  [32mtraining/grad_norm_max: 0.41  [33mtraining/lr: 0.0010  [39m
[titan] 2026-03-15 14:50:02,196 - torchtitan.observability.aggregation - INFO - [31mstep: 10  [31mtraining/loss_mean: 2.54  [32mtraining/grad_norm_max: 0.32  [33mtraining/lr: 0.0010  [39m
[titan] 2026-03-15 14:50:02,360 - torchtitan.observability.aggregation - INFO - [31mstep: 15  [31mtraining/loss_mean: 2.23  [32mtraining/grad_norm_max: 0.26  [33mtraining/lr: 0.0010  [39m
[titan] 2026-03-15 14:50:02,523 - torchtitan.observability.aggregation - INFO - [31mstep: 20  [31mtraining/loss_mean: 2.026  [32mtraining/grad_norm_max: 0.22  [33mtraining/lr: 0.0010  [39m
[titan] 2026-03-15 14:50:02,523 - torchtitan.observability.aggregation - INFO - [logging process] shutting down
```

**Console output - toy_rl**

```
[titan] 2026-03-15 14:52:58,085 - torchtitan.observability.aggregation - INFO - [31mstep:  1  [31mtraining/loss_mean: 3.65  [32mtraining/grad_norm_max: 0.50  [33mtraining/lr: 0.0010  [34mrl/reward_mean: 1.0  [35mrl/completion_len_mean: 16.69  [39m
[titan] 2026-03-15 14:52:58,221 - torchtitan.observability.aggregation - INFO - [31mstep:  3  [31mtraining/loss_mean: 3.33  [32mtraining/grad_norm_max: 0.45  [33mtraining/lr: 0.0010  [34mrl/reward_mean: 1.0  [35mrl/completion_len_mean: 16.69  [39m
[titan] 2026-03-15 14:52:58,354 - torchtitan.observability.aggregation - INFO - [31mstep:  5  [31mtraining/loss_mean: 3.052  [32mtraining/grad_norm_max: 0.39  [33mtraining/lr: 0.0010  [34mrl/reward_mean: 1.0  [35mrl/completion_len_mean: 16.69  [39m
[titan] 2026-03-15 14:52:58,417 - torchtitan.observability.aggregation - INFO - [logging process] shutting down
```

### Output folder (adds experiment_logs/)

```
outputs/toy_spmd/
├── analysis/
│   └── system_metrics_gantt.json
├── experiment_logs/                    ← NEW in PR3
│   ├── trainer_rank_0_experiment.jsonl
│   ├── trainer_rank_1_experiment.jsonl
│   ├── trainer_rank_2_experiment.jsonl
│   └── trainer_rank_3_experiment.jsonl
├── system_logs/
│   └── trainer_rank_{0-3}_system.jsonl
└── wandb/
```

### WandB

toy_spmd: https://wandb.ai/cabernet-team/torchtitan/runs/pi8sosf6
toy_rl: https://wandb.ai/cabernet-team/torchtitan/runs/zcl58f75

### DIFF
--- a/torchtitan/experiments/observability/toy_rl.py
+++ b/torchtitan/experiments/observability/toy_rl.py
@@ -21,6 +21,7 @@ process that spawns GPU workers via Monarch's this_host().spawn_procs().
 
 import asyncio
 import logging
+import multiprocessing
 import os
 import shutil
 import socket
@@ -31,6 +32,7 @@ import torch
 from monarch.actor import Actor, current_rank, endpoint, this_host
 
 from torchtitan.distributed import ParallelDims
+
 from torchtitan.experiments.observability.toy_spmd import (
     BATCH_SIZE,
     DP_SIZE,
@@ -42,10 +44,14 @@ from torchtitan.experiments.observability.toy_spmd import (
 from torchtitan.observability import (
     EventType,
     init_observability,
+    logging_worker,
+    MeanMetric,
+    record_metric,
     record_span,
     set_step,
 )
 from torchtitan.observability.analysis import generate_gantt_trace
+from torchtitan.observability.metrics_processor import MetricsProcessor
 from torchtitan.tools.logging import init_logger
 
 logger = logging.getLogger(__name__)
@@ -146,6 +152,11 @@ class RollouterActor(Actor):
                 )
                 for i in range(len(self.dataset.tokens))
             ]
+        total_completion_len = sum(len(r.completion_text) for r in rollouts)
+        record_metric(
+            "rl/completion_len_mean",
+            MeanMetric(sum=total_completion_len, weight=len(rollouts)),
+        )
         return rollouts
 
 
@@ -176,7 +187,11 @@ class TrainerActor(Actor):
         )
         parallel_dims.build_mesh()
         self.dp_rank = parallel_dims.get_mesh("fsdp").get_local_rank()
-        self.trainer = ToyTrainer(self.device, parallel_dims, OUTPUT_DIR)
+        # Controller owns the logging subprocess — trainer has no backends/console.
+        mp_config = MetricsProcessor.Config(enable_wandb=False)
+        self.trainer = ToyTrainer(
+            self.device, parallel_dims, OUTPUT_DIR, mp_config=mp_config
+        )
 
     @endpoint
     async def set_step(self, step: int):
@@ -225,6 +240,10 @@ class RewardActor(Actor):
         with record_span("reward_time/scoring_s"):
             for rollout in rollouts:
                 rollout.reward = 1.0  # dummy constant reward
+        reward_sum = sum(r.reward for r in rollouts)
+        record_metric(
+            "rl/reward_mean", MeanMetric(sum=reward_sum, weight=len(rollouts))
+        )
         return rollouts
 
 
@@ -242,6 +261,26 @@ async def main():
 
     init_logger()
     init_observability(source="controller", output_dir=OUTPUT_DIR, rank=0)
+
+    # logging_worker reads experiment JSONL from all actors, aggregates
+    # metrics, and flushes to WandB/TB/console. Runs in a separate process.
+    log_queue = multiprocessing.Queue()
+    log_process = multiprocessing.Process(
+        target=logging_worker,
+        args=(log_queue, OUTPUT_DIR),
+        kwargs={
+            "enable_wandb": True,
+            "console_log_metric_keys": [
+                "training/loss_mean",
+                "training/grad_norm_max",
+                "training/lr",
+                "rl/reward_mean",
+                "rl/completion_len_mean",
+            ],
+        },
+        daemon=True,
+    )
+    log_process.start()
 
     # ---- Setup ----
     host = this_host()
@@ -285,13 +324,13 @@ async def main():
             with record_span("rl_time/training_s", EventType.FWD_BWD):
                 await trainer.train_step.call(tokens, labels, loss_mask)
 
-            rewards = [r.reward for r in rollouts]
-            reward_mean = sum(rewards) / len(rewards)
-            logger.info(f"step: {step}  reward_mean: {reward_mean:.5f}")
+            log_queue.put(step)
 
     await run_training()
 
     # ---- Cleanup ----
+    log_queue.put(None)
+    log_process.join(timeout=10)
     await trainer.teardown.call()
 
     sys_logs = os.path.join(OUTPUT_DIR, "system_logs")
--- a/torchtitan/experiments/observability/toy_spmd.py
+++ b/torchtitan/experiments/observability/toy_spmd.py
@@ -38,7 +38,10 @@ from torchtitan.observability import (
     add_step_tag,
     EventType,
     init_observability,
+    MaxMetric,
+    NoOpMetric,
     record_event,
+    record_metric,
     record_span,
 )
 from torchtitan.observability.analysis import generate_gantt_trace
@@ -57,6 +60,7 @@ BATCH_SIZE = 8
 DP_SIZE = 2
 LR = 1e-3
 IGNORE_INDEX = -100
+ENABLE_WANDB = True
 OUTPUT_DIR = os.path.join(os.path.dirname(__file__), "outputs", "toy_spmd")
 
 
@@ -155,7 +159,9 @@ class ToyTrainer:
     - train: owns the training loop
     """
 
-    def __init__(self, device, parallel_dims: ParallelDims, output_dir):
+    def __init__(
+        self, device, parallel_dims: ParallelDims, output_dir, *, mp_config=None
+    ):
         self.device = device
         self.parallel_dims = parallel_dims
         dp_mesh = parallel_dims.get_mesh("fsdp")
@@ -165,7 +171,11 @@ class ToyTrainer:
         self.output_dir = output_dir
         self.step = 0
 
-        self.metrics_processor = MetricsProcessor()
+        if mp_config is None:
+            mp_config = MetricsProcessor.Config()
+        self.metrics_processor = mp_config.build(
+            parallel_dims=parallel_dims, dump_folder=output_dir
+        )
 
         torch.manual_seed(0)
         with record_span("setup/model_build", EventType.BUILD_MODEL):
@@ -277,14 +287,12 @@ class ToyTrainer:
         dist.all_reduce(
             loss_scalar, op=dist.ReduceOp.SUM, group=self.dp_mesh.get_group()
         )
+        record_metric("training/loss_mean", NoOpMetric(value=loss_scalar.item()))
+        record_metric("training/grad_norm_max", MaxMetric(value=grad_norm.item()))
+        record_metric("training/lr", NoOpMetric(value=LR))
         record_event(
             {"train.loss": loss_scalar.item(), "train.grad_norm": grad_norm.item()}
         )
-        if self.rank == 0:
-            print(
-                f"step: {self.step}  loss: {loss_scalar.item():.5f}  "
-                f"grad_norm: {grad_norm.item():.5f}"
-            )
 
     def validate(self, tokens, labels, loss_mask):
         """Run one forward pass for validation (no backward)."""
@@ -298,8 +306,7 @@ class ToyTrainer:
         dist.all_reduce(
             val_loss_scalar, op=dist.ReduceOp.SUM, group=self.dp_mesh.get_group()
         )
-        if self.rank == 0:
-            print(f"  val loss: {val_loss_scalar.item():.4f}")
+        record_metric("validation/loss_mean", NoOpMetric(value=val_loss_scalar.item()))
 
     def train(self, num_steps):
         """Full training loop. Mirrors Trainer.train structure."""
@@ -323,6 +330,8 @@ class ToyTrainer:
                 add_step_tag("eval")
                 with record_span("trainer_time/validation_s", EventType.EVAL):
                     self.validate(tokens, labels, loss_mask)
+
+            self.metrics_processor.log(step)
 
     def close(self):
         """Cleanup."""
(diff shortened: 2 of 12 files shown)


torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -277,14 +287,12 @@ def train_step(self, tokens, labels, loss_mask):
...
+        record_metric("training/grad_norm_max", MaxMetric(value=grad_norm.item()))
+        record_metric("training/lr", NoOpMetric(value=LR))
         record_event(
             {"train.loss": loss_scalar.item(), "train.grad_norm": grad_norm.item()}
         )
-        if self.rank == 0:
-            print(
-                f"step: {self.step}  loss: {loss_scalar.item():.5f}  "
-                f"grad_norm: {grad_norm.item():.5f}"
-            )
```
felipemello1, line 287: no need for printing, since now we can collect metrics an log to console.

torchtitan/observability/__init__.py
```diff
@@ -4,12 +4,17 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-"""
-TorchTitan Observability Library
-
-System metrics: init_observability, set_step, record_span, record_event, EventType
-"""
+"""TorchTitan Observability Library."""
```
felipemello1, line 7: meh, this is a bad docstring. I can delete it later.

torchtitan/observability/aggregation.py
```diff
@@ -0,0 +1,348 @@
...
+    TensorBoardLogger,
+    WandBLogger,
+)
+from torchtitan.observability.metrics import REDUCE_REGISTRY
+from torchtitan.observability.structured_logging import init_observability
+from torchtitan.tools.utils import Color
+
+logger = logging.getLogger(__name__)
+
+_QUEUE_TIMEOUT_S = 600  # 10 minutes — if no signal, assume training crashed
```
felipemello1, line 31: unsure about this. Basically if we dont try to flush anything in 10min, we kill this actor. This is bad if some collective takes >10min, e.g. checkpointing. I put this because if for some reason the parent process dies, then there is no way to kill this child process.

torchtitan/experiments/observability/toy_spmd.py
tianyu-l, line 271: Can we model this as a `MeanMetric`? Not necessarily for logging, but in general it seems a useful abstraction.

torchtitan/observability/aggregation.py
```diff
@@ -0,0 +1,348 @@
...
+    temp = frac
+    while n_decimals < 5 and n_nonzero < 2:
+        n_decimals += 1
+        temp *= 10  # e.g. 0.00123 → 0.0123 → 0.123 → 1.23
+        if int(temp) % 10 != 0 or n_nonzero > 0:
+            n_nonzero += 1
+    return f"{value:.{max(n_decimals, 2)}f}"
+
+
+def _log_to_console(
```
tianyu-l, line 186: what's the relationship between this and `logger.info` call in torchtitan?

torchtitan/experiments/observability/toy_rl.py
```diff
@@ -243,6 +262,26 @@ async def main():
     init_logger()
     init_observability(source="controller", output_dir=OUTPUT_DIR, rank=0)
 
+    # logging_worker reads experiment JSONL from all actors, aggregates
```
tianyu-l, line 265: I'm confused by why you have both `MetricsProcessor` and this multiprocessing of `logging_worker`. It seems that
- metrics processor is more object-oriented, and it can start a subprocess with `logging_worker`
- but the API is more rigid in that you have to pass it around -- I thought the point is to be able to log anywhere, so is the end goal removing `MetricsProcessor`?

torchtitan/experiments/observability/toy_spmd.py
felipemello1: Do you mean a stateful class to help with keeping track of loss and valid tokens, but without` .detach` and `.item()`, so we can still run `.backward()` on it? If so, we could, but i think that it should probably be in a separate PR

torchtitan/observability/aggregation.py
felipemello1: Titan currently logs key metrics to console on every .log call: https://github.com/pytorch/torchtitan/blob/0187d5f992d4edbaf5d350cd5efc154a2bdbcaf0/torchtitan/components/metrics.py#L526

Maybe it should rename to make it more explicit, e.g. "display_key_metrics_on_console"

Functionality is the same, I just turned it into a function, and let the user pass arbitrary keys that they want to monitor on the console. I needed it because 'eval' and 'train' have different keys.

(from pr4) 
image

torchtitan/experiments/observability/toy_rl.py
felipemello1: >I'm confused by why you have both MetricsProcessor and this multiprocessing of logging_worker

I think that this becomes clearer in the follow up PRs, but MetricsProcessor (MP) has multiple methods that help the trainer: 'set_step', 'log', 'record_throughput', 'record_memory', 'reset_counters'. We abstract all of that away to MP, so the trainer is cleaner.

Then, when we call .log(), we trigger the logging_worker.

> but the API is more rigid in that you have to pass it around

The MP is just a helper for the trainer. The rest of logging system is independent. You can call 'record_metric' or 'record_span' from anywhere. No need to "pass the MP around".

Let me know if this makes sense or not.

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -324,6 +331,8 @@ def train(self, num_steps):
                 with record_span("trainer_time/validation_s", EventType.EVAL):
                     self.validate(tokens, labels, loss_mask)
 
+            self.metrics_processor.log(step)
```
joecummings, line 334: is this the same as "set_step"?

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -360,7 +369,15 @@ def main():
     if rank == 0:
         print(f"Toy SPMD: {world_size} GPUs, 2DPx2TP, {NUM_STEPS} steps")
 
-    trainer = ToyTrainer(device, parallel_dims, OUTPUT_DIR)
+    mp_config = MetricsProcessor.Config(
+        enable_wandb=ENABLE_WANDB,
+        console_log_metric_keys=[
```
joecummings, line 374: IIUC these are the keys you want to be logged also to the console?

torchtitan/experiments/observability/toy_spmd.py
```diff
@@ -277,14 +287,12 @@ def train_step(self, tokens, labels, loss_mask):
         dist.all_reduce(
             loss_scalar, op=dist.ReduceOp.SUM, group=self.dp_mesh.get_group()
         )
+        record_metric("training/loss_mean", NoOpMetric(value=loss_scalar.item()))
+        record_metric("training/grad_norm_max", MaxMetric(value=grad_norm.item()))
+        record_metric("training/lr", NoOpMetric(value=LR))
         record_event(
```
joecummings, line 293: I see you're still logging the loss and grad_norm to system metrics along with the record metric. I don't understand why you would need both and the difference between event and metric is blurred. Perhaps this is just a misunderstanding?

torchtitan/experiments/observability/toy_spmd.py
felipemello1: no, 

'set_step' will add the step to every log

`.log' will aggregate and flush all metrics from a given step

torchtitan/experiments/observability/toy_spmd.py
felipemello1: will make it more explicit, good feedback!

PR #2605, opened 2026-03-16
[Observability 5/7] RolloutLogger for RL rollout JSONL with filtering

felipemello1:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #2607
* #2606
* __->__ #2605
* #2604
* #2603
* #2602
* #2601

## How to review

Prioritize the files under /observability. Check how they are used in the toy example. No need to nitpick the toy example.

## Summary

`RolloutLogger` writes RL rollouts to JSONL with optional filtering. Takes a `list[dict]` with no enforced schema — users pass whatever fields their rollouts have (prompt, response, reward, etc). Metadata is nested under `__metadata__` to avoid key collisions.

Optional `filter_fn` for selecting which rollouts to log, e.g. `filter_top_bottom(records, key, k)` keeps the top-k and bottom-k by a given key — useful for logging only the best and worst rollouts per batch.

Integrated into toy_rl: the controller logs filtered rollouts after each RL step.

## Example

```python
from torchtitan.observability import RolloutLogger

rollout_logger = RolloutLogger(
    output_dir="outputs/rollouts",
    filter_fn=lambda records: filter_top_bottom(records, key="reward", k=2),
)

rollouts = [
     {"prompt": "what is 2+2?", "response": "4", "reward": 0.95},
     {"prompt": "write a poem", "response": "roses...", "reward": 0.3}
     ]

# Log rollout dicts with optional metadata
rollout_logger.log(
    rollouts,
    metadata={"step": 42, "batch_size": 64},
)
rollout_logger.close()
```

Output JSONL (metadata stored under `__metadata__` to avoid key collisions):
```json
{"prompt": "what is 2+2?", "response": "4", "reward": 0.95, "__metadata__": {"step": 42, "batch_size": 64}}
{"prompt": "write a poem", "response": "roses...", "reward": 0.3, "__metadata__": {"step": 42, "batch_size": 64}}
```

## Test plan

Run toy_rl: `python -m torchtitan.experiments.observability.toy_rl`

- [x] New unit tests for RolloutLogger: write, filter, metadata nesting, close
- [x] Integration: toy_rl produces rollouts.jsonl with filtered entries
- [x] Integration: rollouts have __metadata__ nesting with step info

### Output folder (toy_rl on PR6 — note rollouts/ directory)

```
outputs/toy_rl/
├── analysis/
│   └── system_metrics_gantt.json
├── experiment_logs/
│   ├── controller_rank_0_experiment.jsonl
[...]

PR #2604, opened 2026-03-16
[Observability 4/7] MetricsProcessor — throughput, memory, validation, log scheduling

felipemello1:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #2607
* #2606
* #2605
* __->__ #2604
* #2603
* #2602
* #2601

## How to review

Prioritize the files under /observability. Check how they are used in the toy example.
No need to nitpick the toy example. Real implementation in trainer are in PRs 6 and 7.

## Summary

(Almost) Full `MetricsProcessor` for the toy trainer. This is the reference implementation that we will mirror for the real Trainer in PRs 6 and 7.

### APIs

- `should_log(step)` — returns True every N steps, on step 1, or when `force_log` is set (for validation).
- `reset_training_counters()` / `reset_val_counters()` — reset timing and token counts.
- `record_throughput()` — computes tokens/sec/device, TFLOPS, and MFU from tokens since last reset.
- `record_memory()` — records GPU peak stats (reserved/active GiB, pct, alloc retries, OOMs).
- `log(step, is_validation)` — signals the logging subprocess to aggregate and flush.

### Validation support

`MetricsProcessor.Config` adds `console_log_validation_keys`.  When `is_validation=True` is passed to `log()`, the subprocess prints a separate validation console line.

## Test plan

Run toy_spmd: `python -m torch.distributed.run --nproc_per_node=4 -m torchtitan.experiments.observability.toy_spmd`
Run toy_rl: `python -m torchtitan.experiments.observability.toy_rl`

- [x] 13 new unit tests for EveryNSteps + logging boundary (80 cumulative)
- [x] Integration: toy_spmd with log_freq=5 only logs on steps {1, 5, 10, 15, 20}
- [x] Integration: validation console line appears on eval steps
- [x] Integration: throughput and memory metrics appear in WandB

### Console output (toy_spmd)

With `log_freq=5`, metrics only appear on steps {1, 5, 10, 15, 20}.
Validation runs on steps 10 and 20 (every 10 steps):

```
[titan] 2026-03-15 14:50:51,749 - torchtitan.observability.aggregation - INFO - step:  1  training/loss_mean: 3.66  training/grad_norm_max: 0.49  trainer_memory/reserved_gib_max: 0.068  trainer_throughput/tps_mean: 11.86  
[...]

PR #2606, opened 2026-03-16
[Observability 6/7] Wire observability into Trainer, Flux, FT, and Forge

felipemello1:
* #2607
* __->__ #2606
* #2605
* #2604
* #2603
* #2602
* #2601

## How to review

Understand the changes in trainer.py. The changes in other trainers (Flux, FT, Forge) are exactly the same.

## Summary

Wires the observability library into TorchTitan's production Trainer and all derived trainers (Flux, FT, Forge).

[x] (This PR) 1st phase - Focuses on 'record_metrics' and excludes 'record_span' and 'record_event'.
[ ] (Next PR) 2nd phase - Adds 'record_span', 'add_step_tag' and 'record_event'.

The old `MetricsProcessor` in `components/metrics.py` is replaced by a new one in `observability/metrics_processor.py`. The old class (~327 lines) is deleted.

**Key improvements over the old MetricsProcessor:**

- **Real loss visible on rank 0 with PP.**: The old code showed sentinel loss (-2.0) on rank 0 because only the last PP stage computed loss. It also had 2 extra functions to assist with that (deleted). 

- **Zero backend overhead on the hot path.** Logging to backends happens in background subprocess

- **Decoupled from the training loop.** Metrics are recorded via `record_metric(key, value)` at the call site. No more passing dicts of extra_metrics between functions.

- **Modularization**: Better separation between training and validation metrics.

**Trainers migrated:**

All four trainers follow the same pattern. Flux and FT inherit `batch_generator` from base Trainer. Forge is standalone.

The changes are the same in all four.

## Numerics

Loss and grad_norm match main exactly on all trainers with `--debug.seed=42 --debug.deterministic`:

| Trainer | Config | Step 1 loss | Step 5/10 loss |
|---------|--------|-------------|----------------|
| Base | PP=2 DP=2 TP=2 | 8.028 | 3.90 (step 10) |
| Flux | DP=4 CP=2 | 1.158 | 1.160 (step 5) |
| Forge | DP=2 TP=2 PP=2 | 8.028 | 5.19 (step 5) |
| FT | — | Not runnable (pre-existing bug on main: torchft renamed `ProcessGroupNCCL` → `ProcessGroupBabyNCCL`, `experiments/ft/manager.py:95` needs updating) |

## Test plan

```bash
# Base trainer (PP=2 DP=2 TP=2)
NGPU=8 LOG_RANK=0 ./run_train.sh --module llama3 --config llama3_debugmodel \
  --training.steps 10 --parallelism.data_parallel_shard_degree=2 \
  --parallelism.tensor_parallel_degree=2 --parallelism.pipeline_parallel_degree=2 \
  --parallelism.pipeline_parallel_schedule=1F1B --debug.seed=42 --debug.deterministic \
  --metrics.enable-wandb

# Flux (DP=4 CP=2)
NGPU=8 LOG_RANK=0 ./run_train.sh --module flux --config flux_debugmodel \
  --training.steps 5 --debug.seed=42 --metrics.enable-wandb \
  --parallelism.data_parallel_shard_degree=4 --parallelism.context_parallel_degree=2 \
  --tokenizer.test_mode --encoder.random_init \
  --encoder.clip_encoder tests/assets/flux_test_encoders/clip-vit-large-patch14/ \
  --encoder.t5_encoder tests/assets/flux_test_encoders/t5-v1_1-xxl/ \
  --tokenizer.t5_tokenizer_path tests/assets/tokenizer \
  --tokenizer.clip_tokenizer_path tests/assets/tokenizer \
  --hf_assets_path tests/assets/tokenizer

# Forge (DP=2 TP=2 PP=2)
python -m torch.distributed.run --nproc_per_node=8 --local-ranks-filter 0 --role rank --tee 3 \
  -m torchtitan.experiments.forge.example_train --module llama3 --config llama3_debugmodel \
  --training.steps 5 --debug.seed=42 --debug.deterministic --metrics.enable-wandb \
  --parallelism.data_parallel_shard_degree=2 --parallelism.tensor_parallel_degree=2 \
  --parallelism.pipeline_parallel_degree=2 --parallelism.pipeline_parallel_schedule=1F1B

```

- [x] pre-commit passes
- [x] Integration: base trainer PP=2 DP=2 TP=2 loss matches main
- [x] Integration: console on rank 0 shows real loss (not PP sentinel -2.0)
- [x] Integration: Flux DP=4 CP=2 loss matches main
- [x] Integration: Forge DP=2 TP=2 PP=2 loss matches main (shows real loss, not -2.0)

### Console output — before (main) vs after

**Main — Base trainer** (rank 0, PP=2 — shows sentinel loss -2.0):
```
[rank0]:[titan] 2026-03-16 08:53:41,731 - root - INFO - step:  1  loss: -2.00000  grad_norm: 50810.4961  memory:  0.19GiB(0.20%)  tps: 421  tflops: 0.03  mfu: 0.00%
[rank0]:[titan] 2026-03-16 08:53:44,233 - root - INFO - step:  5  loss: -2.00000  grad_norm: 83210.2031  memory:  0.19GiB(0.20%)  tps: 11,054  tflops: 0.79  mfu: 0.08%
```

**This PR — Base trainer** (rank 0 — real aggregated loss):
```
[rank0]:[titan] 2026-03-16 08:57:13,807 - torchtitan.observability.aggregation - INFO - step:  1  training/loss_mean: 8.028  training/grad_norm_max: 50810.5  trainer_memory/reserved_gib_max: 0.19  trainer_throughput/tps_mean: 361.4  trainer_throughput/tflops_mean: 0.026  trainer_throughput/mfu_pct_mean: 0.0026
[rank0]:[titan] 2026-03-16 08:57:17,135 - torchtitan.observability.aggregation - INFO - step: 10  training/loss_mean: 3.90  training/grad_norm_max: 52873.5  trainer_memory/reserved_gib_max: 0.20  trainer_throughput/tps_mean: 11729.9  trainer_throughput/tflops_mean: 0.84  trainer_throughput/mfu_pct_mean: 0.085
```

### Output folder
```
outputs/
├── system_logs/
│   └── trainer_rank_{0-7}_system.jsonl
├── experiment_logs/
│   └── trainer_rank_{0-7}_experiment.jsonl
├── tb/20260316-0857/
└── wandb/
```

### WandB runs

| | Main | This PR |
|---|------|------|
| Base (PP=2 DP=2 TP=2) | https://wandb.ai/cabernet-team/torchtitan/runs/9ltdu69z | https://wandb.ai/cabernet-team/torchtitan/runs/uxcx428m |
| Flux (DP=4 CP=2) | https://wandb.ai/cabernet-team/torchtitan/runs/twfvo2ys | https://wandb.ai/cabernet-team/torchtitan/runs/iua7gyq5 |
| Forge (DP=2 TP=2 PP=2) | https://wandb.ai/cabernet-team/torchtitan/runs/35ztpneg | https://wandb.ai/cabernet-team/torchtitan/runs/o1pqmpeb |

Trainer wandb for this PR
image


torchtitan/components/metrics.py
```diff
@@ -195,388 +187,3 @@ def number_of_loggers(self) -> int:
     def close(self) -> None:
         for logger_instance in self._loggers:
             logger_instance.close()
-
-
-def ensure_pp_loss_visible(
```
felipemello1, line 200: this and '_get_metrics_rank' are no longer needed. For PP loss, we would need the head of pp rank to log. Now, we only log on rank0.

Q: What if rank 0 is NOT the head of pp?
A: Thats ok, because we are reading the metrics from logged json files. As long as the correct rank logged, then rank 0 can parse it.

torchtitan/components/metrics.py
```diff
@@ -195,388 +187,3 @@ def number_of_loggers(self) -> int:
...
-    if pp_schedule == "ZBVZeroBubble":
-        return 0
-
-    # Calculate first rank of the last pipeline stage
-    world_size = parallel_dims.world_size
-    pp_size = parallel_dims.pp
-    return (world_size // pp_size) * (pp_size - 1)
-
-
-class MetricsProcessor(Configurable):
```
felipemello1, line 261: moved to /observability. I was going to keep it here so its easier to compare/review, but i was having circular dependencies, so i just moved it to a new file.

torchtitan/trainer.py
```diff
@@ -744,48 +739,63 @@ def train_step(
...
+                            loss_mesh,
+                        ),
+                    )
+                else:
+                    global_avg_loss = global_max_loss = loss.detach().item()
+                    global_ntokens_seen = self.ntokens_seen
+
+                # Only record loss on ranks that computed it (last PP stage has
+                # the real loss; other PP stages have a -1/-2 sentinel).
+                if not parallel_dims.pp_enabled or self.pp_has_last_stage:
```
felipemello1, line 776: we only log it for the right ranks, since 'record_metric' is not aware about which ranks to use to reduce later on.

torchtitan/observability/aggregation.py
```diff
@@ -242,13 +242,17 @@ def _build_metric_logger(
     save_tb_folder: str = "tb",
     config_dict: dict[str, Any] | None = None,
     tag: str | None = None,
+    ft_enable: bool = False,
+    ft_replica_id: int = 0,
```
tianyu-l, line 246: main trainer shouldn't be aware of these fault tolerant fields

torchtitan/models/flux/trainer.py
```diff
@@ -251,36 +255,43 @@ def train_step(
...
+
+                record_metric("training/loss_mean", NoOpMetric(value=global_avg_loss))
+                record_metric("training/loss_max", NoOpMetric(value=global_max_loss))
+                record_metric(
+                    "training/grad_norm_max", NoOpMetric(value=grad_norm.item())
+                )
+                record_metric(
+                    "training/n_tokens_sum", NoOpMetric(value=global_ntokens_seen)
+                )
+                record_event(
```
tianyu-l, line 288: why they are both metrics and events?

torchtitan/observability/metrics_processor.py
```diff
@@ -29,11 +28,43 @@
...
-    and writes to WandB/TB/console.
+    """Metrics lifecycle for distributed training.
+
+    Handles three responsibilities:
+    1. Step tracking and log scheduling (every N steps + validation steps).
+    2. Derived metrics: GPU memory peak stats and throughput (TPS, TFLOPS, MFU).
+    3. Background logging subprocess on rank 0 that reads per-rank experiment
+       JSONL, aggregates across ranks, and writes to WandB/TB/console.
+
+    The trainer calls methods in this order each step:
```
tianyu-l, line 39: orders of record memory and record throughput should be interchangeable?

torchtitan/trainer.py
```diff
@@ -372,12 +377,6 @@ def __init__(self, config: Config):
                     cast(Decoder, m).init_weights(buffer_device=buffer_device)
                 m.train()
 
-            # confirm that user will be able to view loss metrics on the console
-            ensure_pp_loss_visible(
```
tianyu-l, line 376: yes, we should remove this

torchtitan/trainer.py
```diff
@@ -530,22 +526,20 @@ def batch_generator(
...
 
         while True:
-            data_load_start = time.perf_counter()
-            try:
-                batch = next(data_iterator)
-            except StopIteration as ex:
-                # If data runs out during gradient accumulation, that
-                # entire step will not be executed.
-                raise DataloaderExhaustedError() from ex
+            with record_span("trainer_time/data_loading_s", EventType.FETCHING_BATCH):
```
tianyu-l, line 529: what does _s mean? seconds?

torchtitan/trainer.py
tianyu-l: if some ranks are "missing" what happens? How do you guard against the case where all ranks should log something but some ranks didn't?

torchtitan/trainer.py
```diff
@@ -280,18 +287,16 @@ def __init__(self, config: Config):
...
-        color = self.metrics_processor.color
-
         # calculate model size and flops per token
         (
             model_param_count,
             self.metrics_processor.num_flops_per_token,
         ) = model_config.get_nparams_and_flops(model, config.training.seq_len)
 
+        color = utils.Color()
         logger.info(
```
tianyu-l, line 300: This is still directly logging to console right?

tianyu-l (commented):
Several things to clarify:

relationship and separation between `MetricsProcessor` / `RolloutLogger` and directly calling `record_metrics` -- Do we need an object only when something needs to be stateful for derived metrics? Do we need to start `logging_worker` process for both?

relationship between record_span, record_metrics, record_event. It seems
- record_span would log the span to record_metrics
- some metrics are also logged to record_event

(local loss, num_local_tokens) sounds a natural `MeanMetric`, but right now is not modeled that way.

relationship between MetricsProcessor and logger.info. It seems some console logging is handled by `console_log_metric_keys` but not all of them are

torchtitan/trainer.py
felipemello1: > if some ranks are "missing" what happens? 

The aggregation function is simple:
1. read json from all ranks
2. aggregate by metric_name and step

if rank 3 and 6 don't post the metric, they won't be part of the aggregation.

So this is enough:
```
if not parallel_dims.pp_enabled or self.pp_has_last_stage:
    record_metric(....)
```

>How do you guard against the case where all ranks should log something but some ranks didn't?

This would be a tricky bug.  Currently:
- We put a barrier before .log() in the MP, to make sure that all metrics made it to the json before we call aggregate.
- We have a small time.sleep to make sure that the logger had time to flush.
- We always open the file (instead of keeping it open), and claude mentioned something about that helping with having a fresh cache.

So i think we are safe. We could maybe force logger.flush as well before aggregation.

torchtitan/observability/aggregation.py
felipemello1: I copied it from what we have today. Happy to change, but I dont have much context about ft. What do you have in mind?
https://github.com/pytorch/torchtitan/blob/0187d5f992d4edbaf5d350cd5efc154a2bdbcaf0/torchtitan/components/metrics.py#L327-L328

torchtitan/models/flux/trainer.py
felipemello1: this is something that also bothered me a bit. 

This is the original API's docstring.
image

and its used like this in the trainer by sixlib
image

---

I think I could rename 'record_event' so its easier to grasp the difference. Maybe `record_system_event`, I am not sure.

torchtitan/observability/metrics_processor.py
felipemello1: i purposely put record throughput last so that any slow down from previous functions would be captured. But I assume that record_memory is mostly a O(1) lookup. Do you think its worth the edit?

torchtitan/trainer.py
felipemello1: Yes. Happy to change the pattern to make it more explicit if you think its not intuitive.

torchtitan/trainer.py
felipemello1: yes! logger.info is the standard logger. If we intentionally pass some `extra` field, then our filters may act on them. But otherwise, it goes to the console normally.

Image

felipemello1: i just double checked 

Image

torchtitan/trainer.py
tianyu-l: I think internal broadcast PP loss to all PP ranks. I would rather we do the same, instead of relying on "only saving to some ranks". But this is not urgent.

torchtitan/observability/aggregation.py
tianyu-l: We can have extension points, but we shouldn't have them on core torchtitan. I don't mind we remove them entirely for now.

torchtitan/trainer.py
tianyu-l: fine to have, but please make it consistent across all metrics -- I saw some metrics don't have the unit

torchtitan/trainer.py
tianyu-l: > If we intentionally pass some extra field, then our filters may act on them. But otherwise, it goes to the console normally.

This is not accurate? `logger` is not something we register handler to, so even if we put extra field, it won't be "acted on".

torchtitan/models/flux/trainer.py
tianyu-l: I'm sure chatgpt/claude can give us some inspiration :)

PR #2608, opened 2026-03-16
[Module][2/2] Convert remaining nn.Module classes to Module protocol -- 

fegin:
* __->__ #2608

    **Motivation**
    The previous PR, https://github.com/pytorch/torchtitan/pull/2565, doesn't fully convert all the nn.Module to Module. This PR converts the leftover.

    **Design Summary**

    1. Introduce create_class_module so that users can easily create a
       Module class for a nn.Module class For exasmple, users can do `
       Conv2d = create_class_module(nn.Conv2d)` to get a new Cov2d Module
       class that inherits from nn.Conv2d and Module.

    2. Introduce verify_all_module_protocol method to BaseModel to check
       if all sub-nn.Module are Module.

    3. Create container Module classes, nn.ModuleDict, nn.ModuleList,
       nn.Sequence.


torchtitan/protocols/module.py
```diff
@@ -21,3 +23,102 @@ class Module(nn.Module, Configurable):
     def init_weights(self, **kwargs) -> None:
         """Initialize weights. Override in subclasses with learnable parameters."""
         pass
+
+
+class ModuleList(nn.ModuleList, Module):
```
tianyu-l: How necessary is this change?

I can imagine that maybe it's helpful to make the input/output shardings configurable. But for init_weights, it's actually not used https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder.py#L120-L122?

Or is it just for consistency among all modules.

torchtitan/protocols/module.py
```diff
@@ -21,3 +23,102 @@ class Module(nn.Module, Configurable):
...
+# the process lifetime.
+_created_classes: dict[type, type] = {}
+
+
+def _init_weights_from_reset_parameters(self: Any, **kwargs: Any) -> None:
+    """``init_weights`` implementation that delegates to ``reset_parameters``."""
+    self.reset_parameters()
+
+
+def create_module_class(nn_module_cls: type[nn.Module]) -> type[Module]:
```
tianyu-l:
```suggestion
def from_nn_module(nn_module_cls: type[nn.Module]) -> type[Module]:
```
also
- can we make this a class method of `Module`?
- the return type should be both Module and nn.Module? (not sure if python supports it)

torchtitan/protocols/module.py
```diff
@@ -21,3 +23,102 @@ class Module(nn.Module, Configurable):
...
+
+# Cache: maps nn.Module subclass -> created Module wrapper class.
+# Module classes are typically created at import time and live for
+# the process lifetime.
+_created_classes: dict[type, type] = {}
+
+
+def _init_weights_from_reset_parameters(self: Any, **kwargs: Any) -> None:
+    """``init_weights`` implementation that delegates to ``reset_parameters``."""
+    self.reset_parameters()
```
tianyu-l: can this be an inline function?

torchtitan/protocols/module.py
```diff
@@ -21,3 +23,102 @@ class Module(nn.Module, Configurable):
...
+class Sequential(nn.Sequential, Module):
+    """Module-protocol-compatible version of ``nn.Sequential``.
+
+    ``init_weights`` recursively calls ``init_weights`` on each child.
+    """
+
+    def init_weights(self, **kwargs) -> None:
+        for child in self:
+            assert isinstance(child, Module)
+            child.init_weights(**kwargs)
```
tianyu-l: can we put this logic into `create_module_class` so that we don't have to define `class Sequential`, etc.?

torchtitan/protocols/model.py
```diff
@@ -20,6 +20,14 @@ class BaseModel(Module):
     All models must implement ``init_weights`` (from Module).
     """
 
+    def verify_module_protocol(self) -> None:
```
tianyu-l, line 23: how necessary is this function?

If it's necessary, please add why to the docstring.

torchtitan/protocols/module.py
fegin: Once we change `init_states`/`init_wegiths` to be recursive, this change will be helpful. Otherwise, we need to special handle the container type in `init_states`, which is less clean.

I put the change here, not next PR because of the consistency. But I think it will be better to ensure all nn.Module are Module, even for container types.

torchtitan/protocols/module.py
fegin: > can we make this a class method of Module?

sure 

> the return type should be both Module and nn.Module? (not sure if python supports it)

~~why? Module inherits nn.Module, so Module is enough?~~

If you are talking about `type[Module & _T]` where `_T` is `nn_module_cls: _T`, then no, Python currently doesn't support type intersection.

torchtitan/protocols/model.py
fegin: > how necessary is this function?

It depends. This is a guardrail so that if a developer adds a new model/module that contains plain nn.Module, the developer can get explicit error message. Otherwise, the error can be very random, like not defining `init_states` etc.

torchtitan/protocols/module.py
fegin: ~~This is a very specific logic. Only containers will do this. While the lines look very similar, they are actually different -- ModuleList and Sequential are the same but are different from ModuleDict and are different from other nn.Module.~~ 

So this can be done through `children()`. So this is not an issue.

Another motivation of having Sequential, ModuleList, ModuleDict explicitly defined is that these are commonly used containers and more than one model is using these. 

Typing is also an issue as type intersection is not allowed in Python.

torchtitan/protocols/module.py
```diff
@@ -21,3 +28,66 @@ class Module(nn.Module, Configurable):
...
+
+
+def _container_init_weights(self: "Module", **kwargs: Any) -> None:
+    """``init_weights`` for container modules: recursively init each child."""
+    for child in self.children():
+        assert isinstance(child, Module)
+        child.init_weights(**kwargs)
+
+
+class ModuleList(nn.ModuleList, Module):
```
tianyu-l, line 78: > Typing is also an issue as type intersection is not allowed in Python.

I think this can be addressed by creating a new class at runtime, inside `from_nn_module`. But I'm OK with creating these classes.

torchtitan/models/common/decoder.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, config: Config):
         self.rope = config.rope.build()
         self.register_buffer("freqs_cis", self.rope.cache, persistent=False)
 
-        self.layers = torch.nn.ModuleDict()
+        self.layers = ModuleDict()
```
tianyu-l, line 93: fwiw, there is this alternative approach https://fburl.com/code/qkh12cj4

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -633,6 +638,15 @@ def forward(self, *args, **kwargs):
         output = self.model.lm_head(output.last_hidden_state)
         return output
 
+    def verify_module_protocol(self) -> None:
```
tianyu-l, line 641: oh, did you make it an instance method of Model, instead of a static method in Module because of this? I would have preferred the latter if it were not for this use case.

torchtitan/trainer.py
```diff
@@ -270,6 +270,9 @@ def __init__(self, config: Config):
         )
         model_converters.convert(model)
 
+        # Verify all submodules satisfy the Module protocol
+        model.verify_module_protocol()
+
```
tianyu-l, line 280: Ideally this should be part of overall validation logic -- the calling here sounds a bit random.

I understand that putting this in model config `post_init` doesn't solve the problem, because of things like model converter. Ideally we should call convert on model configs at config time, rather than model instances at runtime.

I think it's fine to put it here for now, later on we can group related logic in `Model.post_init` or `Model.validate`.

fegin: I should add a comment here. I remember fully_shard will also remove the Module class :(. That's why I put it here. We need more clean up for sure.

torchtitan/experiments/transformers_modeling_backend/model.py
fegin: Yes, I originally make it as pure function (or staticmethod,  I forgot), but I have to overwrite this method for this model. Unfortunately, transformers_modeling_backend is very hard to support with the Module rewrite.

torchtitan/models/common/decoder.py
fegin: I see. Didn't check this before. ye, we can consider to do this as well.

torchtitan/protocols/module.py
fegin: The problem is that when creating a new class at runtime, we can only type annotating that class to either Module or nn.ModuleList but not both. This is what I meant `Typing is also an issue as type intersection is not allowed in Python.`.

PR #2609, opened 2026-03-16
[graph_trainer] Annotate ac region id for transformer blocks

yiming0416:
Without per-transformer-block AC region IDs, the min-cut partitioner sees the entire model as a single region. In practice, the partitioner can still rely on existing `MUST_SAVE` nodes as anchors to limit recomputation scope. But recomputation could trace all the way back to the beginning of the model when it doesn't hit `MUST_SAVE node.

By assigning a unique `ac_graph_id` to each transformer block, the partitioner is forced to `MUST_SAVE` at region boundaries (i.e., between transformer blocks). This ensures recomputation during the backward pass is always contained within a single block.

This PR:
- Adds `annotate_ac_regions()` to tag each transformer block's forward with a unique `ac_region_id`. 
- Updates `apply_sac_pass` to read the `ac_region_id` from node custom metadata and set it as the `ac_graph_id`.

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -8,12 +8,29 @@ from collections.abc import Callable
 
 import torch
 import torch.distributed as dist
+import torch.nn as nn
 from torch.distributed.tensor import DTensor, Replicate
+from torch.fx.traceback import annotate_fn
 from torch.utils._pytree import register_pytree_node, tree_map
 
 from torchtitan.config import CompileConfig
 from torchtitan.distributed import ParallelDims
 from torchtitan.tools.logging import logger
+
+_AC_REGION_ID = "ac_region_id"
+
+
+def annotate_ac_regions(model: nn.Module) -> None:
+    """Annotate each transformer block with a unique AC region ID.
+
+    This enables apply_sac_pass to assign different ac_graph_id values
+    per block, creating AC region boundaries between transformer blocks.
+    """
+    layers = model.get_submodule("layers")
+    for layer_id, transformer_block in layers.named_children():
+        transformer_block.forward = annotate_fn({_AC_REGION_ID: int(layer_id)})(
+            transformer_block.forward
+        )
 
 
 def parallelize_inputs(parallel_dims, args, kwargs):
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -15,6 +15,7 @@ Pass Types:
 - Compiler passes: Applied to the partitioned forward/backward graphs
 """
 import operator
+from collections import defaultdict
 from collections.abc import Sequence
 from typing import Any
 
@@ -29,6 +30,7 @@ from torch._inductor.fx_passes.overlap_scheduling import schedule_overlap_bucket
 from torch.fx.passes.regional_inductor import regional_inductor
 from torch.utils.checkpoint import CheckpointPolicy
 
+from torchtitan.experiments.graph_trainer.common_utils import _AC_REGION_ID
 from torchtitan.experiments.graph_trainer.reshard_after_forward import (
     annotate_fsdp_all_gather,
 )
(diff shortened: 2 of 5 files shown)


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -205,7 +206,8 @@ def apply_sac_pass(
                 node.meta["ac_graph_id"] = parent.meta.get("ac_graph_id", 0)
             continue
 
-        node.meta["ac_graph_id"] = 0
+        custom_meta = node.meta.get("custom", None) or {}
```
SherlockNoMad: node.meta.get("custom", {})

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -205,7 +206,8 @@ def apply_sac_pass(
                 node.meta["ac_graph_id"] = parent.meta.get("ac_graph_id", 0)
```
SherlockNoMad, line 210: move it here?

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -8,13 +8,30 @@
...
+_AC_REGION_ID = "ac_region_id"
+
+
+def annotate_ac_regions(model: nn.Module) -> None:
+    """Annotate each transformer block with a unique AC region ID.
+
+    This enables apply_sac_pass to assign different ac_graph_id values
+    per block, creating AC region boundaries between transformer blocks.
+    """
+    layers = model.get_submodule("layers")
```
SherlockNoMad, line 29: this is brittle. this is relying on the coincident that models have layers.0, layers.1... 
Can you think of better way to do this? anything we can use from titan's config system?
This also doesn't include embedding or lmhead layer, if any. 

add comment for these limitations, if you end up going this way.

SherlockNoMad: Also add logging for what has been annotated.

torchtitan/experiments/graph_trainer/tests/test_passes.py
```diff
@@ -295,6 +296,40 @@ def test_ac_graph_id_set(self):
...
+        apply_sac_pass(gm)
+
+        self.assertEqual(nodes[0].meta["ac_graph_id"], 1)
+        self.assertEqual(nodes[1].meta["ac_graph_id"], 2)
+
+    def test_ac_graph_id_propagates_to_getitem(self):
+        """getitem nodes should inherit ac_graph_id from their parent."""
+        gm = self._build_gm(
+            [
+                torch.ops.aten.add.Tensor,
```
SherlockNoMad: huh? you need an op that return tuple/list?

SherlockNoMad: also, need test for c10d.wait.

torchtitan/experiments/graph_trainer/passes.py
yiming0416: I didn't get it. move what?

torchtitan/experiments/graph_trainer/common_utils.py
yiming0416: > this is brittle. this is relying on the coincident that models have layers.0, layers.1...

It's relatively safe as torchtitan has its model definition/initialization following this pattern. In eager AC, it also relies on this pattern to wrap each transformer block. I could try to see if there are better ways.

> This also doesn't include embedding or lmhead layer, if any.

embedding and lm_head are intentionally skipped in eager AC. I wonder do we need them?

torchtitan/experiments/graph_trainer/tests/test_passes.py
yiming0416: - updated the graph to insert an op that returns a tuple before a `getitem` node
- added test for `wait_tensor`

torchtitan/experiments/graph_trainer/common_utils.py
yiming0416: Added logging

PR #2553, opened 2026-03-11
[graph_trainer] Add make_fx_tracer utility and unit tests

ydwu4:
- nonstrict_tracer.py: make_fx-based tracer with subclass handling (trace_module, run_traced_module, SubclassMeta, unwrap/wrap helpers)
- test_trace_module.py: simple MLP tests verifying bitwise-exact eager vs traced execution over multiple training steps

cc @SherlockNoMad

PR #2612, opened 2026-03-16
Make MoE models non-strict tracing friendly

ydwu4:
## Summary
- `kernels.py`: wrap `fill_indices` as `custom_op` + `register_fake` for FakeTensor compatibility during `make_fx` tracing
- `test_trace_module.py`: add model-specific tests (llama3, qwen3, qwen3_moe, deepseek_v3, llama4, gpt_oss) and FSDP tests

Stacked on #2553

## Test plan
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_trace_module.py -k TestTraceModels` — all 6 model tests pass
- [x] `pre-commit run --all-files` passes

### DIFF
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -24,31 +24,6 @@ from torch.utils._python_dispatch import is_traceable_wrapper_subclass
 
 
 @contextmanager
-def annotate_flex_attention_for_regional_inductor() -> Generator[None, None, None]:
-    """Annotate FlexAttentionWrapper.forward so regional_inductor compiles flex attention HOPs.
-
-    Uses the same inductor configs as FlexAttentionWrapper._compiled_flex_attn
-    to ensure bitwise-identical kernels between eager and regional_inductor paths.
-    """
-    from torch.fx.traceback import annotate_fn
-
-    from torchtitan.models.common.attention import FlexAttentionWrapper
-
-    orig = FlexAttentionWrapper.forward
-    FlexAttentionWrapper.forward = annotate_fn(
-        {
-            "compile_with_inductor": {
-                "inductor_configs": FlexAttentionWrapper.inductor_configs
-            }
-        }
-    )(orig)
-    try:
-        yield
-    finally:
-        FlexAttentionWrapper.forward = orig
-
-
-@contextmanager
 def _skip_nested_compile() -> Generator[None, None, None]:
     """Tell dynamo to skip torch.compile calls encountered during make_fx tracing.
 
--- a/torchtitan/experiments/graph_trainer/tests/test_trace_module.py
+++ b/torchtitan/experiments/graph_trainer/tests/test_trace_module.py
@@ -15,9 +15,11 @@ from torch.testing._internal.common_fsdp import FSDPTest
 from torchtitan.experiments.graph_trainer.make_fx_tracer import (
     _copy_fwd_metadata_to_bw_nodes,
     _patch_engine_run_backward,
-    annotate_flex_attention_for_regional_inductor,
     run_traced_module,
     trace_module,
+)
+from torchtitan.models.common.attention import (
+    annotate_flex_attention_for_regional_inductor,
 )
 
 
@@ -678,7 +680,9 @@ class TestTraceFSDP(FSDPTest):
         model_test = create_model(config_cls, model_config, "cuda", torch.float32)
         model_test.load_state_dict(model_ref.state_dict())
         model_ref = data_parallel(model_ref, device_mesh=fsdp_mesh, mode="fully_shard")
-        model_test = data_parallel(model_test, device_mesh=fsdp_mesh, mode="fully_shard")
+        model_test = data_parallel(
+            model_test, device_mesh=fsdp_mesh, mode="fully_shard"
+        )
 
         vocab_size = model_config.vocab_size
         seq_len = 128
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -4,7 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from collections.abc import Callable
+from collections.abc import Callable, Generator
+from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import ClassVar, NamedTuple
 
@@ -173,6 +174,29 @@ class FlexAttentionWrapper(Module):
             enable_gqa=enable_gqa,
             return_lse=return_lse,
         )
+
+
+@contextmanager
+def annotate_flex_attention_for_regional_inductor() -> Generator[None, None, None]:
+    """Annotate FlexAttentionWrapper.forward so regional_inductor compiles flex attention HOPs.
+
+    Uses the same inductor configs as FlexAttentionWrapper._compiled_flex_attn
+    to ensure bitwise-identical kernels between eager and regional_inductor paths.
+    """
+    from torch.fx.traceback import annotate_fn
+
+    orig = FlexAttentionWrapper.forward
+    FlexAttentionWrapper.forward = annotate_fn(
+        {
+            "compile_with_inductor": {
+                "inductor_configs": FlexAttentionWrapper.inductor_configs
+            }
+        }
+    )(orig)
+    try:
+        yield
+    finally:
+        FlexAttentionWrapper.forward = orig
 
 
 class ScaledDotProductAttentionWrapper(Module):


torchtitan/models/common/moe/kernels.py
tianyu-l, line 1: again, this file should be removed. cc @pianpwk

torchtitan/models/gpt_oss/moe.py
```diff
@@ -163,31 +230,27 @@ def _run_experts_grouped_mm(
         x.bfloat16(), mlp1_weight.transpose(-2, -1).bfloat16(), offs=offsets
     )
 
-    b1 = torch.cat([mlp1_bias, mlp1_bias.new_zeros(1, mlp1_bias.shape[-1])])
```
tianyu-l, line 166: This is pytorch native op, I'm surprised that we still need any special treatment, even with the padding kernel removed?

torchtitan/experiments/graph_trainer/graph_utils.py
```diff
@@ -538,3 +518,27 @@ def get_joint_custom_passes_from_config(
     )
 
     return joint_custom_passes
+
+
```
yiming0416: Are these functions used anywhere?

ydwu4: Removed!

torchtitan/models/gpt_oss/moe.py
ydwu4: Yeah, you're right, this is not needed. Probably some left over when doing exploration.

torchtitan/models/common/moe/kernels.py
ydwu4: I'll keep the change in this PR for now! Either way works for me.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
         self.assertIs(torch.autograd._engine_run_backward, orig_fn)
 
 
+@contextlib.contextmanager
+def _use_raw_flex_attn():
```
yiming0416: This one is useful in https://github.com/pytorch/torchtitan/pull/2561 as well. 
As I ran into the same issue. Maybe we can put it in `make_fx_tracer.py`?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
...
+            opt_copy.step()
+            opt_copy.zero_grad()
+
+            self.assertTrue(
+                torch.equal(loss_ref, loss_tr), f"Step {step}: loss mismatch"
+            )
+            for gr, gt in zip(grads_ref, grads_tr, strict=True):
+                self.assertTrue(torch.equal(gr, gt), f"Step {step}: grad mismatch")
+
+    def _run_model_test(self, config_cls, model_config, use_attn_masks=False):
```
yiming0416: You removed the simple_fsdp `data_parallel` wrapper here?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
...
+@contextlib.contextmanager
+def _use_raw_flex_attn():
+    """Swap the compiled flex_attention with the raw (uncompiled) version.
+
+    FlexAttentionWrapper uses torch.compile'd flex_attention by default.
+    torch.compile inside make_fx tracing is not supported and raises:
+      "Detected that you are using FX to symbolically trace a
+       dynamo-optimized function."
+    Using the raw version lets make_fx decompose flex_attention into
+    plain aten ops (bmm, softmax, etc.) which trace correctly.
```
SherlockNoMad: IIUC, in eager, compiled flex_attn show up as a HOP?

We should fix make_fx to trace over HOP? 

Also, in Ed's 2-tier graph setup, we are already non-strict tracing per layer, which is torch.compiled.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
...
 
+@contextlib.contextmanager
+def _use_raw_flex_attn():
+    """Swap the compiled flex_attention with the raw (uncompiled) version.
+
+    FlexAttentionWrapper uses torch.compile'd flex_attention by default.
+    torch.compile inside make_fx tracing is not supported and raises:
+      "Detected that you are using FX to symbolically trace a
+       dynamo-optimized function."
+    Using the raw version lets make_fx decompose flex_attention into
```
SherlockNoMad: I don't think it should decompose flex into plain ops?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
...
+                logits_ref = model_ref(*fwd_args)
+            loss_ref = get_loss(logits_ref, labels)
+            loss_ref.backward()
+            grads_ref = [p.grad.clone() for p in model_ref.parameters()]
+            opt_ref.step()
+            opt_ref.zero_grad()
+
+            train_step_copy = TrainStepModule(model_copy, get_loss)
+            pab = _get_params_and_buffers(train_step_copy)
+            wrapped = run_traced_module(traced_result, pab, (*fwd_args, labels))
```
SherlockNoMad: bad name

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
...
+        torch.manual_seed(42)
+        torch.use_deterministic_algorithms(True)
+
+    def tearDown(self):
+        torch.use_deterministic_algorithms(False)
+
+    def _run_bitwise_test(
+        self,
+        model_ref,
+        model_copy,
```
SherlockNoMad: ref vs. test

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +353,224 @@ def test_patch_engine_restores_original(self):
...
+            return
+
+        self._run_bitwise_test(
+            model_ref,
+            model_copy,
+            (tokens,),
+            labels,
+            num_steps=self.NUM_STEPS,
+            lr=self.LR,
+        )
```
SherlockNoMad: dedup

tianyu-l:
cc @danielvegamyhre you might need this change in torchao in the future, if we want to make permutation work with non-strict tracing.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
ydwu4: sure

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
ydwu4: Ok, the real problem is FlexAttentionWrapper is using a different torch.compile wrapper than the original flex_attention.

The graph we traced consists of the flex_attention higher order op. The eager mode runs the torch.compiled version which produces numerical differences with running the hop. So we'll run eager under use_uncompiled_flex_attention, to ensure both graph and eager's flex_attention are the un-compiled version and in this way we get bitwise equivalence.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
ydwu4: adding it back as unit tests!

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
ydwu4: Summarizing the current status:  SPDA is used in llama3, qwen3, it gets decomposed with pre_dispatch=False. For ds, lma4, gpt_oss, flex_attention is used and the  flex_attention hop is preserved in graph. bit-wise equivalence is achieved via running eager under use_uncompiled_flex_attention.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -352,5 +357,422 @@ def test_patch_engine_restores_original(self):
...
+        torch.cuda.manual_seed(42)
+        torch.use_deterministic_algorithms(True)
+        self._setup()
+        fsdp_mesh = self.parallel_dims.get_mesh("fsdp")
+
+        model_ref = create_model(config_cls, model_config, "cuda", torch.float32)
+        model_test = create_model(config_cls, model_config, "cuda", torch.float32)
+        model_test.load_state_dict(model_ref.state_dict())
+        data_parallel(model_ref, device_mesh=fsdp_mesh, mode="fully_shard")
+        data_parallel(model_test, device_mesh=fsdp_mesh, mode="fully_shard")
```
yiming0416: should be
model_ref = data_parallel(model_ref, ...)
model_test = data_parallel(model_test, ...)

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
```diff
@@ -63,7 +63,11 @@ def annotate_deepseekv3(model: GraphTrainerDeepSeekV3Model) -> None:
     MoE.forward = annotate_fn({"EP": "compute"})(MoE.forward)
 
     FlexAttentionWrapper.forward = annotate_fn(
-        {"compile_with_inductor": "flex_attention"}
+        {
```
yiming0416: why do you need this change?

ydwu4: This doesn't need to happen in this PR though. The idea is to keep regional inductor's inductor config to be consistent with the eager FlexAttentionWrapper.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -23,6 +23,48 @@
 from torch.utils._python_dispatch import is_traceable_wrapper_subclass
 
 
+@contextmanager
+def annotate_flex_attention_for_regional_inductor() -> Generator[None, None, None]:
```
SherlockNoMad: implementation looks good, but I don't think it should live in this file, please find a better home for this function. 

maybe somewhere together with regional_inductor pass?

ydwu4: Sure, move it to FlexAttentionWrapper file seems it's directly tied to it.

PR #2617, opened 2026-03-17
[graph_trainer] Propagate forward metadata to backward nodes in make_fx_tracer

SherlockNoMad:
Patch the autograd engine during make_fx tracing to preserve seq_nr on
    backward FX nodes, then copy forward-node custom/nn_module_stack metadata
    to corresponding backward nodes via seq_nr matching.

### DIFF
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -5,9 +5,10 @@
 # LICENSE file in the root directory of this source tree.
 
 import itertools
+from collections.abc import Generator
 from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Generator
+from typing import Any
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -157,6 +161,79 @@ def _remove_cpu_shadow_chains(gm: torch.fx.GraphModule) -> None:
...
+        if node.op != "call_function" or "seq_nr" not in node.meta:
+            continue
+        seq_nr = node.meta["seq_nr"]
+        if seq_nr not in seq_nr_to_fwd_node:
+            seq_nr_to_fwd_node[seq_nr] = node
+        else:
+            fwd_node = seq_nr_to_fwd_node[seq_nr]
+
+            custom = fwd_node.meta.get("custom")
+            if custom:
```
yushangdi, line 231: if the graph module is nested, this'll probably miss some nodes.

SherlockNoMad: will address if we dicided to go 2-tier graph with make_fx trace.

PR #2620, opened 2026-03-18
Only apply grouped GEMM padding for MXFP8 and FP8 non-HybridEP cases

danielvegamyhre:
## Context
- BF16 grouped GEMM no longer requires padding, we can remove it from the BF16 path and only use it for FP8 and MXFP8 grouped GEMMs
- TorchTitan will now only only contain a torch native "rank major to expert major" permutation impl for BF16 grouped GEMM, and not any extra per group padding kernels/logic for FP8/MXFP8 (these will live in torchao, as the quantization library it is a better home for them).

## Summary
There are 7 cases to handle:

- Case 1: BF16 + NoEP  
    - (do nothing)
- Case 2: BF16 + EP
    - Torch native impl handles permute from rank major to expert major (no padding) 
- Case 3: MXFP8 + No EP
    - Handled with pad/unpad kernels in torchao   
- Case 4: MXFP8 + Standard EP
    - torchao `permute_and_pad()` if token_group_alignment_size > 0, in ExpertParallel implementation
- Case 5: MXFP8 + HybridEP
    - HybridEP handles token group padding for MXFP8 grouped GEMM as part of the all2all dispatch
- Case 6: FP8 + No EP
    - Same as case 3  
- Case 7: FP8 + EP
    - Same as case 4 

## Misc changes
- Delete `kernels.py`
- Delete tests for those kernels
- Remove `pad_token_groups_for_grouped_mm` option from MXFP8ConverterConfig, since we can set it correctly automatically 
- Added debug models for float8 and mxfp8 to config registry to speed up future development 

## Tests
- BF16 + No EP (ok): https://www.internalfb.com/phabricator/paste/view/P2244567028
- BF16 + EP (ok): https://www.internalfb.com/phabricator/paste/view/P2244568064
- MXFP8 + No EP (issue): https://www.internalfb.com/phabricator/paste/view/P2244573156
    - Important follow up: nan loss for this case! probably a bug in the new pad/unpad kernels we added for this in torchao, as they have unit tests but this was the first e2e training test 
- MXFP8 + EP (ok): https://www.internalfb.com/phabricator/paste/view/P2244569057
- FP8 + No EP (ok): https://www.internalfb.com/phabricator/paste/view/P2244583497
- FP8 + EP (ok): https://www.internalfb.com/phabricator/paste/view/P2244584583
- I did not test Case 5 (Hybrid EP + MXFP8) because it was recently tested by others, and is a different code path not part of these changes.

FP8 tests were done with fp8 grouped mm only, not fp8 linear. Using both I get this weird tyro error?

```
[rank0]:│ model-converters.converters.0:config was not a match because:                │
[rank0]:│ • Default value Config(enable_fsdp_float8_all_gather=False,                  │
[rank0]:│   precompute_float8_dynamic_scale_for_fsdp=False, recipe_name=None,          │
[rank0]:│   filter_fqns=['output', 'router.gate'], emulate=False) with type Config     │
[rank0]:│   does not match type <class 'torchtitan.components.quantization.float8.Floa │
[rank0]:│   t8GroupedMMConverter.Config'>         
```

### DIFF
--- a/torchtitan/components/quantization/__init__.py
+++ b/torchtitan/components/quantization/__init__.py
@@ -32,5 +32,5 @@ class QuantizationConverter(Configurable):
 
 
 # Module level global constants
-FP8_GROUP_ALIGNMENT_SIZE = 16
+FLOAT8_GROUP_ALIGNMENT_SIZE = 16
 MXFP8_GROUP_ALIGNMENT_SIZE = 32
--- a/torchtitan/components/quantization/mx.py
+++ b/torchtitan/components/quantization/mx.py
@@ -9,14 +9,10 @@ from importlib.util import find_spec
 from typing import ClassVar, Literal
 
 import torch.nn as nn
-from torchtitan.components.quantization import (
-    MXFP8_GROUP_ALIGNMENT_SIZE,
-    QuantizationConverter,
-)
+from torchtitan.components.quantization import QuantizationConverter
 
 from torchtitan.distributed import ParallelDims
 from torchtitan.models.common.linear import Linear
-from torchtitan.models.common.moe.utils import set_token_group_alignment_size
 from torchtitan.tools.logging import logger
 from torchtitan.tools.utils import has_cuda_capability
 
@@ -61,6 +57,7 @@ class MXFP8Converter(QuantizationConverter):
         model_compile_enabled: bool,
     ):
         self.enabled = False
+        self.convert_grouped_gemms = False
 
         # Ensure minimum torchao versions
         if find_spec("torchao") is None:
@@ -86,9 +83,6 @@ class MXFP8Converter(QuantizationConverter):
             )
 
         logger.info("MXFP8 MoE training enabled")
-
-        # Set token group alignment size for MXFP8 grouped GEMMs
-        set_token_group_alignment_size(MXFP8_GROUP_ALIGNMENT_SIZE)
 
         # If EP is enabled, TorchTitan handles the token group padding for MXFP8 grouped GEMM
         # as part of the EP implementation.
@@ -153,3 +147,13 @@ class MXFP8Converter(QuantizationConverter):
         MXFP8 training doesn't require any post-optimizer hooks at the moment
         """
         return
+
+
+def find_mxfp8_config(
+    converters: list,
+) -> MXFP8Converter.Config | None:
+    """Find the MXFP8Converter.Config in a list of converter configs, if any."""
+    return next(
+        (c for c in converters if isinstance(c, MXFP8Converter.Config)),
+        None,
+    )
--- a/torchtitan/components/quantization/utils.py
+++ b/torchtitan/components/quantization/utils.py
@@ -4,7 +4,35 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from enum import auto, Enum
+
 import torch.nn as nn
+from torchtitan.components.quantization.float8 import find_float8_grouped_mm_config
+from torchtitan.components.quantization.mx import find_mxfp8_config
+from torchtitan.protocols import ModelConverter
+
+
+class QuantizationType(Enum):
+    FLOAT8 = auto()
+    MXFP8 = auto()
+
+
+def routed_experts_in_fqns(fqns: list[str]) -> bool:
+    """Helper used to determine if quantization is targeting routed experts (grouped GEMMs)."""
+    for fqn in fqns:
+        if "experts" in fqn and "shared_experts" not in fqn:
+            return True
+    return False
+
+
+def get_grouped_mm_quantization_type(
+    model_converters: list[ModelConverter],
+) -> QuantizationType | None:
+    if find_float8_grouped_mm_config(model_converters):
+        return QuantizationType.FLOAT8
+    elif config := find_mxfp8_config(model_converters):
+        if routed_experts_in_fqns(config.fqns):
+            return QuantizationType.MXFP8
 
 
 def module_filter_fn(mod: nn.Module, fqn: str, filter_fqns: list[str]) -> bool:
--- a/torchtitan/distributed/deepep/hybridep.py
+++ b/torchtitan/distributed/deepep/hybridep.py
@@ -29,13 +29,11 @@ from torch._library.opaque_object import (
 )
 from torch.distributed import ProcessGroup
 
-from torchtitan.models.common.moe.utils import (
-    get_mxfp8_pad_multiple,
-    maybe_align_num_tokens_for_mxfp8,
-)
-
+from torchtitan.components.quantization import MXFP8_GROUP_ALIGNMENT_SIZE
+from torchtitan.components.quantization.utils import QuantizationType
 
 _buffer: Any = None  # Global buffer instance
+_quantization_type: QuantizationType | None = None  # Current quantization type
 
 
 class DispatchHandle(OpaqueBase):
@@ -308,8 +306,6 @@ def _combine_backward(ctx, grad_combined):
 
     # Must pass pad_multiple so backward gradients entering ScaledGroupedMM
     # (torchao MXFP8) also have rows aligned to 32.
-    from torchtitan.models.common.moe.utils import get_mxfp8_pad_multiple
-
     pad_multiple = get_mxfp8_pad_multiple()
 
     grad_x, _, _, _, _ = _buffer.dispatch_with_permute(
@@ -404,6 +400,7 @@ def dispatch_tokens(
     group: ProcessGroup,
     score_before_experts: bool = True,
     non_blocking_expert_capacity_factor: float | None = None,
+    quantization_type: QuantizationType | None = None,
 ) -> tuple[torch.Tensor, torch.Tensor, DispatchState]:
     """Dispatch tokens to experts via HybridEP all-to-all.
 
@@ -419,10 +416,15 @@ def dispatch_tokens(
             float in (0, 1] = non-blocking mode; pre-sizes the permute output
             tensor as num_tokens × ep_size × min(num_local_experts, top_k) × cf,
             aligned for MXFP8.
+        quantization_type: Quantization type (FLOAT8, MXFP8, or None) for
+            determining padding requirements.
 
     Returns:
         (permuted_hidden, tokens_per_expert, state)
     """
+    global _quantization_type
+    _quantization_type = quantization_type
+
     non_blocking = non_blocking_expert_capacity_factor is not None
 
     selected_experts_indices = selected_experts_indices.contiguous()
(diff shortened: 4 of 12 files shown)


danielvegamyhre:
Have not tested yet because of devgpu issues but if you want to take a look feel free @tianyu-l

danielvegamyhre:
I finished testing @tianyu-l this is ready for review

torchtitan/models/common/moe/utils.py
```diff
@@ -12,37 +12,21 @@
 from torchtitan.components.quantization import MXFP8_GROUP_ALIGNMENT_SIZE
 from torchtitan.tools.utils import _round_up
 
-from .kernels import generate_permute_indices
+TOKEN_GROUP_ALIGN_SIZE_M = 0
```
tianyu-l: we should remove this -- setting global variables is error-prone

we should move logic to parallelize functions for various combinations

torchtitan/models/common/moe/utils.py
```diff
@@ -55,38 +39,30 @@ def set_token_group_alignment_size_m(
...
-def _unpermute(out, input_shape, permuted_indices):
-    out_unpermuted = out.new_empty(input_shape)
-    out_unpermuted[permuted_indices, :] = out
-    out = out_unpermuted[:-1]
-    return out
+def maybe_align_num_tokens_for_mxfp8(num_tokens: int) -> int:
+    """Round up token count only when MXFP8 group alignment is active."""
+    if TOKEN_GROUP_ALIGN_SIZE_M != MXFP8_GROUP_ALIGNMENT_SIZE:
+        return num_tokens
+    return _round_up(num_tokens, MXFP8_GROUP_ALIGNMENT_SIZE)
```
tianyu-l: move this logic to hybridep.py, including `_round_up` (as an inline function) which is currently only used once in this repo

torchtitan/distributed/expert_parallel.py
```diff
@@ -135,28 +140,54 @@ def _token_dispatch(
...
-        ) = _permute(
-            routed_input, num_tokens_per_expert_group, ep_degree, num_local_experts
-        )
-
-        return routed_input, num_tokens_per_expert_group
+        # Non-HybridEP case:
+        # FP8/MXFP8 require groups to be permuted to expert major order AND padded to
+        # `alignment_size`.
+        # Otherwise, we only need to permute to expert major order.
+        if self.token_group_alignment > 0:
```
tianyu-l: IMO the proper way is to create e.g. `FP8ExpertParallel` and dispatch to it in parallelize function, instead of making if-else in existing ExpertParallel.

Also the condition should be whether quantization is used, not the `token_group_alignment` size set from somewhere.

torchtitan/models/common/moe/utils.py
```diff
@@ -118,3 +94,67 @@ def wrapper(
...
+        )
+
+    input_shape = x.shape
+    x = x[permuted_indices, :]
+
+    return input_shape, x, permuted_indices, num_tokens_per_expert
+
+
+# Source: https://github.com/pytorch/torchtitan/pull/2255
+def _generate_permute_indices(
```
tianyu-l, line 63: could you verify that before vs. after, we get bitwise identical results under same seed and determinism?

torchtitan/models/common/moe/utils.py
danielvegamyhre: makes sense, i am working on this, the changes are straightforward for standard EP, but for Hybrid EP it seems like it will require (1) refactoring the custom ops, DispatchState etc to pass around the quantization type used, or (2) just a module level variable storing the quantization type, similar to `_buffer`.  I think (2) is a less invasive change, wdyt?

torchtitan/distributed/expert_parallel.py
danielvegamyhre: That works, did a refactor

torchtitan/models/common/moe/utils.py
```diff
@@ -5,116 +5,99 @@
...
 
-from torchtitan.components.quantization import MXFP8_GROUP_ALIGNMENT_SIZE
-from torchtitan.tools.utils import _round_up
-
-from .kernels import generate_permute_indices
-
-TOKEN_GROUP_ALIGN_SIZE_M = 8
-ValidTokenGroupAlignmentSize = Literal[8, 16, 32]
 
+def indices_padding_wrapper(func: Callable) -> Callable:
```
tianyu-l, line 12: I don't think we need this function any more. Please remove this and simplify https://github.com/pytorch/torchtitan/blob/c0c0bf965e32906b5e80f4f42b1f42ebe9851840/torchtitan/models/common/moe/moe.py#L120-L131

torchtitan/models/gpt_oss/moe.py
```diff
@@ -45,10 +45,9 @@ def backward(ctx, grad_output):
 
 def indices_padding_wrapper(func: Callable) -> Callable:
```
tianyu-l, line 46: same

torchtitan/distributed/expert_parallel.py
```diff
@@ -147,16 +151,251 @@ def _token_dispatch(
...
+            routed_input,
+            self.permuted_indices,
+            num_tokens_per_expert_group_padded,
+            group_offsets,
+        ) = permute_and_pad(
+            routed_input,
+            num_tokens_per_expert_group,
+            ep_degree,
+            num_local_experts,
+            FLOAT8_GROUP_ALIGNMENT_SIZE,
```
tianyu-l, line 267: why do you need two different classes? You could just init with different quantization type, which can be used to determine the alignment size, e.g. based on a static dict.

torchtitan/components/quantization/__init__.py
```diff
@@ -32,5 +32,5 @@ class Config(Configurable.Config):
 
 
 # Module level global constants
-FP8_GROUP_ALIGNMENT_SIZE = 16
+FLOAT8_GROUP_ALIGNMENT_SIZE = 16
 MXFP8_GROUP_ALIGNMENT_SIZE = 32
```
tianyu-l, line 36: make this a dict from quantization type to alignment size

torchtitan/components/quantization/utils.py
```diff
@@ -4,7 +4,35 @@
...
+    return False
+
+
+def get_grouped_mm_quantization_type(
+    model_converters: list[ModelConverter],
+) -> QuantizationType | None:
+    if find_float8_grouped_mm_config(model_converters):
+        return QuantizationType.FLOAT8
+    elif config := find_mxfp8_config(model_converters):
+        if routed_experts_in_fqns(config.fqns):
```
tianyu-l, line 34: no need to modularize into multiple small functions which are not used elsewhere -- we can make everything in a single util function for now

torchtitan/components/quantization/utils.py
```diff
@@ -4,7 +4,35 @@
...
 
+from enum import auto, Enum
+
 import torch.nn as nn
+from torchtitan.components.quantization.float8 import find_float8_grouped_mm_config
+from torchtitan.components.quantization.mx import find_mxfp8_config
+from torchtitan.protocols import ModelConverter
+
+
+class QuantizationType(Enum):
```
tianyu-l, line 15: you already have https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/mx.py#L35, can we just use the strings?

torchtitan/components/quantization/mx.py
```diff
@@ -78,21 +70,38 @@ def __init__(
...
                 "torch.compile enablement is required for highest performance of MXFP8 dynamic quantization."
             )
 
+        logger.info("MXFP8 MoE training enabled")
+
+        # If EP is enabled, TorchTitan handles the token group padding for MXFP8 grouped GEMM
+        # as part of the EP implementation.
+        # Otherwise, if EP is not enabled, we need TorchAO to pad the token groups.
+        self.pad_token_groups_for_grouped_mm = not parallel_dims.ep_enabled
+        logger.warning(
```
tianyu-l, line 91: why it's a warning? sounds like a comment to me, especially when both hybridEP is used this warning would still be there

torchtitan/distributed/deepep/hybridep.py
```diff
@@ -404,6 +400,7 @@ def dispatch_tokens(
     group: ProcessGroup,
     score_before_experts: bool = True,
     non_blocking_expert_capacity_factor: float | None = None,
+    quantization_type: QuantizationType | None = None,
```
tianyu-l, line 403: hybridep module doesn't need to know the quantization_type. All it needs to know is pad multiple size.

torchtitan/distributed/expert_parallel.py
```diff
@@ -147,16 +151,251 @@ def _token_dispatch(
...
+            device_mesh,
+            partition_fn=self._partition_fn,
+            # pyrefly: ignore [bad-argument-type]
+            input_fn=self._token_dispatch,
+            # pyrefly: ignore [bad-argument-type]
+            output_fn=self._token_combine,
+        )
+
+
+class Float8ExpertParallel(BaseExpertParallel):
```
tianyu-l, line 184: can you inherit `ExpertParallel` instead of `BaseExpertParallel`, which can save a lot of code?

pianpwk:
Would it also be possible to remove the for-loop padding/unpadding path (for < SM90) in https://github.com/pytorch/torchtitan/blob/b01adfb544b4331ecab090ebdb50b2296cd8eb6a/torchtitan/models/common/moe/moe.py#L24 and https://github.com/pytorch/torchtitan/blob/b01adfb544b4331ecab090ebdb50b2296cd8eb6a/torchtitan/models/gpt_oss/moe.py#L99 , as part of this work?

I think that should resolve issues https://github.com/pytorch/torchtitan/issues/2312 and https://github.com/pytorch/torchtitan/pull/2741

tianyu-l:
I can take this over if you are busy @danielvegamyhre

torchtitan/components/quantization/mx.py
```diff
@@ -78,21 +70,38 @@ def __init__(
...
             logger.warning(
                 "torch.compile enablement is required for highest performance of MXFP8 dynamic quantization."
             )
 
+        logger.info("MXFP8 MoE training enabled")
+
+        # If EP is enabled, TorchTitan handles the token group padding for MXFP8 grouped GEMM
+        # as part of the EP implementation.
+        # Otherwise, if EP is not enabled, we need TorchAO to pad the token groups.
+        self.pad_token_groups_for_grouped_mm = not parallel_dims.ep_enabled
```
tianyu-l, line 90: what happens for float8, where you don't have this flag?

danielvegamyhre:
thank you @tianyu-l yes i have been busy with an urgent NaN loss situation reported by a customer... will take a look!

danielvegamyhre:
Closing in favor of #2774, thanks @tianyu-l !

PR #2626, opened 2026-03-18
Fix Pyrefly type errors for Tensor.item() return type

fegin:
* __->__ #2626

**Safe to merge**
**Root cause (from claude):**

This is due to PyTorch updating the type stubs for Tensor.item(). In older versions, .item() was annotated to return Number (i.e., int | float | complex). But as part of the symbolic shapes work for torch.compile / torch.export, PyTorch added SymInt and SymFloat to the return type.


wwwjn:
I will go ahead and merge this PR to unblock other PRs as all CUDA CI passed, and lint passed

PR #2628, opened 2026-03-18
[RL] Add TorchStore install instructions to README

daniellepintz:
Add torchstore install instructions from main, and pin torchmonarch to 0.3.0.


tianyu-l:
@wwwjn may need to change CI as well

PR #2629, opened 2026-03-19
[DO NOT LAND] Test flux Ci

wwwjn:


wwwjn:
file a bug in `datasets`: https://github.com/huggingface/datasets/issues/8079

break due to recent `datasets` version bump from 4.7.x and 4.8.x

tianyu-l:
can we pin to an old version of datasets in torchtitan to workaround?

wwwjn:
> can we pin to an old version of datasets in torchtitan to workaround?

Yes we can, now changing the stream=False is also a safe workaround, both should work. We can pin to any 4.7.x

lhoestq:
Let me try to do a patch release in `datasets` today - will keep you posted

lhoestq:
`datasets` 4.8.3 is out and includes a fix

wwwjn:
> `datasets` 4.8.3 is out and includes a fix

Thank you for your quick fix, I verified locally it works well. Waiting for our CI  to pick up

PR #2631, opened 2026-03-19
[CI] Pin datasets < 4.8.0 to avoid streaming sharding regression

wwwjn:
`datasets` 4.8.x introduced a bug in StepExamplesIterable._iter_arrow where single-row Arrow batches cause split_dataset_by_node to yield 0 samples for non-zero ranks when **num_shards < world_size**. This breaks Flux training with the cc12m-test and coco-validation datasets.

### DIFF
--- a/.ci/docker/requirements.txt
+++ b/.ci/docker/requirements.txt
@@ -1,5 +1,5 @@
 torchdata >= 0.8.0
-datasets >= 3.6.0
+datasets >= 3.6.0, < 4.8.0
 tensorboard
 wandb
 fsspec


wwwjn:
Close because of https://github.com/pytorch/torchtitan/pull/2629#issuecomment-4092094100, locally verified it  works well

wwwjn:
Not sure why CI haven't pickup `datasets==4..8.3` till today, let land this PR to unblock and revisit the `datasets` version after. cc @tianyu-l

PR #2632, opened 2026-03-19
Fix unreachable links in markdown files

Andrei-Aksionov:
Hi there 👋 

I am planning to start contributing to this repo, so, naturally, I started with reading docs (I know, very old-school).
Noticed broken links.

### This PR:
1. **Fixes broken links:** Resolves several 404s and incorrect internal file references in `.md` and `.py` files.
2. **Automated Link Checking:** Integrates [Lychee](https://github.com/lycheeverse/lychee) via `pre-commit` to prevent future regressions.

### Why Lychee?
It’s a fast, async, Rust-based link checker that handles:
* Async execution and custom User-Agents.
* Anchor validation and rate-limit respect.
* Local file path verification.
* and [more](https://github.com/lycheeverse/lychee?tab=readme-ov-file#features).

### Implementation Details:
The tool is integrated as a **system-level `pre-commit` hook**. I have simplified the design to follow standard `pre-commit` behavior:

* **Zero-Dependency Setup:** To avoid requiring `Cargo` or forcing heavy builds, the hook expects the `Lychee` binary on the system. If missing, it gracefully skips with installation instructions rather than blocking the developer.
* **Git-Aware Filtering:** The hook now uses `pre-commit`'s native file discovery. This automatically respects `.gitignore`, ensuring that local artifacts (`build/`, `.venv/`, `wandb/`, etc.) are never scanned.
* **Validation Strategy:**
    * **Local Commits:** By default, `git commit` only runs against **changed files**. This keeps the dev loop fast but means a renamed file won't trigger a warning for an *untouched* README that links to it.
    * **Full Audit (CI/Manual):** To perform a full repository check (internal + external links across all `.md` and `.py` files), run:
        ```bash
        pre-commit run --all-files
        ```
* **CI Integration:** The GitHub Action is configured to run the **Full Audit** on every PR to ensure total link health and catch any cross-reference breakages that local partial commits might miss.

### Verification:
* Verified that `build/` and `.venv/` are ignored during scans.
* Tested transient error handling (502s) with increased retries and wait times.
* Verified the workflow passes in CI via `pre-commit run --all-files`.
* Tested locally via `pre-commit`, in a containerized environment via `act`, and verified the workflow passes in my GitHub fork.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -80,23 +80,23 @@ repos:
       name: lychee-link-checker
       verbose: true
       language: system
+      require_serial: true
       entry: |
         /bin/bash -c '
         if ! command -v lychee >/dev/null 2>&1; then
             echo "--------------------------------------------------------"
-            echo "Warning: Lychee (link checker) not found. skipping ... "
-            echo "To enable local link checking, install Lychee:"
-            echo "  - macos: brew install lychee"
-            echo "  - linux/windows: https://github.com/lycheeverse/lychee"
+            echo "Warning: Lychee (link checker) not found. Skipping ... "
             echo ""
-            echo "If this error occurred in CI (GitHub actions), ensure the "
-            echo "installation step in lint.yaml is running correctly."
+            echo "To enable local link checking, install Lychee:"
+            echo "  - MacOS: brew install lychee"
+            echo "  - Linux/Windows: https://github.com/lycheeverse/lychee"
             echo "--------------------------------------------------------"
             exit 0
         fi
         lychee "$@"
         ' --
       args:
+        - "--quiet"
         - "--no-progress"
         - "--exclude-path=(^|/)\\.[^/]+/"
         - "--root-dir=."


torchtitan/experiments/forge/README.md
```diff
@@ -9,6 +9,4 @@ The [`ForgeEngine`](engine.py) takes a `ForgeEngine.Config` to
 
 **Primary Use Case**: The engine is designed for building trainers in post-training workflows where multiple specialized components (trainer, generator, replay buffer, parameter server, etc.) work together.
 
-Additionally, the folder provides a model spec registration method [`register_model_spec`](model_spec.py) that allows users to extend beyond the core set of models and training components available in torchtitan, enabling greater flexibility and customization for specific training requirements.
```
Andrei-Aksionov, line 12: While there is `torchtitan/protocols/model_spec.py`, there is no `register_model_spec` inside. That is the reason of deletion.

CONTRIBUTING.md
```diff
@@ -77,6 +77,5 @@ When appropriate, one should consider
   - To add a unit test, put it in the [tests](tests/) folder and follow the existing test files.
   - To add a GPU integration test, create a new `OverrideDefinitions` in [integration_tests](tests/integration_tests/). It will override the default config to run on the Llama 3 debug model (see [config_registry.py](torchtitan/models/llama3/config_registry.py)).
 - Updating [README](README.md) and writing a new note in the [docs](docs/) folder on installation and usage, similar to [float8.md](docs/float8.md).
-- Updating [performance.md](docs/performance.md) with new performance results.
```
Andrei-Aksionov, line 80: Couldn't find `performance.md` anywhere.

README.md
```diff
@@ -85,9 +85,8 @@ We report [performance](benchmarks/llama3_h100_202412_torchtitan.md) on up to 51
 
 You may want to see how the model is defined or how parallelism techniques are applied. For a guided tour, see these files first:
 * [torchtitan/train.py](torchtitan/train.py) - the main training loop and high-level setup code
-* [torchtitan/models/llama3/model/model.py](torchtitan/models/llama3/model/model.py) - the Llama 3.1 model definition
-* [torchtitan/models/llama3/infra/parallelize.py](torchtitan/models/llama3/infra/parallelize.py) - helpers for applying Data Parallel, Tensor Parallel, activation checkpointing, and `torch.compile` to the model
-* [torchtitan/models/llama3/infra/pipeline.py](torchtitan/models/llama3/infra/pipeline.py) - helpers for applying Pipeline Parallel to the model
```
Andrei-Aksionov, line 90: There is no `pipeline.py` inside `llama3` folder.

.github/workflows/check-links.yaml
```diff
@@ -0,0 +1,68 @@
...
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install mistune pytest pytest-check-links pytest-timeout pytest-xdist
+
+      - name: Check links
+        run: |
+          # 1. Create a temporary conftest.py to handle 403/429/503s.
```
tianyu-l: curious, do we have to create this file conftest.py?

tianyu-l (commented):
Thanks, would like to ask some clarifying question.

> Converts absolute paths (e.g., /docs/metrics.md) to relative paths (./docs/metrics.md). While this works on GitHub and VSCode, linters don’t read them.

While "what is this" works on GitHub and VSCode, "which linters" don't read "what are them"?

CONTRIBUTING.md
tianyu-l: this was moved to https://github.com/pytorch/torchtitan/blob/main/benchmarks/llama3_h100_202412_torchtitan.md, we can suggest putting new benchmark results in the folder

README.md
tianyu-l: now in torchtitan/distributed/pipeline_parallel.py, but I think this README is in general a bit outdated.

.github/workflows/check-links.yaml
Andrei-Aksionov: Yes and no :)

Yes, because with external links we can have errors that doesn't mean that the link is incorrect, rather than a webservices blocks it (behind Cloudflare, perhaps just block bots from scraping data).
So we need to somehow treat such HTTP codes as "passed".

No, because I finally found how to properly set up a tool that is designed for checking links.
Pytest-check-links is easy to work with and doesn't throw ambiguous errors, but very limited.

Andrei-Aksionov:
> While "what is this" works on GitHub and VSCode, "which linters" don't read "what are them"?

In my head tools that check links are linters. Sorry for the confusion 🙃.
I'll refer to them as "Link Checkers" from now on.

So these  "Link Checker" might not work properly with absolute paths. Like `pytest-check-links`.

---

After a good sleep I found a way to configure Lychee (a modern "Link Checker" that is designed for this), which gives us:
1. Keep absolute paths
2. We don't need `conftest.py` anymore, Lychee can work with redirects, 403s, ... 
3. And it works in `act` (a tool for running CI actions locally for fast development/debugging).

The config would look like this:

```yaml
name: Check Links

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
  workflow_dispatch:

concurrency:
  group: link-check-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
  cancel-in-progress: true

jobs:
  check-links:
    runs-on: ubuntu-latest

    steps:
      - name: Check out repo
        uses: actions/checkout@v4

      - name: Lychee Link Checker
        uses: lycheeverse/lychee-action@v2
        with:
          # --accept: treats 403, 429, 503 as successful (ignores Cloudflare/Rate limits)
          # --exclude: ignores localhost and internal patterns
          # --max-retries: handles temporary "Connection reset" or 503s
          # --user-agent: mimics a browser to reduce 403 Forbidden errors
          args: >
            --verbose
            --no-progress
            --accept 200,403,429,503
            --exclude "http://localhost:.*"
            --exclude "http://127.0.0.1:.*"
            --max-retries 3
            --retry-wait-time 5
            --user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"
            --root-dir .
            "**/*.md"
        env:
          # By passing the GITHUB_TOKEN, Lychee uses the official GitHub API to check
           # any github.com links. This drastically reduces the 429/503 "Rate Limit" errors
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

```

As you can see, it doesn't look as "ad-hocy" as the current one.

If there is no objection, I'll pivot to this action.

CONTRIBUTING.md
Andrei-Aksionov: Done.

README.md
Andrei-Aksionov: Updated.

Andrei-Aksionov:
Hey @joecummings 

> As you already somewhat address, the biggest concern is the ability to consistently reach these links when CI is running. This can be a good signal if working properly but I'd like to avoid a state wherein CI is frequently red b/c links are inaccessible due to factors outside our control.

Totally agree. That's why I am cautious with the provided solution and try to ~double~ triple-check everything.
So far it works on Act and in my fork.

> you call this out by filtering 403 errors, but currently in the CI I see a failing test due to a Connection Reset By Peer so let's make sure all these cases are covered.

This `Connection Reset By Peer` error didn't happen before.
Apparently, a website sees a high-speed automated request and instantly sever the connection to prevent scraping or DDoS attacks.

Instead of trying to make sure that `pytest-check-links` works in this case, I decided to give one more chance to `Lychee` and this time it worked.

> Can you double check the interaction Lychee and act? You called it out in the PR description as not working but in your updated comment, you say it does. Would love some confirmation either way.

Turns out, it was a skill issue 🫠.
Lychee works fine with act, just config should be correct, otherwise you'll see "tokio" errors.
But once it was configured properly, it worked without issues: both on GitHub and in Act.
(I ran Act on my MacBook with M3 Pro, so even there, with translation layers, it works.)

> While you're here, I think some of the markdown in the README point to the correct location but have incorrect displays e.g. display text says model/model.py but point elsewhere.

Good catch. I changed the path to the correct location, but not the displayed name.
Fixed.

CONTRIBUTING.md
tianyu-l: oh please rewrite this sentence -- people should create new files under this folder, instead of updating the existing files.

Andrei-Aksionov: My bad.
How about now?

Andrei-Aksionov:
Hey @tianyu-l 

Yes, it is possible to put Lychee in [pre-commit](https://github.com/lycheeverse/lychee?tab=readme-ov-file#pre-commit-usage). I agree it’s faster for contributors to catch broken links during `git commit`, but there are a few nuances to handle:

1.  **Force Lychee to run against all markdown files.** By default, `pre-commit` only runs against staged files. if a file is renamed (e.g., `train.py` -\> `training.py`) but the `README.md` referencing it isn't changed, a standard check will skip the README and miss the broken link.

      * **Solution:** We should trigger the hook to scan all markdown files regardless of which files were modified.

2.  **Checking external links is the most time-consuming part.** This will slow down every `git commit` and fail if the developer has no internet.

      * **Proposal:** Locally, we run Lychee in `--offline` mode. It will instantly validate that local file paths exist on disk but ignore `https://` URLs. We can still run a full external check locally via a manual `pre-commit run --all-files` command if needed.

3.  **In CI, add a separate step in `lint.yaml`.** 

      * **Workflow:** Right after the Python linters, we trigger a global Lychee scan. This step will check both internal and external links.

**Note on Behavior Divergence:** This setup means a `git commit` might pass locally (since it only checks local files), but the GitHub CI might still fail if an external link is broken. To avoid this, contributors can manually run `pre-commit run lychee --all-files --hook-stage manual` to verify everything before pushing.

Andrei-Aksionov:
Lychee can parse all text files, not just markdown. For instance, if a Python file contains a comment like 
\# See: https://pytorch.org/docs/…
... it will be verified.
Verifying such links is crucial because they often explain why certain decisions were made.

Should we restrict to markdown files only, or allow checking all text files (Lychee’s default)?

tianyu-l:
> Should we restrict to markdown files only, or allow checking all text files (Lychee’s default)?

Verifying all text files sounds good, as long as there is not much slowdown running `pre-commit` locally.

> To avoid this, contributors can manually run pre-commit run lychee --all-files --hook-stage manual to verify everything before pushing.

Please add a note on this after https://github.com/pytorch/torchtitan/blob/main/CONTRIBUTING.md?plain=1#L17.

torchtitan/experiments/graph_trainer/simple_fsdp.py
```diff
@@ -54,8 +54,6 @@ def _distribute_dtensor(
     This helps enable Simple FSDP + TP/EP, in which
         inner spec/mesh is TP/EP spec/mesh
         outer spec/mesh is FSDP/DDP/HSDP spec/mesh
-    The logic follows
-    https://github.com/pytorch/pytorch/blob/main/torch/distributed/_composable/fsdp/_fsdp_param.py#L261
```
Andrei-Aksionov, line 58: This link is unreachable. Not sure what should be instead of it.

Andrei-Aksionov:
Hey @tianyu-l 

Updated the Lychee integration to work as a **pre-commit hook**. As usual, it required a bit of wrestling to get right.

1. **System-based Hook:** Standard `pre-commit` setups for Lychee require `Cargo` to compile from source. I decided that expecting a system-level install is much cleaner. If Lychee isn't found, the hook gracefully skips and prints a message on how to install it, rather than breaking the dev loop.
2. **Tiered Speed:** Since local hyperlink checks are nearly instant but external URL checks are slow, `git commit` only validates internal links by default. To run the full suite (external URLs + Python source scanning), you can trigger the manual stage: `pre-commit run --all-files --hook-stage manual`. (CI does exactly that.)
3. **CI Troubleshooting:** The CI setup was a bit of a headache. I was seeing "ghost" errors for files like `docs/docs/TROUBLESHOOTING.md` and such that shouldn't exist. Most likely installing the heavy project dependencies was generating extra documentation folders in the environment that Lychee was picking up. Moving the Lychee check to run **before** the dependency installation fixed the noise.

tianyu-l:
I'm seeing lychee linter skipped here https://github.com/pytorch/torchtitan/actions/runs/23456873170/job/68249262800?pr=2632#step:9:42

Is this expected?

Andrei-Aksionov:
Yes, it was expected.

Previously, there were two stages:
1. **Locally check internal hyperlinks against ALL markdown files.** Since `git commit` by default only checks changed files, I originally forced it to scan the whole repo to catch cases where a python file is renamed, but the markdown files linking to it aren't.
2. **Check hyperlinks and external links for all markdown and python files only when run explicitly.**

In the CI, I skipped that check because the step before it was already performing the full scan.

---

Now I have simplified the design of the Lychee pre-commit hooks.

There is now one pre-commit hook that checks both hyperlinks and external links. 
Because `git commit` only runs against **changed files**, it won't warn you if a renamed file breaks a link in an *untouched* README. However, running `pre-commit run --all-files` will still catch those issues and provide a full audit.

The CI is now configured to do exactly that (run against all files), which makes the overall behavior more obvious and consistent.

Andrei-Aksionov:
My first contribution to the repo!



---

Btw, thanks for the comments/questions/suggestions.
The final solution turned out quite nicely.

PR #2633, opened 2026-03-19
[Module] Refactor init_weights to config-based param_init system

fegin:
* #2785
* #2780
* __->__ #2633

## Motivation

Parameter initialization should be configurable without modifying module code. Previously, each module defined its own `init_weights` method with hardcoded init logic (e.g., `nn.init.trunc_normal_` with a fixed std). This couples initialization strategy to module definitions, making it impossible to change init behavior without editing model code.

## Design
- **Expand the config to be per-module config**
- `param_init` is a `dict[str, Callable]` on `Module.Config`, mapping local parameter names to init functions. It is set on every sub-config in the model config registry.
- `init_states` auto-recurses the module tree (children first, then self), calling `_init_self_parameters` and `_init_self_buffers` on each module. Buffer initialization remains module-specific (e.g., RoPE cache, MoE counters) via `_init_self_buffers` overrides.
- `DeferredCallable.Config` wraps init functions that vary by layer depth (e.g., depth-scaled std). `resolve_deferred` walks the config tree and resolves these markers during per-layer config expansion.
- `expand_layer_configs` (per model) deep-copies the layer template N times, resolves deferred markers, and handles model-specific layer interleaving (MoE vs dense, iRoPE, sliding attention).

## Summary of changes

- Remove all `init_weights` methods from modules (Linear, Embedding, RMSNorm, FeedForward, GQAttention, MoE, GroupedExperts, TokenChoiceTopKRouter, RoPE, Decoder).
- Add `param_init` field to `Module.Config`; `Config.build()` propagates it to the instance as `_param_init`.
- Add `init_states` with auto-recursion, `_init_self_parameters` (dict-lookup init), and `_init_self_buffers` (module-specific buffer init).
- Add `DeferredCallable` (in `torchtitan/config/`) for depth-varying init.
- Add `param_init.py` (in `torchtitan/models/common/`) with `resolve_deferred`, `skip_param_init`, and `depth_scaled_std` utilities.
- Remove `init_mean`/`init_std` fields from `Linear.Config` and `Embedding.Config` — init is now expressed entirely through `param_init` dicts.
- Add `layers` list and `expand_layer_configs()` pattern to all model registries, replacing per-layer config construction in `Decoder.__init__`.
- Add `w1`/`w2w3` fields to `FeedForward.Config` and `wqkv`/`wo` fields to `GQAttention.Config` (replacing `linear_bias: bool`), so each linear sub-module gets its own `param_init`.

## Verification

Bitwise identical loss against `origin/main` with `--debug.seed=42 --debug.deterministic`:

| Model  | Steps | Result            |
|--------|-------|-------------------|
| Llama3 | 10    | Bitwise identical |

### DIFF
--- a/torchtitan/models/common/param_init.py
+++ b/torchtitan/models/common/param_init.py
@@ -13,7 +13,7 @@ registry::
     Linear.Config(param_init={"weight": partial(trunc_normal_, std=0.02), "bias": zeros_})
     RMSNorm.Config(param_init={"weight": nn.init.ones_})
 
-``DeferredCallable.Config`` wraps a factory for init that varies by layer
+``Function.Config`` wraps a factory for init that varies by layer
 depth.  ``resolve_deferred`` walks a config tree and resolves all deferred
 markers during layer expansion.
 """
@@ -22,11 +22,11 @@ import dataclasses
 
 import torch.nn as nn
 
-from torchtitan.config import DeferredCallable
+from torchtitan.config import Function
 
 
 def resolve_deferred(cfg: object, layer_id: int) -> None:
-    """Walk a dataclass config tree and resolve all ``DeferredCallable.Config`` fields.
+    """Walk a dataclass config tree and resolve all ``Function.Config`` fields.
 
     Mutates *cfg* in place (intended for use on a deep-copied config).
     """
@@ -37,7 +37,7 @@ def resolve_deferred(cfg: object, layer_id: int) -> None:
             val = getattr(cfg, f.name)
         except AttributeError:
             continue
-        if isinstance(val, DeferredCallable.Config):
+        if isinstance(val, Function.Config):
             object.__setattr__(cfg, f.name, val.build()(layer_id))
         elif dataclasses.is_dataclass(val):
             resolve_deferred(val, layer_id)
--- a/torchtitan/models/llama4/__init__.py
+++ b/torchtitan/models/llama4/__init__.py
@@ -12,7 +12,7 @@ import torch.nn as nn
 
 from torchtitan.components.loss import build_cross_entropy_loss
 from torchtitan.components.optimizer import register_moe_load_balancing_hook
-from torchtitan.config import DeferredCallable
+from torchtitan.config import Function
 from torchtitan.distributed.pipeline_parallel import pipeline_llm
 from torchtitan.models.common import (
     compute_ffn_hidden_dim,
@@ -74,7 +74,7 @@ _LINEAR_INIT = {
     "weight": partial(nn.init.trunc_normal_, std=0.02),
     "bias": nn.init.zeros_,
 }
-_LINEAR_DEPTH_INIT = DeferredCallable.Config(
+_LINEAR_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "weight": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
         "bias": nn.init.zeros_,
@@ -82,7 +82,7 @@ _LINEAR_DEPTH_INIT = DeferredCallable.Config(
 )
 _NORM_INIT = {"weight": nn.init.ones_}
 _EMBEDDING_INIT = {"weight": partial(nn.init.normal_, std=1.0)}
-_EXPERTS_DEPTH_INIT = DeferredCallable.Config(
+_EXPERTS_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "w1": partial(nn.init.trunc_normal_, std=0.02),
         "w2": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
--- a/torchtitan/models/qwen3/__init__.py
+++ b/torchtitan/models/qwen3/__init__.py
@@ -12,7 +12,7 @@ from functools import partial
 import torch.nn as nn
 
 from torchtitan.components.loss import build_cross_entropy_loss
-from torchtitan.config import DeferredCallable
+from torchtitan.config import Function
 from torchtitan.distributed.pipeline_parallel import pipeline_llm
 from torchtitan.models.common import Embedding, FeedForward, GQAttention, Linear, RoPE
 from torchtitan.models.common.attention import FlexAttention, VarlenAttention
@@ -59,7 +59,7 @@ _LINEAR_INIT = {
     "weight": partial(nn.init.trunc_normal_, std=0.02),
     "bias": nn.init.zeros_,
 }
-_LINEAR_DEPTH_INIT = DeferredCallable.Config(
+_LINEAR_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "weight": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
         "bias": nn.init.zeros_,
@@ -68,7 +68,7 @@ _LINEAR_DEPTH_INIT = DeferredCallable.Config(
 _NORM_INIT = {"weight": nn.init.ones_}
 _EMBEDDING_INIT = {"weight": partial(nn.init.normal_, std=1.0)}
 _EMBEDDING_SKIP_INIT = {"weight": skip_param_init}
-_EXPERTS_DEPTH_INIT = DeferredCallable.Config(
+_EXPERTS_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "w1": partial(nn.init.trunc_normal_, std=0.02),
         "w2": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
--- a/torchtitan/protocols/module.py
+++ b/torchtitan/protocols/module.py
@@ -11,7 +11,7 @@ from typing import Any
 import torch
 import torch.nn as nn
 
-from torchtitan.config import Configurable, DeferredCallable
+from torchtitan.config import Configurable, Function
 
 
 # Cache: maps nn.Module subclass -> created Module wrapper class.
@@ -35,7 +35,7 @@ class Module(nn.Module, Configurable):
 
     @dataclass(kw_only=True, slots=True)
     class Config(Configurable.Config):
-        param_init: dict | DeferredCallable.Config | None = None
+        param_init: dict | Function.Config | None = None
 
         def build(self, **kwargs):
             # slots=True prevents super().build() from working; call explicitly.
--- a/tests/unit_tests/test_weight_tying.py
+++ b/tests/unit_tests/test_weight_tying.py
@@ -9,7 +9,7 @@ from functools import partial
 
 import torch.nn as nn
 
-from torchtitan.config import DeferredCallable
+from torchtitan.config import Function
 from torchtitan.models.common.attention import GQAttention
 from torchtitan.models.common.embedding import Embedding
 from torchtitan.models.common.feed_forward import compute_ffn_hidden_dim, FeedForward
@@ -24,7 +24,7 @@ _DUMMY_LINEAR_INIT = {
     "weight": nn.init.zeros_,
     "bias": nn.init.zeros_,
 }
-_LINEAR_DEPTH_INIT = DeferredCallable.Config(
+_LINEAR_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "weight": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
         "bias": nn.init.zeros_,
--- a/torchtitan/config/__init__.py
+++ b/torchtitan/config/__init__.py
@@ -21,13 +21,13 @@ from .configs import (
     TrainingConfig,
 )
 from .configurable import Configurable
-from .deferred_callable import DeferredCallable
+from .function import Function
 from .manager import ConfigManager
 
 __all__ = [
     "ConfigManager",
     "Configurable",
-    "DeferredCallable",
+    "Function",
     "TORCH_DTYPE_MAP",
     # Config dataclasses
     "ActivationCheckpointConfig",
--- a/torchtitan/config/function.py
+++ b/torchtitan/config/function.py
@@ -13,7 +13,7 @@ from torchtitan.config.configurable import Configurable
 R = TypeVar("R")
 
 
-class DeferredCallable(Generic[R], Configurable):
+class Function(Generic[R], Configurable):
     """A Configurable wrapper around any callable.
 
     Wraps a callable in a ``Config`` with ``.build()`` support, fitting into
@@ -22,7 +22,7 @@ class DeferredCallable(Generic[R], Configurable):
     Use for config fields whose values are computed lazily — e.g., per-layer
     parameter initialization that varies by depth::
 
-        DeferredCallable.Config(
+        Function.Config(
             fn=lambda layer_id: {
                 "weight": partial(trunc_normal_, std=0.02 / (2 * (layer_id + 1)) ** 0.5),
             }
--- a/torchtitan/models/deepseek_v3/__init__.py
+++ b/torchtitan/models/deepseek_v3/__init__.py
@@ -12,7 +12,7 @@ import torch.nn as nn
 
 from torchtitan.components.loss import build_cross_entropy_loss
 from torchtitan.components.optimizer import register_moe_load_balancing_hook
-from torchtitan.config import DeferredCallable
+from torchtitan.config import Function
 from torchtitan.distributed.pipeline_parallel import pipeline_llm
 from torchtitan.models.common import Embedding, FeedForward, Linear, RMSNorm, RoPE
 from torchtitan.models.common.attention import FlexAttention
@@ -56,7 +56,7 @@ _LINEAR_INIT = {
     "weight": partial(nn.init.trunc_normal_, std=0.02),
     "bias": nn.init.zeros_,
 }
-_LINEAR_DEPTH_INIT = DeferredCallable.Config(
+_LINEAR_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "weight": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
         "bias": nn.init.zeros_,
@@ -64,7 +64,7 @@ _LINEAR_DEPTH_INIT = DeferredCallable.Config(
 )
 _NORM_INIT = {"weight": nn.init.ones_}
 _EMBEDDING_INIT = {"weight": partial(nn.init.normal_, std=1.0)}
-_EXPERTS_DEPTH_INIT = DeferredCallable.Config(
+_EXPERTS_DEPTH_INIT = Function.Config(
     fn=lambda layer_id: {  # pyrefly: ignore [bad-argument-type]
         "w1": partial(nn.init.trunc_normal_, std=0.02),
         "w2": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
--- a/torchtitan/models/gpt_oss/__init__.py
+++ b/torchtitan/models/gpt_oss/__init__.py
@@ -12,7 +12,7 @@ import torch.nn as nn
 
 from torchtitan.components.loss import build_cross_entropy_loss
 from torchtitan.components.optimizer import register_moe_load_balancing_hook
-from torchtitan.config import DeferredCallable
+from torchtitan.config import Function
 from torchtitan.models.common import Embedding, Linear, RMSNorm, RoPE
 from torchtitan.models.common.moe import TokenChoiceTopKRouter
 from torchtitan.models.common.param_init import depth_scaled_std, resolve_deferred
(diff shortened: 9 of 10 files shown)


torchtitan/config/configurable.py
```diff
@@ -82,6 +82,9 @@ def to_dict(self) -> dict:
                 except AttributeError:
                     # field(init=False) not yet set, ignore this field.
                     continue
+                if callable(val) and not dataclasses.is_dataclass(val):
```
tianyu-l: oh, can a dataclass ever be callable?

torchtitan/models/common/moe/moe.py
```diff
@@ -526,17 +514,10 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             return out_experts.reshape(bs, slen, dim)
         return (out + out_experts).reshape(bs, slen, dim)
 
-    def init_weights(self, **kwargs) -> None:
-        init_std = kwargs.get("init_std")
+    def _init_self_buffers(self, **kwargs) -> None:
```
tianyu-l: should the arg be renamed to `device` and displayed explicitly instead of using `**kwargs`?

torchtitan/trainer.py
```diff
@@ -401,7 +401,7 @@ def __init__(self, config: Config):
 
             model.to_empty(device=init_device)
             with torch.no_grad():
-                cast(BaseModel, model).init_weights(buffer_device=buffer_device)
+                cast(BaseModel, model).init_states(buffer_device=buffer_device)
```
tianyu-l: not related to this PR: why cast to Decoder above but cast to BaseModel here?

torchtitan/models/common/decoder.py
```diff
@@ -101,42 +99,29 @@ def __init__(self, config: Config):
...
+        if "buffer_device" not in kwargs or kwargs["buffer_device"] is None:
+            kwargs = {**kwargs, "buffer_device": self.freqs_cis.device}
+        super().init_states(param_init=param_init, param_prefix=param_prefix, **kwargs)
+
+    def _init_self_buffers(self, **kwargs) -> None:
         buffer_device: torch.device | None = kwargs.get("buffer_device")
-        buffer_device = buffer_device or self.freqs_cis.device
         if self.rope is not None:
-            self.rope.init_weights(buffer_device=buffer_device)
+            # RoPE's _init_self_buffers was already called by auto-recursion
```
tianyu-l, line 126: Not sure about this, but it's possible that it is set on meta device, when `buffer_device` here is cuda? Shall we add an assertion for device?

torchtitan/protocols/module.py
```diff
@@ -4,12 +4,37 @@
...
 from typing import Any
 
 import torch.nn as nn
 
 from torchtitan.config import Configurable
 
+# Type alias for simple parameter initializers: (param) -> Any
+# Uses Any return type because nn.init.* functions return Tensor,
+# but the return value is always ignored by the dispatch layer.
+ParamInitializer = Callable[[nn.Parameter], Any]
```
tianyu-l: If it's only used in common/param_init.py, can we put it there for now?

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
+        while queue:
+            child_prefix, child = queue.pop()
+            if isinstance(child, Module):
+                child.init_states(
+                    param_init=param_init, param_prefix=child_prefix, **kwargs
+                )
+            else:
+                # Plain nn.Module (e.g., CheckpointWrapper) — look inside
+                # for Module descendants.
+                queue.extend(prefixed_children(child, child_prefix))
```
tianyu-l: oh this is tricky

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
+        if param_init is None:
+            raise ValueError(
+                f"No param_init found for {type(self).__name__}. "
+                f"Set param_init on this module's Config or on an "
+                f"ancestor's Config."
+            )
+        for name, param in params:
+            param_init(f"{param_prefix}{name}", param)
+
+    def _init_self_buffers(self, **kwargs) -> None:
```
tianyu-l:
```suggestion
    def _init_self_buffers(self, device) -> None:
```

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
-    def init_weights(self, **kwargs) -> None:
-        """Initialize weights. Override in subclasses with learnable parameters."""
+    param_init: NamedParamInitializer | None = None
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        param_init: NamedParamInitializer | None = None
+
+        def build(self, **kwargs):
+            instance = Configurable.Config.build(self, **kwargs)
```
tianyu-l, line 44: curious what's different from `super().build(**kwargs)`?

torchtitan/models/qwen3/__init__.py
```diff
@@ -25,11 +30,24 @@
...
 
+
+def _qwen3_param_init(
+    dim: int, n_layers: int, weight_tying: bool = False
+) -> RegexInitializer:
+    patterns = make_decoder_param_init(dim=dim, n_layers=n_layers)
+    if weight_tying:
+        # tok_embeddings.weight is tied to output.weight; skip its init
+        # so only output.weight's truncated-normal init applies.
+        patterns[r"tok_embeddings\.weight"] = SKIP_PARAM_INIT
```
tianyu-l: may need this for llama3 as well after https://github.com/pytorch/torchtitan/pull/2580

torchtitan/models/common/param_init.py
```diff
@@ -0,0 +1,145 @@
...
+        assert m is not None, (
+            f"Could not parse layer_id from FQN '{name}'. "
+            f"Expected FQN to contain 'layers.<digit>.'."
+        )
+        layer_id = int(m.group(1))
+        std = self._base_std / (2 * (layer_id + 1)) ** 0.5
+        nn.init.trunc_normal_(param, mean=0.0, std=std, a=self._a, b=self._b)
+
+
+def make_decoder_param_init(
```
tianyu-l:
```suggestion
def decoder_param_init(
```

torchtitan/models/common/param_init.py
```diff
@@ -0,0 +1,145 @@
...
+import re
+from functools import partial
+
+import torch.nn as nn
+
+from torchtitan.protocols.module import NamedParamInitializer, ParamInitializer
+
+# No-op initializer: explicitly skip initialization for a parameter.
+# Useful when a parameter is tied to another (e.g., weight tying).
+SKIP_PARAM_INIT: ParamInitializer = lambda param: None
```
tianyu-l: I'd prefer we make this a subclass of `ParamInitializer`, e.g. named `NoParamInitializer`, instead of a macro.

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
+       Override for device-aware buffer init (e.g., RoPE, MoE).
+
+    Subclasses should NOT override ``init_states`` unless they need custom
+    ordering (e.g., weight tying before init). Override ``_init_self_buffers``
+    for buffer initialization.
     """
 
-    def init_weights(self, **kwargs) -> None:
-        """Initialize weights. Override in subclasses with learnable parameters."""
+    param_init: NamedParamInitializer | None = None
```
tianyu-l:
```suggestion
    _param_init: NamedParamInitializer | None = None
```

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
+            param_prefix: FQN prefix accumulated during recursion, so that
+                ``param_init`` receives fully-qualified parameter names.
+            **kwargs: Forwarded to children and self-init methods
+                (e.g., ``buffer_device``).
+        """
+        # If this module owns a param_init, it becomes the active one
+        # and the prefix resets (FQNs are relative to the owner).
+        if self.param_init is not None:
+            param_init = self.param_init
+            param_prefix = ""
```
tianyu-l: Not sure about this.
E.g. let's say I created a new module to plug in each layer, for which I want to use depth init.
- If I provide a `config.param_init` to this module, I would lose the layer id because of this line
- Of course, I can change the overall decoder init. But if that's the case, what's the point of supporting `param_init` on each module? -- we could always call `get_default_decoder_init` and modify the dictionary.

torchtitan/config/configurable.py
tianyu-l: btw, https://fburl.com/code/8t7xhgna is making `Config.param_init` a config, not a function, so that it should still be serializable. I think we should do this too, to make it reproducible (given a code version).

Right now parallelize_fn and other functions in ModelSpec are not serializable, which should be fixed later.

torchtitan/experiments/vlm/model/model.py
```diff
@@ -72,12 +63,14 @@ def __init__(self, config: Config):
...
-        if self.projector is not None:
-            self.projector.init_weights()
+    def init_states(
+        self,
+        *,
+        param_init: NamedParamInitializer | None = None,
+        param_prefix: str = "",
+        **kwargs,
+    ) -> None:
+        super().init_states(param_init=param_init, param_prefix=param_prefix, **kwargs)
```
tianyu-l: can we remove this function?

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
+    param_init: NamedParamInitializer | None = None
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        param_init: NamedParamInitializer | None = None
+
+        def build(self, **kwargs):
+            instance = Configurable.Config.build(self, **kwargs)
+            if self.param_init is not None:
+                instance.param_init = self.param_init
```
tianyu-l: we could do this in constructor, curious why we need to do it here

torchtitan/protocols/module.py
```diff
@@ -20,13 +45,116 @@ class Module(nn.Module, Configurable):
...
+            return instance
+
+    def init_states(
+        self,
+        *,
+        param_init: NamedParamInitializer | None = None,
```
tianyu-l: This seems different from https://fburl.com/code/ia9w7w5i
- over there, if the param is missing then rely on the parent
- here, if sending the parent param_init to all children. I think this is leaking information from global to local modules, so should be avoided in general.

torchtitan/config/configurable.py
fegin: The `config` is a dictionary that maps from reg to function. So from serializability perspective, the requirement is the same?

torchtitan/trainer.py
fegin: Should be BaseModel

torchtitan/models/common/decoder.py
fegin: Not quite sure if I understand the question. If meta device is used, then everything is meta device. `buffer_device` is only used when trainer parallelizes the model and tries to re-initialize the states again. The comment here just want to explain the order, not about the device though.

Assertion is good, but I want to understand if I miss something.

torchtitan/protocols/module.py
fegin: That will be inconsistent. We will have ParamInitializer in `common/param_init.py` but NamedParamInitializer being here. Ideally, we should put both in `common/param_init.py`. But I cannot do this due to the recursive import.

torchtitan/protocols/module.py
fegin: Do you mean `CheckpointWrapper`? Yes, same as torch.compile.

torchtitan/protocols/module.py
fegin: With `slot=True`, you cannot call `super()`, that's the restriction. I should add comment though.

torchtitan/protocols/module.py
fegin: Because we don't have a common `Module.__init__`, so users who define a new Module needs to remember to do the assignment. Or an alternative is to define a common Module.__init__. I thought about the later case, but that requires some reworks of module inheritance hierarchy and users now have to call `super.__init__(cfg, *args, *kwargs)`. The later constraint is okay guess but the former is something need sometime to verify. If we do want to change this, we can later do another round of change to the Module hierarchy.

torchtitan/protocols/module.py
fegin: I'm not sure though. Parent modules already hold the `param_init`, is that considered as leaking from global to a local? I didn't do the parent walk because unlike the internal code, Node, that has parent, we don't have this implementation in our Module. So I deliberately didn't include that implementation. Adding `parent` will also cause circular dependency and make GC slower. But the later concern should not be a real concern as we never GC modules. Just provide this minor issue.

torchtitan/models/common/param_init.py
fegin: This function itself is NOT a param_init, it creates one. So `make_` prefix seems to be more accurate.

torchtitan/models/common/param_init.py
fegin: okay, I change it to a Function not lambda, ParamInitializer is just a Callable alias.

torchtitan/protocols/module.py
tianyu-l: I see. Can we use `super(Config, self).build(**kwargs)`?

fegin:
I change the parameter init to ancestor-lookup style. But the comment reflects an issue brought by the circular reference. So the original concern of parent reference is legit. This needs some careful handle. But the code looks cleaner for each inherited `init_state`. 

The typing of param_init is still not changed. As long as there is an unnamed function, which the internal code also does, serialization is not sound. `DepthScaledTruncNormal` and `RegexInitializer` are serializable if we correctly implement `to_dict` and `from_dict`. But I think this rule stays even if we change param_init to config. Correct me if I'm wrong. 

@tianyu-l

tianyu-l:
@claude review this pull request. In particular, please comment on where is best to expand general transformer layer config into per-layer configs. I think in general there are three options:
1. There is no per-layer config. The `TransformerBlock` constructor has the logic to create submodules. This is the current behavior before this PR.
2. The model config class has a method to expand into per-layer config. This is the behavior in the current commit of the PR.
3. Alternatively, the model constructor only accepts already-expanded per-layer config. The (optional) logic of "creating the per-layer config based on a general config using some per-model logic" can be put outside the model code, e.g. in `config_registry.py`.

tianyu-l:
@claude Let me clarify.

For option 2, yes we could just remove the logic in `TransformerBlock.__init__()` and only use `expand()` in the config.

For option 3, I didn't mean putting the expansion logic in `config_registry.py`; instead I meant that we can put it in `model/__init__.py` (e.g. in the `model_registry(flavor)` function) where we construct model config.
- This still couples model with expansion logic. Each model has its own expansion logic.
- The logic is flexible and optional. One could have written expanded logic directly.
- I would argue that this provides better encapsulation between model core logic, because what each model looks like is completely determined by the expanded per-layer config, and it doesn't matter how this config is provided, via expansion or not. Putting in each model subclass would prescribe how users can provide the final per-layer configs.
- I disagree with what you said "This breaks the principle that model-specific knowledge should be co-located with the model.", because in the first place although `model.py` holds model-specific components, it doesn't know how to construct the model combining these model-specific components and common components. The actual model composition logic has largely been existing in `model/__init__.py`. So I would say option 3 (moving expansion to `model/__init__.py`) gives better encapsulation, in that `model.py` now only holds model-specific components that's not defined for other models (thus not in `common/`), and `model/__init__.py` holds the composition logic.
- Also if we go with option 3, we can remove classes such as `Llama4TransformerBlock` because the base class `TransformerBlock` can hold general logic.

torchtitan/protocols/module.py
fegin: No, it doesn't work.

torchtitan/trainer.py
```diff
@@ -377,7 +377,9 @@ def __init__(self, config: Config):
             for m in self.model_parts:
                 m.to_empty(device=init_device)
                 with torch.no_grad():
-                    cast(Decoder, m).init_weights(buffer_device=buffer_device)
+                    # TODO: Change this back to init_weights once
```
tianyu-l, line 380: what does this TODO mean? It's already `init_weights`.

torchtitan/models/common/decoder.py
```diff
@@ -79,7 +77,8 @@ class Config(BaseModel.Config):
         # and Attention. Also RoPE itself as a standalone module requires PP special
         # handling, see below.
         rope: RoPE.Config
-        layer: TransformerBlock.Config  # required, no default
+        layer: TransformerBlock.Config
```
tianyu-l, line 80: I had thought we don't need to put this in `Decoder.Config`. It's a pure config time thing, which evaporates after config construction.

torchtitan/models/common/feed_forward.py
```diff
@@ -42,26 +42,21 @@ class FeedForward(Module):
     @dataclass(kw_only=True, slots=True)
     class Config(Module.Config):
         hidden_dim: int
-        linear_bias: bool = False
+        w1: Linear.Config
+        w2w3: Linear.Config
```
tianyu-l, line 46: uh is this because w2w3 share the same param init fn? It sounds unusual to couple them together, but we can revisit this later.

torchtitan/config/deferred_callable.py
```diff
@@ -0,0 +1,40 @@
...
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import Any, Generic, TypeVar
+
+from torchtitan.config.configurable import Configurable
+
+R = TypeVar("R")
+
+
+class DeferredCallable(Generic[R], Configurable):
```
tianyu-l: `DeferredCallable` sounds a bit contrived. I don't have a good name, but maybe we can just call it `Function`, in the sense that it's a generic configurable function, just like `Module` is a generic configurable Module.

torchtitan/models/llama4/__init__.py
```diff
@@ -30,133 +40,257 @@
...
+        if cfg.every_n_layers_nope is not None:
+            if layer_id % cfg.every_n_layers_nope == 0:
+                cfg = replace(cfg, attention=replace(cfg.attention, use_rope=False))
+        # MoE interleaving: keep only the appropriate FFN type per layer
+        moe_enabled = (layer_id + 1) % cfg.interleave_moe_layer_step == 0
+        if moe_enabled:
+            cfg = replace(cfg, feed_forward=None)
+        else:
+            cfg = replace(cfg, moe=None)
+        resolve_deferred(cfg, layer_id)
```
tianyu-l, line 68: Oh I think I understood the design.

Right now you are doing:

1. At initial config construction time:
    - create a single layer config (pending expansion) INSIDE the original model config
    - assign (deferred) init function `_LINEAR_DEPTH_INIT` to each module during 
2. call `expand_layer_configs` which
    - expand single layer config to multi-layer
    - calls `resolve_deferred` to resolve the init function onto each layer's module

I think it has two minor issues:
- the single-layer config gets exposed into final official config, which is unnecessary.
- the deferred init is unnecessarily indirect, and seems overfitting torchtitan's existing depth-init paradigm.
    - I don't know if this "depth init" is a good practice or not, it was introduced by someone and never changed, so we probably shouldn't overfit.
    - If users want something different, not based on layer-id, but e.g. based on if this transformer block has MoE or not (I made this up), then `resolve_deferred` becomes useless to them and they have to change `expand_layer_configs` -- not that they shouldn't, but I think we don't have to program this way.

Concretely I would suggest the following:
- Move config.layer out from `Decoder.Config`. We can first construct `Llama4TransformerBlock` as the thing to be expanded (if that's more convenient than passing dict around), send it to `expand_layer_configs` to construct a `layers` stacked transformer block, and use the result to construct `Llama4Model`.
- Don't assign `param_init` on transformer blocks before layer expansion -- e.g. we can hardcode `param_init` for each module inside the for-loop of layer expansion, as we are anyways providing one expansion function for each model.

torchtitan/models/llama4/__init__.py
```diff
@@ -30,133 +40,257 @@
...
+    n_heads = 40
+    n_kv_heads = 8
+    moe_hidden_dim = compute_moe_hidden_dim(
+        dim,
+        multiple_of=2048,
+        ffn_dim_multiplier=1.2,
+        top_k=1,
+        num_shared_experts=1,
+    )
+    return Llama4Model.Config(
```
tianyu-l, line 230: nit: if you are already doing "first define the shared args, then init the actual config", then it may make sense to move 2nd step out so that all models can share the `Llama4Model.Config(...)` code. But yeah, that's a step further into "config-of-config" land.

tianyu-l:
@claude review the latest commit. In particular, check if @tianyu-l 's latest comments makes sense.

torchtitan/config/deferred_callable.py
fegin: `Callable` is not a good name because it conflicts with `typing.Callable`. So this is the name suggested by Claude.

fegin:
Yes, I don't like the idea of some configurations becoming obsolete later. However, this make the config easier to read in `__init__.py` IMO. So I keep these one layer config in the Config. But we can do the `truly` programming style that remove those dictionary and use another config building blocks in `__init__.py` to construct.  Let's revisit this topic in the `Module Config Programming` PR.

I'll change `DeferredCallable` to `Function`. As for the rest, I think we can solve them in the `Module Config Programming` PR.

PR #2634, opened 2026-03-19
[graph_trainer] Remove eager AC

yiming0416:
Eager AC via `torch.utils.checkpoint` could break graph capture. Since we have graph-based AC support, this PR removes eager AC entirely in `graph_trainer` experiment and uses graph-based SAC exclusively.

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -145,20 +145,25 @@ def get_transformer_block_buckets(model) -> list[list[str] | str]:
     return module_fqns
 
 
-def maybe_disable_eager_ac(
+def apply_graph_ac(
     compile_config: CompileConfig,
     ac_config: "ActivationCheckpointConfig",
 ) -> None:
-    """Disable eager AC when apply_sac graph pass is enabled.
+    """Add apply_sac to compile joint passes for graph-based selective AC.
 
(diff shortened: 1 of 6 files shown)


aditvenk:
@yiming0416  Could this have a memory regression for SimpleFSDP (jit) memory runs? Across TorchTitan scaling runs and Ivan's work on collectives overlap
cc: @anshul-si , @IvanKobzarev

IvanKobzarev:
> @yiming0416  Could this have a memory regression for SimpleFSDP (jit) memory runs? Across TorchTitan scaling runs and Ivan's work on collectives overlap
> cc: @anshul-si , @IvanKobzarev 

In my memory runs I fixed all AC via graph approach for simpleFsdp in reshard_after_forward, it needs changes, but no eager should be fine.

yiming0416:
> > @yiming0416  Could this have a memory regression for SimpleFSDP (jit) memory runs? Across TorchTitan scaling runs and Ivan's work on collectives overlap
> > cc: @anshul-si , @IvanKobzarev
> 
> In my memory runs I fixed all AC via graph approach for simpleFsdp in reshard_after_forward, it needs changes, but no eager should be fine.

That makes sense. SimpleFSDP's `reshard_after_forward` pass is essentially doing the same logic of tagging nodes with different AC policies as the `apply_sac` pass. These two could potentially be merged in follow-ups.

PR #2635, opened 2026-03-19
[graph_trainer] Add Qwen3

yiming0416:
Add Qwen3 (dense + MoE) to the graph_trainer experiment.

Files added
- `torchtitan/experiments/graph_trainer/qwen3/__init__.py` -- Model registry
- `torchtitan/experiments/graph_trainer/qwen3/config_registry.py` -- Debug model configs (dense, flex_attn, MoE)
- `torchtitan/experiments/graph_trainer/qwen3/model.py` -- GraphTrainerQwen3Model
- `torchtitan/experiments/graph_trainer/qwen3/parallelize.py` -- Parallelization function

Handles two Qwen3-specific issues:

- Weight tying: Qwen3 ties `tok_embeddings.weight` and `output.weight`, which means `model.parameters()` returns duplicates. Therefore, we need to use `named_parameters(remove_duplicate=False)`.

- cos_sin RoPE backend: Unlike Llama3/DeepSeek which use a complex backend, Qwen3 uses a cos_sin backend that stores cos/sin caches as float32 buffers. It causes numerics issues but was fixed in #2881

Test Plan:

Since we are making `aot_fx_trace` the default mode. I only added tests for this mode

### DIFF
--- a/torchtitan/experiments/graph_trainer/qwen3/model.py
+++ b/torchtitan/experiments/graph_trainer/qwen3/model.py
@@ -1,0 +1,30 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass
+
+import torch
+
+from torchtitan.models.qwen3 import Qwen3Model
+
+from ..simple_fsdp import disable_active_parametrization
+
+
+class GraphTrainerQwen3Model(Qwen3Model):
+    @dataclass(kw_only=True, slots=True)
+    class Config(Qwen3Model.Config):
+        pass
+
+    def __init__(self, config: Config):
+        super().__init__(config)
+
+    def init_states(
+        self,
+        *,
+        buffer_device: torch.device | None = None,
+    ) -> None:
+        with disable_active_parametrization():
+            super().init_states(buffer_device=buffer_device)
--- a/torchtitan/experiments/graph_trainer/qwen3/parallelize.py
+++ b/torchtitan/experiments/graph_trainer/qwen3/parallelize.py
@@ -1,0 +1,210 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torch.distributed.device_mesh import DeviceMesh
+from torch.fx.traceback import annotate_fn
+
+from torchtitan.components.quantization.float8 import find_float8_linear_config
+from torchtitan.config import (
+    ActivationCheckpointConfig,
+    CompileConfig,
+    ParallelismConfig,
+    TORCH_DTYPE_MAP,
+    TrainingConfig,
+)
+from torchtitan.distributed import ParallelDims
+from torchtitan.distributed.tensor_parallel import maybe_enable_async_tp
+from torchtitan.experiments.graph_trainer.common_utils import (
+    annotate_ac_regions,
+    apply_graph_ac,
+)
+from torchtitan.experiments.graph_trainer.compile import apply_compile
+from torchtitan.experiments.graph_trainer.qwen3.model import GraphTrainerQwen3Model
+
+from torchtitan.experiments.graph_trainer.simple_fsdp import (
+    data_parallel,
+    MixedPrecisionPolicy,
+)
+from torchtitan.models.llama4.parallelize import apply_moe_ep_tp
+from torchtitan.models.qwen3.parallelize import apply_non_moe_tp
+from torchtitan.protocols.model_converter import ModelConvertersContainer
+from torchtitan.tools.logging import logger
+
+
+def annotate_qwen3(model: GraphTrainerQwen3Model) -> None:
+    """Attach annotations to FX graph nodes with ``torch.fx.traceback.annotate_fn``
+
+    - Expert Parallel (EP) annotations: Tags "dispatch", "combine", and "compute"
+      regions in MoE for debugging purposes (only if MoE layers exist).
+    - Flex attention annotation: Tags FlexAttention.forward with
+      {"compile_with_inductor": "flex_attention"} so the compiler can apply
+      regional inductor pass based on the annotation. Regional inductor is now only
+      supported in AOT mode.
+    - AC region annotation: Tags each transformer block's forward with a unique
+      ac_region_id so that apply_sac_pass can assign per-block ac_graph_id
+      boundaries for the min-cut partitioner.
+    """
+    from torchtitan.models.common.attention import FlexAttention
+
+    # Annotate MoE EP regions if any layer has MoE enabled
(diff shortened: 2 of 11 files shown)


torchtitan/experiments/graph_trainer/qwen3/parallelize.py
```diff
@@ -0,0 +1,235 @@
...
+    data_parallel,
+    MixedPrecisionPolicy,
+)
+from torchtitan.models.llama4.parallelize import apply_moe_ep_tp
+from torchtitan.models.qwen3.parallelize import apply_non_moe_tp
+from torchtitan.protocols.model_converter import ModelConvertersContainer
+from torchtitan.tools.logging import logger
+
+# for selective op activation checkpointing
+_op_sac_save_list = {
```
tianyu-l: fyi we've recently refactored this to have a central sac policy. We can evolve if it's too restricted.

yiming0416:
cc @IvanKobzarev @ruisizhang123 for SimpleFSDP related changes

yiming0416:
AutoDev: Review feedback for Qwen3 GraphTrainer support

Overall the PR looks solid — code structure matches the DeepSeek-v3 GraphTrainer pattern, the 3 key design decisions are correct (`init_states` override, `MixedPrecisionPolicy` from `simple_fsdp`, `remove_duplicate=False` for weight tying), and test coverage is comprehensive.

### Issues

1. **MoE import inconsistency** (`parallelize.py`): `annotate_qwen3` imports `from torchtitan.models.common.moe.moe import MoE` (double `moe` submodule), while DeepSeek-v3's `annotate_deepseekv3` uses the shorter `from torchtitan.models.common.moe import MoE`. Should use the shorter re-export path for consistency.

2. **EP degree mismatch between tests**: The integration test uses `expert_parallel_degree=4` while the numerics test uses `expert_parallel_degree=2`. Is this intentional? If so, worth a comment explaining why they differ. If not, they should match.

3. **Gratuitous Llama3 docstring change**: The diff includes a docstring-only change to `graph_trainer/llama3/parallelize.py`. If the intent is to update the docstring to be more accurate, consider doing it in a separate commit or PR to keep this PR focused on Qwen3.

### Nits

4. The conditional EP annotation guard in `annotate_qwen3` (`if any(layer.moe is not None ...)`) is a nice touch for supporting both dense and MoE — DeepSeek-v3 doesn't need this because all its layers are MoE. No change needed, just noting the design is sound.

5. The `remove_duplicate=False` changes to `graph_utils.py` and `trainer.py` are backward-compatible for models without weight tying (Llama3, DeepSeek-v3) since they have no duplicate parameters. Verified this doesn't change behavior for existing models.

torchtitan/experiments/graph_trainer/qwen3/model.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self, config: Config):
+        super().__init__(config)
+
+    def init_states(
+        self,
+        *,
+        buffer_device: torch.device | None = None,
+    ) -> None:
+        with disable_active_parametrization():
```
SherlockNoMad, line 29: I am curious what is this for?

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -464,9 +464,45 @@ def _build_deepseek_v3_tests() -> list[OverrideDefinitions]:
...
+                ],
+            ],
+            "aot_fx_trace qwen3 MoE FSDP+TP+EP",
+            "aot_fx_trace_qwen3_moe_fsdp_tp_ep",
+            ngpu=8,
+        ),
+    ]
+
+
 def build_graph_trainer_test_list() -> list[OverrideDefinitions]:
```
SherlockNoMad, line 507: who uses this? I suppose this is for local developer who is on H100?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -35,7 +35,11 @@ def fwd_bwd_step(
     ):
         pred = model(inputs, **extra_inputs, **extra_kwargs)
         loss = loss_fn(pred, labels) / global_valid_tokens
-        params = [p for p in model.parameters() if p.requires_grad]
+        params = [
+            p
+            for _, p in model.named_parameters(remove_duplicate=False)
```
SherlockNoMad, line 43: comment why it need "remove_duplicate=False”

SherlockNoMad (approved):

SherlockNoMad:
also please include Qwen in test_bitwise_deterministic.py

torchtitan/experiments/graph_trainer/qwen3/model.py
aditvenk: I think this avoids SimpleFSDP's parameterization -- so no redistributes when you access parameter for initializing

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -475,8 +511,8 @@ def build_graph_trainer_default_test_list() -> list[OverrideDefinitions]:
 
 
 def build_graph_trainer_h100_test_list() -> list[OverrideDefinitions]:
-    """DeepSeek-v3 tests only (for H100 machines)."""
-    return _build_deepseek_v3_tests()
+    """DeepSeek-v3 + Qwen3 tests (for H100 machines)."""
```
aditvenk, line 518: Are we specializing for H100 because MoE fails on older due to the sym int issue in looped MoE?

torchtitan/experiments/graph_trainer/qwen3/model.py
yiming0416: > I think this avoids SimpleFSDP's parameterization -- so no redistributes when you access parameter for initializing

Yes exactly.

torchtitan/experiments/graph_trainer/tests/integration_tests.py
yiming0416: Previously when all integration tests are on the same CI workflow. We directly use `build_graph_trainer_test_list `

Now we have `build_graph_trainer_default_test_list` for A10 CI machines and `build_graph_trainer_h100_test_list` for H100 CI machines.

So `build_graph_trainer_test_list ` is not being used in CI. Local developer on H100 can run it. Also I think it's not harmful to keep it.

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: Added:         
```
# remove_duplicate=False to preserve duplicate parameter entries
# from weight tying (e.g. shared embedding/output weights).
```

torchtitan/experiments/graph_trainer/tests/integration_tests.py
yiming0416: Yes. Qwen3 has both dense and moe variants.

Actually instead of hardcode model names, we should have two test classes for dense and moe models, run dense models on default machines, run moe models on h100 machines.

yiming0416:
Qwen3 numerics test added in `test_numerics.py` in this PR. I will add it to `test_bitwise_deterministic.py` in a follow-up PR.

Issue #2636, opened 2026-03-20
Default validator skips a validation batch each round


albertoveneri:
### Bug description

During the default validation loop, the stop condition is checked only after a new validation batch is loaded. As a result, the validation set is reduced by one batch in each validation round.

A simple fix is to move the conditional `break` to the end of the loop, as shown in the following patch:

```bash
diff --git a/torchtitan/components/validate.py b/torchtitan/components/validate.py
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -237,9 +237,6 @@ class Validator(BaseValidator):
         num_steps = 0
 
         for input_dict, labels in self.validation_dataloader:
-            # pyrefly: ignore [missing-attribute, unsupported-operation]
-            if self.config.steps != -1 and num_steps >= self.config.steps:
-                break
 
             self.metrics_processor.ntokens_since_last_log += labels.numel()
             for k, v in input_dict.items():
@@ -308,6 +305,10 @@ class Validator(BaseValidator):
             accumulated_losses.append(loss_sum.detach() / global_valid_tokens)
             num_steps += 1
 
+            # pyrefly: ignore [missing-attribute, unsupported-operation]
+            if self.config.steps != -1 and num_steps >= self.config.steps:
+                break
+
         # Compute average loss
         loss = torch.sum(torch.stack(accumulated_losses))
         loss /= num_steps
```

### Versions

The bug can be reproduced on the latest version of torchtitan by launching a debug run with a command similar to the following:

```bash
NGPU=1 MODULE=llama3 CONFIG=llama3_debugmodel run_train.sh --dataloader.dataset-path=tests/assets/c4_test --training.steps=100 --training.seq-len=16384 --validator.enable --validator.dataloader.dataset-path=tests/assets/c4_test --validator.steps=1 --validator.freq=1 --validator.dataloader.dataset=c4_test
```

This is an extreme case in which each validation round uses a single, large validation step. In this run, the warning message `WARNING - Dataset c4_test is being re-looped` appears only after 14 validation steps, even though fewer than 1,017 sequences have been processed out of the 2,000 sequences in [c4_test](https://github.com/pytorch/torchtitan/tree/main/tests/assets/c4_test).

To count the actual sequences processed, I applied the following patch:

```bash
diff --git a/torchtitan/components/validate.py b/torchtitan/components/validate.py
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
[...]

PR #2637, opened 2026-03-20
Fix validator skipping a batch each round by using itertools.islice

fegin:
* __->__ #2637

Fix https://github.com/pytorch/torchtitan/issues/2636 but using islice instead of moving the check to the end of the loop.

**Safe to merge**

### DIFF
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import itertools
 from collections.abc import Callable
 from contextlib import AbstractContextManager
 from dataclasses import dataclass, field, replace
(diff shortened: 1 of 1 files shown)


fegin:
All ROCm CI are failing. Not related to this PR.

torchtitan/components/validate.py
```diff
@@ -232,14 +233,16 @@ def validate(
...
 
+        config = cast(Validator.Config, self.config)
         accumulated_losses = []
         device_type = utils.device_type
         num_steps = 0
 
-        for input_dict, labels in self.validation_dataloader:
-            # pyrefly: ignore [missing-attribute, unsupported-operation]
-            if self.config.steps != -1 and num_steps >= self.config.steps:
-                break
```
tianyu-l, line 242: I didn't get what it means by
> As a result, the validation set is reduced by one batch in each validation round.

In my mind, the proper behavior is that
**At each validation step, we build an entirely new validation_dataloader, because you would hope each validation step to have the same set of validation data.**

Is the bug that we are not loading validation from the beginning, but loading from where we left last validation step?

cc @albertoveneri

tianyu-l (changes requested):

torchtitan/components/validate.py
albertoveneri: Thank you, @tianyu-l, for pointing this out. I thought this was a different issue and was preparing a separate issue/PR for it.

If I'm not mistaken, in other frameworks such as NeMo (https://github.com/NVIDIA-NeMo/NeMo) the validation dataset is not reset after each validation round, so the model sees different batches each time. For brevity, let's call it "iterative validation."

This PR correctly fixes the off-by-one issue in the iterative validation: https://github.com/pytorch/torchtitan/issues/2636.

However, I strongly prefer rebuilding the validation dataset from scratch at each validation round. That's exactly what I do in my experiments: I simply rebuild the validation dataloader each time `validate` is called. The overhead is usually negligible.

In the new issue, I was going to suggest adding a parameter to the config to decide whether the user wants to use the "iterative validation" or not.

Thank you again.

tianyu-l: let's just (re)build the dataset at each validation step -- no need to add a config option

fegin: If we are going to rebuild the dataset for each validation step, we can close this PR.

PR #2638, opened 2026-03-20
[RL] adopt local map attention for vLLM attention 

acisseJZhong:
**Summary**                                                       
   
  - Adopt LocalMapAttention as the base class for VLLMAttention,
   replacing manual DTensor.to_local() / DTensor.from_local() 
  with local_map for DTensor-to-local conversion. 
  - Add __call__ override that captures seq_len = q.size(2) from
   the DTensor before local_map's to_local() and passes it to   
  forward() via kwargs. This preserves the canonical symbolic   
  shape (s72) that GQAttention uses in its downstream view(bs, 
  seqlen, -1). Capturing from the DTensor's global shape ensures 
  the correct symbolic size is used under torch.compile. 
  See https://github.com/pytorch/pytorch/issues/175690
  - Remove replace_with_vllm_compatible_flash_attention() and 
  its usage in the trainer — the trainer no longer patches its  
  attention module to match vLLM's kernel.
  - Fix the test's vLLM engine creation to use                  
  GeneratorCompileConfig instead of hardcoded CompilationConfig,
   aligning the test with the generator's actual
  compile/CUDA-graph settings. 

**Test attention numerics under both true eager mode and compile mode:** 
```
NCCL_NVLS_ENABLE=0 torchrun --nproc_per_node=2     torchtitan/experiments/rl/tests/test_attn_numerics.py
```

> ============================================================
LOGPROB COMPARISON RESULTS
  Bitwise identical : False
  Tokens checked    : 30
  Tokens different  : 30
  Max delta         : 1.041739e-01
  Avg delta         : 1.816580e-02
  Diff mean         : 2.139650e-03
  Diff max          : 1.041739e-01

### DIFF
--- a/torchtitan/experiments/rl/tests/test_attn_numerics.py
+++ b/torchtitan/experiments/rl/tests/test_attn_numerics.py
@@ -35,7 +35,6 @@ from torch.distributed.checkpoint.state_dict import (
 )
 
 from vllm import EngineArgs, LLMEngine, SamplingParams
-from vllm.config.compilation import CUDAGraphMode
 from vllm.model_executor.layers.batch_invariant import init_batch_invariance
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
@@ -248,9 +247,10 @@ def _test_config() -> RLTrainer.Config:
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            compile=GeneratorCompileConfig(
-                backend="eager", cudagraph_mode=CUDAGraphMode.FULL_AND_PIECEWISE
-            ),
+            # compile=GeneratorCompileConfig(
+            #     backend="eager", cudagraph_mode=CUDAGraphMode.FULL_AND_PIECEWISE
+            # ),
+            compile=GeneratorCompileConfig(backend="none", cudagraph_mode="none"),
             num_samples_per_prompt=1,
             sampling=SamplingConfig(
                 temperature=0.0,
--- a/torchtitan/experiments/rl/config_registry.py
+++ b/torchtitan/experiments/rl/config_registry.py
@@ -13,7 +13,7 @@ Each function returns a complete ``RLTrainer.Config`` and is discoverable by
 
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
-from torchtitan.config.configs import CompileConfig, ParallelismConfig, TrainingConfig
+from torchtitan.config.configs import ParallelismConfig, TrainingConfig
 from torchtitan.experiments.rl.actors.generator import (
     GeneratorCompileConfig,
     SamplingConfig,
@@ -41,13 +41,13 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            compile=CompileConfig(enable=True, backend="aot_eager"),
+            # compile=CompileConfig(enable=True, backend="aot_eager"),
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
             compile=GeneratorCompileConfig(
-                backend="eager",
-                cudagraph_mode="piecewise",
+                backend="none",
+                cudagraph_mode="none",
             ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=4,
@@ -81,13 +81,13 @@ def rl_grpo_qwen3_1_7b() -> RLTrainer.Config:
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            compile=CompileConfig(enable=True, backend="aot_eager"),
+            # compile=CompileConfig(enable=True, backend="aot_eager"),
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
             compile=GeneratorCompileConfig(
-                backend="eager",
-                cudagraph_mode="piecewise",
+                backend="none",
+                cudagraph_mode="none",
             ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=4,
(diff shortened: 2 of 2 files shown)


torchtitan/experiments/rl/models/attention.py
```diff
@@ -275,15 +268,10 @@ def forward(
         # which happens with cudagraph capture for dummy input
         output_flat = output_flat.narrow(0, 0, batch_size * seq_len)
 
-        # Reshape back to titan: (batch, num_heads_local, seq_len, head_dim)
+        # Reshape back to titan: (batch, seq_len, num_heads_local, head_dim)
```
Lucaskabela: Note the transpose (1, 2) - comment is right since we swap num_heads and seq_len on L273

acisseJZhong: yup, I thought the shape annotation is only for the next line. Reverted it back!

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -206,11 +206,8 @@ def _build_model(
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
 
-        # Replace attention with vLLM compatible attention for RL training.
-        # NOTE: Long-term this will be replaced by pytorch attention
-        # supporting paged attention / kv cache.
         if batch_invariant_mode:
-            replace_with_vllm_compatible_flash_attention(
+            replace_with_vllm_attention(
```
wwwjn: I don't think we need this change to make trainer and generator bit-wise identity. To make bit-wise identity, we need the forward path run the same kernels on trainer and generator. This function is replace to vllm.Attention() for kv cache capability.  For trainer, set config to Varlen attention should be enough

Also this might not be trainable as vllm.Attention should not have backward? 

Let's remove this change from this PR.

torchtitan/experiments/rl/tests/test_attn_numerics.py
```diff
@@ -92,13 +95,13 @@ def create_vllm_engine(config):
         tensor_parallel_size=gen_config.parallelism.tensor_parallel_degree,
         distributed_executor_backend="external_launcher",
         gpu_memory_utilization=gen_config.gpu_memory_limit,
+        enforce_eager=gen_config.compile.is_eager,
```
wwwjn, line 98: Why adding this? Do you need enforce_eager = True when compile and cudagraph are disabled?

acisseJZhong: it should use the config from `_test_config` below, otherwise there's two compile config floating around

torchtitan/experiments/rl/actors/trainer.py
acisseJZhong: oh I though vllm.Attention which uses `PyTorchFlashAttentionImpl` has the backward? 

also tried removing this, doesn't seem to change numerics 
```
LOGPROB COMPARISON RESULTS
  Bitwise identical : False
  Tokens checked    : 30
  Tokens different  : 30
  Max delta         : 1.041739e-01
  Avg delta         : 1.816580e-02
  Diff mean         : 2.139650e-03
  Diff max          : 1.041739e-01
```

wwwjn: > oh I though vllm.Attention which uses `PyTorchFlashAttentionImpl` has the backward?

I see, you are right. But in trainer, we don't need the kv cache capability so we can directly using Varlen attention to achieve bit-wise identity

torchtitan/experiments/rl/models/attention.py
```diff
@@ -242,21 +255,7 @@ def forward(
         Returns:
             ``(batch, num_heads, seq_len, head_dim)``
         """
-        # Capture the original symbolic seq_len from the input BEFORE
```
tianyu-l, line 242: Actually I don't know why we were using global `seq_len` here.
- In TP, qkv are sharded on num_head dimension so seq_len should be the same on DTensor / local tensor
- In all-gather CP, qkv are sharded on seq_len dimension before entering CP, but replicate on kv after the hooks (it may not work with varlen attention yet).

In either case
`q = q.transpose(1, 2).reshape(batch_size * seq_len, -1, head_dim)` should work with local qkv's `seq_len`?

acisseJZhong: It's because L270 

>         # vLLM's flash attention backend may pad the token count (e.g.
>         # round up to an even number), which introduces a new symbolic
>         # shape under torch.compile.  Narrow to trim this padding
>         # NOTE: this error only happens when batch_size and seq_len are 1
>         # which happens with cudagraph capture for dummy input

during cudagraph capture with TP=2, the seqlen should be 1, but it's padded to 2, and we need to capture the original symbolic seq_len from the input BEFORE to_local. After using local map, this entire forward is wrapped in local region, therefore I moved this chunk of code before calling forward.

acisseJZhong:
running https://github.com/pytorch/torchtitan/pull/2300/changes#diff-61d04a6e7103debe722a05fcf985c39b2471c332de71b93de4e23e878b9a5d5a the test_numerics.py, all three tests passed. The command I am using is 
```
NCCL_NVLS_ENABLE=0 MASTER_ADDR=localhost MASTER_PORT=29500 MODEL_CHECKPOINT_PATH=Qwen/Qwen3-0.6B pytest         torchtitan/experiments/rl/tests/test_numerics.py -v -s
```
cc @wwwjn 

It means the whole attention module (not only `self_attn`) keep align with vllm's numerics,

torchtitan/experiments/rl/models/attention.py
acisseJZhong: this needs follow up with https://github.com/pytorch/pytorch/issues/175690; @Lucaskabela mentioned:

> I think this is a symbolic shape propagation error - w.r.t to_local(), the symbol we are using has to be divisible by the TP size, so I think it is something like ( 2*(s77 + 1) // 2) or something odd like that - this actually results in some bug I believe in our code that is generated

The approach in this PR is just workaround like before. Can delete once the fix is landed.

torchtitan/experiments/rl/actors/trainer.py
Lucaskabela, line 31: I think this is removed on main if you rebase :)

torchtitan/experiments/rl/models/attention.py
```diff
@@ -239,21 +241,12 @@ def forward(
...
-        # to_local() so that the symbol is the same one GQAttention uses
-        # in its .view(bs, seqlen, -1) call.
+        # This seq_len captured here is wrong probably due to some symbolic shape propagation error wrt to to_local.
+        # Therefore it is breaking compile. We need to fix this in pytorch.
+        # See more details in https://github.com/pytorch/pytorch/issues/175690
+        # TODO(@Lucaskabela): remove this once the issue is fixed in pytorch
         batch_size, _, seq_len, head_dim = q.shape
```
acisseJZhong, line 248: @Lucaskabela I updated this to still capture seq_len in local region. This captured seq_len is wrong and will break compile behavior, but it is the right thing to do to get seq_len in local region instead of global. 

Please help take a look into the fix for compile. Thanks!!

tianyu-l (approved):

PR #2639, opened 2026-03-20
[RL] Extend RLTrainer.setup() to accept an optional host_mesh parameter

daniellepintz:
Extend RLTrainer.setup() to accept an optional host_mesh parameter for multi-node mode. When provided, dedicates whole nodes to trainer/generator roles using HostMesh.slice() and sets CUDA_VISIBLE_DEVICES via Provisioner to isolate GPU allocations.

### DIFF
--- a/torchtitan/experiments/rl/simple_grpo_sum_digits.py
+++ b/torchtitan/experiments/rl/simple_grpo_sum_digits.py
@@ -147,6 +147,31 @@ class RLTrainer(Configurable):
         config.model_spec.parallelize_fn = parallelize_qwen3
 
         self.task = SumDigitsTask(seed=42)
+        self._proc_meshes = []
+
+    async def cleanup(self):
+        """Stop all proc meshes to release GPU memory."""
+        for mesh in self._proc_meshes:
+            try:
+                await mesh.stop()
+            except Exception:
+                pass
+        self._proc_meshes = []
+
+    def _get_rank_0_value(self, result, has_gpus: bool = True):
+        """Extract rank 0 result, handling both single and multi-node meshes.
+
+        Monarch actor endpoints return results from all ranks in the mesh.
+        This picks out rank 0's result by indexing into the host and GPU
+        dimensions as needed (multi-node meshes have an extra host dimension).
+        This should be used in cases where all ranks return the same result.
+        """
+        kwargs = {}
+        if self._multi_node:
+            kwargs["hosts"] = 0
+        if has_gpus:
+            kwargs["gpus"] = 0
+        return result.item(**kwargs)
 
     @staticmethod
     def _compute_world_size(p: "ParallelismConfig") -> int:
@@ -160,12 +185,31 @@ class RLTrainer(Configurable):
             * p.context_parallel_degree
         )
 
-    async def setup(self):
+    async def setup(
+        self,
+        *,
+        host_mesh=None,
+        trainer_nodes: int | None = None,
+        generator_nodes: int | None = None,
+        gpus_per_node: int | None = None,
+    ):
         """Spawn Monarch actors on separate meshes and initialize weights.
 
         Creates separate GPU meshes for trainer and generator, a CPU mesh for
         the grader, and synchronizes initial weights from trainer to generator.
         Must be called before :meth:`train`.
+
+        Args:
+            host_mesh: Optional multi-node HostMesh. When provided,
+                whole nodes are dedicated to trainer vs generator
+                roles instead of partitioning GPUs on a single host.
+            trainer_nodes: Number of nodes for the trainer (required when
+                host_mesh is provided).
+            generator_nodes: Number of nodes for the generator (required when
+                host_mesh is provided).
+            gpus_per_node: GPUs per node, assumed to be the same across all
+                nodes (no heterogeneous node configurations). Required when
+                host_mesh is provided.
         """
         config = self.config
 
@@ -182,19 +226,74 @@ class RLTrainer(Configurable):
 
         self.system_prompt = self.task.get_system_prompt()
 
-        # Allocate non-overlapping GPU ranges for each mesh
-        provisioner = Provisioner(total_gpus=total_gpus)
+        self._multi_node = host_mesh is not None
 
-        # Spawn separate meshes for trainer, generator, and grader
-        trainer_mesh = this_host().spawn_procs(
-            per_host={"gpus": self.trainer_world_size},
-            bootstrap=provisioner.allocate(self.trainer_world_size),
-        )
-        generator_mesh = this_host().spawn_procs(
-            per_host={"gpus": self.generator_world_size},
-            bootstrap=provisioner.allocate(self.generator_world_size),
-        )
-        grader_mesh = this_host().spawn_procs()
+        if host_mesh is not None:
+            # Multi-node mode: dedicate whole nodes to trainer vs generator
+            if (
+                trainer_nodes is None
+                or generator_nodes is None
+                or gpus_per_node is None
+            ):
+                raise ValueError(
+                    "trainer_nodes, generator_nodes, and gpus_per_node are "
+                    "required when host_mesh is provided"
+                )
+            # Validate that world sizes are evenly divisible by node counts
+            assert self.trainer_world_size % trainer_nodes == 0, (
+                f"trainer_world_size ({self.trainer_world_size}) must be "
+                f"evenly divisible by trainer_nodes ({trainer_nodes})"
+            )
+            assert self.generator_world_size % generator_nodes == 0, (
+                f"generator_world_size ({self.generator_world_size}) must be "
+                f"evenly divisible by generator_nodes ({generator_nodes})"
+            )
+
+            # Compute GPUs per node for each role based on the config's
+            # world size and number of nodes allocated to that role
+            trainer_gpus_per_node = self.trainer_world_size // trainer_nodes
+            generator_gpus_per_node = self.generator_world_size // generator_nodes
+
+            trainer_tp = config.trainer.parallelism.tensor_parallel_degree
+            generator_tp = config.generator.parallelism.tensor_parallel_degree
+
+            trainer_host_mesh = host_mesh.slice(hosts=slice(0, trainer_nodes))
+            generator_host_mesh = host_mesh.slice(
+                hosts=slice(trainer_nodes, trainer_nodes + generator_nodes)
+            )
+
+            # Use Provisioner to set CUDA_VISIBLE_DEVICES so each role
+            # only sees its own GPUs and doesn't conflict with other
+            # processes on the node
+            trainer_provisioner = Provisioner(total_gpus=gpus_per_node)
+            generator_provisioner = Provisioner(total_gpus=gpus_per_node)
+
+            trainer_mesh = trainer_host_mesh.spawn_procs(
+                per_host={"gpus": trainer_gpus_per_node},
+                bootstrap=trainer_provisioner.allocate(trainer_gpus_per_node),
+            )
+            generator_mesh = generator_host_mesh.spawn_procs(
+                per_host={"gpus": generator_gpus_per_node},
+                bootstrap=generator_provisioner.allocate(generator_gpus_per_node),
+            )
+            # Grader runs on CPU on the first trainer node
+            grader_mesh = trainer_host_mesh.spawn_procs()
+        else:
+            # Single-node mode: partition GPUs on this_host() via
+            # CUDA_VISIBLE_DEVICES
+            provisioner = Provisioner(total_gpus=total_gpus)
+            trainer_mesh = this_host().spawn_procs(
+                per_host={"gpus": self.trainer_world_size},
+                bootstrap=provisioner.allocate(self.trainer_world_size),
+            )
+            generator_mesh = this_host().spawn_procs(
+                per_host={"gpus": self.generator_world_size},
+                bootstrap=provisioner.allocate(self.generator_world_size),
+            )
+            grader_mesh = this_host().spawn_procs()
+
+        # Store proc meshes for cleanup
+        self._proc_meshes = [trainer_mesh, generator_mesh, grader_mesh]
 
         await setup_torch_elastic_env_async(trainer_mesh)
         await setup_torch_elastic_env_async(generator_mesh)
@@ -258,8 +357,8 @@ class RLTrainer(Configurable):
             eval_questions.append(question)
 
         # Generate on eval prompts
-        episodes = (
-            self.generator.generate.call(eval_prompts, eval_answers).get().item(gpus=0)
+        episodes = self._get_rank_0_value(
+            self.generator.generate.call(eval_prompts, eval_answers).get()
         )
 
         # Score: check first episode per prompt (episodes are ordered by prompt)
@@ -321,14 +420,14 @@ class RLTrainer(Configurable):
             # Fully sync RL loop (GRPO)
             # 1. Generator produces flat list of Episodes with group_id
             # TODO: Create a queue to use all episodes from all GPUs
-            episodes = (
-                self.generator.generate.call(train_prompts, train_answers)
-                .get()
-                .item(gpus=0)
+            episodes = self._get_rank_0_value(
+                self.generator.generate.call(train_prompts, train_answers).get()
             )
 
             # 2. Grader computes rewards per episode
-            episodes = self.grader.score.call(episodes).get().item()
+            episodes = self._get_rank_0_value(
+                self.grader.score.call(episodes).get(), has_gpus=False
+            )
 
             # 3. Controller computes GRPO advantages (normalize within group)
             groups: dict[str, list[int]] = defaultdict(list)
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -160,12 +179,29 @@ def _compute_world_size(p: "ParallelismConfig") -> int:
...
+
+        Args:
+            host_mesh: Optional multi-node HostMesh. When provided,
+                whole nodes are dedicated to trainer vs generator
+                roles instead of partitioning GPUs on a single host.
+            trainer_nodes: Number of nodes for the trainer (required when
+                host_mesh is provided).
+            generator_nodes: Number of nodes for the generator (required when
+                host_mesh is provided).
+            gpus_per_node: GPUs per node (required when host_mesh is provided).
```
tianyu-l: this is assuming no heterogeneity is allowed? please explicit comment this

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -182,19 +218,88 @@ async def setup(self):
...
+
+            # Ensure TP isn't split across nodes (requires fast NVLink)
+            trainer_tp = config.trainer.parallelism.tensor_parallel_degree
+            generator_tp = config.generator.parallelism.tensor_parallel_degree
+            if trainer_gpus_per_node < trainer_tp:
+                raise ValueError(
+                    f"trainer tensor_parallel_degree ({trainer_tp}) exceeds "
+                    f"trainer_gpus_per_node ({trainer_gpus_per_node}). "
+                    f"Use fewer trainer_nodes to keep TP on a single node."
+                )
```
tianyu-l: why do we care if it's user decision?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -182,19 +218,88 @@ async def setup(self):
...
+            trainer_host_mesh = host_mesh.slice(hosts=slice(0, trainer_nodes))
+            generator_host_mesh = host_mesh.slice(
+                hosts=slice(trainer_nodes, trainer_nodes + generator_nodes)
+            )
+
+            # Use Provisioner to set CUDA_VISIBLE_DEVICES so each role
+            # only sees its own GPUs and doesn't conflict with other
+            # processes on the node
+            trainer_provisioner = Provisioner(total_gpus=gpus_per_node)
+            generator_provisioner = Provisioner(total_gpus=gpus_per_node)
```
tianyu-l, line 269: what's the rationale behind this vs. one Provisioner allocate resource for every participant?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -160,12 +179,29 @@ def _compute_world_size(p: "ParallelismConfig") -> int:
...
         )
 
-    async def setup(self):
+    async def setup(
+        self,
+        *,
+        host_mesh=None,
+        trainer_nodes: int | None = None,
+        generator_nodes: int | None = None,
+        gpus_per_node: int | None = None,
```
tianyu-l, line 194: where is this used

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -160,12 +179,29 @@ def _compute_world_size(p: "ParallelismConfig") -> int:
             * p.context_parallel_degree
         )
 
-    async def setup(self):
+    async def setup(
+        self,
+        *,
+        host_mesh=None,
```
tianyu-l, line 191: didn't see this used / tested anywhere

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -160,12 +179,29 @@ def _compute_world_size(p: "ParallelismConfig") -> int:
...
+    ):
         """Spawn Monarch actors on separate meshes and initialize weights.
 
         Creates separate GPU meshes for trainer and generator, a CPU mesh for
         the grader, and synchronizes initial weights from trainer to generator.
         Must be called before :meth:`train`.
+
+        Args:
+            host_mesh: Optional multi-node HostMesh. When provided,
+                whole nodes are dedicated to trainer vs generator
```
acisseJZhong, line 204: does this support multi-hosts but still partition gpus across trainer/generator? For example 2 hosts, with 4 generator gpus, 12 ttrainer gpus.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -147,6 +147,25 @@ def __init__(self, config: Config):
...
+                await mesh.stop()
+            except Exception:
+                pass
+        self._proc_meshes = []
+
+    def _extract_result(self, result, has_gpus: bool = True):
+        """Extract rank 0 result, handling both single and multi-node meshes."""
+        kwargs = {}
+        if self._multi_node:
+            kwargs["hosts"] = 0
```
acisseJZhong, line 171: qq: what will this do?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -321,14 +426,14 @@ async def train(self):
...
-                .get()
-                .item(gpus=0)
+            episodes = self._extract_result(
+                self.generator.generate.call(train_prompts, train_answers).get()
             )
 
             # 2. Grader computes rewards per episode
-            episodes = self.grader.score.call(episodes).get().item()
+            episodes = self._extract_result(
+                self.grader.score.call(episodes).get(), has_gpus=False
```
acisseJZhong, line 429: qq: does this mean reward calculation is purely on cpu?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: it currently does not - currently it is one actor per host. We can consider changing this later

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: yes exactly

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: Monarch actor endpoints return results from all ranks in the mesh. This picks out rank 0's result by indexing into the host and GPU dimensions as needed (multi-node meshes have an extra host dimension). This should be used in cases where all ranks return the same result.

(Added to docstring)

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: This is used in the internal-only code in the linked diff. We could probably add a CI test that uses this, I can follow up with that if you think it makes sense? will require multiple nodes

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: This is used in the internal-only code in the linked diff

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: yes, added comment

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: I wanted to capture cases where TP is *unnecessarily* split across nodes, because of the calculation above:
```
trainer_gpus_per_node = self.trainer_world_size // trainer_nodes
```
Updated the check here to capture that, thanks

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: IMO it's clearer to read if the Provisioner just handles GPUs for one host; if we wanted to use the same Provisioner and do `allocate` calls for different hosts we would need to have some sort of `reset` in between the allocate calls. cc @joecummings in case you have any thoughts here

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -147,6 +147,31 @@ def __init__(self, config: Config):
...
+    async def cleanup(self):
+        """Stop all proc meshes to release GPU memory."""
+        for mesh in self._proc_meshes:
+            try:
+                await mesh.stop()
+            except Exception:
+                pass
+        self._proc_meshes = []
+
+    def _extract_result(self, result, has_gpus: bool = True):
```
tianyu-l: let's be more explicit here
```suggestion
    def _get_rank_0_value(self, result, has_gpus: bool = True):
```

tianyu-l: what's the type of `result`? I can only see that you can call `.item()` on it, is it always a tensor?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
tianyu-l: It's not clear to me how it can work when either trainer / generator spans multiple node.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -182,19 +226,90 @@ async def setup(self):
...
+            # Compute GPUs per node for each role based on the config's
+            # world size and number of nodes allocated to that role
+            trainer_gpus_per_node = self.trainer_world_size // trainer_nodes
+            generator_gpus_per_node = self.generator_world_size // generator_nodes
+
+            if trainer_gpus_per_node * trainer_nodes != self.trainer_world_size:
+                raise ValueError(
+                    f"trainer_world_size ({self.trainer_world_size}) must be "
+                    f"evenly divisible by trainer_nodes ({trainer_nodes})"
+                )
```
tianyu-l: I think these checks should happen in it's own component, e.g. provisioner.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -182,19 +226,90 @@ async def setup(self):
...
+            # Ensure TP isn't unnecessarily split across nodes
+            trainer_tp = config.trainer.parallelism.tensor_parallel_degree
+            generator_tp = config.generator.parallelism.tensor_parallel_degree
+            if trainer_gpus_per_node < trainer_tp and gpus_per_node >= trainer_tp:
+                raise ValueError(
+                    f"trainer tensor_parallel_degree ({trainer_tp}) exceeds "
+                    f"trainer_gpus_per_node ({trainer_gpus_per_node}) but fits "
+                    f"on a single node ({gpus_per_node} GPUs). "
+                    f"Use fewer trainer_nodes to keep TP on a single node."
+                )
```
tianyu-l: > I wanted to capture cases where TP is unnecessarily split across nodes, because of the calculation above:

I don't think we need to do this for users, we are not building auto parallel, so users should be responsible for their own parallel configs.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: `result` is a Monarch [ValueMesh](https://github.com/meta-pytorch/monarch/blob/1000989be29e4184aee05abb4164a8d48a6c3fa0/python/monarch/_src/actor/actor_mesh.py#L738) — the return value of `.get()` on a Monarch actor call. It's a multi-dimensional container indexed by mesh dimensions (e.g., hosts, gpus), holding the return value from each rank. 

The [`.item(**kwargs)`](https://github.com/meta-pytorch/monarch/blob/1000989be29e4184aee05abb4164a8d48a6c3fa0/python/monarch/_src/actor/actor_mesh.py#L811) call extracts a single value by indexing into those dimensions.

The types in the container, and the return type of `.item()` will be whatever type the Monarch endpoint returns.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: Ok, sg will remove

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: If trainer/generator spans multiple nodes, the provisioner will allocate the appropriate GPUs on each node (setting CUDA_VISIBLE_DEVICES); the bootstrap function (allocate) is called _independently_ on each node when we call `spawn_procs`, so this works well.

but if we tried to use the same provisioner for the generator spawn_procs as well, the provisioner would already have the "state" from the trainer spawn_procs, e.g. `next_gpu` would already have a value that is not 0.

It is a bit confusing, pls lmk if you have more questions

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: I am not sure I agree with this, because then the provisioner would need to know about the trainer / generator world size; right now all the provisioner knows about is managing the number of GPUs on one node and knows which are in use and which are free

I guess if we expanded the scope of the Provisioner this could work - how do you envision it?

I do think these checks should be converted to asserts that happen before `trainer_gpus_per_node` is calculated

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: The rationale makes sense - the root of the problem as pointed out by @allenwang28 is that although Monarch does internally keep track of host->gpu offsets, it doesn't expose this API (yet - we should ask them to). 

However, I have a preference for keeping the training loop code cleaner. To that end, we could temporarily hold these offsets in the Provisioner itself rather.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -147,6 +147,31 @@ def __init__(self, config: Config):
...
+    async def cleanup(self):
+        """Stop all proc meshes to release GPU memory."""
+        for mesh in self._proc_meshes:
+            try:
+                await mesh.stop()
+            except Exception:
+                pass
+        self._proc_meshes = []
+
+    def _get_rank_0_value(self, result, has_gpus: bool = True):
```
joecummings, line 161: I'm not sure I understand why this is needed

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: > we could temporarily hold these offsets in the Provisioner itself rather.

could you explain what you mean exactly? Also, can this be done as a follow up?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -182,19 +226,74 @@ async def setup(self):
...
+                trainer_nodes is None
+                or generator_nodes is None
+                or gpus_per_node is None
+            ):
+                raise ValueError(
+                    "trainer_nodes, generator_nodes, and gpus_per_node are "
+                    "required when host_mesh is provided"
+                )
+            # Validate that world sizes are evenly divisible by node counts
+            assert self.trainer_world_size % trainer_nodes == 0, (
```
tianyu-l, line 243: Since we don't allow trainer and generator sharing a node, does it make sense to require each actor to occupy all the gpus on the same node? Like 
```
assert self.trainer_world_size == trainer_nodes * gpus_per_node
```

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: In many cases we need to get the result of a Monarch endpoint from the controller, and the `get()` call returns a Monarch multi-dimensional container, that we need to index into with `.item(gpus=<rank>)` to get the actual result we want. For example, getting the generated episodes:

https://github.com/pytorch/torchtitan/blob/18a46d427e2e6abf5c84073d349d44fd3c520d81/torchtitan/experiments/rl/simple_grpo_sum_digits.py#L324-L328

 We use rank 0 for clarity, since all ranks have the same result, and we are calling this from the controller.

As you can see we have this in existing training loop, in this PR I am just moving it into a separate method

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: Good question, I definitely see your point.

My only concern is then the configs for smaller models like `rl_grpo_qwen3_0_6b` wouldn't work out of the box on multi-node, you would have to manually change the parallelism settings and getting an AssertionError might be annoying.

I would lean towards not enforcing that for now, so we can use existing configs out of the box, but happy to keep discussing this after I land this PR; can always add later.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: Sure this can be a follow up

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: This seems to add extra code just to do the item() call, which is more clear IMO - why do we need it?

daniellepintz: yes, the thing is for multi node the "container" of results becomes more complex / multi dimensional, and we need this code to work for the multi node **and** the single code case

PR #2646, opened 2026-03-21
[graph_trainer] Add FakeTensorMode support for precompilation

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* __->__ #2646
* #2650
* #2644
* #2643
* #2642

Enable precompilation without GPU memory for model weights by using
FakeTensorMode. During init, to_empty() runs under FakeTensorMode so
model parameters become fake CUDA tensors, and init_weights is skipped.
The joint_graph_builder is called directly (not the full forward) since
NCCL cannot communicate fake data. Produces bitwise identical artifacts
to real-tensor precompilation.

@claude

PR #2645, opened 2026-03-20
[graph_trainer] Add Llama3 precompile config, tests, and docs

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #2646
* __->__ #2645
* #2644
* #2643
* #2642

Add graph_trainer_llama3_precompile config function (debugmodel) with
mode=aot, full_inductor_compilation pass, and inductor_decomposition
joint pass — required for serializable compilation.

Add unit tests for DiskStorageAdapter, PrecompiledArtifact pickle
roundtrip, GraphTrainerCompileConfig precompile field defaults, and
config fingerprint validation (deterministic hashing, model shape
sensitivity, parallelism sensitivity, compile config sensitivity,
pass order insensitivity).

Update README with pre-compile usage instructions and mark pre-compile
as supported in the composability table.

## Validation

Validated on Llama3 debugmodel with AOT + inductor passes (DP=2, TP=2,
PP=2) on 8x H100 GPUs using `--debug.deterministic --debug.seed=42`
for 10 steps.

Three runs were compared, all using the same compile passes
(`full_inductor_compilation` + `inductor_decomposition`):
1. **AOT baseline:** same passes, `precompile=false`
2. **Cold run:** `precompile=true` — AOT compiles, serializes the
   artifact to disk, then trains normally.
3. **Warm run:** `precompile=true` — loads the serialized artifact
   from disk (skips compilation), then trains.

**All three runs produce bitwise-identical loss curves** (max |diff| =
0.0 across all pairwise comparisons). This was verified by extracting
full-precision loss values from TensorBoard event files
(`loss_metrics/global_avg_loss` tag) at every step and comparing them
with exact floating-point equality — no tolerance or epsilon was used.

This confirms the serialization/deserialization roundtrip is exact and
the precompile flag has zero numerical impact.

Loss comparison

**Test script and full results:** https://gist.github.com/bobrenjc93/f10a064e64babae5b07c9e217f6eed83

PR #2644, opened 2026-03-20
[graph_trainer] Integrate precompile into AOT compilation path

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #2646
* #2650
* __->__ #2644
* #2643
* #2642

Add precompile and precompile_artifact_dir config fields to
GraphTrainerCompileConfig, and wire serializable compilation through
joint_graph_builder and the AOT compile dispatcher:

- joint_graph_builder: add serializable and on_compile params. When
  serializable=True, passes it to aot_compile_joint_with_descriptors.
  Calls on_compile callback after compilation with the compiled fn and
  tree specs.

- CompiledModule: add precompiled_fn param. When set, uses the
  precompiled function directly instead of calling joint_graph_builder.

- _apply_aot_compile: when precompile=True, checks for existing
  artifacts and loads them (skipping compilation). Otherwise compiles
  with serializable=True and saves the artifact via on_compile callback.

- _make_precompile_callback: builds the on_compile callback, validates
  that full_inductor_compilation is in the pass pipeline (required for
  serializable compilation).

Config fingerprint is threaded through the save and load paths:
fingerprint is computed in _make_precompile_callback (save path) and in
_apply_aot_compile (load path). On fingerprint mismatch, the stale
artifact is automatically discarded and recompilation proceeds — users
just pass --precompile and the right thing happens.

@claude

PR #2643, opened 2026-03-20
[graph_trainer] Add precompile artifact serialization and loading

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #2646
* #2650
* #2644
* __->__ #2643
* #2642

Add precompile_save() and precompile_load() functions that serialize and
deserialize compiled AOT graphs using BundledAOTAutogradSerializableCallable.
Artifacts include the serialized compiled function, parameter/buffer specs,
input/output tree specs, and metadata. Deserialization is deferred to first
call so Triton kernels load on the correct CUDA device.

Includes config fingerprint validation: `compute_config_fingerprint()` hashes
model param/buffer names+shapes+dtypes, parallelism dimensions (world_size,
dp_replicate, dp_shard, cp, tp, pp, ep, etp), and compile config (mode,
backend, passes, joint_passes) into a 16-char SHA-256 digest. The fingerprint
is stored in the artifact on save and validated on load — mismatches produce
a clear error, and legacy artifacts without fingerprints log a warning.

@claude

PR #2647, opened 2026-03-21
[graph_trainer] Address code review comments on precompile stack

bobrenjc93:
* #2648
* __->__ #2647
* #2646
* #2645
* #2644
* #2643
* #2642

----

- storage.py: Accept Path|str for base_dir, use atomic write-to-temp-
  then-rename pattern to avoid partial files on crash
- precompile.py: Validate params/buffers spec on load against current
  model (raise ValueError on mismatch), fix unnecessary f-string concat
- compile.py: Extract _get_precompile_storage_and_key helper to DRY up
  duplicated storage/rank/key construction, fix storage type annotation
  (object -> StorageAdapter), replace silent None lambda with defensive
  RuntimeError for unused joint_graph_builder, add warning when
  precompile is used with non-AOT mode, use direct attribute access
  instead of getattr for passes, fix f-string without interpolation
- configs.py: Expand precompile_artifact_dir docstring to note /tmp is
  ephemeral and users should use a persistent/shared path

### DIFF
--- a/torchtitan/experiments/graph_trainer/compile.py
+++ b/torchtitan/experiments/graph_trainer/compile.py
@@ -40,7 +40,20 @@ from torchtitan.experiments.graph_trainer.graph_utils import (
 from torchtitan.experiments.graph_trainer.jit_backend import (
     get_compile_backend_with_passes,
 )
+from torchtitan.experiments.graph_trainer.storage import (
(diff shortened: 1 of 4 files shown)


bobrenjc93:
@claude

PR #2649, opened 2026-03-22
[rl] Add varlen attention for trainer.

zhxchen17:
Summary:

We previously added varlen attention kernel for generator(vllm) by registering
a CUSTOM backend and using it in Generator Config. In this PR, we added a new
model config with attention_backend set to "varlen".

In the test_attn_numerics.py we are now testing against Qwen variant "0.6B_varlen"
instead of "0.6B" now, which means the generator and trainer should both use varlen
kernel now.

Test Plan:

torchrun --nproc_per_node=2 torchtitan/experiments/rl/tests/test_attn_numerics.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -200,6 +200,12 @@ class PolicyTrainer(Actor, Configurable):
         Returns:
             Initialized model with weights loaded from checkpoint.
         """
+
+        # TODO Also support flex attention backend later.
+        assert (
+            model_spec.model.layer.attention.attn_backend == "varlen"
+        ), "Only varlen attention backend is allowed."
+
         with torch.device("meta"):
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
@@ -285,7 +291,10 @@ class PolicyTrainer(Actor, Configurable):
         with torch.no_grad():
             for prompt_toks, gen_toks in zip(my_prompt_token_ids, my_token_ids):
                 token_lps = compute_token_log_probs(
-                    self.ref_model, prompt_toks, gen_toks, device
+                    self.ref_model,
+                    prompt_toks,
+                    gen_toks,
+                    device,
                 )
                 ref_token_log_probs.append(token_lps)
 
--- a/torchtitan/experiments/rl/actors/utils.py
+++ b/torchtitan/experiments/rl/actors/utils.py
@@ -7,29 +7,55 @@
 import torch
 import torch.nn.functional as F
 
+from torchtitan.models.common.attention import VarlenMetadata
+
+
+# TODO We should either unify all the mask creation for RL, or move them to a
+#      single file.
+def build_varlen_metadata(
+    input_sequences: list[tuple[torch.Tensor, int, int]], device: torch.device
+) -> VarlenMetadata:
+    """Build VarlenMetadata for all sequences in a batch."""
+    cu_seqs = torch.cumsum(
+        torch.tensor(
+            [0] + [token_ids.shape[0] for token_ids, _, _ in input_sequences],
+            dtype=torch.int32,
+            device=device,
+        ),
+        0,
+        dtype=torch.int32,
+    )
+    max_len = max(token_ids.shape[0] for token_ids, _, _ in input_sequences)
+    return VarlenMetadata(
+        cu_seq_q=cu_seqs, cu_seq_k=cu_seqs, max_q=max_len, max_k=max_len
+    )
+
 
 def compute_token_log_probs(
     model: torch.nn.Module,
-    prompt_token_ids: list[int],
-    gen_token_ids: list[int],
+    prompt_ids: list[int],
+    gen_ids: list[int],
     device: torch.device,
 ) -> torch.Tensor:
     """
     Compute per-token log probabilities for generated tokens.
+    TODO Only batch size 1 is supported for now.
 
     Args:
         model: The model to use for computing logits
-        prompt_token_ids: Token IDs for the prompt
-        gen_token_ids: Token IDs for the generated completion
+        prompt_ids: Prompt token IDs
+        gen_ids: Generated token IDs
         device: Device to run computation on
 
     Returns:
         Per-token log probabilities for the generated tokens
     """
-    full_sequence = prompt_token_ids + gen_token_ids
-    full_tensor = torch.tensor(
-        full_sequence, dtype=torch.long, device=device
-    ).unsqueeze(0)
+    token_ids = torch.tensor(prompt_ids + gen_ids, dtype=torch.long, device=device)
+    prompt_len = len(prompt_ids)
+    gen_len = len(gen_ids)
+    attention_masks = build_varlen_metadata([(token_ids, prompt_len, gen_len)], device)
+
+    full_tensor = token_ids.unsqueeze(0)
 
     # NOTE: We should move towards batching to improve efficiency here
     # See https://github.com/pytorch/torchtitan/issues/2674
@@ -38,7 +64,7 @@ def compute_token_log_probs(
     seq_len = full_tensor.shape[1]
     positions = torch.arange(seq_len, device=device).unsqueeze(0)
 
-    logits = model(full_tensor, attention_masks=None, positions=positions)
+    logits = model(full_tensor, attention_masks=attention_masks, positions=positions)
 
     # Convert to float32 for numerical stability
     logits_f32 = logits[:, :-1, :].to(torch.float32)
@@ -46,9 +72,8 @@ def compute_token_log_probs(
     target_tokens = full_tensor[:, 1:]
 
     # Extract log probs for generated tokens only
-    prompt_len = len(prompt_token_ids)
     gen_start_idx = prompt_len - 1
-    gen_end_idx = gen_start_idx + len(gen_token_ids)
+    gen_end_idx = gen_start_idx + gen_len
 
     gen_token_logprobs = log_probs[0, gen_start_idx:gen_end_idx, :]
     gen_token_ids_tensor = target_tokens[0, gen_start_idx:gen_end_idx]
@@ -97,7 +122,12 @@ def compute_policy_gradient_loss(
     batch_token_log_probs = []
 
     for prompt_toks, gen_toks in zip(prompt_token_ids, vllm_token_ids):
-        token_lps = compute_token_log_probs(model, prompt_toks, gen_toks, device)
+        token_lps = compute_token_log_probs(
+            model,
+            prompt_toks,
+            gen_toks,
+            device,
+        )
         batch_token_log_probs.append(token_lps)
 
     # Per-token log ratios and KL, averaged across tokens per sample
--- a/torchtitan/experiments/rl/config_registry.py
+++ b/torchtitan/experiments/rl/config_registry.py
@@ -26,8 +26,9 @@ from torchtitan.models.qwen3 import model_registry
 
 def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
     """GRPO training config for Qwen3-0.6B (6 GPUs: 4 gen + 2 train)."""
+    model_spec = model_registry("0.6B", attn_backend_override="varlen")
     return RLTrainer.Config(
-        model_spec=model_registry("0.6B"),
+        model_spec=model_spec,
         hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
         num_steps=10,
         batch_invariant_mode=True,
@@ -65,8 +66,9 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
 
 def rl_grpo_qwen3_1_7b() -> RLTrainer.Config:
     """GRPO training config for Qwen3-1.7B (6 GPUs: 4 gen + 2 train)."""
+    model_spec = model_registry("1.7B", attn_backend_override="varlen")
     return RLTrainer.Config(
-        model_spec=model_registry("1.7B"),
+        model_spec=model_spec,
         hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-1.7B",
         num_steps=10,
         batch_invariant_mode=True,
@@ -104,8 +106,9 @@ def rl_grpo_qwen3_1_7b() -> RLTrainer.Config:
 
 def rl_grpo_qwen3_debug() -> RLTrainer.Config:
     """Debug config for quick iteration -- small model, few steps (2 GPUs: 1 gen + 1 train)."""
+    model_spec = model_registry("debugmodel", attn_backend_override="varlen")
     return RLTrainer.Config(
-        model_spec=model_registry("debugmodel"),
+        model_spec=model_spec,
         num_steps=5,
         batch_invariant_mode=False,
         trainer=PolicyTrainer.Config(
--- a/torchtitan/experiments/rl/models/attention.py
+++ b/torchtitan/experiments/rl/models/attention.py
@@ -247,6 +247,7 @@ class VLLMAttention(LocalMapAttention):
         *,
         scale: float | None = None,
         enable_gqa: bool = False,
+        attention_masks: None = None,  # Unused but needed for GQA varlen inference.
     ) -> torch.Tensor:
         """Run vLLM paged attention on local (non-DTensor) tensors.
 
@@ -284,9 +285,9 @@ class VLLMAttention(LocalMapAttention):
         # which happens with cudagraph capture for dummy input
         output_flat = output_flat.narrow(0, 0, batch_size * seq_len)
 
-        # Reshape back to titan: (batch, num_heads_local, seq_len, head_dim)
-        output = output_flat.view(batch_size, seq_len, -1, head_dim)
-        output = output.transpose(1, 2)
+        # Reshape back to the format expected by GQAttention.forward()
+        # varlen path expects (bs*seqlen, n_heads, head_dim)
+        output = output_flat.view(batch_size * seq_len, -1, head_dim)
 
         return output
 
@@ -329,6 +330,11 @@ def replace_with_vllm_attention(model, tp_degree=1):
 
         # GQA
         head_dim = model_args.layer.attention.head_dim
+
+        # TODO Support flex attention backend later as well.
+        assert (
+            layer.attention.attn_backend == "varlen"
+        ), "Only varlen attention backend is allowed."
         vllm_attn = VLLMAttention(
             hidden_size=model_args.dim,
             num_heads=model_args.layer.attention.n_heads // tp_degree,
--- a/torchtitan/experiments/rl/tests/test_bitwise_identity.py
+++ b/torchtitan/experiments/rl/tests/test_bitwise_identity.py
@@ -79,7 +79,7 @@ def _log_attention_module(model, label):
 # ---------------------------------------------------------------------------
 
 
-def create_vllm_engine(config):
+def create_vllm_engine(config: RLTrainer.Config) -> LLMEngine:
     """Create a vLLM LLMEngine from the RL config."""
     gen_config = config.generator
     model_path = config.hf_assets_path
@@ -113,7 +113,9 @@ def create_vllm_engine(config):
     return engine
 
 
-def generate_with_vllm(engine, prompt, gen_config):
+def generate_with_vllm(
+    engine: LLMEngine, prompt: str, gen_config: VLLMGenerator.Config
+) -> tuple[list[int], list[int], list[float]]:
     """Generate tokens from *prompt*, return IDs + log-probs."""
     sampling = gen_config.sampling
     sampling_params = SamplingParams(
@@ -135,6 +137,7 @@ def generate_with_vllm(engine, prompt, gen_config):
     assert len(outputs) == 1, f"Expected 1 output, got {len(outputs)}"
     output = outputs[0]
 
+    logger.info('vLLM text output: "%s"', output.outputs[0].text)
     prompt_token_ids = list(output.prompt_token_ids)
     sample = output.outputs[0]
     generated_token_ids = list(sample.token_ids)
@@ -158,8 +161,6 @@ def build_trainer_model(config):
 
     Mirrors PolicyTrainer._build_model() but without the Monarch actor
     framework.
-
-    TODO(zhxchen17) Switch trainer model to use varlen backend with batch-invariant mode.
     """
     model_spec = config.model_spec
     model_config = model_spec.model
@@ -230,8 +231,9 @@ def build_trainer_model(config):
 
 def _test_config() -> RLTrainer.Config:
     """Test-specific config: greedy sampling, fewer tokens, single sample."""
+    model_spec = model_registry("0.6B", attn_backend_override="varlen")
     return RLTrainer.Config(
-        model_spec=model_registry("0.6B"),
+        model_spec=model_spec,
         hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
         batch_invariant_mode=True,
         trainer=PolicyTrainer.Config(
@@ -297,7 +299,10 @@ def main():
 
     with torch.no_grad():
         trainer_token_log_probs = compute_token_log_probs(
-            trainer_model, prompt_token_ids, generated_token_ids, device
+            trainer_model,
+            prompt_token_ids,
+            generated_token_ids,
+            device,
         )
 
     if dist.get_rank() == 0:
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -164,6 +164,9 @@ class VarlenAttentionWrapper(LocalMapAttention):
         attention_masks: VarlenMetadata,
         scale: float | None = None,
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+        assert isinstance(
+            attention_masks, VarlenMetadata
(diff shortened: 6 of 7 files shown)


torchtitan/models/common/attention.py
tianyu-l: shall we move this assert as well?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -7,6 +7,18 @@
 import torch
 import torch.nn.functional as F
 
+from torchtitan.models.common.attention import VarlenMetadata
+
+
+def _build_single_seq_varlen_metadata(
+    seq_len: int, device: torch.device
+) -> VarlenMetadata:
+    """Build VarlenMetadata for a single sequence (batch_size=1, one document)."""
```
wwwjn: This assuming the trainer's batch size always equals one, which is true for now but lack of flexibility, can we make it more generalize? 

And in trainer, we could reuse most components from pretraining. In pretraining we let the model create the varlen metadata https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder.py#L199 Can we reuse the `model.get_attention_masks()`?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -275,9 +278,14 @@ def forward(
...
 
-        # Reshape back to titan: (batch, num_heads_local, seq_len, head_dim)
-        output = output_flat.view(batch_size, seq_len, -1, head_dim)
-        output = output.transpose(1, 2)
+        # Reshape back to the format expected by GQAttention.forward()
+        if self.varlen_output:
+            # varlen path expects (bs*seqlen, n_heads, head_dim)
+            output = output_flat.view(batch_size * seq_len, -1, head_dim)
+        else:
+            # sdpa/flex path expects (batch, num_heads, seq_len, head_dim)
```
wwwjn: Why sdpa/flex will hit this path? `VllmAttention()` should only be patched during inference, not trainer path

torchtitan/models/qwen3/__init__.py
```diff
@@ -117,6 +117,37 @@
             backend="cos_sin",
         ),
     ),
+    "0.6B_varlen": Qwen3Model.Config(
```
wwwjn: Let's don't keep an copy of `0.6B_varlen` in core just for testing purpose? I don't want to end up with keep multiple copy of simliar configs in core. Let's replace the model's config field in rl/config_registry: https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/config_registry.py#L30

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -7,6 +7,18 @@
 import torch
 import torch.nn.functional as F
 
+from torchtitan.models.common.attention import VarlenMetadata
+
+
+def _build_single_seq_varlen_metadata(
```
acisseJZhong: I created some similar functions when integrating Titan into Verl: https://github.com/verl-project/verl/pull/5051/changes#diff-d90b34d27a9c17f5a898247fd8858ab3e61876d67ed2d8653e18bd5f85658a67R156. It uses varlen attention with document mask. Wondering if we could put these kinds of functions in varlen utils? @zhxchen17 @fegin

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -31,8 +43,15 @@ def compute_token_log_probs(
         full_sequence, dtype=torch.long, device=device
     ).unsqueeze(0)
 
+    # Build attention_masks based on attention backend
+    attn_backend = model.config.layer.attention.attn_backend
+    if attn_backend == "varlen":
```
acisseJZhong: just curious is it possible for us to use flex attention in RL?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -188,11 +188,13 @@ def __init__(
         head_dim: int,
         layer_name: str,
         scale: float | None = None,
+        varlen_output: bool = True,
```
acisseJZhong: right now the code assumes in RL we always using varlen attn? there's no need to use this flag?

torchtitan/experiments/rl/models/attention.py
zhxchen17: I can remove this path.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: I think we have plan to use flex in RL, @liangel-02 should have more context on this.

torchtitan/experiments/rl/models/attention.py
zhxchen17: Yes I assumed we have multiple backends in RL but I guess that's not the case for now. Removing.

torchtitan/experiments/rl/actors/utils.py
liangel-02: yeah we need flex attention for more complicated mask mods , ive extended the current vllm flex impl to accept user-defined mask mods (ie from torchtitan)

acisseJZhong: gotcha, then we could potentially also work on aligning the kernels for flex attention? cc @zhxchen17 

btw in Verl using Torchtitan, I am already using flex. Seems both flex and varlen can work. I created the custom mask https://github.com/verl-project/verl/pull/5051/changes#diff-d90b34d27a9c17f5a898247fd8858ab3e61876d67ed2d8653e18bd5f85658a67R156 but maybe we should move this into torchtitan.

torchtitan/experiments/rl/models/attention.py
acisseJZhong: maybe you should check on attn backend? since @liangel-02 mentions flex could also work for RL.

zhxchen17: For now I added an assertion to varlen backend, which we can expand in the future.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: Yeah we should align the kernels for flex. Just want to leave flex out the the current PR's scope.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: I took a look and seems the input format is a bit different from here. Here we already have prompt token and generate tokens, so it should be more straight forward to construct varlen meta with those directly.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: get_attention_masks() seems a bit different from what we need here. I think here the logic is more straight forward. To carve out the structure for future support on batch_size > 1, I just added a helper type TrainerInputBatch which should contain multiple sets of inputs and calculate varlen meta based on that.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -59,10 +66,49 @@ def compute_token_log_probs(
     return token_lps
 
 
+@dataclasses.dataclass
+class TrainerInputBatch:
+    prompt_and_gen_token_ids: list[tuple[list[int], list[int]]]
```
acisseJZhong: why do we have this as a list of a tuple of two things instead of concatenated tensor? For example in verl trainer the input batch should already be concated prompt + gen tokens.

torchtitan/experiments/rl/actors/utils.py
acisseJZhong: They should be similar,  https://fburl.com/1m1c000k the input batch is already prompt + response tokens. Wondering if we should move all mask creation utils into one place for RL.

torchtitan/experiments/rl/config_registry.py
```diff
@@ -26,8 +26,10 @@
 
 def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
     """GRPO training config for Qwen3-0.6B (6 GPUs: 4 gen + 2 train)."""
+    model_spec = model_registry("0.6B")
+    model_spec.model.layer.attention.attn_backend = "varlen"
```
tianyu-l: best practice: also change https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L534 to `block_causal`, although it is only used by pretraining trainer today (to generate proper varlen metadata)

torchtitan/experiments/rl/models/attention.py
```diff
@@ -310,6 +311,9 @@ def replace_with_vllm_attention(model, tp_degree=1):
 
         # GQA
         head_dim = model_args.layer.attention.head_dim
+        assert (
```
wwwjn, line 335: This check should been done in trainer model as well? `replace_with_vllm_attention` is only applied in generator model

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -4,28 +4,35 @@
...
         model: The model to use for computing logits
-        prompt_token_ids: Token IDs for the prompt
-        gen_token_ids: Token IDs for the generated completion
+        inputs: A TrainerInputBatch with batch_size == 1
         device: Device to run computation on
 
     Returns:
         Per-token log probabilities for the generated tokens
     """
+    assert inputs.batch_size == 1, "batch size > 1 not supported yet"
```
acisseJZhong: add a TODO comment to address bz>1?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -59,10 +66,49 @@ def compute_token_log_probs(
     return token_lps
 
 
+@dataclasses.dataclass
+class TrainerInputBatch:
```
tianyu-l: How about let's do this perf improvement in a separate PR.

torchtitan/experiments/rl/config_registry.py
tianyu-l: Maybe we can do: add an argument to `model_registry(flavor, attn_backend_override)` in which you override both attn_backend and mask_type.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -59,10 +66,49 @@ def compute_token_log_probs(
...
+
+def batch_trainer_inputs(
+    prompt_token_ids: list[list[int]], gen_token_ids: list[list[int]], batch: int = 1
+) -> Iterable[TrainerInputBatch]:
+    assert len(prompt_token_ids) == len(gen_token_ids)
+    assert batch == 1, "batch size > 1 not supported yet"
+
+    for i in range(0, len(prompt_token_ids), batch):
+        yield TrainerInputBatch(
+            list(zip(prompt_token_ids[i : i + batch], gen_token_ids[i : i + batch]))
```
acisseJZhong: can we concat them as a tensor instead of using list?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -283,17 +283,18 @@ async def step(self, episodes: list[Episode]) -> dict:
         ref_token_log_probs = []
         device = next(self.model.parameters()).device
         with torch.no_grad():
-            for prompt_toks, gen_toks in zip(my_prompt_token_ids, my_token_ids):
+            for inputs in batch_trainer_inputs(my_prompt_token_ids, my_token_ids):
```
wwwjn: Let's generalize the batched input in a separate PR

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -59,10 +66,49 @@ def compute_token_log_probs(
...
 
 
+@dataclasses.dataclass
+class TrainerInputBatch:
+    prompt_and_gen_token_ids: list[tuple[list[int], list[int]]]
+
+    def build_varlen_metadata(self, device: torch.device) -> VarlenMetadata:
+        """Build VarlenMetadata for all sequences in this batch."""
+
+        cu_seqs = torch.cumsum(
```
wwwjn: This flatten is only needed when batch size >1 , and the trainer is taking a **packed** input, right?

Whereas the model's `create_varlen_metadata_for_document` is taking a **padded** input . 

As @tianyu-l  comment, in RL system, we should use packed input instead of padded input

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -59,10 +66,49 @@ def compute_token_log_probs(
     return token_lps
 
 
+@dataclasses.dataclass
+class TrainerInputBatch:
+    prompt_and_gen_token_ids: list[tuple[list[int], list[int]]]
+
+    def build_varlen_metadata(self, device: torch.device) -> VarlenMetadata:
```
acisseJZhong: is it possible to us to reuse https://github.com/pytorch/torchtitan/blob/cc245f590909ab484a17c073020b364b3ddb9a04/torchtitan/models/common/attention.py#L432 by unifying the input to be torch Tensor instead of `list[tuple[list[int], list[int]]]`

tianyu-l: @joecummings has the idea of modifying `create_varlen_metadata_for_document` and not use eos_id but sequence lengths as input to the function, which can be reused here. But let's not do it in this PR.

acisseJZhong: this will be similar to the util we introduced in verl: creating mask based on positions? maybe we can unify all of them into one.

acisseJZhong: would love to have a separate PR to get rid of this `list[tuple[list[int], list[int]]]`, doesn't seems very intuitive to me.

torchtitan/models/common/attention.py
```diff
@@ -242,6 +243,8 @@ def forward(
         return_lse: bool = False,
         enable_gqa: bool = False,
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+        assert isinstance(block_mask, (BlockMask, type(None))), block_mask
```
Lucaskabela: We should make the error message here a bit more descriptive - like: 

```
        assert isinstance(block_mask, (BlockMask, type(None))), f"block_mask must instance of BlockMask or None, got {type(block_mask)}`
```

torchtitan/models/common/attention.py
```diff
@@ -157,6 +157,7 @@ def forward(
         attention_masks: VarlenMetadata,
         scale: float | None = None,
     ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+        assert isinstance(attention_masks, VarlenMetadata), attention_masks
```
Lucaskabela: We should make the error message here a bit more descriptive - like:
```python
        assert isinstance(attention_masks, (VarlenMetadata), f"attention_masks must be instance of VarlenMetadata but got {type(block_mask)}`
```

torchtitan/experiments/rl/tests/test_bitwise_identity.py
```diff
@@ -114,7 +115,9 @@ def create_vllm_engine(config):
     return engine
 
 
-def generate_with_vllm(engine, prompt, gen_config):
+def generate_with_vllm(
+    engine, prompt, gen_config
+) -> tuple[list[int], list[int], list[float]]:
```
Lucaskabela, line 118: Love type hint on the output :) maybe we can do it on the args too?

Lucaskabela (changes requested):

torchtitan/experiments/rl/actors/utils.py
zhxchen17: No we're not doing perf improvement in this PR, as TrainerInputBatch will always have batch_size == 1 in this PR.

tianyu-l: oh I see, sorry I misread. Then the `TrainerInputBatch` abstraction sounds not necessary? I'd suggest we hardcode minimal VarlenMetadata to unblock. Later we can invent dedicated input processor.

torchtitan/models/common/attention.py
zhxchen17: Sure, this was just a direct code move, but will add it.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: @tianyu-l to clarify: right now this is only a code refactor and we haven't supported batch > 1 in this PR yet. Maybe the code structure is a bit misleading and I can update it to make our intention more clear.

tianyu-l: right, I understand this is code refactor. But I think it's trying to add some abstraction that we don't need yet.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: Sure I will remove this type and replace with a simpler type for representing a single sequence only.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: Yeah makes sense. I will update the PR with concat-ed inputs.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -4,14 +4,66 @@
...
+
+    def __post_init__(self):
+        assert len(self.token_ids.shape) == 1, "Token IDs must have 1D"
+        assert self.token_ids.shape[0] == self.prompt_len + self.gen_len, (
+            f"Token IDs length mismatch: seq length {self.token_ids.shape[0]}"
+            + " != prompt length {self.prompt_len} + generated length {self.gen_len}"
+        )
+
+
+def build_varlen_metadata(
```
acisseJZhong, line 15: nit: can you leave a todo indicating we should either unify all the mask creation for RL or at least put them into the same file?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -200,6 +201,10 @@ def _build_model(
         Returns:
             Initialized model with weights loaded from checkpoint.
         """
+        assert (
```
acisseJZhong, line 205: can you leave a todo to support flex attention? in case we forgot 😄

torchtitan/experiments/rl/models/attention.py
acisseJZhong, line 335: same add a comment?

torchtitan/experiments/rl/config_registry.py
acisseJZhong: qq: why it's block causal instead of causal?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -4,14 +4,66 @@
...
+    )
+    max_len = max(seq.token_ids.shape[0] for seq in input_sequences)
+    return VarlenMetadata(
+        cu_seq_q=cu_seqs, cu_seq_k=cu_seqs, max_q=max_len, max_k=max_len
+    )
+
 
 def compute_token_log_probs(
     model: torch.nn.Module,
-    prompt_token_ids: list[int],
```
wwwjn, line 13: Let's don't introduce this `InputSequence` new abstraction, we can simply keep `prompt_token_ids` and `gen_token_ids` here.

These new abstraction we can discuss and design in separate PR

torchtitan/models/qwen3/__init__.py
```diff
@@ -399,11 +401,16 @@
 }
 
 
-def model_registry(flavor: str) -> ModelSpec:
+def model_registry(flavor: str, attn_backend_override: str | None = None) -> ModelSpec:
+    model = copy.deepcopy(qwen3_configs[flavor])
+    if attn_backend_override is not None:
+        model.layer.attention.attn_backend = attn_backend_override
```
wwwjn, line 412: Need some verify if this `attn_backend_override` is valid. Say we should reject attn_backend_override=`foo` override

torchtitan/models/qwen3/__init__.py
```diff
@@ -399,11 +401,16 @@
 }
 
 
-def model_registry(flavor: str) -> ModelSpec:
+def model_registry(flavor: str, attn_backend_override: str | None = None) -> ModelSpec:
+    model = copy.deepcopy(qwen3_configs[flavor])
```
wwwjn, line 405: Why we need a deep copy here? Can we use dataclass.replace?

torchtitan/experiments/rl/actors/utils.py
tianyu-l: right, let's not introduce this class for now

torchtitan/models/qwen3/__init__.py
zhxchen17: dataclass.replace() helps with shallow field update like `obj.field`. In this case using it would be a bit tedious since the attn_backend field is deeply nested inside model spec like `model.layer.attention.attn_backend`.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -7,29 +7,63 @@
...
+
+# TODO We should either unify all the mask creation for RL, or move them to a
+#      single file.
+def build_varlen_metadata(
+    input_sequences: list[tuple[torch.Tensor, int, int]], device: torch.device
+) -> VarlenMetadata:
+    """Build VarlenMetadata for all sequences in a batch."""
+    cu_seqs = torch.cumsum(
+        torch.tensor(
+            [0] + [token_ids.shape[0] for token_ids, _, _ in input_sequences],
```
acisseJZhong, line 21: curious why do we add 0 in between?

torchtitan/models/qwen3/__init__.py
```diff
@@ -399,11 +401,21 @@
 }
 
 
-def model_registry(flavor: str) -> ModelSpec:
+def model_registry(flavor: str, attn_backend_override: str | None = None) -> ModelSpec:
```
acisseJZhong, line 404: do you need to change the call site in RL? 

also wondering is attn_backend_override the only field we need to overwrite in RL? trying to think of if there's better ways to override from RL side instead of changing the original definition.

torchtitan/experiments/rl/actors/utils.py
zhxchen17: I think this is coming from varlen kernel's requirement. We always go from 0 as the index start.

torchtitan/models/qwen3/__init__.py
zhxchen17: > also wondering is attn_backend_override the only field we need to overwrite in RL?

Hmm I don't have a lot of context over how many fields are needed to be changed. I think a patching mechanism will be useful.

zhxchen17:
@acisseJZhong Created 2 issues for tracking the followups:
- https://github.com/pytorch/torchtitan/issues/2719
- https://github.com/pytorch/torchtitan/issues/2716

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -97,7 +130,13 @@ def compute_policy_gradient_loss(
     batch_token_log_probs = []
 
     for prompt_toks, gen_toks in zip(prompt_token_ids, vllm_token_ids):
-        token_lps = compute_token_log_probs(model, prompt_toks, gen_toks, device)
+        token_lps = compute_token_log_probs(
+            model,
+            [
+                build_full_sequence(prompt_toks, gen_toks, device),
```
wwwjn: Can we remove this helper function? seems it's only used here and it's very short. The name is a little bit confusing

torchtitan/experiments/rl/actors/utils.py
wwwjn: Seems this comment is not addressed, can we revert back to using `prompt_token_ids` and `gen_token_ids` for now? We don't need another `List` around them now, we can assume the input to `compute_token_log_probs` is one episode now.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -7,29 +7,63 @@
...
 
     Returns:
         Per-token log probabilities for the generated tokens
     """
-    full_sequence = prompt_token_ids + gen_token_ids
-    full_tensor = torch.tensor(
-        full_sequence, dtype=torch.long, device=device
-    ).unsqueeze(0)
+    # TODO Support input batching and expand this to batched inputs.
+    assert len(input_sequences) == 1, "batch size > 1 not supported yet"
```
wwwjn: Same, we can remove the check and remove the "batch" idea for now,  you can leave a sentence in `compute_token_log_probs`'s docstring to say this function accept one sequence for now.

torchtitan/experiments/rl/actors/utils.py
acisseJZhong: agree naming is confusing, this function is used in multiple places tho. And we should probably put in some data utils.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -200,6 +201,12 @@ def _build_model(
         Returns:
             Initialized model with weights loaded from checkpoint.
         """
+
+        # TODO Also support flex attention backend later.
+        assert (
```
tianyu-l, line 205: OK to land first, but this should be in the `VLLMGQAAttention` module in https://github.com/pytorch/torchtitan/pull/2709
cc @wwwjn

torchtitan/experiments/rl/actors/utils.py
tianyu-l: This function is 
- only used in `compute_token_log_probs`
- short

so agreed we can just directly expand the function content there.

torchtitan/models/qwen3/__init__.py
tianyu-l: we already have check here https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L502

although since we are using deepcopy and patching, the post init may not be called.

tianyu-l (approved):
please address remaining comments before merge

PR #2669, opened 2026-03-23
[graph_trainer] Add disk storage adapter for precompile artifacts

bobrenjc93:
* #2672
* #2671
* #2670
* __->__ #2669

Add a StorageAdapter abstraction with a DiskStorageAdapter implementation
for persisting compiled training artifacts to disk. This provides a clean
interface for save/load/exists operations on serialized artifacts, keyed
by a string identifier (e.g. "default_rank0").

### DIFF
--- a/torchtitan/experiments/graph_trainer/storage.py
+++ b/torchtitan/experiments/graph_trainer/storage.py
@@ -6,21 +6,33 @@
 
 from __future__ import annotations
 
-import os
 import tempfile
 from abc import ABC, abstractmethod
 from pathlib import Path
 
 
 class StorageAdapter(ABC):
+    """Interface for storing and retrieving precompiled artifacts.
+
+    Keys are flat, opaque strings (no hierarchy or path separators).
+    Implementations map keys to storage locations internally — callers
+    should not embed "/" or other filesystem-specific characters in keys.
+    """
+
     @abstractmethod
     def save(self, key: str, data: bytes) -> str:
-        """Save data under the given key. Returns the path/URI of the saved artifact."""
+        """Save data under the given key. Returns the path/URI of the saved artifact.
+
+        If an artifact already exists for the key, it is silently overwritten.
+        """
         ...
 
     @abstractmethod
     def load(self, key: str) -> bytes:
-        """Load data for the given key."""
+        """Load data for the given key.
+
+        Raises FileNotFoundError if no artifact exists for the key.
+        """
         ...
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/graph_trainer/storage.py
```diff
@@ -0,0 +1,79 @@
...
+
+class StorageAdapter(ABC):
+    @abstractmethod
+    def save(self, key: str, data: bytes) -> str:
+        """Save data under the given key. Returns the path/URI of the saved artifact."""
+        ...
+
+    @abstractmethod
+    def load(self, key: str) -> bytes:
+        """Load data for the given key."""
```
aorenste: Should define what happens if the key isn't found.

torchtitan/experiments/graph_trainer/storage.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+import tempfile
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+
+class StorageAdapter(ABC):
+    @abstractmethod
+    def save(self, key: str, data: bytes) -> str:
+        """Save data under the given key. Returns the path/URI of the saved artifact."""
```
aorenste: Does save of an existing key overwrite or fail? Should be noted and tested.

torchtitan/experiments/graph_trainer/storage.py
```diff
@@ -0,0 +1,79 @@
...
+
+from __future__ import annotations
+
+import os
+import tempfile
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+
+class StorageAdapter(ABC):
```
aorenste, line 14: Needs a comment about what format the keys take - are they flat strings or a hierarchy? Is there a limit on the charset? If you're expecting a hierarchy I have a (minor) concern that "/" isn't always the proper directory separator (windows).

torchtitan/experiments/graph_trainer/storage.py
```diff
@@ -0,0 +1,79 @@
...
+        if not path.is_relative_to(self.base_dir):
+            raise ValueError(f"Key {key!r} resolves outside base directory")
+        return path
+
+    def save(self, key: str, data: bytes) -> str:
+        path = self._path_for(key)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        # Write to a temp file then atomically rename to avoid
+        # leaving partial files if the process crashes mid-write.
+        fd, tmp_path = tempfile.mkstemp(dir=path.parent)
```
aorenste: This might be a little cleaner using a NamedTemporaryFile...

PR #2670, opened 2026-03-23
[graph_trainer] Add precompile artifact serialization and loading

bobrenjc93:
* #2672
* #2671
* __->__ #2670

Add precompile_save() and precompile_load() functions that serialize and
deserialize compiled AOT graphs using BundledAOTAutogradSerializableCallable.
Artifacts include the serialized compiled function, parameter/buffer specs,
input/output tree specs, and metadata. Deserialization is deferred to first
call so Triton kernels load on the correct CUDA device.


torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+
+from torchtitan.experiments.graph_trainer.storage import StorageAdapter
+from torchtitan.tools.logging import logger
+
+
+@dataclass
+class PrecompiledArtifact:
+    serialized_fn: bytes
+    params_spec: list[str]
+    buffers_spec: list[str]
```
aorenste: nit: tuple

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+        return compiled_fn
+    wrapped = getattr(compiled_fn, "__wrapped__", None)
+    if isinstance(wrapped, BundledAOTAutogradSerializableCallable):
+        return wrapped
+    raise TypeError(
+        f"precompile_save requires the compiled function to be a "
+        f"BundledAOTAutogradSerializableCallable, but got "
+        f"{type(compiled_fn).__name__}. Ensure your compiler pass "
+        f"pipeline produces serializable output (e.g. by including "
+        f"'full_inductor_compilation' in --compile.passes)."
```
aorenste: nit: some unnecessary f-prefixes here...

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+    Extract the BundledAOTAutogradSerializableCallable from compiled_fn.
+    PyTorch's aot_compile_joint_with_descriptors wraps the serializable
+    callable in a plain function via functools.wraps, so we check both
+    the direct object and the __wrapped__ attribute.
+    """
+    if isinstance(compiled_fn, BundledAOTAutogradSerializableCallable):
+        return compiled_fn
+    wrapped = getattr(compiled_fn, "__wrapped__", None)
+    if isinstance(wrapped, BundledAOTAutogradSerializableCallable):
+        return wrapped
```
aorenste: Should this be a loop? (A wrapped, wrapped, wrapped, wrapped BundledAOTAutogradSerializableCallable...)

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+    out_spec: pytree.TreeSpec | None
+    metadata: dict[str, Any] = field(default_factory=dict)
+    config_fingerprint: str = ""
+
+
+def compute_config_fingerprint(
+    model: torch.nn.Module,
+    compile_config: GraphTrainerCompileConfig,
+    parallel_dims: ParallelDims,
+) -> str:
```
aorenste: nit: consider using a NewType to make the fingerprint clearer than a simple str?

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def precompile_save(
+    model: torch.nn.Module,
+    compiled_fn: BundledAOTAutogradSerializableCallable,
+    storage: StorageAdapter,
+    artifact_key: str,
+    out_spec: pytree.TreeSpec | None,
+    metadata: dict[str, Any] | None = None,
+    config_fingerprint: str = "",
```
aorenste: Maybe use None as the "no fingerprint" sentinel? For that matter do we want it to be optional at all? Below we treat that as "legacy".

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+            f"Parameter mismatch between saved artifact and current model. "
+            f"Saved: {artifact.params_spec}, Current: {current_params}"
+        )
+    if current_buffers != artifact.buffers_spec:
+        raise ValueError(
+            f"Buffer mismatch between saved artifact and current model. "
+            f"Saved: {artifact.buffers_spec}, Current: {current_buffers}"
+        )
+
+    if expected_fingerprint and artifact.config_fingerprint:
```
aorenste, line 178: Do we want an override envvar so during debugging we can just say "trust me, bro" without having to change the code?

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+                f"current fingerprint: {expected_fingerprint}. "
+                f"Delete the stale artifact and re-run with precompile to "
+                f"generate a fresh one."
+            )
+    elif expected_fingerprint and not artifact.config_fingerprint:
+        logger.warning(
+            "Precompiled artifact has no config fingerprint (legacy artifact). "
+            "Skipping fingerprint validation. Re-save the artifact to enable "
+            "fingerprint checks."
+        )
```
aorenste, line 202: I thought the artifacts were ephemeral - what's a legacy artifact?

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -0,0 +1,237 @@
...
+            compiled_fn = (
+                BundledAOTAutogradSerializableCallable.deserialize_compile_artifacts(
+                    serialized_fn_bytes
+                )
+            )
+
+        # Build the flat input list: params + buffers + user args.
+        # This mirrors the calling convention in joint_graph_builder's
+        # wrapper_fn (graph_utils.py).
+        inputs = [
```
aorenste: nit: I think `tuple(*model.parameters(), *model.buffers(), *args)` would be ever so slightly faster because then python wouldn't have to convert the list into a tuple on the compiled_fn call.

torchtitan/experiments/graph_trainer/precompile.py
bobrenjc93: The precompiled artifacts are persisted on disk/nfs so if you rerun with the same fingerprint you will get reuse by default.

PR #2671, opened 2026-03-23
[graph_trainer] Integrate precompile into AOT compilation path

bobrenjc93:
* #2672
* __->__ #2671

Add precompile and precompile_artifact_dir config fields to
GraphTrainerCompileConfig, and wire serializable compilation through
joint_graph_builder and the AOT compile dispatcher:

- joint_graph_builder: add serializable and on_compile params. When
  serializable=True, passes it to aot_compile_joint_with_descriptors.
  Calls on_compile callback after compilation with the compiled fn and
  tree specs.

- CompiledModule: add precompiled_fn param. When set, uses the
  precompiled function directly instead of calling joint_graph_builder.

- _apply_aot_compile: when precompile=True, checks for existing
  artifacts and loads them (skipping compilation). Otherwise compiles
  with serializable=True and saves the artifact via on_compile callback.

- _make_precompile_callback: builds the on_compile callback that saves
  the compiled artifact to disk.

- regional_inductor_pass: add serializable mode that sets
  force_autograd_cache and overrides the ops filter to allow
  _c10d_functional distributed collective ops.

- simple_fsdp: make dynamically generated SimpleFSDP* classes
  picklable by registering them in sys.modules with unique names.

### DIFF
--- a/torchtitan/experiments/graph_trainer/compile.py
+++ b/torchtitan/experiments/graph_trainer/compile.py
@@ -97,7 +97,7 @@ def _make_precompile_callback(
     artifact_key: str | None = None,
 ):
     """Build the on_compile callback that saves the compiled artifact to disk."""
-    from torchtitan.experiments.graph_trainer.precompile import (
+    from .precompile import (
         compute_config_fingerprint,
         precompile_save,
     )
@@ -145,7 +145,7 @@ def _apply_aot_compile(
         storage, artifact_key = _get_precompile_storage_and_key(compile_config)
 
         if storage.exists(artifact_key):
-            from torchtitan.experiments.graph_trainer.precompile import (
+            from .precompile import (
                 compute_config_fingerprint,
             )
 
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/graph_trainer/compile.py
```diff
@@ -71,6 +136,40 @@ def _apply_aot_compile(
...
+    # When precompile is enabled, compute storage/key once and reuse them
+    # for both the load attempt and the save callback to avoid duplicate
+    # DiskStorageAdapter construction and fingerprint computation.
+    storage: StorageAdapter | None = None
+    artifact_key: str | None = None
+    if compile_config.precompile:
+        storage, artifact_key = _get_precompile_storage_and_key(compile_config)
+
+        if storage.exists(artifact_key):
+            from torchtitan.experiments.graph_trainer.precompile import (
```
aorenste: nit: `from .precompile import...` (so if you move the files it still works) (here and elsewhere in the file)

PR #2672, opened 2026-03-23
[graph_trainer] Add Llama3 precompile config, tests, and docs

bobrenjc93:
* __->__ #2672

----

- Add llama3_precompile and llama3_precompile_regional config registry
  entries for precompilation with full_inductor_compilation and
  regional_inductor passes respectively
- Add integration tests for both precompile variants
- Document precompilation workflow in README

## Validation

Validated on Llama3 debugmodel with AOT compilation (FSDP dp=4, TP=2)
on 8 GPUs using --debug.deterministic --debug.seed=42 for 10 steps.

Three runs compared for each pass (full_inductor_compilation and
regional_inductor):
1. AOT baseline: same passes, precompile=false
2. Cold run: precompile=true — compiles, saves artifact, trains
3. Warm run: precompile=true — loads saved artifact, trains

All three produce bitwise-identical loss curves (max |diff| = 0.0).

### full_inductor_compilation
image

### regional_inductor
image

Test scripts and plotting code:
https://gist.github.com/bobrenjc93/192ea9e481422d68117ef3f1add1892b


torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,49 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
 
+### Pre-compile (Serializable Compilation)
+
+Pre-compile lets you save compiled AOT graphs to disk on the first run and
+reload them on subsequent runs, skipping compilation entirely. This is useful
+for large models where compilation time is significant.
+
+```bash
+# First run: compiles with serializable=True, saves artifacts, then trains
+torchrun --nproc_per_node=8 --virtual-local-rank \
```
ezyang, line 87: This is the non-CooR flow? Because it really shouldn't be necessary to do a torchrun to precompile

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,49 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    --parallelism.tensor_parallel_degree 2
+
+# Subsequent runs: detects existing artifacts, loads them, skips compilation
+torchrun --nproc_per_node=8 --virtual-local-rank \
+    -m torchtitan.train \
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_precompile \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2
+```
```
ezyang, line 103: It would be far better to do a realistic size that is multi-node for the example here (the single node setup is... useful I guess for testing, but it's not really the point of precompile)

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,49 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+
+Pre-compile works with any compiler pass that produces serializable output,
+including `full_inductor_compilation` and `regional_inductor`. Two config
+variants are provided:
+
+- `graph_trainer_llama3_precompile` — full Inductor compilation
+- `graph_trainer_llama3_precompile_regional` — regional Inductor compilation
+
+Artifacts are stored in `/tmp/precompile_artifacts/` by default (configurable
+via `--compile.precompile_artifact_dir`).
```
ezyang: We shouldn't default this. Force the user to say where to put it.

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,49 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2
+
+# Subsequent runs: detects existing artifacts, loads them, skips compilation
+torchrun --nproc_per_node=8 --virtual-local-rank \
+    -m torchtitan.train \
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_precompile \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2
```
ezyang, line 102: Since the parallelism gets baked into the compiled code, it seems weird that we force you to have to specify it again. (Similarly, having to specify the module again; wouldn't it be easier to just point it at the file and it has the module already in it.)

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,49 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
 MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --compile.mode jit --compile.backend inductor
 ```
 
+### Pre-compile (Serializable Compilation)
+
+Pre-compile lets you save compiled AOT graphs to disk on the first run and
+reload them on subsequent runs, skipping compilation entirely. This is useful
+for large models where compilation time is significant.
```
ezyang, line 76: Ephemerality needs to be discussed here

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,49 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+via `--compile.precompile_artifact_dir`).
+
+#### Validation
+
+Pre-compile has been validated for bitwise equivalence on Llama3
+with 2D parallelism (FSDP dp=4, TP=2) on 8 GPUs for both
+`full_inductor_compilation` and `regional_inductor` passes. The three
+paths — baseline (no precompile), precompile-save (first run), and
+precompile-load (subsequent run) — produce identical loss values across all
+training steps.
```
ezyang: This info is useful but I am not so sure it belongs in the README. Anyway, a single node validation isn't that interesting anyway lol.

What might be useful for people, is a recipe for doing this validation yourself when comparing precompile and JIT. Good for debugging!

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: Yeah that's right. As discussed last Friday my goal is to land all the precompile stuff first while Aaron lands his CooR in parallel and then I can integrate afterwards.

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: IMO there's two advantages of this approach:

1) Users are clear with their intent, so we can sanity check if the parallelisms are the same as the ones encoded in the precompiled artifact
2) Since we can detect fingerprint mismatch, we can also automatically redo the precompile. This creates a much more ergonomic flow for users where they can simply change parallelisms in the CLI args and relaunch their jobs with the precompile flag and it "just works"

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: I'll kick off a run today and update accordingly

bobrenjc93: Eh this is just awkward without CooR (since what are we going to do, torchrun and have N GPU nodes write to NFS just to rip all this bespoke logic out once CooR lands?). Maybe I'll set this PR as the assembly point for precompile and CooR and I'll do both in these multi node runs and we can have a single readme update that has the final verison with both precompile and coor.

torchtitan/experiments/graph_trainer/llama3/config_registry.py
```diff
@@ -48,3 +48,26 @@ def graph_trainer_llama3_405b() -> GraphTrainer.Config:
     config = to_graph_trainer_config(llama3_405b(), model_registry)
     config.compile = GraphTrainerCompileConfig(enable=True)
     return config
+
+
+def graph_trainer_llama3_precompile() -> GraphTrainer.Config:
```
yiming0416, line 53: Do we need a separate config here? You can just reuse the existing llama3_debugmodel config and add `--compile.precompile`?

bobrenjc93:
Going to close this out since we are going to straight shot to PCooR

bobrenjc93:
replaced by https://github.com/pytorch/torchtitan/pull/2713

PR #2675, opened 2026-03-23
[Bugfix] Fix missing attribute from compile

Lucaskabela:
Minor fix in https://github.com/pytorch/torchtitan/pull/2568 to move apply_compile to distributed/compile caused a bug due to missing `.named_children()` in this integration.  So, update to fix (avoid crashing) for compile on the trainer

### DIFF
--- a/torchtitan/distributed/compile.py
+++ b/torchtitan/distributed/compile.py
@@ -167,8 +167,9 @@ def apply_compile_dense_rl(model: nn.Module, compile_config: CompileConfig) -> N
     Also fixes some numeric issues we have observed with ``torch.compile()``.
     See https://github.com/pytorch/torchtitan/issues/2673 for more details.
     """
-    # pyrefly: ignore [missing-attribute]
-    for transformer_block in model.layers.named_children.values():
+    # pyrefly: ignore [missing-attribute, not-callable]
+    for _, transformer_block in model.layers.named_children():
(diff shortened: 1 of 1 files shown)


wwwjn (approved):

torchtitan/distributed/compile.py
```diff
@@ -167,8 +167,9 @@ def apply_compile_dense_rl(model: nn.Module, compile_config: CompileConfig) -> N
     Also fixes some numeric issues we have observed with ``torch.compile()``.
     See https://github.com/pytorch/torchtitan/issues/2673 for more details.
     """
-    # pyrefly: ignore [missing-attribute]
-    for transformer_block in model.layers.named_children.values():
+    # pyrefly: ignore [missing-attribute, not-callable]
+    for transformer_block in model.layers.values():
```
acisseJZhong: can we use `for layer_id, transformer_block in model.layers.named_children():` which is what it is before?

Lucaskabela: layer_id won't be used so we can just use `_` - named_children() should work though

acisseJZhong: I see, for compile dense it was used for `model.layers.register_module(layer_id, transformer_block)`, so does that mean model.compile here will change the module in place?

acisseJZhong (approved):

PR #2677, opened 2026-03-23
Fix Config.to_dict() to serialize list fields containing Config objects

anshul-si:
**Summary:** Config.to_dict() did not handle list fields containing Config objects. When a config like llama3_8b_float8 set model_converters with a list of Float8LinearConverter.Config objects, to_dict() passed the raw list through without serializing its elements. This caused json.dumps() to fail with TypeError: Object of type Config is not JSON serializable when maybe_log() attempted to log the config.

In order to reproduce the issue, you can run MODULE=llama3 CONFIG=llama3_debugmodel_float8 ./run_train.sh --debug.print-config. Issue most likely has not surfaced yet because we have debug.print-config disabled by default locally. 

**Test Case** 
1. MODULE=llama3 CONFIG=llama3_debugmodel_float8 ./run_train.sh --debug.print-config.

* __->__ #2677

### DIFF
--- a/torchtitan/config/configurable.py
+++ b/torchtitan/config/configurable.py
@@ -5,8 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 
 import dataclasses
+import logging
 from dataclasses import dataclass, fields, replace
 from typing import ClassVar
+
+logger = logging.getLogger(__name__)
 
 import torch
 
@@ -88,10 +91,11 @@ class Configurable:
                 elif isinstance(val, (str, int, float, bool, type(None))):
                     return val
                 else:
-                    raise TypeError(
+                    logger.warning(
                         f"Config field value of type {type(val).__name__} "
-                        f"is not JSON serializable"
+                        f"may not be JSON serializable"
                     )
+                    return val
 
             result = {}
             for f in fields(self):


torchtitan/config/configurable.py
```diff
@@ -86,6 +86,10 @@ def to_dict(self) -> dict:
                     result[f.name] = val.to_dict()
                 elif dataclasses.is_dataclass(val):
                     result[f.name] = dataclasses.asdict(val)
+                elif isinstance(val, list):
+                    result[f.name] = [
+                        v.to_dict() if hasattr(v, "to_dict") else v for v in val
```
tianyu-l: doesn't sound right -- something doesn't have to_dict doesn't mean it's not serializable

torchtitan/config/configurable.py
```diff
@@ -86,6 +86,10 @@ def to_dict(self) -> dict:
...
+                    ]
                 else:
                     result[f.name] = val
```
tianyu-l: should replace this with warning, too?

torchtitan/config/configurable.py
tianyu-l: also we shouldn't silently skip, we should at least `logger.warning`

torchtitan/config/configurable.py
```diff
@@ -75,19 +75,32 @@ def __repr__(self) -> str:
...
+                elif dataclasses.is_dataclass(val):
+                    return dataclasses.asdict(val)
+                elif isinstance(val, (list, tuple)):
+                    return type(val)(_convert(v) for v in val)
+                elif isinstance(val, dict):
+                    return {k: _convert(v) for k, v in val.items()}
+                elif isinstance(val, (str, int, float, bool, type(None))):
+                    return val
+                else:
+                    raise TypeError(
```
tianyu-l: can quantization converters be included in the dict nicely now?

anshul-si: yes, this solves the problem

torchtitan/config/configurable.py
```diff
@@ -75,19 +75,32 @@ def __repr__(self) -> str:
...
+                        f"Config field value of type {type(val).__name__} "
+                        f"is not JSON serializable"
+                    )
```
fegin, line 97: I think we should just have a warning instead of a hard error. We don't have a strong serializable guarantee yet. Some existing cases or future PRs may be broken due to this error.

tianyu-l: That's reasonable, although I prefer we leave this "not able to serializing" to be identified by CI / later PR, instead of silent CI passing.

PR #2678, opened 2026-03-23
Add pack_samples validation and default to flex+block_causal

joecummings:
> This changes numerics so everything is going to break :)

#### Context

- Adds `pack_samples` field to `BaseDataLoader.Config` (default `False`), overridden to `True` in `HuggingFaceTextDataLoader.Config` since the text dataloader always packs multiple documents per sequence
- Validates at config time (`Trainer.Config.__post_init__`) that `pack_samples=True` requires `flex` or `varlen` attention with `block_causal` masking to prevent cross-document attention leakage
- Updates all model configs (llama3, qwen3, deepseek_v3, gpt_oss) to use `flex` + `block_causal` as the default attention configuration, replacing `sdpa` where it was previously the default

#### Test plan

- [x] Added `TestPackSamplesValidation` with 7 unit tests covering: default passes, False skips validation, sdpa raises, flex+causal raises, flex+block_causal passes, varlen+block_causal passes, model_spec=None skips
- [x] All existing `TestConfigManager` tests pass
- [x] `pre-commit run --all-files` passes (except pre-existing pyrefly env issue)


torchtitan/models/llama3/__init__.py
tianyu-l: Should we remove these for all models, or switch the role and make them `debugmodel_sdpa_attn`? I think we should do latter, and update the CI as well to continue testing both.

torchtitan/components/dataloader.py
```diff
@@ -47,6 +47,10 @@ class Config(Configurable.Config):
         dataset: str = ""
         dataset_path: str | None = None
 
+        pack_samples: bool = False
+        """Whether multiple samples are packed into each training sequence.
+        Requires flex with block_causal masking or varlen attention."""
```
tianyu-l, line 52: I don't think we need it here, because packing sounds txt specific, and maybe an image dataset has nothing to "pack".

joecummings: I'd prefer to keep it so we can access directly in training loop as `dataloader.pack_samples` for validation rather than something ugly like `hasattr(dataloader, "pack_samples")`. As for images, there is a notion of "packing" ([Llava paper](https://arxiv.org/abs/2509.23661)) and it looks like there's a packing strategy already under [experiments/vlm](https://github.com/pytorch/torchtitan/blob/8953d2e39ee113ebbfb38e70294bc4931579a4fe/torchtitan/experiments/vlm/datasets/utils/packing.py#L16)

torchtitan/models/llama3/__init__.py
joecummings: Latter is good

tests/integration_tests/features.py
```diff
@@ -494,7 +494,7 @@ def build_features_test_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
-                    "--module llama3 --config llama3_debugmodel_flex_attn",
+                    "--module llama3 --config llama3_debugmodel",
```
joecummings: default model now uses flex attention so this is equivalent

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -202,6 +202,9 @@ class Config(ParallelAwareDataloader.Config):
         dataset: str = "c4_test"
         """Dataset to use"""
 
+        pack_samples: bool = True
```
tianyu-l, line 205: Shall we assert it somewhere in this file? Users could have changed it but there's no support for not packing.

torchtitan/models/deepseek_v3/config_registry.py
```diff
@@ -57,9 +57,10 @@ def deepseek_v3_debugmodel() -> Trainer.Config:
     )
 
 
-def deepseek_v3_debugmodel_flex_attn() -> Trainer.Config:
+def deepseek_v3_debugmodel_sdpa_attn() -> Trainer.Config:
     config = deepseek_v3_debugmodel()
-    config.model_spec = model_registry("debugmodel_flex_attn")
+    config.model_spec = model_registry("debugmodel_sdpa_attn")
+    config.dataloader.pack_samples = False
```
tianyu-l, line 63: this is hacky -- you are still packing, but changing the config to bypass post_init validation.

torchtitan/trainer.py
```diff
@@ -116,6 +116,28 @@ def __post_init__(self):
                         "Optimizers in backward is not supported with Pipeline Parallel."
                     )
 
+            if self.dataloader.pack_samples and self.model_spec is not None:
```
tianyu-l, line 119: I can see how this is blocking test on sdpa, because packing is always true but sdpa doesn't support block_causal.

One way around may be: we only print a warning instead of exception here, when people combine packing and pure causal.

PR #2680, opened 2026-03-24
[RL] Add parallelism plan for qwen3 30B-A3B MoE to run e2e

acisseJZhong:


Lucaskabela:
https://github.com/pytorch/torchtitan/pull/2688 is WIP to migrate to using `.compile()` for parallelisms so we won't need a unique one - there is a pytorch side change for checkpointing needed to land first though but it is referenced in the PR so you can patch both if you would like to test

PR #2683, opened 2026-03-24
[rl] batch invariant mode under same parallelism

wwwjn:
## What changed
- Currently only support both trainer and generator using bfloat16. (working on consolidating the details of dtype conversion, but currently FA3 kernel only support bf16/f16)
- Tested with compile("eager") and cudagraph("piecewise"), the bitwise identity are still preserved.

## Test
TP=2 on both trainer and generator:
- Verified after enabling batch-invariant mode, run-to-run deterministic is also so enabled as prerequisite .
```
[actor=<root>] Eval: Accuracy=20% (4/20) Format=20% (4/20)
[actor=<root>] Starting RL training for 10 steps
[actor=<root>] Weight sync: push=0.198s, total=0.827s
[actor=<root>] Step  0 | Loss: -0.0041 | Reward: -0.100 | Correct: 18/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 52.1s
[actor=<root>] Weight sync: push=0.203s, total=0.400s
[actor=<root>] Step  1 | Loss: -0.0047 | Reward: -0.350 | Correct: 13/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 44.8s
[actor=<root>] Weight sync: push=0.197s, total=0.384s
[actor=<root>] Step  2 | Loss: -0.0067 | Reward: -0.300 | Correct: 14/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 46.6s
[actor=<root>] Weight sync: push=0.196s, total=0.414s
[actor=<root>] Step  3 | Loss: -0.0064 | Reward: -0.100 | Correct: 18/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 46.3s
[actor=<root>] Weight sync: push=0.204s, total=0.413s
[actor=<root>] Step  4 | Loss: -0.0084 | Reward: -0.400 | Correct: 12/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 46.7s
[actor=<root>] Weight sync: push=0.211s, total=0.413s
[actor=<root>] Step  5 | Loss: -0.0120 | Reward: +0.300 | Correct: 26/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 44.1s
[actor=<root>] Weight sync: push=0.186s, total=0.385s
[actor=<root>] Step  6 | Loss: -0.0051 | Reward: -0.350 | Correct: 13/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 47.6s
[actor=<root>] Weight sync: push=0.223s, total=0.426s
[actor=<root>] Step  7 | Loss: -0.0058 | Reward: -0.600 | Correct:  8/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 47.9s
[actor=<root>] Weight sync: push=0.206s, total=0.408s
[actor=<root>] Step  8 | Loss: -0.0075 | Reward: -0.250 | Correct: 15/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 44.5s
[actor=<root>] Weight sync: push=0.200s, total=0.402s
[actor=<root>] Step  9 | Loss: -0.0124 | Reward: +0.100 | Correct: 22/40 | Avg tokens: 100 | Logprob diff: mean=0.0000e+00, max=0.0000e+00 | Time: 47.8s
[actor=<root>] Eval: Accuracy=35% (7/20) Format=50% (10/20)
[actor=<root>] Pre-training:  Accuracy=20% (4/20) Format=20% (4/20)
[actor=<root>] Post-training: Accuracy=35% (7/20) Format=50% (10/20)
```

## Config for bit-wise identity:
- RLTrainer.debug is set from config_registry. When `batch_invariant_mode=True`, we check `deterministic` must be true, and dtype on both trainer and generator needs to be `bfloat16` in `__post__init`
    - PolicyTrainer.debug is set by RLTrainer.debug
        - PolicyTrainer model's inner_attn (VarlenAttention) has a `batch_invariant` flag, this flag is set by PolicyTrainer
    - Generator.debug is set by RLTrainer.debug
        - Generator model's inner_attn (VLLMAttentionWrapper) enables `batch_invariant` mode by checking `is_batch_invariant_mode_enabled()` global variable. 

## Attention Part
Attention changes (for both VarlenAttention in core torchtitan, and VLLMAttentionWrapper in RL):
- If the hardware is Hopper, always use FA3 attention. If not, use FA2 for Blackwell hardware.
- If we are using FA3 attention, then batch-invariant is enabled by setting `num_split=1`
- If  we are using FA2 attention, batch-invariant is automatically supported.

### DIFF
--- a/torchtitan/experiments/rl/config_registry.py
+++ b/torchtitan/experiments/rl/config_registry.py
@@ -150,7 +150,7 @@ def rl_grpo_qwen3_0_6b_batch_invariant() -> RLTrainer.Config:
 
     Enables deterministic + batch-invariant mode for true on-policy RL training.
     """
-    model_spec = model_registry("0.6B", attn_backend_override="varlen")
+    model_spec = model_registry("0.6B_varlen")
     batch_invariant_config = DebugConfig(batch_invariant=True, deterministic=True)
     return RLTrainer.Config(
         model_spec=model_spec,


wwwjn:
Need to rebase onto #2649 and clean up the varlen patch in trainer side in this PR

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -194,15 +232,39 @@ def _build_model(
...
             Initialized model with weights loaded from checkpoint.
         """
+        # Override attention backend to varlen when batch-invariant mode is
+        # enabled.  The shared model_spec keeps SDPA (needed by the generator's
+        # vLLM wrapper), but the trainer uses varlen attention backed by
+        # ``varlen_attn`` + FA3 for deterministic, batch-invariant computation
+        # that is bitwise-identical to the generator's CUSTOM backend.
+        if batch_invariant_mode:
+            attn_cfg = model_spec.model.layer.attention
+            new_attn = dataclasses.replace(attn_cfg, attn_backend="varlen")
```
wwwjn: NOTE: this attention part need to rebase on to #2649

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -6,6 +6,46 @@
 
 import torch
 import torch.nn.functional as F
+from torchtitan.models.common.attention import VarlenMetadata
+
+
+def _build_single_seq_varlen_metadata(
```
wwwjn: Again, this part needs to be rebased onto #2649

torchtitan/experiments/rl/models/attention.py
```diff
@@ -320,6 +328,12 @@ def replace_with_vllm_attention(model, tp_degree=1):
         )
 
         layer.attention.inner_attention = vllm_attn
+        # Override attn_backend so GQAttention.forward() uses the sdpa
+        # dispatch path (which passes attention_masks=None).  The model_spec
+        # may have attn_backend="varlen" for the trainer, but the generator's
+        # VLLMAttention handles attention independently via vLLM's paged
+        # attention and doesn't need VarlenMetadata.
+        layer.attention.attn_backend = "sdpa"
```
wwwjn: This is hacky but still needs to rebase onto #2649

torchtitan/models/qwen3/__init__.py
```diff
@@ -399,11 +401,34 @@
 }
 
 
-def model_registry(flavor: str) -> ModelSpec:
+def model_registry(
```
wwwjn: Again, this part will be removed after #2649 landed

torchtitan/experiments/rl/models/attention.py
zhxchen17: just heads up #2649 is merged

torchtitan/experiments/rl/tests/test_bitwise_identity.py
```diff
@@ -231,20 +259,23 @@ def build_trainer_model(config):
...
+        model_spec=model_registry(
+            "0.6B", attn_backend_override="varlen", batch_invariant=True
+        ),
         hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
         batch_invariant_mode=True,
         trainer=PolicyTrainer.Config(
+            training=TrainingConfig(dtype="bfloat16"),
             parallelism=ParallelismConfig(
-                tensor_parallel_degree=2,
+                tensor_parallel_degree=1,
```
Lucaskabela: Why tp = 1here?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+
+# ---------------------------------------------------------------------------
+# Enable / disable
+# ---------------------------------------------------------------------------
+
+_SAVED_STATE: dict[str, Any] | None = None
+
+
+def is_batch_invariant_mode_enabled() -> bool:
+    return _upstream_is_enabled()
```
acisseJZhong: qq: why not just use is_batch_invariant_mode_enabled?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+        "tf32_matmul": torch.backends.cuda.matmul.allow_tf32,
+        "tf32_cudnn": torch.backends.cudnn.allow_tf32,
+        "deterministic": torch.are_deterministic_algorithms_enabled(),
+    }
+
+    # Set NCCL env vars for deterministic inter-GPU collectives (all-reduce
+    # in TP/DP). Must be set BEFORE dist.init_process_group so NCCL picks
+    # them up. Without these, NCCL may use Tree/CollNet algorithms or
+    # multiple channels, changing the floating-point summation order.
+    os.environ["NCCL_ALGO"] = "Ring"
```
acisseJZhong: just out of curiosity, how much slower is RING than tree based algorithm? And RING based all reduce is deterministic while tree based is not?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+    os.environ["NCCL_MIN_NCHANNELS"] = "1"
+    os.environ["NCCL_MAX_NCHANNELS"] = "1"
+    os.environ["NCCL_PROTO"] = "Simple"
+    os.environ["NCCL_COLLNET_ENABLE"] = "0"
```
acisseJZhong: what does this do?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+    os.environ["NCCL_NVLS_ENABLE"] = "0"
```
acisseJZhong: I think this should be false regardless of batch inv? I prepend this to my command all the time due to my devgpu.

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # Disable reduced-precision reductions: these allow cuBLAS to use
+    # lower-precision accumulation that can round differently depending
+    # on batch size / tile decomposition.
+    torch.backends.cuda.matmul.allow_bf16_reduced_precision_reduction = False
+    torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction = False
+
+    # Disable TF32 for exact fp32 accumulation
+    torch.backends.cuda.matmul.allow_tf32 = False
+    torch.backends.cudnn.allow_tf32 = False
```
acisseJZhong: wondering why disabling lower-precision accumulation/using higher precision accumulation can help batch inv?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+
+def disable_batch_invariant_mode() -> None:
+    """Unregister batch-invariant ATen overrides and restore settings."""
+    global _SAVED_STATE
+
+    # Unregister upstream ATen overrides
+    _upstream_disable()
+
+    # Restore saved settings
+    if _SAVED_STATE is not None:
```
acisseJZhong: do you also need to recover those os env variables?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+    # on batch size / tile decomposition.
+    torch.backends.cuda.matmul.allow_bf16_reduced_precision_reduction = False
+    torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction = False
+
+    # Disable TF32 for exact fp32 accumulation
+    torch.backends.cuda.matmul.allow_tf32 = False
+    torch.backends.cudnn.allow_tf32 = False
+
+    # Enable deterministic algorithms for run-to-run reproducibility.
+    torch.use_deterministic_algorithms(True)
```
acisseJZhong: do we need to set things like 
```
torch.manual_seed(seed)
torch.cudad.manual_seed(seed)
torch.backends.cudnn.deterministic = True 
torch.backends.cudnn.benchmark = False
```

torchtitan/models/common/attention.py
```diff
@@ -148,6 +148,21 @@ class VarlenAttentionWrapper(LocalMapAttention):
         varlen_attn, mode="max-autotune-no-cudagraphs"
     )
 
+    def __init__(self, *, batch_invariant: bool = False) -> None:
+        super().__init__()
+        self.batch_invariant = batch_invariant
+
+        # num_splits requires FA3. Activate it once when batch-invariant mode
+        # is requested so that varlen_attn can accept the num_splits kwarg.
+        if batch_invariant:
```
acisseJZhong: is it possible to do this inside enable_batch_invariant_mode? same below

torchtitan/models/common/attention.py
```diff
@@ -179,7 +194,16 @@ def forward(
         xk_packed = xk_packed.to(torch.bfloat16)
         xv_packed = xv_packed.to(torch.bfloat16)
 
-        return VarlenAttentionWrapper._compiled_varlen_attn(
+        # In batch-invariant mode, skip compilation and force num_splits=1
+        # to prevent non-deterministic split-k reductions in flash attention.
+        if self.batch_invariant:
+            attn_fn = varlen_attn
```
acisseJZhong: why we need to skip compile wrapper here for batch inv mode? could you add more comment?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -100,6 +111,23 @@ def __init__(
             world_size=world_size,
         )
 
+        # Seed RNG for reproducibility. Deterministic algorithms are already
```
acisseJZhong: is there a way to set this in batch_invariant.py?

torchtitan/experiments/rl/tests/test_bitwise_identity.py
wwwjn: This should be reverted, just for test. Working on rebasing, thanks

torchtitan/experiments/rl/batch_invariant.py
wwwjn: I updated this part, I think we should have a batch_invariant_mode states control ourself, as we are applying more changes than upstream (TML's batch_invariant_ops)

.github/workflows/integration_test_2gpu_h100.yaml
```diff
@@ -0,0 +1,120 @@
...
+        echo "Installed torch version: ${TORCH_VERSION}"
+        echo "torch==${TORCH_VERSION}" > /tmp/torch-constraint.txt
+        PIP_CONSTRAINT=/tmp/torch-constraint.txt pip install --no-build-isolation -v -e .
+        cd -
+
+        # Download Qwen3-0.6B checkpoint
+        python scripts/download_hf_assets.py \
+          --repo_id Qwen/Qwen3-0.6B \
+          --local_dir torchtitan/experiments/rl/example_checkpoint \
+          --all
```
tianyu-l: I don't think you need real weights to run batch invariant test.

.github/workflows/integration_test_2gpu_h100.yaml
tianyu-l, line 1: curious what makes it necessary to use h100? FAv3?

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,133 @@
...
+    # Save current state for restoration
+    _SAVED_STATE = {
+        "bf16": torch.backends.cuda.matmul.allow_bf16_reduced_precision_reduction,
+        "fp16": torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction,
+        "tf32_matmul": torch.backends.cuda.matmul.allow_tf32,
+        "tf32_cudnn": torch.backends.cudnn.allow_tf32,
+        "deterministic": torch.are_deterministic_algorithms_enabled(),
+    }
+
+    # Set NCCL env vars for deterministic inter-GPU collectives (all-reduce
```
tianyu-l: It would be great if you could comment on what each envvar configs.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -173,8 +172,11 @@ def __init__(
         os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
 
         if batch_invariant_mode:
-            os.environ["VLLM_BATCH_INVARIANT"] = "1"
-            init_batch_invariance(AttentionBackendEnum.CUSTOM)
```
tianyu-l, line 177: oh so we don't need this? I guess it's for flash backend of vllm.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         batch_invariant_mode: bool,
         hf_assets_path: str = "",
         transfer_dtype: str = "",
+        seed: int = 42,
```
tianyu-l: Instead of passing into constructor, we should just use https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/utils.py#L92

Both `enable_batch_invariance` and `seed` can just be fields in `DebugConfig` of RL trainer. https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L403

torchtitan/experiments/rl/batch_invariant.py
wwwjn: - Tree has O(log N) latency vs Ring's O(N) for the number of GPUs, so Tree algorithm if N GPUs are huge.
- RING based is deterministic because the communication pattern /  reduction order is deterministic: each GPU sends/receives to/from fixed neighbors in a ring.
- Tree based is not deterministic because NCCL might pick a different tree root across runs depending on topology/load

torchtitan/experiments/rl/batch_invariant.py
wwwjn: This is similar to NCCL_NVLS_ENABLE=0, `NCCL_COLLNET_ENABLE` controls whether NCCL can offload collective operations to network hardware accelerators (SmartNICs, in-network switches like NVIDIA Sharp/ConnectX)

torchtitan/experiments/rl/batch_invariant.py
wwwjn: It could be a devGPU issue I guess? I run the commend without this field on my devmachine and it works fine.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -100,6 +110,23 @@ def __init__(
...
+        # Activate FA3 so that varlen attention on the trainer uses the same
+        # FlashAttention 3 kernel as the generator's CUSTOM backend.  This is
+        # required for bitwise-identical logprobs between the two sides.
+        if batch_invariant_mode:
+            from torch.nn.attention import (
+                activate_flash_attention_impl,
+                current_flash_attention_impl,
+            )
+
+            if current_flash_attention_impl() != "FA3":
```
tianyu-l: we should error out in `set_determinism` / `post_init` of config that batch_invariant doesn't work with <sm90, instead of doing this ad hoc thing.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -210,6 +238,16 @@ def _build_model(
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
 
+        # Disable torch.compile on VarlenAttentionWrapper so the trainer calls
```
tianyu-l: According to @liangel-02, there seems little gain in compiling varlen anyway. Let's just disable varlen compile everywhere.

torchtitan/experiments/rl/models/attention.py
tianyu-l, line 101: if it can't be None, why allowing it to be None in signature -- I guess it's because of other vllm variants?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -164,8 +165,11 @@ def forward(
                 num_seqs + 1, dtype=torch.int32, device=query.device
             )
             cu_seqlens_k[1:] = torch.cumsum(seqused_k, dim=0)
+        # Force num_splits=1 in batch-invariant mode to prevent
+        # non-deterministic split-k reductions in flash attention.
+        num_splits = 1 if is_batch_invariant_mode_enabled() else 0
 
-        return torch.nn.attention.varlen.varlen_attn_out(
+        torch.nn.attention.varlen.varlen_attn_out(
             output[:num_actual_tokens],
```
tianyu-l, line 175: confused here, if `output` is passed in, I'm assuming we don't have to return anything because output is supposed to be mutated inplace. But here varlen is making a copy of it... and you are still returning the original output, what's going on?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -140,13 +144,33 @@ class Config(Configurable.Config):
     def __init__(self, config: Config):
         self.config = config
 
+        # Propagate seed to generator for deterministic sampling
+        config.generator.seed = config.seed
+
+        # Batch-invariant mode requires bf16 on both trainer and generator
+        # so that attention kernels (FA3) run identically on both sides
+        # without any dtype casting that could break bitwise identity.
+        if config.batch_invariant_mode:
```
tianyu-l: should do this in `post_init` of config, rather than constructor of actual trainer.

torchtitan/models/common/attention.py
```diff
@@ -154,6 +154,21 @@ class VarlenAttentionWrapper(LocalMapAttention):
...
+        # num_splits requires FA3. Activate it once when batch-invariant mode
+        # is requested so that varlen_attn can accept the num_splits kwarg.
+        if batch_invariant:
+            from torch.nn.attention import (
+                activate_flash_attention_impl,
+                current_flash_attention_impl,
+            )
+
+            if current_flash_attention_impl() != "FA3":
+                activate_flash_attention_impl("FA3")
```
tianyu-l, line 174: wait, what's relationship between https://github.com/pytorch/torchtitan/pull/2709 and this PR

Should main torchtitan also adapt to reduce the number of transpose?

torchtitan/experiments/rl/batch_invariant.py
wwwjn: Nice catch I think this is what's included in torchtitan set_deterministic(), I didn't set them because:
1. torch.use_deterministic_algorithms(True) as cover torch.backends.cudnn.deterministic = True  + torch.backends.cudnn.benchmark = False
2. Seeds are used to rule out randomness, but in our current implementation, we load weights from the state dict files, and there's should be no randomness in the model?

torchtitan/models/common/attention.py
wwwjn: I put it here because it's a little bit hard to pass the {num_split: 1} parameter from somewhere else to the `Attention` part, and this parameter is specific to Varlen attention

torchtitan/models/common/attention.py
wwwjn: Sure, will add more comments here. 

I actually didn't test if this max-auto compile will change numerics, but from what I asked offline, it might change numerics here, so I disabled it to keep rigorous about numerics.

torchtitan/experiments/rl/actors/trainer.py
wwwjn: See the explanation about seed above. But I think I should remove this change as "randomness" and "batch-invariant" is orthogonal to each other. 

And "RNG for reproducibility" is more for numerics parity when the model has randomness, say if the model is has some randomly dropout.

.github/workflows/integration_test_2gpu_h100.yaml
wwwjn: Nice catch, FAv2 should be able to be batch-invariant as well,

torchtitan/experiments/rl/actors/generator.py
wwwjn: We don't need this, and it's not only because of flash backend. Calling `vllm.init_batch_invariance` actually will let vllm to a bunch of changes:
- replace mm / rmsnorm/ log_softmax to batch-invariant, which is redundant to what we want to do
- set the attention kernel to be batch invariant (but from my test, it should be a no-op for `CUSTOM` backend, because vllm doesn't know how to make CUSTOM attention batch-invariant)

I want to let us explicitly make these changes, instead of depending on vllm to do so

torchtitan/experiments/rl/actors/trainer.py
wwwjn: Sure, let me move `enable_batch_invariance`. And also seed is not needed because we don't have randomness in the qwen3 model (it's needed if we don't load from a checkpoint) so I removed it.

torchtitan/models/common/attention.py
wwwjn: > Should main torchtitan also adapt to reduce the number of transpose?

We can adopt this in main torchtitan for Varlen backend attention

wwwjn: > wait, what's relationship between https://github.com/pytorch/torchtitan/pull/2709 and this PR

This PR and #2709 should be independent, and can be easily rebase onto each other

torchtitan/experiments/rl/batch_invariant.py
acisseJZhong: yeah I saw the point of having our enable batch_invariant_mode, but for this is_batch_inv, seems unnecssary? as you are just retuning _upstream_is_enabled() which is the alias of TML.

torchtitan/experiments/rl/models/attention.py
wwwjn: I think it was introduced in #2198 because the output will be mutated in place to store the output.

torchtitan/experiments/rl/models/attention.py
wwwjn: This is introduced in #2649, I think we don't need to return here

torchtitan/experiments/rl/batch_invariant.py
acisseJZhong: yeah it's devgpu specific. For my local devgpu, I have to prepend this flag. Chien-Chin also adds this for CI machine. It should has nothing to do with batch inv.

torchtitan/experiments/rl/batch_invariant.py
acisseJZhong: Seeds are used to rule out randomness. 

I am not sure if there will be other randomness other than loading ckpt controlled by this seed, safer to set it anyways?

torchtitan/models/common/attention.py
acisseJZhong: without compile this should be way slower, curious to see how much the speed degradation is, and if this actually changes numerics.

torchtitan/experiments/rl/models/attention.py
wwwjn: I checked vllm's FlashAttention, https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/flash_attn.py , it seems using `output[:slice]` and return the output as well. Let me dig more into it

torchtitan/experiments/rl/batch_invariant.py
wwwjn: I didn't realize there's randomness in transformer-based language model, but FLUX model does have randomness. 

Just thinking out loud here:
1. For non-randomness model, we should faithfully follow user's instruction - if user set the seed != None, we will set the seed. We should not pick a seed for user as we don't have to. 
2. For model with randomness, if the user don't set seed, we should set seed for user to ensure bit-wise identity.

But I don't know how to distinguish between 1. and 2, so I assumed the model has no-randomness as we don't support FLUX now...

torchtitan/models/common/attention.py
wwwjn: I think based on @liangel-02  it doesn't bring too much speed up when it is compiled, and it also has start-up overhead. I will just remove it entirely in torchtitan and keep consistent

torchtitan/experiments/rl/batch_invariant.py
wwwjn: Update: 
- We provide `batch-invariant-mode`, this option only enables trainer and generator model's output is batch size invariant.
- We also provide an option `seed` for user to disable randomness between run-to-run

If the user want to debug the RL loop and enable bit-wise parity, the user should decide if they should use `batch-invariant-mode` only, or they would need to fix `seed` as well. I'll add clear instruction in the README to show how to use these tools we provided.

torchtitan/experiments/rl/batch_invariant.py
wwwjn: You are right, accumulation precision should not affect the batch-invariant results. when TF32 is enable, fp32 inputs will be truncated to 19-bits. If both trainer and generator enabled / disabled TF32 at the same time, the bit-wise parity should still be hold

This is originally general reproducibility or reasoning, remove implicitly changes of numerics of truncating.

torchtitan/experiments/rl/batch_invariant.py
wwwjn: these env variables are not recoverable as they will be read by NCCL during `dist.init_process_group`. The only way to recover is to kick off another RL run

I think we should remove the disable_batch_invariant function as there's no real use case for this

torchtitan/experiments/rl/actors/trainer.py
wwwjn: batch_invariant should be able to work if FA2 achieves bit-wise identity on Blackwell. Checking and experimenting, will update later

torchtitan/experiments/rl/batch_invariant.py
acisseJZhong: got it, so we choose to disable tf32 explicitly on both sides?

torchtitan/experiments/rl/models/attention.py
wwwjn: > here varlen is making a copy of it

Do you mean the slice  at line 173 will make a copy of `output`?  If so, I'm not sure if it's creating a copy underlying or create a view of `output` and sharing the underlaying storage

> you are still returning the original output
I agree we can remove the return. But here I think we are following the parent class -- vllm's FlashAttention class API, which returns the output (for no reason?)

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -145,6 +143,9 @@ class Config(Configurable.Config):
         seed: int | None = None
         """Random seed for vLLM engine and sampling. None for non-deterministic."""
 
+        batch_invariant_mode: bool = False
+        """Enable batch-invariant mode for deterministic ops."""
```
tianyu-l: any reason we don't want to use DebugConfig or having a variant of it for generator?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L403

E.g. how do I run determinism without batch invariant?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -92,6 +100,15 @@ def __init__(
         self.device = torch.device(f"{device_type}:{int(os.environ['LOCAL_RANK'])}")
         device_module.set_device(self.device)
 
+        # Enable batch-invariant mode BEFORE init_distributed so NCCL env
+        # vars are set before the first communicator is created.
+        if config.batch_invariant_mode:
+            from torchtitan.experiments.rl.batch_invariant import (
+                enable_batch_invariant_mode,
+            )
```
tianyu-l: I think it's cleaner to do `from batch_invariant_ops import enable_batch_invariant_mode as _upstream_enable` inside `enable_batch_invariant_mode()` so that you don't need to do this optional import -- you can import this at file head.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -107,6 +124,25 @@ def __init__(
...
+        dist_utils.set_determinism(
+            self.parallel_dims,
+            self.device,
+            config.debug,
+            distinct_seed_mesh_dims=["pp"],
+        )
+
+        # Activate FA3 when using Hopper machine so that varlen attention
+        # on the trainer uses the same FAv3 kernel as the generator's CUSTOM backend.
+        if config.batch_invariant_mode and has_cuda_capability(9, 0):
```
tianyu-l: The decision of activating FA3 shouldn't depend on batch_invariant_mode.

You are activating it anyway in common/attention.py. No need to do it here.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -179,8 +183,11 @@ def forward(
             window_size=sliding_window_size,
             block_table=block_table,
             seqused_k=seqused_k,
+            num_splits=num_splits,
```
tianyu-l: Is `num_splits=0` the same as not setting it?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -179,8 +183,11 @@ def forward(
...
         )
 
+        return output
```
tianyu-l: remove?

torchtitan/models/common/attention.py
```diff
@@ -154,7 +154,26 @@ def forward(
 class VarlenAttention(LocalMapInnerAttention):
     @dataclass(kw_only=True, slots=True)
     class Config(LocalMapInnerAttention.Config):
-        pass
+        batch_invariant: bool = False
```
tianyu-l: I don't think you are setting it anywhere, are you?

torchtitan/models/common/attention.py
tianyu-l, line 219: where are you controlling `num_splits=1`?

torchtitan/models/common/attention.py
```diff
@@ -154,7 +154,26 @@ def forward(
...
+        """Enable batch-invariant mode for deterministic attention."""
+
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
+        self.batch_invariant = config.batch_invariant
+
+        # num_splits requires FA3 on Hopper (SM 9.0+). Activate it once when
+        # batch-invariant mode is requested so varlen_attn accepts num_splits.
+        if config.batch_invariant:
```
tianyu-l: As mentioned, we can always turn on FA3 when it's Hopper. Any reason we don't?

torchtitan/experiments/rl/batch_invariant.py
tianyu-l, line 1: I now think it's cleaner if we put this in core, and enhance `DebugConfig` with `enable_batch_invariance` default to `False`. This way the common/attention.py change will be clean.

Of course, TML repo is still optional and we only mention it in RL README.

Happy to discuss.

torchtitan/experiments/rl/tests/test_batch_invariant.py
tianyu-l, line 1: are we guarding on this anywhere?

torchtitan/experiments/rl/models/attention.py
tianyu-l, line 69: It might be more consistent between trainer and generator if we move this to inner attention.

tianyu-l: we don't have to introduce this, we can just use `if current_flash_attention_impl() != "FA3"` if the overhead is low.

tianyu-l: can we `warn_once`? It is warning every layer?

torchtitan/experiments/rl/batch_invariant.py
acisseJZhong: > torch.use_deterministic_algorithms(True) as cover torch.backends.cudnn.deterministic = True + torch.backends.cudnn.benchmark = False

actually I think torch.use_deterministic_algorithms(True) doesn't cover the later two. I was also looking at the doc here https://docs.pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms

titan also sets them all https://github.com/pytorch/torchtitan/blob/2618196882fafa4791816c08560c6da69a021d27/torchtitan/distributed/utils.py#L116

torchtitan/experiments/rl/config_registry.py
```diff
@@ -25,33 +25,34 @@
...
                 decay_type="linear",
             ),
-            training=TrainingConfig(),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
             compile=CompileConfig(enable=True, backend="aot_eager"),
         ),
         generator=VLLMGenerator.Config(
-            model_dtype="bfloat16",
```
tianyu-l, line 48: why deleting, is it because they are default?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -137,6 +145,29 @@ class Config(Configurable.Config):
...
+            # Propagate debug settings to sub-configs
+            self.generator.seed = self.debug.seed
+            self.trainer.batch_invariant_mode = self.debug.batch_invariant_mode
+            self.generator.batch_invariant_mode = self.debug.batch_invariant_mode
+
+            # Batch-invariant mode requires bf16 on both trainer and generator
+            # so that attention kernels (FA3) run identically on both sides
+            # without any dtype casting that could break bitwise identity.
+            if self.debug.batch_invariant_mode:
+                if self.trainer.training.dtype != "bfloat16":
```
tianyu-l, line 142: This sounds tricky.

`training.dtype` is master weight, which should seldom be set to anything not fp32.

In FSDP, `mixed_precision_param` is usually bf16, meaning computation is done in this precision.

I think here you are requiring `training.dtype` because
- batch invariance require the same parallelism
- generator usually runs TP only
- trainer TP doesn't handle mixed precision so not using `mixed_precision_param`, thus you have to change master weight dtype

I think this is fine for now. In the future we will support TP + mixed precision, with FlexShard.

But please create a dedicated config for batch invariance, with training dtype bf16 (I didn't see one).

torchtitan/experiments/rl/tests/test_batch_invariant.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+
+
+def _build_debug_model(device="cuda"):
+    """Build a Qwen3 debugmodel with random weights on the given device."""
+    from torchtitan.models.qwen3 import model_registry
+
+    model_spec = model_registry("debugmodel", attn_backend_override="varlen")
+    model_spec.model.layer.attention.inner_attention.batch_invariant = True
+    with set_default_dtype(torch.bfloat16):
```
tianyu-l: these are too hacky

torchtitan/experiments/rl/config_registry.py
```diff
@@ -25,33 +25,34 @@
 
 
 def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
-    """GRPO training config for Qwen3-0.6B (6 GPUs: 4 gen + 2 train)."""
+    """GRPO training config for Qwen3-0.6B (4 GPUs: 2 gen + 2 train)."""
```
tianyu-l: Create another model/job config for batch invariance testing.
Don't change it here because it's hurting weight sync test.

torchtitan/experiments/rl/tests/test_batch_invariant.py
wwwjn: This is not directly included in our CI now, as we don't do batching for trainer, so the batch-size is always 1 now.  I added this to test these kernels are batch-invariant. I will add this test to CI now, once we have batcher, this part can be covered by e2e test.

torchtitan/models/common/attention.py
wwwjn: In core torchtitan, we will need ask user to install flash-attention-v3 if it's the default backend for varlen. Let me enable it in core torchtitan and add necessary instructions

torchtitan/experiments/rl/simple_grpo_sum_digits.py
wwwjn: You are right, let me create a separate config for bit-wise identity test
Also I'm digging into fp32 weights + fp32 activation (except fp16 activation around inner_attn)

torchtitan/experiments/rl/actors/generator.py
wwwjn: I refactor the config system again, here's the new design:

- RLTrainer.debug is set from config_registry. When `batch_invariant_mode=True`, we check `deterministic` must be true, and dtype on both trainer and generator needs to be `bfloat16` in `__post__init`
    - PolicyTrainer.debug is set by RLTrainer.debug
        - PolicyTrainer model's inner_attn (VarlenAttention) has a `batch_invariant` flag, this flag is set by PolicyTrainer
    - Generator.debug is set by RLTrainer.debug
        - Generator model's inner_attn (VLLMAttentionWrapper) enables `batch_invariant` mode by checking `is_batch_invariant_mode_enabled()` global variable.

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: This seems not resolved.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -172,9 +169,25 @@ def __init__(
...
+        if config.debug.deterministic:
+            torch.use_deterministic_algorithms(
+                True, warn_only=config.debug.deterministic_warn_only
+            )
+            torch.backends.cudnn.deterministic = True
+            torch.backends.cudnn.benchmark = False
+            os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
+
+        if config.debug.seed is not None:
+            torch.manual_seed(config.debug.seed)
```
tianyu-l: can we directly use `set_determinism`?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -107,6 +122,25 @@ def __init__(
...
 
+        # Set determinism flags and seed via core torchtitan utility
+        dist_utils.set_determinism(
+            self.parallel_dims,
+            self.device,
+            config.debug,
+            distinct_seed_mesh_dims=["pp"],
+        )
+
+        # Activate FA3 when using Hopper machine so that varlen attention
```
tianyu-l: should enable this in model code, not here

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -107,6 +122,25 @@ def __init__(
...
+        # on the trainer uses the same FAv3 kernel as the generator's CUSTOM backend.
+        if config.debug.batch_invariant_mode and has_cuda_capability(9, 0):
```
tianyu-l: don't couple fa3 and batch invariance together, unless you have clear reason

torchtitan/experiments/rl/models/attention.py
```diff
@@ -73,6 +74,12 @@ def __init__(self, *args, **kwargs) -> None:
             )
             self._use_fa3 = False
 
+        from torchtitan.experiments.rl.batch_invariant import (
+            is_batch_invariant_mode_enabled,
+        )
+
+        self._batch_invariant = is_batch_invariant_mode_enabled()
```
tianyu-l: why need this variable?

torchtitan/experiments/rl/README.md
```diff
@@ -74,6 +79,32 @@ python torchtitan/experiments/rl/simple_grpo_sum_digits.py --module rl --config
...
+### Batch-invariant mode
+
+Batch-invariant mode guarantees that a model's output for a given input is **identical regardless of what other inputs are in the batch**. This is critical for RL training because the generator computes log-probs in one batch composition (e.g. 8 completions), while the trainer recomputes them in a different batch composition (e.g. 2 completions after DP sharding). Without batch-invariant mode, the same input can produce different log-probs in different batch contexts due to floating-point accumulation order differences.
+
+Enable it in your config:
+```python
+from torchtitan.config.configs import DebugConfig
+
+RLTrainer.Config(
+    debug=DebugConfig(batch_invariant_mode=True),
```
tianyu-l: need to update

torchtitan/models/common/attention.py
```diff
@@ -154,7 +154,25 @@ def forward(
 class VarlenAttention(LocalMapInnerAttention):
     @dataclass(kw_only=True, slots=True)
     class Config(LocalMapInnerAttention.Config):
-        pass
+        batch_invariant: bool = False
```
tianyu-l: how did you feel about https://github.com/pytorch/torchtitan/pull/2683#discussion_r3019581341

torchtitan/experiments/rl/models/attention.py
wwwjn: I double checked here: https://github.com/pytorch/pytorch/pull/178405/changes#diff-0d85d9d89fb11d0cac43164b218d39ca711c7ace58c0c9964f2a965784a2fe77R609 it should be either 1 or none

torchtitan/experiments/rl/actors/trainer.py
wwwjn: Aha sorry I still have some local changes and  haven't finish addressing the comments, let me update this

torchtitan/experiments/rl/batch_invariant.py
wwwjn: I agree on the plan after I tried several other solutions. I tried adding a new RLDebugConfig inheriting from DebugConfig, however this is adding very little value than just changing the core DebugConfig.

torchtitan/experiments/rl/actors/generator.py
wwwjn: I was considering this but set_determinism but it requires parallel_dim not None. I'm trying to make parallel_dim optional

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -107,6 +119,26 @@ def __init__(
...
+            self.parallel_dims,
+            self.device,
+            config.debug,
+            distinct_seed_mesh_dims=["pp"],
+        )
+
+        # Activate FA3 on Hopper so varlen attention on the trainer uses
+        # the same FAv3 kernel as the generator's CUSTOM backend.
+        # Blackwell (SM 10.0+) and older hardware use FA2.
+        if has_cuda_capability(9, 0) and not has_cuda_capability(10, 0):
```
wwwjn: Will remove these to VarlenAttention (inner_attention)

torchtitan/experiments/rl/models/attention.py
wwwjn: This is needed because we need to set `num_splits=1` in foward(). Since vllm's API doesn't let us passing extra config into ``__init__`` function, so I used the global variable within `is_batch_invariant_mode_enabled ` to decided if batch_invariant is enabled.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -215,6 +241,9 @@ def _build_model(
             model_spec.model.layer.attention.inner_attention, VarlenAttention.Config
         ), "Only varlen attention backend is allowed."
 
+        if config.debug.batch_invariant_mode:
+            model_spec.model.layer.attention.inner_attention.batch_invariant = True
```
wwwjn: This is hacky because we only support attn_backend_override to a string. Once we have more flexible override, we can do following:
```
model_spec = model_registry("0.6B", model_override={"layer.attention.inner_attention": VarlenAttention.Config(batch_invariant=True)})
```

torchtitan/models/common/attention.py
wwwjn: I converged to use the method discribed in [#2683 (comment)](https://github.com/pytorch/torchtitan/pull/2683/changes#diff-ba50f95793480d11378e4a2a155b068a5c938553dc3b7b6a870d7e54793dd694R89-R90:~:text=you%20feel%20about-,%232683%20(comment),-Add%20diff%20comment)

torchtitan/experiments/rl/models/attention.py
wwwjn: I thought about this twice, I think activating FA3 is tied to "FlashAttention", even for symmetric, the `inner_attn` for generator model should be VLLMAttentionWrapper.

In the future, we probably need use FlexAttention backend within `VLLMAttentionWrapper`, which no need to call the function to activate FA3

torchtitan/experiments/rl/models/attention.py
tianyu-l: You can just call `is_batch_invariant_mode_enabled()`

torchtitan/config/configs.py
```diff
@@ -417,6 +417,10 @@ class DebugConfig:
     """Enable torch.autograd anomaly detection to help track down NaN/Inf gradients.
     Note: incurs significant overhead; for debugging only."""
 
+    batch_invariant_mode: bool = False
```
tianyu-l: since we only have two mode on vs. off, I'd suggest
```suggestion
   batch_invariant: bool = False
```
to be consistent with e.g. `deterministic`

torchtitan/experiments/rl/actors/generator.py
tianyu-l: That sounds a bit aggressive, maybe for now just put these into a standalone `self._set_determinism(...)`.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -92,6 +98,11 @@ def __init__(
         self.device = torch.device(f"{device_type}:{int(os.environ['LOCAL_RANK'])}")
         device_module.set_device(self.device)
 
+        # Enable batch-invariant mode BEFORE init_distributed so NCCL env
+        # vars are set before the first communicator is created.
+        if config.debug.batch_invariant_mode:
+            enable_batch_invariant_mode()
```
tianyu-l: we can put this in `set_determinism`

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: I don't think we need this if we call `is_in_batch_invariant_mode()` inside model

torchtitan/experiments/rl/models/attention.py
tianyu-l: makes sense

torchtitan/experiments/rl/batch_invariant.py
```diff
@@ -0,0 +1,101 @@
...
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Enable / disable
+# ---------------------------------------------------------------------------
+
+_enabled: bool = False
+
+
+def is_batch_invariant_mode_enabled() -> bool:
```
tianyu-l: nit
```suggestion
def is_in_batch_invariant_mode() -> bool:
```

torchtitan/experiments/rl/batch_invariant.py
tianyu-l, line 1: should put this to core I think

torchtitan/experiments/rl/config_registry.py
```diff
@@ -141,3 +143,50 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
             ),
         ),
     )
+
+
+def rl_grpo_qwen3_0_6b_on_policy() -> RLTrainer.Config:
```
tianyu-l: put batch invariant in name, o/w ambiguous

torchtitan/experiments/rl/simple_grpo_sum_digits.py
tianyu-l: This is worth more discussion.

The only requirement would be
- parallelism other than DP (ddp, fsdp, hsdp) match degree (TP, CP, PP, etc.)
- computation dtype match (so if FSDP is used, we only need to make sure `training.mixed_precision_param` matches generator dtype)

torchtitan/models/common/attention.py
```diff
@@ -211,6 +221,25 @@ def forward(
             #   - (W, 0): Sliding window causal - attend to at most W previous tokens.
             window_size=(-1, 0),
         )
+        if self.batch_invariant:
+            from torch.nn.attention import current_flash_attention_impl
```
tianyu-l: this should have been imported in init?

torchtitan/models/common/attention.py
tianyu-l: I think we can remove this and use `is_in_batch_invariant_mode()`

torchtitan/models/common/attention.py
```diff
@@ -190,14 +207,7 @@ def forward(
         xk_packed = xk_packed.to(torch.bfloat16)
         xv_packed = xv_packed.to(torch.bfloat16)
 
-        out_packed = varlen_attn(
```
tianyu-l, line 193: don't need this change?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -24,6 +26,8 @@
 )
 from vllm.v1.attention.backends.registry import AttentionBackendEnum, register_backend
 
+logger = logging.getLogger(__name__)
+
 
 @register_backend(AttentionBackendEnum.CUSTOM)
 class PyTorchFlashAttentionBackend(FlashAttentionBackend):
```
tianyu-l: this name is not used anywhere, so I assume we can rename to whatever?

If so we should give a more precision name, e.g.
`PyTorchVarlenAttentionBackend` or even `PyTorchVarlenAttentionInferenceBackend`

torchtitan/experiments/rl/models/attention.py
```diff
@@ -59,19 +63,15 @@ class PyTorchFlashAttentionImpl(FlashAttentionImpl):
...
-        # FA3 requires SM 9.0+ (e.g. H100); check capability explicitly because
-        # activate_flash_attention_impl("FA3") succeeds even on SM80.
-        # Fall back to FA2 which requires page_size to be a multiple of 256.
         if has_cuda_capability(9, 0):
             if current_flash_attention_impl() != "FA3":
                 activate_flash_attention_impl("FA3")
-            self._use_fa3 = True
         else:
-            logger.warning(
-                "FA3 not available (requires SM 9.0+), falling back to FA2. "
```
tianyu-l, line 71: why removing the fallback info? If so, at least should also log in the FA3 branch.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -59,19 +63,15 @@ class PyTorchFlashAttentionImpl(FlashAttentionImpl):
...
-                "vLLM block_size must be set to 256 for FA2 paged attention."
+            warn_once(
+                logger,
+                "Using FA2 attention. "
+                "vLLM block_size must be set to 256 for FA2 paged attention.",
```
tianyu-l: remove, user probably don't care

torchtitan/experiments/rl/models/attention.py
tianyu-l: rename to `PyTorchVarlenAttentionImpl`

torchtitan/experiments/rl/models/attention.py
wwwjn: I renamed it to `PyTorchVarlenAttentionBackend`

torchtitan/models/common/attention.py
wwwjn: I was wondering why we need `batch_invariant.py` to be put in core, then I realize here in VarlenAttention we'll need test if it's in batch-invariant mode.

Is above one of the reason to put batch_invariant in core? Or there are more use cases for batch-invariant than in RL.

torchtitan/experiments/rl/batch_invariant.py
wwwjn: Yes, I disable the tf32 on both side

torchtitan/models/common/attention.py
tianyu-l: It's just inconsistent
- we are using this model under batch-invariant model for RL
- general batch-invariant mode requires setting and reading global states
- the reason we can workaround without checking global mode is because the attention part only depends on FA3, but that's not consistent with how you detect batch invariance in general

Setting batch-invariant flag from somewhere outside sounds more error-prone, in the sense that in addition to set the `DebugConfig`, users have to set one more config in trainer model.

Also batch invariance is a property of model / kernels, itself not tied to RL. Just more useful in RL.

I guess your concern is about optional dependency on TML repo in core? E.g. if this mode is in pytorch, would you still think it doesn't make sense to put in `set_determinism()`?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
wwwjn: Both are correct

For 1, my understanding is we already made NCCL reduction order deterministic, this would apply to all communications for parallelism. With same parallelism degree, the communication would not introduce numerics discrepancy between trainer and generator.  In terms of computation on each GPU, TP would affect matmul input size , CP would change sequence length , PP should not affect numerics. 
- If we want to achieve bit-wise identity under different TP degree: We need to make matmul and norm not-sensitive to input dimension 
- If we want to achieve bit-wise identity under different CP degree: We need to make computation ops not sensitive to sequence-length dimension

For 2. As we are comparing log-probability bit-wise parity, we only care about dtype that affects activation's numerics. The master weights are not important in our case, we just need to cast the weights to match generator's dtype before they are used in computation

torchtitan/experiments/rl/batch_invariant.py
wwwjn: Thanks, I added all of them in both trainer and generator

torchtitan/models/common/attention.py
wwwjn: > Also batch invariance is a property of model / kernels, itself not tied to RL. Just more useful in RL

That's correct, I’m also seeing other framework make batch-invarient a field of model, eg `use_batch_invariant_ops`

I'm not objective about moving, or concerning about the dependency, but more curious about other potential use case. The pros of adding this to core is we might need to add more tests and understand what module in torchtitan is batch-invariant and others is not? eg, batch-invariant would not be enabled for FlexAttention now. Do we need to understand if batch-invariant works for each model?

tianyu-l: That's reasonable concern. I think we can error out in `post_init` of main `Trainer.Config` that batch invariant for pretraining / sft is not supported.

titan RL is already a strong enough use case. Potentially, other RL integration with titan (e.g. titan-verl integration) can also use this, considering that they may not use our rl folder.

wwwjn: Sounds good, let me move this

torchtitan/distributed/utils.py
```diff
@@ -207,6 +207,70 @@ def set_determinism(
...
 
+_batch_invariant_enabled: bool = False
+
+
+def is_in_batch_invariant_mode() -> bool:
+    """Return whether batch-invariant mode is active."""
+    return _batch_invariant_enabled
+
+
+def set_batch_invariant() -> None:
```
tianyu-l:
```suggestion
def set_batch_invariance() -> None:
```
also please pass `config.debug.batch_invariant` in, so that we don't have to do if-else outside.

It's also more consistent with `set_determinism()`

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -336,7 +350,8 @@ async def step(self, episodes: list[Episode]) -> dict:
 
         # All-reduce gradients across DP ranks so all ranks have consistent
         # weight updates despite processing different data shards.
-        if self.dp_enabled:
+        # Skip when FSDP is enabled — FSDP already reduces gradients.
+        if self.dp_enabled and not self.parallel_dims.fsdp_enabled:
             for param in self.model.parameters():
                 if param.grad is not None:
                     dist.all_reduce(param.grad, op=dist.ReduceOp.AVG)
```
tianyu-l: can we just remove this codeblock? 

`dist_utils.clip_grad_norm_` should handle DDP/FSDP/HSDP/CP/TP/PP all cases already

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -10,8 +10,6 @@
 from torchtitan.models.common.attention import VarlenMetadata
 
 
-# TODO We should either unify all the mask creation for RL, or move them to a
```
tianyu-l, line 13: why removing this TODO?

torchtitan/experiments/rl/tests/test_batch_invariant.py
```diff
@@ -0,0 +1,218 @@
...
+        ), "Only varlen attention backend is allowed."
+
+    with torch.device("meta"):
+        with set_default_dtype(TORCH_DTYPE_MAP[dtype]):
+            model = model_spec.model.build()
+
+    model.to_empty(device=device)
+    torch.manual_seed(42)
+    with torch.no_grad():
+        model.init_weights(buffer_device=None)
```
tianyu-l: This is the same as
- build model directly on "device", without doing meta-init -> to_empty -> init_weights

torchtitan/experiments/rl/tests/test_bitwise_parity.py
tianyu-l, line 1: Is this test for the case where `batch_invariant = False`?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -141,3 +143,50 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
...
+            deterministic=True,
+        ),
+        trainer=PolicyTrainer.Config(
+            optimizer=OptimizersContainer.Config(lr=2e-6),
+            lr_scheduler=LRSchedulersContainer.Config(
+                warmup_steps=2,
+                decay_type="linear",
+            ),
+            # bfloat16 is needed for trainer to align with generator dtype
+            training=TrainingConfig(dtype="bfloat16"),
```
tianyu-l, line 167: add a todo to replace this with FSDP2+TP2

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -137,6 +137,28 @@ class Config(Configurable.Config):
...
+            self.trainer.debug = self.debug
+            self.generator.debug = self.debug
+
+            # Batch-invariant mode requires bf16 on both trainer and generator
+            # so that attention kernels (FA3) run identically on both sides
+            # without any dtype casting that could break bitwise identity.
+            if self.debug.batch_invariant:
+                if not self.debug.deterministic:
+                    raise ValueError("batch_invariant requires deterministic=True")
+                if self.trainer.training.dtype != "bfloat16":
```
tianyu-l, line 142: add a TODO to replace / supplement this constraint with `mixed_precision_param` under FSDP.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -137,6 +137,28 @@ class Config(Configurable.Config):
         generator: VLLMGenerator.Config = field(default_factory=VLLMGenerator.Config)
         """VLLMGenerator actor configuration (vLLM engine, sampling)."""
 
+        def __post_init__(self):
+            # Propagate debug settings to sub-configs
+            self.trainer.debug = self.debug
+            self.generator.debug = self.debug
```
tianyu-l: let's
- reserve post_init for validation purpose only for now
- don't define debug for RLTrainer; instead, set trainer generator debug config separately. (we could have utils in config_registry.py to make sure they share the same source)

torchtitan/trainer.py
```diff
@@ -229,6 +229,8 @@ def __init__(self, config: Config):
             config.debug,
             distinct_seed_mesh_dims=["pp"],
         )
+        if config.debug.batch_invariant:
```
tianyu-l: this can happen in post_init of `Trainer.Config`

torchtitan/experiments/rl/tests/test_bitwise_parity.py
wwwjn: No, this test is testing against `rl_grpo_qwen3_0_6b_batch_invariant`, so batch_invariant is always True. 

This is a more rigorous test requiring following 3 tests all pass:
 
Test 1: Trainer prefill == vLLM prefill (prompt-only)
Test 2: vLLM decode == vLLM 2nd-pass prefill (generated positions)
Test 3: Trainer full-seq forward == vLLM 2nd-pass prefill

tianyu-l (approved):
The code looks great. The tests I still have trouble internalizing. But I won't block landing this first and maybe revisit tests later, especially because FSDP integration needs its own PR.

torchtitan/experiments/rl/tests/test_bitwise_parity.py
wwwjn: Synced offline: the full test set should be 
Test1: Trainer model prefill (bsz = m) == Trainer model prefill (bsz =n , m !=n). This test is to guard the model kernels are batch-invariant.
Test 2: Trainer prefill == vLLM prefill (prompt-only). This test is to ensure trainer model forward and generator model forward has bit-wise parity, given we have slightly different wrapper around attention module on each sides.
Test 3: vLLM decode == vLLM 2nd-pass prefill (generated positions). This test is to ensure prefill stage v.s. decode stage has bit-wise parity, make sure direct attention calculation v.s. kv-cache does bring numerics discrepency.

wwwjn:
> The code looks great. The tests I still have trouble internalizing. But I won't block landing this first and maybe revisit tests later, especially because FSDP integration needs its own PR.

I updated the test case to align with our offline discussion, and we can refactor test details in next PR

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -107,6 +116,14 @@ def __init__(
             world_size=world_size,
         )
 
+        # Set determinism flags and seed via core torchtitan utility
+        dist_utils.set_determinism(
+            self.parallel_dims,
+            self.device,
+            config.debug,
+            distinct_seed_mesh_dims=["pp"],
```
acisseJZhong, line 113: why we want to have distinct seed on different pp mesh?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -334,13 +346,6 @@ async def step(self, episodes: list[Episode]) -> dict:
         self.optimizers.zero_grad()
         loss.backward()
 
-        # All-reduce gradients across DP ranks so all ranks have consistent
```
acisseJZhong, line 326: why we could remove this?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -157,14 +156,20 @@ def forward(
...
+        if current_flash_attention_impl() == "FA3":
             cu_seqlens_k = None
         else:
             num_seqs = seqused_k.shape[0]
             cu_seqlens_k = torch.zeros(
                 num_seqs + 1, dtype=torch.int32, device=query.device
             )
             cu_seqlens_k[1:] = torch.cumsum(seqused_k, dim=0)
+        # FA3 + batch-invariant: fix num_splits=1 to prevent non-deterministic
+        # split-k reductions. FA2 is automatically batch-invariant and does
```
acisseJZhong, line 168: curious where could I learn more about this non-deterministic split-k reduction? Is there any other way to make it deterministic other than setting num_split to 1? What does num_split mean?

wwwjn: setting nums_split = 1 disables batch-invariance in FA attention, I don't think there's other source of non-batch-invariances. Here's a good explanation: https://thinkingmachines.ai/blog/defeating-nondeterminism-in-llm-inference/#batch-invariant-attention

PR #2684, opened 2026-03-24
Use return_aux in FlexAttention to avoid graph breaking warnings.warn

chelsea0x3b:
Using return_lse currently triggers a warning that breaks graph compilation even if you add warnings.warn to exemptions. The warning is triggered here: https://github.com/pytorch/pytorch/blob/v2.10.0/torch/nn/attention/flex_attention.py#L1559

This PR just swaps to using the newer return_aux, with the addition of pulling out the `aux.lse` so the outputs are easily addressable with TP. This unblocks torch.compile with gpt_oss.

Tested on 8xh100 with gpt_oss 20b

### DIFF
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -152,7 +152,7 @@ class FlexAttentionWrapper(torch.nn.Module):
         # 2. `self._compiled_flex_attn` is not correct, `self` will be passed in
         #    as the first argument, which will cause an error.
         #    `FlexAttentionWrapper._compiled_flex_attn` is correct.
-        out = FlexAttentionWrapper._compiled_flex_attn(
+        out, aux = FlexAttentionWrapper._compiled_flex_attn(
             q,
             k,
             v,
@@ -164,7 +164,7 @@ class FlexAttentionWrapper(torch.nn.Module):
         # Note: return a tuple of Tensor to make converting `lse`
         # to DTensor easier with TP module notation.
         if return_lse:
-            return out[0], out[1].lse
+            return out, aux.lse
         return out
 
 


torchtitan/models/common/attention.py
```diff
@@ -247,43 +152,21 @@ def forward(
...
             q,
             k,
             v,
             block_mask=block_mask,
             scale=scale,
             enable_gqa=enable_gqa,
-            return_lse=return_lse,
+            return_aux=AuxRequest(lse=return_lse),
         )
+        if return_lse:
```
acisseJZhong, line 166: before this line, please add comment like 
```
Used `return_lse` here to return a tuple of Tensor for easier TP module notation to convert `lse` to be DTensor.
```

chelsea0x3b: Added!

torchtitan/models/common/attention.py
```diff
@@ -247,43 +152,23 @@ def forward(
...
             scale=scale,
             enable_gqa=enable_gqa,
-            return_lse=return_lse,
+            return_aux=AuxRequest(lse=return_lse),
         )
+        # Note: return a tuple of Tensor to make converting `lse`
+        # to DTensor easier with TP module notation.
+        if return_lse:
+            return out[0], out[1].lse
+        return out
```
acisseJZhong, line 168: did you verify this path returns the same as before and it works? 

from https://docs.pytorch.org/docs/stable/nn.attention.flex_attention.html, I saw 

Image

the return type seems different.

chelsea0x3b: Oh good catch you are correct! I just verified in the pytorch code too, will update this to be `out[0]`

chelsea0x3b: Fixed - unpacked the tuple above instead since I think thats' cleaner

acisseJZhong:
LGTM, feel free to merge after the CI turns green.

PR #2687, opened 2026-03-24
[Perf] Add model compile support to gpt oss

chelsea0x3b:
Blocked by #2684 

After the above 2 PRs are merged full graph compilation is supported for gpt_oss!

Results:

I ran this code on 8xH100 for gpt_oss_20b using this command for both runs:

```
torchrun --nproc_per_node=8 \
    -m torchtitan.train \
    --module gpt_oss \
    --config gpt_oss_20b \
    --metrics.log_freq 1 \
    --training.steps 10 \
    --compile.enable
```
[compile-gpt-oss.log](https://github.com/user-attachments/files/26222535/compile-gpt-oss.log)

| | main branch | #2686 | this pr + #2684 | this pr + #2686 + #2684 |
| --- | --- | --- | --- | --- |
| MFU | 16.50% | 18.68% | 17.10% |  19.49% |

### DIFF
--- a/torchtitan/models/gpt_oss/parallelize.py
+++ b/torchtitan/models/gpt_oss/parallelize.py
@@ -238,14 +238,6 @@ def apply_non_moe_tp(
             "attention.wq": ColwiseParallel(use_local_output=False),
             "attention.wk": ColwiseParallel(use_local_output=False),
             "attention.wv": ColwiseParallel(use_local_output=False),
-            "attention.inner_attention": PrepareModuleInputOutput(
-                input_layouts=(Shard(1), Shard(1), Shard(1)),
-                desired_input_layouts=(Shard(1), Shard(1), Shard(1)),
-                use_local_input=True,
-                output_layouts=(Shard(1), Shard(1)),
-                desired_output_layouts=(Shard(1), Shard(1)),
-                use_local_output=False,
-            ),
             "attention.wo": rowwise_output_plan,
             "ffn_norm": norm_plan,
         }


tianyu-l:
what's the perf if enable compile + no triton kernels?

chelsea0x3b:
@tianyu-l I added more details to the PR description to include that. Basically this PR + #2684 MFU is at 17.10%. So the cuda kernels in #2686 still add about 2% mfu

Here's the log for that run 
[pr_and_2684.log](https://github.com/user-attachments/files/26245628/pr_and_2684.log)

tianyu-l:
@chelsea0x3b it seems CI failed, does it have to rely on https://github.com/pytorch/torchtitan/pull/2686?

How to land kernel work is still something we are trying to figure out, especially considering hardware heterogenity...

chelsea0x3b:
I will look into it today. My cli tests didn't depend on the kernels, but let me check what the CI does.

chelsea0x3b:
@tianyu-l So the issue is fullgraph=True in `apply_compile_sparse` doesn't work with flex attention when tensor parallelism is enabled (ie DTensors are involved). Note that this is currently an issue with llama4 as well.

Options:
1. fullgraph = False
2. fullgraph = not tp

Any thoughts here?

More details: without TP, q/k/v are plain tensors and Dynamo inlines _compiled_flex_attn. With TP, q/k/v are DTensors that go through LocalMapAttention's local_map path, inside which the nested _compiled_flex_attn call causes an untraceable graph break under fullgraph=True.

tianyu-l:
@acisseJZhong could you take a look at the issue?

tianyu-l:
@chelsea0x3b the issue might have been worked around in https://github.com/pytorch/torchtitan/pull/2723. Could you rebase and try again? Thanks!

chelsea0x3b:
@tianyu-l this doesn't seem to fix the error for me

chelsea0x3b:
@tianyu-l @acisseJZhong So it seems like the `PrepareModuleInputOutput` that I just removed was causing the issue. I don't think its still needed so I removed it. The CI tests pass locally for me now.

The exact error was from this section in `local_map()`:
```python
        for out, spec in zip(flat_out, out_placements_tuple):
            if isinstance(out, torch.Tensor):
                if isinstance(out, DTensor):
                    raise AssertionError(
                        f"torch.Tensor output expected but received {type(out)}: {out}"
                    )
```

Which is basically saying that local_map expects local tensors.

But this conflicts with the`PrepareModuleInputOutput` which uses post hooks to convert output to DTensors (bc of `use_local_output=False`). 

local_map invokes the hooks itself, so the order is:
1. the local_map converts inputs from DTensors to local
2. local_map invokes `super().__call__` which invokes the hooks
3. PrepareModuleInputOutput pre hooks run, which are basically a no-op since tensors are already converted to local
4. Module runs
5. post hooks are run, which convert tensors back to DTensors
6. local_map checks output and sees DTensor and raises the error

acisseJZhong:
this local map error is a new error after you rebase, and your fix looks correct to me! I thought I already get rid of this in https://github.com/pytorch/torchtitan/pull/2557/changes#r2962844786, looks like https://github.com/pytorch/torchtitan/pull/2744 adds it back oops.

PR #2688, opened 2026-03-24
[BE] Migrate torch.compile calls to `.compile()`

### LINKED ISSUE #2622: DCP loading fails when `torch.compile` target is nested inside `CheckpointWrapper`
## Summary

This was a super annoying bug that took a long time to track down. The summary is as follows. Chatbots helped cleaned up this report so that's why it looks nice. I'm putting the issue here because technically `CheckpointWrapper` is not part of the public API yet, and torchtitan is the main reason people use it.

### TLDR: AC(Compile(module)) causes DCP save/load errors, while Compile(AC(module)) does not.

Detailed explanation:

`torch.distributed.checkpoint`'s `get_model_state_dict` (specifically `_get_fqns`) produces state dict keys containing `_orig_mod` when a module has activation checkpointing (`CheckpointWrapper`) applied at the same level where `torch.compile` registers its `OptimizedModule`. This causes a key mismatch when saving a checkpoint from a training model (which uses activation checkpointing) and loading it into an inference model (which does not), resulting in a `RuntimeError: Missing key in checkpoint state_dict`, which is hidden behind another bug which I submitted a PR for https://github.com/pytorch/pytorch/pull/177713

## Root Cause

`_get_fqns` walks the model tree segment-by-segment, using `getattr(curr_obj, segment)` at each step. It strips `_orig_mod` from the FQN **only** when:

```python
isinstance(curr_obj, torch._dynamo.eval_frame.OptimizedModule)
```

is `True` at the moment the `_orig_mod` segment is processed.

The bug manifests when both activation checkpointing and `torch.compile` are applied to the **same** submodule (e.g. `model.layer`), and activation checkpointing runs first. I ran AC first based on torchtitan's flow, but it ended up creating a bug since I was compiling submodules not just the entire block and I didn't have the hacky private attribute access torchtitan uses.

1. `checkpoint_wrapper(original_layer)` -> `model.layer = CheckpointWrapper(original_layer)`
2. `torch.compile(original_layer)` registers `OptimizedModule` **inside** the `CheckpointWrapper`:
   `model.layer._checkpoint_wrapped_module = OptimizedModule(original_layer)`

Resulting tree at the `layer` position:

```
model.layer = CheckpointWrapper
model.layer._checkpoint_wrapped_module = OptimizedModule
model.layer._checkpoint_wrapped_module._orig_mod = original_layer
```

When `_get_fqns` walks a key like `layer._orig_mod.weight`:

| Step | `curr_obj` | `curr_obj_name` | `isinstance(curr_obj, OptimizedModule)`? | Action |
|---|---|---|---|---|
| Start | `model` | - | - | - |
| `layer` | `model` -> `CheckpointWrapper` via `getattr` | `layer` | N/A | `curr_obj = CheckpointWrapper` |
| `_orig_mod` | `CheckpointWrapper` | `_orig_mod` | **False** | `_orig_mod` appended to FQN ✗ |

So the canonical FQN is `layer._orig_mod.weight` instead of the expected `layer.weight`.

For a model with only `torch.compile` (no activation checkpointing):

```
model.layer = OptimizedModule
model.layer._orig_mod = original_layer
```

| Step | `curr_obj` | `curr_obj_name` | `isinstance(curr_obj, OptimizedModule)`? | Action |
|---|---|---|---|---|
| `layer` | `model` -> `OptimizedModule` | `layer` | N/A | `curr_obj = OptimizedModule` |
| `_orig_mod` | `OptimizedModule` | `_orig_mod` | **True** | `_orig_mod` stripped |

Canonical FQN: `layer.weight`

The result is that a training checkpoint (saved with `get_model_state_dict`) has keys like `layer._orig_mod.weight`, while the inference model's `get_model_state_dict` produces clean keys like `layer.weight`. DCP's default load planner does exact key matching, so loading fails.

Note that `model.state_dict()` produces **identical** raw keys (`layer._orig_mod.weight`) in both cases - the divergence only occurs in `_get_fqns`'s remapping.

## Reproduction Script

```
# -*- coding: utf-8 -*-
"""
Created on Wed Mar 18 01:26:01 2026

@author: Shahir
"""

import os
import sys
import tempfile
import traceback
from typing import Self

import torch
import torch.nn as nn
import torch.distributed as dist
import torch.distributed.checkpoint as dcp
import torch.distributed.checkpoint.utils as dcp_utils
from torch.distributed.checkpoint.state_dict import get_model_state_dict
from torch.distributed.checkpoint.filesystem import FileSystemReader, FileSystemWriter
from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import checkpoint_wrapper  # type: ignore
from torch.distributed.checkpoint.state_dict import _get_fqns  # type: ignore


# Workaround for https://github.com/pytorch/pytorch/issues/174669
# See: https://github.com/pytorch/pytorch/pull/177713
# This is not necessary to reproduce the bug, but it just helps show a more
# informative error instead of being hidden by the exception wrapping issue.
if sys.version_info >= (3, 13):
    def _wrap_exception_py313(
        exc: BaseException
    ) -> tuple[BaseException, traceback.StackSummary]:

        frames = traceback.extract_tb(exc.__traceback__)
        safe_summary = traceback.StackSummary.from_list([
            (f.filename, f.lineno, f.name, f.line or '') for f in frames  # type: ignore
        ])
        return (exc.with_traceback(None), safe_summary)

    dcp_utils._wrap_exception = _wrap_exception_py313  # type: ignore

# --------------------------------------------------------------------------------
# Minimal model
# --------------------------------------------------------------------------------


class SubModule(nn.Module):
    def __init__(
        self: Self
    ) -> None:

        super().__init__()
        self.weight = nn.Parameter(torch.randn(4, 4))

    def forward(
        self: Self,
        x: torch.Tensor
    ) -> torch.Tensor:

        return x @ self.weight


class Model(nn.Module):
    def __init__(
        self: Self
    ) -> None:

        super().__init__()
        self.layer = SubModule()

    def forward(
        self: Self,
        x: torch.Tensor
    ) -> torch.Tensor:

        return self.layer(x)

# --------------------------------------------------------------------------------
# Utilities
# --------------------------------------------------------------------------------


def get_submodule_parents(
    root_module: nn.Module,
    parents: dict[nn.Module, set[tuple[nn.Module, str]]] | None = None
) -> dict[nn.Module, set[tuple[nn.Module, str]]]:

    if parents is None:
        parents = {root_module: set()}

    for child_name, submodule in root_module.named_children():
        get_submodule_parents(
            root_module=submodule,
            parents=parents
        )
        if submodule not in parents:
            parents[submodule] = set()
        parents[submodule].add((root_module, child_name))

    return parents


def reregister_module(
    root_module: nn.Module,
    old_child: nn.Module,
    new_child: nn.Module
) -> None:

    parents = get_submodule_parents(root_module)
    if old_child not in parents:
        raise ValueError("Old child module not found in the module hierarchy")
    for parent, child_name in parents[old_child]:
        parent.register_module(child_name, new_child)


# --------------------------------------------------------------------------------
# Case 1: AC first, then compile to the same submodule (training scenario)
# --------------------------------------------------------------------------------

model_train = Model()
original_layer = model_train.layer

# Wrap with CheckpointWrapper
ac_layer = checkpoint_wrapper(original_layer)
reregister_module(model_train, original_layer, ac_layer)

# Compile the original module (now a child of CheckpointWrapper)
#   apply_compile finds original_layer at ac_wrapper._checkpoint_wrapped_module'
#   and registers the OptimizedModule back at that position
# Register the compiled module back into the CheckpointWrapper at the same position
# Note that torchtitan's apply_compile() functions access the private attribute
# ac_wrapper._checkpoint_wrapped_module directly to register the compiled module,
compiled_layer = torch.compile(original_layer)  # type: ignore
reregister_module(model_train, original_layer, compiled_layer)  # type: ignore

# Resulting structure:
#   model_train.layer = CheckpointWrapper
#   model_train.layer._checkpoint_wrapped_module = OptimizedModule
#   model_train.layer._checkpoint_wrapped_module._orig_mod = SubModule


# --------------------------------------------------------------------------------
# Case 2: compile only, no AC (inference scenario)
# --------------------------------------------------------------------------------

model_infer = Model()
original_layer_infer = model_infer.layer

compiled_layer = torch.compile(original_layer_infer)  # type: ignore
reregister_module(model_infer, original_layer_infer, compiled_layer)  # type: ignore

# Resulting structure:
#   model_infer.layer = OptimizedModule
#   model_infer.layer._orig_mod = SubModule


# --------------------------------------------------------------------------------
# Show bug: raw state_dict keys are identical, but _get_fqns remaps differ
# --------------------------------------------------------------------------------

raw_train = sorted(model_train.state_dict().keys())
raw_infer = sorted(model_infer.state_dict().keys())

print("Raw model.state_dict() keys (identical for both - bug NOT here):")
print(f"  AC+compile : {raw_train}")
print(f"  Compile only: {raw_infer}")
print()
print("_get_fqns canonical keys (used by DCP load planner - bug IS here):")
for raw_key in raw_train:
    fqns_train = _get_fqns(model_train, raw_key)
    fqns_infer = _get_fqns(model_infer, raw_key)
    match = "MATCH" if fqns_train == fqns_infer else "MISMATCH"
    print(f"  raw key '{raw_key}':")
    print(f"    AC+compile   -> {fqns_train}   (saved in checkpoint)")
    print(f"    Compile only -> {fqns_infer}   (expected by inference model)  {match}")
print()


# --------------------------------------------------------------------------------
# Demonstrate the actual DCP load failure
# --------------------------------------------------------------------------------

os.environ.setdefault("MASTER_ADDR", 'localhost')
os.environ.setdefault("MASTER_PORT", '8000')
dist.init_process_group(backend='gloo', world_size=1, rank=0)

with tempfile.TemporaryDirectory() as temp_dir:
    # Save: training model checkpoint (keys contain _orig_mod due to the bug)
    train_sd = get_model_state_dict(model_train)
    print(f"get_model_state_dict (AC+compile)   keys: {sorted(train_sd.keys())}")
    dcp.save(train_sd, storage_writer=FileSystemWriter(temp_dir))  # type: ignore

    # Load: inference model state dict has clean keys -> key mismatch -> error
    infer_sd = get_model_state_dict(model_infer)
    print(f"get_model_state_dict (compile only) keys: {sorted(infer_sd.keys())}")
    try:
        dcp.load(infer_sd, storage_reader=FileSystemReader(temp_dir))  # type: ignore
        print("DCP load: succeeded (unexpected)")
    except Exception as e:
        msg = str(e)
        for line in msg.splitlines():
            if "Missing key" in line:
                print(f"DCP load: FAILED (expected) - RuntimeError: {line.strip()}")
                break

dist.destroy_process_group()
```

## Expected Output

```
[Gloo] Rank 0 is connected to 0 peer ranks. Expected number of connected peer ranks is : 0
Raw model.state_dict() keys (identical for both - bug NOT here):
  AC+compile : ['layer._orig_mod.weight']
  Compile only: ['layer._orig_mod.weight']

_get_fqns canonical keys (used by DCP load planner - bug IS here):
  raw key 'layer._orig_mod.weight':
    AC+compile   -> {'layer._orig_mod.weight'}   (saved in checkpoint)
    Compile only -> {'layer.weight'}   (expected by inference model)  MISMATCH

get_model_state_dict (AC+compile)   keys: ['layer._orig_mod.weight']
get_model_state_dict (compile only) keys: ['layer.weight']
[rank0]: Traceback (most recent call last):
... (stack trace of DCP load failure) ...
[rank0]:     raise RuntimeError(f"Missing key in checkpoint state_dict: {fqn}.")
[rank0]: RuntimeError: Missing key in checkpoint state_dict: layer.weight.
```

### Versions

PyTorch 2.10+

### LINKED ISSUE #2673: [RL] Torch.compile() gives logit difference post compilation as opposed to `.compile()`
### Bug description

When enabling compile in https://github.com/pytorch/torchtitan/pull/2568, we discovered that using `torch.compile()` would affect numerics while `.compile()` would not.  This is a known issue with torch.compile - see https://github.com/pytorch/pytorch/blob/7b45702237852c500f256730439dddb726ad59c6/docs/source/user_guide/torch_compiler/compile/programming_model.where_to_apply_compile.md?plain=1#L53

### Versions

torch                                    2.12.0a0+gitbe2c9ee 
vllm                                     0.17.1rc1.dev360+g8b10e4fb3.d20260319

Lucaskabela:
## Summary
Prerequisite: https://github.com/pytorch/pytorch/pull/178330

We migrate calls from `torch.compile` to `.compile` instead.  This is preferred under [PT2 programming model](https://github.com/pytorch/pytorch/blob/e074024e5883f186a1f3c84f690cc8479a56e707/docs/source/user_guide/torch_compiler/compile/programming_model.where_to_apply_compile.md?plain=1#L55) and has benifits that:

1) We modify the module in place (no need for assignment)
2) We retain the original naming (torch.compile prefixes with `._orig_mod`), which may cause issues with weight transfer in RL or AC for instance 
3) We compile 

Fixes https://github.com/pytorch/torchtitan/issues/2622 
Fixes https://github.com/pytorch/torchtitan/issues/2673

## Testing

```bash
pytest tests/unit_tests/test_compile_moe.py
```

```bash
1 passed
```

### DIFF
--- a/torchtitan/distributed/compile.py
+++ b/torchtitan/distributed/compile.py
@@ -6,6 +6,7 @@
 
 import torch
 import torch.nn as nn
+from torch._subclasses.fake_tensor import FakeTensorMode
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
     CheckpointWrapper,
 )
@@ -13,6 +14,14 @@ from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
 from torchtitan.config import CompileConfig
 from torchtitan.models.common.moe import moe as moe_module
 from torchtitan.tools.logging import logger
+
+
+# TODO: Remove this monkeypatch once FakeTensorMode.__init__ is decorated with
+# @torch.compiler.disable(recursive=True) upstream.
+# See https://github.com/pytorch/pytorch/issues/178887
+FakeTensorMode.__init__ = torch.compiler.disable(  # type: ignore[method-assign]
+    FakeTensorMode.__init__, recursive=True
+)
 
 
 def apply_compile_dense(model: nn.Module, compile_config: CompileConfig) -> None:
@@ -33,9 +42,7 @@ def apply_compile_dense(model: nn.Module, compile_config: CompileConfig) -> None
 
     # pyrefly: ignore [missing-attribute]
     for layer_id, transformer_block in model.layers.named_children():
-        transformer_block = torch.compile(
-            transformer_block, backend=compile_config.backend, fullgraph=True
-        )
+        transformer_block.compile(backend=compile_config.backend, fullgraph=True)
         # pyrefly: ignore [missing-attribute]
         model.layers.register_module(layer_id, transformer_block)
 
@@ -93,27 +100,13 @@ def apply_compile_sparse(
                             # NOTE: We don't compile token dispatch and token combine due to an issue on B200:
                             # https://github.com/pytorch/torchtitan/issues/1940
                             continue
-                        setattr(
-                            moe,
-                            attr_name,
-                            torch.compile(
-                                submod, backend=compile_config.backend, fullgraph=True
-                            ),
-                        )
+                        submod.compile(backend=compile_config.backend, fullgraph=True)
                 else:
-                    setattr(
-                        block,
-                        attr_name,
-                        torch.compile(
-                            submod, backend=compile_config.backend, fullgraph=True
-                        ),
-                    )
-
+                    submod.compile(backend=compile_config.backend, fullgraph=True)
         else:
             # If it's not a MoE layer, there is no FSDP(GroupedExperts)
             # So we can compile the whole block
-            transformer_block = torch.compile(
-                transformer_block,
+            transformer_block.compile(
                 backend=compile_config.backend,
                 fullgraph=True,
             )
@@ -154,26 +147,3 @@ def apply_compile_sparse(
     # https://github.com/pytorch/pytorch/issues/166460
 
     logger.info("Compiling each TransformerBlock with torch.compile")
-
-
-def apply_compile_dense_rl(model: nn.Module, compile_config: CompileConfig) -> None:
-    """Apply torch.compile in-place to each TransformerBlock for RL training.
-
-    Uses ``Module.compile()`` (in-place) instead of ``torch.compile()`` +
-    ``register_module()`` to preserve weight naming compatibility between
-    the trainer and vLLM generator (avoids ``_orig_mod`` prefix that
-    ``OptimizedModule`` wrapping introduces).
-
-    Also fixes some numeric issues we have observed with ``torch.compile()``.
-    See https://github.com/pytorch/torchtitan/issues/2673 for more details.
-    """
-    # pyrefly: ignore [missing-attribute, not-callable]
-    for _, transformer_block in model.layers.named_children():
-        # pyrefly: ignore [missing-attribute]
-        transformer_block.compile(
-            backend=compile_config.backend,
-            fullgraph=True,
-            dynamic=True,
-        )
-
-    logger.info("Compiled each TransformerBlock with torch.compile (RL in-place)")
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -29,7 +29,6 @@ from torch.nn.attention.flex_attention import (
     flex_attention,
 )
 from torch.nn.attention.varlen import varlen_attn
-from torch.types import Number
 
 from torchtitan.models.common.linear import Linear
 from torchtitan.models.common.rmsnorm import RMSNorm
@@ -64,8 +63,8 @@ class VarlenMetadata(NamedTuple):
 
     cu_seq_q: torch.Tensor
     cu_seq_k: torch.Tensor
-    max_q: Number
-    max_k: Number
+    max_q: int
+    max_k: int
 
 
 AttentionMasksType = dict[str, BlockMask] | BlockMask | VarlenMetadata
@@ -187,7 +186,8 @@ class VarlenAttentionWrapper(LocalMapAttention):
         xk_packed = xk_packed.to(torch.bfloat16)
         xv_packed = xv_packed.to(torch.bfloat16)
 
-        return varlen_attn(
+        # pyrefly: ignore [bad-assignment]
+        res: torch.Tensor = varlen_attn(
             xq_packed,
             xk_packed,
             xv_packed,
@@ -207,7 +207,8 @@ class VarlenAttentionWrapper(LocalMapAttention):
             #               is_causal=False.
             #   - (W, 0): Sliding window causal - attend to at most W previous tokens.
             window_size=(-1, 0),
-        ).to(xq.dtype)
+        )
+        return res.to(xq.dtype)
 
 
 class FlexAttentionWrapper(LocalMapAttention):
@@ -488,10 +489,11 @@ def create_varlen_metadata_for_document(
         cu_seqlens_list + [torch.tensor([offset], dtype=torch.int32, device=device)]
     )
 
-    max_seqlen = 0
+    max_seqlen: int = 0
     if len(all_seq_lengths) > 0:
         all_seq_lengths = torch.cat(all_seq_lengths)
         # device to host sync but only done once per model forward
+        # pyrefly: ignore[bad-assignment]
         max_seqlen = all_seq_lengths.max().item()
 
     return VarlenMetadata(
--- a/torchtitan/experiments/graph_trainer/compile.py
+++ b/torchtitan/experiments/graph_trainer/compile.py
@@ -80,8 +80,7 @@ def _apply_jit_compile(
         fsdp_reshard_after_forward,
         transformer_block_buckets,
     )
-    model = torch.compile(
-        model,
+    model.compile(
         backend=backend,
         fullgraph=True,
     )
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -80,6 +80,13 @@ class PolicyTrainer(Actor, Configurable):
         requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
         self._transfer_dtype = requested if requested != training_dtype else None
 
+        # The policy and ref models share code objects, so dynamo's
+        # per-code-object cache must hold entries for both grad modes
+        # (grad for policy, no_grad for ref). The default limit of 8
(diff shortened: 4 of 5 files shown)


tianyu-l (approved):
sg, could you verify that compile `aot_eager` backend still gives the identical numerics result as pure eager?

Lucaskabela:
@tianyu-l  will do - that said, there is an issue with the moe + AC (as seen in the CI failure) so I need to address that first before landing

tianyu-l:
maybe https://github.com/pytorch/torchtitan/pull/2281 can address it
cc @weifengpy if you want to land that PR soon

fegin (approved):
Should we also change graph trainer to use the same pattern? This way the checkpointer refactor can be easier.

Lucaskabela:
> Should we also change graph trainer to use the same pattern? This way the checkpointer refactor can be easier.

I will take a spike on this and investigate :)

Lucaskabela:
These failing tests seem to be from some pytorch change - investigating more, see https://github.com/pytorch/torchtitan/pull/2714

torchtitan/distributed/compile.py
```diff
@@ -15,6 +16,20 @@
...
+# FakeTensorMode.__init__ calls traceback.extract_stack() which uses
+# walk_stack() — a Python generator that torch.compile cannot trace.
+# This is hit when module.compile() compiles _call_impl (including DTensor
+# hooks) and the DTensor sharding propagation slow path creates a new
+# FakeTensorMode during tracing. Disabling compile for __init__ lets it
+# run eagerly; the stack trace is only used for debug error messages.
+# recursive=True is required so that the eval frame callback is fully
+# removed during __init__, otherwise walk_stack() generator frames are
+# still intercepted by dynamo.
+FakeTensorMode.__init__ = torch.compiler.disable(  # type: ignore[method-assign]
```
aditvenk, line 22: naive qn: Why should this workaround be in TorchTitan? seems like a generic issue to anyone using compile?

aditvenk (approved):

torchtitan/distributed/compile.py
Lucaskabela: I am planning on upstream fixing in pytorch/pytorch - let me link the issue in this comment so we can remove once it is fixed :)

torchtitan/models/common/attention.py
```diff
@@ -64,8 +63,8 @@ class VarlenMetadata(NamedTuple):
 
     cu_seq_q: torch.Tensor
     cu_seq_k: torch.Tensor
-    max_q: Number
```
Lucaskabela, line 67: Need to fix the type errors

Lucaskabela:
Validated these unit tests passed on local with https://github.com/pytorch/pytorch/pull/178788 patched

PR #2694, opened 2026-03-24
Add dynamic batching for RL trainer forward passes

### LINKED ISSUE #2693: [RL] Dynamic Batching in RL loop
Now the trainer takes a single episode, don't do any padding or packing.

Need to implement Dynamic batching for training efficiency: 

Image

Related discussion:
1. https://github.com/pytorch/torchtitan/pull/2568#discussion_r2970611002

dean-mccoppin:
The trainer currently loops over episodes one by one, doing a separate forward pass for each. This batches them into a
single padded forward pass instead, going from 2N passes per step down to 2

Variable-length sequences are right-padded to the longest in the batch. Causal attention means padding at the end can't
affect earlier positions, so we just extract each episode's log probs from the batched output.

VLLMCompatibleFlashAttention has only ever seen batch_size=1 so this needs GPU testing with the actual model before
merging. No sub-batching for memory yet either, one long outlier episode pads everything to its length

22 CPU unit tests covering causal attention models, padding leak detection, gradient matching, and end-to-end policy
gradient loss. Still needs GPU runs with the debug and 0.6B configs though

closes #2693

### DIFF
--- a/tests/unit_tests/test_batch_token_log_probs.py
+++ b/tests/unit_tests/test_batch_token_log_probs.py
@@ -1,0 +1,667 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Unit tests for compute_batch_token_log_probs.
+
+Verifies that the batched implementation produces identical results to the
+per-episode compute_token_log_probs across various sequence length combinations,
+including models with causal self-attention where padding could leak information.
+"""
+
+import importlib.util
+import math
+import os
+import unittest
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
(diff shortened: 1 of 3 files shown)


tianyu-l (changes requested):
Thanks for contributing. There seem several issues with this implementation:
1. it uses padding, instead of packing, which doesn't sound efficient
2. using default pad_token_id 0 sounds arbitrary can cause collision
3. not utilizing varlen attention's metadata / masking to exclude computation on padding, which is not efficient

dean-mccoppin:
> Thanks for contributing. There seem several issues with this implementation:
> 
> 1. it uses padding, instead of packing, which doesn't sound efficient
> 2. using default pad_token_id 0 sounds arbitrary can cause collision
> 3. not utilizing varlen attention's metadata / masking to exclude computation on padding, which is not efficient

Good points, I didnt do my due diligence with this one. I'll take it slower and revise, appreciate it

PR #2695, opened 2026-03-24
[deepseek_v3/graph_trainer] Add 16B_sdpa and debugmodel_balanced configs

xmfan:
[deepseek_v3/graph_trainer] Add 16B_sdpa and debugmodel_balanced configs

Add model config variants for graph-compilation-compatible benchmarking:
- 16B_sdpa: 16B with sdpa attention (instead of flex) and balanced routing
- debugmodel_balanced: debugmodel with force-balanced MoE routing

These variants avoid flex attention and data-dependent routing ops that
Inductor cannot compile, enabling end-to-end graph compilation.


xmfan:
ghstack incident

PR #2696, opened 2026-03-24
[EP/graph_trainer] Add PaddedExpertParallel and _local_scalar_dense fallback

xmfan:
[EP/graph_trainer] Add PaddedExpertParallel and _local_scalar_dense fallback

Add PaddedExpertParallel, a graph-compilation-compatible alternative to
ExpertParallel that avoids data-dependent ops (_local_scalar_dense).

Standard ExpertParallel uses dynamic split sizes for all-to-all and
dynamic permutation indices, both of which introduce _local_scalar_dense
ops that Inductor cannot compile. PaddedExpertParallel instead:
1. Pads routed_input so each rank sends/receives equal-sized chunks
2. Uses equal-split all-to-all (split_sizes=None)
3. Replaces dynamic _permute with static reshape+transpose

Also updates the graph_trainer DeepSeek V3 parallelize to use
PaddedExpertParallel when EP is enabled, force-balance routing on
all MoE modules, and annotate PaddedExpertParallel for FX tracing.

This is temporary infrastructure that will be removed when non-even
routing support lands in the compiler.

### DIFF
--- a/torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
+++ b/torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from torch import nn
 from torch.distributed.device_mesh import DeviceMesh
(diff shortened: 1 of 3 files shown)


xmfan:
ghstack incident

PR #2699, opened 2026-03-24
[graph_trainer] Wire graph PP into DeepSeek V3 and trainer

xmfan:
[graph_trainer] Wire graph PP into DeepSeek V3 and trainer

Add _pipelining_fn to graph_trainer's DeepSeek V3 model_registry that
dispatches to graph_pp_pipeline_llm when graph_pp_passes are configured,
falling back to standard pipeline_llm otherwise.

Add train_step override to GraphTrainer that handles SymFloat grad_norm
values from fx.Interpreter execution. When graph PP executes subgraphs
via the interpreter, gradient operations may return SymFloat instead of
regular Tensor, which breaks metrics formatting.


xmfan:
ghstack incident

PR #2700, opened 2026-03-24
[graph_trainer] Support DualPipeV in graph PP pipeline

xmfan:
[graph_trainer] Support DualPipeV in graph PP pipeline

Add DualPipeV and ZBV schedule support to the graph PP pipeline:

- Auto-detect V-schedules (DualPipeV, ScheduleZBVZeroBubble) and set
  stages_per_rank=2 for virtual stage splitting
- Auto-enable split_dI_dW pass for V-schedules to enable zero-bubble
  scheduling where dW is interleaved with the next F
- Add BACKWARD_INPUT/BACKWARD_WEIGHT fallback handlers that delegate to
  stage_full_backward when split dI/dW graphs are not available
- Add OVERLAP_F_B handler using multiplexed graphs that fuse F+B into a
  single graph with interleaved compute/comm for true overlap
- Disable donated buffers to avoid "non-empty donated buffers" error
  during AOT compilation with DualPipeV backward


xmfan:
ghstack incident

PR #2701, opened 2026-03-24
[graph_trainer] Add graph PP documentation to README

xmfan:
[graph_trainer] Add graph PP documentation to README

Add Graph-based Pipeline Parallelism section with run commands for
DeepSeek-v3-16B with Interleaved1F1B and DualPipeV schedules. Note
that balanced routing is temporarily required until Inductor supports
non-even routing. Update composability table accordingly.


xmfan:
ghstack incident

PR #2709, opened 2026-03-24
[rl] remove duplicated transpose around VllmAttention in generator

wwwjn:
As titled

 - Previous Standard GQAttention  +  VLLMInnerAttention flow: (bs,seq,heads,dim) → transpose → (bs,heads,seq,dim)  → VLLMInnerAttention transpose back → clone → (bs*seq,heads,dim) → vllm_attn → reshape → transpose → contiguous  (4 layout ops)
Screenshot 2026-03-30 at 1 53 29 PM

- New VLLMGQAttention subclass: (bs,seq,heads,dim) → reshape → (bs*seq,heads,dim)  (1 zero-copy op)→ vllm_attn → view → (bs,seq,hidden)
Screenshot 2026-03-30 at 1 14 26 PM

From Claude's summary of trace before vs. after:
- aten::transpose: 28,090 → 15,486 (-12,604 calls), the redundant transpose round-trip is gone                                                           
- aten::view: 54,508 → 51,707 (-2,801 calls), fewer intermediate views

### DIFF
--- a/torchtitan/experiments/rl/models/attention.py
+++ b/torchtitan/experiments/rl/models/attention.py
@@ -301,8 +301,8 @@ class VLLMGQAttention(GQAttention):
 
     Standard GQAttention + VLLMInnerAttention flow:
       (bs,seq,heads,dim) -> transpose -> (bs,heads,seq,dim)
-      -> VLLMInnerAttention transpose back -> clone -> (bs*seq,heads,dim)
-      -> vllm_attn -> reshape -> transpose -> contiguous  (4 layout ops)
+      -> VLLMInnerAttention transpose back -> (bs*seq,heads,dim)
+      -> vllm_attn -> reshape -> transpose  (4 layout ops)
 
     This subclass:
       (bs,seq,heads,dim) -> reshape -> (bs*seq,heads,dim)  (1 zero-copy op)


torchtitan/experiments/rl/models/attention.py
```diff
@@ -321,7 +406,10 @@ def replace_with_vllm_attention(model, tp_degree=1):
 
         layer.attention.inner_attention = vllm_attn
 
+        # Swap GQAttention to VLLMGQAttention for zero-copy forward path
+        layer.attention.__class__ = VLLMGQAttention
```
tianyu-l: changing `__class__` sounds suspicious -- is it replacing attention with a newly created module?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -220,60 +229,136 @@ def __init__(
...
+        # The inner_attention will only be VLLMInnerAttention class.
+        assert isinstance(
+            self.inner_attention, VLLMInnerAttention
+        ), "inner_attention must be an instance of VLLMInnerAttention"
+        output = self.inner_attention(xq, xk, xv)
+
+        return self.wo(output)
 
 
 def replace_with_vllm_attention(model, tp_degree=1):
```
tianyu-l: This should be done at config time, not after model is built.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -220,60 +229,136 @@ def __init__(
...
+
+        Expects heads sharded on dim 2 (not dim 1 as in base
+        LocalMapAttention), matching VLLMGQAttention's zero-copy path.
+        """
+        if isinstance(q, DTensor):
+            assert isinstance(k, DTensor) and isinstance(v, DTensor)
+            assert q.placements == k.placements == v.placements
+            # qkv are (bs, seq, heads, dim) — heads on dim 2
+            for i, p in enumerate(q.placements):
+                assert p == Shard(2), (
```
tianyu-l: If this is the only reason you need to create another `__call__`, change the original place and only guard on qkv having the same shard dim.

torchtitan/experiments/rl/models/attention.py
wwwjn: nice catch, now we subclass GQAttention to be VLLMGQAttention, we should be able to directly set it during config time.

However, we only have one copy of model_config which will be shard by trainer and generator, so I replaced GQAttention->VLLMGQAttention in generator side model_config, before calling model_config.build()

torchtitan/experiments/rl/models/attention.py
```diff
@@ -225,103 +235,130 @@ def forward(
...
-        vllm_attn = VLLMAttention(
-            hidden_size=model_args.dim,
-            num_heads=model_args.layer.attention.n_heads // tp_degree,
-            num_kv_heads=num_kv_heads,
+    # Counter to auto-assign layer names when built via config
+    _layer_counter: itertools.count = itertools.count()
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(GQAttention.Config):
+        attn_backend: str = "vllm"
```
tianyu-l: This doesn't sound right. `attn_backend` should be "varlen", in the future there will be "flex" as well.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -225,103 +235,130 @@ def forward(
...
+
+        def __post_init__(self):
+            # Skip parent validation which doesn't know about "vllm" backend
+            assert self.n_heads > 0, "n_heads must be > 0"
+
+    def __init__(self, config: Config, *, dim: int):
+        # Use "sdpa" for parent init to avoid ValueError on unknown backend;
+        # inner_attention is immediately replaced below.
+        parent_config = dataclasses.replace(config, attn_backend="sdpa")
+        super().__init__(parent_config, dim=dim)
```
tianyu-l: no this is too hacky, also we shouldn't have vllm attn backend -- vllm is more for KV cache management.

torchtitan/experiments/rl/models/attention.py
tianyu-l: Is this field required? @fegin is doing some config expansion work so that each transformer layer may end up having its own config, after which we probably can put it in config.

torchtitan/experiments/rl/models/attention.py
```diff
@@ -225,103 +235,130 @@ def forward(
...
-            raise ValueError(f"Layer {layer_name} must have .attention attribute")
-
-        # GQA
-        head_dim = model_args.layer.attention.head_dim
-        vllm_attn = VLLMAttention(
-            hidden_size=model_args.dim,
-            num_heads=model_args.layer.attention.n_heads // tp_degree,
-            num_kv_heads=num_kv_heads,
+    # Counter to auto-assign layer names when built via config
+    _layer_counter: itertools.count = itertools.count()
```
tianyu-l: This is hacky, e.g. PP doesn't work this way. I'm OK with it being temporary solution, but leave a TODO here.

torchtitan/experiments/rl/models/attention.py
wwwjn: vLLM requires a unique prefix per Attention layer for static_forward_context registration,  static_forward_context is a dict that vLLM uses during torch.compile piecewise CUDA graph capture

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -150,8 +150,20 @@ def __init__(
...
+        # Swap GQAttention -> VLLMGQAttention at config time so the model
+        # builds with vLLM-compatible attention from the start.
+        model_config = model_spec.model
+        attn_config = model_config.layer.attention
+        vllm_attn_fields = {
+            f.name: getattr(attn_config, f.name)
+            for f in dataclasses.fields(attn_config)
+            if f.init
+        }
+        vllm_attn_fields["attn_backend"] = "varlen"
```
tianyu-l: This is overwriting what's already in model config. Instead, I think we shouldn't change it, and we should assert the attn_backend to be varlen. The reason is that for model config which is using flex attention, we can't model it with varlen so results will be silently wrong.

torchtitan/models/common/attention.py
```diff
@@ -106,15 +106,17 @@ def __call__(
...
-            # on the n_heads dim (dim 1)
-            # TODO: after full DTensor rewrite, the DP mesh will also be
-            # present, update this check to allow Shard(0) for DP and Shard(1) for TP.
             for i, p in enumerate(q.placements):
-                assert p == Shard(1), (
-                    f"LocalMapAttention requires Shard(1) placements "
-                    f"(n_heads dim), but got {p} at position {i}"
+                assert isinstance(p, Shard), (
+                    f"LocalMapAttention requires Shard placements "
+                    f"(heads dim), but got {p} at position {i}"
```
tianyu-l: revert?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -229,118 +243,134 @@ def __init__(
...
+        layer_id = next(VLLMInnerAttention._layer_counter)
         self.vllm_attn = Attention(
             num_heads=num_heads,
             head_size=head_dim,
             scale=self.scale,
             num_kv_heads=num_kv_heads,
             cache_config=cache_config,
             quant_config=None,
-            prefix=f"model.layers.{layer_name}.attention.inner_attention",
+            prefix=f"model.layers.{layer_id}.attention.inner_attention",
```
tianyu-l, line 254: > vLLM requires a unique prefix per Attention layer for static_forward_context registration

Just for my understanding, if uniqueness is the only requirement, in theory can we just pass in a random str for each instance?

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This line is a temporary solution to enforce attn_backend to be varlen before #2649 is landed. I will hold this PR a little bit and after #2649 is landed, the code path will be more clear

torchtitan/experiments/rl/models/attention.py
wwwjn: I asked Claude, here's the reply: Yes, uniqueness is the only requirement, vLLM's own models follow this `model.layers.N.self_attn `pattern and matching it makes debugging/profiling easier, but functionally any unique string works.

PR #2712, opened 2026-03-25
Paged stashing

syed-ahmed:
WIP

Page Stashing is a memory planning pass (similar to SAC/activation offloading) which reduces the number buffers used by MoE models when CUDAGraph'd. The pass improves memory consumption from O(layers x hybrid worst case dispatch size) to O(worst case + actual usage).


torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+### Step 1: The Oversized Buffer is Created at Dispatch
+
+The HybridEP dispatcher pre-sizes the output buffer to worst-case capacity:
+
+```python
+# token_dispatcher.py:1012-1020
+budget = int(
+    routing_map.shape[0]
+    * self.config.moe_router_topk
+    * self.moe_expert_rank_capacity_factor   # e.g. 1.0 (worst case)
```
xmfan: with unlucky routing, doesn't this OOM?

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+
+**Runtime resize** (mirrors Megatron's `allocate_stash_buffers`):
+
+During CUDAGraph warmup (step 1), the `PagedStashObserver` (status='capture') records
+actual and avg token counts via `on_copy`/`on_pop` calls in the `paged_stash.copy` and
+`paged_stash.pop` custom op bodies. These mirror Megatron's `on_save_for_backward` /
+`on_get_saved_tensor` increment/decrement counters.
+
+After step 1, `allocate_stash_buffers` uses the observed
+`max_tokens_across_vp_stages` / `max_avg_tokens_across_vp_stages` to resize buffers
```
xmfan: What consists of a vp_stage? Is it the PP stage? And does this use per-stage avg tokens from only the first training iteration?

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+Combine: weighted sum of expert outputs -> output [num_tokens, dim]
+```
+
+Each expert is a small FFN (typically SwiGLU: `silu(x @ w1) * (x @ w3)` then `h @ w2`). With `_grouped_mm`, all experts run as a single batched matrix multiply using offset indices, avoiding a per-expert for-loop.
+
+### Expert Parallelism (EP)
+
+With EP, experts are distributed across ranks. Each rank owns `num_experts / ep_size` local experts. Before expert computation, tokens must be sent to the rank that owns their assigned expert (**dispatch**), and results sent back afterward (**combine**). This is an all-to-all communication pattern.
+
+In token-dropless MoE training, the number of tokens received by each expert varies, resulting in a **dynamic shaped tensor**. PyTorch handles this naturally in eager mode -- tensors are allocated lazily when their shape is known at runtime. However, dynamic shaped tensors pose a fundamental challenge for CUDA graphs: there is no CUDA API that allows allocating GPU memory within a CUDA graph with a size determined in stream order.
```
xmfan: I haven't understood how this approach is token-dropless. Don't we OOM because of the large static allocations?

xmfan (commented):
Thanks for putting this up. Left some comments to try to understand the approach more.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+
+For simple scheduling where activation allocation/deallocation follows a first-in-last-out pattern, stash and restore can be done with **bump allocation** (a simple stack pointer). To accommodate complex scheduling (e.g., pipeline parallelism with interleaved fwd/bwd), **paging** provides the flexibility to allocate and free pages in arbitrary order -- hence the name "paged stashing."
+
+### Prerequisites: Reducing Compute Fragmentation
+
+Paged stashing works best when paired with techniques that reduce **compute fragmentation** -- skipping padded data in the oversized static buffers:
+
+- **HybridEP**: Token dispatch kernels that work with pre-sized (worst-case) output buffers without CPU-GPU synchronization.
+- **Host-free Grouped GEMM** (`_grouped_mm` with on-device offsets): Expert computation that uses GPU-side offset indices (`torch.cumsum` on device) rather than CPU token counts, avoiding D2H sync.
```
tianyu-l: IIUC, the current `torch._grouped_mm` requires the `offsets` arg to be on CPU. Does it support GPU offsets, or do we have plan to support? @ngimel

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+
+Paged stashing **decouples** the need for oversized buffers during compute from the need for properly-sized buffers for storing activations:
+
+- **Compute buffers** (oversized, temporary): Used during expert forward/backward kernels. Sized to worst-case capacity. Freed after compute completes.
+- **Stash buffers** (right-sized, paged, persistent): Used to store activations for the backward pass. Sized to actual usage. Persist from forward to backward.
+
+The **stash operation** copies the activation from the oversized compute buffer into the paged stash buffer after each expert layer's forward completes. The **restore operation** copies it back during the backward pass. The key memory saving: the stash packs variable-size activations into a contiguous paged buffer, eliminating fragmentation.
+
+For simple scheduling where activation allocation/deallocation follows a first-in-last-out pattern, stash and restore can be done with **bump allocation** (a simple stack pointer). To accommodate complex scheduling (e.g., pipeline parallelism with interleaved fwd/bwd), **paging** provides the flexibility to allocate and free pages in arbitrary order -- hence the name "paged stashing."
```
tianyu-l: > pipeline parallelism with interleaved fwd/bwd

Even for PP without interleaved fwd/bwd, let's say the simplest schedule GPipe (forward microbatch 1 -> 2 -> 3 -> 4, then backward 1 -> 2 -> 3 ->4) it is not LIFO, so paging is needed. Is this accurate?

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+  the concurrent high-water mark across interleaved microbatches. Buffer sizing is exact
+  based on the observed runtime interleaving pattern.
+
+- **torchtitan** (future): Will implement similar schedule-driven coordination.
+  The `PagedStashObserver` already uses the increment/decrement pattern which correctly
+  captures PP interleaving if the observation iteration includes multiple microbatches.
+
+---
+
+## Relationship to PyTorch Activation Offloading
```
tianyu-l: Here you are comparing it with activation offloading, instead of activation checkpointing. I guess it's because the point is to make paged stashing save & load async, is that correct?

torchtitan/experiments/cuda_graphable_moe/README.md
```diff
@@ -0,0 +1,256 @@
...
+| `CUDA_HOME` | Yes | Path to CUDA toolkit (e.g., `/usr/local/cuda`) for DeepEP JIT compilation |
+| `NCCL_GRAPH_REGISTER` | No | Set to `0` to disable NCCL graph registration if needed |
+
+### Paged Stash Buffer Settings
+
+Exposed through `PagedStashActivationCheckpointConfig`:
+
+| Field | Default | Description |
+|---|---|---|
+| `paged_stash_buffer_device` | `"cuda"` | Device for the paged buffer (`"cuda"` or `"cpu"`) |
```
tianyu-l: So it would be either GPU or CPU. What if we picked GPU but the total stashed activations on some GPU is too much?

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+Both Megatron and torchtitan use the same paged buffer design. Each buffer is organized as `[total_tokens, hidden_size]` backed by a circular free list of page IDs:
+
+```
+Buffer:     [page_0 | page_1 | page_2 | ... | page_N]   (each page = page_size tokens)
+Free list:  [2, 5, 0, 3, 1, 4, ...]                      (available page IDs)
+Head ------>                                               (allocate from head)
+                                          <---- Tail       (return to tail)
+```
+
+Page management is handled by lightweight GPU kernels fused with the stash/restore operations. The Triton `_paged_stash_copy_kernel` allocates pages from the head and copies token data. The `_paged_stash_pop_kernel` copies back and returns pages to the tail. Both operate entirely on GPU -- no CPU involvement.
```
tianyu-l: Have you considered comm ops directly put dispatched tokens to the paged buffer, so that following computation (grouped gemm) can be performed on the buffer region?

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
```diff
@@ -0,0 +1,1095 @@
...
+
+**Phase 2 -- `captured` (iteration 2):**
+Buffers are allocated from the high-water mark: `num_tokens = max_tokens[key] * stash_buffer_size_factor`. The capture step doubles as the CUDA graph warmup (needed anyway).
+
+**Phase 3 -- steady state (iteration 3+):**
+CUDA graphs are active. Each step resets buffer free lists, replays the stash/restore pattern recorded during capture.
+
+---
+
+## How It Works in torchtitan
```
tianyu-l: Have you considered making it work in eager? Would love to hear more about the consideration.

syed-ahmed:
Apologies, I was OOTO last week. Will reply to the review soon.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
ngimel: `offsets` has to be only on GPU, cpu offsets are not supported

tianyu-l: oh my bad. I was mistakenly thinking of `output_splits` which was needed by NCCL all-to-all to be on CPU. Both `output_splits` and `offsets` come from `num_tokens_per_expert_group` here https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py#L169

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: The capacity factor of 1.0 sizes the HybridEP dispatch buffer for worst case, so the dispatch itself won't OOM regardless of routing.

For the paged stash buffers, this is a fair concern though. In eager mode, we don't need to pre-budget - dynamic allocations just happen, and if you're unlucky you OOM. But with CUDA graphs, memory is fixed at capture time, so you have to estimate upfront.

Megatron has two strategies: (1) static - assume balanced routing, size to avg_tokens x factor (default), and (2) dynamic - use the first iteration's actual token counts. In practice they've been using the static approach because it works well enough, and the dynamic estimation adds complexity without much benefit.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: VP is virtual pipeline. Apparently it's finer granularity than a PP stage. Please ignore vp stage. I'll remove all instances of it in this implementation since I'm not considering PP at the moment (I believe SOTA for DeepSeek right now is with PP=1).

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: Paged stashing is not solving the token-dropless problem directly. Paged stashing is just a memory management technique or if it helps just think of it as a custom allocator. Without paged stashing, large tensors (as a result of dynamic shapes) would fragment the default CUDACachingAllocator segments. With paged stashing, we are creating a buffer and controlling how the large tensors are packed in that buffer. The "how" using the stash and pop kernels improves the fragmentation and as a result reduces the chances of OOM. The actual large tensor should be free'd (shouldn't cross the fwd->bwd boundary since there shouldn't be consumers in bwd - bwd will pop from the paged buffer).

The paged buffer is sized to balanced estimate, and if we exhaust out this space, we can spillover to host.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: Yes that's correct. Without PP, this feature is called "Packed" offloading/stashing, since we are packing the oversized tensors better using the custom triton kernels. With PP, the concept of pages are needed to account for the order of a PP schedule, since pages are freed individually in arbitrary order. I'm not considering PP changes in this diff at the moment, since PP=1 is what we are using for SOTA deepseek training.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: Yes. I basically followed the activation offloading joint graph pass pattern, since it seemed like the most natural mapping. Instead of D2H, we do a stash operation, and instead of a H2D, we do a pop from the stashing buffer. These operations should also be done in a separate stream and should overlap, and the activation offloading pass was already demonstrating that.

torchtitan/experiments/cuda_graphable_moe/README.md
syed-ahmed: Yes. The CPU device is a fallback path, and basically should work like activation offloading to CPU. Only difference would be you are offloading pages instead of the actual activations to a pre-allocated pinned host buffer.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: That's interesting. I haven't. Are you thinking of a fused comm+paged_buffer+compute type of kernel? The stash and pop should overlap in separate streams, so wonder if overlapping would just give the same performance benefit.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
syed-ahmed: Yes. Megatron currently only supports eager. The main problem is how are we composing CUDAGraph on an e2e eager based training in TorchTitan? The CUDAGraph usage in graph trainer experiment is closest to what Megatron is doing today (FullCUDAGraphWrapper on the fwd/bwd), and that's why primarily went with this approach.

Assuming we can wrap a Hybrid-EP enabled DeepSeek training with CUDAGraphs in eager mode, paged stashing can be implemented as another flavor of activation checkpointing/offloading using saved tensor hooks. Perhaps may be a custom policy function. Megatron implements it as a saved tensor hook but the composability story (e.g. working alongside other activation checkpointing flavors) is lacking.

torchtitan/experiments/cuda_graphable_moe/paged_stashing_guide.md
tianyu-l: There might be some temporary memory overhead but may be fine?

btw the idea of "performing computation directly in the buffer region" was mentioned by the great @danthe3rd

torchtitan/experiments/cuda_graphable_moe/README.md
tianyu-l: Could there be a concept of "page miss", i.e. when the GPU buffer region becomes full, we don't save all activations in forward and instead rely on backward to recompute?

syed-ahmed:
Closing in favor of stacked PR.

PR #2713, opened 2026-03-26
[graph_trainer] Add CooR precompile support for rank-agnostic compilation

bobrenjc93:
* __->__ #2713

Adds single-process precompilation using compile-on-one-rank (CooR) to
generate rank-agnostic compiled artifacts. A single GPU traces through
distributed ops with a fake backend, producing an artifact loadable by
all ranks during torchrun training.

Key changes:
- precompile.py: Artifact serialization/loading with config fingerprint
  validation. DeviceMesh objects flow as opaque graph inputs (via
  DTensor.__tensor_flatten__ -> OpaqueObjectState), so no post-load
  patching or PG aliasing is needed.
- precompile_main.py: Single-process entry point using fake PG backend
- graph_utils.py: Enable CooR in export_joint() when precompile is on,
  so PG names are opaque graph inputs (not hardcoded string constants)
- compile.py: Precompile save/load integration with AOT compilation
- passes.py: Fix inductor_decomposition_pass for CooR opaque inputs

Verification (Llama3-8B, MAST, 2 nodes / 16 GPUs):
  Model: graph_trainer_llama3_8b
  Parallelism: TP=8, SimpleFSDP=2
  Compile: AOT, full_inductor_compilation + inductor_decomposition
  Training: 100 steps, local_batch_size=1, seq_len=2048
  Determinism: --debug.seed=42 --debug.deterministic

  CooR precompile (1 node, 1 GPU, fake PG backend)
    -> train (2 nodes, 16 GPUs) vs baseline (no precompile)
  Result: bitwise identical loss and grad_norm for all 100 steps

image

### DIFF
--- a/torchtitan/experiments/graph_trainer/tests/run_precompile_tests.py
+++ b/torchtitan/experiments/graph_trainer/tests/run_precompile_tests.py
@@ -23,8 +23,8 @@ import os
 import subprocess
 import tempfile
 import time
-from dataclasses import dataclass
 from collections.abc import Sequence
+from dataclasses import dataclass
 
 from torchtitan.tools.logging import logger
 
@@ -121,14 +121,8 @@ def run_precompile_tests(args):
             )
 
         # Step 2: training with the precompiled artifact
-        cmd = (
-            f"NGPU={test.ngpu} LOG_RANK={all_ranks} "
-            f"./{RUN_TRAIN_SCRIPT}"
-        )
-        cmd = (
-            f'TORCH_TRACE="{args.output_dir}/{test.test_name}/compile_trace" '
-            + cmd
-        )
+        cmd = f"NGPU={test.ngpu} LOG_RANK={all_ranks} " f"./{RUN_TRAIN_SCRIPT}"
+        cmd = f'TORCH_TRACE="{args.output_dir}/{test.test_name}/compile_trace" ' + cmd
         cmd += " " + dump_folder_arg
         cmd += " " + " ".join(test.override_args)
 
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -80,13 +80,6 @@ class PolicyTrainer(Actor, Configurable):
         requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
         self._transfer_dtype = requested if requested != training_dtype else None
 
-        # The policy and ref models share code objects, so dynamo's
-        # per-code-object cache must hold entries for both grad modes
-        # (grad for policy, no_grad for ref). The default limit of 8
-        # is not enough; 16 accommodates both without recompile storms.
-        # TODO: @Lucaskabela fix recompiles in general as these increase startup
-        torch._dynamo.config.cache_size_limit = 16
-
         # Device setup
         device_module, device_type = utils.device_module, utils.device_type
         self.device = torch.device(f"{device_type}:{int(os.environ['LOCAL_RANK'])}")
@@ -273,6 +266,13 @@ class PolicyTrainer(Actor, Configurable):
         Returns:
             Training metrics
         """
+        # The policy and ref models share code objects, so dynamo's
+        # per-code-object cache must hold entries for both grad modes
+        # (grad for policy, no_grad for ref). The default limit of
+        # is not enough; 16 accommodates both without recompile storms.
+        # TODO: @Lucaskabela fix recompiles in general as these increase startup
+        torch._dynamo.config.recompile_limit = 16
+
         logger.debug(
             f"{os.getpid()=} PolicyTrainer starting step {self.policy_version} "
         )


torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,52 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    --compile.passes full_inductor_compilation \
+    --compile.joint_passes inductor_decomposition \
+    --compile.precompile \
+    --compile.precompile_artifact_dir /tmp/precompile_artifacts \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2 \
+    --training.seq_len 2048
+
+# Step 2: load and train with torchrun (uses all GPUs)
+torchrun --nproc_per_node=8 --virtual-local-rank \
```
bobrenjc93: I did consider deriving these flags from the precompiled artifact but decided against it for the following reasons:

1) The additional fingerprint check seems useful in the case when the CLI args at runtime are different than the ones used at precompile time.
2) Since some configs come from the artifact (model, parallelism, compile passes) but others come from the user (training steps, batch size, etc), we'd have to implement some sort of partial load + merge config system which makes the interface harder to reason about from a user POV.

yiming0416:
@claude review this PR

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -283,6 +283,42 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
...
+            "aot_llama3_precompile_full_inductor",
+            ngpu=8,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module graph_trainer.llama3",
+                    "--config graph_trainer_llama3_debugmodel",
+                    "--compile.mode aot",
+                    "--compile.passes regional_inductor",
```
yiming0416: regional_inductor only compiles flex attention regions, you'll need to use `--config graph_trainer_llama3_debugmodel_flex_attn`, otherwise it'll be a no-op

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -283,6 +283,42 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
...
+        # === Precompile tests ===
+        OverrideDefinitions(
+            [
+                [
+                    "--module graph_trainer.llama3",
+                    "--config graph_trainer_llama3_debugmodel",
+                    "--compile.mode aot",
+                    "--compile.passes full_inductor_compilation",
+                    "--compile.joint_passes inductor_decomposition",
+                    "--compile.precompile",
```
yiming0416: I wonder how the integration tests work here. 
Is `precompile_main.py` always required for the the precompile workflow? If so, how does the integration test run that?

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+
+    parallelism = config.parallelism
+    dp_replicate = parallelism.data_parallel_replicate_degree
+    dp_shard = parallelism.data_parallel_shard_degree
+    cp = parallelism.context_parallel_degree
+    tp = parallelism.tensor_parallel_degree
+    pp = parallelism.pipeline_parallel_degree
+
+    # dp_shard=-1 means "use remaining ranks" — but in single-process
```
yiming0416: Are all the comments necessary? Did you manually verify them? Any chance we can simplify them?

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,52 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_debugmodel \
+    --compile.mode aot \
+    --compile.passes full_inductor_compilation \
+    --compile.joint_passes inductor_decomposition \
+    --compile.precompile \
+    --compile.precompile_artifact_dir /tmp/precompile_artifacts \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2 \
+    --training.seq_len 2048
```
yiming0416: `llama3_debugmodel` by default has `training.seq_len=2048`

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,52 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+Stale artifacts are detected automatically via config fingerprinting and
+will raise an error at load time. Delete old artifacts and re-run
+precompile when upgrading PyTorch or changing the model/parallelism setup.
+
+```bash
+# Step 1: precompile on a single process (needs only 1 GPU)
+python -m torchtitan.experiments.graph_trainer.precompile_main \
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_debugmodel \
+    --compile.mode aot \
```
yiming0416: `compile.mode` is by default `aot`

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -38,17 +38,16 @@ class GraphTrainerCompileConfig(CompileConfig):
...
-    Enable serializable compilation. On first run, compiles with
-    serializable=True and saves the artifact. On subsequent runs, detects
-    the existing artifact and loads it, skipping compilation entirely.
+    Enable precompiled artifact loading. Requires running precompile_main.py
+    first to generate the artifact. At training time, loads the artifact
+    and skips compilation entirely.
     """
 
-    precompile_artifact_dir: str = "/tmp/precompile_artifacts"
+    precompile_artifact_dir: str = ""
```
aorenste, line 40: Consider killing off `precompile` and just have the singular `precompile_artifact_dir` where a `None` default means no precompile.

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -0,0 +1,249 @@
...
+
+    compile_config = config.compile
+
+    if not compile_config.precompile:
+        raise ValueError("precompile_main requires --compile.precompile true")
+    if not compile_config.precompile_artifact_dir:
+        raise ValueError(
+            "precompile_main requires --compile.precompile_artifact_dir to be set."
+        )
+    if not (_SERIALIZABLE_PASSES & set(compile_config.passes)):
```
aorenste, line 67: So what happens to the non-serializable passes? Did you mean to check that ALL the passes are serializable?

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -0,0 +1,249 @@
...
+            "--parallelism.data_parallel_shard_degree (not -1). "
+            "Set it to the value you will use during torchrun training."
+        )
+    world_size = dp_replicate * dp_shard * cp * tp * pp
+
+    logger.info(f"Initializing single-process precompile with world_size={world_size}")
+
+    # Fake backend produces correct collective output shapes without real
+    # communication, letting us trace distributed ops on a single process.
+    dist.init_process_group("fake", rank=0, world_size=world_size)
```
aorenste, line 98: Should we note that 0 isn't a bad choice but also (for now) rank MUST be 0 because virtual local rank will make all ranks pretend to be local rank 0?

torchtitan/experiments/graph_trainer/graph_utils.py
```diff
@@ -55,27 +61,41 @@ def export_joint(
...
+    import torch.distributed.config as dist_config
 
-    with tracing(tracing_context):
-        return (
-            aot_export_joint_with_descriptors_alone(gm, args, kwargs),
-            tracing_context,
-        )
+    old_coor = dist_config.compile_on_one_rank
+    if precompile:
+        dist_config.compile_on_one_rank = True
```
aorenste: This should definitely be a context manager. You can use it with either ExitStack or nullcontext to make it optional based on the flag.

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -0,0 +1,249 @@
...
+    get_compiler_passes_from_config,
+    get_joint_custom_passes_from_config,
+    joint_graph_builder,
+    make_compiler_with_passes,
+)
+from torchtitan.tools import utils
+from torchtitan.tools.logging import logger
+
+
+def main():
```
aorenste, line 54: There seems to be a lot of overlap between this an training main which could be shared.

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -283,6 +283,68 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
...
+                    "--compile.precompile_artifact_dir /tmp/precompile_test_full_inductor",
```
aorenste: Don't hardcode /tmp - use tempfile instead (here and below)

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: The basic idea is you need a pass that produces a serializable OutputCode, less that all passes need to follow some serializability constraint. I'll update to make this more obvious.

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: Since this will be quite an intrusive refactor I pulled it out into https://github.com/pytorch/torchtitan/pull/2743

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,47 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_debugmodel \
+    --compile.passes full_inductor_compilation \
+    --compile.joint_passes inductor_decomposition \
+    --compile.precompile_artifact_dir /tmp/precompile_artifacts \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2
+```
+
+Pre-compile works with any compiler pass that produces serializable output,
```
SherlockNoMad, line 117: except cudagraph pass.... 

cudagraph pass needs to be re-applied after loading the artifacts. 

Not sure if you have tested together with cudagraph, if not, we can do it in the next PR as a follow up.

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -0,0 +1,254 @@
...
```
SherlockNoMad, line 54: I suppose there are quite a bit of duplication with the trainer's main path... 
I worry it would be costly to maintain them in sync over time. 

For example, we are moving towards using the new minimal make_fx tracer directly tracing the entire fwd+loss+bwd graph, and the following optimization workflow would be different. 

An alternative implementation would be embedding the precompile logic into the trainer's main path. 
Think of it as an packaging step after all the optimization are applied. 

of cause, this is probably over symplified , as CooR would require different init setup, but I do think it's important to reuse the trainer path as much as possible, to reduce divergence risk. 

let me know what you think, and we should have an follow up discussion. I don't want this comment to be blocking tho, so we can work on this as a followup.

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -0,0 +1,254 @@
...
+            "precompile_main requires an explicit "
+            "--parallelism.data_parallel_shard_degree (not -1). "
+            "Set it to the value you will use during torchrun training."
+        )
+    world_size = dp_replicate * dp_shard * cp * tp * pp
+
+    logger.info(f"Initializing single-process precompile with world_size={world_size}")
+
+    # rank must be 0 because --virtual-local-rank maps every torchrun rank
+    # to local rank 0, so the precompiled artifact needs to match that setup.
```
SherlockNoMad, line 95: Naive Q about how CooR work with FSDP reshard? 

Rank 0 - takes first slice of weight
Rank 1 - takes second slice of wegiht
...

In graph, they show up as aten.split() and follows by getitem call. 
the index of getitme needs to be parameterized by rank? 

How do we address this?

bobrenjc93: See @aorenste's post here https://fb.workplace.com/groups/257735836456307/permalink/1071666715063211

> The fix is to modify the code so that we treat the rank as a SymInt (a symbolic integer that holds an expression rather than a concrete value) with a CustomOp to fetch the value for the current rank.

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: @aorenste had a similar concern. Here's a stab at it: https://github.com/pytorch/torchtitan/pull/2743. We can do the bikeshedding on that PR.

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: Correct. I'll follow up with cudagraphs. This just sets the baseline implementation.

torchtitan/experiments/graph_trainer/compile.py
```diff
@@ -133,37 +131,29 @@ def _apply_aot_compile(
...
-    if compile_config.precompile:
-        storage, artifact_key = _get_precompile_storage_and_key(compile_config)
+    config_fingerprint: ConfigFingerprint | None = None
+    if compile_config.precompile_artifact_dir:
+        from .precompile import compute_config_fingerprint
 
-        if storage.exists(artifact_key):
-            from .precompile import compute_config_fingerprint
+        storage, artifact_key = _get_precompile_storage_and_key(compile_config)
+        config_fingerprint = compute_config_fingerprint(
```
SherlockNoMad: artifact_key and config_fingerprint sounds mergeable? 

I am not sure about their difference tho.

bobrenjc93: Good catch. Now that we only have a single artifact (not N artifacts across N ranks) we can get rid of artifact_key entirely. Just updated the PR.

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,47 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_debugmodel \
+    --compile.passes full_inductor_compilation \
+    --compile.joint_passes inductor_decomposition \
+    --compile.precompile_artifact_dir /tmp/precompile_artifacts \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2
+
+# Step 2: load and train with torchrun (uses all GPUs)
+torchrun --nproc_per_node=8 --virtual-local-rank \
```
ezyang: Do you actually want `--virtual-local-rank` in the README here

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -69,6 +69,47 @@ MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --comp
...
+    -m torchtitan.train \
+    --module graph_trainer.llama3 \
+    --config graph_trainer_llama3_debugmodel \
+    --compile.passes full_inductor_compilation \
+    --compile.joint_passes inductor_decomposition \
+    --compile.precompile_artifact_dir /tmp/precompile_artifacts \
+    --parallelism.data_parallel_shard_degree 4 \
+    --parallelism.tensor_parallel_degree 2
```
ezyang, line 106: It is difficult to tell which of these flags are load bearing versus which ones can be dropped / changed.

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: I worry if we don't it will be a footgun since our graphs aren't actually device normalized without it. We could consider maybe turning the flag on by default in the torchtitan trainer? cc @aorenste

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: All of these flags are load bearing since they all feed into the config fingerprint.

bobrenjc93: See https://github.com/pytorch/torchtitan/pull/2713#discussion_r2995701682 for more details

torchtitan/experiments/graph_trainer/README.md
tianyu-l: what does this flag do? It's hard to tell from its pytorch docstring.

tests/integration_tests/__init__.py
```diff
@@ -24,6 +24,8 @@ class OverrideDefinitions:
     ngpu: int = 4
     disabled: bool = False
     skip_rocm_test: bool = False
+    pre_commands: Sequence[str] = ()
+    env_vars: Mapping[str, str] = field(default_factory=dict)
```
tianyu-l: Let's not hack existing no-pre-compile test workflow to adapt pre-compile. They are different enough to worth a separate `run_train.sh` and a separate `run_tests.py`. You can still reuse existing components.

run_train.sh
```diff
@@ -39,9 +39,14 @@ if [ -n "$COMM_MODE" ]; then
     NGPU="${NGPU}" LOCAL_RANK=0 python3 -m torchtitan.train --module ${MODULE} --config ${CONFIG} "$@" --comm.mode=${COMM_MODE} --training.steps 1
 else
     # Normal training with torchrun
+    VIRTUAL_LOCAL_RANK_ARG=""
```
tianyu-l: Let's not merge everything together, especially because full composability of pre-compile and `COMM_MODE`, `TORCHFT_LIGHTHOUSE` are tested.

Let's just have a dedicated run_train script inside graph_trainer, which you can easily refer to in graph_trainer/README.md.

In particular, we shouldn't need to define `VIRTUAL_LOCAL_RANK_ARG` envvar.

torchtitan/experiments/graph_trainer/README.md
bobrenjc93: It makes it so each rank sees cuda:0 by using CUDA_VISIBLE_DEVICES under the hood. We need this since torch.compile by default specializes on device id.

PR #2710, opened 2026-03-25
[RL] Turn compile for generator back on

Lucaskabela:
## Summary

We turned the compile for generator off in https://github.com/pytorch/torchtitan/pull/2638 due to conflict with DTensor and symbolic propogation

We fix this in https://github.com/pytorch/pytorch/pull/178210 so reenable this config (once landed in nightly)

## Test
```bash
python torchtitan/experiments/rl/simple_grpo_sum_digits.py --module rl --config rl_grpo_qwen3_0_6b --hf_assets_path=torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B
```

PR #2718, opened 2026-03-26
[RL][BE] Reapply Compilation in RL

Lucaskabela:
## Summary
This PR 
1) Reapplies https://github.com/pytorch/torchtitan/pull/2710

## Test plan
PREREQ: ensure https://github.com/pytorch/pytorch/pull/178210 is in your torch version
```bash
```
With both compiles on, we expect a 4x speedup over eager (timed from ~400s e2e to ~100s for 10 steps)

```bash
torchrun --nproc_per_node=2 \
      torchtitan/experiments/rl/tests/test_bitwise_identity.py
```

For numerics results in:
```
Trainer computed 30 token log-probs
  vLLM   log-probs[:5]: [-4.129570484161377, -1.795021891593933, -0.71578049659729, -0.2110116183757782, -0.9374725222587585]
  Trainer log-probs[:5]: [-4.129570484161377, -1.795021891593933, -0.71578049659729, -0.2110116183757782, -0.9374725222587585]
LOGPROB COMPARISON RESULTS
  Bitwise identical : True
  Tokens checked    : 30
  Tokens different  : 0
  Max delta         : 0.000000e+00
  Avg delta         : 0.000000e+00
  Diff mean         : 0.000000e+00
  Diff max          : 0.000000e+00
PASS: vLLM and trainer log-probs are bitwise identical.
/home/lucaskabela/.conda/envs/pytorch_build/lib/python3.10
```

### DIFF
--- a/torchtitan/experiments/rl/config_registry.py
+++ b/torchtitan/experiments/rl/config_registry.py
@@ -13,7 +13,7 @@ Each function returns a complete ``RLTrainer.Config`` and is discoverable by
 
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
-from torchtitan.config.configs import ParallelismConfig, TrainingConfig
+from torchtitan.config.configs import CompileConfig, ParallelismConfig, TrainingConfig
 from torchtitan.experiments.rl.actors.generator import (
     GeneratorCompileConfig,
     SamplingConfig,
@@ -41,13 +41,13 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            # compile=CompileConfig(enable=True, backend="aot_eager"),
+            compile=CompileConfig(enable=True, backend="aot_eager"),
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
             compile=GeneratorCompileConfig(
-                backend="none",
-                cudagraph_mode="none",
+                backend="eager",
+                cudagraph_mode="piecewise",
             ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=4,
@@ -80,13 +80,13 @@ def rl_grpo_qwen3_1_7b() -> RLTrainer.Config:
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            # compile=CompileConfig(enable=True, backend="aot_eager"),
+            compile=CompileConfig(enable=True, backend="aot_eager"),
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
             compile=GeneratorCompileConfig(
-                backend="none",
-                cudagraph_mode="none",
+                backend="eager",
+                cudagraph_mode="piecewise",
             ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=4,
@@ -119,12 +119,12 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
                 tensor_parallel_degree=1,
                 data_parallel_replicate_degree=1,
             ),
-            # compile=CompileConfig(enable=True, backend="aot_eager"),
+            compile=CompileConfig(enable=True, backend="aot_eager"),
         ),
         generator=VLLMGenerator.Config(
             compile=GeneratorCompileConfig(
-                backend="none",
-                cudagraph_mode="none",
+                backend="eager",
+                cudagraph_mode="piecewise",
             ),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=1,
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -42,33 +42,31 @@ class GeneratorCompileConfig:
...
     def __post_init__(self) -> None:
         if self.backend == "none" and self.cudagraph_mode in (
-            "piecewise",
-            "full_and_piecewise",
+            CUDAGraphMode.PIECEWISE,
+            CUDAGraphMode.FULL_AND_PIECEWISE,
         ):
             raise ValueError(
-                f"cudagraph_mode='{self.cudagraph_mode}' requires piecewise graph "
+                f"cudagraph_mode={self.cudagraph_mode!r} requires piecewise graph "
```
acisseJZhong: typo？

torchtitan/experiments/rl/tests/test_bitwise_identity.py
```diff
@@ -239,17 +240,17 @@ def _test_config() -> RLTrainer.Config:
...
             gpu_memory_limit=0.5,
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            # compile=GeneratorCompileConfig(
-            #     backend="eager", cudagraph_mode=CUDAGraphMode.FULL_AND_PIECEWISE
-            # ),
-            compile=GeneratorCompileConfig(backend="none", cudagraph_mode="none"),
+            compile=GeneratorCompileConfig(
+                backend="eager", cudagraph_mode=CUDAGraphMode.FULL_AND_PIECEWISE
```
acisseJZhong: curious how to decide to use PIECEWISE or FULL_AND_PIECEWISE?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -42,33 +42,31 @@ class GeneratorCompileConfig:
     NOTE: inductor will offer the best performance, but will impact numerics - use eager for
     bitwise identical results."""
 
-    cudagraph_mode: Literal[
```
tianyu-l, line 45: Fwiw torchtitan's style is using `Literal` than Enum, so I'd prefer we consistently work with `Literal` if possible.

Does `tyro` work with Enum?

Lucaskabela:
@claude does tyro work with enums here?

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Err I don't believe so - !r gets the repr of the enum right?

torchtitan/experiments/rl/tests/test_bitwise_identity.py
Lucaskabela: The difference is around full cudagraph capturability (related to custom attention implementation or not) and memory usage; fullgraph will use more memory so if we don't see much speedup probably better to just default to piecewise. Let me make this consistent here

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: I can change this back then... I thought enum would be preferred since this is already what is used by vLLM

tianyu-l: If this is only used for generator, that's fine, but the general rule should be "whenever torchtitan and vllm have style conflict, bias toward torchtitan".

If we are going to enable these mode for trainer as well, then I strongly think we should consistently use `Literal`.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -42,33 +42,31 @@ class GeneratorCompileConfig:
...
-        "none", "piecewise", "full", "full_and_piecewise"
-    ] = "piecewise"
+    cudagraph_mode: CUDAGraphMode = CUDAGraphMode.PIECEWISE
     """CUDA graph capture mode for vLLM.
     Piecewise capture supports dynamic sizes and splits cudagraphs around non capturable
       ops like attention
     Full capture captures one graph at the expense of less dynamism and requires full
       capturability
```
tianyu-l, line 52: should make these consistent

tianyu-l (approved):

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Hmm since tyro works with Enum, could we actually keep the enum here? I think for interoperability this may be better as if we have some change in vLLM it is downstreamed here for free

Lucaskabela: The trainer callout is a good consideration - let's revert back to literal then

acisseJZhong (approved):
lgtm!

torchtitan/experiments/rl/config_registry.py
```diff
@@ -80,13 +80,13 @@ def rl_grpo_qwen3_1_7b() -> RLTrainer.Config:
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
             ),
-            # compile=CompileConfig(enable=True, backend="aot_eager"),
+            compile=CompileConfig(enable=True, backend="aot_eager"),
```
wwwjn, line 83: Nit: why trainer side has a `enable` field but the generator side don't have it in `CompileConfig`? seems not very symmetrical

acisseJZhong: https://github.com/pytorch/torchtitan/pull/2568#discussion_r2970281777 lol

PR #2724, opened 2026-03-27
Refactor pipeline parallel helpers for graph PP reuse

sanketpurandare:
Stacked PRs:
 * #3090
 * #3089
 * #3088
 * #2727
 * #2726
 * #2725
 * __->__#2724

--- --- ---

### Refactor pipeline parallel helpers for graph PP reuse

Extract pipeline metadata, module splitting, and PP rank-to-stage mapping from pipeline_llm so graph PP can reuse the underlying setup logic without duplicating it. Add backward_requires_autograd to the schedule builder for graph PP, which runs explicit backward graphs instead of autograd. Existing eager PP behavior is unchanged.

How does this refactoring help?

This is the call chain:
  pipeline_llm
    -> _get_pipeline_metadata
    -> _generate_llm_fqn_per_model_part
    -> _build_get_mesh_callback
    -> _pipeline_module_split
         -> _get_pp_rank_to_stage_indices_mapping
         -> _split_module
         -> PipelineStage
    -> parallelize_fn
    -> _build_pipeline_schedule
    
 GraphPP will define:
   **graph_pipeline_llm**
    -> _get_pipeline_metadata
    -> _generate_llm_fqn_per_model_part
    -> _build_get_mesh_callback
    -> _pipeline_module_split
         -> _get_pp_rank_to_stage_indices_mapping
         -> _split_module
         -> **GraphPipelineStage**
    -> parallelize_fn
    -> _build_pipeline_schedule
    -> **GraphPPRunner**

So by just overriding  _pipeline_module_split but by reusing all the other helpers, we minimize code redundancy and maximize code reuse.

Keep pipeline_llm as the only public entrypoint exported by torchtitan.distributed.pipeline_parallel. Make build_pipeline_schedule, generate_llm_fqn_per_model_part, and pipeline_module_split private because they are implementation details with narrower contracts: schedule construction depends on the current PP config shape, LLM FQN generation encodes TorchTitan-specific module naming heuristics, and module splitting assumes models tolerate deleted or empty layer containers.

Update internal and experiment callsites to use the private helper names directly where reuse is still needed. This keeps the reusable code centralized while avoiding accidentally blessing those helpers as stable public API.

### DIFF
--- a/torchtitan/distributed/pipeline_parallel.py
+++ b/torchtitan/distributed/pipeline_parallel.py
@@ -36,12 +36,9 @@ from torchtitan.protocols.model_spec import ParallelizeFunction
 from torchtitan.protocols.module import ModuleDict, ModuleList
 from torchtitan.tools.logging import logger
 
-__all__ = [
-    "pipeline_llm",
-    "build_pipeline_schedule",
-    "generate_llm_fqn_per_model_part",
-    "pipeline_module_split",
-]
+# pipeline_llm is the public entrypoint for model-specific PP setup. Helpers in
+# this module are implementation details and should stay private.
+__all__ = ["pipeline_llm"]
 
 
 def _build_get_mesh_callback(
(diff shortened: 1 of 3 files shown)


torchtitan/distributed/pipeline_parallel.py
```diff
@@ -40,9 +39,12 @@
 
 __all__ = [
     "pipeline_llm",
+    "get_pipeline_metadata",
     "build_pipeline_schedule",
     "generate_llm_fqn_per_model_part",
     "pipeline_module_split",
+    "split_module",
+    "get_pp_rank_to_stage_indices_mapping",
 ]
```
tianyu-l: do all these fields need to be public and exposed?

tianyu-l (changes requested):
Please add more detailed description about why making these changes.

tianyu-l:
@claude review this PR

torchtitan/distributed/pipeline_parallel.py
sanketpurandare: Sorry this should have been marked as draft PR, it is not ready for review yet.

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -210,6 +233,8 @@ def build_pipeline_schedule(
     local_batch_size: int,
     stages: list[PipelineStage],
     loss_fn: Callable,
+    backward_requires_autograd: bool = True,
+    scale_grads: bool = False,
```
tianyu-l: why do you need to expose this? IMO we should deprecate this -- IIRC it's assuming all microbatches have the same number of valid tokens.

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -210,6 +233,8 @@ def build_pipeline_schedule(
     local_batch_size: int,
     stages: list[PipelineStage],
     loss_fn: Callable,
+    backward_requires_autograd: bool = True,
```
tianyu-l, line 227: curious what this field is used for

torchtitan/distributed/pipeline_parallel.py
tianyu-l: question not answered

sanketpurandare: I have updated the PR description to include the change description.

torchtitan/distributed/pipeline_parallel.py
sanketpurandare: scale_grads has been removed

torchtitan/distributed/pipeline_parallel.py
sanketpurandare: This is used when we want to run the backward actions for a microbatch without supplying the loss function to the pp runtime. Currently if we don't supply the loss function the pp_runtime will assume there is no backward. But in graph_pp the forward graph contains the loss function, so explicit loss function is not passed to the pp_runtime.

tianyu-l (approved):
sounds good

PR #2725, opened 2026-03-27
[graph_trainer] AutoParallel AOT FX Trace Integration

sanketpurandare:
Stacked PRs:
 * #3090
 * #3089
 * #3088
 * #2727
 * #2726
 * __->__#2725

--- --- ---

# TorchTitan AutoParallel GraphTrainer PR Summary

## High-Level Intent

This PR moves AutoParallel support out of the old standalone
`torchtitan.experiments.autoparallel` experiment and into GraphTrainer's
`aot_fx_trace` path. The new contract is:

1. AutoParallel solves SPMD placement and returns a placed/sharded module.
2. GraphTrainer traces the normal full train step around that placed module.
3. GraphTrainer applies the same `aot_fx_trace` memory policy, selective remat,
   CPU offload, bucketing, regional/full Inductor, and CUDA graph passes used by
   manually parallelized GraphTrainer models.

This is the cleaner design compared with treating AutoParallel as a separate
TorchTitan backend. AutoParallel is responsible for placement, not for owning the
whole compile stack.

## Files Removed

The PR deletes the standalone AutoParallel experiment and its stale CI:

- `.github/workflows/integration_test_8gpu_autoparallel.yaml`
- `torchtitan/experiments/autoparallel/**`

This is appropriate because the active integration is now under
`torchtitan/experiments/graph_trainer/`. The top-level experiments README also
removes the standalone AutoParallel row.

## Config and Validation

`torchtitan/experiments/graph_trainer/configs.py` adds:

- `GraphTrainerCompileConfig.enable_autoparallel: bool = False`
- `validate_autoparallel_config(...)`

The validation is narrow and non-intrusive: AutoParallel is only rejected when
enabled with a compile mode other than `aot_fx_trace`. This avoids touching the
main trainer and keeps AutoParallel-specific validation in GraphTrainer config
code.

## AutoParallel Graph Adapter

`torchtitan/experiments/graph_trainer/autoparallel_api.py` adds
`AutoParallelGraph`, a small subclass of AutoParallel with one supported entry
point:

- `apply_placement_for_fx_module(...)`

Important behavior:

- Calls AutoParallel's common placement path to obtain sharded params/buffers.
- Builds an inner AOT function via `aot_compile_joint_with_descriptors(...)`.
- Uses AutoParallel's default compiler function for both forward and backward.
- Reconstructs params and buffers from the wrapped module during forward, using
  local tensors for DTensor params so gradients flow back through the wrapped
  module's parameters.
- Supports both positional and keyword model inputs by flattening `(args,
  kwargs)` when positional-only flattening does not match the traced input
  arity.
- Optionally wraps a local output back into a DTensor for Llama loss-parallel
  vocab sharding.

## Llama3 Path

`torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py` adds
the Llama AutoParallel route.

Supported shape:

- FSDP + TP
- no DDP
- no CP
- no PP
- loss parallel required when TP is enabled

The path:

1. Builds the dense mesh from enabled dense axes.
2. Creates an AutoParallel input function with global batch size semantics.
3. Uses TorchTitan mixed precision settings.
4. Resolves `fsdp_reshard_after_forward` through the same helper used by the
   manual FSDP path.
5. Constrains input/output sharding.
6. Runs the AP solver.
7. Converts the local model output back to a TP-sharded DTensor for
   `loss_parallel`.
8. Passes the result into GraphTrainer `apply_compile(...)`.

This is a clean integration point: the model registry only branches on
`compile_config.enable_autoparallel`, and all substantial logic lives in the new
AutoParallel-specific parallelizer.

## DeepSeek V3 Path

`torchtitan/experiments/graph_trainer/deepseek_v3/parallelize_autoparallel.py`
adds the DeepSeek V3 AutoParallel route.

Supported shape:

- EFSDP + EP sparse mesh
- no DDP
- no CP
- no PP
- no TP
- loss parallel disabled

The path intentionally replaces TorchTitan's DSv3 module with
`autoparallel._testing.models.dsv3.DeepSeekV3Model`. This is the one remaining
known workaround in the PR. It is documented in the file: AutoParallel's
solver currently needs the local_map-based DSv3 definition and does not support
TorchTitan's token dispatcher op path yet.

The path preserves TorchTitan-facing MoE attributes needed by downstream
GraphTrainer and optimizer hooks after creating the placed AP model.

## Pass Interaction

The PR keeps AutoParallel composable with GraphTrainer's normal pass list.
The unit test `test_autoparallel_graph_pass_selection_uses_regular_memory_policy`
guards that AutoParallel does not remove:

- `tag_with_memory_policy_pass`
- `selective_activation_remat_pass`
- `apply_cpu_offload_pass`
- `joint_transformer_block_bucketing_reordering_pass`

This matches the intended contract: AP placement happens before full train-step
capture, then GraphTrainer owns memory policy and remat.

## Integration Tests and CI

The PR adds GraphTrainer AutoParallel integration suites:

- `graph_trainer_autoparallel`
  - Llama3 debug model, FSDP + TP, 4 GPUs
- `graph_trainer_autoparallel_h100`
  - DeepSeek V3 debug model, EFSDP + EP, 4 GPUs

CI changes:

- Regular GraphTrainer 8-GPU workflow installs AutoParallel, runs the Llama AP
  integration suite, and runs the Llama AP numerics test.
- H100 GraphTrainer workflow installs AutoParallel, runs the DSv3 AP integration
  suite with `NCCL_NVLS_ENABLE=0`, and runs the DSv3 AP numerics test with
  `NCCL_NVLS_ENABLE=0`.

This placement is sound: Llama stays in regular CI, while DSv3 AP is kept on the
H100 path because the DSv3 numerics/debug model requirements are H100-oriented.

## Numerics Tests

`test_numerics.py` adds close-numerics tests rather than bitwise tests for
AutoParallel:

- Llama3 AP vs eager, 4 GPUs
- DeepSeek V3 AP vs eager, 4 GPUs, `rtol=5e-4`

The close comparison is intentional. AutoParallel can choose different sharding,
collective schedules, operator decompositions, and reduction ordering from the
manual eager/GraphTrainer paths. The test still checks tight agreement by
extracting TensorBoard losses from `scripts/loss_compare.py` output and comparing
all recorded steps.

## Documentation

`torchtitan/experiments/graph_trainer/README.md` adds an
`Experimental AutoParallel Sharding` section.

## Cleanliness Assessment

- No trainer-level AutoParallel validation or core trainer branching.
- Model-specific AutoParallel logic is isolated under GraphTrainer model
  folders.
- Tests cover config validation, integration-suite registration, pass
  composition, FX-module path selection, and numerics.

Intentional caveats:

- DeepSeek V3 still depends on `autoparallel._testing.models.dsv3`. This is
  explicitly documented and is the only significant temporary workaround.
- AutoParallel numerics are close, not bitwise, because AP can choose different
  sharding and collective/reduction schedules.
- The AP graph adapter imports a few AutoParallel internal helpers
  (`_check_forward_args`, `_compute_expected_inputs`). This is acceptable for the
  experiment but should become a supported AutoParallel API before treating the
  integration as stable.

## Validation

- AutoParallel DSv3 equivalence passed with `rtol=5e-4` and `atol=1e-4`.
- AutoParallel pre-commit passed.
- TorchTitan Llama AutoParallel numerics passed.
- TorchTitan DeepSeek V3 AutoParallel numerics passed at `rtol=5e-4`.
- TorchTitan Llama AutoParallel integration passed.
- TorchTitan DeepSeek V3 AutoParallel H100 integration passed.


torchtitan/experiments/autoparallel/local_map_deepseek_v3/model.py
```diff
@@ -14,10 +14,14 @@
...
+        _DeepSeekV3Model.__init__(self, config)
+
+    def verify_module_protocol(self) -> None:
+        # Autoparallel submodules are standard nn.Modules,
+        # not torchtitan Module instances — skip the check.
+        pass
+
+
+# Wire Configurable pattern: build() calls DeepSeekV3Model(config=...)
+DeepSeekV3ModelArgs._owner = DeepSeekV3Model
```
tianyu-l: The "proper" way is to define `DeepSeekV3ModelArgs` inside `DeepSeekV3Model`, and you don't need this wiring.

tianyu-l (approved):

torchtitan/models/deepseek_v3/config_registry.py
```diff
@@ -57,6 +57,12 @@ def deepseek_v3_debugmodel() -> Trainer.Config:
     )
 
 
+def deepseek_v3_debugmodel_sdpa() -> Trainer.Config:
```
yiming0416: isn't the original `deepseek_v3_debugmodel` using sdpa?

SherlockNoMad (approved):

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -523,7 +523,9 @@ def _671b(
 
 deepseekv3_configs = {
     "debugmodel": _debugmodel,
+    "debugmodel_sdpa": partial(_debugmodel, attn_backend="sdpa"),
```
tianyu-l: why do you need this? can be done in config_registry

tianyu-l (changes requested):

xmfan (approved):

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -79,6 +86,56 @@ class GraphTrainerCompileConfig(CompileConfig):
...
+    if compile_config.joint_passes:
+        raise ValueError(
+            "--compile.joint_passes is not supported with "
+            "--compile.inductor_compilation autoparallel_backend."
+        )
+    if compile_config.precompile_artifact_dir:
+        raise ValueError(
+            "--compile.precompile_artifact_dir is not supported with "
+            "--compile.inductor_compilation autoparallel_backend."
+        )
```
SherlockNoMad: just use assert, more compact

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -79,6 +86,56 @@ class GraphTrainerCompileConfig(CompileConfig):
...
+            "--compile.inductor_compilation autoparallel_backend requires "
+            "--compile.enable_passes because the AutoParallel backend is "
+            "installed as a GraphTrainer graph pass."
+        )
+    if compile_config.passes:
+        raise ValueError(
+            "--compile.passes is not supported with "
+            "--compile.inductor_compilation autoparallel_backend."
+        )
+    if compile_config.joint_passes:
```
SherlockNoMad: joint_passes is ignored in GraphTrainer, do don't worry about this .

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -79,6 +86,56 @@ class GraphTrainerCompileConfig(CompileConfig):
...
+            "--compile.inductor_compilation autoparallel_backend requires "
+            "--compile.mode aot_fx_trace."
+        )
+    if not compile_config.enable_passes:
+        raise ValueError(
+            "--compile.inductor_compilation autoparallel_backend requires "
+            "--compile.enable_passes because the AutoParallel backend is "
+            "installed as a GraphTrainer graph pass."
+        )
+    if compile_config.passes:
```
SherlockNoMad: also ignored.

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -79,6 +86,56 @@ class GraphTrainerCompileConfig(CompileConfig):
...
+    return (
+        compile_config.autoparallel
+        and compile_config.inductor_compilation == AUTOPARALLEL_BACKEND
+        and compile_config.mode == "aot_fx_trace"
+    )
+
+
+def validate_autoparallel_backend_config(
+    compile_config: GraphTrainerCompileConfig,
+) -> None:
```
SherlockNoMad, line 106: just share a single error message, and config that it's support atm.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -211,6 +213,17 @@ def _copy_fwd_metadata_to_bw_nodes(fx_g: torch.fx.GraphModule) -> None:
     def _is_backward(node: torch.fx.Node) -> bool:
         return node.meta.get("autograd_backward", False)
 
+    def _metadata_score(node: torch.fx.Node) -> int:
```
SherlockNoMad: huh? this is hacky, let's not do this

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -214,6 +222,16 @@ def compile_time_passes(
     return passes
 
 
+def cleanup_passes() -> list[Callable]:
```
SherlockNoMad: nice refactor

SherlockNoMad: but I suspect you will have a lot of race condition landing this. 
let's do this in a small PR that land this refactor first.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -226,8 +244,19 @@ def construct_default_graph_passes(
     When ``precompile_artifact_dir`` is set, the artifact has graph
     transformed during precompile phase, so only cudagraph is returned.
     """
+    from torchtitan.experiments.graph_trainer.configs import (
+        is_autoparallel_backend_mode,
+    )
     from torchtitan.experiments.graph_trainer.cudagraph import is_cudagraph_compatible
 
+    if is_autoparallel_backend_mode(config.compile):
```
SherlockNoMad: @register_pass_suite(name="auto_parallel")
def you_pass_pipeline
... 


and pick your passes in trainer, replace construct_default_graph_passes

This registry infra doesn't exist yet, but I would appreciate one.

SherlockNoMad: let me know if you would prompt one... it can land first, since Syed also need one.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -889,45 +957,41 @@ def full_inductor_compilation_pass(
...
-    # execution path (gm(*flat_inputs)) works unchanged.
-    def _compiled_forward(*args):
-        return output_code(list(args))
+    functorch_patches, inductor_patches = get_autoparallel_backend_policy_helpers()
+    joint_custom_pass = functorch_patches.pop("joint_custom_pass", None)
 
-    gm.forward = _compiled_forward
-    return gm
+    if joint_custom_pass is not None:
+        gm = joint_custom_pass(gm, example_inputs)
```
SherlockNoMad: I think you would want to maintain your own pass pipeline. 

run joint_custom_pass explicitly, not as a inner step of inductor

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -889,45 +957,41 @@ def full_inductor_compilation_pass(
...
-    gm = _apply_decompositions(gm, example_inputs)
-    output_code = compile_fx_inner(gm, example_inputs)
+    from autoparallel.compile import get_autoparallel_backend_policy_helpers
 
-    # compile_fx_inner returns OutputCode with boxed calling convention
-    # (single list arg). Adapt to positional args so the graph trainer's
-    # execution path (gm(*flat_inputs)) works unchanged.
-    def _compiled_forward(*args):
-        return output_code(list(args))
+    functorch_patches, inductor_patches = get_autoparallel_backend_policy_helpers()
```
SherlockNoMad: let's have a comment on what does the patches do?

torchtitan/trainer.py
```diff
@@ -369,15 +369,17 @@ def __init__(self, config: Config):
...
-                )
+                parallelize_kwargs = {
+                    "parallel_dims": parallel_dims,
+                    "training": config.training,
+                    "parallelism": config.parallelism,
+                    "compile_config": config.compile,
+                    "ac_config": config.activation_checkpoint,
+                    "dump_folder": config.dump_folder,
+                }
+                if getattr(config.compile, "autoparallel", False):
```
SherlockNoMad: shouldn't need to change core?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -43,7 +47,10 @@ def fwd_bwd_step(
             for _, p in model.named_parameters(remove_duplicate=False)
             if p.requires_grad
         ]
-        grads = torch.autograd.grad(loss, params)
+        # Upstream remat uses this user annotation to identify the single
+        # logical backward region; GraphTrainer passes use autograd_backward.
+        with fx_traceback.annotate({"phase": "backward"}):
```
SherlockNoMad: don't need this

autograd_backward should do, if not, it's a tracer bug

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -64,6 +71,16 @@ def __init__(self, config):
...
 
+        if (
+            self.config.compile.autoparallel
+            and self.config.compile.mode != "aot_fx_trace"
+        ):
+            raise ValueError(
+                "AutoParallel graph_trainer integration only supports "
+                "--compile.mode aot_fx_trace"
+            )
+        validate_autoparallel_backend_config(self.config.compile)
```
SherlockNoMad: if self.config.compile.autoparallel: 
    validate_autoparallel_backend_config(self.config.compile)

all validation should fold into the function

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -580,15 +580,91 @@ def _build_async_tp_tests() -> list[OverrideDefinitions]:
...
+            ngpu=4,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module graph_trainer.llama3",
+                    "--config graph_trainer_llama3_debugmodel",
+                    "--compile.mode aot_fx_trace",
+                    "--compile.autoparallel",
+                    "--compile.inductor_compilation autoparallel_backend",
```
SherlockNoMad: I wonder how much delta autoparallel_backend have with full_inducotor? 

They looks mostly the same? maybe time to merge?

torchtitan/experiments/graph_trainer/llama3/__init__.py
```diff
@@ -15,6 +15,21 @@
 from .parallelize import parallelize_llama
 
 
+def _parallelize_fn(model, *, compile_config, loss_fn=None, **kwargs):
+    from torchtitan.experiments.graph_trainer.configs import GraphTrainerCompileConfig
+
+    if (
+        isinstance(compile_config, GraphTrainerCompileConfig)
```
SherlockNoMad: don't need isinstance(compile_config, GraphTrainerCompileConfig)

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize_autoparallel.py
```diff
@@ -0,0 +1,210 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+AutoParallel-based parallelization for DeepSeek V3.
```
SherlockNoMad, line 8: didn't review this file, I assume it's copied over.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+AutoParallel-based parallelization for Llama3.
```
SherlockNoMad, line 8: didn't review, assume code move

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
```diff
@@ -0,0 +1,226 @@
+## GraphTrainer AutoParallel
+
+GraphTrainer can use AutoParallel to solve model placement for `aot_fx_trace`
+training. AutoParallel chooses placements for the model, while GraphTrainer
+still owns the forward, loss, and backward trace through `make_fx`.
```
SherlockNoMad: official name is minimal_fx_tracer

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
```diff
@@ -0,0 +1,226 @@
...
+
+Two execution modes are supported:
+
+- **Native GraphTrainer backend mode**: AutoParallel places the model, then
+  GraphTrainer runs its normal `aot_fx_trace` pass and compile pipeline.
+- **AutoParallel backend mode**: AutoParallel places the model, then
+  GraphTrainer applies AutoParallel's exported backend policy helpers to the
+  traced train-step graph before full Inductor compilation.
```
SherlockNoMad: do you think we need to keep two?

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
```diff
@@ -0,0 +1,226 @@
...
+    --compile.mode aot_fx_trace \
+    --compile.autoparallel \
+    --parallelism.data_parallel_shard_degree 2 \
+    --parallelism.tensor_parallel_degree 2
+```
+
+In this mode:
+
+- AutoParallel solves and applies model placement.
+- GraphTrainer traces forward, loss, and backward with `make_fx`.
```
SherlockNoMad: minimal_fx_tracer

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
```diff
@@ -0,0 +1,226 @@
...
+    --gpu_arch_type cuda \
+    "$RUNNER_TEMP/artifacts-to-be-uploaded/autoparallel" \
+    --ngpu 4
+```
+
+When changing placement, tracing, graph passes, or backend policy behavior,
+compare numerics between native GraphTrainer backend mode and AutoParallel
+backend mode with `--debug.seed=42` and `--debug.deterministic`.
+
+### Current Limitations
```
SherlockNoMad: what about cudagraph?

torchtitan/experiments/graph_trainer/compile.py
```diff
@@ -232,6 +235,7 @@ def apply_compile(
         parallel_dims: Parallel dimensions
         dump_folder: Folder for dumping debug graphs
     """
+    validate_autoparallel_backend_config(compile_config)
```
SherlockNoMad: this path is not run in aot_fx_trace path, don't need changes in this file. 

we should deprecate this latter.

torchtitan/experiments/graph_trainer/deepseek_v3/__init__.py
```diff
@@ -16,6 +16,21 @@
 from .parallelize import parallelize_deepseekv3
 
 
+def _parallelize_fn(model, *, compile_config, loss_fn=None, **kwargs):
+    from torchtitan.experiments.graph_trainer.configs import GraphTrainerCompileConfig
+
+    if (
+        isinstance(compile_config, GraphTrainerCompileConfig)
```
SherlockNoMad: isinstance(compile_config, GraphTrainerCompileConfig)

 don't need this check.

SherlockNoMad (changes requested):

torchtitan/experiments/graph_trainer/configs.py
sanketpurandare: Stale after the latest rewrite. The `autoparallel_backend` validation path was removed entirely. The remaining AutoParallel validation is a single user-facing `ValueError` for unsupported `--compile.mode`, which should stay as `ValueError` rather than `assert`.

torchtitan/experiments/graph_trainer/configs.py
sanketpurandare: Stale after the latest rewrite. The backend-mode validation that checked `joint_passes` was removed.

torchtitan/experiments/graph_trainer/configs.py
sanketpurandare: Stale after the latest rewrite. The backend-mode validation that checked `passes` was removed.

torchtitan/experiments/graph_trainer/configs.py
sanketpurandare: Addressed. The current validation is one compact support check: AutoParallel only supports `--compile.mode aot_fx_trace`. The user-facing flag is now `--compile.enable_autoparallel`.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
sanketpurandare: Addressed. Agreed this was too hacky for this PR. I removed the `_metadata_score()` change and the associated unit test; `make_fx_tracer.py` is no longer touched by this commit.

torchtitan/experiments/graph_trainer/passes.py
sanketpurandare: Thanks. I ended up removing this unrelated refactor from this PR to reduce landing risk and keep the AutoParallel change focused.

sanketpurandare: Agreed. I removed the unrelated pass refactor from this PR. The cleaned AutoParallel commit no longer adds `cleanup_passes`, the memory-policy helper extraction, or the full-Inductor helper extraction.

torchtitan/experiments/graph_trainer/passes.py
sanketpurandare: Stale after the rewrite. AutoParallel no longer owns a special GraphTrainer pass pipeline; it now uses the regular `aot_fx_trace` pass stack. A pass-suite registry still sounds useful, but I think it should be a separate infra PR.

sanketpurandare: Agreed on keeping this separate. Since the AutoParallel-specific pass pipeline was removed, I am not adding registry infra in this PR.

torchtitan/experiments/graph_trainer/passes.py
sanketpurandare: Stale after the rewrite. The AutoParallel backend path was removed. Current design intentionally lets GraphTrainer run its normal `aot_fx_trace` pass stack after AutoParallel placement.

torchtitan/experiments/graph_trainer/passes.py
sanketpurandare: Stale after cleanup. The temporary patch/debug machinery this referred to was removed.

torchtitan/trainer.py
sanketpurandare: Addressed. The current AutoParallel commit does not modify core `torchtitan/trainer.py`.

torchtitan/experiments/graph_trainer/trainer.py
sanketpurandare: Addressed. The hacky backward annotation/wrapper path was removed. The current code relies on the normal `autograd_backward` metadata.

torchtitan/experiments/graph_trainer/trainer.py
sanketpurandare: Mostly stale after removing the backend path. The trainer now calls a simple `validate_autoparallel_config()` directly, and the flag has been renamed to `enable_autoparallel`.

torchtitan/experiments/graph_trainer/tests/integration_tests.py
sanketpurandare: Stale after the rewrite. `autoparallel_backend` was removed, so there is no separate backend delta to compare or merge with full Inductor.

torchtitan/experiments/graph_trainer/llama3/__init__.py
sanketpurandare: Addressed. Removed the `isinstance(compile_config, GraphTrainerCompileConfig)` guard. The registry now checks `compile_config.enable_autoparallel` directly.

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize_autoparallel.py
sanketpurandare: Acknowledged. This remains GraphTrainer-specific glue rather than a pure code move. The only intentional temporary hack left is the dependency on `autoparallel._testing.models.dsv3`, documented in the file, because AutoParallel currently requires the local_map-based DSv3 definition.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
sanketpurandare: Acknowledged. No change requested here; the Llama file remains the GraphTrainer AutoParallel placement glue.

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
sanketpurandare: Addressed. The README now describes the integration as using the `minimal_fx_tracer`-based `aot_fx_trace` flow.

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
sanketpurandare: Addressed by removing the second mode. The AutoParallel backend mode is gone; the integration now uses regular GraphTrainer `regional` / `full` compile choices after AutoParallel placement.

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
sanketpurandare: Addressed. The README now uses the official `minimal_fx_tracer` name.

torchtitan/experiments/graph_trainer/AUTOPARALLEL_README.md
sanketpurandare: Addressed. Added a README note that AutoParallel does not add a separate cudagraph mode; the placed model follows GraphTrainer's regular cudagraph compatibility checks and default `aot_fx_trace` cudagraph behavior.

torchtitan/experiments/graph_trainer/compile.py
sanketpurandare: Addressed. Removed the `validate_autoparallel_config()` import/call from `compile.py`; `compile.py` is no longer touched by this commit. Validation remains in `GraphTrainer.__init__` and the model-specific AutoParallel parallelizers.

torchtitan/experiments/graph_trainer/deepseek_v3/__init__.py
sanketpurandare: Addressed. Removed the `isinstance(compile_config, GraphTrainerCompileConfig)` guard. The registry now checks `compile_config.enable_autoparallel` directly.

torchtitan/experiments/graph_trainer/tests/test_numerics.py
```diff
@@ -152,6 +233,56 @@ def _run_qwen3_moe_loss_compare(test_options_extra: str = "") -> bool:
...
 
+AUTOPARALLEL_LLAMA3_PARALLELISM = (
+    "--parallelism.data_parallel_shard_degree=2"
+    " --parallelism.tensor_parallel_degree=2"
+)
+
+
+def _run_autoparallel_llama3_loss_compare() -> bool:
+    """Run loss_compare for eager llama3 vs graph_trainer AutoParallel llama3."""
+    return run_loss_compare_close(
```
SherlockNoMad, line 244: do we understand why it's not bitwise match?

torchtitan/experiments/graph_trainer/tests/test_numerics.py
```diff
@@ -152,6 +233,56 @@ def _run_qwen3_moe_loss_compare(test_options_extra: str = "") -> bool:
...
+        test_config="graph_trainer_deepseek_v3_debugmodel_ep",
+        baseline_options=AUTOPARALLEL_DSV3_PARALLELISM,
+        test_options=(
+            f"{AUTOPARALLEL_DSV3_PARALLELISM}"
+            " --compile.mode aot_fx_trace"
+            " --compile.enable_autoparallel"
+        ),
+        baseline_ngpus=4,
+        test_ngpus=4,
+        rtol=5e-4,
```
SherlockNoMad, line 282: I can imagine, if autoparallel ended up with a different placement, it would incur different communication, and different reduction, which could change numercis? 

Can we produce 2 tlparse for the same model, one with original parallel, one with autoparallel, to understand better? 

Also, we should caution in the md file, that autoparallel would change numerics.

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize_autoparallel.py
```diff
@@ -0,0 +1,212 @@
...
+
+Uses AutoParallelGraph to apply solver-based SPMD sharding on AutoParallel's
+local_map DSv3 model (whose ops the solver supports), then feeds the pre-built
+JWD into graph_trainer's AOT pipeline. Requires a 2D sparse mesh (EFSDP+EP).
```
SherlockNoMad: I would expect somewhere in the file, this 2D sparse mesh is asserted.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
```diff
@@ -0,0 +1,164 @@
...
+
+Uses AutoParallelGraph to apply solver-based SPMD sharding, then feeds
+the pre-built JointWithDescriptors into graph_trainer's AOT compilation
```
SherlockNoMad: no long using JointWithDescriptors, update docstring.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
```diff
@@ -0,0 +1,164 @@
...
+pipeline (CompiledModule + joint_graph_builder).
```
SherlockNoMad: stale comment

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
```diff
@@ -0,0 +1,164 @@
...
+    graph_trainer's aot_fx_trace path.
+    """
+    validate_autoparallel_config(compile_config)
+
+    if parallel_dims.dp_replicate_enabled:
+        raise ValueError("AutoParallel Llama3 does not support DDP yet")
+    if parallel_dims.cp_enabled:
+        raise ValueError("AutoParallel Llama3 does not support CP yet")
+    if parallel_dims.pp_enabled:
+        raise ValueError("AutoParallel Llama3 does not support PP yet")
```
SherlockNoMad, line 63: nice. I don't remember see similar assertion in dsv3？

torchtitan/experiments/graph_trainer/README.md
```diff
@@ -2,12 +2,12 @@
...
 
-
 This experiment demonstrates graph-based distributed training in torchtitan through toolkit-style usage of PyTorch's compiler technologies. The goal is to give users explicit control over the compiler stack in terms of performance, numerics, and debuggability during large-scale distributed training. See the [Manifesto](MANIFESTO.md) for the motivation and design philosophy behind GraphTrainer.
 
 **Key features:**
 - **Full train step graph capture** — `make_fx`-based `minimal_fx_tracer` traces forward + loss + backward (and optionally `optimizer.step`) into a single FX graph, without AOTAutograd partitioning, giving full visibility and control over the entire computation.
 - **[SimpleFSDP](https://arxiv.org/abs/2411.00284)** — A compiler-based FSDP that represents sharding as parameterized collectives within the computation graph, making it fully tracer-friendly while achieving memory and throughput improvements over eager FSDP2.
+- **AutoParallel** — Automatic model placement for supported `aot_fx_trace` models, while keeping the normal GraphTrainer graph pass pipeline.
```
SherlockNoMad: Given it doesn't support CP on dense, and only support EFSDP+EP for sparse. Do you mind marking this as experimental？and move it as a latter bullet? 

Sry I know this might be an offensive ask, but I need to keep the Key Feature section crisp. This is our user landing page, and I have to gate this. 

After all, this is for power user!

SherlockNoMad (approved):
thanks a lot for merging the experiment, I am really happy that it can fully compose as a parallel module fn!

SherlockNoMad:
fyi, also need to manually tag ciflow/h100.8 to trigger H100 CI.

torchtitan/experiments/graph_trainer/tests/test_numerics.py
sanketpurandare: Yes. The reasons are similar to what we root-caused for DeepSeek V3. The AutoParallel path can produce a different SPMD graph than the manually parallelized eager path: it may choose different sharding, introduce different collective schedules/reductions, and lower some operators differently before GraphTrainer sees the graph. For DSv3, the strongest concrete arithmetic mismatch is RMSNorm: native GraphTrainer preserves fused RMSNorm while AP decomposes it before Inductor. The AP graph also has a different collective surface because it materializes the solver-selected placement. I added the same caveat to the README and gathered detailed evidence in the [DSv3 AP numerics analysis](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/f8ace34b-e300-42c6-84df-0120d546e333/custom/analysis.md?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000).

torchtitan/experiments/graph_trainer/tests/test_numerics.py
sanketpurandare: Done. I produced tlparse reports for the same DSv3 config with [native GraphTrainer parallelization](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/30730055-a956-4e1e-9c1a-8a4a4a76d440/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000) and with [AutoParallel enabled](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/d1549434-dea6-4efe-b71a-b9ea7174802e/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000). I also uploaded the full [DSv3 AP numerics analysis](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/f8ace34b-e300-42c6-84df-0120d546e333/custom/analysis.md?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000).

Summary: native GraphTrainer is bitwise exact against eager for this config, while AP differs by `-3.8147e-06` in first-step loss. The strongest root cause is RMSNorm: native GraphTrainer preserves fused RMSNorm, while AP decomposes it before Inductor. AP also has a different collective surface because it materializes the solver-selected SPMD placement; rank0 has more all-gather/reduce-scatter nodes and integer token-count all-reduces from local-map MoE dispatch. I updated the README wording to caution that AP may choose different sharding, collective schedules, and operator lowerings, so numerics tests use tight close checks rather than strict bitwise equality.

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize_autoparallel.py
sanketpurandare: Addressed. DSv3 now explicitly requires the `("efsdp", "ep")` axes, rejects missing axes, and validates that the resulting sparse mesh is exactly 2D with those mesh axis names.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
sanketpurandare: Addressed. The Llama AutoParallel module docstring no longer mentions `JointWithDescriptors`; it now describes the placed-module contract and the normal GraphTrainer `aot_fx_trace` train-step pipeline.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
sanketpurandare: Addressed. Removed the stale wording about feeding a pre-built `JointWithDescriptors` into GraphTrainer.

torchtitan/experiments/graph_trainer/llama3/parallelize_autoparallel.py
sanketpurandare: Added the DSv3 checks. It now rejects DDP/CP/PP/TP for this path, requires both EFSDP and EP axes, and validates that the sparse mesh is exactly `("efsdp", "ep")`.

torchtitan/experiments/graph_trainer/README.md
sanketpurandare: Agreed. I removed AutoParallel from the key-feature list and moved it into a later `Experimental AutoParallel Sharding` section so the landing page keeps the core GraphTrainer features first. The dedicated section still calls out the supported usage and the numerics caveat.

torchtitan/experiments/graph_trainer/tests/test_numerics.py
sanketpurandare: # DeepSeek V3 GraphTrainer AutoParallel Numerics Report

## Artifacts

- Non-AutoParallel Manifold tlparse:
  https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/30730055-a956-4e1e-9c1a-8a4a4a76d440/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000
- AutoParallel Manifold tlparse:
  https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/d1549434-dea6-4efe-b71a-b9ea7174802e/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000


## Commands

Both runs used 4 GPUs, `NCCL_NVLS_ENABLE=0`, one step, and the same DP/EP shape:

```bash
--parallelism.expert_parallel_degree=2 \
--parallelism.disable_loss_parallel
```

The non-AutoParallel test path added:

```bash
--compile.mode aot_fx_trace
```

The AutoParallel test path added:

```bash
--compile.mode aot_fx_trace --compile.enable_autoparallel
```

## Numerics Result

Non-AutoParallel GraphTrainer DSv3 vs eager is bitwise exact:

```text
baseline: 8.087531089782715
test:     8.087531089782715
diff:     0.0
```

AutoParallel GraphTrainer DSv3 vs eager is close, but not bitwise:

```text
baseline: 8.087531089782715
test:     8.08752727508545
diff:    -3.814697265625e-06
```

## Primary Root Cause

The first concrete arithmetic mismatch is RMSNorm lowering, before Inductor codegen.

The non-AutoParallel GraphTrainer graph preserves fused RMSNorm:

```text
aten._fused_rms_norm.default:          38
aten._fused_rms_norm_backward.default: 19
aten.pow.Tensor_Scalar:                0
aten.mean.dim:                         0
aten.rsqrt.default:                    0
```

The AutoParallel GraphTrainer graph decomposes RMSNorm into primitive arithmetic:

```text
aten._fused_rms_norm.default:          0
aten._fused_rms_norm_backward.default: 0
aten.pow.Tensor_Scalar:                38
aten.mean.dim:                         38
aten.rsqrt.default:                    38
```

This comes from AutoParallel graph construction/lowering using
`torch._inductor.decomposition.select_decomp_table()`. That decomposition table
contains decompositions for `aten.rms_norm.default` and
`aten._fused_rms_norm.default`, so the AP graph is already arithmetically
different before GraphTrainer codegen or Inductor can be blamed.

A standalone check confirms the two RMSNorm forms are not bitwise equivalent:

```text
equal   False
max_abs 0.015625
num_diff 11
```

So the current AP graph cannot be bitwise equal to eager TorchTitan DSv3 unless
AutoParallel preserves the fused RMSNorm op or otherwise exactly matches its
arithmetic.

## Secondary Graph Differences

Counting only the first unprofiled generated function body in rank0 FX:

```text
                     non-AP   AP
all_gather              25    61
reduce_scatter          13    68
all_reduce               0    10
all_to_all              60    55
aten._softmax           10     0
aten.exp                 0    60
```

The extra collectives are expected from the AP contract being different from
TorchTitan's native DSv3 parallelization:

1. AutoParallel solves a full SPMD placement over the 2D `(efsdp, ep)` mesh,
   then materializes that placement as local tensor code plus explicit
   collectives. Native GraphTrainer follows TorchTitan's existing SimpleFSDP +
   EP program and then captures that program.

2. AP introduces more parameter/input redistributions than native SimpleFSDP.
   In the generated AP graph, additional all-gather chains appear around AP
   local tensor boundaries and AP-selected placements. GraphTrainer's FSDP
   overlap pass then rewrites/buckets those chains, which is why the AP graph
   has substantially more all-gathers and waits.

3. AP introduces many more reduce-scatters because AP placement can produce
   partial results/gradients over mesh axes that must be reduced back into the
   solved local layouts. Native GraphTrainer only has the SimpleFSDP/EP gradient
   reduction schedule from the original TorchTitan program.

4. The AP `all_reduce` nodes are integer token-count reductions in the local-map
   DSv3 MoE path, not dense matmul reductions. Representative graph lines point
   at `autoparallel/collectives.py:_all_reduce` and the local-map DSv3 token
   dispatch path:

   ```text
   all_reduce: "i64[8]" = torch.ops._c10d_functional.all_reduce.default(histc, 'sum', '33')
   File: autoparallel/collectives.py:130 in _all_reduce
   ```

   These are used to compute global/per-group expert token counts for local-map
   dispatch. TorchTitan's native DSv3 token dispatcher does not express that
   same token-count exchange as standalone `all_reduce` nodes in the captured
   graph.

5. AP also decomposes router softmax. Native GraphTrainer preserves
   `aten._softmax.default`; AP emits `amax/exp/sum/div`. In the standalone check
   for float32 softmax over the router-sized tensor this was bitwise equal for
   random input, so RMSNorm is the stronger concrete explanation for the
   observed first-step loss delta. Still, this is another place where AP is not
   preserving the same operator surface as TorchTitan eager.

## Conclusion

The AP path is not currently a bitwise-preserving
rewrite of TorchTitan DSv3. The root cause is not Inductor. The AP graph differs
before Inductor because AutoParallel uses Inductor decompositions while building
and lowering the AP graph, and RMSNorm is decomposed into primitive arithmetic
that is not bitwise equivalent to TorchTitan/eager's fused RMSNorm.

The remaining collective schedule differences should be evaluated
separately; those affect distributed ordering and may still prevent strict
bitwise equivalence if AP is allowed to choose a different SPMD schedule than
TorchTitan native parallelization.

tianyu-l (approved):

torchtitan/experiments/graph_trainer/tests/test_numerics.py
SherlockNoMad: Thanks for the investigation. 
I think bitwise different can be addressed by forcing empty decomp table when calling aot_export_joint_with_descriptor in AP. 

I am not sure if AP solver would prefer a decomposed graph tho？ 
Having a bitwise identity is just a very nice property, coz it will make it the numerical doubts of an optimization goes away.

sanketpurandare: AutoP applies decompositions after its first trace even before creating the optimization problem. Even if we do this, the collectives based reduction may not be same due to different placement strategies and Auto-P operating on a finer mesh granularity. @fmassa @xmfan

PR #2726, opened 2026-03-27
[graph_trainer][aot_fx_trace][graph_pp] Add stage graph partitioning

### LINKED ISSUE #3780: [graph_trainer] Graph-PP RFC
# GraphPP GraphTrainer RFC

## TLDR

GraphPP is the GraphTrainer execution path for pipeline-parallel training. It
keeps upstream PyTorch PP schedules responsible for microbatch splitting, stage
metadata initialization, action ordering, P2P communication, loss plumbing, and
eval semantics. It replaces each stage's eager autograd compute with explicit
GraphTrainer FX callables.

The implementation has three layers:

```text
graph_pipeline_llm
  -> build normal TorchTitan/torch.pipelining stages and schedule

GraphTrainerStageGraphProvider
  -> trace one joint stage-local train-step graph per local stage
  -> run GraphTrainer pre-partition passes
  -> extract forward/backward/FSDP/dI-dW/overlap callables
  -> compile extracted callables with boxed execution

GraphPipelineRuntime
  -> register custom handlers on the upstream runtime PP schedule
  -> map FORWARD/BACKWARD/FSDP/OVERLAP actions to graph executor calls
  -> move explicit values through PP caches and per-stage runtime state
```

The key contract is that `GraphPipelineRuntime` does not inspect GraphTrainer
graph metadata. It supplies explicit runtime values to `GraphPPStageGraphs`;
the GraphTrainer-owned executor packs those values according to private graph
metadata and returns the values the PP schedule needs.

Current conservative scope:

```text
- aot_fx_trace only
- upstream PP owns microbatch splitting
- no eager pre-forward for GraphPP trace inference
- extracted graphs use boxed execution
- CUDA graph compilation is disabled with a warning
- FSDP edge work remains exposed through UNSHARD and REDUCE_GRAD actions
- split BlockMask and multiplex ShapeEnv handling are local shims with linked
  upstream PyTorch issues
```

Section index:

```text
Status And Scope              support matrix and goals
Architecture And Files        ownership, files, public APIs
Entry Point                   graph_pipeline_llm construction and validation
Runtime                       registered actions, caches, state, grad flow
Provider And Stage Contracts  graph provider lifecycle and protocols
Tracing                       microbatch ownership and traced stage_step bodies
Partition                     GraphMeta, calling convention, extraction
Graph Construction            pass order, executor metadata, compilation
FSDP Split                    unshard and reduce-grad graph extraction
dI/dW Split                   backward-input/backward-weight extraction
Overlap Graphs                OVERLAP_F_B multiplexing and metadata transfer
Runtime Value Details         boxed calls, wrapping, BlockMask, loss, eval
Validation                    explicit invariants and no-op cases
Numerics                      bitwise target and comparison caveats
Limitations                   unsupported modes
Review Invariants             reviewer-facing design checks
Future Work                   upstream cleanup and planned extensions
```

## Status And Scope

This document describes the current implementation under
`torchtitan/experiments/graph_trainer/graph_pp`. It is an implementation
contract: names, calling conventions, pass boundaries, and runtime ownership
should match the code.

Supported mode:

```text
+-----------------------------+----------------------------------------------+
| Dimension                   | Current status                               |
+-----------------------------+----------------------------------------------+
| Entry point                 | graph_pipeline_llm                           |
| Runtime wrapper             | GraphPipelineRuntime                         |
| Stage wrapper               | GraphPipelineStage                           |
| Compile mode                | aot_fx_trace only                            |
| Schedule family             | _PipelineScheduleRuntime schedules            |
| Graph construction timing   | lazy on first schedule step                  |
| PP microbatch splitting     | upstream torch.pipelining                    |
| Final Inductor placement    | after GraphPP extraction passes              |
| Graph execution             | boxed calls only                             |
| CUDA graphs                 | disabled with a warning                      |
| Precompiled artifacts       | unsupported                                  |
| FSDP reshard_after_forward  | "always" unsupported                         |
+-----------------------------+----------------------------------------------+
```

Design goals:

```text
1. Reuse upstream PP for schedules, P2P, microbatching, metadata, and losses.
2. Keep runtime and graph construction decoupled through protocols.
3. Run metadata-preserving GraphTrainer passes before GraphPP extraction, then
   run final Inductor after extraction.
4. Preserve minimal_fx_tracer tensor-subclass wrapping semantics.
5. Validate graph invariants loudly instead of silently skipping errors.
```

GraphPP owns:

```text
- stage-local train-step tracing
- splitting traced graphs into PP action callables
- FSDP unshard/reduce-grad extraction
- full-backward to dI/dW extraction when possible
- multiplexing forward and full-backward graphs for OVERLAP_F_B
- boxed compilation and execution of extracted graph modules
- mapping upstream schedule actions to graph calls
- accumulating graph-produced grads into live parameters
```

GraphPP does not own:

```text
- full-batch input/target splitting
- upstream PP stage metadata inference
- action ordering or P2P action construction
- model-specific parallelize_fn policy
- core TorchTitan loss definitions
- optimizer, grad clipping, checkpointing, metrics, or logging
```

## Architecture And Files

End-to-end shape:

```text
TorchTitan GraphTrainer
  -> graph_pipeline_llm
  -> split model and build GraphPipelineStage objects
  -> build upstream _PipelineScheduleRuntime schedule
  -> register_graph_pp_schedule
  -> GraphPipelineRuntime
       -> GraphTrainerStageGraphProvider builds/attaches graphs
       -> upstream schedule.step emits actions
       -> registered GraphPP handlers call graph executors
```

File map:

```text
pipeline.py
  graph_pipeline_llm and GraphPP config validation

stage.py
  GraphPipelineStage, GraphPPStageRuntimeState, GraphPPStageGraphs,
  GraphPPOverlapGraphs, GraphPPStageGraphsProvider

runner.py
  GraphPipelineRuntime and register_graph_pp_schedule

graph_builder.py
  GraphTrainerStageGraphProvider, GraphTrainerStageGraphs, tracing,
  extraction orchestration, boxed compilation

partition.py
  partition_joint_graph and public GraphMeta

split_fsdp_collectives.py
  split_forward_fsdp_collectives and split_backward_fsdp_collectives

split_di_dw.py
  split_di_dw_graph

graph_multiplex.py
  multiplex_fw_bw_graph

utils.py
  shared graph helpers, wrapping helpers, BlockMask normalization,
  trace logging, ShapeEnv metadata transfer
```

Public graph-pass API exported by `graph_pp/__init__.py`:

```text
GraphMeta
partition_joint_graph
GraphPPFSDPForwardSplit
GraphPPFSDPBackwardSplit
split_forward_fsdp_collectives
split_backward_fsdp_collectives
GraphPPDiDwSplit
split_di_dw_graph
multiplex_fw_bw_graph
```

The main user-facing entry point is `graph_pipeline_llm` in `pipeline.py`.
Runtime registration is `register_graph_pp_schedule(schedule, graph_provider=...)`.

## Entry Point

`graph_pipeline_llm` validates config, constructs normal PP stage containers,
builds the upstream runtime schedule, then registers GraphPP handlers:

```text
graph_pipeline_llm(...)
  -> _validate_graph_pp_config(...)
  -> pp_mesh = parallel_dims.get_mesh("pp")
  -> _get_pipeline_metadata(...)
  -> choose module_names_per_stage
       parallelism.module_fqns_per_model_part if provided
       otherwise _generate_llm_fqn_per_model_part(...)
  -> _get_pp_rank_to_stage_indices_mapping(...)
  -> for each local stage index:
       _split_module(...)
       parallelize_fn(model_part, ...)
       GraphPipelineStage(model_part, stage_index, num_stages, device, group)
  -> _build_pipeline_schedule(..., backward_requires_autograd=False)
  -> GraphTrainerStageGraphProvider(loss_fn, compile_config, model_config,
                                    parallelism)
  -> register_graph_pp_schedule(schedule, graph_provider=provider)
  -> return (GraphPipelineRuntime, model_parts, has_first_stage, has_last_stage)
```

`backward_requires_autograd=False` is required because backward work is
represented by explicit graph callables, not eager autograd attached to stage
outputs.

Validation rejects:

```text
- compile.mode != "aot_fx_trace"
- compile.precompile_artifact_dir is set
- schedule class is not a _PipelineScheduleRuntime subclass
- parallelism.fsdp_reshard_after_forward == "always"
```

If CUDA graph pass is effectively enabled, GraphPP warns and still compiles
extracted graphs with `use_cudagraph=False`:

```text
warn when:
  compile_config.enable
  compile_config.enable_passes
  "cudagraph_pass" not in compile_config.disable_passes

compile with:
  final_inductor_compile_passes(..., use_cudagraph=False, boxed_codegen=True)
```

## Runtime

`register_graph_pp_schedule` installs:

```text
FORWARD          -> _handle_forward
FULL_BACKWARD    -> _handle_full_backward
UNSHARD          -> _handle_unshard
RESHARD          -> _handle_reshard
REDUCE_GRAD      -> _handle_reduce_grad
BACKWARD_INPUT   -> _handle_backward_input
BACKWARD_WEIGHT  -> _handle_backward_weight
OVERLAP_F_B      -> _handle_overlap_fw_bw
```

`GraphPipelineRuntime.step`:

```text
1. store loss_kwargs = kwargs.get("loss_kwargs") or {}
2. delegate to schedule.step(*args, **kwargs)
3. registered handlers execute GraphPP graph calls as schedule actions run
4. if step succeeded, reduce any remaining grads and accumulate into param.grad
5. clear every GraphPipelineStage.state and loss_kwargs
```

Readiness is idempotent per step:

```text
ensure_ready(ctx)
  if graph_provider exists:
    overlap_graphs = graph_provider.ensure_stage_graphs(schedule, ctx,
                                                        loss_kwargs)
  for each local GraphPipelineStage:
    require stage.graphs
    flatten live params and buffers with flatten_graph_values
    collect trainable params
    initialize unsharded_param_grads slots to None
  mark ready
```

Per-step `GraphPPStageRuntimeState`:

```text
flat_param_values
flat_buffer_values
unsharded_param_values
unsharded_param_grads
sharded_param_grads
trainable_params
```

Forward action:

```text
_handle_forward
  -> ensure_ready
  -> resolve local stage and mb_index
  -> if previous stage is remote, wait schedule.fwd_recv_ops[(stage, mb)]
  -> first stage args = ctx.arg_mbs[mb]
     non-first args = stage._retrieve_recv_activations(mb)
     kwargs = ctx.kwarg_mbs[mb]
     target = ctx.target_mbs[mb] only for last stage
  -> ensure unsharded params
  -> graphs.forward(args, kwargs, target, loss_kwargs,
                    unsharded_param_values, flat_buffer_values)
  -> cache (output_tuple, saved_values_for_backward) in stage.fwd_cache
  -> if last stage, append loss to stage.output_chunks and schedule._internal_losses
  -> if next stage is same rank, set_local_fwd_input on that stage
```

Full backward action:

```text
_handle_full_backward
  -> ensure_ready
  -> resolve local stage and mb_index
  -> if next stage is remote, wait schedule.bwd_recv_ops[(stage, mb)]
  -> schedule.backward_counter[stage] += 1
  -> if not stage.has_backward, return
  -> pop (stage_output, saved_values_for_backward) from fwd_cache
  -> output_grads_from_next = () for last stage
     otherwise stage._retrieve_recv_grads(mb)
  -> graphs.full_backward(...)
  -> accumulate raw unsharded param grads in stage.state
  -> cache input_grads_to_prev in stage.bwd_cache
  -> if previous stage is same rank, set_local_bwd_input on that stage
```

Split backward actions:

```text
BACKWARD_INPUT:
  if graphs.supports_backward_input_weight_split is false, skip
  otherwise run graphs.backward_input and cache saved_values_for_backward_weight

BACKWARD_WEIGHT:
  if split is supported:
    pop saved_values_for_backward_weight and run graphs.backward_weight
  else:
    reroute to FULL_BACKWARD
```

FSDP actions:

```text
UNSHARD:
  stage.state.unsharded_param_values =
      graphs.unshard_params(stage.state.flat_param_values)

RESHARD:
  stage.state.unsharded_param_values = []

REDUCE_GRAD:
  stage.state.sharded_param_grads =
      graphs.reduce_grads(stage.state.unsharded_param_grads)
  scale sharded grads once if schedule.scale_grads
```

If no `REDUCE_GRAD` action ran but unsharded grads exist at the end of the
step, `GraphPipelineRuntime.step` reduces before accumulating into `param.grad`.

`OVERLAP_F_B`:

```text
1. validate sub-actions as FORWARD + FULL_BACKWARD
2. prepare the forward half with normal forward helper
3. prepare the backward half with normal backward helper
4. look up overlap_graphs[(fw_stage_index, bw_stage_index)]
5. run overlap_graph.forward_backward(...)
6. apply normal forward post-effects and backward post-effects
```

Grad accumulation:

```text
unsharded grad slots:
  None grad -> skip
  non-tensor -> first value stored, later value must match
  tensor -> first value stored, later values add_ into it

final param.grad:
  graphs.param_grads_for_accumulation(sharded_param_grads)
  -> accumulate_param_grads_(trainable_params, param_grads)
```

Grad scaling uses:

```text
grad_scale_factor = schedule._n_microbatches if schedule.scale_grads else 1
```

## Provider And Stage Contracts

`GraphTrainerStageGraphProvider.ensure_stage_graphs(schedule, ctx, loss_kwargs)`
ensures all local stages have graph executors and returns overlap graph
executors keyed by `(forward_stage_index, backward_stage_index)`.

Provider lifecycle:

```text
1. cast schedule._stages to GraphPipelineStage
2. maybe_register_blockmask_pytree_node()
3. if ctx.arg_mbs and ctx.kwarg_mbs exist:
     create trace-normalized and runtime-normalized microbatch lists with
     normalize_graph_pp_microbatch_inputs
     mutate ctx.arg_mbs[:] and ctx.kwarg_mbs[:] to runtime-normalized values
     create a trace-only _PipelineContext for graph construction
4. if every stage already has graphs:
     warn if CUDA graph pass was requested
     build/reuse overlap graphs
     compile any uncompiled stage graphs
     return overlap graphs
5. for each stage missing graphs:
     _build_stage_graphs(stage, trace args, trace kwargs, trace target,
                         loss_kwargs, compile_graphs=False)
6. warn if CUDA graph pass was requested
7. build/reuse overlap graphs
8. compile every local stage graph
9. return overlap graphs
```

Trace/runtime BlockMask normalization intentionally creates distinct closure
tensor objects. Tracing should not consume the same closure tensors that later
runtime actions receive.

`GraphPPStageGraphs` protocol:

```text
supports_backward_input_weight_split -> bool
num_unsharded_param_grad_values -> int

unshard_params(flat_param_values) -> unsharded_param_values

forward(args, kwargs, target, loss_kwargs,
        unsharded_param_values=..., flat_buffer_values=...)
  -> (stage_output, saved_values_for_backward)

full_backward(stage_output, saved_values_for_backward, output_grads_from_next)
  -> (input_grads_to_prev, unsharded_param_grads)

backward_input(stage_output, saved_values_for_backward, output_grads_from_next)
  -> (input_grads_to_prev, saved_values_for_backward_weight)

backward_weight(saved_values_for_backward_weight) -> unsharded_param_grads
reduce_grads(unsharded_param_grads) -> sharded_param_grads
param_grads_for_accumulation(sharded_param_grads) -> param_grads for live params
```

`GraphPPOverlapGraphs.forward_backward(...)` returns:

```text
(input_grads_to_prev,
 unsharded_param_grads,
 forward_output,
 forward_saved_values_for_backward)
```

`GraphPipelineStage` adds to upstream `PipelineStage`:

```text
graphs: GraphPPStageGraphs | None
state: GraphPPStageRuntimeState
saved_values_for_backward_weight_cache: dict[int, tuple[Any, ...]]
_graph_pp_grads_scaled: bool

set_graphs(graphs)
scale_grads(grad_scale_factor)
```

The stage is a runtime container. It does not own tracing or compilation.

## Tracing

Upstream PP owns microbatch splitting:

```text
schedule.step(*args, **kwargs)
  -> upstream schedule builds _PipelineContext
  -> ctx.arg_mbs / ctx.kwarg_mbs / ctx.target_mbs are populated
  -> provider normalizes graph-sensitive values such as split BlockMask
  -> runtime handlers consume normalized microbatch entries
```

GraphPP does not run a real eager pre-forward to infer output structure.
Upstream PP metadata supplies non-first stage input metadata, and the traced
`stage_step` output itself supplies output structure.

Trace input helpers:

```text
_trace_args_for_stage:
  if stage.is_first and ctx.arg_mbs exists:
    tuple(ctx.arg_mbs[0])
  else:
    tuple(stage._to_tensor(meta) for meta in stage._stage_meta.inputs)

_trace_kwargs_from_context:
  ctx.kwarg_mbs[0] if present, otherwise {}

_trace_target_from_context:
  ctx.target_mbs[0] only for last stages
```

Non-last output grad metadata comes from `stage._stage_meta.output_grads` and
is converted with `stage._to_tensor`. Missing metadata, or mismatched output
and grad metadata counts, raises.

Before tracing, floating tensor leaves in `args` and `kwargs` are detached and
marked `requires_grad_(True)`. Non-floating tensors are left unchanged. The
minimal FX tracer owns fakeification.

Stage-local trace state:

```text
state_params  = named_parameters(remove_duplicate=False)
state_buffers = named_buffers(remove_duplicate=False)
grad_params   = state params with requires_grad=True
input grads   = differentiable tensor leaves from (stage_args, stage_kwargs)
```

Joint trace signatures:

```text
First stage, non-last:
  joint(params, buffers, microbatch_args, microbatch_kwargs,
        output_grads_from_next)
    -> activation_to_next, param_grads, optional input_grads_to_prev

Intermediate stage:
  joint(params, buffers, activation_from_prev, output_grads_from_next)
    -> activation_to_next, param_grads, input_grads_to_prev

Last stage:
  joint(params, buffers, activation_from_prev, target, loss_kwargs)
    -> scalar_loss, param_grads, input_grads_to_prev
```

First-stage input grads are usually empty because token inputs generally do not
require grad. Last-stage backward has no `output_grads_from_next`; the scalar
loss uses autograd's implicit scalar grad seed.

Traced bodies:

```text
last stage:
  pred = stage.submod(*stage_args, **stage_kwargs)
  loss = compute_annotated_loss(loss_fn, pred, target, loss_kwargs)
  grads = torch.autograd.grad(loss,
                              grad_params + differentiable_input_leaves,
                              allow_unused=True)
  return (loss, tuple(param_grads), tuple(input_grads))

non-last stage:
  output = stage.submod(*stage_args, **stage_kwargs)
  grads = torch.autograd.grad(flatten(output),
                              grad_params + differentiable_input_leaves,
                              grad_outputs=flatten(output_grads_from_next),
                              allow_unused=True)
  return (output, tuple(param_grads), tuple(input_grads))
```

For non-last stages:

```text
flat trace inputs =
  flat module state
  + flat (stage_args, stage_kwargs)
  + flat output_grads_from_next

backward_only_input_indices =
  indices of flat output_grads_from_next in that flat input list
```

Trace output validation:

```text
output spec must be exactly:
  (forward_output_or_loss, param_grads, input_grads_to_prev)

forward output group must have at least one leaf
param grad leaf count must equal trainable param count
input grad leaf count must equal differentiable input leaf count
non-last output grad leaves must match traced forward output leaves
```

## Partition

`partition_joint_graph(traced, num_fwd_outputs, backward_only_input_indices)`
splits a post-pass GraphTrainer joint graph into:

```text
forward graph
full backward graph
GraphMeta
```

Flat traced inputs are ordered as:

```text
params + buffers + forward/user inputs + backward-only schedule inputs
```

Terms:

```text
backward_only_input_indices:
  indices into full traced flat inputs for values forward must not consume

backward_grad_input_names:
  backward graph placeholders supplied by the PP schedule at backward time

backward_grad_input_indices:
  indices into the runtime backward tuple, not the full traced input list

saved_values_for_backward:
  forward-computed values returned by forward and consumed by backward

forward side-effect outputs:
  ignored outputs kept only to materialize forward-side mutations

backward passthrough placeholders:
  metadata-only placeholders needed to rewrap backward outputs, such as DTensor
  layout/mesh metadata
```

Extracted graph signatures:

```text
First stage:
  forward(params, buffers, microbatch_inputs)
    -> activation_to_next, saved_values_for_backward, side_effect_outputs
  backward(saved_values_for_backward, output_grads_from_next)
    -> param_grads, optional input_grads_to_prev

Intermediate stage:
  forward(params, buffers, activation_from_prev)
    -> activation_to_next, saved_values_for_backward, side_effect_outputs
  backward(saved_values_for_backward, output_grads_from_next)
    -> param_grads, input_grads_to_prev

Last stage:
  forward(params, buffers, activation_from_prev, target, loss_kwargs)
    -> scalar_loss, saved_values_for_backward, side_effect_outputs
  backward(saved_values_for_backward)
    -> param_grads, input_grads_to_prev
```

`GraphMeta` fields:

```text
fwd_input_names
fwd_flat_input_indices
num_fwd_user_outputs
saved_for_backward_names
fwd_side_effect_output_names
backward_grad_input_names
backward_grad_input_indices
bwd_input_names
bwd_output_names
```

Partition algorithm:

```text
1. copy traced graph and log graph_pp_partition_joint
2. validate num_fwd_outputs and backward_only_input_indices
3. split flat graph outputs with _extract_fwd_bwd_outputs
4. discover saved_values_for_backward from forward and backward closures
5. add backward passthrough placeholders for subclass metadata
6. preserve forward mutations as side-effect outputs; save mutated bases if
   backward consumes them
7. reject any saved backward-only input
8. flatten tuple-like saved values when backward only uses getitem leaves
9. select forward placeholders from dependencies of forward outputs and saves
10. select schedule-supplied backward grad placeholders from backward-only inputs
11. extract forward and backward graphs with _extract_graph_with_inputs_outputs
12. validate backward inputs are saved values followed by runtime grad inputs
13. validate forward saved output names match backward saved input names
14. build GraphMeta and log graph_pp_partition_forward/backward
```

Forward mutation handling covers:

```text
- chunked loss hidden-state grad accumulator writes consumed by decoder backward
- MoE expert-count buffer updates that affect training state
```

The pass directly uses upstream PyTorch extraction helpers:

```text
_extract_fwd_bwd_outputs
_extract_graph_with_inputs_outputs
torch.fx._lazy_graph_module._make_graph_module
```

## Graph Construction

Per-stage graph construction:

```text
stage_step Python function
  -> minimal_fx_tracer
  -> compile_time_passes(include_inductor=False, use_cudagraph=False)
  -> partition_joint_graph
  -> split_forward_fsdp_collectives
  -> split_backward_fsdp_collectives
  -> split_di_dw_graph
  -> final_inductor_compile_passes(boxed_codegen=True, use_cudagraph=False)
  -> GraphTrainerStageGraphs
```

`_GraphTrainerPassConfigView` exposes only fields read by reused GraphTrainer
pass builders:

```text
compile
parallelism
model_spec.model
```

If compile passes are enabled, `model_config` and `parallelism` are required.

Stage graph module set:

```text
_StageGraphModules(
  fw,
  full_bw,
  bw_di=None or graph,
  bw_dw=None or graph,
  unshard=None or graph,
  reduce_grad=None or graph,
)
```

Each graph is annotated before compilation with:

```text
graph_pp_stage_index
graph_pp_callable
graph_pp_action
graph_pp_slot
```

`_StageGraphMeta` is private to `GraphTrainerStageGraphs` and connects
partition metadata to GraphTrainer execution:

```text
num_user_outputs
num_saved_for_backward
num_param_grad_values
num_input_grad_values
num_flat_param_values
fwd_output_values
param_grad_values
input_grad_values
partition
fwd_input_names
fwd_flat_input_indices
bw_no_fsdp_output_names
reduce_grad_input_names
unshard_flat_param_indices
num_fw_unsharded_param_inputs
is_last_stage
```

`GraphMeta` is the public partition calling convention. `_StageGraphMeta` adds
wrapping specs, FSDP split metadata, flat param counts, and last-stage packing
policy. The runtime never inspects it.

`GraphTrainerStageGraphs` packs runtime values:

```text
forward user inputs:
  last stage    -> flatten ((args, kwargs, target, loss_kwargs), {})
  non-last      -> flatten ((args, kwargs), {})
  then flatten_graph_values to unwrap tensor subclasses

forward graph args:
  leading unsharded parameter prefix consumed by fw_no_fsdp
  + values selected from
      unsharded_param_values + flat_buffer_values + flat_user_inputs
    by fwd_flat_input_indices

forward graph outputs:
  user outputs + saved_values_for_backward + ignored side-effect outputs
  user outputs are rewrapped/unflattened; saved values stay flat

backward graph args:
  saved values by saved_for_backward_names
  + output_grads_from_next selected by backward_grad_input_names/indices
  in partition.bwd_input_names order

full backward outputs:
  param grad flat values first
  input grad flat values second
  input grads are wrapped for PP; param grads stay raw until accumulation
```

Optional behavior:

```text
no unshard graph:
  unshard_params(flat_param_values) returns a copy of flat_param_values

no reduce-grad graph:
  reduce_grads(unsharded_param_grads) returns validated raw grads

no dI/dW split:
  supports_backward_input_weight_split is false
```

Compilation:

```text
if compile disabled or passes disabled:
  ensure_boxed_graph_module(gm)
else:
  example_inputs = node.meta["val"] for each placeholder
  apply final_inductor_compile_passes(..., use_cudagraph=False,
                                      boxed_codegen=True)
  require BOXED_CODEGEN_META is True
```

Overlap graphs must be built before stage graphs are compiled; otherwise
GraphPP raises.

## FSDP Split

`split_forward_fsdp_collectives` extracts all-gather/unshard work from forward.
Inputs:

```text
fw_module
num_params
fwd_input_names
fwd_flat_input_indices
```

It validates input names/indices against placeholders. Since flat traced inputs
are params, buffers, then user inputs:

```text
flat_index < num_params -> parameter input
flat_index >= num_params -> remaining forward input
```

For each parameter placeholder, `find_fsdp_unshard_output` detects the
all-gather/wait/view chain. If no collective exists:

```text
unshard_module = None
fw_no_fsdp_module = original fw_module
num_fw_unsharded_param_inputs = 0
```

If a collective exists:

```text
unshard(param_inputs)
  -> one output per original parameter input

fw_no_fsdp(unsharded_param_outputs, remaining_forward_inputs)
  -> original forward outputs
```

Replicated params pass through `unshard` unchanged so `fw_no_fsdp` receives a
uniform unsharded param prefix. FX extraction temporarily allows selected c10d
`wait_tensor` ops.

`split_backward_fsdp_collectives` extracts reduce-grad work from backward.
Backward outputs are:

```text
parameter grad slots + input_grads_to_prev
```

For each parameter grad output, `find_fsdp_reduce_grad_input` detects the
pre-reduce input to a reduce-scatter/all-reduce chain. If no collective exists:

```text
bw_no_fsdp_module = original bw_module
reduce_grad_module = None
```

If a collective exists:

```text
bw_no_fsdp(original_backward_inputs)
  -> reduce_grad_inputs in original param-grad slot order
     + input_grads_to_prev

reduce_grad(unique_reduce_grad_inputs)
  -> original parameter grad outputs
```

Duplicate reduce-grad inputs are de-duplicated in first-occurrence order.
`None`, unused, non-differentiable, or non-reduced grad slots remain in place.
The pre-reduce dtype cast remains in `bw_no_fsdp`, matching eager FSDP with
gradient sync disabled.

Current runtime placement:

```text
UNSHARD runs all-gather/unshard before stage forwards.
REDUCE_GRAD runs reduce-scatter/all-reduce after stage backward accumulation.
RESHARD only clears cached unsharded params; it has no extracted graph today.
```

## dI/dW Split

`split_di_dw_graph` splits full backward into backward-input and
backward-weight graphs. Input backward outputs are ordered:

```text
param_grads, input_grads_to_prev
```

The pass first rewrites output order to:

```text
input_grads_to_prev, param_grads
```

Then `_extract_fwd_bwd_outputs(num_fwd_outputs=num_input_grads)` treats input
grads as the "forward" side for extraction.

Output callables:

```text
bw_di(original_backward_inputs)
  -> input_grads_to_prev + saved_values_for_backward_weight

bw_dw(saved_values_for_backward_weight)
  -> param_grads
```

Live-ins for `bw_dw` are values computed by the dI closure and later used by
dW. The collector handles:

```text
get_attr nodes:
  keep as graph constants in dW

non-tensor tuple-like call results:
  pass concrete getitem leaves when dW consumes them

tensors used only for symbolic shape reads:
  pass symbolic reads instead of the tensor

symbolic shape values:
  pass saved_sym_nodes as live-ins
```

If `num_input_grads == 0`, the pass returns `None`; runtime skips
`BACKWARD_INPUT` and reroutes `BACKWARD_WEIGHT` to `FULL_BACKWARD`.

Placeholders starting with `tangent` are renamed to
`graph_pp_runtime_grad...` for readable graph dumps.

## Overlap Graphs

`OVERLAP_F_B` uses a multiplexed graph for one
`(forward_stage_index, backward_stage_index)` pair:

```text
placeholders: bw_inputs + fw_inputs
outputs:      bw_outputs + fw_outputs
```

`multiplex_fw_bw_graph(fw_gm, bw_gm)`:

```text
1. deep-copy forward graph as destination
2. find destination FakeTensorMode from forward metadata
3. transfer backward metadata into that fake mode and ShapeEnv
4. copy backward get_attr targets with "bw" prefix
5. insert backward placeholders before forward placeholders
6. insert backward compute before forward compute
7. replace output tuple with backward outputs followed by forward outputs
8. lint, recompile, and trace-log graph_pp_multiplexed_graph
```

The forward graph is the destination because its ShapeEnv owns dynamic
collective-size constraints needed by full Inductor for MoE all-to-all outputs.
The `overlap` argument currently has no behavior. EP-overlap annotations are
not applied inside the multiplexed graph; pass ordering keeps them on
standalone no-FSDP forward/backward graphs.

Required pairs come from:

```text
schedule.pipeline_order_with_comms[schedule.rank]
```

For every `OVERLAP_F_B`, `overlap_fw_bw_sub_actions` requires exactly two
sub-actions:

```text
first:  FORWARD
second: FULL_BACKWARD
```

`BACKWARD_INPUT` overlap is not implemented.

Metadata transfer for multiplexing is handled by `_MetaShapeEnvTransfer`.
It copies backward FX metadata into the forward graph's `FakeTensorMode` and
`ShapeEnv`. Supported leaves:

```text
FakeTensor
torch.SymInt
SymPy expressions
containers of those values
ordinary non-symbolic metadata by shallow sharing
```

Transfer is two phase:

```text
collect:
  find foreign unbacked base symbols in FakeTensor/SymInt metadata
  skip backed symbols with guarding hints
  reject same-named unbacked symbols from multiple foreign ShapeEnvs

seed:
  create destination entries for collected foreign unbacked bases with
  _transfer_foreign_expr_as_unbacked

copy:
  copy FakeTensors through destination FakeTensorMode.from_tensor
  copy SymInts through destination ShapeEnv expression transfer
  copy SymPy expressions by replacing seeded foreign base symbols
```

The TODO links PyTorch issue:

```text
https://github.com/pytorch/pytorch/issues/188723
```

## Runtime Value Details

Boxed execution:

```text
_execute_graph_module(gm, args):
  with torch.no_grad():
    outputs = gm(args)
  require boxed codegen cleared args
  normalize tuple/list/scalar outputs to tuple
```

All GraphPP executor methods use this path. There is no unboxed GraphPP path.
Boxed execution lets generated code clear mutable argument lists and avoid
holding microbatch inputs or saved values longer than explicit returned values.

Tensor subclass wrapping:

```text
flat graph inputs:
  params, buffers, user inputs, targets, loss kwargs, schedule grad inputs

rewrapped graph outputs:
  stage forward outputs
  input_grads_to_prev
  parameter grads before final accumulation

kept flat:
  saved_values_for_backward
  saved_values_for_backward_weight
  unsharded params
  raw unsharded/sharded grad values before accumulation wrapping
  multiplexed graph internals
```

`GraphPPValueSpec` is a slice of `TracedResult.output_subclass_layouts` for
forward outputs, param grads, or input grads. It records semantic value count,
subclass layouts, and optional pytree spec. It computes flat value count,
rewraps flat values, and optionally unflattens pytrees. Count mismatches raise.

BlockMask handling:

```text
Non-PP GraphTrainer:
  traces the full-batch BlockMask directly.

Eager PP:
  upstream _split_block_mask may close over a Python int offset; no non-strict
  graph trace needs to replay different offsets.

GraphPP:
  upstream PP splits microbatches, then GraphPP normalizes each split BlockMask
  so the microbatch batch offset is a scalar tensor closure leaf.
```

`_graphable_split_block_mask` rewrites only recognized upstream PP-split masks:

```text
mask_mod is a function with closure values:
  block_mask: base BlockMask
  idx: Python int batch offset
```

For recognized masks it validates `0 <= idx < base_batch_size`, creates a
scalar tensor `batch_offset`, shallow-copies the `BlockMask`, and replaces only
`mask_mod`:

```text
def graphable_mask_mod(b, h, q_idx, kv_idx):
  offset = batch_offset.to(device=b.device, dtype=b.dtype)
  return base_mask_mod(b + offset, h, q_idx, kv_idx)
```

Unrecognized masks are returned unchanged. `normalize_graph_pp_microbatch_inputs`
maps this across every arg/kwarg microbatch, treating `BlockMask` as a pytree
leaf, and requires equal arg/kwarg microbatch counts. TODO:

```text
https://github.com/pytorch/pytorch/issues/188727
```

Loss handling:

```text
last-stage trace:
  pred = stage.submod(...)
  loss = compute_annotated_loss(loss_fn, pred, target, loss_kwargs)

runtime:
  last-stage forward graph returns scalar loss
  stage.output_chunks.append(loss)
  schedule._internal_losses.append(loss)
```

GraphPP does not change `loss.py` or define a separate loss implementation.
Only the last stage consumes `target` and `loss_kwargs`.

Eval:

```text
GraphPipelineRuntime.eval(*args, **kwargs)
  -> store loss_kwargs
  -> delegate to schedule.eval
  -> clear GraphPP stage state
  -> do not accumulate parameter gradients
```

Trace logging:

```text
trace_graph_pp_graph emits trace_structured("artifact") with gm.print_readable
including stride and device.

Current artifact names:
  graph_pp_partition_joint
  graph_pp_partition_forward
  graph_pp_partition_backward
  graph_pp_fsdp_forward_no_fsdp
  graph_pp_fsdp_unshard
  graph_pp_fsdp_backward_no_fsdp
  graph_pp_fsdp_reduce_grad
  graph_pp_split_di_dw_input
  graph_pp_bw_di
  graph_pp_bw_dw
  graph_pp_multiplexed_graph
```

Shared helpers in `utils.py` include graph output/name helpers, closure and
dependency helpers, mutation detection, placeholder renaming, boxed compile
example extraction, temporary side-effectful FX extraction allowance, overlap
sub-action validation, subclass flattening/wrapping, BlockMask normalization,
trace logging, and metadata transfer.

## Validation

GraphPP raises on violated contracts:

```text
Config:
  wrong compile mode, precompiled artifact dir, non-runtime schedule,
  fsdp_reshard_after_forward="always"

Runtime:
  local stages not GraphPipelineStage, missing stage.graphs, missing remote
  recv op, malformed OVERLAP_F_B, missing overlap graph, boxed args not cleared

Tracing:
  missing non-last output_grad metadata, output metadata mismatch, traced output
  not exactly three groups, grad counts mismatch

Partition:
  invalid backward_only_input_indices, forward depends on backward-only input,
  saved backward-only input, backward input order mismatch, saved-name mismatch

FSDP:
  metadata/placeholder mismatch, invalid flat indices, num_param_grads exceeds
  backward output count

dI/dW:
  invalid num_param_grads, backward output not tuple, unexpected non-tensor
  multi-output users

Wrapping/compile:
  subclass rewrap count mismatch, placeholder metadata missing for examples

BlockMask:
  arg/kwarg microbatch count mismatch, split offset outside base BlockMask
```

Intentional semantic no-ops:

```text
no unshard collective -> unshard_params returns flat params
no reduce-grad collective -> reduce_grads returns raw grads
no dI/dW split -> BACKWARD_INPUT skips and BACKWARD_WEIGHT reroutes
unrecognized BlockMask -> returned unchanged
```

## Numerics

Correctness target: bitwise-identical training numerics against an equivalent
eager PP baseline with the same effective microbatching, topology, schedule,
and grad accumulation semantics.

Important comparison caveats:

```text
- ZBV and DualPipeV eager PP can require torch.compile for FlexAttention in
  tested configs, so eager PP Interleaved1F1B is used as the FlexAttention
  baseline when it is the only runnable eager PP schedule.
- Local-rank gradient clipping can change distributed numerics because clipping
  is applied to local rank gradients. Bitwise comparisons must configure
  clipping consistently or make the clipping threshold unreachable.
- Same-schedule comparisons should not rely on tolerances. If losses or grad
  norms differ, root-cause schedule order, metadata inference, graph passes, or
  distributed reduction differences.
```

## Limitations

```text
- aot_fx_trace only
- precompiled graph artifacts unsupported
- CUDA graph compilation disabled for GraphPP extracted graphs
- fsdp_reshard_after_forward="always" unsupported
- OVERLAP_F_B supports forward + full backward only
- GraphPP depends on upstream PP metadata initialization before tracing
- local BlockMask normalization remains until upstream _split_block_mask is
  non-strict tracing-friendly
- local metadata transfer remains until PyTorch exposes a complete FX metadata
  tree transfer utility for FakeTensor/ShapeEnv metadata
```

## Review Invariants

```text
1. No core TorchTitan changes are required for GraphPP.
2. Runtime does not inspect GraphTrainer graph metadata.
3. Graph construction does not run eager stage forward for trace inference.
4. PP microbatch splitting remains upstream-owned.
5. Partition calling convention is explicit in GraphMeta.
6. FSDP and dI/dW extraction use shared helpers and validate invariants.
7. Extracted/compiled graphs use boxed execution.
8. Graph pass outputs are trace-logged for tlparse.
9. BlockMask normalization is a narrow compatibility shim, not a replacement
   BlockMask implementation.
10. CUDA graph compilation is not silently enabled for GraphPP graphs.
```

## Future Work

```text
1. Upstream non-strict-tracing-friendly _split_block_mask so local mask_mod
   normalization can be deleted.

2. Add a PyTorch ShapeEnv/FakeTensorMode utility that transfers a whole FX
   metadata tree into a destination fake mode for graph multiplexing.

3. Add real GraphPP CUDA graph runtime integration. Current extracted graphs
   compile with use_cudagraph=False and warn when cudagraph_pass is otherwise
   enabled.

4. Extend overlap graph support if schedules need BACKWARD_INPUT overlap pairs.

5. Revisit fsdp_reshard_after_forward="always"; current validation rejects it.

6. Add precompiled artifact loading for stage-local GraphPP graphs; current
   validation rejects --compile.precompile_artifact_dir.

7. Add fused FSDP edge graphs. Today FSDP edges are exposed as schedule actions:
   UNSHARD runs all-gather/unshard before stage forwards, and REDUCE_GRAD runs
   reduce-scatter/all-reduce after stage backward grad accumulation. With fused
   edge graphs those actions would become no-ops for GraphPP execution: unshard
   would fuse into the stage's first forward microbatch graph, iF0, and
   reduce-grad would fuse into the stage's last backward microbatch graph, iBn
   for full backward schedules or iWn for separate weight backward schedules.
   This keeps logical FSDP placement at stage boundaries while removing
   separate boxed-call and launch boundaries for edge collectives.
```

sanketpurandare:
Stacked PRs:
 * #3585
 * #3090
 * #3089
 * #3088
 * #2727
 * __->__#2726

--- --- ---

### [graph_trainer][aot_fx_trace][graph_pp] Add stage graph partitioning

Add the GraphPP stage-local partitioner and the shared graph extraction calling-convention helpers used by later GraphPP passes. The partitioner derives the forward graph, backward graph, runtime input mapping, saved values, and mutation materialization outputs from GraphTrainer traced results without relying on descriptor classes.

Document the first, intermediate, and last-stage calling conventions, keep the partition flow explicit, and validate invalid backward-only input states instead of silently skipping them.

Add unit coverage for last-stage loss graphs, non-last-stage output-gradient graphs, runtime input mapping, mutation preservation, backward passthrough metadata, and partition invariant failures.

### DIFF
--- a/torchtitan/experiments/graph_trainer/graph_pp/partition.py
+++ b/torchtitan/experiments/graph_trainer/graph_pp/partition.py
@@ -1,0 +1,626 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+"""Stage-local GraphPP graph partitioning.
+
+GraphTrainer traces one flat joint graph per pipeline stage. The joint graph
+returns forward user outputs first, then gradients. GraphPP splits that graph
+into a forward callable and a backward callable with a stable calling
+convention for PP runtime schedules. ``GraphMeta`` records that calling
+convention: ordered placeholder names describe each extracted graph API, and
+flat input indices map runtime graph inputs back to the traced GraphTrainer
+input list.
+
+Terms used by this pass:
+
+- flat inputs: traced GraphTrainer inputs. GraphTrainer orders them as params,
+  buffers, then user forward inputs, followed by schedule-supplied backward
+  inputs when the trace contains them.
+- forward user outputs: values returned by stage forward, such as
+  activation_to_next or scalar_loss.
+- backward outputs: gradients returned after forward user outputs; parameter
+  grads come before input_grads_to_prev when input grads exist.
+- backward_only_input_indices: indices into flat inputs for values that must
+  not be forward graph inputs because the PP schedule supplies them only while
+  executing backward.
+- backward_grad_input_indices: indices into the schedule-supplied backward
+  input tuple. They are not indices into the full traced input list.
+- backward grad inputs: the concrete backward-time values named by
+  backward_only_input_indices. For intermediate stages this is
+  output_grads_from_next. The last stage relies on autograd's implicit scalar
+  loss seed and has no schedule-supplied backward grad input.
+- saved_values_for_backward: forward-computed values returned by the forward
+  callable so the backward callable can consume them later.
+- backward passthrough placeholders: metadata-only placeholders needed to
+  rewrap backward outputs, such as DTensor class/layout metadata.
+
+Representative signatures:
+
+First stage, non-last:
+  joint(params, buffers, microbatch_inputs, output_grads_from_next)
+    -> activation_to_next, param_grads
+  forward(params, buffers, microbatch_inputs)
+    -> activation_to_next, saved_values_for_backward, mutation_outputs
+  backward(saved_values_for_backward, output_grads_from_next)
+    -> param_grads
+
+Intermediate stage:
+  joint(params, buffers, activation_from_prev, output_grads_from_next)
+    -> activation_to_next, param_grads, input_grads_to_prev
+  forward(params, buffers, activation_from_prev)
+    -> activation_to_next, saved_values_for_backward, mutation_outputs
+  backward(saved_values_for_backward, output_grads_from_next)
+    -> param_grads, input_grads_to_prev
+
+Last stage:
+  joint(params, buffers, activation_from_prev, target_and_loss_inputs)
+    -> scalar_loss, param_grads, input_grads_to_prev
+  forward(params, buffers, activation_from_prev, target_and_loss_inputs)
+    -> scalar_loss, saved_values_for_backward, mutation_outputs
+  backward(saved_values_for_backward)
+    -> param_grads, input_grads_to_prev
+
+The first stage usually has no input_grads_to_prev because real microbatch
+inputs generally do not require gradients.
+"""
+
+import copy
+from collections.abc import Iterable, Sequence
+from dataclasses import dataclass
+
+import torch.fx as fx
+from torch._functorch.partitioners import (
+    _extract_fwd_bwd_outputs,
+    _extract_graph_with_inputs_outputs,
+)
+from torch.fx._lazy_graph_module import _make_graph_module
+
+from torchtitan.experiments.graph_trainer.graph_pp.utils import (
+    base_tensor_for_mutation_target,
+    is_getitem_node,
+    is_mutation_node,
+    node_closure,
+    node_order,
+    output_names,
+    placeholder_dependencies,
+    placeholder_names,
+    trace_graph_pp_graph,
+    unique_in_order,
+)
+
+from torchtitan.experiments.graph_trainer.make_fx_tracer import TracedResult
+
+
+@dataclass(frozen=True, slots=True)
+class GraphMeta:
+    """Calling convention for GraphTrainer-native GraphPP fwd/bwd graphs.
+
+    ``fwd_input_names`` names the forward graph placeholders, while
+    ``fwd_flat_input_indices`` maps each placeholder to the traced flat input
+    list. Since flat inputs are params, buffers, then user inputs, the runtime
+    can split those indices into state inputs and user inputs using stage
+    metadata.
+
+    ``backward_grad_input_names`` names backward placeholders supplied by the
+    PP schedule at backward time, not values saved by forward.
+    ``backward_grad_input_indices`` indexes into the schedule-provided backward
+    tuple: output_grads_from_next for non-last stages. The last stage has no
+    entries because scalar losses use autograd's implicit grad seed.
+
+    ``fwd_side_effect_output_names`` names extra forward outputs appended after
+    saved values only to keep forward-side mutations live. The runtime ignores
+    their values after the graph has executed.
+
+    Attributes:
+        fwd_input_names (tuple[str, ...]): Forward graph placeholder names.
+        fwd_flat_input_indices (tuple[int, ...]): Indices mapping each forward
+            placeholder back to the traced GraphTrainer flat input list.
+        num_fwd_user_outputs (int): Number of leading forward graph outputs
+            that correspond to real stage forward outputs.
+        saved_for_backward_names (tuple[str, ...]): Forward outputs consumed
+            later as leading backward graph inputs.
+        fwd_side_effect_output_names (tuple[str, ...]): Extra forward outputs
+            used only to materialize forward-side mutations.
+        backward_grad_input_names (tuple[str, ...]): Backward placeholders
+            supplied by the PP schedule, such as output_grads_from_next.
+        backward_grad_input_indices (tuple[int, ...]): Indices into the
+            schedule-supplied backward input tuple.
+        bwd_input_names (tuple[str, ...]): Backward graph placeholder names.
+        bwd_output_names (tuple[str, ...]): Backward graph output names.
+    """
+
+    fwd_input_names: tuple[str, ...]
+    fwd_flat_input_indices: tuple[int, ...]
+    num_fwd_user_outputs: int
+    saved_for_backward_names: tuple[str, ...]
+    fwd_side_effect_output_names: tuple[str, ...]
+    backward_grad_input_names: tuple[str, ...]
+    backward_grad_input_indices: tuple[int, ...]
+    bwd_input_names: tuple[str, ...]
+    bwd_output_names: tuple[str, ...]
+
+    def __post_init__(self) -> None:
+        if len(self.fwd_input_names) != len(self.fwd_flat_input_indices):
+            raise ValueError(
+                "Forward input metadata must pair each placeholder name with "
+                "one traced flat input index: "
+                f"{self.fwd_input_names} vs {self.fwd_flat_input_indices}"
+            )
+        if len(self.backward_grad_input_names) != len(self.backward_grad_input_indices):
+            raise ValueError(
+                "Backward grad input metadata must pair each placeholder name "
+                "with one schedule input index: "
+                f"{self.backward_grad_input_names} vs "
+                f"{self.backward_grad_input_indices}"
+            )
+
+    @property
+    def num_fwd_inputs(self) -> int:
+        """int: Number of forward graph placeholders."""
+        return len(self.fwd_input_names)
+
+    @property
+    def num_saved_for_backward(self) -> int:
+        """int: Number of forward outputs saved for backward."""
+        return len(self.saved_for_backward_names)
+
+    @property
+    def num_fwd_side_effect_outputs(self) -> int:
+        """int: Number of forward outputs kept only for side effects."""
+        return len(self.fwd_side_effect_output_names)
+
+    @property
+    def num_bwd_inputs(self) -> int:
+        """int: Number of backward graph placeholders."""
+        return len(self.bwd_input_names)
+
+    @property
+    def num_backward_grad_inputs(self) -> int:
+        """int: Number of backward placeholders supplied by the PP schedule."""
+        return len(self.backward_grad_input_names)
+
+    @property
+    def num_bwd_outputs(self) -> int:
+        """int: Number of backward graph outputs."""
+        return len(self.bwd_output_names)
+
+
+def _forward_flat_input_indices(
+    gm: fx.GraphModule,
+    placeholder_index_by_name: dict[str, int],
+) -> tuple[int, ...]:
+    """Map extracted forward placeholders to traced flat input positions."""
+
+    flat_input_indices: list[int] = []
+    for node in gm.graph.find_nodes(op="placeholder"):
+        if node.name not in placeholder_index_by_name:
+            raise ValueError(
+                "Extracted graph placeholder does not come from the traced "
+                f"GraphTrainer flat input list: {node.name}"
+            )
+        flat_input_indices.append(placeholder_index_by_name[node.name])
+    return tuple(flat_input_indices)
+
+
+def _validate_backward_only_input_indices(
+    indices: tuple[int, ...],
+    *,
+    num_placeholders: int,
+) -> None:
+    """Validate caller-selected traced inputs that only backward may consume."""
+
+    if len(set(indices)) != len(indices):
+        raise ValueError(
+            "backward_only_input_indices must be unique, got " f"{list(indices)}"
+        )
+    invalid_indices = sorted(
+        index for index in indices if index < 0 or index >= num_placeholders
+    )
+    if invalid_indices:
+        raise ValueError(
+            "backward_only_input_indices must reference traced graph "
+            f"placeholders: {invalid_indices} for {num_placeholders} "
+            "placeholders"
+        )
+
+
+def _invalid_backward_only_value_names(
+    values: Iterable[object],
+    backward_only_names: set[str],
+) -> list[str]:
+    """Return backward-only placeholders that were selected for forward save."""
+
+    return sorted(
+        {
+            value.name
+            for value in values
+            if isinstance(value, fx.Node) and value.name in backward_only_names
+        }
+    )
+
+
+def _saved_values_for_backward(
+    joint: fx.GraphModule,
+    *,
+    fwd_outputs: Sequence[object],
+    bwd_outputs: Sequence[object],
+    backward_only_names: set[str],
+) -> list[fx.Node]:
+    """Select forward-computed values needed by the backward callable.
+
+    A value is saved when it is in the forward dependency closure and has a
+    later user that belongs only to the backward dependency closure. This also
+    covers values used by both forward and backward: the forward callable
+    returns them once, and the PP runtime passes them to backward later.
+    """
+    forward_nodes = node_closure(fwd_outputs)
+    backward_nodes = node_closure(bwd_outputs)
+    saved: list[fx.Node] = []
+    for node in joint.graph.nodes:
+        if node not in forward_nodes or node.name in backward_only_names:
+            continue
+        if any(
+            user in backward_nodes and user not in forward_nodes for user in node.users
+        ):
+            saved.append(node)
+    return saved
+
+
+def _forward_mutations_to_materialize(
+    joint: fx.GraphModule,
+    *,
+    fwd_outputs: Sequence[object],
+    bwd_outputs: Sequence[object],
+    backward_only_names: set[str],
+) -> tuple[list[fx.Node], list[fx.Node]]:
+    """Find forward-side mutations that must execute in the forward callable.
+
+    Chunked loss traces populate the hidden-state gradient accumulator with
+    ``copy_`` mutations during the loss forward. The accumulator is later
+    consumed by decoder backward, but those writes are not visible in the data
+    dependency closure of the loss output. Preserve the mutated base as a saved
+    value and keep the mutation nodes as forward-only outputs so extracted
+    forward graph execution materializes the post-mutation tensor.
+
+    Other forward mutations, such as MoE expert-count buffer updates, are not
+    consumed by backward but still affect training state. Keep those mutation
+    nodes as forward-only outputs as well; the runtime ignores the extra values,
+    but returning them keeps the mutations live in the extracted graph.
+    """
+
+    order = node_order(joint.graph)
+    fwd_output_nodes = [value for value in fwd_outputs if isinstance(value, fx.Node)]
+    if not fwd_output_nodes:
+        return [], []
+    fwd_output_barrier = max(order[node] for node in fwd_output_nodes)
+    backward_nodes = node_closure(bwd_outputs)
+    saved_mutation_bases: list[fx.Node] = []
+    mutation_outputs: list[fx.Node] = []
+
+    for node in joint.graph.nodes:
+        if (
+            order[node] > fwd_output_barrier
+            or node in backward_nodes
+            or not is_mutation_node(node)
+        ):
+            continue
+        if not node.args:
+            continue
+        mutated_base = base_tensor_for_mutation_target(node.args[0])
+        if mutated_base is None:
+            continue
+        if mutated_base.name in backward_only_names:
+            raise ValueError(
+                "Forward mutation cannot target a backward-only input: "
+                f"mutation={node.name}, target={mutated_base.name}"
+            )
+        mutation_outputs.append(node)
+        if mutated_base.op != "placeholder" and mutated_base in backward_nodes:
+            saved_mutation_bases.append(mutated_base)
+
+    return (
+        unique_in_order(saved_mutation_bases),
+        unique_in_order(mutation_outputs),
+    )
+
+
+def _backward_passthrough_placeholders(
+    *,
+    bwd_outputs: Sequence[object],
+    backward_only_names: set[str],
+) -> list[fx.Node]:
+    """Preserve metadata placeholders returned by backward directly.
+
+    minimal_fx_tracer unwraps tensor subclasses into plain graph values. A
+    DTensor gradient, for example, may flatten to ``(local_grad, device_mesh)``,
+    where ``device_mesh`` and layout state are input placeholders carried
+    through only so the runtime can rewrap the output subclass. Since such
+    placeholders have no compute users, dependency-based saved-value discovery
+    will not select them; they still must be available to the extracted
+    backward graph.
+    """
+    return unique_in_order(
+        node
+        for node in bwd_outputs
+        if isinstance(node, fx.Node)
+        and node.op == "placeholder"
+        and node.name not in backward_only_names
+    )
+
+
+def _is_tuple_like_node(node: fx.Node) -> bool:
+    return isinstance(node.meta.get("val"), (tuple, list))
+
+
+def _flatten_saved_values_for_backward(
+    joint: fx.GraphModule,
+    *,
+    saved_values: list[fx.Node],
+    bwd_outputs: Sequence[object],
+) -> list[fx.Node]:
+    """Expose saved tuple intermediates as tensor leaves.
+
+    Higher-order ops such as FlexAttention can return nested tuples whose tensor
+    leaves are consumed by backward. Keeping the raw tuple as a forward output
+    works for interpreted FX, but standalone regional Inductor expects compiled
+    regions to expose plain tensor outputs. If backward only observes the tuple
+    through getitem chains, save those getitem leaves directly.
+    """
+
+    backward_nodes = node_closure(bwd_outputs)
+    order = node_order(joint.graph)
+
+    def flatten_if_backward_uses_items(node: fx.Node) -> list[fx.Node]:
+        if not _is_tuple_like_node(node):
+            return [node]
+
+        backward_users = [user for user in node.users if user in backward_nodes]
+        getitem_users = [user for user in backward_users if is_getitem_node(user)]
+        if not getitem_users or len(getitem_users) != len(backward_users):
+            return [node]
+
+        flattened_items: list[fx.Node] = []
+        for user in sorted(getitem_users, key=order.__getitem__):
+            flattened_items.extend(flatten_if_backward_uses_items(user))
+        return flattened_items
+
+    flattened: list[fx.Node] = []
+    for node in saved_values:
+        flattened.extend(flatten_if_backward_uses_items(node))
+    return unique_in_order(flattened)
+
+
+def _forward_inputs_from_outputs(
+    placeholders: list[fx.Node],
+    *,
+    fw_outputs: Sequence[object],
+    backward_only_names: set[str],
+) -> list[fx.Node]:
+    """Select forward placeholders and reject backward-only dependencies."""
+
+    fw_input_set = placeholder_dependencies(fw_outputs)
+    invalid_names = sorted(
+        node.name for node in fw_input_set if node.name in backward_only_names
+    )
+    if invalid_names:
+        raise ValueError(
+            "Forward graph outputs require backward-only inputs: " f"{invalid_names}"
+        )
+    return [node for node in placeholders if node in fw_input_set]
+
+
+def _backward_grad_inputs_from_schedule(
+    backward_only_inputs: list[fx.Node],
+    *,
+    bwd_outputs: Sequence[object],
+) -> tuple[list[fx.Node], tuple[int, ...]]:
+    """Return schedule-supplied grad placeholders consumed by backward.
+
+    For non-last stages, the PP schedule supplies output_grads_from_next. The
+    last stage has no schedule-supplied backward grad inputs. The returned
+    indices are positions in that backward-time tuple, not positions in the
+    full traced input list.
+    """
+
+    bwd_input_set = placeholder_dependencies(bwd_outputs)
+    backward_grad_inputs = [
+        node for node in backward_only_inputs if node in bwd_input_set
+    ]
+    backward_grad_input_indices = tuple(
+        index
+        for index, node in enumerate(backward_only_inputs)
+        if node in bwd_input_set
+    )
+    return backward_grad_inputs, backward_grad_input_indices
+
+
+def partition_joint_graph(
+    traced: TracedResult,
+    *,
+    num_fwd_outputs: int,
+    backward_only_input_indices: tuple[int, ...] = (),
+) -> tuple[fx.GraphModule, fx.GraphModule, GraphMeta]:
+    """Partition a post-pass GraphTrainer joint graph into GraphPP callables.
+
+    The caller passes a flat ``minimal_fx_tracer`` graph after any
+    metadata-preserving GraphTrainer passes. ``num_fwd_outputs`` counts the raw
+    forward user outputs after tensor subclasses are unwrapped. Semantic pytree
+    structure is restored by the GraphPP runtime at the runtime boundary.
+
+    Args:
+        traced (TracedResult): Flat stage-local joint graph produced by
+            ``minimal_fx_tracer``.
+        num_fwd_outputs (int): Number of leading traced outputs that belong to
+            the stage forward result.
+        backward_only_input_indices (tuple[int, ...]): Flat input indices for
+            values supplied only during backward by the PP schedule. These
+            inputs are excluded from the forward graph.
+
+    Returns:
+        tuple[fx.GraphModule, fx.GraphModule, GraphMeta]: Forward graph,
+        backward graph, and their GraphPP calling-convention metadata.
+
+    Raises:
+        ValueError: If the requested output/input partition violates the
+            GraphPP calling convention or would silently save backward-only
+            inputs through the forward graph.
+    """
+    if num_fwd_outputs < 1:
+        raise ValueError(f"num_fwd_outputs must be positive, got {num_fwd_outputs}")
+
+    joint = copy.deepcopy(traced.gm)
+    trace_graph_pp_graph("graph_pp_partition_joint", joint)
+    placeholders = list(joint.graph.find_nodes(op="placeholder"))
+    placeholder_index_by_name = {
+        node.name: index for index, node in enumerate(placeholders)
+    }
+    _validate_backward_only_input_indices(
+        backward_only_input_indices,
+        num_placeholders=len(placeholders),
+    )
+    backward_only_index_set = set(backward_only_input_indices)
+    backward_only_inputs = [
+        node
+        for index, node in enumerate(placeholders)
+        if index in backward_only_index_set
+    ]
+    backward_only_names = {node.name for node in backward_only_inputs}
+
+    (
+        fwd_outputs,
+        bwd_outputs,
+        fwd_output_descs,
+        bwd_output_descs,
+    ) = _extract_fwd_bwd_outputs(joint, num_fwd_outputs=num_fwd_outputs)
+    if len(fwd_outputs) != num_fwd_outputs:
+        raise ValueError(
+            "num_fwd_outputs exceeds traced graph output count: "
+            f"requested {num_fwd_outputs}, found {len(fwd_outputs) + len(bwd_outputs)}"
+        )
+
+    # 1. Select values produced by forward and later consumed by backward.
+    saved_values = _saved_values_for_backward(
+        joint,
+        fwd_outputs=fwd_outputs,
+        bwd_outputs=bwd_outputs,
+        backward_only_names=backward_only_names,
+    )
+
+    # 2. Add metadata-only placeholders needed to rewrap backward outputs.
+    saved_values.extend(
+        _backward_passthrough_placeholders(
+            bwd_outputs=bwd_outputs,
+            backward_only_names=backward_only_names,
+        )
+    )
+
+    # 3. Preserve forward tensor mutations whose return values are otherwise
+    # dead from the perspective of forward user outputs.
+    (mutation_saved_values, fwd_mutation_outputs,) = _forward_mutations_to_materialize(
+        joint,
+        fwd_outputs=fwd_outputs,
+        bwd_outputs=bwd_outputs,
+        backward_only_names=backward_only_names,
+    )
+    saved_values = unique_in_order([*saved_values, *mutation_saved_values])
+    invalid_saved_names = _invalid_backward_only_value_names(
+        saved_values,
+        backward_only_names,
+    )
+    if invalid_saved_names:
+        raise ValueError(
+            "saved_values_for_backward must not include backward-only inputs: "
+            f"{invalid_saved_names}"
+        )
+
+    # 4. Expose tuple saved values as leaves when backward only observes the
+    # leaves through getitem chains.
+    saved_values = _flatten_saved_values_for_backward(
+        joint,
+        saved_values=saved_values,
+        bwd_outputs=bwd_outputs,
+    )
+
+    # 5. Select the concrete calling convention and extract both subgraphs.
+    fw_outputs = fwd_outputs + saved_values + fwd_mutation_outputs
+    fw_output_descs = fwd_output_descs + [None] * (
+        len(saved_values) + len(fwd_mutation_outputs)
+    )
+    fw_inputs = _forward_inputs_from_outputs(
+        placeholders,
+        fw_outputs=fw_outputs,
+        backward_only_names=backward_only_names,
+    )
+
+    (
+        backward_grad_inputs,
+        backward_grad_input_indices,
+    ) = _backward_grad_inputs_from_schedule(
+        backward_only_inputs,
+        bwd_outputs=bwd_outputs,
+    )
+    bw_inputs = saved_values + backward_grad_inputs
+
+    fw_graph = _extract_graph_with_inputs_outputs(
+        joint.graph,
+        fw_inputs,
+        fw_outputs,
+        fw_output_descs,
+        "forward",
+        ignore_must_be_in_fw_bw=True,
+    )
+    bw_graph = _extract_graph_with_inputs_outputs(
+        joint.graph,
+        bw_inputs,
+        bwd_outputs,
+        bwd_output_descs,
+        "backward",
+        ignore_must_be_in_fw_bw=True,
+    )
+    fw_module = _make_graph_module(joint, fw_graph)
+    bw_module = _make_graph_module(joint, bw_graph)
+    fw_module.graph.lint()
+    bw_module.graph.lint()
+    fw_module.recompile()
+    bw_module.recompile()
+    trace_graph_pp_graph("graph_pp_partition_forward", fw_module)
+    trace_graph_pp_graph("graph_pp_partition_backward", bw_module)
+
+    saved_for_backward_names = output_names(fw_module)[
+        num_fwd_outputs : num_fwd_outputs + len(saved_values)
+    ]
+    bwd_input_names = placeholder_names(bw_module)
+    expected_bwd_input_names = tuple(node.name for node in bw_inputs)
+    if bwd_input_names != expected_bwd_input_names:
+        raise ValueError(
+            "Backward graph inputs must be saved values followed by "
+            "schedule-supplied backward grad inputs: "
+            f"expected {expected_bwd_input_names}, got {bwd_input_names}"
+        )
+    if bwd_input_names[: len(saved_for_backward_names)] != saved_for_backward_names:
+        raise ValueError(
+            "Forward saved output names must match backward saved input names: "
+            f"saved={saved_for_backward_names}, backward={bwd_input_names}"
+        )
+
+    return (
+        fw_module,
+        bw_module,
+        GraphMeta(
+            fwd_input_names=placeholder_names(fw_module),
+            fwd_flat_input_indices=_forward_flat_input_indices(
+                fw_module, placeholder_index_by_name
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/graph_trainer/graph_pp/boxed.py
```diff
@@ -0,0 +1,26 @@
...
+    """Execute a GraphPP callable with boxed argument lifetime semantics.
+
+    GraphPP schedules keep many per-microbatch activations alive.  Always
+    executing through a mutable boxed input list lets the runner drop its input
+    references immediately after the graph consumes them, instead of retaining a
+    long-lived Python argument tuple.
+    """
+    try:
+        if isinstance(graph, fx.GraphModule):
+            return fx.Interpreter(graph).boxed_run(boxed_args)
```
aditvenk: Can we have boxed codegen applied to each fwd/bwd graph and have this directly execute that? Instead of creating a new instance of Interpreter.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,288 @@
...
+    for bwd_placeholder in bw_module.graph.find_nodes(op="placeholder"):
+        if bwd_placeholder.name in backward_only_names:
+            continue
+        if bwd_placeholder.name not in saved_outputs_by_name:
+            raise ValueError(
+                "Backward placeholder is neither a forward saved output nor a "
+                "runtime backward input: "
+                f"{bwd_placeholder.name}"
+            )
+        ordered_saved_outputs.append(saved_outputs_by_name[bwd_placeholder.name])
```
aditvenk: maybe dumb q: can we assume that name of bwd_placeholder is guaranteed to match with name in fw_output ?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,288 @@
...
+        ordered_saved_descs.append(saved_descs_by_name[bwd_placeholder.name])
+
+    kept_outputs = user_outputs + ordered_saved_outputs
+    kept_output_descs = user_output_descs + ordered_saved_descs
+    needed_fw_placeholders = _placeholder_dependencies(kept_outputs)
+    fw_inputs = [
```
aditvenk: If there are inputs that are consumed by side-effecting ops, is it guaranteed by `default_partition()` that it will be captured by either fw_outputs or saved_outputs?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,288 @@
...
+        node
+        for node in fw_placeholders
+        if node in needed_fw_placeholders and node.name not in backward_only_names
+    ]
+
+    if list(fw_inputs) != list(fw_placeholders) or list(fw_outputs) != kept_outputs:
+        fw_graph = _extract_graph_with_inputs_outputs(
```
aditvenk: Maybe a general question -- why do we do this pruning (e.g., saved fw outputs that are not consumed in bw, etc.) ? Is it commonly seen?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,288 @@
...
+        strict=True,
+    ):
+        if isinstance(output, fx.Node):
+            saved_outputs_by_name[output.name] = output
+            saved_descs_by_name[output.name] = desc
+
+    ordered_saved_outputs = []
+    ordered_saved_descs = []
+    for bwd_placeholder in bw_module.graph.find_nodes(op="placeholder"):
+        if bwd_placeholder.name in backward_only_names:
```
aditvenk: Should we be validating that backward_only_names is not interleaved with saved_from_forward inputs? Or, is that guaranteed?

aditvenk (approved):
LGTM -- some naive questions on what behaviors are guaranteed by the paritioner.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: Yes have the side-effect part handled

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+class GraphPPInputSource:
+    """Runtime source for one extracted graph placeholder."""
+
+    name: str
+    kind: GraphPPInputSourceKind
+    index: int
+
+
+@dataclass(frozen=True, slots=True)
+class GraphMeta:
```
wconstab, line 97: just wondering- i've seen calling conventions for graphs floating around in previous iterations of things (autop, probably) - do we have vary many versions of these conventions existing in parallel, or are we standardizing on one or a couple that can be shared by different use cases?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+        and hasattr(node.target, "is_view")
+        and node.target.is_view
+        and node.args
+        and isinstance(node.args[0], fx.Node)
+    ):
+        node = node.args[0]
+    return node
+
+
+def _forward_side_effects(
```
wconstab: print is a side effect, this helper seems to only target mutations (possibly rename it?)


also, this is similar to what functionalization does in core aot_autograd isn't it?  i guess it's hard to reuse any of that infra?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: All the Auto-Parallel stuff was deleted. Its a full port. The non-strict graphs don't have descriptors, so the calling convention is what we have defined with the minimal fx tracer.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: yeah the aot-autograd infra is not very much compatible with non-strict traced graphs due to missing metadata. And then we have our own remat, offload, backward tagging.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+    in the forward dependency closure and has a backward-only user.
+    """
+    forward_nodes = _node_closure(fwd_outputs)
+    backward_nodes = _node_closure(bwd_outputs)
+    saved: list[fx.Node] = []
+    for node in joint.graph.nodes:
+        if node not in forward_nodes or node.name in backward_only_names:
+            continue
+        if any(
+            user in backward_nodes and user not in forward_nodes for user in node.users
```
aditvenk, line 265: Hmm. what about a node that has a backward use AND a forward use? Shouldn't we be saving that too?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: print side effect won't be captured here, its a non-strict trace not a dynamo trace, so tensor mutations and wait_tensor calls

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+    if not fwd_output_nodes:
+        return [], []
+    fwd_output_barrier = max(order[node] for node in fwd_output_nodes)
+    backward_nodes = _node_closure(bwd_outputs)
+    saved_bases: list[fx.Node] = []
+    side_effect_outputs: list[fx.Node] = []
+
+    for node in joint.graph.nodes:
+        if (
+            order[node] > fwd_output_barrier
```
aditvenk, line 304: what is the purpose of this check? Can't we just restrict to checking if node in backward_nodes or is a fwd-but-not-mutating-node?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+            or node in backward_nodes
+            or not _is_mutation_node(node)
+        ):
+            continue
+        if not node.args:
+            continue
+        base = _alias_base(node.args[0])
+        if base is None or base.name in backward_only_names:
```
aditvenk: Shouldn't `base.name in backward_only_names` be a cause to throw? It is unexpected?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+            continue
+        side_effect_outputs.append(node)
```
aditvenk: Can a view op be a mutation op? When can `node` be a mutation op in fwd, but `base` is different than `node` ?

torchtitan/experiments/graph_trainer/graph_pp/utils.py
```diff
@@ -0,0 +1,284 @@
...
+        BLOCK_SIZE=block_mask.BLOCK_SIZE,
+        mask_mod=graph_pp_batch_offset_mask_mod,
+        dq_write_order=block_mask.dq_write_order,
+        dq_write_order_full=block_mask.dq_write_order_full,
+        dq_kv_order=block_mask.dq_kv_order,
+        dq_kv_order_spt=block_mask.dq_kv_order_spt,
+    )
+
+
+def normalize_graph_pp_microbatch_inputs(
```
wconstab: we're baking flexattn blockmask into pp graph infra. i guess it'd be nicer if we could tell users 'here's how to tell us how to not specialize on a particular input, provide a wrapper for it like this' rather than hiding that process and not having it be extensible to other types of inputs. not sure how that would look though.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+        backward_only_names=backward_only_names,
+    )
+    saved_values = list(dict.fromkeys([*saved_values, *side_effect_saved_values]))
+    saved_values = _flatten_tuple_saved_values(
+        joint,
+        saved_values=saved_values,
+        bwd_outputs=bwd_outputs,
+    )
+
+    fw_outputs = fwd_outputs + saved_values + fwd_side_effect_outputs
```
aditvenk: Could there be fwd-passthroughs similar to bwd-passthroughs solely for the purpose of re-wrapping?

torchtitan/experiments/graph_trainer/graph_pp/utils.py
sanketpurandare: This is temporary, we are going to supply PP with pre-split microbatches, so PP never has to deal with splitting the micro-batches in the first place. This will be an upstream change in torch.pipelining and also in torchtitan

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+    fw_output_descs = fwd_output_descs + [None] * (
+        len(saved_values) + len(fwd_side_effect_outputs)
+    )
+    fw_input_set = _placeholder_dependencies(fw_outputs)
+    fw_inputs = [
+        node
+        for node in placeholders
+        if node in fw_input_set and node.name not in backward_only_names
```
aditvenk: similar to above, shouldn't it be cause for error if `node` in `fw_input_set` and it is in `backward_only_names` ?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: Nodes used in forward and backward are a strict subset of nodes in forward that have a backward user.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
```diff
@@ -0,0 +1,465 @@
...
+    ]
+
+    bwd_input_set = _node_closure(bwd_outputs)
```
aditvenk: Why not _placeholder_dependencies(bwd_outputs) ?

sanketpurandare: So just renaming?

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: I agree, this should probably be an invariant error, not a silent skip.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: A view op itself is not a mutation. The mutation op can target a view, e.g. slice = aten.slice(...); copy_(slice, value). In that case _alias_base() walks from the mutated view back to the actual base tensor/buffer.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: This check keeps the helper scoped to mutations that occur in the forward region before the forward outputs. Checking only node in backward_nodes is not enough because the joint graph can contain recompute or backward-side nodes after the forward outputs; those must not be materialized by the forward callable.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: Agrred, this should also be an explicit invariant error.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: Forward outputs already carry their rewrap metadata as part of fwd_outputs, so no separate fwd-passthrough helper is needed. Backward outputs also include metadata-only leaves in bwd_outputs, but unlike forward, backward graph inputs are reconstructed from saved forward values plus runtime backward inputs. A metadata-only backward output can be a passthrough placeholder with no users, so it must be explicitly added to saved values to become an input to the extracted backward graph.

torchtitan/experiments/graph_trainer/graph_pp/partition.py
sanketpurandare: yes like buffer mutations

PR #2728, opened 2026-03-27
Add async logging

drisspg:
Stacked PRs:
 * __->__#2728
 * #2171

--- --- ---

Add async logging

I set this by default  but curious if there is interest  

```Py
~/.venvs/nightly/bin/torchrun --nproc-per-node=1 -m torchtitan.train \
    --module llama3 --config llama3_debugmodel \
    --training.steps 200 \
    --training.local-batch-size 64 \
    --training.seq-len 8192 \
    --metrics.log_freq 5
```
    
Running before after on my b200 gives me non trivial speedups:
image
However in real workloads you can do other things ( log less frequently)

### DIFF
--- a/torchtitan/components/metrics.py
+++ b/torchtitan/components/metrics.py
@@ -258,6 +258,24 @@ def _get_metrics_rank(
     return (world_size // pp_size) * (pp_size - 1)
 
 
+@dataclass
+class PendingMetrics:
+    """Bundles all metric tensors from a single logging step with one CUDA event.
+
+    A single event guards all non-blocking D2H copies so we pay for one
+    synchronization point rather than one per metric. Used by
+    ``MetricsProcessor.log_async`` / ``flush`` to defer the GPU-CPU sync
+    until the next train step, where the event has already completed.
+    """
+
+    step: int
+    global_avg_loss: torch.Tensor
+    global_max_loss: torch.Tensor
+    grad_norm: torch.Tensor
+    extra_metrics: dict[str, Any]
+    event: torch.cuda.Event
+
+
 class MetricsProcessor(Configurable):
     """Metrics processor to processes the metrics and log metrics.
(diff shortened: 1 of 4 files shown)


wwwjn:
@felipemello1 has a stack of PR working on observability, it might also be covered 

Update: After reading this PR and I realize this PR and @felipemello1  are doing different level of "Async". This PR is doing non-blocking D2H copy, but @felipemello1 's is async file aggregation. Maybe can consider consolidating? Both are interesting

tianyu-l:
Agreed that we should merge with @felipemello1 's work. In particular, the metrics logger should be able to handle both GPU metrics and CPU metrics.

felipemello1:
SixLib has something like this, but for dtensors. They use a context manager to store references to the dtensor, so you can just call "add_summary(key, tensor)" from anywhere. At the end of training, they call "replicate_to_host", that groups and does all_reduce only once.

The key difference is that since its a dtensor, they dont have to pass the mesh.

I had a PR for this, but realized that our activations are not dtensor as in sixlib, so their APIs didnt work quite well. I put it on hold until the dtensor changes in Titan are finalized.

I am going to come back to this soon, and the idea is to enable computing metrics more deep into titan stack, so we can have more visibility into tensor related metrics, e.g. in MoE paths, like llama4x has.

If we can wait a bit more, I would prefer to share an RFC and we can think about it, otherwise, I am happy to review this PR and can land it in between.

tianyu-l:
@felipemello1 
IMO we can and should wait.

One thing I didn't understand from your comment:
> At the end of training, they call "replicate_to_host", that groups and does all_reduce only once.

Is this happening every iteration, or only once per training job? If it's the former, that's exactly what this PR is trying to avoid? i.e. the sync between GPU and CPU due to per-iteration logging.

PR #2732, opened 2026-03-27
Add bf16 optimizer state support via step pre-hook

vivekgoe:
Use a step pre-hook to pre-populate Adam/AdamW optimizer states (exp_avg, exp_avg_sq) in bfloat16 before the optimizer's lazy initialization runs. This causes _init_group to skip its own fp32 allocation and the fused CUDA kernel to dispatch to its mixed-precision path (FusedAdamMathFunctorMP), reducing optimizer memory usage while maintaining convergence.
Enabled via --optimizer.bf16_optimizer_states flag (requires fused implementation and Adam/AdamW).

### DIFF
--- a/torchtitan/components/optimizer.py
+++ b/torchtitan/components/optimizer.py
@@ -354,7 +354,7 @@ class OptimizersInBackwardContainer(OptimizersContainer):
                     "implementation='fused_opt_states_bf16' is not supported with "
                     "OptimizersInBackwardContainer"
                 )
-            super().__post_init__()
+            OptimizersContainer.Config.__post_init__(self)
 
     def __init__(self, config: Config, *, model_parts: list[nn.Module]) -> None:
         optimizer_cls = self._resolve_optimizer_cls(config.name)


vivekgoe:
@tianyu-l @janeyx99 Changes to initialize optimizer states in bf16 from TorchTitan such that we can use Pytorch optimizer Fused kernels that support mixed precision optimizer states. Please review, if this looks acceptable then I will go ahead and work on cleaning up corresponding Pytorch PR.

torchtitan/components/optimizer.py
```diff
@@ -87,6 +88,14 @@ class Config(Configurable.Config):
         - more info: https://pytorch.org/docs/stable/optim.html
         """
 
+        bf16_optimizer_states: bool = False
```
tianyu-l: should name this to `states_dtype`, mimic https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L50

torchtitan/trainer.py
```diff
@@ -419,6 +420,18 @@ def __init__(self, config: Config):
...
+            if config.optimizer.name not in ("Adam", "AdamW"):
+                raise ValueError(
+                    f"bf16_optimizer_states is only supported for Adam/AdamW, "
+                    f"got optimizer '{config.optimizer.name}'"
+                )
+            if config.optimizer.implementation != "fused":
+                raise ValueError(
+                    f"bf16_optimizer_states requires fused implementation, "
+                    f"got '{config.optimizer.implementation}'"
+                )
```
tianyu-l: throw in optimizer.config `__post_init__`

torchtitan/trainer.py
```diff
@@ -419,6 +420,18 @@ def __init__(self, config: Config):
...
+            register_bf16_optimizer_state_hook(self.optimizers)
```
tianyu-l: can we do this inside optimizer constructor?

torchtitan/components/optimizer.py
vivekgoe: Done.

torchtitan/trainer.py
vivekgoe: Done.

torchtitan/trainer.py
vivekgoe: Done.

vivekgoe:
@tianyu-l Pushed patch with changes to take care of review comments, please let me know if any other changes are required. 
If everything looks ok, we can merge this once https://github.com/pytorch/pytorch/pull/175230 is completed and merged.

torchtitan/components/optimizer.py
```diff
@@ -119,6 +141,8 @@ def __init__(self, config: Config, *, model_parts: list[nn.Module]) -> None:
             params = [p for p in model.parameters() if p.requires_grad]
             self.optimizers.append(optimizer_cls(params, **optimizer_kwargs))
             all_params.extend(params)
+        if config.states_dtype == "bfloat16":
+            register_bf16_optimizer_state_hook(self)
```
tianyu-l: we can make this a instance method so you can call
```suggestion
            self.register_bf16_optimizer_state_hook()
```

vivekgoe: Done

vivekgoe:
@tianyu-l This PR is now ready. Please help review. Results shared are with latest nightly torch build. 
Some plots from deepseek-v3-16b training to show memory saving from the feature, also shows there is no adverse impact on convergence & e2e performance.
optimizer_states_comparison

Added snapshots from torch traces showing performance gain only for optimizer.step calls, average wall duration of call reduced by ~25%.
**FP32:**
image
**BF16**
image

docs/bf16_optimizer_states.md
```diff
@@ -0,0 +1,52 @@
...
+
+These constraints are validated at config time.
+
+### How it works
+
+A step pre-hook is registered on each optimizer instance. Before Adam's lazy state initialization runs on the first step, the hook pre-populates `exp_avg` and `exp_avg_sq` as bfloat16 tensors. When `_init_group` finds non-empty state, it skips its own fp32 allocation. The fused kernel detects the dtype mismatch between fp32 parameters and bf16 states and dispatches to the mixed-precision code path.
+
+### Interaction with other features
+
+- **`training.dtype`**: `states_dtype` controls only the optimizer state precision. It is independent of `training.dtype`, which controls parameter/gradient precision. Both can be set to `bfloat16` simultaneously for maximum memory savings.
```
tianyu-l: Not sure about the "independent part" -- if we set `training.dtype` to bf16 and `states_dtype` to fp32, would the fp32 take effect? I somehow think the optimizer states would follow the parameters to be still bf16.

If so, should we error out (in post_init) that training dtype bf16, optimizer states dtype fp32 is not supported?

docs/bf16_optimizer_states.md
```diff
@@ -0,0 +1,52 @@
+## BF16 Optimizer States
+
+By default, Adam/AdamW maintain their momentum (`exp_avg`) and variance (`exp_avg_sq`) buffers in float32, which doubles the memory footprint relative to model parameters. Setting `optimizer.states_dtype` to `bfloat16` stores these buffers in bfloat16 instead, reducing optimizer state memory by approximately 50%.
```
tianyu-l: worth clarifying that the default is when model params are in fp32?

docs/bf16_optimizer_states.md
vivekgoe: Good suggestion. Although PyTorch Fused Adam MP kernel did support training.dtype bf16 and states_dtype fp32, I choose to only allow training.dtype fp32 and states_dtype bf16 combination because I it did not seem logical to keep optimizer states in higher precision compared to params/grads. I will add extra check in post_init.

docs/bf16_optimizer_states.md
vivekgoe: Yes makes sense, will add it.

vivekgoe:
@tianyu-l Updated as per your review comments. Note I had to make a small change to optimizer.states_dtype default configuration to maintain backward compatibility 
`states_dtype: Literal["float32", "bfloat16"] | None = None`
This ensures that someone who was already using training.dtype=bfloat16 does not hit the new assert added to check training.dtype=bfloat16 and optimizer.states_dtype=float32 combination.

torchtitan/components/optimizer.py
```diff
@@ -211,6 +275,10 @@ def optim_hook(param) -> None:
 
         self.optimizers = list(optim_dict.values())
 
+        assert (
+            config.states_dtype != "bfloat16"
+        ), "states_dtype='bfloat16' is not supported with OptimizersInBackwardContainer"
+
```
tianyu-l: shouldn't error out if `training.dtype` is also bf16?

torchtitan/components/optimizer.py
```diff
@@ -119,6 +141,8 @@ def __init__(self, config: Config, *, model_parts: list[nn.Module]) -> None:
             params = [p for p in model.parameters() if p.requires_grad]
             self.optimizers.append(optimizer_cls(params, **optimizer_kwargs))
             all_params.extend(params)
+        if config.states_dtype == "bfloat16":
+            self._register_bf16_optimizer_state_hook()
```
tianyu-l: if `training.dtype` is already bf16, is this registration nearly no-op?

torchtitan/components/optimizer.py
```diff
@@ -87,6 +87,28 @@ class Config(Configurable.Config):
         - more info: https://pytorch.org/docs/stable/optim.html
         """
 
+        states_dtype: Literal["float32", "bfloat16"] | None = None
```
tianyu-l: Since we don't ever need any setup other than bf16 or None, do you think it would be less confusing (and error checks) if we remove the fp32 option?

torchtitan/components/optimizer.py
```diff
@@ -87,6 +87,28 @@ class Config(Configurable.Config):
...
+        'bfloat16' reduces memory usage. Requires fused implementation and
+        Adam/AdamW optimizer. The fused CUDA kernel handles mixed-precision
+        (fp32 params + bf16 states) natively. When None (default), optimizer
+        states follow PyTorch's default behavior.
+        """
+
+        def __post_init__(self):
+            if self.states_dtype == "bfloat16":
+                if self.name not in ("Adam", "AdamW"):
+                    raise ValueError(
```
tianyu-l, line 136: what if training.dtype is already bf16?

torchtitan/trainer.py
```diff
@@ -115,6 +115,15 @@ def __post_init__(self):
                     raise NotImplementedError(
                         "Optimizers in backward is not supported with Pipeline Parallel."
                     )
+            if (
+                self.training.dtype == "bfloat16"
+                and self.optimizer.states_dtype == "float32"
```
tianyu-l: see the other comments: if we remove the fp32 option, we don't need this check in main trainer.py

torchtitan/components/optimizer.py
vivekgoe: If training.dype is already bf16, ideally we expect user to leave optimizer.states_dtype to default (which is None). In case someone still sets states.dtype to bf16, it should not result in any change in behavior because it will only result in optimizer states initialization happening in the hook code (instead of happening within Pytorch Adam optimizer code).

torchtitan/components/optimizer.py
vivekgoe: It will move optimizer state initialization to hook (instead of being done in pytorch Adam optimizer code), but end result should be the same.

torchtitan/components/optimizer.py
vivekgoe: We are not registering the hook for BackwardContainer, so I was not sure if it will work or not, therefore added assert for bf16. 
I also do not clearly understand the use-case where this class will be used and how to test it, therefore I am hesitant to add bf16 states support to this. 
If you could suggest some test or use-case for this class, I can add the hook and check if it works as expected or not.

torchtitan/components/optimizer.py
vivekgoe: Yes, I agree. I will make the change.

torchtitan/trainer.py
vivekgoe: Yes, I agree. We can remove fp32 option and this check as well. I will make the change.

vivekgoe:
@tianyu-l Responded to your review comments. Need some guidance from you on OptimizerBackwardContainer change. Thanks.

torchtitan/components/optimizer.py
tianyu-l: OK thinking over it twice, I feel we shouldn't have this generic `states_dtype` config.

As your work is targeting a specific important scenario (param master dtype fp32, optimizer states dtype bf16), I think it's better to have a flag to explicitly control this behavior.

How about adding an option `"fused_bf16"` in `implementation`, and clearly document what happens if training dtype is bf16 or fp32.

torchtitan/components/optimizer.py
tianyu-l: oh we don't need to support this container. If we use my latest suggestion, we can just error out in `BackwardContainer.Config`'s `post_init`, if the implementation is `fused_bf16`.

torchtitan/components/optimizer.py
vivekgoe: Sounds reasonable. I will make the change, but will name the new option as "fused_opt_states_bf16" instead to align closely with feature.

tianyu-l:
need to rebase

vivekgoe:
I have pushed latest version which aligns to all review comments. But Pytorch patch for mixed precision optimizer states has been reverted due to some failures with ROCm builds, so this one will need to wait until those failures get fixed. I will let you know. 
Please review latest version and let me know if any other changes are required. Thanks.

tianyu-l:
lgtm, let me know when it's ready to be tested against CI. Thanks a lot!

vivekgoe:
@tianyu-l This PR is now ready to be tested against CI. I have rebased it to latest. Thanks.

PR #2740, opened 2026-03-27
[ROCm][CI] Use MI350 label for all workflows.

akashveramd:
This PR replaces MI325 label and uses MI350 label for all workflows.


jithunnair-amd:
@akashveramd https://github.com/pytorch/torchtitan/actions/runs/24062922281/job/70182799508?pr=2740 launched the workflow successfully (but failed during execution) on an 8-GPU MI350 test runner. It's not clear yet if the failure is due to MI350 or not.

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -105,7 +105,13 @@ jobs:
           # But this is more an unknown issue that AMD people may want to
           # figure out the root cause or confirm that this is an expected
           # behavior.
-          LOSS_FILE="tests/assets/losses/llama3_rocm.txt"
+          rocm_device_name=$(python -c "import torch; print(torch.cuda.get_device_name(None))")
```
tianyu-l, line 108: why do we need this if the goal is to
> replaces MI325 label and uses MI350 label for all workflows.

akashveramd: @jithunnair-amd: Do you want to answer as it was your suggestion to preserve loss numbers for MI325X

jithunnair-amd: @tianyu-l Mostly for coverage reasons, but if you don't see any value in retaining the values for MI300s/MI325, we can just keep the MI350 ones.

tianyu-l: @jithunnair-amd since we switch CI to MI350, the values for 325 are not guarded anyways, so I think we should just keep the MI350 one.

jithunnair-amd: @tianyu-l Do you have any objections to moving the `tests/assets/losses/llama3_rocm.txt` to `tests/assets/losses/MI350X/llama3_rocm.txt`? I think this keeps the source code much more clearer in terms of which device the losses are for. I'd suggest the same be done for CUDA, if different CUDA GPUs would have different losses. This also allows adding other GPUs to CI coverage down the line, with their respective txt files.

tianyu-l: sure, alternatively you could do `losses/llama3_rocm_mi350x.txt` as there's only one model right now.

jithunnair-amd: @akashveramd Can you please follow the above suggestion, and rename `tests/assets/losses/llama3_rocm.txt` to `tests/assets/losses/llama3_rocm_mi350x.txt` with updated loss values?

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -105,7 +105,13 @@ jobs:
...
+          rocm_device_tag=$(echo "$rocm_device_name" | grep -oE 'MI[0-9]+X')
+          if [[ -z "${rocm_device_tag}" ]]; then
+            echo "Error: Could not extract ROCm GPU tag from device name: ${rocm_device_name}"
+            exit 1
+          fi
+          LOSS_FILE="tests/assets/losses/llama3_rocm_${rocm_device_tag,,}.txt"
```
tianyu-l: we don't need this complicated logic anymore?

jithunnair-amd: updated

tianyu-l:
lgtm, I can stamp once it runs and completes.

tianyu-l (approved):

PR #2743, opened 2026-03-28
[graph_trainer] Extract shared model setup helpers into torchtitan/model_setup.py

bobrenjc93:
* __->__ #2743
* #2713

## Summary

Extract `materialize_model` into `torchtitan/model_setup.py` so that
`precompile_main` and `trainer` share a single implementation of the
`to_empty` → `init_weights` → `train()` materialization path.

This is the one piece of the build → parallelize → init_weights sequence with
enough logic to justify extraction (meta→device move, `no_grad` context,
weight init, training mode). The simpler steps (`update_from_config`,
`model_config.build()`, `parallelize_fn(...)`) remain inlined at their
callsites since they're 2–3 lines each.

Changes:
- `materialize_model` takes `model: BaseModel` directly (no internal cast).
- Removed `init_weights_context` parameter — the single caller
  (`precompile_main`) now wraps externally with `dist_config.patch(...)`.
- Dropped `prepare_model_config`, `build_model_on_meta`, `parallelize_model`
  (review consensus: trivial wrappers adding indirection without value).

## Test plan

- `tests/unit_tests/test_model_setup.py` covers `materialize_model`:
  - Verifies parameters move from meta → cpu
  - Verifies `init_weights` is called with correct `buffer_device`
  - Verifies weights are actually initialized (not uninitialized)
  - Verifies model is in training mode after materialization
  - Tests the PP loop pattern (calling `materialize_model` per-stage)
- Tests run in CI via `unit_test_cpu.yaml`


torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+
+def prepare_model_config(
+    *,
+    model_spec: ModelSpec,
+    trainer_config: Any,
+) -> BaseModel.Config:
+    """Update the selected model config from the trainer config."""
+    model_config = model_spec.model
+    model_config.update_from_config(trainer_config=trainer_config)
+    return model_config
```
SherlockNoMad: Melchior: `prepare_model_config` is a 3-line function that just calls `update_from_config`. This adds indirection without real value — the original two lines at each callsite were already clear. If the goal is dedup, at least make the helper do something non-trivial (e.g. validate the config, log what changed). As-is, this is speculative abstraction for a trivial operation.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+    """Apply the model spec's non-pipelined parallelization path."""
+    return cast(
+        Module,
+        model_spec.parallelize_fn(
+            model,
+            parallel_dims=parallel_dims,
+            training=training,
+            model_converters=model_converters,
+            parallelism=parallelism,
+            compile_config=compile_config,
```
SherlockNoMad: Melchior: `parallelize_model` is just a `cast()` wrapper around `model_spec.parallelize_fn(...)`. It doesn't add validation, logging, or any logic — just passes kwargs through. The callsites already had the same call with the same arguments. This is adding a layer of indirection that makes it harder to grep for who actually calls `parallelize_fn`. If you want a typed wrapper, consider adding type hints to the `parallelize_fn` signature in `ModelSpec` instead.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+        init_weights_context
+        if init_weights_context is not None
+        else contextlib.nullcontext()
+    ):
+        with torch.no_grad():
+            # TODO: Change this back to init_weights once
+            # autoparallel contains the wrap_init_states
+            cast(BaseModel, model).init_weights(buffer_device=buffer_device)
+    model.train()
+    return model
```
SherlockNoMad, line 30: Melchior: The TODO comment `Change this back to init_weights once autoparallel contains the wrap_init_states` was copied verbatim from the callsites. This is fine to keep, but the cast to `BaseModel` here is a code smell — if `materialize_model` takes a `Module`, but internally needs `BaseModel.init_weights`, the type signature is lying. Should the parameter type be `BaseModel` instead of `Module`?

torchtitan/trainer.py
```diff
@@ -375,12 +379,11 @@ def __init__(self, config: Config):
...
-                m.to_empty(device=init_device)
-                with torch.no_grad():
-                    # TODO: Change this back to init_weights once
-                    # autoparallel contains the wrap_init_states
-                    cast(BaseModel, m).init_weights(buffer_device=buffer_device)
-                m.train()
+                materialize_model(
+                    cast(BaseModel, m),
+                    init_device=init_device,
+                    buffer_device=buffer_device,
```
SherlockNoMad, line 379: Melchior: The PP path still calls `m.to_empty()` / `init_weights()` / `m.train()` inline instead of using `materialize_model`. This means the dedup is incomplete — you still have two implementations of the same logic. Was this intentional? If `materialize_model` is the canonical path, it should be used here too (and this is the more complex case since it iterates over `model_parts`).

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -163,15 +169,12 @@ def main():
...
-        with torch.no_grad():
-            model.init_weights(buffer_device=None)
-    finally:
-        dist_config.compile_on_one_rank = True
-    model.train()
+    model = materialize_model(
+        model,
+        init_device=utils.device_type,
+        buffer_device=None,
+        init_weights_context=dist_config.patch("compile_on_one_rank", False),
```
SherlockNoMad: Melchior: `dist_config.patch("compile_on_one_rank", False)` — where is `.patch()` defined? I don't see it in the diff. Is this a new method being added elsewhere, or does it already exist? If it's new, it should be in this PR or referenced.

tests/unit_tests/test_model_setup.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.linear(x)
+
+    def init_weights(self, buffer_device: torch.device | None = None, **kwargs) -> None:
+        self.saw_init_weights_context = bool(getattr(self, "_context_active", False))
+        self.init_buffer_device = buffer_device
+
+
+@dataclass(kw_only=True, slots=True)
```
SherlockNoMad: Melchior: `init_weights` sets `self.saw_init_weights_context` and `self.init_buffer_device` but never actually initializes any weights. The test asserts that the context manager was entered and that `buffer_device` was passed through, but it doesn't verify that `to_empty` actually moved parameters from meta → real device. Consider asserting `next(model.parameters()).requires_grad` or that param values are not uninitialized after `init_weights`.

tests/unit_tests/test_model_setup.py
```diff
@@ -0,0 +1,176 @@
...
+            init_weights_context=tracking_context,
+        )
+
+        assert seen_kwargs["parallel_dims"] is parallel_dims
+        assert seen_kwargs["compile_config"] is compile_config
+        assert tracking_context.entered
+        assert tracking_context.exited
+        assert model.saw_init_weights_context
+        assert model.init_buffer_device == torch.device("cpu")
+        assert next(model.parameters()).device.type == "cpu"
```
SherlockNoMad: Melchior: No test for the PP path (`model_parts` loop in trainer.py), which is the path that was *not* refactored. Since the stated goal is preventing the two paths from diverging, you should at minimum test that `materialize_model` works when called in a loop (the PP pattern). Also — does this test actually run in CI? Check the test config.

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: lol? this is just code movement

torchtitan/model_setup.py
bobrenjc93: also just code movement :P

torchtitan/model_setup.py
SherlockNoMad: Balthasar: prefer to not have this function, given how short it is — it's 2 lines (`model_config = model_spec.model` + `update_from_config`). No need to modularize into small functions which are not reused in a meaningful way. The callers can do this inline.

Same concern applies to `build_model_on_meta` below — it's a `with` block + `build()`. The indirection doesn't reduce cognitive load, it just moves 3 lines behind a name.

`materialize_model` is the one extraction here that actually consolidates non-trivial logic (`to_empty` + context manager + `no_grad` + `init_weights` + `train()`) across multiple callsites. I'd keep that and drop the other thin wrappers.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+from torchtitan.protocols import BaseModel, ModelSpec, Module
+from torchtitan.protocols.model_converter import ModelConvertersContainer
+from torchtitan.tools import utils
+
+
+def prepare_model_config(
+    *,
+    model_spec: ModelSpec,
+    trainer_config: Any,
+) -> BaseModel.Config:
```
SherlockNoMad: Balthasar: what's the consideration of typing this as `Any`? If we do keep this function, the type should be more specific — either a Protocol that declares the fields this function actually reads from, or the concrete config type.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+    dump_folder: str,
+) -> Module:
+    """Apply the model spec's non-pipelined parallelization path."""
+    return cast(
+        Module,
+        model_spec.parallelize_fn(
+            model,
+            parallel_dims=parallel_dims,
+            training=training,
+            model_converters=model_converters,
```
SherlockNoMad: Balthasar: This is essentially a pass-through that wraps `model_spec.parallelize_fn(...)` with a `cast(Module, ...)`. Is the indirection worth it? The caller already has the model_spec — wrapping a 1-line delegation in a function doesn't reduce duplication in a meaningful way. If the goal is type safety via the `cast`, that should probably live in the protocol/spec definition, not in a wrapper.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+            parallelism=parallelism,
+            compile_config=compile_config,
+            ac_config=ac_config,
+            dump_folder=dump_folder,
+        ),
+    )
+
```
SherlockNoMad, line 10: Balthasar: Parameter is typed as `Module` but line 91 casts to `BaseModel` to call `init_weights`. If this function always calls `init_weights`, why not type the parameter as `BaseModel` directly? The internal cast is hiding a type requirement from the caller.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+
+def materialize_model(
+    model: Module,
```
SherlockNoMad: Balthasar: Is `init_weights_context` only used by `precompile_main`? If so, we're adding API surface (optional context manager parameter + the `nullcontext` fallback) to serve a single caller. Could precompile_main handle the context externally and call a simpler `materialize_model` without this parameter? e.g.:
```python
with dist_config.patch("compile_on_one_rank", False):
    materialize_model(model, init_device=..., buffer_device=...)
```

torchtitan/trainer.py
```diff
@@ -375,12 +379,11 @@ def __init__(self, config: Config):
             del model
 
             for m in self.model_parts:
-                m.to_empty(device=init_device)
-                with torch.no_grad():
-                    # TODO: Change this back to init_weights once
-                    # autoparallel contains the wrap_init_states
-                    cast(BaseModel, m).init_weights(buffer_device=buffer_device)
-                m.train()
+                materialize_model(
```
SherlockNoMad, line 376: Balthasar: `cast(BaseModel, m)` here, and then `materialize_model` internally casts to `BaseModel` again (line 91 of model_setup.py). The double cast suggests the typing is confused — if the function parameter were typed as `BaseModel` (as I suggested on the other file), the callsite cast would be the only one needed and the intent is clearer.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+from torchtitan.distributed import ParallelDims
+from torchtitan.protocols import BaseModel, ModelSpec, Module
+from torchtitan.protocols.model_converter import ModelConvertersContainer
+from torchtitan.tools import utils
+
+
+def prepare_model_config(
+    *,
+    model_spec: ModelSpec,
+    trainer_config: Any,
```
SherlockNoMad: Casper: `trainer_config: Any` is pretty loose — is there a protocol or base type we could use here? iiuc the only real contract is that `update_from_config` knows how to pull fields from it. `Any` makes it easy for callers to pass something that doesn't satisfy the implicit requirements.

torchtitan/model_setup.py
```diff
@@ -0,0 +1,96 @@
...
+    parallel_dims: ParallelDims,
+    training: TrainingConfig,
+    model_converters: ModelConvertersContainer.Config,
+    parallelism: ParallelismConfig,
+    compile_config: CompileConfig,
+    ac_config: ActivationCheckpointConfig,
+    dump_folder: str,
+) -> Module:
+    """Apply the model spec's non-pipelined parallelization path."""
+    return cast(
```
SherlockNoMad: Casper: why do we need the `cast(Module, ...)` here? is the return type of `parallelize_fn` not `Module` already? if it's a type annotation gap upstream, might be cleaner to fix it there rather than casting at every call site.

torchtitan/model_setup.py
SherlockNoMad: Casper: thinking out loud — the signature says `model: Module` but we call `cast(BaseModel, model).init_weights(...)` on L94. should this just take `BaseModel`?

i see the PP path in `trainer.py` passes `cast(BaseModel, m)` where `m` might only be a `Module` at the type level — so maybe the loose type is intentional for that case. if so, a brief docstring noting the `BaseModel` assumption would help.

torchtitan/experiments/graph_trainer/precompile_main.py
SherlockNoMad: Casper: nice, much cleaner than the old try/finally. just checking — `dist_config.patch` restores the original value on `__exit__` even if `init_weights` raises?

torchtitan/model_setup.py
bobrenjc93: Done — dropped `prepare_model_config`, `build_model_on_meta`, and `parallelize_model`. All three were trivial wrappers. Only `materialize_model` is kept since it consolidates non-trivial logic (`to_empty` + `no_grad` + `init_weights` + `train()`).

torchtitan/model_setup.py
bobrenjc93: Done — dropped this function. Callsites now call `model_spec.parallelize_fn(...)` directly.

torchtitan/model_setup.py
bobrenjc93: Fixed — `materialize_model` now takes `model: BaseModel` directly, so the internal cast is gone.

torchtitan/trainer.py
bobrenjc93: The PP path was already using `materialize_model` — it's the non-PP path that previously had the inline version. Both paths use `materialize_model` now. The PP callsite does `cast(BaseModel, m)` since `model_parts` is typed as `list[torch.nn.Module]`.

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: `.patch()` comes from `install_config_module(sys.modules[__name__])` at the bottom of `torch/distributed/config.py`. It's a standard PyTorch config module utility that temporarily sets a config value and restores it on `__exit__` (even if an exception is raised). That said, this is now moot — I removed the `init_weights_context` parameter and the callsite in `precompile_main` wraps externally instead:
```python
with dist_config.patch("compile_on_one_rank", False):
    materialize_model(model, init_device=..., buffer_device=...)
```

tests/unit_tests/test_model_setup.py
bobrenjc93: Rewrote the tests. They now verify:
- Parameters actually move from meta → cpu
- `init_weights` is called with correct `buffer_device`
- Weights are actually initialized (not uninitialized garbage)
- Model is in training mode after materialization

tests/unit_tests/test_model_setup.py
bobrenjc93: Added `test_works_in_loop_for_pp_pattern` which calls `materialize_model` in a loop over multiple models. And yes, this test runs in CI — `unit_test_cpu.yaml` runs `pytest tests/unit_tests` which picks up this file.

torchtitan/model_setup.py
bobrenjc93: Agreed — dropped all three thin wrappers. Only `materialize_model` remains.

torchtitan/model_setup.py
bobrenjc93: Moot now — function removed.

torchtitan/model_setup.py
bobrenjc93: Agreed — dropped it. Callsites call `model_spec.parallelize_fn(...)` directly.

torchtitan/model_setup.py
bobrenjc93: Fixed — `materialize_model` now takes `model: BaseModel` directly. No more internal cast.

torchtitan/model_setup.py
bobrenjc93: Good call — removed the `init_weights_context` parameter. `precompile_main` now wraps externally exactly as you suggested.

torchtitan/trainer.py
bobrenjc93: Fixed — `materialize_model` now takes `BaseModel` directly, so there's only the single `cast(BaseModel, m)` at the callsite in the PP loop.

torchtitan/model_setup.py
bobrenjc93: Moot now — function removed.

torchtitan/model_setup.py
bobrenjc93: Moot now — function removed. The `parallelize_fn` return type is untyped (`Callable`), so the cast lives at the callsite in `trainer.py` now.

torchtitan/model_setup.py
bobrenjc93: Fixed — `materialize_model` now takes `model: BaseModel` directly. No more mismatch between the signature and internal usage.

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: Yes — `patch()` comes from `install_config_module` in `torch/distributed/config.py` and uses a standard context manager that restores the original value on `__exit__` regardless of exceptions. The `init_weights_context` parameter is now removed though — `precompile_main` wraps externally with `with dist_config.patch(...):` around the `materialize_model` call.

torchtitan/experiments/graph_trainer/precompile_main.py
SherlockNoMad, line 16: This is a directional comment, without me knowing the underlying constrain of the precompile implementation. 

I imagine an ideal entry for precompile is "not introducing a new precompile_main, but instead a flag for enabling precompile in the existing path".

So it should be something like "<original_graph_trainer_commands> --compile.precompile_artifact_dir /tmp/precompile_artifacts" 

Under the hood, it's just reusing the entire GraphTrainer init path, swapping the FakeProcessGroup so it's single gpu only, apply the necessary parameterization pass so the graph is rank symbolic, finally branch off from before running the 2nd training step. 
1 step is still ran, which triggers the compile, and dump.

SherlockNoMad: I wonder what are the implementation gaps that prevents us from doing so...

PR #2744, opened 2026-03-28
Gh/pianpwk/4/base

shuhuayu:
Resubmit the enable/disable sp for tp PR (https://github.com/pytorch/torchtitan/pull/2535) to unblock.


shuhuayu:
For `qk_norm_plan`, disabled `use_local_output=enable_sp` and set it as default `False`. This is because unlike other norms, qk norm happens within tp region and need to stay as dtensor. Can you help check? @tianyu-l .

tianyu-l:
@shuhuayu sounds fine, we should anyway use all dtensor after https://github.com/pytorch/torchtitan/pull/2149

PR #2750, opened 2026-03-30
[RL] Make reference model and KL penalty optional via kl_coef

daniellepintz:
## Summary
  - Add `kl_coef` config field to `RLTrainer.Config` (default 0.0)
  - When `kl_coef > 0`, a frozen reference model is built and KL divergence with PPO clipping is added to the policy gradient loss
  - When `kl_coef == 0` (default), no reference model is built, saving GPU memory and compute
  - Make `ref_token_log_probs` optional in `compute_policy_gradient_loss` — when not provided, uses a simple policy gradient loss with entropy bonus

  ## Test plan
  - [x] Run with default config (`kl_coef=0`): verify no ref model is built and training works without KL penalty
  - [x] Run with `--kl_coef 0.1`: verify ref model is built and KL divergence appears in loss metrics
  - [x] Verify loss metrics include `kl_div`, `ratio_mean`, `ratio_clipped_frac` only when `kl_coef > 0`

### DIFF
--- a/.github/workflows/integration_test_4gpu_rl_h100.yaml
+++ b/.github/workflows/integration_test_4gpu_rl_h100.yaml
@@ -61,17 +61,20 @@ jobs:
         # 2. Install batch-invariant ops
         uv pip install --no-deps "git+https://github.com/thinking-machines-lab/batch_invariant_ops.git@main"
 
-        # 3. Install PyTorch nightly, vllm, and xformers
+        # 3. Install Flash Attention 3 (required for H100/H200)
+        uv pip install flash-attn-3 --extra-index-url=https://download.pytorch.org/whl/test/cu128
+
+        # 4. Install PyTorch nightly, vllm, and xformers
         uv pip install torch vllm xformers --pre \
           --extra-index-url https://download.pytorch.org/whl/nightly/cu128 \
           --index-strategy unsafe-best-match
         uv pip install torchdata==0.12.0.dev20260327 --extra-index-url https://download.pytorch.org/whl/nightly/cpu
 
-        # 4. Install torchtitan in editable mode because integration test spawns
+        # 5. Install torchtitan in editable mode because integration test spawns
         # python simple_grpo_sum_digits.py as a subprocess
         uv pip install -e .
 
-        # 5. Download HF model checkpoint for tests
+        # 6. Download HF model checkpoint for tests
         MODEL_PATH=$(python -c "from huggingface_hub import snapshot_download; print(snapshot_download('Qwen/Qwen3-0.6B'))")
 
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
--- a/torchtitan/experiments/rl/models/attention.py
+++ b/torchtitan/experiments/rl/models/attention.py
@@ -63,6 +63,8 @@ class PyTorchVarlenAttentionImpl(FlashAttentionImpl):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
+        self.enable_gqa = self.num_heads > self.num_kv_heads
+
         # Hopper (SM 9.0) uses FA3
         if has_cuda_capability(9, 0):
             # activate_flash_attention_impl() will restore internal global state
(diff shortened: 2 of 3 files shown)


torchtitan/experiments/rl/actors/utils.py
```diff
@@ -89,13 +89,13 @@ def compute_policy_gradient_loss(
     vllm_token_ids: list[list[int]],
     prompt_token_ids: list[list[int]],
     advantages: torch.Tensor,
-    ref_token_log_probs: list[torch.Tensor],
+    ref_token_log_probs: list[torch.Tensor] | None = None,
```
wwwjn, line 91: It's ok to remove the ref_token_log_probs in this field and switch to REINFORCE entirely. We should update all the file names including README, remove `GRPO` and explicitly tell user we are using REINFORCE

daniellepintz: @wwwjn if we just remove the ref model logprobs, it will still be GRPO I believe, as we are still doing group-relative rewards. and GRPO without ref model is valid. 

sorry for the confusion, I have changed this wording in PR summary.

but yes I will remove the ref model code entirely

wwwjn: The reference model output is used to calculate the KL divergence between policy model and ref model, can you send me some reference of GRPO without ref model?

I think we can also switch to REINFORCE entirely, generating 1 completion per sample instead of generating `n_samples_per_group` can also reduce memory pressure on generator side.

daniellepintz: Yes, here is a reference from [TRL docs](https://huggingface.co/docs/trl/main/en/grpo_trainer): 

> Note that compared to the original formulation in [DeepSeekMath: Pushing the Limits of Mathematical Reasoning in Open Language Models](https://huggingface.co/papers/2402.03300), we use β=0.0 by default, meaning that the KL divergence term is not used. This choice is motivated by several recent studies (e.g., [Open-Reasoner-Zero: An Open Source Approach to Scaling Up Reinforcement Learning on the Base Model](https://huggingface.co/papers/2503.24290)) which have shown that the KL divergence term is not essential for training with GRPO. As a result, it has become common practice to exclude it (e.g. [Understanding R1-Zero-Like Training: A Critical Perspective](https://huggingface.co/papers/2503.20783), [DAPO: An Open-Source LLM Reinforcement Learning System at Scale](https://huggingface.co/papers/2503.14476)). If you wish to include the KL divergence term, you can set beta in [GRPOConfig](https://huggingface.co/docs/trl/main/en/grpo_trainer#trl.GRPOConfig) to a non-zero value.

regarding:
> I think we can also switch to REINFORCE entirely

I am open to that, but can that be in a later PR, as it requires more testing and updating more places? This PR can be a first step?

wwwjn: Yeah we can switch to REINFORCE later. 

> If you wish to include the KL divergence term, you can set beta in [GRPOConfig](https://huggingface.co/docs/trl/main/en/grpo_trainer#trl.GRPOConfig) to a non-zero value.


Thanks for the reference and after ready this,  I think we can also quickly make this configurable like TRL, eg: have a `grpo_beta` field in RLTrainer, and if it's 0, we skill the reference model initialization. The rational of doing this is make sure we faithfully follow the GRPO algorithm which aligns with users' expectation

daniellepintz: Done! Used the name `kl_coef`, lmk if that looks good :)

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -77,9 +77,11 @@ def __init__(
         model_spec: ModelSpec,
         hf_assets_path: str = "",
         transfer_dtype: str = "",
+        kl_coef: float = 0.0,
```
tianyu-l, line 80: Is it general enough to be put in trainer constructor kwargs? I somehow feel `kl_coef` should be part of `Loss` which itself should be a `Configurable` class.

Previous we decided not to worry about configuring algorithm / loss, now it seems we've come to the point (or have we?). If so, I think we need better encapsulation, as trainer should be general enough to handle configurable loss (in this case, based on `kl_coef`) rather than hardcoded fields.

cc @wwwjn @joecummings

joecummings: 100% - put this in #2906

PR #2758, opened 2026-03-30
Remove varlen compile

wwwjn:
Benchmarked on qwen3 debug model (might be too small)
- Peak MFU is ~17% both with and without torch.compile on varlen

### DIFF
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -150,8 +150,6 @@ class LocalMapAttention(Module):
 
 
 class VarlenAttentionWrapper(LocalMapAttention):
-    _varlen_attn: ClassVar[Callable] = varlen_attn
-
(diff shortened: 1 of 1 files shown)


wwwjn:
@liangel-02  do you think we need more benchmarking to see if removing compile hurts perf? Or do you have some benchmarking data? Thanks

torchtitan/models/common/attention.py
```diff
@@ -153,9 +150,7 @@ def forward(
 
 
 class VarlenAttentionWrapper(LocalMapAttention):
-    _compiled_varlen_attn: ClassVar[Callable] = torch.compile(
-        varlen_attn, mode="max-autotune-no-cudagraphs"
-    )
+    _varlen_attn: ClassVar[Callable] = varlen_attn
```
tianyu-l: can we just remove this class var and call `varlen_attn` directly?

PR #2761, opened 2026-03-31
refactor inner attention module

tianyu-l:
for several things:
- Save a pair of transpose for varlen attention, by moving the transpose for sdpa / flex inside their own module. I had to change `_ContextParallel(seq_dim=2, ...)` to `seq_dim=1` -> I also had to change flux model impl to adapt.
- The above change follows https://github.com/pytorch/torchtitan/pull/2709 which is only for the generator. With this PR I'm able to remove `VLLMGQAttention` module added in that PR.
- Make `inner_attention` a python config, instead of string `attn_backend` which is now used in qwen3 `model_registry` as a convenient way to switch config.
- Thanks to that, we can support FlexAttention with configurable options, including `block_size` and `kernel_options` to enable Flash attention backend https://github.com/pytorch/torchtitan/pull/2171. Added a config `qwen3_debugmodel_flex_flash` to the registry.
- Move `annotate_flex_attention_for_regional_inductor` from general attention file to `graph_trainer` where it's used.
- remove outdated artificial restrictions on CP / CP + TP
- some naming improvement

NOTE:
- GraphTrainer tests fail on main as well
- Model test on Llama4+PP+compile fails on main as well


tianyu-l:
For the RL CI failure
> ValueError: use_strided_shard_as_shard_order is True, but placements: (_StridedShard(dim=1, sf=8),) is unable to be interpreted into a corresponding shard_order

I don't know why strided shard showed up, and I couldn't reproduce on pretraining Qwen3 0.6B with NGPU=2, TP=2.
cc @wwwjn @weifengpy @zpcore

weifengpy:
> > ValueError: use_strided_shard_as_shard_order is True, but placements: (_StridedShard(dim=1, sf=8),) is unable to be interpreted into a corresponding shard_order

a quick fix can be https://github.com/pytorch/pytorch/pull/178735

> I don't know why strided shard showed up, and I couldn't reproduce on pretraining Qwen3 0.6B with NGPU=2, TP=2. cc @wwwjn @weifengpy @zpcore

what's the config you are using to repro?

torchtitan/models/common/attention.py
```diff
@@ -187,14 +189,14 @@ def forward(
         xk_packed = xk_packed.to(torch.bfloat16)
         xv_packed = xv_packed.to(torch.bfloat16)
 
-        return varlen_attn(
+        out = varlen_attn(
```
wwwjn: What is the shape of varlen attention output? Is it (bs*seq, heads, dim)?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -290,84 +299,7 @@ def forward(
...
-            if self.rope_backend == "cos_sin":
-                xq, xk = apply_rotary_emb_cos_sin(xq, xk, rope_cache, positions)
-            else:
-                xq, xk = apply_rotary_emb_complex(
-                    xq, xk, freqs_cis=rope_cache, positions=positions
-                )
-
-        output = self.inner_attention(xq, xk, xv)
+        # Reshape back to the format expected by GQAttention.forward()
+        output = output_flat.view(batch_size, seq_len, -1, head_dim)
```
wwwjn, line 303: Here VLLMAttentionWrapper return's a 4D output, while VarlenAttention returns a 3D output.

I think the bug might because RL is using full DTensor in TP region , and used LocalMapAttention https://github.com/pytorch/torchtitan/blob/b01adfb544b4331ecab090ebdb50b2296cd8eb6a/torchtitan/models/common/attention.py#L74. In this function, the `out_placement` is inferred from input (q/k/v)'s placement.  


Previous:
- In RL's trainer model, when TP is enabled, the input placement is Shard(1) (inner_attention gets (bsz, heads, seq, dim)), so the output varlen's placement will also be shard(1). 
- The VarlenAttention's output has shape (bsz* seq, heads, dim) with shard(1) placement.


After this PR:
- Inner attention now gets (bsz, seq, heads, dim) as input, the input placement is Shard(2). So the LocalMapAttention will infer the `output_placement` to be Shard(2) as well. 
- The VarlenAttention's output still has shape (bsz* seq, heads, dim), but Shard(2) is a wrong placement

torchtitan/models/common/attention.py
wwwjn: From the doc here: https://docs.pytorch.org/docs/stable/nn.attention.varlen.html yes it's (total_tokens, heads, dim)

torchtitan/distributed/context_parallel.py
```diff
@@ -27,54 +27,43 @@
...
     cp_mesh: DeviceMesh,
-    attention_type: str,
 ) -> None:
     """
     Apply context parallelism to attention modules.
 
     CP splits the sequence dimension across devices to enable training with
     longer sequences. This function applies CP to the provided attention
-    modules.
+    modules. The attention type is inferred via isinstance on the first module.
```
drisspg, line 36: hmm any issues with this for hybrid models

torchtitan/models/common/attention.py
```diff
@@ -238,83 +246,74 @@ class FlexAttentionWrapper(LocalMapAttention):
...
-        out, aux = FlexAttentionWrapper._compiled_flex_attn(
+        #    `FlexAttention._compiled_flex_attn` is correct.
+        out, aux = FlexAttention._compiled_flex_attn(
             q,
             k,
             v,
-            block_mask=block_mask,
+            block_mask=attention_masks,
             scale=scale,
             enable_gqa=enable_gqa,
```
drisspg, line 289: want to double check, we are now universally setting enbale_gqa = True instead of manually running repeat interleave right?

drisspg:
I think one nit is it would be nice to have Some Readme somewhere in Attention that describes a little about how inner config passing works / where they are expected to be constructed. As well like lets say you want to write a new model def that uses attention impl Y, here is how you should do it instead of reverse engineering from other defs

torchtitan/models/common/attention.py
tianyu-l: Yes, we are now running `enable_gqa` as long as `self.n_heads > self.n_kv_heads`.

torchtitan/distributed/context_parallel.py
tianyu-l: Right now in torchtitan we don't have hybrid attention. Also we are going to refactor CP so this part will be gone.

torchtitan/experiments/rl/models/attention.py
tianyu-l: oh mismatch should be the reason, let me retry.

tianyu-l:
> I think one nit is it would be nice to have Some Readme somewhere in Attention that describes a little about how inner config passing works / where they are expected to be constructed. As well like lets say you want to write a new model def that uses attention impl Y, here is how you should do it instead of reverse engineering from other defs

@drisspg Definitely! But maybe after this wave of refactor? We are doing massive change to the codebase, especially on config system, so any doc would easily become outdated and a debt to maintain.

torchtitan/experiments/rl/tests/integration_tests.py
```diff
@@ -38,6 +38,7 @@ def build_rl_test_list() -> list[OverrideDefinitions]:
                     "--generator.parallelism.tensor_parallel_degree 2",
                     "--generator.num_samples_per_prompt 2",
                     "--no_batch_invariant_mode",
+                    "--trainer.compile.no-enable",
```
wwwjn, line 41: Why explicitly disable trainer's compile here? cause it's  by default enabled?

torchtitan/models/llama3/config_registry.py
```diff
@@ -226,11 +226,13 @@ def process_sample(sample):
             {"role": "assistant", "content": sample["answer"]},
         ]
 
+    from torchtitan.models.common import FlexAttention
+
     model_spec = model_registry("debugmodel")
```
wwwjn, line 231: We already have attention_backend_override in `model_registry`, we should override

tianyu-l: I don't think we have that in llama3. We only have it in qwen3. We can work in later PR to make style consistent.

torchtitan/experiments/rl/tests/integration_tests.py
tianyu-l: Because this test is about no_compile. Generator is already no compile, this PR makes trainer consistent.

xmfan (approved):

torchtitan/experiments/rl/models/attention.py
weifengpy: meaning _StridedShard shouldn't show up?

tianyu-l: Right

PR #2766, opened 2026-03-31
[graph_trainer] Add torch.no_grad() and graph-based SAC to traced execution

tugsbayasgalan:
* __->__ #2766

Execute traced graph under torch.no_grad() since it already contains
explicit backward ops (from torch.autograd.grad traced by make_fx).
Without this, PyTorch builds a redundant autograd graph keeping all
forward intermediates alive via grad_fn references.

Adds SAC option on aot_fx_trace on GraphTrainer. 

Adds option to dump the peak memory from CUDA caching allocator. 

Adds similar test as test_bitwise_equivalency.py for peak memory 

Here is the run comparison between eager and aot_fx_trace with SAC
Screenshot 2026-04-14 at 1 18 41 PM

Llama3B 8b with dp=4 tp=2 on 10 step run before SAC **with cuda graph** seq_len=1024
Screenshot 2026-04-14 at 10 39 41 PM

Llama3B 8b with dp=4 tp=2 on 10 step run after SAC **with cuda graph** seq_len=1024
Screenshot 2026-04-14 at 10 39 26 PM

Llama3B 8b with dp=4 tp=2 on 10 step run before SAC **without cuda graph** with seq_len=8192
Screenshot 2026-04-15 at 4 07 50 PM

Llama3B 8b with dp=4 tp=2 on 10 step run after SAC **without cuda graph** with seq_len=8192
Screenshot 2026-04-15 at 4 07 43 PM

Deepseek3 16B before SAC 
Screenshot 2026-04-14 at 10 54 31 PM

Deepseek3 16B after SAC 

Screenshot 2026-04-14 at 10 58 30 PM


torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -480,8 +542,23 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
+        ac_id = node.meta.get("custom", {}).get("ac_graph_id")
+        if ac_id is not None:
+            node.meta["ac_graph_id"] = ac_id
+
     _remove_cpu_shadow_chains(traced)
 
+    # Rematerialize activations tagged PREFER_RECOMPUTE by selective AC.
+    # Duplicates recomputable forward ops before the backward region and
+    # DCEs the original copies, reducing peak memory.
+    traced = remat_using_tags_for_fwd_loss_bwd_graph(traced)
```
SherlockNoMad: No, user should apply this if they want. 

keep the tracing simple.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -480,8 +542,23 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
 
+    # FIXME: Promote ac_graph_id from custom dict to top-level meta so that
+    # cleanup_recompute_tags (called by the remat pass) can find it.
+    # torch.fx.traceback.annotate puts values under meta["custom"], but
+    # cleanup_recompute_tags reads meta["ac_graph_id"] directly.
+    # Remove once https://github.com/pytorch/pytorch/pull/175348 lands.
+    for node in traced.graph.nodes:
+        ac_id = node.meta.get("custom", {}).get("ac_graph_id")
+        if ac_id is not None:
+            node.meta["ac_graph_id"] = ac_id
```
SherlockNoMad: again, remove this. should be a post processing pass in user's control

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: Isn't it kinda weird that user annotates nodes with AC and don't see any improvement in the end? It feels like silent incorrectness to me.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -27,6 +30,46 @@
 # nodes/constants or captured in fn's closure.
 _ALLOWED_LEAF_TYPES = (torch.Tensor, int, float, bool, str, type(None))
 
+# Counter for assigning unique ac_graph_id to each CheckpointWrapper region.
+_ac_graph_id_counter = 0
+
+
+@contextmanager
+def _patch_checkpoint_wrapper_ac_graph_id() -> Generator[None, None, None]:
```
SherlockNoMad: No, let's have a clear separation of "tracer code" and "trainer code". 

ac related optimization is clearly a user decision, and should live in the trainer code.

SherlockNoMad (changes requested):
The only change needed is 

```
        with torch.no_grad():
            flat_outputs = self.gm(*flat_inputs)
``` 

All other changes shouldn't live in tracer code.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -27,6 +30,46 @@
...
+    )
+
+    global _ac_graph_id_counter
+    _ac_graph_id_counter = 0
+    _orig_forward = CheckpointWrapper.forward
+
+    def _patched_forward(self, *args, **kwargs):
+        global _ac_graph_id_counter
+        _ac_graph_id_counter += 1
+        with torch.fx.traceback.annotate({"ac_graph_id": _ac_graph_id_counter}):
```
anijain2305: @tugsbayasgalan  can you check if we don't do anything - does the tracer do the right thing to just recompute the forward into the backward? Because if yes, then we dont need anything.

tugsbayasgalan: Here is the full summary: https://github.com/pytorch/pytorch/issues/178935

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -9,16 +9,58 @@
...
 from torchtitan.experiments.graph_trainer.cudagraph import cudagraph_teardown
 from torchtitan.experiments.graph_trainer.make_fx_tracer import (
     minimal_fx_tracer,
     TracedResult,
 )
+from torchtitan.experiments.graph_trainer.passes import apply_sac_pass
 from torchtitan.trainer import Trainer
 
 
+def _strip_recompute_from_backward_nodes(gm: torch.fx.GraphModule) -> None:
```
yiming0416: Not sure if this function should stay in `trainer.py`

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -9,16 +9,58 @@
...
+    for node in gm.graph.nodes:
+        if node.op != "call_function":
+            continue
+        custom = node.meta.get("custom", {})
+        if custom.get("remat_pass_tag") == "is_backward":
+            node.meta.pop("recompute", None)
+            node.meta.pop("ac_graph_id", None)
+
+
+def apply_ac_remat_pass(traced: TracedResult) -> None:
```
yiming0416: Similarly, this shouldn't be in `trainer.py`, we have a dedicated `passes.py` in `experiments/graph_trainer/`, which contains the passes for the traditional JIT and AOT graphs (dynamo+AOT Autograd).

I imagine the passes will be different for the make_fx tracer, but they might share common components.

tugsbayasgalan: Sounds good.

torchtitan/experiments/graph_trainer/trainer.py
tugsbayasgalan: moved to passes

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -244,6 +244,47 @@ def apply_sac_pass(
...
+    for node in gm.graph.nodes:
+        if node.op != "call_function":
+            continue
+        custom = node.meta.get("custom", {})
+        if custom.get("remat_pass_tag") == "is_backward":
+            node.meta.pop("recompute", None)
+            node.meta.pop("ac_graph_id", None)
+
+
+def apply_ac_remat_pass(traced: "TracedResult") -> None:
```
SherlockNoMad: follow the pass signature, always use graph module

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -235,7 +241,8 @@ def _patched(t_outputs, *args, **kwargs):  # type: ignore[no-untyped-def]
         ]
         if roots:
             setup_stacktrace_preservation_hooks(roots)
-        return _orig_fn(t_outputs, *args, **kwargs)
+        with torch.fx.traceback.annotate({"remat_pass_tag": "is_backward"}):
```
SherlockNoMad: hmm... 
Doesn't this add the annotation for forward nodes?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -244,6 +244,47 @@ def apply_sac_pass(
     return gm
 
 
+def _strip_recompute_from_backward_nodes(gm: torch.fx.GraphModule) -> None:
+    """Remove recompute tags from backward nodes.
+
+    apply_sac_pass tags ALL call_function nodes, but backward nodes
```
SherlockNoMad: The root of problem is "apply_sac_pass tags ALL call_function", let's avoid adhoc point fixes. 

Fix at apply_sac_pass, to avoid tagging backward nodes in the first place.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -111,6 +114,7 @@ def _make_fx_forward_backward_step(
                         extra_kwargs,
                     ),
                 )
+            apply_ac_remat_pass(self._traced_step)
```
SherlockNoMad: @yiming0416 , do we have a graph pass manager? 

Abstract this into an `def optimize(...)` function

torchtitan/experiments/graph_trainer/make_fx_tracer.py
SherlockNoMad: also, we need to get systematic about annotations, as many graph passes are now relying on annotations. 

I think marking node as backward is generally useful, and doesn't only benefit "remat_pass_tag“. 
So I would think of a more general annotation.

SherlockNoMad (changes requested):

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: This is only invoked under backward engine and yeah i think we should upstream this to pytorch.

torchtitan/experiments/graph_trainer/passes.py
tugsbayasgalan: done

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -111,6 +129,8 @@ def _make_fx_forward_backward_step(
                         extra_kwargs,
                     ),
                 )
+            self._traced_step.gm = apply_ac_remat_pass(self._traced_step.gm)
+            apply_regional_inductor(self._traced_step)
```
yiming0416: Are the ac pass and regional inductor always on? Shall we have flag or config for user to control?

We have `compile.joint_passes` and `compile.passes` configs for the old AOT/JIT workflow, so that users can explicit indicate which pass they want to apply

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -244,6 +251,59 @@ def apply_sac_pass(
...
+    """
+    from torch._functorch._activation_checkpointing.remat_using_tags_for_fwd_loss_bwd_graph_pass import (
+        remat_using_tags_for_fwd_loss_bwd_graph,
+    )
+
+    apply_sac_pass(gm)
+    return remat_using_tags_for_fwd_loss_bwd_graph(gm)
+
+
+def apply_regional_inductor(traced: "TracedResult") -> None:
```
yiming0416: it's a bit confusing to have another `apply_regional_inductor` as we already have `regional_inductor_pass`.

The root problem here is that graphs captured by `aot_export_joint` and `make_fx` are different, and it requires changes for passes to work on different graphs. I am concerned that each existing pass will have a variant for the `make_fx` tracer and I wonder what's a good way to manage this. cc @SherlockNoMad

torchtitan/experiments/graph_trainer/trainer.py
yiming0416: We have `get_compiler_passes_from_config` in `graph_utils.py`

I don't think the passes should be explicitly applied in the trainer, they should be abstracted.

torchtitan/experiments/graph_trainer/passes.py
tugsbayasgalan: Sorry this is claude doing when i was debugging sth else, accidentally pushed. i don't think we need this.

torchtitan/experiments/graph_trainer/trainer.py
tugsbayasgalan: no this was accidentally added.

tugsbayasgalan:
@claude Review this PR. Did I address all the feedbacks?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -111,6 +122,10 @@ def _make_fx_forward_backward_step(
                         extra_kwargs,
                     ),
                 )
+            self._traced_step.gm = optimize_fwd_bwd_graph(
```
tugsbayasgalan: @SherlockNoMad , @yiming0416  do you guys have any suggestions on how to make it into a proper config?

tugsbayasgalan:
@claude Give review. 
1) Is the sac peak test minimal enough? 
2) Am i following torchtitan coding format? 
3) Does the backward phase tagging make sense? 
4) Is the sac pass plumbing make sense?

tests/utils.py
```diff
@@ -11,6 +11,11 @@
 from torch.distributed.tensor import DTensor
 
 
+def _canonicalize_tensor_name(name: str) -> str:
```
tugsbayasgalan: needed when comparing eager + AC vs graph + SAC because eager wraps the layers with _checkpoint_wrapped_module

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -401,18 +352,22 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
 
         ctx = TracingContext(fake_mode)
         # preserve_node_meta propagates fx.traceback.annotate metadata to traced nodes
-        with fake_mode, tracing(ctx), preserve_node_meta(), _skip_nested_compile():
+        with (
+            fake_mode,
+            tracing(ctx),
+            preserve_node_meta(),
+            _skip_nested_compile(),
+            torch.compiler._non_strict_tracing_context(),
```
SherlockNoMad, line 385: comment on why it needs torch.compiler._non_strict_tracing_context() ? 
 and why it needs _skip_nested_compile...

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -594,6 +610,25 @@ def apply_sac_pass(
     return gm
 
 
+def apply_ac_on_fwd_bwd_graph(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
```
SherlockNoMad: did you author this with claude? 

it should follow the graph signature in clude.md closely.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -249,6 +253,13 @@ def apply_graph_passes(
     return gm
 
 
+def graph_ac_pass(
+    gm: torch.fx.GraphModule, example_inputs: tuple | None = None
+) -> torch.fx.GraphModule:
+    """Apply graph-based SAC to a traced fwd+loss+bwd graph."""
+    return apply_ac_on_fwd_bwd_graph(gm)
```
SherlockNoMad: why dummy wrapping

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -128,7 +135,10 @@ def _make_fx_forward_backward_step(
                 )
 
             if self.config.compile.enable_passes:
-                passes = construct_default_graph_passes(self._traced_step)
+                passes = []
+                if enable_graph_ac:
+                    passes.append(graph_ac_pass)
```
SherlockNoMad: move apply_ac_on_fwd_bwd_graph inside construct_default_graph_passes

I am fine with no introducing the option of enable_graph_ac for now, we can worry about he UX of exposing configs for pass latter.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -116,6 +119,10 @@ def _make_fx_forward_backward_step(
     ) -> torch.Tensor:
         if self._traced_step is None:
             fwd_bwd_fn = make_fwd_bwd_step(self.loss_fn)
+            ac_mode = self.config.activation_checkpoint.mode
+            enable_graph_ac = enable_graph_ac_for_mode(ac_mode)
+            if enable_graph_ac:
+                annotate_ac_regions(model)
```
SherlockNoMad: we should not have per-pass specific code in this file, keep trainer clean

SherlockNoMad: don't need annotate_ac_regions, this should have been called in parallelize.py

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -542,6 +553,13 @@ def apply_sac_pass(
         if node.op != "call_function":
             continue
 
+        custom_meta = node.meta.get("custom", {})
```
SherlockNoMad: move back for better locality

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -450,6 +405,10 @@ def run_traced(
     This is a reference implementation of traced-graph execution. It keeps the
     state handling, subclass unwrapping, and output reconstruction logic
     explicit instead of baking those semantics into ``TracedResult`` itself.
+    Runs under ``torch.no_grad()`` because the graph already contains explicit
```
SherlockNoMad, line 427: nice comment

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -155,20 +155,29 @@ def convert_modules_to_fqns(modules, module_to_fqn_mapping):
     return module_fqns
 
 
+def enable_graph_ac_for_mode(ac_mode: str) -> bool:
```
SherlockNoMad: remove this, just always assume selective for now. 

we will worry about what's the right UX for "what to AC" latter

SherlockNoMad:
I am looking for memory snapshots figures for runs before and after apply SAC.

tests/utils.py
SherlockNoMad: IIUC, SAC shouldn't change numerics. 

So comparing Eager without AC vs. Graph with SAC is also fine.  

Let's not have this adhoc string replacement, as it's brittle.

tests/integration_tests/run_tests.py
```diff
@@ -193,6 +278,12 @@ def main():
         default=None,
         help="Comma-separated list of test names to skip",
     )
+    parser.add_argument(
+        "--collect_peak_memory",
+        default=False,
+        action="store_true",
+        help="Collect peak reserved/active CUDA memory from TensorBoard logs.",
```
SherlockNoMad: do you know how memory is measured in tensorboard? 
I am not sure if we want to offload the memory measurement to tensorboard. I think it can be done with < 10 lines of code. See MoE_Sol for reference. 

reading memory from tensorboard seems indirect.

tests/integration_tests/run_tests.py
```diff
@@ -35,11 +41,60 @@ def _run_cmd(cmd, timeout=None):
     )
 
 
+def _read_peak_memory(tb_root: Path) -> dict[str, float | int]:
```
SherlockNoMad: I think the diff is landable without the "reading memory from tensorboard bits". 
I would do it in another PR, so the core pass functionality can land first. 

I would also like to see a proof of SAC working in a memory snapshot.

tests/utils.py
tugsbayasgalan: Yeah good point.

tests/integration_tests/run_tests.py
tugsbayasgalan: Removed

torchtitan/experiments/graph_trainer/common_utils.py
tugsbayasgalan: ok

tugsbayasgalan:
@claude What are the failures?

tugsbayasgalan:
@claude what are the failures

tugsbayasgalan:
@claude what are the failures?

tugsbayasgalan:
@claude I am pretty sure there was no RNG ops to be saved few days ago, is this a recent change?

tugsbayasgalan:
@claude What is the test command to reproduce these errors?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -401,18 +352,27 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
-
             flat_outs, output_spec = pytree.tree_flatten(result)
             num_flat_outputs = len(flat_outs)
             unwrapped_outs, output_layouts = _unwrap_subclasses(flat_outs)
             return unwrapped_outs
 
         ctx = TracingContext(fake_mode)
         # preserve_node_meta propagates fx.traceback.annotate metadata to traced nodes
-        with fake_mode, tracing(ctx), preserve_node_meta(), _skip_nested_compile():
+        # _skip_nested_compile lets this tracer run when an outer dynamo trace
```
SherlockNoMad: not "outer dynamo trace", it's the "current make_fx trace"?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -68,6 +72,7 @@ def construct_default_graph_passes(
 
     passes: list[Callable] = [
         functools.partial(tlparse_log_graph_pass, graph_name="make_fx_graph_traced"),
+        apply_ac_on_fwd_bwd_graph,
```
SherlockNoMad: apply after graph clean up pass

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -461,6 +470,27 @@ def apply_sac_pass(
     return gm
 
 
+def apply_ac_on_fwd_bwd_graph(
```
SherlockNoMad: please ask claude to come up with a more tasteful name, also follow naming convention, ends with *_pass

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -461,6 +470,27 @@ def apply_sac_pass(
...
+
+    Tags forward nodes with recompute policy via apply_sac_pass (backward
+    nodes are skipped automatically via ``node.meta["autograd_backward"]``), then
+    applies remat_using_tags_for_fwd_loss_bwd_graph to duplicate
+    PREFER_RECOMPUTE forward ops before backward and DCE originals.
+
+    The model must have been annotated with annotate_ac_regions before
+    tracing so that nodes have custom["ac_region_id"] metadata.
+    """
+    from torch._functorch._activation_checkpointing.remat_using_tags_for_fwd_loss_bwd_graph_pass import (
```
SherlockNoMad, line 529: follow up in future diff, I also think it' better to have remat_using_tags_for_fwd_loss_bwd_graph_pass in titan, since no one it's using it in core, and it's applicable senario is niche.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -401,18 +352,27 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
+        # reaches torch.compile'd FlexAttention kernels.
+        # _non_strict_tracing_context is required by _patch_autograd_grad() and
```
SherlockNoMad, line 376: I am not a fan of needing `_non_strict_tracing_context` coupling with `_patch_autograd_grad`. 
What breaks if we don't have this context? 
Is there other ways to fix this?

tests/integration_tests/run_tests.py
```diff
@@ -91,9 +91,13 @@ def run_single_test(
             )
 
 
-def run_tests(args, test_list: list[OverrideDefinitions], module=None, config=None):
+def run_tests(
```
SherlockNoMad: I don't see meaningful change in this file

SherlockNoMad (changes requested):

torchtitan/experiments/graph_trainer/passes.py
tugsbayasgalan: we do use it in torch.compile(fullgraph=True) today.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: We only want to enable patching autograd in the cases we know it is carefully coupled with rest of the stack. So we added check in patch_autograd that it should only succeed if this specific context is on. The context is _non_strict_tracing_context. For example, we don't want to do this non-strict export.

SherlockNoMad (approved):

tests/integration_tests/run_tests.py
```diff
@@ -93,7 +93,6 @@ def run_single_test(
 
 def run_tests(args, test_list: list[OverrideDefinitions], module=None, config=None):
     """Run all integration tests to test the core features of TorchTitan"""
-
```
fegin, line 96: nit: why delete this line?

fegin (approved):

PR #2769, opened 2026-03-31
Multi-Turn Chat Dataset SFT

haydn-jones:
# Multi-turn SFT with renderers

Added multi-turn SFT support using the renderers library (https://github.com/PrimeIntellect-ai/renderers) as `0.1.11` no longer requires transformers as a dependency. The RL experiment was already using renderers to some degree, so I moved the config / adapter out into `components` so it could be shared by SFT. I also needed to do a slight small serialization change to ensure that the renderers config would be properly saved. For now this is opt-in and the old template based path is still there.

If thinking gets dropped by the template for any given turn, then the model is not trained on that thinking (there is no conversation expansion). To enable that, the *dataset* needs to include N-turn conversations expanded to multiple data points.

Validation:

- Relevant CPU tests: 183 passed, 2 skipped.
- Checked with the real Qwen3 tokenizer including tool calls/results, unicode, newline boundaries / etc. Everything lines up.
- The new multi turn integration test passes through the same fake PG setup used by PR CI. I only ran it on one physical GPU locally.
- Single-turn loss and `grad_norm` match main over 10 steps.


haydn-jones:
~Geeze, ok there's actually an issue with how I'm detecting conversation boundaries. Because BPE is not compositional I tried to get around this by progressively tokenizing the chat, but with models like Qwen which add and drop thinking tokens based on position in chat my span detection is wrong.~

~I'm not sure exactly what to do about this. I know transformers now uses special jinja directives (`{% generation %}`) to detect the assistant spans, but plenty of models do not use them (like GPT-OSS :roll_eyes:). Marking as WIP until I figure this out.~

joecummings:
Thanks for the thorough work here @haydn-jones 

Couple things that give me pause:

1. Per-model maintenance: Every new model needs a hand-written regex, and the GPT-OSS pattern is already complex enough to be hard to audit. I'd rather have something that works with *at least most* templates automatically.
2. Readability: The char_spans_to_token_spans offset-walking logic is difficult to follow. I had to trace through it multiple times to convince myself it was correct. 

I've been looking at how https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/prompt_strategies/chat_template.py does it where they find where assistant turn i lives in the token stream, render the conversation with that turn's content replaced by an empty string, tokenize both, and diff.

 Some claude pseudocode:

```python
  def _get_assistant_spans(self, messages, full_tokens):
      spans = []
      for i, msg in enumerate(messages):
          if msg["role"] != "assistant":
              continue

          # Same conversation structure, just empty content for this turn.
          # Strips extra fields (e.g. reasoning_content) so template
          # conditionals like Qwen3's <think> insertion also differ.
          dummy_messages = [
              {"role": m["role"], "content": ""} if j == i else m
              for j, m in enumerate(messages)
          ]
          dummy_tokens = tokenize(apply_chat_template(dummy_messages))

          # Scan from front to find where they diverge
          start = first index where full_tokens[start] != dummy_tokens[start]

          # Scan from back to find where they reconverge
          end = last index (in full_tokens) where they still differ

          spans.append((start, end + 1))
      return spans
```

I think this handles the Qwen3 <think> problem because both renderings have the same conversation structure - the template's conditional logic fires identically. The only difference is the content of one turn. It should also handle any future template quirks without needing new regexes.

 The tradeoff is 2N extra tokenization calls per sample (N = assistant turns). Tokenization is fast relative to a training step, but I acknowledge it's not free. For train_on="last_assistant" we can reduce this to a single extra call.
 
 Thoughts? Something I might be missing here?

haydn-jones:
I think your suggestion is the better approach, and I’ve implemented it. I've pushed so you can look at it.

With respect to the token diffing method: it correctly finds the assistant content span, but I don’t think the diff alone can reliably recover the end of turn token.

Concrete example:
```
full:   ...<|im_start|>assistant\nHello there!<|im_end|>\n<|im_start|>user...
blank:  ...<|im_start|>assistant\n            <|im_end|>\n<|im_start|>user...
```

When scanning from the back, the two sequences reconverge at `<|im_end|>\n<|im_start|>user...`, because that suffix is identical in both renderings. So the diff isolates `Hello there!`, but not `Hello there!<|im_end|>`.

As far as I can tell, Axolotl also needs more than just the diff for this. Their `find_turn(...)` path renders twice per turn (`turns_with_empty` and `turns_with_content`) to find the content boundary, and then they have separate logic to locate `EOT` / `EOS` near that boundary.

How do you feel about this? If we want to also supervise the end of turn token, I think that requires additional machinery beyond the diff itself, and I’d be happy to add that once we agree on the right approach.

Edit: Not sure how commit hygiene is enforced here, if we settle on an approach I'd be happy to open a separate PR with cleaner diffs.

joecummings:
Thanks for implementing - looks straightforward. Yeah we have to supervise eot. Could it be as simple as a post-hoc extension that checks whether full_tokens[end] is a special token and include it?

Don't worry about commit hygiene, we'll squash on merge and unless you have 50 reverting commits, it should be fine

haydn-jones:
I checked gpt-oss, llama3, llama4, qwen3, and deepseek, and yes the heuristic is correct for all of them. The last things on my mind are:

1. Push the EOT heuristic
2. Probably should implement `train_on=last_assistant` (so we have `all`, `assistant`, `last_assistant`). Or whatever names we want to use.
3. Reasoning field. `gpt-oss` expects the reasoning content in a `thinking` field, qwen3 expects it in a `reasoning_content` field. Not sure if it should be the responsibility of the dataset to try and handle that mapping at all (e.g., expose  `reasoning_field_name` for the field in the dataset and `reasoning_template_name` for the chat template expected name). Would allow you to use the same dataset on Qwen / gpt-oss without having to like re-upload to HF. I don't mind either way.

If you have any comments on the above or anything else, let me know and I'll take care of them so we can hopefully close this out.

joecummings:
1 and 2 are fine - maybe use an Enum for 2. 

WRT 3, my current thinking is to make this the responsibility of the user constructing the dataset (see how I set up the Qwen3 experiment config).

haydn-jones:
Looks like I've done everything I intended to. Merged main in and hit some issues in the validation pipeline. I'm not certain this is correct for context parallel, so this might need some scrutiny.

https://github.com/pytorch/torchtitan/pull/2769/changes/0582ecc8aa18f3219e58fd04b9044c70932ca354#diff-a05647646c2ceafcb0c9312bebbefe6a7952fbb16e6ffa1c0bf7367680cba184R201

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -53,6 +54,20 @@ def _process_c4_text(sample: dict[str, Any]) -> str:
...
+    Controls which tokens receive supervised labels during training.
+
+    ALL: Supervise the full rendered conversation.
+    ASSISTANT: Supervise all assistant turns.
+    LAST_ASSISTANT: Supervise only the final assistant turn.
+    """
+
+    ALL = "all"
+    ASSISTANT = "assistant"
+    LAST_ASSISTANT = "last_assistant"
```
tianyu-l: As someone who doesn't know sft well in general, can I ask when would we want the three modes, respectively? E.g. my naive thinking would be `ASSISTANT` mode should always be used.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -487,6 +609,11 @@ class Config(ParallelAwareDataloader.Config):
         sample_processor: Annotated[Callable, tyro.conf.Suppress]
         """Callable(sample_dict) -> list[message_dict]. Set in config functions."""
 
+        train_on: Annotated[SupervisionMode, tyro.conf.EnumChoicesFromValues] = (
+            SupervisionMode.ALL
+        )
+        """Which tokens to supervise: 'all', 'assistant', or 'last_assistant'."""
```
tianyu-l: In PR summary, you said "To do assistant-only loss (which is the default)", but here the default seems `ALL`.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -53,6 +54,20 @@ def _process_c4_text(sample: dict[str, Any]) -> str:
 }
 
 
+class SupervisionMode(str, Enum):
```
tianyu-l: torchtitan convention is to use `Literal` instead of `Enum`, if that works for you. E.g. activation checkpointing mode https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L283. Not trying to argue which is better, but just to keep consistency throughout the repo.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -487,6 +609,11 @@ class Config(ParallelAwareDataloader.Config):
         sample_processor: Annotated[Callable, tyro.conf.Suppress]
         """Callable(sample_dict) -> list[message_dict]. Set in config functions."""
 
+        train_on: Annotated[SupervisionMode, tyro.conf.EnumChoicesFromValues] = (
```
tianyu-l: In terms of naming, can we do
`supervision_mode: Literal["assistant", ...]`

torchtitan/hf_datasets/text_datasets.py
haydn-jones: Generally, you do only want to train on assistant completions only as far as I know, though I don't have any insight on real CPT after standard pretraining. I definitely always default to assistant only loss.

I have the other two to match TRL from huggingface and Axolotl. TRL defaults to `ALL` and axolotl technically supports `LAST_ASSISTANT` but it's not really advertised (you have to set `roles_to_train: []`).

torchtitan/hf_datasets/text_datasets.py
haydn-jones: Ok, sounds good.

torchtitan/hf_datasets/text_datasets.py
haydn-jones: Good catch, thanks.

haydn-jones:
Need to do a training run on my silly dataset to make sure everything is looking good.

torchtitan/hf_datasets/text_datasets.py
tianyu-l: Thanks. I would lean on the side of "keeping it simple unless we know it's going to be useful". But maybe @joecummings knows more.

torchtitan/hf_datasets/text_datasets.py
tianyu-l: oops, name seems not addressed -- but we can wait until the decision is made on whether or not to have this field.

haydn-jones:
A couple of things:

1. The VLM CI is broken because the VLM dataset doesn't emit positions, and because of some changes that happened while I was working on this my fix no longer works around that. The correct fix would be to add position info in to the VLM dataset, and maybe a simple fast fix is emit dummy position information from the VLM collator.
2. My E2E experiment isn't working properly with the token-span diffing method of finding assistant spans. For Qwen3, we cant supervise the opening `<think>` token in the assistant message because it shows up in both the blanked render and the real render. When serving, vLLM will use something like `add_generation_prompt=True` to start off the assistant message, but this does *not* include a `<think>` token, so my trained model seems to be getting confused and not thinking / responding as it should. I can get it to do what it should by prefilling the assistant response with a thinking token but not great. Thoughts @joecummings?
3. ~~Will fix the `SupervisionMode` thing once we decide if we want to keep configurable supervision modes.~~

torchtitan/hf_datasets/text_datasets.py
joecummings: It's not uncommon to be able to specify weighting per-turn (OpenAI also allows this). The best reasoning would be some datasets wherein there's a lot of "setup" and we're really only interested in having the model predict the last turn. 

It is definitely not as common as just training on all assistant turns. 

In the spirit of "as minimal as possible", we can omit from this PR but please leave a TODO to revisit at some point. Sorry, I got overeager here :)

haydn-jones:
> A couple of things:
> 
> 1. The VLM CI is broken because the VLM dataset doesn't emit positions, and because of some changes that happened while I was working on this my fix no longer works around that. The correct fix would be to add position info in to the VLM dataset, and maybe a simple fast fix is emit dummy position information from the VLM collator.
> 2. My E2E experiment isn't working properly with the token-span diffing method of finding assistant spans. For Qwen3, we cant supervise the opening `<think>` token in the assistant message because it shows up in both the blanked render and the real render. When serving, vLLM will use something like `add_generation_prompt=True` to start off the assistant message, but this does _not_ include a `<think>` token, so my trained model seems to be getting confused and not thinking / responding as it should. I can get it to do what it should by prefilling the assistant response with a thinking token but not great. Thoughts @joecummings?
> 3. ~Will fix the `SupervisionMode` thing once we decide if we want to keep configurable supervision modes.~

Marked this as draft so CI won't run. Points (1) and (2) still need to be addressed. Will wait a few days for comments before checking back in.

haydn-jones:
@joecummings I'm a bit swamped with NeurIPS deadline stuff, but I do think I have a solution to the small tokenization issue here. Will implement when I have time. Maybe this weekend?

haydn-jones:
@joecummings Long story short the issue I saw with my toy dataset and  `Qwen3-0.6B` was more an issue with `Qwen3-0.6B`'s specific chat template (newer Qwen templates inject an opening `<think>` token when setting `add_generation_prompt=True` which solves it). I fixed the VLM experiment by adding positions into the dataloader (I don't think there is anything I need to do for the VLM in the main repo).

I accidentally ran ufmt on too many files and touched the forge experiment :( can revert.

The last question I have before I feel like I'm done is: Should chat template arguments (like `enable_thinking`) be exposed somehow to the user through something like `process_sample`?

joecummings:
> The last question I have before I feel like I'm done is: Should chat template arguments (like `enable_thinking`) be exposed somehow to the user through something like `process_sample`?

`process_sample` should largely be user defined, which means if they are loading in a chat dataset that utilizes reasoning traces, they should insert reasoning tags [like so](https://github.com/pytorch/torchtitan/blob/786e26f8ee47ffecb523a661535e71031583ff60/torchtitan/models/qwen3/config_registry.py#L328). For our current setup, maybe we need to pass along `chat_template_kwargs` from the dataloader. Eventually we should really have better staging between: 1) getting data into a loose "message" structure and 2) applying model transform (tokenization) for training.

haydn-jones:
@joecummings Ok yeah I'm with you.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -343,43 +351,121 @@ def _get_data_iter(self):
...
-            raise ValueError(
-                f"Expected single-turn [user, assistant], got {len(messages)} messages"
-            )
-        if messages[0]["role"] != "user":
+        """Validate conversation structure.
+
+        Allows an optional leading 'system' message, then requires
+        alternating user/assistant turns ending with 'assistant'.
+        """
+        if len(messages) < 2:
```
tianyu-l: this exception is covered by the one below, can we just keep that one?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -319,6 +320,13 @@ def __init__(
         self.seq_len = seq_len
         self.infinite = infinite
         self._sample_processor = sample_processor
+        self._assistant_end_token_ids = {self._eos_id}
+        if isinstance(tokenizer, HuggingFaceTokenizer):
+            self._assistant_end_token_ids.update(
+                token_id
+                for token_id, token in tokenizer.tokenizer.get_added_tokens_decoder().items()
+                if token.special
```
tianyu-l, line 333: how robust is this -- does a special token always belong to "end tokens"?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -343,43 +351,121 @@ def _get_data_iter(self):
...
+            # Strips extra fields (e.g. reasoning_content) so template
+            # conditionals like Qwen3's <think> insertion also differ.
+            blanked_messages = [
+                {"role": msg["role"], "content": ""} if idx == assistant_idx else msg
+                for idx, msg in enumerate(messages)
+            ]
+
+            _, dummy_tokens = self._render_conversation(blanked_messages)
+
+            # Scan from front to find where they diverge.
```
tianyu-l: hmm this algorithm sounds correct, but
- it's hard to read
- the complexity seems Omega(num assistant * length of full tokens)

Do you think we can extend the previous algorithm
- The messages is alternating between users and assistants
- To figure out the start of assistant `i` (with in a user-assistant pair), we `apply_chat_template(add_generation_prompt=True)` to user `i`
- To figure out the length of assistant `i`, we `apply_chat_template` to user i + assistant i
- and we can iterate over all pairs sequentially

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -396,29 +482,40 @@ def _tokenize_sample(
...
+        # Start with everything masked
+        masked_labels = [IGNORE_INDEX] * len(label_ids)
+
+        for start, end in spans:
+            # In label space: first supervised position is start - 1
+            # (predicting full_tokens[start] from position start-1),
+            # and last supervised position is end - 2
+            # (predicting full_tokens[end-1] from position end-2).
+            label_start = max(start - 1, 0)
+            label_end = min(end - 1, len(label_ids))
```
tianyu-l: sounds strange to use `len(labels_ids)` in multi-turn -- `end` should suffice

torchtitan/hf_datasets/text_datasets.py
haydn-jones: A token being special does not imply that it is some kind of turn terminator, so we're adding more tokens than necessary to this list. To select only the necessary or "correct" tokens, we would need model-specific information. For example, GPT-OSS uses two different turn terminators (`<|return|>`/`<|end|>`), deepseek uses `<｜end▁of▁sentence｜>`, and so on. I think the goal has been to avoid hard coding model specific information.

For robustness, I think it's fine. I'm having a hard time imagining a scenario where we'd run into an issue because of this with any of the chat templates in the models supported by the code.

torchtitan/hf_datasets/text_datasets.py
haydn-jones: Ok, had to think about this for a while because I did it this way for a specific reason and forgot what it was. 

Qwen3's template renders assistant turns differently depending on positions. Only the turn after the last user message gets `<think>...</think> `and earlier turns have reasoning dropped from history. But `add_generation_prompt=True` always emits `<|im_start|>assistant\n<think>\n`, so for any non-final assistant turn:

```
prompt = '... <|im_start|>assistant\n<think>\n'
full   = '... <|im_start|>assistant\n4<|im_end|>'
```

Prefix check fails, sample dropped. The `messages[:i+1]` end check fails the same way. The blanking approach handles this fine because both renderings have identical structure, so the template's lookahead cancels in the diff.

With respect to readability, tried to make things more clear in this commit: https://github.com/pytorch/torchtitan/pull/2769/commits/05d8a8b5bfd371e0b28695de15c8f34fc5853c51

felipemello1:
hey @haydn-jones , thanks for the PR! Apologies, I haven't read it completely, but could you please summarize the behavior of the dataset when stripping thinking?

E.g.
```
T1:
{user: "how much is 1+1"}
T2:
{assistant: "<think>This is a math problem</think> it's 2"}
T3
{user: "how much is 4+4+"}
T4:
{assistant: "<think>This is a math problem</think> it's 8"}
```

1. We want the model to train on both T2 and T4 

2. If we do `apply_chat_template` on the whole thing, models like Qwen and Deepseek will **strip** the `think` portion from T2. 

This is a problem, because now the model will lose all of the training related to `think` in T2

3. An alternative then is to branch:
```
Sample 1 is `apply_chat_template(all_turns[:T2])
Sample 2 is `apply_chat_template(all_turns[:T4])
```

And to save resources on unnecessary branching, we check if tokens(sample_1) match tokens(sample_2[:T2])). If they do, no need to branch.

Is this what we are doing?

4. Also, i see many replies mentioning qwen, but every model (e.g. gpt oss) has their own caveats. If my understanding above is correct, it seems that we have two options:

a) Branch as described in (3) and do a naive "prefix" check, which would fail very frequently for Qwen, increasing training cost;
b) Have a dedicated renderer per model to avoid unnecessary branching.

tianyu-l:
```T1:
{user: "how much is 1+1"}
T2:
{assistant: "<think>This is a math problem</think> it's 2"}
T3
{user: "how much is 4+4+"}
T4:
{assistant: "<think>This is a math problem</think> it's 8"}
```

Let's put training efficiency aside, what's the idea behavior we should train on? I could imagine it's:
- full thinking everywhere -- during inference, the engine should see the entire history with all thinking, in this case T1 user -> T2 including thinking -> T3 user -> T4 including thinking -> ...

> An alternative then is to branch:

wouldn't this skew the weight of examples? A data point should be trained once (whose weights is roughly the number of tokens). With branching, are you saying creating two individual training examples?

> b) Have a dedicated renderer per model to avoid unnecessary branching.

this sounds more robust design to me

felipemello1:
>full thinking everywhere -- during inference, the engine should see the entire history with all thinking, in this case T1 user -> T2 including thinking -> T3 user -> T4 including thinking -> ...

"during inference, the engine should see the entire history with all thinking," -> This would be wrong. In inference, when generating T4, the model will **not** have access to the thinking tokens of T2. So, if you train the model to generate T4 seeing "T2 including thinking", it will **not** match the inference behavior. 

Now, you may ask: "is that a big problem?" -- idk, maybe it is, maybe it isnt. Other libraries in RL give this option to the users to `never-branch` and keep all tokens as they were on Tn (no stripping). It is an efficiency <-> correctness tradeoff.

> wouldn't this skew the weight of examples? A data point should be trained once (whose weights is roughly the number of tokens). With branching, are you saying creating two individual training examples?

Yes, two different samples. It wouldn't skew from a valid token perspective. You are still training on the same number of **valid** tokens. When you do [t1,t2 without thinking, t3, t4], the training mask is **only** on T4.

haydn-jones:
> hey @haydn-jones , thanks for the PR! Apologies, I haven't read it completely, but could you please summarize the behavior of the dataset when stripping thinking?
> 
> E.g.
> 
> ```
> T1:
> {user: "how much is 1+1"}
> T2:
> {assistant: "<think>This is a math problem</think> it's 2"}
> T3
> {user: "how much is 4+4+"}
> T4:
> {assistant: "<think>This is a math problem</think> it's 8"}
> ```
> 
> 1. We want the model to train on both T2 and T4
> 2. If we do `apply_chat_template` on the whole thing, models like Qwen and Deepseek will **strip** the `think` portion from T2.
> 
> This is a problem, because now the model will lose all of the training related to `think` in T2
> 
> 3. An alternative then is to branch:
> 
> ```
> Sample 1 is `apply_chat_template(all_turns[:T2])
> Sample 2 is `apply_chat_template(all_turns[:T4])
> ```
> 
> And to save resources on unnecessary branching, we check if tokens(sample_1) match tokens(sample_2[:T2])). If they do, no need to branch.
> 
> Is this what we are doing?
> 
> 4. Also, i see many replies mentioning qwen, but every model (e.g. gpt oss) has their own caveats. If my understanding above is correct, it seems that we have two options:
> 
> a) Branch as described in (3) and do a naive "prefix" check, which would fail very frequently for Qwen, increasing training cost; b) Have a dedicated renderer per model to avoid unnecessary branching.

Ok, I think I understand what you're saying: take a `K` assistant-turn conversation with thinking and turn it into `K` training data points, each ending at `assistant[i]` so we don't waste the intermediate thinking tokens. IMO this should be the responsibility of user constructing the dataset by inserting those extra examples in (though that comes at the cost of making storage `>K` times larger). AFAIK most SFT training frameworks (TRL, Axolotl, etc) will not do this for you. That's not to say I don't think we should add support for it, I'll happily write whatever you guys want me to.

For dedicated rendering, at one point I had something like that to detect assistant spans, but it does increase the burden of per-model maintenance https://github.com/pytorch/torchtitan/pull/2769#issuecomment-4171072807.

felipemello1:
> IMO this should be the responsibility of user constructing the dataset by inserting those extra examples in

It is a foot gun. Say that there are 30 rounds of assistant/user discussion and it ends with "thank you" / "no problem.". All of the intermediate thinking would be wasted haha.

> AFAIK most SFT training frameworks (TRL, Axolotl, etc) will not do this for you.

Hmm, i would need to take a better look. I haven't checked SFT in a while, but i imagine that the points i made hold.

> That's not to say I don't think we should add support for it, I'll happily write whatever you guys want me to.

Thank you! So, what is your current algorithm proposal? is there branching or do we tokenize all messages at once? I saw some words like prefix and a for loop on all assistant messages, so i thought that you were doing branching.

perhaps we can align on the algorithm before we align on the code

> For dedicated rendering, at one point I had something like that to detect assistant spans, but it does increase the burden of per-model maintenance

It does, unfortunately. Luckily we don't have many models to maintain, so maybe it is feasible.

haydn-jones:
> > IMO this should be the responsibility of user constructing the dataset by inserting those extra examples in
> 
> It is a foot gun. Say that there are 30 rounds of assistant/user discussion and it ends with "thank you" / "no problem.". All of the intermediate thinking would be wasted haha.
> 
> > AFAIK most SFT training frameworks (TRL, Axolotl, etc) will not do this for you.
> > Hmm, i would need to take a better look. I haven't checked SFT in a while, but i imagine that the points i made hold.

Sorry yeah, I think my last reply was confusing and missed one of your questions. The current code doesn't branch anywhere. What it does:

1. Render the full conversation once via `apply_chat_template(messages)`. For `Qwen3`/`DeepSeek`, this template strips intermediate thinking and we inherit that.
2. For each assistant turn `i`, find its content span in that single token stream by re-rendering with turn `i`'s content blanked and diffing. The for-loop is purely for assistant span detection.
3. Mask everything outside those spans with `IGNORE_INDEX`.

Result: one sample per conversation, supervised on whatever the template kept. So yes, your foot gun is real. For `Qwen3`, only the last assistant turn's thinking gets gradient, though the visible (non-thinking) assistant content in earlier turns is still supervised.

For the `K`-sample expansion, I'm happy to implement it if that's what you guys want. On that, I have at least two questions:

1. Default on or off? Default-off keeps the foot gun for anyone who doesn't read docs but default-on silently multiplies dataset size by avg #assistant-turns.
2. Would the preferred implementation be template agnostic or with custom renderers per-model?

tianyu-l:
> Yes, two different samples. It wouldn't skew from a valid token perspective. You are still training on the same number of valid tokens. When you do [t1,t2 without thinking, t3, t4], the training mask is only on T4.

@felipemello1 
In terms of what tokens to be trained on, it sounds correct. However, I can imagine it creates more complexity for multi-source datasets https://github.com/pytorch/torchtitan/pull/3063. The ratio probably should depend on the valid tokens being trained in each source, not packed data lengths. cc @slimfrkha

felipemello1:
thanks for being flexible @haydn-jones !

> For the K-sample expansion, I'm happy to implement it if that's what you guys want. On that, I have at least two questions:
1 Default on or off? Default-off keeps the foot gun for anyone who doesn't read docs but default-on silently multiplies dataset size by avg #assistant-turns.
2 Would the preferred implementation be template agnostic or with custom renderers per-model?

I am going to work on renderers for our RL experiments and will have a PR in probably 1-2 weeks. The steps i would take are the following:

a. Confirm how other libraries handle the SFT for chatdataset -- I can launch a research with claude since we have the tokens at meta;

b. Then, we have two options:
i) try to land this multi-turn chat dataset with a flag: "branch_on_assistant_turn". If true, we have K samples, if false, we do a single encode pass, like you are already doing it.
ii) hold off, wait for the renderer work, see how it fits here

I think that (ii) minimizes work, but i don't want to lose your momentum on this PR. @tianyu-l, you have iterated on this PR a few times. Do you think that (i) is doable/close to land, and we take our time with (ii) later?

tianyu-l:
> Do you think that (i) is doable/close to land, and we take our time with (ii) later?

I personally think we should hold for (ii)

felipemello1:
@haydn-jones , can i ping you when i have an RFC/PR for renderers?

rakkit:
TBH i am not convinced we should truncted thinking trajectory from past conversation.

>  Confirm how other libraries handle the SFT for chatdataset 

Nemotron 3 Nano (super) introduced this as well, maybe we could check Nvidia's implement

haydn-jones:
> @haydn-jones , can i ping you when i have an RFC/PR for renderers?

@felipemello1 Yes of course, happy to look at anything when it's ready.

slimfrkha:
Hey folks,

My 2 cents: I would keep the design simple and push data management responsibility to the user via the jinja template.

1. **Loss mask:** Use `return_assistant_tokens_mask=True` in `apply_chat_template`. It keys off `{% generation %}` in the template to identify assistant turns, and most standard models already support it: [https://github.com/huggingface/trl/issues/5471](https://github.com/huggingface/trl/issues/5471). Users verify and tweak the template if needed. Going from `assistant` to `last_assistant` is trivial from there — find the last assistant turn (backward-scan for the first 1→0 transition) and mask everything before it.

2. **Thinking traces:** Train on all intermediate traces. Dropping them is an inference optimization with negligible impact on model performance, and keeping them helps sample efficiency and token distribution. Users ensure the template doesn't strip prior traces. If someone wants to train on the last thinking trace only, they can tweak the jinja template and pick the right strategy (`assistant` / `last_assistant`). Template tweaks + `train_on` values give users full flexibility.

TL;DR:
- `return_assistant_tokens_mask` for the loss mask
- Train on all thinking traces in a sample is up to the user
- Users own the jinja template

Torchtitan stays clean — no per-model special-casing, not now not for a future model.

haydn-jones:
> Hey folks,
> 
> 
> 
> My 2 cents: I would keep the design simple and push data management responsibility to the user via the jinja template.
> 
> 
> 
> 1. **Loss mask:** Use `return_assistant_tokens_mask=True` in `apply_chat_template`. It keys off `{% generation %}` in the template to identify assistant turns, and most standard models already support it: [https://github.com/huggingface/trl/issues/5471](https://github.com/huggingface/trl/issues/5471). Users verify and tweak the template if needed. Going from `assistant` to `last_assistant` is trivial from there — find the last assistant turn (backward-scan for the first 1→0 transition) and mask everything before it.
> 
> 
> 
> 2. **Thinking traces:** Train on all intermediate traces. Dropping them is an inference optimization with negligible impact on model performance, and keeping them helps sample efficiency and token distribution. Users ensure the template doesn't strip prior traces. If someone wants to train on the last thinking trace only, they can tweak the jinja template and pick the right strategy (`assistant` / `last_assistant`). Template tweaks + `train_on` values give users full flexibility.
> 
> 
> 
> TL;DR:
> 
> - `return_assistant_tokens_mask` for the loss mask
> 
> - Train on all thinking traces in a sample is up to the user
> 
> - Users own the jinja template
> 
> 
> 
> Torchtitan stays clean — no per-model special-casing, not now not for a future model.

Personally I 100% agree with you on both points, and in particular the jinja template. Assistant span detection is hard, jinja directives make it easy (at least for torchtitan) and flexible. It does make it harder for the user, but I personally would eat that cost, and in fact do with my internal chat dataset which uses the generation directives.

felipemello1:
thanks @slimfrkha !

I leveraged claude/codex to write an RFC here: https://github.com/pytorch/torchtitan/pull/3304

There are two families of libraries:
a) Uses a renderer for SFT for more precise control;
Examples: torchtune, Tinker, LLaMA-Factory, internal reference, verifiers (renderer package), slime.

b) Relies on jinja --> possible footguns and more burden on users;
Examples: NeMo-RL, prime-rl, Verl, Open-Instruct, OpenRLHF, Axolotl, Megatron-LM,TRL

TLDR: I think we should enable it with jinja+footguns, but we should follow up with renderers based on the RL work.

-------

> 1) Loss mask: Use return_assistant_tokens_mask=True in apply_chat_template. It keys off {% generation %}

This seems pretty straight forward indeed, but this is in TRL only. We would need to bring over their per-model jinja to titan. Is that true?

> 2) Thinking traces: Train on all intermediate traces. [...] Users ensure the template doesn't strip prior traces.

I agree on training on intermediate traces, but I find it burdensome to put this responsibility on users. If this is the standard in training, i think we should fix it.

cc: @tianyu-l @haydn-jones

felipemello1:
@haydn-jones , lets go ahead with leveraging HF jinja as is and let the user edit it if they want to. We should raise a warning on prefix mismatch though, so the user is at least aware of the footgun. When you have the time, can you please let us know how far this PR is from the A1/B1 in the RFC (you can go straight to the section and read A1 and B1 if you are short in time)? We can proceed with that design and later refine for a more per-model renderer solution. Does that work for you?

haydn-jones:
@felipemello1 This diverges a bit from A1 (it aligns with B1 just fine) in the sense that my current span recovery skips the incremental-prefix fast path and goes straight to the message-payload-diff fallback for every assistant turn. I can align it to A1 by checking for prefix mismatch, emitting a warning, then falling back to what I have now if you'd like.

felipemello1:
> This diverges a bit from A1 (it aligns with B1 just fine) in the sense that my current span recovery skips the incremental-prefix fast path and goes straight to the message-payload-diff fallback for every assistant turn.

awesome! I would need to think a bit about what you said. Do you have an opinion on `incremental-prefix fast path` vs ` straight to the message-payload-diff fallback`. Do you see any cons in doing the incremental prefix?

haydn-jones:
> > This diverges a bit from A1 (it aligns with B1 just fine) in the sense that my current span recovery skips the incremental-prefix fast path and goes straight to the message-payload-diff fallback for every assistant turn.
> 
> awesome! I would need to think a bit about what you said. Do you have an opinion on `incremental-prefix fast path` vs ` straight to the message-payload-diff fallback`. Do you see any cons in doing the incremental prefix?

No downside other than having two paths, and as far as I can tell the fast path will fail on any reasoning model.

As an aside, I've benchmarked my code and it can actually get pretty slow when there are a large number of turns:

| K (assistant turns) | tok/conv | conv/s (current) | conv/s (render-only ceiling) |
|---:|---:|---:|---:|
| 1 | 454 | 3029 | 4656 |
| 2 | 908 | 1040 | 2725 |
| 4 | 1810 | 291 | 1333 |
| 8 | 3623 | 74 | 643 |
| 16 | 7233 | 19 | 322 |
| 32 | 14476 | 4.6 | 159 |

`render-only ceiling` would be something like using `{% generation %}` JINJA directives.

felipemello1:
> > > This diverges a bit from A1 (it aligns with B1 just fine) in the sense that my current span recovery skips the incremental-prefix fast path and goes straight to the message-payload-diff fallback for every assistant turn.
> > 
> > 
> > awesome! I would need to think a bit about what you said. Do you have an opinion on `incremental-prefix fast path` vs ` straight to the message-payload-diff fallback`. Do you see any cons in doing the incremental prefix?
> 
> No downside other than having two paths, and as far as I can tell the fast path will fail on any reasoning model.
> 
> As an aside, I've benchmarked my code and it can actually get pretty slow when there are a large number of turns:
> 
> K (assistant turns)	tok/conv	conv/s (current)	conv/s (render-only ceiling)
> 1	454	3029	4656
> 2	908	1040	2725
> 4	1810	291	1333
> 8	3623	74	643
> 16	7233	19	322
> 32	14476	4.6	159
> `render-only ceiling` would be something like using `{% generation %}` JINJA directives.

maybe `renderers` solves this for us and we can skip the short term and jump straight into the final solution? https://github.com/PrimeIntellect-ai/renderers/tree/main

The only issue is the dependency. I think that they require transformers while we only need tokenizers. Transformers is a heavier dependency that we would probably want to avoid. I opened an issue with them asking about it.

haydn-jones:
@felipemello1 I think thats reasonable, especially given all of the issues they pointed out in their blog post (https://www.primeintellect.ai/blog/renderers) and subsequently fixed.

slimfrkha:
I would push also for renderers (even though it may be an overkill for SFT?). My only concern is limited support for other models. Renderers support a handful of models that support agentic behaviour (though it may grow for sure in the future). SFTing is not only for agentic models.
Maybe have the below 2 options:
1. default option: renderers
2. fall back for complete flexibility (destined for power users): `return_assistant_tokens_mask` + `{% generation %}`. use case for non supported models / funky loss mask use cases.

Also would be interesting to profile renderers against the same profiling table above.

felipemello1:
@haydn-jones , would you like to try writing some small code with renderers and see how it looks like? reference from prime-rl for sft: https://github.com/PrimeIntellect-ai/renderers/issues/36#issuecomment-4455150938

It seems that they might remove the hard requirement of `transformers` for text-only models, so that shouldn't be a blocker for us.

> Also would be interesting to profile renderers against the same profiling table above.

@slimfrkha  , yeah, me too. Hopefully its not painfully slow for SFT needs

haydn-jones:
> @haydn-jones , would you like to try writing some small code with renderers and see how it looks like? reference from prime-rl for sft: https://github.com/PrimeIntellect-ai/renderers/issues/36#issuecomment-4455150938
> 
> 
> 
> It seems that they might remove the hard requirement of `transformers` for text-only models, so that shouldn't be a blocker for us.
> 
> 
> 
> > Also would be interesting to profile renderers against the same profiling table above.
> 
> 
> 
> @slimfrkha  , yeah, me too. Hopefully its not painfully slow for SFT needs

Sure I can give it a shot, and profile as well.

haydn-jones:
| K (assistant turns) | tok/conv | conv/s (current) | conv/s (render-only ceiling) | `renderers` library |
|--------------------:|---------:|-----------------:|-----------------------------:|------------------:|
| 1                   | 454      | 3029             | 4656                         | 4362              |
| 2                   | 908      | 1040             | 2725                         | 2465              |
| 4                   | 1810     | 291              | 1333                         | 1184              |
| 8                   | 3623     | 74               | 643                          | 571               |
| 16                  | 7233     | 19               | 322                          | 281               |
| 32                  | 14476    | 4.6              | 159                          | 148               |

Renderers is within ~10% of the current ceiling. 

Chat dataset with renderers:
```python
from renderers import create_renderer
from transformers import AutoTokenizer

class ChatDataset(IterableDataset, Stateful):
    def __init__(self, dataset, tokenizer_path, *, renderer_name="auto", ...):
        hf_tokenizer = AutoTokenizer.from_pretrained(tokenizer_path)
        self._renderer = create_renderer(hf_tokenizer, renderer=renderer_name)
        # ... rest of init: dataset, seq_len, buffers, etc.

    ...

    def _tokenize_sample(self, sample):
        messages = self._sample_processor(sample)
        self._validate_messages(messages)
      
        out = self._renderer.render(messages) # one render, no diffing
        full_tokens = out.token_ids
      
        if len(full_tokens) - 1 > self.seq_len:
            return None
      
        input_ids = full_tokens[:-1]
        label_ids = full_tokens[1:]
      
        # Per-token: True if this token's source message is an assistant turn.
        # message_indices[i] == -1 marks structural scaffolding (never supervised).
        supervise = [
            idx >= 0 and messages[idx]["role"] == "assistant"
            for idx in out.message_indices
        ]
  
        # Labels are shifted by one: label_ids[j] predicts full_tokens[j+1],
        # so it's supervised iff position j+1 in the full stream is supervised.
        masked_labels = [
            tok if supervise[j + 1] else IGNORE_INDEX
            for j, tok in enumerate(label_ids)
        ]
        return input_ids, masked_labels
```

We delete:
  - _render_conversation (HF + BOS + EOS bookkeeping). The renderer handles it per-model.
  - _get_assistant_spans (K re-renders not needed)
  - _content_diff (longest-common-prefix/suffix not needed)
  - _assistant_end_token_ids and the EOT-extension logic
  
If we want to emit `k` turns for a `k`-turn chat sample, we could just loop over `_tokenize_sample`.

@felipemello1 @slimfrkha

felipemello1:
thanks @haydn-jones! that looks pretty clean and performant to me. Assuming others agree, then I think that the only blockers would be:
a. some sanity checks on some dummy samples (print the tokens, masks, encode->decode)
b. renderers removing the pre-requisit of transformers for text renderes. Perhaps we could open propose an RFC or open a PR there https://github.com/PrimeIntellect-ai/renderers/issues/31

wdyt? do you like this approach?

haydn-jones:
@felipemello1 I think its a nice idea, seems like a convenient way to solve a lot of problems at once.

Aside: this PR (https://github.com/PrimeIntellect-ai/renderers/pull/53) probably would have changed the way I wrote the example code above. Notably the `is_content` field.

HosseinKaviani-H:
@haydn-jones curious on the state of this PR? Any current blockers?

haydn-jones:
> @haydn-jones curious on the state of this PR? Any current blockers?

To be honest I entirely forgot where we ended up with this. I know @felipemello1 made a different RFC (https://github.com/pytorch/torchtitan/pull/3304).

If we could just use renderers from prime-intellect this would be a very straightforward impl but it would pull in transformers as a dep.

felipemello1:
> > @haydn-jones curious on the state of this PR? Any current blockers?
> 
> To be honest I entirely forgot where we ended up with this. I know @felipemello1 made a different RFC (#3304).
> 
> If we could just use renderers from prime-intellect this would be a very straightforward impl but it would pull in transformers as a dep.

they have a PR to remove the dependency, but never landed :/ . I think we should wait a bit longer, and i can ping them again

HosseinKaviani-H:
> > > @haydn-jones curious on the state of this PR? Any current blockers?
> > 
> > 
> > To be honest I entirely forgot where we ended up with this. I know @felipemello1 made a different RFC (#3304).
> > If we could just use renderers from prime-intellect this would be a very straightforward impl but it would pull in transformers as a dep.
> 
> they have a PR to remove the dependency, but never landed :/ . I think we should wait a bit longer, and i can ping them again

Which PR is this? @felipemello1

felipemello1:
> thanks @haydn-jones! that looks pretty clean and performant to me. Assuming others agree, then I think that the only blockers would be: a. some sanity checks on some dummy samples (print the tokens, masks, encode->decode) b. renderers removing the pre-requisit of transformers for text renderes. Perhaps we could open propose an RFC or open a PR there [PrimeIntellect-ai/renderers#31](https://github.com/PrimeIntellect-ai/renderers/issues/31)
> 
> wdyt? do you like this approach?

check the PRs tagged on this issue, please @HosseinKaviani-H

haydn-jones:
@felipemello1 https://github.com/PrimeIntellect-ai/renderers/pull/70

felipemello1:
@haydn-jones , do you want to double check if APIs are still up to date? renderers changed some things. Let me know if you want to make the changes and when the PR is ready. Lets land it :)

tianyu-l:
@haydn-jones are you still interested in working on this? please let us know, thanks!

haydn-jones:
@tianyu-l Right, sorry. I'll work through it this weekend!

haydn-jones:
@felipemello1 @tianyu-l

Ok, I moved the code onto renderers and updated the branch. I kinda started over and pushed with --force-with-lease because of how far behind that branch got and all of the changes I didn't want to navigate.

Right now this is the most simple and literal port onto renderers I could do (see the `Aside` at the end for this).

I did a number of manual checks against Qwen3s official tokenizer (BOS / EOS, thinking tokens in multi turn, tool calling, etc.) and everything looks right. There was previously discussion about how to handle multi-turn data in SFT (e.g., do we automatically expand an N (user, assistant) turn conversation into N data points to train on thinking at every step), for now this just renders the data as it appears in the dataset. No conversation expansion.

Aside: I think leaning on renderers fully would make things nicer and cleaner. If we did so we could remove

- The branches for template-based vs. renderer-based tokenization and masking.
- Manual BOS/EOS handling and newline stripping.
- Rendering / tokenizing the prompt separately to find the assistant boundary, along with `_require_token_prefix`.
- The single-turn `(user, assistant)` restriction.

Tool calls / responses already work with a supporting renderer. We'd also need to pass the tool definitions through `build_training_sample(..., tools=...)` for it to inject the schemas into the prompt. Doing all of this should be a relatively straightforward thing, I just didn't because I wasn't sure exactly how deeply you wanted this integrated.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -91,13 +94,15 @@ def _require_token_prefix(full_tokens: list[int], prompt_tokens: list[int]) -> N
...
     class Config(SampleProcessor.Config):
         messages_fn: Annotated[
-            Callable[[dict[str, Any]], list[dict[str, str]]], tyro.conf.Suppress
+            Callable[[dict[str, Any]], list[Message]], tyro.conf.Suppress
         ]
+        renderer: RendererConfig | None = None
+        """Model renderer; None uses the tokenizer's single-turn chat template."""
 
     def __init__(self, config: Config, *, context: DatasetBuildContext) -> None:
         if context.tokenizer.eos_id is None:
```
jinsooihm: `context.tokenizer.eos_id` assigned to `self._eos_id` is only used when `renderer` is not present, so maybe this check should only be done when `renderer` is none?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -163,19 +188,24 @@ def _tokenize_sample(self, sample: dict[str, Any]) -> TextSequence | None:
...
-        )
-        prompt_tokens = self._tokenizer.encode(prompt_text, add_bos=True, add_eos=False)
-        _require_token_prefix(full_tokens, prompt_tokens)
-        prompt_len = len(prompt_tokens)
-
         tokens = np.asarray(full_tokens, dtype=np.int64)
         input_ids = tokens[:-1]
         labels = tokens[1:].copy()
-        labels[: max(prompt_len - 1, 0)] = IGNORE_INDEX
+        if loss_mask is None:
```
jinsooihm: It seems that if `renderer` was used, then `loss_mask` will always be not `None`. Would it be better to separate as two functions, one solely `renderer` path and the other solely fallback path to make the logic cleaner?

jinsooihm (commented):
Thank you. I think it looks good.

Codex has one concrete complaint
```
TorchTitan:          Python >=3.11
renderers 0.1.11:   Python >=3.10,<3.14
Effective range:     Python >=3.11,<3.14

On Python 3.14, pip install torchtitan sees that the exact required renderer version does not support the interpreter and fails during dependency resolution:

ERROR: Ignored renderers 0.1.11 because it requires Python <3.14
ERROR: No matching distribution found for renderers==0.1.11
```

torchtitan/hf_datasets/text_datasets.py
haydn-jones: Yes, good catch.

torchtitan/hf_datasets/text_datasets.py
haydn-jones: Good catch. I've separated the two.

haydn-jones:
> Thank you. I think it looks good.
> 
> Codex has one concrete complaint
> 
> ```
> TorchTitan:          Python >=3.11
> renderers 0.1.11:   Python >=3.10,<3.14
> Effective range:     Python >=3.11,<3.14
> 
> On Python 3.14, pip install torchtitan sees that the exact required renderer version does not support the interpreter and fails during dependency resolution:
> 
> ERROR: Ignored renderers 0.1.11 because it requires Python <3.14
> ERROR: No matching distribution found for renderers==0.1.11
> ```

Addressed your two comments. For the 3.14 thing, I checked with renderers and 3.14 works fine, I've opened a PR there (https://github.com/PrimeIntellect-ai/renderers/pull/155).

jinsooihm (approved):

.ci/docker/requirements.txt
```diff
@@ -4,6 +4,7 @@ tensorboard
 wandb
 tyro >= 1.0.5
 tokenizers >= 0.15.0
+renderers == 0.1.11
```
tianyu-l, line 7: let's remove https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/requirements.txt#L6

torchtitan/components/renderer.py
```diff
@@ -11,7 +11,7 @@
 
 import tyro
 from renderers import create_renderer, Renderer
-from renderers.configs import BaseRendererConfig
+from renderers.configs import RendererConfig as LibraryRendererConfig
```
tianyu-l: curious why import changes from Base to non-Base?

also the naming more-or-less collides with https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/renderer.py#L33

torchtitan/config/configurable.py
```diff
@@ -50,7 +50,7 @@ def _convert(val):
                 if hasattr(val, "to_dict"):
                     return val.to_dict()
                 elif dataclasses.is_dataclass(val):
-                    return _convert(dataclasses.asdict(val))
+                    return {f.name: _convert(getattr(val, f.name)) for f in fields(val)}
```
tianyu-l, line 53: does this solve it https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/renderer.py#L57

torchtitan/components/renderer.py
haydn-jones: It started failing type checking once I moved it into core (`experiments` is excluded from `pyrefly`). `BaseRendererConfig` doesn't define name and `create_renderer()` expects the union of concrete configs. I could rename to like `ExternalRendererConfig` or something if you want.

torchtitan/config/configurable.py
haydn-jones: That works when called directly but the SFT config nests it inside plain dataset/packing dataclasses. `dataclasses.asdict()` recursively flattens those before we get to the renderers `to_dict()` so that method never gets called and the `Pydantic` config ends up as a string.

.ci/docker/requirements.txt
haydn-jones: Yes

tianyu-l:
@claude give renaming / aliasing suggestions, as
- we already have `RendererConfig` as core abstraction in repo, `RenderersLibraryConfig` as a concrete impl of the abstraction to work with prime-rl `renderers` library configs
- Now this PR requires importing `RendererConfig` from prime-rl repo.

torchtitan/components/renderer.py
tianyu-l: OK, let's do `PrimeRendererConfig`. Thanks!

haydn-jones:
> @claude give renaming / aliasing suggestions, as
> 
> * we already have `RendererConfig` as core abstraction in repo, `RenderersLibraryConfig` as a concrete impl of the abstraction to work with prime-rl `renderers` library configs
> * Now this PR requires importing `RendererConfig` from prime-rl repo.

Renamed to `PrimeRendererConfig`

tianyu-l (approved):

torchtitan/components/data/README.md
```diff
@@ -213,14 +213,39 @@ config.dataloader = GrainDataLoader.Config(
...
+For multi-turn conversations, select the model's renderer explicitly:
+
+```python
+from renderers import Qwen3RendererConfig
+
+from torchtitan.components.renderer import RenderersLibraryConfig
+
+processor = ChatProcessor.Config(
+    messages_fn=lambda row: row["messages"],
+    renderer=RenderersLibraryConfig(renderers_config=Qwen3RendererConfig()),
```
slimfrkha, line 233: Is there a reason why we expose `RenderersLibraryConfig` as a config ? I checked its occurence in the repo: `RenderersLibraryConfig` takes only one attr `renderers_config` and its value is always a renderer config. 
`RenderersLibraryConfig` can still exist but not as an exposed config to the user. something like
```python
processor = ChatProcessor.Config(
    messages_fn=lambda row: row["messages"],
    renderer=renderers_config=Qwen3RendererConfig(), # no more RenderersLibraryConfig
)
```
and `RenderersLibraryConfig` is instantiated inside `ChatProcessor.Config` or somewhere else.
I feel `RenderersLibraryConfig` is in the config as a legacy code from the RL examples and now can be cleaned out from the config. 
Maybe not part of this PR but as a follow up as it touches several files unrelated to multiturn SFT.
I can create the PR once this one is merged.
cc @felipemello1 @tianyu-l

tianyu-l: I think it sounds "accidental" that `renderers` lib also uses `config.build()` style. But that doesn't mean it'd always be seamlessly usable, e.g.
- what if they use `config.create()`
- our config build would wire non-config fields to the owner of the config https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configurable.py#L160 -- what if theirs don't
- even if we use their `config.create()` for this specific integration, what if another renderer library gets popular and uses `config.generate()`

So from maintenance perspective, every config should be torchtitan native, hence the seemingly unnecessary wrapper. WDYT?

slimfrkha: oh yeah that's right !
However, `RenderersLibraryConfig` is technically a TorchTitan config, but conceptually it is an adapter (renderer to torchtian config conversion). Exposing it makes users wonder what TorchTitan-specific behavior they are supposed to configure. There is none: the only meaningful choice is already expressed by Qwen3RendererConfig. So I’d rename `RenderersLibraryConfig` to `RenderersConfigAdapter` and make these changes:

```python
  def from_renderers(config: PrimeRendererConfig) -> RendererConfig:
      """Adapt a renderers-library config to TorchTitan's renderer config interface."""
      return RenderersConfigAdapter(renderers_config=config)
```
  Then:
```
  processor = ChatProcessor.Config(
      messages_fn=lambda row: row["messages"],
      renderer=from_renderers(Qwen3RendererConfig(enable_thinking=False)),
  )
```
Keep `RenderersConfigAdapter` public for compatibility and internal cases such as Verifiers, but make from_renderers(...) the documented API.

tianyu-l: @slimfrkha 
I think that's great idea. Please submit a PR.

tianyu-l:
merging as errors are irrelevant, being fixed by @acisseJZhong

PR #2770, opened 2026-03-31
[MoE Rewrite 1/n] Use local map for torch.histc and torch.gather, and use DTensor for router 

acisseJZhong:
* __->__ #2770

**Summary**                                                           
                                                                                                
  - Refactored MoE router to run natively in DTensor (Replicate on TP mesh) instead of          
  converting to local tensors before routing. Distribute_module automatically handle parameter and buffer distribution.                                                                      
  - Replaced torch.histc (no DTensor support) with torch.scatter + sum to compute
  num_tokens_per_expert, and wrapped torch.gather and torch.scatter with local_map as DTensor  
  prop doesn't work for these two ops.
  - Updated NoParallel to handle tuple outputs (the router returns a 3-tuple of scores, indices,
   and token counts). 
  - Moved expert_bias and load_balance_coeff from MoE to TokenChoiceTopKRouter, so
  distribute_module in the parallelize plan distributes expert_bias as a DTensor buffer         
  alongside the gate weight. The to_local conversion now happens after the router in
  MoE.forward.                                                                                      
  - Updated the optimizer load-balancing hook to access expert_bias via moe.router.expert_bias
  and handle DTensor in-place updates using _local_tensor.                                      
  - Updated state dict adapters (Llama4, DeepSeek V3) and all parallelize callsites (including
  experiments) to reflect the new expert_bias path.    

**Test:**
```
NCCL_NVLS_ENABLE=0  NGPU=4 ./run_train.sh --module llama4 --config llama4_debugmodel     --parallelism.pipeline_parallel_degree 1     --parallelism.data_parallel_shard_degree 2     --parallelism.tensor_parallel_degree 2    --parallelism.expert_parallel_degree 2     --parallelism.expert_tensor_parallel_degree 2
```
**Run loss compare:**
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 2" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 2" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```

### DIFF
--- a/torchtitan/models/common/moe/moe.py
+++ b/torchtitan/models/common/moe/moe.py
@@ -317,6 +317,38 @@ class TokenChoiceTopKRouter(Module):
                 - num_tokens_per_expert (torch.Tensor):
                     Number of tokens assigned to each expert with shape ``(num_experts,)``.
         """
+        # === DEBUG: Check gate weight consistency across ranks ===
+        import torch.distributed as dist
+
+        if dist.is_initialized() and dist.get_world_size() > 1:
+            rank = dist.get_rank()
+            gate_weight = self.gate.weight
+            if isinstance(gate_weight, DTensor):
+                gate_weight_local = gate_weight.to_local()
+            else:
+                gate_weight_local = gate_weight
+            # allgather gate weights from all ranks
+            gathered = [
+                torch.zeros_like(gate_weight_local)
+                for _ in range(dist.get_world_size())
+            ]
+            dist.all_gather(gathered, gate_weight_local.contiguous())
+            # check if all ranks have the same gate weight
+            for other_rank, other_weight in enumerate(gathered):
+                if other_rank != rank:
+                    if not torch.equal(gate_weight_local, other_weight):
+                        max_diff = (gate_weight_local - other_weight).abs().max().item()
+                        print(
+                            f"[DEBUG] GATE WEIGHT DIVERGED! rank {rank} vs rank {other_rank}, "
+                            f"max_diff={max_diff:.6e}, "
+                            f"rank_{rank}_norm={gate_weight_local.norm().item():.6f}, "
+                            f"rank_{other_rank}_norm={other_weight.norm().item():.6f}"
+                        )
+                    else:
+                        print(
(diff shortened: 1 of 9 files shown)


torchtitan/models/common/moe/moe.py
```diff
@@ -278,7 +324,22 @@ def forward(
         # top scores shape (bs*slen, top_k)
         # NOTE: The expert_bias is only used for routing. The gating value
         #       top_scores is still derived from the original scores.
-        top_scores = scores.gather(dim=1, index=selected_experts_indices)
+        # gather is wrapped with local_map for DTensor support.
+        if isinstance(selected_experts_indices, DTensor):
```
tianyu-l, line 379: question 1: are they DTensor today? If not do we have test showing these impl are correct?

question 2: IIRC DTensor supports `gather` and `scatter`, so why using local_map?

question 3: what's the motivation of using `routing_map`, not clear from this PR

acisseJZhong: for question 1: I am trying to make router uses DTensor(except torch.gather and torch.scatter), today the entire MoE is wrapped in local region.

tianyu-l: They should happen in the same PR, because otherwise the DTensor branch is dead and I can't verify correctness of this PR.

acisseJZhong: yeah it's not ready yet, I converted it to draft.

Issue #2771, opened 2026-03-31
PP + Compile breaking CI


xmfan:
### Bug description

Reported by @tianyu-l, breaks CI

```python
NGPU=8 LOG_RANK=0,1,2,3,4,5,6,7 ./run_train.sh --module llama4 --config llama4_debugmodel --parallelism.pipeline_parallel_degree 2 --parallelism.pipeline_parallel_schedule Interleaved1F1B --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 4 --parallelism.expert_tensor_parallel_degree 1 --compile.enable
```

Logs: https://gist.github.com/xmfan/8ebfad161cd02af9c1ae5c5818799f43

```python
[rank1]:[rank1]: Traceback (most recent call last):
[rank1]:[rank1]:   File "/home/xmfan/core/a/pytorch/torch/distributed/pipelining/_backward.py", line 368, in stage_backward
[rank1]:[rank1]:     torch.autograd.backward(
[rank1]:[rank1]:   File "/home/xmfan/core/a/pytorch/torch/autograd/__init__.py", line 379, in backward
[rank1]:[rank1]:     _engine_run_backward(
[rank1]:[rank1]:   File "/home/xmfan/core/a/pytorch/torch/autograd/graph.py", line 877, in _engine_run_backward
[rank1]:[rank1]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank1]:[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:[rank1]:   File "/home/xmfan/core/a/pytorch/torch/utils/checkpoint.py", line 1168, in unpack_hook
[rank1]:[rank1]:     frame.check_recomputed_tensors_match(gid)
[rank1]:[rank1]:   File "/home/xmfan/core/a/pytorch/torch/utils/checkpoint.py", line 899, in check_recomputed_tensors_match
[rank1]:[rank1]:     raise CheckpointError(
[rank1]:[rank1]: torch.utils.checkpoint.CheckpointError: torch.utils.checkpoint: Recomputed values for the following tensors have different metadata than during the forward pass.
[rank1]:[rank1]: tensor at position 3:
[...]

PR #2774, opened 2026-04-01
[refactor] remove unnecessary padding in MoE

tianyu-l:
including
- remove triton kernel for permutation+padding as it is not required by default bf16 grouped mm.
- replace the kernel with a torch-native `generate_permutation_indices` impl, which doesn't do padding.
- added `TorchAOExpertParallel` to do permutation+padding during EP using the same triton kernel moved to torchao.
- added a `llama4_debugmodel_fp8` in llama4 `config_registry` to demonstrate
- refactored HybridEP to take `pad_multiple` as input so that it works with MXFP8.
- flatten `common/moe/` folder by moving `moe.py` and `moe_deepep.py` to `common/` and remove `utils.py`
- remove `dual_pipe_v.py` as it doesn't compose with SAC and is not maintained.
- disabled `compile.enable` in llama4 and gpt-oss CI as they break the CI even on main, tracking in https://github.com/pytorch/torchtitan/issues/2771 and https://github.com/pytorch/torchtitan/issues/2776, respectively

Numerics of before vs. after on llama4 debugmodel [dp8, ep4] are bitwise identical.


torchtitan/components/quantization/__init__.py
```diff
@@ -31,6 +31,27 @@ class Config(Configurable.Config):
...
-# Module level global constants
-FP8_GROUP_ALIGNMENT_SIZE = 16
-MXFP8_GROUP_ALIGNMENT_SIZE = 32
+# Mapping from quantization type to the pad_multiple needed for grouped GEMMs.
+# FP8: 16 byte alignment / 1 byte per elem = 16 elements.
+# MXFP8: scaling block size is (1 x 32), so contracting dim must be divisible by 32.
+PAD_MULTIPLE_MAP: dict[str, int] = {
+    "float8": 16,
+    "mxfp8": 32,
+}
```
fegin, line 40:
```suggestion
class PAD_MULTIPLE_MAP(IntEnum):
    float8: 16
    mxfp8: 32 
```

torchtitan/experiments/autoparallel/deepseek_v3/parallelize_deepseekv3.py
```diff
@@ -218,7 +218,7 @@ def monkey_patch_local_map_moe(model, sparse_mesh):
     """
     from torch.distributed._tensor.experimental import local_map
 
-    # from torchtitan.models.common.moe import moe
+    # from torchtitan.models.common import moe
```
fegin: Just remove this comment?

torchtitan/distributed/expert_parallel.py
```diff
@@ -22,7 +22,59 @@
 )
 from torch.distributed.tensor.parallel import ParallelStyle
 
-from torchtitan.models.common.moe.utils import _permute, _unpermute
+
+def _generate_permute_indices(
+    tokens_per_expert_group: torch.Tensor,  # shape [num_ranks * experts_per_rank]
```
fegin: It's better to move this comment to docstring.

torchtitan/distributed/expert_parallel.py
```diff
@@ -22,7 +22,59 @@
...
+def _permute(x, num_tokens_per_expert, ep_degree, num_local_experts):
+    permuted_indices, num_tokens_per_expert = _generate_permute_indices(
+        num_tokens_per_expert, num_local_experts, ep_degree
+    )
+    return x.shape, x[permuted_indices, :], permuted_indices, num_tokens_per_expert
+
+
+def _unpermute(out, input_shape, permuted_indices):
+    out_unpermuted = out.new_empty(input_shape)
+    out_unpermuted[permuted_indices, :] = out
```
danielvegamyhre, line 79: did you test mxfp8, or only bf16? in my impl i found that for the standard EP case + mxfp8, the extra padding row added at index -1 by the EP per-group padding logic (`generate_permute_indices` generates -1 indexes, then permute() implements padding by selecting this row in permuted = `tokens[permuted_indices, :]`) needs to be removed as part of unpermute. However, for the non-EP case, padding and unpadding will have been done by torchao CUDA kernels, and no such row removal is needed. Does your implementation handle this differently?

```python
    def _token_combine(
        self, mod: nn.Module, routed_output: Tensor, device_mesh: DeviceMesh
    ) -> Tensor:
        # If per group padding was done to prepare for MXFP8 grouped mm, there is an extra 'padding' row that
        # `permuted_indices` selects from to add padding into the routed_input in dispatch.
        routed_output = _unpermute(
            routed_output,
            self.input_shape,
            self.permuted_indices,
            remove_padding_row=True,
        )
```

tianyu-l: @danielvegamyhre 
This is the bf16 path. For mxfp8 please take a look at the `_unpermute` in `TorchAOExpertParallel`.

No I haven't tested mxfp8 because I don't have Blackwell dev machine. But I probably should try to get one.

torchtitan/components/quantization/__init__.py
tianyu-l: I'm trying to
- be consistent with https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/float8.py#L38
- be consistent with `https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/__init__.py#L9`
- in general torchtitan uses `str` / `Literal` instead of `Enum`

torchtitan/distributed/expert_parallel.py
danielvegamyhre: i see now - lgtm

PR #1974, opened 2025-10-31
remove scatter_add in MoE implementation

garrett361:
PR for removing `scatter_add` in the MoE implementation. `scatter_add` is somewhat problematic as it is non-deterministic due to the necessity of [atomic adds](https://discuss.pytorch.org/t/why-does-index-add-and-scatter-add-induce-non-deterministic-behavior-on-the-cuda-backend/45544/2) for correctness.

Determinism, correctness, and performance tests using scripts under `torchtitan/moe_bench_and_test`:

```
# Determinism: run same forward 100x and compute standard deviations
pytest -rsfP torchtitan/moe_bench_and_test/test_moe.py -k test_determinism

out_old_std=tensor(0.0297, device='cuda:0', dtype=torch.bfloat16)
out_std=tensor(0., device='cuda:0', dtype=torch.bfloat16)
out_old_std/out_moe_old.abs().mean()=tensor(0.0006, device='cuda:0', dtype=torch.bfloat16)
out_std/out_moe.abs().mean()=tensor(0., device='cuda:0', dtype=torch.bfloat16)
```

```
# Accuracy: compare MoE outputs to FFN outputs, with weights set such that outputs should be the same
# Relative error decreased by 3x
pytest -rsfP torchtitan/moe_bench_and_test/test_moe.py -k test_moe_ffn_equivalence

moe_old_rel_err=0.009754068047048696
moe_rel_err=0.002507858727736454
moe_old_rel_err/moe_rel_err=3.8894009216589858
```

```
# Timing: triton do_bench for DSv3 16B layer fwd + bwd. ~3% faster runtime
python torchtitan/moe_bench_and_test/moe_timing.py moe_old && python torchtitan/moe_bench_and_test/moe_timing.py moe

args=Namespace(cls='moe_old', perf_reps=1000, perf_warmups=100, seqlen=4096, bsz=4)
moe_time_ms=19.712812881469727

args=Namespace(cls='moe', perf_reps=1000, perf_warmups=100, seqlen=4096, bsz=4)
moe_time_ms=19.03301840562087

```

```
# Memory: for DSv3 16B layer fwd + bwd. ~15% reduction in active mem, ~18% in reserved mem.
python torchtitan/moe_bench_and_test/moe_memory.py moe_old && python torchtitan/moe_bench_and_test/moe_memory.py moe

args=Namespace(cls='moe_old', iters=1, seqlen=4096, bsz=4)
peak_stats.max_active_gib=5.926029682159424
[...]

PR #2775, opened 2026-04-01
[MoE] change torch.bmm back to scatter add 

acisseJZhong:
### Change
scatter_add was replaced by torch.bmm in https://github.com/pytorch/torchtitan/pull/1974 due to its non determinism. However, bmm backward kernel was reported to be slow in https://github.com/pytorch/torchtitan/issues/2225. Although @drisspg has https://github.com/pytorch/pytorch/pull/176552 which greatly optimized the bmm backward time under compile, eager mode backward is still slow. 

Therefore we try to switch back to scatter_add with `torch.use_deterministic_algorithms(True)`, see more in https://docs.pytorch.org/docs/stable/notes/randomness.html#avoiding-nondeterministic-algorithms. We tested out the determinism part using [this test script.  ](https://gist.github.com/acisseJZhong/c56dbc331d3248e488edb24894dc4565), updated [test](https://gist.github.com/acisseJZhong/562b635e6ef96d4f2ca68769dbdbad82) 

### Test with topk=2
llama4 debug model(topk=2), qwen3 debug model(topk=8)
```
NCCL_NVLS_ENABLE=0 NGPU=8 ./run_train.sh     --module llama4     --config llama4_debugmodel     --parallelism.pipeline_parallel_degree 1     --parallelism.data_parallel_shard_degree 4     --parallelism.tensor_parallel_degree 2     --parallelism.expert_parallel_degree 2     --parallelism.expert_tensor_parallel_degree 2      --metrics.enable_wandb --training.steps 100
```
image

### Loss compare

```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 4 --parallelism.expert_tensor_parallel_degree 1" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 4 --parallelism.expert_tensor_parallel_degree 1" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```
Testing with llama4 debug model, scatter_add with torch deterministic mode on seems still slightly faster than torch.bmm. 

cc @garrett361 @chelsea0x3b @rakkit @drisspg

### DIFF
--- a/torchtitan/models/common/moe.py
+++ b/torchtitan/models/common/moe.py
@@ -14,35 +14,9 @@ from torch.distributed.tensor import DTensor, Partial
 
 from torchtitan.models.common.feed_forward import FeedForward
 from torchtitan.models.common.linear import Linear
+
+from torchtitan.ops.scatter_add import deterministic_scatter_add
 from torchtitan.protocols.module import Module
-
-
-class _DeterministicScatterAdd(torch.autograd.Function):
-    @staticmethod
-    def forward(
-        ctx: torch.autograd.function.FunctionCtx,
-        out: torch.Tensor,
-        index: torch.Tensor,
-        src: torch.Tensor,
-    ) -> torch.Tensor:
-        ctx.save_for_backward(index)
-        prev = torch.are_deterministic_algorithms_enabled()
-        prev_warn_only = torch.is_deterministic_algorithms_warn_only_enabled()
-        torch.use_deterministic_algorithms(True, warn_only=False)
-        try:
-            out = out.scatter_add(dim=0, index=index, src=src)
-        finally:
-            torch.use_deterministic_algorithms(prev, warn_only=prev_warn_only)
-        return out
-
-    @staticmethod
-    def backward(
-        ctx: torch.autograd.function.FunctionCtx,
-        grad_output: torch.Tensor,
-    ) -> tuple[torch.Tensor, None, torch.Tensor]:
-        (index,) = ctx.saved_tensors
-        grad_input = torch.gather(grad_output, dim=0, index=index)
-        return grad_output, None, grad_input
 
 
 # NOTE: keeping this for-loop implementation for comparison
@@ -506,7 +480,7 @@ class MoE(Module):
                 * top_scores_experts_sorted.reshape(-1, 1)
             ).to(x.dtype)
 
-        out = _DeterministicScatterAdd.apply(
+        out = deterministic_scatter_add(
             out,
             token_indices_experts_sorted.reshape(-1, 1).expand(-1, dim),
             routed_output,
--- a/torchtitan/ops/scatter_add.py
+++ b/torchtitan/ops/scatter_add.py
@@ -1,0 +1,48 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+
+@torch.library.custom_op("torchtitan::deterministic_scatter_add", mutates_args=())
+def deterministic_scatter_add(
+    out: torch.Tensor, index: torch.Tensor, src: torch.Tensor
+) -> torch.Tensor:
+    prev = torch.are_deterministic_algorithms_enabled()
+    prev_warn_only = torch.is_deterministic_algorithms_warn_only_enabled()
+    torch.use_deterministic_algorithms(True, warn_only=False)
+    try:
+        return out.scatter_add(dim=0, index=index, src=src)
+    finally:
+        torch.use_deterministic_algorithms(prev, warn_only=prev_warn_only)
+
+
+@deterministic_scatter_add.register_fake
+def _(out: torch.Tensor, index: torch.Tensor, src: torch.Tensor) -> torch.Tensor:
+    return torch.empty_like(out)
+
+
+def _backward(
+    ctx: torch.autograd.function.FunctionCtx, grad_output: torch.Tensor
+) -> tuple[torch.Tensor, None, torch.Tensor]:
+    (index,) = ctx.saved_tensors  # pyrefly: ignore[missing-attribute]
+    grad_src = torch.gather(grad_output, dim=0, index=index)
+    return grad_output, None, grad_src
+
+
+def _setup_context(
+    ctx: torch.autograd.function.FunctionCtx,
+    inputs: tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+    output: torch.Tensor,
+) -> None:
+    _out, index, _src = inputs
+    ctx.save_for_backward(index)
+
+
+deterministic_scatter_add.register_autograd(
+    _backward,
+    setup_context=_setup_context,
+)
--- a/torchtitan/ops/__init__.py
+++ b/torchtitan/ops/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.


torchtitan/models/common/moe/moe.py
```diff
@@ -498,33 +499,25 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            ).to(x.dtype)
 
-        if out is None:
-            return out_experts.reshape(bs, slen, dim)
-        return (out + out_experts).reshape(bs, slen, dim)
+        out = out.scatter_add(
+            dim=0,
+            index=token_indices_experts_sorted.reshape(-1, 1).expand(-1, dim),
+            src=routed_output,
+        )
```
tianyu-l: Can we localize the `torch.use_deterministic_algorithms(True)` usage to just around this `scatter_add`, instead of relying on `--debug.deterministic` for the entire run?

e.g. https://www.internalfb.com/metamate/M4AK5M not sure if context manager works with compile

torchtitan/models/common/moe/moe.py
```diff
@@ -482,9 +482,10 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             token_indices_experts_sorted,
             num_tokens_per_expert,
         ) = self.reorderer(top_scores, selected_experts_indices)
+        token_indices_experts_sorted = token_indices_experts_sorted // self.router.top_k
```
tianyu-l: In general the code looks a bit different from what's before https://github.com/pytorch/torchtitan/pull/1974

Please make sure numerics are not changed. In particular, we had to modify `ReordererSequenceParallel` to make tp2ep correct.

tianyu-l: also, in this series of refactor, it would be good if you can help rename variables using https://medium.com/@NoamShazeer/shape-suffixes-good-coding-style-f836e72e24fd

garrett361:
Do you have performance numbers comparing before/after @acisseJZhong?

chelsea0x3b:
I've been doing some testing and found another way to disappear the slow bmm dispatch with just extracting the bmm into a separate function with torch.compile annotation. Boosts qwen3 30b-a30b from 22.9% mfu to 24.7%. Would ya'll be interested in a PR? its a pretty simple one without any triton kernels.

tianyu-l:
> just extracting the bmm into a separate function with torch.compile annotation.

@chelsea0x3b how's that compared with just enable compile in general? For compile, our goal is to enable it for entire transformer block, instead of on individual functions.

garrett361:
[Related discussion here](https://github.com/pytorch/torchtitan/issues/2225#issuecomment-4021847286) @chelsea0x3b

chelsea0x3b:
the MoE.forward is not compiled currently, [only its submodules are](https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/compile.py#L98):

```python
if isinstance(submod, moe_module.MoE):
    # avoid graph breaking on the GroupedExperts' FSDP hooks
    # by wrapping each submod's forward instead of their __call__
    moe = submod
    for attr_name, submod in moe.named_children():
        if attr_name == "experts":
            # NOTE: We don't compile token dispatch and token combine due to an issue on B200:
            # https://github.com/pytorch/torchtitan/issues/1940
            continue
        submod.compile(backend=compile_config.backend, fullgraph=True)
```

I *think* you could probably add 

```python
    moe.compile(..., fullgraph=False)
```

to the end and it would be equivalent? You'd need the `fullgraph=False` because the experts are skipped

I will test it out as well

acisseJZhong:
> Do you have performance numbers comparing before/after
@garrett361 yes see the screenshot in description section. Using scatter_add(with torch deterministic wrapping around) has higher throughput and mfu than using bmm, but slower than no deterministic scatter_add.

garrett361:
Oh, missed the image somehow, sorry. And weird, I found in https://github.com/pytorch/torchtitan/pull/1974#issue-3575069876 that bmm was faster, fwd + bwd. Maybe inequivalent tests.

torchtitan/models/common/moe/moe.py
acisseJZhong: I think previously we also changed `ReordererSequenceParallel` because we did `token_indices_experts_sorted = token_indices_experts_sorted // self.router.top_k` inside the reorderer. Now I move this line right after the reorderer. @garrett361 maybe could you also help confirm we don't need other changes here? 

For numerics, I ran the loss compare script with dp2tp2ep4etp1(should have tp2ep), loss is matching exactly.
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 4 --parallelism.expert_tensor_parallel_degree 1" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 4 --parallelism.expert_tensor_parallel_degree 1" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```

acisseJZhong:
> And weird, I found in https://github.com/pytorch/torchtitan/pull/1974#issue-3575069876 that bmm was faster, fwd + bwd. Maybe inequivalent tests.

maybe worth re-running your script @garrett361 

~~I also just tested qwen3 debug model, in addition to the llama4 debug model. For qwen3 debug, I think the throughput and MFU of scatter add with deterministic was slightly better than bmm / almost the same.~~

see https://github.com/pytorch/torchtitan/pull/2775#issuecomment-4173555048

torchtitan/models/common/moe/utils.py
```diff
@@ -62,6 +62,22 @@ def maybe_align_num_tokens_for_mxfp8(num_tokens: int) -> int:
     return _round_up(num_tokens, MXFP8_GROUP_ALIGNMENT_SIZE)
 
 
+def scatter_add_with_local_determinism(
```
tianyu-l: maybe just
```suggestion
def _deterministic_scatter_add(
```
inside moe.py

Please don't put in this file because I'm deleting it in https://github.com/pytorch/torchtitan/pull/2774

acisseJZhong:
> Is the perf you tested with this local determinism?

yup I just doubled check and found out I forgot to delete --deterministic flag in the command. I deleted this flag, and retested llama4 and qwen3 debug model, with topk=2, and topk=8. 
image

For qwen3 debug moe both scatter add with local determinism and bmm have similar MFU and wps. However, for llama4 debug moe bmm initially has slightly higher but later almost same MFU and wps.

tianyu-l:
hmm, how about if we compile scatter_add? I think if we compile bmm it will be quite faster.

acisseJZhong:
@tianyu-l compile scatter add and compile bmm seems to be on par
image
image

torchtitan/models/common/moe/moe.py
```diff
@@ -19,6 +19,23 @@
 from .utils import indices_padding_wrapper
 
 
+@torch.compiler.disable
```
acisseJZhong: i added this because otherwise will hit compile error. cc @xmfan not sure if this is okay.

```
 traceback : Traceback (most recent call last):        
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/distributed/elastic/multiprocessing/error 
s/__init__.py", line 367, in wrapper                    
      return f(*args, **kwargs)                         
             ^^^^^^^^^^^^^^^^^^                         
    File "/pytorch/torchtitan/torchtitan/trainer.py",   
line 835, in train                                      
      self.train_step(data_iterator)                    
    File "/pytorch/torchtitan/torchtitan/trainer.py",   
line 745, in train_step                                 
      loss = self.forward_backward_step(                
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^                
    File "/pytorch/torchtitan/torchtitan/experiments/gr 
aph_trainer/trainer.py", line 69, in                    
forward_backward_step                                   
      return super().forward_backward_step(             
    File "/pytorch/torchtitan/torchtitan/trainer.py",   
line 693, in forward_backward_step                      
      pred = model_parts[0](inputs, **extra_inputs,     
**extra_kwargs)                                         
                                                        
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/nn/modules/module.py", line 1776, in      
_wrapped_call_impl                                      
      return self._compiled_call_impl(*args, **kwargs)  
 # type: ignore[misc]                                   
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/_dynamo/eval_frame.py", line 1046, in     
compile_wrapper                                         
      raise e.with_traceback(None) from e.__cause__  #  
User compiler error                                     
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^     
  torch._dynamo.exc.Unsupported: torch.* op returned    
non-Tensor                                              
    Explanation: torch.* ops that return a non-Tensor   
cannot be traced into the Dynamo FX graph output        
                                                        
                                                        
    Developer debug context: example_value type: bool;  
op: call_function; target: <function                    
is_deterministic_algorithms_warn_only_enabled at        
0x713510aa6c00>                                         
                                                        
   For more details about this graph break, please      
visit: https://meta-pytorch.github.io/compile-graph-bre 
ak-site/gb/gb0208.html                                  
                                                        
  from user code:                                       
     File "/pytorch/torchtitan/torchtitan/models/common 
/decoder.py", line 153, in forward                      
      h = layer(h, self.freqs_cis, attention_masks,     
positions)                                              
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/nn/modules/module.py", line 1778, in      
_wrapped_call_impl                                      
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/nn/modules/module.py", line 1789, in      
_call_impl                                              
      return forward_call(*args, **kwargs)              
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/fx/traceback.py", line 384, in wrapper    
      return func(*args, **kwargs)                      
    File "/pytorch/torchtitan/torchtitan/models/deepsee 
k_v3/model.py", line 218, in forward                    
      x = x + self.moe(self.ffn_norm(x))                
    File "/opt/conda/envs/py_3.12/lib/python3.12/site-p 
ackages/torch/fx/traceback.py", line 384, in wrapper    
      return func(*args, **kwargs)                      
    File "/pytorch/torchtitan/torchtitan/models/common/ 
moe/moe.py", line 530, in forward                       
      out = _deterministic_scatter_add(                 
    File "/pytorch/torchtitan/torchtitan/models/common/ 
moe/moe.py", line 30, in _deterministic_scatter_add     
      prev_warn_only =                                  
torch.is_deterministic_algorithms_warn_only_enabled()   
                                                        
  Set TORCHDYNAMO_VERBOSE=1 for the internal stack      
trace (please do this especially if you're reporting a  
bug to PyTorch). For even more developer context, set   
TORCH_LOGS="+dynamo"   
```

garrett361:
> For qwen3 debug moe both scatter add with local determinism and bmm have similar MFU and wps. However, for llama4 debug moe bmm initially has slightly higher but later almost same MFU and wps.

> compile scatter add and compile bmm seems to be on par

If the perf is similar, is scatter_add still desirable over bmm for other reasons? I might not have all the context here.

torchtitan/models/common/moe/moe.py
tianyu-l: If you add this, we don't get the benefit of compile scatter_add? IIRC you told me compiled scatter_add has similar perf to compiled bmm, which are faster than their eager versions.

acisseJZhong:
> If the perf is similar, is scatter_add still desirable over bmm for other reasons? I might not have all the context here.

I slightly prefer scatter_add approach as the code is more explainable, and it changes in place. For the bmm approach, the `token_indices_experts_sorted` is not real indexes, and we need to have an extra unsort buffer, reshape to group by token, bmm, and squeeze, which is what scatter_add does in one step.

Curious what you think? @garrett361 @tianyu-l

acisseJZhong:
mfu drops around 0.02% for deterministic on for full run; deterministic only on scatter add is roughly on par with deterministic mode off. Not sure if we are okay with the regression, or prefer to keep deterministic on for only scatter add. 

For the later, we need to fix https://github.com/pytorch/pytorch/issues/179194 
image

garrett361:
> Curious what you think? 

I don't have very strong feelings, but don't see much reason for the change if there's not a perf advantage and we hit the graph-break issue above . I imagine that if I were seeing this for the first time/were newish to ML, I'd have many questions about the determinism wrapper.  Mildly concerned about the proposed `deterministic: bool = True`; surprised that doesn't seem to be very costly.

> it changes in place.

Do you also compare memory usage between the two impls? Curious.

> For the bmm approach, the token_indices_experts_sorted is not real indexes

Not sure what you mean? I'm sure the variable names could be better.

acisseJZhong:
deterministic = True is costly if we enabled for the whole run. But if we only wrap deterministic over scatter_add, it doesn't make much difference. See the graph above. 

I didn't compare memory exactly, but the bmm approach seems to allocating `routed_output_unsorted`, while scatter add doesn't need? 

regarding the index problem, this is where I think bmm is less intuitive than scatter add. 
```
routed_input = x[token_indices_experts_sorted // self.router.top_k]
```
the `token_indices_experts_sorted` is not in the range of [0, num_tokens], it's pending divide by topk, ideally this intermediate value should not be exposed. It should be handled inside reorderer, and this is exactly what scatter add does.

tianyu-l:
> mfu drops around 0.02% for deterministic on for full run

0.02% or 2%?
0.02% sounds negligible

acisseJZhong:
the wandb graph shows from 0.36% to 0.34%, so 0.02%.

tianyu-l:
Then you are not testing on a realistic setup. Are you using a debug model? We should test in a more realistic setup, say large enough model on multi-node GPUs.

acisseJZhong:
yeah I think qwen3 debug and llama4 debug model are the only two moe models that can fit within 8gpus. I could test tmr with multi node setup if they are more representative.

garrett361:
There are some non-debug models that I believe should work fine on 8gpus. Chelsea was using gpt-oss-20B for testing on 8xH100 in https://github.com/pytorch/torchtitan/pull/2717.

> deterministic = True is costly if we enabled for the whole run. But if we only wrap deterministic over scatter_add, it doesn't make much difference. See the graph above.

Yeah, I was a little confused about this. I see the local wrapper, but [this commit](https://github.com/pytorch/torchtitan/commit/9a16bf887a73d3aa773e173b5848622ceaa37f30) changed the global setting as well, right?

> the token_indices_experts_sorted is not in the range of [0, num_tokens], it's pending divide by topk, ideally this intermediate value should not be exposed

Agreed this is confusing. I remember it took me a while to understand what the reorderer was doing and why this `top_k` division is needed. [But doesn't the present PR still have similar code?](https://github.com/pytorch/torchtitan/pull/2775/changes#diff-70f05588a4e66cf3577cb7ad51ba15dd73d0fb5d463343c7ace2009789426c45R358)

acisseJZhong:
> I see the local wrapper, but [this commit](https://github.com/pytorch/torchtitan/commit/9a16bf887a73d3aa773e173b5848622ceaa37f30) changed the global setting as well, right?

@garrett361 yes in this commit, I set the whole run to be deterministic instead of just wrap scatter add with deterministic. 

And thanks for the reminder, I re-tested on gpt-oss 20b. Turns out the mfu regression for whole run deterministic is pretty large from 10% to 7% drop. So i'd prefer keep the deterministic wrapper only for scatter add(already updated the commit). @tianyu-l 
image

> I remember it took me a while to understand what the reorderer was doing and why this top_k division is needed. [But doesn't the present PR still have similar code?](https://github.com/pytorch/torchtitan/pull/2775/changes#diff-70f05588a4e66cf3577cb7ad51ba15dd73d0fb5d463343c7ace2009789426c45R358)

regarding the index problem, with this approach now the index returned from the token reorderer is already from zero to number of tokens. We did the division inside the token reorderer, and hided the details from users in moe forward.

drisspg:
btw I still plan to land this: https://github.com/pytorch/pytorch/pull/179082

acisseJZhong:
@drisspg yeah sounds good, could you help test out before and after this bmm optimization?

garrett361:
> Turns out the mfu regression for whole run deterministic is pretty large from 10% to 7% drop

Ok, yeah, that's more in line with what I expected, thanks for running that test.

> We did the division inside the token reorderer, and hided the details from users in moe forward

Ah, right, I remember now. That was my questionable change 😅

acisseJZhong:
> Ah, right, I remember now. That was my questionable change 😅

haha yeah would be great if you could verify if the change in expert_parallel.py looks correct to you.

garrett361:
will look, but please also quantitatively test, thanks

torchtitan/models/common/moe.py
```diff
@@ -16,6 +16,28 @@
...
+
+def _deterministic_scatter_add(
+    out: torch.Tensor,
+    token_indices: torch.Tensor,
+    dim: int,
+    src: torch.Tensor,
+) -> torch.Tensor:
+    prev = torch.are_deterministic_algorithms_enabled()
+    prev_warn_only = torch.is_deterministic_algorithms_warn_only_enabled()
+    torch.use_deterministic_algorithms(True, warn_only=False)
```
xmfan: this only works for the forward, do you care about the backward numerics?

garrett361: That's good to know! Is a full `autograd.Function` for fwd + bwd the right way to implement this, then?

acisseJZhong: I think for backward the gather should have no non-determinism accumulation. 

quoting Claude:

> In the forward, multiple source values map to the
>   same output index — e.g. tokens 3, 7, 15 all scatter
>   into out[5]. CUDA launches these additions in
>   parallel across threads, and floating-point addition
>   isn't associative, so the order of accumulation
>   affects the result. Different kernel launches can sum
>    in different orders, giving slightly different
>   results.
> 
>   In the backward, it's reversed — each src[i] came
>   from exactly one out[index[i]], so the gradient for
>   src[i] is just grad_out[index[i]] (a gather). There's
>    no accumulation into a shared location, just a
>   1-to-1 read. No race, no ordering ambiguity, so it's
>   deterministic.
> 
>   TL;DR: Forward is many-to-one (non-deterministic
>   accumulation), backward is one-to-one (deterministic
>   read).

garrett361: Oh, yeah, true. But that's a good general callout that the determinism flags don't enter the computational graph.

acisseJZhong: added a torch.autograd.Function.

torchtitan/ops/scatter_add.py
tianyu-l, line 1: `model/common` so far mainly holds shared modules used by various models. For ops, which can be potentially used by other components than model, maybe it's worth its own folder, such as `torchtitan/ops/`. Wdyt?

torchtitan/ops/scatter_add.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
acisseJZhong, line 1: custom op scatter add seems to have slightly lower mfu initially, but later on par. @xmfan please help review if this file looks good to you. 

Image

torchtitan/ops/scatter_add.py
tianyu-l, line 1: let's not call it `ops/custom_ops.py` which sounds redundant. I'd suggest
- either put it in __init__.py for now since there's only one op
- put it in `scatter_add.py`

torchtitan/ops/scatter_add.py
xmfan: custom op introduces more cpu overhead

torchtitan/models/common/moe.py
```diff
@@ -465,33 +468,25 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-            dtype=routed_output.dtype,
-            device=routed_output.device,
-        )
-        routed_output_unsorted[token_indices_experts_sorted] = routed_output
-        routed_output_unsorted = routed_output_unsorted.reshape(
-            -1, self.router.top_k, dim
+        out = (
+            self.shared_experts(x)
+            if self.shared_experts is not None
+            else torch.zeros_like(x)
```
xmfan, line 474: ?

acisseJZhong: see https://github.com/pytorch/torchtitan/pull/1974

xmfan: I don't see how that PR means we need to have a torch.zeros_like(x) that ends up being unused

acisseJZhong: sorry wdym? out is used here https://github.com/pytorch/torchtitan/pull/2775/changes#diff-9c3e17fd22b0e249ef06806990a6a33b74d45f9ce009100d3a040e27ef074a3eR483, and we don't always have shared expert right

acisseJZhong:
locally ran the failed test, it's passing 
```
NCCL_NVLS_ENABLE=0 NGPU=8 LOG_RANK=0,1,2,3,4,5,6,7 ./run_train.sh --module graph_trainer.llama3 --config graph_trainer_llama3_debugmodel_flex_attn --compile.mode aot --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 2 --compile.passes auto_bucketing,regional_inductor
```

```
[rank5]:[titan] 2026-04-06 13:33:04,407 - root - INFO - Training completed
[rank0]:[titan] 2026-04-06 13:33:04,408 - root - INFO - step: 10  loss:  4.08641  grad_norm:  1.7635  memory:  0.54GiB(0.56%)  tps: 116,880  tflops: 8.37  mfu: 0.85%
[rank0]:[titan] 2026-04-06 13:33:04,408 - root - INFO - Sleeping 2 seconds for other ranks to complete
[rank7]:[titan] 2026-04-06 13:33:04,406 - root - INFO - step: 10  loss:  4.08641  grad_norm:  1.7635  memory:  0.54GiB(0.56%)  tps: 118,494  tflops: 8.48  mfu: 0.86%
```

PR #2779, opened 2026-04-01
fix: ensure non-zero exit code on unhandled exceptions

yuxinwan-amd:
Fixes issue where training failures (e.g., ValueError from external libraries) would exit with code 0 in CI, masking errors. Now wraps main() in try/except that explicitly calls sys.exit(1) on any uncaught exception

### DIFF
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -65,4 +65,8 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        logger.error(f"Training failed with error: {e}", exc_info=True)
+        sys.exit(1)
--- a/torchtitan/train.py
+++ b/torchtitan/train.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import os
+import sys
 
 import torch
 


torchtitan/train.py
```diff
@@ -65,4 +65,8 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        logger.error(f"Training failed with error: {e}", exc_info=True)
+        sys.exit(1)
```
jagadish-amd, line 73: might need import sys

yuxinwan-amd: Good catch. I missed one local commit. adding here: fb4e97da326c370170d5b0bcd06e41e65a29d563

torchtitan/train.py
```diff
@@ -65,4 +66,8 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
```
tianyu-l, line 70: We are already doing try-catch in `main`. Do we still need to do it here?

yuxinwan-amd: Thanks for the comment. I'm closing this draft for now while I do a deeper root-cause pass. Please see the new comment

yuxinwan-amd:
Closing this draft for now while I do a deeper root-cause pass.

Summary of current status:


1. Reproduced failure: AutoParallel raises ValueError: Unsupported device on ROCm.
2. Observed mismatch: traceback is printed but command still exits 0, so CI can report green.
3. Initial mitigation attempts in test/train entrypoints did not fully address propagation in this path.

I’d like to pause this PR until I can validate the right long-term fix and add proper end-to-end verification.

PR #2741, opened 2026-03-27
Enable per-layer compile with or without MoE

weifengpy:
per-layer compile becomes possible after applying fully_shard at layer-level (no more moe level): https://github.com/pytorch/torchtitan/pull/2281 

`apply_compile`: consolidate apply_compile_dense and apply_compile_sparse into one.                             
  The only difference was capture_scalar_outputs which is harmless for dense models.                            
  Also removed the _run_experts_grouped_mm separate compile boundary and EP wrapper.                            

FSDP2 + EP: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --compile.enable --compile.components model,loss --parallelism.expert_parallel_degree 4`

FSDP2 + EP + TP: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --compile.enable --compile.components model,loss --parallelism.expert_parallel_degree 4 --parallelism.tensor_parallel_degree 2`

FSDP2 + EP + TP&ETP: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --compile.enable --compile.components model,loss --parallelism.expert_parallel_degree 4 --parallelism.tensor_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 2`

qwen3-vl: `NGPU=8 MODULE=qwen3_vl CONFIG=qwen3_vl_debugmodel_moe ./run_train.sh --compile.enable --compile.components  model,loss --parallelism.expert_parallel_degree 4 --training.steps 20`

H100 works
Screenshot 2026-04-02 at 16 28 08

Screenshot 2026-04-03 at 00 32 26

A100 also works
Screenshot 2026-04-03 at 00 26 14

PR #2781, opened 2026-04-01
[Perf] Adds missing moe.compile in apply_compile_sparse

chelsea0x3b:
Related #2775  #2225 

This PR addresses the slow bmm dispatch when compile is turned on at least, by compiling MoE.forward, which wasn't compiled before even when compilation was enabled.

Perf results with Qwen3-30B-A3B on 8xH100:
| main | this pr |
| --- | --- |
| 22.7% MFU | 27.6% MFU |

Log files:
[qwen3_30b_a3b_mfu.txt](https://github.com/user-attachments/files/26418928/qwen3_30b_a3b_mfu.txt)

### DIFF
--- a/torchtitan/distributed/compile.py
+++ b/torchtitan/distributed/compile.py
@@ -101,6 +101,7 @@ def apply_compile_sparse(
                             # https://github.com/pytorch/torchtitan/issues/1940
                             continue
                         submod.compile(backend=compile_config.backend, fullgraph=True)
+                    moe.compile(backend=compile_config.backend, fullgraph=False)
                 else:
                     submod.compile(backend=compile_config.backend, fullgraph=True)
         else:
--- a/torchtitan/distributed/compile.py
+++ b/torchtitan/distributed/compile.py
@@ -99,6 +99,7 @@ def apply_compile_sparse(
                         if attr_name == "experts":
                             # NOTE: We don't compile token dispatch and token combine due to an issue on B200:
                             # https://github.com/pytorch/torchtitan/issues/1940
+                            torch.compiler.disable(submod)
                             continue
                         submod.compile(backend=compile_config.backend, fullgraph=True)
                     moe.compile(backend=compile_config.backend, fullgraph=False)


acisseJZhong (approved):

acisseJZhong:
curious did full_graph=True break?

chelsea0x3b:
So both fullgraph=True & fullgraph=False break when EP or TP is set.

The EP failure is that mark_dynamic is forbidden by dynamo apparently:

```
File "/data/clo/torchtitan/.venv/lib/python3.12/site-packages/torch/_dynamo/symbolic_convert.py", line 1265, in call_function
      raise AssertionError(f"Attempt to trace forbidden callable {inner_fn}")
  AssertionError: Attempt to trace forbidden callable <function mark_dynamic at 0x7e7409ccf880>
  
  from user code:
     File "/data/clo/torchtitan/torchtitan/models/common/moe/moe.py", line 496, in forward
      routed_output = self.experts(routed_input, num_tokens_per_expert)
    File "/data/clo/torchtitan/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
    File "/data/clo/torchtitan/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1881, in _call_impl
      return inner()
    File "/data/clo/torchtitan/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1832, in inner
    File "/data/clo/torchtitan/torchtitan/models/common/moe/moe.py", line 131, in forward
      return run_experts_fn(w1, w2, w3, x, num_tokens_per_expert)
    File "/data/clo/torchtitan/torchtitan/distributed/compile.py", line 145, in _run_experts_grouped_mm_dynamic
      torch._dynamo.mark_dynamic(x, 0)
```

and then for TP the:
```
routed_input = x[token_indices_experts_sorted // self.router.top_k]
```

fails in the backward step with this weird error:

```
    File "/data/clo/torchtitan/.venv/lib/python3.12/site-packages/torch/_inductor/lowering.py", line 2460, in make_fallback
      assert op not in check_decomps or override_decomp, (
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  torch._inductor.exc.InductorError: AssertionError: both a fallback and a decomp for same op: aten.index_add.default
```

chelsea0x3b:
so the workaround would basically be:
```python
if not ep and not tp:
    moe.compile(..., fullgraph=True)
```

if we wanted to add this

acisseJZhong:
feels weird and maybe we shouldn't guard compile under certain parallelism.

tianyu-l (changes requested):
We are actively working on making fullgraph=True for the entire transformer block (with MoE) in it, so I don't mind waiting for that.

cc @weifengpy @xmfan @aditvenk

chelsea0x3b:
The EP case at least can be fixed by adding a `torch.compiler.disable(submod)` where we skip experts, so when we later call moe.compile the experts are still skipped. The TP thing I'm less sure about.

weifengpy:
> The EP case at least can be fixed by adding a `torch.compiler.disable(submod)` where we skip experts, so when we later call moe.compile the experts are still skipped. The TP thing I'm less sure about.

we are tracing into moe without mark_dynamic: https://github.com/pytorch/torchtitan/pull/2741

@chelsea0x3b let me know your case not contained in my test plan

FSDP2 + EP: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --compile.enable --compile.components model,loss --parallelism.expert_parallel_degree 4`

FSDP2 + EP + TP: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --compile.enable --compile.components model,loss --parallelism.expert_parallel_degree 4 --parallelism.tensor_parallel_degree 2`

FSDP2 + EP + TP&ETP: `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --compile.enable --compile.components model,loss --parallelism.expert_parallel_degree 4 --parallelism.tensor_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 2`

PR #2782, opened 2026-04-01
[rl] [1/n] Refactor policy trainer

joecummings:
Goals of this refactor:

Simplify, do less.

Manifests as:
1. Match the existing patterns of Torchtitan's trainer for the pieces we use. If the same things look different in two places, it's just confusing
2. Solidify API and introduce data types. We borrow ideas from forge and tinker on how a training client should be set up. We introduce some datatypes that are **very flexible** just to show that we should be fitting clean boundaries here

Everything else is downstream of these changes.

### DIFF
--- a/torchtitan/experiments/rl/types.py
+++ b/torchtitan/experiments/rl/types.py
@@ -5,41 +5,109 @@
 # LICENSE file in the root directory of this source tree.
 
 from dataclasses import dataclass
+from typing import Any
+
+import torch
+
+
+# ---------------------------------------------------------------------------
+# Legacy type -- used by existing code during migration. Will be removed once
+# all callers have been updated to use the new stage-typed data model.
+# ---------------------------------------------------------------------------
 
 
 @dataclass
-class Episode:
-    """
-    A single prompt + completion pair with reward and advantage.
+class LegacyEpisode:
+    """A single prompt + completion pair with reward and advantage.
 
-    The generator creates Episodes (with group_id, no reward yet).
+    The generator creates LegacyEpisodes (with group_id, no reward yet).
     The grader fills in the reward.
     The controller computes advantages across episodes sharing a group_id.
-    The trainer consumes the final Episodes with advantages set.
-
-    Attributes:
-        policy_version: Version of policy that produced this episode.
-        prompt_token_ids: Token IDs for the prompt.
-        text: Decoded completion text.
-        token_ids: Completion token IDs.
-        token_log_probs: Per-token log probabilities from the generator.
-        expected_answer: Expected answer for reward computation.
-            Passed to Episode by the generator — the generator
-            does not read this field.
-        reward: Scalar reward assigned by the grader.
-        group_id: Identifies which group this episode belongs to.
-            Episodes with the same group_id share a prompt and have
-            their advantages normalized together.
-        advantage: Advantage value computed by the controller (GRPO:
-            reward minus group mean reward).
+    The trainer consumes the final LegacyEpisodes with advantages set.
     """
 
     policy_version: int
     prompt_token_ids: list[int]
     text: str
     token_ids: list[int]
-    token_log_probs: list[float]
+    logprobs: list[float]
     expected_answer: str = ""
     reward: float = 0.0
     group_id: str = ""
     advantage: float = 0.0
+
+
+# Keep backward-compatible alias so existing imports still work.
+Episode = LegacyEpisode
+
+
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -59,184 +48,192 @@ class Config(Configurable.Config):
...
-        # Build optimizer and LR scheduler
+        # Build optimizer and LR scheduler (after parallelize, matches main Trainer)
         self.optimizers = config.optimizer.build(model_parts=self.model_parts)
         self.lr_schedulers = config.lr_scheduler.build(
             optimizers=self.optimizers,
             training_steps=config.training.steps,
         )
 
+        # RL-specific state
         self.policy_version = 0
```
tianyu-l, line 120: nit: let's put this towards the end of this function

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -6,51 +6,40 @@
...
-
-    Uses ModelSpec for model construction, parallelization, and weight loading.
-
-    Args:
-        config: PolicyTrainer.Config for model/optimizer/parallelism settings.
-        model_spec: Model specification (model config, parallelize_fn, state_dict_adapter).
-        hf_assets_path: Path to HF assets folder for checkpoint loading.
-    """
-
+class PolicyTrainer(Actor, Stateful, Configurable):
```
tianyu-l, line 35: should this just inherit main trainer (and actor)?

They don't share the same set of functions, but we may be able to reduce the constant effort to update both sides. We may explicit inherit and ban the functions we don't need here.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -59,184 +48,192 @@ class Config(Configurable.Config):
...
         parallelism: ParallelismConfig = field(default_factory=ParallelismConfig)
-        comm: CommConfig = field(default_factory=CommConfig)
-        """Communication configuration for distributed initialization."""
         compile: CompileConfig = field(default_factory=CompileConfig)
+        comm: CommConfig = field(default_factory=CommConfig)
+        checkpoint: CheckpointManager.Config = field(
+            default_factory=CheckpointManager.Config
+        )
+
+    weight_sync_transfer_dtype: Literal["bfloat16", "float32"] = "bfloat16"
```
tianyu-l, line 57: this seems hardcoded, should it be put in config?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -59,184 +48,192 @@ class Config(Configurable.Config):
...
 
-        return model
+        return OptimStepResult(
+            grad_norm=(grad_norm.item() if hasattr(grad_norm, "item") else grad_norm),
+            policy_version=self.policy_version,
+        )
 
     @endpoint
-    async def push_model_state_dict(self) -> None:
+    async def push_weights(self) -> None:
```
tianyu-l, line 256: I had the comment that "weights" doesn't sound formal -- if we use weights, what about "bias"?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -254,126 +251,16 @@ async def push_model_state_dict(self) -> None:
...
-                    dist.all_reduce(param.grad, op=dist.ReduceOp.AVG)
-
-        # Gradient clipping
-        grad_norm = dist_utils.clip_grad_norm_(
-            [p for m in self.model_parts for p in m.parameters()],
-            self.config.training.max_norm,
-            foreach=True,
-            pp_mesh=self.parallel_dims.get_optional_mesh("pp"),
+            transfer_dtype=TORCH_DTYPE_MAP[self.weight_sync_transfer_dtype]
+            if self._cast_dtypes_for_weight_sync
```
tianyu-l, line 277: why do we need this variable?

torchtitan/experiments/rl/actors/trainer.py
joecummings: I'd like to move in that direction, yes - this is the first step. That would require some updates to the main trainer so starting by just making them as similar as possible. That way it's easier for an LLM to pick up changes and apply regardless.

tianyu-l: Sg. Just fyi we are still doing refactor to main trainer. @tugsbayasgalan and I will be adding eager chunked ce loss, which will necessitate changing trainer a bit.

PR #2783, opened 2026-04-01
[RL] Add support for Compiled loss

Lucaskabela:
## Summary

Enable compiled loss matching general torchtitan trainer

### Testplan
```bash
```

### Results:
```
[actor=<root>] Eval: Accuracy=60% (12/20) Format=95% (19/20)
[actor=<root>] Post-training: Accuracy=60% (12/20) Format=95% (19/20)
```

Per step timing data:
| Metric | Inductor (compiled loss) | No compile |
  |--------|--------------------------|------------|
  | Step 1 | 8.9s | 9.0s |
  | Step 2 | 8.0s | 8.0s |
  | Step 3 | 8.4s | 8.8s |
  | Step 4 | 8.2s | 8.1s |
  | Step 5 | 8.3s | 7.9s |
  | Step 6 | 9.0s | 9.5s |
  | Step 7 | 8.6s | 8.5s |
  | Step 8 | 8.2s | 7.7s |
  | Step 9 | 9.1s | 8.9s |
  | **P50** | **8.4s** | **8.5s** |
  | **Total (1-9)** | **76.7s** | **76.4s** |

So no real change yet; however with batching forward (coming soon from @joecummings we could expect to see much larger improvements

### DIFF
--- a/torchtitan/experiments/rl/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/models/vllm_wrapper.py
@@ -54,6 +54,32 @@ def _dtensor_safe_weak_ref_tensor(tensor):
 _torch_utils.weak_ref_tensor = _dtensor_safe_weak_ref_tensor
 
 
+# NOTE: Monkeypatch DTensor placement types' __repr__ to use fully
+# qualified names. vLLM's compilation/codegen.py falls back to repr()
+# when stringifying non-primitive graph args, producing bare names like
+# "Shard(dim=2)". The generated execution_fn is exec'd in a namespace
+# that only does "import torch", so these bare references raise
+# NameError. qwen3's TP parallelize plan emits redistribute(placements=
+# (Shard(1),...)) which hits this path. Returning a qualified name like
+# "torch.distributed.tensor.placement_types.Shard(dim=2)" resolves
+# cleanly against the torch-only globals.
+from torch.distributed.tensor.placement_types import (  # noqa: E402
+    Partial as _Partial,
+    Replicate as _Replicate,
+    Shard as _Shard,
+)
+
+_Shard.__repr__ = (
+    lambda self: f"torch.distributed.tensor.placement_types.Shard(dim={self.dim})"
+)
+_Replicate.__repr__ = (
+    lambda self: "torch.distributed.tensor.placement_types.Replicate()"
+)
+_Partial.__repr__ = (
+    lambda self: f"torch.distributed.tensor.placement_types.Partial({self.reduce_op!r})"
+)
+
+
 def create_torchtitan_config_from_vllm_config(
     vllm_config: VllmConfig,
 ) -> tuple[ParallelDims, ParallelismConfig]:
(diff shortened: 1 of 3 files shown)


torchtitan/experiments/rl/actors/utils.py
```diff
@@ -84,12 +88,98 @@ def compute_token_log_probs(
...
+
+
+def build_policy_gradient_loss(
+    compile_config: CompileConfig,
+) -> Callable[..., tuple[torch.Tensor, ...]]:
+    """Optionally compile ``policy_gradient_loss`` following the build_*_loss pattern."""
+    loss_fn = policy_gradient_loss
+    if compile_config.enable and "loss" in compile_config.components:
+        # if compile_config.enable and "loss" in compile_config.components:
+        loss_fn = torch.compile(loss_fn, backend="inductor", fullgraph=True)
```
acisseJZhong: do we need to make backend and fullgraph configurable? Or is it we always use inductor and full graph for loss compilation?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -84,12 +88,98 @@ def compute_token_log_probs(
     return token_lps
 
 
+def policy_gradient_loss(
```
acisseJZhong: wondering why do you only compile this part of the loss not the entire compute_policy_gradient_loss?

tianyu-l (commented):
> So no real change yet; however with batching forward (coming soon from @joecummings we could expect to see much larger improvements

Thanks. Personally I would prefer if we could wait until batching lands and test again.

torchtitan/experiments/rl/actors/utils.py
Lucaskabela: We can make it configurable, but fullgraph=True is best for performance so shouldn't change

And since there aren't numerics (no concern of how vLLM does it since this is strictly in the trainer) I would default backend inductor here to get perf improvements; with eager and aot_eager, I don't think there would be much/any perf improvements to be had here

torchtitan/experiments/rl/actors/utils.py
Lucaskabela: The entire `compute_policy_gradient_loss` has calls that will cause graph breaks (like .item() for metrics) which may result in worse performance than no compile at all - this is a good starting point in that we group all the heavy tensor ops to give opportunity for fusions, etc while not losing speed from graph breaks

Lucaskabela:
Quick timing test results:

| shape  | batch | max_len | eager   | compiled (dynamic) | vs eager |                                   
  |--------|-------|---------|---------|--------------------|----------|
  | small  | 5     | 120     | 0.94 ms | **0.80 ms**        | **1.17x** |                                  
  | medium | 8     | 300     | 0.98 ms | **0.80 ms**        | **1.22x** |                                  
  | large  | 16    | 490     | 1.28 ms | **1.10 ms**        | **1.17x** |

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -54,6 +54,32 @@ def _dtensor_safe_weak_ref_tensor(tensor):
 _torch_utils.weak_ref_tensor = _dtensor_safe_weak_ref_tensor
 
 
+# NOTE: Monkeypatch DTensor placement types' __repr__ to use fully
```
Lucaskabela, line 57: NOTE: I am working on fixing this upstream

PR #2785, opened 2026-04-01
[Module] Model Config Programming

fegin:
* __->__ #2785

**Summary**
The title is a bit click-baity, but it captures the goal of this and previous PRs. After this PR, all model modules' build() require no kwargs. Each model spec should be fully defined, calculated, and inferred before build() is called — basically during configuration authoring time.

There are many different ways to program the configurations; this (and previous) PRs use one design that I think balances readability, explicitness, and verbosity. It may not fit other people's taste. We can further discuss how default core model config programming style should look (**after this PR is landed, rebasing is a nightmare for such a PR, lol**). For experimental models, authors/owners can decide however they want to orchestrate the model construction.

**Key changes**
- Remove all `field(init=False)` fields.
- No template fields.
- All the configs fields must be constructed before `build()` is called.
- No, kwargs is allowed unless the kwargs is not in the config. Overwriting config must happen before `build()`.

### DIFF
--- a/torchtitan/experiments/graph_trainer/tests/test_passes.py
+++ b/torchtitan/experiments/graph_trainer/tests/test_passes.py
@@ -37,9 +37,7 @@ class ToyModel(Module):
         super().__init__()
 
         def _make_linear():
-            cfg = Linear.Config(bias=True)
-            cfg.in_features = dim
-            cfg.out_features = dim
+            cfg = Linear.Config(in_features=dim, out_features=dim, bias=True)
             return cfg.build()
 
         self.layers = ModuleList([_make_linear() for _ in range(n_layers)])


torchtitan/config/configurable.py
tianyu-l: Not sure if I've asked before:
I think replace will run `post_init` again which can serve as validation for config classes. Is there a way we can run this check with the overrides?

tianyu-l: actually would `_replace` ever be called with non-empty `overrides`, after this PR?

torchtitan/models/common/attention.py
```diff
@@ -552,8 +552,8 @@ class GQAttention(BaseAttention):
     @dataclass(kw_only=True, slots=True)
     class Config(BaseAttention.Config):
         n_heads: int
-        wqkv: Linear.Config
-        wo: Linear.Config
+        wqkv: Linear.Config  # template for wq, wk, wv (expand_layer_configs creates wq/wk/wv)
```
tianyu-l: similar as objection in first PR: we shouldn't have both expanded and unexpanded config in the same parent config.

torchtitan/models/common/attention.py
```diff
@@ -587,24 +594,16 @@ def __init__(self, config: Config, *, dim: int):
...
-        )
-        self.wv = config.wqkv.build(
-            in_features=dim, out_features=self.n_kv_heads * self.head_dim
-        )
-        self.wo = config.wo.build(
-            in_features=self.n_heads * self.head_dim, out_features=dim
-        )
+        self.wq = config.wq.build()
+        self.wk = config.wk.build()
+        self.wv = config.wv.build()
```
tianyu-l: still reviewing, but what's the motivation of having seperate q,k,v config?

torchtitan/models/common/feed_forward.py
```diff
@@ -36,27 +36,25 @@ class FeedForward(Module):
...
+    by ``expand_layer_configs()`` before ``build()`` is called.
     """
 
     @dataclass(kw_only=True, slots=True)
     class Config(Module.Config):
         hidden_dim: int
-        w1: Linear.Config
-        w2w3: Linear.Config
+        w1: Linear.Config  # template for w1 (base init)
+        w2w3: Linear.Config  # template for w2, w3 (depth-scaled init)
```
tianyu-l: same

torchtitan/models/qwen3/__init__.py
```diff
@@ -567,43 +602,8 @@ def _235b_a22b():
 }
 
 
-def model_registry(flavor: str, attn_backend_override: str | None = None) -> ModelSpec:
```
tianyu-l, line 570: I thought this was convenient, what's the motivation of reverting back to old style?

torchtitan/models/llama4/__init__.py
tianyu-l, line 180: Decoder doesn't need to own this, right?

Instead of utils like `fill_decoder_fields`, I would prefer setting them in the function just like `dim` and `n_heads` and directly apply them in `Llama4Model.Config` -- it doesn't have to be `init=False`.

torchtitan/models/qwen3/__init__.py
fegin: If you are already doing the config programming, why do you need to set `attn_backend_override during model_registry time? We can just create config_flex, config_varlen, config_sdpa and all three share the same base config and the difference is attn_backend_override.

tianyu-l: Oh the reason is that:
In torchtitan RL effort we only support "varlen" for now, while the default is SDPA. So for any existing model, we'll have to create a varlen variation, and later flex variation, where the only difference is inner attention type. So providing this `attn_backend` arg saves cartesian product number of model configs.

The idea is similar to Ailing's argument -- users only want to change certain arg more often than others. So it's purely out of convenience concern. In some sense, attention is pretty much the only thing in model code user need to tune, and maybe also EP token dispatcher that @acisseJZhong is working on. So I don't expect the args to `model_registry` to explode much.

torchtitan/config/configurable.py
fegin: No, lol

torchtitan/models/qwen3/__init__.py
fegin: But we still create `qwen3_debugmodel_flex` and other similar configs anyway. If so, why do we have two layers of programming? Why not just centralize the same programming flow in once place, `__init__.py`? Now we need to check `config_registry.py` to understand someone will pass arguments to `mode_registry` and we expand it based on that arguments.

We can just pre-register all the model config functions like `qwen3_debugmodel_flex: _qwen3_debugmodel_flex` in __init__.py. `model_registry(flavor: str)` just calls `config = llama3_configs[flavor]()` and it will guarantee different combinations. I don't see why this is less convenient. And we can leave the arguments for `model_registry` for runtime information that cannot be decided during `__init__.py` is imported.

tianyu-l: I think the actual reason was code dedup for cartesian product number of configs [flavors] x [sdpa, varlen, flex]. If your proposal somehow share the code for different combinations, that's OK.

However, this will still be creating cartesian number of model config names, `"modelname_[# of flavor]_[#_attn_backends]"`, which I'm OK with but don't see the benefit.

fegin: The current solution also causes the cartesian product problem, doesn't it? `config_registry.py` needs to do the cartesian product definitions, no? I'm not strongly against that, but I don't see the difference of the two approaches because both have the cartesian product problem. Am I missing anything?

fegin: More concretely, the current solution defines `_qwen3_debugmodel_flex`, `_qwen3_debugmodel_varlen`, and `_qwen3_debugmodel_sdpa` in `config_registry.py`. The approach here is to define the same configurations in `__init__.py` using shared base configs, similar to what the current solution does. But it achieve this without passing `attn_backend_override` to `model_registry()`. 

The config number are the same. If the current solution only defines `debugmodel + #_attn_backends`, then with this approach, the same number of configurations will be defined in `_qwen3.__init__.py`. If the current solution enumerates all, this approaches will also enumerate all.

torchtitan/models/llama4/__init__.py
fegin: `dim` and `vocab_size` are required. we do have some downstream code directly accesses these two. `n_layers` can be inferred from `len(layers)`, so I removed it. As for `dim` and `vocab_size`, we can reorganize the downstream usage, but I think it is out of the scope of this PR.

torchtitan/models/qwen3/__init__.py
tianyu-l: model config and job config are two different space, the former is limited, the latter is unlimited

e.g. usually you won't call a job config `_qwen3_debugmodel_flex `, instead it will be
- `_qwen3_debugmodel_flex_1_5_trillions_tokens_xyz_data_mixing`
- `_qwen3_debugmodel_varlen_amd_gpu_midtraining_phase_3`
- etc.

fegin: Implementation wise, there are not difference since they are both functions, so if you have _qwen3_debugmodel_flex as a training config, which is a function, you will also have qwen3_debugmodel_flex function for this, which will only be executed if users use it. The examples you gave include non-model configurations, but we have argued this last year, attention backend should be a model config.

tianyu-l: @claude review this conversion and give your opinions

tianyu-l:
@claude review the conversations under https://github.com/pytorch/torchtitan/pull/2785#discussion_r3024735101, list the tradeoffs, and give your opinion

torchtitan/config/configurable.py
```diff
@@ -99,81 +63,27 @@ def _convert(val):
...
-                            raise ValueError(
-                                f"{type(self).__name__}.build() conflict for "
-                                f"'{key}': config has {existing!r} "
-                                f"but got {value!r}"
-                            )
-                return self._owner(config=self._replace(**kwargs))
-
-            # TODO: Old style, will be deprecated.
-            return self._owner(config=self._replace(), **kwargs)
+            return self._owner(config=replace(self), **kwargs)
```
fegin, line 93: TODO: we need a check that kwargs contains zero arguments from config.

torchtitan/config/configurable.py
```diff
@@ -192,8 +102,5 @@ def __init_subclass__(cls, **kwargs):
                             f"{cls.__name__}.Config field '{f.name}' "
                             "must be keyword-only"
                         )
-                # Override @dataclass-generated __repr__ with our safe
```
fegin, line 195: TODO: remove the comment. It is outdated.

torchtitan/experiments/vlm/model/siglip2.py
```diff
@@ -121,22 +115,20 @@ class Attention(Module):
 
     @dataclass(kw_only=True, slots=True)
     class Config(Module.Config):
-        qkv: Linear.Config  # shared config for q, k, v projections
+        qkv_proj: Linear.Config  # shared config for q/k/v (build() copies)
```
fegin: TODO: the comment is verbose, remove.

torchtitan/experiments/vlm/model/siglip2.py
```diff
@@ -121,22 +115,20 @@ class Attention(Module):
...
 
-        self.q_proj = config.qkv.build(in_features=self.dim, out_features=self.dim)
-        self.k_proj = config.qkv.build(in_features=self.dim, out_features=self.dim)
-        self.v_proj = config.qkv.build(in_features=self.dim, out_features=self.dim)
-        self.out_proj = config.out_proj.build(
-            in_features=self.dim, out_features=self.dim
-        )
+        self.q_proj = config.qkv_proj.build()  # build() copies
+        self.k_proj = config.qkv_proj.build()  # build() copies
+        self.v_proj = config.qkv_proj.build()  # build() copies
```
fegin: TODO: the comments are verbose, remove.

torchtitan/models/common/attention.py
```diff
@@ -552,16 +552,16 @@ class GQAttention(BaseAttention):
     @dataclass(kw_only=True, slots=True)
     class Config(BaseAttention.Config):
         n_heads: int
-        wqkv: Linear.Config
-        wo: Linear.Config
+        dim: int
+        wq: Linear.Config  # query projection
+        wkv: Linear.Config  # shared config for key + value (build() copies)
+        wo: Linear.Config  # output projection
```
fegin: TODO: the comments are verbose, remove.

torchtitan/models/common/attention.py
```diff
@@ -587,24 +589,16 @@ def __init__(self, config: Config, *, dim: int):
...
-        )
-        self.wv = config.wqkv.build(
-            in_features=dim, out_features=self.n_kv_heads * self.head_dim
-        )
-        self.wo = config.wo.build(
-            in_features=self.n_heads * self.head_dim, out_features=dim
-        )
+        self.wq = config.wq.build()
+        self.wk = config.wkv.build()  # build() copies — independent module
+        self.wv = config.wkv.build()  # build() copies — independent module
```
fegin: TODO: the comments are verbose, remove.

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Shared config builder helpers for model registries.
+
+These helpers construct fully-specified sub-configs with all dimensional
+fields set at config creation time.  Module ``__init__`` methods call
+``build()`` directly — no expansion or mutation needed.
```
fegin: TODO:
```suggestion
These helpers construct fully-specified sub-configs with all dimensional
fields set at config creation time.
```

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+from torchtitan.models.common.moe import GroupedExperts, MoE, TokenChoiceTopKRouter
+from torchtitan.models.common.rmsnorm import RMSNorm
+
+
+def make_gqa_config(
+    *,
+    dim: int,
+    n_heads: int,
+    wqkv_param_init: dict,
+    wo_param_init: dict,
```
fegin: TODO: typing should be `dict[str, Callable]`

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+def make_ffn_config(
+    *,
+    dim: int,
+    hidden_dim: int,
+    w1_param_init: dict,
+    w2w3_param_init: dict,
+) -> FeedForward.Config:
+    """Build a fully-specified FeedForward.Config.
+
+    ``w1_param_init`` is for w1.  ``w2w3_param_init`` is shared for w2 and w3.
```
fegin: TODO: remove this obvious comment.

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+    use_rope: bool = True,
+    mask_type: str = "causal",
+    rope_backend: str = "complex",
+    q_norm: RMSNorm.Config | None = None,
+    k_norm: RMSNorm.Config | None = None,
+) -> GQAttention.Config:
+    """Build a fully-specified GQAttention.Config.
+
+    ``wqkv_param_init`` is shared for wq and wkv (wk/wv).
+    ``wo_param_init`` is for the output projection.
```
fegin: TODO: remove these obvious comments.

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+        rope_backend=rope_backend,
+    )
+
+
+def make_ffn_config(
+    *,
+    dim: int,
+    hidden_dim: int,
+    w1_param_init: dict,
+    w2w3_param_init: dict,
```
fegin: TODO: typing should be dict[str, Callable]

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+        experts=experts,
+        shared_experts=shared_experts,
+    )
+
+
+def make_router_config(
+    *,
+    dim: int,
+    num_experts: int,
+    gate_param_init: dict,
```
fegin: TODO: typing should be dict[str, Callable]

torchtitan/models/common/config_utils.py
```diff
@@ -0,0 +1,158 @@
...
+        num_limited_groups=num_limited_groups,
+    )
+
+
+def make_experts_config(
+    *,
+    dim: int,
+    hidden_dim: int,
+    num_experts: int,
+    param_init: dict,
```
fegin: TODO: typing should be dict[str, Callable]

tianyu-l:
@claude I partly disagree with your comments. Here are my arguments.

> Architectural consistency. The PR's core thesis is that model configs should be fully determined at configuration authoring time, not at model_registry() call time.

The real constraint we would like to have is not "model configs should be fully determined at configuration authoring time", but "model configs should be fully determined before calling config.build()". As long as the config we obtain from config_registry.py is fully ready, that's good enough.

> As fegin correctly points out, both approaches produce the same number of named configurations. In Approach A, config_registry.py still defines qwen3_debugmodel_flex, qwen3_debugmodel_varlen, etc. — the combinatorics aren't avoided, just split across two files.

I disagree, because in reality you may not need the full cartesian product of model configs in config_registry. The attn_backend arg in model_registry gives you the option to not pre-define a model config if it's not used by any config_registry. One could argue that we don't have to have all the full cartesian product of model configs such as "_0_6b_varlen" in __init__.py, either. But it would be hard to predict what would be needed, what wouldn't be.

Moreover, the risk of having exploded amount of model configs is real. E.g. if we add more options in addition to attn_backend, let's say three different ep_comm_backend options, then the number of pre-registered model configs would be num_model_size_variants * num_attn_backends * num_ep_comm_backend, which will soon become a burden to maintain. You could introduce helper functions that's shared by multiple model configs, but that seems not very different from introducing args in `model_registry()` function.

torchtitan/models/deepseek_v3/model.py
```diff
@@ -217,15 +199,13 @@ def update_from_config(
...
                 assert isinstance(layer_cfg.attention, Attention.Config)
-                layer_cfg.attention = dataclasses.replace(
-                    layer_cfg.attention,
-                    rope_max_seq_len=seq_len,
-                    rope_factor=self.rope.rope_factor,
-                    rope_original_seq_len=self.rope.original_seq_len,
-                )
+                layer_cfg.attention.rope_max_seq_len = seq_len
+                layer_cfg.attention.rope_factor = self.rope.rope_factor
+                layer_cfg.attention.rope_original_seq_len = self.rope.original_seq_len
```
fegin, line 211: TODO: we should still use replace.

torchtitan/models/flux/model/model.py
```diff
@@ -131,30 +127,14 @@ def __init__(self, config: Config):
...
-        self.single_blocks = ModuleList(
-            [cfg.build() for cfg in config.single_blocks_expanded]
-        )
+        self.txt_in = config.txt_in.build()
+
+        self.double_blocks = ModuleList([cfg.build() for cfg in config.double_blocks])
+
+        self.single_blocks = ModuleList([cfg.build() for cfg in config.single_blocks])
 
         self.final_layer = config.final_layer_config.build()
```
fegin, line 139: TODO: remove the unnecessary blank lines.

torchtitan/models/flux/__init__.py
```diff
@@ -43,60 +43,114 @@
...
-    ]
-
-
 def _make_double_block_config(
     hidden_size: int,
     num_heads: int,
     mlp_ratio: float,
     qkv_bias: bool,
 ) -> DoubleStreamBlock.Config:
+    hs = hidden_size
```
fegin, line 52: TODO: do not rename hidden_size to hs.

torchtitan/models/flux/__init__.py
```diff
@@ -105,43 +159,93 @@ def _make_single_block_config(
     num_heads: int,
     mlp_ratio: float,
 ) -> SingleStreamBlock.Config:
+    hs = hidden_size
```
fegin, line 162: TODO: do not rename hidden_size to hs.

torchtitan/models/qwen3/__init__.py
```diff
@@ -36,529 +40,591 @@
...
+        "w2": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
+        "w3": partial(nn.init.trunc_normal_, std=depth_scaled_std(0.02, layer_id)),
+    }
+
+
+def _qwen3_norm(dim: int) -> RMSNorm.Config:
+    return RMSNorm.Config(normalized_shape=dim, eps=_EPS, param_init=_NORM_INIT)
+
+
+def _qwen3_q_norm() -> RMSNorm.Config:
```
fegin: TODO: same as _qwen3_norm, we should pass the dim instead of using _HEAD_DIM.

torchtitan/models/qwen3/__init__.py
```diff
@@ -36,529 +40,591 @@
...
+    for layer_id in range(n_layers):
+        layers.append(
+            Qwen3TransformerBlock.Config(
+                attention_norm=_qwen3_norm(dim),
+                ffn_norm=_qwen3_norm(dim),
+                attention=make_gqa_config(
+                    dim=dim,
+                    n_heads=n_heads,
+                    n_kv_heads=n_kv_heads,
+                    head_dim=_HEAD_DIM,
```
fegin: TODO: head_dim should be an argument of `_build_qwen3_layers`

torchtitan/models/qwen3/__init__.py
```diff
@@ -36,529 +40,591 @@
...
```
fegin: TODO: head_dim should be an argument of _build_qwen3_moe_layers

torchtitan/models/qwen3/__init__.py
```diff
@@ -36,529 +40,591 @@
...
+                    ),
+                ),
+            )
+        )
+    return layers
+
+
+def _debugmodel() -> Qwen3Model.Config:
     dim = 256
-    head_dim = 128
```
fegin, line 90: TODO: do not remove head_dim, define it explicitly here. Apply this rule to all other model definitions.

torchtitan/models/common/decoder.py
```diff
@@ -77,34 +76,22 @@ class Config(BaseModel.Config):
         # and Attention. Also RoPE itself as a standalone module requires PP special
         # handling, see below.
         rope: RoPE.Config
-        layer: TransformerBlock.Config
-        layers: list | None = None
+        layers: list  # list[TransformerBlock.Config] or subclass configs
```
tianyu-l, line 82: can we just type it with `list[TransformerBlock.Config]`? I feel it should be compatible as long as each element in the list is a subclass of `TransformerBlock.Config`

torchtitan/models/common/decoder.py
```diff
@@ -202,5 +189,5 @@ def get_attention_masks(
 
     @property
     def attn_config(self):
-        """Convenience accessor for the attention config from layer."""
-        return self.config.layer.attention
+        """Convenience accessor for the attention config from the first layer."""
+        return self.config.layers[0].attention
```
tianyu-l: I'm thinking if we should just delete this function. But if we don't, the key should be str, not int. In PP, you may not always have 0 as key.

I would prefer we remove this because it's leaking / relying on each layer's implementation details.

torchtitan/models/common/decoder.py
tianyu-l: btw the actual layers is a module Dict. Do we want to make this config a dict, to be consistent?

torchtitan/models/deepseek_v3/model.py
```diff
@@ -71,33 +72,18 @@ def __init__(self, config: Config, *, dim: int):
...
-        self.wo = config.wo.build(
-            in_features=self.n_heads * self.v_head_dim, out_features=self.dim
-        )
+            self.wq_a = config.wq_a.build()
+            self.q_norm = config.q_norm.build()
+            self.wq_b = config.wq_b.build()
+        self.wkv_a = config.wkv_a.build()
+        self.kv_norm = config.kv_norm.build()
+        self.wkv_b = config.wkv_b.build()
+        self.wo = config.wo.build()
```
tianyu-l, line 89: ooops, this is much less readable

Maybe we can introduce some post_init validation on the config?

torchtitan/models/common/decoder.py
fegin: Yes, I tried, but this causes way too many typing errors. I will revisit this later.

torchtitan/models/common/decoder.py
fegin: It should be 0 not "0", because this is a config, so list, not ModuleDict. Let's remove this though.

torchtitan/models/deepseek_v3/model.py
fegin: Curious, what's your idea of post_init validation? This is an actual build.

tianyu-l: e.g. the shapes of consecutive matmal would match against each other

In traditional attention, it was simpler to reason. For MLA it's very hard to infer what each weight's shape is, by only looking at model code in this file.

joecummings (approved):

PR #2786, opened 2026-04-01
[ROCm] Fix ROCm GPU name resolution in autoparallel workflow

jagadish-amd:
Copy PyTorch's bundled amdgpu.ids into /usr/share/libdrm/amdgpu.ids in the CI job so torch can resolve GPU names correctly on runners.


jagadish-amd:
@tianyu-l autoparallel test is passing with correct detection.
2026-04-03T13:21:01.3094857Z [rank1]:[titan] 2026-04-03 13:21:01,288 - root - INFO - CUDA capacity: AMD Instinct MI325X with 255.98GiB memory
Earlier it was failing due to "ValueError: Unsupported device:"

PR #2788, opened 2026-04-01
[graph_trainer] Start deprecating jit and aot compile modes in graph_trainer

yiming0416:
Summary:
Going forward, we want to make the minimal `make_fx` tracer the default and only graph capture path in the graph trainer to minimize the attack surface for graph capture. So we will deprecate the existing `aot` and `jit` modes that rely on Dynamo and AOT Autograd. 

In this PR:
- Start deprecating the `aot` and `jit` compile modes in the graph_trainer experiment, making `aot_fx_trace` the new default. Users can still use the old modes by explicitly passing `--compile.mode aot` or `--compile.mode jit`, which will emit a FutureWarning. 
- Update README to document `aot_fx_trace` as the primary mode and add deprecation notice for the old modes

### DIFF
--- a/torchtitan/experiments/graph_trainer/README.md
+++ b/torchtitan/experiments/graph_trainer/README.md
@@ -10,9 +10,10 @@ This experiment demonstrates graph-based distributed training in torchtitan thro
 - Graph optimization via FX graph passes
 
 The goal is to give users more explicit control over the compiler stack in terms of performance, numerics, and debuggability during large-scale distributed training. Three compilation modes are currently supported:
+- **AOT FX trace mode** (`--compile.mode aot_fx_trace`): Non-strict tracing of the full forward + loss + backward via `make_fx`, producing a single end-to-end graph without AOTAutograd partitioning.
 - **AOT mode** (`--compile.mode aot`): Explicit joint graph export with a custom graph pass pipeline.
 - **JIT mode** (`--compile.mode jit`): Standard `torch.compile()` with graph passes registered to custom backends.
-- **AOT FX trace mode** (`--compile.mode aot_fx_trace`): Non-strict tracing of the full forward + loss + backward via `make_fx`, producing a single end-to-end graph without AOTAutograd partitioning.
+> **Deprecation notice:** The `aot` and `jit` compile modes are deprecated and will be removed in the future. Please migrate to `aot_fx_trace`.
 
 ### Prerequisites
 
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/graph_trainer/README.md
```diff
@@ -48,22 +50,20 @@ NGPU=8 MODULE=graph_trainer.deepseek_v3 CONFIG=graph_trainer_deepseek_v3_16b ./r
...
 
-By default, the graph is captured with the AOT mode (you can switch to JIT mode via `--compile.mode jit`) and compiled with the `aot_eager` backend.
-
-Graph passes can be applied to further optimize the graph by using the `--compile.joint_passes` and `--compile.passes` flags.
+> **Note:** Graph pass support for `aot_fx_trace` mode is a work in progress. The `--compile.passes` and `--compile.joint_passes` flags currently only apply to the deprecated `aot` and `jit` modes.
 
 ```bash
 # Auto bucketing for comm/compute overlap
-MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --compile.passes auto_bucketing
+MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_8b ./run_train.sh --compile.mode aot --compile.passes auto_bucketing
```
SherlockNoMad, line 57: do you meant to use aot_fx_trace？

yiming0416: No this is for the deprecated `aot` mode.

yiming0416: Since we change default mode to `aot_fx_trace`, we need to explicitly add `--compile.mode aot`.

For `aot_fx_trace`, the pass itself and pass config are not working yet.

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/compile.py
```diff
@@ -277,6 +278,14 @@ def apply_compile(
         logger.info("No compile mode set, skipping compilation")
         return model
 
+    if mode in ("aot", "jit"):
+        warnings.warn(
```
aditvenk, line 244: I fully support making minimal tracer as the default, however I am not sure about this warning that we will remove aot/jit in the future. Are there future developments we anticipate that will preclude us from continuing to support dynamo aot/jit graph capture on the frontend?

tianyu-l (commented):
This PR is not making any real change, so I wouldn't block. But I would like to learn the reasoning behind deprecation, and in particular hear what @xmfan thinks.

yiming0416:
@tianyu-l @aditvenk 
The reasoning behind deprecation:
1. Based on recent tracing exercises in different repos, we conclude that using the minimal `make_fx` tracer to capture the (forward + loss + backward) step is the direction we want to pursue for graph-based large-scale pre-training.
2. Once chunked loss is added, the existing AOT mode cannot support it.
3. Reducing the burden of maintaining the existing AOT/JIT modes. We have already observed that these two modes remain brittle due to the large surface in Dynamo and AOT Autograd. CI can be easily broken by changes in PyTorch.
cc @SherlockNoMad

SherlockNoMad:
> @tianyu-l @aditvenk The reasoning behind deprecation:
> 
> 1. Based on recent tracing exercises in different repos, we conclude that using the minimal `make_fx` tracer to capture the (forward + loss + backward) step is the direction we want to pursue for graph-based large-scale pre-training.
> 2. Once chunked loss is added, the existing AOT mode cannot support it.
> 3. Reducing the burden of maintaining the existing AOT/JIT modes. We have already observed that these two modes remain brittle due to the large surface in Dynamo and AOT Autograd. CI can be easily broken by changes in PyTorch.
>    cc @SherlockNoMad

This is the correct summary.

aditvenk (approved):

tianyu-l (approved):

PR #2793, opened 2026-04-02
[Bugfix] Move recompile limit for now

Lucaskabela:
## Summary

Lower the `recompile_limit`/`cache_size` change to inside the trainer step as it was not getting picked up at top level resulting in some crash

## Test plan
```
```

### DIFF
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -80,13 +80,6 @@ class PolicyTrainer(Actor, Configurable):
         requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
         self._transfer_dtype = requested if requested != training_dtype else None
 
-        # The policy and ref models share code objects, so dynamo's
-        # per-code-object cache must hold entries for both grad modes
-        # (grad for policy, no_grad for ref). The default limit of 8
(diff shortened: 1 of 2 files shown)


Lucaskabela:
NOTE: I also tried changing the disable llmv to only `disable-slp-vectorization` in v1 of this PR, but that failed, so we need to wait for the upstream (pytorch) fix

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -80,13 +80,6 @@ def __init__(
         requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
         self._transfer_dtype = requested if requested != training_dtype else None
 
-        # The policy and ref models share code objects, so dynamo's
-        # per-code-object cache must hold entries for both grad modes
-        # (grad for policy, no_grad for ref). The default limit of 8
```
tianyu-l, line 85: why removing these helpful notes?

tianyu-l (approved):
one comment

torchtitan/experiments/rl/actors/trainer.py
Lucaskabela: I didn't want as much spam inside the step function - but I suppose it is better to have this detail especially since it is temporary (I am actively working on reducing these recompiles)

Issue #2505, opened 2026-03-05
Llama4 debug_model  with DeepEP - ValueError: not enough values to unpack (expected 5, got 2)


githubsgi:
### Bug description

Runs without error with expert_parallel_comm_backend="standard". With "deeep" see the following error. 

log:
```
[titan] 2026-03-05 21:42:01,095 - root - INFO - Building device mesh with parallelism: pp=1, dp_replicate=1, dp_shard=4, cp=1, tp=1, ep=4, etp=1
[titan] 2026-03-05 21:42:01,102 - root - INFO - Successfully created meshes with active dimensions: ['batch', 'loss', 'fsdp', 'ep']
.
.
.
[rank0]:   File "torchtitan/torchtitan/models/llama4/model.py", line 128, in forward
[rank0]:     out = h + self.moe(self.ffn_norm(h))
[rank0]:   File "/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:   File "/lib/python3.10/site-packages/torch/nn/modules/modx3205c0s7b1n0.hsn.cm.polaris.alcf.anl.gov 0: ule.py", line 1789, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:   File "torchtitan/torchtitan/models/common/moe/moe.py", line 524, in forward
[rank0]:     routed_output = self.experts(routed_input, num_tokens_per_expert)
[rank0]:   File "/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:   File "/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1884, in _call_impl
[rank0]:     return inner()
[rank0]:   File "/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1821, in inner
[rank0]:     args_result = hook(self, args)
[rank0]:   File "/lib/python3.10/site-packages/torch/distributed/tensor/_api.py", line 1125, in <lambda>
[rank0]:     lambda mod, inputs: input_fn(mod, inputs, device_mesh)
[rank0]:   File "torchtitan/torchtitan/distributed/expert_parallel.py", line 307, in _token_dispatch
[rank0]:     hidden_states, _, selected_experts_indices, top_scores, num_experts = inputs
[...]

PR #2794, opened 2026-04-02
[DeepEP] hide buffer init from sac

shuhuayu:
DeepEP `Buffer.__init__ ` calls `all_gather_object()`, which internally does `aten._to_copy` (CUDA→CPU). Our SAC policy marks `aten._to_copy` as `MUST_SAVE`, so these infrastructure ops enter the FIFO cache and later ops consume wrong saved tensors during recompute, causing shape mismatch errors. This pr wraps `get_buffer()` and `get_dispatch_layout()` in `_disable_current_modes()` in both `deepep.py` and `hybridep.py`'s `dispatch_tokens()`, hiding these ops from SAC's `__torch_dispatch__`. This should fix https://github.com/pytorch/torchtitan/issues/2505.

### DIFF
--- a/torchtitan/distributed/deepep/deepep.py
+++ b/torchtitan/distributed/deepep/deepep.py
@@ -15,6 +15,7 @@ from dataclasses import dataclass
 
 import torch
 from torch.distributed import ProcessGroup
+from torch.utils._python_dispatch import _disable_current_modes
 
 try:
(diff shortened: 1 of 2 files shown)


tianyu-l (approved):

shuhuayu:
Let's merge this first to work with SAC on A100.

PR #2799, opened 2026-04-02
[graph_trainer] Add bitwise deterministic guardrail test

SherlockNoMad:
* __->__ #2799

Add test_bitwise_deterministic.py that verifies GraphTrainer (aot_fx_trace)
and Trainer (eager) produce bitwise identical losses, model weights, and
gradients on both Llama3 and DeepSeek-v3 debug models. Uses hash_model and
hash_gradient utilities for efficient whole-model comparison.

This serves as a guardrail — all graph_trainer changes must pass this
test before proceeding with other tests.

Also adds hash_model/hash_gradient to torchtitan/tools/utils.py and
documents the guardrail in graph_trainer's CLAUDE.md.

TODO:
1. Use distributed setup for multi-GPU testing

### DIFF
--- a/torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
+++ b/torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
@@ -155,6 +155,9 @@ class TestLlama3BitwiseDeterministic(BitwiseDeterministicBase):
 
     model_registry = staticmethod(llama3_model_registry)
 
+    # TODO: Re-enable once upstream PyTorch numerical change is resolved.
+    # Broken by https://github.com/pytorch/pytorch/pull/160509
+    @unittest.skip("Upstream PyTorch change broke expected numerics")
     def test_eager_self_deterministic(self):
         """Eager mode: results match hardcoded expected values.
 
@@ -179,6 +182,9 @@ class TestDSv3BitwiseDeterministic(BitwiseDeterministicBase):
 
     model_registry = staticmethod(dsv3_model_registry)
 
+    # TODO: Re-enable once upstream PyTorch numerical change is resolved.
+    # Broken by https://github.com/pytorch/pytorch/pull/160509
+    @unittest.skip("Upstream PyTorch change broke expected numerics")
     def test_eager_self_deterministic(self):
         """Eager mode: results match hardcoded expected values.
 


torchtitan/tools/utils.py
```diff
@@ -168,6 +168,116 @@ def get_peak_flops(device_name: str) -> float:
...
+        return ""
+
+    h = hashlib.new(algo)
+    hashes: dict[str, str] = {}
+
+    def hash_named_tensor(name: str, obj) -> None:
+        if isinstance(obj, torch.Tensor):
+            if isinstance(obj, DTensor):
+                t = obj.to_local().cpu().contiguous()
+            else:
```
wconstab: just curious, how long does this take for a model like 8B or 70B? (or what size models have you been using it on)?

SherlockNoMad: I have been using debug configs, which is tiny model, and the test is <20s. 
I have previously uses the same hash for internal T2 models, perf doesn't seems to be an issue.

.ci/docker/requirements.txt
```diff
@@ -8,3 +8,5 @@ tokenizers >= 0.15.0
 safetensors
 einops
 pillow
+numpy < 2
+pytest
```
tianyu-l: could you put this in https://github.com/pytorch/torchtitan/blob/main/.ci/docker/requirements-dev.txt

also, why numpy has to <2?

torchtitan/tools/utils.py
```diff
@@ -168,6 +168,116 @@ def get_peak_flops(device_name: str) -> float:
...
+                t = obj.cpu().contiguous()
+            if per_tensor:
+                tensor_hash = hashlib.new(algo)
+                tensor_hash.update(t.numpy().tobytes())
```
tianyu-l: just curious -- do we have to depend on numpy to get bytes? Ideally we don't need to introduce numpy dependency if we can workaround it here.

torchtitan/tools/utils.py
```diff
@@ -168,6 +168,116 @@ def get_peak_flops(device_name: str) -> float:
         return 312e12
 
 
+def _hash_model_impl(
```
tianyu-l: since these functions are only used in testing, can we put them in [repo_root]/test/utils.py for now?

.github/workflows/integration_test_8gpu_graph_trainer.yaml
```diff
@@ -76,4 +76,7 @@ jobs:
         # TODO: Currently failing, re-enable once fixed.
         # pytest torchtitan/experiments/graph_trainer/tests/test_precompile.py -v
 
+        # Run bitwise deterministic guardrail test
+        pytest torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py -v
```
tianyu-l, line 80: just so I understand correctly, this is running identical test on all 8 ranks.

SherlockNoMad: this is still a single gpu test atm. 
We can make it a distribute test to cover deterministic of distributed init/ops. 
If we don't have such coverage in titan core, we should add one.

torchtitan/tools/utils.py
SherlockNoMad: done.

torchtitan/tools/utils.py
SherlockNoMad: alternative is to use tensor.untyped_storage(), but call is sinificantly slower than the numpy().tobytes() path. 

Added a todo to investigate.

SherlockNoMad: looks like this is a known issue

https://github.com/pytorch/pytorch/issues/108565#issuecomment-3067330004

shows benchmark

.ci/docker/requirements.txt
SherlockNoMad: done.

PR #2803, opened 2026-04-03
[RL] Add graph_trainer AOT compilation support for PolicyTrainer

Lucaskabela:
## Summary

We want to make torchtitan RL compatible with graph_trainer - this would enable:

1) Flexible traceability options(AOT, make_fx, etc)
2) Out of the box support for cudagraphs, compile-on-one-rank, and more (that we were already hoping to support in torchtitan RL, unifying these efforts)

To do so, we need to:
1) Fix wrapper_fn in graph_utils.py to use spec-based param/buffer lookups instead of model.parameters()/model.buffers(), which deduplicate weight-tied parameters (e.g. Qwen3's tok_embeddings.weight / output.weight). The AOT joint graph expects both names in params_spec, so the old approach caused an index shift and DTensor assertion failures like:
```
 AssertionError: expected DTensor, got Tensor
```

2) Wire up GraphTrainerCompileConfig in PolicyTrainer so the RL trainer can opt into graph_trainer's AOT compilation path. When enabled, per-layer compile in parallelize_fn is skipped (apply_compile handles the whole model) and as we instead use apply_compile from graph_trainer

NOTE: We are not yet enabling by default as this requires more testing

## Test plan
```bash
```

Output:
```
[actor=<root>] Eval: Accuracy=70% (14/20) Format=95% (19/20)
[actor=<root>] Post-training: Accuracy=70% (14/20) Format=95% (19/20)
```
So convergence to same performance
Authored with Claude.


torchtitan/experiments/rl/config_registry.py
```diff
@@ -43,6 +44,9 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
                 tensor_parallel_degree=2,
             ),
             compile=CompileConfig(enable=True, backend="aot_eager"),
+            graph_trainer_compile=GraphTrainerCompileConfig(
```
Lucaskabela: Here for testing for now - should probably comment this out but wanted to prove it works!

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -28,6 +28,7 @@
     TrainingConfig,
 )
 from torchtitan.distributed import ParallelDims, utils as dist_utils
+from torchtitan.experiments.graph_trainer.configs import GraphTrainerCompileConfig
```
wwwjn, line 31: What's the purpose of integrating with GraphTrainer? From the PR description mainly 2 benefits:
1. Flexible traceability options(AOT, make_fx, etc)
2. support for Trainer's cudagraphs, other graph trainer tools

It's a good exploration, but I'm hesitate should RL depend on graph_trainer, which means we can not enable model post-training unless graph_trainer supports it. I would prefer to not depending on graph trainer until they graduate to core torchtitan, so the 2 workstreams won't block each other

To enable CUDAgraph in trainer, can we directly working on enabling it in core torchtitan? How many overlap does it have with graph-trainer

Lucaskabela: I think these two are fairly significant though - they let us have access to a number of desirable techniques out of the box (cudagraphs is most immedaite, but also we get compile on one rank/precompile work, AC, etc).

In regards to the dependency, all changes in this PR in rl are optional and off by default- there are no hard dependencies at all, so I think for the purpose of unblocking exploration and prototyping this PR is incredibly beneficial; there are no blocking risks here 

> To enable CUDAgraph in trainer, can we directly working on enabling it in core torchtitan? How many overlap does it have with graph-trainer

graph-trainer already has this enabled as an fx-graph pass today; if we want to enable this we will likely be reimplementing portions of that work

tianyu-l: Eventually we should make RL and graph trainer work together, although IMO it's not urgent compared with other works.

Graph trainer, like eager main trainer, is an upper stream component, which is still going through active development (and even path finding) and may end up being very different from the main trainer, in terms of script structure. E.g.
- precompile requires first running a precompile script, then running a different training script
- tracing requires putting loss and model into the same compile region

This makes integration point with RL not 100% clear (at least to me). I could even imagine that eventually we may still end up having two different RL training workflow / scripts, one for eager and one for graph trainer.

tianyu-l: Speaking of cudagraph enablement
- If we still do per-layer compilation in main trainer, is there no hope that JIT torch.compile can help with CUDA graph?
- btw some people are working on getting full graph from FSDP
    - Yidi is working on tracing existing FSDP (minimal viable tracing)
    - @weifengpy is thinking of rewriting FSDP logic so that maybe even existing torch.compile can trace through (this is not urgent priority though)

Lucaskabela: Re:

* We can likely still enable cudagraph on main trainer - I can prioritize this and put this work on the back burner for the time being.  
* It is worth keeping this PR as reference though so if we want to know the required changes for enabling graph trainer we can reference here

Going to put this PR on draft mode for a bit

PR #2804, opened 2026-04-03
[graph_trainer] Add SAC correctness and peak memory tests for all decoder models

tugsbayasgalan:
* #2875
* __->__ #2804
* #2766

Add TestGraphBasedSAC tests covering llama3, qwen3, deepseek_v3, llama4, and gpt_oss. Each test compares eager+AC against traced+graph SAC, asserting bitwise-identical loss and peak memory within 10%.

For flex_attention models (llama4, gpt_oss), regional_inductor is applied to the traced graph before SAC because eager uses torch.compile(flex_attention) internally, so the traced graph needs compilation to match numerics.

Authored with Claude.


tugsbayasgalan:
@claude review this PR

PR #2816, opened 2026-04-03
[graph_trainer] Add fsdp2_sac activation checkpointing mode

IvanKobzarev:
The problem:
Using SimpleFSDP/graph_trainer AC modes do not give equivalent AC strategy vs FSDP2 (baseline).

Solution
Add a new `fsdp2_sac` AC mode to graph_trainer that closely matches FSDP2's eager per-op selective activation checkpointing policy.
This allows to have clean comparison of performance.

Key differences from the existing `selective` (apply_sac) graph pass:
- MUST_SAVE / MUST_RECOMPUTE instead of PREFER — removes partitioner flexibility to ensure deterministic save/recompute decisions.
- Per-AC-region mm counters — each transformer block resets its own counter, matching FSDP2's per-module checkpoint_wrapper behavior.
- mm_ops includes aten.linear — some backends keep linear as a leaf op.
- CUDA→CPU _to_copy always saved — for MoE D2H sync metadata.
- Only annotates nodes inside AC regions — embedding, output, and loss nodes are left for the partitioner to decide.

### DIFF
--- a/torchtitan/config/configs.py
+++ b/torchtitan/config/configs.py
@@ -231,7 +231,7 @@ class ParallelismConfig:
 @dataclass(kw_only=True, slots=True)
 class ActivationCheckpointConfig:
     mode: Literal["selective", "full", "memory_budget", "none"] = "selective"
-    """Type of activation checkpointing to use"""
+    """Type of activation checkpointing to use."""
 
     per_op_sac_force_recompute_mm_shapes_by_fqns: list[str] = field(
         default_factory=lambda: ["moe.router.gate"]
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -703,6 +703,48 @@ def _make_eager_memory_policy(save_ops: set | None = None) -> Callable:
     return policy_fn
 
 
+def _propagate_recompute_tag(node: torch.fx.Node) -> bool:
+    """Propagate recompute/ac_graph_id from parent for getitem and wait_tensor.
+
+    getitem: When a node returns a tuple/list (e.g., rmsnorm, sdpa), it is
+    followed by getitem nodes that extract individual elements. They inherit
+    the parent's recompute tag, otherwise they will be exposed as graph outputs
+    and saved for backwards unnecessarily.
+
+    wait_tensor: Semantically tied to its parent async collective (e.g.,
+    reduce_scatter_tensor, all_gather_into_tensor) and must share the same
+    save/recompute decision.
+
+    Returns True if the node was handled (caller should ``continue``).
+    """
+    if node.target not in (
+        operator.getitem,
+        torch.ops._c10d_functional.wait_tensor.default,
+    ):
+        return False
+    parent = node.args[0]
+    if isinstance(parent, torch.fx.Node) and "recompute" in parent.meta:
+        node.meta["recompute"] = parent.meta["recompute"]
+        node.meta["ac_graph_id"] = parent.meta.get("ac_graph_id", 0)
+    return True
+
+
+def _log_ac_region_stats(
+    pass_name: str,
+    ac_region_stats: dict[int, dict[str, int]],
+    save_label: str,
+    recompute_label: str,
+) -> None:
+    """Log per-AC-region annotation statistics."""
+    for ac_region_id in sorted(ac_region_stats):
+        stats = ac_region_stats[ac_region_id]
+        logger.info(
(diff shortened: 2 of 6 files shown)


torchtitan/config/configs.py
```diff
@@ -288,8 +288,17 @@ def __post_init__(self):
...
 @dataclass(kw_only=True, slots=True)
 class ActivationCheckpointConfig:
-    mode: Literal["selective", "full", "memory_budget", "none"] = "selective"
-    """Type of activation checkpointing to use"""
+    mode: Literal[
+        "selective", "full", "memory_budget", "fsdp2_sac", "none"
+    ] = "selective"
+    """Type of activation checkpointing to use.
+
+    "fsdp2_sac": graph_trainer mode that closely matches FSDP2's eager per-op
```
aditvenk: If this is a graph-trainer specific option, can we follow similar pattern to GraphTrainerCompileConfig ? e.g., GraphTrainerActivationCheckpointConfig

IvanKobzarev: Yes, we can extend it.

torchtitan/experiments/graph_trainer/llama3/config_registry.py
```diff
@@ -48,3 +48,28 @@ def graph_trainer_llama3_405b() -> GraphTrainer.Config:
...
+def graph_trainer_llama3_8b_lc_force() -> GraphTrainer.Config:
+    config = to_graph_trainer_config(llama3_8b(), model_registry)
+    config.compile = GraphTrainerCompileConfig(
+        enable=True,
+        mode="jit",
+        backend="inductor",
+        passes=["auto_bucketing"],
+        low_contention_collectives=True,
+        low_contention_allow_nvlink_contention=True,
+    )
```
SherlockNoMad: we are not doing jit mode anymore.

torchtitan/config/configs.py
```diff
@@ -288,8 +288,17 @@ def __post_init__(self):
 
 @dataclass(kw_only=True, slots=True)
 class ActivationCheckpointConfig:
```
SherlockNoMad, line 290: GraphTrainer is currently ignoring this class, please use memory_policy instead. 

Latter we can consider unifying them, but graphtrainer has a independent config system for sac+cpu_offload for now.

torchtitan/experiments/graph_trainer/jit_backend.py
```diff
@@ -87,6 +87,14 @@ def inductor_autobucketing_reordering_pass(
...
                 inductor_autobucketing_reordering_pass
             )
+
+            if compile_config.low_contention_collectives:
+                dist_opts.enable_overlap_scheduling = True
+                dist_opts.enable_low_contention_collectives = True
+                dist_opts.low_contention_min_bytes_per_rank = 0
+                dist_opts.low_contention_allow_nvlink_contention = (
+                    compile_config.low_contention_allow_nvlink_contention
+                )
```
SherlockNoMad, line 97: this is jit backend, which is not used now.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -234,13 +261,105 @@ def apply_sac_pass(
...
+
+        if _propagate_recompute_tag(node):
+            continue
+
+        custom_meta = node.meta.get("custom", {})
+        # Only annotate nodes inside AC regions (transformer blocks).
+        if _AC_REGION_ID not in custom_meta:
+            continue
+        has_ac_regions = True
+        ac_region_id = custom_meta[_AC_REGION_ID]
```
SherlockNoMad, line 316: ac_region_id is deprecated globally.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -234,13 +261,105 @@ def apply_sac_pass(
...
-        stats = ac_region_stats[ac_region_id]
-        logger.info(
-            f"  AC region {ac_region_id}: "
-            f"{stats['save']} nodes annotated with MUST_SAVE, "
-            f"{stats['recompute']} nodes annotated with PREFER_RECOMPUTE"
+    _log_ac_region_stats("apply_sac", ac_region_stats, "MUST_SAVE", "PREFER_RECOMPUTE")
+    return gm
+
+
+def fsdp2_sac_pass(
```
SherlockNoMad, line 268: this seems like memory_policy="eager", we can redirect eager policy to this

SherlockNoMad (changes requested):

PR #2823, opened 2026-04-03
[RL] Fix recompiles (From 10+ to only 1)

Lucaskabela:
## Summary


We make the following changes:

*torchtitan/models/common/attention.py*

 - VarlenMetadata: max_q/max_k changed from int → 0-dim torch.Tensor. Prevents dynamo from specializing on the exact value so per-rollout seq-length changes don't recompile.
  - VarlenAttention.forward: .item() the tensors to yield unbacked SymInts (works because capture_scalar_outputs=True is set in apply_compile).
  - create_varlen_metadata_for_document: stops calling .item() on the d2h-synced max; returns a 0-dim tensor.
  - LocalMapInnerAttention.init_local_map: new method to eagerly initialize _local_map_fn so dynamo never sees the None → callable transition.

*torchtitan/experiments/rl/models/parallelize.py*

  - _make_dynamic_seq_len_forward: overrides model.forward to mark_dynamic(h, 1) and mark_dynamic(positions, 1) before each compiled block call. Keeps the mark outside the compiled region (required since mark_dynamic is                
  forbid_in_graph).
  - parallelize_qwen3: after TP is applied but before apply_compile, calls init_local_map on each block's inner attention with placements derived from the parallelized wq weight.

*torchtitan/experiments/rl/actors/trainer.py*

  - Removed the old cache_size_limit = 16 and recompile_limit = 16 bandaids.
  - _build_model gains an isolate_code_objects flag; ref model gets its own per-block forward code objects so its cache doesn't fight the policy's under recompile_limit=1.
  - After step 0, locks recompile_limit = 1 as a regression guard (one expected recompile: the AsyncCollectiveTensor vs plain-tensor split between layer 0 and layers 1+).
                                                                                                                                                                                                                                           *torchtitan/experiments/rl/actors/utils.py*

  - create_varlen_metadata: wraps max(seq_lens) as a 0-dim int32 tensor on device (matches new VarlenMetadata contract).                                                                                                                   
   
*torchtitan/experiments/rl/tests/test_bitwise_parity.py*

  - _build_padded_varlen_metadata: same wrapping for the test helper.                                                                                                                                                                      
   
## Net effect                                                                                                                                                                                                                               

From 9+ recompiles (grad-mode splits + AsyncCollectiveTensor + lazy _local_map_fn + seq-len shape + varying max_k) down to 1 expected recompile (the AsyncCollectiveTensor between layer-0 and layer-1+), locked down after step 0. 

## Testplan

```bash
```

Results in 
```
[__recompiles]     - 1/0: ___check_type_id(x._local_tensor, 140561999080752), type=<class 'torch.distributed._functional_collectives.AsyncCollectiveTensor'>  # return inner()  # nn/modules/module.py:1881 in _call_impl (HINT: type AsyncCollectiveTensor)
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     - User stack trace:
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     -   File "/home/lucaskabela/torchtitan/torchtitan/models/qwen3/model.py", line 65, in forward
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     -     self.attention_norm(x), freqs_cis, attention_masks, positions
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     -   File "/home/lucaskabela/pytorch/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     -     return self._call_impl(*args, **kwargs)
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     -   File "/home/lucaskabela/pytorch/torch/nn/modules/module.py", line 1881, in _call_impl
[rank0]:V0403 13:52:56.045000 788326 /home/lucaskabela/pytorch/torch/_dynamo/guards.py:5142] [1/1] [__recompiles]     -     return inner()
...
[actor=<root>] Step  9 | Loss: +0.9693 | Reward: +0.350 | Correct: 27/40 | Avg tokens: 100 | Logprob diff: mean=-1.0396e-01, max=1.0704e+01 | Time: 16.3s
[actor=<root>] Eval: Accuracy=65% (13/20) Format=90% (18/20)
[actor=<root>] Post-training: Accuracy=65% (13/20) Format=90% (18/20)
```

TIMING RESULTS:

| | Baseline | This PR | Delta |
  |---|---|---|---|
  | **Step 0 (warmup)** | 35.3s | 28.7s | **-6.6s (-19%)** |                                                                                                                                   
  | — ref | 11.8s | 8.3s | -3.5s |
  | — policy | 17.7s | 14.7s | -3.0s |                                                                                                                                                         
  | — backward | 2.7s | 2.6s | -0.1s |                            
  | **Avg step 1–9** | 9.3s | 9.2s | ~same |                                                                                                                                                   
  | **E2E** | 119.1s | 112.0s | **-7.1s (-6%)** |

### DIFF
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -89,13 +89,6 @@ class PolicyTrainer(Actor, Configurable):
         requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
         self._transfer_dtype = requested if requested != training_dtype else None
 
-        # The policy and ref models share code objects, so dynamo's
-        # per-code-object cache must hold entries for both grad modes
-        # (grad for policy, no_grad for ref). The default limit of 8
-        # is not enough; 16 accommodates both without recompile storms.
-        # TODO: @Lucaskabela fix recompiles in general as these increase startup
-        torch._dynamo.config.cache_size_limit = 16
-
         # Device setup
         device_module, device_type = utils.device_module, utils.device_type
         self.device = torch.device(f"{device_type}:{int(os.environ['LOCAL_RANK'])}")
@@ -134,7 +127,11 @@ class PolicyTrainer(Actor, Configurable):
         # TODO: @joecummings remove ref entirely, this is hacky and we don't need it
         if getattr(config.loss, "kl_coef", 0) > 0:
             ref_model = self._build_model(
-                model_spec, config, device_type, hf_assets_path
+                model_spec,
+                config,
+                device_type,
+                hf_assets_path,
+                isolate_code_objects=True,
             )
             ref_model.eval()
             ref_model.requires_grad_(False)
@@ -207,6 +204,7 @@ class PolicyTrainer(Actor, Configurable):
         config: Config,
         device_type: str,
         hf_assets_path: str,
+        isolate_code_objects: bool = False,
     ):
         """Build, parallelize, and initialize a model from checkpoint.
         Will be used to build trainer's policy model and reference model.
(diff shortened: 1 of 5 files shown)


Lucaskabela:
PR is still not land ready so I will leave as draft, but it is okay if people TAL since this is a step in the right direction

Lucaskabela:
NOTE: https://github.com/vllm-project/vllm/pull/38933 is a fix on vLLM side for one of these recompiles we observe in this RL use case

torchtitan/models/common/attention.py
```diff
@@ -124,18 +160,24 @@ def __call__(
                 f"All Shard placements must shard on the same dim, "
                 f"but got dims {shard_dims}"
             )
-            # return_lse=True (e.g. gpt_oss attention sinks) produces
```
Lucaskabela, line 127: Just move this into where it is actually needed

torchtitan/models/common/attention.py
```diff
@@ -91,6 +97,36 @@ def __init__(self, config: Config) -> None:
         super().__init__()
         self._local_map_fn: Callable | None = None
 
+    def init_local_map(
```
Lucaskabela, line 100: Note: This is used in parallelize to avoid the lazy init before compile

Lucaskabela:
@claude review

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -232,6 +230,21 @@ def _build_model(
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
 
+        if isolate_code_objects:
```
Lucaskabela, line 233: Wondering if there is a better way to do/isolate this (trying to isolate the code object from ref_model for KL divergence and the trainer model (difference is running under inference mode or not)

torchtitan/models/common/attention.py
```diff
@@ -60,12 +61,17 @@ class VarlenMetadata(NamedTuple):
...
+    are converted to unbacked SymInts via ``.item()`` before being passed to
+    the varlen attention op, avoiding per-seq-length recompiles in RL-style
+    workloads where the longest sequence varies each step.
     """
 
     cu_seq_q: torch.Tensor
     cu_seq_k: torch.Tensor
-    max_q: int
-    max_k: int
+    max_q: torch.Tensor
```
Lucaskabela, line 73: Validated no dowsntream usages rn so this change is safe

Lucaskabela:
Closing this PR as we have decided that this particular change needs to wait until our codebase is more stable

PR #2825, opened 2026-04-03
Integration tests always run all tests instead of exiting on the first failure

yiming0416:
- Integration tests now always run all tests to completion instead of exiting on the first failure
  - Failed tests are logged immediately and a summary of all failures with error messages is reported at the end
  - The process still exits with a non-zero code when any test fails

### DIFF
--- a/tests/integration_tests/run_tests.py
+++ b/tests/integration_tests/run_tests.py
@@ -25,7 +25,7 @@ _TEST_SUITES_FUNCTION = {
 
 
 def _run_cmd(cmd):
-    return subprocess.run([cmd], text=True, shell=True)
+    return subprocess.run([cmd], text=True, shell=True, capture_output=True)
 
 
 def run_single_test(
@@ -68,10 +68,13 @@ def run_single_test(
             )
 
         result = _run_cmd(cmd)
-        logger.info(result.stdout)
+        if result.stdout:
+            logger.info(result.stdout)
         if result.returncode != 0:
-            raise Exception(
-                f"Integration test failed, flavor : {test_flavor.test_descr}, command : {cmd}"
+            raise RuntimeError(
+                f"\nFailed test flavor: {test_flavor.test_descr}.\n"
+                f"Command: {cmd}\n"
+                f"stderr: {result.stderr}\n"
             )
 
 
@@ -83,6 +86,7 @@ def run_tests(args, test_list: list[OverrideDefinitions], module=None, config=No
         exclude_set = {name.strip() for name in args.exclude.split(",")}
 
     ran_any_test = False
+    failed_tests: list[tuple[str, str]] = []
     for test_flavor in test_list:
         # Filter by test_name if specified
         if args.test_name != "all" and test_flavor.test_name != args.test_name:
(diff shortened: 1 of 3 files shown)


tests/integration_tests/run_tests.py
```diff
@@ -105,9 +106,21 @@ def run_tests(args, test_list: list[OverrideDefinitions], module=None, config=No
...
-            run_single_test(test_flavor, args.output_dir, module, config)
+            try:
+                run_single_test(test_flavor, args.output_dir, module, config)
+            except Exception as e:
+                logger.error(str(e))
+                failed_tests.append((test_flavor.test_name, str(e)))
             ran_any_test = True
 
+    if failed_tests:
+        failure_summary = "\n".join(
```
tianyu-l, line 120: I'd be interested in looking at examples of failures. Not sure if this aggregated error message would make it harder to locate the original command / error message.

tests/integration_tests/run_tests.py
tianyu-l, line 1: you might have to do the same for https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/graph_trainer/tests/run_precompile_tests.py, as they are separate scripts for now.

tianyu-l (approved):
One downside is that we have to wait until all tests to finish before we can get signal.

tests/integration_tests/run_tests.py
fegin: I have the same concern. I think it is possible to capture the stderr/stdout from _run_cmd and output them together. But overall this is a nice improvement compared the original test infra.

yiming0416: This PR should provide an example of failures in graph_trainer CI, as the fixing PR is not landed.

will wait and see the message format

yiming0416: I adopted @fegin's suggestion and set `capture_output=True` in `_run_cmd`, so stderr will be displayed as the error message.

Example failure:
```
2026-04-03T23:46:43.7230186Z RuntimeError: 1 integration test(s) failed:
2026-04-03T23:46:43.7230375Z   aot_llama3_fsdp_tp_flexattn_autobucketing_regional_inductor: 
2026-04-03T23:46:43.7230615Z Failed test flavor: AOT llama3 FSDP+TP+FlexAttn autobucketing regional_inductor.
2026-04-03T23:46:43.7233295Z Command: TORCH_TRACE="/home/ec2-user/actions-runner/_work/_temp/artifacts-to-be-uploaded/aot_llama3_fsdp_tp_flexattn_autobucketing_regional_inductor/compile_trace" NGPU=8 LOG_RANK=0,1,2,3,4,5,6,7 ./run_train.sh --dump_folder /home/ec2-user/actions-runner/_work/_temp/artifacts-to-be-uploaded/aot_llama3_fsdp_tp_flexattn_autobucketing_regional_inductor --module graph_trainer.llama3 --config graph_trainer_llama3_debugmodel_flex_attn --compile.mode aot --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 2 --compile.passes auto_bucketing,regional_inductor
2026-04-03T23:46:43.7233389Z stderr: .....(full error message from stderr)
```

PR #2833, opened 2026-04-03
[rl] [1/5] [dnr] Add batched RL training materials

joecummings:
* #2837
* #2836
* #2835
* #2834
* __->__ #2833

### DIFF
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -6,6 +6,7 @@
 
 import logging
 import os
+from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/rl/types.py
```diff
@@ -43,3 +45,15 @@ class Episode:
     reward: float = 0.0
     group_id: str = ""
     advantage: float = 0.0
+
+
+@dataclass
+class TrainBatch:
```
joecummings, line 51: Basic container for training. This is NOT meant to be the final form - can evolve over time, but I want to start adding some structure.

PR #2841, opened 2026-04-04
[graph_trainer] Add apply_default_graph_passes entry point and propagate stac…

SherlockNoMad:
Add `apply_default_graph_passes` in passes.py as the entry point for graph-level optimizations on the traced fwd+bwd graph, called by GraphTrainer after trace_module(). Currently applies tlparse graph logging.

Also propagate stack_trace metadata from forward nodes to backward nodes in `_copy_fwd_metadata_to_bw_nodes`, alongside the existing custom and nn_module_stack propagation, with a corresponding test.

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -40,6 +40,19 @@ from torchtitan.experiments.graph_trainer.reshard_after_forward import (
 from torchtitan.tools.logging import logger
 
 
+def apply_default_graph_passes(
+    gm: torch.fx.GraphModule, example_inputs: tuple
+) -> torch.fx.GraphModule:
+    """Entry point for optimizing the traced fwd+bwd graph.
+
+    Called by GraphTrainer after tracing to apply graph-level optimization
+    passes before execution. Individual passes are defined below.
+    """
+    gm = tlparse_log_graph_pass(gm, example_inputs, graph_name="make_fx_graph_traced")
+
+    return gm
+
+
 def autobucketing_reordering_pass(
     gm: torch.fx.GraphModule, example_inputs=None
 ) -> torch.fx.GraphModule:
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -40,6 +40,19 @@
 from torchtitan.tools.logging import logger
 
 
+def apply_graph_passes(
+    gm: torch.fx.GraphModule, example_inputs: tuple
+) -> torch.fx.GraphModule:
+    """Entry point for optimizing the traced fwd+bwd graph.
+
+    Called by GraphTrainer after tracing to apply graph-level optimization
```
yiming0416, line 48: Are we going to always turn on all available passes here? Or give users options to choose which passes they want to opt in?

xmfan, line 48: how does this interact with other pass ordering? like inductor pass which will not return a gm

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -109,6 +110,11 @@ def _make_fx_forward_backward_step(
                     (inputs, labels, global_valid_tokens, extra_inputs, extra_kwargs),
                 )
 
+            self._traced_step.gm = apply_graph_passes(
+                self._traced_step.gm,
+                self._traced_step.example_inputs,
+            )
```
xmfan: why is this not parametrized via --compile.passes like all the other ones?

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: This is the central pass where we apply passes. 

we will standardize pass signatures as （gm, example_inputs, *, kwargs） -> gm, so that we can apply passes in a loop.

ideally, passes should be idempotent. but in practice, order matters, and we need to control that.

SherlockNoMad: Yes, this is our curated passes. we demonstrate application of all passes here. 
Let's worry about customization latter.

torchtitan/experiments/graph_trainer/trainer.py
SherlockNoMad: This is a set of curated passes we would demonstrate. 
Think of it as the default passes to apply. Later we can support customization via  --compile.passes

PR #2842, opened 2026-04-04
[MoE][1/n] Introduce token dispatcher and replace token reorderer

acisseJZhong:
* #2960
* __->__ #2842

TODO tracking here https://github.com/pytorch/torchtitan/issues/2956
## Summary

Introduce a token dispatcher abstraction (`token_dispatcher.py`) that encapsulates the full token routing lifecycle — dispatch (reorder + optional EP all-to-all) and combine (reverse dispatch + shared_experts overlap + score application + scatter_add).

**Dispatcher hierarchy:**
- `LocalTokenDispatcher` — EP=1, local token reordering only. Base class for all dispatchers.
- `AllToAllTokenDispatcher` — EP>1, adds NCCL all-to-all dispatch/combine. Calls `super().combine()` for shared score+scatter_add logic. Supports sequence parallel via `sp_size` config.
- `TorchAOTokenDispatcher` — EP>1, extends AllToAllTokenDispatcher with `permute_and_pad` for FP8/MXFP8 quantized grouped GEMMs.
- `DeepEPTokenDispatcher` — EP>1, uses DeepEP/HybridEP library kernels instead of standard all-to-all.

**Key changes:**
- `GroupedExperts.forward()` now owns the full dispatch → expert compute → combine pipeline, accepting `shared_experts` to overlap with async combine communication.
- `MoE.forward()` simplified to: router → `self.experts(x, top_scores, indices, shared_experts=self.shared_experts)` → reshape.
- `DeepEPMoE` no longer overrides `forward()` — the dispatcher handles both NCCL and DeepEP paths.
- Removed `TokenReorderer` (folded into dispatchers) and `moe_deepep.py` (folded into dispatchers + simplified `DeepEPMoE`).
- `ReordererSequenceParallel` input/output hooks replaced by `sp_size` config on `AllToAllTokenDispatcher` — sequence parallel token splitting now happens inside the dispatcher's `dispatch()` and `combine()`.
- `ExpertParallel` / `ExpertTensorParallel` now only handle weight sharding + setting `ep_group` on the dispatcher.
- Removed `score_before_experts` from `MoE.Config` (now lives on the dispatcher).

## Motivation

**Previously,** token routing was split across three places:
1. `TokenReorderer` — reorder tokens from token-index to expert-index order
2. `MoE.forward()` — score application, shared_experts, scatter_add
3. `ExpertParallel` input/output hooks — all-to-all dispatch/combine (EP>1 only)

This made it hard to reason about the async overlap pattern (shared_experts should run while combine communication is in flight), and `DeepEPMoE` had to duplicate most of `MoE.forward()`.

**Now,** the dispatcher encapsulates the full lifecycle. Benefits:
- Async overlap is explicit: `combine()` starts async communication, runs `shared_experts`, then forces sync via scatter_add or `sync_combine()`.
- Single place to swap dispatch/combine implementations (NCCL, DeepEP, HybridEP, TorchAO).
- `AllToAllTokenDispatcher.combine()` reuses `LocalTokenDispatcher.combine()` via `super()`, eliminating code duplication.
- `DeepEPMoE.forward()` is no longer needed — the base `MoE.forward()` works for all paths.
- Sequence parallel logic consolidated into `AllToAllTokenDispatcher` — no more `ReordererSequenceParallel` class with `distribute_module` input/output hooks. The dispatcher splits tokens in `dispatch()` and `combine()` when `sp_size > 1`.
- Enables `local_map` usage: single wrapper around `GroupedExperts` (which contains the dispatcher).

## Test Plan
Losses matched under EP=1, EP>1, `--parallelism.expert_parallel_comm_backend=deepep`, and torchao cases.
```
NCCL_NVLS_ENABLE=0 NGPU=8 ./run_train.sh     --module llama4     --config llama4_debugmodel     --parallelism.pipeline_parallel_degree 1     --parallelism.data_parallel_shard_degree 4     --parallelism.tensor_parallel_degree 2     --parallelism.expert_parallel_degree 2     --parallelism.expert_tensor_parallel_degree 2      --metrics.enable_wandb --training.steps 100
```

```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 1 --parallelism.expert_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 1" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 1 --parallelism.expert_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 1" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```

cc @fegin for parallelism refactor, @elfiegg @shuhuayu for removing DeepEP MoE.

### DIFF
--- a/torchtitan/models/common/token_dispatcher.py
+++ b/torchtitan/models/common/token_dispatcher.py
@@ -408,6 +408,10 @@ class AllToAllTokenDispatcher(LocalTokenDispatcher):
             routed_output, metadata.input_shape, metadata.permuted_indices
         )
 
+        assert self.ep_group is not None, (
+            "ep_group must be set before dispatch. "
+            "ExpertParallel._partition_fn() should set it."
+        )
         # All-to-all combine: returns AsyncCollectiveTensor — the a2a runs
         # on the NCCL stream and won't block until the tensor is accessed.
         routed_output = all_to_all_single_autograd(


torchtitan/models/llama4/parallelize.py
```diff
@@ -552,11 +550,13 @@ def apply_moe_ep_tp(
                 ),
             }
             if ep_mesh is not None and etp_mesh is None:
-                # If TP is borrowed for EP, then split the tokens across TP ranks so that
```
acisseJZhong, line 568: will work on supporting this probably in the next PR as we want shared experts to be in Dtensor region. Only Experts and dispatcher should be in local map.

torchtitan/distributed/expert_parallel.py
```diff
@@ -340,127 +238,6 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
...
-        return distribute_module(
-            module,
-            device_mesh,
-            partition_fn=DeepEPExpertParallel._partition_fn,
-            input_fn=self._token_dispatch,  # pyrefly: ignore [bad-argument-type]
-            output_fn=self._token_combine,  # pyrefly: ignore [bad-argument-type]
-        )
-
-
 class TorchAOExpertParallel(ExpertParallel):
```
tianyu-l: I think you need a dispatcher for this class too, which does "all-to-all + permute_and_pad" instead of the default "all-to-all + permute".

torchtitan/models/common/config_utils.py
```diff
@@ -148,4 +151,9 @@ def make_experts_config(
         num_experts=num_experts,
         use_grouped_mm=use_grouped_mm,
         param_init=param_init,
+        token_dispatcher=BaseTokenDispatcher.Config(
```
tianyu-l: doesn't really make sense to make a abstract base class config

torchtitan/models/common/moe_deepep.py
tianyu-l: side question: do we not need `scatter_add` here?

torchtitan/models/gpt_oss/moe.py
```diff
@@ -149,11 +150,14 @@ def __init__(self, config: Config):
         )  # (num_experts, out_dim, in_dim)
         self.mlp2_bias = nn.Parameter(torch.empty((num_experts, dim)))
 
-    def forward(
+        self.token_dispatcher = LocalTokenDispatcher(config.token_dispatcher)
```
tianyu-l: use config.build instead of calling the class constructor directly?

torchtitan/models/common/moe.py
```diff
@@ -91,12 +93,14 @@ def __init__(self, config: Config):
             torch.empty(config.num_experts, config.hidden_dim, config.dim)
         )
         self.use_grouped_mm = config.use_grouped_mm
+        self.token_dispatcher = LocalTokenDispatcher(config.token_dispatcher)
```
tianyu-l: same: use config.build, which means you should put LocalTokenDispatcher.Config in the config, not BaseTokenDispatcher.Config

torchtitan/models/llama4/parallelize.py
```diff
@@ -633,3 +627,34 @@ def apply_moe_ep_tp(
...
+                token_dispatcher = DeepEPTokenDispatcher(
+                    td_config,
+                    comm_backend=comm_backend,
+                    hybridep_non_blocking_expert_capacity_factor=hybridep_non_blocking_expert_capacity_factor,
+                    pad_multiple=pad_multiple,
+                )
+            else:
+                from torchtitan.models.common.token_dispatcher import TokenDispatcher
+
+                token_dispatcher = TokenDispatcher(td_config)
```
tianyu-l: Here we need a "big" style change.
- previous we first create the model, then apply parallelism to add dispatch/combine logic
- now all the dispatcher are config objects, we should apply config changes, which probably should happen in `model/__init__.py` or `config_registry.py` rather than `parallelize.py` (which receives a created model)

torchtitan/models/common/moe_deepep.py
acisseJZhong: I moved the scatter add to combine. https://fburl.com/ptbnwyf5

torchtitan/models/common/config_utils.py
acisseJZhong: I was thinking about having `BaseTokenDispatcher` which defines the abstract dispatch/combine functions, and the __init__. `LocalTokenDispatcher`(EP=1),  `TokenDispatcher`(EP>1), `DeepEPTokenDispatcher`(deepep), and maybe the torchao dispatcher just inherits  `BaseTokenDispatcher`. Are you saying this seems uneccessary?

tianyu-l: This is good. I'm saying that when we actually create a model config, it's better to specify a concrete subclass that is runnable, not something that's base class. Strictly speaking, an abstract class (and its config class) should never be instantiated.

torchtitan/distributed/expert_parallel.py
acisseJZhong: done, ended up move the _permute/_unpermute function over from expert_parallel.py as well.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+            ).to(routed_output.dtype)
+
+        return deterministic_scatter_add(
+            outputs,
+            metadata.token_indices_experts_sorted.reshape(-1, 1).expand(-1, dim),
+            routed_output,
+        )
+
+
+class TorchAoTokenDispatcher(TokenDispatcher):
```
tianyu-l:
```suggestion
class TorchAOTokenDispatcher(TokenDispatcher):
```

torchtitan/models/common/moe_deepep.py
```diff
@@ -40,15 +38,12 @@ class Config(MoE.Config):
 
     def __init__(self, config: Config):
         super().__init__(config)
```
tianyu-l: remove

torchtitan/models/common/moe_deepep.py
```diff
@@ -86,15 +81,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        # must run in parallel with the async combine — it hasn't computed yet.
+        outputs = torch.zeros_like(x)
         routed_output = self.experts(
             x,
             num_tokens_per_expert,
-            selected_experts_indices,
             top_scores,
-            self.experts.num_experts,
+            selected_experts_indices,
+            outputs=outputs,
```
tianyu-l: requiring `outputs` seems unnecessary -- what if there's no shared expert at all, e.g. in gpt-oss.

torchtitan/models/deepseek_v3/model.py
```diff
@@ -223,6 +226,27 @@ def update_from_config(
                     layer_cfg.moe.router._debug_force_load_balance = (
                         debug.moe_force_load_balance
                     )
+
+                    # Replace the default LocalTokenDispatcher config with the
```
tianyu-l: can this be done in model registry / config registry? Eventually we should remove `update_from_configs` function, as it violates encapsulation (this function can see all job config, which doesn't make sense)

torchtitan/models/gpt_oss/model.py
```diff
@@ -217,6 +220,25 @@ def update_from_config(
                             "Failed to use grouped mm, which is only supported on SM90 or later",
                         )
                         layer_cfg.moe.experts.use_grouped_mm = False
+                    # Replace the default LocalTokenDispatcher config with the
```
tianyu-l: similar

torchtitan/models/common/moe.py
```diff
@@ -427,54 +388,23 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-        # shared expert
-        # Note: we execute the shared expert before scoring the output of the routed expert
-        # to "implicitly" overlap the shared expert compute with token combine communication
-        out = (
-            self.shared_experts(x)
-            if self.shared_experts is not None
-            else torch.zeros_like(x)
+            top_scores,
+            selected_experts_indices,
+            outputs=torch.zeros_like(x),
```
tianyu-l: why do you need this `outputs`? I think one motivation was that you can send `shared_experts` results to it, but here it doesn't seem you are doing so (for the sake of overlapping).

torchtitan/models/common/moe_deepep.py
```diff
@@ -17,21 +17,19 @@
...
-    3. Experts process tokens
-    4. DeepEP combine starts (async) - returns immediately
-    5. shared_experts runs IN PARALLEL with combine communication
-    6. sync_combine() waits for combine to complete
-    7. Addition of shared_experts output and routed_output
+    2. GroupedExperts.forward() runs the full pipeline via DeepEPTokenDispatcher:
+       a. dispatch — sends tokens to expert-owning ranks (sync)
+       b. expert computation
+       c. combine — starts sending results back (async, returns immediately)
+    3. shared_experts runs IN PARALLEL with the async combine communication
```
tianyu-l: I think for the default moe with NCCL all-to-all, it is also async and can overlap with `shared_experts`. It might be that DeepEP requires extra `sync_combine()`. cc @shuhuayu to confirm.

The current docstring sounds that the default moe doesn't have overlapping.

torchtitan/models/common/moe_deepep.py
tianyu-l: I think we should try to consolidate this file with moe.py. One direction could be, check tokenizer class / comm_backend, if it's deepep then call this `sync_combine`.

torchtitan/models/llama4/__init__.py
```diff
@@ -124,10 +124,17 @@ def _build_llama4_layers(
                 num_experts=num_experts,
                 gate_param_init=_depth_init(layer_id),
             )
+            # Currently create experts with LocalTokenDispatcher (EP=1) by default.
+            # The dispatcher is replaced in update_from_config() which has
+            # access to the training config's EP degree and comm backend.
```
tianyu-l: You'll have access in config_registry.py and we can do the replacement there. Concretely, you can have a config pass `apply_ep` which replaces all local dispatchers to EP dispatchers.

torchtitan/models/llama4/parallelize.py
```diff
@@ -523,7 +519,6 @@ def apply_moe_ep_tp(
     etp_mesh: DeviceMesh | None,
     ep_etp_mesh: DeviceMesh | None,
     comm_backend: str = "standard",
-    hybridep_non_blocking_expert_capacity_factor: float | None = None,
     pad_multiple: int | None = None,
```
tianyu-l: We don't need `comm_backend` and `pad_multiple`. They are only used for logging (not necessary) and erroring (can be done at config place, too).

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+class DispatchMetadata(LocalDispatchMetadata):
+    """Metadata returned by TokenDispatcher.dispatch() for use in combine()."""
+
+    input_shape: tuple  # for _unpermute
+    permuted_indices: torch.Tensor  # for _unpermute
+    input_splits: list[int]
+    output_splits: list[int]
+
+
+class BaseTokenDispatcher:
```
tianyu-l: subclass `Configurable`?

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+    """Abstract base for token dispatchers in MoE.
+
+    A token dispatcher handles the full token routing lifecycle:
+    dispatch (reorder + optional EP all-to-all) and combine (reverse).
+
+    Not an nn.Module — dispatchers have no learnable parameters or buffers.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config:
```
tianyu-l: same

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+                GEMMs (e.g. 16 for FP8, 32 for MXFP8). Only supported with hybridep.
+                None means no padding.
+        """
+
+        comm_backend: str
+        hybridep_non_blocking_expert_capacity_factor: float | None = None
+        pad_multiple: int | None = None
+
+        def build(self) -> "DeepEPTokenDispatcher":
+            return DeepEPTokenDispatcher(self)
```
tianyu-l: no need to implement `build()` if inheriting `Configurable`

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+        # Apply scores before expert computation if configured
+        if self.score_before_experts:
+            routed_input = (
+                routed_input.to(torch.float32)
+                * top_scores_experts_sorted.reshape(-1, 1)
+            ).to(x.dtype)
+
+        metadata = LocalDispatchMetadata(
+            token_indices_experts_sorted=token_indices_experts_sorted,
+            top_scores=top_scores,
```
tianyu-l: not used?

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+class TokenDispatcher(BaseTokenDispatcher):
+    """Token dispatcher for EP>1. Handles token reorder + all-to-all dispatch/combine.
+
+    ep_group is set by the parallelization code after construction.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(BaseTokenDispatcher.Config):
+        def build(self) -> "TokenDispatcher":
+            return TokenDispatcher(self)
```
tianyu-l: same

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+        )
+        return routed_input, num_tokens_per_expert_group, metadata
+
+    def _permute(
+        self, routed_input, num_tokens_per_expert_group, ep_degree, num_local_experts
+    ):
+        """Reorder tokens from rank-major to expert-major layout.
+
+        Input layout:  (e0,r0), (e1,r0), ..., (e0,r1), (e1,r1), ...  (rank-major)
+        Output layout: (e0,r0), (e0,r1), ..., (e1,r0), (e1,r1), ...  (expert-major)
```
tianyu-l, line 335: Let's use this succinct format for the comments in `dispatch()` as well

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+        ) = self._permute(
+            routed_input,
+            num_tokens_per_expert_group,
+            ep_degree,
+            num_local_experts,
+        )
+
+        metadata = DispatchMetadata(
+            token_indices_experts_sorted=token_indices_experts_sorted,
+            top_scores=top_scores,
```
tianyu-l: not used

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+        raise NotImplementedError
+
+
+class LocalTokenDispatcher(BaseTokenDispatcher):
+    """Token dispatcher for EP=1. Handles local token reordering only."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(BaseTokenDispatcher.Config):
+        def build(self) -> "LocalTokenDispatcher":
+            return LocalTokenDispatcher(self)
```
tianyu-l: same

torchtitan/distributed/expert_parallel.py
```diff
@@ -139,101 +64,18 @@ def _partition_fn(self, name: str, mod: nn.Module, device_mesh: DeviceMesh) -> N
...
             partition_fn=self._partition_fn,
-            # pyrefly: ignore [bad-argument-type]
-            input_fn=self._token_dispatch,
-            # pyrefly: ignore [bad-argument-type]
-            output_fn=self._token_combine,
         )
 
 
 # This class is for dp2ep with TP (without TP we can just use ExpertParallel)
 class ExpertTensorParallel(ExpertParallel):
```
tianyu-l, line 78: Does this still work? I'm OK with removing this class, as it's not efficient and in general not useful (no one really applies TP to group experts)

torchtitan/models/llama4/parallelize.py
```diff
@@ -552,11 +547,13 @@ def apply_moe_ep_tp(
...
             if ep_mesh is not None and etp_mesh is None:
-                # If TP is borrowed for EP, then split the tokens across TP ranks so that
-                # the reorderer, the all-to-all comms, and routed experts computation
-                # are effectively running Sequence Parallel (split along the folded bs*slen dim)
-                # pyrefly: ignore [no-matching-overload]
-                moe_layer_plan.update({"moe.reorderer": ReordererSequenceParallel()})
+                # ETP=1: EP borrows from TP, requires ReordererSequenceParallel
+                # to split tokens across TP ranks. Not yet supported with
+                # the new token dispatcher.
+                raise NotImplementedError(
```
tianyu-l: This sounds a big regression. How about we work on it in next PR, but land the two PRs together.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,565 @@
...
+        self.ep_group: dist.ProcessGroup
+
+    def dispatch(
+        self,
+        x: torch.Tensor,
+        top_scores: torch.Tensor,
+        selected_experts_indices: torch.Tensor,
+        num_tokens_per_expert: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, DispatchMetadata]:
+        ep_degree = dist.get_world_size(self.ep_group)
```
tianyu-l: can we put this in config, rather at run time? Not sure how much overhead it has, but must be non-zero, and this code will be run every single layer / iteration.

torchtitan/models/llama4/parallelize.py
```diff
@@ -633,3 +619,13 @@ def apply_moe_ep_tp(
...
+
+        # Set ep_group on the token dispatcher so it can perform
+        # all-to-all communication. The dispatcher type (including
+        # TorchAoTokenDispatcher for pad_multiple) is already selected
+        # at config time in update_from_config.
+        if ep_mesh is not None:
+            # pyrefly: ignore [missing-attribute]
+            transformer_block.moe.experts.token_dispatcher.ep_group = (
+                ep_mesh.get_group()
+            )
```
tianyu-l: This seems hacky ... I don't have a great solution at hand. One option is to use a global state / context manager in the experts / dispatcher code.

For now, how about we plug this logic into `ExpertParallel()`? I.e. we assign the mesh / group to `experts.token_dispatcher` when doing `partition` here https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py#L137

torchtitan/models/common/moe_deepep.py
shuhuayu: Yes. DeepEPMoE's `forward()` needs to call `sync_combine()` explicitly right after `shared_experts` computation, and it will get errors if not.

torchtitan/models/common/moe.py
acisseJZhong: you are right, I was initially doing this
```
        out = self.experts(
            x,
            num_tokens_per_expert,
            top_scores,
            selected_experts_indices,
            outputs=torch.zeros_like(x) if self.shared_experts is None else self.shared_experts(x),
        )
```
if we don't do this, it is equivalent to either allocating output outside token dispatcher combine or inside combine. 

but this is no longer numerics identical to baseline because baseline calculates experts forward first and then add shared expert. Therefore I change it back in this PR, maybe we can optimize in later PR? wdty?

torchtitan/models/common/moe_deepep.py
acisseJZhong: yeah reason is replied in the other thread, seems unnecessary for now, but could be useful later.

If there's no shared expert, we pass in zeros in MoE forward.

torchtitan/models/deepseek_v3/model.py
acisseJZhong: yeah I believe this could be move out to config registry once we know parallelism info at config time(related to @fegin's parallelism config refactor). For deepep and torchao dispatcher, we need to think about how to get `pad_multiple` from config tho. 

added a todo comment for this.

torchtitan/models/common/moe.py
```diff
@@ -427,54 +388,23 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
         )
 
-        if not self.score_before_experts:
-            routed_output = (
-                routed_output.to(torch.float32)
-                * top_scores_experts_sorted.reshape(-1, 1)
-            ).to(x.dtype)
+        # shared expert
+        if self.shared_experts is not None:
+            out = out + self.shared_experts(x)
```
tianyu-l: I feel this might be killing the overlap between `combine` and `shared_experts` computation.

torchtitan/models/common/moe.py
tianyu-l: Oh I think I understand.

- I think it's very important to have the same numerics in this PR, because we are doing a lot of changes.
- It's OK to change the order of dispatcher and shared experts in later PRs. I don't know how we can keep the overlapping between dispatcher and shared experts if we let `outputs=shared_experts(x)`.
- In this PR, let's not send `outputs` to dispatcher -- `combine` can create this zero tensor and it's meaningless to create it here.
- I left another comment on the perf regression this PR might might be causing which I think could be avoided.

torchtitan/models/llama4/__init__.py
acisseJZhong: makes sense, move apply_ep to Trainer Config post init.

torchtitan/distributed/expert_parallel.py
acisseJZhong: yeah ETP is working, and in fact, only ETP is working. Turning TP entirely into EP is not working as this PR since we don't have the reorderer. 

curious why ETP in general is not efficient? I thought dp2ep and keep etp>1 is common?

torchtitan/models/llama4/parallelize.py
acisseJZhong: let me try using moe_layer_plan.update({"moe.experts.token_dispatcher": ReordererSequenceParallel()}) to see if this work.

torchtitan/distributed/expert_parallel.py
tianyu-l: In the current impl, when ETP>1 all TP ranks (per EP rank) are doing the same all-to-all, which is a waste. The optimal impl could shard the communication into 1/TP_degree.

Also, usually experts are thin, and using ETP would make the experts shape too small so that gemm kernels become inefficient.

Can you find a tech report where their main training is using ETP > 1?

torchtitan/models/llama4/parallelize.py
acisseJZhong: yeah sounds good let me note this down and clean up in a separate PR? 
a few todos:
- change tensor shape suffix namings
- simplify routing map generation logic, extracting helper functions
- clean up logging in parallelize.py
- unify moe and deepep moe

torchtitan/models/common/moe.py
acisseJZhong: yes I just realize the NCCL a2a is async, therefore even though we wrote experts forward before shared_experts forward, the combine won't finish until we access the result. And we execute the shared expert before scoring the output of the routed expert, therefore we are implicitly overlapping shared expert and experts comp.

However, with this PR change, I move the score multiply part inside dispatch, so now the async a2a will force to finish before shared expert comp start. This is no longer overlapping. 

Let me think about how to make it overlap again without changing numerics.

torchtitan/models/llama4/parallelize.py
acisseJZhong: ended up introducing `ExpertSequenceParallel` which is almost the same as `ReordererSequenceParallel`(+inheriting the Expert Parallel weight partition), it's used on experts when ETP=1. I've verified loss parity with baseline for both ETP> and =1.

torchtitan/models/common/moe_deepep.py
acisseJZhong: Let me do it in next PR.

torchtitan/models/common/moe.py
acisseJZhong: moved the scatter add outside dispatcher as before. This is a bit ugly/unintuitive but is the easiest way I can think of to overlap shared expert and experts.

torchtitan/distributed/expert_parallel.py
```diff
@@ -4,99 +4,25 @@
...
-    return x.shape, x[permuted_indices, :], permuted_indices, num_tokens_per_expert
-
-
-def _unpermute(out, input_shape, permuted_indices):
-    out_unpermuted = out.new_empty(input_shape)
-    out_unpermuted[permuted_indices, :] = out
-    return out_unpermuted
-
-
 class BaseExpertParallel(ParallelStyle, ABC):
```
tianyu-l: we can delete this class now (and keep `ExpertParallel`)

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,512 @@
...
+@dataclass(frozen=True, kw_only=True)
+class LocalDispatchMetadata:
+    """Metadata returned by LocalTokenDispatcher.dispatch() for use in combine()."""
+
+    token_indices_experts_sorted: torch.Tensor  # (N*top_k,)
+    top_scores_experts_sorted: torch.Tensor  # (N*top_k,) scores in expert-sorted order
+
+
+@dataclass(frozen=True, kw_only=True)
+class DispatchMetadata(LocalDispatchMetadata):
```
tianyu-l: Maybe call this `AllToAllDispatchMetadata`? o/w sounds too generic

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,512 @@
...
+    """Metadata returned by TokenDispatcher.dispatch() for use in combine()."""
+
+    input_shape: tuple  # for _unpermute
+    permuted_indices: torch.Tensor  # for _unpermute
+    input_splits: list[int]
+    output_splits: list[int]
+
+
+class BaseTokenDispatcher(Configurable):
```
tianyu-l: Maybe fine to just use `LocalTokenDispatcher` as base class, like you are doing for metadata.

o/w Base doesn't have dispatch / combine functions, whereas Local inherits trivial config

Nothing fundamental here though.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def combine(
+        self,
+        routed_output: torch.Tensor,
+        metadata: LocalDispatchMetadata,
+    ) -> torch.Tensor:
+        return routed_output
+
+
+class TokenDispatcher(BaseTokenDispatcher):
```
tianyu-l: Similar, maybe
```suggestion
class AllToAllTokenDispatcher(BaseTokenDispatcher):
```
which is really short for `NCCLAllToAllTokenDispatcher`

torchtitan/trainer.py
```diff
@@ -115,6 +116,18 @@ def __post_init__(self):
                     raise NotImplementedError(
                         "Optimizers in backward is not supported with Pipeline Parallel."
                     )
+            if (
+                self.model_spec is not None
+                and hasattr(self.model_spec.model, "layers")
+                and self.parallelism.expert_parallel_degree > 1
+            ):
+                apply_ep(
```
tianyu-l, line 137: let's do this in config_registry instead of in trainer

torchtitan/distributed/expert_parallel.py
```diff
@@ -138,102 +64,22 @@ def _partition_fn(self, name: str, mod: nn.Module, device_mesh: DeviceMesh) -> N
...
-            self.input_splits,
-            self.output_splits,
-            device_mesh.get_group(),
-        )
-        return routed_output
+        # Set ep_group on the token dispatcher for all-to-all communication.
+        # device_mesh here is the 1D EP mesh.
+        mod.token_dispatcher.ep_group = (
+            device_mesh.get_group()
+        )  # pyrefly: ignore [missing-attribute]
```
tianyu-l: Need to assert it's instance of `TokenDispatcher` or `DeepEPTokenDispatcher`; Base doesn't have this attribute.

torchtitan/models/llama4/parallelize.py
tianyu-l: Can we put these TODO in code / issue, in case we forget?

torchtitan/models/common/moe.py
```diff
@@ -461,20 +421,20 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             if self.shared_experts is not None
             else torch.zeros_like(x)
         )
-
+        # Score multiply + scatter_add force the a2a to sync
         if not self.score_before_experts:
```
tianyu-l: hmm, it's weird that `score_before_experts` code is scattered inside and outside grouped experts module.

For overlapping, do you think we can pass `shared_experts` to `self.experts()` call, so that shared experts computation can be done inside, thus achieving overlapping as before?

torchtitan/models/llama4/parallelize.py
```diff
@@ -551,12 +547,6 @@ def apply_moe_ep_tp(
                     local_output_grad_placements=(Partial(),),
                 ),
             }
-            if ep_mesh is not None and etp_mesh is None:
-                # If TP is borrowed for EP, then split the tokens across TP ranks so that
-                # the reorderer, the all-to-all comms, and routed experts computation
-                # are effectively running Sequence Parallel (split along the folded bs*slen dim)
-                # pyrefly: ignore [no-matching-overload]
-                moe_layer_plan.update({"moe.reorderer": ReordererSequenceParallel()})
```
tianyu-l, line 572: what happened here if DeepEP moe is used? I think we should error out?

torchtitan/distributed/expert_parallel.py
tianyu-l: Let's add a TODO to remove this or mark it removable (if it becomes messy to maintain down the road).

torchtitan/distributed/expert_parallel.py
```diff
@@ -255,33 +101,35 @@ def _partition_fn(self, name: str, mod: nn.Module, device_mesh: DeviceMesh) -> N
...
-            output_fn=self._token_combine,
         )
 
 
-# This class is to support Sequence Parallel for ETP=1
-# when EP borrows from all TP and part of DP
-class ReordererSequenceParallel(ParallelStyle):
-    def __init__(self):
-        super().__init__()
+class ExpertSequenceParallel(ExpertParallel):
```
tianyu-l: Can we do this inside `AllToAllTokenDispatcher`? I guess you just need to give it the TP group size and rank.

I feel if we make proper helper function, it may be fine. If it hurts readability too much, we can have another dispatcher.

acisseJZhong: wondering why not keeping `ExpertSequenceParallel` here which inherits from `ExpertParallel` who does the weight partition? We simply choose to apply either `ExpertParallel` or `ExpertSequenceParallel` https://github.com/pytorch/torchtitan/blob/c2a134e420cc31161f4d2fd63f98a18d301e2611/torchtitan/models/llama4/parallelize.py#L596 or `ExpertTensorParallel`.

torchtitan/models/llama4/parallelize.py
acisseJZhong: for DeepEP MoE we simply fallback to `ExpertParallel`(it does not handle ExpertParallel). 
```
            if tp_mesh is not None and comm_backend not in ("deepep", "hybridep"):
                # ETP=1: EP borrows from TP. Each EP rank processes a
                # disjoint token subset (sequence parallel).
                experts_plan = ExpertSequenceParallel()
            else:
                # Weight sharding only — dispatch/combine handled by
                # the token dispatcher.
                experts_plan = ExpertParallel()
```

torchtitan/trainer.py
acisseJZhong: discussed offline, I will change both attn backend and moe backend in the next PR.

torchtitan/distributed/expert_parallel.py
acisseJZhong: nvm I recalled! we don't want keep the specific routing logic in parallelize.py. Moved.

torchtitan/models/common/moe.py
```diff
@@ -424,57 +398,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         # TODO: Activation Checkpointing has the side effect of double counting tokens_per_expert --
         #       first in the forward pass, and then in the backward pass. However, this has no
         #       effect on the expert bias update thanks to the torch.sign() operator.
-        with torch.no_grad():
-            self.tokens_per_expert.add_(num_tokens_per_expert)
+        if self.load_balance_coeff is not None:
```
tianyu-l: I think this was deliberated kept even when load balancing is off, because the routing data could be valuable anyways. Let's keep it.

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -99,19 +99,13 @@ def parallelize_deepseekv3(
...
-
-        pad_multiple = find_pad_multiple(model_converters.converters)
-
         apply_moe_ep_tp(
             model,
             tp_mesh=parallel_dims.get_optional_mesh("tp"),
             ep_mesh=parallel_dims.get_optional_mesh("ep"),
             etp_mesh=parallel_dims.get_optional_mesh("etp"),
             ep_etp_mesh=parallel_dims.get_optional_mesh(["ep", "etp"]),
             comm_backend=comm_backend,
```
tianyu-l, line 108: can remove this as well? (can be done in separate PR)

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,617 @@
...
+        super().__init__(config)
+        self.ep_degree = config.ep_degree
+        # Set by ExpertParallel / ExpertTensorParallel._apply()
+        self.ep_group: dist.ProcessGroup
+        self.sp_size = config.sp_size
+
+    def _split_along_sp(self, *tensors: torch.Tensor) -> list[torch.Tensor]:
+        """Split tensors along the first dim across EP ranks for sequence parallel."""
+        sp_size = self.sp_size
+        sp_rank = dist.get_rank(self.ep_group)
```
tianyu-l: wait, this rank shouldn't be from ep_group, should be from tp_group
https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py#L301

did you verify that this works with identical numerics?

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,617 @@
...
+        Returns:
+            routed_input: (R, dim) tokens in expert-major order for local experts
+            num_tokens_per_expert_local: (num_local_experts,) token counts
+            metadata: AllToAllDispatchMetadata for combine()
+        """
+        if self.sp_size > 1:
+            # NOTE: If needed, we can pad tokens in case bs*slen is not divisible by TP degree
+            # if top_scores.shape[0] % device_mesh.size() != 0:
+            #     num_tokens = top_scores.shape[0]
+            #     tp_size = device_mesh.size()
```
tianyu-l: outdated comments, maybe update

torchtitan/models/common/token_dispatcher.py
acisseJZhong: oops yeah I mixed the moe_layer_plan(applied with tp mesh https://github.com/pytorch/torchtitan/blob/42170d8abfc5825f5dda586676e411f000e2d695/torchtitan/models/llama4/parallelize.py#L584) and expert_plan(applied on ep mesh https://github.com/pytorch/torchtitan/blob/42170d8abfc5825f5dda586676e411f000e2d695/torchtitan/models/llama4/parallelize.py#L630). 

the loss parity cases I ran when ETP=1 has TP=EP=2, so didn't cover this bug.

torchtitan/models/deepseek_v3/parallelize.py
acisseJZhong: updated item in https://github.com/pytorch/torchtitan/issues/2956

torchtitan/models/common/token_dispatcher.py
acisseJZhong: passed in sp_rank here https://github.com/pytorch/torchtitan/blob/0bc04ba6eacb2dd1635911a38cbcc74f9b8502d0/torchtitan/models/gpt_oss/parallelize.py#L300

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,612 @@
...
+        if self.sp_size > 1:
+            # NOTE: If needed, we can pad tokens in case bs*slen is not divisible by TP degree
+            # shape (batch_size * seq_len // ep_size, top_k)
+            x, top_scores, selected_experts_indices = self._split_along_sp(
+                x, top_scores, selected_experts_indices
+            )
+
+        ep_size = self.ep_size
+
+        # TODO: Extract this local reordering block (histc, argsort, score
```
tianyu-l, line 235: I thought you said EP dispatcher doesn't necessarily need to have the same reorder steps.
Is it still needed for nccl AllToAll dispatcher, or are you going to change the implementation later?

acisseJZhong: I think before this PR, the torch.hist + torch.argsort(aka routing map generation) logic exists both in token reorderer https://github.com/pytorch/torchtitan/blob/5242bdf7151d8a57ca59ea54c656165d855772be/torchtitan/models/common/moe.py#L314 anddd moe forward https://github.com/pytorch/torchtitan/blob/5242bdf7151d8a57ca59ea54c656165d855772be/torchtitan/models/common/moe.py#L271 so it's duplicated. 

After this PR, it only exists once in the dispatcher. The todo is more about refactoring this part to be a helper function(and maybe mimic six's routing map generation logic, but don't have to) instead of duplicating code.

tianyu-l: I see. Sounds good.

torchtitan/distributed/expert_parallel.py
```diff
@@ -259,324 +96,18 @@ def _partition_fn(self, name: str, mod: nn.Module, device_mesh: DeviceMesh) -> N
...
-            return output
-
-        top_scores = _split_along_first_dim(top_scores)
-        selected_experts_indices = _split_along_first_dim(selected_experts_indices)
-
-        # shape (batch_size * seq_len // ep_degree, top_k)
-        return top_scores, selected_experts_indices
-
-    def _prepare_output_fn(self, mod, outputs, device_mesh):
-        # shape (batch_size * seq_len * top_k // ep_degree)
```
acisseJZhong, line 319: @tianyu-l double check, this comment should be `tp_degree` right? 

I updated the logic here since your last review https://github.com/pytorch/torchtitan/blob/622060ce7c54310550a1a794ee98b50da8f41bec/torchtitan/models/common/token_dispatcher.py#L424, let me know if it makes sense to you.

torchtitan/distributed/expert_parallel.py
```diff
@@ -259,324 +96,18 @@ def _partition_fn(self, name: str, mod: nn.Module, device_mesh: DeviceMesh) -> N
...
-        self.score_before_experts = score_before_experts
-        self.comm_backend = comm_backend
-        self.hybridep_non_blocking_expert_capacity_factor = (
-            hybridep_non_blocking_expert_capacity_factor
+        # Set ep_group on the token dispatcher for all-to-all communication.
+        # device_mesh is the 2D (EP, ETP) mesh; slice the EP dimension.
+        from torchtitan.models.common.token_dispatcher import (
+            AllToAllTokenDispatcher,
+            DeepEPTokenDispatcher,
         )
```
fegin: nit: since multiple function use this import, should we move it to module-level import? Claude likes to put the import into local sometimes. But since there multiple usages, I'm wondering if this is a good strategy.

torchtitan/models/common/config_utils.py
```diff
@@ -131,13 +135,108 @@ def make_router_config(
...
+    ep_size: int = 1,
+    sp_size: int = 1,
+    hybridep_non_blocking_expert_capacity_factor: float | None = None,
+    pad_multiple: int | None = None,
+) -> (
+    LocalTokenDispatcher.Config
+    | AllToAllTokenDispatcher.Config
+    | TorchAOTokenDispatcher.Config
+    | DeepEPTokenDispatcher.Config
+):
```
fegin: nit: lol, this is a bit verbose. Should we have a type alias defined in somewhere?

torchtitan/models/common/moe.py
```diff
@@ -278,69 +300,25 @@ def forward(
...
+
+    The forward pass proceeds as:
+    1. Router computes expert assignments
+    2. GroupedExperts.forward() handles:
+       a. dispatch — reorder tokens + optional EP communication to expert-owning ranks
+       b. expert computation
+       c. combine — starts async EP communication (NCCL all-to-all or DeepEP combine),
+          runs shared_experts in parallel with the async communication,
+          then forces sync (scatter_add for NCCL AllToAll, sync_combine for DeepEP)
+          and produces final output
```
fegin: Maybe add description saying that step a and step c are handled by TokenDispatcher. Also async EP communication is optional (if LocalTokenDispatcher is used).

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,631 @@
...
+    output_splits: list[int]
+
+
+class LocalTokenDispatcher(Configurable):
+    """Token dispatcher for EP=1. Handles local token reordering only.
+
+    Also serves as the base class for EP dispatchers (AllToAllTokenDispatcher,
+    DeepEPTokenDispatcher) which override dispatch() and combine().
+
+    Not an nn.Module — dispatchers have no learnable parameters or buffers.
```
fegin, line 46: ~We may have to make it as a Module because we want to use LocalMap sharding configuration. We can discuss this later.~

Please ignore my comment, I confused myself.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,631 @@
...
+        # Sequence parallel size. When EP borrows from TP (ETP=1),
+        # set to TP group size to split tokens across ranks.
+        sp_size: int = 1
+
+    def __init__(self, config: Config):
+        super().__init__(config)
+        self.ep_size = config.ep_size
+        self.sp_size = config.sp_size
+        # Set at runtime by ExpertParallel / ExpertTensorParallel._partition_fn()
+        self.ep_group: dist.ProcessGroup
```
fegin: Probably make it optional and actually check that it is not None before using it. It is safer and matches the reason why we introduce pyrefly.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -0,0 +1,631 @@
...
+        # Set at runtime by apply_moe_ep_tp from tp_mesh.get_local_rank()
+        self.sp_rank: int = 0
```
fegin: Set as -1 and check before using it?

torchtitan/models/common/config_utils.py
tianyu-l: Just use the base class?

torchtitan/models/common/moe.py
acisseJZhong: thanks, modified the comments. Async EP communication is not needed when EP=1.

torchtitan/models/common/token_dispatcher.py
acisseJZhong: yeah sounds good, better to set default to sth not valid, to make sure it's being properly set when used.

PR #2843, opened 2026-04-04
[graph_trainer] Standardize Graph Pass Signature

SherlockNoMad:
## Graph Pass Signature

All graph passes must follow this signature:
```python
def my_pass(gm: torch.fx.GraphModule, example_inputs, *, other_kwargs) -> torch.fx.GraphModule:
```
- The first two positional args are always `(gm, example_inputs)`.
- Any additional parameters must be **keyword-only** (after `*`).
- The pass must return the (possibly transformed) `GraphModule`.
- Passes that don't need `example_inputs` should still accept it (use `example_inputs=None`).

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -17,8 +17,6 @@ Pass Types:
 
 import operator
 from collections import defaultdict
-from collections.abc import Sequence
-from typing import Any
 
 import torch
 from torch._functorch.aot_autograd import JointWithDescriptors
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -370,7 +378,7 @@ def inductor_decomposition_pass(
 
 
 def full_inductor_compilation_pass(
-    gm: torch.fx.GraphModule, example_inputs
+    gm: torch.fx.GraphModule, example_inputs: tuple
 ) -> torch.fx.GraphModule:
```
xmfan: this annotation is wrong, compile_fx doesn't return a gm

Issue #2757, opened 2026-03-30
[RFC] Multi-turn SFT


haydn-jones:
Currently multi-turn SFT isn't supported, I assume the primary reason is because the code uses EOS tokens for detecting document boundaries, which won't work for tokenizers like Qwen3:

```py
In [6]: tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-0.6B")

In [7]: chat = [
   ...:     {"role": "user", "content": "Hi!"},
   ...:     {"role": "assistant", "content": "Hello there!"}
   ...: ]

In [8]: tokenizer.apply_chat_template(chat, tokenize=False)
Out[8]: '<|im_start|>user\nHi!<|im_end|>\n<|im_start|>assistant\n<think>\n\n</think>\n\nHello there!<|im_end|>\n'

In [9]: tokenizer.eos_token
Out[9]: '<|im_end|>
```

Just allowing multi-turn as it stands would make the 2nd assistant turn into a separate document.

This can probably be fixed pretty simply: both dataset styles (HF and Chat) are emitting positions that reset to 0 at packed conversation boundaries. Mask builders can just look at position resets instead of scanning for EOS:
```python
positions[i] <= positions[i-1] 
```
means "new document here," and will work with tokenizers like Qwen that use EOS as end of turn.

Outline would be like:

- Mask builder (`attention.py`)
  - Add `get_document_mask_from_positions(positions)` for flex path
  - Add `create_varlen_metadata_from_positions(positions)` for varlen path
  - Delete `get_document_mask_mod` and `create_varlen_metadata_for_document`
- Decoder (decoder.py)
  - `get_attention_masks` / `_get_flex_attention_masks`: read `extra_inputs["positions"]` instead of `(input_batch, tokenizer.eos_id)`
- Model-specific (`gpt_oss/model.py`, `llama4/model.py`)
  - Same swap in their `get_attention_masks`
- ChatDataset (`text_datasets.py`)
  - `_validate_messages`: accept multi-turn (optional system, alternating user/assistant, ends with assistant)
  - New `_get_assistant_spans`: prefix re-tokenization loop returning [(start, end), ...] like what is already in there
  - `_tokenize_sample`: mask everything, unmask assistant spans
- Tests
  - Multi-turn masking correctness
  - Cross-turn attention within a conversation
  - Cross-conversation isolation
  - Regression against existing single-turn configs

[...]

PR #2849, opened 2026-04-04
Fix RoPE position inconsistencies across chunked samples

francesco-bertolotti:
### Summary

This PR removes the modulo-based logic previously used to wrap RoPE positions:

```python
self._positions_buffer.extend(
    i % self.seq_len for i in range(len(sample_tokens))
)
```

Although this prevented RoPE indices from exceeding the cache size, it introduced inconsistencies in positional encoding across split samples.

---

### Problem

Using modulo to reset positions can break positional continuity when a single logical sequence is split across multiple chunks.

For example:

```text
tokens:    [tx ,ty ,tz ,t0 ,t1 ,t2 ,t3 ,t4 ,t5 ,t6 ,t7 ]
positions: [x  ,y  ,z  ,0  ,1  ,2  ,3  ,4  ,5  ,6  ,7  ]

tokens:    [t8 ,t9 ,t10,t11,t12,t13,t14,t0 ,t1 ,t2 ,t3 ]
positions: [8  ,9  ,10 ,0  ,1  ,2  ,3  ,0  ,1  ,2  ,3  ]
```

Here, the sequence `t0–t14` is split across two chunks. In the second chunk, positions are reset mid-sequence due to the modulo operation,even though the tokens belong to the same original sequence. This results in inconsistent positional encodings.

Since different samples in a batch cannot attend to each other, positional resets should align strictly with sample boundaries, not occur within a continued sequence.

---

#### 1. Removed modulo-based position wrapping

Eliminated `% seq_len` to preserve positional continuity within sequences.

#### 2. Introduced `_normalize_positions`

Adds a normalization step that ensures each sequence starts at position 0. This is applied in the dataloader.

#### 3. Added position validation tests

New test verify that:

* Positions are monotonically increasing within a sequence
* Positions reset to 0 after an EOS token
* Each sequence starts at position 0
* BOS tokens are always assigned position 0
* Positions remain within valid bounds (≤ sequence length)

### DIFF
--- a/torchtitan/hf_datasets/text_datasets.py
+++ b/torchtitan/hf_datasets/text_datasets.py
@@ -32,6 +32,7 @@ def _process_c4_text(sample: dict[str, Any]) -> str:
     """Process C4 dataset sample text."""
     return sample["text"]
 
+
 # Add your dataset here - more information at docs/datasets.md
 DATASETS = {
     "c4": DatasetConfig(
@@ -114,8 +115,8 @@ class HuggingFaceTextDataset(IterableDataset, Stateful):
     def _normalize_positions(self, positions: list[int]) -> list[int]:
         offset = positions[0]
         if offset > 0:
-            for i,p in enumerate(positions):
-                if p == 0: 
+            for i, p in enumerate(positions):
+                if p == 0:
                     break
                 positions[i] = p - offset
         return positions
@@ -143,9 +144,11 @@ class HuggingFaceTextDataset(IterableDataset, Stateful):
 
                 while len(self._inputs_buffer) >= max_buffer_token_len:
                     x = torch.LongTensor(self._inputs_buffer[:max_buffer_token_len])
-                    pos = torch.LongTensor(self._normalize_positions(
-                        self._positions_buffer[:max_buffer_token_len]
-                    ))
+                    pos = torch.LongTensor(
+                        self._normalize_positions(
+                            self._positions_buffer[:max_buffer_token_len]
+                        )
+                    )
                     # update buffers to the remaining tokens
                     self._inputs_buffer = self._inputs_buffer[max_buffer_token_len:]
                     self._positions_buffer = self._positions_buffer[
--- a/tests/unit_tests/test_dataloader.py
+++ b/tests/unit_tests/test_dataloader.py
@@ -79,7 +79,7 @@ class TestParallelAwareDataloader(unittest.TestCase):
         last_batch_input, last_batch_label = batches[-1]
         self.assertEqual(last_batch_input["input"].tolist(), [96, 97, 98, 99])
         self.assertEqual(last_batch_label.tolist(), [96, 97, 98, 99])
-        
+
     def test_validate_kwargs_rejects_invalid_kwargs(self):
         """Test that passing invalid kwargs raises ValueError."""
         dataset = DummyDataset()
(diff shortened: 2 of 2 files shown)


torchtitan/hf_datasets/text_datasets.py
```diff
@@ -122,28 +150,26 @@ def __iter__(self):
...
-                # positions at 0 per request.  Positions wrap at seq_len
-                # to stay within the RoPE cache, effectively chunking
-                # long documents into seq_len-sized segments.
-                # TODO: make overflow policy configurable (chunk / truncate / drop).
-                self._positions_buffer.extend(
-                    i % self.seq_len for i in range(len(sample_tokens))
-                )
                 self._sample_idx += 1
 
+                sample_tokens = self._apply_overflow_policy(sample_tokens, max_buffer_token_len)
```
tianyu-l: This doesn't sound right in terms of how the `overflow_policy` works with packing. Here's an example:
- `sample_tokens` is of length 90.
- `max_buffer_token_len` is of length 100.
- `self._inputs_buffer`, before adding this particular `sample_tokens`, already has length 99.

According to the policy, we would still get full `sample_tokens` because 90 < 100. However, it will be truncated below, regardless of the policy.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -211,6 +237,18 @@ class Config(ParallelAwareDataloader.Config):
         infinite: bool = True
         """Whether to loop the dataset infinitely"""
 
+        overflow_policy: Literal["chunk", "truncate", "drop"] = "chunk"
+        """
+        Policy for handling samples that exceed seq_len: 
+        'chunk' (default) splits into multiple samples, 'truncate' truncates to fit, and 'drop' skips the sample.
+        """
+
+        reset_positions_on_chunk: bool = True
```
tianyu-l: I'm worried about out-of-index error if we set it to False, and we get an ultra-long sequence whose length is more than `training.seq_len`.
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/rope.py#L232

Maybe we can use `torch._assert_async` to do assertion there.

joecummings:
Thanks for being proactive about this one - we discussed it briefly in #2769 / #2757. 

I'd like to get this landed but simplified.
1. Can we move the drop/truncate/chunk to a new PR? That's not technically part of this fix. Leave the TODO in there
2. I think we can let the positions buffer extend monotonically with `self._positions_buffer.extend(range(len(sample_tokens)))` and then normalize the leading document fragement if the first position id AFTER slicing is > 0.

torchtitan/hf_datasets/text_datasets.py
francesco-bertolotti: I kept this optional because some users may want direct access to the true token positions within the model and handle positional embeddings using a custom implementation. That said, keeping customization simple might be preferable, especially since it’s straightforward to extend `HuggingFaceTextDataset` and override `_normalize_positions` if needed.

francesco-bertolotti:
Happy to help @joecummings.
> Can we move the drop/truncate/chunk to a new PR? That's not technically part of this fix. Leave the TODO in there.

No problem, as @tianyu-l pointed out, the current implementation was a bit lacking anyway.

> I think we can let the positions buffer extend monotonically with `self._positions_buffer.extend(range(len(sample_tokens)))` and then normalize the leading document fragment if the first position id *after slicing* is > 0.

In the latest commit, `_normalize_positions` implements exactly this behavior.

francesco-bertolotti:
@joecummings, thanks for the feedback, I have updated the PR description and added a test regarding positions.

francesco-bertolotti:
I don’t think the linting error is related to this PR, as it appears to be tied to the training and validation loss stack. I’ll take a closer look at it tomorrow.

tianyu-l:
oh might be fixed in https://github.com/pytorch/torchtitan/commit/3057525dc6a88e3c8fe5d961530989fea39b9183

PR #2855, opened 2026-04-06
Skip GPU CI workflows for draft PRs to save compute resources

SherlockNoMad:
Draft PRs don't need to run expensive H100/ROCm/A10 integration tests. CI will automatically trigger when the PR is marked ready for review. Lightweight jobs (lint, CPU unit tests, docker builds) still run on drafts for fast feedback.

### DIFF
--- a/.github/workflows/set-matrix.yaml
+++ b/.github/workflows/set-matrix.yaml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   set:
-    if: github.event.pull_request.draft == false
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.set.outputs.matrix }}
--- a/.github/workflows/integration_test_8gpu_graph_trainer.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer.yaml
@@ -30,7 +30,7 @@ permissions:
 
 jobs:
   build-test:
-    if: github.event.pull_request.draft == false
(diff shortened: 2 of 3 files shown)


.github/workflows/integration_test_4gpu_rl.yaml
```diff
@@ -29,6 +29,7 @@ permissions:
 jobs:
   # Step 1: Dynamically compute the matrix based on conditions
   set-matrix:
+    if: github.event.pull_request.draft == false
```
wwwjn: Is it possible to put `if: github.event.pull_request.draft == false` into `set-matrix.yaml`?

SherlockNoMad: done.

.github/workflows/set-matrix.yaml
```diff
@@ -18,6 +18,7 @@ on:
 
 jobs:
   set:
+    if: github.event.pull_request.draft == false
```
wwwjn: Wait, this will skip all "push" and "scheduled" events on main branch, right? We still need to run these kind of jobs.

SherlockNoMad:
> Is there a way we can make a PR draft from the very beginning, instead of making a non-draft PR and then marking it as draft?

ChatGPT: No — GitHub does not appear to have a repository setting that makes all newly created PRs default to draft. The docs describe draft PRs as a choice made when creating the PR, via the UI dropdown or CLI/API, rather than as a repo-level default.  ￼

SherlockNoMad:
@copilot 

I intend to skip triggering GPU CIs for all draft PR. Does this change looks good to you?

PR #2857, opened 2026-04-07
[graph_trainer] Add CI failure monitoring to nightly.md

yiming0416:
Summary: 
- Adds a new "CI Failure Monitoring" section (step 6) to the graph_trainer nightly scout instructions (`nightly.md`)

- The new section instructs the nightly scout to check the two GraphTrainer periodic CI workflows via `gh run list`, parse failure logs for root-cause errors, identify the PyTorch nightly regression range by extracting commit hashes from wheels, and generate a minimal local repro.


yiming0416:
> How do we run the nightly scout? Is it a cron job on someone's dev server at the moment?

I think using cron job is the plan. But for `gh` related commands, we have to run it from a devserver with proper permission.

PR #2859, opened 2026-04-07
introduce param_groups to optimizer config

tianyu-l:
It is common to have different learning rates and weight decay rates for different parameters in the model.

Requested in https://github.com/pytorch/torchtitan/issues/2599, and as an alternative approach to https://github.com/pytorch/torchtitan/pull/2600.

In this PR we put `param_groups` as a `list[ParaGroupConfig]` into the optimizer config, to override lr and weight decay with multipliers based on regex matching.

### DIFF
--- a/tests/integration_tests/models.py
+++ b/tests/integration_tests/models.py
@@ -111,13 +111,13 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
-                    "--module qwen3 --config qwen3_debugmodel",
+                    "--module qwen3 --config qwen3_debugmodel_param_groups",
                     "--parallelism.data_parallel_shard_degree 2",
                     "--parallelism.tensor_parallel_degree 2",
                 ],
             ],
-            "Qwen3 FSDP+TP",
-            "qwen3_fsdp+tp",
+            "Qwen3 FSDP+TP (param groups)",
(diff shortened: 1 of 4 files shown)


tianyu-l:
cc @francesco-bertolotti

francesco-bertolotti:
This approach works well when the intended use is limited to overriding learning rate and weight decay for specific parameters. It would be great if other hyperparameters were also supported, such as `betas`. More generally, it would be even more powerful if the design allowed assigning entirely different optimizer types to different parameter groups, for example, using Muon for some parameters and AdamW for others.

For what it's worth, I've been working on exactly this in the latest revision of #2600, which introduces a `BaseOptimizer` class hierarchy with typed configs per optimizer, and supports mixed-optimizer parameter groups. 

Happy to discuss whether it makes sense to consolidate the two approaches!

tianyu-l:
@francesco-bertolotti 

> More generally, it would be even more powerful if the design allowed assigning entirely different optimizer types to different parameter groups, for example, using Muon for some parameters and AdamW for others.

That makes sense (actually we have plans to support that).

Since the scope is larger than the original request in https://github.com/pytorch/torchtitan/issues/2599, maybe we can land this first (I'll add `betas` to the `ParamGroupConfig`) and address mixed optimizers in later PR.

wwwjn (approved):

Issue #2864, opened 2026-04-07
ChatDataset mid training resume not reproducible with multi epoch


slimfrkha:
### Bug description
Shuffling the dataset is different between
1. training relooping
2. state dict loading to resume training
This problem appears on ChatDataset in multi epoch training if resuming after epoch > 1

Basically the below is happening right now:
```python
from datasets import load_dataset
ds = load_dataset("cornell-movie-review-data/rotten_tomatoes", split="validation")

# in training
shuffled_ds_training = ds
for epoch in range(4):
        shuffled_ds_training =shuffled_ds_training.shuffle(seed=42 + epoch)
print(shuffled_ds_training['label'][:10])

# when loading from state to resume training
shuffled_ds_resume_training = ds.shuffle(seed=42 + epoch)
print(shuffled_ds_resume_training['label'][:10])
```

## Solution
Either on state dict load, do a perform to replay all the previous epochs shuffling or dissociate shuffled data from original dataset and shuffle always from original dataset.

### Versions

main branch latest commit `0e0590c137599276d36128abc1702efe9e091607` as of this writing.

PR #2865, opened 2026-04-07
fix(data-ckpting-resume): fix broken shuffling

slimfrkha:
Solve #2864
dissociate original dataset from shuffled dataset.
cc @joecummings as you worked on ChatDataset.

### DIFF
--- a/torchtitan/hf_datasets/text_datasets.py
+++ b/torchtitan/hf_datasets/text_datasets.py
@@ -272,7 +272,8 @@ class ChatDataset(IterableDataset, Stateful):
                 "ChatDataset requires a tokenizer with a valid EOS token."
             )
 
-        self._data = split_dataset_by_node(dataset, dp_rank, dp_world_size)
+        self._original_data = split_dataset_by_node(dataset, dp_rank, dp_world_size)
+        self._data = self._original_data
         self._tokenizer = tokenizer
         self._eos_id = tokenizer.eos_id
         self.seq_len = seq_len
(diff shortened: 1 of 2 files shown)


torchtitan/hf_datasets/text_datasets.py
```diff
@@ -381,8 +382,8 @@ def _iter_greedy_packed(self):
             for sample in self._get_data_iter():
                 # pyrefly: ignore [bad-argument-type]
                 result = self._tokenize_sample(sample)
-                self._sample_idx += 1
                 if result is None:
+                    self._sample_idx += 1
```
joecummings, line 386: I don't think you need to move around the sample_idx to achieve this change - can you remove these changes?

joecummings (commented):
Just one comment, but thanks for looking into this!

torchtitan/hf_datasets/text_datasets.py
slimfrkha: test fail without this change.
When the generator yields at the "doesn't fit" branch (current sample's tokens don't fit in the remaining space of the buffer), it suspends after flushing the buffers but before adding the current sample's tokens to the now-empty buffer. If `_sample_idx` was already incremented, the saved state says "I'm done with sample S" while the buffers don't contain sample S's tokens yet. On restore, the dataset skips to sample S+1 and sample S is lost forever, causing the two streams to diverge.
By moving the increment to after the extend calls, we guarantee that at every possible yield point, `_sample_idx` and the buffer contents agree: every sample counted by the index is either already yielded in a previous batch or present in the current buffer.

joecummings: I see - thank you for the clarification. 

Once you fix the linter, happy to merge!

tianyu-l (approved):

torchtitan/hf_datasets/text_datasets.py
slimfrkha: the linter is failing on 2 files not involved in this PR.

PR #2872, opened 2026-04-07
GQAttention: Combine `q_norm` and `k_norm` into `qk_norm`

Andrei-Aksionov:
Hey there 👋 

Query and Key norm was introduced in Qwen 3 model architecture.
The current code provides an ability to:
- Configure them separately
- Set one of them to None

There are reasons why to apply them both:
- Better training stability
- Lower sensitivity to actiovations scale
- Dot product of normalized values behaves similarly to cosine similarity

At the same time there is no real reason, appart from a research or an ablation study, to apply them asymmetricaly, e.g.:
- Query norm is None, but Key norm isn't
- Query norm isn't None, but Key is

(If a research group decides to do such a study, the needed changes will require only modifying a couple of lines of code.)

---

This PR collapses separate `q_norm` and `k_norm` into a combined `qk_norm` and reuses `_qwen_3_norm` function
https://github.com/pytorch/torchtitan/blob/8328876dbe5843a6fc6aeef7480b1535e17c6b66/torchtitan/models/qwen3/__init__.py#L78-L79

since parameters are exactly the same.

### DIFF
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -641,7 +641,8 @@ class GQAttention(BaseAttention):
         xv = xv.view(bs, seqlen, -1, self.head_dim)
 
         # Optional QK normalization (before RoPE, per Qwen3)
-        if self.q_norm is not None and self.k_norm is not None:
+        if self.q_norm is not None or self.k_norm is not None:
+            assert self.q_norm is not None and self.k_norm is not None
             xq = self.q_norm(xq)
             xk = self.k_norm(xk)
 


torchtitan/models/common/attention.py
```diff
@@ -613,11 +613,8 @@ def __init__(self, config: Config):
         self.rope_backend = config.rope_backend
 
         # Optional QK normalization (Qwen3-style)
-        self.q_norm: RMSNorm | None = None
-        self.k_norm: RMSNorm | None = None
-        if config.q_norm is not None and config.k_norm is not None:
```
tianyu-l, line 618: I agree this code doesn't look right (e.g. what if `q_norm is not None` but `k_norm is None`). However, in reality have you seen use cases where only one of q / k gets norm?

If not, we should modify the code to disallow that.

Andrei-Aksionov: > However, in reality have you seen use cases where only one of q / k gets norm?

No, I haven't. 
I can imagine that someone might want to do a research or an ablation study on importance of using both in a large-scale scenario, so this repo might be used. But in this case that imaginary research group can easily tweak just a couple of lines of code, if necessary.

> If not, we should modify the code to disallow that.

I think we can collapse `q_norm` and `k_norm` into `qk_norm`.

tianyu-l: I see. Yeah if they always share the same init, merging the config sounds valid. If users would like to research more combinations, they can always create their own config and module.

torchtitan/models/common/attention.py
```diff
@@ -647,9 +641,8 @@ def forward(
         xv = xv.view(bs, seqlen, -1, self.head_dim)
 
         # Optional QK normalization (before RoPE, per Qwen3)
-        if self.q_norm is not None:
+        if self.q_norm is not None and self.k_norm is not None:
```
tianyu-l: better have a safety check
```suggestion
        if self.q_norm is not None or self.k_norm is not None:
            assert self.q_norm is not None and self.k_norm is not None
```

tianyu-l (approved):
please address remaining comment

PR #2874, opened 2026-04-07
[ROCm][CI] Turn off ROCm CI on experimental workflows

jithunnair-amd:
The push tag `ciflow/8gpu` trigger [does not respect](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore) the paths qualification:
> Path filters are not evaluated for pushes of tags.

This results in these experimental workflows getting triggered for all PRs that have the `ciflow/8gpu` tag, which severely overloads the limited ROCm CI capacity.

### DIFF
--- a/.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
@@ -7,6 +7,7 @@ on:
       - 'torchtitan/experiments/graph_trainer/**'
       - '.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml'
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
     paths:
       - 'torchtitan/experiments/graph_trainer/**'
       - '.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml'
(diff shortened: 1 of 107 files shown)


.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
tianyu-l, line 1: if possible, let's keep this in the test, since it's not for experimental features

jithunnair-amd: Oh, I saw the paths containing `experiments`:
```
    paths:
```
Should only this one be retained? How about `.github/workflows/integration_test_8gpu_graph_trainer.yaml`?

tianyu-l: Oh sorry, I misread. I thought this was .github/workflows/integration_test_8gpu_h100.yaml which tests core.

You are right, let's remove the ciflow/8gpu test for it.

jithunnair-amd:
@tianyu-l The currently running ROCm jobs on this PR won't succeed either, because they are using the gfx942.8 runner labels, which have very high queueing. We have [another PR](https://github.com/pytorch/torchtitan/pull/2740) which will use the new MI350 ecosystem runner labels.

jithunnair-amd:
@tianyu-l Thanks for approving the PR. I rebased it since the one ROCm CI job failure should be resolved by latest main. Please approve the CI workflows.

jithunnair-amd:
@tianyu-l Okay, doesn't look like the rebase helped, but this PR has no bearing on the non-experimental workflows, so can we merge this to reduce the queueing on ROCm CI jobs?

Issue #375, opened 2024-05-31
about reference of weight init according to layer depth or layer id 


SeunghyunSEO:
hi, first of all, thank you for the nice opensource project !
i just have been reading your model code and found it initialize model weights following num_layers or layer_id.
it is not conventional like kaming init (std=1/\sqrt{fan_in}) or GPT-2 init (std=0.02).
and it also dost not look like MuP or something.
so i just want to know if there are any references or it's just empirical for training stability.

edit) i forgot std=0.02/\sqrt{depth} init for output layers of residual block is from GPT-2 paper. sry ! just wondering where depth_init is from

wanchaol:
@lessw2020

lessw2020:
Hi @SeunghyunSEO  - sorry for the delay, didn't see this earlier.
To your question - the depth init came about from research last summer when we we were doing work on parallel attention blocks. I did a comparison sweep and adding the depth init was the winner, so have continued to use that. 
I'm not sure where the concept came from though - it bubbled up in discussions with IBM research a while back. 
I did see that Olmo was also using this, and they referred to it as a "mitchell init" but I was not able to find anything in arxiv on it. 
Anyway, short answer is it's empirically based.  We haven't done a sweep on it though since llama3 came out so maybe we will revisit it in the future but it continues to perform well in our training runs.
Hope that helps!

SeunghyunSEO:
thank you for the kind answer @lessw2020 !
i guess it makes sense because Mitchell init's layerwise output variance would be more consistent compared to GPT init

garrett361:
I had this same question. Some analysis, in case anyone is interested:

* `depth_init = False`: rescales output `nn.Linear` layers by `~ 1 / sqrt(n_layers)`. This is what gpt2 does: [Sec 2.3](https://cdn.openai.com/better-language-models/language_models_are_unsupervised_multitask_learners.pdf). The idea, presumably, is that the final output from the sum of all residuals is $z \sim \sum_{\ell=0}^{L-1} r_\ell$ for $L$ layers. The sum vanishes in expectation over model param init, $E[z]=0$, so its size is characterized by its second moment, $E[z^2]$ and the $r_\ell \sim 1 / \sqrt{L}$ initialization enforces that $E[z^2] \sim  \mathcal{O}(1)$ is independent of $L$ at leading order.  
[...]

PR #2878, opened 2026-04-07
Fused QKV GQAttention implementation

wwwjn:
## Implementation
Replaces separate wq, wk, wv with single wqkv linear of shape [n_kv_heads * R * head_dim, dim], where R = heads_per_kv + 2. Fused q/k/v project to reduce number of matmul kernel launches, and also reduce the number of TP communication kernel launches.

Model-wise: 
- DeepSeek-v3 is untouched as the Attention has 2 different computation path (depend on `q_lora_rank`), makes it not easy to interact with fusing decision. 
- Flux is also untouched
 

## Test 
using llama3 debug model:
- Numerics test: When initialized using the same seed checkpoint: 
Screenshot 2026-04-07 at 6 02 38 PM

- Performance test fused v.s. unfused: 
    - CPU bounded model: Screenshot 2026-04-07 at 6 04 39 PM
    - GPU (memory-bandwidth bounded, or GPU compute bounded): Screenshot 2026-04-07 at 7 22 47 PM

- Test with DCP + state_dict_adapter: 
Screenshot 2026-04-09 at 2 19 35 PM

### DIFF
--- a/torchtitan/models/qwen3_vl/parallelize.py
+++ b/torchtitan/models/qwen3_vl/parallelize.py
@@ -39,6 +39,7 @@ from torchtitan.distributed.activation_checkpoint import apply_ac
 from torchtitan.distributed.compile import apply_compile
 from torchtitan.distributed.fsdp import get_fsdp_reshard_after_forward_policy
 from torchtitan.distributed.tensor_parallel import NoParallel
+from torchtitan.models.common.attention import FusedQKVLinear
 from torchtitan.models.llama4.parallelize import apply_fsdp, apply_moe_ep_tp
 from torchtitan.protocols.model_converter import ModelConvertersContainer
 from torchtitan.tools.logging import logger
@@ -79,8 +80,28 @@ def _apply_non_moe_tp_to_decoder(
     # tp_mesh (needed for consistent (fsdp, tp) mesh after FSDP). Its input
     # hook wraps plain tensors as DTensor(Replicate); its output stays as
     # DTensor (no unwrap) so downstream modules receive DTensors.
+
+    # Detect whether fused QKV is used by checking the first layer
+    # pyrefly: ignore [not-callable]
+    first_block = next(iter(model.layers.values()))
+    use_fused_qkv = isinstance(
+        first_block.attention.qkv_linear,  # pyrefly: ignore [missing-attribute]
+        FusedQKVLinear,
+    )
+
     # pyrefly: ignore [not-callable]
     for transformer_block in model.layers.values():
+        if use_fused_qkv:
+            qkv_plan = {
+                "attention.qkv_linear.wqkv": ColwiseParallel(use_local_output=False),
+            }
+        else:
+            qkv_plan = {
+                "attention.qkv_linear.wq": ColwiseParallel(use_local_output=False),
+                "attention.qkv_linear.wk": ColwiseParallel(use_local_output=False),
+                "attention.qkv_linear.wv": ColwiseParallel(use_local_output=False),
+            }
+
         layer_plan = {
             "attention_norm": NoParallel(),
             "ffn_norm": NoParallel(),
@@ -92,9 +113,7 @@ def _apply_non_moe_tp_to_decoder(
                 input_layouts=(Replicate(), Replicate(), None, None),
                 desired_input_layouts=(Replicate(), Replicate(), None, None),
             ),
-            "attention.wq": ColwiseParallel(use_local_output=False),
-            "attention.wk": ColwiseParallel(use_local_output=False),
-            "attention.wv": ColwiseParallel(use_local_output=False),
+            **qkv_plan,
             # Not actual sequence parallelism — SequenceParallel(sequence_dim=2)
             # tells DTensor that dim 2 (the head dimension) is the sharded dim,
             # so the per-head RMSNorm operates on each rank's local heads without


torchtitan/models/common/attention.py
```diff
@@ -622,12 +619,15 @@ def __init__(self, config: Config):
         # Scaling factor (needed when head_dim differs from dim // n_heads)
         self.scaling = self.head_dim**-0.5 if config.head_dim is not None else None
 
-        self.wq = config.wq.build()
-        self.wk = config.wkv.build()
-        self.wv = config.wkv.build()
-        self.wo = config.wo.build()
+    def _project_qkv(
```
tianyu-l: I think here composition makes more sense than inheritance, because essentially we are just swapping the qkv projection, but keeping all other implementation untouched.

Can we introduce a single `wqkv` in the config, of class `QKVLinear`, which can have a subclass `FusedQKVLinear`?

torchtitan/models/common/attention.py
```diff
@@ -675,4 +669,72 @@ def forward(
...
+        return self.wo(output)  # pyrefly: ignore [not-callable]
+
+
+class GQAttention(BaseGQAttention):
+    """Grouped-Query Attention with separate Q, K, V projections."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(BaseGQAttention.Config):
+        wq: Linear.Config
+        wkv: Linear.Config
```
tianyu-l: The reason of keeping them separately might be that they have different `param_init`? I think we should remove depth_init and thus may be able to make them into one `wqkv`.

See discussions in https://github.com/pytorch/torchtitan/issues/375#issuecomment-4187484549

It will change numerics, but it's fine.

torchtitan/models/gpt_oss/state_dict_adapter.py
tianyu-l, line 1: are these changes tested with TP applied on DCP side?

torchtitan/models/llama3/parallelize.py
```diff
@@ -217,18 +217,33 @@ def apply_tp(
         output_layouts=sp_layout, use_local_output=enable_sp
     )
 
+    # Detect whether fused QKV is used by checking the first layer
+    # pyrefly: ignore [not-callable]
+    first_block = next(iter(model.layers.values()))
+    use_fused_qkv = hasattr(
+        first_block.attention, "wqkv"  # pyrefly: ignore [missing-attribute]
```
tianyu-l: use `isinstance` not `hasattr`

torchtitan/models/common/attention.py
wwwjn: I see, the refined design would be:

- Only one attention GQAttention class, has a field `wqkv: BaseQKVLinear`
- Class `QKVLinear` is a subclass of `BaseQKVLinear`,  and has fields `wq` / `wk` / `wv`
- Class `FusedQKVLinear` is a subclass of `BaseQKVLinear`,  and has fields `wqkv`

torchtitan/models/common/attention.py
wwwjn: The reason is only about `depth_init`, also becuase the wk and wv has the same number of heads (num_kv_heads), but in GQAttention, `num_q_heads` not necessarily equal to num_kv_heads. So wq has different Linear.out_features

torchtitan/models/common/attention.py
```diff
@@ -565,23 +568,106 @@ def __post_init__(self):
...
+    :class:`FusedQKVLinear` for a single fused projection.
     """
 
     @dataclass(kw_only=True, slots=True)
     class Config(BaseAttention.Config):
         n_heads: int
         dim: int
-        wq: Linear.Config
-        wkv: Linear.Config
+        qkv: BaseQKVLinear.Config
```
tianyu-l: qkv sounds ambiguous, how about
```suggestion
        qkv_linear: BaseQKVLinear.Config
```

torchtitan/models/common/attention.py
```diff
@@ -565,23 +568,106 @@ def __post_init__(self):
...
+        wkv: Linear.Config
+
+    def __init__(self, config: Config):
+        super().__init__()
+        self.wq = config.wq.build()
+        self.wk = config.wkv.build()
+        self.wv = config.wkv.build()
+
+    def forward(
+        self, x: torch.Tensor, bs: int, seqlen: int, head_dim: int
```
tianyu-l: Let's not pass bs / seqlen / head_dim in forward. Instead
- put `head_dim` in config
- infer bs and seqlen from x shape

torchtitan/models/common/attention.py
```diff
@@ -565,23 +568,106 @@ def __post_init__(self):
...
+
+    def __init__(self, config: Config):
+        super().__init__()
+        self.wqkv = config.wqkv.build()
+        self.heads_per_kv = config.heads_per_kv
+        # R = Q-heads-per-KV-group + 1 (K) + 1 (V)
+        self.r_dim = config.heads_per_kv + 2
+
+    def forward(
+        self, x: torch.Tensor, bs: int, seqlen: int, head_dim: int
```
tianyu-l: similarly, I think it's better to store `head_dim` (if we put it in Base class config, we automatically get it here).

torchtitan/models/qwen3/__init__.py
```diff
@@ -456,6 +458,44 @@ def _8b_varlen() -> Qwen3Model.Config:
     return config
 
 
+def _4b_fused_qkv() -> Qwen3Model.Config:
```
tianyu-l: let's create a debug model with fused_qkv and swap one qwen3 model CI test (maybe the moe one) to guard on it

wwwjn: Nice catch, this config is for perf testing and I removed it.

tests/integration_tests/models.py
```diff
@@ -147,6 +147,18 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
             "qwen3_fsdp+tp+ep+etp",
             ngpu=4,
         ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module qwen3 --config qwen3_debugmodel_fused_qkv",
```
tianyu-l: Can we piggyback on existing tests instead of creating a new one? I'm worried about CI pressure in general.

PR #2881, opened 2026-04-07
Set cast_forward_inputs=False for LLM model's MixedPrecisionPolicy 

yiming0416:
Summary

- Set `cast_forward_inputs=False` on all `MixedPrecisionPolicy` instances. By default, FSDP2's `cast_forward_inputs=True` casts all forward inputs to param_dtype (e.g. bfloat16). In torchtitan, `TransformerBlock.forward(x, freqs_cis, attention_masks, positions)` is wrapped by `fully_shard`, and the `freqs_cis` (RoPE cache) is precomputed as float32 (cos/sin RoPE). Casting `freqs_cis` to bfloat16 would lose necessary precision in the cos/sin case. 

- For the other forward inputs: x (intermediate activations) is already in param_dtype from the preceding FSDP-wrapped module's output or the embedding layer, so the cast is a no-op. `attention_masks` and `positions` are not floating-point tensors, so FSDP2's cast_forward_inputs skips them.

-  `cast_forward_inputs=False` is only safe for LLM models where all float inputs to FSDP-wrapped modules are already in param_dtype. Image models (Flux, VLM) receive external float32 inputs (pixel values, timesteps, text embeddings) from preprocessing that genuinely need to be cast to param_dtype, so they retain the default `cast_forward_inputs=True`.

Test Plan:

```
pytest torchtitan/experiments/graph_trainer/tests/test_numerics.py 
```

Will wait the numerics tests in graph_trainer to be re-enabled first and then land this PR.

### DIFF
--- a/torchtitan/experiments/autoparallel/llama3/parallelize_llama.py
+++ b/torchtitan/experiments/autoparallel/llama3/parallelize_llama.py
@@ -96,7 +96,11 @@ def parallelize_llama(
 
     param_dtype = TORCH_DTYPE_MAP[training.mixed_precision_param]
     reduce_dtype = TORCH_DTYPE_MAP[training.mixed_precision_reduce]
-    mp_policy = MixedPrecisionPolicy(param_dtype=param_dtype, reduce_dtype=reduce_dtype)
+    mp_policy = MixedPrecisionPolicy(
+        param_dtype=param_dtype,
+        reduce_dtype=reduce_dtype,
+        cast_forward_inputs=False,
+    )
     with AutoParallel(
         model,
         input_fn,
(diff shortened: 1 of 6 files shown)


torchtitan/experiments/graph_trainer/simple_fsdp.py
```diff
@@ -20,6 +19,7 @@
     Shard,
 )
 from torch.distributed.device_mesh import DeviceMesh
+from torch.distributed.fsdp import MixedPrecisionPolicy
```
tianyu-l: Let simple_fsdp depend on FSDP2 class sounds unnecessary, especially when SimpleFSDP doesn't support `cast_forward_inputs` and `output_dtype`.

I prefer we stay as is for simple fsdp, since we are not trying to unify the API right now.

yiming0416: Good point. I'll keep simple_fsdp untouched

yiming0416:
`cast_forward_inputs=False` is only safe for LLMs where all float inputs to FSDP-wrapped modules are already in param_dtype. 

Image models (Flux, VLM) receiving external float32 inputs (pixel values, timesteps, text embeddings) from preprocessing still need to be cast to param_dtype, so they retain the default `cast_forward_inputs=True`.

PR #2883, opened 2026-04-08
[graph_trainer] Enable precompile with regional_inductor

bobrenjc93:
* __->__ #2883

Two changes are needed to make regional_inductor precompile work with
distributed DTensor ops (FSDP + TP, flex_attn):

1. Extend the GraphPickler ops filter to allow _dtensor and device_mesh
   ops.  DTensor's backward reconstructs DeviceMesh submeshes from
   tracked ancestor meshes via _get_submesh, and derives ProcessGroups
   via mesh_get_process_group.  These ops appear in the backward graph
   as call_function nodes.  Without the filter, GraphPickler rejects
   them during serialization.

2. Add a metadata key filter that strips "val" and "eager_input_vals"
   from node metadata before serialization.  These keys can contain
   unpicklable distributed objects (DeviceMesh, ProcessGroup) that are
   produced as side effects of tracing DTensor ops.  They are not needed
   at load time — the ops will produce fresh outputs when re-executed.

Also adds the regional_inductor precompile test definition (replacing
the previous TODO) in run_precompile_tests.py.

Requires PyTorch fix: https://github.com/pytorch/pytorch/pull/179660

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -106,8 +106,7 @@ def _node_metadata_key_filter_distributed(key: str) -> bool:
     Distributed ops (e.g. _get_submesh, mesh_get_process_group) produce
     opaque values (DeviceMesh, ProcessGroup) in node.meta["val"] and
     node.meta["eager_input_vals"] that cannot be pickled.  We strip
-    both — they are not needed at load time since the ops will produce
-    fresh outputs when re-executed.
+    both — they are not needed at runtime.
     """
     if key in ("val", "eager_input_vals"):
         return False


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -85,17 +85,35 @@ def _ops_filter_with_distributed(name: str) -> bool:
...
 
 
+def _node_metadata_key_filter_distributed(key: str) -> bool:
+    """Metadata key filter for regional_inductor with distributed ops.
+
+    Distributed ops (e.g. _get_submesh, mesh_get_process_group) produce
+    opaque values (DeviceMesh, ProcessGroup) in node.meta["val"] and
+    node.meta["eager_input_vals"] that cannot be pickled.  We strip
+    both — they are not needed at load time since the ops will produce
+    fresh outputs when re-executed.
```
xmfan: during compile time on load?

PR #2884, opened 2026-04-08
Add A6000 and B300 peak FLOPS to get_peak_flops

Johnsonms:
Add missing NVIDIA GPU entries:
- A6000 (Ampere): 309.7 TFLOPS BF16
- B300 (Blackwell): 2.25 PFLOPS BF16

B300 is placed after the GB200/GB300 check to prevent the substring "B300" from falsely matching "GB300".

### DIFF
--- a/torchtitan/tools/utils.py
+++ b/torchtitan/tools/utils.py
@@ -91,6 +91,11 @@ def get_peak_flops(device_name: str) -> float:
     if "A100" in device_name:
         # data from https://www.nvidia.com/en-us/data-center/a100/
         return 312e12
+    elif "A6000" in device_name:
+        # data from https://www.nvidia.com/content/dam/en-zz/Solutions/design-visualization/
+        # quadro-product-literature/proviz-print-nvidia-rtx-a6000-datasheet-us-nvidia-1454980-r9-web%20(1).pdf
+        # NOTE: 309.7 TFLOPS is with sparsity; dense value is half.
+        return 154.85e12
     elif "H100" in device_name:
         # data from https://www.nvidia.com/en-us/data-center/h100/
         # NOTE: Specifications are one-half lower without sparsity.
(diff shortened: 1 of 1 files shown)


torchtitan/tools/utils.py
```diff
@@ -118,6 +121,8 @@ def get_peak_flops(device_name: str) -> float:
         # GB200 data from https://www.nvidia.com/en-us/data-center/dgx-gb200
         # GB300 data from https://www.nvidia.com/en-us/data-center/dgx-gb300
         return 2.5e15
+    elif "B300" in device_name:
+        return 2.25e15
```
tianyu-l: maybe merge this with the entry below

torchtitan/tools/utils.py
```diff
@@ -91,6 +91,9 @@ def get_peak_flops(device_name: str) -> float:
     if "A100" in device_name:
         # data from https://www.nvidia.com/en-us/data-center/a100/
         return 312e12
+    elif "A6000" in device_name:
+        # data from https://www.nvidia.com/en-us/design-visualization/rtx-a6000/
```
tianyu-l: I didn't see this number from the link. I got it from https://www.nvidia.com/content/dam/en-zz/Solutions/design-visualization/quadro-product-literature/proviz-print-nvidia-rtx-a6000-datasheet-us-nvidia-1454980-r9-web%20(1).pdf

torchtitan/tools/utils.py
```diff
@@ -91,6 +91,9 @@ def get_peak_flops(device_name: str) -> float:
...
+        return 309.7e12
```
tianyu-l: this number seems assuming "sparsity", whose value should be halved for non-sparsity use cases.

torchtitan/tools/utils.py
Johnsonms: Done, merged into elif "B300" in device_name or "B200" in device_name.

torchtitan/tools/utils.py
Johnsonms: Updated to the PDF datasheet link. Thanks for the correct reference!

torchtitan/tools/utils.py
Johnsonms: Good catch! Fixed — updated to 154.85e12 (309.7 TFLOPS with sparsity halved for dense),  consistent with how other entries like H100 are handled.

PR #2887, opened 2026-04-08
TODO Debt Tracker

### LINKED ISSUE #2866: Track PR/Issue status mentioned in TODOs
Hey there 👋 

It's common to mention links to PR/Issue on which current code rely, like here:

https://github.com/pytorch/torchtitan/blob/0e0590c137599276d36128abc1702efe9e091607/torchtitan/models/common/attention.py#L265-L267

(Right now this PR is closed.)

I think there should be an action that:
- Scans codebase for comments that has TODO and a github link in it
- Pulls their statuses
- Creates an action summary
- Runs on daily basis

Nothing fancy.

So anyone can open actions tab, open the action, view the summary, decide what to do about it.
~(Or just close it and pretend that saw nothing.)~

Andrei-Aksionov:
Hey there 👋 

As mentioned in #2866 it would be nice to track the status of a PR/Issue mentioned in the TODO.

This PR adds exactly that:
1. Scans `.py` files for comments that contain a link to a GitHub
3. Pulls statuses
4. Creates a markdown table, that contains:
    a) **Author**: The author of the comment
    b) **Content**: The body of the comment
    c) **When**: When the linked PR/Issue was Merged, Closed or Opened
    d) **File Location**: The placement of the comment

This table is written into the action summary and into an issue.
(Issue number is hard-coded into the action.)

---

Example of an action summary: [link](https://github.com/Andrei-Aksionov/torchtitan/actions/runs/24256008377)
Example of an issue: [link](https://github.com/Andrei-Aksionov/torchtitan/issues/6)

---

The action is triggered:
- Once a day
- When there is a change in the checker's files
- And can also be triggered manually (once it's landed to the main branch).

---

### Caveats

Since comments with a link to a GitHub PR/Issue will be added to the body of an issue, the aforementioned PR/Issue will contain record that it was mentioned by the GitHub bot.

Fixes #2866

### DIFF
--- a/.github/workflows/todo_debt_tracker.yaml
+++ b/.github/workflows/todo_debt_tracker.yaml
@@ -48,6 +48,7 @@ jobs:
           python .github/scripts/todo_debt_tracker.py > report.md
 
       - name: Update the Issue
+        if: github.event_name != 'pull_request'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ISSUE_NUMBER: 2936


Andrei-Aksionov:
> But on the other hand, in reality if a TODO is left there for long time, it probably means it's low priority.

In general yes. At least when I create a TODO.
Usually it means that the work is done, but it would be nice to implement what is described in TODO sometime in the future.
But when there is a link to a github PR or an issue, as I see it, it should mean that this should have been implemented this way from the get go, but something else has blocked it. So, once this blocking is gone, revisit the code and do it as it was intended initially.

To me it doesn't sound like a low priority.
But I might be wrong.

> While it'd be good if everyone could follow strict rules and always add TODOs in the code, my slight worry is that this tracker will add noise to the current workflow, and I prefer to have a single source of priority list.

I see where your concern comes from. I totally get it.
But right now you have two places which you have to monitor:
1. The project page
2. The issue page: you do your triage of incoming problems and put it in the project page, if needed. So, having a dedicated issue for tracking status of github PR/Issues mentioned in TODO section fits kinda nicely - you can check which PRs/Issues are already closed/merged and perhaps assign the person to it, if not already done. 

------------

Essentially, there are two options.

**Option A**.

Add this feature, keep it for some time.
If within a month or so it doesn't prove it's usefulness - remove it.

**Option B**.

Close the PR.
If one day you'll notice that something was missed (no issue, no record on the project page), and this tracker could have prevented it - we can come back to it.

This is my prefered way of work - solve only existing problem, don't add something just to be there.

The problem in this case is that without this tracker you might not know that you've missed something in the first place 🙃.

tianyu-l:
I think Option A sounds reasonable. I do have a question

> So, having a **dedicated issue** for tracking status of github PR/Issues mentioned in TODO section fits kinda nicely

Is it true that this PR is creating an action which will have a dedicated page that one has to click and check. Will it auto create / update a github issue? That would be more helpful, because we could track that aggregated issue in the project page.

Andrei-Aksionov:
> Is it true that this PR is creating an action which will have a dedicated page that one has to click and check.

Right now it creates an action that is triggered automatically once a day and could be triggered manually, once it's landed on the default (main) branch.

The action will have a summary. You can see how it will look [here](https://github.com/Andrei-Aksionov/torchtitan/actions/runs/24141360517).

> Will it auto create / update a github issue?

Nope, it won't 🙂.

I asked a [question](https://github.com/pytorch/torchtitan/issues/2866#issuecomment-4208459250) whether it makes sense to do so in the issue.

But since you had concerns regarding this tracker in general, I stopped any work on it.

> That would be more helpful, because we could track that aggregated issue in the project page.

I will make changes to the PR tomorrow and ping you to re-review it.

Andrei-Aksionov:
Hey @tianyu-l 

I've made changes.
Please take a look at the updated PR description. 
There are links with examples.

If it looks good enough:
- Create a new label `todo-debt-report`
- Create a new issue and assign this label to it
- Allow this PR to run actions

.github/workflows/todo_debt_tracker.yaml
```diff
@@ -0,0 +1,59 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python .github/scripts/todo_debt_tracker.py > report.md
+
+      - name: Update the Issue
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # 1. Find the issue with the specific label
```
tianyu-l: what if humans create multiple issues with `todo-debt-report` label (maybe by accident)?

tianyu-l: > Why a label?
The action searches for an opened issue with such label.
If one decides to archive (by closing) the issue with the report due to comments contamination, a new issue with such label will give a fresh start for the report.

I would prefer we start simple.
How about
- we just create an issue manually and use this fixed issue number here
- put this issue in the current project dashboard, so no need to pin it

.github/scripts/todo_debt_tracker.py
```diff
@@ -0,0 +1,418 @@
...
+
+    emojis = {"merged": "💜 Merged", "closed": "🔴 Closed", "open": "🟢 Open"}
+    has_any = False
+
+    for status in ("merged", "closed", "open"):
+        if groups[status]:
+            has_any = True
+            output += f"### {emojis[status]}\n"
+            status_title = "opened" if status == "open" else status
+            output += f"| Comment Author | Comment Content | Blocker {status_title.capitalize()} | Comment Location |\n"
```
tianyu-l, line 331: The links in https://github.com/Andrei-Aksionov/torchtitan/issues/6 seem not right

.github/workflows/todo_debt_tracker.yaml
Andrei-Aksionov: Sure.
Will do.

.github/scripts/todo_debt_tracker.py
Andrei-Aksionov: Could you clarify what doesn't seem right?
I manually re-checked rows in the table and links in them are correct.

Perhaps something confusing with the naming?

tianyu-l: oh now it seems correct, maybe I was looking at an old version, sorry

.github/workflows/todo_debt_tracker.yaml
Andrei-Aksionov: Issue is created.
Code is changed.

All set.
Now we need to start the action.

.github/workflows/todo_debt_tracker.yaml
```diff
@@ -0,0 +1,74 @@
...
+on:
+  # 1. Manual trigger
+  workflow_dispatch:
+
+  # 2. Automatic trigger (Once a day at midnight UTC)
+  schedule:
+    - cron: '0 0 * * *'
+
+  # 3. Triggered if the script or the action configuration is changed
+  push:
```
tianyu-l, line 12: change this to be triggered only when pushed to main

also need to add pull_request, e.g. right now it's not triggered

see https://github.com/pytorch/torchtitan/blob/main/.github/workflows/integration_test_8gpu_graph_trainer.yaml for an example

Andrei-Aksionov: Done

tianyu-l:
@claude CI test of the yaml workflow got the following problem at "Update the Issue" step. What can we do?
```
  set -e
  
  echo "Checking if issue #$ISSUE_NUMBER exists..."
  
  if ! STATE=$(gh issue view "$ISSUE_NUMBER" --json state --jq '.state' 2>/dev/null); then
    echo "Error: Issue #$ISSUE_NUMBER does not exist in this repository."
    exit 1
  fi
  
  if [ "$STATE" != "OPEN" ]; then
    echo "Error: Issue #$ISSUE_NUMBER is currently $STATE. "
    exit 1
  fi
  
  if [ ! -s report.md ]; then
    echo "Error: report.md is missing or empty."
    exit 1
  fi
  
  if gh issue edit "$ISSUE_NUMBER" --body-file report.md; then
    echo "Successfully updated issue #$ISSUE_NUMBER"
  else
    echo "Failed to update issue #$ISSUE_NUMBER. Check permissions."
    exit 1
  fi
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.10.20/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.10.20/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.20/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.20/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.20/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.20/x64/lib
    GH_TOKEN: ***
    ISSUE_NUMBER: 2936
Checking if issue #2936 exists...
failed to update https://github.com/pytorch/torchtitan/issues/2936: GraphQL: Resource not accessible by integration (updateIssue)
failed to update 1 issue
Failed to update issue #2936. Check permissions.
```

tianyu-l:
@Andrei-Aksionov 

It makes sense that Pull Request action is not allowed to overwrite the issue. So I think we should remove the `pull_request` trigger in workflow. (apologize for the back and forth).

Since you've already tested in your own fork, it should be fine to land this directly.

Andrei-Aksionov:
Actually, I like the solution proposed by Claude.
PR that changes those files should trigger the action, but shouldn't be able to overwrite the issue.
This way anyone who might want to modify those files at least can see how markdown table would look like in the action summary.
And since the table in the action summary and in the issue are identical, it would be a great proxy.

PR #2916, opened 2026-04-09
[graph_trainer] Add CooR precompile support for DeepSeek V3

bobrenjc93:
* __->__ #2916

Extend the CooR precompile infrastructure to support MoE models
with expert parallelism (EP).

E2E validation (1 node, 8x H100, --debug.deterministic --debug.seed=42):

  DSv3 AOT precompile, DP=4 TP=2 EP=4 ETP=1, 5 steps:
    step 1: loss=7.94874 grad_norm=3.7722
    step 2: loss=6.06074 grad_norm=4.1036
    step 3: loss=5.01055 grad_norm=3.2918
    step 4: loss=4.91436 grad_norm=2.7785
    step 5: loss=4.67803 grad_norm=2.7366
  Result: bitwise identical baseline vs precompile
  MFU: ~0.90% baseline vs ~0.89% precompile (equivalent)


torchtitan/components/forward_utils.py
```diff
@@ -0,0 +1,82 @@
...
+import torch
+import torch.nn as nn
+
+from torchtitan.components.tokenizer import BaseTokenizer
+from torchtitan.distributed import ParallelDims
+from torchtitan.models.common.decoder import Decoder
+from torchtitan.protocols import BaseModel
+
+
+def build_forward_extra_kwargs(
```
SherlockNoMad: good refactoring

torchtitan/distributed/expert_parallel.py
```diff
@@ -153,7 +153,7 @@ def _token_dispatch(
                 num_tokens_per_expert,
                 None,
                 None,
-                group=device_mesh.get_group(),
+                group=device_mesh,
```
SherlockNoMad: seems weird? I didn't check the signature for all_to_all_single tho

torchtitan/distributed/expert_parallel.py
```diff
@@ -303,7 +303,12 @@ def _split_along_first_dim(x: torch.Tensor) -> torch.Tensor:
                     "Requires EP degree dividing batch size * seq len."
                 )
             local_num_tokens = num_tokens // device_mesh.size()
-            local_rank = device_mesh.get_local_rank()
+            # Use _sym_get_coordinate instead of get_local_rank so that
```
SherlockNoMad: This is non trivial requirement on modeling code, and needs go into a precompile documents and skill.md for converting a model into pre-compile ready.

SherlockNoMad (approved):
lgtm on the precompile and graph trainer part. 

need @tianyu-l on the modeling change.

torchtitan/distributed/expert_parallel.py
bobrenjc93: Needed as part of CooR since get_group() would bake the pg into the graph. Passing in device mesh allows us to use the `torch.ops._dtensor.mesh_get_process_group` custom op to be rank agnostic.

torchtitan/distributed/expert_parallel.py
bobrenjc93: Will follow up in another PR

torchtitan/distributed/expert_parallel.py
tianyu-l, line 1: this seems an old version, please rebase

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -318,7 +321,14 @@ def _precompile_aot_fx_trace(
     )
     dummy_global_valid_tokens = float(global_batch_size * seq_len)
     extra_inputs: dict[str, torch.Tensor] = {}
-    extra_kwargs: dict[str, torch.Tensor] = {}
+
+    extra_kwargs = build_forward_extra_kwargs(
```
tianyu-l: I would like to understand the necessity of this change:
- whether it's because you need to make things "functional"
- or because you are not organize CooR as a trainer structure

E.g. if it's the latter, you could directly reuse / override the `post_dataloading_process` method in eager trainer.

torchtitan/trainer.py
```diff
@@ -589,50 +589,17 @@ def post_dataloading_process(
...
+            self.model_config,
+            self.model_parts[0],
+            inputs,
+            positions=positions,
+            tokenizer=self.tokenizer,
+            parallel_dims=self.parallel_dims,
+            extra_inputs=extra_inputs,
+        )
 
         if self.parallel_dims.cp_enabled:
```
tianyu-l, line 630: I'm not convinced why you don't need this. Is it because you are not targeting CP?

tianyu-l (dismissed):

torchtitan/experiments/graph_trainer/precompile_main.py
bobrenjc93: Precompile needs the exact same extra_kwargs as training so that trace time inputs matches runtime inputs. This is primarily a refactoring so we don't duplicate the logic. We can't use post_dataloading_process directly since precompile doesn't instantiate a full Trainer (we don't want dataloader init since for LLMs the fake input is 100% knowable).

torchtitan/trainer.py
bobrenjc93: Not targeting CP with this PR. I added a TODO and an assert and the next PR adds support for CP.

torchtitan/models/llama4/parallelize.py
```diff
@@ -618,7 +618,10 @@ def apply_moe_ep_tp(
             if tp_mesh is not None:
                 if isinstance(dispatcher, AllToAllTokenDispatcher):
                     dispatcher.sp_size = tp_mesh.size()
-                    dispatcher.sp_rank = tp_mesh.get_local_rank()
+                    # Use _sym_get_coordinate so the rank is a SymInt
+                    # under CooR precompile instead of a concrete int
+                    # that gets baked into the FX graph.
+                    dispatcher.sp_rank = tp_mesh._sym_get_coordinate(0)
```
tianyu-l, line 624: I wish to see identical numerics on eager trainer, before and after this change. Please merge after you verify so.

tianyu-l (approved):
one nit

torchtitan/models/llama4/parallelize.py
bobrenjc93: image

https://gist.github.com/bobrenjc93/3f3c60b4fad4104ce762ca259ff55c52

PR #2917, opened 2026-04-09
[GraphTrainer][AutoDev] Add remove_detach_pass graph pass

yiming0416:
## Summary
- Add `remove_detach_pass` graph pass that removes `aten.detach.default` nodes from traced fwd+bwd graphs, since detach is semantically a no-op without an autograd context.
- Register the pass early in `construct_default_graph_passes` (before bucketing/scheduling passes).
- Add comprehensive unit tests covering removal, no-op on clean graphs, numerics preservation, multiple users, and nested detach chains.

## Test plan
- [x] `test_detach_nodes_removed` — verifies detach nodes are removed from a graph containing them
- [x] `test_graph_without_detach_unchanged` — verifies graphs without detach are returned unchanged
- [x] `test_numerics_preserved` — verifies forward outputs match before/after the pass
- [x] `test_detach_with_multiple_users` — verifies all uses of a multi-user detach are replaced
- [x] `test_nested_detach_chain` — verifies chains of consecutive detach nodes are fully removed
- [x] All tests pass: `pytest torchtitan/experiments/graph_trainer/tests/test_passes.py -x`
- [x] Lint passes: `pre-commit run --all-files`

## Board item
Project 164 item: Graph pass: remove_detach_pass

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -44,6 +44,37 @@ from torchtitan.experiments.graph_trainer.reshard_after_forward import (
 from torchtitan.tools.logging import logger
 
 
+def remove_detach_pass(
+    gm: torch.fx.GraphModule, example_inputs=None
+) -> torch.fx.GraphModule:
+    """Remove ``aten.detach.default`` nodes from the graph.
+
+    In a traced fwd+bwd graph there is no autograd context, so detach is
+    semantically a no-op.  Removing these nodes simplifies the graph for
+    downstream passes.
+
+    Args:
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -60,6 +91,7 @@ def construct_default_graph_passes(
         An ordered list of graph passes ready to apply.
     """
     passes: list[Callable] = [
+        remove_detach_pass,
```
SherlockNoMad: move this after tlparse_log_graph_pass

SherlockNoMad (approved):

PR #2922, opened 2026-04-09
[graph_trainer] Fix CUDAGraph warmup to stay on current stream

bobrenjc93:
* __->__ #2922

The warmup phase was running on `_cg_manager.stream` via
`_use_cuda_memory_pool_manager`, which caused NCCL collectives to
execute on a non-default stream. On multi-node IB/RoCE setups this
leads to illegal memory access errors because NCCL expects to run
on the stream it was initialized with.

Fix by passing `torch.cuda.current_stream()` to
`_use_cuda_memory_pool_manager` instead of `_cg_manager.stream`. This
keeps NCCL collectives on the correct stream while still directing
warmup allocations into the graph memory pool (avoiding fragmentation
between eager and graph pools).

Also remove a redundant `torch.cuda.synchronize()` before graph
recording — `torch.cuda.graph.__enter__` already performs a full
device sync. Add `capture_error_mode="thread_local"` for better
error reporting during graph capture.

### DIFF
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -17,7 +17,6 @@ from collections.abc import Callable, Sequence
 from typing import Any
 
 import torch
-from torch._inductor.cudagraph_trees import _use_cuda_memory_pool_manager
 from torch._library.opaque_object import is_opaque_value
 from torch.utils._ordered_set import OrderedSet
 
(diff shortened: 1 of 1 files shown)


bobrenjc93:
@claude

torchtitan/experiments/graph_trainer/cudagraph.py
```diff
@@ -198,15 +201,13 @@ def __call__(self, *args):
...
                 x.data_ptr() if isinstance(x, torch.Tensor) else None for x in args
             ]
-
             self._cudagraph = torch.cuda.CUDAGraph()
-
             with torch.cuda.graph(
                 self._cudagraph,
                 pool=_cg_manager.graph_pool,
                 stream=_cg_manager.stream,
+                capture_error_mode="thread_local",
```
BoyuanFeng, line 209: would this line be sufficient? If we don't have the change above in 
```
            with _use_cuda_memory_pool_manager(
                device, _cg_manager.graph_pool, torch.cuda.current_stream()
            ):
```

bobrenjc93: I don't think so? This just gives better error reporting but the underlying issue is IMAs stemming from the warmup run on the wrong stream (which gets solved with the quoted code). 

Is there a reason you think this one line would be sufficient?

BoyuanFeng: 1. warmup on capture stream has been working with nccl for a while. can we find the root cause why torch.cuda.current_stream() is needed to avoid IMA?
2. if using `torch.cuda.current_stream()`, there would be a high memory overhead due to fragmentation between capture stream and default stream in private pool.

bobrenjc93:
talked offline, will try and get precise profile data first and then we can figure out best workaround

PR #2923, opened 2026-04-09
[LoRA]Add LoRA model converter with dynamic class inheritance

mori360:
### Summary                                                                                                                 
                                                                  
  - Add LoRAConverter model converter that applies low-rank adaptation (LoRA) to all nn.Linear layers in a model          
  - LoRA uses dynamic subclass creation (__class__ swap) to wrap existing Linear layers, preserving compatibility with any nn.Linear subclass (e.g., Float8Linear, FakeQuantizedLinear)                                                           
  - Base model weights are frozen (requires_grad=False), only LoRA adapter weights (lora_a, lora_b) are trainable
  - Patches init_weights on the model to reinitialize LoRA adapters during the meta-device → real-device init flow        
  - Add llama3_debugmodel_lora debug config                                                                               
  - Add converter ordering validation: quantization must come before LoRA 

### Parallelism Compatibility                                                                                               
                                                                                                                          
  FSDP (Data Parallel) — Works out of the box. LoRA adapters (lora_a, lora_b) are registered as child nn.Modules of the wrapped linear, so fully_shard() treats them like any other submodule. Base weights are frozen (requires_grad=False) and LoRA weights are trainable — FSDP only all-gathers/reduces gradients for trainable parameters, so no wasted communication. 
                                                                                                                          
  Pipeline Parallel (PP) — Works. LoRA doesn't change the model's module structure or the number of layers, so pipeline stage splitting is unaffected. Each stage's linear layers independently get LoRA adapters during convert().
                                                                                                                          
  Context Parallel (CP) — Works. CP operates on the attention module's sequence dimension, not on linear layer internals. LoRA adapters participate in the same forward pass as the base linear.                                                                
                                                                                                                        
  Tensor Parallel (TP) — Not supported yet. PyTorch's ColwiseParallel._partition_linear_fn (in                            
  torch/distributed/tensor/parallel/style.py) uses named_parameters(recurse=True), which yields dotted names like
  lora_a.weight from child modules. register_parameter() rejects names containing ".", causing a KeyError. RowwiseParallel
   is safe (it explicitly accesses module.weight and module.bias only). This requires an upstream PyTorch fix to        
  ColwiseParallel._partition_linear_fn — either switching to recurse=False or explicitly accessing
  module.weight/module.bias like RowwiseParallel does. Until then, LoRA should not be used with TP. The debug config
  llama3_debugmodel_lora uses TP degree 1.

  Expert Parallel (EP) — Works in principle. LoRA applies to nn.Linear subclasses, and MoE expert layers use              
  GroupedExperts (not nn.Linear), so LoRA would only wrap the shared/dense linear layers (attention projections, shared
  expert FFN), not the routed experts. EP sharding of the expert dimension is unaffected.        
                                                                                                                          
### Test Plan                                                                                                               
                                                                                                                          
  Unit tests (pytest tests/unit_tests/test_model_converter.py):                                                           
  - test_lora_freeze_and_training: base params frozen, LoRA params trainable, base weights unchanged after 5 training steps                                                                                                                   
  - test_lora_filter_fqns: filter_fqns correctly skips specified modules from LoRA conversion
  - test_lora_before_quantization_raises: converter ordering validation rejects LoRA before quantization 
  - Integration tests, 3D compile test (PP+DP+TP, 8 GPU), 3D compile test (PP+DP+CP, 8 GPU)
  - Loss tests, running full finetune, lora finetune(rank=8), lora finetune(rank=64) on llama3 debugmodel
<img width="1279" height="822" alt="loss_curves" src="https://github.com/user-attachments/assets/d62da84a-0564-4dba-8476-2735f045f8eb" />

### DIFF
--- a/tests/unit_tests/test_model_converter.py
+++ b/tests/unit_tests/test_model_converter.py
@@ -3,8 +3,13 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
-import pytest
+from collections import OrderedDict
 
+import pytest
+import torch
+import torch.nn as nn
+
+from torchtitan.components.lora import LoRAConverter
 from torchtitan.components.quantization.float8 import Float8LinearConverter
 from torchtitan.config import ConfigManager
 from torchtitan.distributed import ParallelDims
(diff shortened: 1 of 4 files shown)


mori360:
@claude review the pr

PR #2928, opened 2026-04-10
Skip parallelize_fn for seed checkpoint creation

weifengpy:
Follow up on CI timeout https://github.com/pytorch/torchtitan/actions/runs/24217297179/job/70700751227 — TestGraphTrainerNumerics hangs during seed checkpoint creation 
                                                                                                                                                                                                                                 
[PR #2900](https://github.com/pytorch/torchtitan/pull/2900) applies fully_shard when world size = 1. This triggered a pytorch side DTensor/nn.init bug: https://github.com/pytorch/pytorch/issues/180088

Beyond pytorch side fix, it also makes sense to skip parallelize_fn for seed checkpoints — nothing from it is needed (AC, compile, nD parallelism, mixed precision, etc.). Seed checkpoints only initialize weights and save.                                                    
   
Also adds a seed checkpoint test with a 30s timeout

### DIFF
--- a/torchtitan/trainer.py
+++ b/torchtitan/trainer.py
@@ -392,17 +392,19 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful, Configurable):
                 color=color,
             )
         else:
-            # apply Tensor/Context/Expert Parallel, activation checkpointing, torch.compile, Data Parallel
-            model = model_spec.parallelize_fn(
-                model,
-                parallel_dims=parallel_dims,
-                training=config.training,
-                model_converters=config.model_converters,
-                parallelism=config.parallelism,
-                compile_config=config.compile,
-                ac_config=config.activation_checkpoint,
-                dump_folder=config.dump_folder,
-            )
+            if not config.checkpoint.create_seed_checkpoint:
+                # Skip parallelize_fn for seed checkpoints — nothing from
+                # it is needed (AC, compile, nD parallelism, mixed precision, etc.).
+                model = model_spec.parallelize_fn(
+                    model,
+                    parallel_dims=parallel_dims,
+                    training=config.training,
+                    model_converters=config.model_converters,
+                    parallelism=config.parallelism,
+                    compile_config=config.compile,
+                    ac_config=config.activation_checkpoint,
+                    dump_folder=config.dump_folder,
+                )
 
             model.to_empty(device=init_device)
             with torch.no_grad():
--- a/tests/integration_tests/__init__.py
+++ b/tests/integration_tests/__init__.py
@@ -24,6 +24,7 @@ class OverrideDefinitions:
     ngpu: int = 4
     disabled: bool = False
     skip_rocm_test: bool = False
+    timeout: int | None = None
 
     def __repr__(self):
         return self.test_descr
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -607,6 +607,18 @@ def build_features_test_list() -> list[OverrideDefinitions]:
             "sft",
             ngpu=2,
         ),
+        OverrideDefinitions(
+            [
+                [
+                    "--checkpoint.enable",
+                    "--checkpoint.create_seed_checkpoint",
+                ],
+            ],
+            "Seed checkpoint creation",
+            "seed_checkpoint",
+            ngpu=1,
+            timeout=30,
+        ),
     ]
 
     return integration_tests_flavors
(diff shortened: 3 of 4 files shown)


torchtitan/trainer.py
```diff
@@ -392,17 +392,20 @@ def __init__(self, config: Config):
...
-                model_converters=config.model_converters,
-                parallelism=config.parallelism,
-                compile_config=config.compile,
-                ac_config=config.activation_checkpoint,
-                dump_folder=config.dump_folder,
-            )
+            if not config.checkpoint.create_seed_checkpoint:
+                # Seed checkpoint skips parallelization: fully_shard is
+                # applied even at world_size=1 for mixed precision, but
+                # seed checkpoints need raw unsharded weights on CPU.
```
tianyu-l: hmm ... why?

weifengpy: good question - I am digging this further. might be related to running DTensor random init op cpu device, without initialing cpu/gloo backend

https://github.com/pytorch/torchtitan/blob/7ef10559fd3145587a7eca890867db8ab2bf0dda/torchtitan/trainer.py#L313-L315

torchtitan/trainer.py
```diff
@@ -313,6 +313,11 @@ def __init__(self, config: Config):
         if config.checkpoint.create_seed_checkpoint:
             init_device = "cpu"
             buffer_device = None
+            # Use gloo/cpu mesh so DTensor collectives (e.g. trunc_normal_)
+            # don't hang on CPU tensors — NCCL requires CUDA tensors.
+            import torchtitan.distributed.parallel_dims as _pd
+
+            _pd.device_type = "cpu"
```
weifengpy: `ParallelDims.build_mesh()` reads `device_type` from `torchtitan.tools.utils` at import time: 

https://github.com/pytorch/torchtitan/blob/7ef10559fd3145587a7eca890867db8ab2bf0dda/torchtitan/distributed/parallel_dims.py#L15

let me know if this patching is enough. otherwise I can change `build_mesh` to accept `device_type` as an explicit argument. That will touch more files

weifengpy: @tianyu-l updated the PR to address nccl hang directly 
but do we actually want fully_shard for seed checkpoint? it feels a little bit wired. my previous solution on skipping parallelize_fn for seed checkpont might still worth consideration

torchtitan/trainer.py
```diff
@@ -313,6 +313,11 @@ def __init__(self, config: Config):
         if config.checkpoint.create_seed_checkpoint:
             init_device = "cpu"
             buffer_device = None
+            # Use gloo/cpu mesh so DTensor collectives (e.g. trunc_normal_)
```
tianyu-l: wait, why there's collective in `trunc_normal` or anything in init?

torchtitan/trainer.py
tianyu-l: Yeah it may be more straighforward to just skip parallelize, because nothing from it is needed in seed checkpoint (including AC, compile, nD parallelism, mixed precision, etc.), but let's figure out why there's this bug.

tianyu-l: > ParallelDims.build_mesh() reads device_type from torchtitan.tools.utils at import time

It sounds tricky. Thanks for debugging.

weifengpy: > Yeah it may be more straighforward to just skip parallelize

updated PR to skip parallelize for seed checkpoint

> but let's figure out why there's this bug

it's actually a infinite while loop in nn.init because RNG does not advance: https://github.com/pytorch/pytorch/issues/180088  . no collective on DTensor random ops (my earlier speculation was wrong)

tianyu-l (approved):
Thanks a lot for investigating the issue beyond the scope of this change!!

not sure why but it seems several cuda CI got skipped

weifengpy:
> not sure why but it seems several cuda CI got skipped

will trigger a dummy PR to see if this is related

weifengpy:
check other PRs - they are also skipped

weifengpy:
> not sure why but it seems several cuda CI got skipped

those might be skipped becuase of draft mode. "Ready for review" + a new commit are both needed

```
 Yes — all GPU integration tests are skipped for draft PRs. They all have:

  if: github.event_name != 'pull_request' || github.event.pull_request.draft == false                                                                                                                                           
   
  This includes:                                                                                                                                                                                                                
  - integration_test_8gpu_features.yaml                                                                                                                                                                                       
  - integration_test_8gpu_models.yaml                                                                                                                                                                                           
  - integration_test_8gpu_graph_trainer.yaml                                                                                                                                                                                  
  - integration_test_8gpu_h100.yaml         
  - All other integration tests                                                                                                                                                                                                 
                               
```

PR #2931, opened 2026-04-10
Fused linear projection within MLP 

wwwjn:


torchtitan/models/common/feed_forward.py
```diff
@@ -52,3 +52,21 @@ def __init__(self, config: Config):
...
+        w13: Linear.Config
+        w2: Linear.Config
+
+    def __init__(self, config: Config):
+        super().__init__()
+        self.w13 = config.w13.build()
+        self.w2 = config.w2.build()
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        w1_out, w3_out = self.w13(x).chunk(2, dim=-1)
```
tianyu-l, line 71: `chunk` may not give great performance, IIRC. Did you benchmark this against an efficient impl, e.g. vllm?

torchtitan/models/llama3/parallelize.py
```diff
@@ -238,11 +239,17 @@ def apply_tp(
...
                 desired_input_layouts=(Replicate(),),
             ),
-            "feed_forward.w1": colwise_parallel(),
-            "feed_forward.w2": rowwise_output_plan,
-            "feed_forward.w3": colwise_parallel(),
         }
 
+        # pyrefly: ignore [missing-attribute]
+        if isinstance(transformer_block.feed_forward, FusedFeedForward):
+            layer_plan["feed_forward.w13"] = colwise_parallel()
```
tianyu-l, line 246: This doesn't sound right. Sharding and stacking happening in the same tensor dim causes problems, e.g. when resharding.

See the trick here https://fburl.com/code/8ou37gda

tianyu-l (changes requested):

torchtitan/models/llama3/parallelize.py
tianyu-l: maybe we should wait until @fegin 's config change, before changing it away from nn.Linear, which is assumed by the current parallelize plan

PR #2932, opened 2026-04-10
[rl] Enable batch-invariant using FSDP to apply mixed precision 

wwwjn:
Using FSDP to apply mixed precision (bfloat16) to match vllm's computation dtype, and achieve bit-wise identity between trainer and generator.


torchtitan/experiments/rl/models/attention.py
```diff
@@ -179,7 +179,14 @@ def forward(
...
         if self.enable_gqa:
-            extra_kwargs["enable_gqa"] = True
+            # enable_gqa kwarg requires a recent PyTorch build; FA3 auto-
+            # detects GQA from the head count mismatch when the kwarg is
+            # absent, so we only pass it when the API supports it.
+            import inspect
+
+            sig = inspect.signature(torch.nn.attention.varlen.varlen_attn_out)
+            if "enable_gqa" in sig.parameters:
+                extra_kwargs["enable_gqa"] = True
```
tianyu-l, line 189: what if enable_gqa is not available in kwargs? If we send in q and kv of different numbers of heads, wouldn't it be wrong?

Like, in the past, if we don't do enable_gqa, we would do repeat_kv beforehands

torchtitan/experiments/rl/models/attention.py
```diff
@@ -179,7 +179,14 @@ def forward(
             extra_kwargs["num_splits"] = 1
 
         if self.enable_gqa:
-            extra_kwargs["enable_gqa"] = True
+            # enable_gqa kwarg requires a recent PyTorch build; FA3 auto-
```
tianyu-l, line 182: we always require latest pytorch, is this worth an exception?

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -39,12 +40,14 @@ def parallelize_qwen3(
...
 ):
     """
-    Apply tensor parallelism to the Qwen3 dense model for RL training/inference.
+    Apply tensor parallelism (and optionally FSDP) to the Qwen3 dense model
+    for RL training/inference.
 
     NOTE: The function signature is intentionally simpler than core torchtitan's
-    parallelize_qwen3 — it only accepts the configs needed for TP.
+    parallelize_qwen3 — it only accepts the configs needed for TP and FSDP.
     TODO: Change to core torchtitan's Qwen3 parallel plan when full DTensor is ready
```
tianyu-l, line 51: let's do this soon to avoid maintaining this function

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -141,12 +137,23 @@ def __post_init__(self):
...
-                #  training enabled by FSDP.
-                if self.trainer.training.dtype != "bfloat16":
+                # Trainer computation must be in bfloat16 to match generator.
+                # This can be achieved either by:
+                #   1. Setting training.dtype="bfloat16" (full bf16 training), or
+                #   2. Using FSDP mixed precision (data_parallel_shard_degree > 1)
+                #      with training.mixed_precision_param="bfloat16" to cast fp32
+                #      weights to bf16 during forward.
+                dp_shard = self.trainer.parallelism.data_parallel_shard_degree
+                fsdp_provides_bf16 = dp_shard > 1 and (
```
tianyu-l, line 147: After https://github.com/pytorch/torchtitan/pull/2900, we are always applying mixed precision via FSDP, even when dp_shard == 1.

Please rebase.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -141,12 +137,23 @@ def __post_init__(self):
             if self.trainer.debug.batch_invariant:
                 if not self.trainer.debug.deterministic:
                     raise ValueError("batch_invariant requires deterministic=True")
-                # TODO: Replace trainer dtype constraint to use mixed
-                #  training enabled by FSDP.
-                if self.trainer.training.dtype != "bfloat16":
+                # Trainer computation must be in bfloat16 to match generator.
+                # This can be achieved either by:
+                #   1. Setting training.dtype="bfloat16" (full bf16 training), or
+                #   2. Using FSDP mixed precision (data_parallel_shard_degree > 1)
```
tianyu-l, line 143: It is definitely worth noting down the rope_cache behavior in both cases for trainer, and for generator. E.g. we can have a BATCH_INVARIANCE.md or BITWISE_NUMERICS.md to put everything related there.

torchtitan/experiments/rl/models/attention.py
wwwjn: Oh sorry, this part is a quick shim and should be removed. Fixed

tianyu-l (approved):

PR #2926, opened 2026-04-10
[graph_trainer] Add CUDA graph kernel annotation pass

yushangdi:
Need https://github.com/pytorch/pytorch/pull/179867  and cuda 13.1 driver/compat 

### Summary

Adds a compiler pass that automatically labels CUDA graph kernels with their originating `nn.Module` path in profiler traces. When enabled, Perfetto/chrome traces show which component each kernel belongs to (e.g., `layers.0.attention.wq`, `layers.0.feed_forward.w1`, `norm`). Both forward and backward kernels are annotated.

No model code changes required — module paths are derived from the existing `nn.Module` hierarchy.

### How it works

1. **`annotate_module_fqns(model)`** walks the `nn.Module` tree and wraps each submodule's `forward` with `torch.fx.traceback.annotate_fn({"module_fqn": fqn})`. This sets `node.meta["custom"]["module_fqn"]` on all FX nodes during tracing — surviving dynamo, `make_fx`, and export tracers. Backward nodes receive the metadata via `_copy_fwd_metadata_to_bw_nodes`.

2. **`insert_kernel_annotations_pass`** reads the `module_fqn` metadata from FX nodes and inserts `mark_kernels()` enter/exit calls at module boundaries in the graph. Bundled with `cudagraph_pass` — runs automatically whenever cudagraph is used. No-op when `cuda-python` or `cuda-compat >= 13.1` is unavailable.

3. **`CUDAGraphWrapper`** captures the annotations during CUDA graph recording when the pass is active, and registers a trace post-processor with the core profiler.

4. **Profiler integration** — `profiling.py` exposes `register_trace_post_processor()` so experiment code can hook into trace export without core importing from experiments. The annotation post-processor merges kernel annotations into the trace automatically.

### Usage

The annotation pass runs automatically in `aot_fx_trace` mode when cudagraph is compatible. Profiler traces come out pre-annotated:

```bash
NGPU=1 MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_debugmodel ./run_train.sh \
    --compile.mode aot_fx_trace \
    --profiling.enable_profiling \
    --profiling.profile_freq 4
```

Open the exported trace at `outputs/.../rank0_trace.json` in https://ui.perfetto.dev to see module labels on each kernel.

### Example output (Llama3 debug model, 6 layers)

```
tok_embeddings                      → embedding gather
layers.0.attention_norm             → RMSNorm
layers.0.attention.wq/.wk/.wv      → Q/K/V projections
[...]

PR #2934, opened 2026-04-10
[GraphTrainer] Support transformer_block_bucketing in aot_fx_trace mode

yiming0416:
## Summary

Enable `transformer_block_bucketing` pass in `aot_fx_trace` compile mode. Previously this only worked in `aot` and `jit` modes because `aot_fx_trace` traces a joint fwd+bwd graph where `nn_module_stack` metadata alone isn't sufficient to correctly identify and separate forward vs backward nodes for bucketing.                                                                                                                                                                                                                                                    

**Root cause:** In the joint graph produced by `aot_fx_trace`, forward and backward collectives coexist. The existing `manual_overlap_bucketing` pass needs to bucket each direction independently — forward collectives in forward execution order, backward collectives in reverse order — but had no way to distinguish them.

**Solution:** Leverage the existing `annotate_fn` mechanism (already used for AC region tagging) to:
  1. Tag each bucket-level submodule's forward with its FQN (`annotate_module_fqns`)
  2. Mark backward nodes with `is_bwd=True` during metadata propagation                                                                                                                                                                                                                
  3. Provide a direction-aware `module_stack_fn` to `manual_overlap_bucketing` that filters by fwd/bwd

A new `joint_transformer_block_bucketing_reordering_pass` applies bucketing in two passes — once for forward nodes, once for backward — so each direction's collectives are bucketed and reordered independently.

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -18,6 +18,18 @@ from torchtitan.distributed import ParallelDims
 from torchtitan.tools.logging import logger
 
 _AC_REGION_ID = "ac_region_id"
+_MODULE_FQN = "module_fqn"
+
+
+def annotate_module_fqns(model: nn.Module) -> None:
+    """Annotate all modules' forward with their fully-qualified names.
+
+    Every named submodule (excluding the root) gets its forward method wrapped
+    with annotate_fn so that FX nodes carry the module_fqn metadata.
+    """
+    for fqn, submodule in model.named_modules():
+        if fqn:  # skip root module
+            submodule.forward = annotate_fn({_MODULE_FQN: fqn})(submodule.forward)
 
 
 def annotate_ac_regions(model: nn.Module) -> None:
--- a/torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
+++ b/torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
@@ -19,6 +19,7 @@ from torchtitan.distributed import ParallelDims
 from torchtitan.distributed.tensor_parallel import maybe_enable_async_tp
 from torchtitan.experiments.graph_trainer.common_utils import (
     annotate_ac_regions,
+    annotate_module_fqns,
     apply_graph_ac,
 )
 from torchtitan.experiments.graph_trainer.compile import apply_compile
@@ -55,6 +56,7 @@ def annotate_deepseekv3(model: GraphTrainerDeepSeekV3Model) -> None:
     )
     MoE.forward = annotate_fn({"EP": "compute"})(MoE.forward)
 
+    annotate_module_fqns(model)
     annotate_ac_regions(model)
 
 
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -273,6 +273,9 @@ class TracedResult:
         num_flat_outputs: Number of flat graph outputs before subclass rewrapping.
         output_subclass_layouts: Subclass unwrap/rewrap metadata for outputs.
         output_spec: Original output pytree spec used during reconstruction.
+        model: The original nn.Module, cached for downstream passes that need
+            module structure (e.g. transformer block bucketing). Only set when
+            traced via :func:`trace_train_step`.
     """
 
     gm: torch.fx.GraphModule
@@ -284,6 +287,10 @@ class TracedResult:
     output_subclass_layouts: dict[int, SubclassLayout]
     output_spec: pytree.TreeSpec
     tensor_input_indices: list[int] = field(default_factory=list)
+    # TODO: Remove model from TracedResult. TracedResult is becoming a
+    # contract between the trainer and pre-compile, so carrying the original
+    # model is not feasible.
+    model: nn.Module | None = None
 
     @property
     def num_static_inputs(self) -> int:
@@ -380,6 +387,7 @@ def minimal_fx_tracer(fn: Callable) -> Callable[..., TracedResult]:
             user_list = pytree.tree_unflatten(list(user_flat), user_args_spec)
             with torch.compiler._patch_autograd_grad():
                 result = fn(state_for_fn, *user_list)
+
             flat_outs, output_spec = pytree.tree_flatten(result)
             num_flat_outputs = len(flat_outs)
             unwrapped_outs, output_layouts = _unwrap_subclasses(flat_outs)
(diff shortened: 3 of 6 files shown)


torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -268,6 +271,7 @@ def _copy_fwd_metadata_to_bw_nodes(fx_g: torch.fx.GraphModule) -> None:
             custom = fwd_node.meta.get("custom")
             if custom:
                 node.meta.setdefault("custom", {}).update(custom)
+            node.meta.setdefault("custom", {})[_IS_BWD] = True
```
SherlockNoMad: we won't need to do this when @tugsbayasgalan land the change in the upstream pytorch. 
backward graph would then have a standardized metadata tag.

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -18,6 +18,31 @@
...
+    result = []
+    for item in bucket_plans:
+        if isinstance(item, list):
+            result.extend(item)
+        else:
+            result.append(item)
+    return result
+
+
+def annotate_module_fqns(model: nn.Module) -> None:
```
SherlockNoMad, line 24: cc @yushangdi, who is also adding the annotation for module path

yushangdi: thx! I can probably just re-use this.

yushangdi: I updated my PR to use the same function now 
https://github.com/pytorch/torchtitan/pull/2926

yushangdi: btw I think this currently doesn't work in the following situation. **The bug only manifests with parameterless modules.**

```
        class Block(torch.nn.Module):
            # The bug only manifests with parameterless modules. 
            def forward(self, x):
                return x + 1

        class Model(torch.nn.Module):
            def __init__(self):
                super().__init__()
                self.a = Block()
                self.b = Block()

            def forward(self, x):
                return self.b(self.a(x))
```

it works for make_fx, but something is wrong with minimal_fx_tracer/trace_train_step


 cc @SherlockNoMad @yiming0416 

I think the `annotate_module_fqns` function is fine. Maybe a fix need to be made in `minimal_fx_tracer/trace_train_step`?

yushangdi: ah, I find the root cause for this, there is a bug in `_copy_fwd_metadata_to_bw_nodes`.  I think it should be fix by @tugsbayasgalan's PR https://github.com/pytorch/torchtitan/pull/2875 

```

  Root cause: _copy_fwd_metadata_to_bw_nodes in make_fx_tracer.py matches nodes by seq_nr. Parameterless ops all get seq_nr=-1, so the second forward
  node was treated as a "backward" node and had its custom metadata overwritten with the first forward node's values.
```

torchtitan/experiments/graph_trainer/deepseek_v3/parallelize.py
```diff
@@ -55,6 +56,7 @@ def annotate_deepseekv3(model: GraphTrainerDeepSeekV3Model) -> None:
     )
     MoE.forward = annotate_fn({"EP": "compute"})(MoE.forward)
 
+    annotate_module_fqns(model)
```
SherlockNoMad, line 59: This is duplicated with Shangdi's, which is fine. 
Shangdi's can land first, and you can rebase.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -146,6 +180,48 @@ def transformer_block_bucketing_reordering_pass(
...
+        insert_overlap_deps=False,
+        module_stack_fn=_make_module_fqn_stack_fn(bwd=False),
+    )
+
+    # Clear manual_bucket_node_type metadata left by the forward pass.
+    # The reordering is already applied via topological sort, but stale
+    # metadata would confuse the second call's _manual_reorder_graph
+    # which iterates all graph nodes regardless of direction.
+    for node in gm.graph.nodes:
+        node.meta.pop("manual_bucket_node_type", None)
```
SherlockNoMad, line 229: this adhoc patch is an indication that the pass is touching something it shouldn't touch in the first place. 

I think you can fix inside manual_overlap_bucketing, so it doesn't leak the implementation here. 
e.g. only touch forward node in the first call

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -51,6 +51,7 @@
 
 def construct_default_graph_passes(
     traced_result: "TracedResult",
+    model: torch.nn.Module,
```
SherlockNoMad: Keep the interface clean if possible. 

1. trace_result has some internal GraphModule, can we use that? 
2. if not, alternative is to cache an instance of model inside TracedResult

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -130,6 +139,31 @@ def autobucketing_reordering_pass(
     return gm
 
 
+def _make_module_fqn_stack_fn(
```
SherlockNoMad: make this inner function of joint_transformer_block_bucketing_reordering_pass

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: or clean up the node.meta when its done.

yiming0416: yes, this should be fixed in pytorch

torchtitan/experiments/graph_trainer/passes.py
yiming0416: I attached an instance of model to `TracedResult`.

We can use the internal GraphModule from `TracedResult` as they has been flattened and lost the module hierarchy info

torchtitan/experiments/graph_trainer/passes.py
yiming0416: Fix in https://github.com/pytorch/pytorch/pull/180579

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -273,6 +273,9 @@ class TracedResult:
         num_flat_outputs: Number of flat graph outputs before subclass rewrapping.
         output_subclass_layouts: Subclass unwrap/rewrap metadata for outputs.
         output_spec: Original output pytree spec used during reconstruction.
+        model: The original nn.Module, cached for downstream passes that need
+            module structure (e.g. transformer block bucketing). Only set when
+            traced via :func:`trace_train_step`.
```
SherlockNoMad, line 278:
```
Only set when traced via :func:`trace_train_step`.
```
I am not a fan of conditional behavior. 

Also TracedResult is now becoming a contract between Trainer and pre-compile, so adding model is not feasible now. 

So this is pointing towards something more fundamental, we probably need a different way to come up with bucketing than `get_transformer_block_buckets(traced_result.model)`. 

Let's land for now, but add to TODO here to remove the `model` field from Traced Result.

SherlockNoMad (approved):

PR #2935, opened 2026-04-10
[AutoDev] Default FlexAttn max_autotune to False

SherlockNoMad:
## Summary
- Changed `max_autotune` from `True` to `False` in `FlexAttention.inductor_configs`
- Changed `coordinate_descent_tuning` from `True` to `False` (only useful when `max_autotune` is enabled)
- Added a comment explaining the recommended workflow

## Why
The recommended workflow is:
1. Run once with `max_autotune=True` to discover good `kernel_options`
2. Set `kernel_options` explicitly in the config
3. Disable `max_autotune` for faster compilation in subsequent runs

Defaulting to `False` avoids unnecessary autotuning overhead for users who have already determined their kernel options, and makes the first-run experience faster for users who haven't yet tuned.

`coordinate_descent_tuning` is a sub-feature of `max_autotune` and does no useful work when `max_autotune` is disabled, so it is also defaulted to `False` to match.

## Test plan
- [ ] Verify that FlexAttn still works correctly with default config (max_autotune=False)
- [ ] Verify that setting max_autotune=True via inductor config override still enables autotuning
- [ ] Confirm no numerical difference when kernel_options are explicitly set

### DIFF
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -274,8 +274,12 @@ class FlexAttention(LocalMapInnerAttention):
         # good kernel_options, then set kernel_options explicitly in the config
         # and keep max_autotune disabled for faster compilation.
         "max_autotune": False,
-        # Sub-feature of max_autotune: applies coordinate descent to further
-        # optimize the kernel configs found by max_autotune's initial search.
+        # When enabled, after max_autotune selects the best kernel config,
+        # coordinate descent iteratively tunes individual parameters (block
+        # sizes, num_warps, num_stages) one at a time -- doubling/halving each
+        # and accepting changes that improve runtime by >0.1%. This can also
+        # run without max_autotune but starts from a weaker baseline config.
+        # See torch/_inductor/runtime/coordinate_descent_tuner.py.
         "coordinate_descent_tuning": False,
         "triton.cudagraphs": False,
     }
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -274,6 +274,8 @@ class FlexAttention(LocalMapInnerAttention):
         # good kernel_options, then set kernel_options explicitly in the config
         # and keep max_autotune disabled for faster compilation.
         "max_autotune": False,
+        # Sub-feature of max_autotune: applies coordinate descent to further
+        # optimize the kernel configs found by max_autotune's initial search.
         "coordinate_descent_tuning": False,
         "triton.cudagraphs": False,
     }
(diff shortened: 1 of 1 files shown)


torchtitan/models/common/attention.py
```diff
@@ -270,8 +270,11 @@ class Config(LocalMapInnerAttention.Config):
         # TODO: turn on wrap_inductor_compiled_regions after PyTorch fix is
         # landed again: https://github.com/pytorch/pytorch/pull/175733.
         "wrap_inductor_compiled_regions": False,
-        "max_autotune": True,
-        "coordinate_descent_tuning": True,
+        # Recommended workflow: run once with max_autotune=True to discover
+        # good kernel_options, then set kernel_options explicitly in the config
+        # and keep max_autotune disabled for faster compilation.
+        "max_autotune": False,
+        "coordinate_descent_tuning": False,
```
tianyu-l, line 283: comment that this is a sub feature of `max_autotune`; would be great if we could comment what it does, too

tianyu-l (approved):

torchtitan/models/common/attention.py
SherlockNoMad: AutoDev: Added a comment above `coordinate_descent_tuning` explaining that it is a sub-feature of `max_autotune` and that it applies coordinate descent to further optimize the kernel configs found by max_autotune's initial search.

SherlockNoMad: look up coordinate_descent_tuning form pytorch/pytorch repo, and explain more what is it and how it works.

SherlockNoMad: AutoDev: Looked up the implementation in `torch/_inductor/runtime/coordinate_descent_tuner.py` and `triton_heuristics.py`. Here's what I found:

- `coordinate_descent_tuning` is an iterative optimization pass that runs after the initial autotuning selects the best kernel config. It tunes individual Triton kernel parameters (block sizes like `XBLOCK`/`BLOCK_M`/`BLOCK_N`, `num_warps`, `num_stages`) one at a time by doubling/halving each value and accepting changes that improve runtime by >0.1%.
- It can run independently of `max_autotune` -- the difference is just the starting config. With `max_autotune` enabled, it starts from the best of the full search; without it, it starts from the single heuristic-chosen config.
- The docstring in `CachingAutotuner.coordinate_descent_tuning()` explicitly calls this out: *"regular autotune only gets one config C1; while max-autotune gets 4 configs C1, C2, C3, C4 and figures out C3 is the best. Then coordinate descent tuning starts from C1 vs C3 respectively."*

Updated the comment to explain what it actually does and referenced the source file.

wwwjn:
@SherlockNoMad  @tianyu-l  This PR causing CI failure , can we revert this PR?

PR #2937, opened 2026-04-10
ChunkedCELoss

wwwjn:
ChunkedCELoss implementation in eager.

## Implementation

When FSDP enabled:
~- separate [norm, lm_head] FSDP group: separate backward (lm_head, decoder) will double fire reduce-scatters on the same FSDP module~ Need https://github.com/pytorch/pytorch/pull/180428 to be landed
- set_reshard_after_forward = false
- set_reshard_after_backward = false
- Reduce-scatter (SUM reduce) fires per-chunk for lm_head weights’ gradients
- h_chunk.grad is populated via plain autograd, collected by GradAccumulator

When TP is enabled:
- If SP is applied, all-gather on TP mesh to be Replicate(). Local tensor shape[B, L/N, D]
- Then Detach
- Apply chunk: chunk on sequence dimension. Local shape: [B, L/N, D]. 
- Calculate CE loss on each chunk sequentially. Loss parallel is applied when calculating each chunk. 

## Memory Profiling

Qwen3-0.6B model FSDP=4

w/o ChunkedLoss:
Screenshot 2026-04-14 at 3 47 00 PM

With ChunkedCELoss:
Screenshot 2026-04-14 at 3 45 44 PM

## Compatibility Test
(llama3 debug model)

TP + no SP

TP + SP

CP 
Screenshot 2026-04-14 at 2 38 59 PM

PP
- Not supported for now.  
- Waiting for https://github.com/pytorch/torchtitan/pull/2972 to be fixed

Compile 
Screenshot 2026-04-14 at 4 14 19 PM


tests/integration_tests/features.py
```diff
@@ -607,6 +607,62 @@ def build_features_test_list() -> list[OverrideDefinitions]:
...
+                    "--training.loss_num_chunks 4",
+                    "--parallelism.tensor_parallel_degree 2",
+                    "--parallelism.context_parallel_degree 2",
+                ],
+            ],
+            "Chunked CE loss with FSDP+TP+CP",
+            "chunked_loss_fsdp+tp+cp",
+            ngpu=8,
+        ),
+        OverrideDefinitions(
```
wwwjn: Need to consolidate to one single compound test

torchtitan/components/loss.py
```diff
@@ -18,7 +19,39 @@
 LossFunction: TypeAlias = Callable[..., torch.Tensor]
 
 
-def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
+class CrossEntropyLoss:
```
wwwjn: Need a cleaner refactor for loss part

torchtitan/components/loss.py
```diff
@@ -28,26 +61,271 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        reference: torch.Tensor,
+        num_chunks: int,
+        *,
+        seq_dim: int = 1,
+        dtype: torch.dtype | None = None,
+    ):
+        """Initialize the gradient accumulator.
+
+        Args:
+            reference: Reference tensor to get shape, device, and dtype from.
```
wwwjn: pass the shape direclty

torchtitan/components/loss.py
```diff
@@ -28,26 +61,271 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            scaled_chunk_loss.backward()
+
+            # Accumulate gradient for this chunk
+            grad_accumulator.add(h_chunk.grad)
+
+            # Free memory before next chunk
+            h_chunk.grad = None
+            del scaled_chunk_loss, chunk_loss, logits
+
+        # Get the accumulated gradient and backward through the decoder.
```
wwwjn: Revisit this FSDP trigger bug

torchtitan/components/loss.py
```diff
@@ -28,26 +61,271 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            h_detached, num_chunks=len(h_chunks), seq_dim=1, dtype=torch.float32
+        )
+
+        total_loss = hidden_states.new_zeros((), dtype=torch.float32)
+
+        for h_chunk, label_chunk in zip(h_chunks, label_chunks):
+            # Use F.linear instead of self.lm_head(h_chunk) to bypass FSDP2's
+            # module forward/backward hooks. This ensures FSDP2's one-shot
+            # backward hooks remain available for the decoder backward below.
+            logits = torch.nn.functional.linear(h_chunk, lm_weight)
```
wwwjn: Refactor, avoid using functional but trigger FSDP hooks at correct time

torchtitan/components/loss.py
```diff
@@ -28,26 +61,271 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        # We use (h * grad).sum() instead of h.backward(grad) because FSDP2's
+        # backward hooks are one-shot and only fire during a proper loss.backward()
+        # chain. h.backward(grad) would attempt a second backward through FSDP
+        # modules whose hooks have been consumed, causing "data not allocated" errors.
+        # The (h * grad).sum() trick creates a scalar loss connected to h's
+        # autograd graph, and its backward properly triggers FSDP2's hooks.
+        accumulated_grad = grad_accumulator.result()
+        assert accumulated_grad.dtype == torch.float32
+
+        decoder_loss = (hidden_states * accumulated_grad.to(hidden_states.dtype)).sum()
```
wwwjn: This too tricky, find how to trigger FSDP hooks explicitly

torchtitan/components/loss.py
```diff
@@ -28,26 +61,271 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        decoder_loss.backward()
+
+        return total_loss / global_valid_tokens
+
+
+class ChunkedCELossFactory:
+    """Factory for creating ChunkedCELoss after model construction.
+
+    Since ChunkedCELoss needs the model's lm_head, and the model is not available
+    at loss builder time, this factory is returned by build_chunked_cross_entropy_loss
```
wwwjn: Remove the build function

tests/integration_tests/features.py
```diff
@@ -629,6 +629,21 @@ def build_features_test_list() -> list[OverrideDefinitions]:
             ngpu=1,
             timeout=30,
         ),
+        # Chunked CE loss tests use --module llama3 --config llama3_debugmodel_chunked_loss
```
tianyu-l: I wonder if we should set this as default, instead of only in one config.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
 
 
-def build_mse_loss(compile_config: CompileConfig, **kwargs):
-    del kwargs  # delete any unused arguments
-    loss_fn = mse_loss
-    if compile_config.enable and "loss" in compile_config.components:
-        logger.info("Compiling the loss function with torch.compile")
-        loss_fn = torch.compile(loss_fn, backend=compile_config.backend)
-    return loss_fn
+class CrossEntropyLoss(Configurable):
```
tianyu-l: should inherit a `BaseLoss` class or `Function` https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/function.py, instead of Configurable directly

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        self._fn: LossFunction = cross_entropy_loss
+        if (
+            compile_config is not None
+            and compile_config.enable
+            and "loss" in compile_config.components
+        ):
+            logger.info("Compiling the loss function with torch.compile")
+            self._fn = torch.compile(self._fn, backend=compile_config.backend)
+
+    def __call__(self, pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
```
tianyu-l: shouldn't `global_valid_tokens` be in the args?

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+    6. Assemble chunk gradients into a full gradient [B, L, D] via GradAccumulator
+    7. Backward through the decoder via hidden_states.backward(accumulated_grad)
+
+    FSDP2 composability:
+        The lm_head's FSDP reshard-after-forward and reshard-after-backward are
+        temporarily disabled during the chunked loop so that the weight stays
+        unsharded across all chunks (avoiding repeated all-gathers). Reduce-scatter
+        fires per-chunk, and FSDP2 accumulates the sharded gradients correctly.
+
+    TP with Loss Parallel:
```
tianyu-l: mention composability with SP

torchtitan/trainer.py
```diff
@@ -415,6 +415,10 @@ def __init__(self, config: Config):
 
             self.model_parts = [model]
 
+        # Set lm_head reference for ChunkedCELoss after model construction
+        if isinstance(self.loss_fn, ChunkedCELoss):
+            self.loss_fn.lm_head = self.model_parts[-1].lm_head
```
tianyu-l: Have a public API in ChunkedCELoss, to set lm_head, e.g. `self.loss_fn.set_lm_head(lm_head)`.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        """Number of chunks to split the sequence into."""
+
+    def __init__(
+        self,
+        config: Config,
+        *,
+        compile_config: CompileConfig | None = None,
+    ):
+        self.num_chunks = config.num_chunks
+        self.loss_fn = CrossEntropyLoss(
```
tianyu-l: why loss_fn in this class, but _fn in others? Also, strange to init this way -- maybe inherit CELoss or just build the loss fn the same way as in CELoss

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            CrossEntropyLoss.Config(), compile_config=compile_config
+        )
+        self.lm_head: nn.Module | None = None
+
+    def __call__(
+        self,
+        pred: torch.Tensor,
```
tianyu-l, line 209: nit: `pred` should be the states after lm_head, so there could be better name here -- but if the inheritance require the same name, I'm OK with it.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        is_fsdp = isinstance(lm_head, FSDPModule)
+
+        # If SP is enabled, hidden states are Shard(1) on the TP mesh.
+        # Redistribute to Replicate before chunking so that the lm_head
+        # receives Replicate input and produces correct Shard(-1) output
+        # for loss parallel.
+        if isinstance(hidden_states, DTensor):
+            placements = hidden_states.placements
+            new_placements = tuple(
+                Replicate() if not isinstance(p, Replicate) else p for p in placements
```
tianyu-l: This is not right after full DTensor rewrite, when CP mesh is in DTensor. Let's pick the TP mesh out and only make that `Replicate`.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        # Detach hidden states to stop gradient propagation at this boundary.
+        h_detached = hidden_states.detach().requires_grad_(True)
+
+        # Split hidden states and labels into chunks along seq dim.
+        # Use .contiguous() to break shared storage from torch.chunk(),
+        # otherwise all chunks share the same underlying memory and the
+        # autograd graph retains references, preventing memory from being freed.
+        h_chunks = torch.chunk(h_detached, num_chunks, dim=1)
+        h_chunks = [c.contiguous().detach().requires_grad_(True) for c in h_chunks]
+        label_chunks = torch.chunk(labels, num_chunks, dim=1)
```
tianyu-l, line 255: without CP mesh in the DTensor, it might be OK; but with CP, this wouldn't work -- you need some local map to disregard the CP Shard(1)

We can leave a TODO to tackle later. cc @fegin

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+
+        # Pre-allocate gradient accumulator in fp32 for numerical stability.
+        grad_accumulator = GradAccumulator(
+            h_detached.shape,
+            device=h_detached.device,
+            dtype=torch.float32,
+            num_chunks=len(h_chunks),
```
tianyu-l:
```suggestion
            num_chunks=num_chunks,
```

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            lm_head.set_reshard_after_backward(False)
+
+        for h_chunk, label_chunk in zip(h_chunks, label_chunks):
+            logits = lm_head(h_chunk)
+
+            chunk_loss = self.loss_fn(logits, label_chunk)
+            total_loss = total_loss + chunk_loss.detach()
+
+            # Scale loss before backward so gradients are properly normalized
+            scaled_chunk_loss = chunk_loss / global_valid_tokens
```
tianyu-l: `global_valid_tokens` should be in loss_fn call args

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        # per-chunk, and FSDP2 accumulates the sharded gradients correctly.
+        if is_fsdp:
+            lm_head.set_reshard_after_forward(False)
+            lm_head.set_reshard_after_backward(False)
+
+        for h_chunk, label_chunk in zip(h_chunks, label_chunks):
+            logits = lm_head(h_chunk)
+
+            chunk_loss = self.loss_fn(logits, label_chunk)
+            total_loss = total_loss + chunk_loss.detach()
```
tianyu-l, line 279: why keeping `total_loss` without dividing `global_valid_tokens`?

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        )
+
+        total_loss = hidden_states.new_zeros((), dtype=torch.float32)
+
+        # Disable FSDP reshard on lm_head to keep weight unsharded across
+        # all chunks, avoiding repeated all-gathers. Reduce-scatter fires
+        # per-chunk, and FSDP2 accumulates the sharded gradients correctly.
+        if is_fsdp:
+            lm_head.set_reshard_after_forward(False)
+            lm_head.set_reshard_after_backward(False)
```
tianyu-l, line 272: do we need this when lm_head is wrapper with other modules together in one bucket, e.g. norm?

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+    """
+
+    def __init__(
+        self,
+        shape: tuple[int, ...],
+        *,
+        device: torch.device | str,
+        dtype: torch.dtype,
+        num_chunks: int,
+        seq_dim: int = 1,
```
tianyu-l: order is strange, should be num_chunks -> seq_dim -> dtype ->

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+
+        accumulated_grad = grad_accumulator.result().to(hidden_states.dtype)
+
+        # Wrap as DTensor if hidden_states is a DTensor (TP enabled)
+        if isinstance(hidden_states, DTensor):
+            accumulated_grad = DTensor.from_local(
+                accumulated_grad,
+                device_mesh=hidden_states.device_mesh,
+                placements=hidden_states.placements,
+            )
```
tianyu-l, line 301: If you pass `hidden_states` to `grad_accumulator`, we can do this transparently in it.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+
+    Backward returns accumulated_grad as the gradient for hidden_states.
+    Autograd then propagates this through the decoder layers automatically —
+    no explicit hidden_states.backward() needed.
+    """
+
+    @staticmethod
+    def forward(
+        ctx,
+        hidden_states: torch.Tensor,
```
tianyu-l: I'm surprised that you don't call `hidden_states.backward(accumulated_grad)` anywhere -- how are you passing the gradient back to the decoder graph?

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+class _DeferredBackward(torch.autograd.Function):
+    """Bridges chunked lm_head backward with decoder backward via autograd.
+
+    Forward takes hidden_states (connected to decoder graph), the accumulated
+    gradient from chunked lm_head backward, and the loss value. Returns the
+    loss value as a differentiable tensor.
+
+    Backward returns accumulated_grad as the gradient for hidden_states.
+    Autograd then propagates this through the decoder layers automatically —
+    no explicit hidden_states.backward() needed.
```
tianyu-l: wait, why this could happen automatically, if the chunked parts are always detached from original graph.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,300 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+
+        # Return a differentiable loss via _DeferredBackward. When
+        # .backward() is called (by the trainer or PP schedule), autograd
+        # calls _DeferredBackward.backward which returns accumulated_grad
+        # as the gradient for hidden_states, propagating through the decoder.
+        loss = total_loss / global_valid_tokens
+        return _DeferredBackward.apply(hidden_states, accumulated_grad, loss)
+
+
+class _DeferredBackward(torch.autograd.Function):
```
tianyu-l: I can see how this could help bridge pytorch PP automatically, but I wonder if this creates difficulty for full fwd+loss+bwd compile in graph_trainer @tugsbayasgalan

torchtitan/protocols/model_spec.py
```diff
@@ -32,14 +31,14 @@ class ModelSpec:
     name: str
     flavor: str
     model: BaseModel.Config
+    loss: Configurable.Config
```
tianyu-l: Let's move this out of model-spec and put in Trainer.Config, which will also make it more consistent with RL

torchtitan/experiments/vlm/model/model.py
```diff
@@ -115,5 +115,5 @@ def forward(
             )
 
         h_BSD = self.norm(h_BSD) if self.norm else h_BSD
-        output = self.output(h_BSD) if self.output else h_BSD
+        output = self.lm_head(h_BSD) if self.lm_head else h_BSD
```
tianyu-l, line 121: skip option?

torchtitan/models/llama3/config_registry.py
```diff
@@ -97,6 +97,14 @@ def llama3_debugmodel_float8() -> Trainer.Config:
     return config
 
 
+def llama3_debugmodel_chunked_loss() -> Trainer.Config:
```
tianyu-l: we should make chunked loss default and (reversely) have a dedicated normal CE loss to be tested in CI

torchtitan/trainer.py
tianyu-l: If there's no lm_head on the local PP rank, then this will be assigning None, which seems the right behavior (but maybe worth commenting out?)

When there is lm_head on this local PP rank, are we sure it's always the **last** model_part which has lm_head? I think it's possible, but I feel we should use the `pp_has_last_stage` flag to explicitly assert.

torchtitan/trainer.py
```diff
@@ -662,10 +666,17 @@ def forward_backward_step(
 
         if parallel_dims.pp_enabled:
             # Pipeline Parallel forward / backward inside step() call
+            is_chunked = isinstance(self.loss_fn, ChunkedCELoss)
+            if is_chunked:
+                self.loss_fn._global_valid_tokens = global_valid_tokens
```
tianyu-l: not valid

torchtitan/trainer.py
```diff
@@ -687,26 +698,32 @@ def forward_backward_step(
...
-                # Rescale PP loss to be "local loss sum / global valid tokens"
-                # because each microbatch could have different number of valid tokens
-                loss = (torch.sum(torch.stack(losses)) / global_valid_tokens).to(
-                    self.device
-                )
+                if is_chunked:
+                    # ChunkedCELoss already returns scaled loss per microbatch
+                    loss = torch.sum(torch.stack(losses)).to(self.device)
+                else:
+                    loss = (torch.sum(torch.stack(losses)) / global_valid_tokens).to(
```
tianyu-l: don't need this divide any more after PP change right?

torchtitan/trainer.py
```diff
@@ -687,26 +698,32 @@ def forward_backward_step(
...
+                    **({"skip_lm_head": True} if is_chunked else {}),
+                )
 
-                # Scale the loss by the inverse of the total weight denominator before backward
-                # This ensures gradients are properly normalized across all microbatches
-                loss = loss_sum / global_valid_tokens
+                if is_chunked:
+                    loss = self.loss_fn(pred, labels, global_valid_tokens)
+                else:
+                    loss = self.loss_fn(pred, labels) / global_valid_tokens
```
tianyu-l: unify both by supporting global_valid_tokens in normal CE loss

torchtitan/trainer.py
```diff
@@ -687,26 +698,32 @@ def forward_backward_step(
...
+            is_chunked = isinstance(self.loss_fn, ChunkedCELoss)
             with self.train_context():
-                pred = model_parts[0](inputs, **extra_inputs, **extra_kwargs)
-                # Compute loss sum (reduction='sum')
-                loss_sum = self.loss_fn(pred, labels)
+                pred = model_parts[0](
+                    inputs,
+                    **extra_inputs,
+                    **extra_kwargs,
+                    **({"skip_lm_head": True} if is_chunked else {}),
```
tianyu-l: nit: use `if isinstance(self.loss_fn, ChunkedCELoss): extra_kwargs["skip_lm_head"]=True`

or simply `extra_kwargs["skip_lm_head"]=isinstance(self.loss_fn, ChunkedCELoss)`

tests/integration_tests/features.py
wwwjn: I have no preference of that, ChunkedLoss would not be needed for small models, we could enable it for large models, eg 32B qwen3 model?

torchtitan/components/loss.py
wwwjn: Fixed, now MSELoss and CrossEntropyLoss will do the rescale before returning, similar to ChunkedCELoss

tests/integration_tests/features.py
tianyu-l: I think it's fine to default to non chunked loss, but enable it in some big models + test.

> ChunkedLoss would not be needed for small models

Is this a fact or guess? The counter-argument in my head was "small models are usually trained on small number of GPUs, by the GPU-poor, so memory will be critical to them".

torchtitan/trainer.py
wwwjn: > If there's no lm_head on the local PP rank, then this will be assigning None

Agreed the current behavior is too vague, let me update

> When there is lm_head on this local PP rank, are we sure it's always the last model_part which has lm_head? I think it's possible, but I feel we should use the pp_has_last_stage flag to explicitly assert.

No matter how we split PP stages and how we assign each stage to PP ranks, the ChunkedCELoss will always be the last module of the whole model, so my assumption is it will be the **last** model_part 

> but I feel we should use the pp_has_last_stage flag to explicitly assert

We can only call `self.loss_fn.set_lm_head(lm_head)` when current rank has last PP stage, when PP is enabled

torchtitan/components/loss.py
wwwjn: I refactored to BaseLoss and inheritance mode, but add a comment to say "pred" should be the hidden-states before lm_head

torchtitan/components/loss.py
wwwjn: Do you mean when [norm, lm_head] are wrapped into a single FSDP Module? 

Yes we still need these to avoid repeated all-gather then reshard of parameters. Previously we need Wei's fix is because of another issue: Partial FSDPmodule forward will mess up the FSDP state machine, which causing the non-forwarded module missing a reduce-scatter during backward.

torchtitan/components/loss.py
wwwjn: With `_DeferredBackward` , decoder graph's backward is implicitly done by autograph.

The process is: 
1. After forward, Loss is returned from ChunkedCELoss  by `return _DeferredBackward.apply(hidden_states, accumulated_grad, loss)`, so loss is a differentiable variable. Because hidden_states is a input to _DeferredBackward.apply.apply(hidden_states), so the autograd graph is: decoder model -> hidden_states -> _DeferredBackward.forward()
2. Trainer call loss.backward(), which actually is calling _DeferredBackward.backward(), then triggers the cascade backward to decoder graph.

I'm doing this because of PP: PP is calling `torch.autograd.backward(loss)`[here](https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/_backward.py#L402) instead of `loss.backward()`, so we can not call `hidden_states.backward(accumulated_grad)` in `ChunkedCELoss.backward()` by simply have a backward() function in class  ChunkedCELoss. If we call `hidden_states.backward(accumulated_grad)` within `ChunkedCELoss.__call__`, it will cause double backward.

torchtitan/components/loss.py
wwwjn: Because hidden_states is a input to `_DeferredBackward.apply.apply(hidden_states)`, so the autograd graph is: decoder model -> hidden_states -> _DeferredBackward. 

hidden_states has requires_grad=True (it's connected to the decoder graph).

result = _DeferredBackward.apply(hidden_states, accumulated_grad, loss)                 
- result.grad_fn = <_DeferredBackwardBackward>  ← autograd added this grad_fn                
- result.item() = loss value   ← from loss.detach().clone() 

So conceptually, we use_DeferredBackward to: 1) returning the loss value during forward, this is for logging purpose, 2) return the `accumulated_grad` as the gradient of `hidden_states` during backward

torchtitan/trainer.py
wwwjn: Oh yes, this would only used for logging purpose. I refactor this togather with CELoss , MSELoss to do the scaling during MSELoss.forward()

torchtitan/components/loss.py
wwwjn: Updated: I find this claim is wrong: `hidden_states.backward(accumulated_grad)` doesn't work with PP calling torch.autograd.backward(loss)[here](https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/_backward.py#L402)

It works with PP with following steps:
1. torch.autograd.backward(loss) is a no-op, because loss is detached
2. The val.grad has already been populated by `hidden_states.backward(accumulated_grad)`. https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/_backward.py#L409-L411

Update current complicated design to what we discussed previously.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,276 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+    ) -> torch.Tensor:
+        return self.fn(pred, labels) / global_valid_tokens
+
+
+class CrossEntropyLoss(BaseLoss):
+    """Cross-entropy loss with sum reduction for token-based normalization."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Function.Config):
+        fn: LossFunction = cross_entropy_loss
```
tianyu-l, line 72: no shouldn't make this `cross_entropy_loss` in the config -- can you imagine creating `loss_config = CrossEntropyLoss.Config(mse_loss)`?

torchtitan/components/loss.py
```diff
@@ -28,26 +30,276 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        # Wrap as DTensor if hidden_states is a DTensor (TP enabled)
+        if isinstance(hidden_states, DTensor):
+            accumulated_grad = DTensor.from_local(
+                accumulated_grad,
+                device_mesh=hidden_states.device_mesh,
+                placements=hidden_states.placements,
+            )
+
+        # Backward through the decoder with accumulated gradients.
+        hidden_states.backward(accumulated_grad)
```
tianyu-l: what happens in PP?

wwwjn: Steps happened on last PP stages:
1. loss = _compute_loss(), which will hit ChunkedCELoss.__call__, so the gradients of parameters and activations on last stage has already been populated . 
2. torch.autograd.backward(loss) -> this is a no-op, because returned loss is detached
3. Collect input's gradients of last stage, which will be sent to previous stage: https://github.com/pytorch/pytorch/blob/main/torch/distributed/pipelining/_backward.py#L411 The `input_values`'s gradient is already populated as we manually trigger the backward.


Non-PP
1. loss = loss_fn(pred, label) returns a detached loss, and already calls the `hidden_states.backward(accumulated_grad)` triggers the backward 
2. loss.backward() which is a no-op

torchtitan/components/loss.py
```diff
@@ -28,26 +30,299 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        )
+
+        total_loss = hidden_states.new_zeros((), dtype=torch.float32)
+
+        # Disable FSDP reshard on lm_head to keep weight unsharded across
+        # all chunks, avoiding repeated all-gathers. Reduce-scatter fires
+        # per-chunk, and FSDP2 accumulates the sharded gradients correctly.
+        if fsdp_enabled:
+            lm_head.set_reshard_after_forward(False)
+            lm_head.set_reshard_after_backward(False)
```
weifengpy, line 272: another option is `set_requires_gradient_sync` to achieve one reduce-scatter after N chunks

for `NGPU=8 CONFIG=llama3_8b ./run_train.sh`, seems one reduce-scateter after N chunks give us the same peak memory, because activation memory is big. for shorter sequence, we might have different answer

anyway, totally up to you if you want to expose set_requires_gradient_sync as an option for user

```
265          total_loss = hidden_states.new_zeros((), dtype=torch.float32)
266  
267          # Disable FSDP reshard on lm_head to keep weight unsharded across
268 -        # all chunks, avoiding repeated all-gathers. Reduce-scatter fires                     
269 -        # per-chunk, and FSDP2 accumulates the sharded gradients correctly.                   
268 +        # all chunks, avoiding repeated all-gathers. Coalesce per-chunk                       
269 +        # grad sync into a single reduce-scatter at the last chunk by                         
270 +        # disabling gradient sync for chunks 1..N-1 — matches the microbatch                  
271 +        # grad-accumulation pattern.                                                          
272          if fsdp_enabled:
273              lm_head.set_reshard_after_forward(False)
274              lm_head.set_reshard_after_backward(False)
275 +            lm_head.set_requires_gradient_sync(False, recurse=False)                          
276  
274 -        for h_chunk, label_chunk in zip(h_chunks, label_chunks):                              
277 +        last_idx = len(h_chunks) - 1                                                          
278 +        for i, (h_chunk, label_chunk) in enumerate(zip(h_chunks, label_chunks)):              
279 +            if fsdp_enabled and i == last_idx:                                                
280 +                lm_head.set_requires_gradient_sync(True, recurse=False)                       
281 +                                                                                              
282              logits = lm_head(h_chunk)
283  
284              chunk_loss_sum = self.fn(logits, label_chunk)
```

wwwjn: Thanks! To understand it, set_reshard_after_forward/backward() will disable the multiple forward all-gather then reshard, but it doesn't batch multiple gradients RS.  set_requires_gradient_sync() disables the gradient sync (RS) for each chunk, and we only have one batched RS at the end of all chunk are finished.

weifengpy: that's correct

one controls AG for parameter, one controls RS for gradient. they are separate

weifengpy: btw, I always see same memory usage for doing 1 RS at the end, tried reducing seq length and debug model

Let me know find a counter example that N RS saves memory

weifengpy: I tried hard to let claude to find a case but seems 1 RS is just strictly better than N RS, and we should alwasy set `set_requires_gradient_sync`

from claude
```
TL;DR of the full memory investigation: I found a config where chunked-loss is the memory peak (tiny backbone
  + big vocab → 5.44 GiB peak dominated by lm_head buffers). Even there, baseline vs batched produces identical
  peak. The caching allocator makes RS-batching memory-neutral in every regime tested.
```

config with chunked-loss as the memory peak
```
110  def _debugmodel() -> Llama3Model.Config:
      111 -    dim = 2048                                                                                
      111 +    dim = 256                                                                                 
      112      n_heads = 16
      113 -    n_layers = 1                                                                              
      114 -    vocab = 128256                                                                            
      113 +    n_layers = 6                                                                              
      114      return Llama3Model.Config(
      115          dim=dim,
      117 -        vocab_size=vocab,                                                                     
      116 +        vocab_size=2048,                                                                      
      117          tok_embeddings=Embedding.Config(
      119 -            num_embeddings=vocab, embedding_dim=dim, param_init=_EMBEDDING_INIT               
      118 +            num_embeddings=2048, embedding_dim=dim, param_init=_EMBEDDING_INIT                
      119          ),
      120          norm=RMSNorm.Config(normalized_shape=dim, param_init=_NORM_INIT),
      121          lm_head=Linear.Config(
      123 -            in_features=dim, out_features=vocab, param_init=_output_linear_init(dim)          
      122 +            in_features=dim, out_features=2048, param_init=_output_linear_init(dim)           
      123          ),
      124          rope=RoPE.Config(
      125              dim=dim // n_heads,
```

wwwjn: Thank you, I changed to set_requires_gradient_sync

tests/integration_tests/features.py
```diff
@@ -629,6 +629,17 @@ def build_features_test_list() -> list[OverrideDefinitions]:
             ngpu=1,
             timeout=30,
         ),
+        # Test standard (non-chunked) CE loss (default is ChunkedCELoss)
+        OverrideDefinitions(
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_ce_loss",
```
tianyu-l: CI resource is limited, how about piggyback on some other test, e.g. TP with loss_parallel

torchtitan/models/flux/trainer.py
```diff
@@ -36,6 +37,10 @@ class Config(Trainer.Config):
         tokenizer: FluxTokenizerContainer.Config = (  # pyrefly: ignore [bad-override]
             field(default_factory=FluxTokenizerContainer.Config)
         )
+        # Overwrite parent class loss (CE → MSE for diffusion)
+        loss: MSELoss.Config = field(  # pyrefly: ignore [bad-override]
```
tianyu-l: technically not needed because you could always specify in config_registry.py, but sounds fine as it's the only loss that's supposed to be used with diffusion models right now.

torchtitan/trainer.py
```diff
@@ -409,6 +410,15 @@ def __init__(self, config: Config):
 
             self.model_parts = [model]
 
+        # Set lm_head reference for ChunkedCELoss after model construction.
+        # - Non-PP: single model part always has lm_head.
+        # - PP: only the last stage has lm_head; other stages have None
+        #   so ChunkedCELoss.set_lm_head is never called for non-last stage.
+        if isinstance(self.loss_fn, ChunkedCELoss):
+            lm_head = self.model_parts[-1].lm_head
+            if lm_head is not None:
```
tianyu-l: if self.pp_has_last_stage: assert self.model_parts[-1] has lm_head and it's not None

torchtitan/trainer.py
```diff
@@ -657,8 +667,14 @@ def forward_backward_step(
...
         )
 
+        if isinstance(self.loss_fn, ChunkedCELoss):
+            extra_kwargs["skip_lm_head"] = True
+
         if parallel_dims.pp_enabled:
             # Pipeline Parallel forward / backward inside step() call
+            self.pp_schedule._loss_kwargs = {
+                "global_valid_tokens": global_valid_tokens,
+            }
```
tianyu-l: This shouldn't be a static field in pp_schedule. Instead, we should pass in `self.pp_schedule.step()`.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,346 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            and "loss" in compile_config.components
+        ):
+            logger.info("Compiling the loss function with torch.compile")
+            self.fn = torch.compile(self.fn, backend=compile_config.backend)
+
+    def __call__(
+        self,
+        pred: torch.Tensor,
+        labels: torch.Tensor,
+        global_valid_tokens: torch.Tensor,
```
tianyu-l: should be optional

torchtitan/components/loss.py
```diff
@@ -28,26 +30,346 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        # Return a differentiable loss via DecoderOutputGradientBackProp. When
+        # .backward() is called (by the trainer or PP schedule), autograd
+        # calls DecoderOutputGradientBackProp.backward which returns accumulated_grad
+        # as the gradient for hidden_states, propagating through the decoder.
+        return DecoderOutputGradientBackProp.apply(
+            hidden_states, accumulated_grad, total_loss
+        )
+
+
+class DecoderOutputGradientBackProp(torch.autograd.Function):
```
tianyu-l:
```suggestion
class _DecoderOutputGradientBackProp(torch.autograd.Function):
```

torchtitan/components/loss.py
```diff
@@ -28,26 +30,346 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            lm_head.set_requires_gradient_sync(False, recurse=False)
+
+        last_idx = len(h_chunks) - 1
+        for i, (h_chunk, label_chunk) in enumerate(zip(h_chunks, label_chunks)):
+            if fsdp_enabled and i == last_idx:
+                lm_head.set_requires_gradient_sync(True, recurse=False)
+
+            logits = lm_head(h_chunk)
+
+            chunk_loss_sum = self.fn(logits, label_chunk)
```
tianyu-l: I doubt if this is correct for TP + loss parallel. Have you tested numerics for TP?

What I think would happen
1. hidden states sharded on vocab dim before entering chunked loss
2. redistribute -> Replicate
3. this `self._placements` is Replicate
4. each chunk has Replicate

However, the following loss should expecte a Shard(vocab_dim) from each chunk.

After you fix this, let's be careful about backward as well.

wwwjn: Yes the TP + loss parallel numerics matches: comparison up to 5 digits.
Screenshot 2026-04-21 at 7 43 02 PM

I don't know if I understand your concern here. The DTensor placements around `lm_head` is:
 ```
 h_chunk: DTensor(Replicate) on TP     ← Input to Chunked Loss. Because we already AG on TP mesh, no matter if SP enabled or not
      ↓
  logits = lm_head(h_chunk)             ← ColwiseParallel hooks fire
      ↓
  logits: DTensor(Shard(-1)) on TP      ← sharded on vocab dim
      ↓
  self.fn(logits, label_chunk)          ← loss parallel receives Shard(-1) ✓
```

Note the lm_head parallel plan: 
```
  "lm_head": ColwiseParallel(
      input_layouts=sp_layout,
      output_layouts=Shard(-1) if enable_loss_parallel else Replicate(),
      use_local_output=not enable_loss_parallel,
  )
```

And because input (`h_chunk`) is already a DTensor (Replicate), not a plain tensor, it won't case a redistribute to `sp_layout` cause ColwiseParallel will only redistribute to `desired_input_layout`

File: torchtitan/components/loss.py (hunk 40 above)
Comment (tianyu-l) [reply torchtitan/components/loss.py]: I think we can keep this required for now.

File: torchtitan/components/loss.py (hunk 42 above)
Comment (tianyu-l) [reply torchtitan/components/loss.py]: oh I think you are right. What I said was not correct.

> And because input (h_chunk) is already a DTensor (Replicate), not a plain tensor, it won't case a redistribute to sp_layout cause ColwiseParallel will only redistribute to desired_input_layout

Oh the Colwise API doesn't sound good -- having `input_layouts` but doesn't match inputs.

The code sounds OK for now, but as we move to (erasable) global spmd, we need the `sp_layout` to match what's in the chunked input, which means the sp_layout needs to depend on whether chunked loss is used. Let's add a TODO. cc @fegin

File: torchtitan/trainer.py [hunk 43]
```diff
@@ -670,40 +691,32 @@ def forward_backward_step(
                         **extra_kwargs,
                         target=targets,
                         losses=losses,
+                        loss_kwargs=loss_kwargs,
```
wwwjn, line 700: Here's the callsite of `loss_kwargs` cc @tugsbayasgalan

torchtitan/trainer.py
```diff
@@ -409,6 +410,22 @@ def __init__(self, config: Config):
...
+        if isinstance(self.loss_fn, ChunkedCELoss):
+            if parallel_dims.pp_enabled:
+                if self.pp_has_last_stage:
+                    lm_head = self.model_parts[-1].lm_head
+                    assert (
+                        lm_head is not None
+                    ), "Last PP stage must have lm_head for ChunkedCELoss"
+                    self.loss_fn.set_lm_head(lm_head)
+            else:
+                lm_head = self.model_parts[-1].lm_head
```
tianyu-l: more readable if using
```suggestion
                lm_head = self.model_parts[0].lm_head
```
also, can assert size == 1 for model_parts (optional)

tianyu-l (approved):
sgtm

torchtitan/components/loss.py
```diff
@@ -28,26 +30,345 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+
+    def __init__(
+        self,
+        reference: torch.Tensor,
+        *,
+        num_chunks: int,
+        seq_dim: int = 1,
+        dtype: torch.dtype,
+    ):
+        from torch.distributed.tensor import DTensor
```
joecummings: Why does this have to be nested?

joecummings: Applies to all of the defensive imports in this file

wwwjn:
CI is failing becuase PyTorch PR [#180428](https://github.com/pytorch/pytorch/pull/180428) is not in 0424 nightly yet. After patched locally, the CI test works

tests/assets/losses/llama3_cuda.txt
wwwjn, line 1: This is needed because we change default loss calculation to chunked loss, which changes the numerics slightly (around 1e-6 level). cc @fegin @tianyu-l

torchtitan/components/loss.py
```diff
@@ -28,26 +30,357 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
 
 
-def build_mse_loss(compile_config: CompileConfig, **kwargs):
-    del kwargs  # delete any unused arguments
-    loss_fn = mse_loss
-    if compile_config.enable and "loss" in compile_config.components:
-        logger.info("Compiling the loss function with torch.compile")
-        loss_fn = torch.compile(loss_fn, backend=compile_config.backend)
-    return loss_fn
+class BaseLoss(ABC, Configurable):
```
wwwjn, line 40: On major change after approved: Make BaseLoss an abstract class, incase of some user forget to pass loss=MSELoss.Config() in `config_registry.py` which will result in calling `BaseLoss.__init__`

torchtitan/components/loss.py
```diff
@@ -28,26 +30,369 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+            mesh = hidden_states.device_mesh
+            if mesh.mesh_dim_names is not None and "tp" in mesh.mesh_dim_names:
+                tp_dim = mesh.mesh_dim_names.index("tp")
+                placements = list(hidden_states.placements)
+                if not isinstance(placements[tp_dim], Replicate):
+                    placements[tp_dim] = Replicate()
+                    hidden_states = hidden_states.redistribute(mesh, tuple(placements))
+
+        # Check if it's training model or validation mode
+        requires_grad = hidden_states.requires_grad
```
wwwjn, line 288: Another major change after approved: Add this `requires_grad` path to bypass backward for validator usage. Now validator's loss function is passed from trainer, which is the same loss function instance trainer used. So I added this short path to skip the backward for validation usecase.

Another way is to switch to CELoss before constructing the validator, however, it doesn't works well with PP. Because the PP schedule has its own internal `_loss_fn` that's set once during construction.  cc @tianyu-l

tianyu-l: Chunked loss should be used regardless of training or validation. It's just that in validation, we don't need to accumulate gradient.

There are two ways to organize the code
1. just like what you are doing right now: put if requires_grad around grad computation and its accumulation
2. parametrize grad computation and accumulation with `requires_grad` so that they are no-ops when the flag is False.

torchtitan/components/loss.py
```diff
@@ -28,26 +30,369 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        h_chunks = torch.chunk(hidden_states, num_chunks, dim=1)
+        label_chunks = torch.chunk(labels, num_chunks, dim=1)
+
+        if requires_grad:
+            # Detach hidden states to stop gradient propagation at this boundary.
+            h_detached = hidden_states.detach().requires_grad_(True)
+            # Use .contiguous() to break shared storage from torch.chunk(),
+            # TODO: When CP mesh is in DTensor, chunking along dim=1 won't work
+            # directly with Shard(1) on CP. Need local_map to operate on local tensors
+            h_chunks = torch.chunk(h_detached, num_chunks, dim=1)
```
tianyu-l: this should be used in validation, so technically shouldn't be inside this branch?

wwwjn: Let me rewrite it to be more clear. Now it's overriding the variable`h_chunks` at line 291

torchtitan/models/common/decoder.py
```diff
@@ -134,7 +134,13 @@ def forward(
             h = layer(h, self.freqs_cis, attention_masks, positions)
 
         h = self.norm(h) if self.norm is not None else h
-        output = self.output(h) if self.output is not None else h
+
+        # _skip_lm_head is an attribute rather than a forward kwarg because PP backward
+        # calls .requires_grad on all stage inputs, which fails on bool kwargs.
+        # TODO: fix PP backward upstream to skip non-tensor inputs
+        if getattr(self, "_skip_lm_head", False):
+            return h
```
wwwjn, line 148: One last major change after approved: We have to put `skip_lm_head` an attributed of the model, instead of a forward kwargs. From design perspective, which method do you think is better? cc @tianyu-l

wwwjn:
@yiming0416  @SherlockNoMad @tugsbayasgalan This PR breaks GraphTrainer CI because we switched all config to use ChunkedCELoss by default. Can you help check if we can make ChunkedCELoss compatible with GraphTrainer? Or you could switch back to non-chunked CE Loss.

torchtitan/models/common/decoder.py
tianyu-l: 1. we should fix PP behavior, but I'm OK with using this workaround to unblock
2. As workaorund, this field should be put in the attribute section of Decoder, instead of always using getattr / setattr.

wwwjn: I put a issue to tracking this change: https://github.com/pytorch/torchtitan/issues/3112 (temporary in torchtitan, bc titan needs to remove `_skip_lm_head` after upstream fix

PR #2942, opened 2026-04-13
[RL] Add `gpu_direct_weight_sync` flag to control whether we use GPU direct or not

daniellepintz:
Add a configurable flag to control whether weight sync between trainer and generator uses GPU-direct RDMA or goes through TorchStore StorageVolumes. Previously this was hardcoded to `is_rdma_available()`.

Defaults to True. Disable with --no_gpu_direct_weight_sync in CLI.

# Test Plan
## With GPU Direct (default):
```
python torchtitan/experiments/rl/simple_grpo_sum_digits.py --module rl --config rl_grpo_qwen3_0_6b
```
```
Weight sync: push=0.010s, total=0.086s
```

## Without GPU direct:
```
python torchtitan/experiments/rl/simple_grpo_sum_digits.py --module rl --config rl_grpo_qwen3_0_6b --no-gpu_direct_weight_sync
```
```
Weight sync: push=0.096s, total=4.109s
```

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -160,9 +160,11 @@ class VLLMGenerator(Actor, Configurable):
         *,
         model_spec: ModelSpec,
         model_path: str,
+        gpu_direct_weight_sync: bool = False,
     ):
         self.config = config
         self.model_spec = model_spec
+        self.gpu_direct_weight_sync = gpu_direct_weight_sync
 
         # Register TorchTitan model with vLLM before any engine creation
         register_model_to_vllm_model_registry(model_spec)
@@ -324,24 +326,20 @@ class VLLMGenerator(Actor, Configurable):
     async def pull_model_state_dict(self, version: int) -> None:
         """Pull latest weights from TorchStore.
 
-        When ``direct_rdma=True``, weights are read directly from the
-        trainer's GPU memory via one-sided RDMA, bypassing StorageVolumes.
-        When ``False``, data is fetched through StorageVolumes (which may
-        themselves use RDMA as their transport internally).
-
-        See ``push_model_state_dict`` for more details on the distinction.
+        When ``direct_rdma=True``, weights are transferred directly from
+        GPU to GPU via one-sided RDMA reads, bypassing StorageVolumes
+        entirely. When ``False``, data goes through StorageVolumes
+        (which may themselves use RDMA as a transport internally).
 
         Args:
             version: New policy version number.
         """
-        from monarch.rdma import is_rdma_available
-
         model_sd = self._get_model().model.state_dict()
         await ts.get_state_dict(
             "model_state_dict",
             user_state_dict=model_sd,
             strict=False,
-            direct_rdma=is_rdma_available(),
+            direct_rdma=self.gpu_direct_weight_sync,
         )
         self.policy_version = version
         logger.debug(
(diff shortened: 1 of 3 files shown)


torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -131,6 +131,11 @@ class Config(Configurable.Config):
         """KL divergence penalty coefficient. When > 0, a frozen reference model
         is built and KL divergence is added to the policy gradient loss."""
 
+        gpu_direct_weight_sync: bool = True
```
tianyu-l, line 134: If we make this a config, how should user determine whether to set it to True or False?
- does it work when trainer / generator spans multiple node?
- does it work when user's hardware doesn't have RDMA capability?
- etc.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -324,24 +326,20 @@ async def generate(
...
-        When ``direct_rdma=True``, weights are read directly from the
-        trainer's GPU memory via one-sided RDMA, bypassing StorageVolumes.
-        When ``False``, data is fetched through StorageVolumes (which may
-        themselves use RDMA as their transport internally).
-
-        See ``push_model_state_dict`` for more details on the distinction.
+        When ``direct_rdma=True``, weights are transferred directly from
+        GPU to GPU via one-sided RDMA reads, bypassing StorageVolumes
+        entirely. When ``False``, data goes through StorageVolumes
+        (which may themselves use RDMA as a transport internally).
```
tianyu-l, line 332: what if this is set to False (`StorageVolumes` "may" use RDMA) but RDMA is not available?

tianyu-l (changes requested):

torchtitan/experiments/rl/simple_grpo_sum_digits.py
daniellepintz: > does it work when trainer / generator spans multiple node?

yes

> does it work when user's hardware doesn't have RDMA capability?

no, TorchStore should throw an error in this case

> how should user determine whether to set it to True or False?

if the user has RDMA on their machine they should generally set it to True (default). if they don't have it they will get an error and then should set it to False. Also, currently gpu direct puts more memory pressure on the machine, so if this becomes an issue it is useful for the user to be able to set it to False.

torchtitan/experiments/rl/actors/generator.py
daniellepintz: then TorchStore will fall back to using gloo or plain RPC calls as a transport. If this comment is confusing lmk and I can remove it

torchtitan/experiments/rl/simple_grpo_sum_digits.py
tianyu-l: How about let's rename this to `weight_sync_backend: Literal["gpu", "cpu"]` or maybe `gpu_direct` / `gpu_rdma` / whatever set of options that make sense?

Also please add docstring on how it works under the hood. E.g.
- more memory pressure if using gpu direct
- when rdma would be used, if available
- what if rdma is not available
- the potential difference between torchcomms backend and monarch backend (right now it's very easy for power users to be confused)
- etc.

torchtitan/experiments/rl/actors/generator.py
tianyu-l: would be good to be more transparent about
- differences among all possible options 
- selection / fall back logic
- how to observe which backend is used for transport

Right now it's too blackbox. Potentially we can expose them to users, with an optional "auto" select mode where torchstore makes the decision.

PR #2943, opened 2026-04-13
[rl] Add torchcomms installation to setup instructions and remove xformers

daniellepintz:
Add torchcomms installation to README instructions, for testing with TorchStore weight sync.
Also remove xformers installation as it does not appear to be used (please correct me if I'm wrong).

# Test Plan
- Create a new env following README

RL training works:
```
python torchtitan/experiments/rl/simple_grpo_sum_digits.py --module rl --config rl_grpo_qwen3_0_6b
```

```
uv pip show torchcomms

Name: torchcomms
Version: 0.2.0.dev20260407+cu128
```


d4l3k (approved):
LGTM

PR #2948, opened 2026-04-13
Minor improvements to the tlparse info 

tugsbayasgalan:
We also want to log the TracedResult artifact which explains what are the state fqns, input layouts, and torch version we are running. Also added extra tlparse entry after all default passes run. 

Screenshot 2026-04-13 at 3 18 47 PM

### DIFF
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -509,7 +509,17 @@ def trace_train_step(fn: Callable) -> Callable[..., TracedResult]:
             with stateless._reparametrize_module(module, state):
                 return fn(module, *user_args)
 
-        return minimal_fx_tracer(_stateless_fn)(extract_module_state(module), *args)
+        traced_result = minimal_fx_tracer(_stateless_fn)(
+            extract_module_state(module), *args
+        )
+
+        # Log the raw traced graph before GraphTrainer constructs default passes.
+        from torchtitan.experiments.graph_trainer.tlparse_utils import (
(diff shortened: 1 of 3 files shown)


torchtitan/experiments/graph_trainer/passes.py
tugsbayasgalan, line 229: Moved this to right after traced_train-step API as it seems more likely to catch the real unmodified graph module.

yiming0416 (approved):

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad, line 229: keep this?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -61,6 +63,34 @@ class SubclassLayout:
     meta: SubclassMeta | None
 
 
+def _tlparse_log_string_artifact(name: str, payload: str) -> None:
```
SherlockNoMad: all changes in this file are non-essential, move it to another file, and import it here. 

Keep this file within 300 lines, so it's header only portable.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -67,7 +67,6 @@ def construct_default_graph_passes(
     from torchtitan.models.common.attention import FlexAttention
 
     passes: list[Callable] = [
-        functools.partial(tlparse_log_graph_pass, graph_name="make_fx_graph_traced"),
```
SherlockNoMad, line 68: don't remove this?

PR #2839, opened 2026-04-03
[TorchFT] Extract TorchFT checkpoint logic from core into experiments/ft/

mori360:
## Summary
  - Move all TorchFT fault-tolerance checkpoint logic out of core `CheckpointManager` into a new `FTCheckpointManager` subclass in                         
  `torchtitan/experiments/ft/checkpoint.py`                                                                                       
  - Extract `_should_purge()` as an overridable hook so the FT subclass can add `participating_rank` guards without duplicating the purge loop             
  - Remove FT-specific test fixtures from core unit tests

## Test plan
Test the main checkpoint by training 50 steps first, save, then load and train the next 50 steps.
```
$ NGPU=4 LOG_RANK=0 ./run_train.sh --checkpoint.enable --checkpoint.interval 50 --training.steps 50 --debug.seed 42 --debug.deterministic  --lr_scheduler.total_steps 100 
[rank0]:[titan] 2026-04-06 10:32:53,780 - root - INFO - step:  1  loss:  8.14122  grad_norm:  1.5181  memory:  0.91GiB(0.95%)  tps: 3,632  tflops: 0.26  mfu: 0.03%
[rank0]:[titan] 2026-04-06 10:32:54,517 - root - INFO - step: 10  loss:  3.58616  grad_norm:  1.3104  memory:  0.96GiB(1.01%)  tps: 193,943  tflops: 13.88  mfu: 1.40%
[rank0]:[titan] 2026-04-06 10:32:55,328 - root - INFO - step: 20  loss:  3.01840  grad_norm:  0.4317  memory:  0.96GiB(1.01%)  tps: 196,353  tflops: 14.06  mfu: 1.42%
[rank0]:[titan] 2026-04-06 10:32:56,103 - root - INFO - step: 30  loss:  2.86744  grad_norm:  0.4638  memory:  0.96GiB(1.01%)  tps: 194,395  tflops: 13.92  mfu: 1.41%
[rank0]:[titan] 2026-04-06 10:32:56,812 - root - INFO - step: 40  loss:  2.86680  grad_norm:  0.3045  memory:  0.96GiB(1.01%)  tps: 241,692  tflops: 17.30  mfu: 1.75%
[rank0]:[titan] 2026-04-06 10:32:57,549 - root - INFO - step: 50  loss:  2.80785  grad_norm:  0.2986  memory:  0.96GiB(1.01%)  tps: 172,155  tflops: 12.32  mfu: 1.25%


[...]

PR #2954, opened 2026-04-13
Add HF adapter checkpoint tests and loss_compare checkpoint resume su…

mori360:
### Summary                                                         
                                                                                                             
  - Add unit tests for the HF save/load adapter integration in CheckpointManager, covering the  sd_adapter.to_hf/from_hf roundtrip that previously had no test coverage
  - Add --test-checkpoint-resume to loss_compare.py for verifying checkpoint save/load doesn't corrupt training state                                                                                             
  
  ### Test plan                                                                                                  
                                                                  
  - pytest tests/unit_tests/test_checkpoint.py -x — all tests pass (including 2 new HF adapter tests)        
  - pre-commit run --all-files — no new lint failures
  - loss_compare.py . . --test-checkpoint-resume --steps=20 --assert-equal — verify checkpoint resume produces identical loss                                                                                    
                                                                                                             
  ### Details                                                                                                    
                                                                  
  New checkpoint tests:                                                                                      
  - test_hf_save_adapter_receives_all_model_keys — verifies to_hf receives all model keys and dcp.save gets HF-keyed state dict                                                                                        
  - test_hf_load_roundtrip_to_hf_then_from_hf — verifies the full load sequence: to_hf (build container) → dcp.load → from_hf → load_state_dict, including call ordering and data flow through each step              
                                                                                                             
  loss_compare.py checkpoint resume:
  - --test-checkpoint-resume runs training in two phases: train to checkpoint interval and save, then resume to final step                                                                                             
  - --test-checkpoint-interval controls the save point (defaults to steps // 2)                              
  - Merged losses from both phases are compared against baseline to verify bit-exact resume

### DIFF
--- a/scripts/loss_compare.py
+++ b/scripts/loss_compare.py
@@ -573,6 +573,8 @@ def run_training(
 
     run_with_realtime_output(full_cmd, log_file, env)
 
+    return log_file
+
 
 # =============================================================================
 # LOG PROCESSING AND ANALYSIS
--- a/scripts/loss_compare.py
+++ b/scripts/loss_compare.py
@@ -156,9 +156,9 @@ def extract_losses_from_tensorboard(
 ) -> dict[int, float]:
     """Extract full-precision loss values from TensorBoard event files.
 
-    Reads all timestamped subdirectories under the TB folder and merges
-    their losses.  Normally there is one subdir (single training run),
-    but checkpoint resume produces two (phase 1 + phase 2).
+    The TB directory is cleared before each run (see ``run_training``), so
+    there is exactly one timestamped subdirectory.  We find it and point
+    ``EventAccumulator`` at it directly.
 
     Args:
         job_dump_folder: The --job-dump-folder value (e.g., "outputs")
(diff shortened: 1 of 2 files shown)


scripts/loss_compare.py
```diff
@@ -992,6 +1005,24 @@ def parse_arguments() -> argparse.Namespace:
...
+    )
+    parser.add_argument(
+        "--test-checkpoint-interval",
+        type=int,
+        default=0,
+        help=(
+            "Checkpoint interval for --test-checkpoint-resume. "
+            "Defaults to steps // 2 if not specified."
+        ),
+    )
```
fegin, line 1025: Do we actually need this? This seems to be not necessary since checkpoint should always guarantee bit-wise parity. So we can just use the default interval?

fegin (commented):
Do we plan to use `loss_compare.py` to do checkpoint integration test or no?

tests/unit_tests/test_checkpoint.py
```diff
@@ -701,6 +701,121 @@ def fake_load(state_dict: dict, checkpoint_id=None):
...
+        to_hf_keys = set(mock_adapter.to_hf.call_args[0][0].keys())
+        self.assertEqual(to_hf_keys, {"weight", "bias"})
+
+        mock_adapter.from_hf.assert_called_once()
+        from_hf_keys = set(mock_adapter.from_hf.call_args[0][0].keys())
+        self.assertEqual(from_hf_keys, {"hf_weight", "hf_bias"})
+
+        mock_set_sd.assert_called_once()
+        loaded_sd = mock_set_sd.call_args[1]["model_state_dict"]
+        self.assertEqual(set(loaded_sd.keys()), {"weight", "bias"})
```
Andrei-Aksionov, line 816: Not like this is super important, but it could be done simpler:
```python
from unittest.mock import ANY
...
mock_adapter.to_hf.assert_called_once_with({"weight": ANY, "bias": ANY})
mock_adapter.from_hf.assert_called_once_with({"hf_weight": ANY, "hf_bias": ANY})
mock_set_sd.assert_called_once_with(
    ANY, model_state_dict={"weight": ANY, "bias": ANY}, options=ANY
)

```

Btw, shouldn't this test also check that `get_hf_storage_reader` was called?

tests/unit_tests/test_checkpoint.py
```diff
@@ -701,6 +701,121 @@ def fake_load(state_dict: dict, checkpoint_id=None):
...
+        manager.save(curr_step=10, last_step=True)
+
+        mock_adapter.to_hf.assert_called_once()
+        self.assertIn("weight", received_keys)
+        self.assertIn("bias", received_keys)
+
+        mock_save.assert_called_once()
+        saved_sd = mock_save.call_args[0][0]
+        self.assertIn("hf_weight", saved_sd)
+        self.assertIn("hf_bias", saved_sd)
```
Andrei-Aksionov, line 749: I guess it's better to do a strict comparison instead of `assertIn`.

```python
mock_adapter.to_hf.assert_called_once_with({"weight": ANY, "bias": ANY})

mock_save.assert_called_once()
self.assertEqual(
    mock_save.call_args.args[0], {"hf_weight": ANY, "hf_bias": ANY}
)
```

(With that `recieved_keys` is no longer needed.)

scripts/loss_compare.py
Andrei-Aksionov: Hello @fegin 
Could you please clarify what did you mean? 😊 

> This seems to be not necessary since checkpoint should always guarantee bit-wise parity.

I haven't noticed anywhere in the code that verifies bit-wise parity.
I assume that you've meant that PyTorch DCP gives that guarantee, but since `CheckpointManager` class is basically a wrapper around it, it might be a good idea to check the correctness via this integration test.

That should also make it easier to give answer to questions like https://github.com/pytorch/torchtitan/pull/2839#pullrequestreview-4059108990 whenever there are significant changes to the checkpointing logic.

PR #2957, opened 2026-04-14
[Qwen3VL] remove local_map for dtensor interpolation

shuhuayu:
Remove the `local_map` workaround in the vision encoder's learned position embedding computation. Pushed a pr (https://github.com/pytorch/pytorch/pull/180311) to make DTensor natively supports `F.interpolate` for both forward and backward.


tianyu-l (approved):
unblock

PR #2963, opened 2026-04-14
[Module][Full DTensor] Config-based sharding infrastructure with llama3 adoption

fegin:
* #2987
* #3113
* #3080
* #2986
* #2969
* __->__ #2963

**Summary**
Replace string-keyed plan dicts passed to `parallelize_module()` with `ShardingSpec` attached to each `Module.Config`. Sharding is co-located with the rest of a submodule's configuration and applied by `Module.parallelize(mesh)` via auto-recursion. 

**NOTE that we have to do some compromise designs due to the backward compatibility issues, these should be removed either full DTensor is applied or at some time in the future** 

**Known TODOs for follow-up:**
   
   - apply_cp_to_forward: This is a workaround to keep the current CP work. Remove once all models use config-based sharding + full DTensor
   - input_layouts: Another workaround. We can remove it once all inputs flow as DTensors 
   - LocalMapInnerAttention: remove entire class once all models use config-based sharding (#2986)
   - parallelize(tp_mesh): pass full SPMD mesh once FSDP/CP also uses DTensor (#2987)
   - _shard_outputs: extend to support nested outputs (tuples/dicts) for multi-tensor forward


torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,131 @@
...
+    """Preserve existing placement — no redistribution."""
+
+    def __init__(self) -> None:
+        # Placement.__init__ requires no args for custom subclasses
+        super().__init__()
+
+
+@dataclass(kw_only=True, slots=True)
+class LocalMapSpec:
+    """Spec for modules computing on local tensors (inner attention).
```
acisseJZhong: nit: inner attention and MoE routed experts.

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Completely dtype-agnostic — quantization (Float8/MXFP8) is orthogonal.
+
+    Attributes:
+        state_shardings: Parameter/buffer placements for ``distribute_tensor``.
+            Outer dict keys are param names.
+            e.g. ``{"weight": {TP: Shard(0)}}`` for colwise.
+        input_layouts: How to annotate plain tensor inputs as DTensors,
+            keyed by ``forward()`` arg name. Used when inputs arrive as
+            plain tensors (e.g., from dataloader or FSDP-only path).
```
acisseJZhong, line 92: did we already enable FSDP full dtensor? remembered seeing Wei's PR somewhere.

fegin: https://github.com/pytorch/torchtitan/pull/2987 will do so. But the key issue is that unless we completely delete non-full-dtensor path, we need `input_layouts` to ensure the legacy workflow is correct.

torchtitan/distributed/context_parallel.py
```diff
@@ -75,6 +80,79 @@ def apply_cp_to_attention_module(
     logger.info("Applied Context Parallel to the model")
 
 
+def apply_cp_to_forward(
+    attention_modules: Sequence[nn.Module],
+    cp_mesh: DeviceMesh,
+) -> None:
```
fegin, line 86: This is a workaround. We will remove it in the future (after fully migrated to DTensor). See the docstring.

torchtitan/distributed/sharding.py
```diff
@@ -0,0 +1,58 @@
...
+
+def set_decoder_sharding_spec(config, loss_parallel: bool) -> None:
+    """Set sharding on tok_embeddings, norm, output — shared by all decoders."""
+    config.tok_embeddings.sharding_spec = ShardingSpec(
+        state_shardings={"weight": {TP: Shard(1)}},
+        input_layouts={"input": {TP: Replicate()}},
+        in_shardings={"input": {TP: Replicate()}},
+        out_shardings={TP: Shard(1)},
+    )
+    config.norm.sharding_spec = sequence_parallel_spec()
```
fegin: TODO: the original code base, SP is decoupled from TP. We lost the ability here. Fix this.

torchtitan/distributed/sharding.py
```diff
@@ -0,0 +1,58 @@
...
+def colwise_spec() -> ShardingSpec:
+    """ColwiseParallel: weight Shard(0), output Shard(-1)."""
+    return ShardingSpec(
+        state_shardings={"weight": {TP: Shard(0)}, "bias": {TP: Shard(0)}},
+        out_shardings={TP: Shard(-1)},
+    )
+
+
+def rowwise_spec(
+    out_shardings: dict | None = None,
```
fegin: TODO: do we actually need this? Is it because optional SP? If it is, should we be more explicit to just say make it a boolean? We know what SP expects the out_shardings is.

torchtitan/models/llama3/sharding.py
```diff
@@ -0,0 +1,81 @@
...
+    - ColwiseParallel outputs DTensor Shard(-1)
+    - RowwiseParallel outputs DTensor Shard(1) (sequence parallel)
+    - SequenceParallel norms keep DTensor
+    - rope_cache is wrapped as DTensor Replicate via in_shardings
+    """
+    TP = MeshDimName.TP
+
+    # Norms: SequenceParallel
+    layer_cfg.attention_norm.sharding_spec = sequence_parallel_spec()
+    layer_cfg.ffn_norm.sharding_spec = sequence_parallel_spec()
```
fegin: TODO: the same as the above comments, sequence_parallel shouldn't be always on.

torchtitan/models/llama3/sharding.py
```diff
@@ -0,0 +1,81 @@
...
+    colwise_spec,
+    rowwise_spec,
+    sequence_parallel_spec,
+    set_decoder_sharding_spec,
+)
+from torchtitan.protocols.sharding import MeshDimName, ShardingSpec
+
+
+def set_llama3_sharding_spec(
+    config,
```
fegin: TODO: can we have a proper typing for this?

torchtitan/models/llama3/sharding.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    if not parallel_dims.tp_enabled:
+        return
+
+    set_decoder_sharding_spec(config, loss_parallel)
+    for layer_cfg in config.layers:
+        _set_llama3_layer_sharding(layer_cfg)
+
+
+def _set_llama3_layer_sharding(layer_cfg) -> None:
```
fegin: TODO: can we have a proper typing for layer_cfg

torchtitan/protocols/model_spec.py
```diff
@@ -44,8 +44,9 @@ class ModelSpec:
     pipelining_fn: Callable | None
     post_optimizer_build_fn: Callable | None
     state_dict_adapter: type[BaseStateDictAdapter] | None
+    set_sharding_spec_fn: Callable | None = None
```
fegin: TODO: move this before pipelining_fn and don't set the default value. Just ask all users to set the value to None if not using the new way to do sharding.

torchtitan/protocols/model_spec.py
```diff
@@ -44,8 +44,9 @@ class ModelSpec:
...
 
 
 @dataclass
 class FaultTolerantModelSpec(ModelSpec):
-    fragment_fn: Callable | None
+    fragment_fn: Callable | None = None
```
fegin: TODO: no default value

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -158,6 +159,16 @@ def pipeline_llm(
...
 
+    # Config-based TP keeps DTensors on the PP boundary. The PipelineStage
+    # strips DTensors to local tensors for P2P send and reconstructs them on
+    # recv by looking up the mesh in its ``_mesh_cache``. Register the TP
+    # mesh so the reconstruction can find it.
+    if parallel_dims.tp_enabled:
+        tp_mesh = parallel_dims.get_mesh("tp")
+        get_mesh_cb: GetMeshCallback = lambda mesh_dim_names, mesh_layout: tp_mesh
+        for stage in stages:
+            stage._mesh_cache._get_mesh_cb = get_mesh_cb
```
tianyu-l, line 170: I haven't figured out why we need this https://github.com/pytorch/torchtitan/pull/2149#discussion_r3113791418

torchtitan/distributed/utils.py
```diff
@@ -44,14 +44,25 @@ def _dist_reduce(
             Defaults to None. If provided, this all_reduce will be called for the extra
             process group, and then the result will be all_reduced for the mesh.
     """
+    skip_mesh_reduce = False
```
tianyu-l, line 47: introduction of this flag doesn't help readability -- can we just have two branches, one for DTensor one for plain Tensor?

torchtitan/models/common/attention.py
```diff
@@ -84,6 +84,13 @@ class LocalMapInnerAttention(Module):
     DTensors **after** all ``forward_hook``s complete.
 
     Placements and device mesh are inferred from the input DTensors.
+
+    TODO: This entire class is a legacy path for models that have not yet
+    adopted config-based sharding. Once all models use config-based
+    sharding, ``local_map`` will be applied statically by
+    ``Module.parallelize()`` via ``LocalMapSpec``, and this class
+    should be removed — subclasses should inherit from ``Module``
+    directly.
```
tianyu-l: putting a placeholder, as i don't understand the comment

torchtitan/protocols/model_spec.py
```diff
@@ -41,6 +41,7 @@ class ModelSpec:
     # signatures elsewhere in the codebase.
     build_loss_fn: Callable
     parallelize_fn: Callable
+    set_sharding_spec_fn: Callable | None
```
tianyu-l, line 44: Oh this is different from what I thought. I had thought that ShardingSpec should be part of `model` config; this function's role should be similar to setting up param init, which can all be done in config registry.

torchtitan/trainer.py
```diff
@@ -249,6 +249,15 @@ def __init__(self, config: Config):
         )
         self.model_config = model_config
 
+        # Fill sharding specs based on parallelism settings (before build)
+        if model_spec.set_sharding_spec_fn is not None:
```
tianyu-l, line 253: this should happen in config registry not in trainer -- trainer should get a model config which already has param init config and sharding config

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,130 @@
...
+from torch.distributed.tensor import Placement
+
+
+class StrEnum(str, Enum):
+    """str + Enum for Python < 3.11 compatibility."""
+
+    pass
+
+
+class MeshDimName(StrEnum):
```
tianyu-l, line 27: why do we need this enum, compared with str / Literal?

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,130 @@
...
+            e.g. ``{"x": {TP: Replicate()}}`` for all-gather.
+            ``None`` means no input redistribution.
+        out_shardings: Desired output placement.
+            e.g. ``{TP: Shard(1)}`` for reduce-scatter to sequence-parallel.
+            ``None`` means no output redistribution.
+        local_map: If set, wraps forward with ``local_map()``.
+    """
+
+    state_shardings: dict[str, NamedPlacement] = field(default_factory=dict)
+    # TODO: Remove once all inputs flow as DTensors (full DTensor regime).
```
tianyu-l, line 107: we shouldn't remove, because in global spmd types (erased), we have to hold the source placements

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,130 @@
...
+    orthogonal.
+
+    Attributes:
+        state_shardings: Parameter/buffer placements for ``distribute_tensor``.
+            Outer dict keys are param names.
+            e.g. ``{"weight": {TP: Shard(0)}}`` for colwise.
+        input_layouts: A workaround to annotate plain tensor inputs as DTensors,
+            keyed by ``forward()`` arg name. Used when inputs arrive as
+            plain tensors (e.g., from dataloader or FSDP-only path).
+            e.g. ``{"x": {TP: Shard(1)}}`` means the plain tensor is a
```
tianyu-l, line 93: `NamedPlacement` makes shard order ambiguous. After global spmd types integration, we can have `TP: V` + `PartitionSpec` (containing shard order info).

Right now I'm OK with implicitly always assuming outer -> inner trivial shard order, but let's add note & TODO explicitly.

The only non-trivial shard order should be FSDP + TP both sharding on dim 0, but I think it doesn't need to be annotated so should be fine.

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,130 @@
...
+            e.g. ``{TP: Shard(1)}`` for reduce-scatter to sequence-parallel.
+            ``None`` means no output redistribution.
+        local_map: If set, wraps forward with ``local_map()``.
+    """
+
+    state_shardings: dict[str, NamedPlacement] = field(default_factory=dict)
+    # TODO: Remove once all inputs flow as DTensors (full DTensor regime).
+    input_layouts: dict[str, NamedPlacement] | None = None
+    in_shardings: dict[str, NamedPlacement] | None = None
+    out_shardings: NamedPlacement | None = None
```
tianyu-l, line 110: similarly, for output, we have to have output source and output dest

tianyu-l: output could also be tuple?

torchtitan/models/llama3/sharding.py
tianyu-l, line 1: Is this llama3 specific? Can we associate it with decoder.py?

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,130 @@
...
+    local_map: LocalMapSpec | None = None
```
tianyu-l, line 111: this seems not applied in this PR, can we introduce it later?

torchtitan/distributed/sharding.py
tianyu-l, line 1: This should be put around common/decoder.py, reason being that they are specific to the model definition. E.g. colwise being shard(0) vs. shard(1) purely depends on the module definition.

torchtitan/distributed/context_parallel.py
```diff
@@ -75,6 +80,79 @@ def apply_cp_to_attention_module(
...
+
+                def cp_forward(q, k, v, **kwargs):
+                    if not isinstance(q, DTensor):
+                        q = DTensor.from_local(q, mesh, placement, run_check=False)
+                    if not isinstance(k, DTensor):
+                        k = DTensor.from_local(k, mesh, placement, run_check=False)
+                    if not isinstance(v, DTensor):
+                        v = DTensor.from_local(v, mesh, placement, run_check=False)
+                    output = orig_fn(q, k, v, **kwargs)
+                    return output.to_local() if isinstance(output, DTensor) else output
```
tianyu-l, line 139: can we use local_map here?

torchtitan/protocols/module.py
```diff
@@ -114,6 +127,209 @@ def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> No
...
+    def parallelize(self, mesh: DeviceMesh) -> None:
+        """Parallelize this module and all Module children recursively.
+
+        For each module with a ``sharding_spec``:
+
+        1. ``distribute_tensor`` on params and buffers per ``state_shardings``.
+        2. Wrap ``self.forward`` with redistribution (+ ``local_map`` if needed).
+
+        The wrapping order is:
+            ``shard_inputs -> [optional local_map] fn -> shard_outputs``.
```
tianyu-l, line 182: you mean reshard inputs, reshard outputs?

torchtitan/protocols/module.py
```diff
@@ -114,6 +127,209 @@ def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> No
...
+        if spec is None:
+            return
+
+        assert mesh.mesh_dim_names is not None, "DeviceMesh must have named dims"
+        mesh_dim_names = mesh.mesh_dim_names
+
+        # Distribute parameters and buffers per state_shardings.
+        # Unconstrained dims default to Replicate for state distribution
+        # (distribute_tensor requires concrete placements).
+        from torch.distributed.tensor import Replicate as _Replicate
```
tianyu-l, line 209: why this indirection?

torchtitan/protocols/module.py
```diff
@@ -114,6 +127,209 @@ def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> No
...
+            # Step 1: Annotate plain tensor as DTensor using input_layouts
+            if not isinstance(value, DTensor) and name in input_layouts:
+                from torch.distributed.tensor import Replicate as _Replicate
+
+                layout = resolve_placements(input_layouts[name], mesh_dim_names)
+                # Plain tensor needs concrete placements; Unconstrained -> Replicate
+                layout = tuple(
+                    _Replicate() if isinstance(p, Unconstrained) else p for p in layout
+                )
+                value = DTensor.from_local(value, mesh, layout, run_check=False)
```
tianyu-l, line 302: do we have concrete use case in torchtitan for this? would prefer if we don't add complexity before it's actually needed.

torchtitan/protocols/module.py
```diff
@@ -114,6 +127,209 @@ def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> No
...
+        if spec.in_shardings is None and spec.input_layouts is None:
+            return args, kwargs
+
+        # Use pre-cached positional arg names (populated in parallelize()) to
+        # merge positional args into a unified kwargs dict.
+        pos_arg_names = [
+            name for name in self._cache_pos_arg_names() if name not in kwargs
+        ]
+        new_kwargs = dict(zip(pos_arg_names, args))
+        new_kwargs.update(kwargs)
```
tianyu-l, line 282: why calling it "new"? should we call it "updated"?

torchtitan/protocols/module.py
```diff
@@ -114,6 +127,209 @@ def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> No
...
+
+        # Distribute parameters and buffers per state_shardings.
+        # Unconstrained dims default to Replicate for state distribution
+        # (distribute_tensor requires concrete placements).
+        from torch.distributed.tensor import Replicate as _Replicate
+
+        def _resolve_state(named_pl):
+            pl = resolve_placements(named_pl, mesh_dim_names)
+            return tuple(
+                _Replicate() if isinstance(p, Unconstrained) else p for p in pl
```
tianyu-l, line 214: why would we have `Unconstrained`? Can we always be explicit?

torchtitan/protocols/model_spec.py
fegin: `ShardingSpec` is a part of `model` config. `set_sharding_spec_fn` is not. config registry will call it based on the ParallelismDim to set the `ShardingSpec` to `model`.

torchtitan/models/llama3/sharding.py
fegin: This one is. There are common sharding helper:
```
    replicate_norm_spec,
    sequence_parallel_spec,
    set_decoder_sharding_spec,
    set_dense_ffn_sharding,
    set_gqa_attention_sharding,
```

We can probably further unify the logic. We can wait until the future PRs to see how common they are.  I think some duplications can help the readability. But it is just a minor personal perference.

torchtitan/protocols/module.py
fegin: Yes. But it is not just `reshard` in our case because currently we also have the plain tensors. So we need to wrap the plain tensor into DTensor. We can change it to `reshard` if that make more sense to you.

torchtitan/protocols/module.py
fegin: fair enough. Let's call it updated_kwargs.

[TODO]: update the variable name to updated_kwargs.

torchtitan/protocols/module.py
fegin: Short answer, yes, any non-full dtensor path needs this. 

Think about our input is a plain tensor, we will need to wrap it into a DTensor, what's the sharding of that tensor? 

**This is a workaround until we have convert everything into full DTensor, but you mentioned that SPMD type may need it**

torchtitan/protocols/sharding.py
fegin: I prefer `enum`. It's more readable and has less chances to produce typo. More importantly, it gives you the typing benefit.  pyre should give you the error before running time errors. `StrEnum` also gives you the `str` ability (compare and repr).

I prefer to 1) make TorchTitan only support 3.11 and 2) change all the str to Enum.

torchtitan/protocols/sharding.py
fegin: Yes, it is used in PR3 (assuming this is PR1). The reason it is not used is that I split 1 PR into 3 PRs.  I prefer to keep this here as it keeps the `parallelize()` logic more complete. Unless you have a strong preference, I personally think it is easier for me to look how the `parallelize()` will look like if I put the local map logic here.

torchtitan/trainer.py
fegin: Do you mean we should put it into `update_from_config()`? We need the parallelism config from traing_config.

torchtitan/distributed/context_parallel.py
fegin: We are doing the opposite of what local_map does. local_map convert DTensors to plain tensors so that users can do logic with plain tensors. And the output of local_map region is DTensor.

Here, the input is a plain tensor and the output is also a plain tensor.

torchtitan/distributed/pipeline_parallel.py
fegin: I'll remove this once #2149  merges. Let's move the discussion there.

torchtitan/distributed/utils.py
fegin: [TODO:] Change the logic to have a dedicate branch to handle DTensor.

torchtitan/models/common/decoder_sharding.py
fegin: As long as there are no circular dependency, I'm fine.
[TODO:] Move this as common/decoder_sharding.py

torchtitan/protocols/module.py
fegin: [TODO:] Change the comment to reshard inputs, ..., reshard outputs.

torchtitan/protocols/module.py
fegin: Okay, I renamed to Keep and narrowed to the activation flowing contexts where "preserve existing placement" is a meaningful semantic. `state_sharding` default to `Replicate` if not specificed.

torchtitan/distributed/utils.py
```diff
@@ -45,15 +45,29 @@ def _dist_reduce(
...
+        # for Partial placements and is a no-op for Replicate. Skipping the
+        # subsequent mesh all-reduce is required to avoid double-counting.
+        # Shard placements are not supported — semantics are undefined since
+        # the reduction target is ambiguous.
+        assert all(p.is_replicate() or p.is_partial() for p in x.placements), (
+            f"_dist_reduce received a DTensor with unsupported placements "
+            f"{x.placements}; only Replicate/Partial are supported."
+        )
+        if extra_pg is not None:
+            raise ValueError(
```
tianyu-l, line 58: technically we could local_map this? not sure if ft will be blocked o/w

torchtitan/models/common/decoder_sharding.py
```diff
@@ -0,0 +1,198 @@
...
+
+def dense_param_placement(*, tp: Placement) -> NamedPlacement:
+    """Placement for dense-path params/buffers.
+
+    DP/CP dims are ``Replicate`` at ``distribute_tensor`` time; FSDP reshards
+    DP_SHARD post-parallelize. TP placement is caller-specified.
+    """
+    return {
+        DP_REPLICATE: Replicate(),
+        DP_SHARD: Replicate(),
```
tianyu-l, line 31: Ideally we define on computation mesh which only has DP and does not differentiate DP_REPLICATE vs. DP_SHARD. This requires a mesh switch between fsdp storage and computation. Fine to move forward and fix later (potentially with flex_shard).

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,144 @@
...
+            arg name. Used to annotate plain tensors as DTensors via
+            ``DTensor.from_local`` when inputs arrive plain (e.g. from
+            dataloader or FSDP-only path). Also declares the src side of
+            the input redistribute pair.
+            e.g. ``{"x": {TP: Shard(1)}}``.
+        in_dst_shardings: Desired input placements after redistribution,
+            keyed by ``forward()`` arg name.
+            e.g. ``{"x": {TP: Replicate()}}`` for all-gather.
+            ``None`` means no input redistribution.
+        out_dst_shardings: Desired output placement after redistribution.
```
tianyu-l, line 112: also need out_src with spmd_types, add a todo if you don't want it now

default.profraw
tianyu-l, line 1: what's this, lol

fegin: All mighty Claude secretly add this. I don't even know why it exists, lol. You reviewed too fast, I already removed it, lol.

torchtitan/distributed/utils.py
fegin: yup, we can technically use local_map. But it is TorchFT, which currently no one is maintaining it. I would ignore it as long as the CI is not broken.

torchtitan/models/common/decoder_sharding.py
fegin: If we want to remove DP_SHARD and DP_REPLICATE, we will need FlexShard? That's my understanding. We can add a TODO to revisit this.

torchtitan/protocols/sharding.py
fegin: Claude forgot, lol: `No — I forgot. Adding it now as part of PR1:`.

torchtitan/models/common/decoder_sharding.py
```diff
@@ -0,0 +1,203 @@
...
+
+def replicate_norm_spec() -> ShardingSpec:
+    """Plain-TP norm (no SP): weight Replicate, activations pass through.
+
+    Needed so the norm's weight becomes a DTensor alongside DTensor
+    activations; otherwise we'd mix plain Tensor and DTensor inside the op.
+    """
+    return ShardingSpec(
+        state_shardings={"weight": dense_param_placement(tp=Replicate())},
+    )
```
tianyu-l: maybe more consistent if we combine them into `norm_spec(sp:bool)`

torchtitan/models/common/decoder_sharding.py
```diff
@@ -0,0 +1,203 @@
...
+    feed_forward_cfg.sharding_spec = ShardingSpec(
+        in_src_shardings={"x": dense_activation_placement(tp=attn_x_placement)},
+        in_dst_shardings={"x": dense_activation_placement(tp=Replicate())},
+    )
+    feed_forward_cfg.w1.sharding_spec = colwise_spec()
+    feed_forward_cfg.w3.sharding_spec = colwise_spec()
+    feed_forward_cfg.w2.sharding_spec = rowwise_spec(output_sp=enable_sp)
+
+
+def set_decoder_sharding_spec(config, *, loss_parallel: bool, enable_sp: bool) -> None:
```
tianyu-l: add comment that this excludes per layer sharding

torchtitan/models/llama3/__init__.py
```diff
@@ -379,6 +380,7 @@ def model_registry(
     attn_backend: str = "sdpa",
 ) -> ModelSpec:
     config = llama3_configs[flavor](attn_backend=attn_backend)
+    set_llama3_sharding_spec(config, loss_parallel=True, enable_sp=False)
```
tianyu-l: can we not set it here? aka can we make ShardingSpec optional in a module / config, and error out if it's used inside `.parallelize(mesh)`?

Setting it here doesn't take any effect anyways, because it will be override in `update_from_config`.

torchtitan/models/llama3/parallelize.py
```diff
@@ -68,39 +58,23 @@ def parallelize_llama(
...
             [block.attention.inner_attention for block in model.layers.values()],
             parallel_dims.get_mesh("cp"),
         )
 
+    # TODO: We pass tp_mesh here because TP is the only parallelism
+    # using DTensor currently. Once we move to full DTensor (e.g.,
+    # FSDP via DTensor, CP via DTensor), pass the full SPMD mesh instead.
+    if parallel_dims.tp_enabled:
+        tp_mesh = parallel_dims.get_mesh("tp")
+        model.parallelize(tp_mesh)
```
tianyu-l, line 75: Sounds OK for now. Eventually, I hope we can move this to trainer.py and delete per-model `parallelize_fn` in `ModelSpec`

torchtitan/models/llama3/sharding.py
tianyu-l, line 1: maybe need to update `torchtitan/models/README.py` since we are modifying folder structure -- can defer to later when the stack is merged

torchtitan/protocols/module.py
```diff
@@ -72,7 +84,24 @@ def init_states(
...
 
         self._init_self_parameters()
+
+        # Buffers declared in state_shardings must remain DTensors after
+        # _init_self_buffers reassignment; snapshot placements and restore.
+        dtensor_meta = {
+            name: (buf.device_mesh, buf.placements)
+            for name, buf in self._buffers.items()
+            if isinstance(buf, DTensor)
+        }
```
tianyu-l: I'm confused by what this is trying to do.
- this code is getting dtensor meta
- the code below is assigning dtensor

How could we get before set?

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,148 @@
...
+    in_placements: tuple[NamedPlacement, ...]
+    out_placements: tuple[NamedPlacement, ...]
+    in_grad_placements: tuple[NamedPlacement, ...]
+
+    def to_dict(self) -> dict:
+        return {"repr": repr(self)}
+
+
+@dataclass(kw_only=True, slots=True)
+class ShardingSpec:
```
tianyu-l: I wonder if we should rename to `ShardingConfig` to be consistent within the repo; similar for `LocalMapConfig`

torchtitan/models/llama3/parallelize.py
fegin: We need to do this after all models are migrated, no exceptions. But that's the ultimate goal.

torchtitan/protocols/module.py
fegin: The issue is that:
1) create states on Meta device, 2) shard the buffers and parameters, so they are DTensor, 3) init_states, which parameters are done in-place but buffers are done using `assignment`. So we lose the DTensor or spmd_type annotation in this step.

We can get the dtensor_meta because during `init_states` we already shard the buffers. This ensure that buffers are replicated. I should add a TODO that this only work for buffers that are replicated.

torchtitan/protocols/sharding.py
fegin: Sounds good.

torchtitan/models/llama3/__init__.py
fegin: The reason why I added here is to ensure that if users change the trainer, there is still a default value. But I guess it is a premature decision. Let's remove this.

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,154 @@
...
+from torch.distributed.tensor import Placement
+
+
+class StrEnum(str, Enum):
+    """str + Enum for Python < 3.11 compatibility."""
+
+    pass
+
+
+class MeshDimName(StrEnum):
```
tianyu-l, line 27: the convention in jax and spmd_types is
- mesh axis
- tensor dim

so if it's OK to you, let's follow convention.

torchtitan/protocols/sharding.py
```diff
@@ -0,0 +1,154 @@
...
+self-documenting and support multi-dim meshes.
+"""
+
+from dataclasses import dataclass, field
+from enum import Enum
+
+from torch.distributed.tensor import Placement
+
+
+class StrEnum(str, Enum):
```
tianyu-l, line 21: this may need to be put in e.g. protocols/types.py later

torchtitan/protocols/module.py
```diff
@@ -72,7 +81,31 @@ def init_states(
...
+                continue
+            replicated = DTensor.from_local(
+                new_buf, mesh, (Replicate(),) * mesh.ndim, run_check=False
+            )
+            persistent = name not in self._non_persistent_buffers_set
+            self.register_buffer(
+                name,
+                replicated.redistribute(mesh, list(placements)),
+                persistent=persistent,
+            )
```
tianyu-l: Partial is supported by distribute_tensor, so let's use that to simplify a bit 
https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_api.py#L950

tianyu-l (approved):
lgtm, final comments

torchtitan/protocols/module.py
```diff
@@ -114,6 +147,196 @@ def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> No
...
+    def _shard_inputs(
+        self,
+        mesh: DeviceMesh,
+        args: tuple,
+        kwargs: dict,
+    ) -> tuple[tuple, dict]:
+        """Redistribute inputs to desired placements.
+
+        Two-step process per input:
+        1. If plain tensor, wrap as DTensor using ``in_src_shardings``
```
tianyu-l, line 273: 1. do we need this?
2. with spmd_types, there should be assertion. Up to you when to introduce this assertion -- may need to add a couple of explicit redistribute

torchtitan/protocols/sharding.py
```diff
@@ -78,7 +78,7 @@ def to_dict(self) -> dict:
 
 
 @dataclass(kw_only=True, slots=True)
-class ShardingConfig:
+class ShardingSpec:
```
tianyu-l: curious why changing config back to spec?

torchtitan/protocols/module.py
fegin: 1. Yes for non-full dtensor BC.
2. Later.

torchtitan/protocols/sharding.py
fegin: sounds good, will do that later.

torchtitan/protocols/sharding.py
fegin: sounds good, I'll change this in a later PR.

PR #2945, opened 2026-04-13
Update pytorch nightly to cu13

tugsbayasgalan:
CUDA 12.8 is deprecated so we need 13 to get new nightlies in TorchTitan

PR #2967, opened 2026-04-14
[CI] Upgrade CUDA nightly from cu128 to cu130

weifengpy:
The cu128 nightly index is lagging ~6 days behind cu130, causing CI to test against stale PyTorch builds.


weifengpy:
close because it' duplicate ot https://github.com/pytorch/torchtitan/pull/2945

PR #2968, opened 2026-04-14
Enable graph trainer passes for agnostic  accelerators

jemitche1:
This PR keeps the existing CUDA path and adds a non-CUDA fallbacks for graph trainer passes to avoid scheduler estimation/alignment issues.


torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -386,7 +386,7 @@ def _build_deepseek_v3_tests() -> list[OverrideDefinitions]:
             "jit_fsdp+ep",
             ngpu=8,
             disabled=_JIT_AOT_DISABLED,
-        ),
+        ),_``
```
xmfan: typo?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -233,10 +233,73 @@ def autobucketing_reordering_pass(
...
+        if original_estimate_runtime_analytical is not None:
+            overlap_scheduling.estimate_runtime_analytical = lambda node: 1e-3
+        if original_log_compute_estimations is not None:
+            node_runtime_estimation._log_compute_estimations = (
+                lambda compute_nodes, benchmarked_estimations, analytical_estimations: None
+            )
+        if scheduler_cls is not None and original_align is not None:
+            scheduler_cls._align_compute_nodes_runtime_estimations_across_all_distributed_ranks = (
+                lambda self: None
+            )
```
xmfan: This seems much more like a change that should happen upstream in OverlapScheduler, and patching it from here is brittle. So consider making this change there.

If not, please all least extract all these out into a helper, it's okay to assume the OverlapScheduler will contain those methods since refactors upstream will break this code anyways

jemitche1: Agreed; this belongs in overlap scheduler rather than patched in torchtitan. I removed the duplicated inline patching and created a PR to torch. Once it's merged, I will ping for further review.

PR #2969, opened 2026-04-14
[Module][Full DTensor]  Config-based sharding for Qwen3, Llama4, DeepSeek V3, GPT-OSS

fegin:
* #2987
* #3113
* #3080
* #2986
* __->__ #2969
* #2963

Convert dense-layer TP and attention for all remaining models to config-based `ShardingSpec`, leaving MoE part unchanged, still using `apply_moe_ep_tp()`. After this PR, every model's attention uses `Module.parallelize()`.

**We will remove LocalInnerAttention in the next PR.**
**Models that are not converted: Flux and qwen3-vl.**

1. Flux doesn't support TP. So it is out of the current scope. We can convert it during the full dtensor PRs.
2. qwen3-vl vision part is hard to convert without full dtensor support. It is doable but let's not do the hacky way and wait until the full dtensor PRs.


torchtitan/models/README.md
```diff
@@ -24,9 +24,14 @@ The folder should be organized as follows
...
+- `sharding.py`
+  - Define `set_<model>_sharding_spec(config, *, loss_parallel, enable_sp, ...)` that populates `sharding_spec` on each `Module.Config` in the model config (embeddings, norms, attention, feed-forward, output). TP, SP, and inner-attention `LocalMapConfig` placements are expressed declaratively via `ShardingConfig` instead of a runtime `parallelize_module` plan.
+  - Call the helper from `Model.Config.update_from_config()` so placements depend on the trainer's `parallelism` settings.
+  - Reuse shared helpers from `torchtitan/models/common/decoder_sharding.py` (`set_decoder_sharding_spec`, `set_dense_ffn_sharding`, `set_gqa_attention_sharding`, `norm_spec`, `dense_param_placement`, `dense_activation_placement`) where possible.
 - `parallelize.py`
   - apply training techniques in the following order
-    - TP (and EP if the model has MoE architecture)
+    - `model.parallelize(mesh)` — auto-recursive declarative sharding driven by `sharding_spec` (TP, SP, attention `local_map`). Replaces per-model `parallelize_module` plan dicts.
+    - (MoE models) `apply_moe_ep_tp` for expert-parallel + TP on MoE experts (not yet config-based).
```
tianyu-l, line 34: Right now the support of EP + ETP should be poor in perf, any no one is doing that really. So I think we should
- move the ETP > 1, EP=1 code to sharding.py
- deprecate the `ExpertTensorParallel` class. 
- rename  `apply_moe_ep_tp` to `apply_ep`, which itself should be killed after @acisseJZhong 's refactoring. The local_map application around `GroupedExperts` should also happen in set sharding.
 
Can be done in separate PR.

tianyu-l (approved):
didn't check actual plan

torchtitan/models/README.md
fegin: Sparse related change will come later. I only focus on the dense part. Mesh selection needs to be addressed first before I play with the sparse part.

PR #2607, opened 2026-03-16
[Observability 7/7] Timing spans and step tags across all trainers

felipemello1:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #2607
* #2606
* #2605
* #2604
* #2603
* #2602
* #2601

## How to review

Understand the changes in trainer.py. The changes in other trainers (Flux, FT, Forge) are exactly the same.

## Summary

Wraps every training phase in `record_span` so the gantt chart shows the full training timeline. Adds `add_step_tag` annotations for GC, validation, and checkpoint events.

[ ] (Previous PR) 1st phase - Focuses on 'experiment metrics' and excludes 'record_span' and 'record_event'.
[x] (This PR) 2nd phase - Adds 'record_span', 'add_step_tag' and 'record_event'.

All four trainers (base, Flux, FT, Forge) have the same span coverage (Trainer, FT, Flux, Forge)

- `checkpoint.save()` and `GarbageCollection.run()` return `bool` for step tagging 

image

## Test plan

Flux, Forge and Trainer.py were test in main vs this PR. Losses and gradnorm are the same.

```bash
# Base trainer with PP
NGPU=8 LOG_RANK=0 ./run_train.sh --module llama3 --config llama3_debugmodel \
  --training.steps 10 --parallelism.data_parallel_shard_degree=2 \
  --parallelism.tensor_parallel_degree=2 --parallelism.pipeline_parallel_degree=2 \
  --parallelism.pipeline_parallel_schedule=1F1B --debug.seed=42 --debug.deterministic \
  --metrics.enable-wandb

# Base trainer with validation (no PP)
NGPU=8 LOG_RANK=0 ./run_train.sh --module llama3 --config llama3_debugmodel \
  --training.steps 10 --parallelism.data_parallel_shard_degree=4 \
  --parallelism.tensor_parallel_degree=2 --debug.seed=42 --debug.deterministic \
  --validator.enable --validator.freq=5 --validator.steps=2 --metrics.enable-wandb

# Flux (DP=4 CP=2)
NGPU=8 LOG_RANK=0 ./run_train.sh --module flux --config flux_debugmodel \
  --training.steps 5 --debug.seed=42 --metrics.enable-wandb \
  --parallelism.data_parallel_shard_degree=4 --parallelism.context_parallel_degree=2 \
  --tokenizer.test_mode --encoder.random_init \
[...]

PR #2974, opened 2026-04-15
Collect peak memory directly in integration tests

tugsbayasgalan:
* __->__ #2974
* #2766

Add direct peak-memory collection for integration runs by having MetricsProcessor write a JSON summary at the end of training.

The test runner now passes TORCHTITAN_PEAK_MEMORY_JSON into each launched training job, forces metrics logging every step, and reads the emitted summary file back for reporting. MetricsProcessor tracks the maximum reserved and active CUDA memory it observes across log and validation calls and writes a single summary on close from the metrics rank.

This keeps the measurement path local to the training run, avoids depending on TensorBoard event parsing for memory collection, and preserves the integration-test UX via --collect_peak_memory. The graph-trainer integration entrypoint and 8-GPU workflow are wired to use the flag.

### DIFF
--- a/.github/workflows/integration_test_8gpu_graph_trainer.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer.yaml
@@ -61,7 +61,7 @@ jobs:
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
         sudo chown -R $(id -u):$(id -g) "$RUNNER_TEMP/artifacts-to-be-uploaded"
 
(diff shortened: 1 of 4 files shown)


tianyu-l:
> Add direct peak-memory collection for integration runs by having MetricsProcessor write a JSON summary at the end of training.

@felipemello1 has WIP changes https://github.com/pytorch/torchtitan/pull/2607 to move towards this direction in general. Would prefer we wait for that if this is not urgent to graph trainer workstream.

PR #2975, opened 2026-04-15
[graph_trainer] Add CooR precompile support for aot_fx_trace compile mode

bobrenjc93:
* __->__ #2975

Introduce aot_fx_trace as an alternate precompilation path alongside
the existing AOT joint graph export. aot_fx_trace uses make_fx to
trace the full fwd+loss+bwd step as a single FX graph, then serializes
the GraphModule for loading on all ranks during training. Because
cudagraphs is a default pass, we also include cudagraph composability
related changes in this PR to get to bitwise equivalence between
aot_fx_trace with and without precompile.

Changes:
- precompile_main.py: Refactor into _common_setup() + _precompile_aot()
  + _precompile_aot_fx_trace(), route by --compile.mode. Use Python
  float for global_valid_tokens (matching dist_sum's return type) so
  make_fx bakes it as a graph constant, identical to the runtime path.
- precompile.py: PrecompiledFxTraceArtifact dataclass, save/load fns.
  Uses GraphPickler (not plain pickle) to preserve SymInt expressions
  (_runtime_compute_coordinate_on_dim) through serialization — plain
  pickle evaluates SymInts to concrete trace-time values, baking in
  rank-specific embedding vocab offsets.
- compile.py: Mode-aware validation (aot_fx_trace doesn't need
  serializable passes), logging for precompile load vs runtime trace
- trainer.py: _load_precompiled_fx_trace(), branching in
  _make_fx_forward_backward_step()
- make_fx_tracer.py: Use set_multithreading_enabled(False) so backward
  tracing runs on the calling thread, keeping the compile_on_one_rank
  ContextVar visible
- cudagraph.py: Defer _input_indices_to_copy when example_inputs is
  empty (precompile load path), accept opaque values (DeviceMesh) in
  input validation, skip non-tensors in static address checks.
  Squashed the separate cudagraph CooR PR into this one because
  cudagraph activates automatically for aot_fx_trace graphs (all
  inputs are desugared plain tensors + opaques, no HOPs), so bitwise
  equivalence testing required both fixes together.
- passes.py: Guard cudagraph_pass against non-GraphModule inputs

## Test plan
- `pytest torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py -x` — 6 passed, 4 skipped (Llama3, DSv3 × {eager self-deterministic, eager vs aot_fx_trace, precompile vs trace})
- `pytest torchtitan/experiments/graph_trainer/tests/test_precompile.py -x` — all pass
- `pytest torchtitan/experiments/graph_trainer/tests/test_trace_module.py -x` — all pass
- 8-GPU integration tests via CI (ciflow/8gpu)

### DIFF
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -132,6 +132,10 @@ class CUDAGraphWrapper:
         self._static_input_indices = OrderedSet(
             static_input_indices if static_input_indices is not None else []
         )
+        # example_inputs is empty when loading a precompiled artifact:
+        # the artifact is deserialized before the first real forward
+        # call, so no example inputs are available yet. In that case
+        # we defer computation to the first __call__ (warmup phase).
         self._input_indices_to_copy: list[int] | None = (
             [
                 i
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -188,8 +188,6 @@ class CUDAGraphWrapper:
 
     def _check_static_inputs_address(self) -> None:
         for i in self._static_input_indices:
-            if not isinstance(self._args[i], torch.Tensor):
-                continue
             actual = self._args[i].data_ptr()
             expected = self._input_addresses[i]
             assert expected == actual, (
@@ -212,17 +210,22 @@ class CUDAGraphWrapper:
 
         if self._cudagraph is None:
             self._validate_inputs(args)
-            if self._input_indices_to_copy is None:
-                self._input_indices_to_copy = [
-                    i
-                    for i, inp in enumerate(args)
-                    if isinstance(inp, torch.Tensor)
-                    and i not in self._static_input_indices
-                ]
             self._args = args
-            self._input_addresses = [
-                x.data_ptr() if isinstance(x, torch.Tensor) else None for x in args
-            ]
+            if self._input_indices_to_copy is None:
+                self._input_indices_to_copy = []
+                self._input_addresses = []
+                for i, x in enumerate(args):
+                    if isinstance(x, torch.Tensor):
+                        self._input_addresses.append(x.data_ptr())
+                        if i not in self._static_input_indices:
+                            self._input_indices_to_copy.append(i)
+                    else:
+                        self._input_addresses.append(None)
+            else:
+                self._input_addresses = [
+                    x.data_ptr() if isinstance(x, torch.Tensor) else None
+                    for x in args
+                ]
 
             self._cudagraph = torch.cuda.CUDAGraph()
 
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -134,8 +134,10 @@ class CUDAGraphWrapper:
         )
         # example_inputs is empty when loading a precompiled artifact:
         # the artifact is deserialized before the first real forward
-        # call, so no example inputs are available yet. In that case
-        # we defer computation to the first __call__ (warmup phase).
+        # call, so no example inputs are available yet. We defer
+        # _input_indices_to_copy to the first __call__ because we
+        # need actual inputs to distinguish tensors from opaque
+        # values (e.g. DeviceMesh from CooR/DTensor unwrapping).
         self._input_indices_to_copy: list[int] | None = (
             [
                 i
@@ -223,8 +225,7 @@ class CUDAGraphWrapper:
                         self._input_addresses.append(None)
             else:
                 self._input_addresses = [
-                    x.data_ptr() if isinstance(x, torch.Tensor) else None
-                    for x in args
+                    x.data_ptr() if isinstance(x, torch.Tensor) else None for x in args
                 ]
 
             self._cudagraph = torch.cuda.CUDAGraph()
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -124,6 +124,7 @@ class CUDAGraphWrapper:
         example_inputs: Sequence[Any],
         static_input_indices: tuple[int] | None = None,
         should_check_address: bool = False,
+        tensor_input_indices: list[int] | None = None,
     ):
         _cg_manager.maybe_initialize()
         _cg_manager.register_wrapper(self)
@@ -132,21 +133,16 @@ class CUDAGraphWrapper:
         self._static_input_indices = OrderedSet(
             static_input_indices if static_input_indices is not None else []
         )
-        # example_inputs is empty when loading a precompiled artifact:
-        # the artifact is deserialized before the first real forward
-        # call, so no example inputs are available yet. We defer
-        # _input_indices_to_copy to the first __call__ because we
-        # need actual inputs to distinguish tensors from opaque
-        # values (e.g. DeviceMesh from CooR/DTensor unwrapping).
-        self._input_indices_to_copy: list[int] | None = (
-            [
+        if tensor_input_indices is not None:
+            self._input_indices_to_copy = [
+                i for i in tensor_input_indices if i not in self._static_input_indices
+            ]
+        else:
+            self._input_indices_to_copy = [
                 i
                 for i, inp in enumerate(example_inputs)
                 if isinstance(inp, torch.Tensor) and i not in self._static_input_indices
             ]
-            if example_inputs
-            else None
-        )
         self._cudagraph: torch.cuda.CUDAGraph | None = None
         self._has_warmup = False
 
@@ -213,20 +209,9 @@ class CUDAGraphWrapper:
         if self._cudagraph is None:
             self._validate_inputs(args)
             self._args = args
-            if self._input_indices_to_copy is None:
-                self._input_indices_to_copy = []
-                self._input_addresses = []
-                for i, x in enumerate(args):
-                    if isinstance(x, torch.Tensor):
-                        self._input_addresses.append(x.data_ptr())
-                        if i not in self._static_input_indices:
-                            self._input_indices_to_copy.append(i)
-                    else:
-                        self._input_addresses.append(None)
-            else:
-                self._input_addresses = [
-                    x.data_ptr() if isinstance(x, torch.Tensor) else None for x in args
-                ]
+            self._input_addresses = [
+                x.data_ptr() if isinstance(x, torch.Tensor) else None for x in args
+            ]
 
             self._cudagraph = torch.cuda.CUDAGraph()
 
--- a/torchtitan/experiments/graph_trainer/cudagraph.py
+++ b/torchtitan/experiments/graph_trainer/cudagraph.py
@@ -132,11 +132,15 @@ class CUDAGraphWrapper:
         self._static_input_indices = OrderedSet(
             static_input_indices if static_input_indices is not None else []
         )
-        self._input_indices_to_copy = [
-            i
-            for i, inp in enumerate(example_inputs)
-            if isinstance(inp, torch.Tensor) and i not in self._static_input_indices
-        ]
+        self._input_indices_to_copy: list[int] | None = (
+            [
+                i
+                for i, inp in enumerate(example_inputs)
+                if isinstance(inp, torch.Tensor) and i not in self._static_input_indices
+            ]
+            if example_inputs
+            else None
+        )
         self._cudagraph: torch.cuda.CUDAGraph | None = None
         self._has_warmup = False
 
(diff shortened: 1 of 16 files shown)


torchtitan/experiments/graph_trainer/cudagraph.py
```diff
@@ -157,22 +165,31 @@ def _copy_non_static_inputs(self, *args):
...
+                "args must be tensor, integer (for dynamic shapes), "
+                "float (for scalar constants), "
+                "Generator (for random number generator), "
+                "or opaque object, "
+                f"but found {type(inp)} with value {inp!r} at index {i}"
+            )
 
     def _check_static_inputs_address(self) -> None:
         for i in self._static_input_indices:
+            if not isinstance(self._args[i], torch.Tensor):
```
xmfan: do we add static input indices for non-tensors?

torchtitan/experiments/graph_trainer/cudagraph.py
```diff
@@ -194,7 +211,14 @@ def __call__(self, *args):
             return out
 
         if self._cudagraph is None:
-            self._check_input_types(args)
+            self._validate_inputs(args)
+            if self._input_indices_to_copy is None:
```
xmfan: piggy back off the earlier loop over _static_input_indices

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -82,9 +78,18 @@ def construct_default_graph_passes(
         regional_inductor_pass,
     ]
 
-    # cudagraph should be the last pass.
+
```
xmfan, line 81: isn't this still true?

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -83,6 +87,22 @@ def compute_config_fingerprint(
...
+    device_mesh._runtime_compute_coordinate_on_dim) that are lazily
+    registered. The ops module uses @torch.library.custom_op with
+    DeviceMesh, which requires DeviceMesh to be registered as an
+    opaque type first. Must be called before deserializing any
+    CooR-compiled artifact.
+    """
+    from torch.distributed.device_mesh import _register_distributed_opaque_types
+
+    _register_distributed_opaque_types()
+    from torch.distributed._ops import device_mesh as _dm_ops  # noqa: F401
```
xmfan, line 103: this was a try/catch ImportError before

bobrenjc93: This was intentional since the ops must exist for CooR to work. I want to surface the error immediately rather than silently skipping.

torchtitan/experiments/graph_trainer/cudagraph.py
bobrenjc93: no I think this was overly defensive. will remove

torchtitan/experiments/graph_trainer/passes.py
bobrenjc93: added it back below

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
```diff
@@ -150,6 +152,107 @@ def _run_steps(
 
         return loss.detach().clone(), hash_model(model), hash_gradient(model)
 
+    def _run_steps_with_precompile(
```
SherlockNoMad, line 155: Nice, I like the flow of this test, this matches what I had in mind. 

The open question is, how would this flow fit into GraphTrainer, the Trainer Class? 
Do you think it should fit? or should I think of this workflow as a subclass of GraphTrainer?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -412,7 +412,20 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
+        # engine dispatches backward to a worker thread that has a
+        # fresh contextvars.Context, making the compile_on_one_rank
+        # ContextVar invisible and causing _sym_get_coordinate to
+        # bake rank 0's concrete coordinates into the backward graph.
+        with (
+            fake_mode,
+            tracing(ctx),
+            preserve_node_meta(),
+            _skip_nested_compile(),
+            torch.autograd.set_multithreading_enabled(False),
```
SherlockNoMad, line 431: Add a TODO to move this to the global init function. 

I think this good global default to force having backward on the same cpu thread as main, not just for the tracing time, but also for the run time. 

cc @tianyu-l

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -49,24 +49,20 @@
...
-    Per-pass configuration (e.g. ``static_input_indices`` for cudagraph) is
-    bound here via ``functools.partial`` so that ``apply_graph_passes``
-    stays a generic pass runner with no pass-specific parameters.
+    """Passes to apply at precompile time (before serialization).
 
-    Args:
-        traced_result: The traced graph and metadata from ``trace_train_step``.
+    Includes cleanup, FlexAttention annotation, and regional_inductor
+    compilation. The compiled Triton kernels (AOTCompiledArtifact nodes)
+    are serialized into the artifact by GraphPickler.
```
SherlockNoMad: condition this on " if pre-compile is enabled, ... " 
Precompile is optional, all graph passes are optional optimizations. 

We are showcasing the most optimized workflow, but hope not to enforce an optimization on user.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -82,9 +78,19 @@ def construct_default_graph_passes(
         regional_inductor_pass,
     ]
 
-    # cudagraph should be the last pass.
+
+def runtime_passes(
```
SherlockNoMad: If a user is not aware of the pre-compile workflow, it would be confused by meaning of "runtime pass"... 

If it's just only cudagraph, we can just directly apply cudagraph_pass in construct_default_graph_passes

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -285,6 +303,13 @@ def cudagraph_pass(
         static_input_indices: Explicit list of input indices with stable tensor
             addresses. When provided, ``is_forward`` is not used for inference.
     """
+    if not isinstance(gm, torch.fx.GraphModule):
```
SherlockNoMad, line 307: nice.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -115,25 +144,35 @@ def _make_fx_forward_backward_step(
...
-                self._traced_step.gm = apply_graph_passes(
-                    self._traced_step.gm,
-                    self._traced_step.example_inputs,
-                    passes,
-                )
+                if self.config.compile.precompile_artifact_dir:
+                    # Precompiled artifact already has cleanup +
+                    # regional_inductor baked in; only apply
+                    # load-time passes (cudagraph).
+                    passes = runtime_passes(self._traced_step)
```
SherlockNoMad: ah.... I see why you need `runtime_passes` now

SherlockNoMad: let's fold the complexity into construct_default_graph_passes, by taking an additional config and decide if you just need cudagraph pass there? 

Keep the cognitive load small in the trainer path.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -115,25 +144,35 @@ def _make_fx_forward_backward_step(
...
+                else:
+                    passes = construct_default_graph_passes(self._traced_step)
+                if passes:
```
SherlockNoMad: if passes is not necessary, if pass is empty, the apply would loop over empty list

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -104,6 +105,34 @@ def forward_backward_step(
...
+                f"Precompiled fx_trace artifact not found at "
+                f"'{compile_config.precompile_artifact_dir}/{_FX_TRACE_ARTIFACT_KEY}'. "
+                f"Run precompile_main with --compile.mode aot_fx_trace first."
+            )
+
+        config_fingerprint = compute_config_fingerprint(
+            model, compile_config, self.parallel_dims
+        )
+
+        self._traced_step = precompile_fx_trace_load(
```
SherlockNoMad, line 130: This is an important contact point between precompile loader and trainer. 

TracedResult is currently lightweight, and should stay lightweight. 

It's a serializable container, without fancy __call__ method. Let's keep it this way. 

cc @tugsbayasgalan, @yiming0416

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -269,3 +249,175 @@ def wrapper_fn(args, kwargs):
...
+        f"generate a fresh one. Set TORCHTITAN_SKIP_FINGERPRINT_CHECK=1 "
+        f"to bypass this check."
+    )
+
+
+_FX_TRACE_ARTIFACT_KEY = "fx_trace_default"
+
+
+@dataclass
+class PrecompiledFxTraceArtifact:
```
SherlockNoMad, line 300: Should we think of this class as the serialized container for TracedResult? 

TracedResult seems to be our contract now.

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -269,3 +249,175 @@ def wrapper_fn(args, kwargs):
...
+    """Artifact for aot_fx_trace precompilation.
+
+    Stores the traced FX graph (as a pickled GraphModule) alongside the
+    TracedResult metadata needed to unwrap DTensor inputs and rewrap
+    outputs at runtime. Inductor compilation happens at load time on
+    each rank.
```
SherlockNoMad: `Inductor compilation happens at load time on each rank.`

Is this accurate? are we really re-compiling triton kernel during artifact load time?

torchtitan/experiments/graph_trainer/cudagraph.py
```diff
@@ -132,11 +132,19 @@ def __init__(
...
-        ]
+        # example_inputs is empty when loading a precompiled artifact:
+        # the artifact is deserialized before the first real forward
+        # call, so no example inputs are available yet. In that case
+        # we defer computation to the first __call__ (warmup phase).
+        self._input_indices_to_copy: list[int] | None = (
+            [
+                i
+                for i, inp in enumerate(example_inputs)
+                if isinstance(inp, torch.Tensor) and i not in self._static_input_indices
```
SherlockNoMad, line 144: num_static_inputs is available from TracedResult, which is serialzied, so precompile shouldn't run into a case where _input_indices_to_copy couldn't be determined.

torchtitan/experiments/graph_trainer/cudagraph.py
```diff
@@ -157,19 +165,26 @@ def _copy_non_static_inputs(self, *args):
...
-                raise ValueError(
-                    "args must be tensor, integer (for dynamic shapes), "
-                    "float (for scalar constants), "
-                    "Generator (for random number generator), "
-                    "or opaque object, "
-                    f"but found {type(inp)} with value {inp!r} at index {i}"
-                )
+            if isinstance(inp, (torch.Tensor, int, float, torch._C.Generator)):
+                continue
+            if is_opaque_value(inp):
```
SherlockNoMad, line 177: For my own education, when does opaque input show up in precompile path?

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -247,6 +266,157 @@ def main():
...
+    logger.info(
+        f"Applied {len(passes)} precompile graph passes, "
+        f"graph now has {len(list(traced_result.gm.graph.nodes))} nodes"
+    )
+
+    # Augment compile_config with AC joint passes to match the training
+    # path, which calls apply_graph_ac during parallelization. Without
+    # this the config fingerprint will differ from training.
+    if config.activation_checkpoint.mode != "none":
+        apply_graph_ac(compile_config, config.activation_checkpoint)
```
SherlockNoMad: not needed. 

all passes should be applied via apply_graph_passes

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
bobrenjc93: It's integrated in _make_fx_forward_backward_step where _load_precompiled_fx_trace is called instead of retracing/compiling

torchtitan/experiments/graph_trainer/cudagraph.py
bobrenjc93: will refactor

torchtitan/experiments/graph_trainer/cudagraph.py
bobrenjc93: CooR hoists device meshes as opaque object inputs

torchtitan/experiments/graph_trainer/precompile.py
bobrenjc93: typo, will fix

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -49,24 +49,20 @@
...
 
-    Args:
-        traced_result: The traced graph and metadata from ``trace_train_step``.
+    If precompile is enabled, these are applied before serialization so
+    that compiled Triton kernels are baked into the artifact. Otherwise
+    they run at trace time via ``construct_default_graph_passes``.
 
-    Returns:
-        An ordered list of graph passes ready to apply.
+    cudagraph is excluded — it needs real tensors and devices at runtime.
```
SherlockNoMad, line 61: cudagraph is excluded, coz it need to re-capture the graph into in-memory cudagraph during run time, not due to "real tensors and devices"

torchtitan/experiments/graph_trainer/cudagraph.py
SherlockNoMad: I trust this is tracked somewhere, but if not, let's leave a TODO here

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -115,20 +143,26 @@ def _make_fx_forward_backward_step(
...
+                        global_valid_tokens,
+                        extra_inputs,
+                        extra_kwargs,
+                    )
 
             if self.config.compile.enable_passes:
-                passes = construct_default_graph_passes(self._traced_step)
+                passes = construct_default_graph_passes(
+                    self._traced_step,
+                    precompiled=bool(self.config.compile.precompile_artifact_dir),
```
SherlockNoMad, line 164: nit, I think we will latter need to refactor this into passing "self.config.compile" or even "self.config" as a more general interface, but it's fine for now.

SherlockNoMad (approved):

SherlockNoMad:
Do we have clear instruction in claude.md for agent to use the pre-compile workflow? 

I would validate by prompting ”run graph_trainer precompile training of llama_8b for me, show me the MFU, profiling result and memory snapshot as proof of work" in a /new session. 

this should work in one shot. 
If it doesn't, "reflect on instruction from claude.md, improve based on the experiment you just did"

bobrenjc93:
For posterity - claude one shotted it

CleanShot 2026-04-16 at 11 13 09@2x

PR #2977, opened 2026-04-15
[CI] Fix graph_trainer CI triggering on every PR

weifengpy:
▎ Summary

  ▎ Graph trainer workflows (integration_test_8gpu_graph_trainer.yaml,                      
  integration_test_8gpu_graph_trainer_h100.yaml) triggered on every PR, even those not 
  touching graph_trainer code. Two issues:                                                  
                                                                                          
  ▎ - Wrong scope: Both listened for ciflow/8gpu/* tags, which fire on any PR touching      
  torchtitan/**. The paths filter doesn't properly gate tag push events, so the workflows 
  ran unconditionally.                                                                      
  ▎ - Double-trigger: Each graph_trainer PR update triggered twice — once via pull_request 
  (synchronize) and once via ciflow/8gpu/* tag push, in different concurrency groups so     
  cancel-in-progress doesn't help.
                                                                                            
  ▎ Fixes:                                                                                
  ▎ 1. Replace ciflow/8gpu/* with a new ciflow/graph_trainer/* tag. Auto-labeling in 
  labeler.yml only matches torchtitan/experiments/graph_trainer/**, so the tag is only      
  pushed for graph_trainer PRs.
  ▎ 2. Restrict pull_request trigger to ready_for_review only — the ciflow tag path already 
  handles opened/synchronize/reopened; ready_for_review is kept because the ciflow probot   
  doesn't listen for that event.
                                                                                            
  ▎ Non-graph_trainer workflows are unchanged. The main H100 workflow                       
  (integration_test_8gpu_h100.yaml) keeps both triggers since they run different test 
  matrices (CUDA via pull_request, ROCm via ciflow/8gpu/* tag per set-matrix.yaml).         
                                                                                          
  ▎ Test plan

  ▎ - Verify graph trainer CI no longer triggers on PRs that don't touch                    
  torchtitan/experiments/graph_trainer/**
  ▎ - Verify graph trainer CI triggers exactly once per graph_trainer PR update (via ciflow 
  tag, no duplicate from pull_request synchronize)                                          
  ▎ - Verify draft→ready transition still triggers CI via ready_for_review

### DIFF
--- a/.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main ]
     tags:
-      - ciflow/8gpu/*
+      - ciflow/graph_trainer/*
     paths:
       - 'torchtitan/experiments/graph_trainer/**'
       - '.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,3 +4,8 @@
   - scripts/**
   - tests/**
   - torchtitan/**
+
+"ciflow/graph_trainer":
+  - torchtitan/experiments/graph_trainer/**
+  - .github/workflows/integration_test_8gpu_graph_trainer.yaml
+  - .github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
--- a/.github/pytorch-probot.yml
+++ b/.github/pytorch-probot.yml
@@ -1,3 +1,4 @@
 ciflow_push_tags:
   - ciflow/8gpu
+  - ciflow/graph_trainer
 labeler_config: labeler.yml
--- a/.github/workflows/integration_test_8gpu_graph_trainer.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main ]
     tags:
-      - ciflow/8gpu/*
+      - ciflow/graph_trainer/*
     paths:
       - 'torchtitan/experiments/graph_trainer/**'
       - '.github/workflows/integration_test_8gpu_graph_trainer.yaml'


.github/workflows/integration_test_8gpu_h100.yaml
tianyu-l, line 7: this seems auto-triggered for every commit from people who have write access https://github.com/pytorch/torchtitan/blob/main/.github/pytorch-probot.yml

I wonder if we should disable it for h100 tests as well.

weifengpy: Good call! H100 workflows now use a separate ciflow/h100/* tag, and requires mannual tagging

ciflow/8gpu/* is reduced to A10G and continues to auto-trigger. A10G runners have no queue issue

let me know if you want to be more aggressive

.github/workflows/integration_test_8gpu_h100.yaml
```diff
@@ -4,11 +4,11 @@ on:
   push:
     branches: [ main ]
     tags:
-      - ciflow/8gpu/*
+      - ciflow/h100/*
```
tianyu-l: We have the 8gpu tag because rocm tests would rely on this tag to be triggered https://github.com/pytorch/torchtitan/blob/main/.github/workflows/set-matrix.yaml#L29

If we change this tag, a downside could be we can't trigger rocm tests on a PR.

weifengpy: good question! claude told me this
```
ciflow/8gpu is auto-labeled on every PR touching torchtitan/** and continues to trigger ROCm
```

one evidence is this PR is triggering ROCm and ciflow/8gpu is auto attached

weifengpy: The labeler auto-applies ciflow/8gpu
```
From .github/labeler.yml:                                                                 
                                                                                            
  "ciflow/8gpu":                                                                            
    - .ci/docker/**                                                                         
    - .github/workflows/**                                                                  
    - scripts/**                                                                          
    - tests/**                                                                              
    - torchtitan/**                                                                       

  And from .github/pytorch-probot.yml:                                                      
   
  ciflow_push_tags:                                                                         
    - ciflow/8gpu                  

```

tianyu-l: Yes, ciflow/8gpu will be auto-triggered by any PR, because of the labeler.

But ciflow/8gpu tag will no longer trigger h100.yaml tests, which was meant to run on both cuda h100 AND rocm CI hardware. (Sorry, h100 test is just a "name" which should also on rocm)

IIUC the purpose of this PR is to save the resource for cuda h100, not rocm.

weifengpy: thanks for explaining this

ciflow/8gpu is now kept in the "h100" workflows to ensure ROCm coverage

.github/workflows/integration_test_4gpu_rl_h100.yaml
tianyu-l, line 4: can we add the tags for this rl test as well?

.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
tianyu-l, line 1: maybe not related to this PR:

I'm observing that the following jobs are trigger for every PR
- [GraphTrainer 8 GPU H100 Integration Tests / build-test / linux-job (push)](https://github.com/pytorch/torchtitan/actions/runs/24439489743/job/71400965583?pr=2977)
- [GraphTrainer 8 GPU Integration Tests / build-test / linux-job (push)](https://github.com/pytorch/torchtitan/actions/runs/24439489759/job/71401043647?pr=2977)

which is unexpected (e.g. for this PR). Could you help debug if possible?

.github/workflows/integration_test_4gpu_rl_h100.yaml
```diff
@@ -7,7 +7,7 @@ on:
       - 'torchtitan/experiments/rl/**'
       - '.github/workflows/integration_test_4gpu_rl_h100.yaml'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [ready_for_review]
```
tianyu-l: when you say
> non-H100 workflows unchanged

curious why non-H100 doesn't need dedup? IIUC they also have both `ciflow/8gpu` and `pull_requests`

.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
weifengpy: opened a new PR for graph trainer fix: https://github.com/pytorch/torchtitan/pull/2980

.github/workflows/integration_test_4gpu_rl_h100.yaml
weifengpy: actually this whole PR might just be wrong. There is no double triggering on "h100" workflows
* ciflow/8gpu was triggering rocm tests
* pull_requests was trigger cuda

Screenshot 2026-04-15 at 00 59 38

the other graph trainer PR is actually worth fixing

weifengpy:
after removing ciflow/8gpu/* frome experimental tests (https://github.com/pytorch/torchtitan/pull/2874 ), graph trainer are correcttly triggered, base on my following verification
* trigger from inside graph_trainer https://github.com/pytorch/torchtitan/pull/3004
* no trigger from outside graph_trainer https://github.com/pytorch/torchtitan/pull/3005

close this PR as it's resolved

PR #2980, opened 2026-04-15
[CI] Scope graph_trainer CI to graph_trainer files

weifengpy:
1. PR touching `torchtitan/experiments/graph_trainer/**`
  - the normal `pull_request` path still runs the main graph_trainer workflow on CUDA
  - the dedicated `ciflow/graph_trainer/*` tag-triggered path goes through `set-matrix`, which adds ROCm coverage for graph_trainer CI
  - the separate `GraphTrainer 8 GPU H100 Integration Tests` workflow in `integration_test_8gpu_graph_trainer_h100.yaml` still runs for
  graph-trainer-related changes, but PR updates no longer double-trigger it because its direct PR trigger is limited to
  `ready_for_review`

  2. PR touching files unrelated to graph_trainer CI
  - no graph_trainer workflow should run from `ciflow/graph_trainer/*`, because that label/tag is scoped to graph_trainer-related paths
  - no accidental graph_trainer trigger should come from broad `ciflow/8gpu/*`, because the graph_trainer workflows no longer listen to
  that tag

### DIFF
--- a/.github/workflows/integration_test_8gpu_graph_trainer.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer.yaml
@@ -4,12 +4,34 @@ on:
   push:
     branches: [ main ]
     paths:
+      - '.ci/docker/**'
+      - 'scripts/**'
+      - 'tests/**'
       - 'torchtitan/experiments/graph_trainer/**'
+      - 'torchtitan/__init__.py'
+      - 'torchtitan/trainer.py'
+      - 'torchtitan/components/**'
+      - 'torchtitan/config/**'
+      - 'torchtitan/distributed/**'
+      - 'torchtitan/models/**'
+      - 'torchtitan/protocols/**'
+      - 'torchtitan/tools/**'
       - '.github/workflows/integration_test_8gpu_graph_trainer.yaml'
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
+      - '.ci/docker/**'
+      - 'scripts/**'
+      - 'tests/**'
       - 'torchtitan/experiments/graph_trainer/**'
+      - 'torchtitan/__init__.py'
+      - 'torchtitan/trainer.py'
+      - 'torchtitan/components/**'
+      - 'torchtitan/config/**'
+      - 'torchtitan/distributed/**'
+      - 'torchtitan/models/**'
+      - 'torchtitan/protocols/**'
+      - 'torchtitan/tools/**'
       - '.github/workflows/integration_test_8gpu_graph_trainer.yaml'
   schedule:
     # Runs every 12 hours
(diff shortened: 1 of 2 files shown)


weifengpy:
seems we can only verify after merging

```
1. pull_request trigger uses main's config, not the PR's. The workflow on main still has  
  types: [opened, synchronize, reopened, ready_for_review]. Our change to [ready_for_review]
   doesn't take effect until it's merged. So pull_request (opened) fires using main's       
  trigger config.                                                                         
  
2. This PR modifies the workflow files themselves, which are in the paths filter:
  paths:
    - '.github/workflows/integration_test_8gpu_graph_trainer.yaml'  # ← we changed this file
```

.github/workflows/integration_test_8gpu_graph_trainer.yaml
```diff
@@ -4,12 +4,12 @@ on:
   push:
     branches: [ main ]
     tags:
-      - ciflow/8gpu/*
```
tianyu-l: https://github.com/search?q=repo%3Apytorch%2Ftorchtitan%20ciflow%2F8gpu%2F*&type=code

This pattern exists for almost every CI yaml. Actually I don't know what's going on. The idea should be about taking the "AND" of this `tags` and `paths`, rather than creating dedicated paths inside the labler.yaml.

weifengpy: claude gives me this evidence that "Path filters are not evaluated for pushes of tags" using official github doc. any ciflow/8gpu/* tag push triggers graph_trainer regardless of paths

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

ciflow/graph_trainer/* tag fixes this: labeler.yml only pushes that tag for PRs touching torchtitan/experiments/graph_trainer/**


> https://github.com/search?q=repo%3Apytorch%2Ftorchtitan%20ciflow%2F8gpu%2F*&type=code
> This pattern exists for almost every CI yaml.

other 8gpu workflows don't have this problem because they use set-matrix.yaml to route ciflow/8gpu/* tags to ROCm-only runs. Graph_trainer doesn't use set-matrix.yaml as it does not run on ROCm

tianyu-l: > Graph_trainer doesn't use set-matrix.yaml as it does not run on ROCm

Oh I see! Can we just use set-matrix.yaml for all CI, including graph trainer? Ideally all tests should pass for rocm.

weifengpy: Switched graph_trainer to \set-matrix.yaml` so it can use the shared CI routing logic, including adding ROCm coverage

Some tests are still gated on ROCm for now. I want to get the CI green first, then call out the CUDA/ROCm parity gap

weifengpy:
after removing ciflow/8gpu/* frome experimental tests (https://github.com/pytorch/torchtitan/pull/2874 ), graph trainer are correcttly triggered, base on my following verification

trigger from inside graph_trainer https://github.com/pytorch/torchtitan/pull/3004
no trigger from outside graph_trainer https://github.com/pytorch/torchtitan/pull/3005
close this PR as it's resolved

PR #2984, opened 2026-04-15
Refactor Loss with Scale to fix PP last stage gradients

### LINKED ISSUE #2972: PP and non-PP produce differently scaled gradients
## Summary

Pipeline Parallel (PP) and non-PP training produce differently scaled gradients for the same model and data. In non-PP, the loss is divided by `global_valid_tokens` before backward, so gradients are normalized. In PP, the schedule backwards the raw sum loss and the `/ global_valid_tokens` division happens after backward (for metrics only). This means PP gradients are `global_valid_tokens` times larger than non-PP gradients.

## Root Cause

**Non-PP path** (`trainer.py`):
```python
loss_sum = self.loss_fn(pred, labels)       # sum reduction
loss = loss_sum / global_valid_tokens       # scale BEFORE backward
loss.backward()                             # gradients are normalized
```

**PP path** (`trainer.py` + PyTorch `schedules.py`):
```python
# Inside pp_schedule.step(), on the last stage:
#   schedules.py:285 — loss = loss_fn(output, target)        # sum reduction, unscaled
#   schedules.py:766 — stage.backward_one_chunk(loss=loss)   # gradients are NOT normalized

# After pp_schedule.step(), in trainer.py:
loss = torch.sum(torch.stack(losses)) / global_valid_tokens  # metrics only, backward already done
```

Additionally:
- `scale_grads=False` is set in `build_pipeline_schedule` (`pipeline_parallel.py:234`)
- `gradient_divide_factor=1.0` via `disable_fsdp_gradient_division`

So neither the PP schedule nor FSDP normalizes the gradients.

## Reproduce

```bash
# Non-PP: 4 GPUs, FSDP only
NGPU=4 LOG_RANK=0 ./run_train.sh \
  --training.steps 3 --debug.seed 42 --debug.deterministic \
  --dump_folder /tmp/grad_scale_test/non_pp

# PP: 4 GPUs, PP2 + FSDP2
NGPU=4 LOG_RANK=0,2 ./run_train.sh \
  --training.steps 3 --debug.seed 42 --debug.deterministic \
  --parallelism.pipeline_parallel_degree 2 \
  --dump_folder /tmp/grad_scale_test/pp
```

Compare `grad_norm` between the two runs:

| Step | Non-PP grad_norm | PP grad_norm | Ratio |
|------|-----------------|-------------|-------|
| 1 | 1.5181 | 50,291.5 | ~33,000x |
| 2 | 1.5738 | 53,092.8 | ~33,700x |
| 3 | 1.9868 | 67,011.1 | ~33,700x |

The ratio matches `global_valid_tokens` (`local_batch_size=8 * seq_len=2048 * dp=2 = 32,768`, minus IGNORE_INDEX tokens).

### Versions

torchtitan main

wwwjn:
fix #2972 

There's 2 option to fix:
1. Add scaling to the loss function classes (clean, self-contained)
```
  class CrossEntropyLoss:
      def __init__(self, ...):
          self._scale = None

      def set_scale(self, scale: torch.Tensor) -> None:
          self._scale = scale

      def __call__(self, pred, labels):
          loss = self._fn(pred, labels)
          if self._scale is not None:
              return loss * self._scale
          return loss
```
Then in the trainer, before each step:
  `self.loss_fn.set_scale(1.0 / global_valid_tokens)`

Works for both PP and non-PP. Non-PP can also use this instead of the explicit loss_sum / global_valid_tokens in the trainer, unifying both paths.

2. Upstream fix in PyTorch's PP schedule: Add a loss_scale parameter to the schedule or allow _compute_loss to accept additional context. This is the right long-term fix but requires a PyTorch PR.

This PR go with Option 1. 

## Test
Screenshot 2026-04-15 at 10 57 36 AM

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -8,6 +8,7 @@ from collections.abc import Callable
 from typing import TypeAlias
 
 import torch
+import torch.nn as nn
 
 from torchtitan.config import CompileConfig
 from torchtitan.tools.logging import logger
@@ -51,7 +52,7 @@ class ScaledLoss:
         self._fn = loss_fn
         self._scale: torch.Tensor | None = None
 
-    def set_scale(self, scale: torch.Tensor) -> None:
+    def set_scale(self, scale: torch.Tensor | None) -> None:
         """Set a scaling factor applied to the loss output."""
         self._scale = scale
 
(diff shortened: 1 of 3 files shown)


wwwjn:
Closed because of #2937

PR #2985, opened 2026-04-15
[rl] Trainer refactor

joecummings:
### Context

**Main goal**: Refactor and clean up the Trainer interface to be flexible and similar to familiar Tinker's API

**Other things done**:
* Added `init_logging` to trainer so you can actually see the logs from PolicyTrainer - before was no output
* Switched to use default Qwen3 parallelize function for PolicyTrainer which means you can actually now use FSDP, TP, DDP, etc. Before you could only use TP
* Added ability to save a checkpoint - just standard
* Removed KL calculation - we weren't using it

### Testing
* CI
* Ran 0.6B script - same speed, same result
* Ran 0.6B script with FSDP - faster speed, same result
* Ran 0.6B script and saved a DCP checkpoint - checkpoint saves :)


torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,31 +74,41 @@ class Config(Configurable.Config):
         compile: CompileConfig = field(default_factory=CompileConfig)
         debug: DebugConfig = field(default_factory=DebugConfig)
         loss: Configurable.Config = field(default_factory=Configurable.Config)
+        ac_config: ActivationCheckpointConfig = field(
```
joecummings, line 77: These additions are needed when properly doing parallelisms

torchtitan/experiments/rl/tests/test_bitwise_identity.py
```diff
@@ -341,12 +343,21 @@ def main():
     _log_attention_module(trainer_model, "Trainer")
 
     with torch.no_grad():
-        trainer_token_log_probs = compute_token_log_probs(
-            trainer_model,
-            prompt_token_ids,
-            generated_token_ids,
-            device,
+        all_ids = prompt_token_ids + generated_token_ids
```
joecummings: I made the choice to follow the pattern set before rather than keep functions that are only used in tests

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -36,23 +38,24 @@
...
 logger = logging.getLogger(__name__)
 
 
 class PolicyTrainer(Actor, Configurable):
-    """
-    Updates policy based on collected Episode using TorchTitan components.
+    """Updates policy based on collected Episode using TorchTitan components.
 
-    Uses ModelSpec for model construction, parallelization, and weight loading.
+    Exposes a Tinker-style API with separate ``forward_backward`` and
```
felipemello1: maybe just "Exposes separate ``foward_...`

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,31 +74,41 @@ class Config(Configurable.Config):
         compile: CompileConfig = field(default_factory=CompileConfig)
         debug: DebugConfig = field(default_factory=DebugConfig)
         loss: Configurable.Config = field(default_factory=Configurable.Config)
+        ac_config: ActivationCheckpointConfig = field(
+            default_factory=lambda: ActivationCheckpointConfig(mode="none")
+        )
+        model_converters: ModelConvertersContainer.Config = field(
+            default_factory=ModelConvertersContainer.Config
+        )
+        model_spec: ModelSpec | None = None
```
felipemello1: Do you know if there was a reason for model_spec to be in the args instead of config previously?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,31 +74,41 @@ class Config(Configurable.Config):
...
+        transfer_dtype = config.transfer_dtype
         requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
         self._transfer_dtype = requested if requested != training_dtype else None
 
-        # The policy and ref models share code objects, so dynamo's
-        # per-code-object cache must hold entries for both grad modes
-        # (grad for policy, no_grad for ref). The default limit of 8
-        # is not enough; 16 accommodates both without recompile storms.
-        # TODO: @Lucaskabela fix recompiles in general as these increase startup
-        torch._dynamo.config.cache_size_limit = 16
```
felipemello1, line 97: why remove?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -124,32 +137,38 @@ def __init__(
         else:
             self.sd_adapter = None
 
+        # Build model converters (e.g. float8). No-op if converters list is empty.
+        model_compile_enabled = (
+            config.compile.enable and "model" in config.compile.components
+        )
+        model_converters = config.model_converters.build(
```
felipemello1, line 144: nit: is model converter our own abstractation or titans? When i read "converter" i thought of hf<->titan format, but this seems to be related to parallelism? It doesnt fit this PR, but we should rename it

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -232,11 +252,20 @@ def _build_model(
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
 
+        # Apply pre-parallelize conversions (e.g. float8)
+        model_converters.convert(model)
+
+        # Pass Config (not built instance) to parallelize_fn, matching core trainer.
```
felipemello1: probably dont need this comment

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -232,11 +252,20 @@ def _build_model(
...
+        # The parallelize function inspects converter configs for float8/MoE decisions.
```
felipemello1: prob dont need this either

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -369,23 +378,43 @@ async def step(self, train_data: list[TrainBatch]) -> dict:
...
+        dcp.save(state_dict, checkpoint_id=path)
+        logger.info(f"Saved checkpoint to {path}")
+
+    @endpoint
+    async def push_model_state_dict(self) -> None:
+        """Publish model weights for generator consumption via TorchStore.
+
+        Uses get_model_state_dict() to produce an unsharded state dict
+        that is compatible with the generator's different parallelization.
+        """
```
felipemello1, line 417: you changed the docstring, please check if that was intentional

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -369,23 +378,43 @@ async def step(self, train_data: list[TrainBatch]) -> dict:
...
+
+    @endpoint
+    async def save_checkpoint(self, path: str) -> None:
+        """Save model state dict to disk via DCP.
+
+        Args:
+            path: Directory to save the checkpoint to.
+        """
+        # TODO: Add checkpoint loading for resume support.
+        state_dict = {"model": get_model_state_dict(self.model)}
```
felipemello1, line 397: is "get_model_state_dict" expensive? If so, and if we call 'push_model_state_dict' on every step, i wonder if we should combine them.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -369,23 +378,43 @@ async def step(self, train_data: list[TrainBatch]) -> dict:
...
-                "logprob_bitwise_identical"
-            ],
-            **loss_metrics,
+            "policy_version": self.policy_version,
         }
-        logger.debug(
-            f"{os.getpid()=} PolicyTrainer finished step {self.policy_version}"
+
+    @endpoint
+    async def save_checkpoint(self, path: str) -> None:
```
felipemello1, line 389: where is this called? Nit: I wonder if ckpt should be a different PR

torchtitan/experiments/rl/tests/test_bitwise_identity.py
```diff
@@ -341,12 +343,21 @@ def main():
...
+        token_ids = torch.tensor([all_ids], dtype=torch.long, device=device)
+        seq_lens = [len(all_ids)]
+        prompt_lens = [len(prompt_token_ids)]
+        response_lens = [len(generated_token_ids)]
+        attention_masks = create_varlen_metadata(seq_lens, device)
+        positions = create_positions_from_seq_lens(seq_lens, device)
+        logits = trainer_model(
+            token_ids, attention_masks=attention_masks, positions=positions
         )
+        all_logprobs = compute_logprobs(logits, token_ids)
```
felipemello1: Do we still get same results?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -33,10 +33,11 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
     """GRPO training config for Qwen3-0.6B (6 GPUs: 4 gen + 2 train)."""
     model_spec = model_registry("0.6B_varlen")
     return RLTrainer.Config(
-        model_spec=model_spec,
         hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
         num_steps=10,
         trainer=PolicyTrainer.Config(
+            model_spec=model_spec,
+            transfer_dtype="bfloat16",
```
felipemello1, line 39: nit: should this be hardcoded?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -45,6 +46,7 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
             training=TrainingConfig(),
             parallelism=ParallelismConfig(
                 tensor_parallel_degree=2,
+                disable_loss_parallel=True,
```
felipemello1, line 48: nit: should this be hardcoded?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -415,15 +382,20 @@ async def setup(
             "trainer",
             PolicyTrainer,
             config.trainer,
-            model_spec=config.model_spec,
             hf_assets_path=config.hf_assets_path,
-            transfer_dtype=config.generator.model_dtype,
+        )
+
+        # Generator uses RL-specific parallelize (TP-only, no FSDP, vLLM-compatible)
+        from torchtitan.experiments.rl.models.parallelize import parallelize_qwen3
```
felipemello1, line 395: nit1: Are we hardcoding this script to qwen?
nit2: i saw multiple local imports. Is there any reason for them to be local?

felipemello1 (commented):
First pass, left some comments and nits.

regarding testing, you said that the different setting "passes". Whats the criteria? should we be expecting the same final value? same speed?

torchtitan/experiments/rl/actors/trainer.py
joecummings: It was b/c the `model_spec` was being overridden with the RL specific parallelize function, which was not actually the desired behavior.

torchtitan/experiments/rl/actors/trainer.py
joecummings: No reference model anymore - cc @Lucaskabela

torchtitan/experiments/rl/actors/trainer.py
joecummings: This is a titan abstraction following the exact pattern of the trainer.py file. They apply a "conversion" to a PyTorch model most often changing FQNs

torchtitan/experiments/rl/actors/trainer.py
joecummings: It's not called anywhere yet, but it was upsetting me that we didn't have the ability to save a ckpt. Let my omnibus pass!

torchtitan/experiments/rl/tests/test_bitwise_identity.py
joecummings: Yes

torchtitan/experiments/rl/config_registry.py
joecummings: I'd rather this be explicit, but maybe @daniellepintz has thoughts

torchtitan/experiments/rl/config_registry.py
joecummings: When we were using the TP-only parallelize plan, it had hardcoded the output to return plain tensors.  For now, we can just keep it the same by disabling loss_parallel in the standard parallelize plan.

torchtitan/experiments/rl/actors/trainer.py
joecummings: At our scale, it's negligible, but yeah a more performant synchronous checkpointing thing would likely be to do something like have the option to save an extra copy of the gathered weights to disk WHILE pushing to store. However, I don't really want this baked into our Trainer b/c it doesn't make as much sense in the async scenario where we may want to checkpoint and update weights are different points.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: For now, its a qwen3 parallelization script so implicitly yes

Local imports seem to be claude's defensive coding style and there's lots not introduced in this PR. Maybe a follow up and something to put in the Titan CLAUDE.md to stop doing?

joecummings:
> regarding testing, you said that the different setting "passes". Whats the criteria? should we be expecting the same final value? same speed?

Updated the description to be more clear

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -232,11 +252,17 @@ def _build_model(
...
+        model_converters.convert(model)
+
         model = model_spec.parallelize_fn(
             model,
             parallel_dims=self.parallel_dims,
+            training=config.training,
+            model_converters=config.model_converters,
             parallelism=config.parallelism,
             compile_config=config.compile,
+            ac_config=config.ac_config,
```
wwwjn, line 264: Previously I explicitly remove some fields from core torchtitan trainer that not being tested / supported in RL side. Do we add these fields because of API reason?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -369,23 +375,52 @@ async def step(self, train_data: list[TrainBatch]) -> dict:
...
+    @endpoint
+    async def save_checkpoint(self, path: str) -> None:
+        """Save model state dict to disk via DCP.
+
+        Args:
+            path: Directory to save the checkpoint to.
+        """
+        # TODO: Add checkpoint loading for resume support.
+        state_dict = {"model": get_model_state_dict(self.model)}
+        dcp.save(state_dict, checkpoint_id=path)
```
wwwjn, line 398: We might want to reuse torchtitan's CheckpointerManager for better composibility

torchtitan/experiments/rl/actors/trainer.py
wwwjn: I put `model_spec` a config for RLTrainer (the top level RL loop), and passed as args to PolicyTrainer. This is because the "unified" model path, we share the same model_spec for trainer and generator, so it's not only a sub-field of PolicyTrainer

joecummings: That's fair. Unfortunately the parallelize plans that are part of the model_spec are *not* actually shared which is why I did this moving around. However, I recognize the overall intention. 

I'll move it back and we can have a specific override for the Tp-only inference parallelize plan.

torchtitan/experiments/rl/actors/trainer.py
joecummings: Yeah that's completely fair and _was_ the right approach. It seems these are needed in order to use the "core" version of Qwen3's parallelize functionality, which is why I added them back here.

torchtitan/experiments/rl/actors/trainer.py
joecummings: I agree, but I'd like to do this as a follow up if that's alright? ATM all I wanted to have is the ability to save a quick checkpoint.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,31 +74,38 @@ class Config(Configurable.Config):
         compile: CompileConfig = field(default_factory=CompileConfig)
         debug: DebugConfig = field(default_factory=DebugConfig)
         loss: Configurable.Config = field(default_factory=Configurable.Config)
+        ac_config: ActivationCheckpointConfig = field(
+            default_factory=lambda: ActivationCheckpointConfig(mode="none")
+        )
+        model_converters: ModelConvertersContainer.Config = field(
+            default_factory=ModelConvertersContainer.Config
+        )
+        transfer_dtype: str = ""
```
tianyu-l: shall the type be https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L50

we can use None for "not requested"

tianyu-l: This should be combined with https://github.com/pytorch/torchtitan/pull/2942 in a combined `WeightSyncConfig`, depending on which lands first. cc @daniellepintz

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: This model converter is for applying quantization, mostly. We will remove this abstraction soon, and move quantization to config. cc @acisseJZhong to also refactor RL.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -124,32 +134,38 @@ def __init__(
...
+        model = self._build_model(
+            model_spec, config, model_converters, device_type, hf_assets_path
+        )
         model.train()
         self.model = model
         self.model_parts = [model]
 
-        # Conditionally build frozen reference model for KL penalty
-        # TODO: @joecummings remove ref entirely, this is hacky and we don't need it
-        if getattr(config.loss, "kl_coef", 0) > 0:
```
tianyu-l, line 135: IIRC @wwwjn had some concerns here, e.g. if we remove ref model entirely, then is it still GRPO?

I'm OK with removing it entirely for now to focus on the core part.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -124,32 +134,38 @@ def __init__(
...
         self.optimizers = config.optimizer.build(model_parts=self.model_parts)
         self.lr_schedulers = config.lr_scheduler.build(
             optimizers=self.optimizers,
             training_steps=config.training.steps,
         )
 
+        # Post optimizer step model converters hook.
+        # e.g. calculate float8 dynamic amax/scale for FSDP2
+        self.optimizers.register_step_post_hook(
+            lambda *args, **kwargs: model_converters.post_optimizer_hook(
```
tianyu-l, line 167: @acisseJZhong in your refactor, I don't think we need this hook anymore -- let's discuss this later.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -249,48 +272,18 @@ def _build_model(
...
+    async def forward_backward(self, train_data: list[TrainBatch]) -> dict:
+        """Run forward pass, compute loss, and call backward.
 
         Args:
-            train_data (list[TrainBatch]): List of batches, one per DP rank.
+            train_data: List of batches, one per DP rank.
 
         Returns:
-            dict: Training metrics (loss, policy version, etc.).
+            dict with loss metrics, advantage stats, and logprob verification.
```
tianyu-l, line 285: Do you think we should make this into a dataclass as well? Please leave a TODO if so.

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: Please add a TODO

torchtitan/experiments/rl/config_registry.py
tianyu-l: @wwwjn pls confirm this

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -417,13 +389,17 @@ async def setup(
             config.trainer,
             model_spec=config.model_spec,
             hf_assets_path=config.hf_assets_path,
-            transfer_dtype=config.generator.model_dtype,
         )
+
+        # Generator uses RL-specific parallelize (TP-only, no FSDP, vLLM-compatible)
+        from torchtitan.experiments.rl.models.parallelize import parallelize_qwen3
+
+        gen_model_spec = replace(config.model_spec, parallelize_fn=parallelize_qwen3)
```
tianyu-l, line 397: wait, do we still need this because full DTensor change has not landed? @wwwjn can we just land it, despite PP work or not? We've waited for too long.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -552,19 +528,28 @@ async def train(self):
...
                 self._collate_episodes(per_rank_episodes)
                 for per_rank_episodes in maybe_sharded_episodes
             ]
-            metrics = self._get_rank_0_value(self.trainer.step.call(batches).get())
+            fwd_bwd_metrics = self._get_rank_0_value(
+                self.trainer.forward_backward.call(batches).get()
+            )
+
+            # 5. Optimizer step
+            optim_metrics = self._get_rank_0_value(self.trainer.optim_step.call().get())
```
tianyu-l, line 542: curious what's the example benefits of separating optim and fwd/bwd for trainer?

torchtitan/experiments/rl/config_registry.py
tianyu-l: would this ever be different from `generator.model_dtype`? If not, we shouldn't give user this option only to enable the program to fail.

Another option is to allow different transfer dtype and generator dtype, so that the synced weight is converted one more time at receiver side.

torchtitan/experiments/rl/actors/trainer.py
joecummings: It depends on the direction we go with Observability. We could record all these metrics here, directly with `record_metrics` (or something similar). If we end up forgoing that and returning metrics here (not my preferred option), then we might consider a dataclass. Punting on this to @felipemello1's work

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: There's a few:
* controller control over gradient accumulation over multiple rollout batches
* controller owned LR scheduler (future direction where we can pass in optimizer params per step - helpful for interesting things in RL)

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,31 +74,41 @@ class Config(Configurable.Config):
...
+
         self.config = config
         self.loss_fn = config.loss.build()
-        self.model_spec = model_spec
+
         # Only cast if transfer dtype differs from training dtype, otherwise
         # staging buffers would be allocated for a no-op cast.
         training_dtype = TORCH_DTYPE_MAP[config.training.dtype]
-        requested = TORCH_DTYPE_MAP[transfer_dtype] if transfer_dtype else None
+        requested = (
```
tianyu-l: let's use a more clear name

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -71,31 +74,41 @@ class Config(Configurable.Config):
         compile: CompileConfig = field(default_factory=CompileConfig)
         debug: DebugConfig = field(default_factory=DebugConfig)
         loss: Configurable.Config = field(default_factory=Configurable.Config)
+        ac_config: ActivationCheckpointConfig = field(
+            default_factory=lambda: ActivationCheckpointConfig(mode="none")
+        )
+        model_converters: ModelConvertersContainer.Config = field(
+            default_factory=ModelConvertersContainer.Config
+        )
+        transfer_dtype: str | None = None
```
tianyu-l: still not sure about if we should have this! https://github.com/pytorch/torchtitan/pull/2985/changes#r3097562813

at least we should assert on the dtype being the same as generator.model_dtype

tianyu-l (approved):

torchtitan/experiments/rl/actors/trainer.py
joecummings: Okay no response yet from the TS team but my understanding is that we may eventually want to be able to do quantized weight transfer. However, this seems like exposing a parameter way before it's needed so let's hardcode to the generator dtype for now (similar to what we were doing originally, but now explicitly referring to it as generator dtype)

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -417,13 +389,18 @@ async def setup(
             config.trainer,
             model_spec=config.model_spec,
             hf_assets_path=config.hf_assets_path,
-            transfer_dtype=config.generator.model_dtype,
+            generator_dtype=config.generator.model_dtype,
```
daniellepintz, line 392: yea this looks good

tianyu-l (approved):

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -218,13 +197,6 @@ def __post_init__(self):
 
     def __init__(self, config: Config):
         self.config = config
-
-        # Patch model_spec to use the RL-specific parallelize function.
-        # TODO: Switch to canonical Qwen3 parallel plan
-        from torchtitan.experiments.rl.models.parallelize import parallelize_qwen3
-
-        config.model_spec.parallelize_fn = parallelize_qwen3
```
wwwjn, line 226: Have you verified switching to Qwen3 core parallel plan works for trainer and generator? Especially generator has odd number of tokens? My impression is that it won't work on generator side with odd  number of tokens when TP is applied. Did CI catch this?

Note the major difference between parallelize_qwen3 and core torchtitan's qwen3 parallel plan is whether DTensor is applied within TP region, not only the loss_parallel

torchtitan/experiments/rl/simple_grpo_sum_digits.py
wwwjn: The PR works and under review, should be land soon

wwwjn: @joecummings After this change, you are using core torchtitan qwen3's parallel plan for trainer, only patching the parallelize_qwen3 for generator, which should work. I take my words back.

torchtitan/experiments/rl/config_registry.py
wwwjn: Becase this PR switches trainer's parallel plan to be core torchtitan's qwen3 parallelism plan , yes we would need to have this change to make sure trainer's parallelism is applied the same before and after. 

However, I'm a little bit doubt that loss parallel won't work for trainer. Would need to dig more into the how loss parallel works in RL settings

FYI after https://github.com/pytorch/torchtitan/pull/2149 this PR landed, we can switch both trainer and generator's parallel plan to core torchtitan's qwen3 plan

PR #2986, opened 2026-04-15
[Module] Remove LocalMapInnerAttention, use static LocalMapSpec

fegin:
* #2987
* #3113
* #3080
* __->__ #2986

Replace runtime DTensor detection in `LocalMapInnerAttention` with static `LocalMapSpec` on inner attention configs via sharding.py.  All inner attention types (SDPA, FlexAttention, Varlen) now use `Module.parallelize()` local_map for TP DTensor→local conversion.

- Set LocalMapSpec on inner attention in all model sharding.py files
- Migrate Flux CP from apply_cp_to_attention_module to apply_cp_to_forward
- Remove LocalMapInnerAttention class; subclasses inherit Module directly
- Remove apply_cp_to_attention_module (no remaining callers)
- Migrate experiments/rl to config-based sharding:
      - VLLMAttentionWrapper inherits Module directly (no runtime DTensor bridge).
      - parallelize_qwen3 replaced with model.parallelize(tp_mesh); legacy parallelize_module plan deleted.
      - set_qwen3_sharding_spec(..., include_positions=True) installed on the config before build() in vllm_wrapper and in trainer.PolicyTrainer._build_model.
      - compute_logprobs materializes DTensor logits to plain via full_tensor() (downstream RL uses plain-tensor indexing).
- Add include_positions kwarg to set_gqa_attention_sharding so the generator path annotates the positions argument.


torchtitan/distributed/sharding.py
```diff
@@ -99,6 +99,29 @@ def set_gqa_attention_sharding(attention_cfg, *, enable_sp: bool) -> None:
     attention_cfg.wo.sharding_spec = rowwise_spec(output_sp=enable_sp)
 
 
+def set_gqa_inner_attention_local_map(
+    inner_attention_cfg, *, num_outputs: int = 1
```
tianyu-l, line 103: `num_outputs` is vague, how about just passing `return_lse`

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -35,6 +35,13 @@ def compute_logprobs(logits: torch.Tensor, token_ids: torch.Tensor) -> torch.Ten
     Returns logprobs for positions 1..N (the predicted tokens).
     Output shape is ``[batch, seq_len - 1]``.
     """
+    from torch.distributed.tensor import DTensor
+
+    # Config-based TP returns logits as a Replicate DTensor. Downstream RL
+    # code (gather with plain-tensor indices, slicing per-sample) expects a
+    # plain tensor — materialize once here.
+    if isinstance(logits, DTensor):
+        logits = logits.full_tensor()
```
tianyu-l, line 44: @wwwjn pls review this

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -228,6 +228,16 @@ def _build_model(
             model_spec.model.layers[0].attention.inner_attention, VarlenAttention.Config
         ), "Only varlen attention backend is allowed."
 
+        # Fill sharding specs on the config BEFORE build (RL's trainer bypasses
+        # the standard torchtitan Trainer which would otherwise do this).
+        if model_spec.set_sharding_spec_fn is not None:
```
tianyu-l, line 233: similarly to PR1 comment: sharding config should be done at config time

torchtitan/experiments/rl/models/parallelize.py
tianyu-l, line 26: @wwwjn will remove this soon

torchtitan/distributed/sharding.py
```diff
@@ -99,6 +99,29 @@ def set_gqa_attention_sharding(attention_cfg, *, enable_sp: bool) -> None:
...
+
+    q/k/v arrive as ``(bs, seq, heads, head_dim)`` DTensors with heads
+    TP-sharded (``Shard(2)``), regardless of SP. ``local_map`` converts them
+    to local tensors before the kernel runs, then wraps outputs back.
+
+    ``num_outputs`` controls the number of tensor outputs: 1 for the usual
+    (attn_out) case; 2 for kernels that return ``(output, lse)`` like
+    GPT-OSS's flash attention with ``return_lse=True``.
+    """
+    qkv_placements = (Shard(2),)
```
tianyu-l, line 115: not necessarily always 2 -- varlen is different from flex (and sdpa), so this probably should depend on the inner_attn config class

fegin: I think it is the same. Our `VarlenAttention` reshape to `[bs * seqlen, n_heads, head_dim]` within the local_map region. So `in_placements` should be `(Shard(2),)`.  Correct me if I'm wrong.

torchtitan/models/common/decoder_sharding.py
```diff
@@ -99,6 +99,29 @@ def set_gqa_attention_sharding(attention_cfg, *, enable_sp: bool) -> None:
     attention_cfg.wo.sharding_spec = rowwise_spec(output_sp=enable_sp)
 
 
+def set_gqa_inner_attention_local_map(
+    inner_attention_cfg, *, num_outputs: int = 1
```
fegin: [TODO:] change naming to `return_lse`.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -245,6 +245,17 @@ def _build_model(
...
+        # bypasses torchtitan.Trainer and does not call update_from_config,
+        # so we call set_qwen3_sharding_spec directly here.
+        from torchtitan.models.qwen3.sharding import set_qwen3_sharding_spec
+
+        set_qwen3_sharding_spec(
+            model_spec.model,
+            loss_parallel=not config.parallelism.disable_loss_parallel,
+            enable_sp=config.parallelism.enable_sequence_parallel,
+        )
+
```
tianyu-l: instead of this I think we should call `update_from_config(trainer_config)`. Right now the only thing needed is trainer_config.parallelism and trainer_.training which are available in both core trainer and RL policy trainer.

Update: I think for RL, the trainer rope cache length is not supposed to be changed, either, so we don't need to send in `trainer_config.training.seq_len`

torchtitan/models/common/decoder_sharding.py
```diff
@@ -148,6 +153,30 @@ def set_gqa_attention_sharding(attention_cfg, *, enable_sp: bool) -> None:
...
+
+    q/k/v arrive as ``(bs, seq, heads, head_dim)`` DTensors with heads
+    TP-sharded (``Shard(2)``), regardless of SP. ``local_map`` converts them
+    to local tensors before the kernel runs, then wraps outputs back.
+
+    ``return_lse=True`` is for kernels that return ``(output, lse)`` (e.g.,
+    GPT-OSS's flash attention with ``return_lse=True``); both outputs share
+    the same heads-sharded placement.
+    """
+    qkv_placements: NamedPlacement = {TP: Shard(2)}
```
tianyu-l, line 169: Not always Shard(2), it depends on whether the inner attention is varlen vs. sdpa/flex. Maybe need `isinstance` here.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -205,27 +206,35 @@ def __init__(
...
         # vLLM config contains the tensor_parallel_size from command-line args
         # and this will be consistent across all worker processes.
         self.parallel_dims, parallelism = create_torchtitan_config_from_vllm_config(
             vllm_config
         )
 
+        # Fill sharding specs on the config BEFORE build so every sub-module
+        # (including VLLMAttentionWrapper) is constructed with its
+        # ShardingConfig / LocalMapConfig attached.
+        set_qwen3_sharding_spec(
```
tianyu-l: Need to call update_from_config to make it model-agnostic.

In RL generator we never need to extend sequence length for RoPE, so we can make it optional.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -367,6 +376,11 @@ def compute_logits(
 
         logits = self.model.output(hidden_states)
 
+        # Config-based sharding returns logits as a Replicate DTensor; vLLM
+        # expects a plain tensor.
+        if isinstance(logits, DTensor):
+            logits = logits.to_local()
```
tianyu-l, line 394: @wwwjn also this

torchtitan/experiments/rl/models/parallelize.py
wwwjn: #3070  is the PR to remove this function. I cal rebase later

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -245,6 +245,17 @@ def _build_model(
             model_spec.model.layers[0].attention.inner_attention, VarlenAttention.Config
         ), "Only varlen attention backend is allowed."
 
+        # Fill sharding specs on the config BEFORE build. RL's trainer
+        # bypasses torchtitan.Trainer and does not call update_from_config,
+        # so we call set_qwen3_sharding_spec directly here.
+        from torchtitan.models.qwen3.sharding import set_qwen3_sharding_spec
+
+        set_qwen3_sharding_spec(
```
wwwjn: Is this overriding `loss_parallel` and `enable_sp` fields in qwen3 sharding spec

torchtitan/experiments/rl/actors/utils.py
wwwjn: I think this change looks correct to me, previously we return a ColwiseParallel(input_layouts=sp_layout, output_layouts=Replicate(), use_local_output=True), so the explicated redistribute() makes sense

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This materialize makes sense to me

torchtitan/experiments/rl/models/parallelize.py
fegin: I'm not sure, maybe you will land the PR first, lol.

torchtitan/models/common/decoder_sharding.py
fegin: Please check my comment above. The varlen flatten the dim-0 and dim-1 INSIDE the local map region. So this should always be Shard(2)?

torchtitan/experiments/rl/actors/trainer.py
fegin: qwen3 sharding spec doesn't have loss_parallel and enable_sp. They are in the parallelism config. That's why we defer to set this later.

fegin (commented):
@tianyu-l I think it is always `Shard(2)` regardless of the attention kernel. Please check my comment above.

torchtitan/models/common/decoder_sharding.py
tianyu-l: Seems you are right. There has been some refactor, so it is possible that either
- the refactor makes it always Shard(2)
- I hallucinated

torchtitan/experiments/rl/actors/utils.py
tianyu-l: why `full_tensor` not `to_local`? I'm worried when full DTensor on DP mesh is enabled, this is correct or not

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -205,27 +210,37 @@ def __init__(
...
+        # Fill sharding configs on the config BEFORE build so every sub-module
+        # (including VLLMAttentionWrapper) is constructed with its
+        # ShardingConfig / LocalMapConfig attached. Uses the model-agnostic
+        # ``update_from_config`` hook. vLLM inference doesn't need RoPE
+        # cache extension here (that happens explicitly below via
+        # ``max_model_len``) nor MoE force-load-balance, so ``training`` and
+        # ``debug`` are omitted.
+        self.config.update_from_config(
+            trainer_config=SimpleNamespace(parallelism=parallelism)
+        )
```
tianyu-l, line 230: sounds hacky, please leave a TODO for @wwwjn to refactor `TrainingConfig`

torchtitan/models/qwen3/model.py
```diff
@@ -92,22 +92,29 @@ def update_from_config(
...
-                    f"Sequence length {seq_len} exceeds original maximum {self.rope.max_seq_len}."
-                )
-            # Sync rope max_seq_len
-            self.rope = dataclasses.replace(self.rope, max_seq_len=seq_len)
-
-            for layer_cfg in self.layers:
-                if layer_cfg.moe is not None:
-                    layer_cfg.moe.router._debug_force_load_balance = (
-                        debug.moe_force_load_balance
+            # ``training`` and ``debug`` are optional: the RL vLLM generator
```
tianyu-l, line 96: add a TODO: at least we shouldn't call it `trainer_config` because it's also used in generator

tianyu-l (approved):

torchtitan/experiments/rl/models/vllm_wrapper.py
fegin: yes, I agree this is very hacky. Claude suggests that `update_from_config` can accept `TrainingConfig | ParallelismConfig`. But I don't have high confidence this is a good alternative for RL. Will leave a TODO and let @wwwjn work on this. Thanks.

torchtitan/experiments/rl/actors/utils.py
fegin: `to_local` should be the correct one.

PR #2993, opened 2026-04-16
[GraphTrainer] Fix custom_codegen_pass hang on large models

SherlockNoMad:
## Summary
- Fix O(lines × nodes) regex hang in `_generate_profiled_forward` that made `custom_codegen_pass` unusable on large models (Llama 8B+)
- The inner loop compiled a new regex per node per code line (~5k × ~5k = ~25M regex ops for Llama 8B), causing an effective hang at `[CUSTOM_CODEGEN] Saving code to ...`
- Replace with a single pre-compiled regex that extracts the leading identifier + O(1) dict lookup. Codegen pass now completes in ~1s for Llama 8B.
- Re-enables `custom_codegen_pass` in the default pass list (was disabled as a workaround in #2992)

## Test plan
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_passes.py -x` — 37/37 passed (including 4 codegen tests)
- [x] Reproduced hang: Llama 8B profiling run stuck at codegen for 3+ minutes, `py-spy` confirmed stuck in `re.match` at `custom_codegen.py:338`
- [x] Verified fix: same command completes in ~1s for codegen, full 15-step profiling run finishes cleanly

### DIFF
--- a/torchtitan/experiments/graph_trainer/custom_codegen.py
+++ b/torchtitan/experiments/graph_trainer/custom_codegen.py
@@ -320,6 +320,12 @@ def call(self, *args, **kwargs):
             if node.op in ("call_function", "call_method", "call_module")
         }
 
+        # Pre-compile a single regex to extract the leading identifier from
+        # assignment lines (``name = ...``) or tuple-unpacking lines
+        # (``name, ...``).  This replaces the previous O(lines × nodes)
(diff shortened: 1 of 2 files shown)


SherlockNoMad:
@copilot  are you sure this is equivalent rewrite?

xmfan (approved):

PR #2994, opened 2026-04-16
Fix batch invariant mode: using NCCL tree based all-reduce

wwwjn:
As titled. Previously we used ring based algorithm which is not deterministic


tianyu-l (commented):
shall we error out if SP + batch invariance are used together?

torchtitan/distributed/utils.py
```diff
@@ -243,16 +243,33 @@ def set_batch_invariance(enable: bool) -> None:
 
     # Set NCCL env vars for deterministic inter-GPU collectives.
     # Must be set BEFORE dist.init_process_group.
-    os.environ["NCCL_ALGO"] = "Ring"  # Fixed summation order (Tree may vary)
-    os.environ["NCCL_MIN_NCHANNELS"] = "1"  # Single channel to avoid split interleaving
-    os.environ["NCCL_MAX_NCHANNELS"] = "1"
-    os.environ["NCCL_PROTO"] = "Simple"  # LL/LL128 may reorder reductions
+    # Reference: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/batch_invariant.py
```
acisseJZhong, line 246: for batch inv, do we need to set `s.environ["VLLM_ALLREDUCE_USE_SYMM_MEM"] = "0"`?  https://github.com/vllm-project/vllm/blob/219bb5b8c0dcc6a5d5f894e9168fa5b8c2f8255a/vllm/model_executor/layers/batch_invariant.py#L1031

torchtitan/experiments/rl/README.md
```diff
@@ -97,7 +97,7 @@ Batch-invariant mode guarantees that a model's output for a given input is **ide
 
 When enabled, batch-invariant mode will:
 - Replaces `mm`, `addmm`, `log_softmax`, and `mean.dim` with Triton kernels that use a fixed tile iteration order (via [batch_invariant_ops](https://github.com/thinking-machines-lab/batch_invariant_ops))
-- Forces NCCL to use Ring all-reduce with a single channel for deterministic inter-GPU collectives
+- Forces deterministic NCCL collectives (single channel, simple protocol, tree allreduce) matching vLLM's settings
```
acisseJZhong, line 100: curious previously why we see Ring also deterministic?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -215,6 +215,12 @@ def __post_init__(self):
                         f"batch_invariant requires bfloat16 generator dtype, "
                         f"got {self.generator.model_dtype!r}"
                     )
+                if self.trainer.parallelism.enable_sequence_parallel:
+                    raise ValueError(
+                        "batch_invariant mode doesn't support SP now. "
+                        "SP uses reduce-scatter which only supports Ring in NCCL "
```
joecummings, line 221: Doesn't FSDP also need reduce-scatter?

torchtitan/distributed/utils.py
wwwjn: No, this is controlling vllm's all-reduce kernel, which we are not relying on

torchtitan/experiments/rl/README.md
wwwjn: Because previously we only tested on TP=2 with smaller sequence length, within a single node. For larger size test, eg, cross node test, Tree algorithm is deterministic to best of my knowledge

torchtitan/experiments/rl/simple_grpo_sum_digits.py
wwwjn: For bit-wise identity check, we comparing **forward** results between trainer and generator. When FSDP is enabled, the RS is applied during **backward**  to sync gradients, which won't affect forward results

tianyu-l: batch variant mode is about forward only, anything bwd is not required
- inference is fwd only
- there is no hope to achieve batch invariant for backward because the batch dim will be reduced too

tianyu-l (approved):

PR #2061, opened 2025-11-19
[moe] brings batch/sequence-wise load balance loss

rakkit:
This is a draft PR for:
1) Make the moe's load_balance_coeff configurable 
2) add the batch and seq-wise aux loss for load balance. [ref: [dpskv3](https://arxiv.org/pdf/2412.19437#page=9.60) eqn. 17~20]

For now, it only applies to the DeepSeek model, but I can add it for all other moe models at the end.
(also, we dont log the aux loss, but i can add it in optimizer hook to do this if you want)

The main concern is that the aux loss does not work well with PP. From what I have tested, it works well only with 1F1B.  And it is broken for ZBV or interleaved 1f1b.

To test it:
[sequence_wise, by default]
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --training.extra_losses.load_balance_loss_weight=0.001`
image

[batch_wise, need to pick this in ModelArgs]
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh --training.extra_losses.load_balance_loss_weight=0.001`
image

(turn it off)
`CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" NGPU=4 ./run_train.sh `
image

PR #3000, opened 2026-04-16
[moe] load-balancing aux loss

pianpwk:
Adds MoE load-balancing aux loss following https://github.com/pytorch/torchtitan/pull/2061.

For PP/compile compatibility, the aux_loss is implemented as an autograd.Function for loss gradient injection, following https://github.com/mlc-ai/pith-train/blob/main/pithtrain/modules/load_balance.py. Added a unit test to show this approach matches in numerics, to standard `(loss + aux_loss).backward()`. This is configured as a local_map region

We support 2 aux loss functions: batch-wise (batch-level stats) and seq-wise (seq-level stats). For parallelism compatibility, we allreduce routing expert counts across meshes:
- for batch-wise, we allreduce counts across DP (batch-dim), CP (seq-dim), and TP (seq-dim when EP enabled), as a [# experts] tensor
- for seq-wise, we allreduce counts across CP & TP on same conditions, as a [BS, # experts] tensor, as we cannot aggregate across batch dim.

The unsupported cases for batch-wise aux loss are gradient accumulation, and PP (w/ microbatching, but titan bans non-microbatched PP anyways), as the expert counts are approximate. Currently we error out for this (too strict?)

seq-wise aux loss (aux loss settings aren't in CLI): `MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel NGPU=4 ./run_train.sh`
```
step:  1  loss:  8.04721  grad_norm:  4.0597  memory: 22.19GiB(28.04%)  tps: 4,794  tflops: 2.60  mfu: 0.83%
step:  5  loss:  4.42904  grad_norm:  2.3887  memory: 22.25GiB(28.12%)  tps: 19,275  tflops: 10.47  mfu: 3.35%
step: 10  loss:  3.85238  grad_norm:  1.5835  memory: 22.25GiB(28.12%)  tps: 19,077  tflops: 10.36  mfu: 3.32%
```

batch-wise aux loss (Qwen3 MoE debug uses batch-wise aux loss): `MODULE=qwen3 CONFIG=qwen3_moe_debug NGPU=4 ./run_train.sh`
```
step:  1  loss:  8.07451  grad_norm:  2.5818  memory:  3.84GiB(4.85%)  tps: 2,089  tflops: 2.32  mfu: 0.74%
step:  5  loss:  5.07399  grad_norm:  2.6846  memory:  4.43GiB(5.60%)  tps: 23,166  tflops: 25.74  mfu: 8.25%
step: 10  loss:  4.51730  grad_norm:  2.5816  memory:  4.43GiB(5.60%)  tps: 24,027  tflops: 26.70  mfu: 8.56%
```

seq-wise, compile: `MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel NGPU=4 ./run_train.sh --compile.enable`
```
step:  1  loss:  8.13759  grad_norm:  4.0312  memory: 16.86GiB(21.31%)  tps: 360  tflops: 0.20  mfu: 0.06%
step:  5  loss:  4.35701  grad_norm:  2.3747  memory: 16.90GiB(21.35%)  tps: 28,007  tflops: 15.21  mfu: 4.87%
step: 10  loss:  3.82011  grad_norm:  1.4143  memory: 16.90GiB(21.35%)  tps: 28,781  tflops: 15.63  mfu: 5.01%
```

seq-wise, compile + PP=4 (1F1B): `LOG_RANK=3 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel NGPU=4 ./run_train.sh --compile.enable --parallelism.pipeline_parallel_degree=4 --parallelism.pipeline_parallel_schedule="1F1B" --parallelism.pipeline_parallel_microbatch_size=1`
```
step:  1  loss:  8.14354  grad_norm:  3.2394  memory:  0.78GiB(0.98%)  tps: 71  tflops: 0.04  mfu: 0.01%
step:  5  loss:  4.65961  grad_norm:  2.6585  memory:  0.80GiB(1.01%)  tps: 9,895  tflops: 5.37  mfu: 1.72%
step: 10  loss:  4.15489  grad_norm:  2.1032  memory:  0.80GiB(1.01%)  tps: 10,878  tflops: 5.91  mfu: 1.89%
```

seq-wise, compile + FSDP=2, PP=2 (Interleaved1F1B): `LOG_RANK=3 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel NGPU=4 ./run_train.sh --compile.enable --parallelism.pipeline_parallel_degree=2 --parallelism.data_parallel_shard_degree=2 --parallelism.pipeline_parallel_schedule="Interleaved1F1B" --parallelism.pipeline_parallel_microbatch_size=1`
```
step:  1  loss:  8.09411  grad_norm:  3.5881  memory:  2.64GiB(3.34%)  tps: 212  tflops: 0.12  mfu: 0.04%
step:  5  loss:  4.43106  grad_norm:  2.5952  memory:  2.69GiB(3.40%)  tps: 14,186  tflops: 7.70  mfu: 2.47%
step: 10  loss:  3.84682  grad_norm:  1.7512  memory:  2.69GiB(3.40%)  tps: 13,517  tflops: 7.34  mfu: 2.35%
```

seq-wise, FSDP=2, PP=2 (ZBV). no compile, no AC (ZBV + compile/AC fails): `LOG_RANK=1 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel NGPU=4 ./run_train.sh --parallelism.pipeline_parallel_degree=2 --parallelism.data_parallel_shard_degree=2 --parallelism.pipeline_parallel_schedule="ZBVZeroBubble" --parallelism.pipeline_parallel_microbatch_size=1 --activation_checkpoint.mode=none`
```
step:  1  loss:  8.20286  grad_norm:  3.3025  memory:  2.67GiB(3.38%)  tps: 1,364  tflops: 0.74  mfu: 0.24%
step:  5  loss:  4.45793  grad_norm:  2.4340  memory:  2.71GiB(3.43%)  tps: 9,726  tflops: 5.28  mfu: 1.69%
step: 10  loss:  3.96624  grad_norm:  1.8390  memory:  2.71GiB(3.43%)  tps: 9,671  tflops: 5.25  mfu: 1.68%
```

batch-wise + PP errors, as aux loss is not additive over microbatches: `NGPU=2 COMM_MODE=fake_backend MODULE=qwen3 CONFIG=qwen3_moe_debug ./run_train.sh --training.steps 1 --parallelism.pipeline_parallel_degree 2 --parallelism.pipeline_parallel_microbatch_size 4`
```
ValueError: batch_wise MoE aux loss is incompatible with pipeline parallelism because TorchTitan PP schedules split batches into pipeline microbatches. Use sequence_wise instead.
```

batch-wise + gradient accumulation errors, as aux loss is not additive over gradient accumulation microbatches: `NGPU=2 COMM_MODE=fake_backend MODULE=qwen3 CONFIG=qwen3_moe_debug ./run_train.sh --training.steps 1 --parallelism.data_parallel_shard_degree 2 --training.global_batch_size 16`
```
ValueError: batch_wise MoE aux loss is incompatible with gradient accumulation. Use sequence_wise instead or set training.global_batch_size to local_batch_size * data-parallel degree.
```


torchtitan/models/common/moe.py
```diff
@@ -140,6 +140,74 @@ def forward(
...
+                    bins=ctx.num_experts,  # pyrefly: ignore [missing-attribute]
+                    min=0,
+                    max=ctx.num_experts,  # pyrefly: ignore [missing-attribute]
+                )
+                aux_loss = MoE._batch_wise_aux_loss(
+                    scores_detached,
+                    num_tokens_per_expert,
+                    ctx.top_k,  # pyrefly: ignore [missing-attribute]
+                    ctx.aux_loss_weight,  # pyrefly: ignore [missing-attribute]
+                )
```
tianyu-l: why don't we compute these in forward?

torchtitan/models/qwen3_vl/model.py
```diff
@@ -80,6 +81,11 @@ def update_from_config(
                     layer_cfg.moe.router._debug_force_load_balance = (
                         debug.moe_force_load_balance
                     )
+                    apply_moe_load_balance_config(
```
tianyu-l: put this at config time, not run time

torchtitan/models/common/moe.py
```diff
@@ -140,6 +140,74 @@ def forward(
...
+            (aux_grad,) = torch.autograd.grad(aux_loss, scores_detached)
+        return grad_scores + aux_grad, None, None, None, None, None, None, None
```
pianpwk: note (@tianyu-l): prefer compute aux_loss in fwd & save

torchtitan/config/configs.py
```diff
@@ -79,6 +79,16 @@ class TrainingConfig:
...
 
+    moe_aux_loss_weight: float = 0.0
+    """Weight for the MoE auxiliary load-balance loss. 0 disables it."""
+
+    moe_aux_loss_type: Literal["sequence_wise", "batch_wise"] = "sequence_wise"
+    """Type of MoE auxiliary load-balance loss."""
+
+    moe_load_balance_coeff: float | None = 1e-3
+    """Coefficient for aux-loss-free bias-based MoE load balancing.
+    Overrides the model config default when set. None disables it."""
```
tianyu-l: do this in model_registry not allow to override per CLI job

torchtitan/models/common/moe.py
```diff
@@ -325,16 +421,29 @@ class Config(Module.Config):
         router: TokenChoiceTopKRouter.Config
         load_balance_coeff: float | None = 1e-3
         shared_experts: FeedForward.Config | None = None
+        aux_loss_weight: float = 0.0
+        """Weight for the auxiliary load-balance loss. 0 disables it."""
+        aux_loss_type: Literal["sequence_wise", "batch_wise"] = "sequence_wise"
+        """Type of auxiliary load-balance loss."""
+        aux_loss_local_batch_size: int | None = None
```
tianyu-l: abstract them into its own class

rakkit:
would be good if we add a buffer of `load_balance_loss` together with `tokens_per_expert` so one can log it out in pre-optim step hook.

tianyu-l:
> would be good if we add a buffer of load_balance_loss together with tokens_per_expert so one can log it out in pre-optim step hook.

@rakkit Logging capability will be separately addressed by @felipemello1 in a PR similar to https://github.com/pytorch/torchtitan/pull/2607. Will tag you for review.

torchtitan/models/common/moe.py
shuhuayu: iirc, this is something (nested autograd) @xmfan wanted to avoid for graph breaks? We can 1) write the grad function analytically, can be tedious or 2) use `torch.func.grad` to transform the aux loss function, which is perhaps the simplest approach, or 3) use `torch.library.custom_op`+ `register_autograd` as in `deepep.py`. I think all of these approaches can avoid the global config added in this pr `torch._dynamo.config.trace_autograd_ops = True`, which can have side effects globally. Maybe Simon can correct me if anything not making sense.

pianpwk: I went with `torch.func.grad` to remove the flag, but it seems the loss computation is still in backwards. I don't think we save any memory by computing the loss in FWD.

shuhuayu: Yeah, that's by design. We do not compute loss in the forward to avoid coupling with modeling loss which breaks pp. All we need is its gradients in the backward. On the other hand, if we want to log load balancing loss to monitor the healthiness of training, we can do it under `torch.no_grad()`in the forward.

torchtitan/models/common/moe.py
```diff
@@ -140,6 +140,122 @@ def forward(
...
+        max=num_experts,
+    )
+    p_i = scores.mean(dim=0)
+    f_i = num_tokens_per_expert.to(scores.dtype) * (
+        num_experts / (top_k * total_tokens)
+    )
+    return (f_i * p_i).sum() * aux_loss_weight
+
+
+class _AuxLossBackward(torch.autograd.Function):
```
shuhuayu: I think we can just use some regular name like `LoadBalanceAuxLoss`? Did not get why we need to put `Backward` only in the name.

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -285,6 +288,7 @@ def _debugmodel(
         router_top_k=3,
         router_score_func="softmax",
         score_before_experts=False,
+        aux_loss=MoELoadBalanceAuxLoss.Config(weight=1e-4),
```
tianyu-l: check technical report of dsv3 and see if it's actually used.

IIRC qwen3 used it so let's enable it for qwen3. Need to check gpt-oss as well.

torchtitan/trainer.py
```diff
@@ -245,6 +245,18 @@ def __init__(self, config: Config):
...
+        for layer_cfg in model_config.layers:  # pyrefly: ignore [missing-attribute]
+            if hasattr(layer_cfg, "moe") and layer_cfg.moe is not None:
+                aux_loss = layer_cfg.moe.aux_loss
+                if aux_loss.weight > 0:
+                    if aux_loss.type == "batch_wise" and pp_enabled:
+                        raise ValueError(
+                            "batch_wise MoE aux loss is incompatible with pipeline "
+                            "parallelism. Use sequence_wise instead."
+                        )
+                    aux_loss.local_batch_size = config.training.local_batch_size
```
tianyu-l: should be done at config time not here.

torchtitan/trainer.py
```diff
@@ -245,6 +245,18 @@ def __init__(self, config: Config):
         )
         self.model_config = model_config
 
+        pp_enabled = config.parallelism.pipeline_parallel_degree > 1
+        for layer_cfg in model_config.layers:  # pyrefly: ignore [missing-attribute]
+            if hasattr(layer_cfg, "moe") and layer_cfg.moe is not None:
+                aux_loss = layer_cfg.moe.aux_loss
+                if aux_loss.weight > 0:
+                    if aux_loss.type == "batch_wise" and pp_enabled:
+                        raise ValueError(
```
tianyu-l: should be done in `update_from_configs` fn, not here

torchtitan/models/common/moe.py
```diff
@@ -335,6 +511,8 @@ def __init__(self, config: Config):
         self.shared_experts = (
             config.shared_experts.build() if config.shared_experts is not None else None
         )
+        self.aux_loss = MoELoadBalanceAuxLoss(config.aux_loss)
+        self.aux_loss.top_k = config.router.top_k
```
tianyu-l: should be set in config time

torchtitan/models/common/moe.py
```diff
@@ -335,6 +511,8 @@ def __init__(self, config: Config):
         self.shared_experts = (
             config.shared_experts.build() if config.shared_experts is not None else None
         )
+        self.aux_loss = MoELoadBalanceAuxLoss(config.aux_loss)
```
tianyu-l: use build, not constructor

torchtitan/models/common/moe.py
```diff
@@ -297,7 +420,59 @@ def forward(
...
+
+    Injects aux loss gradients into router scores without changing the model
+    output (PP-safe). Call instance with router outputs to apply.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config:
+        weight: float = 0.0
+        """Weight for the auxiliary load-balance loss. 0 disables it."""
+        type: Literal["sequence_wise", "batch_wise"] = "sequence_wise"
```
tianyu-l: Let's have two subclasses, rather than using a str factory

torchtitan/models/common/moe.py
```diff
@@ -297,7 +420,59 @@ def forward(
...
+
+    def __call__(
+        self,
+        top_scores: torch.Tensor,
+        scores: torch.Tensor,
+        selected_experts_indices: torch.Tensor,
+        bs: int,
+        slen: int,
+    ) -> torch.Tensor:
+        if self.weight <= 0:
```
tianyu-l: why it would be < 0?

torchtitan/models/common/moe.py
```diff
@@ -297,7 +420,59 @@ def forward(
...
+    @dataclass(kw_only=True, slots=True)
+    class Config:
+        weight: float = 0.0
+        """Weight for the auxiliary load-balance loss. 0 disables it."""
+        type: Literal["sequence_wise", "batch_wise"] = "sequence_wise"
+        """Type of auxiliary load-balance loss."""
+        local_batch_size: int | None = None
+        """Total local batch size (before microbatching). Used to normalize aux
+        loss gradients across pipeline-parallel microbatches so they match the
+        non-PP case. Set automatically by the trainer; None means use the
```
tianyu-l: oh, could you remind me why not using global batch size?

torchtitan/models/common/moe.py
```diff
@@ -233,22 +352,26 @@ def _get_node_limited_routing_scores(
...
-            tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+            tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
                 - top_scores (torch.Tensor):
                     Routing scores for selected experts with shape ``(bs*slen, top_k)``.
                 - selected_experts_indices (torch.Tensor):
                     Expert indices selected for each token with shape ``(bs*slen, top_k)``.
                 - num_tokens_per_expert (torch.Tensor):
                     Number of tokens assigned to each expert with shape ``(num_experts,)``.
+                - scores (torch.Tensor):
+                    Full router scores for all experts with shape ``(bs*slen, num_experts)``.
```
tianyu-l: add reasons to have this field returned (for load balance loss)

torchtitan/models/common/moe.py
```diff
@@ -140,6 +140,125 @@ def forward(
...
+    f_i = num_tokens_per_expert.to(scores.dtype) * (
+        num_experts / (top_k * total_tokens)
+    )
+    return (f_i * p_i).sum() * aux_loss_weight
+
+
+class _AuxLossBase(torch.autograd.Function):
+    """Injects auxiliary load-balance loss gradients at the router scores level.
+
+    Identity in forward (returns ``top_scores`` unchanged). In backward,
```
tianyu-l: > Identity in forward (returns ``top_scores`` unchanged)

Is this even a goal?

torchtitan/models/common/moe.py
pianpwk: IIUC this is to match the non-microbatch-pipelining variant; the loss should be averaged across all tokens this sees, but this autograd fn will accumulate gradient across microbatches, so we normalize by micro_bs / local_bs .

torchtitan/models/common/moe.py
pianpwk: this is so it stays on the autograd graph, will update the docstring

torchtitan/models/deepseek_v3/__init__.py
pianpwk: yes this is from the hyperparams section; I'll check the other 2

pianpwk: gpt-oss seems to set 1e-3: https://github.com/huggingface/transformers/blob/6ffbb07f93d9e44457450d1150136309b0dc966b/src/transformers/models/gpt_oss/configuration_gpt_oss.py#L61, though I need to look closely if the loss is equivalent, it might be the batch-wise but with softmax scores: https://github.com/huggingface/transformers/blob/a8f43eca15b8d1c63deb33f6b97dfab30419e5da/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L504

torchtitan/models/common/moe.py
```diff
@@ -233,22 +348,27 @@ def _get_node_limited_routing_scores(
...
         Returns:
-            tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+            tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
                 - top_scores (torch.Tensor):
                     Routing scores for selected experts with shape ``(bs*slen, top_k)``.
                 - selected_experts_indices (torch.Tensor):
                     Expert indices selected for each token with shape ``(bs*slen, top_k)``.
                 - num_tokens_per_expert (torch.Tensor):
                     Number of tokens assigned to each expert with shape ``(num_experts,)``.
+                - scores (torch.Tensor):
```
shuhuayu: nit: can we put `scores` before `top_scores`? just put scores together.

torchtitan/models/qwen3/model.py
```diff
@@ -120,6 +121,13 @@ def update_from_config(
                             debug.moe_force_load_balance
                         )
 
+            if training is not None:
```
shuhuayu: why qwen3 has this condition check, what's the difference with other models?

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -199,6 +199,7 @@ def _build_gptoss_layers(
                 ),
                 top_k=top_k,
             ),
+            aux_loss=BatchWiseAuxLoss.Config(weight=1e-3),
```
shuhuayu: can you put some reference here where we know it uses batchwise

torchtitan/models/llama4/__init__.py
```diff
@@ -150,6 +151,7 @@ def _build_llama4_layers(
                 router=router,
                 experts=experts,
                 shared_experts=shared_experts,
+                aux_loss=BatchWiseAuxLoss.Config(weight=1e-3),
```
shuhuayu: similar

torchtitan/models/qwen3/__init__.py
```diff
@@ -172,6 +173,7 @@ def _build_qwen3_moe_layers(
                         comm_backend=moe_comm_backend,
                         non_blocking_capacity_factor=non_blocking_capacity_factor,
                     ),
+                    aux_loss=BatchWiseAuxLoss.Config(weight=1e-3),
                 ),
```
shuhuayu, line 193: similar, we need some references, may be helpful these two
1. https://arxiv.org/pdf/2505.09388 
2. https://arxiv.org/pdf/2501.11873

Is our batchwise auxloss equivalent to the global batch lbl? How do we compute average loss over dp ranks? Like the equation (4) in the paper 2 above.

torchtitan/models/common/moe.py
```diff
@@ -140,6 +141,120 @@ def forward(
...
+
+class _AuxLossBase(torch.autograd.Function):
+    """Injects auxiliary load-balance loss gradients at the router scores level.
+
+    Identity in forward (returns ``top_scores`` unchanged). In backward,
+    computes ``d(aux_loss)/d(scores)`` via ``torch.func.grad`` and adds it
+    to ``scores``'s gradient. ``top_scores`` is a pass-through so this node
+    remains in the autograd graph.
+
+    Uses ``torch.func.grad`` instead of nested ``torch.autograd.grad`` to
```
shuhuayu: I think avoiding this global config is part our code iterating process and no need to reveal it to the users.

torchtitan/models/common/moe.py
```diff
@@ -140,6 +141,120 @@ def forward(
...
+        ctx.top_k = top_k  # pyrefly: ignore [missing-attribute]
+        ctx.aux_loss_weight = aux_loss_weight  # pyrefly: ignore [missing-attribute]
+        return top_scores
+
+    @staticmethod
+    def _backward_impl(loss_fn, ctx, grad_top_scores):
+        (
+            scores,
+            selected_experts_indices,
+        ) = ctx.saved_tensors
```
shuhuayu: maybe add one line comment if we want to be explicit. 
```suggestion
        ) = ctx.saved_tensors
        # torch.func.grad avoids the graph break that torch.autograd.grad causes under torch.compile
```

torchtitan/models/qwen3/model.py
pianpwk: gonna remove this

torchtitan/models/common/moe.py
```diff
@@ -140,6 +141,118 @@ def forward(
...
+    """Sequence-wise auxiliary load-balance loss (DeepSeek-V3 Eqs 17-20)."""
+    num_experts = scores.size(-1)
+    scores_per_seq = scores.view(bs, slen, num_experts)
+    denom = scores_per_seq.sum(dim=-1, keepdim=True) + 1e-20
+    probs_per_seq = scores_per_seq / denom
+    p_i = probs_per_seq.mean(dim=1)
+    indices_per_seq = selected_experts_indices.view(bs, -1)
+    offset = torch.arange(bs, device=indices_per_seq.device).unsqueeze(1) * num_experts
+    flat_indices = (indices_per_seq + offset).reshape(-1)
+    counts = torch.bincount(flat_indices.long(), minlength=bs * num_experts)
```
shuhuayu: Can you double check whether `torch.bincout` leads to a graph break under torch.compile? Maybe we need `.scatter_add` on a sparse tensor.

torchtitan/models/gpt_oss/__init__.py
pianpwk: I actually couldn't find arxiv references, so left hf urls - maybe that's inaccurate/weird. also don't have something for llama4 yet.

torchtitan/models/qwen3/__init__.py
pianpwk: this is batch-wise aux loss yes, currently we ban with PP because it's not additive across microbatches, maybe we should ban with DP too until the count accumulator is implemented.

torchtitan/models/common/moe.py
pianpwk: This runs under compile, I believe `capture_scalar_outputs` enables capturing the output shape.

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -415,11 +420,12 @@ def _236b(
         num_experts=num_experts,
         num_shared_experts=num_shared_experts,
         router_top_k=6,
-        router_score_func="softmax",
+        router_score_func="sigmoid",
```
tianyu-l, line 439: Why changing default? If aux loss is only supported with sigmoid, let's delete this model variant.

torchtitan/models/qwen3/__init__.py
```diff
@@ -132,7 +133,13 @@ def _build_qwen3_moe_layers(
     moe_comm_backend: str,
     non_blocking_capacity_factor: float | None = None,
 ) -> list[TransformerBlock.Config]:
-    """Build per-layer configs for MoE Qwen3 models with depth-scaled inits."""
+    """
+    Build per-layer configs for MoE Qwen3 models with depth-scaled inits.
```
tianyu-l, line 145: should `load_balance_coeff` be None for qwen and llama4, as they use aux loss?

torchtitan/models/common/moe.py
```diff
@@ -297,7 +415,80 @@ def forward(
...
+
+    Subclass to select loss variant (sequence-wise or batch-wise).
+    """
+
+    _autograd_fn: type[_AuxLossBase]
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        weight: float = 0.0
+        """Weight for the auxiliary load-balance loss. 0 disables it."""
```
tianyu-l: too many ways to disable it -- let's assert `weight>0` here. If users want to disable it, they can set moe.aux_loss to None.

tianyu-l: also, should it be called `coeff` to be consistent with loss-free load balance coeff and with other repo?

torchtitan/models/common/moe.py
```diff
@@ -297,7 +415,80 @@ def forward(
...
+            slen,
+            self.top_k,
+            scaled_weight,
+        )
+
+
+class SequenceWiseAuxLoss(MoELoadBalanceAuxLoss):
+    """Sequence-wise auxiliary load-balance loss (DeepSeek-V3 Eqs 17-20)."""
+
+    _autograd_fn = _SequenceWiseAuxLoss
```
tianyu-l: I didn't get why we need so many layers of nesting.
Can we have `MoELoadBalanceAuxLoss` inheriting both `Configurable` and `torch.autograd.Function` and just have one layer?

torchtitan/models/common/moe.py
```diff
@@ -140,6 +141,118 @@ def forward(
         return self.token_dispatcher.combine(routed_output, metadata, x, shared_experts)
 
 
+def _sequence_wise_aux_loss(
```
tianyu-l: can we inline these inside the autograd function?

torchtitan/trainer.py
```diff
@@ -241,11 +241,24 @@ def __init__(self, config: Config):
...
+            f"data-parallel degree ({global_batch_size} "
+            f"% ({config.training.local_batch_size} * {batch_degree}) != 0)"
+        )
+
         # build model (using meta init)
         model_config = model_spec.model
         # set the model args from training job configs
         model_config.update_from_config(
             trainer_config=config,
+            global_batch_size=global_batch_size,
```
tianyu-l, line 239: let's not pass this in, we can have some redundant logic for getting `global_batch_size` (although, not clear to me what the correct math yet, would appreciate a discussion first)

torchtitan/trainer.py
```diff
@@ -241,11 +241,24 @@ def __init__(self, config: Config):
...
+        if global_batch_size < 0:
+            global_batch_size = config.training.local_batch_size * batch_degree
+        assert global_batch_size > 0
+        assert (
+            global_batch_size % (config.training.local_batch_size * batch_degree) == 0
+        ), (
+            f"global batch size must be multiple of local batch size times "
+            f"data-parallel degree ({global_batch_size} "
+            f"% ({config.training.local_batch_size} * {batch_degree}) != 0)"
+        )
```
tianyu-l, line 232: prefer to not change the location for now

torchtitan/models/common/moe.py
```diff
@@ -297,7 +415,80 @@ def forward(
...
+    Subclass to select loss variant (sequence-wise or batch-wise).
+    """
+
+    _autograd_fn: type[_AuxLossBase]
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Configurable.Config):
+        weight: float = 0.0
+        """Weight for the auxiliary load-balance loss. 0 disables it."""
+        top_k: int = 1
```
tianyu-l: shouldn't have a default value. It should be either set in config setup, or passed from other module at run time.

torchtitan/models/common/moe.py
```diff
@@ -297,7 +415,80 @@ def forward(
...
+        self,
+        top_scores: torch.Tensor,
+        scores: torch.Tensor,
+        selected_experts_indices: torch.Tensor,
+        bs: int,
+        slen: int,
+    ) -> torch.Tensor:
+        if self.weight == 0:
+            return top_scores
+        global_bs = self.global_batch_size or bs
```
tianyu-l: this is unclear what it is doing

torchtitan/models/common/moe.py
```diff
@@ -297,7 +415,80 @@ def forward(
...
+        scaled_weight = self.weight * bs / global_bs
```
tianyu-l: let's use consistent variable name

torchtitan/models/common/moe.py
tianyu-l: Returning top_scores unchanged sounds necessary, but is it necessary to save all activations for backward and do no-op in forward?

torchtitan/models/deepseek_v3/__init__.py
pianpwk: need another pair of eyes, but I think default might been wrong? paper looks like sigmoid + route_norm

Image

torchtitan/models/common/moe.py
pianpwk: think this means no normalization when global_batch_size isn't passed (e.g. no microbatching).

torchtitan/models/common/moe.py
```diff
@@ -297,7 +303,169 @@ def forward(
...
+        """Batch-wise auxiliary load-balance loss."""
+        num_experts = scores.size(-1)
+        total_tokens = scores.size(0)
+        num_tokens_per_expert = torch.histc(
+            selected_experts_indices.view(-1).float(),
+            bins=num_experts,
+            min=0,
+            max=num_experts,
+        )
+        p_i = scores.mean(dim=0)
```
shuhuayu: this seems to be wrong. When computing batch-wise mean, for expert `i`, we should first compute each token's probability of of choosing expert `i`, then compute the average over all tokens.

torchtitan/models/common/moe.py
```diff
@@ -297,7 +303,169 @@ def forward(
...
+        top_scores: torch.Tensor,
+        scores: torch.Tensor,
+        selected_experts_indices: torch.Tensor,
+        bs: int,
+        slen: int,
+        top_k: int,
+        aux_loss_coeff: float,
+        loss_type: str,
+    ) -> torch.Tensor:
+        ctx.save_for_backward(scores, selected_experts_indices)
```
shuhuayu: we do not need to save `selected_exeprts_indices`, we only need to save the values of `f_i`.

shuhuayu (changes requested):

torchtitan/models/common/moe.py
```diff
@@ -292,6 +294,268 @@ def forward(
...
+                self.top_k,
+                scaled_coeff,
+                dp_mesh=self.dp_mesh,
+                cp_mesh=self.cp_mesh,
+                tp_mesh=self.tp_mesh,
+            )
+        return _AuxLossAutograd.apply(topk_scores_BLK, aux_loss)
+
+
+def configure_moe_aux_loss_reduction(model: nn.Module, parallel_dims) -> None:
```
tianyu-l, line 532: we can now get these meshes from current mesh global context, right?

torchtitan/models/common/moe.py
```diff
@@ -292,6 +294,268 @@ def forward(
...
+class MoELoadBalanceAuxLoss(Module):
+    """MoE auxiliary load-balance loss.
+
+    Injects aux loss gradients into router scores without changing the model
+    output (PP-safe). Call instance with router outputs to apply.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        type: Literal["sequence_wise", "batch_wise"]
```
tianyu-l, line 456: more extensible way should be letting two variants inherit `MoELoadBalanceAuxLoss` which serves as a base class.

torchtitan/models/common/moe.py
```diff
@@ -292,6 +294,268 @@ def forward(
...
+            aux_loss = _sequence_wise_load_balance_loss(
+                scores_BLE,
+                topk_expert_ids_BLK,
+                self.top_k,
+                scaled_coeff,
+                cp_mesh=self.cp_mesh,
+                tp_mesh=self.tp_mesh,
+            )
+        else:
+            aux_loss = _batch_wise_load_balance_loss(
```
tianyu-l, line 520: Should we error out when PP / gradient accumulation is used?

Then I'm worried because our RL stack would almost always need gradient accumulation, so we probably should try to make it work.

tianyu-l (commented):
Is the plan to land this or https://github.com/pytorch/torchtitan/pull/3864?

tianyu-l:
following up in https://github.com/pytorch/torchtitan/pull/3864

PR #3001, opened 2026-04-16
[rl] Generator refactor

joecummings:
### Summary

  Narrow the RL `VLLMGenerator` to pure generation. Reward, advantage, and grouping
  now live in the controller + `Grader`, not in the generator's output.

  ### Changes

  **Types** (`types.py`)
  - Split the old `Episode` into three: `Completion` (generator output — no reward,
    no advantage, no group), `ScoredCompletion` (Completion + scalar reward from
    Grader), and `Episode` (what the trainer consumes — flat scored sample with
    advantage). ;`token_log_probs` → `token_logprobs` (matches `TrainBatch`).

  **Generator** (`actors/generator.py`)
  - `generate(prompts) -> list[Completion]`. No `expected_answers` arg, no group
    stamping. `prompt_idx` is pinned to `int(output.request_id)`.
  - `prompt_logprobs=1` dropped (no consumer, saves compute).
  - `num_samples_per_prompt` moved from `VLLMGenerator.Config` into
    `SamplingConfig.n` to match vLLM.
  - `generate(... *, sampling_config=None)`: per-call sampling override is now
    kw-only and disambiguated from the `SamplingConfig` type.

  **Grader** (`actors/grader.py`)
  - `score(completions: list[Completion], expected_answer: str) -> list[ScoredCompletion]`.
    Scores one prompt's group per call so `reward_fn` sees its natural batched
    shape. Controller loops over prompts, issuing one RPC each.

  **Controller** (`simple_grpo_sum_digits.py`)
  - Groups `Completion`s by `prompt_idx`, scores each group, computes mean-baseline
    advantage per group in a single pass, builds `Episode`s.
  - `evaluate()` uses the new `Completion` type; `completions[i * n]` stride picks
    the first sample per prompt.
  - `_log_samples` refocused on debug output only: prints `prompt_idx`, optional
    reward, and truncated text. Accepts `Episode`s or `Completion`s.

  **Task** (`sum_digits.py`)
  - `reward_function(completions: list[Completion], expected_answer)` consumes the
    new `Completion` type directly.

  **Configs** (`config_registry.py`, `tests/integration_tests.py`)
  - All five configs migrated from `num_samples_per_prompt` to
    `SamplingConfig.n`. Integration-test CLI flag updated.

  ### Test Plan

  - Trained `rl_grpo_qwen3_0_6b` end-to-end (10 steps, TP=2 trainer + TP=4
    generator on 6 H100s). Clean exit, eval accuracy 40% → 50%. https://www.internalfb.com/phabricator/paste/view/P2282561759


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -107,15 +110,14 @@ class VLLMGenerator(Actor, Configurable):
...
-    Maintains a vLLM engine that is synchronized with the Trainer
-    via weight sync. Generates completions for given prompts and
-    computes rewards/advantages.
+    Maintains a vLLM engine synchronized with the Trainer via weight
+    sync. ``generate()`` produces a flat list of Completions; reward
+    and advantage computation live in the controller and Grader.
 
     Args:
         config: Generator-specific configuration.
+        model_spec: TorchTitan model specification.
```
joecummings, line 119: This was undocumented before.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -243,82 +237,74 @@ def _get_model(self):
     @endpoint
     async def generate(
         self,
-        prompt_texts: list[str],
-        expected_answers: list[str],
-    ) -> list[Episode]:
-        """Generate completions and return a flat list of Episodes.
+        prompts: list[str],
```
joecummings, line 240: This now matches the standard vLLM generate: https://docs.vllm.ai/en/stable/getting_started/quickstart/#offline-batched-inference

joecummings: Just keeping "config" instead of "params" to be in line with Titan

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -243,82 +237,74 @@ def _get_model(self):
...
+        *,
+        sampling_config: SamplingConfig | None = None,
```
joecummings, line 242: Allowing this here b/c we eventually WILL want to support different sampling configs e.g. per task and also for eval versus training.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -431,43 +432,56 @@ async def setup(
...
-            extracted = extract_answer(ep.text)
-            is_correct = extracted == int(answer)
-            has_tag = bool(re.search(r"\[ANSWER\]", ep.text))
-            correct += int(is_correct)
-            format_ok += int(has_tag)
+        for i, answer in enumerate(eval_answers):
+            c = completions[i * n]
+            extracted = extract_answer(c.text)
+            correct += int(extracted == int(answer))
+            format_ok += int(bool(re.search(r"\[ANSWER\]", c.text)))
```
joecummings, line 462: In a follow up this should be the same as the task itself. That's out of scope for this refactor.

torchtitan/experiments/rl/types.py
```diff
@@ -10,41 +10,45 @@
...
+    completion: Completion
+    reward: float
+
+
+@dataclass
+class Episode:
+    """Training sample: flattened scored completion + GRPO advantage.
+
+    Flat shape (rather than composition) because the trainer collate
+    path and logging read these fields directly.
```
tianyu-l, line 41: This doesn't sound fundamental?

Also, even if we keep it flat, it may be better to inherit `Completion` for maintainability?

tianyu-l (approved):
Looks very neat. Had one comment.

daniellepintz:
in the output from your test command, why is the weight sync so slow? are you not using gpu direct? something seems off there, for .6B it should be ~.1 seconds

torchtitan/experiments/rl/actors/grader.py
```diff
@@ -35,31 +36,30 @@ def __init__(
...
         """
-        logger.debug(f"Grader scoring {len(episodes)} episodes...")
-
-        # Score each episode individually
-        for ep in episodes:
-            rewards = self.reward_fn([ep.text], ep.expected_answer)
-            ep.reward = rewards[0].item()
+        rewards = self.reward_fn(completions, expected_answer)
+        scored = [
+            ScoredCompletion(completion=c, reward=r.item())
```
felipemello1, line 55: not a huge fan of ScoredCompletion. I would prefer that after we get a completion, we start putting data in an Episode.

Would this dataclass be used only here? If so, then i think it might be too ephemeral.

i wouldnt block the PR on this though.

felipemello1: a simple solution could be to build episodes, pass them here and have the function update and return them, or just return the rewards themselves, to reduce data transfer

torchtitan/experiments/rl/sum_digits.py
```diff
@@ -75,30 +77,32 @@ def get_system_prompt(self) -> str:
 
     def reward_function(
```
felipemello1, line 78: this is confusing. We have another reward function here. Whats the difference between sum_digits and simple_sum_digits? can we delete one of them?

torchtitan/experiments/rl/types.py
felipemello1: Personally, I would prefer to not inherit from Completion. I would consider adding it as a argument, e.g. Episode.completion.text, but my intuition is that leaving it flat is the best call. 

People may have all sort of logic in different places, and i don't think that we should enforce the coupling, e.g. let's say that the generator outputs `CompletionWithLargeMetadata` that i don't want to add to my Episode.

felipemello1 (approved):
left some comments, nothing blocking

torchtitan/experiments/rl/sum_digits.py
joecummings: Next refactor! Yeah I don't love this either but didn't want to change reward structure in this PR

torchtitan/experiments/rl/actors/grader.py
joecummings: Ack - I wanted a data structure to hold the reward per completion but with a better handling of rewards in general this is not as important. 

My general sense is Grader doesn't do a ton here overall.

PR #2982, opened 2026-04-15
Increase timeout for features test to 60 minutes.

akashveramd:
In this PR, we have increased the timeout for features test to 60 minutes. This is the timeout failure we are seeing for features test https://github.com/pytorch/torchtitan/actions/runs/24430483234/job/71373655959. The increased timeout is helping features test to pass on ROCm CI https://github.com/pytorch/torchtitan/actions/runs/24541348555/job/71747650589. 

This is a temporary fix for ROCm CI timeout issue. We are working towards enabling MI350 label and replace the existing MI325 label. Hopefully it shouldn't result in a timeout. But we need to enable it and check. 
This is the draft PR for enabling MI350 label https://github.com/pytorch/torchtitan/pull/2740

The failing transformers test seems unrelated to the timeout change made in features test https://github.com/pytorch/torchtitan/actions/runs/24466095345/job/71493479334?pr=2982.

PR #3002, opened 2026-04-16
ci: install torchvision alongside torch in transformers backend workflow

rishisinhanj:
## Summary

The ROCm path of `integration_test_8gpu_transformers_modeling_backend.yaml` reinstalls `torch` from the ROCm nightly index but doesn't reinstall `torchvision`, leaving the docker image's stale `torchvision==0.26.0` (built against `torch==2.11.0`). pip's resolver flags this explicitly:

```
torchvision 0.26.0 requires torch==2.11.0,
  but you have torch 2.12.0.dev20260411+rocm7.1 which is incompatible.
```

When `transformers.modeling_utils` eagerly imports torchvision (via `loss_for_object_detection.py` → `image_transforms.py` → `image_utils.py`), the registration `@torch.library.register_fake("torchvision::nms")` fails because the op doesn't exist in the freshly-installed torch nightly:

```
RuntimeError: operator torchvision::nms does not exist
```

This crashes the test at import time, before any GPU code runs.

## Reproduction

Failing scheduled run on `main`: https://github.com/pytorch/torchtitan/actions/runs/24270213946/job/70873527337

Reproduced locally on MI325X with `rocm/pytorch:latest`:
```bash
pip install --force-reinstall --pre torch \
  --index-url https://download.pytorch.org/whl/nightly/rocm7.1
python -c "import torchvision"
# RuntimeError: operator torchvision::nms does not exist
```

After the fix:
```bash
pip install --force-reinstall --pre torch torchvision \
  --index-url https://download.pytorch.org/whl/nightly/rocm7.1
python -c "import torchvision; print(torchvision.__version__)"
# OK
```

## Test plan

- [ ] CI's ROCm `build-test (rocm, linux.rocm.gpu.gfx942.8, ...)` job goes green on this PR
- [ ] CUDA `build-test (cuda, ...)` job stays green (no regression)

## Note on a separate latent issue

While reproducing, I also hit an unrelated `TypeError: HFTransformerModel.Config field 'transformers_version' must be keyword-only` at `model.py:83` with `transformers >= 5.5.x` (inherited from `PretrainedConfig`). That's downstream of this fix — CI's docker image currently has an older `transformers` so it doesn't trigger yet. Worth filing separately once this lands and the ROCm path actually gets that far.

### DIFF
--- a/.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
+++ b/.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
@@ -64,6 +64,8 @@ jobs:
         if [ -n "${{ matrix.torch-version }}" ]; then
           TORCH_SPEC="torch==${{ matrix.torch-version }}"
         fi
+        # NOTE: reinstall torchvision alongside torch so their ABIs stay in sync;
+        # otherwise transformers' transitive `import torchvision` fails at test time.
         python -m pip install --force-reinstall --pre \
           "${TORCH_SPEC}" torchvision --index-url ${{ matrix.index-url }}
 
--- a/.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
+++ b/.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
@@ -65,7 +65,7 @@ jobs:
           TORCH_SPEC="torch==${{ matrix.torch-version }}"
         fi
         python -m pip install --force-reinstall --pre \
-          "${TORCH_SPEC}" --index-url ${{ matrix.index-url }}
+          "${TORCH_SPEC}" torchvision --index-url ${{ matrix.index-url }}
 
         if [[ "${{ matrix.gpu-arch-type }}" == "rocm" ]]; then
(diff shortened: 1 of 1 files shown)


rishisinhanj:
Also seen on #2982 — same traceback, same job: https://github.com/pytorch/torchtitan/actions/runs/24466095345/job/71493479334

.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
```diff
@@ -65,7 +65,7 @@ jobs:
           TORCH_SPEC="torch==${{ matrix.torch-version }}"
         fi
         python -m pip install --force-reinstall --pre \
-          "${TORCH_SPEC}" --index-url ${{ matrix.index-url }}
+          "${TORCH_SPEC}" torchvision --index-url ${{ matrix.index-url }}
```
tianyu-l, line 70: Sounds reasonable to me. Could you add a NOTE comment nearby, so that people who read / maintain the code can get context? Thanks!

tianyu-l (approved):

.github/workflows/integration_test_8gpu_transformers_modeling_backend.yaml
rishisinhanj: Thanks @tianyu-l - I added the note for context. Let me know if there are any other issues!

PR #3008, opened 2026-04-17
Fix: reproducible training resume across epoch boundaries for map and streaming datasets

slimfrkha:
### Problem
Resuming from a checkpoint taken after at least one full epoch re-loop produced wrong data for streaming datasets in both `HuggingFaceTextDataset` and `ChatDataset`.

Two bugs fixed:

**1. HuggingFace epoch guard mismatch** — HF's `IterableDataset` restores shard position only if `self.epoch == self._starting_state_dict["epoch"]`. A freshly constructed dataset starts at epoch 0, so the guard silently fails for checkpoints taken at epoch N > 0, restarting iteration from shard 0. Fixed by calling `set_epoch(saved_epoch)` before `load_state_dict`.

**2. In-flight sample lost at yield boundary (`ChatDataset`)** — when a sample didn't fit the current buffer, it lived only in the suspended generator frame. On resume, a fresh generator lost it, skipping one sample. Fixed by persisting the in-flight sample in `_pending_input_ids`/`_pending_label_ids` and consuming it at generator startup.

### Changes
- `HuggingFaceTextDataset.load_state_dict`: sync HF epoch before restoring state
- `ChatDataset.load_state_dict`: same fix for streaming path
- `ChatDataset`: add pending sample to checkpoint state
- Tests: consume >1 epoch to exercise the re-loop resume path

### DIFF
--- a/torchtitan/hf_datasets/text_datasets.py
+++ b/torchtitan/hf_datasets/text_datasets.py
@@ -208,7 +208,12 @@ class HuggingFaceTextDataset(IterableDataset, Stateful):
                 )
         else:
             assert "data" in state_dict
-            self._data.load_state_dict(state_dict["data"])
+            data_state = state_dict["data"]
+            # HuggingFace IterableDataset sync epoch
+            saved_epoch = data_state.get("epoch", 0)
+            if saved_epoch != self._data.epoch:
+                self._data.set_epoch(saved_epoch)
+            self._data.load_state_dict(data_state)
 
     def state_dict(self):
         _state_dict: dict[str, Any] = {
@@ -319,6 +324,8 @@ class ChatDataset(IterableDataset, Stateful):
         self._inputs_buffer: list[int] = []
         self._labels_buffer: list[int] = []
         self._positions_buffer: list[int] = []
+        self._pending_input_ids: list[int] = []
+        self._pending_label_ids: list[int] = []
 
         self._logged_first_sample = False
 
@@ -409,6 +416,18 @@ class ChatDataset(IterableDataset, Stateful):
         The model's flex/varlen attention mask uses these EOS positions to
         prevent cross-document attention.
         """
+        # resume from ckpt edge case
+        if self._pending_input_ids:
+            input_ids = self._pending_input_ids
+            label_ids = self._pending_label_ids
+            self._pending_input_ids = []
+            self._pending_label_ids = []
+            self._inputs_buffer.extend(input_ids)
+            self._labels_buffer.extend(label_ids)
+            self._positions_buffer.extend(range(len(input_ids)))
+            self._sample_idx += 1
+            if len(self._inputs_buffer) == self.seq_len:
+                yield self._flush_buffers()
         while True:
             for sample in self._get_data_iter():
                 # pyrefly: ignore [bad-argument-type]
@@ -426,8 +445,14 @@ class ChatDataset(IterableDataset, Stateful):
                     self._inputs_buffer.extend([self._eos_id] * pad_len)
                     self._labels_buffer.extend([IGNORE_INDEX] * pad_len)
                     self._positions_buffer.extend(range(pad_len))
-
+                    self._pending_input_ids = input_ids
+                    self._pending_label_ids = label_ids
                     yield self._flush_buffers()
+                    # resumed generator continues here or fresh generator handles pending at top (resume path)
+                    input_ids = self._pending_input_ids
+                    label_ids = self._pending_label_ids
+                    self._pending_input_ids = []
+                    self._pending_label_ids = []
 
                 # Add example to buffer with positions resetting to 0
                 self._inputs_buffer.extend(input_ids)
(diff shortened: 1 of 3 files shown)


slimfrkha:
## Suggestion: simplify `ChatDataset` iteration

The fixes in this PR expose a deeper issue: `_iter_greedy_packed` is complex enough to hide subtle checkpointing bugs (in-flight sample loss, pending state across yield boundaries). This complexity comes from greedy packing, which is the right strategy for pretraining but questionable for SFT.

For SFT, the simplest and most correct approach is to yield one sample per sequence, padded to `seq_len`:
- No buffer state → checkpointing is trivial
- No cross-document attention concerns
- Matches inference distribution (single sample per forward pass)
- Naturally extends to multi-turn conversations (just update `_validate_messages` and `_tokenize_sample`)

GPU utilization is lower due to padding, but for SFT dataset sizes this is rarely the bottleneck and is a reasonable tradeoff for correctness and maintainability.

Worth considering as a follow-up. cc @joecummings

joecummings:
@slimfrkha Thanks for the work here - reviewing more in-depth now!

To your point about removing packing entirely from SFT, I'd like to address what seems to be the main concern: This packing code is somewhat complex to handle all the edge cases. That's very fair

Right now, we're doing packing entirely with streaming datasets from Hugging Face, which is nice for our quick experiments. In a true production environment, we might like to rely on a pre-packed setup for pretraining and sft data so we mimic a map-based dataset with set indices. This is actually the direction we'd like to move in using `Grain` for Torchtitan. 

In the meantime though, packing is a valuable part of any reasonable sized training solution including for SFT. The [Olmo3 Think sft dataset](https://huggingface.co/datasets/allenai/Dolci-Think-SFT-Olmo-Hybrid
) is roughly 6B tokens. Assuming we need a set seq_len (for nice torch.compile support - even faster),  we'd waste a ton of flops on padding and training time would increase significantly.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -188,7 +188,12 @@ def load_state_dict(self, state_dict):
             self._sample_idx = state_dict["sample_idx"]
         else:
             assert "data" in state_dict
-            self._data.load_state_dict(state_dict["data"])
+            data_state = state_dict["data"]
+            # HuggingFace IterableDataset sync epoch
+            saved_epoch = data_state.get("epoch", 0)
```
joecummings, line 193: The `set_epoch` is lightweight, can just use it directly: https://github.com/huggingface/datasets/blob/b0de7a8d787b83c1fe0780e4960364e2171c8c36/src/datasets/iterable_dataset.py#L2880-L2881

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -406,8 +425,14 @@ def _iter_greedy_packed(self):
                     self._inputs_buffer.extend([self._eos_id] * pad_len)
                     self._labels_buffer.extend([IGNORE_INDEX] * pad_len)
                     self._positions_buffer.extend(range(pad_len))
-
+                    self._pending_input_ids = input_ids
+                    self._pending_label_ids = label_ids
                     yield self._flush_buffers()
+                    # resumed generator continues here or fresh generator handles pending at top (resume path)
+                    input_ids = self._pending_input_ids
```
joecummings, line 432: These will still maintain their state after the yield, why do we need to reset input_ids (etc) to pending_input_ids here?

torchtitan/hf_datasets/text_datasets.py
slimfrkha: used `set_epoch` just like what was done in `HuggingFaceTextDataset.__iter__`

torchtitan/hf_datasets/text_datasets.py
slimfrkha: After a checkpoint restore, a fresh generator is created and the suspended frame, including input_ids, is gone. `_pending_input_ids` persists the in-flight sample across that restart; the reassignment after the yield is a no-op for the live generator but ensures the fresh generator's startup path (top of `_iter_greedy_packed`) and the live generator's resume path share the same code.

slimfrkha:
cc @joecummings @tianyu-l 
Can you please have a look ? should be all good.

Edit:
@tianyu-l  i am also following #3023 
Should we rebase here as well as some test are not triggered / failing?

torchtitan/hf_datasets/text_datasets.py
joecummings: Sorry, I should've been more clear - I think we can remove the `if saved_epoch != self._data.epoch`

joecummings (approved):
Could you rebase before we merge?

torchtitan/hf_datasets/text_datasets.py
slimfrkha: got it ! it is done now !

slimfrkha:
> Could you rebase before we merge?

rebased. ready to merge when CI clears.

slimfrkha:
FYI @tianyu-l both failing tests are irrelevant to this PR.

tianyu-l (approved):

PR #2990, opened 2026-04-15
fix CP `n_tokens_seen` overcounting

pianpwk:
Fixes https://github.com/pytorch/torchtitan/issues/2983

Before (DP=4, CP=1): `torchrun --nproc_per_node=4 --local-rank 0 -m torchtitan.train --module llama3 --config llama3_debugmodel_flex_attn --training.steps 5 --metrics.log_freq 1 --parallelism.data_parallel_shard_degree 4 --parallelism.context_parallel_degree 1 --training.global_batch_size 32 --training.seq_len 2048`
```
[titan] 2026-04-15 16:43:45,087 - root - INFO - [DEBUG] n_tokens_seen: 327680.0
[titan] 2026-04-15 16:43:45,087 - root - INFO - [31mstep:  5  [32mloss:  5.59510  [38;2;180;60;0mgrad_norm:  2.4778  [38;2;54;234;195mmemory:  1.09GiB(1.38%)  [34mtps: 104,455  [36mtflops: 7.48  [35mmfu: 2.40%[39m
```

Before (DP=2, CP=2): `torchrun --nproc_per_node=4 --local-rank 0 -m torchtitan.train --module llama3 --config llama3_debugmodel_flex_attn --training.steps 5 --metrics.log_freq 1 --parallelism.data_parallel_shard_degree 2 --parallelism.context_parallel_degree 2 --training.global_batch_size 32 --training.seq_len 2048`
```
[titan] 2026-04-15 16:37:32,439 - root - INFO - [DEBUG] n_tokens_seen: 655360.0
[titan] 2026-04-15 16:37:32,440 - root - INFO - [31mstep:  5  [32mloss:  5.63218  [38;2;180;60;0mgrad_norm:  2.4113  [38;2;54;234;195mmemory:  0.65GiB(0.82%)  [34mtps: 50,305  [36mtflops: 3.60  [35mmfu: 1.15%[39m
```

After (DP=4, CP=1):
```
[titan] 2026-04-15 16:44:35,135 - root - INFO - [DEBUG] n_tokens_seen: 327680.0
[titan] 2026-04-15 16:44:35,135 - root - INFO - [31mstep:  5  [32mloss:  5.60330  [38;2;180;60;0mgrad_norm:  2.2935  [38;2;54;234;195mmemory:  1.09GiB(1.38%)  [34mtps: 102,444  [36mtflops: 7.33  [35mmfu: 2.35%[39m
```

After (DP=2, CP=2):
```
[...]

PR #3009, opened 2026-04-17
fix: normalize n_tokens_seen by cp_degree when context parallelism is…

### LINKED ISSUE #2983: n_tokens_seen metric inflated by context_parallel_degree when CP is enabled
### Bug description

When context parallelism is enabled, the `n_tokens_seen` metric reported in logs/wandb is inflated by `cp_degree`. This makes it impossible to compare training curves (e.g., loss vs tokens seen) across runs with different CP configurations.

`tps` and `mfu` are correctly normalized via `non_data_parallel_size`, so this only affects `n_tokens_seen`.

## Possible root cause

In `trainer.py` `batch_generator`, `labels.numel()` is counted before `context_parallel` chunks the labels:

https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L542-L544

```python
ntokens_batch = labels.numel()          # full seq_len, before CP split
self.ntokens_seen += ntokens_batch
self.metrics_processor.ntokens_since_last_log += ntokens_batch
```

Then `n_tokens_seen` is aggregated via `dist_sum` over `loss_mesh`:

https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L793-L802

```python
global_avg_loss, global_max_loss, global_ntokens_seen = (
    dist_utils.dist_sum(loss, loss_mesh),
    dist_utils.dist_max(local_avg_loss, loss_mesh),
    dist_utils.dist_sum(
        torch.tensor(
            self.ntokens_seen, dtype=torch.int64, device=self.device
        ),
        loss_mesh,
    ),
)
```

Where `loss_mesh` size = `dp_replicate x dp_shard x cp` (includes CP). Both CP ranks in the same DP group load the same full-size batch from the dataloader, so the same tokens are counted `cp_degree` times.

## Reproduction

Same model, same `global_batch_size=32`, same `seq_len=98304`:

| Config | grad_accum | per rank / step | dist_sum (loss_mesh, 8 ranks) | actual tokens / step | inflation |
|---|---|---|---|---|---|
| dp=8, cp=1 | 4 | 4 x 98304 = 393K | **3.15M** ✓ | 3.15M | 1x |
| dp=4, cp=2 | 8 | 8 x 98304 = 786K | **6.29M** ✗ | 3.15M | 2x |
| dp=2, cp=4 | 16 | 16 x 98304 = 1,573K | **12.58M** ✗ | 3.15M | 4x |

The inflation factor equals `cp_degree` exactly. When `dp` shrinks to make room for CP, `grad_accum` increases proportionally, and each CP rank counts the full sequence before the split.

On wandb, the CP=2 loss curve is stretched 2x and CP=4 is stretched 4x along the x-axis when plotted against `n_tokens_seen`, making it appear that CP needs more tokens to converge.

## NB

`tps` and `mfu` are **not affected** — they divide by `non_data_parallel_size` (which includes `cp`), correctly canceling the inflation. Only the cumulative `n_tokens_seen` counter lacks this normalization.

### Versions

- torchtitan: `main` branch (commit `9f055b4`)
- PyTorch: 2.11.0+cu128
- Reproduced on latest torchtitan main, the code path is unchanged

### Config (cp=2 example)

```toml
[training]
local_batch_size = 1
global_batch_size = 32
seq_len = 98304

[parallelism]
data_parallel_replicate_degree = 1
data_parallel_shard_degree = 4
tensor_parallel_degree = 1
pipeline_parallel_degree = 1
context_parallel_degree = 2
```

TryingtobeingNikhil:
### Problem
The training metric `n_tokens_seen` counts the tokens across the entire batch per Context Parallel (CP) rank. When `dist_sum` aggregates this metric over the `loss_mesh` (which includes the CP dimension), the total token count gets inflated by `cp_degree`.

### Root Cause
In `trainer.py`, `ntokens_batch` is measured using the raw `labels.numel()` before any sequence slicing. Both CP ranks load the same dataloader batch. The metrics are then blindly aggregated over `loss_mesh` regardless of CP. 

### Fix
Divide `ntokens_batch` by `self.parallel_dims.cp` before logging and accumulating. This cleanly counters the `dist_sum` multiplication, matching the existing paradigm exactly as seen in throughput calculations using `non_data_parallel_size`.

### How I Verified It
I wrote a new dedicated unit test (`test_tokens_seen.py`) using `unittest.mock` to simulate `Trainer` behavior across different `cp_degree` environments dynamically. The test formally proves the division prevents inflation post-dist_sum accumulation.

### Related Issues/PRs
Fixes #2983
Note: I saw that PR #2990 was also opened for this issue. I wanted to contribute a formal unit test that explicitly maps out the mock-distributed math for future-proofing, so I opened this PR. Happy to consolidate or defer to the maintainers' preference!

### DIFF
--- a/scripts/checkpoint_conversion/numerical_tests_example.py
+++ b/scripts/checkpoint_conversion/numerical_tests_example.py
@@ -12,7 +12,7 @@ from torchtitan.components.checkpoint import ModelWrapper
 from torchtitan.config import ConfigManager
 from torchtitan.tools.logging import logger
 
-from transformers import AutoModelForCausalLM
+from transformers import AutoModelForCausalLM  # pyrefly: ignore
 
 device_type = "cuda" if torch.cuda.is_available() else "cpu"
 
--- a/scripts/checkpoint_conversion/numerical_tests_qwen3_vl.py
+++ b/scripts/checkpoint_conversion/numerical_tests_qwen3_vl.py
@@ -35,7 +35,7 @@ torch._dynamo.config.disable = True
 
 from torchtitan.components.checkpoint import ModelWrapper
 from torchtitan.models.qwen3_vl import model_registry
-from transformers import AutoProcessor
+from transformers import AutoProcessor  # pyrefly: ignore
 
 
 # ============================================================
@@ -207,7 +207,7 @@ def print_pixel_comparisons(comparisons):
 @torch.no_grad()
 def run_hf(model_path, hf_inputs, device):
     """Run HF model, return last-token logits per sample."""
-    from transformers import AutoModelForImageTextToText
+    from transformers import AutoModelForImageTextToText  # pyrefly: ignore
 
     print(f"Loading HuggingFace model on {device} ...")
     model = AutoModelForImageTextToText.from_pretrained(
--- a/tests/unit_tests/test_tokens_seen.py
+++ b/tests/unit_tests/test_tokens_seen.py
@@ -1,0 +1,61 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+from unittest.mock import MagicMock, patch
+
+import torch
+
+from torchtitan.trainer import Trainer
+
+
+class TestTokensSeen(unittest.TestCase):
+    @patch("torchtitan.trainer.prepare_context_parallel_input")
+    def test_cp_token_normalization(self, mock_prepare_cp: MagicMock) -> None:
+        """
+        Verify that ntokens_seen is accumulated *after* context parallel sharding
+        inside post_dataloading_process to prevent metric inflation across the loss mesh.
+        """
+        seq_len = 100
+        global_batch_size = 8
+        original_labels = torch.zeros(global_batch_size, seq_len)
+        actual_tokens_in_batch = original_labels.numel()  # 800
+
+        for cp_degree in [1, 2, 4]:
+            # 1. Mock the trainer and its dependencies
+            mock_trainer = MagicMock(spec=Trainer)
+            mock_trainer.ntokens_seen = 0
+
+            # 2. Mock ParallelDims with our cp_degree
+            mock_trainer.parallel_dims = MagicMock()
+            mock_trainer.parallel_dims.cp_enabled = cp_degree > 1
+            mock_trainer.parallel_dims.cp = cp_degree
+            mock_trainer.model_config = MagicMock()
+            mock_trainer.model_parts = []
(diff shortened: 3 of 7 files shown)


torchtitan/trainer.py
```diff
@@ -539,7 +539,8 @@ def batch_generator(
                 # entire step will not be executed.
                 raise DataloaderExhaustedError() from ex
             input_dict, labels = batch
-            ntokens_batch = labels.numel()
+            cp_degree = self.parallel_dims.cp
+            ntokens_batch = labels.numel() // cp_degree
```
tianyu-l: I prefer the solution in https://github.com/pytorch/torchtitan/pull/2990 and merged that. If you still want to contribute the test, please rebase.

TryingtobeingNikhil:
Makes total sense! I have rebased the PR against main so it now purely contributes the isolated unit test for batch_generator

tianyu-l:
seems test failed

TryingtobeingNikhil:
Apologies for that! The test was failing after the rebase because I was still testing batch_generator instead of the new post_dataloading_process location. I have updated the unit test to properly mock prepare_context_parallel_input. I verified it passes cleanly against the latest main, and I just pushed the fix!

tianyu-l:
CPU test passed, please also fix linting

TryingtobeingNikhil:
Awesome, thanks for the heads up! I just ran pre-commit locally to inject the missing license headers and correct the spacing. Just pushed the lint fix, so everything should be completely green now!

tianyu-l:
linting still failed, could you check if you are using the same version of pyrefly?

TryingtobeingNikhil:
Ah, good catch! I missed the explicit type annotations for Pyrefly in the test signature. I've just pushed a commit adding (self, mock_prepare_cp: MagicMock) -> None:, so Pyrefly should be perfectly happy now!

TryingtobeingNikhil:
Hey Tianyu! I just checked my PR only adds the test_tokens_seen.py file, and I haven't added or modified any # pyre-ignore suppressions myself. Since the CI pyrefly-check hook runs with pass_filenames: false, it looks like it's automatically sweeping the whole repository and found 2 stale error suppressions in a completely unrelated file (possibly due to an upstream type update in main?). Could you take a look to see if main is currently triggering this unused suppression sweep?

tianyu-l:
oh, that could be possible. Could you try to help trigger global pyrefly check and contribute the fixes in this PR as well?

TryingtobeingNikhil:
I've run the global pyrefly check across the repository and found a few missing import errors (primarily transformers) and unexpected kwargs. I've added the appropriate # pyrefly: ignore suppressions in this PR to resolve all 7 of them. pyrefly check is now exiting with 0 errors.

PR #3010, opened 2026-04-17
[Typing] Pyrefly: --remove-unused-ignores during pre-commit

Andrei-Aksionov:
Hey there 👋 

I've noticed that pyrefly can do cleanup of unused ignores with:

```bash
pyrefly check --remove-unused-ignores
```

It can be integrated as a part of `pre-commit` step via adding this flag as an arg.

> [!Important]
> Ensure `pass_filenames: false` is set. 
> Pyrefly is designed to check the entire project state, and running it only on a subset of staged files can lead to inconsistent results where it "thinks" an ignore is unused because it can't see the rest of the project.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,6 +71,7 @@ repos:
       name: Pyrefly (type checking)
       pass_filenames: false
       language: system
+      args: ["--remove-unused-ignores"]
 
 
 - repo: local
@@ -78,9 +79,9 @@ repos:
     # Check hyperlinks and external links
     - id: lychee-link-checker
       name: lychee-link-checker
-      verbose: true
       language: system
       require_serial: true
+      verbose: true
       entry: |
         /bin/bash -c '
         if ! command -v lychee >/dev/null 2>&1; then
@@ -103,6 +104,7 @@ repos:
         - "--accept=200,403,429,503"
         - "--exclude=http://localhost:.*"
         - "--exclude=http://127.0.0.1:.*"
+        - "--exclude=tcp://localhost:.*"
         - "--max-retries=5"
         - "--retry-wait-time=10"
         - "--user-agent=Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"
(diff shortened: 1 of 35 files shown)


.pre-commit-config.yaml
```diff
@@ -71,16 +71,17 @@ repos:
       name: Pyrefly (type checking)
       pass_filenames: false
       language: system
+      args: ["--remove-unused-ignores"]
```
tianyu-l, line 74: Nice. I understand this is useful for developer before submitting PR.

I wonder what happens in CI which doesn't modify code, really. When CI calls this, would it hide errors when it could've been caught before this change?

Andrei-Aksionov: This behavior is standard for pre-commit.

When you commit changes locally, pre-commit runs and returns a non-zero error code if a formatter modifies files. To complete the process, you must stage those new changes and run git commit again.

In CI, the logic is similar: the workflow calls pre-commit, and if a tool like Pyrefly finds and removes unused ignores, the resulting file changes cause the hook to exit with an error, failing the action. 

---

A couple of things that I didn't understand from you question.

> I wonder what happens in CI which doesn't modify code, really. 

Did you mean will Pyrefly find unused ignores in files that aren't changed, since the `lint` workflow runs only on modified code? 
If so, then the answer is yes. 

```yaml
pass_filenames: false
```
Setting `pass_filenames: false` in the [Pyrefly pre-commit hook](https://github.com/facebook/pyrefly-pre-commit) tells pre-commit not to pass specific modified filenames as arguments.
(pre-commit sends to the hook files that are changed and are not in .gitignore.)
Instead, Pyrefly runs against the entire repository by default.

> When CI calls this, would it hide errors when it could've been caught before this change?

If your question was whether Pyrefly could have showed unused ignores before or not, then the answer is yes.
There is even a dedicated [error](https://pyrefly.org/en/docs/error-kinds/#unused-ignore) for this.
However, it likely wasn't displayed because of its low default severity. In version 0.45.1 (the one that this repo uses) there isn't a clear way to override this log level directly in the configuration; newer versions typically offer better control over diagnostic reporting. (TBH I haven't checked the newest version 0.61.1.)

tianyu-l: > When you commit changes locally, pre-commit runs and returns a non-zero error code if a formatter modifies files. To complete the process, you must stage those new changes and run git commit again.
>
> In CI, the logic is similar: the workflow calls pre-commit, and if a tool like Pyrefly finds and removes unused ignores, the resulting file changes cause the hook to exit with an error, failing the action.

This addressed me question. Thanks for the thorough response!

tianyu-l (approved):

PR #3016, opened 2026-04-17
Gate H100.8 CI workflows behind ciflow/h100.8 label

SherlockNoMad:
## Summary

H100.8 runners are scarce and the job queue is very long, blocking PRs. This PR gates H100 CI workflows behind a `ciflow/h100.8` label so they only run on PRs when a maintainer explicitly requests it.

**Changes to 2 workflows:**
- `integration_test_8gpu_h100.yaml` — PR trigger narrowed from `[opened, synchronize, reopened, ready_for_review]` to `[labeled]` only, with a job-level `if` checking for the `ciflow/h100.8` label
- `integration_test_8gpu_graph_trainer_h100.yaml` — same trigger change, replacing the draft-PR check with the `ciflow/h100.8` label check

**Trigger behavior after this PR:**

| Event | `h100.yaml` | `graph_trainer_h100.yaml` |
|---|---|---|
| Push to `main` (merge) | Runs (with `paths-ignore: experiments/`) | Runs (with `paths: graph_trainer/`) |
| Cron schedule | Runs (every 6h) | Runs (every 12h) |
| PR labeled `ciflow/h100.8` | Runs if PR touches core (non-experiments) files | Runs if PR touches `graph_trainer/` files |
| PR opened / pushed / reopened | **Skipped** | **Skipped** |

**Note:** The `ciflow/h100.8` label needs to be created in the repo settings.

## Test plan
- [ ] Verify that opening a new PR does NOT trigger the two H100 workflows
image

- [ ] Verify that adding the `ciflow/h100.8` label to a PR triggers the appropriate H100 workflow(s)
- [ ] Verify that push-to-main and cron schedules still trigger normally

### DIFF
--- a/.github/pytorch-probot.yml
+++ b/.github/pytorch-probot.yml
@@ -1,3 +1,4 @@
 ciflow_push_tags:
   - ciflow/8gpu
+  - ciflow/h100.8
 labeler_config: labeler.yml
--- a/.github/workflows/integration_test_8gpu_h100.yaml
+++ b/.github/workflows/integration_test_8gpu_h100.yaml
@@ -8,7 +8,7 @@ on:
     paths-ignore:
       - 'torchtitan/experiments/**'
   pull_request:
-    types: [labeled]
+    types: [labeled, synchronize]
     branches: [ main ]
     paths-ignore:
       - 'torchtitan/experiments/**'
--- a/.github/workflows/integration_test_8gpu_h100.yaml
+++ b/.github/workflows/integration_test_8gpu_h100.yaml
@@ -8,7 +8,7 @@ on:
     paths-ignore:
       - 'torchtitan/experiments/**'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [labeled]
     branches: [ main ]
     paths-ignore:
       - 'torchtitan/experiments/**'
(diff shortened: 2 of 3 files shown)


.github/pytorch-probot.yml
```diff
@@ -1,3 +1,4 @@
 ciflow_push_tags:
   - ciflow/8gpu
+  - ciflow/h100.8
```
SherlockNoMad, line 3: I am not sure what this do....

.github/workflows/integration_test_8gpu_h100.yaml
```diff
@@ -8,7 +8,7 @@ on:
     paths-ignore:
       - 'torchtitan/experiments/**'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [labeled]
```
malfet: I would propose to add `on: push: tags: ciflow/h100.8/*` to a dispatch tigger rather than the complex condition down below

fegin:
How much do we actually save if we still have when PR merged into main ?

fegin (approved):
Thanks for the improvement. Can we verify if this PR is not related to the failing CI? I don't think it is but just want confirm. Thanks!

malfet (approved):

.github/workflows/integration_test_8gpu_h100.yaml
SherlockNoMad:
```
Claude: Tag-based triggering and path filters are incompatible — GitHub ignores
  paths/paths-ignore for tag pushes. To get both label gating and path filtering,
  we need to go back to the pull_request approach.
  ```
  I don't know how true is this, but what I have now seems to work, so I am shipping this.

SherlockNoMad:
> Thanks for the improvement. Can we verify if this PR is not related to the failing CI? I don't think it is but just want confirm. Thanks!

verified, not related.

PR #2959, opened 2026-04-14
[CI] Fix torchvision::nms error in RL integration tests

weifengpy:
fix https://github.com/pytorch/torchtitan/actions/runs/24377404846/job/71193671209?pr=2741

## Summary
- vllm pulls in torchvision transitively via transformers, but the PyPI torchvision doesn't match the nightly torch ABI, causing `operator torchvision::nms does not exist` at import time.
- Install torchvision explicitly from the nightly channel alongside torch in both RL CI workflows (A10G and H100).

## Test plan
- [ ] RL integration tests pass on A10G (`integration_test_4gpu_rl.yaml`)
- [ ] RL integration tests pass on H100 (`integration_test_4gpu_rl_h100.yaml`)

```
+ python -m torchtitan.experiments.rl.tests.integration_tests /home/ec2-user/actions-runner/_work/_temp/artifacts-to-be-uploaded --ngpu 4 --hf_assets_path /var/lib/ci-user/.cache/huggingface/hub/models--Qwen--Qwen3-0.6B/snapshots/c1899de289a04d12100db370d81485cdf75e47ca
Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/pytorch/torchtitan/torchtitan/experiments/rl/__init__.py", line 15, in <module>
    from torchtitan.experiments.rl.models.vllm_wrapper import TorchTitanVLLMModelWrapper
  File "/pytorch/torchtitan/torchtitan/experiments/rl/models/vllm_wrapper.py", line 27, in <module>
    from torchtitan.experiments.rl.models.attention import VLLMAttentionWrapper
  File "/pytorch/torchtitan/torchtitan/experiments/rl/models/attention.py", line 20, in <module>
    from vllm.model_executor.layers.attention import Attention
  File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/vllm/model_executor/layers/attention/__init__.py", line 4, in <module>
    from vllm.model_executor.layers.attention.attention import Attention
  File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/vllm/model_executor/layers/attention/attention.py", line 10, in <module>
    from vllm.config import CacheConfig, get_current_vllm_config
  File "/opt/conda/envs/py_3.12/lib/python3.12/site-packages/vllm/config/__init__.py", line 19, in <module>
    from vllm.config.model import (
[...]

PR #3022, opened 2026-04-18
[ci fix] torchvision channel

yushangdi:
Similar to the fix in https://github.com/pytorch/torchtitan/pull/2959

Fix ` RuntimeError: operator torchvision::nms does not exist`. 
https://github.com/pytorch/torchtitan/actions/runs/24593052832/job/71917713051?pr=3020

The integration_test_8gpu_transformers_modeling_backend.yaml installs only torch (no torchvision), so the old torchvision 0.26.0 stays and is incompatible with the nightly torch.


weifengpy (approved):

tianyu-l:
fixed in https://github.com/pytorch/torchtitan/pull/3002

PR #3023, opened 2026-04-18
fix(hf_datasets): shuffle HuggingFaceTextDataset on re-loop and replay on resume

### LINKED ISSUE #2733: HuggingFaceTextDataset does not shuffle data on re-loop
When `HuggingFaceTextDataset` re-loops (infinite=True), it replays the data in the exact same order every epoch. For `Dataset` (map-style) instances, there is no shuffle between epochs — the model sees identical batches each time through.

The re-loop path for iterable datasets calls `set_epoch`, but the map-style `Dataset` path is a no-op:

https://github.com/pytorch/torchtitan/blob/d84e83dc4ef4eb7b7e7b7e0b8d7c8b3f3f3f3f3f/torchtitan/hf_datasets/text_datasets.py#L148-L165

A fix would be to shuffle with a deterministic seed per epoch, e.g. `self._data.shuffle(seed=42 + epoch)`, and replay shuffles on checkpoint resume.

CrepuscularIRIS:
Fixes #2733. @joecummings — tagging as requested.

## Problem

When `HuggingFaceTextDataset` re-looped (`infinite=True`) against a map-style `Dataset` source, it replayed the data in the exact same order every epoch. Only the iterable-dataset path called `set_epoch`; the `isinstance(self._data, Dataset)` branch was a no-op. Training runs longer than one epoch of a finite dataset therefore saw identical batches on every pass — the exact behavior you described in the issue.

## Approach (matches #2733 guidance and the existing `ChatDataset` pattern in the same file)

- Store an unshuffled `self._original_data` reference in `__init__`.
- On re-loop, when `_data` is a map-style `Dataset`, set
  `self._data = self._original_data.shuffle(seed=42 + self._epoch)` so each epoch gets a deterministic but distinct permutation. **Epoch 0 stays unshuffled**, preserving bit-identical behavior for single-epoch training runs.
- `load_state_dict` now restores `self._epoch` (previously dropped) and replays the same shuffle when `epoch > 0`, so checkpoint resume matches the order the checkpoint was taken at.
- **Backward compatible**: missing `epoch` key in older checkpoints defaults to 0, which maps to the unshuffled path — no breakage for in-flight runs.

The seed constant (`42`) mirrors what `ChatDataset` already uses — I left it hardcoded for parity. Happy to route it through config if you'd prefer that shape.

## Test

Added `test_map_style_shuffle_on_reloop` to `tests/unit_tests/test_dataset_checkpointing.py`. It covers four things without having to drain a full epoch of c4_test (which would be slow for a unit test):

1. After an epoch boundary, `_data` is a shuffled copy of `_original_data` (identity check).
2. `state_dict()` carries `epoch` so resume knows the seed.
3. `load_state_dict()` replays the same `shuffle(seed=42+epoch)` — verified by comparing the first 5 sample texts between the two instances.
4. Legacy checkpoints missing `epoch` still load cleanly and default to the unshuffled path.

Verified locally:

```
$ python -m pytest tests/unit_tests/test_dataset_checkpointing.py -v
tests/unit_tests/test_dataset_checkpointing.py::TestDatasetCheckpointing::test_c4_resumption PASSED
tests/unit_tests/test_dataset_checkpointing.py::TestDatasetCheckpointing::test_map_style_shuffle_on_reloop PASSED
============================== 2 passed in 10.01s ==============================
```

The existing `test_c4_resumption` still passes, confirming the bit-identical guarantee for non-reloop resume paths.


tests/unit_tests/test_dataset_checkpointing.py
```diff
@@ -72,3 +72,72 @@ def _build_dataloader(self, dataset_name, batch_size, seq_len, world_size, rank)
...
+        Validates three things end-to-end without having to drain a full epoch
+        of c4_test (which would be slow for a unit test):
+          1. After an epoch boundary, ``_data`` is a *shuffled* copy of
+             ``_original_data`` — not the same object.
+          2. ``state_dict()`` carries ``epoch`` so resume knows the shuffle
+             seed.
+          3. ``load_state_dict()`` replays the same ``shuffle(seed=42+epoch)``
+             so a resumed run observes the identical sample order.
+        """
+        from torchtitan.hf_datasets.text_datasets import HuggingFaceTextDataset
```
joecummings, line 90: Nit: This can be imported at the top-level. Should simplify things across this file.

joecummings (approved):
Thanks for the fix, looks good. Please address the linting and comments and we can get this merged

WRT hardcoded seed, we can parameterize that in a follow up in users ask for it.

tianyu-l:
Is this change supposed to change loss curve?

tests/unit_tests/test_dataset_checkpointing.py
CrepuscularIRIS: Addressed in [`8f693e59`](https://github.com/CrepuscularIRIS/torchtitan/commit/8f693e59) — `HuggingFaceTextDataset` is now imported alongside the other names in the existing top-level import line, so the `_build_ds` helper and all four assertions reference it without any local-import boilerplate.

CrepuscularIRIS:
@tianyu-l yes, the loss curve will look different for multi-epoch runs — intentionally so.

**Before this PR:** map-style datasets (`Dataset`) replayed the identical sample order every epoch because `_data` was never reshuffled on re-loop. The model saw the same sequence of batches on epoch 1, epoch 2, etc., which can cause it to overfit to presentation order and produces a loss curve that repeats the same pattern each epoch.

**After this PR:** at each epoch boundary `_data` is replaced with `_original_data.shuffle(seed=42 + epoch)`, so every pass through the data uses a different order. The per-step loss values will differ from the old code for any run that crosses an epoch boundary, but the overall learning trend should be the same or better.

**Single-epoch runs are bit-identical:** epoch 0 is never shuffled, so if your run finishes within one pass through the dataset the loss curve is unchanged.

The iterable-style path (`set_epoch`) is also unchanged; this only affects `isinstance(self._data, Dataset)` (map-style).

---

@joecummings the import nit is fixed in [`8f693e59`](https://github.com/CrepuscularIRIS/torchtitan/commit/8f693e59). Should be ready to merge whenever CI clears.

tianyu-l (commented):
Since this will change loss for multi-epoch, the current failure makes sense. Please update the loss in https://github.com/pytorch/torchtitan/tree/main/tests/assets/losses (for both cuda and rocm).
cc @akashveramd rocm test hasn't run.

Also please fix linting.

akashveramd:
> Since this will change loss for multi-epoch, the current failure makes sense. Please update the loss in https://github.com/pytorch/torchtitan/tree/main/tests/assets/losses (for both cuda and rocm). cc @akashveramd rocm test hasn't run.
> 
> Also please fix linting.

@tianyu-l: Seems the ROCm runner hasn't picked up jobs to run from queue.

tianyu-l:
@akashveramd 
This PR will likely violate loss_rocm_mi350 numerics. Ideally the test could run and we can update the txt accordingly, like the cuda one. But since it's now running despite a day's wait, do you think we can merge this PR and update the file later, whenever it runs?

CrepuscularIRIS:
Linting and CUDA losses are both addressed in [`e9a3fd46`](https://github.com/CrepuscularIRIS/torchtitan/commit/e9a3fd46):

- **Lint**: `ufmt` (`black` + `usort`) was applied to the import block in the test file and a multi-line `elif` in `text_datasets.py`.
- **CUDA losses** (`llama3_cuda.txt`): updated with values from CI run #24769318999. Steps 1–23 (epoch 0) are bit-identical to the old file; divergence starts at step 24 where the dataset re-loops and is reshuffled — which is the intended behavior.

**ROCm losses** (`llama3_rocm.txt`): I can't update this without the runner producing new numbers. I'm happy to push an update to `llama3_rocm.txt` the moment the MI350 job completes. If the maintainers prefer to merge now and land the ROCm fix in a follow-up commit once the runner catches up, that works for me too — just let me know which route you'd prefer.

tianyu-l:
@CrepuscularIRIS
oh I just realized that this PR didn't rebase onto https://github.com/pytorch/torchtitan/pull/2740 thus still using the old rocm hardware in CI.

Please rebase and submit again so that I can trigger run against latest hardware.

CrepuscularIRIS:
Got it, will rebase onto #2740 and push shortly.

CrepuscularIRIS:
Linting and CUDA reference losses are updated in e9a3fd46. For the ROCm file — happy to push the updated numbers once the runner produces them, or to leave that for a follow-up if merging first is easier.

tianyu-l:
> Linting and CUDA reference losses are updated in https://github.com/pytorch/torchtitan/commit/e9a3fd46768c38ba832e51bdfea7b0ca3c306a84. For the ROCm file — happy to push the updated numbers once the runner produces them, or to leave that for a follow-up if merging first is easier.

OK I'm merging this now. Please follow up with a rocm loss PR later.

tianyu-l (approved):

CrepuscularIRIS:
Thanks for merging! Will open a follow-up PR with the updated ROCm loss values once the runner produces them.

tianyu-l:
queued here https://github.com/pytorch/torchtitan/actions/runs/24823421351/job/72653343303

PR #3030, opened 2026-04-20
Change logging level for model config

wwwjn:
The model config print is too long in CI, which is hard to scroll and read the real error message.
Eg: https://github.com/pytorch/torchtitan/actions/runs/24666450657/job/72125280092#step:17:73090


tianyu-l:
would https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L393 cover this? If so we don't need to explicit log it again any more.

tianyu-l (approved):

Issue #2981, opened 2026-04-15
tyro CLI subcommand collision between Float8LinearConverter.Config and Float8GroupedMMConverter.Config breaks DeepSeek-V3-671B training


janbernloehr:
### Bug description

The `deepseek_v3_671b` config in `torchtitan/models/deepseek_v3/config_registry.py` specifies two model converters:

```python
model_converters=ModelConvertersContainer.Config(
    converters=[
        Float8LinearConverter.Config(filter_fqns=["output", "router.gate"]),
        Float8GroupedMMConverter.Config(fqns=["experts"]),
    ],
)
```

Both `Float8LinearConverter` and `Float8GroupedMMConverter` (in `torchtitan/components/quantization/float8.py`) define inner dataclasses named `Config`. Since they share the same `__name__` and `__module__`, `tyro` generates identical subcommand names for both, causing a collision at CLI parsing:

```
UserWarning: Duplicate subcommand name detected: 'model-converters.converters.0:config'
is already used for torchtitan.components.quantization.float8.Config but will be
overwritten by torchtitan.components.quantization.float8.Config.

╭─ Invalid input to tyro.cli() ──────────────────────────────────────────────╮
│ The default value of the model-converters.converters.0 field could not be  │
│ matched to any of its subcommands.                                          │
│ model-converters.converters.0:config was not a match because:               │
│ • Default value Config(enable_fsdp_float8_all_gather=False, ...             │
│   filter_fqns=['output', 'router.gate'], emulate=False)                     │
│   with type Config does not match type                                      │
│   <class '...Float8GroupedMMConverter.Config'>                              │
╰────────────────────────────────────────────────────────────────────────────╯
```

The process exits with code 2 before any training logic runs. This affects only the DeepSeek-V3-671B config since it is the only one that uses both Float8 converter types simultaneously.

**Suggested fix**: Rename the inner `Config` dataclasses to unique names (e.g., `Float8LinearConfig` and `Float8GroupedMMConfig`), or use `tyro.conf.subcommand()` annotations to provide explicit distinct subcommand names.

### Versions

Reproducible on torchtitan `main` at commit `74485ea` (April 14, 2026). Confirmed across multiple PyTorch nightly versions (torch `2.12.0a0+0291f960b6` and later). Uses the new Python dataclass registry system (`--module deepseek_v3 --config deepseek_v3_671b`).

tianyu-l:
[...]

PR #3032, opened 2026-04-20
quantize on config instead of on model

acisseJZhong:
Fixes https://github.com/pytorch/torchtitan/issues/2981

## Summary

Apply quantization at model config construction time by swapping config types, instead of post-build model mutation.

Previously, the trainer built the model, then walked the module tree to swap `nn.Linear` with `Float8Linear`, requiring save/restore of `_param_init` and the `Module` protocol. Now, quantization is a config pass in `model_registry()`:

```python
model_registry("405B", quantization=[
    Float8LinearQuant(filter_fqns=["output"]),
    Float8MoEQuant(),
])
```

Each `QuantizationConfig` subclass calls a standalone function that walks the config tree:
- `convert_to_float8()` swaps `Linear.Config` → `Float8LinearConfig` (whose `build()` produces `Float8Linear` directly)
- `convert_moe_to_float8()` / `convert_moe_to_mxfp8()` walk `GroupedExperts.Config` by type and set `_convert_fn`

**Removed:**
- `ModelConverter`, `ModelConvertersContainer`, `Float8LinearConverter`, `Float8GroupedMMConverter`, `MXFP8Converter`, `QuantizationConverter` classes and `model_converter.py`, `float8.py`, `mx.py` files
- `enable_fsdp_float8_all_gather`, `precompute_float8_dynamic_scale_for_fsdp`, `enable_float8_tensorwise_tp`
- `model_converters` from `Trainer.Config` and all parallelize functions
- `parallel_dims` dependency from converters

**Added:**
- `QuantizationConfig` hierarchy (`Float8LinearQuant`, `Float8MoEQuant`, `MXFP8MoEQuant`, `MXFP8Quant`)
- `Float8LinearConfig(Linear.Config)` — builds `Float8Linear` directly
- `Configurable.Config.walk()` yields `(fqn, config, parent, attr)` for config tree traversal and replacement
- `has_quantization()` — type-based check
- Token dispatcher swap and `pad_multiple` handled at config time

## Test plan
- [ ] Full unit test suite passes (219 passed)
- [ ] Verify float8 training produces identical loss:
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama3' --baseline-config='llama3_debugmodel_float8' --baseline-ngpus=8 --test-module='llama3' --test-config='llama3_debugmodel_float8' --test-ngpus=8 --assert-equal --no-seed-checkpoint
```


torchtitan/components/quantization/mx.py
```diff
@@ -76,11 +76,13 @@ def __init__(
...
 
     def convert_config(self, model_config) -> None:
-        """Convert MoE layers in the model config to use MXFP8 grouped GEMMs.
+        """Convert MoE layers in the model config to use MXFP8 scaled grouped GEMMs.
 
         Walks the model config tree and sets ``_convert_fn`` on modules
-        matching the target FQNs so that ``build()`` applies dynamic
-        MXFP8 quantization on grouped GEMM operations.
+        matching the target FQNs so that ``build()`` replaces instances of
+        nn.Parameter with ScaledGroupedMMTensor. This will use low precision grouped
```
danielvegamyhre: >ScaledGroupedMMTensor

it's called `MXFP8TrainingWeightWrapperTensor` now

torchtitan/protocols/model.py
```diff
@@ -59,6 +59,10 @@ class Config(Module.Config):
         Subclasses define model-specific hyperparameters.
         """
 
+        quantization: object | None = None
```
tianyu-l, line 62: This doesn't sound a fundamental property to be put in `Model` class. We can easily build a util function to check.

torchtitan/protocols/model_converter.py
tianyu-l, line 1: need to remove https://github.com/pytorch/torchtitan/blob/main/docs/extension.md#modelconverter as well

torchtitan/components/quantization/__init__.py
```diff
@@ -39,19 +35,252 @@ class Config(Configurable.Config):
...
+def convert_to_float8(
+    model_config,
+    *,
+    recipe_name: str = "rowwise",
+    filter_fqns: list[str] | None = None,
+    emulate: bool = False,
+) -> None:
+    """Convert Linear.Config instances in the model config to produce Float8Linear.
+
+    Walks the model config tree and sets ``_convert_fn`` on matching
```
tianyu-l, line 110: I think there's still too much indirection. IMO we should build a pass / walk on the model config to convert configs to quantization config, rather than convert modules. The filter logic can still apply if we convert configs.

torchtitan/components/quantization/__init__.py
```diff
@@ -39,19 +35,252 @@ class Config(Configurable.Config):
...
+    from torchao.quantization.quant_api import quantize_
+    from torchtitan.protocols.module import Module
+
+    def convert_fn(mod: nn.Module) -> nn.Module:
+        recipe = MXFP8TrainingRecipe(recipe_name)
+        mxfp8_op_config = MXFP8TrainingOpConfig.from_recipe(recipe)
+        mxfp8_op_config.pad_token_groups_for_grouped_mm = (
+            pad_token_groups_for_grouped_mm
+        )
+        quantize_(mod, config=mxfp8_op_config)
```
tianyu-l, line 276: @danielvegamyhre how to create a quantized module out of config, rather than converting a pre-built regular module?

torchtitan/components/quantization/__init__.py
```diff
@@ -39,19 +35,252 @@ class Config(Configurable.Config):
     "mxfp8": 32,
 }
 
+AUTO_FILTER_SMALL_KN_FLAG = "auto_filter_small_kn"
+
+
+@dataclass
+class QuantizationConfig:
```
tianyu-l, line 42: We should avoid putting everything into a giant config class.

You could let float8 / mxfp8 share the same base config class, which has an apply / map method that applies what's in `convert_to_float8` / `convert_moe_to_mxfp8`.

torchtitan/components/quantization/__init__.py
```diff
@@ -39,19 +35,252 @@ class Config(Configurable.Config):
...
+        if filter_fn(linear_config, fqn):
+            linear_config._convert_fn = convert_fn
+
+    logger.info("Swapped to Float8Linear layers")
+
+
+def convert_moe_to_float8(
+    model_config,
+    *,
+    fqns: list[str],
```
tianyu-l, line 193: I think using fqn was because torchao doesn't have access to torchtitan classes. Here we can use `isinstance(..., MoE.Config)`.

torchtitan/components/quantization/float8.py
tianyu-l, line 1: need to modify docs https://github.com/pytorch/torchtitan/blob/main/docs/float8.md and https://github.com/pytorch/torchtitan/blob/main/docs/mxfp8.md as well -- and let's put docs locally to this folder

torchtitan/components/quantization/__init__.py
acisseJZhong: > we should build a pass / walk on the model config to convert configs to quantization config, rather than convert modules

I think this is exactly what this function is doing: walking the config tree and convert configs. We are annotating configs with `_convert_fn` so `build()` produces the right module. The `_convert_fn` is a workaround for torchao's API requiring [`from_float`](https://github.com/pytorch/ao/blob/990d15548447bc7828c9afdfb36c16b101943518/torchao/float8/float8_linear_utils.py#L114). Otherwise we could replace `Linear.Config` with `Float8Linear.Config` at config time.

acisseJZhong: introduced `Float8LinearConfig` to make this more readable.

torchtitan/components/quantization/float8.py
acisseJZhong: sounds good, will update the doc after all code has been finalized.

torchtitan/components/quantization/utils.py
```diff
@@ -6,20 +6,30 @@
 
 import torch.nn as nn
 
+from torchtitan.models.common.linear import Linear
 
-def module_filter_fn(mod: nn.Module, fqn: str, filter_fqns: list[str]) -> bool:
+
+def module_filter_fn(
+    mod: nn.Module | Linear.Config, fqn: str, filter_fqns: list[str]
```
tianyu-l: do we need to keep the `nn.Module` option?

torchtitan/models/common/config_utils.py
```diff
@@ -225,13 +224,7 @@ def make_token_dispatcher_config(
...
             non_blocking_capacity_factor=non_blocking_capacity_factor,
         )
-    elif comm_backend == "torchao":
-        return TorchAOTokenDispatcher.Config(
-            num_experts=num_experts,
-            top_k=top_k,
-            score_before_experts=score_before_experts,
-        )
-    elif comm_backend == "standard":
+    elif comm_backend in ("standard", "torchao"):
```
tianyu-l: I don't think we need the `"torchao"` backend here.

torchtitan/trainer.py
```diff
@@ -277,19 +262,15 @@ def __init__(self, config: Config):
         # have this guanrantee, e.g., fully_shard.
         model.verify_module_protocol()
 
-        # Check if any converter uses quantization (FP8, MX, etc.)
-        has_quantization = any(
-            isinstance(cc, QuantizationConverter.Config)
-            for cc in config.model_converters.converters
-        )
+        has_quant = has_quantization(model_config)
```
tianyu-l: make this inline?

torchtitan/protocols/model_spec.py
```diff
@@ -48,4 +48,4 @@ class ModelSpec:
 
 @dataclass
 class FaultTolerantModelSpec(ModelSpec):
-    fragment_fn: Callable | None
+    fragment_fn: Callable | None = None
```
tianyu-l: Not related to this PR: this should be put into experiments/ft folder

torchtitan/protocols/model.py
```diff
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from abc import abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
```
tianyu-l: accidental change?

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+@dataclass(kw_only=True, slots=True)
+class Float8LinearConfig(Linear.Config):
+    """Drop-in replacement for Linear.Config that builds Float8Linear directly."""
+
+    _torchao_config: object = None
+
+    def build(self, **kwargs):
+        from torchao.float8.float8_linear import Float8Linear
+
+        instance = Float8Linear(
```
tianyu-l: why don't we create `Float8Linear` in torchtitan which inherits torchao's `Float8Linear` and define `Float8Linear.Config(Linear.Config)`, so that we can remove `inject_module_protocol`

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+
+    emulate: bool = False
+    """
+    If True, emulation is used instead of hardware accelerated gemm. This is for test purpose only,
+    as the current CI does not have sm_89 capability, required by Float8.
+    Not compatible with torch.compile.
+    """
+
+    def apply(self, model_config) -> None:
+        convert_to_float8(
```
tianyu-l: why two layers, you can just convert inside `apply`

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+
+    if not model_compile_enabled:
+        logger.warning(
+            "torch.compile enablement is required for highest performance "
+            "of MXFP8 dynamic quantization."
+        )
+    convert_fn = _make_mxfp8_convert_fn(recipe_name, pad_token_groups_for_grouped_mm)
+
+    # MXFP8: scaling block size is (1 x 32), so contracting dim must be divisible by 32.
+    pad_multiple = 32
```
tianyu-l: this should come from https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/__init__.py#L37

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+
+    model_compile_enabled: bool = False
+    """Whether torch.compile is enabled for the model."""
+
+    def apply(self, model_config) -> None:
+        raise NotImplementedError
+
+
+@dataclass
+class Float8LinearQuant(QuantizationConfig):
```
tianyu-l: I see that you would like to keep the configurability of the quantization.

Instead of letting a config own the apply function, you can do this alternative to make it consistent with the rest config system:
- Make a `QuantizationConverter` which has `QuantizationConverter.Config`
- `QuantizationConverter` has a method `convert(Model.Config)`

It sounds what we had before, but actually is different, because this config only works on top of config, not actual model.

Sorry if I denied this idea from you before.

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+        raise ImportError(
+            "torchao installation does not have MoE training support. Please install torchao nightly build."
+        ) from e
+
+    # FP8: 16 byte alignment / 1 byte per elem = 16 elements.
+    pad_multiple = 16
+
+    def convert_fn(mod: nn.Module) -> nn.Module:
+        config = Float8TrainingOpConfig()
+        quantize_(mod, config=config)
```
tianyu-l: I'm not convinced that we have to call `quantize_` on a pre-existing module (and the necessity of this `convert_fn`). Please work with @danielvegamyhre to convert config only.

torchtitan/protocols/module.py
```diff
@@ -37,13 +37,23 @@ class Module(nn.Module, Configurable):
     class Config(Configurable.Config):
         param_init: dict | None = None
 
+        _convert_fn: Callable[[nn.Module], nn.Module] | None = field(
```
tianyu-l: I prefer we don't introduce this.

torchtitan/models/llama3/config_registry.py
```diff
@@ -91,25 +90,23 @@ def llama3_debugmodel_opt_in_bwd() -> Trainer.Config:
...
-        converters=[
-            Float8LinearConverter.Config(
-                enable_fsdp_float8_all_gather=True,
-                precompute_float8_dynamic_scale_for_fsdp=True,
+    config.model_spec = model_registry(
+        "debugmodel",
+        quantization=[
+            Float8LinearQuant(
                 emulate=True,
+                model_compile_enabled=(
```
tianyu-l, line 123: This looks not nice .. if it's for validation / error msg only, we can put it in post_init of model config.

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+        convert_to_mxfp8(
+            model_config,
+            fqns=self.fqns,
+            recipe_name=self.recipe_name,
+            pad_token_groups_for_grouped_mm=self.pad_token_groups,
+            model_compile_enabled=self.model_compile_enabled,
+        )
+
+
+def has_quantization(model_config) -> bool:
```
tianyu-l: This can be put into utils.py?

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+    ), "MXFP8 is only supported on SM100 or architectures"
+
+    if not model_compile_enabled:
+        logger.warning(
+            "torch.compile enablement is required for highest performance "
+            "of MXFP8 dynamic quantization."
+        )
+    convert_fn = _make_mxfp8_convert_fn(recipe_name, pad_token_groups_for_grouped_mm)
+
+    for fqn, config, _parent, _attr in model_config.walk(Module.Config):
```
tianyu-l: Not sure: should only targeting Linear.Config, o/w overlapped with `convert_moe_to_mxfp8`?

torchtitan/config/configurable.py
```diff
@@ -63,7 +64,40 @@ def _convert(val):
...
-            return {f.name: _convert(getattr(self, f.name)) for f in fields(self)}
+            return {
+                f.name: _convert(getattr(self, f.name))
+                for f in fields(self)
+                if not f.name.startswith("_")
+            }
+
+        def walk(
+            self, config_cls: type, *, _prefix: str = ""
+        ) -> Iterator[tuple[str, "Configurable.Config", object, str]]:
```
tianyu-l, line 75: can we just do
```suggestion
        ) -> Iterator[tuple[str, Config, object, str]]:
```

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+    if not has_cuda_capability(8, 9):
+        raise ValueError("Float8 MoE training only supported on SM89 or later.")
+
+    if not model_compile_enabled:
+        logger.warning(
+            "Compile is required for high performance float8 MoE training; "
+            "enable it with --compile.enable"
+        )
+
+    from torchao.quantization.quant_api import quantize_
```
joecummings: Nit: In general, do we have to be defensive with stable APIs? Applies to other cases as well

torchtitan/components/quantization/__init__.py
```diff
@@ -14,44 +14,430 @@
...
+def convert_to_mxfp8(
+    model_config,
+    *,
+    fqns: list[str] | None = None,
+    recipe_name: str = "mxfp8_rceil",
+    pad_token_groups_for_grouped_mm: bool = True,
+    model_compile_enabled: bool = False,
+) -> None:
+    """Convert modules matching FQNs to use MXFP8 quantization (e.g. Flux blocks)."""
+    if find_spec("torchao") is None:
```
joecummings: In order to do pretty much any of these conversions, don't we need torchao?

torchtitan/protocols/model_spec.py
acisseJZhong: since it's just moving these three lines, I made the change with this PR.

torchtitan/components/quantization/__init__.py
acisseJZhong: l deleted the mapping as we can just assign different values for different quantization(in total 2).

torchtitan/protocols/module.py
acisseJZhong: @danielvegamyhre  do you know how should we remove this? similar comment https://github.com/pytorch/torchtitan/pull/3032#discussion_r3121771349

torchtitan/models/llama3/config_registry.py
acisseJZhong: it's also used here https://github.com/pytorch/torchtitan/blob/e0a926f4f0a81d243aecc0a5a3507aa6a5940524/torchtitan/components/quantization/float8.py#L70, all the rest usage could be moved to post init.

torchtitan/components/quantization/__init__.py
danielvegamyhre: By the way, we would also like to support multiple of 128, not just multiple of 32, enabling more efficient quantization kernels and overall net performance benefit. If you can add this in your PR that would be great, or I can do it in a follow up later.

torchtitan/components/quantization/__init__.py
acisseJZhong: not added by this PR.

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,484 @@
...
+                "Compile is required for high performance float8 MoE training; "
+                "enable it with --compile.enable"
+            )
+
+        # FP8: 16 byte alignment / 1 byte per elem = 16 elements.
+        pad_multiple = 16
+
+        for _fqn, config, parent, attr in model_config.walk(GroupedExperts.Config):
+            dispatcher_handles_padding = swap_token_dispatcher(config, pad_multiple)
+            if not dispatcher_handles_padding:
```
tianyu-l: let's move this raise inside `swap_token_dispatcher` (and we don't need to return bool)

torchtitan/components/quantization/utils.py
```diff
@@ -4,24 +4,79 @@
...
+
+    has_quant_linear = any(
+        isinstance(lc, (Float8Linear.Config, MXFP8Linear.Config))
+        for _fqn, lc, _parent, _attr in model_config.walk(Linear.Config)
+    )
+    has_quant_moe = any(
+        getattr(type(config), "_is_quantized", False)
+        for _fqn, config, _parent, _attr in model_config.walk(GroupedExperts.Config)
+    )
+    return has_quant_linear or has_quant_moe
```
tianyu-l, line 86: nit:you don't need to walk twice?

torchtitan/config/configurable.py
```diff
@@ -63,7 +66,40 @@ def _convert(val):
...
                     return repr(val)
 
-            return {f.name: _convert(getattr(self, f.name)) for f in fields(self)}
+            return {
+                f.name: _convert(getattr(self, f.name))
+                for f in fields(self)
+                if not f.name.startswith("_")
+            }
+
+        def walk(
```
tianyu-l: maybe `traverse` as a more formal name than `walk`

torchtitan/experiments/ft/llama3/__init__.py
```diff
@@ -12,7 +15,12 @@
     Llama3StateDictAdapter,
     parallelize_llama,
 )
-from torchtitan.protocols.model_spec import FaultTolerantModelSpec
+from torchtitan.protocols.model_spec import ModelSpec
+
+
+@dataclass
+class FaultTolerantModelSpec(ModelSpec):
```
tianyu-l: let's put in https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/ft/config/job_config.py instead of here which looks model specific

torchtitan/models/common/config_utils.py
tianyu-l, line 218: how do you think of the idea of separating them into two dispatcher classes, for modularity concern?

default.profraw
tianyu-l, line 1: remove?

torchtitan/protocols/model.py
```diff
@@ -70,6 +70,7 @@ def update_from_config(
         ) -> None:
             pass
 
+
```
tianyu-l: seems accidental

torchtitan/models/llama3/config_registry.py
```diff
@@ -189,15 +186,9 @@ def llama3_405b() -> Trainer.Config:
...
-            converters=[
-                Float8LinearConverter.Config(
-                    enable_fsdp_float8_all_gather=True,
-                    precompute_float8_dynamic_scale_for_fsdp=True,
-                    filter_fqns=["output"],
-                ),
-            ],
+        model_spec=model_registry(
+            "405B",
+            quantization=[Float8LinearConverter.Config(filter_fqns=["output"])],
```
tianyu-l: also need model_compile_enabled?

torchtitan/models/common/linear.py
```diff
@@ -27,9 +27,11 @@ class Config(Module.Config):
         out_features: int
         bias: bool = False
 
-    def __init__(self, config: Config):
-        super().__init__(
-            config.in_features,
-            config.out_features,
-            bias=config.bias,
-        )
+    def __init__(self, config_or_in_features=None, out_features=None, bias=False, *, config=None):
```
tianyu-l: why this change?

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,484 @@
...
 from torchtitan.config import Configurable
+from torchtitan.models.common.linear import Linear
+from torchtitan.models.common.moe import GroupedExperts
+from torchtitan.tools.logging import logger
+from torchtitan.tools.utils import has_cuda_capability
+
+from .utils import module_filter_fn, swap_token_dispatcher
+
+
+class Float8Linear:
```
tianyu-l: inherit torchtitan Linear -- actually, need to discuss

torchtitan/components/quantization/__init__.py
tianyu-l, line 1: can we deprecate https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/module_utils.py

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+            def build(self, **kwargs):
+                instance = Float8Linear(
+                    self.in_features,
+                    self.out_features,
+                    bias=self.bias,
+                    config=self._torchao_config,
+                )
+                if self.param_init is not None:
+                    instance._param_init = self.param_init
+                return instance
```
tianyu-l: instead of implementing `build`, let's just implement it in `__init__`. Also I don't see why you need to manually do anything to `param_init`

torchtitan/components/quantization/utils.py
acisseJZhong: used `_is_quantized` instead.

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+                bias=self.bias,
+                recipe_name=self._recipe_name,
+                pad_token_groups=self._pad_token_groups,
+            )
+            if self.param_init is not None:
+                instance._param_init = self.param_init
+            return instance
+
+    def __init__(
+        self, in_features, out_features, bias=False, *, recipe_name="mxfp8_rceil", pad_token_groups=True
```
tianyu-l: signature should be `Config`

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+            instance = MXFP8Linear(
+                self.in_features,
+                self.out_features,
+                bias=self.bias,
+                recipe_name=self._recipe_name,
+                pad_token_groups=self._pad_token_groups,
+            )
+            if self.param_init is not None:
+                instance._param_init = self.param_init
+            return instance
```
tianyu-l: don't need this `build` function?

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+    class Float8Linear(TorchAOFloat8Linear, Linear):
+        """Diamond inheritance (TorchAOFloat8Linear + Linear) so the built module
+        satisfies the Module protocol (init_states, _param_init).
+        """
+
+        @dataclass(kw_only=True, slots=True)
+        class Config(Linear.Config):
+            """Drop-in replacement for Linear.Config that builds Float8Linear."""
+
+            _is_quantized: ClassVar[bool] = True
```
tianyu-l: I don't think we need this field. We can just let all quantization modules inherit the same base `QuantizedLinear` class.

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+class MXFP8Linear(Linear):
+    """Linear that applies MXFP8 quantization in its constructor."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Linear.Config):
+        """Drop-in replacement for Linear.Config that builds MXFP8Linear."""
+
+        _is_quantized: ClassVar[bool] = True
+        _recipe_name: str = "mxfp8_rceil"
+        _pad_token_groups: bool = True
```
tianyu-l: remove this field

torchtitan/config/configurable.py
acisseJZhong: Renamed it to traverse as it makes more sense in ML context. I think Claude choose walk similar to os.walk.

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+            logger.warning(
+                "Compile is required for high performance float8 MoE training; "
+                "enable it with --compile.enable"
+            )
+
+        from torchao.prototype.moe_training.config import Float8TrainingOpConfig
+        from torchao.quantization.quant_api import quantize_
+
+        # FP8: 16 byte alignment / 1 byte per elem = 16 elements.
+        pad_multiple = 16
```
tianyu-l: should be a frozen class variable?

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+        cache: dict[type, type] = {}
```
tianyu-l: maybe name it
```suggestion
        _converted_config_cache: dict[type, type] = {}
```

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+                    "Float8 MoE quantization requires a token dispatcher that handles "
+                    "padding (TorchAOTokenDispatcher or DeepEP hybridep). "
+                    "Enable expert parallelism or use a compatible comm backend."
+                )
+
+            base_cls = type(config)
+            if base_cls not in cache:
+
+                @dataclass(kw_only=True, slots=True)
+                class QuantizedConfig(base_cls):
```
tianyu-l:
```suggestion
                class Float8GroupedExpertsConfig(base_cls):
```

torchtitan/models/common/config_utils.py
acisseJZhong: yes on my to do list.

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+                    _is_quantized: ClassVar[bool] = True
```
tianyu-l: also don't need this -- we can have a `QuantizedGroupedExpertsConfig` as base class to share with mxfp8

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,432 @@
...
+
+        fqns: list[str] = field(default_factory=list)
+        """
+        *Prototype feature, performance optimization still in progress*
+        Comma-separated list of fully qualified names of MoE modules to apply MXFP8 dynamic quantization
+        on grouped GEMM operations.
+        This is a prototype feature that requires the torchao nightly build.
+        """
+
+        pad_token_groups: bool = True
```
tianyu-l: remove and raise at the place where it's used with True

torchtitan/components/quantization/utils.py
```diff
@@ -4,24 +4,83 @@
...
+
+    quant_types = (MXFP8Linear.Config,)
+    if Float8Linear is not None:
+        quant_types = (Float8Linear.Config, MXFP8Linear.Config)
+
+    has_quant_linear = any(
+        isinstance(lc, quant_types)
+        for _fqn, lc, _parent, _attr in model_config.walk(Linear.Config)
+    )
+    has_quant_moe = any(
```
tianyu-l, line 82: may be more efficient to use `next()` to early exit upon seeing the first instance

torchtitan/components/quantization/__init__.py
acisseJZhong: yes this would work if we no longer inherit from Linear. 

I need to override Config.build if we inherit from Linear. The default Module.Config.build() would call Float8Linear(config=<Config>), which is not what torchao expects the parameters to be.

torchtitan/components/quantization/__init__.py
acisseJZhong: yeah you are right, this build is not needed.

torchtitan/components/quantization/__init__.py
acisseJZhong: yeah I think both way work? I don't know which one is better tho. We can define base QuantizedLinearConfig and _QuantizedMoEConfig.

torchtitan/components/quantization/__init__.py
acisseJZhong: we still need this for MXFP8Linear? no? I removed the one for MoE.

tianyu-l: according to https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/mx.py#L76, linear doesn't need it

acisseJZhong: I see, this should be only for moe grouped GEMM.

torchtitan/components/quantization/__init__.py
acisseJZhong: simply removed, torchao default this field to False. We will never set it to True in titan.

torchtitan/components/quantization/utils.py
acisseJZhong: TIL: any() already short-circuits — it returns True as soon as the first truthy element is found.

default.profraw
tianyu-l, line 1: remove!

torchtitan/components/quantization/float8.py
```diff
@@ -3,308 +3,223 @@
...
-                        "To use _auto_filter_for_recipe, please install torchao nightly build."
-                    )
+                    "Using default module_filter_fn for float8 model conversion. "
+                    "To use _auto_filter_for_recipe, please install torchao nightly build."
                 )
+                filter_fn = partial(module_filter_fn, filter_fqns=clean_fqns)
+        else:
+            filter_fn = partial(module_filter_fn, filter_fqns=clean_fqns)
+
+        # Walk config tree and swap Linear.Config → Float8Linear.Config
```
tianyu-l: let's put everything above this into `__init__()` of this converter

torchtitan/components/quantization/__init__.py
```diff
@@ -11,47 +11,51 @@
...
-    All quantization converter classes should inherit from this so they can be
-    identified via isinstance checks.
-    """
+    pass
 
-    @dataclass(kw_only=True, slots=True)
-    class Config(Configurable.Config):
-        _quantization_type: ClassVar[str]
 
+class _QuantizedGroupedExpertsConfig:
```
tianyu-l: makes sense

torchtitan/components/quantization/float8.py
```diff
@@ -3,308 +3,223 @@
...
-        ), "Float8 MoE training prototype does not yet support context parallelism"
-
-        self.enabled = True
-
-    def convert(self, model: nn.Module):
-        """
-        Mutates the model inplace replacing instances of nn.Parameter with ScaledGroupedMMTensor,
-        to perform dynamic float8 rowwise quantization + scaled grouped GEMMs for the target MoE FQNs.
-        """
+        from torchao.prototype.moe_training.config import Float8TrainingOpConfig
```
tianyu-l, line 200: everything above should be in constructor

torchtitan/components/quantization/mx.py
```diff
@@ -48,87 +63,118 @@ class Config(QuantizationConverter.Config):
...
-        ), "MXFP8 is only supported on SM100 or architectures"
+        ), "MXFP8 is only supported on SM100 or later architectures"
 
-        if not model_compile_enabled:
+        if not self.config.model_compile_enabled:
             logger.warning(
-                "torch.compile enablement is required for highest performance of MXFP8 dynamic quantization."
+                "torch.compile enablement is required for highest performance "
+                "of MXFP8 dynamic quantization."
             )
```
tianyu-l, line 83: put in constructor

torchtitan/components/quantization/mx.py
```diff
@@ -48,87 +63,118 @@ class Config(QuantizationConverter.Config):
...
-            model, ["_init_mean", "_init_std"], nn_module_cls=nn.Linear
-        )
+        recipe_name = self.config.recipe_name
 
-        recipe = MXFP8TrainingRecipe(self.config.recipe_name)
-        mxfp8_op_config = MXFP8TrainingOpConfig.from_recipe(recipe)
-        mxfp8_op_config.pad_token_groups_for_grouped_mm = (
-            self.pad_token_groups_for_grouped_mm
-        )
+        _converted_config_cache: dict[type, type] = {}
```
tianyu-l, line 149: similar for above

tianyu-l (approved):
nice! please address final nits

torchtitan/components/quantization/float8.md
tianyu-l, line 1: can we move these docs to be under components/quantization? The long term goal is to move all docs to be close to where relevant.

torchtitan/components/quantization/module_utils.py
acisseJZhong, line 1: cc @fegin as I am deleting this file.

acisseJZhong:
auto parallel tests failing due to https://github.com/meta-pytorch/autoparallel/pull/415, update in https://github.com/pytorch/torchtitan/pull/3114

graph trainer issue was known: same as https://github.com/pytorch/torchtitan/actions/runs/25016687204/job/73266229364

PR #2337, opened 2026-02-06
Remove `_ScaledPartial` placement

Aidyn-A:
A follow up on https://github.com/pytorch/torchtitan/pull/2313#discussion_r2771922920. This PR removes `_ScaledPartial` placement in favor of `Partial(reduce_op="sum")` placement.

cc @tianyu-l, @wwwjn

PR #3035, opened 2026-04-20
Enable lora training with simplefsdp

jialei777:
Added ignored_modules to data_parallel API in simpleFSDP: 

It allows explicitly passing parameters (like lora adaptor) into ignored_modules that FSDP should completely skip over. We will then do DDP over those parameter via another wrapper of data_parallel API.

### DIFF
--- a/torchtitan/experiments/graph_trainer/simple_fsdp.py
+++ b/torchtitan/experiments/graph_trainer/simple_fsdp.py
@@ -45,37 +45,6 @@ class MixedPrecisionPolicy:
     reduce_dtype: torch.dtype | None = None
 
 
-class _ScaledPartial(Partial):
-    # A subclass of Partial placement that allows user to perform reduction with a custom
-    # factor (reduction_divide_factor) other than the default world size.
-    def __init__(
-        self,
-        reduction_divide_factor: float,
-    ):
-        self.reduction_divide_factor = reduction_divide_factor
-        super().__init__(reduce_op="sum")
-
-    def _reduce_value(
-        self, tensor: torch.Tensor, mesh: DeviceMesh, mesh_dim: int
-    ) -> torch.Tensor:
-        # for all_reduce in DDP
-        tensor.div_(self.reduction_divide_factor)
-        reduced = super()._reduce_value(tensor, mesh, mesh_dim)
-        return reduced
-
-    def _reduce_shard_value(
-        self,
-        tensor: torch.Tensor,
-        mesh: DeviceMesh,
-        mesh_dim: int,
-        shard_spec: Placement,
-    ) -> torch.Tensor:
-        # for reduce_scatter in FSDP
-        tensor.div_(self.reduction_divide_factor)
-        reduced = super()._reduce_shard_value(tensor, mesh, mesh_dim, shard_spec)
-        return reduced
-
-
 def _distribute_dtensor(
     tensor: DTensor,
     device_mesh: DeviceMesh,
@@ -203,7 +172,6 @@ class ReplicateComputation(Module):
         mode: str,
         mp_policy: MixedPrecisionPolicy | None,
         full_dtensor: bool = False,
-        reduction_divide_factor: float | None = None,
     ) -> None:
         super().__init__()
         self.device_mesh = device_mesh
@@ -211,11 +179,7 @@ class ReplicateComputation(Module):
         self.mode = mode
         self.compute_placements: list[Placement] = [Replicate()] * self.device_mesh.ndim
         self.grad_placements: list[Placement] = [
-            _ScaledPartial(
-                reduction_divide_factor=reduction_divide_factor,
-            )
-            if reduction_divide_factor is not None
-            else Partial(reduce_op="avg")
+            Partial(reduce_op="sum")
         ] * self.device_mesh.ndim
         mp_policy = mp_policy or MixedPrecisionPolicy()
         self.param_dtype: torch.dtype | None = mp_policy.param_dtype
@@ -300,7 +264,6 @@ def data_parallel(
     mode: str = "replicate",
     mp_policy: MixedPrecisionPolicy | None = None,
     shard_dim: int = 0,
-    reduction_divide_factor: float | None = None,
     full_dtensor: bool = False,
     ignored_params: set[nn.Parameter] | None = None,
 ) -> nn.Module:
@@ -371,7 +334,6 @@ def data_parallel(
                 param_sharding,
                 mode,
                 mp_policy=mp_policy,
-                reduction_divide_factor=reduction_divide_factor,
                 full_dtensor=full_dtensor,
             ),
         )
--- a/torchtitan/experiments/graph_trainer/simple_fsdp.py
+++ b/torchtitan/experiments/graph_trainer/simple_fsdp.py
@@ -45,6 +45,37 @@ class MixedPrecisionPolicy:
     reduce_dtype: torch.dtype | None = None
 
 
+class _ScaledPartial(Partial):
+    # A subclass of Partial placement that allows user to perform reduction with a custom
+    # factor (reduction_divide_factor) other than the default world size.
+    def __init__(
+        self,
+        reduction_divide_factor: float,
+    ):
+        self.reduction_divide_factor = reduction_divide_factor
+        super().__init__(reduce_op="sum")
+
+    def _reduce_value(
+        self, tensor: torch.Tensor, mesh: DeviceMesh, mesh_dim: int
+    ) -> torch.Tensor:
+        # for all_reduce in DDP
+        tensor.div_(self.reduction_divide_factor)
+        reduced = super()._reduce_value(tensor, mesh, mesh_dim)
+        return reduced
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/graph_trainer/simple_fsdp.py
```diff
@@ -45,6 +45,37 @@ class MixedPrecisionPolicy:
     reduce_dtype: torch.dtype | None = None
 
 
+class _ScaledPartial(Partial):
```
tianyu-l: This was removed in https://github.com/pytorch/torchtitan/pull/2337 in favor of trainer refactor.

> Added _ScaledPartial Class: This enables custom scaling during DDP/FSDP gradient reduction across the cluster (via reduction_divide_factor), which is often needed in LoRA to manage learning rate scaling.

How is it related to LoRA?

tianyu-l (changes requested):

torchtitan/experiments/graph_trainer/simple_fsdp.py
jialei777: I was using an old version of the codebase. Forwarded now.

torchtitan/experiments/graph_trainer/simple_fsdp.py
```diff
@@ -265,10 +265,11 @@ def data_parallel(
     mp_policy: MixedPrecisionPolicy | None = None,
     shard_dim: int = 0,
     full_dtensor: bool = False,
+    ignored_params: set[nn.Parameter] | None = None,
 ) -> nn.Module:
     param_sharding: tuple[Placement, ...]
     if mode == "replicate":
-        param_sharding = (Replicate(),)
+        param_sharding = (Replicate(),) * device_mesh.ndim
```
tianyu-l, line 272: why this change?

torchtitan/experiments/graph_trainer/simple_fsdp.py
```diff
@@ -289,6 +290,14 @@ def data_parallel(
         if "SimpleFSDP" in mod.__class__.__name__:
             continue
 
+        # Skip parameters that are explicitly ignored (e.g., shared or frozen weights).
+        if ignored_params is not None:
+            params_dict = {
+                p_name: p for p_name, p in params_dict.items() if p not in ignored_params
+            }
+            if not params_dict:
+                continue
```
tianyu-l, line 299: this is applying FSDP sharding to entire module as long as any param is not ignored?

jialei777: yes, if we do mix sharding, i.e., lora with ddp and main model with fsdp, we want to ignore the lora adaptor here

torchtitan/experiments/graph_trainer/simple_fsdp.py
jialei777: if using HSDP, the DP mesh is 2D. So one for data_parallel_replicate_degree, and one for data_parallel_shard_degree.

tianyu-l: hsdp is 3 lines below

torchtitan/experiments/graph_trainer/simple_fsdp.py
tianyu-l: then it's ignoring module, not ignoring individual params

tianyu-l (commented):
cc @mori360

torchtitan/experiments/graph_trainer/simple_fsdp.py
jialei777: ok, i guess our use case is forcing lora adaptor to Replicate no matter what device_mesh is specified for main model. If it is not the desired use case i can revert this line.

torchtitan/experiments/graph_trainer/simple_fsdp.py
jialei777: ok, just to ensure i understand things correctly, you are suggest to rename it from `ignored_params` to `ignored_modules`?

tianyu-l (commented):
This proposal is missing details and thus making review difficult:
The goal of this PR is to support / unblock LoRA, but the PR didn't show how you plan to do LoRA. Instead, it changes capability of simple fsdp API, which could've been used for other cases than LoRA.

jialei777:
@tianyu-l the API should be kept the same for other use case. I am just proposing to add a `ignored_modules` input arg.

Our use case is, for LoRA training, we would like to do mix sharding, i.e., main model weights via simplefsdp, where lora adpator via DDP (i.e., simplefsdp with mode=replicate or just standard DDP without simplefsdp). In both case,  data_parallel API needs to support ignored_modules, where we are flexible to shard only partial of model weights. 

Does this make sense to you? Any other suggestions on doing lora training with simplefsdp?

mori360:
We currently have the plan to add lora as converter, freezing base weights via requires_grad_(False), which     integrates with existing parallelism. https://github.com/pytorch/torchtitan/pull/2484
The LoRA adapters are so tiny that the trade-off between sharding them (to save memory) vs replicating them (to save gather/scatter overhead) is negligible either way. Given this, are there any motivation for needing ignored_modules?

jialei777:
@mori360 The change we proposed is specifically for simplefsdp, which with our benchmark is much faster than fsdp2 with torch.compile. 

Currently, if we also shard lora parameters with simplefsdp we see very poor performance, that is why we want to add this arg to explicitly do ddp for lora adaptor while keep the frozen whole model fsdp sharded with simplefsdp. Is this doable with your lora implementation?

jialei777:
To be explicit, we are doing lora with something like

```python
dp_mesh = parallel_dims.get_mesh(["dp_replicate", "fsdp"])
model = data_parallel(
    model,
    dp_mesh,
    mode="fully_shard",
    ignored_params=lora_params,
)


adapter_group = torch.nn.ModuleList([m.adapters for m in adapted_layers])

torch.distributed._composable.replicate(
    adapter_group,
    device_mesh=dp_mesh,
    bucket_cap_mb=100,
)

# OR doing DDP with simpleFSDP, which is less performant as bucket is not supported
# model = data_parallel(
#     model,
#     dp_mesh,
#     mode="replicate",
#     ignored_params=non_lora_params,
# )
```

mori360:
> Is this doable with your lora implementation?

Thanks for the reply — it's doable to add LoRA adapters as ignored_params. Our LoRA converter already provides a key_filter that identifies all adapter parameters by name, which can directly feed into the ignore list.
You mentioned "SimpleFSDP is much faster than FSDP2 with torch.compile" — is that performance gap with or without LoRA? I'm curious whether the gap is from the base infrastructure or specific to the LoRA adapter sharding overhead.
Also, do you have numbers on the improvement from replicating adapters vs sharding them with SimpleFSDP? That would help us understand whether to prioritize this integration.

jialei777:
@mori360:
re "SimpleFSDP is much faster than FSDP2 with torch.compile": we benchmarked and it is the case with lora and full finetuning. We observed about 15%-25% gain with simplefsdp+compile over fsdp2+compile.

We also benchmarked: simplefsdp for both model weights and adapter is like 40% slower than simplefsdp for main model and ddp for lora adaptor. this actaully makes me wonder if I was doing something incorrect. 

Our use case is not standard, we can chat a bit if you are interested.

re this PR: how do you suggest we move forward? I would hope we can merge this so we don't need to patch our fork of torchtitan.

torchtitan/experiments/graph_trainer/simple_fsdp.py
fegin: If it is `ignored_modules`, the typing should be `ignored_modules: Iterable[nn.Module] | None = None`. It is confusing to use parameters to decided modules. `Iterable[nn.Parameter]` should be preserved for truely `ignored_parameters`, which is used by fsdp2.

tianyu-l:
> we benchmarked and it is the case with lora and full finetuning. We observed about 15%-25% gain with simplefsdp+compile over fsdp2+compile.

Do you use any graph optimizations from simplefsdp + compile, e.g. bucketing / reordering? It sounds surprising that simplefsdp could be 25% faster -- is it not transformer architecture?

> We also benchmarked: simplefsdp for both model weights and adapter is like 40% slower than simplefsdp for main model and ddp for lora adaptor. this actaully makes me wonder if I was doing something incorrect.

Let's be careful here. Is it because simplefsdp still runs reduce-scatter for non-adapter params? I would prefer to understand the root cause before making API changes.

jialei777:
> Do you use any graph optimizations from simplefsdp + compile

No, just simplefsdp + per-layer compile vs fsdp2 + per-layer compile on a transformer model. 

> Is it because simplefsdp still runs reduce-scatter for non-adapter params?

That should be the same for simplefsdp with lora adaptor and simplefsdp only for non-lora adaptor right?

We did not test very carefully with simplefsdp sharding for both main model and lora adaptor. Because we see performance gain when we do main model fsdp and lora ddp with eager. And it feel reasonable to do the same thing with simpleFSDP with compile.

PR #3038, opened 2026-04-21
[profiler] Suppress Callable field from tyro CLI parsing

saforem2:
## Summary

Fixes a crash introduced by #2926 where `trace_post_processor: Function.Config | None` in `Profiler.Config` causes `tyro.cli()` to fail because tyro cannot parse `Callable[..., Any]` types from CLI arguments.

**Error:**
```
Invalid input to tyro.cli()
Unsupported type annotation for field profiler.trace-post-processor.fn
with type collections.abc.Callable[..., typing.Any]
```

**Fix:** Annotate `trace_post_processor` with `tyro.conf.Suppress`, matching the existing pattern used for:
- `model_spec` in `trainer.py:62`
- `sample_processor` in `text_datasets.py:502`

The field defaults to `None` and is only set programmatically by `CUDAGraphWrapper`, never via CLI.

## Test plan

- [ ] Verify `python -m torchtitan.train ...` no longer crashes at config parsing
- [ ] Verify graph_trainer still works (trace_post_processor set programmatically)

### DIFF
--- a/torchtitan/tools/profiler.py
+++ b/torchtitan/tools/profiler.py
@@ -10,9 +10,8 @@ import time
 from dataclasses import dataclass
 from typing import Annotated
 
-import tyro
 import torch
-
+import tyro
 from torchtitan.config import Configurable
 from torchtitan.config.function import Function
 from torchtitan.tools.logging import logger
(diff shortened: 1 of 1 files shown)


wwwjn (commented):
Thanks this PR looks good to me and I also verified it fixed locally, can you fix the lint before merging?

wwwjn (approved):

PR #3041, opened 2026-04-21
Move RL batch-invariant tests to A10G + Fix RL CI

wwwjn:
Move batch-invariant tests from H100 to A10G to relief H100 CI starvation.
Depend on pytorch side change: https://github.com/pytorch/pytorch/pull/179760

We will need `VLLM_USE_FLASHINFER_SAMPLER=0` because  https://github.com/vllm-project/vllm/pull/40376 landed Apr. 29. For our CI environment, we didn't install nvcc so it won't support FlashInfer to be JIT compiled.


tianyu-l (approved):

felipemello1 (approved):

wwwjn:
This PR will be separate into 2:
1. Fix CI only: https://github.com/pytorch/torchtitan/pull/3232
2. Move bit-wise parity test to A10G.  TBA

PR #3042, opened 2026-04-21
[graph_trainer] Add CP support to CooR precompile

bobrenjc93:
* __->__ #3042
* #2916

Add context parallel support to the precompile path by calling
prepare_context_parallel_input after build_forward_extra_kwargs,
matching the trainer's post_dataloading_process.

prepare_context_parallel_input uses _context_parallel_shard which
is already CooR-safe (no rank-specific operations), so no additional
adaptation is needed.


bobrenjc93:
Doing a big refactor and will stack CP support on top of that PR

PR #3046, opened 2026-04-21
Support mark_unbacked in minimal_fx_tracer

tugsbayasgalan:
minimal_fx_tracer now supports mark_unbacked() on plain tensor inputs.

Because this non-strict make_fx path fakeifies inputs directly, it must call
compute_unbacked_bindings() itself after fakeification; otherwise fresh
unbacked symbols remain pending and tracing fails.

Scope is intentionally narrow:
  - support plain-tensor mark_unbacked()
  - reject mark_dynamic()
  - reject wrapper-subclass mark_unbacked() such as DTensor

  Tests cover replay, multiple inputs, shape_id/strict/min-max metadata,
  data-dependent branching failure on unbacked dims, and DTensor rejection.

  Co-authored-by: Codex <codex@openai.com>

Authored with Codex

### DIFF
--- a/torchtitan/experiments/graph_trainer/dynamic_shapes.py
+++ b/torchtitan/experiments/graph_trainer/dynamic_shapes.py
@@ -1,0 +1,121 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Dynamic-shape helpers used by minimal_fx_tracer.
+
+Covers:
+- detection of mark_unbacked() metadata on plain and wrapper-subclass tensors,
+- input fakeification with a StatelessSymbolicContext built from that metadata,
+- post-trace materialization of deferred ShapeEnv runtime asserts and
+  symbolic-input guards.
+"""
+
+from typing import Any
+
+import torch
+from torch._subclasses import FakeTensorMode
+from torch.utils._python_dispatch import is_traceable_wrapper_subclass
+
+
+def _tensor_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+    return bool(
+        getattr(tensor, "_dynamo_unbacked_indices", None)
+        or getattr(tensor, "_dynamo_strict_unbacked_indices", None)
+    )
+
+
+def _wrapper_subclass_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+    if not is_traceable_wrapper_subclass(tensor):
+        return False
+    if _tensor_has_mark_unbacked(tensor):
+        return True
+
+    attrs, _ = tensor.__tensor_flatten__()
+    for attr in attrs:
+        inner_value = getattr(tensor, attr)
+        if not isinstance(inner_value, torch.Tensor):
+            continue
+        if _tensor_has_mark_unbacked(
+            inner_value
+        ) or _wrapper_subclass_has_mark_unbacked(inner_value):
+            return True
+    return False
+
+
+def _symbolic_context_for_marked_dims(tensor: torch.Tensor) -> Any | None:
+    from torch.fx.experimental.symbolic_shapes import (
+        DimDynamic,
+        RelaxedUnspecConstraint,
+        StatelessSymbolicContext,
+    )
+
+    marked_unbacked_indices = getattr(tensor, "_dynamo_unbacked_indices", set())
+    marked_strict_unbacked_indices = getattr(
+        tensor, "_dynamo_strict_unbacked_indices", set()
+    )
+    has_outer_marked_dims = bool(
+        marked_unbacked_indices or marked_strict_unbacked_indices
+    )
+    if not has_outer_marked_dims:
+        return None
+
+    dynamic_sizes = [DimDynamic.STATIC] * tensor.dim()
+    constraint_sizes = [None] * tensor.dim()
+    specialize_on = [
+        list(getattr(tensor, "_specialize_on", {}).get(i, []))
+        for i in range(tensor.dim())
+    ]
+
+    for dim in range(tensor.dim()):
+        if dim in marked_unbacked_indices:
+            dynamic_sizes[dim] = DimDynamic.UNBACKED
+        elif dim in marked_strict_unbacked_indices:
+            dynamic_sizes[dim] = DimDynamic.UNBACKED
+            constraint_sizes[dim] = RelaxedUnspecConstraint(warn_only=False)
+
+    return StatelessSymbolicContext(
+        dynamic_sizes=dynamic_sizes,
+        constraint_sizes=constraint_sizes,
+        specialize_on=specialize_on,
+        shape_ids=getattr(tensor, "_dynamo_shape_ids", None),
+        unbacked_bounds=getattr(tensor, "_dynamo_unbacked_bounds", None),
+    )
+
+
+def _fakeify_input(
+    fake_mode: FakeTensorMode, arg: torch.Tensor, *, input_name: str
+) -> torch.Tensor:
+    from torch._dynamo.source import LocalSource
+
+    symbolic_context = _symbolic_context_for_marked_dims(arg)
+    if symbolic_context is None:
+        return fake_mode.from_tensor(arg, static_shapes=True)
+
+    source = LocalSource(input_name, is_input=True)
+    return fake_mode.from_tensor(
+        arg,
+        source=source,
+        symbolic_context=symbolic_context,
+    )
+
+
+def _insert_runtime_asserts(
+    gm: torch.fx.GraphModule, fake_mode: FakeTensorMode
+) -> None:
+    """Materialize the ShapeEnv's deferred runtime asserts into the graph.
+
+    Constraints accumulated during fake-tensor propagation (e.g. ``u0 >= 0``,
+    ``torch._check(...)``) only live in the ShapeEnv. Without this pass they
+    are silently dropped at runtime, so unbacked-symbol assumptions go
+    unchecked. Mirrors the call Dynamo and HOP-subgraph proxy tracing make
+    after capture.
+    """
+    if fake_mode.shape_env is None:
+        return
+    from torch.fx.passes.runtime_assert import insert_deferred_runtime_asserts
+
+    insert_deferred_runtime_asserts(gm, fake_mode.shape_env, "minimal_fx_tracer")
+    gm.recompile()
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -6,7 +6,7 @@
 
 import copy
 from collections.abc import Callable, Generator
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
 from dataclasses import dataclass, field
 from typing import Any
 
@@ -19,6 +19,13 @@ from torch.fx.experimental.proxy_tensor import make_fx
 from torch.fx.traceback import preserve_node_meta
 from torch.nn.utils import stateless
 from torch.utils._python_dispatch import is_traceable_wrapper_subclass
+
+from torchtitan.experiments.graph_trainer.dynamic_shapes import (
+    _fakeify_input,
+    _insert_runtime_asserts as _insert_runtime_asserts_pass,
+    _tensor_has_mark_unbacked,
+    _wrapper_subclass_has_mark_unbacked,
+)
 
 # Tensors and make_fx-safe primitives are allowed as pytree leaves in args.
 # Everything else (callables, custom objects) should be registered as pytree
@@ -82,7 +89,8 @@ def _unwrap_subclass(t: torch.Tensor) -> tuple[list[torch.Tensor], SubclassMeta 
 
 
 def _wrap_to_subclass(
-    plain_tensors: list[torch.Tensor], meta: SubclassMeta
+    plain_tensors: list[torch.Tensor],
+    meta: SubclassMeta,
 ) -> torch.Tensor:
     inner_dict = {}
     idx = 0
@@ -94,8 +102,12 @@ def _wrap_to_subclass(
             inner_dict[attr] = inner_tensors[0]
         else:
             inner_dict[attr] = _wrap_to_subclass(list(inner_tensors), inner_meta)
+
     return meta.cls.__tensor_unflatten__(
-        inner_dict, meta.ctx, meta.outer_size, meta.outer_stride
+        inner_dict,
+        meta.ctx,
+        meta.outer_size,
+        meta.outer_stride,
     )
 
 
@@ -302,7 +314,11 @@ class TracedResult:
         )
 
 
-def minimal_fx_tracer(fn: Callable) -> Callable[..., TracedResult]:
+def minimal_fx_tracer(
+    fn: Callable,
+    *,
+    _insert_runtime_asserts: bool = False,
+) -> Callable[..., TracedResult]:
     """Return a tracer for a stateless ``fn`` with explicit ``state`` input.
 
     ``fn`` must be a plain callable (not an ``nn.Module``). The returned
@@ -321,6 +337,12 @@ def minimal_fx_tracer(fn: Callable) -> Callable[..., TracedResult]:
     Tensor subclasses (for example ``DTensor``) are recursively unwrapped into
     plain tensors for tracing, and the layouts needed to rewrap them are stored
     in the returned :class:`TracedResult`.
+
+    ``_insert_runtime_asserts`` opts into materializing the ShapeEnv's deferred
+    runtime asserts (from ``mark_unbacked()`` bounds and ``torch._check()``
+    calls) into the graph as ``_assert_scalar`` nodes. Off by default because
+    cudagraph capture does not evaluate these nodes, and downstream graph
+    passes generally don't need them.
     """
 
     def _trace_with_args(state: Any, *args: Any) -> TracedResult:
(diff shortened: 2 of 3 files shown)


tugsbayasgalan:
@claude Pls review

tugsbayasgalan:
@claude Do another round

tugsbayasgalan:
@claude Do review

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -110,6 +141,148 @@ def forward(self, x):
...
+        x = torch.randn(3, 4)
+        mark_unbacked(x, 0, strict=True, min=2, max=5)
+
+        traced = minimal_fx_tracer(forward)({}, x)
+        fake_x = next(
+            node.meta["val"]
+            for node in traced.gm.graph.nodes
+            if node.op == "placeholder"
+        )
+        x_min = torch.randn(2, 4)
```
laithsakka, line 193: what are you trying to test here in this test did you want to add 
if (x.size()[0]>=2 and  x.size()[0]<=5) int the forward and ensure it does not DDE?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -110,6 +141,148 @@ def forward(self, x):
...
+        x_max = torch.randn(5, 4)
+
+        self.assertIsInstance(fake_x.size(0), torch.SymInt)
+        self.assertEqual(fake_x.size(1), 4)
+        self.assertTrue(torch.equal(forward({}, x_min), run_traced(traced, {}, x_min)))
+        self.assertTrue(torch.equal(forward({}, x_max), run_traced(traced, {}, x_max)))
+
+    def test_mark_unbacked_preserves_unbacked_placeholder_dim(self):
+        from torch._dynamo.decorators import mark_unbacked
```
laithsakka, line 203: this test basically ensure when you call mark_unbacked dim is dynamic right?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -110,6 +141,148 @@ def forward(self, x):
...
+
+    def test_mark_unbacked_shape_id_multiple_inputs_replay(self):
+        from torch._dynamo.decorators import mark_unbacked
+
+        def forward(_state, x, y):
+            return x.sin() + y.cos()
+
+        x = torch.randn(2, 4)
+        y = torch.randn(2, 4)
+        mark_unbacked(x, 0, shape_id="batch")
```
laithsakka, line 305: better you can add 
if (x.size()[0]==y.size([0])) and ensure it does not DDE. but its fine as is also.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -349,20 +470,39 @@ def _trace_with_args(state: Any, *args: Any) -> TracedResult:
...
+            _fakeify_input(fake_mode, a, input_name=f"minimal_fx_tracer_input_{i}")
+            for i, a in enumerate(unwrapped_args)
         )
+        if fake_mode.shape_env is not None:
+            # Input fakeification can allocate fresh unbacked symbols. In the
+            # non-strict tracer path we don't go through Dynamo's placeholder
+            # binding logic, so bind them here explicitly. mark_unbacked() is
+            # only supported on plain tensor inputs, so binding after subclass
+            # unwrapping is sufficient.
+            torch.fx.experimental.symbolic_shapes.compute_unbacked_bindings(
```
tugsbayasgalan: cc: @laithsakka  does this make sense?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -119,6 +124,115 @@ def _unwrap_subclasses(
     return flat, layouts
 
 
+def _tensor_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+    return bool(
+        getattr(tensor, "_dynamo_unbacked_indices", None)
+        or getattr(tensor, "_dynamo_strict_unbacked_indices", None)
+        or getattr(tensor, "_dynamo_shape_ids", None)
```
laithsakka: not needed exisits iff _dynamo_unbacked_indices exisit

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -119,6 +124,115 @@ def _unwrap_subclasses(
...
+        or getattr(tensor, "_dynamo_unbacked_bounds", None)
```
laithsakka: ditto  not needed

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: updated the  test to be more rigorous

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: yeah

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: yep added.

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+            inner_value
+        ) or _wrapper_subclass_has_mark_unbacked(inner_value):
+            return True
+    return False
+
+
+def _symbolic_context_for_marked_dims(
+    tensor: torch.Tensor,
+    *,
+    source: Any,
```
sanketpurandare: What is source used for?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -83,22 +83,53 @@ def _apply_regional_inductor(traced_result):
...
+        (
+            val.fake_mode
+            for val in fake_inputs
+            if isinstance(val, torch.Tensor) and hasattr(val, "fake_mode")
+        ),
+        None,
+    )
+
+
+def _standalone_compile_graph(gm):
```
sanketpurandare: Where is this function used?

tugsbayasgalan: Good catch not needed.

torchtitan/experiments/graph_trainer/dynamic_shapes.py
tugsbayasgalan: I think it is internally used for guard generation etc. @laithsakka  would know more i think.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -83,22 +83,53 @@ def _apply_regional_inductor(traced_result):
...
+    fake_inputs = _graph_placeholder_fake_inputs(traced_result.gm)
+    fake_mode = _graph_fake_mode(fake_inputs)
+    with torch._guards.tracing(torch._guards.TracingContext(fake_mode)):
         traced_result.gm = regional_inductor(traced_result.gm)
 
     traced_result.gm.graph.set_codegen(CodeGen())
     traced_result.gm.recompile()
 
 
+def _graph_placeholder_fake_inputs(gm):
```
laithsakka, line 95: does the new api support List[Tensor] ..etc? do we need to test that?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -110,6 +141,193 @@ def forward(self, x):
...
+
+        traced = minimal_fx_tracer(forward)({}, x)
+        fake_x = next(
+            node.meta["val"]
+            for node in traced.gm.graph.nodes
+            if node.op == "placeholder"
+        )
+        x_other = torch.randn(3, 4)
+
+        self.assertIsInstance(fake_x.size(0), torch.SymInt)
```
laithsakka, line 220: nit: check it has unbacked symbol? if you want.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -279,6 +497,78 @@ def test_core_explicit_state_train_step(self):
         for gr, gt in zip(grads_ref, grads_tr, strict=True):
             self.assertTrue(torch.equal(gr, gt))
 
+    def test_flex_attention_block_mask_mark_unbacked(self):
+        from torch._dynamo.decorators import mark_unbacked
+        from torch.nn.attention.flex_attention import AuxRequest, BlockMask, flex_attention
+
+        maybe_register_blockmask_pytree_node()
+
+        def make_mask_fn(attn_regions, document_ids):
```
laithsakka, line 499: i will not review this test,

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+
+import torch
+from torch._subclasses import FakeTensorMode
+from torch.utils._python_dispatch import is_traceable_wrapper_subclass
+
+
+def _tensor_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+    return bool(
+        getattr(tensor, "_dynamo_unbacked_indices", None)
+        or getattr(tensor, "_dynamo_strict_unbacked_indices", None)
```
laithsakka, line 26: yeh this one when we do mark_ubacked(strict=True) you are not really testing it, but i guess thats fine. 
I want to delete this mode, this only allow throwing compile time errors if traced code have
torch._check(u0== 100(any integr)

torchtitan/experiments/graph_trainer/dynamic_shapes.py
laithsakka: source not used in this function, we shall remove it probably?

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+    *,
+    source: Any,
+) -> Any | None:
+    from torch.fx.experimental.symbolic_shapes import (
+        DimDynamic,
+        RelaxedUnspecConstraint,
+        StatelessSymbolicContext,
+    )
+
+    marked_unbacked_indices = getattr(tensor, "_dynamo_unbacked_indices", set())
```
laithsakka: do you want to error if mark_dynamic is called to ensure its NOT supported. ?

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+    return StatelessSymbolicContext(
+        dynamic_sizes=dynamic_sizes,
+        constraint_sizes=constraint_sizes,
+        specialize_on=specialize_on,
+        shape_ids=getattr(tensor, "_dynamo_shape_ids", None),
+        unbacked_bounds=getattr(tensor, "_dynamo_unbacked_bounds", None),
+    )
+
+
+def _fakeify_input(fake_mode: FakeTensorMode, arg: Any, *, input_name: str) -> Any:
```
laithsakka: arg: Any -> arg: torch.Tensor?

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+    if not isinstance(arg, torch.Tensor):
+        return arg
+
+    from torch._dynamo.source import LocalSource
+    from torch.fx.experimental.symbolic_shapes import TrackedFake
+
+    source = LocalSource(input_name, is_input=True)
+    symbolic_context = _symbolic_context_for_marked_dims(arg, source=source)
+    if symbolic_context is not None:
```
laithsakka: nit: early exit first

if symbolic_context is None: 
    # nothing to make dynamic. 
    fake_mode.from_tensor(arg, static_shapes=True)

but i mean 
...
https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+        shape_env = fake_mode.shape_env
+        assert shape_env is not None and shape_env.tracked_fakes is not None
+        shape_env.tracked_fakes.append(TrackedFake(fake, source, symbolic_context))
+        return fake
+    return fake_mode.from_tensor(arg, static_shapes=True)
+
+
+def _produce_symbolic_input_guards(fake_mode: FakeTensorMode) -> None:
+    shape_env = fake_mode.shape_env
```
laithsakka: why do you need to produce guards do we have a path in the non strict API that evaluate guards. ?
also for unabcked there shall not be guards  (well now we have some but its a bug).

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+
+    from torch._dynamo.source import LocalSource
+    from torch.fx.experimental.symbolic_shapes import TrackedFake
+
+    source = LocalSource(input_name, is_input=True)
+    symbolic_context = _symbolic_context_for_marked_dims(arg, source=source)
+    if symbolic_context is not None:
+        fake = fake_mode.from_tensor(
+            arg,
+            source=source,
```
laithsakka: I believe if you dont care about guards  you do not need the source at all you can pass none. 
but you can keep it as is its ok also.

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,146 @@
...
+            symbolic_context=symbolic_context,
+        )
+        shape_env = fake_mode.shape_env
+        assert shape_env is not None and shape_env.tracked_fakes is not None
+        shape_env.tracked_fakes.append(TrackedFake(fake, source, symbolic_context))
```
laithsakka: same thing here do you need to track if you are not calling _produce_symbolic_input_guards? probably not in theory.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
laithsakka: yeh 
since you are ignoring compute_unbacked_bindings result you can instead use
```
with shape_env.ignore_fresh_unbacked_symbols():
    fake_args = tuple(
    fakeify_input(fake_mode, a, input_name=f"minimal_fx_tracer_input{i}")
    for i, a in enumerate(unwrapped_args)
```

laithsakka: since you are ignoring compute_unbacked_bindings result you can instead use 
with shape_env.ignore_fresh_unbacked_symbols():
    fake_args = tuple(
            _fakeify_input(fake_mode, a, input_name=f"minimal_fx_tracer_input_{i}")
            for i, a in enumerate(unwrapped_args)

tugsbayasgalan: Hmm we shouldn;t ignore them, no?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: this is just test api so it doens't matter i think

torchtitan/experiments/graph_trainer/dynamic_shapes.py
tugsbayasgalan: delketed now. we don't need guards

torchtitan/experiments/graph_trainer/dynamic_shapes.py
tugsbayasgalan: yep deleted.

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -0,0 +1,126 @@
...
+
+    source = LocalSource(input_name, is_input=True)
+    fake = fake_mode.from_tensor(
+        arg,
+        source=source,
+        symbolic_context=symbolic_context,
+    )
+    shape_env = fake_mode.shape_env
+    assert shape_env is not None and shape_env.tracked_fakes is not None
+    shape_env.tracked_fakes.append(TrackedFake(fake, source, symbolic_context))
```
laithsakka: probably this is not needed also.

laithsakka (approved):

torchtitan/experiments/graph_trainer/dynamic_shapes.py
ydwu4: Need verify whether tracked_fakes are needed.

ydwu4 (approved):
Implementation looks reasonable.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -421,6 +452,7 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
         _copy_fwd_metadata_to_bw_nodes(traced)
 
         _remove_cpu_shadow_chains(traced)
+        _insert_runtime_asserts(traced, fake_mode)
```
SherlockNoMad: what does this do to the graph if no dynamic shape is used?

we will need a pass to remove these runtime asserts. 
In cudagraph setup, these nodes are not evaluate anyway. 

I would have a flag to add runtime assert conditionally, and default to false.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -349,20 +367,33 @@ def _trace_with_args(state: Any, *args: Any) -> TracedResult:
...
+            shape_env=torch.fx.experimental.symbolic_shapes.ShapeEnv(
+                tracked_fakes=[],
+            ),
         )
-        fake_args = tuple(
-            (
-                fake_mode.from_tensor(a, static_shapes=True)
+        with fake_mode.shape_env.ignore_fresh_unbacked_symbols():
+            fake_args = tuple(
+                _fakeify_input(fake_mode, a, input_name=f"minimal_fx_tracer_input_{i}")
```
SherlockNoMad: How would this name show up in graph? 
can we make it shorter, more readable?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -349,20 +367,33 @@ def _trace_with_args(state: Any, *args: Any) -> TracedResult:
...
             allow_non_fake_inputs=True,
-            shape_env=torch.fx.experimental.symbolic_shapes.ShapeEnv(),
+            shape_env=torch.fx.experimental.symbolic_shapes.ShapeEnv(
+                tracked_fakes=[],
+            ),
         )
-        fake_args = tuple(
-            (
-                fake_mode.from_tensor(a, static_shapes=True)
+        with fake_mode.shape_env.ignore_fresh_unbacked_symbols():
```
SherlockNoMad: What does this do when dynamic shape is not used? 

Our principle for the tracer is, if a feature it not used, it cannot break the main code path.

SherlockNoMad (changes requested):

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: This is a dummy string that comes with LocalSource from dynamo as it requires string to be initialized. I think it is only used in debugging/logging in torch.compile context. This won't change the graph.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: it won't do anything but to be safer i made this ctx manager bit more explicit (check if we have unbacked)

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -82,7 +89,8 @@ def _unwrap_subclass(t: torch.Tensor) -> tuple[list[torch.Tensor], SubclassMeta
 
 
 def _wrap_to_subclass(
-    plain_tensors: list[torch.Tensor], meta: SubclassMeta
+    plain_tensors: list[torch.Tensor],
```
SherlockNoMad, line 92: nit, some new line changes seems linter config difference.

SherlockNoMad (approved):
didn't read dynamic_shape.py, I assume you have the blessing from Laith.

PR #3049, opened 2026-04-21
[GraphTrainer] Add enable_cudagraph config flag

SherlockNoMad:
when profiling, cudagraph mess up the cuda streams. This flag is for the ease of having a human readable profile.

## Summary
- Add `compile.enable_cudagraph` config option (default `True`) to allow disabling the cudagraph pass via `--compile.no-enable_cudagraph`
- Rename `cudagraph_compatible` to `use_cudagraph` for clarity — it now combines the config flag with the graph compatibility check

## Test plan
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_passes.py -x` — all pass
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py -x -k "not test_eager_self_deterministic"` — 6/6 pass

### DIFF
--- a/torchtitan/experiments/graph_trainer/configs.py
+++ b/torchtitan/experiments/graph_trainer/configs.py
@@ -43,6 +43,9 @@ class GraphTrainerCompileConfig(CompileConfig):
     debug_graph_passes: bool = False
     """Log timing, op-count diffs, and before/after graphs for each pass to tlparse."""
 
+    enable_cudagraph: bool = True
+    """When False, skip the cudagraph pass even if the graph is compatible."""
+
     precompile_artifact_dir: str = ""
     """
     Directory for precompiled artifacts. Setting this enables precompile:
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -43,6 +43,9 @@ class GraphTrainerCompileConfig(CompileConfig):
     debug_graph_passes: bool = False
     """Log timing, op-count diffs, and before/after graphs for each pass to tlparse."""
 
+    enable_cudagraph: bool = True
```
aditvenk, line 46: Why would we want to disable CG if the graph is compatible? For ablations?

SherlockNoMad: when profiling, cudagraph mess up the streams. 
This flag is for the ease of having a human readable profile.

aditvenk (approved):

yiming0416 (approved):
i guess we need `--compile.enable_cudagraph `

PR #3050, opened 2026-04-21
[GraphTrainer] Refactor SAC pass: use module_fqn for layer boundaries

SherlockNoMad:
## Summary
- Remove `annotate_ac_regions` and `_AC_REGION_ID` — SAC now derives layer IDs by parsing `module_fqn` (e.g., `layers.3.attention` → layer 3), eliminating the redundant annotation step
- Remove the arbitrary mm alternating save/recompute heuristic from `apply_sac_pass` 
- Add layer boundary detection directly in `apply_sac_pass`: nodes whose output crosses a layer boundary (layer N → layer N+1) are forced to `MUST_SAVE`
- Non-layer nodes (tok_embeddings, norm, output) are tracked separately in logging for clarity

## Test plan
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_passes.py -x` — all pass
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py -x -k "not test_eager_self_deterministic"` — 6/6 pass

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -17,7 +17,6 @@ from torchtitan.config import CompileConfig
 from torchtitan.distributed import ParallelDims
 from torchtitan.tools.logging import logger
 
-_AC_REGION_ID = "ac_region_id"
 _MODULE_FQN = "module_fqn"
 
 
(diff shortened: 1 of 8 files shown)


tugsbayasgalan (approved):
Looks good! Is it an invariant that we always do checkpointing per layer? It seems so today but if we do end up changing that, there is a risk our AC pass logic can break (like inside a layer we could have two AC regions in which you need to mark the boundary values as MUST_SAVE)

SherlockNoMad:
> Looks good! Is it an invariant that we always do checkpointing per layer? It seems so today but if we do end up changing that, there is a risk our AC pass logic can break (like inside a layer we could have two AC regions in which you need to mark the boundary values as MUST_SAVE)

In eager, it's safety guard that output of every layer is saved, so that it doesn't run into the case where you rereun the entire forward to get an activation for the last layer (this happened to me before!). but in theory, we can do better than this. 

I don't think we need to rely on the concept of ac_region in graph. with graph, we can easily detect if a policy is "saving enough" to avoid excessive recompute, and warn user if so.

yiming0416 (approved):

PR #3057, opened 2026-04-22
[Don't review] [rl] MoE TP2EP

wwwjn:
Based on: #3070


torchtitan/experiments/rl/scripts/test_ep_numerics.py
wwwjn, line 1: Need to clean up test scripts

PR #3063, opened 2026-04-23
[data] Add weighted interleaved multi-source dataloader

slimfrkha:
## Summary

Adds `InterleavedHuggingFaceTextDataLoader` and `InterleavedChatDataLoader` to support training on multiple data sources simultaneously with configurable sampling weights.

## Changes

- `torchtitan/hf_datasets/utils/interleave.py`: new `InterleavedDataset`, a lightweight weighted sampler that interleaves multiple `IterableDataset` sources
- `torchtitan/hf_datasets/text_datasets.py`: adds `InterleavedHuggingFaceTextDataLoader`, `InterleavedChatDataLoader`
- `tests/`: new and updated tests for interleaving, checkpointing, and dataloader construction
- `docs/datasets.md`: updated with SFT, interleaved, and multimodal documentation

## Design choices

**`InterleavedDataset` sits above `HuggingFaceTextDataset` / `ChatDataset`, not below.**
Each source is a fully constructed `HuggingFaceTextDataset` or `ChatDataset`. `InterleavedDataset` interleaves already-packed `(input, label)` tensor sequences. This keeps tokenization, packing, shuffling, and epoch logic entirely inside each source and makes `InterleavedDataset` a pure weighted sampler (~100 lines). see discussion with @rakkit in #2385.

**Weights are sampling probabilities, not pass multipliers.**
A weight of `3.0` alongside `1.0` means source A is drawn ~75% of the time at each step. This is easy to reason about as a token mixture ratio and requires no knowledge of dataset sizes upfront.

**Epoch boundary is defined by first source exhaustion.**
When any source raises `StopIteration`, `InterleavedDataset` stops. Re-looping and shuffling are handled by the parent sources exactly as in the single-source case — no new concepts introduced. This is consistent with how `HuggingFaceTextDataset` and `ChatDataset` already define an epoch.

**All sources must share the same `infinite` setting**, enforced at config validation time (`__post_init__`). Mixed infinite/finite sources create ambiguous epoch semantics and are rejected with a clear error message.

**Checkpointing** saves the interleaver RNG state and delegates source state to each child's existing `state_dict` / `load_state_dict`, giving byte-identical resumption with no additional machinery.

---

**N.B**: some resume from checkpoints test now failing but will pass once #3008 is merged.

@tianyu-l  you can now close #2385.


rakkit:
thx @slimfrkha 

in this case, we might want to patch `snapshot_every_n_steps` set it to be `checkpoint_interval * gradient_accumulation_steps` (or 1 if there is only one dataset) for `ParallelAwareDataloader`

the reason is, each `HuggingFaceTextDataset` will maintain a $seq-len buffer, by default snapshot_every_n_steps=1 that dataloader worker(s) will sync theses buffers together with all states with main dataloader worker.  it is slow and cpu-bound.  so if we have lots of dataset, this dataloder state sync will be super slow and significantly slow down training.

slimfrkha:
> thx @slimfrkha
> 
> in this case, we might want to patch `snapshot_every_n_steps` set it to be `checkpoint_interval * gradient_accumulation_steps` (or 1 if there is only one dataset) for `ParallelAwareDataloader`
> 
> the reason is, each `HuggingFaceTextDataset` will maintain a $seq-len buffer, by default snapshot_every_n_steps=1 that dataloader worker(s) will sync theses buffers together with all states with main dataloader worker. it is slow and cpu-bound. so if we have lots of dataset, this dataloder state sync will be super slow and significantly slow down training.

Nice catch @rakkit ! just addressed.

torchtitan/trainer.py
```diff
@@ -239,6 +239,7 @@ def __init__(self, config: Config):
             tokenizer=self.tokenizer,
             seq_len=config.training.seq_len,
             local_batch_size=config.training.local_batch_size,
+            snapshot_every_n_steps=config.checkpoint.interval * self.gradient_accumulation_steps,
```
tianyu-l: what's this and where is it consumed?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -568,3 +643,78 @@ def __init__(
...
+@dataclass(kw_only=True, slots=True)
+class ChatDataSource(ChatDataLoader.Config):
+    """Represent one chat dataset source and its sampling weight"""
+
+    weight: float = 1
+    """Data Source sampling weight"""
+
+
+class InterleavedChatDataLoader(ParallelAwareDataloader):
+    """Configurable chat dataloader that wraps multiple ChatDataset."""
```
tianyu-l, line 712: Does it not make sense to mix multiple types of datasets, e.g. chat, txt, and even multimodal?

torchtitan/trainer.py
slimfrkha: pytorch `StatefulDataLoader`. parent class of `ParallelAwareDataloader`

From docstring
```
        snapshot_every_n_steps (int, optional): Defines how often the state is
            transferred from the dataloader workers to the dataloader. By default, it is set to ``1``, i.e., state is transferred every step. If the state is large, this value can be increased (and ideally set to the frequency of training checkpointing) to reduce the overhead of transferring state every step.
```

torchtitan/hf_datasets/text_datasets.py
slimfrkha: For simplicity, maybe we restrict to mixing same modalities only first ? My intent was first to introduce mixing per class (HuggingFaceTextDataset / ChatDataset). otherwise PR would be too long and complicated. we can iterate on top to make mixing more flexible.

---

In text modality, about mixing `HuggingFaceTextDataset` and `ChatDataset`, my thoughts below:
1. in principle, we can mix chat and pretraining data. This can be resolved by customizing the `sample_process` func for each source
2. `HuggingFaceTextDataset` and `ChatDataset` dont only represent chat vs txt data. they also manage packing and tokenization differently today.

In theory, we can make mixing chat and txt data possible. but the code needs first a full reorg to delete duplicated and repeated code and make it more modular.
FYI: once this PR is finalized, i will push a new pr to factor code of chat pretain and interleaved in a clean way. too much repeated code currently.

slimfrkha:
Rebased on main after #3008 merge so that tests pass.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -600,3 +675,78 @@ def __init__(
...
+            seed=config.seed,
+        )
+
+        dataloader_kwargs = {
+            "num_workers": config.num_workers,
+            "persistent_workers": config.persistent_workers,
+            "pin_memory": config.pin_memory,
+            "prefetch_factor": config.prefetch_factor,
+            "batch_size": local_batch_size,
+        }
```
rakkit, line 778: i think we need to add `snapshot_every_n_steps` here to actually use it

slimfrkha: Yep actually `snapshot_every_n_steps ` flows all the way until here and gets lost. 2 solutions possible:
1. add `snapshot_every_n_steps ` manually in `dataloader_kwargs`. it means we need to add it in the init signature as well
2. or pass `**kwargs` inside `dataloader_kwargs`. Notice that `kwargs` is actually not used in all dataloaders (except `ParallelAwareDataloader`). also this will expose all `StatefulDataLoader` init args to `build()`.

I am inclined to go with option 2.

rakkit: for (1) you mean add `snapshot_every_n_steps` in `HuggingFaceTextDataLoader.__init__` ?  i think it is ok to let user know what happend there, and optionally set it to `snapshot_every_n_steps: Optional[int] = 1` can match the default behaviors.    

up to you.  I dont have preference : -  )

slimfrkha: sure thing. should we delete `kwargs` from the init def in case of (1) ? i am just bugged about the silent error here: `snapshot_every_n_steps` (or any future StatefulDataLoader arg) passed in kwargs but not used.
@tianyu-l @joecummings any preference?

torchtitan/trainer.py
rakkit: @tianyu-l see comment [here](https://github.com/pytorch/torchtitan/pull/2385#issuecomment-3909030954) and [here](https://github.com/pytorch/torchtitan/pull/3063#issuecomment-4303116560)


here is log (throughput token-per-gpu-per-second) from our previous run. (defualt behaviour is snapshot_every_n_steps=1, i.e. sync every step)
image


after set `snapshot_every_n_steps=config.checkpoint.interval * self.gradient_accumulation_steps` :

image

the throughput is remains same as no snapshot sync, and only slow up a bit at (few) steps of checkpoint saving

rakkit:
out of curiosity, (maybe not in this pr). does HF's `InterleavedDataset` support to change data mix weights on fly? so that people can do multiple phase training (assuming  its same datasets just change mix).  in this case we may need to carefully deal with the InterleavedDataset's state for mix weights (like need some flag to load all datasets's state, but without load mix weights from state when we want to loading ckpt and switch mix)

torchtitan/trainer.py
felipemello1: Not for this PR, but this seems to be a perfect feature to be done async

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -283,6 +284,79 @@ def __init__(
         )
 
 
+@dataclass(kw_only=True, slots=True)
+class HFDataSource(HuggingFaceTextDataLoader.Config):
+    """Represent one dataset source and its sampling weight"""
+
+    weight: float = 1
+    """Data Source sampling weight"""
```
felipemello1, line 294: Is there an elegant way to avoid inheritance here? It makes it hard to read and understand the library over time

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -283,6 +284,79 @@ def __init__(
...
+
+
+class InterleavedHuggingFaceTextDataLoader(ParallelAwareDataloader):
```
felipemello1, line 297: Do we need this class?

Could we have:
a) Dataloader just needs an iterable
b) Datasets can be a chain of iterables, e.g.

ds = PackedIterable(InterleavedIterable([DatasetIterable]))
dataloader(ds)

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -600,3 +675,78 @@ def __init__(
             dp_world_size=dp_world_size,
             **dataloader_kwargs,
         )
+
+
+@dataclass(kw_only=True, slots=True)
+class ChatDataSource(ChatDataLoader.Config):
```
felipemello1, line 704: similar comment as before, is there an elegant way to avoid adding a new class?

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -600,3 +675,78 @@ def __init__(
...
+    """Represent one chat dataset source and its sampling weight"""
+
+    weight: float = 1
+    """Data Source sampling weight"""
+
+
+class InterleavedChatDataLoader(ParallelAwareDataloader):
```
felipemello1, line 711: similar as before, I am not sure why we would need a `InterleavedChatDataLoader` and `InterleavedHuggingFaceTextDataLoader`. We can probably just have `ParallelAwareDataloader` require an Iterable as input and *args, **kwargs that we pass to it, so its agnostic to the iterable type.

felipemello1 (commented):
Hey, thanks for the PR! A few points:

1. I left a few comments regarding adding new classes. It seems that we can reduce complexity there by having a single dataloader that can accept any iterable + kwargs. Do you think we can avoid adding the extra dataloader classes?
 
2. PR description says: "InterleavedDataset interleaves already-packed (input, label) tensor sequences. " --> Can you please clarify? When I sample from a dataset, am I getting 1 sample or N packed samples? In other words, are we doing: Packed(Interlevead(Datasets))), or are we doing Interleaved(Packed(datasets))?

3. When I proposed this in torchtune, I made dataset add metadata to the output sample. The idea is that we could use it to log i) distribution of sources; ii) other key data, e.g. prompt length, sequence_length, num_turns, img_size, etc. We could then just dump whatever keys were available into wandb.

I am not saying we should do it in this PR, but to verify correctness, we should try to log the data mix distribution. I am ok if we just hack it for now to prove it works, and later add proper logging.

4. The perf impact of 'snapshot_every_n_steps' is concerning. Not for this PR, but we should check if we could have this state sync be done in a non-blocking way.

5. Regarding a) datamix scheduling, b) datamix on-the-fly-changes, c) text+image mixing;

my intuition is that they would require a proper RFC/follow up

Thanks for taking us a step in this direction!

rakkit:
@felipemello1

actually my local branch more or less have what you asked. i have my own implementation (rather than HF's interleaved dataset).  but i dont have bandwith to clean up and open PR for it.

To  make this dataload, pack, mix, track and mix-schueudler we do
- HuggingFaceDataset only load data via HF api and tokenzier it 
- GreedyPackedDataset: pack multiple (short) sequences into a batch
- MixedDataset: sample one dataset.

that allows us to do free combination of MixedDataset (GreedyPackedDataset) or GreedyPackedDataset(MixedDataset)

and we inject doc/tokens log in MixedDataset.   ([here for dataset](https://github.com/SDLAML/torchtitan/blob/llm-0.4.0/torchtitan/hf_datasets/text_datasets.py))

(the major diff is decouple current torchtitan's HuggingFaceDataset into two parts for  load/tokenizer +  pack)


and we have [DataMixScheduler](https://github.com/SDLAML/torchtitan/blob/llm-0.4.0/torchtitan/components/data_mix_scheduler.py), bind with MixedDataset to reset mix weights.

in general saying its not easy to get a perfect datasets.

PS. another thing we observed is we need trade off between "having precisely expected data mix distribution" v.s. efficiency.  the actual sampled tokens/docs will always be  +- 0.5% around expected value.  to have exactly precisely expectd distribution we may needs lots of extra communication to set it

torchtitan/trainer.py
rakkit: my feeling is the design of `snapshot_every_n_steps` is kind of "anti-async` or saying its the tradeoff args to have dataloader worker running in async.  and to have this update in non-block way one might need to refactor   StatefulDataLoader like a lot

slimfrkha:
> Hey, thanks for the PR! A few points:
> 
> 1. I left a few comments regarding adding new classes. It seems that we can reduce complexity there by having a single dataloader that can accept any iterable + kwargs. Do you think we can avoid adding the extra dataloader classes?
> 2. PR description says: "InterleavedDataset interleaves already-packed (input, label) tensor sequences. " --> Can you please clarify? When I sample from a dataset, am I getting 1 sample or N packed samples? In other words, are we doing: Packed(Interlevead(Datasets))), or are we doing Interleaved(Packed(datasets))?
> 3. When I proposed this in torchtune, I made dataset add metadata to the output sample. The idea is that we could use it to log i) distribution of sources; ii) other key data, e.g. prompt length, sequence_length, num_turns, img_size, etc. We could then just dump whatever keys were available into wandb.
> 
> I am not saying we should do it in this PR, but to verify correctness, we should try to log the data mix distribution. I am ok if we just hack it for now to prove it works, and later add proper logging.
> 
> 4. The perf impact of 'snapshot_every_n_steps' is concerning. Not for this PR, but we should check if we could have this state sync be done in a non-blocking way.
> 5. Regarding a) datamix scheduling, b) datamix on-the-fly-changes, c) text+image mixing;
> 
> my intuition is that they would require a proper RFC/follow up
> 
> Thanks for taking us a step in this direction!

Thanks for the thorough review @felipemello1. my answers below

1. I am totally aware of code duplication. I just wanted to separate the scope into 2 PRs (1 PR for interleaving sources as a feature, 1 PR for refactoring code for better readability) , otherwise it would a very complicated PR. As I mentioned, in a previous comment, I am planning to push a new PR once we resolve this one. sneak peak of [my personal draft PR](https://github.com/slimfrkha/torchtitan/pull/1). How do you want to proceed about your 4 comments (which would be addressed in the second PR) ?

2. we are doing Interleaved(Packed(datasets)) otherwise weights would be wrong or you need sources to all yield samples of fixed length tokens. check discussion with @rakkit in #2385.

3. yes possible to add source idx in interleaved class. btw there is some basic tests to test source indexes in `test_interleaved_dataset.py`. but still pretty basic.

4. I honestly don't know about this specific one. @rakkit pinpointed

felipemello1:
> I am totally aware of code duplication. I just wanted to separate the scope into 2 PRs (1 PR for interleaving sources as a feature, 1 PR for refactoring code for better readability) , otherwise it would a very complicated PR

 i feel like this PR is not really heavy. A lot of it is unit test + readme changes + classes that would be deleted in the refactoring, IIUC.

I would vote for a single refactor PR. I would just suggest that maybe writing a small RFC to get alignment before writing the refactor might be more time efficient. Or maybe invert the order: refactor first, add interleaved later?

cc: @tianyu-l

felipemello1:
>2.  we are doing Interleaved(Packed(datasets)) otherwise weights would be wrong or you need sources to all yield samples of fixed length tokens

Sorry, i still don't follow. How would Interleaved(Packed()) be possible?
Lets say we have:
ds1: [t1,t2], [t3,t4,t5]
ds: [t6], [t7,t8]

Lets say that packing budget is 7 tokens.
I would expect to have something like:

a. sample from ds1:
samples: = [[t1,t2]]

b. sample from ds2
samples: = [[t1,t2], [t6]]

c. sample from ds1
samples: = [[t1,t2], [t6], [t3,t4,t5]]

stop and pack(samples)

output:
samples: [t1,t2, t6, t3,t4,t5]
doc_mask: [1,1, 2, 3,3,3,3]

---

if i sample an already packed sample, what would this look like?

ps: i also think that we should be doing on-the-fly packing if we arent already

rakkit:
@felipemello1 

Interleaved(Packed()) never mixed multiple datasource into one batch (long sequence),
"Packed" will return a whole "batch" (long sequence)

in your example, 
ds1: [t1,t2], [t3,t4,t5]
ds2: [t6], [t7,t8]

and given 7 token budget,

we have pack-1 for dataset 1 to be  { [t1,t2], [t3,t4,t5], [t1, t2]} (start repeat a bit)
and we have pack-2 for  dataset 2 to be  {  [t6], [t7,t8],  [t6], [t7,t8], [t6]} (repeat 2.3 epoch)

and samples/mix only tell us with a% to take whole pack-1 or b% to take pack-2 

so output will be either 
  { [t1,t2], [t3,t4,t5], [t1, t2]} 
  or
  {  [t6], [t7,t8],  [t6], [t7,t8], [t6]}

felipemello1:
I see, so the interleaved would be on the **bsz dimension**, not on the **sample dimension**, e.g.

interleaved_batch =
[
   [N samples from DS1],
   [M samples form DS2],
]

as opposed to:

[
  [ds1, ds2, ds1, ds2, ds1],
  [ds2, ds1, ds2, ds1, ds2],
]

is that right? Idk how i feel about it. Sounds like there would be many issues with misrepresenting the datamix on a sample level, but i guess it would be more accurate on a token level? For small bsz regime though, this sounds like a bad per-step bias.

torchtitan/hf_datasets/interleave.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import Any
+
+from torch.utils.data import IterableDataset
+
+
+class InterleavedDataset(IterableDataset):
+    """Randomly interleaves multiple IterableDatasets weighted by sampling probability.
+
+    At each step a source is drawn proportionally to its weight.  Iteration
+    stops as soon as any source is exhausted, defining a natural epoch
```
felipemello1: if i remember it correctly, HF had this option "on_first_exhausted" or "on_all_exhausted". Since you are masking the datasets as infinite, then it wouldn't make a difference. I think it is also fine to have it in a follow up, just making a comment for future prs.

rakkit:
there are three cases in total 

|                         | Mix data in sequence | Mix data in batch |
|-------------------------|----------------------|-------------------|
| Prob. over docs         | Case A               | x            |
| Prob. over tokens       | Case B               | Case C            |


Mix data in sequence means, one batch (long -sequnce) have data from difference source
Mix data in batch, one batch data comes from same source, and differnt batch comes from differnt source 

`pack[interleaved]` is  **Case A**: sample packed docs by probability.  **Case B**  is do-able if we estimated tokens per doc.

`interleaved[pack]` is **Case C**: 



pack[interleaved] don't have the problem of this `snapshot_every_n_steps`, but it only work "accurately"  on mix by prob of docs mode.  (unless one pre-process the data to make each document have same amount of tokens)

interleaved[pack] have this thing of `snapshot_every_n_steps`, but it can work "accurately"  on mix  by prob of tokens.  and this should be more or less what people expected.     (not sure for one) but in the case of trianing with docuement masking,  this can work without problem

felipemello1:
ty! I did some research on 3 production ready libraries, and all 3 did Interleaved[packed], so i think we are good to keep this pattern. Thanks for clarifying it.

I guess main open questions are:
1. If we should refactor before merging this PR, to avoid adding more classes than we need. My opinion is that we should. Wdyt?
2. Perf related to snapshot_every_n_steps.

I guess main blocker would be if @joecummings is going to land changes on how dataloading looks in Titan, or if we should proceed with this PR.

slimfrkha:
Summarizing raised points and giving my opinion on each:


| # | Feature | Sub-topic | My Opinion |
|---|---|---|---|
| 1 | **`snapshot_every_n_steps`** | How to pass to dataloader? (`dataloader_kwargs` / `kwargs`) | Pending decision |
| 1 | **`snapshot_every_n_steps`** | Performance implications | To be addressed in a separate PR |
| 2 | **Interleaving additional features** | Expose source index in yield | Straightforward to implement, should be included in this PR |
| 2 | **Interleaving additional features** | Additional stopping criteria: `on_first_exhausted` / `all_exhausted` | Straightforward to implement, should be included in this PR |
| 3 | **Dataset / dataloader refactor** | Should this happen before, within, or after this PR? | Out of scope for this PR. Should happen either before or after to keep the review focused. Existing draft proposal [here](https://github.com/slimfrkha/torchtitan/pull/1), open to revisiting after discussion. |
| 4 | **Support for multiple data source types** | — | Defer to a future PR / discussion |

slimfrkha:
Hey @tianyu-l @joecummings , just bumping this PR in case it got buried. happy to address any feedback or make changes if needed. Let me know if there's anything blocking the review!

felipemello1:
hey @slimfrkha , sorry. I chatted with Joe offline. @joecummings said that he still plans on adding a dataloader refactor soon. I think that we should wait for that and go from there. Joe, can you confirm if thats still the case?

felipemello1:
hey @slimfrkha , Joe went on PTO and didn't have time to address this PR. I will take a look at your refactor later today. Is it ready for review (https://github.com/slimfrkha/torchtitan/pull/1)? Depending on how it goes, i think that we should refactor first and add new features later, if thats ok with you, so we dont have more tech debt later on.

slimfrkha:
> hey @slimfrkha , Joe went on PTO and didn't have time to address this PR. I will take a look at your refactor later today. Is it ready for review ([slimfrkha#1](https://github.com/slimfrkha/torchtitan/pull/1))? Depending on how it goes, i think that we should refactor first and add new features later, if thats ok with you, so we dont have more tech debt later on.

Hey @felipemello1, 
Yes it is ready for review but it is built on top of this branch. I can push an additional commit to delete all the interleaved stuff (pretty isolated thanks to the refactor) and open an official PR on torchtitan. would that work for you ?

felipemello1:
@slimfrkha i want to avoid extra work for you. Let me take a look at it first, but thanks for offering!

slimfrkha:
> @slimfrkha i want to avoid extra work for you. Let me take a look at it first, but thanks for offering!

Sure. thanks for reviving this 🙏

felipemello1:
I did a first pass on the draft PR checking the base classes. I will get back to it tomorrow and check the implementation of the other classes.

felipemello1:
did another pass in the draft, ty!

slimfrkha:
Hi @felipemello1 

addressed the below comments:
* `snapshot_every_n_steps` pass through `kwargs` in `dataloader_kwargs`
* Expose `source_idx` in `yield`
*  Stopping strategy: `on_first_exhausted` / `all_exhausted`

The following points were not addressed due to future introduction of `grain` agreed on in https://github.com/slimfrkha/torchtitan/pull/1
* performance implication of `snapshot_every_n_steps`
* dataset / dataloader refactor
* Support for multiple data source types

N.B: I did not put an extra effort to try to refactor the several dataset/dataloader classes. I an aware of code duplication. not worth the effort since Torchtitan would move to `grain` soon.

felipemello1:
> Hi @felipemello1
> 
> addressed the below comments:
> 
> * `snapshot_every_n_steps` pass through `kwargs` in `dataloader_kwargs`
> * Expose `source_idx` in `yield`
> * Stopping strategy: `on_first_exhausted` / `all_exhausted`
> 
> The following points were not addressed due to future introduction of `grain` agreed on in [slimfrkha#1](https://github.com/slimfrkha/torchtitan/pull/1)
> 
> * performance implication of `snapshot_every_n_steps`
> * dataset / dataloader refactor
> * Support for multiple data source types
> 
> N.B: I did not put an extra effort to try to refactor the several dataset/dataloader classes. I an aware of code duplication. not worth the effort since Torchtitan would move to `grain` soon.

thank you! I couldn't get to it today, but i will prioritize it tomorrow.

torchtitan/hf_datasets/interleave.py
```diff
@@ -0,0 +1,133 @@
...
+
+from torch.utils.data import IterableDataset
+
+
+class InterleavedDataset(IterableDataset):
+    """Randomly interleaves multiple IterableDatasets weighted by sampling probability.
+
+    At each step a source is drawn proportionally to its weight.
+
+    ``stopping_strategy`` controls when iteration ends:
```
felipemello1: should we add proper "Args:" field?

torchtitan/hf_datasets/interleaved.py
```diff
@@ -0,0 +1,133 @@
...
+                f"Missing on: {missing}"
+            )
+
+        self._datasets = list(datasets)
+        self._probs = [w / total for w in weights]
+        self._rng = random.Random(seed)
+        self._stopping_strategy = stopping_strategy
+
+    @staticmethod
+    def _add_source_idx(sample: Any, source_idx: int) -> Any:
```
felipemello1: should we have a type for sample, even if generic, e.g. define at the top of the doc "SampleType = tuple[dict, str] | ..."

felipemello1: should we also add a small dummy example of what a sample would look like? Just by reading this, its unclear to me why its a tuple, and what sample[1] is, and why there isnt a sample[2]

e.g:
```
Example:
	sample = ({"prompt": "blablabla}, something)
	_add_source_idx(sample, 1)
```

torchtitan/hf_datasets/interleave.py
```diff
@@ -0,0 +1,133 @@
...
+        return {
+            "rng_state": self._rng.getstate(),
+            "sources": [ds.state_dict() for ds in self._datasets],
+        }
+
+    def load_state_dict(self, sd: dict[str, Any]) -> None:
+        """Restore from *sd*."""
+        self._rng.setstate(sd["rng_state"])
+        for ds, src_sd in zip(self._datasets, sd["sources"]):
+            ds.load_state_dict(src_sd)
```
felipemello1: i guess loading and saving based on order works, but seems a bit brittle. Can we save a keyed dict by source idx? 

nit: it would be even nicer if datasets had names instead of idx, but we can leave that for the refactor

torchtitan/hf_datasets/text_datasets.py
felipemello1: gosh, i really dislike this haha

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -283,6 +285,84 @@ def __init__(
...
+            "batch_size": local_batch_size,
+            **kwargs,
+        }
+
+        super().__init__(
+            ds,
+            dp_rank=dp_rank,
+            dp_world_size=dp_world_size,
+            **dataloader_kwargs,
+        )
```
felipemello1, line 369: sorry, stupid question, just sanity checking this:

this is a parallel aware dataloader, but the datasets also are...

Does this just work? Are we confident? could you ask claude to do some dummy test in a distributed setting, maybe 2dp, 2tp, standalone file, 2 datasets, where ds1 outputs 1,2,3,4 and ds2 outputs 10001, 10002, 10003, and we print(sample, self.dp_rank, self.tp_rank), and we sanity check that:

dp has unique samples, tp has repeated samples, things make sense, checkpointer works?

torchtitan/trainer.py
```diff
@@ -582,6 +585,9 @@ def post_dataloading_process(
         """
         inputs = input_dict["input"]
         extra_inputs = {k: v for k, v in input_dict.items() if k != "input"}
+        # in case of interleaved dataset
+        # TODO: log source_idx to track per-source sample distribution
+        source_idx = extra_inputs.pop("source_idx", None)
```
felipemello1: why do we need this line?

torchtitan/trainer.py
```diff
@@ -446,6 +435,20 @@ def __init__(self, config: Config):
...
+        self.tokenizer = config.tokenizer.build(tokenizer_path=config.hf_assets_path)
+
+        # build dataloader
+        self.dataloader = config.dataloader.build(
+            dp_world_size=batch_degree,
+            dp_rank=batch_rank,
+            tokenizer=self.tokenizer,
+            seq_len=config.training.seq_len,
+            local_batch_size=config.training.local_batch_size,
+            snapshot_every_n_steps=config.checkpoint.interval * self.gradient_accumulation_steps,
```
felipemello1: i didnt see `snapshot_every_n_steps` in any file. Was this removed?

felipemello1 (changes requested):
PR is clean, great implementation! I added a couple of nits that should be easy to address.

Then, do you think you could provide the sanity check test i asked in one of the comments?

i am happy to approve after that. Thanks for much for the effort in this PR and commenting on other PRs as well!

torchtitan/hf_datasets/text_datasets.py
slimfrkha: It should because I did not touch the sharding itself as interleaved is just a list of the old datasets. One of the reasons i went with this design is offloading all that complexity the datasets elements themselves which meant 0 code change for dp sharding, epoch management etc etc
Nevertheless here is a script and its result confirming it:
```python
"""
Distributed sanity check for InterleavedDataset.

Tests that with 2 DP × 2 TP:
  - DP ranks see disjoint sample sets (data is properly sharded)
  - TP ranks within the same DP group see identical samples
  - Checkpoint save/load produces the same tail on resume

Run with:
    torchrun --nproc_per_node=4 tests/distributed/test_interleaved_distributed.py

No GPUs required (uses gloo backend).
"""

from __future__ import annotations

import os
from copy import deepcopy

import torch
import torch.distributed as dist
from torch.utils.data import IterableDataset as TorchIterableDataset

from torchtitan.hf_datasets.interleave import InterleavedDataset

DP_SIZE = 2
TP_SIZE = 2
# Global rank layout:
#   rank 0 → dp=0, tp=0
#   rank 1 → dp=0, tp=1
#   rank 2 → dp=1, tp=0
#   rank 3 → dp=1, tp=1


class ShardedRangeDataset(TorchIterableDataset):
    """Yields a subset of *values* determined by dp_rank, mimicking split_dataset_by_node.

    dp=0 gets values at indices 0, 2, 4, …
    dp=1 gets values at indices 1, 3, 5, …
    """

    def __init__(self, values: list, dp_rank: int, dp_world_size: int) -> None:
        self._values = [v for i, v in enumerate(values) if i % dp_world_size == dp_rank]
        self._pos = 0

    def __iter__(self):
        while self._pos < len(self._values):
            yield self._values[self._pos]
            self._pos += 1
        self._pos = 0

    def state_dict(self):
        return {"pos": self._pos}

    def load_state_dict(self, sd):
        self._pos = sd["pos"]


def build_interleaved(dp_rank: int) -> InterleavedDataset:
    ds1 = ShardedRangeDataset([1, 2, 3, 4], dp_rank, DP_SIZE)
    ds2 = ShardedRangeDataset([10001, 10002, 10003], dp_rank, DP_SIZE)
    return InterleavedDataset([ds1, ds2], weights=[1.0, 1.0], seed=42)


def main():
    dist.init_process_group(backend="gloo")
    rank = dist.get_rank()
    world_size = dist.get_world_size()
    assert world_size == DP_SIZE * TP_SIZE, f"Expected {DP_SIZE * TP_SIZE} processes, got {world_size}"

    dp_rank = rank // TP_SIZE
    tp_rank = rank % TP_SIZE

    # --- Iteration check ---
    interleaved = build_interleaved(dp_rank)
    samples = list(interleaved)
    print(f"[rank={rank} dp={dp_rank} tp={tp_rank}] samples={samples}", flush=True)

    all_samples: list[list | None] = [None] * world_size
    dist.all_gather_object(all_samples, samples)

    if rank == 0:
        # TP invariant: ranks sharing the same dp_rank must produce identical output
        assert all_samples[0] == all_samples[1], (
            f"TP ranks (dp=0) diverged!\n  tp=0: {all_samples[0]}\n  tp=1: {all_samples[1]}"
        )
        assert all_samples[2] == all_samples[3], (
            f"TP ranks (dp=1) diverged!\n  tp=0: {all_samples[2]}\n  tp=1: {all_samples[3]}"
        )
        print("[OK] TP ranks within the same DP group see identical samples")

        # DP invariant: different dp_ranks must produce disjoint sets
        dp0 = set(all_samples[0])  # type: ignore[arg-type]
        dp1 = set(all_samples[2])  # type: ignore[arg-type]
        assert dp0.isdisjoint(dp1), (
            f"DP ranks share samples!\n  dp=0: {dp0}\n  dp=1: {dp1}"
        )
        print(f"[OK] DP ranks see disjoint samples — dp=0: {dp0}, dp=1: {dp1}")

    dist.barrier()

    # --- Checkpoint check ---
    interleaved2 = build_interleaved(dp_rank)
    it = iter(interleaved2)
    n_consume = max(1, len(samples) // 2)
    for _ in range(n_consume):
        next(it)
    ckpt = deepcopy(interleaved2.state_dict())
    tail_original = list(it)

    interleaved3 = build_interleaved(dp_rank)
    interleaved3.load_state_dict(ckpt)
    tail_resumed = list(interleaved3)

    assert tail_original == tail_resumed, (
        f"[rank={rank} dp={dp_rank} tp={tp_rank}] Checkpoint resume mismatch!\n"
        f"  original: {tail_original}\n  resumed:  {tail_resumed}"
    )
    print(f"[rank={rank} dp={dp_rank} tp={tp_rank}] [OK] Checkpoint resume matches")

    dist.barrier()
    if rank == 0:
        print("\nAll checks passed.")
    dist.destroy_process_group()


if __name__ == "__main__":
    main()
```
and bash prints
```bash
[rank=2 dp=1 tp=0] samples=[10002, 2, 4]
[rank=0 dp=0 tp=0] samples=[10001, 1, 3]
[rank=1 dp=0 tp=1] samples=[10001, 1, 3]
[rank=3 dp=1 tp=1] samples=[10002, 2, 4]
[OK] TP ranks within the same DP group see identical samples
[OK] DP ranks see disjoint samples — dp=0: {10001, 3, 1}, dp=1: {10002, 2, 4}
[rank=2 dp=1 tp=0] [OK] Checkpoint resume matches
[rank=3 dp=1 tp=1] [OK] Checkpoint resume matches
[rank=1 dp=0 tp=1] [OK] Checkpoint resume matches
[rank=0 dp=0 tp=0] [OK] Checkpoint resume matches

All checks passed.
```

torchtitan/trainer.py
slimfrkha: extra_inputs gets unpacked directly into the model's forward() at line 704 (and into pp_schedule.step() in the PP branch at line 674). The model's forward doesn't accept source_idx, so leaving it in would raise a TypeError.
exatcly same behaviour as positions.
source_idx is just popped and discarded for now (pending the TODO to log it).

torchtitan/trainer.py
slimfrkha: `snapshot_every_n_steps` is an arg from `StatefulDataLoader` (parent class of `ParallelAwareDataloader`). It is to `StatefulDataLoader` through the `kwargs` in `dataloader_kwargs`.
(passing kwargs in dataloader_kwargs exposes for free all StatefulDataLoader args).

torchtitan/hf_datasets/interleaved.py
slimfrkha: maybe overkill with code. I just listed the 3 cases of sample format in the docstring.

torchtitan/hf_datasets/interleave.py
slimfrkha: made class docstring shorter and moved the args under init.

torchtitan/hf_datasets/interleave.py
slimfrkha: fixed.
about dataset names: today HFDataset and ChatDataset dont use the same convention for dataset names.
So not worth given the future refactor.

torchtitan/hf_datasets/text_datasets.py
slimfrkha: I know it is ugly. believe me i tried to find a way to not do it xD
It is 2 lines duplicated, no clean way to remove.
- either we create a mixin class that defines weight field only
- or we make the weights not pinned to source but a list in the class Config (below seed and stopping_strategy) but you lose the expressivity of defining the weight at the source level.

torchtitan/trainer.py
felipemello1: i see. Can you please add a line explaining that?

felipemello1 (approved):
LGTM! Great work

just 1 nit about adding a comment to the trainer

@tianyu-l , feel free to merge

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -273,6 +274,7 @@ def __init__(
             "pin_memory": config.pin_memory,
             "prefetch_factor": config.prefetch_factor,
             "batch_size": local_batch_size,
+            **kwargs,
```
tianyu-l: let's use `snapshot_every_n_steps` explicitly here

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -283,6 +285,86 @@ def __init__(
...
+            stopping_strategy=config.stopping_strategy,
+        )
+
+        dataloader_kwargs = {
+            "num_workers": config.num_workers,
+            "persistent_workers": config.persistent_workers,
+            "pin_memory": config.pin_memory,
+            "prefetch_factor": config.prefetch_factor,
+            "batch_size": local_batch_size,
+            **kwargs,
```
tianyu-l: similar

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -598,6 +681,7 @@ def __init__(
             "pin_memory": config.pin_memory,
             "prefetch_factor": config.prefetch_factor,
             "batch_size": local_batch_size,
+            **kwargs,
```
tianyu-l: similar

torchtitan/trainer.py
```diff
@@ -447,6 +436,21 @@ def __init__(self, config: Config):
...
+
+        # build dataloader
+        self.dataloader = config.dataloader.build(
+            dp_world_size=batch_degree,
+            dp_rank=batch_rank,
+            tokenizer=self.tokenizer,
+            seq_len=config.training.seq_len,
+            local_batch_size=config.training.local_batch_size,
+            snapshot_every_n_steps=config.checkpoint.interval
+            * self.gradient_accumulation_steps,
```
tianyu-l: if `config.checkpoint.enable=False`, can we set this to infinite?

torchtitan/trainer.py
```diff
@@ -585,6 +589,9 @@ def post_dataloading_process(
         """
         inputs = input_dict["input"]
         extra_inputs = {k: v for k, v in input_dict.items() if k != "input"}
+        # in case of interleaved dataset
+        # TODO: log source_idx to track per-source sample distribution
+        source_idx = extra_inputs.pop("source_idx", None)
```
tianyu-l: Let's not add it into `extra_inputs`, especially as it's not consumed at all.

`Trainer` should not know such details. Ideally, logging could be done within dataloader.

torchtitan/hf_datasets/interleaved.py
tianyu-l, line 1: maybe rename to interleaved.py

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -283,6 +285,86 @@ def __init__(
...
+
+        super().__init__(
+            ds,
+            dp_rank=dp_rank,
+            dp_world_size=dp_world_size,
+            **dataloader_kwargs,
+        )
+
+
 class ChatDataset(IterableDataset, Stateful):
```
tianyu-l, line 372: Now that we have interleaved. Does it make sense to put this into `hf_datasets/chat.py` and rename the current file to `hf_datasets/text.py`?

torchtitan/hf_datasets/text_datasets.py
tianyu-l: I'm assuming you discussed this somewhere else and had a plan.

torchtitan/hf_datasets/interleaved.py
```diff
@@ -0,0 +1,140 @@
...
```
tianyu-l: could logging be done right here with https://github.com/pytorch/torchtitan/tree/main/torchtitan/observability/structured_logger

torchtitan/hf_datasets/interleaved.py
```diff
@@ -0,0 +1,140 @@
...
+        return sample
+
+    def __iter__(self) -> Iterator:
+        """Yield weighted-interleaved samples across all sources."""
+        indices = list(range(len(self._datasets)))
+        iterators = [iter(ds) for ds in self._datasets]
+        exhausted = [False] * len(self._datasets)
+
+        while True:
+            idx = self._rng.choices(indices, weights=self._probs, k=1)[0]
```
tianyu-l, line 116: please add a comment that output from each source is already packed, achieving
> This makes weights easy to reason about as a token mixture ratio: if source A has weight 3 and source B has weight 1, roughly 75 % of training tokens will come from A and 25 % from B, regardless of the absolute dataset sizes.

docs/datasets.md
```diff
@@ -30,47 +31,146 @@ def load_wikipedia_dataset(dataset_path: str, **kwargs):
...
+            infinite=True,
+        ),
+    ],
+    seed=42,
+),
+```
+
+### Weight semantics
+
+Weights are **sampling probabilities**, normalised internally. A weight of `3.0` alongside `1.0` means the first source is drawn three times as often on average — it does not mean the source is iterated three times per epoch. The epoch boundary is defined by whichever source exhausts first.
```
tianyu-l, line 158: Mention that when each source is drawn, it returns a packed sample, potentially consisting multiple data points in the source.

torchtitan/hf_datasets/interleaved.py
```diff
@@ -0,0 +1,140 @@
...
+            try:
+                yield self._add_source_idx(next(iterators[idx]), idx)
+            except StopIteration:
+                if self._stopping_strategy == "on_first_exhausted":
+                    return
```
tianyu-l: is this having the same effect as `break` here? https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/text_datasets.py#L168

If so, let's use the same syntax for consistency.

torchtitan/hf_datasets/interleaved.py
```diff
@@ -0,0 +1,140 @@
...
+                exhausted[idx] = True
+                if all(exhausted):
+                    return
+                iterators[idx] = iter(self._datasets[idx])
```
tianyu-l, line 134: trying to understand what's going on
- `StopIteration` is caught because all source datasets have set `infinite=False`
- If not all exhausted, but some source is exhausted, the exhausted dataset will reenter the sampling as if it's completely newly added like the first time. In particular, there's no shuffling-across-epoch behavior https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/text_datasets.py#L169

Is it correct? Are they (different shuffling strategy between infinite vs. multi-source finite) expected?

torchtitan/hf_datasets/interleaved.py
felipemello1: i dont think that we would want this on the trace output. This would need to be all_reduced, turned into a metric and passed to the MetricsProcessor, something like:

{"dataset_a": sum_samples_seen, "dataset_b", sum_samples_seen}

It is still useful for debugging/unit tests, but logging to metrics would be ideal. Perhaps we could add as a TODO for when we add the distributed metrics?

tianyu-l: Sure. I'd prefer we have the TODO in dataloader for now than in trainer.

torchtitan/hf_datasets/text_datasets.py
slimfrkha: yes it is in https://github.com/slimfrkha/torchtitan/pull/1
the tldr is that the effort not worth it to refactor the current code as torchtitan will move on to grain soon.

torchtitan/trainer.py
slimfrkha: yes good ctach. 
it's snapshot_every_n_steps=None.

torchtitan/hf_datasets/interleaved.py
slimfrkha: okay so there are 3 constraints that don't align. we need to drop one:

- constraint 1: source_idx stays on the yield (debug/unit tests) → if we ignore it , solution= drop _add_source_idx and the pop entirely; __iter__ yields the sample untouched with the TODO sitting next to the already-in-scope 'idx'
- constraint 2: the trainer knows nothing about source_idx. →if we ignore it, solution= current code
- constraint 3: no bloat code that exists only to relocate the pop (this is my constraint) → if we ignore it , solution= keep source_idx on the yield and add a collate_fn on the interleaved dataloaders that strips it (and carries the TODO) before the batch reaches the trainer.

what do we do?

torchtitan/hf_datasets/text_datasets.py
slimfrkha: it can make sense but in a different PR. (https://github.com/slimfrkha/torchtitan/pull/1)
I would argue also that it is not worth the effort as torchtitan is moving to grain soon.

torchtitan/hf_datasets/interleaved.py
slimfrkha: Actually 'InterleavedDataset' does weighted interleaving of arbitrary datasets (packed / not packed). I would add this comment for 'InterleavedHuggingFaceTextDataLoader' and 'InterleavedChatDataLoader'.

torchtitan/hf_datasets/interleaved.py
slimfrkha: Good catch, you're right: with all_exhausted the interleaver re-enters exhausted sources via a bare iter(ds), which skips the per-epoch reshuffle, and for map-style sources it is even worse: it replays nothing since _sample_idx never reset. the stopping strategy was added later on and this is definitely an edge case.

Fix (pushed): I extracted the reset+reshuffle into an explicit reloop() on both datasets and use reloop method for infinite=true and inside interleaver.

torchtitan/hf_datasets/interleaved.py
tianyu-l: I'm fine as long as trainer.py code is not touched.

slimfrkha: sure just dropped constraint 1 and added a comment.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -255,6 +256,7 @@ def __init__(
         tokenizer: BaseTokenizer,
         seq_len: int,
         local_batch_size: int,
+        snapshot_every_n_steps: int | None = 1,
```
tianyu-l, line 259: Trying to be explicit, is it the same as
```suggestion
        snapshot_every_n_steps: int = 1,
```

Same for other occurrences.

tianyu-l (approved):
Looks wonderful, thank you!

Had one more nit comment

torchtitan/hf_datasets/text_datasets.py
slimfrkha: no it is not. when checkpointing is disabled, snapshot_every_n_steps=None (or as you said infinity) to disable snapshoting.
it is the same signature as StatefulDataloder:
```python
snapshot_every_n_steps: Optional[int] = 1
```

slimfrkha:
race-win

PR #3066, opened 2026-04-23
[rl] Remove call to get_model_state_dict in push_model_state_dict

daniellepintz:
The trainer refactor (#2985) changed push_model_state_dict to call get_model_state_dict() instead of `model.state_dict()`. While the end result of these calls are the same (returns sharded state_dict with DTensors), calling `get_model_state_dict` adds some overhead.

Average push time over 10 steps for Qwen3 0.6B:
with `get_model_state_dict()`: 0.0170s
with `model.state_dict()`: 0.008s

Revert to previous way of calling self.model.state_dict() directly.

### DIFF
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -410,14 +410,11 @@ class PolicyTrainer(Actor, Configurable):
         means "skip StorageVolumes and let the destination read directly
         from the source's GPU memory".
 
-        Uses get_model_state_dict() to unshard FSDP state before pushing.
         """
         from monarch.rdma import is_rdma_available
 
-        full_sd = get_model_state_dict(self.model)
-
         await ts.put_state_dict(
-            full_sd,
+            self.model.state_dict(),
             "model_state_dict",
             direct_rdma=is_rdma_available(),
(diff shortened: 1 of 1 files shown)


joecummings (approved):
Great catch thank you! Would you mind opening a GitHub Issue as well to add in timing tests for our RL work? That would've caught this immediately

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -410,14 +410,11 @@ async def push_model_state_dict(self) -> None:
         means "skip StorageVolumes and let the destination read directly
         from the source's GPU memory".
 
-        Uses get_model_state_dict() to unshard FSDP state before pushing.
```
wwwjn, line 413: Now why we don't need unshard FSDP state?

daniellepintz: actually, even before this PR, we weren't unsharding the FSDP state; the comment was wrong. by default `get_model_state_dict()` returns the **sharded** state

we don't need to unshard FSDP state because torchstore handles sharded DTensors. each trainer rank can put its own shard of a tensor to TorchStore, and when the generator pulls, it will handle the resharding if it needs to

Issue #2515, opened 2026-03-06
Support LoRA at TorchTitan


mori360:
This stack adds parameter-efficient fine-tuning (PEFT) to torchtitan through the existing ModelConverter protocol, enabling LoRA,  QLoRA (NF4-quantized base weights), and QAT (quantization-aware training) — all composable with each other and with existing parallelism strategies (FSDP, TP, PP, CP).

### Motivation

Full fine-tuning of large models is expensive. LoRA reduces trainable parameters by orders of magnitude by injecting low-rank adapters into frozen base weights. QLoRA further reduces memory by quantizing those frozen base weights to NF4. QAT allows training with simulated quantization so the model learns to compensate for quantization error at deployment time. All three are widely used in practice and compose naturally.

  What's included

  PR 1 — LoRA adapters (https://github.com/pytorch/torchtitan/pull/2484)

  Adds LoRAConverter in torchtitan/components/lora.py, a ModelConverter that:
  - Dynamically subclasses any nn.Linear (including FakeQuantizedLinear, etc.) to inject lora_a / lora_b adapter parameters
  - Freezes all base model parameters (requires_grad_(False)) and keeps only LoRA params trainable
  - Patches init_weights() to properly reinitialize adapters (Kaiming for A, zeros for B)
  - Uses a class cache so the same parent class is only subclassed once
  - Registered config: llama3_debugmodel_lora

  PR 2 — LoRA-aware checkpointing (https://github.com/pytorch/torchtitan/pull/2485)

  Extends CheckpointManager and ModelWrapper in torchtitan/components/checkpoint.py to support:
  - Adapter-only saving: when save_adapter_only=True (default), only LoRA weights are saved, not the full model — checkpoints stay
  small
  - Partial loading: uses DefaultLoadPlanner(allow_partial_load=True) so base model can be loaded from HF and adapters from a separate
   DCP checkpoint
  - Multi-source loading: new additional_load_paths config for loading from multiple checkpoints sequentially (e.g., base model from
  HF + adapters from a previous run)
  - Converter key filtering: ModelWrapper delegates to model.converter_key_filter() to partition state dicts without knowing LoRA
  internals

  PR 3 — QLoRA (NF4 quantization) (https://github.com/pytorch/torchtitan/pull/2487)

  Extends LoRAConverter with quantize_base="nf4" config:
  - Registers a one-shot forward_pre_hook that quantizes base weights to NF4 (via torchao to_nf4) after checkpoint load but before the
[...]

PR #3068, opened 2026-04-23
Add tests for `torchtitan/components/checkpoint.py`

Andrei-Aksionov:
Hey there 👋 

I've been working on refactoring (or more accurately, some "opinionated polishing") of the logic in `torchtitan/components/checkpoint.py`. 

*In my opinion, the most effective approach to gaining a comprehensive understanding of a codebase is to actively engage in its improvement, even if the primary focus is on enhancing readability and structure.*

To make the upcoming refactor easier to review and to provide "peace of mind," I’ve added missing tests to `tests/unittests/test_checkpoint.py`. The goal is to establish a robust baseline: if the upcomming refactored code passes these new tests, we can be confident that the core functionality remains intact.

### **Coverage Statistics**

By running:
```bash
pytest --cov=torchtitan.components.checkpoint tests/unit_tests/test_checkpoint.py
```

**This PR:**
| Name | Stmts | Miss | Cover |
| :--- | :--- | :--- | :--- |
| torchtitan/components/checkpoint.py | 363 | 10 | **97%** |

**vs. Current Main:**
| Name | Stmts | Miss | Cover |
| :--- | :--- | :--- | :--- |
| torchtitan/components/checkpoint.py | 363 | 47 | **87%** |

While a 10 percent point increase might seem modest given the volume of test code added, it covers the most complex logic branches that were previously untested. 
(To be frank, if I had realized coverage was already at 87%, I might not have started this 🫠, but since the tests are now written, they provide a much stronger safety net for the project.)

### **Implementation Details**
* **Preservation:** I have kept the existing tests exactly as they were to ensure no regressions in the testing logic itself (again, it is simpler to review: there is no need to verify that the logic in existing tests is preserved.). 
* **Refactoring:** I only extracted common `setUp` and `tearDown` logic to keep the test file maintainable.
* **Add-ons:** These new tests should be viewed as non-breaking additions that focus on high-level orchestration, async staging, and checkpoint purging logic.

This PR serves as the foundation for the follow-up refactoring of the checkpoint manager.


fegin:
What refactor are you planing to do? We also have some refactor plan internally. Would love to learn the request.

Andrei-Aksionov:
Maybe “refactor” isn’t quite the right term for what I’m trying to do.

The core idea will remain the same. My goal is to simplify the code by organizing/grouping code inside methods, adding docstrings where they’re needed, updating existing comments, adding my own, ...

Some of the logic I understood only after I read a couple of blogpost (like [this](https://pytorch.org/blog/distributed-checkpoint-efficient-checkpointing-in-large-scale-jobs/), [this](https://pytorch.org/blog/6x-faster-async-checkpointing/) and [this](https://pytorch.org/blog/reducing-checkpointing-times/)) and this [discussion](https://discuss.pytorch.org/t/distributed-w-torchtitan-optimizing-checkpointing-efficiency-with-pytorch-dcp/211250) (great writing, btw 🚀 )
I’ve tried to make the code easier to understand, even if you haven’t had a chance to read those articles. 
It's not like the code is bad (quite contrary), just the topic isn't easy. 

----

> What refactor are you planing to do?

Well, it's actually almost done.
I just saw how many lines of code are changed and wanted to make reviewing easier by first introducing these tests.

If you want you can take a sneak peek at the code in my fork.
- The [file](https://github.com/Andrei-Aksionov/torchtitan/blob/components_checkpoint_refactor/torchtitan/components/checkpoint.py)
- The [diff](https://github.com/pytorch/torchtitan/compare/main...Andrei-Aksionov:torchtitan:components_checkpoint_refactor?expand=1)

----

> We also have some refactor plan internally. Would love to learn the request.

And I would love to learn what are you planning to do ☺️.

Would like to introduce changes that are more challenging to implement.

fegin:
@Andrei-Aksionov 
We are adding LoRA support which makes checkpointing more challenging, there will be more branches and state_dict transformation. We would like to consolidate the state_dict transformation and how a state_dict being saved and loaded.

@mori360 
As we discussed before, the very first step to do checkpointing refactor is to add tests, like this PR to cover more cases. However, I think one key test we need to add is the correctness, whether that being an unittest or an integration test. Please help review this PR.

Andrei-Aksionov:
> We are adding LoRA support which makes checkpointing more challenging, there will be more branches and state_dict transformation. We would like to consolidate the state_dict transformation and how a state_dict being saved and loaded.

Oh, I see.
- For loading: load base model, load LoRA adapter, merge them.
- For saving: 
    - During training: save full checkpoint
    - Final save: unmerge adapter, save model and adapter separately.

I saw in #2515 that you guys are working on LoRA, but haven't yet looked closely into it.
After I finish with this - I'll help reviewing code that is already written by @mori360. 

> However, I think one key test we need to add is the correctness, whether that being an unittest or an integration test.

I think unittests are mostly done.
These tests check the internal logic of the checkpointer.

To really validate the correctness we need integration tests. The ones that test all three async modes, on a single and on mutli-GPU machines. So we can be sure that:
- `dcp.load/save` worked correctly with the provided parameters
-  stager, returned futures, all handled without issues

Check `loading -> saving -> loading` scenario.

I can do this. Well, mostly. I don't have access to multi-GPU machine. 
So what I can do is to write such tests on a single-GPU one, simulate a multi-GPU scenario with spawning multiple processes. But with the rest I might need a hand.

-----

If there are no objections (after @mori360 checked the PR), I would like to:
1. Merge this PR
2. Work on integration tests
3. Merge the future PR with ~refactored~ prettyfied code
4. Try to help with LoRA

mori360:
Thanks for adding test coverage for the checkpoint manager! A few concerns about overlap with existing tests — I think we can make this PR more focused by trimming the redundant parts.
Several new test classes re-test logic that's already covered end-to-end in TestCheckpointManager:                                                                                                                                                                              
  - TestShouldSave — _should_save() interval/first-step/last-step/load-only logic is already tested by test_interval_respects_interval, test_enable_first_step_checkpoint, and test_load_only_prevents_saving
  - TestAsyncWait — async future wait/clear is covered by test_async_save_calls_async_wait                                                  
  - TestMaybeWaitForStaging — staging flag behavior is covered by test_async_save_with_pinned_mem_sets_staging_flag                         
  - TestShouldPurge / TestPurgeStaleCheckpoints — purge behavior is covered by test_save_and_purge_keeps_last_k_checkpoints and test_nonzero_rank_does_not_purge_or_save                                                                                                  
  - TestFindLoadStep — step discovery is covered by test_load_finds_latest_and_calls_dcp_load                                               
  - TestSave / TestLoad orchestration — largely mirrors what the existing integration-style tests already exercise 
  
Also I have tests at https://github.com/pytorch/torchtitan/pull/2954, they verify actual key transformations through the full orchestration path.

Andrei-Aksionov:
Hello @mori360 

Thank you for reviewing this PR.

> A few concerns about overlap with existing tests.

Oh yes, let's discuss it.

----

> TestShouldSave — _should_save() interval/first-step/last-step/load-only logic is already tested by test_interval_respects_interval, test_enable_first_step_checkpoint, and test_load_only_prevents_saving

Regarding the overlap, I’d like to clarify the intent. 
I don't see new tests as replacements for the existing suite. I believe both styles can and should co-exist. While they cover the same logic, they do so at different levels of granularity.

To illustrate my point, let’s look at `test_interval_respects_interval`:
https://github.com/pytorch/torchtitan/blob/ef8e2820e714471657c484008a62027ed91da15f/tests/unit_tests/test_checkpoint.py#L313-L344

Currently, this test verifies `_should_save` indirectly by checking how many times the `torch.distributed.save` mock is called. The execution flow looks like this:
`self.save` -> `self._should_save` -> `self.dcp_save` -> `torch.distributed.checkpoint.save` (mocked)

Because the assertion happens at the very end of this chain, a failure requires you to untangle the entire stack to find the culprit. It confirms *that* something went wrong, but not necessarily *where*.

In contrast, my proposed tests target `self._should_save` directly. If the logic for intervals or step-counts breaks, these tests will "scream" exactly which method is failing. While they cover the same logic as the integration tests, they provide a much higher resolution of error reporting. That is why I don’t see them as a replacement, but as a complementary layer that makes debugging significantly faster.

While the root cause of a failure might be immediately obvious to you, given your deep familiarity with this codebase, it’s often much harder for a newcomer or an external contributor to diagnose. For someone still learning the architecture, having tests that are "closer to true unit nature" provides a safety net. It transforms a vague integration failure into an explicit pointer to a specific method, which is incredibly helpful for maintaining the project as it grows beyond its original authors.

---

> Also I have tests at https://github.com/pytorch/torchtitan/pull/2954, they verify actual key transformations through the full orchestration path.

Since both of our approaches rely on mocking functions and methods, the core objective remains the same: ensuring that the correct state dicts are being passed around and that the overall order of execution is preserved. Whether we accomplish that by mocking return values (as I did in my tests), or by replacing the function body with a `side_effect` as you’ve done, the underlying verification logic is consistent.

That said, I have to admit I really like your approach. Using `side_effect` to manually track the call stack is a great touch; it allows you to cleanly and explicitly assert the execution sequence with:

```python
self.assertEqual(call_order, ["to_hf", "dcp_load", "from_hf"])
```
It’s a very elegant way to handle orchestration testing. Nice job on that.

----

I am quite new to this repo and still getting up to speed with the existing architecture and patterns. 
If you feel the current integration-style coverage is sufficient and that these additional unit tests are excessive (at least the ones that overlap), I am happy to remove them.
However, I believe they add significant value for long-term maintainability and easier debugging.

PR #3070, opened 2026-04-23
[rl] Removed patched qwen3 parallel plan to core torchtitan Qwen3 parallel plan

wwwjn:
As titled. 

We need to skip FSDP because FSDP's forward hooks (unshard/reshard) use in-place operations that conflict with torch.inference_mode(). vLLM runs the model forward under @torch.inference_mode() during KV cache profiling and generation. 

Also inference don't need mixed precision (now) because all computation / weights are in bf16

### DIFF
--- a/torchtitan/experiments/graph_trainer/precompile_main.py
+++ b/torchtitan/experiments/graph_trainer/precompile_main.py
@@ -43,7 +43,6 @@ import torch.distributed as dist
 
 from torchtitan.config import ConfigManager, TORCH_DTYPE_MAP
 from torchtitan.distributed import ParallelDims
-from torchtitan.models.common.decoder import Decoder
 from torchtitan.experiments.graph_trainer.common_utils import (
     apply_graph_ac,
     parallelize_inputs,
@@ -66,6 +65,7 @@ from torchtitan.experiments.graph_trainer.precompile import (
     _FX_TRACE_ARTIFACT_KEY,
 )
 from torchtitan.experiments.graph_trainer.storage import DiskStorageAdapter
+from torchtitan.models.common.decoder import Decoder
 from torchtitan.tools import utils
 from torchtitan.tools.logging import logger
 
--- a/torchtitan/experiments/graph_trainer/precompile_main.py
+++ b/torchtitan/experiments/graph_trainer/precompile_main.py
@@ -43,6 +43,7 @@ import torch.distributed as dist
 
 from torchtitan.config import ConfigManager, TORCH_DTYPE_MAP
 from torchtitan.distributed import ParallelDims
+from torchtitan.models.common.decoder import Decoder
 from torchtitan.experiments.graph_trainer.common_utils import (
     apply_graph_ac,
     parallelize_inputs,
@@ -65,7 +66,6 @@ from torchtitan.experiments.graph_trainer.precompile import (
     _FX_TRACE_ARTIFACT_KEY,
 )
 from torchtitan.experiments.graph_trainer.storage import DiskStorageAdapter
-from torchtitan.models.common.decoder import Decoder
 from torchtitan.tools import utils
 from torchtitan.tools.logging import logger
 
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -147,7 +147,7 @@ class VLLMGenerator(Actor, Configurable):
             # Generator only supports TP. vLLM handles its own parallelism
             # and we only apply TP via the core parallelize function.
             p = self.parallelism
-            if p.data_parallel_shard_degree not in (1, -1):
+            if p.data_parallel_shard_degree != 1:
                 raise ValueError(
                     f"Generator does not support data parallel sharding, "
                     f"got dp_shard={p.data_parallel_shard_degree}"
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -147,11 +147,6 @@ class VLLMGenerator(Actor, Configurable):
             # Generator only supports TP. vLLM handles its own parallelism
             # and we only apply TP via the core parallelize function.
             p = self.parallelism
-            if p.data_parallel_shard_degree != 1:
-                raise ValueError(
-                    f"Generator does not support data parallel sharding, "
-                    f"got dp_shard={p.data_parallel_shard_degree}"
-                )
             if p.data_parallel_replicate_degree != 1:
                 raise ValueError(
                     f"Generator does not support data parallel replication, "
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -144,14 +144,34 @@ class VLLMGenerator(Actor, Configurable):
         """Debug and determinism settings."""
 
         def __post_init__(self):
-            assert self.parallelism.data_parallel_shard_degree in (1, -1), (
-                f"Generator does not support data parallel sharding, "
-                f"got dp_shard={self.parallelism.data_parallel_shard_degree}"
-            )
-            assert self.parallelism.data_parallel_replicate_degree == 1, (
-                f"Generator does not support data parallel replication, "
-                f"got dp_replicate={self.parallelism.data_parallel_replicate_degree}"
-            )
+            # Generator only supports TP. vLLM handles its own parallelism
+            # and we only apply TP via the core parallelize function.
+            p = self.parallelism
+            if p.data_parallel_shard_degree not in (1, -1):
+                raise ValueError(
+                    f"Generator does not support data parallel sharding, "
+                    f"got dp_shard={p.data_parallel_shard_degree}"
+                )
+            if p.data_parallel_replicate_degree != 1:
+                raise ValueError(
+                    f"Generator does not support data parallel replication, "
+                    f"got dp_replicate={p.data_parallel_replicate_degree}"
+                )
+            if p.pipeline_parallel_degree > 1:
+                raise ValueError(
+                    f"Generator does not support pipeline parallelism, "
+                    f"got pp={p.pipeline_parallel_degree}"
+                )
+            if p.context_parallel_degree > 1:
+                raise ValueError(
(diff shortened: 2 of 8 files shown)


torchtitan/experiments/graph_trainer/precompile_main.py
wwwjn, line 1: Will revert changes to this file - Just import order change

torchtitan/models/qwen3/parallelize.py
```diff
@@ -108,6 +109,11 @@ def parallelize_qwen3(
             parallel_dims.get_mesh("cp"),
         )
 
+    # Skip FSDP, AC, and compile for inference. FSDP is incompatible
```
wwwjn: For FSDP, here's why FSDP doesn't work with inference mode:

FSDP's pre_forward hook calls unshard() which uses _unsafe_preserve_version_counter to do in-place modifications on parameter tensors (swapping sharded ↔ unsharded views). torch.inference_mode() disables version counting on tensors entirely, so when FSDP tries to access tensor._version, it raises RuntimeError: Inference tensors do not track version counter.      

This happens even at dp_shard=1 because FSDP still registers the hooks — it just doesn't do actual communication. The hook still runs the version counter logic.

cc @weifengpy

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -144,14 +144,34 @@ class Config(Configurable.Config):
...
-                f"got dp_replicate={self.parallelism.data_parallel_replicate_degree}"
-            )
+            # Generator only supports TP. vLLM handles its own parallelism
+            # and we only apply TP via the core parallelize function.
+            p = self.parallelism
+            if p.data_parallel_shard_degree not in (1, -1):
+                raise ValueError(
+                    f"Generator does not support data parallel sharding, "
+                    f"got dp_shard={p.data_parallel_shard_degree}"
+                )
```
tianyu-l: This doesn't guarantee no FSDP, when value is -1. You want to assert this after parallel_dims is generated.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -218,14 +191,26 @@ def __init__(
...
+            ActivationCheckpointConfig,
+            CompileConfig,
+            TrainingConfig,
+        )
+        from torchtitan.protocols.model_converter import ModelConvertersContainer
+
+        self.parallelize_fn(
+            self.model,
             parallel_dims=self.parallel_dims,
+            training=TrainingConfig(),
```
tianyu-l, line 246: we probably shouldn't pass in `TrainingConfig()`
- with full DTensor we probably don't need this assert any more, so can remove seq_len? https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/parallelize.py#L64
- for training.enable_cpu_offload, I think we should move it to `parallelism.enable_fsdp_cpu_offload`

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -218,14 +191,26 @@ def __init__(
...
+            model_converters=ModelConvertersContainer.Config(),
             parallelism=parallelism,
-            has_position_id=True,  # vLLM always passes positions explicitly
+            compile_config=CompileConfig(),
+            ac_config=ActivationCheckpointConfig(),
```
tianyu-l: let's create one with mode being `None` here

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -144,14 +144,34 @@ class Config(Configurable.Config):
...
+            if p.data_parallel_replicate_degree != 1:
+                raise ValueError(
+                    f"Generator does not support data parallel replication, "
+                    f"got dp_replicate={p.data_parallel_replicate_degree}"
+                )
```
tianyu-l, line 154: When EP is used, you will need DP, maybe dp_replicate > 1 should be allowed. But today, torchtitan's dp2ep only works with dp_shard, so there might be bigger changes needed.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -218,14 +191,26 @@ def __init__(
...
+            CompileConfig,
+            TrainingConfig,
+        )
+        from torchtitan.protocols.model_converter import ModelConvertersContainer
+
+        self.parallelize_fn(
+            self.model,
             parallel_dims=self.parallel_dims,
+            training=TrainingConfig(),
+            model_converters=ModelConvertersContainer.Config(),
```
tianyu-l, line 247: @acisseJZhong 's latest PR will deprecate this field

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: Let's make these optional in parallelize_fn.

In particular, let's kill the `"none"` mode in `ActivationCheckpointConfig`.

torchtitan/models/qwen3/parallelize.py
```diff
@@ -108,6 +109,11 @@ def parallelize_qwen3(
...
+    # with torch.inference_mode() and unnecessary for inference.
+    if inference:
```
tianyu-l: mention that generator compilation is handled by passing compile config to vllm

torchtitan/experiments/rl/actors/generator.py
wwwjn: Here we hard coded fsdp degree=1, so I removed FSDP degree check in generator https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/experiments/rl/models/vllm_wrapper.py#L123

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: Sure I could also rebase after ModelConverter is dropped

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This would involve a larger refactor across all the models, we will need to refactor all the parallelize function signature, let me put up another PR to remove these fields from the parallel_plan

wwwjn: https://github.com/pytorch/torchtitan/pull/3108 Here's the other PR refactor the parallelize function signature

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: Done in #3108

torchtitan/experiments/rl/actors/generator.py
wwwjn: > When EP is used, you will need DP, maybe `dp_replicate > 1` should be allowed.

I'm not sure why this statement stands. But beside that, I need to investigate more: In inference, do we use vllm's native parameter data_parallel_replicate_degree to do data parallel? Or should we just spin more vllm instances for data parallel.  In today's implementation, I assumed the latter so add this check.

tianyu-l: you may need more sparse EP degree than dense TP degree, so need additional dense parallelism, typically DP

torchtitan/experiments/rl/actors/generator.py
tianyu-l: can we just assert 1, not -1?

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: Let's init ac config with "none" here. Same for compile. So that you only need to skip FSDP in parallelize fn. What to do in  #3108 needs to be discussed further.

torchtitan/models/qwen3/parallelize.py
tianyu-l: I think you can make this flag `skip_dp` for now (mark it as temporary), and use none mode for compile and ac. In particular, I wonder if later we can switch compilation from vllm to torchtitan. Need to discuss with @Lucaskabela

tianyu-l (approved):

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -205,27 +207,50 @@ def __init__(
...
+        # declarative model.parallelize() API).
+        # TODO: Refactor update_from_config to accept ParallelismConfig
+        # directly instead of requiring a trainer_config wrapper.
+        from types import SimpleNamespace
+
+        from torchtitan.config import DebugConfig, TrainingConfig
+
+        if hasattr(self.config, "update_from_config"):
+            self.config.update_from_config(
+                trainer_config=SimpleNamespace(
```
wwwjn: This is idea from @fegin 's PR and discussion here: https://github.com/pytorch/torchtitan/pull/2986#discussion_r3140649420

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -205,27 +207,50 @@ def __init__(
...
+        # Fill sharding configs on the config BEFORE build so every sub-module
+        # is constructed with its ShardingConfig attached (required by the
+        # declarative model.parallelize() API).
+        # TODO: Refactor update_from_config to accept ParallelismConfig
+        # directly instead of requiring a trainer_config wrapper.
+        from types import SimpleNamespace
+
+        from torchtitan.config import DebugConfig, TrainingConfig
+
+        if hasattr(self.config, "update_from_config"):
```
tianyu-l: should we inherit https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/model.py#L13 so that we don't need this `hasattr`?

torchtitan/models/qwen3/parallelize.py
```diff
@@ -36,6 +36,7 @@ def parallelize_qwen3(
     compile_config: CompileConfig,
     ac_config: ActivationCheckpointConfig,
     dump_folder: str,
+    inference: bool = False,
```
tianyu-l: I think `skip_dp` is more accurate than `inference`

PR #3073, opened 2026-04-23
[rl] Env rollout based + controller refactor

joecummings:
Okay this PR changes several things, but the biggest is how we structure the rollout loop. Generally, we have two options:

1. **Dataset + reward functions.** The dataset yields (prompt, target) pairs; grading is a pure callable (completion, target) -> float,composed via weighted rubrics when you have multiple signals. Simplest shape for single-turn verifiable reward. Used by TRL (GRPOTrainer.reward_funcs), verifiers (Rubric), mitra.
2. **Gym-inspired environments.** An Env exposes reset() / step(action) (or construct-and-go + step); the env owns problem generation, reward, and termination. Scales naturally to multi-turn and tool use. Used by Tinker (tinker-cookbook/rl/types.py: Env.initial_observation + step) and OpenEnv (HTTP-served envs with gym-style reset/step).

The goal of Titan RL is to be used for non-trivial "agentic" training, which fits the pattern of Environments neatly. Forcing a little structure now makes it easier for us to expand to multi-turn later. 

Other things changed:

* Updated reward function to 0.3 b/c if the rewards can add up to the group size, then the advantage can be zero, but you could probably get some information from these trajectories.
* Simplified the train loop w/ more concise logging
* Deleted grader - no longer needed
* Added trajectory and step types
* Changed some variable named (`num_prompts_per_step`) to be more clear

**Tests**: [[P2286462966](https://www.internalfb.com/phabricator/paste/view/P2286462966)](https://www.internalfb.com/phabricator/paste/view/P2288041049)
* Timing looks the same, reward before and after is good


torchtitan/experiments/rl/types.py
```diff
@@ -9,6 +9,19 @@
 import torch
 
 
+@dataclass
```
tianyu-l: For all the dataclasses in this file, it may be good if we follow other config dataclass and mark it `kw_only=True`, `slots=True`
https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L26

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -438,28 +477,23 @@ async def evaluate(self, num_samples: int = 20) -> dict:
...
-        for i, answer in enumerate(eval_answers):
-            c = completions[i * n]
-            extracted = extract_answer(c.text)
-            correct += int(extracted == int(answer))
-            format_ok += int(bool(re.search(r"\[ANSWER\]", c.text)))
+        for i, env in enumerate(envs):
+            c = completions[i * group_size]
+            step = env.step(c.text)
+            # Correct iff reward includes the +1.0 correctness component.
+            correct += int(step.reward >= 1.0)
```
tianyu-l: hmm why not use `correctness_reward`? o/w unnecessarily leaking details of reward fn

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -438,28 +477,23 @@ async def evaluate(self, num_samples: int = 20) -> dict:
...
+        envs = [SumDigitsEnv(eval_rng) for _ in range(num_samples)]
         completions = self._get_rank_0_value(
-            self.generator.generate.call(eval_prompts).get()
+            self.generator.generate.call([env.prompt for env in envs]).get()
         )
 
-        # Score the first sample per prompt. Completions come back in
-        # prompt_idx order, length num_samples * n.
-        n = self.config.generator.sampling.n
+        # Score the first completion per prompt via the env. Completions come
```
tianyu-l: "scoring the first completion" sounds arbitrary. Is there logic behind -- what if I score everything?

torchtitan/experiments/rl/sum_digits.py
```diff
@@ -9,43 +9,33 @@
 import random
 import re
 
-import torch
+from torchtitan.experiments.rl.types import Step
 
-from torchtitan.experiments.rl.types import Completion
 
+def correctness_reward(completion: str, target: int) -> float:
```
tianyu-l: should they not be part of the Env?

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -420,6 +413,52 @@ async def setup(
         # pull weights for policy version 0 (initial weights)
         self.generator.pull_model_state_dict.call(0).get()
 
+    def _collect_rollout(self, num_prompts: int) -> list[Trajectory]:
+        """Collect group rollouts: one single use env per prompt, scored and returned"""
```
tianyu-l: if multi-turn, I imagine the logic is more complicated -- each sample in each group can stop at different number of turns?

torchtitan/experiments/rl/grpo.py
```diff
@@ -420,6 +413,52 @@ async def setup(
         # pull weights for policy version 0 (initial weights)
         self.generator.pull_model_state_dict.call(0).get()
 
+    def _collect_rollout(self, num_prompts: int) -> list[Trajectory]:
+        """Collect group rollouts: one single use env per prompt, scored and returned"""
+        group_size = self.config.generator.sampling.n
+        envs = [SumDigitsEnv(self._train_rng) for _ in range(num_prompts)]
+        completions = self._get_rank_0_value(
```
tianyu-l, line 449: https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L247
The comment there seems wrong -- this is ordered by prompt_idx, but not sample idx (which doesn't exist?)

torchtitan/experiments/rl/grpo_sync.py
```diff
@@ -420,6 +413,52 @@ async def setup(
...
+        """Collect group rollouts: one single use env per prompt, scored and returned"""
+        group_size = self.config.generator.sampling.n
+        envs = [SumDigitsEnv(self._train_rng) for _ in range(num_prompts)]
+        completions = self._get_rank_0_value(
+            self.generator.generate.call([env.prompt for env in envs]).get()
+        )
+
+        trajectories: list[Trajectory] = []
+        for sample_idx, env in enumerate(envs):
+            group = completions[sample_idx * group_size : (sample_idx + 1) * group_size]
```
tianyu-l: what's the purpose of this grouping?

torchtitan/experiments/rl/grpo.py
```diff
@@ -420,6 +413,52 @@ async def setup(
...
+            for c in group:
+                step = env.step(c.text)  # Score the Completion
+                trajectories.append(
+                    Trajectory(sample_idx=sample_idx, transitions=[(c, step)])
+                )
+        return trajectories
+
+    @staticmethod
+    def _build_episodes(trajectories: list[Trajectory]) -> list[Episode]:
+        """Group trajectories by sample, apply mean-baseline advantage, flatten to Episodes."""
```
tianyu-l, line 462: curious: in multi-turn, for each Trajectory, how many Episode do we usually build -- is it typically just one, or num_steps many

torchtitan/experiments/rl/simple_grpo_sum_digits.py
```diff
@@ -438,28 +477,23 @@ async def evaluate(self, num_samples: int = 20) -> dict:
...
-            c = completions[i * n]
-            extracted = extract_answer(c.text)
-            correct += int(extracted == int(answer))
-            format_ok += int(bool(re.search(r"\[ANSWER\]", c.text)))
+        for i, env in enumerate(envs):
+            c = completions[i * group_size]
+            step = env.step(c.text)
+            # Correct iff reward includes the +1.0 correctness component.
+            correct += int(step.reward >= 1.0)
+            format_ok += int(format_reward(c.text) > 0)
```
tianyu-l: how would you imagine refactoring this, if this script is to be made env-agnostic?

tianyu-l (commented):
nice, having some questions

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: It is arbitrary, remnant of previous iterations - there's a TODO to make this legit (greedy decoding, pass@k).

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: In that case, we would drop the batched generation contract we're using right now and opt for one in which each rollout would call generate individually (and continue to do so until the Env reports done). We'd span this out across multiple generator replicas. This would require some reworking of generator to schedule across different requests but return as completed, rather than waiting for all to be done.

torchtitan/experiments/rl/grpo_sync.py
joecummings: Calculating advantage

torchtitan/experiments/rl/grpo.py
joecummings: Answered offline, but can be changed according to the goals of the researcher. We'd start with just one.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: Will be partially done here. Env returns rewards broken down by reward functions, user passes in entire Env class to config in order to train. Then this script can just be called `grpo_sync.py` or something similar.

torchtitan/experiments/rl/simple_grpo_sum_digits.py
joecummings: I actually just decided to update it here to just do greedy decoding

torchtitan/experiments/rl/grpo.py
joecummings: Fixed

torchtitan/experiments/rl/grpo_sync.py
```diff
@@ -420,195 +441,146 @@ async def setup(
...
-        pre-refactor eval signal.
-
-        TODO: report both greedy (temperature=0, n=1) and pass@k
-        (training temp, n=k) so mode-sharpening vs. coverage
-        regressions are both visible. A single-sample score hides
-        both signals.
-
-        Args:
-            num_samples: Number of eval prompts to generate.
+    def _collect_rollout(self, step: int, num_prompts: int) -> list[Trajectory]:
```
tianyu-l: wonder if it is more consistent to call it
```suggestion
    def _collect_trajectories(self, step: int, num_prompts: int) -> list[Trajectory]:
```

torchtitan/experiments/rl/grpo.py
```diff
@@ -124,6 +127,22 @@ def _bootstrap():
...
+
+def _format_rewards(components: dict[str, float]) -> str:
+    return ", ".join(f"{k}={v:+.3f}" for k, v in components.items())
+
+
+def _format_validation(result: dict) -> str:
+    return (
+        f"mean_reward={result['mean_reward']:+.3f} "
+        f"({_format_rewards(result['components'])})"
+    )
```
tianyu-l, line 143: these look somewhat ad hoc, but I don't if it's too early to worry about structures on them

torchtitan/experiments/rl/grpo.py
```diff
@@ -420,195 +441,146 @@ async def setup(
...
-            correct_count = sum(1 for r in all_rewards if r > 0)
-            total_count = len(all_rewards)
-
+            self.generator.pull_model_state_dict.call(metrics["policy_version"]).get()
+            t_sync = time.perf_counter() - t0
+            logger.info(f"Weight sync: push={t_push:.3f}s, total={t_sync:.3f}s")
+
+            # --- Logging --- #
+            steps = [t.transitions[0][1] for t in trajectories]
+            components = _mean_rewards(steps)
```
tianyu-l, line 558: e.g. this could be put in a method in `Trajectories`

torchtitan/experiments/rl/grpo_sync.py
tianyu-l: IIUC grouping is only used in `_build_episodes`, achieved by `groups.setdefault(t.sample_idx, []).append(t)`

I believe here you don't need this grouping.

torchtitan/experiments/rl/README.md
```diff
@@ -8,7 +8,7 @@ This directory contains code for RL training using TorchTitan model definitions
 The integration consists of the following components:
 
 1. **vLLM Model Wrapper** (`models/vllm_wrapper.py`): Adapts TorchTitan models for vLLM's inference engine
-2. **RL Training Loop** (`simple_grpo_sum_digits.py`): GRPO-based RL training with Monarch actors
+2. **RL Training Loop** (`grpo_sync.py`): GRPO-based RL training with Monarch actors
```
tianyu-l: maybe mention this is single turn only

tianyu-l (approved):

daniellepintz:
Not a huge fan of the name `grpo_sync` - could we just call it `grpo.py`? with "sync" it isn't immediately clear to me that it's talking about synchronous RL

torchtitan/experiments/rl/grpo_sync.py
joecummings: It's more type consistent but rollouts are the canonical RL literature reference

torchtitan/experiments/rl/grpo.py
joecummings: Yeah was just trying to declutter the main training loop a little

torchtitan/experiments/rl/grpo_sync.py
joecummings: Ah I misread, this is correct - will remove this extra grouping.

joecummings:
> Not a huge fan of the name `grpo_sync` - could we just call it `grpo.py`? with "sync" it isn't immediately clear to me that it's talking about synchronous RL

I'm fine with doing grpo.py right now and async will be called `grpo_async.py`.

PR #3075, opened 2026-04-23
[GraphTrainer][AutoDev] Support float8 linear in aot_fx_trace path

SherlockNoMad:
## Summary

- Add `graph_trainer_llama3_debugmodel_float8` config to the graph_trainer llama3 config registry, reusing the existing `llama3_debugmodel_float8` base config
- Add float8 + aot_fx_trace integration test (`aot_fx_trace_llama3_float8_fsdp_tp`) to the H100 test suite covering FSDP+TP with float8 tensorwise scaling
- Include float8 tests in both `graph_trainer` (all) and `graph_trainer_h100` test suites

## Why this should work

The aot_fx_trace path already has the infrastructure to handle Float8Linear:

1. **Subclass unwrapping**: `make_fx_tracer.py`'s `_unwrap_subclass()` recursively flattens any traceable wrapper subclass (including Float8Tensor nested inside DTensor) using the standard `__tensor_flatten__`/`__tensor_unflatten__` protocol
2. **Float8 TP detection**: `graph_trainer/llama3/parallelize.py` already detects `Float8LinearConverter.Config` and passes `enable_float8_tensorwise_tp=True` to `apply_tp`, which uses `Float8ColwiseParallel`/`Float8RowwiseParallel` from torchao
3. **Graph passes**: The default passes (SAC, regional_inductor, cudagraph) operate on aten ops and are dtype-agnostic -- float8 casts and scaled matmuls become explicit aten ops after tracing

The `enable_fsdp_float8_all_gather` and `precompute_float8_dynamic_scale_for_fsdp` flags from the base float8 config are no-ops with graph_trainer's `simple_fsdp` (which doesn't use `fully_shard()`), so they are harmless.

## Test plan

- [ ] Run `aot_fx_trace_llama3_float8_fsdp_tp` integration test on H100 (8 GPUs):
  ```bash
  python torchtitan/experiments/graph_trainer/tests/integration_tests.py <output_dir> \
      --test_suite graph_trainer_h100 --test_name aot_fx_trace_llama3_float8_fsdp_tp --ngpu 8
  ```
- [ ] Numerical validation: compare loss/grad_norm against eager float8 training with `--debug.seed=42 --debug.deterministic`
- [ ] Verify pre-commit passes (pyrefly failures are pre-existing in core files, not from this change)

## Note

This is a first-time exercise of float8 + aot_fx_trace. CI validation on H100 machines is required to confirm correctness. Cannot run GPU tests locally without GPU access.

### DIFF
--- a/.github/workflows/integration_test_8gpu_graph_trainer.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer.yaml
@@ -54,6 +54,7 @@ jobs:
 
         python -m pip install --force-reinstall --pre \
           torch --index-url https://download.pytorch.org/whl/nightly/cu130
+        USE_CPP=0 python -m pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cu130
         python -m pip install torchdata==0.12.0.dev20260327 --extra-index-url https://download.pytorch.org/whl/nightly/cpu
 
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
--- a/.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
+++ b/.github/workflows/integration_test_8gpu_graph_trainer_h100.yaml
@@ -54,6 +54,7 @@ jobs:
 
         python -m pip install --force-reinstall --pre \
           torch --index-url https://download.pytorch.org/whl/nightly/cu130
+        USE_CPP=0 python -m pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cu130
         python -m pip install torchdata==0.12.0.dev20260327 --extra-index-url https://download.pytorch.org/whl/nightly/cpu
 
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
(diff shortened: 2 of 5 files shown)


yiming0416 (approved):

tianyu-l (approved):

aditvenk:
> The enable_fsdp_float8_all_gather and precompute_float8_dynamic_scale_for_fsdp flags from the base float8 config are no-ops with graph_trainer's simple_fsdp (which doesn't use fully_shard()), so they are harmless.

@SherlockNoMad  -- I am not sure these are harmless if they dictate the dtype of the bytes sent over the wire -- SimpleFSDP will be sending bf16 which is 2x the bandwidth usage as what FSDP2 would be doing. 
cc @weifengpy

SherlockNoMad:
> > The enable_fsdp_float8_all_gather and precompute_float8_dynamic_scale_for_fsdp flags from the base float8 config are no-ops with graph_trainer's simple_fsdp (which doesn't use fully_shard()), so they are harmless.
> 
> @SherlockNoMad -- I am not sure these are harmless if they dictate the dtype of the bytes sent over the wire -- SimpleFSDP will be sending bf16 which is 2x the bandwidth usage as what FSDP2 would be doing. cc @weifengpy

yes, this is composability gap between simplefsdp and low percision support. we should fix this . Tracked in kanban.

Issue #2716, opened 2026-03-26
[RL] Unify mask creation logic for RL.


zhxchen17:
Today on trainer we create attention mask using the helper function.

We should investigate whether this piece can be merged and put into a dedicated place with code from https://github.com/verl-project/verl/pull/5051/changes#diff-d90b34d27a9c17f5a898247fd8858ab3e61876d67ed2d8653e18bd5f85658a67R156 (`_create_varlen_metadata_for_document`).

Context: following up https://github.com/pytorch/torchtitan/pull/2649

cc @acisseJZhong

acisseJZhong:
cc @joecummings

PR #3077, opened 2026-04-23
add Flex Attention to torchtitan RL loop

### LINKED ISSUE #2719: [RL] Support flex attention kernel for generator and trainer.
Today both generator (vllm) and trainer (titan) use PyTorch's varlen kernel from the core to make sure attentions are numerically matching. For achieving full model coverage we need to support alternative attention backend for both generator and trainer for cases varlen doesn't cover. This should be similar to how varlen backend is used in generator and varlen today.

Creating this issue for tracking purposes (in case we forgot) and some context are also being tracked in https://docs.google.com/document/d/1nxsjXrfdv9oWxZ6i2onOc2Bi0XdO4Pdygeuce3YHnxA/edit?tab=t.0#heading=h.7ysj2n82t39u (sorry for the internal link).

cc @wwwjn @liangel-02 @tianyu-l @acisseJZhong

liangel-02:
i added FlexAttention to the torchtitan RL loop. for the generator side we can call vLLM's flex attention impl directly. i tested with the grpo script: `python torchtitan/experiments/rl/simple_grpo_sum_digits.py --module rl --config rl_grpo_qwen3_0_6b`

batch invariant:
Screenshot 2026-06-01 at 6 42 29 PM

not batch invariant: 
Screenshot 2026-06-01 at 6 45 34 PM

also updated `torchtitan/experiments/rl/tests/test_bitwise_parity.py` to include flex
Screenshot 2026-06-01 at 6 43 02 PM

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -26,7 +26,7 @@ from torchtitan.experiments.rl.models.vllm_registry import (
 )
 from torchtitan.experiments.rl.observability import metrics as m
 from torchtitan.experiments.rl.types import Completion
-from torchtitan.models.common.attention import FlexAttention
+from torchtitan.models.common.attention import FlexAttention, VarlenAttention
 from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.model_spec import ModelSpec
 from torchtitan.tools.logging import init_logger
@@ -288,12 +288,11 @@ class VLLMGenerator(Actor, Configurable):
 
         # Set vLLM environment variables from config before any vLLM initialization
         inner_attn = model_spec.model.layers[0].attention.inner_attention
-        self._use_flex = isinstance(inner_attn, FlexAttention.Config)
+        assert isinstance(
+            inner_attn,
+            (VarlenAttention.Config, FlexAttention.Config),
+        ), "Only varlen and flex attention backends are allowed."
 
-        if self._use_flex:
-            os.environ["VLLM_ATTENTION_BACKEND"] = "FLEX_ATTENTION"
-        else:
-            os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
         os.environ["VLLM_USE_V2_MODEL_RUNNER"] = "1"
         set_batch_invariance(config.debug.batch_invariant)
 
@@ -324,7 +323,7 @@ class VLLMGenerator(Actor, Configurable):
             attention_config=AttentionConfig(
                 backend=(
                     AttentionBackendEnum.FLEX_ATTENTION
-                    if self._use_flex
+                    if isinstance(inner_attn, FlexAttention.Config)
                     else AttentionBackendEnum.CUSTOM
                 ),
             ),
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -211,9 +211,6 @@ class VLLMGenerator(Actor, Configurable):
         debug: DebugConfig = field(default_factory=DebugConfig)
         """Debug and determinism settings."""
 
-        vllm_attn_backend: str = "varlen"
-        """Which attention impl to use inside vLLM: 'varlen' or 'flex'."""
-
         def __post_init__(self):
             # VLLMGenerator only supports TP. vLLM handles its own parallelism;
             # we only apply TP via the core parallelize function.
@@ -288,26 +285,17 @@ class VLLMGenerator(Actor, Configurable):
             checkpoint_config=config.checkpoint,
         )
 
-        # Set vLLM environment variables from config before any vLLM initialization
-        if config.vllm_attn_backend == "flex":
+        from torchtitan.models.common.attention import FlexAttention
+
+        inner_attn = model_spec.model.layers[0].attention.inner_attention
+        self._use_flex = isinstance(inner_attn, FlexAttention.Config)
+
+        if self._use_flex:
             os.environ["VLLM_ATTENTION_BACKEND"] = "FLEX_ATTENTION"
-            if config.debug.batch_invariant:
-                os.environ["VLLM_BATCH_INVARIANT"] = "1"
         else:
             os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
         os.environ["VLLM_USE_V2_MODEL_RUNNER"] = "1"
-
         set_batch_invariance(config.debug.batch_invariant)
-
-        # Set vLLM attention backend
-        # if config.vllm_attn_backend == "flex":
-        #     os.environ["VLLM_ATTENTION_BACKEND"] = "FLEX_ATTENTION"
-        #     if config.debug.batch_invariant:
-        #         os.environ["VLLM_BATCH_INVARIANT"] = "1"
-        # else:
-        #     os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
-        #     if config.debug.batch_invariant:
-        #         set_batch_invariance(True)
 
         self._set_determinism(config.debug)
 
@@ -336,7 +324,7 @@ class VLLMGenerator(Actor, Configurable):
             attention_config=AttentionConfig(
                 backend=(
                     AttentionBackendEnum.FLEX_ATTENTION
-                    if config.vllm_attn_backend == "flex"
+                    if self._use_flex
                     else AttentionBackendEnum.CUSTOM
                 ),
             ),
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -85,9 +85,9 @@ def _prepare_generation_request_metrics(
         if stats.num_generation_tokens > 1:
             first_to_last_token_ms = (stats.last_token_ts - stats.first_token_ts) * 1000
             metric_values[f"{prefix}/decode_time_ms"] = first_to_last_token_ms
-            metric_values[
-                f"{prefix}/inter_token_latency_ms"
-            ] = first_to_last_token_ms / (stats.num_generation_tokens - 1)
+            metric_values[f"{prefix}/inter_token_latency_ms"] = (
+                first_to_last_token_ms / (stats.num_generation_tokens - 1)
+            )
 
     # Emit each value with both Mean and Max aggregators.
     return [
@@ -211,6 +211,9 @@ class VLLMGenerator(Actor, Configurable):
         debug: DebugConfig = field(default_factory=DebugConfig)
         """Debug and determinism settings."""
 
+        vllm_attn_backend: str = "varlen"
+        """Which attention impl to use inside vLLM: 'varlen' or 'flex'."""
+
         def __post_init__(self):
             # VLLMGenerator only supports TP. vLLM handles its own parallelism;
             # we only apply TP via the core parallelize function.
@@ -286,10 +289,25 @@ class VLLMGenerator(Actor, Configurable):
         )
 
         # Set vLLM environment variables from config before any vLLM initialization
-        os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
+        if config.vllm_attn_backend == "flex":
+            os.environ["VLLM_ATTENTION_BACKEND"] = "FLEX_ATTENTION"
+            if config.debug.batch_invariant:
+                os.environ["VLLM_BATCH_INVARIANT"] = "1"
+        else:
+            os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
         os.environ["VLLM_USE_V2_MODEL_RUNNER"] = "1"
 
         set_batch_invariance(config.debug.batch_invariant)
+
+        # Set vLLM attention backend
+        # if config.vllm_attn_backend == "flex":
+        #     os.environ["VLLM_ATTENTION_BACKEND"] = "FLEX_ATTENTION"
+        #     if config.debug.batch_invariant:
+        #         os.environ["VLLM_BATCH_INVARIANT"] = "1"
+        # else:
+        #     os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
+        #     if config.debug.batch_invariant:
+        #         set_batch_invariance(True)
 
         self._set_determinism(config.debug)
 
@@ -316,16 +334,24 @@ class VLLMGenerator(Actor, Configurable):
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=not config.cudagraph.enable,
             attention_config=AttentionConfig(
-                backend=AttentionBackendEnum.CUSTOM,
+                backend=(
+                    AttentionBackendEnum.FLEX_ATTENTION
+                    if config.vllm_attn_backend == "flex"
+                    else AttentionBackendEnum.CUSTOM
+                ),
             ),
             # Enables RequestOutput.metrics, so generator metrics can be returned
             disable_log_stats=False,
         )
         engine_kwargs["max_model_len"] = model_spec.model.rope.max_seq_len
         engine_kwargs["max_num_seqs"] = self._max_num_seqs
-        # FA2 requires block_size to be a multiple of 256
-        if not has_cuda_capability(9, 0):
+
+        if config.vllm_attn_backend == "flex":
+            engine_kwargs["enable_chunked_prefill"] = False
+        elif not has_cuda_capability(9, 0):
+            # FA2 requires block_size to be a multiple of 256
             engine_kwargs["block_size"] = 256
+
         vllm_compilation_config = config.cudagraph.get_vllm_compilation_config(
             max_num_seqs=self._max_num_seqs,
         )
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -334,7 +334,7 @@ class VLLMGenerator(Actor, Configurable):
         engine_kwargs["max_model_len"] = model_spec.model.rope.max_seq_len
         engine_kwargs["max_num_seqs"] = self._max_num_seqs
 
-        if config.vllm_attn_backend == "flex":
+        if self._use_flex:
             engine_kwargs["enable_chunked_prefill"] = False
         elif not has_cuda_capability(9, 0):
             # FA2 requires block_size to be a multiple of 256
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -26,6 +26,7 @@ from torchtitan.experiments.rl.models.vllm_registry import (
 )
 from torchtitan.experiments.rl.observability import metrics as m
 from torchtitan.experiments.rl.types import Completion
+from torchtitan.models.common.attention import FlexAttention
 from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.model_spec import ModelSpec
 from torchtitan.tools.logging import init_logger
@@ -85,9 +86,9 @@ def _prepare_generation_request_metrics(
         if stats.num_generation_tokens > 1:
             first_to_last_token_ms = (stats.last_token_ts - stats.first_token_ts) * 1000
             metric_values[f"{prefix}/decode_time_ms"] = first_to_last_token_ms
-            metric_values[f"{prefix}/inter_token_latency_ms"] = (
-                first_to_last_token_ms / (stats.num_generation_tokens - 1)
-            )
+            metric_values[
+                f"{prefix}/inter_token_latency_ms"
+            ] = first_to_last_token_ms / (stats.num_generation_tokens - 1)
 
     # Emit each value with both Mean and Max aggregators.
     return [
@@ -285,8 +286,7 @@ class VLLMGenerator(Actor, Configurable):
             checkpoint_config=config.checkpoint,
         )
 
-        from torchtitan.models.common.attention import FlexAttention
-
+        # Set vLLM environment variables from config before any vLLM initialization
         inner_attn = model_spec.model.layers[0].attention.inner_attention
         self._use_flex = isinstance(inner_attn, FlexAttention.Config)
 
@@ -333,13 +333,9 @@ class VLLMGenerator(Actor, Configurable):
         )
         engine_kwargs["max_model_len"] = model_spec.model.rope.max_seq_len
         engine_kwargs["max_num_seqs"] = self._max_num_seqs
-
-        if self._use_flex:
-            engine_kwargs["enable_chunked_prefill"] = False
-        elif not has_cuda_capability(9, 0):
-            # FA2 requires block_size to be a multiple of 256
+        # FA2 requires block_size to be a multiple of 256
+        if not has_cuda_capability(9, 0):
             engine_kwargs["block_size"] = 256
-
         vllm_compilation_config = config.cudagraph.get_vllm_compilation_config(
             max_num_seqs=self._max_num_seqs,
         )
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -15,7 +15,6 @@ import torch.distributed.distributed_c10d as c10d
 import torch.nn.functional as F
 import torchstore as ts
 from monarch.actor import Actor, current_rank, endpoint
-from torch.nn.attention.flex_attention import and_masks
 from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.components.loss import IGNORE_INDEX
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
@@ -33,13 +32,7 @@ from torchtitan.config import (
 from torchtitan.distributed import ParallelDims, utils as dist_utils
 from torchtitan.distributed.utils import set_batch_invariance
 from torchtitan.experiments.rl.types import OptimStepOutput, TrainingBatch
-from torchtitan.models.common.attention import (
-    create_attention_mask,
-    create_varlen_metadata_for_document,
-    FlexAttention,
-    get_causal_mask_mod,
-    get_document_mask_mod,
-)
+from torchtitan.models.common.attention import FlexAttention
 from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.model_spec import ModelSpec
 from torchtitan.tools import utils
@@ -214,13 +207,6 @@ class PolicyTrainer(Actor, Configurable):
         else:
             self.sd_adapter = None
 
-        # Detect attention backend for mask creation
-        inner_attn_cfg = model_spec.model.layers[0].attention.inner_attention
-        self._use_flex = isinstance(inner_attn_cfg, FlexAttention.Config)
-        self._batch_invariant = config.debug.batch_invariant
-        if self._use_flex:
-            self._flex_block_size = inner_attn_cfg.block_size
-
         # Create training policy model
         model = self._build_model(model_spec, config, device_type)
         model.train()
@@ -336,20 +322,6 @@ class PolicyTrainer(Actor, Configurable):
 
         return model
 
-    def _create_flex_masks(self, positions: torch.Tensor):
-        """Create a flex attention BlockMask from packed positions."""
-        B, S = positions.shape
-        mask_mod = and_masks(get_causal_mask_mod(), get_document_mask_mod(positions))
-        return create_attention_mask(
-            mask_mod,
-            B,
-            None,
-            S,
-            S,
-            BLOCK_SIZE=self._flex_block_size,
-            separate_full_blocks=not self._batch_invariant,
-        )
-
     @endpoint
     async def sync_log_step(self, step: int, relative_step: int | None = None) -> None:
         """Sync the structured-logger step counter from the controller."""
@@ -435,10 +407,7 @@ class PolicyTrainer(Actor, Configurable):
         generator_logprobs = local_batch.generator_logprobs.to(device)
         advantages = local_batch.advantages.to(device)
 
-        if self._use_flex:
-            attention_masks = self._create_flex_masks(positions)
-        else:
-            attention_masks = create_varlen_metadata_for_document(positions)
+        attention_masks = model.get_attention_masks(positions)
 
         with sl.log_trace_span("model_forward"):
             logits = model(
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -31,8 +31,21 @@ from torchtitan.config import (
 )
 from torchtitan.distributed import ParallelDims, utils as dist_utils
 from torchtitan.distributed.utils import set_batch_invariance
+from torchtitan.experiments.rl.actors.utils import (
+    compute_logprobs,
+    create_flex_block_mask,
+    create_positions_from_seq_lens,
+    create_varlen_metadata,
+    extract_response_logprobs,
+    pad_to_block_aligned,
+    unpad_from_block_aligned,
+    verify_logprob_identity,
+)
 from torchtitan.experiments.rl.types import OptimStepOutput, TrainingBatch
-from torchtitan.models.common.attention import create_varlen_metadata_for_document
+from torchtitan.models.common.attention import (
+    create_varlen_metadata_for_document,
+    FlexAttention,
+)
 from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.model_spec import ModelSpec
 from torchtitan.tools import utils
@@ -207,6 +220,22 @@ class PolicyTrainer(Actor, Configurable):
         else:
             self.sd_adapter = None
 
+        # Detect attention backend for mask creation
+        inner_attn_cfg = model_spec.model.layers[0].attention.inner_attention
+        self._use_flex = isinstance(inner_attn_cfg, FlexAttention.Config)
+        self._batch_invariant = config.debug.batch_invariant
+        if self._use_flex:
+            self._flex_block_size = inner_attn_cfg.block_size
+
+        # Build model converters (e.g. float8). No-op if converters list is empty.
+        model_compile_enabled = (
+            config.compile.enable and "model" in config.compile.components
+        )
+        model_converters = config.model_converters.build(
+            parallel_dims=self.parallel_dims,
+            model_compile_enabled=model_compile_enabled,
+        )
+
         # Create training policy model
         model = self._build_model(model_spec, config, device_type)
         model.train()
@@ -218,6 +247,14 @@ class PolicyTrainer(Actor, Configurable):
         self.lr_schedulers = config.lr_scheduler.build(
             optimizers=self.optimizers,
             training_steps=config.training.steps,
+        )
+
+        # Post optimizer step model converters hook.
+        # e.g. calculate float8 dynamic amax/scale for FSDP2
+        self.optimizers.register_step_post_hook(
+            lambda *args, **kwargs: model_converters.post_optimizer_hook(
+                self.model_parts
+            )
         )
 
         self.policy_version = 0
@@ -289,12 +326,13 @@ class PolicyTrainer(Actor, Configurable):
             Model with random-initialized weights.
         """
 
-        # TODO: Also support flex attention backend later.
         from torchtitan.models.common.attention import VarlenAttention
 
+        inner_attn = model_spec.model.layers[0].attention.inner_attention
         assert isinstance(
-            model_spec.model.layers[0].attention.inner_attention, VarlenAttention.Config
-        ), "Only varlen attention backend is allowed."
+            model_spec.model.layers[0].attention.inner_attention,
+            (VarlenAttention.Config, FlexAttention.Config),
+        ), "Only varlen and flex attention backends are allowed."
 
         # Fill sharding configs on the config BEFORE build via the
         # model-agnostic `update_from_config` hook (RL's trainer bypasses
@@ -305,10 +343,13 @@ class PolicyTrainer(Actor, Configurable):
             with utils.set_default_dtype(TORCH_DTYPE_MAP[config.training.dtype]):
                 model = model_spec.model.build()
 
+        model_converters.convert(model)
+
         model = model_spec.parallelize_fn(
             model,
             parallel_dims=self.parallel_dims,
             training=config.training,
+            model_converters=config.model_converters,
             parallelism=config.parallelism,
             compile_config=self.compile_config,
             ac_config=config.ac_config,
@@ -413,6 +454,50 @@ class PolicyTrainer(Actor, Configurable):
                 token_ids, attention_masks=attention_masks, positions=positions
             )
         policy_logprobs = compute_logprobs(logits, labels)
+        # max_seq_len = max(seq_lens)
+        # rope_cache_len = self.model.freqs_cis.shape[0]
+        # if max_seq_len > rope_cache_len:
+        #     raise ValueError(
+        #         f"Episode length {max_seq_len} exceeds rope cache size "
+        #         f"{rope_cache_len}. Increase model max_seq_len or reduce "
+        #         f"generation max_tokens."
+        #     )
+
+        # if self._use_flex and self._batch_invariant:
+        #     attention_masks, maybe_padded_seq_lens = create_flex_block_mask(
+        #         seq_lens, device, self._flex_block_size
+        #     )
+        #     padded_ids, positions = pad_to_block_aligned(
+        #         token_ids.squeeze(0), seq_lens, maybe_padded_seq_lens, device
+        #     )
+        #     logits = self.model(
+        #         padded_ids, attention_masks=attention_masks, positions=positions
+        #     )
+        #     logits = unpad_from_block_aligned(logits, seq_lens, maybe_padded_seq_lens)
+        #     token_ids = unpad_from_block_aligned(
+        #         padded_ids, seq_lens, maybe_padded_seq_lens
+        #     )
+        # elif self._use_flex:
+        #     attention_masks, _ = create_flex_block_mask(
+        #         seq_lens, device, self._flex_block_size
+        #     )
+        #     positions = create_positions_from_seq_lens(seq_lens, device)
+        #     logits = self.model(
+        #         token_ids, attention_masks=attention_masks, positions=positions
+        #     )
+        # else:
+        #     positions = torch.cat(
+        #         [torch.arange(l, device=device) for l in seq_lens]
+        #     ).unsqueeze(0)
+        #     attention_masks = create_varlen_metadata_for_document(positions)
+        #     logits = self.model(
+        #         token_ids, attention_masks=attention_masks, positions=positions
+        #     )
+
+        # all_policy_logprobs = compute_logprobs(logits, token_ids)
+        # policy_logprobs = extract_response_logprobs(
+        #     all_policy_logprobs, seq_lens, prompt_lens, response_lens
+        # )
 
         with sl.log_trace_span("loss_fn"):
             loss, loss_metrics = self.loss_fn(
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -15,6 +15,7 @@ import torch.distributed.distributed_c10d as c10d
 import torch.nn.functional as F
 import torchstore as ts
 from monarch.actor import Actor, current_rank, endpoint
+from torch.nn.attention.flex_attention import and_masks
 from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.components.loss import IGNORE_INDEX
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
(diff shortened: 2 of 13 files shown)


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -166,10 +169,15 @@ def __init__(
...
 
-        # Set vLLM environment variables from config before any vLLM initialization
-        os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
-
-        set_batch_invariance(config.debug.batch_invariant)
+        # Set vLLM attention backend
+        if config.vllm_attn_backend == "flex":
+            os.environ["VLLM_ATTENTION_BACKEND"] = "FLEX_ATTENTION"
+            if config.debug.batch_invariant:
+                os.environ["VLLM_BATCH_INVARIANT"] = "1"
```
wwwjn: During batch-invariant, I avoid to use VLLM_BATCH_INVARIANT, which will causing vllm patching a lot of fields (eg, patching batch-invariant ops) more than just attention part. Instead we want to control batch-invariant ourself, so we use this function to explicitly do that: https://github.com/pytorch/torchtitan/blob/bd6aec6195cbe86838c03a567f0638b5900528a9/torchtitan/distributed/utils.py#L218

Do we know for flexAttention, what parameters / changes will vllm do it we set `VLLM_BATCH_INVARIANT=1`?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -304,12 +315,35 @@ async def forward_backward(self, train_data: list[TrainBatch]) -> dict:
                 f"generation max_tokens."
             )
 
-        attention_masks = create_varlen_metadata(seq_lens, device)
-        positions = create_positions_from_seq_lens(seq_lens, device)
+        if self._use_flex and self._batch_invariant:
```
wwwjn: Can we have a unified https://github.com/pytorch/torchtitan/blob/bd6aec6195cbe86838c03a567f0638b5900528a9/torchtitan/models/common/decoder.py#L170 `get_attention_masks`, or reuse the `model.get_attention_mask`? Basically can we further unify these 3 paths?

torchtitan/experiments/rl/actors/generator.py
liangel-02: changed to use set_batch_invariance()

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -143,6 +143,9 @@ class Config(Configurable.Config):
         debug: DebugConfig = field(default_factory=DebugConfig)
         """Debug and determinism settings."""
 
+        vllm_attn_backend: str = "varlen"
```
tianyu-l: this should be coming from model_spec, not generator config

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -166,8 +169,11 @@ def __init__(
         # Register TorchTitan model with vLLM before any engine creation
         register_model_to_vllm_model_registry(model_spec)
 
-        # Set vLLM environment variables from config before any vLLM initialization
-        os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
+        # Set vLLM attention backend
```
tianyu-l: use model_spec to figure out

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -188,12 +194,16 @@ def __init__(
...
+                backend=AttentionBackendEnum.FLEX_ATTENTION
+                if config.vllm_attn_backend == "flex"
+                else AttentionBackendEnum.CUSTOM,
             ),
             disable_log_stats=True,
         )
-        # FA2 requires block_size to be a multiple of 256
-        if not has_cuda_capability(9, 0):
+        if config.vllm_attn_backend == "flex":
+            engine_kwargs["enable_chunked_prefill"] = False
```
tianyu-l: does flex not work with chunked prefill? should be noted somewhere in docstring / readme?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -30,13 +32,58 @@
...
+    )
+
+
+def rl_grpo_qwen3_0_6b_flex() -> RLTrainer.Config:
+    """GRPO training config for Qwen3-0.6B with flex attention (4 GPUs: 2 gen + 2 train)."""
+    spec = model_registry("0.6B", attn_backend="flex")
+    for layer_cfg in spec.model.layers:
+        layer_cfg.attention.inner_attention.kernel_options[
+            "FORCE_USE_FLEX_ATTENTION"
+        ] = True
```
tianyu-l: what's this for?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -73,6 +120,68 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
+def rl_grpo_qwen3_0_6b_flex_batch_invariant() -> RLTrainer.Config:
+    """GRPO training config for Qwen3-0.6B with flex attention and batch invariance
+    for bitwise-identical numerics between trainer and generator (4 GPUs: 2 gen + 2 train).
+    """
+    spec = model_registry("0.6B", attn_backend="flex")
+    for layer_cfg in spec.model.layers:
+        layer_cfg.attention.inner_attention.kernel_options[
+            "FORCE_USE_FLEX_ATTENTION"
+        ] = True
+        layer_cfg.attention.inner_attention.kernel_options["SPLIT_KV"] = 1
```
tianyu-l: does this mean num_split = 1?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -73,6 +120,68 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
+        # match vllm flex block size
+        layer_cfg.attention.inner_attention.kernel_options["BLOCK_M"] = 16
+        layer_cfg.attention.inner_attention.kernel_options["BLOCK_N"] = 16
```
tianyu-l: why 16?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -304,12 +315,35 @@ async def forward_backward(self, train_data: list[TrainBatch]) -> dict:
...
+            attention_masks, maybe_padded_seq_lens = create_flex_block_mask(
+                seq_lens, device, self._flex_block_size
+            )
+            padded_ids, positions = pad_to_block_aligned(
```
tianyu-l: do we need this because we haven't implemented proper batcher for trainer?

tianyu-l (commented):
Thanks, left some questions to help me understand what is happening

torchtitan/experiments/rl/actors/trainer.py
liangel-02: i didn't use the existing function because it uses tokenizer and EOS token which isn't available in the RL loop. i didn't modify it cus i felt like it would be cleaner to keep it separate, but what do you guys think?

torchtitan/experiments/rl/actors/generator.py
liangel-02: oh mb, i forgot to remove this i was trying different ways to get bitwise parity earlier this is not needed

torchtitan/experiments/rl/config_registry.py
liangel-02: this is to force the trainer to use the standard flex attention kernel instead of the optimized decode one, vllm sets it [here](https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/flex_attention.py#L1140) so i added it so that it would match the generator. i think theres specific conditions in which the decode kernel would be triggered cc @drisspg

torchtitan/experiments/rl/config_registry.py
liangel-02: let me remove this this isnt used

torchtitan/experiments/rl/config_registry.py
liangel-02: this is to align block m and n with vLLM size https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/flex_attention.py#L1158-L1160

torchtitan/experiments/rl/actors/trainer.py
liangel-02: no this is because vLLM's kv cache table is organized per request so we need to match this on the trainer side by padding sequences so that they take up a full flex block size (or vLLM page size).

torchtitan/experiments/rl/config_registry.py
tianyu-l: how is different from specifying `block_size` https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L270

also 16 sounds very small, wouldn't it depend on vllm block size, which itself seems configurable?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L212

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: We should refactor the existing one from using `tokenzier.eos_id` to using `positions`. IIRC there was a WIP doing this, but I couldn't find anymore. Would you please take it and refactor altogether, for both varlen and flex?

related issue https://github.com/pytorch/torchtitan/issues/2716

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: hmmm why varlen doesn't need it but flex needs it?

torchtitan/experiments/rl/config_registry.py
liangel-02: this is the flex attention block size defined in vllm, i think the line you linked in the generator is for vllm kv cache page size. @drisspg is the block size smth we want to expose on the vllm side?

torchtitan/experiments/rl/actors/trainer.py
liangel-02: varlen doesnt have concept of "blocks" since everything is packed together

torchtitan/experiments/rl/actors/trainer.py
liangel-02: yes, let me do this in a separate pr

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: This doesn't answer the "root" question. Why varlen is able to achieve batch invariance without isolating each requests, but flex is not?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -8,8 +8,109 @@
...
+        def mask_mod(b, h, q_idx, kv_idx):
+            q_doc = doc_ids[q_idx]
+            q_local = local_pos[q_idx]
+            kv_local = local_pos[kv_idx]
+            q_len = original_seq_lens_tensor[q_doc]
+            return (
+                (q_doc == doc_ids[kv_idx])
+                & (q_local < q_len)
+                & (kv_local < q_len)
+                & (kv_local <= q_local)
```
drisspg: Nit; I feel like compsoing some of these via named funcs  e.g. `same_doc`, `causal` .. etc is helpful for maintenance

torchtitan/experiments/rl/tests/test_bitwise_parity.py
```diff
@@ -240,32 +241,71 @@ def compute_trainer_prefill_logprobs(model, token_ids, device):
...
-
-    results = []
-    for i, t in enumerate(input_tensors):
-        seq_len = t.shape[0]
-        seq_lps = log_probs[i, : seq_len - 1]
-        seq_lps = seq_lps.gather(1, t[1:seq_len].unsqueeze(-1)).squeeze(-1)
-        results.append(seq_lps)
+    seq_lens = [t.shape[0] for t in input_tensors]
+
+    if attn_backend == "flex":
```
drisspg, line 370: nit: this is a pretty big if else, idk if how people prefer this but I would ideally put this into a func

torchtitan/experiments/rl/tests/test_bitwise_parity.py
```diff
@@ -240,32 +241,71 @@ def compute_trainer_prefill_logprobs(model, token_ids, device):
...
+        from torchtitan.experiments.rl.actors.utils import (
+            create_flex_block_mask,
+            create_positions_from_seq_lens,
+            pad_to_block_aligned,
+            unpad_from_block_aligned,
+        )
+
+        flat = torch.cat(input_tensors)
+        attention_masks, maybe_padded_seq_lens = create_flex_block_mask(
```
drisspg: if you know hwo many generation steps it can very much be beneficial to preallocate and slice; https://github.com/meta-pytorch/gpt-fast/blob/6ecad9b5b6b987d17ac4303965545873d0192086/generate.py#L85

not sure how dynamic num  requests are though here which can make that alil more annoying / increase memory

torchtitan/experiments/rl/config_registry.py
drisspg: that is actually a staleish option, you can just direclty set the kernel backend, decode uses splitkv which can lead to non bitwise numerics we in fact should not force though if we dont need to since it is in general much slower

torchtitan/experiments/rl/config_registry.py
drisspg: Yeah so values need to basically match what page-size you are using in vllm

torchtitan/experiments/rl/actors/trainer.py
drisspg: This is due to how we parallelize work across CTAs. Varlen fundamentally is intra seqlen, when scheduling the kernel we basically launch (max m blocks, num_heads, num_batch) work so and therefore its not possible to have mixing for requests between sequences. This is not true for Flex-attention. Markus had a huge PR stack to do add 1 off api for this https://github.com/pytorch/pytorch/pull/173016 but it did not land

torchtitan/experiments/rl/config_registry.py
tianyu-l: how to "just direclty set the kernel backend"? I think it's fine to separate batch invariant behavior and faster behavior.

torchtitan/experiments/rl/config_registry.py
tianyu-l: iiuc we need to let flex block size | vllm page size -- how do we know what that size is? 16 sounds arbitrary (like, I would assume 2 also works?)

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: sorry I don't think I understand what's happening for varlen, and in particular why the following leads to varlen being able to achieve batch invariance with mixed requests

> when scheduling the kernel we basically launch (max m blocks, num_heads, num_batch) work so and therefore its not possible to have mixing for requests between sequences

could you use more layman terms?

torchtitan/experiments/rl/config_registry.py
drisspg: https://github.com/pytorch/pytorch/blob/9b6f47cda030e993b8f80bb854509dd0b48671e3/torch/nn/attention/flex_attention.py#L207-L219

torchtitan/experiments/rl/tests/test_bitwise_parity.py
```diff
@@ -240,32 +241,71 @@ def compute_trainer_prefill_logprobs(model, token_ids, device):
...
-    results = []
-    for i, t in enumerate(input_tensors):
-        seq_len = t.shape[0]
-        seq_lps = log_probs[i, : seq_len - 1]
-        seq_lps = seq_lps.gather(1, t[1:seq_len].unsqueeze(-1)).squeeze(-1)
-        results.append(seq_lps)
+    seq_lens = [t.shape[0] for t in input_tensors]
+
+    if attn_backend == "flex":
+        from torchtitan.experiments.rl.actors.utils import (
```
drisspg: why lazy imports why?

torchtitan/experiments/rl/actors/trainer.py
drisspg: the way you break up the problem for varlen on the gpu works enforces an invariant that you never mix tokens between different sequences. This means their relative ordering can not have an effect on each other's outputs

torchtitan/experiments/rl/config_registry.py
liangel-02: talked to @drisspg offline he confirmed this is fine, we just want the trainer side to match the generator in vllm.

torchtitan/experiments/rl/actors/trainer.py
liangel-02: on vllm side each request has its own independent kv cache but on trainer side everything is grouped together, so for each sequence we need to maintain the independence by padding so that it also takes up the whole block

torchtitan/experiments/rl/config_registry.py
liangel-02: spoke to driss offline, let me open a PR in vllm to expose this

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: would appreciate a meeting on this

wwwjn: So do we have a conclusion why we need pad to block size here? It worth have a note somewhere to explain

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -8,8 +8,109 @@
...
+    seq_lens: list[int],
+    device: torch.device,
+    block_size: int | tuple[int, int] = _DEFAULT_SPARSE_BLOCK_SIZE,
+) -> tuple[BlockMask, list[int]]:
+    original_seq_lens_tensor = torch.tensor(seq_lens, dtype=torch.int32, device=device)
+    maybe_padded_seq_lens = seq_lens
+    maybe_padded_seq_lens_tensor = original_seq_lens_tensor
+
+    if is_in_batch_invariant_mode():
+        block_size_ = block_size if isinstance(block_size, int) else block_size[0]
```
wwwjn: In `create_flex_block_mask` we already pad attention mask to align with block size when batch_invariant mode are enabled. And the in Trainer, we padded the token_ids to match attention_mask padding, is this correct?

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -194,3 +195,9 @@ def apply_non_moe_tp(
             # pyrefly: ignore [bad-argument-type]
             parallelize_plan=layer_plan,
         )
+
+        inner_attn = (
+            transformer_block.attention.inner_attention
+        )  # pyrefly: ignore [missing-attribute]
+        if hasattr(inner_attn, "_sharding_config") and inner_attn._sharding_config:
```
wwwjn: Why we need to check _sharding_config here?

torchtitan/experiments/rl/tests/test_bitwise_parity.py
```diff
@@ -105,28 +109,13 @@ def build_trainer_model(
...
-
-    model_spec.parallelize_fn(
-        model,
-        parallel_dims=parallel_dims,
-        training=config.trainer.training,
-        parallelism=parallelism,
-        compile_config=CompileConfig(enable=False),
-        ac_config=ActivationCheckpointConfig(mode="none"),
-        dump_folder="",
+    model = parallelize_qwen3(
```
wwwjn: This function should be updated in #3070, maybe need rebase

torchtitan/experiments/rl/config_registry.py
```diff
@@ -78,6 +125,64 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
+def rl_grpo_qwen3_0_6b_flex_batch_invariant() -> RLTrainer.Config:
+    """GRPO training config for Qwen3-0.6B with flex attention and batch invariance
+    for bitwise-identical numerics between trainer and generator (4 GPUs: 2 gen + 2 train).
+    """
+    spec = model_registry("0.6B", attn_backend="flex")
+    for layer_cfg in spec.model.layers:
+        layer_cfg.attention.inner_attention.kernel_options[
+            "FORCE_USE_FLEX_ATTENTION"
+        ] = True
+        layer_cfg.attention.inner_attention.kernel_options["BLOCK_M"] = 16
```
wwwjn: We should handle this in attention extra option logic in model code, not config registry? Ideally when user set batch-invariant mode on, these fields should always been set. 

Curious can we have simliar wrapper like https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/experiments/rl/models/attention.py#L178C12-L178C38, and we can intercept the forward() function to handle these extra configs and padding / unpadding?

torchtitan/experiments/rl/models/parallelize.py
liangel-02: iiuc if _sharding_config is set then the model is configured with TP so we need to use parallelize() to wrap Flex with local_map (since flex kernel only understands plain tensors)

torchtitan/experiments/rl/actors/utils.py
liangel-02: yep!

torchtitan/experiments/rl/actors/generator.py
tianyu-l: shall we remove?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -234,14 +241,20 @@ def __init__(
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=not config.cudagraph.enable,
             attention_config=AttentionConfig(
-                backend=AttentionBackendEnum.CUSTOM,
+                backend=AttentionBackendEnum.FLEX_ATTENTION
```
tianyu-l: Do we need to set both `AttentionConfig` and `os.environ["VLLM_ATTENTION_BACKEND"]`? E.g. can we remove the os.environ setting?

If we have to keep both, let's put `attention_backend = AttentionBackendEnum.FLEX_ATTENTION if ... else ...` at the same location above to reduce branching.

torchtitan/experiments/rl/models/parallelize.py
```diff
@@ -39,6 +39,7 @@ def parallelize_qwen3(
     parallelism: ParallelismConfig,
     compile_config: CompileConfig | None = None,
     has_position_id: bool = False,
+    **kwargs,
```
tianyu-l: accidental?

torchtitan/experiments/rl/models/parallelize.py
tianyu-l: wait, @wwwjn do we still need this file? We should use core for parallelize_fn

torchtitan/experiments/rl/config_registry.py
tianyu-l: We do use `BACKEND`, e.g. at https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/config_utils.py#L53

I saw 
> `# TODO: remove FORCE_USE_FLEX_ATTENTION once BACKEND is fully adopted.`

https://github.com/pytorch/pytorch/blob/9b6f47cda030e993b8f80bb854509dd0b48671e3/torch/nn/attention/flex_attention.py#L1713

Can we avoid `"FORCE_USE_FLEX_ATTENTION"` in torchtitan?

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: The conclusion is that we should move padding to batcher (which doesn't exist today), instead of put inside fwd_bwd.

Does it make sense for @wwwjn to create a batcher abstraction to unblock @liangel-02 ? We can start with very simple packing (no padding).

torchtitan/experiments/rl/config_registry.py
tianyu-l: I think we should consider provide a batch invariance model converter like https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama4/config_registry.py#L73
which, given a model config, traverse all flex attention modules and add the kernel options

torchtitan/experiments/rl/config_registry.py
```diff
@@ -20,19 +20,66 @@
     TrainingConfig,
 )
 from torchtitan.experiments.rl.actors.generator import SamplingConfig, VLLMGenerator
+
+_BATCH_INVARIANT_DEBUG = DebugConfig(batch_invariant=True, deterministic=True)
```
tianyu-l: fine to have this, but the position can be after all import?

tianyu-l (changes requested):
I think a prerequisite to this PR should be a batcher for RL.

torchtitan/experiments/rl/models/parallelize.py
wwwjn: This file has been removed and you can rebase to latest main

torchtitan/experiments/rl/config_registry.py
wwwjn: Sounds good, currently we have a lot of repeated configuration in our config_registry

torchtitan/experiments/rl/actors/trainer.py
wwwjn: I'm working on a quick RFC by comparing some RL framework on market, will put up a PR soon

torchtitan/experiments/rl/models/parallelize.py
tianyu-l: @wwwjn https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/models/parallelize.py

torchtitan/models/common/attention.py
```diff
@@ -227,6 +227,17 @@ def forward(
...
+        # In batch-invariant mode, pin Triton tile sizes to prevent the
+        # autotuner from picking input-dependent block dimensions.
+        kernel_options = self.kernel_options
+        if is_in_batch_invariant_mode():
+            kernel_options = {
+                "BLOCK_M": 16,
+                "BLOCK_N": 16,
+                "IS_DIVISIBLE": False,
+                **self.kernel_options,
+            }
```
tianyu-l: - Would this not be covered by the model config converters?
- Do block_m = block_n = 32 / 64 / etc. work? If so here we just need to assert, not set.
- what's each field for? please explain

torchtitan/experiments/rl/config_registry.py
```diff
@@ -26,19 +28,85 @@
...
+
+
+def rl_grpo_qwen3_0_6b_flex() -> RLTrainer.Config:
+    """GRPO training config for Qwen3-0.6B with flex attention (4 GPUs: 2 gen + 2 train)."""
+    spec = model_registry(
+        "0.6B",
+        attn_backend="flex",
+        converters=[
+            FlexKernelOptionsConverter.Config(
+                kernel_options={"SPLIT_KV": 1, "FORCE_USE_FLEX_ATTENTION": True},
```
tianyu-l: what are they for? This is not for batch invariance right?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -26,19 +28,85 @@
...
 from torchtitan.models.qwen3 import model_registry
+from torchtitan.protocols.model import ModelConfigConverter
 
+_BATCH_INVARIANT_DEBUG = DebugConfig(batch_invariant=True, deterministic=True)
 
-def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
-    """GRPO training config for Qwen3-0.6B (6 GPUs: 4 gen + 2 train)."""
-    group_size = 8
+
+class FlexKernelOptionsConverter(ModelConfigConverter):
```
tianyu-l: I'm not sure if we need to expose such flexibility. I was thinking we just need a `BatchInvarianceConverter` where we force a batch inv setup. Is there incentive to expose kernel_options in such a converter? They are already exposed when constructing the original flex attention module config.

torchtitan/experiments/rl/grpo.py
```diff
@@ -315,7 +315,22 @@ def __init__(self, config: Config):
...
+        use_flex = isinstance(inner_attn, FlexAttention.Config)
+        batch_invariant = config.trainer.debug.batch_invariant
+        batcher_kwargs: dict = {}
+        if use_flex and batch_invariant:
+            batcher_kwargs["batch_invariant"] = True
+            batcher_kwargs["block_size"] = (
+                inner_attn.block_size
+                if isinstance(inner_attn.block_size, int)
+                else inner_attn.block_size[0]
+            )
```
tianyu-l: Let's do this at config time, not in constructor.

E.g. the batch invariance converter could update Batcher, together with Flex kernel_options.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -316,7 +322,11 @@ def __init__(
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=not config.cudagraph.enable,
             attention_config=AttentionConfig(
-                backend=AttentionBackendEnum.CUSTOM,
+                backend=(
+                    AttentionBackendEnum.FLEX_ATTENTION
+                    if self._use_flex
+                    else AttentionBackendEnum.CUSTOM
+                ),
```
tianyu-l, line 328: Do we need to update both `os.environ["VLLM_ATTENTION_BACKEND"]` and `AttentionConfig`? Sounds suspicious. Can we kill one of them? cc @wwwjn

torchtitan/experiments/rl/observability/metrics/processor.py
```diff
@@ -55,6 +55,7 @@ class Config(Configurable.Config):
             default_factory=lambda: [
                 "loss/mean",
                 "loss/ratio/clipped_frac",
+                "bit_wise/logprob_diff/max",
```
tianyu-l, line 58: where is this logged?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -321,6 +336,20 @@ def _build_model(
...
+        mask_mod = and_masks(get_causal_mask_mod(), get_document_mask_mod(positions))
+        return create_attention_mask(
+            mask_mod,
+            B,
+            None,
+            S,
+            S,
+            BLOCK_SIZE=self._flex_block_size,
+            separate_full_blocks=not self._batch_invariant,
+        )
```
tianyu-l: can we not reuse https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder.py#L268?

torchtitan/experiments/rl/batcher.py
```diff
@@ -115,11 +115,20 @@ class Batcher(Configurable):
...
         batch: BatchConfig = field(default_factory=BatchConfig)
 
-    def __init__(self, config: Config, *, pad_id: int):
+    def __init__(
+        self,
+        config: Config,
+        *,
+        pad_id: int,
+        batch_invariant: bool = False,
+        block_size: int = 128,
```
tianyu-l: this should be in the config, as something like `per_sample_pad_multiple: int | None = None` and adjusted by `BatchInvarianceConverter`

torchtitan/experiments/rl/batcher.py
```diff
@@ -115,11 +115,20 @@ class Batcher(Configurable):
     class Config(Configurable.Config):
         batch: BatchConfig = field(default_factory=BatchConfig)
 
-    def __init__(self, config: Config, *, pad_id: int):
+    def __init__(
+        self,
+        config: Config,
+        *,
+        pad_id: int,
+        batch_invariant: bool = False,
```
tianyu-l: no need for this flag

torchtitan/experiments/rl/batcher.py
```diff
@@ -203,13 +212,28 @@ def batch(
 
         return microbatches, num_global_valid_tokens, packing_metrics
 
+    @staticmethod
+    def _align_to(length: int, alignment: int) -> int:
+        return ((length + alignment - 1) // alignment) * alignment
```
tianyu-l: maybe should be an inline helper

torchtitan/experiments/rl/observability/metrics/processor.py
tianyu-l: I noticed that the printed diff is 4.8e-07. Is it expected to be not 0, and why?

torchtitan/models/common/attention.py
tianyu-l: @wwwjn we should have a dedicated .md file for what we enabled to achieve batch invariance

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -286,9 +287,10 @@ def __init__(
         )
 
         # Set vLLM environment variables from config before any vLLM initialization
-        os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
-        os.environ["VLLM_USE_V2_MODEL_RUNNER"] = "1"
+        inner_attn = model_spec.model.layers[0].attention.inner_attention
+        self._use_flex = isinstance(inner_attn, FlexAttention.Config)
```
tianyu-l: Instead of introducing `_use_flex` and translate it to `backend` below, you could just set define `attention_backend = ` here with if-else.

Also worth erroring out when it's neither Flex nor Varlen? Not sure if we already error out somewhere.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -289,12 +289,13 @@ def _build_model(
...
-        # TODO: Also support flex attention backend later.
         from torchtitan.models.common.attention import VarlenAttention
 
+        inner_attn = model_spec.model.layers[0].attention.inner_attention
         assert isinstance(
-            model_spec.model.layers[0].attention.inner_attention, VarlenAttention.Config
-        ), "Only varlen attention backend is allowed."
+            model_spec.model.layers[0].attention.inner_attention,
+            (VarlenAttention.Config, FlexAttention.Config),
+        ), "Only varlen and flex attention backends are allowed."
```
tianyu-l, line 298: oh yeah, we are erroring out in trainer. I think worth doing it again in generator.

torchtitan/experiments/rl/config_registry.py
```diff
@@ -26,19 +28,94 @@
...
+
+    Sets FORCE_USE_FLEX_ATTENTION and fixed BLOCK_M/BLOCK_N on all
+    FlexAttention layers so the Triton kernel produces deterministic
+    results regardless of batch composition.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(ModelConfigConverter.Config):
+        block_m: int = 16
+        block_n: int = 16
```
tianyu-l: looks they don't have to be the same; as long as they are fixed, we are good, right?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -26,19 +28,94 @@
...
+        self.block_m = config.block_m
+        self.block_n = config.block_n
+
+    def convert(self, model_config) -> None:
+        for layer_cfg in model_config.layers:
+            inner = layer_cfg.attention.inner_attention
+            if isinstance(inner, FlexAttention.Config):
+                q_bs = inner.block_size if isinstance(inner.block_size, int) else inner.block_size[0]
+                kv_bs = inner.block_size if isinstance(inner.block_size, int) else inner.block_size[1]
+                if q_bs % self.block_m != 0 or kv_bs % self.block_n != 0:
```
tianyu-l: - what's the difference between `inner.block_size` and `block_m` / `block_n`
- what happens when they are both set, assuming compatible (but not necessarily the same)

torchtitan/experiments/rl/config_registry.py
```diff
@@ -86,6 +164,72 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
+def rl_grpo_qwen3_0_6b_flex_batch_invariant() -> RLTrainer.Config:
+    """GRPO training config for Qwen3-0.6B with flex attention and batch invariance
+    for bitwise-identical numerics between trainer and generator (4 GPUs: 2 gen + 2 train).
+    """
+    spec = model_registry(
+        "0.6B",
+        attn_backend="flex",
+        converters=[BatchInvariantFlexConverter.Config()],
+    )
+    return RLTrainer.Config(
```
tianyu-l: for length's sake, can we create a `rl_grpo_qwen3_0_6b_flex` and replace the `model_spec`?

torchtitan/models/common/attention.py
```diff
@@ -227,6 +227,8 @@ def forward(
         # Transpose to (bs, heads, seq, dim) for flex_attention
         q, k, v = q.transpose(1, 2), k.transpose(1, 2), v.transpose(1, 2)
 
+        kernel_options = self.kernel_options
```
tianyu-l: motivation for this change?

torchtitan/models/common/decoder.py
```diff
@@ -263,6 +264,7 @@ def _get_flex_attention_masks(
             seq_len,
             seq_len,
             BLOCK_SIZE=attn_config.inner_attention.block_size,
+            separate_full_blocks=not is_in_batch_invariant_mode(),
```
tianyu-l, line 272: From the docstring
> If True, fully unmasked blocks are stored
            separately so kernels can skip mask_mod on those blocks. If False,
            all non-empty blocks are stored as partial blocks and mask_mod is
            applied to every block.

It's not clear why we are turning it off when batch invariance is on. Could you explain + adding the explanation to comment in code?

torchtitan/experiments/rl/batcher.py
```diff
@@ -114,12 +118,17 @@ class Batcher(Configurable):
     @dataclass(kw_only=True, slots=True)
     class Config(Configurable.Config):
         batch: BatchConfig = field(default_factory=BatchConfig)
+        per_sample_pad_multiple: int = 0
```
tianyu-l: divisor should never default to 0 :)
```suggestion
        per_sample_pad_multiple: int | None = None
```

torchtitan/experiments/rl/batcher.py
```diff
@@ -17,6 +17,10 @@
 from torchtitan.experiments.rl.types import Episode, TrainingBatch
 
 
+def _align_to(length: int, alignment: int) -> int:
```
tianyu-l: nit: inline this helper into `_pack_episodes`

torchtitan/experiments/rl/config_registry.py
```diff
@@ -86,6 +164,72 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
...
+        num_steps=10,
+        num_prompts_per_step=5,
+        num_validation_samples=20,
+        env=SumDigitsEnv.Config(seed=42, correctness_reward=1.0, format_reward=0.3),
+        validation_env=SumDigitsEnv.Config(
+            seed=99, correctness_reward=1.0, format_reward=0.3
+        ),
+        batcher=Batcher.Config(
+            batch=BatchConfig(local_batch_size=2, global_batch_size=8, seq_len=2048),
+            per_sample_pad_multiple=128,
```
tianyu-l: Oh I see why https://github.com/pytorch/torchtitan/pull/3077#discussion_r3337865573 doesn't make sense. It was because the the converter only works on Model, not the entire job config.

It might be fine to explicitly set it in `Batcher`. But what's the guidance on this value `per_sample_pad_multiple` vs. block_m / block_n vs. inner_attention.block_size?

Could you start a BATCH_INVARIANce.md to document flex (and varlen) related things? @wwwjn can add to it later.

torchtitan/experiments/rl/config_registry.py
liangel-02: yea fixed and a divisor of logical block size (i added a check in titan + it gets checked in flex attention impl later), this is just the default

torchtitan/experiments/rl/config_registry.py
liangel-02: `inner.block_size` is the flex attention BlockMask block size and `block_m`/`block_n` are the triton kernel tile sizes. when theyre both set the constraint is that the kernel tiles must evenly divide the BlockMask blocks

tianyu-l: This is helpful, please add comment in code.

I'm assuming block_m / block_n would have an impact on perf. What's the trade off of larger / smaller block? What are the values of block_m / block_n if we don't set it explicitly -- I know they are not fixed, but what ranges make sense?

It's just not clear to me if it should be configurable in `FlexConfigConverter`. Maybe should default to 16 until someone asks.

torchtitan/experiments/rl/config_registry.py
liangel-02: `inner_attention.block_size` is the logical flex attention BlockMask block size so `inner_attention.block_size` should match `per_sample_pad_multiple` so that one document takes up one block.

torchtitan/experiments/rl/config_registry.py
liangel-02: yeah larger tile block size i think means more hardware utilization but maybe more wasted memory while smaller block sizes means more kernel launches but less wasted compute on partial blocks. 

the triton autotuner today picks a range between 64-128 for the fwd and 32-128 for the backward depending on input

i can default to 16 for now

tianyu-l: > larger tile block size i think means ... more wasted memory

why triton block size can affect memory?

> smaller block sizes means ... less wasted compute on partial blocks

I can imagine smaller Flex block size wastes less on partial blocks. But why smaller triton blocks matter? Within a Flex block, wouldn't every sub triton block be computed?

> i can default to 16 for now

If autotuner picks between 32-128, should we default to something in that range?

torchtitan/experiments/rl/models/attention.py
```diff
@@ -135,7 +135,7 @@ def forward(
         ), "Encoder-only attention not supported yet."
 
         # For decoder and cross-attention, use KV cache as before
-        key_cache, value_cache = kv_cache.unbind(0)
+        key_cache, value_cache = kv_cache.unbind(1)
```
tianyu-l: what happened?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -26,19 +29,89 @@
...
+    """GRPO training config for Qwen3-0.6B with flex attention (4 GPUs: 2 gen + 2 train)."""
+    spec = model_registry("0.6B", attn_backend="flex")
+    group_size = 8
+    return RLTrainer.Config(
+        model_spec=spec,
+        hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
+        num_steps=10,
         num_prompts_per_step=5,
         num_validation_samples=20,
-        compile=CompileConfig(enable=True, backend="aot_eager"),
```
tianyu-l, line 41: is this for debugging?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -53,7 +126,7 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
                 warmup_steps=2,
                 decay_type="linear",
             ),
-            training=TrainingConfig(),
+            training=TrainingConfig(dtype="bfloat16"),
```
tianyu-l, line 153: similar

torchtitan/experiments/rl/config_registry.py
```diff
@@ -69,8 +142,9 @@ def rl_grpo_qwen3_0_6b() -> RLTrainer.Config:
         ),
         generator=VLLMGenerator.Config(
             model_dtype="bfloat16",
+            gpu_memory_limit=0.7,
```
tianyu-l: why need this?

tianyu-l (commented):
Code-wise LGTM!

It seems there are some debugging changes remaining, I can stamp after everything settles.

torchtitan/experiments/rl/models/attention.py
liangel-02: oh sorry this is to be compatible w/ local vllm version let me revert

torchtitan/experiments/rl/config_registry.py
liangel-02: dtype needs to be bf16 to match generator dtype

tianyu-l: @wwwjn you left a TODO here for varlen batch invariance mode. Is it done?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/config_registry.py#L232-L234

torchtitan/experiments/rl/config_registry.py
tianyu-l: From @liangel-02 :
> if i use aot_eager the numerics don't match and if i use default inductor i get a vllm error so i removed it

For Flex I guess it's because we don't have the "scooping Flex inductor region in aot_eager" in torchtitan? cc @SherlockNoMad 

But why did you remove it for varlen as well?

liangel-02: fixed varlen

tianyu-l (approved):
stamp, but please follow up on
- adding issue and TODO for Flex + aot_eager compile backend
- further numerics debugging

torchtitan/experiments/rl/config_registry.py
wwwjn: Previously I started a PR but needs to rebase / followup on it: https://github.com/pytorch/torchtitan/pull/2932/changes. Will request review once I update it

torchtitan/models/common/attention.py
wwwjn: Add a TODO. We have the same discussion in https://github.com/pytorch/torchtitan/pull/2932/changes this PR, will pick up this PR soon.

PR #3081, opened 2026-04-24
Add MoE loss comparison guard to CI workflow

wwwjn:
Add CI step to verify Qwen3 MoE deterministic losses match a reference file, following the same pattern as the dense llama3 loss comparison. 

Uses FSDP8 on 8 GPUs with deterministic seed=42 and seed checkpoint.

TODO: Generate reference loss files on CI machines:
  # CUDA (8×A10G or H100):
  python3 scripts/loss_compare.py . . \
    --baseline-module=qwen3 --baseline-config=qwen3_moe_debug \
    --baseline-options="--compile.no-enable --activation_checkpoint.mode=none" \
    --export-result=tests/assets/losses/qwen3_moe_cuda.txt --steps=100

  # ROCm (8×MI350X):
  python3 scripts/loss_compare.py . . \
    --baseline-module=qwen3 --baseline-config=qwen3_moe_debug \
    --baseline-options="--compile.no-enable --activation_checkpoint.mode=none" \
    --export-result=tests/assets/losses/qwen3_moe_rocm_mi350x.txt --steps=100


wwwjn:
This PR is ready to be reviewed, but I want to discuss for MoE , which parallelism settings we should guard numerics on. Based on my understanding , run-to-run deterministic can be also achieved with FSDP + EP, given FSDP is already run-to-run deterministic.  cc @fegin @tianyu-l

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
       docker-image: ${{ matrix.docker-image }}
       repository: pytorch/torchtitan
       upload-artifact: outputs
-      timeout: 45
+      timeout: 60
```
tianyu-l, line 49: why this change?

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -122,6 +122,22 @@ jobs:
         python3 scripts/loss_compare.py . . --baseline-options="${baseline_options}" --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/accuracy_comparison_outputs" --import-result="${LOSS_FILE}" --assert-equal --steps=100
         rm -rf $RUNNER_TEMP/artifacts-to-be-uploaded/*
 
+        # MoE loss comparison: verify Qwen3 MoE with FSDP+EP deterministic losses match reference
```
tianyu-l: not sure which is better, but we are only testing dense llama3 in features.yaml, so I had thought we can put in model test to balance. I'm not sure which is slower today though.

tianyu-l (commented):
I think we should test dense fsddp 4 + tp 2 to sparse efsdp 2 + ep 4

wwwjn:
> I think we should test dense fsddp 4 + tp 2 to sparse efsdp 2 + ep 4

That sounds comprehensive, but curious why dense model doesn't test TP in loss comparison? Is this because of TP all-reduce (but it would be RS when SP enabled, and RS is already used in FSDP backward, so RS should be run-to-run determinstic). I ask Claude and given the same communicator topology and message sizes, the NCCL communication are deterministic

tianyu-l:
> That sounds comprehensive, but curious why dense model doesn't test TP in loss comparison?

We should. We should guard on all FSDP, TP, CP, PP

.github/workflows/integration_test_8gpu_features.yaml
wwwjn: I took a look at Features / Model CI, both of them takes ~30 mins, so adding to features/ models doesn't make too much difference here

tianyu-l: OK. Can we revert this to 45?

tianyu-l (approved):

PR #3089, opened 2026-04-24
[graph_trainer][aot_fx_trace][graph_pp] Add GraphPP runner integration

### LINKED ISSUE #3780: [graph_trainer] Graph-PP RFC

sanketpurandare:
Stacked PRs:
 * #3585
 * #3090
 * __->__#3089
 * #3088
 * #2727
 * #2726

--- --- ---

### [graph_trainer][aot_fx_trace][graph_pp] Add GraphPP runner integration

Add the GraphPP pipeline entry point, GraphPipelineStage wrapper, and GraphPipelineRuntime schedule wrapper that executes prebuilt stage-local graph executors for PyTorch runtime PP schedule actions. Graph construction uses upstream PP metadata initialization, GraphTrainer tracing, the metadata-preserving compile-time pass pipeline without terminal Inductor compilation, and the GraphPP partition/FSDP/dI-dW extraction passes.

Reuse GraphTrainer tracer subclass wrapping for values crossing PP boundaries, keep internal GraphPP calling-convention values flat, avoid eager pre-trace forwards, normalize PP-split BlockMask inputs, and share boxed graph execution plus explicit parameter-gradient accumulation helpers with the non-PP path. OVERLAP_F_B is registered but intentionally unsupported until the multiplex/DualPipeV follow-up commit.

Add runner unit coverage for graph construction, forward/backward execution, FSDP unshard/reduce actions, BlockMask handling, metadata contracts, compile hooks, and supported runtime schedule validation. Add DeepSeek V3 GraphPP numerics and integration coverage for Interleaved1F1B and ZBVZeroBubble.

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -6,8 +6,9 @@
 
 import fnmatch
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from contextlib import contextmanager
+from typing import Any, TypeAlias
 
 import torch
 import torch.nn as nn
@@ -24,6 +25,11 @@ from torchtitan.experiments.graph_trainer.simple_fsdp import (
 )
 from torchtitan.models.common.decoder import Decoder, TransformerBlock
 from torchtitan.tools.logging import logger
+
+
+BOXED_CODEGEN_META = "graph_trainer_boxed_codegen"
+LossResult: TypeAlias = torch.Tensor | tuple[torch.Tensor, dict[str, Any]]
+AnnotatedLossFn: TypeAlias = Callable[..., LossResult]
 
 
 @contextmanager
@@ -89,12 +95,86 @@ def _maybe_materialize_grad_for_param_layout(
     return materialized_grad
 
 
+def set_graph_module_boxed_codegen(
+    gm: torch.fx.GraphModule,
+    *,
+    boxed: bool,
+) -> torch.fx.GraphModule:
+    """Set the FX calling convention and record it in graph metadata.
+
+    Args:
+        gm: Graph module whose generated Python wrapper should be updated.
+        boxed: Whether the wrapper should accept one mutable argument list and
+            clear it after placeholder extraction.
+
+    Returns:
+        The same graph module after recompilation if the calling convention
+        changed.
+    """
+
+    if gm.meta.get(BOXED_CODEGEN_META) is boxed:
+        return gm
+    codegen = torch.fx.graph._BoxedCodeGen() if boxed else torch.fx.graph.CodeGen()
+    gm.graph.set_codegen(codegen)
+    gm.recompile()
+    gm.meta[BOXED_CODEGEN_META] = boxed
+    return gm
+
+
+def ensure_boxed_graph_module(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
+    """Use boxed FX codegen for runtime-owned graph modules.
+
+    Args:
+        gm: Graph module to box.
+
+    Returns:
+        The same graph module with boxed FX codegen enabled.
+    """
+
+    return set_graph_module_boxed_codegen(gm, boxed=True)
+
+
 _MODULE_FQN = "module_fqn"
 _EP_TOKEN_COUNT_EXCHANGE = "EP_token_count_exchange"
 _EP_TOKEN_COUNT_SYNC = "EP_token_count_sync"
 _EP_TOKEN_EXCHANGE = "EP_token_exchange"
 _EP_TOKEN_EXCHANGE_WAIT = "EP_token_exchange_wait"
 _NOT_IN_LAYERS = -1
+
+
+def compute_annotated_loss(
+    loss_fn: AnnotatedLossFn,
+    pred: torch.Tensor,
+    labels: torch.Tensor,
+    loss_kwargs: dict[str, Any] | None = None,
+) -> torch.Tensor:
+    """Compute the loss tensor with the same FX metadata convention as GraphTrainer."""
+    annotated_loss_fn = annotate_fn({_MODULE_FQN: "loss"})(loss_fn)
+    result = annotated_loss_fn(pred, labels, **(loss_kwargs or {}))
+    if isinstance(result, tuple):
+        if len(result) != 2:
+            raise ValueError(
+                "GraphTrainer loss functions must return a loss tensor or "
+                "(loss tensor, metrics)."
+            )
+        loss, _metrics = result
+        return loss
+    return result
+
+
+def accumulate_param_grads_(
+    params: Iterable[torch.Tensor],
+    grads: Iterable[torch.Tensor | None],
+) -> None:
+    """Accumulate explicit graph-produced gradients into live parameters."""
+    for param, grad in zip(params, grads, strict=True):
+        if grad is None:
+            continue
+        grad = _maybe_materialize_grad_for_param_layout(param, grad)
+        if param.grad is None:
+            param.grad = grad
+        else:
+            param.grad += grad
 
 
 def _is_backward_node(node: torch.fx.Node) -> bool:
--- a/torchtitan/experiments/graph_trainer/deepseek_v3/config_registry.py
+++ b/torchtitan/experiments/graph_trainer/deepseek_v3/config_registry.py
@@ -4,6 +4,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from dataclasses import replace
+
+from torchtitan.distributed.pipeline_parallel import pipeline_llm
 from torchtitan.experiments.graph_trainer.configs import (
     GraphTrainerCompileConfig,
     to_graph_trainer_config,
@@ -43,6 +46,18 @@ def graph_trainer_deepseek_v3_debugmodel_minimal_async_ep() -> GraphTrainer.Conf
         model_registry,
     )
     config.compile = GraphTrainerCompileConfig(enable=True)
+    return config
+
+
+def graph_trainer_deepseek_v3_debugmodel_eager_pp() -> GraphTrainer.Config:
+    """Test-only FlexAttention baseline that runs through eager pipeline parallelism."""
+    config = graph_trainer_deepseek_v3_debugmodel()
+    config.compile = GraphTrainerCompileConfig(
+        enable=True,
+        components=["loss"],
+        mode=None,
+    )
+    config.model_spec = replace(config.model_spec, pipelining_fn=pipeline_llm)
     return config
 
 
--- a/torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
+++ b/torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
@@ -1,0 +1,1352 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+"""GraphTrainer-backed graph construction for GraphPP stages.
+
+Flat calling convention and wrapping contract:
+1. Extracted graphs execute on the flat values produced by
+   ``minimal_fx_tracer``. Tensor subclasses are unwrapped into plain leaves by
+   the tracer before FX execution.
+2. Only values that cross the PP/runtime boundary are rewrapped: stage forward
+   outputs, input gradients sent to the previous stage, and parameter gradients
+   before assigning to live ``param.grad``.
+3. Internal graph values stay flat because they never escape GraphPP graph
+   execution: saved-for-backward values, unsharded FSDP params, raw grad
+   leaves, and reduce-grad inputs.
+4. DTensor and other traceable tensor subclasses use the existing tracer layout
+   metadata. GraphPP must not add a separate DTensor-specific wrapping path.
+"""
+
+import dataclasses
+import types
+import warnings
+from collections.abc import Callable
+from typing import Any, cast
+
+import torch
+import torch.fx as fx
+import torch.utils._pytree as pytree
+from torch.distributed.pipelining.schedules import (
+    _PipelineContext,
+    _PipelineScheduleRuntime,
+    OVERLAP_F_B,
+)
+
+from torchtitan.config import ParallelismConfig
+from torchtitan.experiments.graph_trainer.common_utils import (
+    BOXED_CODEGEN_META,
+    compute_annotated_loss,
+    ensure_boxed_graph_module,
+    maybe_register_blockmask_pytree_node,
+)
+from torchtitan.experiments.graph_trainer.configs import GraphTrainerCompileConfig
+from torchtitan.experiments.graph_trainer.graph_pp.split_fsdp_collectives import (
+    split_backward_fsdp_collectives,
+    split_forward_fsdp_collectives,
+)
+from torchtitan.experiments.graph_trainer.graph_pp.partition import (
+    GraphMeta as PartitionGraphMeta,
+    partition_joint_graph,
+)
+from torchtitan.experiments.graph_trainer.graph_pp.split_di_dw import (
+    GraphPPDiDwSplit,
+    split_di_dw_graph,
+)
+from torchtitan.experiments.graph_trainer.graph_pp.stage import (
+    GraphPPOverlapGraphs,
+    GraphPPStageGraphs,
+    GraphPipelineStage,
+)
+from torchtitan.experiments.graph_trainer.graph_pp.utils import (
+    example_inputs_from_placeholders,
+    flatten_graph_values,
+    GraphPPValueSpec,
+    graph_pp_value_spec,
+    normalize_graph_pp_microbatch_inputs,
+    overlap_fw_bw_sub_actions,
+)
+from torchtitan.experiments.graph_trainer.make_fx_tracer import (
+    extract_module_state,
+    minimal_fx_tracer,
+    TracedResult,
+)
+from torchtitan.experiments.graph_trainer.passes import (
+    apply_graph_passes,
+    canonicalize_graph_pass,
+    compile_time_passes,
+    deduplicate_fsdp_unshard_chains_pass,
+    eliminate_dead_code_pass,
+    final_inductor_compile_passes,
+)
+from torchtitan.protocols.model import BaseModel
+from torchtitan.tools.logging import logger
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class _GraphTrainerPassConfigView:
+    """Minimal config-shaped view needed by reused GraphTrainer pass builders.
+
+    GraphPP is entered through TorchTitan's generic pipelining function API,
+    which passes decomposed config fields instead of the full
+    ``GraphTrainer.Config``. The pre-partition GraphTrainer passes read only
+    ``compile``, ``parallelism``, and ``model_spec.model``, so GraphPP exposes
+    exactly those fields instead of synthesizing a fake full trainer config.
+    """
+
+    compile: GraphTrainerCompileConfig
+    parallelism: ParallelismConfig
+    model_spec: types.SimpleNamespace
+
+
+@dataclasses.dataclass(slots=True)
+class _StageGraphModules:
+    """FX graph modules produced by GraphTrainer stage graph construction."""
+
+    fw: fx.GraphModule
+    full_bw: fx.GraphModule
+    bw_di: fx.GraphModule | None = None
+    bw_dw: fx.GraphModule | None = None
+    unshard: fx.GraphModule | None = None
+    reduce_grad: fx.GraphModule | None = None
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class _StageGraphMeta:
+    """GraphTrainer metadata for one bound GraphPP stage graph executor.
+
+    The fields describe how the flat FX graph signatures map to the PP calling
+    convention:
+
+    1. Forward: ``fwd_input_names`` and ``fwd_flat_input_indices`` select the
+       parameter, buffer, user-input, target, and loss-kwarg leaves consumed by
+       the forward graph.
+    2. Saved values: ``num_saved_for_backward`` counts forward outputs that are
+       hidden from PP users but fed to the backward graph.
+    3. Backward: ``partition`` names saved values and output gradients from the
+       next stage so runtime can pack backward placeholders deterministically.
+    4. Grad outputs: ``param_grad_values`` and ``input_grad_values`` describe
+       how flat graph outputs rewrap into parameter grads and input grads sent
+       to the previous stage.
+    5. FSDP edges: ``unshard_flat_param_indices`` and
+       ``reduce_grad_input_names`` bind optional ``UNSHARD`` and ``REDUCE_GRAD``
+       graphs to the same flat calling convention.
+
+    Counts ending in ``_values`` refer to flat graph values, not privacy. This
+    metadata is private to ``GraphTrainerStageGraphs``; the PP runtime must not
+    inspect it.
+    """
+
+    num_user_outputs: int
+    num_saved_for_backward: int
+    num_param_grad_values: int
+    num_input_grad_values: int
+    num_flat_param_values: int
+    fwd_output_values: GraphPPValueSpec
+    param_grad_values: GraphPPValueSpec
+    input_grad_values: GraphPPValueSpec
+    partition: PartitionGraphMeta
+    fwd_input_names: tuple[str, ...]
+    fwd_flat_input_indices: tuple[int, ...]
+    bw_no_fsdp_output_names: tuple[str, ...] = ()
+    reduce_grad_input_names: tuple[str, ...] = ()
+    unshard_flat_param_indices: tuple[int, ...] = ()
+    num_fw_unsharded_param_inputs: int = 0
+    is_last_stage: bool = False
+
+
+def _execute_graph_module(
+    gm: fx.GraphModule,
+    args: list[Any],
+) -> tuple[Any, ...]:
+    """Execute one boxed FX graph module and normalize its result to a tuple."""
+
+    with torch.no_grad():
+        outputs = gm(args)
+    if args:
+        raise ValueError(
+            "GraphPP graph call expected boxed FX codegen to clear its mutable "
+            f"argument list, but {len(args)} entries remain."
+        )
+    if isinstance(outputs, tuple):
+        return outputs
+    if isinstance(outputs, list):
+        return tuple(outputs)
+    return (outputs,)
+
+
+@dataclasses.dataclass(slots=True)
+class GraphTrainerStageGraphs(GraphPPStageGraphs):
+    """GraphTrainer-backed bound graph executor for one GraphPP stage.
+
+    The object owns the GraphTrainer-specific FX modules and metadata needed to
+    pack flat graph inputs and unwrap graph outputs. ``GraphPipelineRuntime``
+    calls this object through the generic ``GraphPPStageGraphs`` protocol and
+    never inspects the private metadata directly.
+
+    Args:
+        modules: Stage-local FX graph modules after GraphPP graph passes.
+        meta: GraphTrainer calling-convention metadata for those modules.
+        compiled: Whether the FX modules have already been compiled.
+    """
+
+    modules: _StageGraphModules
+    meta: _StageGraphMeta
+    compiled: bool = False
+
+    def __post_init__(self) -> None:
+        num_unsharded_inputs = self.meta.num_fw_unsharded_param_inputs
+        if num_unsharded_inputs > self.meta.num_flat_param_values:
+            raise ValueError(
+                "GraphPP forward graph needs more unsharded params than "
+                f"metadata has: {num_unsharded_inputs} > "
+                f"{self.meta.num_flat_param_values}"
+            )
+        if len(self.meta.fwd_input_names) != (
+            num_unsharded_inputs + len(self.meta.fwd_flat_input_indices)
+        ):
+            raise ValueError(
+                "GraphPP forward input metadata must be an unsharded-param "
+                "prefix followed by traced flat input indices: "
+                f"names={self.meta.fwd_input_names}, "
+                f"num_unsharded={num_unsharded_inputs}, "
+                f"flat_indices={self.meta.fwd_flat_input_indices}"
+            )
+
+    @property
+    def supports_backward_input_weight_split(self) -> bool:
+        return self.modules.bw_di is not None and self.modules.bw_dw is not None
+
+    @property
+    def num_unsharded_param_grad_values(self) -> int:
+        return self.meta.num_param_grad_values
+
+    def unshard_params(
+        self,
+        flat_param_values: list[Any],
+        *,
+        runtime_validate: bool = False,
+    ) -> list[Any]:
+        """Run the optional FSDP unshard graph.
+
+        ``flat_param_values`` is the live stage parameter list flattened with
+        the tracer's subclass rules. The unshard graph consumes only the flat
+        parameters that own an all-gather chain and returns one flat value for
+        every original parameter input. Replicated parameters pass through
+        unchanged so later forward calls can use a uniform parameter prefix.
+        """
+
+        if (
+            runtime_validate
+            and len(flat_param_values) != self.meta.num_flat_param_values
+        ):
+            raise ValueError(
+                "GraphPP unshard expected one runtime value per flat param: "
+                f"{len(flat_param_values)} != {self.meta.num_flat_param_values}"
+            )
+        if self.modules.unshard is None:
+            return list(flat_param_values)
+        unshard_args = []
+        for param_index in self.meta.unshard_flat_param_indices:
+            if runtime_validate and (
+                param_index < 0 or param_index >= len(flat_param_values)
+            ):
+                raise ValueError(
+                    "GraphPP unshard parameter index is out of range: "
+                    f"index {param_index}, but runtime has "
+                    f"{len(flat_param_values)} flat params"
+                )
+            unshard_args.append(flat_param_values[param_index])
+        unsharded_param_values = list(
+            _execute_graph_module(self.modules.unshard, unshard_args)
+        )
+        if (
+            runtime_validate
+            and len(unsharded_param_values) != self.meta.num_flat_param_values
+        ):
+            raise ValueError(
+                "GraphPP unshard graph output count must match flat parameter "
+                "count: "
+                f"{len(unsharded_param_values)} != {self.meta.num_flat_param_values}"
+            )
+        return unsharded_param_values
+
+    def _flat_user_forward_inputs(
+        self,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+        target: Any,
+        loss_kwargs: dict[str, Any],
+    ) -> list[Any]:
+        """Flatten the non-state runtime inputs supplied by PP for forward."""
+
+        if self.meta.is_last_stage:
+            flat_user_inputs, _ = pytree.tree_flatten(
+                ((args, kwargs, target, loss_kwargs), {})
+            )
+        else:
+            flat_user_inputs, _ = pytree.tree_flatten(((args, kwargs), {}))
+        return flatten_graph_values(list(flat_user_inputs))
+
+    def _forward_args(
+        self,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+        target: Any,
+        loss_kwargs: dict[str, Any],
+        *,
+        unsharded_param_values: list[Any],
+        flat_buffer_values: list[Any],
+        runtime_validate: bool = False,
+    ) -> list[Any]:
+        """Pack the extracted forward graph inputs in placeholder order."""
+
+        num_unsharded_inputs = self.meta.num_fw_unsharded_param_inputs
+        if (
+            runtime_validate
+            and len(unsharded_param_values) != self.meta.num_flat_param_values
+        ):
+            raise ValueError(
+                "GraphPP forward expected one unsharded value per flat param: "
+                f"{len(unsharded_param_values)} != {self.meta.num_flat_param_values}"
+            )
+
+        flat_user_inputs = self._flat_user_forward_inputs(
+            args,
+            kwargs,
+            target,
+            loss_kwargs,
+        )
+        flat_inputs = [
+            *unsharded_param_values,
+            *flat_buffer_values,
+            *flat_user_inputs,
+        ]
+        # Forward placeholders are a prefix of unsharded parameter values
+        # followed by explicit indices into params, buffers, and user inputs.
+        fw_args = list(unsharded_param_values[:num_unsharded_inputs])
+        flat_input_names = self.meta.fwd_input_names[num_unsharded_inputs:]
+        for name, flat_index in zip(
+            flat_input_names,
+            self.meta.fwd_flat_input_indices,
+            strict=True,
+        ):
+            if runtime_validate and (
+                flat_index < 0 or flat_index >= len(flat_inputs)
+            ):
+                raise ValueError(
+                    "GraphPP forward placeholder index is out of range: "
+                    f"{name} indexes {flat_index}, but runtime has "
+                    f"{len(flat_inputs)} flattened inputs"
+                )
+            fw_args.append(flat_inputs[flat_index])
+        return fw_args
+
+    def _split_forward_outputs(
+        self,
+        fw_outputs: tuple[Any, ...],
+    ) -> tuple[Any, tuple[Any, ...]]:
+        """Split user-visible outputs from saved values for backward."""
+
+        user_outputs = fw_outputs[: self.meta.num_user_outputs]
+        saved_start = self.meta.num_user_outputs
+        saved_end = saved_start + self.meta.num_saved_for_backward
+        saved_values_for_backward = tuple(fw_outputs[saved_start:saved_end])
+        output = self.meta.fwd_output_values.unflatten(user_outputs)
+        return output, saved_values_for_backward
+
+    def forward(
+        self,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+        target: Any,
+        loss_kwargs: dict[str, Any],
+        *,
+        unsharded_param_values: list[Any],
+        flat_buffer_values: list[Any],
+        runtime_validate: bool = False,
+    ) -> tuple[Any, tuple[Any, ...]]:
+        """Return ``(stage_output, saved_values_for_backward)``."""
+
+        fw_args = self._forward_args(
+            args,
+            kwargs,
+            target,
+            loss_kwargs,
+            unsharded_param_values=unsharded_param_values,
+            flat_buffer_values=flat_buffer_values,
+            runtime_validate=runtime_validate,
+        )
+        placeholders = self.modules.fw.graph.find_nodes(op="placeholder")
+        if runtime_validate and len(fw_args) != len(placeholders):
+            raise ValueError(
+                "GraphPP forward graph input mismatch: "
+                f"expected {len(placeholders)} args, got {len(fw_args)}. "
+                f"Placeholders: {[node.name for node in placeholders]}. "
+                f"Graph inputs: {list(self.meta.fwd_input_names)}."
+            )
+        return self._split_forward_outputs(
+            _execute_graph_module(self.modules.fw, fw_args)
+        )
+
+    def _backward_args(
+        self,
+        stage_output: tuple[Any, ...],
+        saved_values_for_backward: tuple[Any, ...],
+        output_grads_from_next: tuple[Any, ...],
+        *,
+        runtime_validate: bool = False,
+    ) -> list[Any]:
+        """Pack backward graph inputs from saved tensors and next-stage grads."""
+
+        if runtime_validate and self.meta.is_last_stage:
+            if len(stage_output) != 1:
+                raise ValueError(
+                    "GraphPP last stage backward expects the traced forward "
+                    f"graph to return one loss tensor, got {len(stage_output)} "
+                    "outputs."
+                )
+            if output_grads_from_next:
+                raise ValueError(
+                    "GraphPP last stage backward must not receive "
+                    "output_grads_from_next."
+                )
+        raw_output_grads_from_next = flatten_graph_values(
+            list(output_grads_from_next)
+        )
+        # The partitioner names every backward placeholder. At runtime those
+        # placeholders are supplied either by forward-saved values or by the
+        # output gradients received from the next PP stage.
+        saved_by_name = dict(
+            zip(
+                self.meta.partition.saved_for_backward_names,
+                saved_values_for_backward,
+                strict=True,
+            )
+        )
+        backward_grad_by_name = dict(
+            zip(
+                self.meta.partition.backward_grad_input_names,
+                [
+                    raw_output_grads_from_next[index]
+                    for index in self.meta.partition.backward_grad_input_indices
+                ],
+                strict=True,
+            )
+        )
+        bwd_args = []
+        for name in self.meta.partition.bwd_input_names:
+            if name in saved_by_name:
+                bwd_args.append(saved_by_name[name])
+            elif name in backward_grad_by_name:
+                bwd_args.append(backward_grad_by_name[name])
+            else:
+                raise ValueError(f"Missing GraphPP backward input {name}")
+        return bwd_args
+
+    def _split_full_backward_outputs(
+        self,
+        bw_outputs: tuple[Any, ...],
+    ) -> tuple[list[Any], list[Any]]:
+        """Split ``full_bw`` outputs into input grads and param grads."""
+
+        num_param_grads = self.meta.num_param_grad_values
+        param_grads = list(bw_outputs[:num_param_grads])
+        input_grads = self.meta.input_grad_values.wrap_flat_values(
+            bw_outputs[num_param_grads:]
+        )
+        return input_grads, param_grads
+
+    def full_backward(
+        self,
+        stage_output: tuple[Any, ...],
+        saved_values_for_backward: tuple[Any, ...],
+        output_grads_from_next: tuple[Any, ...],
+        *,
+        runtime_validate: bool = False,
+    ) -> tuple[list[Any], list[Any]]:
+        """Return ``(input_grads_to_prev, unsharded_param_grads)``."""
+
+        return self._split_full_backward_outputs(
+            _execute_graph_module(
+                self.modules.full_bw,
+                self._backward_args(
+                    stage_output,
+                    saved_values_for_backward,
+                    output_grads_from_next,
+                    runtime_validate=runtime_validate,
+                ),
+            )
+        )
+
+    def backward_input(
+        self,
+        stage_output: tuple[Any, ...],
+        saved_values_for_backward: tuple[Any, ...],
+        output_grads_from_next: tuple[Any, ...],
+        *,
+        runtime_validate: bool = False,
+    ) -> tuple[list[Any], tuple[Any, ...]]:
+        """Return input grads and saved values for later weight backward."""
+
+        if self.modules.bw_di is None:
+            raise ValueError("GraphPP stage does not have a backward-input graph")
+        outputs = _execute_graph_module(
+            self.modules.bw_di,
+            self._backward_args(
+                stage_output,
+                saved_values_for_backward,
+                output_grads_from_next,
+                runtime_validate=runtime_validate,
+            ),
+        )
+        input_grad_outputs = outputs[: self.meta.num_input_grad_values]
+        saved_values_for_backward_weight = outputs[self.meta.num_input_grad_values :]
+        input_grads = self.meta.input_grad_values.wrap_flat_values(input_grad_outputs)
+        return input_grads, tuple(saved_values_for_backward_weight)
+
+    def backward_weight(
+        self,
+        saved_values_for_backward_weight: tuple[Any, ...],
+    ) -> list[Any]:
+        """Return flat unsharded parameter gradients."""
+
+        if self.modules.bw_dw is None:
+            raise ValueError("GraphPP stage does not have a backward-weight graph")
+        return list(
+            _execute_graph_module(
+                self.modules.bw_dw,
+                list(saved_values_for_backward_weight),
+            )
+        )
+
+    def _validate_unsharded_param_grads(
+        self,
+        unsharded_param_grads: list[Any],
+        *,
+        runtime_validate: bool = False,
+    ) -> list[Any]:
+        raw_grads = list(unsharded_param_grads)
+        if (
+            runtime_validate
+            and len(raw_grads) != self.meta.num_param_grad_values
+        ):
+            raise ValueError(
+                "GraphPP raw unsharded grad count mismatch: "
+                f"expected {self.meta.num_param_grad_values}, got "
+                f"{len(raw_grads)}"
+            )
+        return raw_grads
+
+    def reduce_grads(
+        self,
+        unsharded_param_grads: list[Any],
+        *,
+        runtime_validate: bool = False,
+    ) -> list[Any]:
+        """Run the optional FSDP reduce-grad graph."""
+
+        raw_grads = self._validate_unsharded_param_grads(
+            unsharded_param_grads,
+            runtime_validate=runtime_validate,
+        )
+        if self.modules.reduce_grad is None:
+            return raw_grads
+        # ``bw_no_fsdp`` returns one raw grad slot per trainable parameter.
+        # ``reduce_grad`` consumes the subset/name order selected by the FSDP
+        # split pass and returns the original sharded/reduced grad slots.
+        grad_values_by_name = dict(
+            zip(
+                self.meta.bw_no_fsdp_output_names[: self.meta.num_param_grad_values],
+                raw_grads,
+                strict=True,
+            )
+        )
+        reduce_grad_args = [
+            grad_values_by_name[name] for name in self.meta.reduce_grad_input_names
+        ]
+        return list(_execute_graph_module(self.modules.reduce_grad, reduce_grad_args))
+
+    def param_grads_for_accumulation(
+        self,
+        sharded_param_grads: list[Any],
+    ) -> list[Any]:
+        """Return parameter gradients in optimizer accumulation structure."""
+
+        return self.meta.param_grad_values.wrap_flat_values(sharded_param_grads)
+
+
+@dataclasses.dataclass(slots=True)
+class GraphTrainerOverlapGraphs(GraphPPOverlapGraphs):
+    """GraphTrainer-backed executor for one multiplexed ``OVERLAP_F_B`` pair."""
+
+    fw_graphs: GraphTrainerStageGraphs
+    bw_graphs: GraphTrainerStageGraphs
+    multiplexed_graph: fx.GraphModule
+
+    def forward_backward(
+        self,
+        *,
+        backward_stage_output: tuple[Any, ...],
+        backward_saved_values_for_backward: tuple[Any, ...],
+        output_grads_from_next: tuple[Any, ...],
+        forward_args: tuple[Any, ...],
+        forward_kwargs: dict[str, Any],
+        forward_target: Any,
+        forward_loss_kwargs: dict[str, Any],
+        forward_unsharded_param_values: list[Any],
+        forward_flat_buffer_values: list[Any],
+        runtime_validate: bool = False,
+    ) -> tuple[list[Any], list[Any], Any, tuple[Any, ...]]:
+        """Run the multiplexed graph and split backward/forward outputs."""
+
+        bw_args = self.bw_graphs._backward_args(
+            backward_stage_output,
+            backward_saved_values_for_backward,
+            output_grads_from_next,
+            runtime_validate=runtime_validate,
+        )
+        fw_args = self.fw_graphs._forward_args(
+            forward_args,
+            forward_kwargs,
+            forward_target,
+            forward_loss_kwargs,
+            unsharded_param_values=forward_unsharded_param_values,
+            flat_buffer_values=forward_flat_buffer_values,
+            runtime_validate=runtime_validate,
+        )
+        multiplex_args: list[Any] = []
+        multiplex_args.extend(bw_args)
+        multiplex_args.extend(fw_args)
+        multiplexed_outputs = _execute_graph_module(
+            self.multiplexed_graph,
+            multiplex_args,
+        )
+        num_bw_outputs = (
+            self.bw_graphs.meta.num_param_grad_values
+            + self.bw_graphs.meta.num_input_grad_values
+        )
+        input_grads, param_grads = self.bw_graphs._split_full_backward_outputs(
+            multiplexed_outputs[:num_bw_outputs]
+        )
+        output, saved_values_for_backward = self.fw_graphs._split_forward_outputs(
+            multiplexed_outputs[num_bw_outputs:]
+        )
+        return input_grads, param_grads, output, saved_values_for_backward
+
+
+def _requires_grad_like(value: Any) -> Any:
+    if isinstance(value, torch.Tensor) and value.is_floating_point():
+        value = value.detach().requires_grad_(True)
+    return value
+
+
+def _grad_input_leaves(
+    stage_args: tuple[Any, ...],
+    stage_kwargs: dict[str, Any],
+) -> list[torch.Tensor]:
+    flat_inputs, _ = pytree.tree_flatten((stage_args, stage_kwargs))
+    return [
+        value
+        for value in flat_inputs
+        if isinstance(value, torch.Tensor) and value.requires_grad
+    ]
+
+
+def _compile_graph_pp_module(
+    gm: fx.GraphModule,
+    *,
+    compile_config: GraphTrainerCompileConfig,
+    graph_name: str,
+) -> fx.GraphModule:
+    """Compile one extracted GraphPP callable with GraphTrainer Inductor passes."""
+    if not compile_config.enable or not compile_config.enable_passes:
+        return ensure_boxed_graph_module(gm)
+
+    example_inputs = example_inputs_from_placeholders(gm)
+    gm = apply_graph_passes(
+        gm,
+        example_inputs,
+        final_inductor_compile_passes(
+            compile_config,
+            use_cudagraph=False,
+            boxed_codegen=True,
+        ),
+        compile_config=compile_config,
+    )
+    if gm.meta.get(BOXED_CODEGEN_META) is not True:
+        raise ValueError(
+            "GraphPP compiled graph did not use boxed codegen. Check that the "
+            "terminal Inductor pass was not disabled."
+        )
+    logger.info(
+        "GraphPP compiled %s with %s inductor",
+        graph_name,
+        compile_config.inductor_compilation,
+    )
+    return gm
+
+
+def _compile_stage_graphs(
+    stage: GraphPipelineStage,
+    *,
+    compile_config: GraphTrainerCompileConfig,
+) -> None:
+    """Compile the GraphTrainer graphs attached to ``stage`` once."""
+
+    if stage.graphs is None:
+        raise ValueError(
+            "GraphPP cannot compile missing stage graphs for "
+            f"stage {stage.stage_index}."
+        )
+    graphs = cast(GraphTrainerStageGraphs, stage.graphs)
+    if graphs.compiled:
+        return
+    compiled_modules: dict[str, fx.GraphModule | None] = {}
+    for name, gm in (
+        ("fw", graphs.modules.fw),
+        ("full_bw", graphs.modules.full_bw),
+        ("bw_di", graphs.modules.bw_di),
+        ("bw_dw", graphs.modules.bw_dw),
+        ("unshard", graphs.modules.unshard),
+        ("reduce_grad", graphs.modules.reduce_grad),
+    ):
+        compiled_modules[name] = (
+            None
+            if gm is None
+            else _compile_graph_pp_module(
+                gm,
+                compile_config=compile_config,
+                graph_name=f"stage_{stage.stage_index}_{name}",
+            )
+        )
+    graphs.modules = _StageGraphModules(
+        fw=cast(fx.GraphModule, compiled_modules["fw"]),
+        full_bw=cast(fx.GraphModule, compiled_modules["full_bw"]),
+        bw_di=compiled_modules["bw_di"],
+        bw_dw=compiled_modules["bw_dw"],
+        unshard=compiled_modules["unshard"],
+        reduce_grad=compiled_modules["reduce_grad"],
+    )
+    graphs.compiled = True
+
+
+def _annotate_graph_pp_graph(
+    gm: fx.GraphModule,
+    *,
+    stage_index: int,
+    callable_name: str,
+    action_name: str,
+) -> None:
+    for node in gm.graph.nodes:
+        node.meta = dict(node.meta)
+        node.meta["graph_pp_stage_index"] = stage_index
+        node.meta["graph_pp_callable"] = callable_name
+        node.meta["graph_pp_action"] = action_name
+        if node.op == "placeholder":
+            node.meta["graph_pp_slot"] = f"input:{node.name}"
+        elif node.op == "output":
+            node.meta["graph_pp_slot"] = "output"
+
+
+def _annotate_graph_pp_modules(
+    modules: _StageGraphModules,
+    *,
+    stage_index: int,
+) -> None:
+    graph_specs = (
+        (modules.fw, "fw", "FORWARD"),
+        (modules.full_bw, "full_bw", "FULL_BACKWARD"),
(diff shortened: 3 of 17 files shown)


torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -136,6 +134,22 @@ def forward_backward_step(
                 labels=labels,
                 global_valid_tokens=global_valid_tokens,
             )
+        if self.parallel_dims.pp_enabled:
+            from torchtitan.experiments.graph_trainer.graph_pp.runner import (
+                GraphPPRunner,
+            )
+
+            if not isinstance(self.pp_schedule, GraphPPRunner):
```
aditvenk: When will this be true?

sanketpurandare: So GraphPPRunner is a subclass of _PipelineScheduleRuntime in upstream torch.pipelining. All new schedules (Interleaved1F1B, ZeroBubbleV and DualPipeV) are subclassed from _PipelineScheduleRuntime.But we should error instead of fallback

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -67,6 +68,36 @@ def build_decoder_config_for_backend(
...
+    if set(loss_kwargs) == {"global_valid_tokens"}:
+        return annotated_loss_fn(pred, labels, loss_kwargs["global_valid_tokens"])
+    return annotated_loss_fn(pred, labels, **loss_kwargs)
+
+
+def accumulate_param_grads_(
+    params: Iterable[torch.Tensor],
+    grads: Iterable[torch.Tensor | None],
+    *,
+    strict: bool = True,
```
aditvenk: nit: don't see any use for strict = False, we can remove this param

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -157,6 +171,57 @@ def forward_backward_step(
...
+    ) -> torch.Tensor:
+        from torchtitan.experiments.graph_trainer.graph_pp.runner import (
+            build_graph_pp_graph_bundles,
+        )
+
+        inputs, labels, extra_kwargs = self.post_dataloading_process(input_dict, labels)
+        loss_kwargs = {"global_valid_tokens": global_valid_tokens}
+        with self.train_context():
+            targets, losses = (labels, []) if self.pp_has_last_stage else (None, None)
+            if self.pp_has_first_stage:
```
aditvenk: Can we avoid duplication by building the positional args once (inputs if pp_has_first_stage, else empty) and reusing for both cases?

torchtitan/experiments/graph_trainer/graph_pp/runner.py
```diff
@@ -0,0 +1,1454 @@
...
+    Pseudocode:
+      split the full batch with the upstream PP microbatch splitter
+      normalize trace-reused BlockMask inputs
+      run eager PP metadata inference on representative microbatch 0
+      build missing stage graph bundles
+    """
+    graph_stages = [cast(GraphPipelineStage, stage) for stage in schedule._stages]
+    if all(stage.graph_callables is not None for stage in graph_stages):
+        for stage in graph_stages:
+            _compile_stage_graph_bundle(stage)
```
aditvenk: This feels unnecssary. We should never be in a case where callable exists but it is not compiled.

torchtitan/experiments/graph_trainer/graph_pp/runner.py
```diff
@@ -0,0 +1,1454 @@
...
+      if graph is missing or compile passes are disabled: return graph
+      collect placeholder example inputs
+      run regional or full Inductor according to compile_config
+      return compiled GraphModule
+    """
+    if gm is None or not compile_config.enable or not compile_config.enable_passes:
+        return gm
+
+    example_inputs = _example_inputs_from_placeholders(gm)
+    if compile_config.inductor_compilation == "regional":
```
aditvenk: Can we convert this into an util and reuse here and in the non-Graph PP case? 
There is a risk of this diverging from the non-graph PP case

torchtitan/experiments/graph_trainer/graph_pp/runner.py
```diff
@@ -0,0 +1,1454 @@
...
+        )
+
+    passes = compile_time_passes(
+        traced,
+        _GraphPPPassConfig(
+            compile=compile_config,
+            parallelism=parallelism,
+            model_spec=_GraphPPPassModelSpec(model=model_config),
+        ),
+        use_cudagraph=False,
```
aditvenk: Can we ensure we error out if cudagraph and pp are enabled?

torchtitan/experiments/graph_trainer/graph_pp/runner.py
```diff
@@ -0,0 +1,1454 @@
...
+    apply_graph_passes,
+    compile_time_passes,
+    full_inductor_compilation_pass,
+    regional_inductor_pass,
+)
+from torchtitan.tools.logging import logger
+
+
+@dataclasses.dataclass(slots=True)
+class GraphCallables:
```
aditvenk: Can we add docstrings explaining what each dataclass in this file is for? I know the RFC mentions this, but it will help future code readers.

torchtitan/experiments/graph_trainer/tests/test_graph_pp_runner.py
```diff
@@ -0,0 +1,371 @@
...
+from torchtitan.config import ParallelismConfig
+from torchtitan.experiments.graph_trainer.chunked_loss import (
+    ChunkedCELossWithParamGrads,
+)
+from torchtitan.experiments.graph_trainer.configs import GraphTrainerCompileConfig
+from torchtitan.experiments.graph_trainer.graph_pp.pipeline import (
+    _validate_graph_pp_config,
+)
+
+from torchtitan.experiments.graph_trainer.graph_pp.runner import (
```
aditvenk, line 37: Can we have an agent to vet that this test is covering sufficient cases of the Graph PP runtime state machine? 
For e.g., I don't see a unit test for the case dW falls back to full backward. 
I suspect there may be more.

aditvenk (changes requested):

torchtitan/experiments/graph_trainer/graph_pp/pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+    schedule_class = get_schedule_class(parallelism.pipeline_parallel_schedule)
+    if not issubclass(schedule_class, _PipelineScheduleRuntime):
+        raise ValueError(
+            "GraphPP currently requires a runtime PP schedule such as "
+            "Interleaved1F1B, ZBVZeroBubble, or DualPipeV. "
+            f"Got {parallelism.pipeline_parallel_schedule}."
+        )
+
+
+def graph_pp_llm(
```
wconstab: this is meant to be the main user entrypoint to share across models right? probably worth adding a real docstring

torchtitan/experiments/graph_trainer/graph_pp/runner.py
```diff
@@ -0,0 +1,1454 @@
...
+            flex_compile_config=FlexAttention.inductor_configs,
+        )
+        if compile_config.numerics_changing_optim:
+            from torchtitan.experiments.graph_trainer.performance_passes import (
+                annotate_rmsnorm_for_regional_inductor_pass,
+            )
+
+            gm = annotate_rmsnorm_for_regional_inductor_pass(gm, example_inputs)
+        gm = regional_inductor_pass(gm, example_inputs)
+    elif compile_config.inductor_compilation == "full":
```
wconstab: worth calling out what the difference is between regional and full? it seems more obvious in the non-pp case, but i might assume regional == layer-by-layer which approximately == full pipeline stage sometimes. i also note something special cased for rmsnorm.

torchtitan/experiments/graph_trainer/tests/test_graph_pp_runner.py
sanketpurandare: This happens automatically for all the first stages since the input to the first stage do no require grads so in ZeroBubble and DualPipeV dW will fall back to fullBW, if it doesn't it won't work at all. But I agree with the general direction, will post a comment on the PR about test coverage generated using an agent.

torchtitan/experiments/graph_trainer/graph_pp/runner.py
sanketpurandare: regional is not layer by layer, regional is compiling only annotated nodes in your fx graph with inductor, full is compiling the the entire fx graph using inductor.

torchtitan/experiments/graph_trainer/graph_pp/runner.py
sanketpurandare: Yep, I will aim for maximal reuse of pre-existing inductor passes in GraphTrainer.

torchtitan/experiments/graph_trainer/graph_pp/runner.py
sanketpurandare: Instead of erroring out, I will show a warning that cuda graphs are not enabled yet and fall back to non-cuda graph case

torchtitan/components/loss.py
tianyu-l, line 1: seems meaningless to core

tianyu-l (dismissed):

torchtitan/components/loss.py
sanketpurandare: It will not be there, I am refactoring some stuff, core won't be touched in this entire stack

torchtitan/experiments/graph_trainer/tests/test_graph_pp_runner.py
sanketpurandare: @aditvenk 

# GraphPP Test Coverage Review Note

This stack has deliberately layered GraphPP coverage. The tests are not a large
synthetic matrix; they are split across the invariants that matter for review:
graph transformation correctness, runner calling conventions, distributed
numerics, and full-Inductor runtime coverage.

## Summary

The current GraphPP stack is covered by:

| Layer | Test file | Coverage |
| --- | --- | --- |
| Graph partitioning | `torchtitan/experiments/graph_trainer/tests/test_graph_pp_passes.py` | Real DSV3 MoE decoder block with FlexAttention, plus partition invariant failures |
| dI/dW backward split | `torchtitan/experiments/graph_trainer/tests/test_graph_pp_passes.py` | Real DSV3 MoE backward graph replayed through split dI and dW graphs |
| FSDP collective split | `torchtitan/experiments/graph_trainer/tests/test_graph_pp_passes.py` | Forward unshard extraction, backward reduce-grad extraction, malformed collective chains, no-FSDP no-op paths |
| FSDP memory policy interaction | `torchtitan/experiments/graph_trainer/tests/test_passes.py` | `reshard_after_forward` policy saves the helper-selected FSDP unshard boundary |
| GraphPP graph builder and runner contracts | `torchtitan/experiments/graph_trainer/tests/test_graph_pp_runner.py` | 26 focused tests for tracing, graph building, calling conventions, loss handling, block masks, overlap graphs, and runtime state |
| Distributed numerics | `torchtitan/experiments/graph_trainer/tests/test_numerics.py` | Exact loss comparison for DSV3 MoE + FlexAttention + EP + DP-shard + PP across GraphPP schedules |
| Full-Inductor integration | `torchtitan/experiments/graph_trainer/tests/integration_tests.py` | 8-GPU DSV3 GraphPP full-Inductor runs for Interleaved1F1B, ZBVZeroBubble, and DualPipeV |

The highest-risk graph passes are not tested only with toy graphs. Partitioning and dI/dW splitting both exercise a real DeepSeek V3 MoE decoder block using FlexAttention and the real block-mask tracing path.

## Commit-by-Commit Coverage

| Commit | Main tests added | What the tests prove |
| --- | --- | --- |
| `[graph_pp] Add stage graph partitioning` | `test_graph_pp_passes.py` +386 lines | Partitioning preserves the real joint graph semantics for a DSV3 MoE FlexAttention block; pass-through saved values are preserved; invalid backward-only inputs fail loudly. |
| `[graph_pp] Add dI/dW backward splitting` | `test_graph_pp_passes.py` +61 lines | Splitting full backward into input-grad and weight-grad graphs reconstructs the original backward outputs on a real DSV3 MoE graph. |
| `[graph_pp] Add FSDP collective splitting` | `test_graph_pp_passes.py` +208 lines, `test_passes.py` +59 lines | FSDP unshard and reduce-grad collectives are extracted into separate graphs; malformed chains are rejected; no-FSDP paths remain no-op; default memory policy handles `reshard_after_forward` correctly. |
| `[graph_pp] Add GraphPP runner integration` | `test_graph_pp_runner.py` +693 lines, integration/numerics additions | GraphPP graph construction avoids real eager pre-trace forward/loss, preserves buffers, handles loss kwargs, builds callable graphs, and integrates with distributed DSV3 loss comparison. |
| `[graph_pp] Add multiplexed graph coverage and DualPipeV validation` | `test_graph_pp_runner.py` +312 lines, `test_numerics.py` +150 lines, integration updates | Overlap FW/BW multiplex graphs are built, cached, metadata-correct, and validated through DSV3 numerics and full-Inductor integration for ZBV/DualPipeV. |
| `[graph_pp] Add DualPipeV scheduling placeholder` | minimal runner test update | Keeps placeholder scheduling validation in the stack without broad test churn. |

## Pass-Level Coverage

### Partition Pass

`GraphPPPartitionTest` has five tests:

- `test_real_dsv3_moe_block_partition_matches_joint_graph`
- `test_partition_saves_backward_passthrough_placeholders`
- `test_invalid_backward_only_input_indices_raise`
- `test_backward_only_input_required_by_forward_raises`
- `test_forward_mutation_of_backward_only_input_raises`

The real DSV3 test traces a MoE decoder block with FlexAttention and then checks:

1. The forward partition returns the same user forward output as the original
   joint graph.
2. The backward partition consumes saved forward values plus output grads and
   reconstructs the original joint backward outputs.
3. The metadata counts line up with the calling convention: user forward outputs,
   saved values for backward, side-effect outputs, backward grad inputs, and
   backward outputs.

The synthetic partition tests are intentionally small. They cover invariants
that would otherwise fail silently:

- duplicate or out-of-range backward-only input indices,
- backward-only inputs accidentally required by the forward graph,
- mutation of a backward-only input by the forward graph,
- pass-through values such as DTensor layout metadata that are needed by
  backward even though they are not user-visible forward outputs.

### dI/dW Split

`GraphPPSplitDiDwTest` has two tests:

- `test_real_dsv3_moe_block_split_reconstructs_backward`
- `test_real_dsv3_moe_block_without_input_grad_skips_split`

The first test starts from the real partitioned DSV3 MoE backward graph, applies
the dI/dW split, and verifies:

1. The dI graph produces the same input grads as the full backward graph.
2. The dI graph forwards the live values needed by dW.
3. The dW graph produces the same parameter grads as the full backward graph.

The second test covers the first-stage calling convention: if the stage does not
produce input grads, dI/dW splitting is correctly skipped.

### FSDP Collective Split

`GraphPPFSDPCollectiveSplitTest` has five tests:

- `test_forward_split_extracts_unshard_and_replicated_params`
- `test_forward_split_no_fsdp_is_noop`
- `test_forward_split_requires_wait_after_all_gather`
- `test_backward_split_extracts_reduce_grad_epilogues`
- `test_backward_split_no_fsdp_is_noop_and_validates_grad_count`

These tests cover the pass contracts directly:

1. Forward FSDP all-gather/wait/reconstruction chains are moved into an
   explicit unshard graph.
2. Replicated params remain on the correct calling convention.
3. Malformed all-gather chains fail loudly when the expected wait is missing.
4. Backward reduce-grad epilogues are separated from the main backward graph.
5. Both reduce-scatter and all-reduce collectives are handled in the reduce-grad
   extraction path.
6. No-FSDP graphs remain no-op instead of being rewritten unnecessarily.

There is also targeted coverage in `test_passes.py` that the default memory
policy saves the FSDP unshard boundary when `reshard_after_forward` semantics
require it. This matters because GraphPP's FSDP splitting and graph trainer's
standard FSDP memory policy must agree on the same unshard boundary.

## Runner and Graph Builder Coverage

`GraphPPRunnerTraceTest` has 26 tests. The important coverage groups are:

### Tracing and Graph Construction

- Non-last and last stage graph builds do not run a real eager pre-trace
  forward.
- Last-stage graph build does not run a real eager pre-trace loss.
- The graph builder relies on fakeification from the minimal FX tracer instead
  of doing an eager forward just to infer output structure.
- Stage graph construction preserves module buffer mutations.
- GraphPP node metadata is annotated for TLParse and debugging.

### Calling Convention and Loss Handling

- Non-last stage graphs match eager gradients.
- Last-stage graphs return scalar loss and input grads correctly.
- Chunked loss preserves the hidden-state grad accumulator and checks exact
  tensor equality in the local test.
- Loss kwargs, such as `global_valid_tokens`, are passed through correctly.
- Last-stage forward leaves loss ownership to upstream PP loss handling.

### Runtime State

- Only trainable params get accumulated grads.
- Upstream PP owns microbatch splitting; GraphPP does not wrap or replace
  upstream split inputs.
- `eval()` forwards to the schedule and clears GraphPP runtime state.
- GraphPP validates unsupported schedule shape early.
- Runtime graph readiness errors are explicit.

### BlockMask Handling

- Split BlockMask batch offsets remain dynamic for replay.
- Existing stage graphs normalize split BlockMasks in place before graph
  construction.

This directly targets FlexAttention replay risk: GraphPP should not specialize
the graph to microbatch 0's batch-offset closure.

### Overlap and Multiplexed Graphs

- Existing overlap graphs are reused instead of rebuilt.
- Full-Inductor overlap builds multiplexed FW/BW graphs.
- Unsupported overlap backward-input subactions fail explicitly.
- Shape environment metadata is transferred into the forward fake mode.
- Multiplexed graph outputs are ordered as backward outputs followed by forward
  outputs.
- Multiplexed overlap graphs are prebuilt for overlap actions.

These tests are focused on the highest-risk runner feature: DualPipeV and
overlapped FW/BW execution need the runner to call the multiplexed graph with a
precise convention and output split.

## Distributed Numerics Coverage

`test_graph_pp_moe_dsv3_aot_fx_trace_vs_eager` runs exact loss comparison for:

- model: `graph_trainer.deepseek_v3`,
- config: DSV3 debug model,
- attention: FlexAttention,
- pipeline parallel degree: 2,
- data parallel shard degree: 4,
- expert parallel degree: 2,
- local batch size: 8,
- global batch size: 32,
- GraphPP schedules:
  - `Interleaved1F1B`,
  - `ZBVZeroBubble`,
  - `DualPipeV`.

The baseline is eager PP `Interleaved1F1B`. This is intentional:

1. FlexAttention needs `torch.compile`.
2. Eager PP cannot use `ZBVZeroBubble` or `DualPipeV` with `torch.compile`.
3. Therefore the meaningful comparison is GraphPP schedule execution against the
   eager Interleaved1F1B PP baseline.

The test disables gradient clipping with `--training.max_norm=inf`. This is not
a GraphPP workaround. TorchTitan clipping is applied per local rank, so different
pipeline schedules can produce different local clipping coefficients even when
the pre-clip gradients are bitwise equal. Disabling clipping isolates the graph
execution numerics from that schedule-level clipping effect.

This is the strongest end-to-end correctness coverage in the stack because it
tests the actual distributed mode reviewers care about: DSV3 MoE with
FlexAttention, EP, DP shard, PP, graph tracing, and GraphPP schedule execution.

## Full-Inductor Integration Coverage

The integration test matrix includes 8-GPU full-Inductor GraphPP runs for:

- DSV3 GraphPP `Interleaved1F1B`,
- DSV3 GraphPP `ZBVZeroBubble`,
- DSV3 GraphPP `DualPipeV`.

Each uses:

- `--module graph_trainer.deepseek_v3`,
- `--config graph_trainer_deepseek_v3_debugmodel`,
- `--compile.mode aot_fx_trace`,
- `--compile.inductor_compilation full`,
- `--parallelism.pipeline_parallel_degree 2`,
- `--parallelism.data_parallel_shard_degree 4`,
- `--parallelism.expert_parallel_degree 2`.

The distributed numerics test uses regional Inductor for exact loss comparison,
while the integration tests use full Inductor for runtime coverage. This split is
intentional: exact numerics catches semantic regressions; full-Inductor
integration catches backend/runtime integration failures.

## Why This Coverage Is Meaningful

The stack does not rely on shallow smoke tests alone.

1. The graph partition and dI/dW passes are replayed against the original joint
   graph, including a real DSV3 MoE FlexAttention trace.
2. The runner tests assert the GraphPP calling convention directly, including
   last-stage loss behavior, saved backward values, input grads, param grads,
   chunked loss, and overlap graph output ordering.
3. Distributed numerics compares actual training losses for DSV3 MoE with EP,
   DP-shard, PP, and FlexAttention across all GraphPP schedules in the stack.
4. Full-Inductor integration separately verifies the final compiled runtime path
   for those schedules.
5. Error-path tests check that invalid graph-pass invariants fail loudly instead
   of being silently skipped.

## Known Bounded Gaps

There are still some gaps, but they are bounded and do not invalidate the current
coverage story:


1. The runner unit tests use local/fake schedules for focused contract coverage.
   End-to-end schedule execution is covered by distributed numerics and
   integration tests rather than a small unit-level `_PipelineScheduleRuntime`
   replay test.
2. CUDA graph capture itself is not directly tested in this stack. The current
   stack makes GraphPP graph execution structured enough for future capture work;
   capture/replay tests should be added with the actual CUDA graph enablement.
3. Exact numerics currently uses regional Inductor, while full Inductor is
   covered by integration smoke tests. A full-Inductor exact numerics test could
   be added later if test runtime is acceptable.

These gaps are appropriate for the current stack. The tests added here cover the
core semantic risk: graph transformations preserve the original graph behavior,
the runner uses the intended calling convention, and distributed DSV3 GraphPP
matches eager PP numerics for the supported schedules.

## Reviewer-Facing Test Plan

The concise test plan for the stack is:

```bash
pytest torchtitan/experiments/graph_trainer/tests/test_graph_pp_passes.py -q
pytest torchtitan/experiments/graph_trainer/tests/test_graph_pp_runner.py -q
pytest torchtitan/experiments/graph_trainer/tests/test_numerics.py -k graph_pp_moe_dsv3_aot_fx_trace_vs_eager -q
```

The integration matrix additionally includes:

- `aot_fx_trace deepseek_v3 GraphPP Interleaved1F1B full_inductor`
- `aot_fx_trace deepseek_v3 GraphPP ZBVZeroBubble full_inductor`
- `aot_fx_trace deepseek_v3 GraphPP DualPipeV full_inductor`

## Bottom Line

The coverage is strong where this stack is risky:

- real DSV3 MoE FlexAttention graph pass coverage,
- exact distributed DSV3 GraphPP numerics,
- full-Inductor runtime integration,
- explicit runner calling-convention tests,
- explicit invariant/error-path tests.

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -97,6 +98,41 @@ def _maybe_materialize_grad_for_param_layout(
 _NOT_IN_LAYERS = -1
 
 
+def compute_annotated_loss(
+    loss_fn: Callable[..., torch.Tensor],
+    pred: Any,
```
aditvenk: nit: can use torch.Tensor for pred and labels.

Also, `loss_fn` annotation seems wrong, as it returns multiple outputs (loss, metrics)

Suggested:

```
from typing import TypeAlias

LossResult: TypeAlias = torch.Tensor | tuple[torch.Tensor, dict[str, torch.Tensor]]
AnnotatedLossFn: TypeAlias = Callable[..., LossResult]
```

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -163,6 +166,32 @@ def forward_backward_step(
...
+                **extra_kwargs,
+                target=targets,
+                losses=losses,
+                loss_kwargs=loss_kwargs,
+                return_outputs=False,
+            )
+
+        if self.pp_has_last_stage:
+            assert losses is not None
+            return torch.sum(torch.stack(losses)).to(self.device)
```
aditvenk, line 192: Why is this to(device) required? Are there any cases where losses is on CPU? isn't that a bug?

torchtitan/experiments/graph_trainer/inductor_passes.py
```diff
@@ -76,10 +79,20 @@ def regional_inductor_pass(
...
             filter to allow distributed collective ops.
+        boxed_codegen: When True, compile the returned FX graph with boxed
+            calling convention so its mutable runtime arg list is cleared after
+            placeholder extraction.
     """
     import torch._inductor.config as ic
     from torch._subclasses.fake_tensor import FakeTensor
 
+    if serializable and boxed_codegen:
+        raise ValueError(
```
aditvenk, line 91: Few questions:

1) Is boxed_codegen expected to be a win even without PP? 

2) Will this restriction be removed once Precompile support for PP is enabled?

cc: @bobrenjc93

torchtitan/experiments/graph_trainer/deepseek_v3/config_registry.py
```diff
@@ -26,6 +29,12 @@ def graph_trainer_deepseek_v3_debugmodel() -> GraphTrainer.Config:
     return config
 
 
+def graph_trainer_deepseek_v3_debugmodel_ep() -> GraphTrainer.Config:
```
aditvenk: This is redundant with graph_trainer_deepseek_v3_debugmodel

torchtitan/experiments/graph_trainer/graph_pp/utils.py
```diff
@@ -324,17 +323,16 @@ def _wrap_unwrapped_values(
...
 ) -> BlockMask:
-    """Use PyTorch PP's BlockMask splitter for GraphPP microbatch inputs.
+    """Make a PyTorch PP-split BlockMask replayable by one traced graph.
 
-    Older split BlockMask objects can carry a closure pointing back to the
-    original full-batch BlockMask plus this microbatch's index. Re-split that
-    original value with PyTorch PP's splitter instead of maintaining a separate
-    GraphPP BlockMask construction path.
+    ``_split_block_mask`` correctly creates per-microbatch tensor metadata, but
+    its ``mask_mod`` closure captures the batch offset as a Python int. ``make_fx``
```
aditvenk, line 333: Isn't this a bug in _split_block_mask?

torchtitan/experiments/graph_trainer/graph_pp/utils.py
```diff
@@ -345,17 +343,50 @@ def _resplit_block_mask_with_torch_pipelining(
...
+    def graphable_mask_mod(
+        b: torch.Tensor,
+        h: torch.Tensor,
+        q_idx: torch.Tensor,
+        kv_idx: torch.Tensor,
+    ) -> torch.Tensor:
+        offset = batch_offset.to(device=b.device, dtype=b.dtype)
+        return base_mask_mod(b + offset, h, q_idx, kv_idx)
+
+    return BlockMask(
```
aditvenk: nit: can't we shallow-copy and mutate just mask_mod field? Will make this smaler.

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+from torchtitan.experiments.graph_trainer.passes import (
+    apply_graph_passes,
+    compile_time_passes,
+    final_inductor_compile_passes,
+)
+from torchtitan.tools.logging import logger
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class _GraphPPPassConfig:
```
aditvenk: can we add docstring for this? It is not clear what parallelism/model_spec are and the typing annotation of Any doesn't provide clues either.

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+    unshard: fx.GraphModule | None = None
+    reduce_grad: fx.GraphModule | None = None
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class _StageGraphMeta:
+    """GraphTrainer metadata for one bound GraphPP stage graph executor.
+
+    Counts ending in ``_values`` refer to flat tensor values in extracted FX
+    graphs. This metadata is private to ``GraphTrainerStageGraphs``; the PP
```
aditvenk: nit: if something is private, prefix with underscore to make it clear. Better than suffix with _values

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+    full_bw: fx.GraphModule
+    bw_di: fx.GraphModule | None = None
+    bw_dw: fx.GraphModule | None = None
+    unshard: fx.GraphModule | None = None
+    reduce_grad: fx.GraphModule | None = None
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class _StageGraphMeta:
+    """GraphTrainer metadata for one bound GraphPP stage graph executor.
```
aditvenk, line 117: Can we expand this docstring to capture at a higher-level what kind of metadata is captured per stage? While the individual parameter names are intelligible, we can capture higher-level organizing principle in the docstring

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+            f"argument list, but {len(args)} entries remain."
+        )
+    if isinstance(outputs, tuple):
+        return outputs
+    if isinstance(outputs, list):
+        return tuple(outputs)
+    return (outputs,)
+
+
+def _ensure_boxed_graph_module(gm: fx.GraphModule) -> fx.GraphModule:
```
aditvenk: minor nit: this feels like something we can move to common_utils.py and reuse in inductor_passes.py too.

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+                raise ValueError(
+                    "GraphPP unshard parameter index is out of range: "
+                    f"index {param_index}, but runtime has "
+                    f"{len(flat_param_values)} flat params"
+                )
+            unshard_args.append(flat_param_values[param_index])
+        unsharded_param_values = list(
+            _execute_graph_module(self.modules.unshard, unshard_args)
+        )
+        if len(unsharded_param_values) != self.meta.num_flat_param_values:
```
aditvenk: Should we also error out if `flat_param_values` contains _more_ values than expected?

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+        loss_kwargs: dict[str, Any],
+        *,
+        unsharded_param_values: list[Any],
+        flat_buffer_values: list[Any],
+    ) -> list[Any]:
+        """Pack the extracted forward graph inputs in placeholder order."""
+
+        num_unsharded_inputs = self.meta.num_fw_unsharded_param_inputs
+        if len(unsharded_param_values) != self.meta.num_flat_param_values:
+            raise ValueError(
```
aditvenk, line 310: The checks based solely on `self.meta` values could be done once instead of in every time _forward_args is called -- e.g., post init of the per-stage object?

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
```diff
@@ -0,0 +1,1309 @@
...
+    if stage.graphs is None:
+        raise ValueError(
+            "GraphPP cannot compile missing stage graphs for "
+            f"stage {stage.stage_index}."
+        )
+    graphs = cast(GraphTrainerStageGraphs, stage.graphs)
+    if graphs.compiled:
+        return
+    if not compile_config.enable or not compile_config.enable_passes:
+        graphs.modules = _ensure_boxed_graph_modules(graphs.modules)
```
aditvenk: nit: This is redundant given _compile_graph_pp_module already takes care of this case. This is an earlier out, but at the cost of more code.

torchtitan/experiments/graph_trainer/graph_pp/runner.py
```diff
@@ -0,0 +1,735 @@
...
+            accumulated[index] = grad
+        else:
+            accumulated[index] += grad
+
+
+def _require_stage_graphs(
+    stage: GraphPipelineStage,
+    action_name: str,
+) -> GraphPPStageGraphs:
+    if stage.graphs is None:
```
aditvenk: This is redundant with _require_graph_trainer_stage_graphs

Generally - we can use agent to scrub some of these utility functions in this big PR, either inline or use common ones in utils.py

aditvenk (approved):
I have tried to be thorough in places where we are interfacing wiht larger graph trainer. The inner details of pipeline/runner -- I tried to mentally map with RFC and have not reviewed as thoroughly.

torchtitan/experiments/graph_trainer/trainer.py
sanketpurandare: Since loss functions are provided by user in case of PP there is no good way to know this. In our case it will be on GPU and the to device call will be a no-op

torchtitan/experiments/graph_trainer/graph_pp/utils.py
sanketpurandare: It is a bug for tracing not for execution, the split-block mask was never traced before hence it having python int works fine. Now we need something more from it so its more of scope expansion. Btw we are eliminiating the need for PP to split microbatches in the first place, so won't be an issue. 

https://github.com/pytorch/pytorch/pull/188500

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
sanketpurandare: Strict equality is necessary here else graph calling convention fails

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
sanketpurandare: these are runtime checks since metadata inference is done for first micro-batch and reused throughout. For perf reasons I will guard it agains't this variable. 

https://github.com/pytorch/pytorch/blob/0f99b003def75673557701e08848354a51ff15fe/torch/distributed/pipelining/stage.py#L298

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
sanketpurandare: Will add typing for it

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
sanketpurandare: Makes sense.

torchtitan/experiments/graph_trainer/graph_pp/graph_builder.py
sanketpurandare: will remove it.

torchtitan/experiments/graph_trainer/inductor_passes.py
sanketpurandare: 1. I don't think so there is any with for boxed_codegen without PP since inputs of non-PP graphs are params and user inputs and outputs are loss and grads, all need to be retained.
2. Precompile and save with boxed codegen disabled and can do boxed_codegen at runtime, I guess the boxed_codegen is a very lightweight pass?

torchtitan/experiments/graph_trainer/graph_pp/utils.py
sanketpurandare: python int offsets are baked into it so mutating will take the same time. Anyways this is a very short term patch

PR #3092, opened 2026-04-24
Use `current_accelerator` for device in AutoParallel calls to enable XPU

frost-intel:
In order to enable AutoParallel for XPU, this changes the `torch.device("cuda")` device to `torch.accelerator.current_accelerator()`.


torchtitan/experiments/autoparallel/deepseek_v3/parallelize_deepseekv3.py
```diff
@@ -306,7 +306,7 @@ def input_fn():
                 0,
                 model.config.vocab_size,
                 (global_batch_size, training.seq_len),
-                device=torch.device("cuda"),
+                device=torch.accelerator.current_accelerator(),
```
tianyu-l: could you use https://github.com/pytorch/torchtitan/blob/main/torchtitan/tools/utils.py#L41 instead, to be more consistent with the repo style

tianyu-l (approved):

PR #3095, opened 2026-04-24
add reset_prefix_cache

liangel-02:
need to add self._engine.reset_prefix_cache() so that old KV cache values arent used after weight updates
Screenshot 2026-04-24 at 6 41 33 PM


tianyu-l (approved):
makes a lot of sense!

PR #3098, opened 2026-04-24
Fix SAC test compatibility with PyTorch indexed storage

mori360:
### Summary                                                                                                    
                                                                  
- Fix test_flops failure caused by PyTorch PR #176455 which switches SAC from re-running the policy function during recompute to indexed storage replay
- Wrap forward with FlopCounterMode so SAC indexes ops under the same dispatch mode stack as backward recompute                                                                                                  
- Backward flops are still measured separately via a dedicated bwd_mode instance — expected values
  unchanged                                                                                                  
                                                                  
### Test plan                                                                                                  
                                                                  
`python -m pytest tests/unit_tests/test_activation_checkpoint.py -xvs`


tianyu-l (approved):
thanks

PR #3108, opened 2026-04-27
Simplify parallelize function signature

wwwjn:
## Why change
This PR makes the parallelize API self-contained: everything it actually uses lives on ParallelismConfig, and the few training-only checks move to the trainer.

## What changed

Here's changes:
1. **Parallelize signatures**:  training: TrainingConfig removed from all core model parallelize_* (llama3, llama4, qwen3, qwen3_vl, deepseek_v3, gpt_oss, flux, transformers_modeling_backend). Bodies now read parallelism.fsdp_* instead of training.*.                                                                                        
2. **Optional defaults**: compile_config and ac_config are now | None = None (None means skip), and dump_folder defaults to "". Inference callers can omit them.
                                                                                                 
3. **Seq-len assert moved**: the seq_len % seq_len_divisor == 0 check moves from each parallelize_*  to Trainer.__init__, where both training and parallel_dims are naturally available.


torchtitan/models/qwen3/parallelize.py
```diff
@@ -45,22 +44,18 @@ def parallelize_qwen3(
...
-    training: TrainingConfig,
     model_converters: ModelConvertersContainer.Config,
     parallelism: ParallelismConfig,
-    compile_config: CompileConfig,
-    ac_config: ActivationCheckpointConfig,
-    dump_folder: str,
+    compile_config: CompileConfig | None = None,
+    ac_config: ActivationCheckpointConfig | None = None,
+    dump_folder: str = "",
+    inference: bool = False,
```
tianyu-l: Is qwen3 the only model to support inference?

torchtitan/trainer.py
```diff
@@ -339,6 +339,15 @@ def __init__(self, config: Config):
         )
         assert self.gradient_accumulation_steps > 0
 
+        # Validate sequence length is compatible with parallelism dimensions
+        assert (
+            config.training.seq_len % parallel_dims.seq_len_divisor == 0
+        ), (
+            f"Sequence length {config.training.seq_len} must be divisible by "
+            f"the product of TP degree ({parallel_dims.tp}) and "
```
tianyu-l: after migrating to full DTensor, I don't think we have the TP degree divisibility limitation.

tianyu-l (commented):
actually it's not clear to me what's the best way to approach this
- maybe fsdp should handle inference?
- not sure if we have to delegate compile to vllm -- what if an inference engine doesn't handle it for us
- should we introduce `inference` flag to parallelize, or just not introduce degree 1 FSDP in parallel_dims

Need to discuss more

torchtitan/models/qwen3/parallelize.py
wwwjn: Yes, but this `inference` parameter should not be added in this PR. It was introduced in https://github.com/pytorch/torchtitan/pull/3070 because the explaination here: https://github.com/pytorch/torchtitan/pull/3070#discussion_r3133278434.  Even with fsdp degree =1, FSDP wrap won't work with inference mode.

torchtitan/trainer.py
```diff
@@ -351,6 +357,7 @@ def __init__(self, config: Config):
                     model_config=model_config,
                     parallelize_fn=model_spec.parallelize_fn,
                     loss_fn=self.loss_fn,
+                    local_batch_size=config.training.local_batch_size,
```
wwwjn, line 353: This is rebase issue

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -69,15 +68,15 @@ def pipeline_llm(
...
-    ac_config: ActivationCheckpointConfig,
-    dump_folder: str,
+    compile_config: CompileConfig | None = None,
+    ac_config: ActivationCheckpointConfig | None = None,
+    dump_folder: str = "",
     device: torch.device,
     model_config: BaseModel.Config,
     parallelize_fn: ParallelizeFunction,
     loss_fn: LossFunction,
+    local_batch_size: int,
```
wwwjn, line 79: Local_batch_size is needed for the pipeline_fn.

We only add `local_batch_size` for pipeline_function, not the parallel_fn

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
tianyu-l, line 237: why change?

torchtitan/models/llama3/parallelize.py
```diff
@@ -42,11 +41,10 @@ def parallelize_llama(
...
     *,
     parallel_dims: ParallelDims,
-    training: TrainingConfig,
     parallelism: ParallelismConfig,
-    compile_config: CompileConfig,
-    ac_config: ActivationCheckpointConfig,
-    dump_folder: str,
+    compile_config: CompileConfig | None = None,
+    ac_config: ActivationCheckpointConfig | None = None,
+    dump_folder: str = "",
```
tianyu-l, line 47: why making them optional?
- At least we shouldn't make compile optional -- unified model would rely on this function to compile the model (layer wise)
- For AC, since it has a "none" mode, I would suggest we use that in generator. O/w we are supporting two ways of doing no AC. (Earlier I commented that we could remove "none" mode and just use ACConfig=None, but that would make it impossible to switch AC mode from CLI.)
- dump_folder, IIRC, is for AC related logging, I think we can just keep https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/models/vllm_wrapper.py#L221, but NOT provide a default here. Having an empty default makes pretraining config error-prone.

torchtitan/trainer.py
tianyu-l: should we explore making this and seq_len into data loader config?

torchtitan/distributed/pipeline_parallel.py
```diff
@@ -217,7 +215,7 @@ def _get_pipeline_metadata(
     return num_virtual_stages, num_layers, input_weight, output_weight
 
 
-def _build_pipeline_schedule(
+def build_pipeline_schedule(
```
tianyu-l, line 218: why changing?

PR #3111, opened 2026-04-27
fix: proper cp global_max_loss accounting

garrett361:
The current `global_max_loss` computation is inaccurate for context-parallelism because it relies on
`local_valid_tokens`, which is computed prior to CP-sharding.  

Relates to https://github.com/pytorch/torchtitan/pull/2314. CC @wwwjn @Shagun-G @tianyu-l

## Problem

Current computation:
```
local_avg_loss  = local_loss_sum / local_valid_tokens
global_max_loss = max(local_avg_loss)   # all-reduce max over ranks
```

With context parallelism (CP), the sequence is sharded across CP ranks *after* `local_valid_tokens` is computed. That means every rank in a CP group has the same `local_valid_tokens` value, even though each rank only processes approx `local_valid_tokens / cp_degree` tokens. This results in `global_max_loss` being too-small by a factor of  `~cp_degree`. See before/after screenshots below.

See the screenshots of the max and
mean losses from wandb and the CLI printout of the loss scale.

## Fix

Track `_rank_local_valid_tokens_per_step` by accumulating `(labels != IGNORE_INDEX).sum()` inside `post_dataloading_process`, *after* CP sharding has been applied. This gives the true count of valid tokens on each rank's CP shard. Use this rank-local count as the denominator for `local_avg_loss`.

A `clamp(min=1)` guard is also added to handle the degenerate case where a CP shard contains only
masked tokens (all `IGNORE_INDEX`), which would otherwise produce a 0/0.

#### Before Fix

Scherm­afbeelding 2026-04-27 om 9 38 54 AM
Scherm­afbeelding 2026-04-27 om 9 40 36 AM

#### After Fix

Scherm­afbeelding 2026-04-27 om 10 24 11 AM
Scherm­afbeelding 2026-04-27 om 10 23 39 AM

### DIFF
--- a/torchtitan/trainer.py
+++ b/torchtitan/trainer.py
@@ -445,6 +445,7 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful, Configurable):
         # These attributes must be initialized before checkpoint loading.
         self.step = 0
         self.ntokens_seen = 0
+        self._rank_local_valid_tokens_per_step = 0
 
         self.checkpointer = config.checkpoint.build(
             dataloader=self.dataloader,
@@ -637,9 +638,12 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful, Configurable):
                 self.config.parallelism.context_parallel_load_balancer,
             )
 
-        # Accumulate after CP sharding so labels.numel() reflects the actual
-        # unique tokens this rank processes (not the full pre-split sequence).
+        # Accumulate after CP sharding so counts reflect the actual unique
+        # tokens this rank processes (not the full pre-split sequence).
         self.ntokens_seen += labels.numel()
+        self._rank_local_valid_tokens_per_step += int(
+            (labels != IGNORE_INDEX).sum()
+        )
 
         return inputs, labels, extra_inputs, extra_kwargs
 
@@ -739,6 +743,7 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful, Configurable):
             global_valid_tokens = local_valid_tokens.float()
 
         # Process each microbatch: move to GPU, forward/backward, then free
+        self._rank_local_valid_tokens_per_step = 0
         accumulated_losses = []
         for input_dict, labels in microbatches:
             # Move tensors to GPU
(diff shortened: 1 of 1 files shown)


garrett361:
`_rank_local_valid_tokens_per_step` isn't a great name. Maybe just `_rank_local_valid_tokens`?

fegin:
Thanks for raising the issue. However, I feel the proper fix is actually moving `self.post_dataloading_process` callsite out of `forward_backward_step` and put it directly in `train_step`. That should solve the issue. I thought `post_dataloading_process` should not be inside `forward_backward_step` but didn't have a strong motivation to do the refactor. But this issue seems to justify the thought.

The `local_valid_tokens` can be calculated directly based on the new labels. `global_valid_tokens` calculation may need to be adjusted according to this change (need to consider CP sharding the sequence). 

wdyt?

garrett361:
> However, I feel the proper fix is actually moving self.post_dataloading_process callsite out of forward_backward_step and put it directly in train_step.

I agree @fegin , the above is my minimal change to get this working on my end while also trying to produce the minimal diff, but I think a refactor of how `forward_backward_step` is the cleaner overall solution.

I originally started  down that path, but it involves changing `forward_backward_step`'s signature, which appears at a bunch of call sites and didn't want to invest the time until we agreed upon a solution:
```
❯ rg forward_backward_step
torchtitan/trainer.py
646:    def forward_backward_step(
750:            loss = self.forward_backward_step(

torchtitan/models/flux/trainer.py
142:    def forward_backward_step(
156:                This field is a placeholder for now as we rescale the loss within forward_backward_step for FLUX.
280:        loss = self.forward_backward_step(input_dict=input_dict, labels=labels)

torchtitan/experiments/ft/trainer.py
427:            loss = self.forward_backward_step(

torchtitan/experiments/graph_trainer/trainer.py
70:    def forward_backward_step(
78:            return super().forward_backward_step(
97:        return self._make_fx_forward_backward_step(
135:    def _make_fx_forward_backward_step(

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
134:            loss = trainer.forward_backward_step(

torchtitan/experiments/graph_trainer/compile.py
286:        # aot_fx_trace traces fwd+loss+bwd together inside forward_backward_step,
289:        # GraphTrainer._make_fx_forward_backward_step.

torchtitan/experiments/forge/example_train.py
199:    def forward_backward_step(
298:            loss = self.forward_backward_step(

torchtitan/experiments/graph_trainer/tests/test_sac_peak_memory.py
67:    loss = trainer.forward_backward_step(

torchtitan/experiments/graph_trainer/tests/test_passes.py
528:        # One forward_backward_step triggers _make_fx_forward_backward_step
530:        trainer.forward_backward_step(
```

So, you think something like changing from
```py
 def forward_backward_step(
    self,
    *,
    input_dict: dict[str, torch.Tensor],
    labels: torch.Tensor,
    global_valid_tokens: torch.Tensor,
) -> torch.Tensor:
    model_parts = self.model_parts
    parallel_dims = self.parallel_dims

    inputs, labels, extra_inputs, extra_kwargs = self.post_dataloading_process(
        input_dict, labels
    )

[...]

loss = self.forward_backward_step(
    input_dict=input_dict,
    labels=labels,
    global_valid_tokens=global_valid_tokens,
)
```
to
```py
def forward_backward_step(
    self,
    *,
    inputs, torch.Tensor,
    labels: torch.Tensor,
    extra_inputs: dict,
    extra_kwargs: dict,
    global_valid_tokens: torch.Tensor,
) -> torch.Tensor:
    model_parts = self.model_parts
    parallel_dims = self.parallel_dims
[...]

inputs, labels, extra_inputs, extra_kwargs = self.post_dataloading_process(
    input_dict, labels
)
loss = self.forward_backward_step(
    inputs=inputs,
    labels=labels,
    extra_inputs=extra_inputs,
    extra_kwargs=extra_kwargs,
    global_valid_tokens=global_valid_tokens,
)
```
?

torchtitan/trainer.py
```diff
@@ -782,11 +787,23 @@ def train_step(
...
+            #                = (loss * global_valid_tokens) / rank_local_valid_tokens
             # global_max_loss = max(local_avg_loss)
-            local_avg_loss = loss * global_valid_tokens / local_valid_tokens
+            rank_local_valid_tokens = torch.tensor(
+                self._rank_local_valid_tokens_per_step,
+                dtype=torch.int64,
+                device=self.device,
+            )
+            # clamp avoids 0/0 when, e.g.,  a CP shard contains only masked tokens (loss=0)
+            local_avg_loss = (
```
wwwjn, line 804: The `local_avg_loss` is for logging purpose only, I think the actually backward numerics are still correct but this PR is fixing the loss logging on each rank, right?

 I agree with @fegin 's comment we should probably split CP early enough

tianyu-l:
@claude review this PR and give your suggestion on code organization

torchtitan/trainer.py
garrett361: This should be a logging-only change, yeah.

PR #3115, opened 2026-04-27
[graph_trainer] Disable chunked CE loss in graph trainer

yiming0416:
As titled. Turn it on after we support chunked CE loss in graph trainer


sanketpurandare (approved):

tianyu-l (approved):
maybe create an issue to track?

PR #3118, opened 2026-04-27
[GraphTrainer] Unified framework for activation memory management

SherlockNoMad:
## Summary
- Add `tag_with_memory_policy_pass` that unifies SAC and CPU offload tagging into a single pass, dispatching based on `config.compile.memory_policy`
- Add `cpu_offload_all` memory policy that tags all eligible activations for CPU offload (WIP, for dev/testing)
- Restructure pass pipeline: `apply_cpu_offload_pass` and `selective_activation_remat_pass` always run after tagging (no-ops when no nodes are tagged)

## Test plan
- [x] Pre-commit passes
- [x] CPU offload unit tests pass (`test_cpu_offload.py`)
- [x] Passes unit tests pass (`test_passes.py`)
- [x] Bitwise deterministic tests pass (default policy unchanged)
- [x] Manual 8-GPU run with `--compile.memory_policy cpu_offload_all` completes successfully, logs show 40 activations offloaded (222.81 MB)
- [ ] CI integration test `aot_fx_trace_llama3_fsdp_tp_cpu_offload_all`

### DIFF
--- a/torchtitan/experiments/graph_trainer/configs.py
+++ b/torchtitan/experiments/graph_trainer/configs.py
@@ -44,12 +44,14 @@ class GraphTrainerCompileConfig(CompileConfig):
     debug_graph_passes: bool = False
     """Log timing, op-count diffs, and before/after graphs for each pass to tlparse."""
 
-    memory_policy: Literal["default", "eager"] = "default"
+    memory_policy: Literal["default", "eager", "cpu_offload_all"] = "default"
     """
     Memory optimization policy for activation management (SAC, offload).
         default: save all compute-intensive ops and FSDP all_gathers.
         eager: alternate mm ops between save/recompute, matching the eager
             AC policy in torchtitan.distributed.activation_checkpoint.
+        cpu_offload_all: offload all eligible activations to CPU.
+            Work in progress — for development and testing only.
     """
 
     enable_cudagraph: bool = True
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -43,7 +43,10 @@ from torchtitan.experiments.graph_trainer.common_utils import (
     _MODULE_FQN,
     _NOT_IN_LAYERS,
 )
-from torchtitan.experiments.graph_trainer.cpu_offload import cpu_offload_pass
+from torchtitan.experiments.graph_trainer.cpu_offload import (
+    apply_cpu_offload_pass,
+    tag_all_offloadable_activations,
+)
 from torchtitan.experiments.graph_trainer.custom_codegen import custom_codegen_pass
 from torchtitan.experiments.graph_trainer.debug_utils import (
     log_graph_diff,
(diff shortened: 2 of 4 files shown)


torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -43,12 +43,14 @@ class GraphTrainerCompileConfig(CompileConfig):
     debug_graph_passes: bool = False
     """Log timing, op-count diffs, and before/after graphs for each pass to tlparse."""
 
-    memory_policy: Literal["default", "eager"] = "default"
+    memory_policy: Literal["default", "eager", "cpu_offload_all"] = "default"
```
xmfan, line 47: how should a user express a combination of offload + SAC

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -43,7 +43,11 @@
     _MODULE_FQN,
     _NOT_IN_LAYERS,
 )
-from torchtitan.experiments.graph_trainer.cpu_offload import cpu_offload_pass
+from torchtitan.experiments.graph_trainer.cpu_offload import (
```
yiming0416, line 46: IIUC, `cpu_offload_pass` = `tag_all_offloadable_activations` + `apply_cpu_offload_pass`

In the current code, we didn't call `cpu_offload_pass` anymore. Shall we remove it? And we can also clean up the APIs in `cpu_offload.py`

torchtitan/experiments/graph_trainer/configs.py
yiming0416, line 47: same question: if user use `"default"`, `tag_all_offloadable_activations` won't be called, then `apply_cpu_offload_pass` is a no-op?

SherlockNoMad: I expect user to implement a policy_fn, and register it. The framework doesn't exist yet, but I expect something like this to work 

```
@register_policy
def my_tagging_pass(gm)
    ... 
```

We should provide our own, e.g. based on Sanket's ILP solver.

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: removed.

torchtitan/experiments/graph_trainer/configs.py
SherlockNoMad: yes, it's ok to run a no-op pass.

SherlockNoMad: annotate_flex_attention_for_regional_inductor_pass is also an no-op pass for model not using flex

yiming0416: so right now we can't have SAC and CPU offloading at the same time?

SherlockNoMad: yea, we don't have a policy_fn that would use SAC and CPU offload at the same time yet...

SherlockNoMad: added comment.

yiming0416 (approved):

Issue #3082, opened 2026-04-24
moe_comm_backend requires EP > 1


githubsgi:
### Bug description

Looks like this  [commit](https://github.com/pytorch/torchtitan/commit/7cec16600a6bb9b0f4be91bfdcaa60a4eb55f5cd) made significant changes in EP.  

One consequence is that EP has to be >1 where moe_comm_backend is explicitly added to config_registry.    This I would consider a bug, 

[rank0]: AssertionError: ep_mesh must be set before dispatch. ExpertParallel._partition_fn() should set it.

### Versions

head

tianyu-l:
@acisseJZhong we should still allow MoE to be used without EP, if that's what's being complained.

acisseJZhong:
@githubsgi can you share command and config to repro? EP=1 should pair with moe backend None which is the default.

githubsgi:
`REPO=$HOME/torchtitan MODEL=deepseek_v3 FLAVOR=16b  N=1 PPN=8 TP=8 CP=1 EP=1;   PYTHONPATH=$REPO:$PYTHONPATH torchrun   --nnodes $N --nproc-per-node $PPN --no-python python -m torchtitan.train --module $MODEL --config ${MODEL}_${FLAVOR}  --parallelism.tensor_parallel_degree ${TP}   --parallelism.context_parallel_degree ${CP} --parallelism.expert_parallel_degree ${EP} --parallelism.expert_parallel_comm_backend=standard  --training.steps 10  --dataloader.dataset=c4_test`

acisseJZhong:
you don't need to pass in --parallelism.expert_parallel_comm_backend=standard. It should be None/default.

githubsgi:
Yes, that is another question i had.  Why  duplicate configs for EP ( e.g. parallelism.expert_parallel_comm_backend and  moe_comm_backend="standard" ) ?  

Also , in the deepseek config_registry it is set to standard  ( "16B", attn_backend="flex", moe_comm_backend="standard" ) . I know it is easy to change that, but default  config should work OOB . 

Same result after removing --parallelism.expert_parallel_comm_backend=standard .

acisseJZhong:
> Why duplicate configs for EP ( e.g. parallelism.expert_parallel_comm_backend and moe_comm_backend="standard" )

[...]

PR #2960, opened 2026-04-14
[MoE][2/n]Move EP setup from trainer to config registry and add model_registry params

acisseJZhong:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #2960
* * __->__ #2842

## Summary

Move EP-aware token dispatcher setup from runtime (`apply_ep()` in `Trainer.__post_init__`) to config time (`model_registry(moe_comm_backend=...)`), so config registries build the correct dispatcher upfront.

- **Remove `apply_ep()` from trainer** — Delete the `apply_ep()` call in `Trainer.__post_init__` and the function itself from `config_utils.py`. Config registries now pass `moe_comm_backend` directly to `model_registry()` → `make_experts_config()` → `make_token_dispatcher_config()`.
- **Simplify `make_token_dispatcher_config()`** — Replace `ep_size`/`sp_size`/`comm_backend`/`pad_multiple`/`hybridep_non_blocking_expert_capacity_factor` params with a single `moe_comm_backend` string (`"local"`, `"standard"`, `"torchao"`, `"deepep"`, `"hybridep"`).
- **Remove `ep_size` from dispatcher Configs** — `AllToAllTokenDispatcher`, `TorchAOTokenDispatcher`, and `DeepEPTokenDispatcher` no longer store `ep_size`; derived at runtime from `ep_group.size()`. `sp_size` and `pad_multiple` similarly moved to runtime-set attributes.
- **Add `get_attention_config()` helper** — Centralizes `attn_backend` → `(inner_attention, mask_type)` mapping, replacing duplicated logic across model families.
- **Add `model_registry(moe_comm_backend=...)` to all models** — llama3, llama4, deepseek_v3, qwen3, qwen3_vl, gpt_oss all accept `moe_comm_backend` to select the dispatcher at config time.

PR #3125, opened 2026-04-28
[MoE][3/n] consolidate EP=1 and EP>1 to all use All2AllTokenDispatcher

acisseJZhong:
* __->__ #3125
* * __->__ #2960

## Summary

Unify EP=1 and EP>1 token dispatching by making `AllToAllTokenDispatcher` fall back to `LocalTokenDispatcher` behavior when `ep_mesh is None`.

### Changes

- **`AllToAllTokenDispatcher.dispatch/combine`**: When `ep_mesh is None` (EP=1), falls back to `super().dispatch()`/`super().combine()` (LocalTokenDispatcher) — no all-to-all communication, just local token reordering.
- **`make_token_dispatcher_config`**: Default `comm_backend` changed from `None` to `"standard"`. Removed the `None` → `LocalTokenDispatcher` path. All MoE models now use `AllToAllTokenDispatcher` by default.
- **All model `__init__.py` files**: `moe_comm_backend` default changed from `str | None = None` to `str = "standard"`.
- **All `model_registry` functions**: Removed `if moe_comm_backend is not None` guards — `moe_comm_backend` is always passed through directly.
- **All config_registry files**: Removed redundant `moe_comm_backend="standard"` since it's now the default.

### Why

Previously, users had to manually set `moe_comm_backend=None` when overriding `expert_parallel_degree=1` on the command line, otherwise the `AllToAllTokenDispatcher` would assert `ep_group is not None`(see issue in https://github.com/pytorch/torchtitan/issues/3082,). Now the dispatcher handles EP=1 gracefully by falling back to local dispatch.

### Test command:
```
NCCL_NVLS_ENABLE=0 torchrun --nnodes 1 --nproc-per-node 8 -m torchtitan.train --module deepseek_v3 --config deepseek_v3_16b --parallelism.tensor_parallel_degree 8 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 1 --training.steps 10 --dataloader.dataset c4_test
```


1
tianyu-l, line 1: remove?

torchtitan/experiments/graph_trainer/llama3/__init__.py
```diff
@@ -18,12 +18,11 @@
 def model_registry(
     flavor: str,
     attn_backend: str = "sdpa",
-    moe_comm_backend: str | None = None,
+    moe_comm_backend: str = "standard",
```
tianyu-l: non moe model shouldn't have this field

torchtitan/experiments/graph_trainer/qwen3/__init__.py
```diff
@@ -18,12 +18,11 @@
 def model_registry(
     flavor: str,
     attn_backend: str = "sdpa",
-    moe_comm_backend: str | None = None,
+    moe_comm_backend: str = "standard",
```
tianyu-l: maybe need to be optional for qwen3, as it has non moe variants

we can assert its existence inside qwen3 moe variants

torchtitan/models/common/config_utils.py
```diff
@@ -199,23 +198,17 @@ def make_token_dispatcher_config(
     num_experts: int,
     top_k: int,
     score_before_experts: bool = True,
-    comm_backend: str | None = None,
+    comm_backend: str = "standard",
     non_blocking_capacity_factor: float | None = None,
-) -> LocalTokenDispatcher.Config:
+) -> AllToAllTokenDispatcher.Config:
```
tianyu-l: return type should be common base class, or union of classes

torchtitan/models/common/token_dispatcher.py
```diff
@@ -451,14 +455,15 @@ def combine(
...
+    """Token dispatcher with token group padding for quantized grouped GEMMs.
 
     Uses torchao's ``permute_and_pad`` instead of the standard ``_permute`` to
     reorder tokens into expert-major order and pad each expert's token group to
     a multiple of ``pad_multiple``. This alignment is required by FP8/MXFP8
     quantized grouped GEMM kernels (e.g. 16 for FP8, 32 for MXFP8).
 
-    ep_mesh is set by ExpertParallel / ExpertTensorParallel._apply().
+    When EP is enabled (ep_mesh is set), uses all-to-all communication.
+    When EP=1 (ep_mesh is None), falls back to local dispatch.
```
tianyu-l: When EP=1, there's no permute / unpermute, so I don't think it works with quantized grouped gemm. We just need to error out somewhere properly.

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -171,7 +171,7 @@ def _build_dsv3_layers(
     router_route_norm: bool = False,
     score_before_experts: bool = False,
     attn_backend: str,
-    moe_comm_backend: str | None = None,
+    moe_comm_backend: str = "standard",
```
tianyu-l: don't need default here?

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -447,7 +447,7 @@ def _236b(
 
 def _671b(
     attn_backend: str = "flex",
-    moe_comm_backend: str | None = None,
+    moe_comm_backend: str = "standard",
```
tianyu-l: also, don't need these defaults, as they are always given by `model_registry`

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -124,7 +124,7 @@ def _make_gptoss_experts_config(
     layer_id: int,
     top_k: int,
     score_before_experts: bool,
-    moe_comm_backend: str | None = None,
+    moe_comm_backend: str = "standard",
```
tianyu-l: also don't need default

torchtitan/models/llama3/__init__.py
```diff
@@ -104,7 +104,7 @@ def _build_llama3_layers(
     return layers
 
 
-def _debugmodel(attn_backend: str = "sdpa") -> Llama3Model.Config:
+def _debugmodel(attn_backend: str = "sdpa", moe_comm_backend: str = "standard") -> Llama3Model.Config:
```
tianyu-l: there shouldn't be moe backend for dense models

1
acisseJZhong: oops claude is doing sth weird. I removed it once, but seems everytime I commit, it re appears :(

torchtitan/experiments/graph_trainer/qwen3/__init__.py
acisseJZhong: restore this back to `moe_comm_backend: str | None = None`; and only MoE model builders will have this `moe_comm_backend` which defaults to `standard`.

torchtitan/models/common/token_dispatcher.py
acisseJZhong: if we want to support quantization for MoE EP=1, I believe it's possible if we set padding needs to be handled by torchao? However I think per last discussion, we would prefer banning EP=1 with torchao dispatcher? I will error out for now. cc @danielvegamyhre

danielvegamyhre: Yeah that sounds good to me

torchtitan/models/deepseek_v3/__init__.py
acisseJZhong: rethinking this, maybe we should get rid of the default here for both attn_backend and moe_comm_backend as both are provided through model_registry.

torchtitan/models/common/config_utils.py
```diff
@@ -199,23 +199,17 @@ def make_token_dispatcher_config(
     num_experts: int,
     top_k: int,
     score_before_experts: bool = True,
-    comm_backend: str | None = None,
+    comm_backend: str = "standard",
```
tianyu-l: model_registry should be the only function that has default for this field? o/w when would the default be used here?

torchtitan/models/qwen3_vl/__init__.py
```diff
@@ -283,7 +289,8 @@ def _debugmodel() -> Qwen3VLModel.Config:
 
 
 def _debugmodel_moe(
-    moe_comm_backend: str | None = None,
+    attn_backend: str,
+    moe_comm_backend: str = "standard",
```
tianyu-l, line 293: similar

tianyu-l (approved):

acisseJZhong:
failing test not caused by this PR, landing

PR #3126, opened 2026-04-28
call correct parallelize function for trainer

liangel-02:
after https://github.com/pytorch/torchtitan/pull/2969 refactor, the core `parallelize_qwen3` depends on sharding configs being set via `update_from_config()` [here](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/qwen3/model.py#L125) (which calls `set_qwen3_sharding_config`). The RL trainer never calls update_from_config(), so the trainer gets no TP. but since TP is set for the generator, you get different numerics.


tianyu-l (changes requested):
is this being addressed in https://github.com/pytorch/torchtitan/pull/3070

PR #3127, opened 2026-04-28
quantize on config instead of on model

acisseJZhong:
Continuation of #3032 (accidentally merged to wrong base branch). See #3032 for full description.


tianyu-l (approved):

torchtitan/experiments/transformers_modeling_backend/parallelize.py
```diff
@@ -168,33 +151,11 @@ def apply_non_moe_tp(
     if root_plan:  # Only call if there's something to parallelize
         parallelize_module(model, tp_mesh, root_plan)
 
-    # Parallel styles used for transformer block linear weights and their
-    # inputs may be different for float8 linears with tensorwise scaling.
-    if enable_float8_tensorwise_tp:
```
zpw1337, line 173: Sorry for posting comment on merged PR.

But I'm curious why tensorwise was removed? We currently run an older version of torchtitan and use this, and we observe MFU drop when switching to rowwise, consistent with documentation: https://docs.pytorch.org/ao/stable/workflows/training.html

@tianyu-l @acisseJZhong

tianyu-l:
@zpw1337 for sure tensorwise can give faster MFU, so it's not surprise. The interesting part is the tradeoff between MFU and numerics (tensorwise is "less accurate" compared with rowwise / blockwise).

Can I ask if you have evidence that tensorwise gives better model quality?

zpw1337:
> @zpw1337 for sure tensorwise can give faster MFU, so it's not surprise. The interesting part is the tradeoff between MFU and numerics (tensorwise is "less accurate" compared with rowwise / blockwise).
> 
> Can I ask if you have evidence that tensorwise gives better model quality?

Thanks for the reply. I'm not implying tensorwise gives better model quality. It's just that we've been training models with preference of higher MFU. I'm wondering it would reasonable to expose both and let the user choose like before?

tianyu-l:
> It's just that we've been training models with preference of higher MFU. I'm wondering it would reasonable to expose both and let the user choose like before?

This is where I'm interested in learning convincing benefits over blindly supporting as many features as possible.

Could you show at least one tech report / blog post which says one could train a good model with tensorwise fp8?

YassineYousfi:
@tianyu-l 

> Could you show at least one tech report / blog post which says one could train a good model with tensorwise fp8?

We trained video DiTs up to 4B params dense with tensorwise fp8, we are writing a report + open sourcing the code for them in the next couple weeks. We haven't seen much numerics degradation with tensorwise, but we don't mind moving to row/blockwise for our next bigger runs if that does not hurt MFU much. 

With that said, the diff to add back support for tensorwise + enable_fsdp_float8_all_gather + precompute_float8_dynamic_scale_for_fsdp is fairly small.

PR #3129, opened 2026-04-28
[graph_trainer] Add async tensor parallel (micro-pipeline TP) graph pass

IvanKobzarev:
Adds `enable_async_tensor_parallel` config to GraphTrainerCompileConfig and applies the upstream `micro_pipeline_tp_pass` as an FX graph pass in `aot_fx_trace` mode.

The pass fuses all-gather + matmul (AG+MM) and matmul + reduce-scatter (MM+RS) patterns using symmetric memory for overlapped communication and computation.

Key implementation detail: the make_fx traced graph produces `aten.view.default` and `aten._unsafe_view.default`, but the upstream micro_pipeline_tp_pass pattern matcher expects `aten.reshape.default`. A normalization step (`_normalize_views_to_reshape`) converts these before running the pass, enabling successful pattern matching.

Verified on llama3 8B TP=4: 576 AG+MM and 489 MM+RS fusions applied.

Usage:
  --compile.enable_async_tensor_parallel

Authored with Claude.

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -64,6 +64,66 @@ from torchtitan.experiments.graph_trainer.reshard_after_forward import (
 )
 from torchtitan.tools.logging import logger
 
+aten = torch.ops.aten
+c10d = torch.ops._c10d_functional
+
+
+def normalize_view_ops_as_reshape(
+    gm: torch.fx.GraphModule,
+    example_inputs=None,
+) -> torch.fx.GraphModule:
+    """Replace aten.view and aten._unsafe_view with aten.reshape.
+
+    Downstream passes expect aten.reshape.default for pattern matching.
+    """
+    view_targets = {aten.view.default, aten._unsafe_view.default}
+    for node in gm.graph.nodes:
+        if node.op == "call_function" and node.target in view_targets:
+            node.target = aten.reshape.default
+    gm.graph.lint()
+    gm.recompile()
+    return gm
+
+
+def async_tensor_parallel_pass(
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple,
+) -> torch.fx.GraphModule:
+    """Pipeline TP collectives with matmuls via symmetric memory.
+
+    Fuses all-gather + matmul into ``symm_mem.fused_all_gather_matmul``
+    and matmul + reduce-scatter into
+    ``symm_mem.fused_matmul_reduce_scatter``.
+    """
+    import warnings
+
+    from torch._inductor.fx_passes.micro_pipeline_tp import micro_pipeline_tp_pass
+    from torch._inductor.fx_passes.overlap_scheduling import get_group_name
+    from torch.distributed._symmetric_memory import enable_symm_mem_for_group
+
+    # Ensure symmetric memory is registered for every collective PG in
+    # the graph.  The upstream API is deprecated but the auto-registration
+    # it promises has not landed yet, so the explicit call is still needed.
+    collective_targets = {
+        c10d.all_gather_into_tensor.default,
+        c10d.reduce_scatter_tensor.default,
+    }
+    registered: set[str] = set()
+    for node in gm.graph.nodes:
+        if node.target not in collective_targets:
+            continue
+        pg = get_group_name(node)
+        if pg not in registered:
+            registered.add(pg)
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", FutureWarning)
+                enable_symm_mem_for_group(pg)
+
+    micro_pipeline_tp_pass(gm.graph)
+    gm.graph.lint()
+    gm.recompile()
+    return gm
+
 
 def compile_time_passes(
     traced_result: "TracedResult",
@@ -90,6 +150,7 @@ def compile_time_passes(
         remove_detach_pass,
         remove_identity_view_pass,
         remove_identity_slice_pass,
+        normalize_view_ops_as_reshape,
         functools.partial(
             tag_with_memory_policy_pass,
             config=config,
@@ -103,6 +164,8 @@ def compile_time_passes(
             module_bucket_plans=get_default_transformer_block_buckets(n_layers),
         ),
     ]
+    if config.parallelism.enable_async_tensor_parallel:
+        passes.append(async_tensor_parallel_pass)
 
     inductor_compilation = config.compile.inductor_compilation
     if inductor_compilation == "full":
--- a/torchtitan/experiments/graph_trainer/tests/integration_tests.py
+++ b/torchtitan/experiments/graph_trainer/tests/integration_tests.py
@@ -320,6 +320,7 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
             ngpu=8,
             skip_rocm_test=True,
         ),
+        # async_tp test lives in graph_trainer_h100 suite (needs NVLink).
         OverrideDefinitions(
             [
                 [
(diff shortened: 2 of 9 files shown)


torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -54,6 +54,9 @@ class GraphTrainerCompileConfig(CompileConfig):
             Work in progress — for development and testing only.
     """
 
+    enable_async_tensor_parallel: bool = False
```
SherlockNoMad: do we need to introduce a new flag, shall we reuse --parallelism.enable_async_tensor_parallel

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+    view_targets = {aten.view.default, aten._unsafe_view.default}
+    count = 0
+    for node in graph.nodes:
+        if node.op == "call_function" and node.target in view_targets:
+            node.target = aten.reshape.default
+            count += 1
+    return count
+
+
+def async_tensor_parallel_pass(
```
SherlockNoMad, line 88: Let's also have unit test

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple,
+    **kwargs,
```
SherlockNoMad: don't need kwargs?

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad, line 88: Let's also add some instructions to claude.md, on when to use async tp. share your heuristics.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+        micro_pipeline_tp_pass,
+    )
+    from torch._inductor.fx_passes.overlap_scheduling import (
+        get_group_name,
+    )
+    from torch.distributed._symmetric_memory import (
+        enable_symm_mem_for_group,
+    )
+
+    _normalize_views_to_reshape(gm.graph)
```
SherlockNoMad: let's not modify the entire graph to replace view into reshape. 

Pass should act on local region to maximize compatibility with other passes. 

Other passes may rely on the fact that it has views...

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+
+    c10d = torch.ops._c10d_functional
+    collective_targets = {
+        c10d.all_gather_into_tensor.default,
+        c10d.reduce_scatter_tensor.default,
+    }
+    seen_groups: set[str] = set()
+    for node in gm.graph.nodes:
+        if node.op != "call_function":
```
SherlockNoMad: redundant check? 
target in collective_targets is sufficient.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+            continue
+        if node.target not in collective_targets:
+            continue
+        gn = get_group_name(node)
+        if gn not in seen_groups:
+            seen_groups.add(gn)
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", FutureWarning)
+                enable_symm_mem_for_group(gn)
```
SherlockNoMad: what's the assumption of AG overlapping with RS in the backward region, when reshard_after_forward is true? 

Does this require AG and RS to be on different PG, or single PG ? 

FYI, see reassign_to_pg_pass, how's the composability with it?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+    seen_groups: set[str] = set()
+    for node in gm.graph.nodes:
+        if node.op != "call_function":
+            continue
+        if node.target not in collective_targets:
+            continue
+        gn = get_group_name(node)
+        if gn not in seen_groups:
+            seen_groups.add(gn)
+            with warnings.catch_warnings():
```
SherlockNoMad, line 118: what is this needed? add comment?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,62 @@
...
+            count += 1
+    return count
+
+
+def async_tensor_parallel_pass(
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple,
+    **kwargs,
+) -> torch.fx.GraphModule:
+    """Fuse AG+MM and MM+RS patterns via micro-pipeline TP."""
```
SherlockNoMad: a few more word on how the pass will transform the graph? 
should user expect new fused op, make the contract explicit.

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -321,6 +321,26 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
             ngpu=8,
             skip_rocm_test=True,
         ),
+        OverrideDefinitions(
```
SherlockNoMad, line 324: In addition to integration test, we should also add unit test.

SherlockNoMad (changes requested):
thanks!!

torchtitan/experiments/graph_trainer/configs.py
IvanKobzarev: Yes, we can reuse.

torchtitan/experiments/graph_trainer/passes.py
IvanKobzarev: Existing inductor post_passes rely on decompositions, that we will have reshape only.

torchtitan/experiments/graph_trainer/passes.py
IvanKobzarev: This should be agnostic to how the graph was made, with reshard_after_forward or etc. It just looks for patterns ag+mm and mm+rs on any group that can use symmetric memory (nvlink connected).
I will recheck what is reassign_to_pg_pass

torchtitan/experiments/graph_trainer/passes.py
IvanKobzarev: Technically there is some autoregistration of pgs with symmetric memory, that we do not need enable_symm_mem(pg), and it produces warning on calling it. But in reality autoregistration did not happen for me, and patterns were not applied without this enable_symm_mem. So I am suppressing warning to avoid annoying logs every time :)

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,65 @@
 from torchtitan.tools.logging import logger
 
 
+def normalize_view_ops_as_reshape(
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple = (),
```
SherlockNoMad: follow other passes signature, default to None

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,65 @@
...
+) -> torch.fx.GraphModule:
+    """Replace aten.view and aten._unsafe_view with aten.reshape.
+
+    Downstream passes expect aten.reshape.default for pattern matching.
+    """
+    aten = torch.ops.aten
```
SherlockNoMad: aten = torch.ops.aten 

lift as common, will be useful for other passes

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,65 @@
...
+    import warnings
+
+    from torch._inductor.fx_passes.micro_pipeline_tp import micro_pipeline_tp_pass
+    from torch._inductor.fx_passes.overlap_scheduling import get_group_name
+    from torch.distributed._symmetric_memory import enable_symm_mem_for_group
+
+    # Ensure symmetric memory is registered for every collective PG in
+    # the graph.  The upstream API is deprecated but the auto-registration
+    # it promises has not landed yet, so the explicit call is still needed.
+    c10d = torch.ops._c10d_functional
```
SherlockNoMad: same here

yiming0416 (commented):
If you look at the `parallelize.py` of models in graph trainer. We have the `maybe_enable_async_tp` path, which reads `parallelism.enable_async_tensor_parallel`.  But that path was never tested.

Given this PR, shall we remove `maybe_enable_async_tp` function in the `parallelize.py`?

torchtitan/experiments/graph_trainer/tests/test_passes.py
```diff
@@ -1263,6 +1263,68 @@ def test_insert_kernel_annotations_pass_noop_without_metadata(self):
...
+
+
+class TestAsyncTensorParallelPass(FSDPTest):
+    """Verify async_tensor_parallel_pass produces fused ops."""
+
+    @property
+    def world_size(self):
+        return 2
+
+    def test_ag_mm_becomes_fused_op(self):
```
SherlockNoMad, line 1339: what about reduce scatter case?

torchtitan/experiments/graph_trainer/tests/test_passes.py
```diff
@@ -1263,6 +1263,68 @@ def test_insert_kernel_annotations_pass_noop_without_metadata(self):
...
+        from torch.distributed._symmetric_memory import _test_mode
+
+        from torchtitan.experiments.graph_trainer.passes import (
+            async_tensor_parallel_pass,
+        )
+
+        pg = torch.distributed.distributed_c10d._get_default_group().group_name
+        aten, c10d = torch.ops.aten, torch.ops._c10d_functional
+
+        # shard[2048,4096] -> all_gather -> wait -> mm(w[4096,1024])
```
SherlockNoMad, line 1349: also, I didn't see use of reshape? 
include the pattern here?

SherlockNoMad (approved):
thanks, lgtm per comments.

SherlockNoMad:
also fyi, will need to add "ciflow/h100.8" to manually trigger h100.8 CI

IvanKobzarev:
> follow other passes signature, default to None

Yes, good catch, `maybe_enable_async_tp` was based on running full inductor compilation, so it enabled inductor config and enabled symm_memm for pgs. I will remove it.

PR #3131, opened 2026-04-28
fix(hf_datasets): shuffle ChatDataset before splitting across nodes

garrett361:
Adds an initial shuffle to `ChatDataset` (mirroring the end-of-epoch shuffles the class already does) to promote even token distribution across nodes.

Without this, I was finding that `split_dataset_by_node` can give highly-imbalanced examples to different nodes, due to the default behavior where it splits map-style datasets sequentially.  This is problematic when the dataset has some non-trivial ordering bias, e.g. if shorter examples tend to be early in the dataset. Can lead to node-0, say, getting way fewer tokens than other nodes and completing its initial epoch early, an offset that persists for the remainder of training.

### DIFF
--- a/torchtitan/hf_datasets/text_datasets.py
+++ b/torchtitan/hf_datasets/text_datasets.py
@@ -311,7 +311,7 @@ class ChatDataset(IterableDataset, Stateful):
         # datasets, split_dataset_by_node assigns contiguous data chunks to consecutive nodes, which
         # can lead to token imbalances, causing some nodes' epoch_idx to run ahead of others.
         self._original_data = split_dataset_by_node(
-            dataset.shuffle(seed=42), dp_rank, dp_world_size
+            cast(Dataset, dataset.shuffle(seed=42)), dp_rank, dp_world_size
         )
         self._data = self._original_data
         self._tokenizer = tokenizer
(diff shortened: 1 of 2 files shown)


garrett361:
Tested w/ custom manual print statements.

torchtitan/hf_datasets/text_datasets.py
```diff
@@ -300,19 +300,26 @@ def __init__(
         dp_rank: int = 0,
         dp_world_size: int = 1,
         infinite: bool = False,
+        seed: int = 42,
```
tianyu-l: Add initial shuffling and introducing this seed are two different things. Do you need both in this PR? If not let's not add this seed in this PR -- Don't want too many sources of randomness before they prove to be useful.

garrett361: This just moves the hard code of seed=42 that already existed in the end-of-epoch shuffle to a default arg value. And then reuses it for the initial shuffle. 

Could also hard code 42 as the base seed at both shuffle sites, but those kind of hard codes seem strictly worse to me. Wdyt?

tianyu-l: The seed is not used by any config, so with / without hardcoding the behavior is the same. I would rather delay the introduction of this separate seed until its first meaningful usage.

garrett361: Hard coded to 42 everywhere.

tianyu-l (approved):

tianyu-l:
pls take a look at CI failures, CPU and lint

garrett361:
Fixed the lint failure related to this change (also seeing qwen3/qwen3_vl lint failure, but didn't address those).

Also adjusted the failing test. All of `tests/unit_tests/test_chat_dataset.py` is passing for me locally now.

PR #3132, opened 2026-04-28
[GraphTrainer][AutoDev] Fuse RMSNorm kernels via regional Inductor compilation

SherlockNoMad:
## Summary

- Adds `annotate_rmsnorm_for_regional_inductor_pass`, a graph pass that tags RMSNorm forward and backward ops for regional Inductor compilation via `compile_with_inductor` node metadata
- RMSNorm appears as fused `_fused_rms_norm` / `_fused_rms_norm_backward` ops in the traced graph. Compiling each norm region with Inductor fuses them into a single Triton kernel, closing the performance gap vs Megatron's fused TransformerEngine RMSNorm
- Identifies RMSNorm nodes by `node.target` (simple and direct) and tags their `getitem` users so the full fused op is compiled together

## How it works

1. `annotate_rmsnorm_for_regional_inductor_pass` walks the traced graph and tags nodes whose `node.target` is `_fused_rms_norm.default` or `_fused_rms_norm_backward.default`
2. `getitem` users of each fused norm op are also tagged (they extract the tuple outputs)
3. The existing `regional_inductor_pass` compiles tagged regions with Inductor's `standalone_compile`
4. The `CapabilityBasedPartitioner` in `regional_inductor` automatically separates disconnected norm subgraphs into distinct compiled partitions
5. Inductor configs match FlexAttention: `max_autotune`, `coordinate_descent_tuning`, `wrap_inductor_compiled_regions`

Sample TLP: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/.tmp27Bxjy/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

## Microbenchmark

Isolated RMSNorm fwd+bwd (single layer, H100):

| Path | ms/iter (run 1) | ms/iter (run 2) |
|------|-----------------|-----------------|
| Eager | 26.13 | 25.73 |
| make_fx (uncompiled) | 21.99 | 22.03 |
| regional_inductor | 14.35 | 14.17 |
| **Speedup vs eager** | **1.82x** | **1.82x** |

## Benchmark

Llama3 8B, 8×H100, FSDP4+TP2, `c4_test`, 20 steps (last step steady-state):

| Branch | tps | tflops | MFU | Memory |
|--------|-----|--------|-----|--------|
| `main` | 6,538 | 378.66 | 38.29% | 30.92 GiB |
| this PR | 6,600 | 382.26 | **38.65%** | 30.92 GiB |

**Delta**: +0.36% MFU, +0.9% tps, no memory regression.

## Numerics

This may change numerics slightly (Inductor kernel numerics vs native ATen fused ops), which is expected and acceptable.

## Test plan

- [x] 8 CPU unit tests in `TestAnnotateRMSNormForRegionalInductorPass` covering:
  - Correct tagging of fused RMSNorm nodes and getitem users
  - Non-RMSNorm nodes are not tagged
  - Multiple norms all tagged
  - Forward and backward both tagged
  - Empty graph is a no-op
  - None compile config produces empty annotation
  - Config dict propagation
- [x] Bitwise deterministic tests pass (RMSNorm Inductor compilation patched to no-op for eager-vs-traced comparisons since the pass intentionally changes numerics)
- [x] GPU benchmark comparison showing fused kernel performance


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -565,6 +601,84 @@ def cudagraph_pass(
     return gm
 
 
+def annotate_rmsnorm_for_regional_inductor_pass(
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple | None = None,
+    *,
+    rmsnorm_compile_config: dict | None,
+    rmsnorm_fqns: set[str],
+) -> torch.fx.GraphModule:
```
SherlockNoMad: Too complicated. 

Just use node.target to identify, see sample. 

Remember to include the subsequent getitem node. 

```
        _fused_rms_norm_6_recomputed = torch.ops.aten._fused_rms_norm.default(add_6, [256], _unsafe_view_101, 1e-05)
        getitem_66_recomputed: "f32[8, 1024, 1][1024, 1, 1]cuda:0" = _fused_rms_norm_6_recomputed[1];  _fused_rms_norm_6_recomputed = None
```
 and 

```
        _fused_rms_norm_backward_6 = torch.ops.aten._fused_rms_norm_backward.default(wait_tensor_127, add_7_recomputed, [256], getitem_81_recomputed, _unsafe_view_106, [True, True]);  wait_tensor_127 = add_7_recomputed = getitem_81_recomputed = _unsafe_view_106 = None
        getitem_189: "bf16[8, 1024, 256][262144, 256, 1]cuda:0" = _fused_rms_norm_backward_6[0]
        getitem_190: "bf16[256][1]cuda:0" = _fused_rms_norm_backward_6[1];  _fused_rms_norm_backward_6 = None```

File: torchtitan/experiments/graph_trainer/passes.py [hunk 2]
```diff
@@ -565,6 +601,84 @@ def cudagraph_pass(
     return gm
 
 
+def annotate_rmsnorm_for_regional_inductor_pass(
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple | None = None,
+    *,
+    rmsnorm_compile_config: dict | None,
+    rmsnorm_fqns: set[str],
```
SherlockNoMad: remove rmsnorm_fqns.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -109,6 +136,15 @@ def compile_time_passes(
             annotate_flex_attention_for_regional_inductor_pass,
             flex_compile_config=FlexAttention.inductor_configs,
         ),
+        # RMSNorm decomposes into small ATen ops (pow, mean, rsqrt, mul)
```
SherlockNoMad: This is incorrect, we see _fused_rms_norm

SherlockNoMad:
Benchmark llama3_8b, compare MFU before and after this pass. Show proof of work.

```
    --compile.mode aot_fx_trace \
    --comm.trace_buf_size=0 \
    --training.steps 20
```

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: AutoDev: Simplified. Now matches `torch.ops.aten._fused_rms_norm.default` and `torch.ops.aten._fused_rms_norm_backward.default` via `node.target`, plus their `getitem` user nodes. Removed `_collect_rmsnorm_fqns` and all FQN-based identification.

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: AutoDev: Removed. The `rmsnorm_fqns` parameter and `_collect_rmsnorm_fqns` function are both deleted. The pass now identifies RMSNorm ops purely by `node.target`.

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: AutoDev: Fixed. The comment now correctly states that RMSNorm appears as fused `_fused_rms_norm` / `_fused_rms_norm_backward` ops in the traced graph, not decomposed small ATen ops.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -64,6 +64,18 @@
...
 
+# Default Inductor configs for RMSNorm regional compilation.
+# Same tuning flags as FlexAttention: Inductor max_autotune + coordinate
+# descent discover good Triton kernels, then wrap them for cudagraph
+# compatibility.
+_RMSNORM_INDUCTOR_CONFIGS: dict[str, bool] = {
+    "wrap_inductor_compiled_regions": True,
+    "max_autotune": True,
+    "coordinate_descent_tuning": True,
+    "triton.cudagraphs": False,
```
aditvenk: why is this False?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -109,6 +121,14 @@ def compile_time_passes(
             annotate_flex_attention_for_regional_inductor_pass,
             flex_compile_config=FlexAttention.inductor_configs,
         ),
+        # RMSNorm appears as fused _fused_rms_norm / _fused_rms_norm_backward
```
aditvenk: Isn't this the case only if model definition uses torch.nn.functional.rms_norm?

SherlockNoMad:
add test for run-to-run bitwise deterministic.

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
```diff
@@ -465,6 +518,24 @@ def test_precompile_vs_trace(self):
 
         self._assert_runs_match(run_traced, run_precompile, "trace vs precompile: ")
 
+    # TODO: OOMs during flex_attention compilation on A100 GPUs.
```
aditvenk, line 525: can we file an issue for this?  it's surprising to OOM for debug model

torchtitan/experiments/graph_trainer/performance_passes.py
```diff
@@ -0,0 +1,71 @@
...
+            Default is None (no inductor_configs in the annotation).
+    """
+    compile_annotation: dict = (
+        {"inductor_configs": rmsnorm_compile_config}
+        if rmsnorm_compile_config is not None
+        else {}
+    )
+
+    _RMSNORM_TARGETS = {
+        torch.ops.aten._fused_rms_norm.default,
```
aditvenk, line 49: Why key off of these nodes being present instead of regional inductor annotation around whatever module is doing RMSNorm? This approach will also work for cases where the rms norm is written in decomposed fashion

aditvenk (approved):

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: yes, it only cover such case.  
We currently don't decompose in minimal_make_fx. Will update comment to mention this.

torchtitan/experiments/graph_trainer/performance_passes.py
SherlockNoMad: Let's see if we can find an reliable way to annotation around RMSNorm. cc tried, and didn't find so.

SherlockNoMad: First commit in this PR [[GraphTrainer] Add RMSNorm regional Inductor annotation pass](https://github.com/pytorch/torchtitan/pull/3132/changes/b1294c0cdb7b95c17cc3ca10002c284715d7d925)

was the attempt, and I found it unsatisfactory.

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
SherlockNoMad: https://github.com/orgs/pytorch/projects/161/views/4?filterQuery=assignee%3A&pane=issue&itemId=182831982

PR #3134, opened 2026-04-28
[GraphTrainer][AutoDev] Fuse RoPE kernels via regional Inductor compilation

SherlockNoMad:
## Summary

- Annotate RoPE free functions (`apply_rotary_emb_complex`, `apply_rotary_emb_cos_sin`, `apply_rotary_emb_single_complex`) at tracing time via `annotate_fn` so traced FX nodes carry a `rope_region` tag in `node.meta["custom"]`
- Add `annotate_rope_for_regional_inductor_pass` graph pass that converts `rope_region` tags into `compile_with_inductor` annotations consumed by `regional_inductor_pass`
- Each attention layer's RoPE ops get a separate `inductor_region` ID so Inductor compiles them as independent fused kernels

### Why not FQN-based matching?

Unlike RMSNorm (which is an `nn.Module` with its own FQN), RoPE computation is done by free functions called inside `GQAttention.forward()`. All RoPE ops inherit the attention module's FQN (e.g., `layers.0.attention`), making FQN-based identification impossible. The two-step approach (tracing-time annotation + graph pass) solves this by tagging at the source.

### Backward propagation

`_copy_fwd_metadata_to_bw_nodes` already does `copy.deepcopy(custom)` from forward to backward nodes, so the `rope_region` tag automatically propagates to backward RoPE ops.

### Models covered

- **Llama3/4** (`GQAttention` with `apply_rotary_emb_complex` via `attention.py`)
- **Qwen3, GPT-OSS** (`GQAttention` with `apply_rotary_emb_cos_sin` via `attention.py`)
- **DeepSeek V3** (MLA attention with `apply_rotary_emb_single_complex` via `deepseek_v3/model.py`)

### Numerics note

This could be changing numerics, which is fine per the board item.

## Test plan

- [x] 7 new unit tests in `test_passes.py::TestAnnotateRopeForRegionalInductor`
  - Verifies RoPE nodes tagged, non-RoPE nodes untouched
  - Verifies separate regions per layer, shared region within layer
  - Verifies inductor config propagation
  - Verifies no-op when no rope annotations present
  - Verifies None config produces empty annotations
- [x] All 52 existing non-FSDP tests in `test_passes.py` still pass
- [ ] GPU integration test with Llama3 debug model (FSDP+TP)
- [ ] Benchmark with `c4_test` dataset to measure RoPE kernel fusion speedup

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -18,6 +18,7 @@ from torchtitan.distributed import ParallelDims
 from torchtitan.tools.logging import logger
 
 _MODULE_FQN = "module_fqn"
+_ROPE_REGION = "rope_region"
 _NOT_IN_LAYERS = -1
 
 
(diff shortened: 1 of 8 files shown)


SherlockNoMad:
AutoDev: Benchmark results on 8×H100, Llama3 8B, FSDP4+TP2, `c4_test`, 20 steps (last step steady-state):

| Branch | tps | tflops | MFU | Memory |
|--------|-----|--------|-----|--------|
| `main` (baseline) | 6,891 | 399.11 | 40.36% | 40.51 GiB |
| `graph_trainer/rope-regional-inductor` | 6,918 | 400.67 | 40.51% | 41.09 GiB |

**Delta**: +0.15% MFU — essentially flat.

**Blocker**: Inductor emitted this warning during compilation:
```
Torchinductor does not support code generation for complex operators. Performance may be worse than eager.
```

Llama3 uses the **complex RoPE backend** (`apply_rotary_emb_complex` with `view_as_complex`/`view_as_real`). Inductor can't generate fused Triton kernels for complex number ops, so the regional compilation falls back to eager — negating the fusion benefit.

**Options**:
1. This pass would be more effective for models using the **cos/sin backend** (Qwen3, GPT-OSS) which doesn't use complex ops.
2. For the complex backend, we may need a different approach (e.g., a dedicated fused RoPE kernel, or converting the complex implementation to real arithmetic before Inductor compilation).

Command used:
```bash
    --compile.mode aot_fx_trace \
    --comm.trace_buf_size=0 \
    --training.steps 20
```

PR #3139, opened 2026-04-28
[torchtitan] Fix deepseek_v3_16b config to use deepep comm backend

anshul-si:
**Summary:** [D101435913](https://www.internalfb.com/diff/D101435913) introduced a token dispatcher hierarchy that requires moe_comm_backend (model construction) to match parallelism.expert_parallel_comm_backend (parallelization). The deepseek_v3_16b config used moe_comm_backend="standard" but the gb200 conveyor test passes expert_parallel_comm_backend="deepep",  causing an AssertionError in apply_moe_ep_tp. This has broken
sandcastle-test-titan-gb200-package since release 195 (April 18).

**Test Case**
Run MAST job with deepseek_v3_16b on GB200 and verify training starts without assertion failure. mast jobs fails at the same point as h100 equivalent due to a separate issue.

* __->__ #3139

### DIFF
--- a/torchtitan/models/deepseek_v3/config_registry.py
+++ b/torchtitan/models/deepseek_v3/config_registry.py
@@ -83,7 +83,7 @@ def deepseek_v3_16b() -> Trainer.Config:
         loss=ChunkedCELoss.Config(),
         hf_assets_path="./assets/hf/deepseek-moe-16b-base",
         model_spec=model_registry(
-            "16B", attn_backend="flex", moe_comm_backend="standard"
+            "16B", attn_backend="flex", moe_comm_backend="deepep"
         ),
         dataloader=HuggingFaceTextDataLoader.Config(
             dataset="c4",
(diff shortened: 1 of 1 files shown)


torchtitan/models/deepseek_v3/config_registry.py
```diff
@@ -83,7 +83,7 @@ def deepseek_v3_16b() -> Trainer.Config:
         loss=ChunkedCELoss.Config(),
         hf_assets_path="./assets/hf/deepseek-moe-16b-base",
         model_spec=model_registry(
-            "16B", attn_backend="flex", moe_comm_backend="standard"
+            "16B", attn_backend="flex", moe_comm_backend="deepep"
```
tianyu-l, line 120: Instead of changing default, I think it may be good to have two job config, one with deepep backend, the other with standard. The conveyor one can depend on the deepep one.

anshul-si: sure sounds good, was worried if adding more configs would clutter registry

tianyu-l (approved):

torchtitan/models/deepseek_v3/config_registry.py
acisseJZhong: you can do sth like https://github.com/pytorch/torchtitan/blob/7da65eceebab762279ee042e95a556ffae35dcf2/torchtitan/models/gpt_oss/config_registry.py#L63, just swap out model spec

PR #3141, opened 2026-04-28
[GraphTrainer] Add full inductor compilation pass

SherlockNoMad:
## Summary
- Add `compile.inductor_compilation` config (`"regional"` | `"full"`) to select between regional inductor (existing default) and full inductor compilation.
- Full inductor applies inductor decompositions (e.g. `aten.t` → `aten.permute`) via `make_fx` retrace, then compiles the entire graph into Triton kernels via `compile_fx_inner`, adapting the boxed calling convention back to positional args.
- Remove dead `inductor_decomposition_pass` code from the deprecated AOT path in `graph_utils.py`.

## Benchmark

Llama 8B, 8×H100, FSDP(4)+TP(2), `c4_test`, 20 steps:
```bash
    --compile.mode aot_fx_trace \
    --compile.inductor_compilation full \  # or regional
    --comm.trace_buf_size=0 \
    --training.steps 20
```

| Metric | full_inductor | regional_inductor | Delta |
|---|---|---|---|
| **MFU** | **44.05%** | 40.26% | **+3.79%** |
| **tflops** | **435.61** | 398.19 | **+9.4%** |
| **tps** | **7,522** | 6,876 | **+9.4%** |
| **memory** | **38.40 GiB** | 40.51 GiB | **-2.11 GiB** |

## Test plan
- [x] `run_graph_trainer_llama3_8b.sh` (debug model, 10 steps) passes
- [x] Integration test suite `graph_trainer_default` passes (includes new `aot_fx_trace_llama3_fsdp_tp_full_inductor` test)
- [x] Llama 8B benchmark above completes successfully


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -1058,7 +1008,6 @@ def tlparse_log_graph_pass(
 
 # Registry for joint custom passes (applied before partitioning, AOT mode only)
 AVAILABLE_JOINT_PASSES = {
-    "inductor_decomposition": inductor_decomposition_pass,
```
yiming0416, line 1003: `inductor_decomposition` is still being used in the deprecated AOT path. 

If we remove it here, I suggest also removing the entire full inductor option from the deprecated AOT path.

yiming0416 (commented):
Also FYI, llama3_8b doesn't use flex attention so regional-inductor is a no-op

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: yes, I removed it in torchtitan/experiments/graph_trainer/graph_utils.py

SherlockNoMad:
> Also FYI, llama3_8b doesn't use flex attention so regional-inductor is a no-op
yea, so in such the case. the comparison is "no compile" vs. "full compile"

torchtitan/experiments/graph_trainer/passes.py
yiming0416: In the deprecated AOT path, `inductor_decomposition` will be added automatically when `full_inductor_compilation` is on. So we need to clean up scenarios where we have `--compile.mode aot` and `--compile.passes full_inductor_compilation`

For example, in `integration_test.py` and `run_precompile_tests.py`

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -143,8 +155,10 @@ def construct_default_graph_passes(
     """
     from torchtitan.experiments.graph_trainer.cudagraph import is_cudagraph_compatible
 
-    use_cudagraph = config.compile.enable_cudagraph and is_cudagraph_compatible(
-        traced_result.gm
+    use_cudagraph = (
```
aditvenk: In full inductor path, can we defer to Inductor's "reduce_overhead" mode which does CG on inductor compiled code?

SherlockNoMad: yea, we can, but it won't be compatible with precompile. 

precompile need to reapply cudagraph at artifact loading time, so ideally we have an cudagraph solution that is decoupled from compiler.

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: ic.. let me clean up

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: I retested, full_inductor is actually compatible with cudagraph. 

so this is not an issue.

torchtitan/experiments/graph_trainer/passes.py
SherlockNoMad: done

aditvenk (approved):

torchtitan/experiments/graph_trainer/tests/run_precompile_tests.py
```diff
@@ -56,7 +56,6 @@ def _build_precompile_tests() -> list[PrecompileTestDefinition]:
                 " --config graph_trainer_llama3_debugmodel"
                 " --compile.mode aot"
                 " --compile.passes full_inductor_compilation"
-                " --compile.joint_passes inductor_decomposition"
```
yiming0416, line 59: I think you should just remove these two test cases. For `aot`, having `--compile.passes full_inductor_compilation` will automatically have `--compile.joint_passes inductor_decomposition`

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -279,7 +279,6 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
                     "--compile.mode aot",
                     "--parallelism.data_parallel_shard_degree 4",
                     "--parallelism.tensor_parallel_degree 2",
-                    "--compile.joint_passes inductor_decomposition",
```
yiming0416, line 282: You can just remove this test case. `auto_bucketing` only is already covered

yiming0416 (approved):

torchtitan/experiments/graph_trainer/tests/run_precompile_tests.py
SherlockNoMad: `full_inductor_compilation` is internally calling `inductor_decomposition`, so this is not an issue.

PR #3142, opened 2026-04-28
[rl] Enable TP2EP for unified MoE model in vLLM wrapper

wwwjn:
* #3236
* __->__ #3142

Enable Expert Parallelism for MoE models in the vLLM inference path, with TP on dense layers. Includes meta-device init, init_states() fix for RoPE cache, EP mesh creation, enable_sp-based output layout, and MoE debug configs.

Key changes:
- Meta-device init + to_empty() + init_states() for large MoE models
- EP mesh: ep = tp_size (TP ranks become EP ranks for experts)
- Use enable_sp (not inference flag) for output layout and SP splitting
- enable_sequence_parallel=True, disable_loss_parallel=True for inference
- Remove stale ModelConvertersContainer references
- Add MoE debug and 30B-A3B RL configs

Similar as #3057 

### Dependency
- [x] https://github.com/pytorch/pytorch/pull/172792/changes
- [x] https://github.com/pytorch/torchtitan/pull/3604 

## Verification

Qwen3-30B-A3B model on 8 GPU (Trainer TP4 EP4, generator TP4 EP4): 

```
wandb: Detected [openai] in use.
wandb: Use W&B Weave for improved LLM call tracing. Install Weave with `pip install weave` then add `import weave` to the top of your script.
wandb: For more information, check out the docs at: https://weave-docs.wandb.ai
WandB logging enabled
Using renderer qwen3, of type <class 'renderers.configs.Qwen3RendererConfig'>, with args {'enable_thinking': False}
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
4 generator GPUs + 4 trainer GPUs = 8 total
[actor=<root>] Initializing TorchStoreStrategy with default_transport_type=<TransportType.Unset: 1>
[actor=<root>] Pre-training validation; then 10 steps of RL training
[actor=<root>] ----------
Validation | Step:  0  validation/response_length/mean: 30.10  timing/validate: 29.13
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  1  loss/mean: 0.013  bit_wise/logprob_diff/max: 55.49  reward/_mean: 0.79  rollout_reward/_mean: 0.79  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.68  rollout/response_length/max: 42.0  train/grad_norm/mean: 0.46  train/lr: 5.0e-07  perf/tokens_per_second: 141.1  timing/step: 120.0
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  2  loss/mean: 0.0055  bit_wise/logprob_diff/max: 56.85  reward/_mean: 0.70  rollout_reward/_mean: 0.70  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.80  rollout/response_length/max: 44.0  train/grad_norm/mean: 0.37  train/lr: 1.0e-06  perf/tokens_per_second: 158.5  timing/step: 110.2
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  3  loss/mean: 0.016  bit_wise/logprob_diff/max: 40.089  reward/_mean: 0.73  rollout_reward/_mean: 0.73  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.76  rollout/response_length/max: 47.0  train/grad_norm/mean: 0.45  train/lr: 1.0e-06  perf/tokens_per_second: 156.9  timing/step: 110.0
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  4  loss/mean: 0.011  bit_wise/logprob_diff/max: 40.92  reward/_mean: 0.73  rollout_reward/_mean: 0.73  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.72  rollout/response_length/max: 45.0  train/grad_norm/mean: 0.43  train/lr: 1.0e-06  perf/tokens_per_second: 163.3  timing/step: 108.7
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  5  loss/mean: 0.0076  bit_wise/logprob_diff/max: 96.53  reward/_mean: 0.71  rollout_reward/_mean: 0.71  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.76  rollout/response_length/max: 42.0  train/grad_norm/mean: 0.36  train/lr: 1.0e-06  perf/tokens_per_second: 165.3  timing/step: 109.2
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  6  loss/mean: 0.011  bit_wise/logprob_diff/max: 48.36  reward/_mean: 0.79  rollout_reward/_mean: 0.79  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.84  rollout/response_length/max: 49.0  train/grad_norm/mean: 0.44  train/lr: 1.0e-06  perf/tokens_per_second: 147.1  timing/step: 117.7
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  7  loss/mean: 0.0055  bit_wise/logprob_diff/max: 54.41  reward/_mean: 0.65  rollout_reward/_mean: 0.65  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.72  rollout/response_length/max: 43.0  train/grad_norm/mean: 0.12  train/lr: 1.0e-06  perf/tokens_per_second: 158.1  timing/step: 110.1
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  8  loss/mean: 0.0035  bit_wise/logprob_diff/max: 51.040  reward/_mean: 0.66  rollout_reward/_mean: 0.66  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.72  rollout/response_length/max: 45.0  train/grad_norm/mean: 0.31  train/lr: 1.0e-06  perf/tokens_per_second: 168.6  timing/step: 101.4
[actor=<root>] Dropping 1 packed rows (9 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step:  9  loss/mean: 0.023  bit_wise/logprob_diff/max: 58.88  reward/_mean: 0.56  rollout_reward/_mean: 0.56  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.64  rollout/response_length/max: 46.0  train/grad_norm/mean: 0.73  train/lr: 1.0e-06  perf/tokens_per_second: 175.6  timing/step: 103.8
[actor=<root>] Dropping 2 packed rows (10 -> 8) to fit global_batch_size
[actor=<root>] ----------
Train | Step: 10  loss/mean: 0.020  bit_wise/logprob_diff/max: 50.84  reward/_mean: 0.52  rollout_reward/_mean: 0.52  reward/_max: 1.0  rollout_reward/_max: 1.0  reward/zero_std_frac: 0.64  rollout/response_length/max: 47.0  train/grad_norm/mean: 0.23  train/lr: 1.0e-06  perf/tokens_per_second: 167.2  timing/step: 109.2
[actor=<root>] ----------
Validation | Step: 10  validation/response_length/mean: 31.85  timing/validate: 25.94
[actor=<root>] ============================================================
[actor=<root>] Validation summary (pre / post):
[actor=<root>]   validation_reward/_max:  +1.000  /  +1.000
[actor=<root>]   validation_reward/_mean:  +0.639  /  +0.664
[actor=<root>]   validation_reward/_min:  +0.052  /  +0.056
[actor=<root>]   validation_reward/_std:  +0.406  /  +0.386
[actor=<root>]   validation_reward/_sum:  +12.776  /  +13.289
[actor=<root>]   validation_reward/component/RewardAlphabetSort/mean:  +0.639  /  +0.664
[actor=<root>] ============================================================
[actor=<root>] Closing: tearing down actors and process meshes.

```
NOTE: this locally run didn't patch #3604, so the trainer are actually producing wrong numerics

### Batch-invariant verification
- To achieve batch invariant in MoE model, we need extra `bmm`  invariant kernel: dense doesn’t meaningfully exercise a 3D batched matmul path in this parity setup, while MoE/EP introduces batched/grouped per-expert shapes and routing-dependent layouts; mm/addmm fixes 2D linear layers, but not 3D aten::bmm/batched matmul reductions.

```
Validation | Step:  0  validation/response_length/mean: 46.70  timing/validate: 7.059  
[actor=<root>] ----------
Train | Step:  1  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00040  perf/tokens_per_second: 1253.0  timing/step: 14.12  
[actor=<root>] ----------
Train | Step:  2  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1542.7  timing/step: 12.10  
[actor=<root>] ----------
Train | Step:  3  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1667.2  timing/step: 10.78  
[actor=<root>] ----------
Train | Step:  4  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1641.3  timing/step: 11.034  
[actor=<root>] ----------
Train | Step:  5  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1641.3  timing/step: 11.0032  
[actor=<root>] ----------
Train | Step:  6  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1586.3  timing/step: 11.52  
[actor=<root>] ----------
Train | Step:  7  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1768.6  timing/step: 10.43  
[actor=<root>] ----------
Train | Step:  8  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1685.5  timing/step: 10.56  
[actor=<root>] ----------
Train | Step:  9  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1822.5  timing/step: 10.44  
[actor=<root>] ----------
Train | Step: 10  loss/mean: 0  bit_wise/logprob_diff/max: 9.5e-07  reward/_mean: 0  rollout_reward/_mean: 0  reward/_max: 0  rollout_reward/_max: 0  reward/zero_std_frac: 1.0  rollout/response_length/max: 50.0  train/grad_norm/mean: 0  train/lr: 0.00080  perf/tokens_per_second: 1341.5  timing/step: 13.36  
```

NOTE:  the loss is zero is because it's debug MoE model, not meaningful weights loaded so  reward is constantly 0


torchtitan/experiments/rl/grpo.py
```diff
@@ -416,11 +416,25 @@ async def setup(
             generator_dtype=config.generator.model_dtype,
         )
 
+        # When EP is enabled on the generator, rebuild the model spec with
+        # AllToAllTokenDispatcher (moe_comm_backend="standard") since the
+        # shared model_spec uses LocalTokenDispatcher (for EP=1 trainer).
```
wwwjn: Might need to better config this field

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -431,6 +462,16 @@ def _initial_load_weights(self, checkpoint_path):
 
         # Load HF state dict using DCP
         hf_state_dict = adapter.to_hf(self.model.state_dict())
+
+        # Filter out keys not in the HF checkpoint (e.g. expert_bias is a
```
wwwjn: This need to be double check

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -395,6 +422,10 @@ def compute_logits(
         if isinstance(logits, DTensor):
             logits = logits.to_local()
 
+        # Ensure logits are plain tensors for vLLM's sampling code
+        if isinstance(logits, DTensor):
+            logits = logits.full_tensor()
```
tianyu-l: don't need this? we are doing to_local one line above

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: I think this should be case by case. For models that don't have expert_bias in checkpoint, they are probably trained with aux loss for load balancing -- we should use the same (being added in https://github.com/pytorch/torchtitan/pull/3000) and not introduce bias in persistent buffer.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -291,8 +291,16 @@ def dispatch(
             input_splits_list = input_splits.tolist()
             output_splits_list = output_splits.tolist()
 
-        # All-to-all dispatch tokens to EP ranks
-        routed_input = all_to_all_single_autograd(
+        # All-to-all dispatch tokens to EP ranks.
+        # Use the non-autograd version when running in inference mode
+        # (e.g. vLLM), since _c10d_functional_autograd ops don't dispatch
+        # correctly without an active autograd context.
```
tianyu-l: this sounds strange, let's ask comms expert if this is expected.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -101,39 +101,55 @@ def create_torchtitan_config_from_vllm_config(
         Tuple of (ParallelDims, ParallelismConfig) mapped from vLLM config
 
     Note:
-        vLLM doesn't use FSDP sharding (dp_shard=1) or expert parallelism (ep=1, etp=1)
-        in inference. These are set to default values.
+        vLLM doesn't use FSDP sharding (dp_shard=1) in inference.
+        Expert parallelism is enabled when vLLM's enable_expert_parallel
+        flag is set, repurposing TP ranks for EP (ep=tp_size, etp=1).
```
acisseJZhong: nit: we should be deprecating ETP in titan soon. So ETP will always be 1.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -101,39 +101,55 @@ def create_torchtitan_config_from_vllm_config(
...
+    else:
+        ep_size = 1
+        etp_size = 1
+
     parallel_dims = ParallelDims(
         dp_replicate=parallel_config.data_parallel_size,
         dp_shard=1,
-        cp=parallel_config.decode_context_parallel_size,
-        tp=parallel_config.tensor_parallel_size,
+        cp=1,
```
acisseJZhong: why hardcoding?

torchtitan/models/llama4/parallelize.py
```diff
@@ -408,15 +407,19 @@ def apply_moe_ep_tp(
...
-                # all-reduce (Partial→Replicate) for output.
                 "moe": PrepareModuleInputOutput(
                     input_layouts=(sp_layout,),
                     desired_input_layouts=(Replicate(),),
                     use_local_input=False,
-                    output_layouts=(Partial(),),
+                    # TP-sharded experts or SP token splitting produce
+                    # Partial outputs that need reduce-scatter. EP without
+                    # SP produces Replicate (all-to-all combine already
+                    # returns the full result on every rank).
```
tianyu-l: I don't think this is right. With / without SP for dense, we should always run SP inside MoE on tp mesh. cc @acisseJZhong

torchtitan/models/qwen3/state_dict_adapter.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, model_config: Qwen3Model.Config, hf_assets_path: str | None):
             "model.layers.{}.mlp.experts.{}.up_proj.weight": "layers.{}.moe.experts.w3",
             "model.layers.{}.mlp.experts.{}.down_proj.weight": "layers.{}.moe.experts.w2",
             "model.layers.{}.mlp.gate.weight": "layers.{}.moe.router.gate.weight",
+            "model.layers.{}.mlp.expert_bias": "layers.{}.moe.expert_bias",
```
tianyu-l: why this change?

torchtitan/experiments/rl/grpo.py
tianyu-l: No we should use `moe_comm_backend="standard"` for both EP and non-EP case after https://github.com/pytorch/torchtitan/pull/3125

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -395,6 +422,10 @@ def compute_logits(
         if isinstance(logits, DTensor):
             logits = logits.to_local()
 
+        # Ensure logits are plain tensors for vLLM's sampling code
+        if isinstance(logits, DTensor):
```
acisseJZhong, line 326: what's the diff between this and L422?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -221,6 +221,133 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
...
+            compile=CompileConfig(enable=True, backend="aot_eager"),
+            loss=GRPOLoss.Config(),
+        ),
+        generator=VLLMGenerator.Config(
+            compile=GeneratorCompileConfig(
+                backend="eager",
+                cudagraph_mode="piecewise",
+            ),
+            parallelism=ParallelismConfig(
+                tensor_parallel_degree=2,
```
acisseJZhong: wondering how do we solve the imbalance seq split problem if turning on TP?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -191,11 +186,13 @@ def __init__(
         self.model_path = model_path
 
         # Build vLLM engine
+        enable_ep = config.parallelism.expert_parallel_degree > 1
         engine_kwargs = dict(
             model=model_path,
             trust_remote_code=True,
             dtype=config.model_dtype,
             tensor_parallel_size=config.parallelism.tensor_parallel_degree,
+            enable_expert_parallel=enable_ep,
```
tianyu-l, line 475: it sounds OK if this is boolean in vllm, but we have to have safety guards on the torchtitan config https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L128

torchtitan/experiments/rl/grpo.py
acisseJZhong: why don't we just build the AllToAllTokenDispatcher config for generator from the start?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -221,6 +221,133 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
...
+            sampling=SamplingConfig(
+                temperature=1.0,
+                top_p=0.95,
+                max_tokens=50,
+            ),
+        ),
+    )
+
+
+def rl_grpo_qwen3_moe_debug_ep() -> RLTrainer.Config:
```
tianyu-l: is batch invariance achievable?

torchtitan/models/llama4/parallelize.py
```diff
@@ -394,7 +394,6 @@ def apply_moe_ep_tp(
         ep_etp_mesh: Combined EP/ETP mesh.
         comm_backend: MoE communication backend.
         enable_sp: Whether sequence parallelism is enabled.
-        pad_multiple: Optional token padding multiple for compatible expert kernels.
```
acisseJZhong: should be gone if you rebase

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: I think it's fine to deprecate ETP field.

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This is not enabled now, but previous this is just for testing purpose. And we are leak of knowledge of `decode_context_parallel_size` in vllm, and I think hard codinf it to be 1 is safer before we officially support CP in inference

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This is a rebase issue after Chien-Chin's full DTensor change, let me update

torchtitan/models/llama4/parallelize.py
acisseJZhong: should be handled by https://github.com/pytorch/torchtitan/pull/3167/changes#diff-629cab747b607577819f02d9305b3f2c3e5a99121a3b7376346a3f0f92a3b4aaR444

torchtitan/experiments/rl/models/vllm_wrapper.py
acisseJZhong: deprecating in https://github.com/pytorch/torchtitan/pull/3167, @wwwjn you might need to rebase.

wwwjn: Sure, thanks! Let me rebase

torchtitan/models/llama4/parallelize.py
wwwjn: So SP should always be enabled within MoE, and `enable_sp` controls the dense part enable SP or not. Let me change this part

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This was an issue because of rebase. but I changed the to_local() to fulld_tenser() for safer conversion, especially someone wants to try loss parallel

acisseJZhong: what's different between to_local and full_tensor? both convert from dtensor to plain tensor?

torchtitan/models/llama4/parallelize.py
acisseJZhong: yes! enable SP only controls dense part. MoE will always have SP on as long as it has TP mesh in dense part.

torchtitan/experiments/rl/config_registry.py
wwwjn: Testing in later PR. Since we need 8 GPUs for both trainer and generator, I was using six_cluster to do so

wwwjn: Yes TP2EP batch-invariant is achievable. I added CI to this commit. And I tested groupGEMM locally, it is batch-invariant

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: Are you suggesting we should infer which load balancing method they are using by checking if the checkpoint contains an `expert_bias` field?
- If no `expert_bias` field, they highly likely used aux-loss based load balancing?

I'm hesitate because this would be vague for users: If they specify a LB method, but it wash 

We could explicitly error out if user specify expert-bias based LB , but the checkpoint doesn't have `expert_bias` field. In our RL config_registry, I will switch to loss based LB.

tests/assets/qwen3_moe_debug/tokenizer_config.json
wwwjn, line 1: This file is a soft link

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -258,8 +295,13 @@ def __init__(
                 self.model.freqs_cis, max_model_len
             )
 
-        # Initial load model weights from HuggingFace checkpoint path
-        self._initial_load_weights(checkpoint_path=vllm_config.model_config.model)
+        # Initial load model weights from HuggingFace checkpoint path.
+        import os as _os
+
+        if _os.environ.get("TORCHTITAN_SKIP_INITIAL_HF_LOAD") != "1":
```
wwwjn: I used a env variable to random init in vllm_wrapper because it's impossible to change the vllm_wrapper init signature, which is hard to pass as a kwarg or config

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -427,6 +470,29 @@ def _initial_load_weights(self, checkpoint_path):
...
+        # mismatches are not silently masked.
+        # TODO: once aux-loss-based load balancing lands (#3000), switch the
+        # affected models to ``load_balance_coeff=None`` + aux loss. Then
+        # ``expert_bias`` will not exist in the model state dict and this
+        # special case can be removed.
+        hf_keys_in_checkpoint = set(
+            storage_reader.read_metadata().state_dict_metadata.keys()
+        )
+        missing = set(hf_state_dict.keys()) - hf_keys_in_checkpoint
+        unexpected_missing = {k for k in missing if not k.endswith(".expert_bias")}
```
wwwjn: Need to remove expert_bias from the check after the loss based load balancing: https://github.com/pytorch/torchtitan/pull/3000. If expert_bias is in our model, we should load from checkpoint

tests/assets/qwen3_moe_debug/config.json
wwwjn, line 1: This file is a fake config.json (HF format) which is required by vllm. Will check vllm in detail why this file is needed and can we relief that

tests/assets/qwen3_moe_debug/generation_config.json
wwwjn, line 1: Same, needed by vLLM

torchtitan/models/common/token_dispatcher.py
```diff
@@ -308,8 +314,15 @@ def dispatch(
...
             output_splits_list = output_splits.tolist()
 
-        # All-to-all dispatch tokens to EP ranks
-        routed_input = all_to_all_single_autograd(
+        # All-to-all dispatch tokens to EP ranks.
+        # Use the non-autograd version under inference (vLLM), since
+        # _c10d_functional_autograd ops don't dispatch correctly without
+        # an active autograd context. Gated by a Python bool so the choice
+        # is stable at trace time.
+        a2a_fn = (
```
wwwjn: @tianyu-l Who should I ask this comm issue?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -155,10 +155,18 @@ def __post_init__(self):
                     f"Generator does not support context parallelism, "
                     f"got cp={p.context_parallel_degree}"
                 )
-            if p.expert_parallel_degree > 1:
+            # vLLM ties EP to TP (enable_expert_parallel=True reuses the TP
+            # group as the EP group), so the only valid values are EP=1
+            # (disabled) or EP=TP.
+            if (
+                p.expert_parallel_degree != 1
+                and p.expert_parallel_degree != p.tensor_parallel_degree
```
tianyu-l: why this is valid? we are removing this on trainer side, but even if we don't, it's not legit in vllm?

tests/assets/qwen3_moe_debug/config.json
tianyu-l, line 1: why this file?

tests/assets/qwen3_moe_debug/generation_config.json
tianyu-l, line 1: why this file?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +201,166 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
...
+            sampling=SamplingConfig(
+                temperature=1.0,
+                top_p=0.95,
+                max_tokens=50,
+            ),
+        ),
+    )
+
+
+def rl_grpo_qwen3_moe_debug_ep_batch_invariant() -> RLTrainer.Config:
```
tianyu-l: is batch invariance achieved?

torchtitan/experiments/rl/trainer.py
```diff
@@ -165,7 +167,7 @@ class RLTrainer(Configurable):
     class Config(Configurable.Config):
         """Top-level config for RL training."""
 
-        model_spec: ModelSpec | None = None
+        model_spec: Annotated[ModelSpec | None, tyro.conf.Suppress] = None
```
tianyu-l: why we didn't need this before?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -188,7 +188,10 @@ def _load_initial_hf_weights(self, model, checkpoint_path: str) -> None:
         set_model_state_dict(
             model=model,
             model_state_dict=torchtitan_state_dict,
-            options=StateDictOptions(strict=True),
+            # strict=False: HF MoE checkpoints don't carry expert_bias buffers
```
tianyu-l: then expert_bias should be non-persistent buffer?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/moe.py#L349

I believe qwen is not using loss-free-load-balancing, so we should enable load balance loss for it.
This is blocked by @pianpwk 's https://github.com/pytorch/torchtitan/pull/3000

torchtitan/config/configs.py
```diff
@@ -363,6 +363,14 @@ class DebugConfig:
     """Enable batch-invariant mode to use batch-invariant ops in model
     forward and deterministic NCCL collective reduction order"""
 
+    random_init: bool = False
```
tianyu-l: `Debug` is shared by pre-training trainer, rl trainer, and rl generator, so the name should be made much narrower, e.g. `skip_rl_trainer_initial_hf_weight_load`.

Later we should find a better home for this config, e.g. under `WeightSyncConfig` in policy trainer.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -189,18 +197,26 @@ def __init__(
         # Set vLLM environment variables from config before any vLLM initialization
         os.environ["VLLM_ATTENTION_BACKEND"] = "CUSTOM"
 
+        # Signal the wrapper to skip the HF weight load when running with
+        # random init. The trainer pushes its random-init weights to TorchStore
+        # at startup, so the generator picks up matching weights from there.
+        if config.debug.random_init:
+            os.environ["TORCHTITAN_SKIP_INITIAL_HF_LOAD"] = "1"
```
tianyu-l: I didn't get why we need this

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: how is `vllm_config.model_config.max_model_len` determined? It should never surpass torchtitan model len. In fact, the capability should be intrinsic to model and shouldn't be "extended" in RL.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -272,8 +306,13 @@ def __init__(
                 self.model.freqs_cis, max_model_len
             )
 
-        # Initial load model weights from HuggingFace checkpoint path
-        self._initial_load_weights(checkpoint_path=vllm_config.model_config.model)
+        # Initial load model weights from HuggingFace checkpoint path.
+        import os as _os
+
+        if _os.environ.get("TORCHTITAN_SKIP_INITIAL_HF_LOAD") != "1":
+            self._initial_load_weights(checkpoint_path=vllm_config.model_config.model)
```
tianyu-l: I'm confused here. Why do we need to initial load weights at all?
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/grpo.py#L453
this will init the weights on generator. The flow should be
- trainer build model, load weights from HF (which we now can skip with a flag for debug purpose)
- trainer push model state dict to TS
- generator pull model state dict from TS and load into model

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -104,37 +104,48 @@ def create_torchtitan_config_from_vllm_config(
...
+    ep_size = tp_size if parallel_config.enable_expert_parallel else 1
+
     parallel_dims = ParallelDims(
         dp_replicate=parallel_config.data_parallel_size,
         dp_shard=1,
-        cp=parallel_config.decode_context_parallel_size,
-        tp=parallel_config.tensor_parallel_size,
+        cp=1,
+        tp=tp_size,
         pp=parallel_config.pipeline_parallel_size,
```
tianyu-l: then you should hardcode this to 1 as well?

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: we should assert vllm CP / PP degrees to be 1

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -104,37 +104,48 @@ def create_torchtitan_config_from_vllm_config(
...
         pipeline_parallel_degree=parallel_config.pipeline_parallel_size,
-        expert_parallel_degree=1,
+        expert_parallel_degree=ep_size,
         disable_loss_parallel=True,  # vLLM handles sampling and expects plain tensor logits.
         enable_sequence_parallel=False,
     )
 
+    # Build the full device mesh so all dimensions (tp, ep, efsdp, etc.)
+    # are available to the core parallelize function.
+    parallel_dims.build_mesh()
```
tianyu-l: why do we not need this before?
I think it builds itself implicitly https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/parallel_dims.py#L244

torchtitan/models/common/token_dispatcher.py
```diff
@@ -308,8 +314,15 @@ def dispatch(
             input_splits_list = input_splits.tolist()
             output_splits_list = output_splits.tolist()
 
-        # All-to-all dispatch tokens to EP ranks
-        routed_input = all_to_all_single_autograd(
+        # All-to-all dispatch tokens to EP ranks.
+        # Use the non-autograd version under inference (vLLM), since
+        # _c10d_functional_autograd ops don't dispatch correctly without
+        # an active autograd context. Gated by a Python bool so the choice
+        # is stable at trace time.
```
tianyu-l: I'm quite uncomfortable with this. An autograd function should be able to run fwd-only under inference mode, unless there is strong reason why it can't.
cc @ezyang @fegin

tests/assets/qwen3_moe_debug/config.json
tianyu-l: oh, I'm guessing you need it because it's required here? https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L200

still very blackbox to me -- could you explain how vllm uses this file? It feels not fundamental to me.

torchtitan/experiments/rl/config_registry.py
wwwjn: Yes, Full TP degree to EP can achieve  batch invariant .

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: Yes you are right. I guess it's more from a historical reason: we started from inference only, so we will need a way to load weights . Now the initial_load_weights are unconditionally called when generator are initialized. I agree we should skip either 1) initial_load_weights (but passing an extra flag to vllm_wrapper need env variable), , or 2) skip the first weight sync 

I'd prefer go with 1)

tests/assets/qwen3_moe_debug/config.json
wwwjn: Yes, the generator still needs to point to a HF format checkpoint folder. I added some github comments but forgot to submit my comment, I was trying to figure out VLLM init path and understand why we need this

If we skip the initial load weight, we might be able to get rid of this config as well.

tests/assets/qwen3_moe_debug/generation_config.json
wwwjn: I think we can remove this file and config.json, and totally remove the dependency of HuggingFace. Previously the config.json file is needed because by default it hit https://github.com/vllm-project/vllm/blob/6467213a9f118fcf5ebf65d5e1ee8d64d2e101e4/vllm/transformers_utils/config.py#L172C7-L184 HFConfigParser . 

I will refactor to register a customized ConfigParser to vllm.

torchtitan/models/common/token_dispatcher.py
tianyu-l: ask in group chat and we can start tag people

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: what's 1) -- is it generator side or trainer side?

I think we can just remove any initial load weight from generator (and only rely on the first weight sync to get proper weight)

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -154,10 +154,18 @@ def __post_init__(self):
                     f"Generator does not support context parallelism, "
                     f"got cp={p.context_parallel_degree}"
                 )
-            if p.expert_parallel_degree > 1:
+            # vLLM ties EP to TP (enable_expert_parallel=True reuses the TP
```
acisseJZhong: so vllm doesn't support EP>TP? by also reusing DP group if we turn on FSDP.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -202,6 +217,7 @@ def __init__(
             config_format=TORCHTITAN_CONFIG_FORMAT,
             dtype=config.model_dtype,
             tensor_parallel_size=config.parallelism.tensor_parallel_degree,
+            enable_expert_parallel=enable_ep,
```
acisseJZhong, line 475: make this inline?

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -104,37 +104,48 @@ def create_torchtitan_config_from_vllm_config(
...
+    # (each rank holds a shard of experts): ep_size = tp_size.
+    tp_size = parallel_config.tensor_parallel_size
+    ep_size = tp_size if parallel_config.enable_expert_parallel else 1
+
     parallel_dims = ParallelDims(
         dp_replicate=parallel_config.data_parallel_size,
         dp_shard=1,
-        cp=parallel_config.decode_context_parallel_size,
-        tp=parallel_config.tensor_parallel_size,
+        cp=1,
```
acisseJZhong: maybe leave a comment on why hardcoding cp to 1? also for fsdp, wondering if we could turn it on and later have both tp2ep and dp2ep

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -269,6 +285,24 @@ def __init__(
...
+        # vLLM runs under torch.inference_mode(); the autograd functional
+        # collectives don't dispatch correctly there. Flip MoE token
+        # dispatchers to the non-autograd a2a path. Read at trace time.
+        from torchtitan.models.common.token_dispatcher import AllToAllTokenDispatcher
+
+        for layer in self.model.layers.values():
+            moe = getattr(layer, "moe", None)
+            if moe is None:
+                continue
+            dispatcher = getattr(moe.experts, "token_dispatcher", None)
```
acisseJZhong: can you raise ValueError when it's None?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +205,161 @@ def rl_grpo_qwen3_debug() -> RLTrainer.Config:
...
+    Generator uses TP=4 for dense layers and EP=4 for MoE experts.
+    Trainer uses TP=4 for all layers.
+
+    Note: Qwen3-30B-A3B has 4 KV heads, so TP degree cannot exceed 4.
+    """
+    return RLTrainer.Config(
+        model_spec=model_registry("30B-A3B", attn_backend="varlen"),
+        hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-30B-A3B",
+        num_steps=10,
+        compile=CompileConfig(enable=True, backend="aot_eager"),
```
acisseJZhong: why this doesn't need to be False?

torchtitan/experiments/rl/trainer.py
```diff
@@ -168,7 +170,7 @@ class RLTrainer(Configurable):
...
```
acisseJZhong: why this?

torchtitan/models/common/token_dispatcher.py
```diff
@@ -308,8 +314,15 @@ def dispatch(
             input_splits_list = input_splits.tolist()
             output_splits_list = output_splits.tolist()
 
-        # All-to-all dispatch tokens to EP ranks
-        routed_input = all_to_all_single_autograd(
+        # All-to-all dispatch tokens to EP ranks.
+        # Use the non-autograd version under inference (vLLM), since
+        # _c10d_functional_autograd ops don't dispatch correctly without
```
acisseJZhong: just curious why can't torch.inference_mode automatically switch between these two if this is the only difference.

torchtitan/experiments/rl/actors/generator.py
wwwjn: When EP > TP , we will borrow DP degrees, which is achieved in next PR. And it's pure DP , not FSDP

torchtitan/models/common/token_dispatcher.py
wwwjn: Image
I dig into more details with claude: Under inference mode, it (`all_to_all_single_autograd`) never got dispatched to the only autograd kernel

My questions is why we use `all_to_all_single_autograd` for training? Can we use `all_to_all_single` for training?

xmfan: sry why do we think all_to_all_autograd didn't dispatch properly? autograd.Functions have some early exit conditions in inference mode, but I don't think that prevents you from using it in inference

UPDATE: we need to revive https://github.com/pytorch/pytorch/pull/149411

torchtitan/config/configs.py
wwwjn: Sg, this makes a lot of sense.

To be clear for weight initialization in RL:
- For Trainer: This flag is only skip trainer's initial_hf_weights_load, which is used for CI to random init weights, without loading from an actual HF checkpoint folder. 

- For generator: 
   - Previously we have duplicate generator initial_load_weights, even the weights will be pushed from trainer to generator at the beginning of RL. We should remove the redundant generator's `initial_load_weight` in RL.
   - For eval / inference, we still need VLLMWrapper to load weights from somewhere. This can be done by adding explicit field in VLLMWrapper.Config(): `inital_hf_weights_path` default to be None. We only load weights if it's not None.

torchtitan/experiments/rl/actors/generator.py
wwwjn: This is a bug after rebase. Removed

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: Replied here: https://github.com/pytorch/torchtitan/pull/3142#discussion_r3203974062

> I think we can just remove any initial load weight from generator (and only rely on the first weight sync to get proper weight)

In RL yes.

torchtitan/models/common/token_dispatcher.py
wwwjn: For my current PR, I will leave the current branching as is, and create a issue in torchtitan to track the progress - The ultimate goal for us is to consolidate to a single ops for pretrain and RL.

https://github.com/pytorch/torchtitan/issues/3268

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -408,9 +462,10 @@ def compute_logits(
 
         logits = self.model.lm_head(hidden_states)
 
-        # Full DTensor path returns logits as DTensor; vLLM expects plain tensors.
+        # vLLM's sampling expects full plain tensors. full_tensor() handles both
+        # Replicate (no-op) and Shard(-1) (all-gather) lm_head output placements.
         if isinstance(logits, DTensor):
-            logits = logits.to_local()
+            logits = logits.full_tensor()
```
tianyu-l: should work with disable_loss_parallel already?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +205,164 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
     )
 
 
+def rl_grpo_qwen3_moe_debug_ep() -> RLTrainer.Config:
```
tianyu-l: maybe put varlen in the name, as flex is being added

torchtitan/models/common/token_dispatcher.py
tianyu-l: any progress? could you follow up?
add a TODO and issue if you think it's not possible to get this done quickly

torchtitan/config/configs.py
```diff
@@ -383,6 +383,18 @@ class DebugConfig:
     """Enable batch-invariant mode to use batch-invariant ops in model
     forward and deterministic NCCL collective reduction order"""
 
+    skip_rl_trainer_initial_hf_weight_load: bool = False
```
tianyu-l: remove?

torchtitan/models/common/token_dispatcher.py
tianyu-l: is checking "inference mode" good enough? Do we have to introduce this config?

.github/workflows/integration_test_8gpu_rl.yaml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [ main ]
     paths:
       - 'torchtitan/experiments/rl/**'
-      - '.github/workflows/integration_test_4gpu_rl.yaml'
+      - '.github/workflows/integration_test_8gpu_rl.yaml'
```
wwwjn, line 8: This integration test is actually occupying 8 GPU, but running a RL integration test using 4 GPUs only. Rename to 8 GPUs

torchtitan/config/configs.py
```diff
@@ -385,6 +385,18 @@ class DebugConfig:
...
```
wwwjn: This flag can be removed because we can use trainer.checkpointer to control inital_weight_loading now

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -317,6 +318,7 @@ def __init__(
...
```
wwwjn, line 475: Added a NOTE to indicate vllm will build EP degree = DP* TP

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -124,10 +124,7 @@ def __init__(
 
         assert vllm_config is not None, "vllm_config is required"
 
-        # PP and CP are not supported on this inference path. User-facing
```
wwwjn, line 127: Reverted

torchtitan/models/common/token_dispatcher.py
```diff
@@ -297,9 +303,16 @@ def dispatch(
...
-        # All-to-all dispatch tokens to EP ranks
-        routed_input_RD = all_to_all_single_autograd(
-            routed_input_ND,
+        # All-to-all dispatch tokens to EP ranks.
+        # Use the non-autograd version under inference (vLLM), since
+        # _c10d_functional_autograd ops don't dispatch correctly without
+        # an active autograd context. Gated by a Python bool so the choice
+        # is stable at trace time.
+        a2a_fn = (
+            all_to_all_single if self.use_inference_a2a else all_to_all_single_autograd
```
wwwjn: https://github.com/pytorch/pytorch/pull/172792 After this PR merged, we can use `all_to_all_single` for both path

torchtitan/experiments/rl/config_registry.py
```diff
@@ -317,6 +321,164 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
...
+                max_tokens=50,
+            ),
+        ),
+    )
+
+
+def rl_grpo_qwen3_moe_debug_ep_batch_invariant() -> RLTrainer.Config:
+    """Batch-invariant MoE EP config for bitwise parity testing (8 GPUs).
+
+    Trainer: TP=4, EP=4 (4 GPUs). Generator: TP=4, EP=4 (4 GPUs).
```
wwwjn, line 402: In next PR which supports DP2EP, this config should be upgrade to DP2TP2, EP4

torchtitan/experiments/rl/tests/integration_tests.py
```diff
@@ -67,6 +67,31 @@ def build_rl_test_list() -> list[OverrideDefinitions]:
...
+        # End-to-end MoE debug model with TP=4, EP=4 (4 trainer + 4
+        # generator = 8 GPUs). Uses skip_rl_trainer_initial_hf_weight_load
+        # + the bundled tests/assets/qwen3_moe_debug bootstrap dir so no
+        # real HF checkpoint is needed.
+        OverrideDefinitions(
+            [
+                [
+                    "--module rl",
+                    "--config rl_grpo_qwen3_moe_debug_ep",
+                    "--hf_assets_path tests/assets/qwen3_moe_debug",
```
wwwjn: This points to a fake tokenizer folder, should be able to point to our test tokenizer folder with RL's ConfigParser PR merged

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -218,8 +220,27 @@ class _InferenceConfig:
...
+        # dispatchers to the non-autograd a2a path. Read at trace time.
+        from torchtitan.models.common.token_dispatcher import AllToAllTokenDispatcher
+
+        for layer in self.model.layers.values():
+            moe = getattr(layer, "moe", None)
+            if moe is None:
+                continue
+            dispatcher = getattr(moe.experts, "token_dispatcher", None)
+            if isinstance(dispatcher, AllToAllTokenDispatcher):
+                dispatcher.use_inference_a2a = True
```
wwwjn: I will remove this part, and land this PR after https://github.com/pytorch/pytorch/pull/172792, which is ready to be land

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -303,9 +324,10 @@ def compute_logits(
 
         logits = self.model.lm_head(hidden_states)
 
-        # Full DTensor path returns logits as DTensor; vLLM expects plain tensors.
+        # vLLM's sampling expects full plain tensors. full_tensor() handles both
+        # Replicate (no-op) and Shard(-1) (all-gather) lm_head output placements.
         if isinstance(logits, DTensor):
-            logits = logits.to_local()
+            logits = logits.full_tensor()
```
wwwjn: This line needs to be reverted. We use `disable_loss_parallel=True` in RL and the tensor will be Replicate(). Just to_local() is enough

wwwjn (commented):
Left comments for myself to be update after rebasing to last main

tests/assets/tokenizer/tokenizer.json
```diff
@@ -2040,7 +2040,14 @@
       "<|vision_start|>": 2005,
       "<|vision_end|>": 2006,
       "<|image_pad|>": 2007,
-      "<|video_pad|>": 2008
+      "<|video_pad|>": 2008,
```
wwwjn, line 2043: These special tokens are needed becuase external render (AutoTokenizer in RL) checks these and they are needed by Qwen3 model

wwwjn: hf_assets_path="tests/assets/tokenizer"
  renderer=RendererConfig(name="qwen3", enable_thinking=False)

torchtitan/experiments/rl/environment/token.py
```diff
@@ -177,7 +177,6 @@ async def step(self, completion: Completion) -> TokenEnvOutput:
             completion_message["reasoning_content"] = parsed.reasoning_content
         if parsed.tool_calls:
             completion_message["tool_calls"] = parsed.tool_calls
-
```
wwwjn, line 180: Need to revert

torchtitan/experiments/rl/config_registry.py
```diff
@@ -324,6 +328,205 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
...
+
+    Generator uses TP=4 for dense layers and EP=4 for MoE experts.
+    Trainer uses TP=4 for all layers.
+
+    Note: Qwen3-30B-A3B has 4 KV heads, so TP degree cannot exceed 4.
+    """
+    group_size = 8
+    return RLTrainer.Config(
+        model_spec=model_registry("30B-A3B", attn_backend="varlen"),
+        hf_assets_path="/data/users/jianiw/model/Qwen3-30B-A3B",
```
wwwjn: This needs to be reverted

torchtitan/experiments/rl/generate.py
wwwjn, line 1: Changes in this files just adding basic kwargs and make this script runnable with different configs, and align with the generator setup in RL.

torchtitan/models/common/moe.py
```diff
@@ -382,8 +382,18 @@ def forward(self, x_BLD: torch.Tensor) -> torch.Tensor:
...
         """
         B, L, D = x_BLD.shape
-        T = B * L
         sp_size = getattr(self.experts.token_dispatcher, "sp_size", 1)
+
+        # EP routes over sequence-parallel token shards. Very short decode
+        # forwards cannot shard across all SP ranks, so pad to ``sp_size`` and
+        # trim before returning. This is a no-op for training/prefill and EP-off;
+        # padded tokens only affect the inference-only load-balance counter.
+        seq_pad = sp_size - L if L < sp_size else 0
```
wwwjn, line 394: This is to solve https://github.com/pytorch/torchtitan/issues/3622 

TLDR  the issue is: when `seq_len >= TP_degree`, `num_local_tokens_per_expert_E = routing_map_BLE.sum(dim=(0, 1))`  suppose to be 
`V (tokens vary across TP ranks)  --sum(seq)-->  P (partial per-rank count)`. But when seq_len < TP, it degraded to R --> sum --> R, which can not be converted to P later. 

The simple fix there is to pad when `seq_len < TP_degree`

torchtitan/distributed/utils.py
```diff
@@ -265,6 +267,12 @@ def set_batch_invariance(enable: bool) -> None:
 
     _upstream_enable()
 
+    from vllm.model_executor.layers.batch_invariant import bmm_batch_invariant
```
wwwjn: This needs more discussion: We shouldn't put vllm as dependency in core. I will move these part into generator, as it's generator sepecific

Meanwhile I'm asking Claude to check the trace / read the code, where exactly is bmm used in MoE

torchtitan/models/common/moe.py
wwwjn: And this path should be only used when inference (however the pretraining might also hit seq_len < SP rank, we can just abandon it) - During inference the token_per_exepert_E would NOT matter much for inference. We need to enforce it's inference only

torchtitan/distributed/utils.py
wwwjn: Screenshot 2026-06-11 at 4 50 10 PM
Only generator model dispatch self.gate() to `bmm`, trainer dispatch it to `mm`

.github/workflows/integration_test_8gpu_rl_h100.yaml
tianyu-l, line 1: do you plan to delete this?

torchtitan/distributed/utils.py
tianyu-l: 1. Does TML package not support batch-invariant bmm?
2. which exact tensor is not contiguous? I'd like to check if we could just `.contiguous()` on it before doing linear. I think it's the ``x_BLD` here https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/moe.py#L402. Maybe let's call contiguous on it to avoid further complicating the code, and leave a TODO to investigate if we could refactor later.

torchtitan/experiments/rl/rollout_recorder.py
```diff
@@ -96,7 +106,9 @@ def record(
                     "step": step,
                     "is_validation": is_validation,
                     **self._encode_rollout(rollout),
-                }
+                },
+                # Validation rollouts can contain renderer-owned message objects.
```
tianyu-l: I didn't understand the comment and why we need this. cc @felipemello1

torchtitan/models/common/moe.py
tianyu-l: Let's add a TODO.

Please wrap the affected region with `-------------` so people know this is temporary code which needs to be removed.

torchtitan/models/common/moe.py
```diff
@@ -479,6 +489,10 @@ def _generate_routing_map(
 
         if shared_out_BLD is not None:
             out_BLD = out_BLD + shared_out_BLD
+
+        if seq_pad:
+            # Drop the tokens padded on for SP sharding, restoring (B, L, D).
+            out_BLD = out_BLD[:, : L - seq_pad, :]
```
tianyu-l, line 513: similar here

tianyu-l (commented):
one comment on the batch-invariant bmm, and some nits

.github/workflows/integration_test_8gpu_rl_h100.yaml
wwwjn: Yes! I plan to delete this - One blocker is 0.6B dense model OOM on A10G under batch-invariant mode. So I want to dig a little bit into this OOM in another PR

torchtitan/distributed/utils.py
wwwjn: 1. No, TML's version doesn't include batch-invariant bmm. vllm has batch-invariant bmm implemented.
2. I'm testing locally if a continous tensor will be dispatched to mm instead of bmm
Update: with only x_BLD.contiguous() and no bmm override, the drift stayed large (0.016, 2e-05, 0.031, ...),

The Aten dispatch rule for matmul is:

  if      (dim1==1 && dim2==1) → dot
  else if (dim1==2 && dim2==1) → mv
  else if (dim1==1 && dim2==2) → mm (unsqueeze)
  else if (dim1==2 && dim2==2) → mm                       // line 2044
  else if (should_fold(tensor1, tensor2, has_out)) → mm   // line 2046  (fold batch 
  into rows)
  else → bmm                                              // line 2112  
  (expand+broadcast batch)

Our gate is 3‑D activation @ 2‑D weight (dim1=3, dim2=2), so it lands on the should_fold decision. bmm is taken iff should_fold returns false. What should_fold do is decided at graph_tracing time, and it always returns false no matter we call `contiguous` or not.

In eager, a contiguous 3-D @ 2-D never produced bmm in eager, DTensor (Shard(1)/Replicate), or backed/unbacked-symbolic tests — they all fold to mm. **So the bmm is specifically the traced-inference-graph activation being non-contiguous** , which eager .contiguous() doesn't reach.

Another potential solution is fold the 3D activation into 2D (and change sharing config accrodingly)
```
  B, L, D = x_BLD.shape
  scores_BLE = self.gate(x_BLD.reshape(B * L, D)).view(B, L, -1)
```
I'm checking if we can implement this cleanly

torchtitan/experiments/rl/rollout_recorder.py
wwwjn: I hit something (tool calling results) is not serializable error when json dump

tianyu-l: comment that this is to bypass tool calling results.

torchtitan/distributed/utils.py
tianyu-l: this will introduce problem when spmd_types migration happens, so preferring patching vllm bmm for now.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -104,6 +104,31 @@ def _prepare_generation_request_metrics(
...
+def _patch_bmm_for_batch_invariance() -> None:
+    """Override ``aten::bmm`` with vLLM's batch-invariant bmm kernel.
+
+    torchtitan's batch-invariant mode (``batch_invariant_ops``, applied by
+    ``set_batch_invariance``) overrides ``mm``/``addmm``/``_log_softmax``/
+    ``mean.dim`` but not ``bmm``. The MoE router gate lowers to ``aten::bmm`` in
+    the vLLM inference graph (3-D activation @ 2-D weight), so without this the
+    generator's gate scores drift from the trainer's ``mm`` gate and flip top-k
+    expert routing, breaking on-policy logprob parity. Only the generator needs
+    it (the trainer's gate is ``mm``), so it lives here rather than in core.
```
tianyu-l: What I still don't understand is why trainer gate is folded to mm. In the [eager, compile, cudagraph] x [training, inference] matrix, possibly furthur expanded by TP/EP/SP configs, when do we hit `bmm`?

https://github.com/pytorch/torchtitan/pull/3142#discussion_r3406260330 explained some, but it's not entirely clear.

My point is that
- if this is e.g. aot_eager compile vs. eager mismatch, then we should ask compile experts to help resolve.
- basically I would like to learn "Unless xxx is willing to change, we couldn't really solve this problem". The way you describe the issue sounds to me that it's some mismatch, and we can fix it.

I don't mind leaving a TODO for investigation.

tianyu-l (approved):

wwwjn:
Model Test failure is not related

RL test passed: However, for MoE model the loss is 0 (which is expected because we random initialized weights)

PR #3143, opened 2026-04-28
[WIP] Enable DP-to-EP for MoE inference

wwwjn:
* __->__ #3143
* #3142

In development. Currently it doesn't work with weight loading

Map vLLM's data_parallel_size to dp_shard in TorchTitan's mesh math, enabling EP to span both DP and TP ranks (ep = dp * tp). For inference, the skip_dp path computes 2D meshes (fsdp + dp_replicate) then calls apply_fsdp which uses shard_placement_fn to route expert params to efsdp mesh and dense params to fsdp mesh.

Changes:
- parallel_dims: dp_replicate mesh always exists (needed for 2D mesh)
- vllm_wrapper: ep_size = dp_size * tp_size, dp_shard = dp_size
- qwen3/parallelize: inference path computes 2D meshes for apply_fsdp
- llama4/parallelize: clarify shard_placement_fn comments

### DIFF
--- a/torchtitan/distributed/parallel_dims.py
+++ b/torchtitan/distributed/parallel_dims.py
@@ -81,6 +81,10 @@ class ParallelDims:
             # Always keep fsdp mesh with real backend so fully_shard()
             # can apply MixedPrecisionPolicy even at degree 1.
             return True
+        if name == "dp_replicate":
+            # Always keep dp_replicate mesh so replicate() / fully_shard()
+            # can use it for data parallel replication even at degree 1.
+            return True
         if name == "efsdp":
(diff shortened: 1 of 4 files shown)


torchtitan/distributed/parallel_dims.py
```diff
@@ -81,6 +81,10 @@ def _mesh_exist(self, name: str, degree: int) -> bool:
             # Always keep fsdp mesh with real backend so fully_shard()
             # can apply MixedPrecisionPolicy even at degree 1.
             return True
+        if name == "dp_replicate":
```
wwwjn, line 84: We need to make dp_replicate always exist because we need a 2D mesh to apply DDP via fully_shard

PR #3144, opened 2026-04-28
[CI] Run integration tests in parallel

yuankaichen-amd:
Pack feature/model integration tests onto a fixed GPU pool so that tests
whose `ngpu` sums to the available budget run concurrently, instead of
sequentially monopolizing all 8 GPUs each. With the default `features`
suite this roughly halves wall-clock time on an 8-GPU host since most
flavors only need 4 GPUs.
- Add a small GPUPool (mutex+condvar) that hands out disjoint physical
  GPU id sublists and blocks when not enough are free.
- Submit runnable tests to a ThreadPoolExecutor sorted largest-first to
  reduce head-of-line blocking; each worker acquires `test_flavor.ngpu`
  GPUs, runs the full override_args sequence pinned to that subset via
  `CUDA_VISIBLE_DEVICES` / `HIP_VISIBLE_DEVICES`, then releases them.
  Sum of in-flight `ngpu` is bounded by `--ngpu` by construction;
  `torchrun --rdzv_endpoint=localhost:0` already auto-allocates ports.
- Replace `subprocess.run(..., capture_output=True)` with a Popen-based
  capture helper plus a global output lock: each test's merged
  stdout/stderr is buffered and emitted as one contiguous block under
  the lock, so concurrent tests never interleave their lines. Headers
  and footers carry the test name, start/end timestamps, command, and
  return code.
- Factor test filtering (test_name / exclude / disabled / arch /
  insufficient ngpu) into `_filter_tests` shared by both modes.
- New `--parallel` / `--no-parallel` flag (BooleanOptionalAction,
  default True). Sequential execution is still available via
  `--no-parallel`.


tests/integration_tests/flux.py
```diff
@@ -6,11 +6,21 @@
...
+
+def _run_cmd(cmd, timeout=None):
+    return subprocess.run(
+        [cmd],
+        encoding="utf-8",
+        errors="replace",
+        shell=True,
+        capture_output=True,
+        timeout=timeout,
+    )
```
fegin: Any reason why we treat Flux differently?

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+            futures: dict[Future, OverrideDefinitions] = {
+                ex.submit(_runner, t): t for t in scheduled
+            }
+            for fut in futures:
+                test_flavor = futures[fut]
+                try:
+                    fut.result()
+                except Exception as e:
+                    logger.error(str(e))
+                    failed_tests.append((test_flavor.test_name, str(e)))
```
fegin, line 272: Do we actually need a ThreadPoolExecutor? Will it be simpler if we just sort the the tests and do bucketing first and then run the tests deterministically? Another reason why I think we can pre-plan is that if all the tests do not use much memory, we can even let some tests share GPUs. It seems to be easier for a controller (main thread) to decide tests bucketing strategy.

fegin (commented):
Overall, I like the idea, but I'm not too sure about a GPUPool versus a controller style. For now, GPUPool may be more performant but if we want to further parallelize the runs, it seems that controller style is a better approach.

wdyt?

yuankaichen-amd:
> Overall, I like the idea, but I'm not too sure about a GPUPool versus a controller style. For now, GPUPool may be more performant but if we want to further parallelize the runs, it seems that controller style is a better approach.
> 
> wdyt?

Thanks for the comment! Can you elaborate what a "controller style" mean?

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+        gpu_env_prefix = ""
+
     for idx, override_arg in enumerate(test_flavor.override_args):
         cmd = ""
         if module is not None:
             cmd += f"MODULE={module} "
         if config is not None:
             cmd += f"CONFIG={config} "
-        cmd = f"NGPU={test_flavor.ngpu} LOG_RANK={all_ranks} ./run_train.sh"
+        cmd = (
```
shuhuayu: seems to have a previously existing bug here, it should be `cmd += ...`

tests/integration_tests/flux.py
shuhuayu: `flux` has a standalone entry of tests which is not using the parallel tests in this pr. Should be minor since there are only two tests in this category.

tests/integration_tests/run_tests.py
shuhuayu: hmm, i agree this static pre-planed way is better for resource sharing, but how can we estimate the memory usage for each test and handle the possible oom? Without gpu sharing, the current dynamic way seems fine to me. Suppose using a static bucket with 2 tests A and B, A takes 5 mins, B takes 20 mins, then there will GPUs being idle for 15 mins. Using the dynamic approach in this pr, gpus will be used as soon as they are freed up. What do you think?

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+            while len(self._free) < n:
+                self._cond.wait()
+            chosen = sorted(self._free[:n])
+            self._free = self._free[n:]
+            return chosen
+
+    def release(self, gpus: list[int]) -> None:
+        with self._cond:
+            self._free.extend(gpus)
+            self._cond.notify_all()
```
shuhuayu, line 57: iiuc, only the first batch of tests will follow the larget-first rule, after that, all tests with different sizes are equally lucky to get the available gpus? If this is true, we may need to make it explicit when defining `scheduled`.

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+            sys.stderr.write(footer)
+        sys.stderr.flush()
+
 
 def run_single_test(
     test_flavor: OverrideDefinitions,
     output_dir: str,
     module: str | None = None,
     config: str | None = None,
+    gpu_ids: list[int] | None = None,
```
shuhuayu, line 112: maybe add a comment here saying that `gpu_ids` is not `None` only in the parallel mode.

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+            captured.append(raw.rstrip("\n"))
+    finally:
+        proc.wait()
+        if timer is not None:
+            timer.cancel()
+
+    if timed_out:
+        raise subprocess.TimeoutExpired(cmd, timeout)
+
+    return proc.returncode, "\n".join(captured)
```
shuhuayu:
```suggestion
def _run_cmd(cmd, timeout=None):
    try:                                                                                                                                       
        return subprocess.run(                            
            [cmd],
            encoding="utf-8",                                                                                                                  
            errors="replace",
            shell=True,                                                                                                                        
            stdout=subprocess.PIPE,                       
            stderr=subprocess.STDOUT,
            timeout=timeout,
        )
    except subprocess.TimeoutExpired as e:
        stdout = e.stdout.decode("utf-8", errors="replace") if isinstance(e.stdout, bytes) else (e.stdout or "")                               
        return subprocess.CompletedProcess(cmd, -1, stdout=stdout, stderr=None)
```
We may use a two-line change to do the same work here:  replace `capture_output=True` with `stdout=subprocess.PIPE, stderr=subprocess.STDOUT`. The added `except` branch can help us store information in case of timeout.

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+                f"{gpu_env_prefix}NGPU={test_flavor.ngpu} LOG_RANK={all_ranks} "
                 f"CHECKPOINT_DIR={output_dir}/{test_name}/checkpoint/step-10 "
                 "PROMPT='What is the meaning of life?' "
                 f"./scripts/generate/run_llama_generate.sh --out > {output_dir}/{test_name}/generated_output.json"
             )
 
+        start_ts = time.strftime("%Y-%m-%d %H:%M:%S")
         try:
-            result = _run_cmd(cmd, timeout=test_flavor.timeout)
+            returncode, captured = _run_cmd_capture(
```
shuhuayu: We can stick to the updated `_run_cmd`.

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+                cmd, prefix=test_name, timeout=test_flavor.timeout
+            )
         except subprocess.TimeoutExpired as e:
             raise RuntimeError(
                 f"\nTest timed out after {test_flavor.timeout}s: {test_flavor.test_descr}.\n"
                 f"Command: {cmd}\n"
             ) from e
```
shuhuayu: Since `timeout` is handled, no longer need `try/except` here. `result = _run_cmd(cmd, timeout=test_flavor.timeout)` is enough.

tests/integration_tests/flux.py
shuhuayu: With the new `_run_cmd` function, this function may be unnecessary, and we can keep the import above.

tests/integration_tests/run_tests.py
```diff
@@ -24,109 +27,270 @@
...
+            self._free = self._free[n:]
+            return chosen
+
+    def release(self, gpus: list[int]) -> None:
+        with self._cond:
+            self._free.extend(gpus)
+            self._cond.notify_all()
+
+
+_OUTPUT_LOCK = threading.Lock()
```
shuhuayu: nit: can we move this line to the beginning of the file?

shuhuayu (changes requested):

tests/integration_tests/flux.py
yuankaichen-amd: I adopted @shuhuayu 's _run_cmd, and the changes in this file are reverted.

tests/integration_tests/run_tests.py
yuankaichen-amd: I share @shuhuayu 's argument -- I feel the pre-plan idea, in theory more efficient, may be an overkill for this purpose.

tests/integration_tests/run_tests.py
yuankaichen-amd: Good catch. Fixed.

tests/integration_tests/run_tests.py
yuankaichen-amd: You are right on this. Added a comment in the code to describe the behavior more clearly.

tests/integration_tests/run_tests.py
yuankaichen-amd: Done.

tests/integration_tests/run_tests.py
yuankaichen-amd: Thanks for the suggestion! Adopted.

tests/integration_tests/run_tests.py
yuankaichen-amd: Done.

tests/integration_tests/run_tests.py
yuankaichen-amd: Done.

tests/integration_tests/run_tests.py
yuankaichen-amd: Done.

tests/integration_tests/flux.py
yuankaichen-amd: Done.

tests/integration_tests/run_tests.py
yuankaichen-amd: I'm closing this for now. But happy to reopen if @fegin has additional inputs.

tests/integration_tests/run_tests.py
```diff
@@ -23,110 +26,253 @@
...
+    runnable, skipped_ngpu = _filter_tests(args, test_list)
+    for test_flavor in skipped_ngpu:
+        logger.info(
+            f"Skipping test {test_flavor.test_name} that requires {test_flavor.ngpu} gpus,"
+            f" because --ngpu arg is {args.ngpu}"
+        )
+
+    failed_tests: list[tuple[str, str]] = []
+
+    if args.parallel and runnable:
```
shuhuayu: I just realized the `run_tests` function is imported into multiple experiments folders, where `args` does not have the `parallel` attribute. Maybe we can make it an optional arg for this function with default value `True`?  And also i found `module` and `config` are always `None` in our use cases so there are dead code. But this is not related to this pr, and we can revisit later.

shuhuayu (changes requested):

tests/integration_tests/run_tests.py
yuankaichen-amd: This is a good point. Done.

shuhuayu (approved):
LGTM, thanks!

fegin (approved):
Memory and execution time planning is not hard for such a small problem set.  We just need to do a simple profiling and get a rough estimation is okay. Since we are using debug models, it is almost certain that running two tests on the same GPU will not be OOM. Doing this dynamically can complicate the test code. Given that this is just CI planning, we don't even to find the best solution, we just need a find solution, like packing two tests per time or three per time.

That being said, I agree this dynamic approach is good enough for now and is strictly better than what we have.

PR #3145, opened 2026-04-28
Improve compilation time (reduce from ~50 seconds to ~15s for vLLM)

### LINKED ISSUE #2509: [RL] Improve vLLM/Generator startup time with cudagraphs and support_torch_compile
As noted in initial PR https://github.com/pytorch/torchtitan/pull/2486 - the time with our vLLMWrapper is quite a bit larger (4x) than vLLM Native.  We should investigate this startup time and see how to reduce it

### LINKED ISSUE #3071: [rl] [feature] Cache cudagraphs for unified model generator
As title

### LINKED ISSUE #3119: replace vllm compile with trainer model compile?
investigate how impossible it is to move generator compilation logic from vllm to torchtitan 
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L200
->
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/parallelize.py#L92

Lucaskabela:
Fixes https://github.com/pytorch/torchtitan/issues/3119 and https://github.com/pytorch/torchtitan/issues/3071

## Summary
We make significant improvements to the vLLM compilation, saving ~40s (20s from cudagraph, 1s per step, and ~13s from Dynamo) from the following changes:

1) Since we are using FA based attention which is compatible with traceability, we can use FullGraph cudagraphs (which yields -1s per step)
2) We adjust the max cudagraph capture size based on default from configs - this cuts cudagraph capture time from 30s to 11s
3) We move compilation to use the same compile pipeline that trainer model does.  While this misses out on some of vLLM's custom passes, we have more observability and controllability to ensure unified definition.  This has the added benefit of leveraging the regional compile - reducing compile from O(n) to O(1) (as we compile one transformer layer then reuse it) cutting Dynamo compile time from 17s to 4s

## Test plan

```bash
python torchtitan/experiments/rl/grpo.py --module rl --config rl_grpo_qwen3_0_6b
```

### Before
```bash
INFO 04-28 16:20:20 [backends.py:1128] [actor=<root>.<torchtitan.experiments.rl.actors.generator.VLLMGenerator generator{'gpus': 0/4}>] Dynamo bytecode transform time: 15.26 s
Capturing CUDA graphs (mixed prefill-decode, PIECEWISE): 100%|███████████████████████████████████████████████████████████████████████████████████████████| 35/35 [00:31<00:00,  1.13it/s]
...
[actor=<root>] Step  9 | Loss: +0.0020 | Reward: +0.743 (correctness=+0.450, format=+0.293) | Avg tokens: 100 | Logprob diff: mean=-6.4715e-05, max=2.2314e-01 | Time: 2.7s
[actor=<root>] Post-training validation
[actor=<root>] Summary:
  Pre:  mean_reward=+0.365 (correctness=+0.200, format=+0.165)
  Post: mean_reward=+0.700 (correctness=+0.400, format=+0.300)
```

### After (this PR)
```bash
Capturing CUDA graphs (mixed prefill-decode, FULL): 100%|██████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [00:09<00:00,  1.34s/it]
...
[actor=<root>] Step  9 | Loss: +0.0020 | Reward: +0.743 (correctness=+0.450, format=+0.293) | Avg tokens: 100 | Logprob diff: mean=-6.4715e-05, max=2.2314e-01 | Time: 1.8s
[actor=<root>] Post-training validation
[actor=<root>] Summary:
  Pre:  mean_reward=+0.365 (correctness=+0.200, format=+0.165)
  Post: mean_reward=+0.700 (correctness=+0.400, format=+0.300)
```


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -31,60 +32,58 @@
...
+class VLLMCudagraphConfig:
+    """CUDA graph capture settings for the vLLM inference engine.
+
+    torch.compile is intentionally not configured here — the generator
+    reuses the trainer's ``CompileConfig`` (passed via ``compile_config``
+    at construction time) so both sides compile identically.  Only CUDA
+    graph capture, which is vLLM-specific, is controlled here.
+    """
+
+    cudagraph_mode: Literal["none", "full"] = "full"
```
tianyu-l: let's follow compile config style and have `enable` since it's a boolean option https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L323

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -31,60 +32,58 @@
...
-            )
+    "full" captures the entire forward as one graph.
+    "none" disables CUDA graph capture entirely.
+    "piecewise" and "full_and_piecewise" are intentionally excluded: they
+    require vLLM's whole-model torch.compile to split the graph around
+    non-capturable ops, which conflicts with per-layer compile.
+    See https://docs.vllm.ai/en/latest/design/cuda_graphs/#cudagraphmodes"""
 
     @property
     def is_eager(self) -> bool:
```
tianyu-l: don't need this aliasing field anymore

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -137,8 +139,15 @@ class Config(Configurable.Config):
...
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        compile: GeneratorCompileConfig = field(default_factory=GeneratorCompileConfig)
-        """Compilation and CUDA graph settings for the vLLM engine."""
+        max_num_seqs: int | None = None
+        """Maximum number of concurrent sequences the engine will batch.
+        When set, this is passed to vLLM EngineArgs and used to derive
+        CUDA graph capture sizes, avoiding captures for batch sizes that
+        will never be reached.  Auto-computed by RLTrainer from
+        num_prompts_per_step * sampling.n when not explicitly set."""
```
tianyu-l: what's the benefit of not using this default?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -174,12 +183,16 @@ def __init__(
         *,
         model_spec: ModelSpec,
         model_path: str,
+        compile_config: CompileConfig | None = None,
```
tianyu-l: should be part of Generator.Config? If we believe trainer and generator should always use the same compilation config (which I think is fine to assume for now), then it should be part of RLTrainer.Config?

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -170,6 +173,7 @@ def __init__(
         model_spec: ModelSpec,
         vllm_config: VllmConfig,
         prefix: str = "",
+        compile_config: CompileConfig | None = None,
```
tianyu-l: let's make it required in the call chain

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -200,17 +213,21 @@ def __init__(
...
-            enforce_eager=config.compile.is_eager,
+            enforce_eager=config.cudagraph.is_eager,
             hf_overrides={"architectures": [VLLM_MODEL_NAME]},
             attention_config=AttentionConfig(
                 backend=AttentionBackendEnum.CUSTOM,
             ),
             disable_log_stats=True,
         )
+        if config.max_num_seqs is not None:
+            engine_kwargs["max_num_seqs"] = config.max_num_seqs
```
tianyu-l: what does this field do?
what if we don't set it here?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -31,60 +32,58 @@
...
-            # Disable torch.compile but keep CUDA graphs (e.g. full mode).
-            # mode=0 (CompilationMode.NONE) prevents vLLM from inferring
-            # VLLM_COMPILE based on the default optimization level.
-            kwargs["mode"] = 0
-        else:
-            kwargs["backend"] = self.backend
+        kwargs: dict = dict(cudagraph_mode=self.cudagraph_mode, mode=0)
+
+        if max_num_seqs is not None and self.cudagraph_mode != "none":
+            kwargs["cudagraph_capture_sizes"] = self._compute_cudagraph_capture_sizes(
```
tianyu-l: what if we don't set it when cudagraph is enabled?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -31,60 +32,58 @@
...
-                f"cudagraph_mode='{self.cudagraph_mode}' requires piecewise graph "
-                "capture which depends on torch.compile. Set backend "
-                "to 'eager' or 'inductor'."
-            )
+    "full" captures the entire forward as one graph.
+    "none" disables CUDA graph capture entirely.
+    "piecewise" and "full_and_piecewise" are intentionally excluded: they
+    require vLLM's whole-model torch.compile to split the graph around
+    non-capturable ops, which conflicts with per-layer compile.
+    See https://docs.vllm.ai/en/latest/design/cuda_graphs/#cudagraphmodes"""
```
tianyu-l: What happens when we enable cudagraph for per-layer compile?

We can save compile time, but what's the impact on run time, e.g. when going to GB200 with significant CPU overhead.

tianyu-l: Also, it seems we move compile back to torchtitan, but cudagraph application still in vllm. How far are we from moving cudagraph application also in torchtitan?

Lucaskabela: The test plan shows the time impact - we observe speedup over piecewise in this particular setup

Lucaskabela: I would estimate 2-3weeks but I can leave a TODO here that we should unify cudagraph config once we have it on trainer side

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Defaults to the 256 which captures ~35 different sizes (ranging from 1 to 256) so no incorrectness, just more memory and startup time used

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Better flexibility - if we scale either of num_prompts_per_step or sampling.n this automatically scales with us

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Will up level to RLTrainer.Config then

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Commented above - this controls cudagraphs we capture, not setting it defaults to the behavior today on main

I can go ahead and make this set by default to avoid any sort of silent slowdown

torchtitan/experiments/rl/actors/generator.py
tianyu-l: please elaborate a bit more on what it means by "if we scale ..., this scales with us". My naive thought is that we will change them per job, which will change this cudagraph setting.

torchtitan/experiments/rl/actors/generator.py
tianyu-l: wait these are two different kwargs -- the other is for vllm cudagraph behavior, what's this additional kwarg for?

torchtitan/experiments/rl/actors/generator.py
tianyu-l: also would like to check if it works with EP, being enabled in https://github.com/pytorch/torchtitan/pull/3142

MoE has dynamic shape, despite being full graph torch-compilable

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: Sorry that is what I meant here - if we change num_prompts_per_step for a given job, we do not have to remember to also bump this value

torchtitan/experiments/rl/actors/generator.py
Lucaskabela: max_num_seqs controls other things like the padding for max size (used in kv cache)

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -89,11 +88,13 @@ def __init__(
         model_spec: ModelSpec,
         hf_assets_path: str = "",
         generator_dtype: str = "",
+        compile_config: CompileConfig,
```
tianyu-l: does this work in python? putting a arg without default to be after args with defaults

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -31,62 +31,53 @@
...
+    enable: bool = True
+    """Whether to enable CUDA graph capture (vLLM "full" mode)."""
 
-        return CompilationConfig(**kwargs)
+    # TODO: Validate CUDA graph capture with MoE / Expert Parallelism.
+    # MoE routing produces dynamic shapes that may conflict with full
+    # CUDA graph capture despite being torch.compile-compatible.
+
+    @staticmethod
+    def _compute_cudagraph_capture_sizes(max_num_seqs: int) -> list[int]:
```
tianyu-l: maybe inline this function into `get_vllm_compilation_config`

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -31,62 +31,53 @@
...
-                "to 'eager' or 'inductor'."
-            )
+class VLLMCudagraphConfig:
+    """CUDA graph capture settings for the vLLM inference engine.
 
-    @property
-    def is_eager(self) -> bool:
-        """Inferred from backend and cudagraph_mode."""
-        return self.backend == "none" and self.cudagraph_mode == "none"
+    torch.compile is configured separately via ``CompileConfig`` at the
```
tianyu-l, line 37: shall we add a TODO to explore moving cudagraph application to torchtitan model as well?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -200,17 +203,25 @@ def __init__(
...
                 backend=AttentionBackendEnum.CUSTOM,
             ),
             disable_log_stats=True,
         )
+        if config.max_num_seqs is None:
+            raise ValueError(
+                "max_num_seqs must be set (auto-computed by RLTrainer, or "
+                "set explicitly when using VLLMGenerator standalone)"
+            )
+        engine_kwargs["max_num_seqs"] = config.max_num_seqs
```
tianyu-l: add comment to educate
> max_num_seqs controls other things like the padding for max size (used in kv cache)

and what happens if not set explicitly

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -236,15 +240,22 @@ def __init__(
...
-        from torchtitan.config import ActivationCheckpointConfig, CompileConfig
+        # AC is disabled; skip_dp=True skips FSDP. compile_config is passed
+        # through so apply_compile runs per-layer after TP.
+        from torchtitan.config import ActivationCheckpointConfig
+
+        # CUDA graph warmup runs the model at each capture size with
+        # varying tensor types (AsyncCollectiveTensor vs plain tensor from
+        # TP collectives). Each combination triggers a dynamo recompile.
+        if compile_config.enable:
+            torch._dynamo.config.recompile_limit = 12
```
tianyu-l: despite the comments above, 12 still looks a magical number. Could you elaborate further
- where 12 comes from
- what happens if we don't set it

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -137,8 +130,14 @@ class Config(Configurable.Config):
         gpu_memory_limit: float = 0.9
         """Fraction of GPU memory to use for the vLLM engine (0.0 to 1.0)."""
 
-        compile: GeneratorCompileConfig = field(default_factory=GeneratorCompileConfig)
-        """Compilation and CUDA graph settings for the vLLM engine."""
+        max_num_seqs: int | None = None
+        """Maximum number of concurrent sequences the engine will batch.
+        When set, this is passed to vLLM EngineArgs.  Auto-computed by
+        RLTrainer from num_prompts_per_step * sampling.n when not
+        explicitly set."""
```
tianyu-l: > if we change num_prompts_per_step for a given job, we do not have to remember to also bump this value

Not sure if I understand. My (probably wrong) understanding was
- "Maximum number of concurrent sequences the engine will batch" should always be obtained from `num_prompts_per_step * sampling.n` per RL job, and there won't be any room to set it smaller.
- If not set explicitly in vllm kwargs, it will be inferred by vllm and causing potential overhead.
- There is no benefit to give control to users, and we should always set it to be `num_prompts_per_step * sampling.n` explicitly.

Where does the argument break?

Lucaskabela: Ah I think I misunderstood this line of questioning (thought it was more why do we need this at all) - I agree we can just hide this from users

torchtitan/experiments/rl/grpo.py
```diff
@@ -422,6 +427,8 @@ async def setup(
             config.generator,
             model_spec=config.model_spec,
             model_path=config.hf_assets_path,
+            compile_config=config.compile,
+            max_num_seqs=config.num_prompts_per_step * config.generator.sampling.n,
```
tianyu-l: generator is used to do generation both during training and validation, so should we take the maximum between this and https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/grpo.py#L487

tianyu-l (approved):
nice! one nit

PR #3146, opened 2026-04-28
[MoE][SAC] Use deterministic ops in MoE routing

songhappy:
MoE routing uses torch.histc and relies on torch.topk being recomputed under selective activation checkpointing. Neither op is deterministic on all backends (notably Intel XPU), so SAC recompute can produce different expert assignments than the original forward, silently corrupting gradients.

This PR:

Replaces torch.histc with torch.bincount in TokenChoiceTopKRouter and TokenReorderer — bincount is deterministic and equivalent for integer indices.
Adds aten.topk.default to the SAC save list so its output is reused on recompute.
Both changes are no-ops on backends where these ops are already deterministic (e.g. CUDA).

Changes
torchtitan/models/common/moe/moe.py: histc → bincount (×2)
torchtitan/distributed/activation_checkpoint.py: save aten.topk.default
tests/unit_tests/test_moe_routing.py: routing-count and recompute-consistency tests

Testing
`python -m unittest tests.unit_tests.test_moe_routing` — 3/3 pass on CUDA and Intel XPU. DeepSeek-V3 training on CUDA/XPU now runs cleanly under SAC.

Risk
No API changes. bincount and histc produce identical integer counts for valid inputs (covered by tests). Saving topk adds negligible memory.


acisseJZhong:
for the torch.hist -> bincount change, wondering could you measure speed / mfu before and after?

another approach we've been seeing used is `torch.scatter(zeros, -1, topk_ids, 1)` to replace `torch.hist`, curious to learn what's the performance comparison between these three options.

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -44,6 +44,9 @@ def _get_save_ops() -> set:
         # FlexAttention (torch.ops.higher_order.flex_attention is the same object)
         torch._higher_order_ops.flex_attention,
         torch.ops.aten.linear.default,
+        # topk can be non-deterministic on some backends; save to keep MoE
```
acisseJZhong: wondering did you try turning on torch determinism mode? is torch.topk not deterministic even under determinism mode? 

wondering is it possible to only add to the list for non-deterministic backend(e.g. XPU) instead of doing this for all backends.

songhappy: Good questions! I've tested both scenarios:

  1. Determinism mode testing:

  Yes, I tried --debug.deterministic --debug.seed=42 and it does work - the training completes
  without the tensor size mismatch error. This confirms that torch.topk is deterministic under
  determinism mode on XPU.

  However, the issue is that XPU's topk is non-deterministic by default (without explicit determinism
   mode), unlike CUDA. During activation checkpointing, when topk is recomputed in the backward pass
  without determinism enabled, it produces slightly different results (e.g., 275 vs 276 tokens routed
   to an expert), causing the MoE routing mismatch.

  2. Backend-specific approach:

  Yes, it's definitely possible to add topk only for non-deterministic backends. We could do
  something like:
` from torchtitan.tools.utils import device_type
  compute_ops = [
      # ... existing ops ...
  ]
  if device_type() == "xpu":
      compute_ops.append(torch.ops.aten.topk.default)`
  However, I'd argue for the simpler approach of saving topk unconditionally because:

  1. Negligible cost: Saving topk outputs (just indices + scores) has minimal memory overhead
  compared to saving full activation tensors
  2. Simpler code: Avoids backend-specific logic and future maintenance burden
  3. Safer: Protects against any potential non-determinism on other backends
  4. Cheap recompute anyway: topk is already fast, so recomputing vs saving has similar cost

  Which approach would you prefer? I'm happy to implement backend-specific logic if you feel strongly
   about it, but recommend keeping it simple and unconditional given the negligible overhead.

songhappy:
> for the torch.hist -> bincount change, wondering could you measure speed / mfu before and after?
> 
> another approach we've been seeing used is `torch.scatter(zeros, -1, topk_ids, 1)` to replace `torch.hist`, curious to learn what's the performance comparison between these three options.

  Great catch, I'm not changing from histc to bincount. After investigation:
  Root cause: topk is non-deterministic on XPU during activation checkpointing only, not the counting
  method.

  Solution: Save topk outputs in the AC save list (cheap: just scores + indices). This fixes the
  issue without requiring bincount which is deterministic.

 Performance on XPU (with topk fix):
  | Method | Step Time | vs histc |
  |--------|-----------|----------|
  | **histc** (keeping this) | 19.7 sec | baseline |
  | **scatter_reduce** | 19.4 sec | +1.5% (negligible) |
  | **bincount** | 23.6 sec | **-20%** ⚠️ |


  Conclusion: bincount is 20% slower with no benefit. Keeping histc.

  All methods produce identical loss values ✓

songhappy:
@acisseJZhong many thanks for reviewing, all comments are addressed. Please merge if no further issues.

acisseJZhong:
@songhappy thanks for the detailed experiments! Wondering how much memory increase/speed up do you observe when saving the indices and scores(although I thought in the code we are not using scores https://github.com/pytorch/torchtitan/blame/19c567f768fcffc4710ae4efe2441f575f823693/torchtitan/models/common/moe.py#L236?) from topk? Just curious to know, I don't have strong opinions against adding topk to save list. cc @tianyu-l

tests/unit_tests/test_moe_routing.py
```diff
@@ -0,0 +1,86 @@
...
+        self.assertEqual(counts.numel(), num_experts)
+        self.assertEqual(int(counts.sum().item()), n_tokens * top_k)
+        torch.testing.assert_close(
+            counts.to(torch.int64),
+            self._reference_counts(selected, num_experts),
+        )
+
+    def test_reorderer_counts(self):
+        num_experts, top_k, n_tokens = 8, 2, 64
+        reorderer = TokenReorderer(num_experts=num_experts, top_k=top_k)
```
tianyu-l: this is deprecated I think

tests/unit_tests/test_moe_routing.py
```diff
@@ -0,0 +1,86 @@
...
+        self, selected_experts_indices: torch.Tensor, num_experts: int
+    ) -> torch.Tensor:
+        return torch.histc(
+            selected_experts_indices.view(-1).float(),
+            bins=num_experts,
+            min=0,
+            max=num_experts,
+        ).to(torch.int64)
+
+    def test_router_counts(self):
```
tianyu-l: what's this testing?

tests/unit_tests/test_moe_routing.py
```diff
@@ -0,0 +1,86 @@
...
+        dim, num_experts, top_k, n_tokens = 32, 16, 4, 128
+
+        cfg = TokenChoiceTopKRouter.Config(top_k=top_k, score_func="softmax")
+        cfg.dim = dim
+        cfg.num_experts = num_experts
+        router = cfg.build(dim=dim, num_experts=num_experts)
+
+        x = torch.randn(n_tokens, dim)
+        _, sel1, c1 = router(x, expert_bias=None)
+        _, sel2, c2 = router(x, expert_bias=None)
```
tianyu-l: The fact that they could be different is the reason why we checkpoint topk results. I don't know what this test is trying to guard.

torchtitan/distributed/activation_checkpoint.py
tianyu-l: https://docs.pytorch.org/docs/main/generated/torch.topk.html
According to this official doc, topk may not be deterministic regardless of backend being xpu or not
```suggestion
        # topk can be non-deterministic; save to keep MoE
```

songhappy:
> @songhappy thanks for the detailed experiments! Wondering how much memory increase/speed up do you observe when saving the indices and scores(although I thought in the code we are not using scores https://github.com/pytorch/torchtitan/blame/19c567f768fcffc4710ae4efe2441f575f823693/torchtitan/models/common/moe.py#L236?) from topk? Just curious to know, I don't have strong opinions against adding topk to save list. cc @tianyu-l

| Metric | With topk save | Without topk save | Diff |
  |--------|---------------|-------------------|------|
  | Peak memory | 49.51 GiB (77.38%) | 49.51 GiB (77.38%) | 0 |
  | TPS @ step 10 | 485 | 476 | +1.9% |
  | TPS @ step 20 | 486 | 489 | -0.6% |
  | TPS @ step 30 | 506 | 506 | 0% |

  Config: DeepSeek-V3 16B, 8x Intel Max 1550, seq_len=2048, selective AC, 30 steps. Zero memory overhead, no speed impact. topk outputs are indices+scores of shape [num_tokens, top_k]
  (~64KB per MoE layer) — negligible

acisseJZhong (approved):

tianyu-l (approved):

PR #3148, opened 2026-04-29
[graph_trainer] Apply simple_fsdp unconditionally for NGPU=1 parity

aditvenk:
The main trainer applies fully_shard unconditionally so MixedPrecisionPolicy's param_dtype cast runs even at degree 1 — the `fsdp` mesh is deliberately kept with a real backend for this (ParallelDims._mesh_exist). graph_trainer's llama3/deepseek_v3/qwen3 parallelize fns instead gated simple_fsdp on fsdp/ep/dp_replicate being enabled, so on a single GPU parameters stayed in fp32 and any bf16-only kernel downstream (e.g. MXFP8) failed.
    
Drop the guard and consolidate the three near-identical data_parallel blocks into apply_simple_fsdp in common_utils. MoE experts handling folds in via ep_enabled + moe_enabled checks (no-op for dense models).
    
Adds tests/test_simple_fsdp.py asserting the bf16 param cast at NGPU=1: the parametrized weight access yields bf16 while the underlying storage stays fp32, and a forward pass produces bf16 activations. Verified the test fails when the gating is reintroduced inside apply_simple_fsdp.

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -13,7 +13,12 @@ from torch.distributed.tensor import DTensor, Replicate
 from torch.fx.traceback import annotate_fn
 from torch.utils._pytree import register_constant, register_pytree_node, tree_map
 
+from torchtitan.config import TORCH_DTYPE_MAP, TrainingConfig
 from torchtitan.distributed import ParallelDims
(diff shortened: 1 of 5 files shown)


aditvenk:
@claude Review this PR

SherlockNoMad (approved):

yiming0416 (approved):
Let's wait for the H100 CI results before landing

PR #3149, opened 2026-04-29
refactor get_attention_masks to use positions instead of eos_id

### LINKED ISSUE #2716: [RL] Unify mask creation logic for RL.

### LINKED ISSUE #3130: SFT eot_id and causality bugs
### Bug description

Believe I found two issues with how `get_document_mask_mod` is used for SFT today. Discussed in the order I found them.  Didn't see these discussed in existing issue, but maybe missed it.

## Issue 1: `get_document_mask_mod` Assumes Single `eos_id` 

`get_document_mask_mod` in `torchtitan/models/common/attention.py:400-422` uses the tokenizer's `eos_id` to detect document boundaries in packed sequences. This is a bug for models whose `eos_id` also serves other roles, specifically multiple important models use the same token for `eos_id` and for the closing tag for the role markers in their chat templates.

In these examples, a typical chat example for these models will look like:
```
<|im_start|>system\nYou are a...<|im_end|>\n      <- false boundary
<|im_start|>user\nProve sqrt(2)...<|im_end|>\n        <- false boundary
<|im_start|>assistant\n<think>...\n</think>\n...<|im_end|>  <- actual doc end
```
Rather than treating this as a single example which should be attended over in whole, it's instead treated as three disjoint sequences, which corrupts training. Keying off of `eos_id` creates false causal boundaries, in this scenario. Since only the assistant response is trained on, the model never learns to react to the chat template's system/user format (the default `ChatDataset` only supports a user field today, and not the three turn system/user/assistant format, but I'm describing the more general scenario).

Model examples:
* Recent Qwen models use `<|im_end|>` [as the eos token](https://huggingface.co/Qwen/Qwen3.6-35B-A3B/blob/main/tokenizer_config.json#L287) and as the [role-end marker](https://huggingface.co/Qwen/Qwen3.6-35B-A3B/blob/main/chat_template.jinja#L88)
* Nemotron models follow the same pattern:  `<|im_end|>` [as the eos token](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16/blob/main/special_tokens_map.json#L10) and as the [role-end marker](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16/blob/main/chat_template.jinja#L101)

### Code path and Root Cause

Code path:

```
config: attn_backend="flex"
  -> config_utils.py:44: get_attention_config("flex") -> mask_type="block_causal"
  -> trainer.py:627: model.get_attention_masks(input_batch, tokenizer)
    -> decoder.py:154: get_document_mask_mod(input_batch, tokenizer.eos_id)
      -> attention.py:411: eos_mask = batch == eos_id  # fires at every <|im_end|>
```

`get_document_mask_mod` assumes `eos_id` appears exactly once per document (at the very end):

```python
def get_document_mask_mod(batch: torch.Tensor, eos_id: int) -> _mask_mod_signature:
    eos_mask = batch == eos_id  # <-- fires at every role close
    eos_mask[:, -1] = True
    cumulative_mask = torch.cumsum(torch.where(eos_mask, 1, 0), dim=1)
    sequence_indices = torch.zeros_like(cumulative_mask, dtype=torch.int32)
    sequence_indices[:, 1:] = cumulative_mask[:, :-1]
    ...
```
When this assumption is violated, the attention bug appears.

## Issue 2: EOS is stripped from `input_ids`

Even without the false-positive problem above, `get_document_mask_mod` would still miss the actual boundary between packed samples. `_tokenize_sample` (`hf_datasets/text_datasets.py:396`) produces:

```python
input_ids = full_tokens[:-1]  # EOS stripped — lives only in label_ids
label_ids = full_tokens[1:]
```

`full_tokens` is guaranteed to end with `eos_id` (lines 385-386), so the boundary EOS is always the token that gets cut. The packed `input_batch` passed to `get_document_mask_mod` looks like:

```
[<s>, ...sample_1..., <s>, ...sample_2..., EOS, EOS, ..., EOS]
                    ^--- no EOS here; only padding EOS at the end
```

`eos_mask = batch == eos_id` fires on the padding tokens but not at sample boundaries, so all content tokens from every packed sample land in `doc_id=0` and cross-sample attention is not blocked.

### Reproducer

Uses the test tokenizer and `ChatDataset` from the repo (`tests/unit_tests/test_chat_dataset.py` setup):
```python

import torch
from datasets import Dataset

from torchtitan.components.tokenizer import HuggingFaceTokenizer
from torchtitan.hf_datasets.text_datasets import ChatDataset
from torchtitan.models.common.attention import get_document_mask_mod

tokenizer = HuggingFaceTokenizer(tokenizer_path="tests/assets/tokenizer/")

ds = Dataset.from_json("tests/assets/sft_test/data.json")
assert len(ds) >= 2, "Need at least two samples"


def _process(sample):
    return [
        {"role": "user", "content": sample["question"]},
        {"role": "assistant", "content": sample["answer"]},
    ]


chat_ds = ChatDataset(
    dataset=ds,
    tokenizer=tokenizer,
    sample_processor=_process,
    seq_len=2048,
    infinite=False,
)

# _tokenize_sample returns input_ids = full_tokens[:-1]; EOS lives only in label_ids.
r0 = chat_ds._tokenize_sample(ds[0])
r1 = chat_ds._tokenize_sample(ds[1])
assert r0 is not None and r1 is not None
input_ids_0, label_ids_0 = r0
input_ids_1, _ = r1

assert input_ids_0[-1] != tokenizer.eos_id  # EOS absent from inputs
assert label_ids_0[-1] == tokenizer.eos_id  # EOS present only as prediction target

# Pack two samples exactly as _iter_greedy_packed does: concatenate input_ids directly.
packed = input_ids_0 + input_ids_1
boundary = len(input_ids_0)

# No EOS at or near the packed-sample boundary — the doc separator is invisible.
assert tokenizer.eos_id not in packed[boundary - 2 : boundary + 2]

# get_document_mask_mod finds no eos_id at the boundary, so both samples
# are assigned doc_id=0. Cross-sample attention is not blocked.
batch = torch.tensor([packed])  # [1, seq_len]
mask_mod = get_document_mask_mod(batch, tokenizer.eos_id)

# The first tok of doc 1 can attend to the last tok of doc 0:
b, h = torch.tensor(0), torch.tensor(0)
q = torch.tensor(boundary)       # first token of sample 1
kv = torch.tensor(boundary - 1)  # last token of sample 0
# Should be False (different documents); is True due to the bug.
assert mask_mod(b, h, q, kv).item()

# And the last tok of doc 1 can attend to the very first tok of doc 0:
q_last = torch.tensor(len(packed) - 1)  # last token of sample 1
kv_first = torch.tensor(0)              # first token of sample 0
# Should be False (different documents); is True due to the bug.
assert mask_mod(b, h, q_last, kv_first).item()

print("All assertions passed — reproducer confirmed the bug")
```

---

## Position Fix?

I think a robust fix is to use the `positions` field from the data post-processing method (and ultimately generated during dataset creation) to mark doc boundaries, rather than relying on `eos_id`: https://github.com/pytorch/torchtitan/blob/b9e33527530dd16d6e435b4dbf208bdb3ca08f83/torchtitan/trainer.py?plain=1#L595-L603

Whenever the positions field is not sequentially increasing we have hit a new doc (at least in the simplest cases, not sure if there are more complex scenarios that need consideration).  Could modify `get_document_mask_mod` or create a new `get_document_mask_mod_from_positions`? I've implemented the latter locally and verified that SFT is effective w/ these changes, and badly broken without them.

The causality invariants appear to be untested in `tests/unit_tests/test_chat_dataset.py`, also.

CC @fegin @tianyu-l @wwwjn @drisspg @rakkit

### Versions

```
❯ uv pip freeze
absl-py==2.4.0
aiohappyeyeballs==2.6.1
aiohttp==3.13.5
aiosignal==1.4.0
annotated-doc==0.0.4
annotated-types==0.7.0
anyio==4.13.0
attrs==26.1.0
certifi==2025.8.3
cfgv==3.5.0
charset-normalizer==3.4.3
click==8.3.3
coverage==7.13.5
cuda-bindings==13.0.3
cuda-pathfinder==1.2.2
cuda-toolkit==13.0.2
datasets==4.7.0
debugpy==1.8.20
dill==0.4.0
distlib==0.4.0
docstring-parser==0.18.0
dtest @ git+https://github.com/garrett361/dtest@2d4c5cf78da82d977d529605d35cc2ce5eb2a0ca
einops==0.8.2
expecttest==0.3.0
filelock==3.25.2
frozenlist==1.8.0
fsspec==2026.2.0
gitdb==4.0.12
gitpython==3.1.47
grpcio==1.80.0
h11==0.16.0
hf-xet==1.4.3
httpcore==1.0.9
httpx==0.28.1
huggingface-hub==1.12.0
identify==2.6.19
idna==3.10
iniconfig==2.3.0
jinja2==3.1.6
markdown==3.10.2
markdown-it-py==4.0.0
markupsafe==3.0.2
mdurl==0.1.2
mpmath==1.3.0
multidict==6.7.1
multiprocess==0.70.18
networkx==3.6.1
nodeenv==1.10.0
numpy==2.4.4
nvidia-cublas==13.1.1.3
nvidia-cuda-cupti==13.0.85
nvidia-cuda-nvrtc==13.0.88
nvidia-cuda-runtime==13.0.96
nvidia-cudnn-cu13==9.20.0.48
nvidia-cufft==12.0.0.61
nvidia-cufile==1.15.1.6
nvidia-curand==10.4.0.35
nvidia-cusolver==12.0.4.66
nvidia-cusparse==12.6.3.3
nvidia-cusparselt-cu13==0.8.1
nvidia-nccl-cu13==2.29.7
nvidia-nvjitlink==13.0.88
nvidia-nvshmem-cu13==3.4.5
nvidia-nvtx==13.0.85
packaging==26.2
pandas==3.0.2
pillow==12.2.0
platformdirs==4.9.6
pluggy==1.6.0
pre-commit==4.6.0
propcache==0.4.1
protobuf==7.34.1
pyarrow==24.0.0
pydantic==2.13.3
pydantic-core==2.46.3
pygments==2.20.0
pyrefly==0.45.1
pytest==7.3.2
pytest-cov==7.1.0
python-dateutil==2.9.0.post0
python-discovery==1.2.2
python-dotenv==1.0.1
pyyaml==6.0.3
regex==2026.4.4
requests==2.32.4
rich==15.0.0
ruff==0.15.12
safetensors==0.7.0
sentry-sdk==2.58.0
setuptools==78.1.0
shellingham==1.5.4
six==1.17.0
smmap==5.0.3
sympy==1.14.0
tensorboard==2.20.0
tensorboard-data-server==0.7.2
tokenizers==0.22.2
torch==2.13.0.dev20260422+cu130
torchdata==0.12.0.dev20250220
-e file:///proj/checkpoints/goon/github/garrett361/torchtitan-g41-sft
tqdm==4.67.3
transformers==5.6.2
triton==3.7.0+gitb4e20bbe
ty==0.0.32
typeguard==4.5.1
typer==0.24.2
typing-extensions==4.15.0
typing-inspection==0.4.2
tyro==1.0.13
urllib3==2.5.0
virtualenv==21.2.4
wandb==0.26.1
werkzeug==3.1.8
xxhash==3.6.0
yarl==1.23.0
```

liangel-02:
1. fixes #3130, use positions instead of eos_id incorrectly to avoid incorrectly stripping it or assuming it correctly determines the end of a sequence
2. fixes #2716, removes tokenizer from get_attention_masks so the rl loop can reuse it and we can have a unified definition

added test to `tests/unit_tests/test_chat_dataset.py`, ran existing `experiments/graph_trainer/tests/test_trace_module.py`

varlen GRPO loop numerics are identical with and without refactor 
Screenshot 2026-04-30 at 12 17 24 PM

flex llama 3 training run numerics are also identical 
Screenshot 2026-04-30 at 1 18 14 PM
Screenshot 2026-04-30 at 1 17 31 PM

### DIFF
--- a/torchtitan/experiments/rl/actors/utils.py
+++ b/torchtitan/experiments/rl/actors/utils.py
@@ -4,29 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import itertools
-
 import torch
 import torch.nn.functional as F
-
-from torchtitan.models.common.attention import VarlenMetadata
-
-# TODO We should either unify all the mask creation for RL, or move them to a
-#      single file.
-def create_varlen_metadata(seq_lens: list[int], device: torch.device) -> VarlenMetadata:
-    """Build VarlenMetadata from sequence lengths.
-
-    Example:
-        seq_lens = [3, 5, 2]
-        -> cu_seqs = [0, 3, 8, 10], max_len = 5
-    """
-    cu_seqs = torch.tensor(
-        [0] + list(itertools.accumulate(seq_lens)), dtype=torch.int32, device=device
-    )
-    max_len = max(seq_lens)
-    return VarlenMetadata(
-        cu_seq_q=cu_seqs, cu_seq_k=cu_seqs, max_q=max_len, max_k=max_len
-    )
 
 
 def compute_logprobs(logits: torch.Tensor, token_ids: torch.Tensor) -> torch.Tensor:
@@ -66,18 +45,6 @@ def extract_response_logprobs(
         result.append(packed_logprobs[0, s:e])
         seq_start += seq_lens[i]
     return result
-
-
-def create_positions_from_seq_lens(
-    seq_lens: list[int], device: torch.device
-) -> torch.Tensor:
-    """Build a ``[1, total_tokens]`` positions tensor that resets at each sequence boundary.
-
-    Example:
-        seq_lens = [3, 5, 2]
-        -> positions = [[0, 1, 2, 0, 1, 2, 3, 4, 0, 1]]
-    """
-    return torch.cat([torch.arange(l, device=device) for l in seq_lens]).unsqueeze(0)
 
 
 def verify_logprob_identity(
--- a/torchtitan/models/common/attention.py
+++ b/torchtitan/models/common/attention.py
@@ -312,27 +312,25 @@ def get_causal_mask_mod() -> _mask_mod_signature:
     return _causal_mask
 
 
-def get_document_mask_mod(batch: torch.Tensor, eos_id: int) -> _mask_mod_signature:
+def get_document_mask_mod(positions: torch.Tensor) -> _mask_mod_signature:
     """Creates a document mask that prevents attention across document boundaries.
 
+    Document boundaries are detected where ``positions`` resets to 0, which
+    marks the start of a new packed document.
+
     Args:
-        batch: Input batch tensor with shape [b, s, h, d]
-        eos_id: End-of-sequence token ID that marks document boundaries
+        positions: Per-token position tensor with shape ``[b, s]``. Positions
+            reset to 0 at each document start.
 
     Returns:
         A mask modifier function that implements document-level masking.
     """
-    # batch is [b, s, h, d] shape
-    eos_mask = batch == eos_id
-    eos_mask[:, -1] = True
-    cumulative_mask = torch.cumsum(torch.where(eos_mask, 1, 0), dim=1)
-    sequence_indices = torch.zeros_like(cumulative_mask, dtype=torch.int32)
-    sequence_indices[:, 1:] = cumulative_mask[:, :-1]
+    doc_ids = torch.cumsum((positions == 0).int(), dim=1) - 1
 
     def document_mask(
         b: torch.Tensor, h: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
     ) -> torch.Tensor:
-        return sequence_indices[b, q_idx] == sequence_indices[b, kv_idx]
+        return doc_ids[b, q_idx] == doc_ids[b, kv_idx]
 
     return document_mask
 
@@ -403,35 +401,30 @@ def create_attention_mask(*args, **kwargs):
 
 
 def create_varlen_metadata_for_document(
-    input_batch: torch.Tensor, eos_id: int
+    positions: torch.Tensor,
 ) -> VarlenMetadata:
-    """
-    Creates cumulative sequence length indices needed for variable length attention
+    """Creates cumulative sequence length indices needed for variable length attention.
+
+    Document boundaries are detected where ``positions`` resets to 0 (same
+    convention as :func:`get_document_mask_mod`).
 
     Args:
-        input_batch
-        eos_id: the EOS id marker
+        positions: Per-token position tensor with shape ``[b, s]``. Positions
+            reset to 0 at each document start.
 
     Returns:
         VarlenMetadata containing cumulative sequence length indices for q, k, and max_seq_len
     """
-    batch_size, seq_len = input_batch.shape
-    device = input_batch.device
+    batch_size, seq_len = positions.shape
+    device = positions.device
     cu_seqlens_list, all_seq_lengths = [], []
     offset = 0
-    max_seqlen = 0
 
     for b in range(batch_size):
-        tokens = input_batch[b]
-        eos_positions = (tokens == eos_id).nonzero(as_tuple=True)[0].to(torch.int32)
+        doc_starts = (positions[b] == 0).nonzero(as_tuple=True)[0].to(torch.int32)
         sample_cu_seqlens = torch.cat(
-            [
-                torch.tensor([0], dtype=torch.int32, device=device),
-                eos_positions + 1,
-                torch.tensor([seq_len], dtype=torch.int32, device=device),
-            ]
+            [doc_starts, torch.tensor([seq_len], dtype=torch.int32, device=device)]
         )
-        sample_cu_seqlens = torch.unique_consecutive(sample_cu_seqlens)
 
         seq_lengths = torch.diff(sample_cu_seqlens)
         all_seq_lengths.append(seq_lengths)
--- a/torchtitan/models/common/decoder.py
+++ b/torchtitan/models/common/decoder.py
@@ -9,7 +9,6 @@ from dataclasses import dataclass
 import torch
 from torch.nn.attention.flex_attention import and_masks
 
-from torchtitan.components.tokenizer import BaseTokenizer
 from torchtitan.models.common.attention import (
     AttentionMasksType,
     BaseAttention,
@@ -151,8 +150,7 @@ class Decoder(BaseModel):
 
     def _get_flex_attention_masks(
         self,
-        input_batch: torch.Tensor,
-        tokenizer: BaseTokenizer,
+        positions: torch.Tensor,
         attn_config: BaseAttention.Config,
     ) -> AttentionMasksType:
         mask_mods = [get_causal_mask_mod()]
@@ -161,42 +159,39 @@ class Decoder(BaseModel):
             case "causal":
                 B = 1
             case "block_causal":
-                B = input_batch.shape[0]
-                assert tokenizer.eos_id is not None
-                mask_mods.append(get_document_mask_mod(input_batch, tokenizer.eos_id))
+                B = positions.shape[0]
+                mask_mods.append(get_document_mask_mod(positions))
             case _:
                 raise ValueError(
                     f"Unknown attention mask type: {attn_config.mask_type}"
                 )
 
+        seq_len = positions.shape[1]
         assert isinstance(attn_config.inner_attention, FlexAttention.Config)
         return create_attention_mask(
             and_masks(*mask_mods),
             B,
             None,
-            input_batch.shape[1],
-            input_batch.shape[1],
+            seq_len,
+            seq_len,
             BLOCK_SIZE=attn_config.inner_attention.block_size,
         )
 
     def get_attention_masks(
         self,
-        input_batch: torch.Tensor,
-        tokenizer: BaseTokenizer,
-        extra_inputs: dict[str, torch.Tensor] | None = None,
+        positions: torch.Tensor,
     ) -> AttentionMasksType:
         attn_config = self.config.layers[0].attention
         inner_attn = attn_config.inner_attention
         if isinstance(inner_attn, FlexAttention.Config):
-            return self._get_flex_attention_masks(input_batch, tokenizer, attn_config)
+            return self._get_flex_attention_masks(positions, attn_config)
         elif isinstance(inner_attn, VarlenAttention.Config):
             if attn_config.mask_type != "block_causal":
                 raise ValueError(
                     f"varlen attention is only supported with block_causal "
                     f"attention mask type, got {attn_config.mask_type}"
                 )
-            assert tokenizer.eos_id is not None
-            return create_varlen_metadata_for_document(input_batch, tokenizer.eos_id)
+            return create_varlen_metadata_for_document(positions)
         else:
             raise TypeError(
                 f"Only VarlenAttention and FlexAttention support attention masks, "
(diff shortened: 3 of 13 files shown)


torchtitan/models/common/attention.py
```diff
@@ -311,27 +317,25 @@ def _causal_mask(
     return _causal_mask
 
 
-def get_document_mask_mod(batch: torch.Tensor, eos_id: int) -> _mask_mod_signature:
+def get_document_mask_mod(positions: torch.Tensor) -> _mask_mod_signature:
     """Creates a document mask that prevents attention across document boundaries.
 
+    Document boundaries are detected where ``positions`` resets to 0, which
```
tianyu-l, line 318: how valid is this? For training examples constructed from torchtitan, this seems correct. How about continuous batched samples in generator?

E.g.
- request 1 iteration 1 positions [0, 1, 2], iteration 2 positions [3]
- request 2 iteration 2 positions [0, 1, 2, 3], iteration 2 [4]

torchtitan/models/gpt_oss/model.py
```diff
@@ -253,8 +252,7 @@ def __init__(self, config: Config):
     def get_attention_masks(
         self,
         input_batch: torch.Tensor,
-        tokenizer: BaseTokenizer,
-        extra_inputs: dict[str, torch.Tensor] | None = None,
+        positions: torch.Tensor | None = None,
```
tianyu-l: let's make this required, we can raise ValueError at callsites (e.g. trainer)

torchtitan/models/common/decoder.py
```diff
@@ -182,21 +180,25 @@ def _get_flex_attention_masks(
     def get_attention_masks(
         self,
         input_batch: torch.Tensor,
-        tokenizer: BaseTokenizer,
-        extra_inputs: dict[str, torch.Tensor] | None = None,
+        positions: torch.Tensor | None = None,
```
tianyu-l: similar

torchtitan/experiments/vlm/model/model.py
tianyu-l, line 1: @shuhuayu I think we can deprecate `vlm`

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -4,29 +4,27 @@
...
-from torchtitan.models.common.attention import VarlenMetadata
+from torchtitan.models.common.attention import (
+    create_varlen_metadata_for_document,
+    VarlenMetadata,
+)
+
 
-# TODO We should either unify all the mask creation for RL, or move them to a
-#      single file.
 def create_varlen_metadata(seq_lens: list[int], device: torch.device) -> VarlenMetadata:
```
tianyu-l: I think we can remove this function, and simplify the callsite https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/trainer.py#L284

torchtitan/experiments/rl/actors/utils.py
tianyu-l: remove this function as well

torchtitan/experiments/vlm/model/model.py
shuhuayu: Sounds good. should i make a pr to deprecate it given that now we have a vlm (qwen3_vl) in core.

tianyu-l: @shuhuayu Yes please.

torchtitan/models/common/attention.py
liangel-02: discussed offline, this is fine for generator side since vllm will handle this

torchtitan/models/common/decoder.py
```diff
@@ -162,8 +161,7 @@ def _get_flex_attention_masks(
                 B = 1
             case "block_causal":
                 B = input_batch.shape[0]
```
tianyu-l: can we remove input_batch from signature of `_get_flex_attention_masks` and `get_attention_masks`? The shape should be equivalently inferred from `positions`.

tianyu-l (approved):
LGTM. Could you show that numerics are bitwise the same as prior to this change, for both varlen and flex? (we probably should have this in CI, but can do that later)

torchtitan/trainer.py
```diff
@@ -613,14 +612,16 @@ def post_dataloading_process(
...
-                assert (
-                    self.tokenizer is not None
-                ), "tokenizer is required for flex/varlen attention"
+                # Flex/varlen always need positions for mask creation.
+                # For block_causal they're already in extra_kwargs; for
+                # causal mode synthesise sequential positions.
+                if extra_kwargs.get("positions") is None:
+                    extra_kwargs["positions"] = torch.arange(
+                        0, inputs.shape[1], dtype=torch.int32, device=self.device
+                    ).expand(inputs.shape)
```
tianyu-l: logic here is complicated and repeated. Maybe worth cleaning globally?
- e.g. when Flex / Varlen is used, we usually require block_causal masking? In which case we should error out if `positions` is not given? 
- what happens if gpt-oss uses attention-sink which is Flex only?

can be done in a separate PR, but if you want to do that in separate PR, we can be conservatite here -- error out as much as possible

liangel-02: i removed the repeated logic. for ur first point, i added an error check. for the second point, this would've been caught in the decoder but i added an error check so we can catch it earlier

torchtitan/trainer.py
```diff
@@ -594,35 +596,41 @@ def post_dataloading_process(
...
-                    self.tokenizer is not None
-                ), "tokenizer is required for flex/varlen attention"
-                model = cast(Decoder, self.model_parts[0])
-                extra_kwargs["attention_masks"] = model.get_attention_masks(
-                    input_batch=inputs,
-                    tokenizer=self.tokenizer,
-                    extra_inputs=extra_inputs,
+                raise ValueError(
+                    "VarlenAttention requires block_causal mask_type, "
+                    f"got '{mask_type}'."
```
tianyu-l: this error should be raised inside `VarlenAttention` (if not already) -- let's keep the trainer succinct

torchtitan/trainer.py
```diff
@@ -594,35 +596,41 @@ def post_dataloading_process(
...
+            raise ValueError(
+                "block_causal mask_type requires per-document positions "
+                "from the dataloader, but none were provided."
+            )
+        elif positions is not None:
+            extra_kwargs["positions"] = positions
+        elif self.parallel_dims.cp_enabled or needs_attention_masks:
+            extra_kwargs["positions"] = torch.arange(
+                0, inputs.shape[1], dtype=torch.int32, device=self.device
+            ).expand(inputs.shape)
```
tianyu-l: I think the current text/chat dataloader always generate positions
https://github.com/pytorch/torchtitan/blob/main/torchtitan/hf_datasets/text_datasets.py#L164

so let's assert that and kill these branches

tianyu-l (commented):
some comments on simplification in trainer

PR #3156, opened 2026-04-29
[graph_trainer] FSDP AG RS overlap

yiming0416:
Introduced `overlap_fsdp_ag_rs_pass` to overlap AG and RS in FSDP by creating a separate PG for AG (renaming and refactoring the old `reassign_to_pg_pass`).

The pass:
- Discovers the FSDP PG by walking the graph for FSDP wait_tensor nodes (is_wait_tensor_from_fsdp).
- Lazily creates the extra PG on first use via an internal _EXTRA_FSDP_PG_REGISTRY keyed by source PG name. 
- Rewrites all matching all-gather nodes to the extra PG, putting them on a separate CUDA stream from reduce-scatters so AG/RS overlap in backward. 
- Is a no-op when the graph contains no FSDP all-gathers.

w/o AG RS overlap:
Screenshot 2026-04-29 at 1 30 22 PM

w/ AG RS overlap:
Screenshot 2026-04-29 at 1 31 20 PM

### DIFF
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -64,6 +64,7 @@ from torchtitan.experiments.graph_trainer.remove_noop_passes import (
 )
 from torchtitan.experiments.graph_trainer.reshard_after_forward import (
     annotate_fsdp_all_gather,
+    is_wait_tensor_from_fsdp,
 )
 from torchtitan.tools.logging import logger
 
@@ -141,7 +142,13 @@ def compile_time_passes(
     they run at trace time via ``construct_default_graph_passes``.
 
     cudagraph is excluded because it needs to re-capture the graph into
-    an in-memory CUDA graph at runtime
+    an in-memory CUDA graph at runtime.
+
+    ``overlap_fsdp_ag_rs_pass`` runs immediately before
+    ``joint_transformer_block_bucketing_reordering_pass`` so that
+    forward+backward all-gathers end up on a separate CUDA stream from
+    reduce-scatters (enabling AG/RS overlap in backward). It is a no-op
+    when the graph contains no FSDP all-gathers.
     """
     from torchtitan.experiments.graph_trainer.common_utils import (
         get_default_transformer_block_buckets,
@@ -162,6 +169,7 @@ def compile_time_passes(
         # same time. Composability between these two passes is untested.
         apply_cpu_offload_pass,
         selective_activation_remat_pass,
+        overlap_fsdp_ag_rs_pass,
         functools.partial(
             joint_transformer_block_bucketing_reordering_pass,
             module_bucket_plans=get_default_transformer_block_buckets(n_layers),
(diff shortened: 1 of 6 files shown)


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -917,37 +925,71 @@ def _compiled_forward(*args):
...
+    )
+    return extra_pg.group_name
+
+
+def overlap_fsdp_ag_rs_pass(
     gm: torch.fx.GraphModule,
     example_inputs: tuple | None = None,
-    *,
-    source_pg_name: str,
-    target_pg_name: str,
```
SherlockNoMad, line 938: nice simplification of API

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -917,37 +925,71 @@ def _compiled_forward(*args):
...
+    This separates all-gathers from reduce-scatters onto different streams,
+    enabling AG/RS overlap in backward.
 
-    Args:
-        gm: The graph module (forward or backward)
-        example_inputs: Example inputs (unused, required by pass interface)
-        source_pg_name: The group_name of the process group to match
-        target_pg_name: The group_name of the process group to assign
+    No-op when the graph has no FSDP all-gathers. Must be applied BEFORE
+    bucketing passes so bucketed all-gathers inherit the new PG name.
```
SherlockNoMad, line 986: Wondering if this should just be an inner step of joint_transformer_block_bucketing_reordering_pass

Can you think of reasons that ”it‘s bad to do such coupling?"

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -61,6 +61,7 @@
 )
 from torchtitan.experiments.graph_trainer.reshard_after_forward import (
     annotate_fsdp_all_gather,
```
SherlockNoMad, line 66: side q, do we still need fsdp_reshard_after_fwd_pass, now that we have get_fsdp_reshard_after_forward_policy

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -917,37 +925,71 @@ def _compiled_forward(*args):
     return gm
 
 
-def reassign_to_pg_pass(
+# Maps an FSDP group_name to an extra group_name created by this pass.
+# Each NCCL PG gets its own CUDA stream, so the extra PG is what enables
+# AG/RS overlap in backward.
+_EXTRA_FSDP_PG_REGISTRY: dict[str, str] = {}
```
SherlockNoMad, line 944: as a follow up, I think we should split fsdp related passes into its own file. 
same for test. 

passes.py is getting crowded.

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/passes.py
yiming0416: I originally thought about putting it to be an inner step. 

But we may have other FSDP bucketing passes other than the existing `joint_transformer_block_bucketing_reordering_pass` (e.g. auto bucketing). And this `overlap_fsdp_ag_rs_pass` should be a general pass before any kind of bucketing pass.

torchtitan/experiments/graph_trainer/passes.py
yiming0416: These are two different things. 

`fsdp_reshard_after_fwd_pass` is still used in deprecated `aot` and `jit` paths. When `get_fsdp_reshard_after_forward_policy` returns True, we apply `fsdp_reshard_after_fwd_pass` in `aot` and `jit` paths

weifengpy:
I discussed with Andrew on using separte PG for AG and RS: https://github.com/pytorch/pytorch/pull/177015#issuecomment-4032932307

the hypothesis is, AG and RS fundmentally share the same network resource. overlapping AG and RS is up to benchmarking on a bigger scale

it's like allocating compute kernels to multiple cuda streams, - it's not necessarily better if kernels use the same resource

PR #3158, opened 2026-04-29
[CI] Move lychee link checker to nightly workflow

fegin:
* __->__ #3158

The lychee link checker is flaky due to transient external network failures, causing spurious PR CI failures. Move it out of per-PR lint into a dedicated nightly workflow so flakes no longer block PRs but broken links still get surfaced.

### DIFF
--- a/.github/workflows/link_check.yaml
+++ b/.github/workflows/link_check.yaml
@@ -1,0 +1,56 @@
+name: Link Check
+
+on:
+  schedule:
+    # Run nightly at 07:00 UTC
+    - cron: '0 7 * * *'
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash -l -eo pipefail {0}
+
+jobs:
+  link-check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.10']
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v6
+
+      - name: Setup python
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Update pip
+        run: python -m pip install --upgrade pip
+
+      - name: Install Lychee (Latest Release)
+        run: |
+          # Query GitHub API for the latest release tag
+          LATEST_TAG=$(curl -s https://api.github.com/repos/lycheeverse/lychee/releases/latest | jq -r .tag_name)
+          echo "Installing Lychee version: $LATEST_TAG"
+
+          URL="https://github.com/lycheeverse/lychee/releases/download/${LATEST_TAG}/lychee-x86_64-unknown-linux-gnu.tar.gz"
+
+          # Load the archive, extract into temp folder, move the binary, delete the rest
+          mkdir -p lychee_install
+          curl -sLO "$URL"
+          tar -xzf "lychee-x86_64-unknown-linux-gnu.tar.gz" -C lychee_install --strip-components=1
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -59,7 +41,3 @@ jobs:
 
       - name: Lint modified files
         run: pre-commit run --show-diff-on-failure --files ${{ steps.changed-files.outputs.all_changed_files }}
```
tianyu-l, line 67: I think this also requires lychee check, just not checking all-files / external links?
https://github.com/pytorch/torchtitan/blob/main/.pre-commit-config.yaml#L77

So I don't think we can remove the install in this yaml.

.github/workflows/link_check.yaml
tianyu-l, line 1: wonder how team could get signal from this nightly run. If not, I'd rather delete it. Instead we can get fix when people occasionally run `pre-commit run --all-files`

fegin: I think people monitor it quite often? But if you don't think so, then we can delete it.

.github/workflows/lint.yaml
fegin: It will show lychee is not installed and skip.

fegin (commented):
I think the main issue is that lychee is quite flaky due to external network issues. I'm not sure how much signals lychee can give us for per-PR run if we only run on the modified files, which 99% of the time don't contain links.

.github/workflows/lint.yaml
tianyu-l: lychee can check both in-repo links and internet links, I think it's still reasonable to keep the internal link checking

tianyu-l:
@Andrei-Aksionov wdyt?

fegin:
I update the per-PR lychee to run only the local links. I still keep the full run nightly check. Let me know if you want to delete it.

tianyu-l (approved):
can give it a try

Andrei-Aksionov:
Oh, I see.
It looks like we're hitting transient GitHub availability issues. 
I’ve experienced this myself: sometimes when opening a GitHub link, I get a "Unicorn" error page instead of code 🙃.

To ensure consistent behavior between local environments and CI, and to prevent transient network failures from blocking the workflows, I suggest the following options:

### 1. Fail only on 404 (Not Found)

We can configure Lychee to accept most HTTP error codes except for `404`. This ensures that if an external resource is temporarily offline (e.g., a `502` or `503` error), it won't fail the pre-commit check. The downside is that a legitimate `404` could theoretically be masked by a `502` during a check, but I believe the trade-off for CI stability is worth it.

### 2. Implement Lychee Caching

Lychee supports caching results in a simple CSV-formatted file (`URL, Status_Code, Timestamp`). This provides several benefits:
*   **Resilience:** If a URL was reachable previously but is currently down, Lychee can rely on the cached status.
*   **Self-Healing:** If a real `404` is hidden behind a `502`, Lychee will eventually reach the URL once it's back online and update the cache accordingly.
*   **Performance:** It significantly speeds up link checking by skipping recently verified URLs.
*   **CI Integration:** We can configure GitHub Actions to share and update a "global" cache across all PRs and runs.

### 3. Maintain a "Nightly" Action

Even if we decide to fail only on `404` for standard PRs, we should still monitor other errors like `502` or `503`. A nightly action is perfect for this:
*   We can set a much higher number of retries (e.g., retrying for up to 30 minutes) to account for longer outages. 
*   Once the nightly run successfully verifies all links, it will refresh the Lychee cache, which all other workflows can then reuse for the next day.

PR #3162, opened 2026-04-29
[graph_trainer] Re-enable ChunkedCELoss via compute_traceable_grads

tugsbayasgalan:
graph_trainer's make_fx flow uses torch.autograd.grad to capture gradients as explicit graph outputs, which is incompatible with eager ChunkedCELoss: chunk_loss.backward() inside __call__ consumes lm_head parameters via in-place .grad writes, and the _DecoderOutputGradientBackProp autograd Function only emits a gradient for hidden_states. autograd.grad on the loss can't reach lm_head and raises "One of the differentiated Tensors appears to not have been used in the graph."

The previous workaround (#3115) sidestepped this by swapping ChunkedCELoss for CrossEntropyLoss inside to_graph_trainer_config. Revert that and put the chunked path on a single source of truth instead.

Review order:

  1. torchtitan/components/loss.py adds compute_traceable_grads on ChunkedCELoss, which expresses the chunked forward + per-chunk backward via either chunk_loss.backward() (mode="backward", the default eager path used by __call__) or torch.autograd.grad (mode="autograd_grad", used by graph_trainer). Both modes share _compute_chunked_lm_head for the per-chunk loop, so __call__ and compute_traceable_grads run identical chunked-CE math; only the backward dispatch differs. accumulated_h_grad is re-wrapped with hidden_states.placements (Replicate-labeled, mirroring the fp32 GradAccumulator path) so the redistribute backward downstream does not all-reduce the seed under TP.

  2. torchtitan/experiments/graph_trainer/trainer.py wires ChunkedCELoss into make_fwd_bwd_step: when loss_fn is a ChunkedCELoss it partitions params into lm_head/decoder by id(), calls compute_traceable_grads(mode="autograd_grad"), and reorders grads back into named_parameters() order. Other losses keep the existing single autograd.grad(loss, params) path.

     compute_traceable_grads(mode="autograd_grad") includes a /tp_size compensation on the grad_outputs seed: under TP, autograd.grad's redistribute backward over-counts a Replicate seed by tp_size relative to the same op invoked via loss.backward(). Dividing here cancels the factor and produces grads that match eager chunked CE numerics under FSDP-only and converge similarly under FSDP+TP.

  3. torchtitan/experiments/graph_trainer/passes.py forces MUST_SAVE on non-layer nodes in apply_sac_pass. Without this, every per- chunk lm_head + ce_loss backward region is tagged for recompute, and the upstream remat pass (forward-loss-backward mode) only supports a single recompute region. Confining remat to the decoder also matches eager SAC, which checkpoints decoder layers only.

  4. The graph_trainer config wrappers and test_numerics baselines drop the "*_ce_loss" workaround variants so the chunked-CE loss_compare runs apples-to-apples against eager chunked CE.

  5. New llama3_8b_flex_attn / graph_trainer_llama3_8b_flex_attn configs route 8B through flex_attention; the default cuDNN SDPA backend currently fails ("No valid execution plans built") on this nightly torch + CUDA combo.

Verification:

  - debugmodel_flex_attn (FSDP=4 x TP=2, deterministic seed=42): eager __call__ refactored via _compute_chunked_lm_head matches the upstream eager __call__ bitwise (5/5 steps, all 57 params). compute_traceable_grads(mode="backward") matches __call__ bitwise on the same setup.
  - 8B flex_attn (FSDP=4 x TP=2, deterministic seed=42): make_fx- traced compute_traceable_grads(mode="autograd_grad") and the plain-Python execution of the same algorithm match bitwise on all 291 parameters at step 1, confirming make_fx itself introduces no numerical drift.

This commit was authored with Claude.

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -7,7 +7,7 @@
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import TypeAlias
+from typing import Literal, TypeAlias
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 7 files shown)


tugsbayasgalan:
@claude Give me review. Did I preserve the eager trainer behaviour?

tugsbayasgalan:
This is still a draft, just turning it non-draft to run the H100 tests

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -31,20 +32,53 @@ def make_fwd_bwd_step(loss_fn):
...
     ``loss_fn`` is captured in the closure so it is not a graph input.
+
+    For :class:`ChunkedCELoss` the per-chunk forward and backward live in
+    ``ChunkedCELoss.compute_traceable_grads`` so the chunking + TP redistribute
+    are shared with the eager ``__call__`` path; this wrapper only handles
+    parameter partitioning, the decoder-only path's ``torch.autograd.grad``,
+    and reordering grads back into ``named_parameters`` order.
     """
 
     def fwd_bwd_step(
```
SherlockNoMad, line 43: seems natural to just have a fwd_chunkloss_bwd_step

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -31,20 +32,53 @@ def make_fwd_bwd_step(loss_fn):
...
             p
             for _, p in model.named_parameters(remove_duplicate=False)
             if p.requires_grad
         ]
-        grads = torch.autograd.grad(loss, params)
-        return [loss] + list(grads)
+
+        if isinstance(loss_fn, ChunkedCELoss):
+            lm_head_param_ids = {
+                id(p) for p in loss_fn.lm_head.parameters() if p.requires_grad
```
SherlockNoMad, line 57: The impl is intrusive here, coz loss_fn break tracer's assumption that it's stateless. 

Need to revisit this assumption.

torchtitan/components/loss.py
```diff
@@ -245,71 +245,105 @@ def set_lm_head(self, lm_head: nn.Module) -> None:
...
+        """Shared chunked forward + per-chunk lm_head backward.
+
+        Returns ``(total_loss, accumulated_h_grad, lm_head_grads, hidden_states)``.
+
+        - ``total_loss``: scaled scalar loss.
+        - ``accumulated_h_grad``: per-chunk ``d(chunk_loss)/d(h_chunk)`` cat'd
+          along the seq axis — the seed for the decoder backward.
+        - ``lm_head_grads``: explicit grads for ``lm_head_params`` (for
+          ``mode="autograd_grad"``); ``None`` for ``mode="backward"`` because
+          they live on ``leaf.grad`` after the eager backward.
```
SherlockNoMad, line 318: why we need two modes? 

can eager also use autograd_grad path? sounds like you are hinting no... but it's not explicitly called-out why? 

is it implementation limitation? or something foundametal?

PR #3164, opened 2026-04-29
[graph_trainer] Generalize minimal_fx_tracer to module + optimizer roots

tugsbayasgalan:
Reshapes the public API so a single entry point handles three cases — plain stateless functions, module-rooted train steps, and full module+optimizer train steps:

    minimal_fx_tracer(fn)(*args)
    minimal_fx_tracer(fn, module=m)(*args)
    minimal_fx_tracer(fn, module=m, optimizer=opt)(*args)

`fn` no longer takes an explicit `state` argument; module parameters and `optimizer.state_dict()` are obtained from the live objects and threaded into the graph as inputs, with `stateless._reparametrize_module` and a new `swap_in_optim_state` context wrapping the user fn during tracing. The optimizer must already be initialized before tracing (so its lazy per-parameter state has been populated). `run_traced` mirrors the same arguments as `minimal_fx_tracer`.

`trace_train_step` and `run_traced_train_step` are removed — the new form subsumes both. 

Authored with Claude.

### DIFF
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import contextlib
 import copy
 from collections.abc import Callable, Generator
 from contextlib import contextmanager, nullcontext
@@ -272,16 +273,59 @@ def extract_module_state(mod: nn.Module) -> dict[str, torch.Tensor]:
     }
 
 
-def _check_no_extra_module_in_user_inputs(
-    args: tuple, kwargs: dict, fn_name: str
+def extract_train_state(
+    module: nn.Module | None = None,
+    optimizer: "torch.optim.Optimizer | None" = None,
+) -> tuple[dict[str, torch.Tensor], dict[str, Any]]:
+    """Return ``(model_state, optim_state)`` for ``minimal_fx_tracer``.
+
+    Both are dicts (empty when the corresponding object is ``None``) and are
+    sampled from the live module/optimizer, so callers can reuse this helper
+    to refresh state at runtime.
+    """
+    model_state = extract_module_state(module) if module is not None else {}
+    optim_state = optimizer.state_dict() if optimizer is not None else {}
+    return model_state, optim_state
+
+
+def _check_optimizer_has_module(
+    module: nn.Module | None,
+    optimizer: "torch.optim.Optimizer | None",
 ) -> None:
-    """Reject extra nn.Module instances in user args/kwargs (args[0] is the
-    module-rooted entry point)."""
-    if any(isinstance(v, nn.Module) for v in (*args, *kwargs.values())):
+    """Optimizer parameters align with module.named_parameters() by order, so
+    requiring a module guarantees that alignment."""
+    if optimizer is not None and module is None:
         raise ValueError(
-            f"{fn_name} supports exactly one nn.Module at args[0]. "
-            "Additional nn.Module instances found in args[1:] or kwargs."
+            "minimal_fx_tracer: when 'optimizer' is provided, 'module' must also "
+            "be provided so optimizer parameters align with the module's parameters."
         )
+
+
+@contextlib.contextmanager
+def _reparametrize_train_state(
+    module: nn.Module | None,
+    optimizer: "torch.optim.Optimizer | None",
+    model_state: dict[str, torch.Tensor],
+    optim_state: dict[str, Any],
+):
+    """Reparametrize module and optimizer with explicit tensor state for tracing."""
+    with contextlib.ExitStack() as stack:
+        if module is not None:
+            stack.enter_context(stateless._reparametrize_module(module, model_state))
+        if optimizer is not None:
+            # swap_in_optimizer_params_and_state aligns swapin_parameters values
+            # with optimizer.param_groups by order, so pass only parameters (in
+            # module.named_parameters() order).
+            params_for_optim = {
+                name: model_state[name]
+                for name, _ in module.named_parameters(remove_duplicate=False)
+            }
+            stack.enter_context(
+                torch.optim.swap_in_optimizer_params_and_state(
+                    optimizer, params_for_optim, optim_state
+                )
+            )
+        yield
 
 
 @dataclass
@@ -297,7 +341,7 @@ class TracedResult:
         num_flat_outputs: Number of flat graph outputs before subclass rewrapping.
         output_subclass_layouts: Subclass unwrap/rewrap metadata for outputs.
         output_spec: Original output pytree spec used during reconstruction.
-        state_fqns: Trace-time state keys.
+        state_fqns: Trace-time module parameter/buffer FQNs.
     """
 
     gm: torch.fx.GraphModule
@@ -322,8 +366,13 @@ class TracedResult:
         """Number of leading graph inputs with stable tensor addresses.
 
         Parameters and buffers (the state entries) have fixed addresses across
-        training steps.  Each may expand to multiple plain tensors after
+        training steps. Each may expand to multiple plain tensors after
         subclass unwrapping (e.g. DTensor -> inner tensors).
+
+        TODO: graph_trainer does not trace optimizers yet, so optimizer state
+        tensors are not counted here. When optimizer tracing is enabled, they
+        should be included since their addresses are also stable across steps,
+        avoiding cudagraph re-copying them every step.
         """
         num_state = len(self.state_fqns)
         return sum(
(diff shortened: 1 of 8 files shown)


torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -302,30 +516,66 @@ def num_static_inputs(self) -> int:
         )
 
 
-def minimal_fx_tracer(fn: Callable) -> Callable[..., TracedResult]:
-    """Return a tracer for a stateless ``fn`` with explicit ``state`` input.
+def minimal_fx_tracer(
+    fn: Callable,
+    module: nn.Module | None = None,
```
ydwu4, line 388: Using explicit module/optimizer references is OK but seems a bit restrictive. like what if we have a global constants/ module attributes that's non-buffer. Something more generic is like:
```python
def minimal_fx_tracer(
  fn: Callable, *, 
  implicit_states = pytree[Source]
): 
```

The idea is the implicit_states can be tracked a dynamo source like system to look the states up at tracing time, this allows us to codegen something like:
"implicit_states = self.module.token_per_expert_count" at runtime

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
ydwu4: we probably don't need the _state in  some of these tests.  They are empty dict anyway?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -530,7 +475,7 @@ def forward(state, tokens, *, scale):
         state = extract_module_state(model)
         traced = minimal_fx_tracer(forward)(state, tokens, scale=scale)
         with self.assertRaisesRegex(ValueError, "input spec mismatch"):
-            run_traced(traced, state, tokens, _validate_runtime=True, factor=scale)
+            run_traced(traced, _strict_check_inputs=True)(state, tokens, factor=scale)
```
ydwu4: nit: it would be better if we could do one thing per PR. the _validate_runtime -> _strict_check_inputs change probably worth a new pr.  Mixing changes make it harder to review.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -648,19 +593,92 @@ def forward(_state, q, k, v, block_mask):
...
         other_model = SimpleMLP().to(device=self.DEVICE, dtype=self.DTYPE)
         tokens = torch.randint(0, 256, (2, 32), device=self.DEVICE)
 
-        with self.assertRaises(ValueError, msg="Additional nn.Module"):
-            trace_train_step(forward)(model, other_model, tokens)
+        def forward(other, tokens):
+            return other(tokens)
+
+        with self.assertRaises(ValueError, msg="nn.Module"):
+            minimal_fx_tracer(forward, module=model)(other_model, tokens)
```
ydwu4, line 652: This might be OK for current PR but I think we shouldn't force no nn module input in args. it's quite common to trace a module forward. Can handle them in a similar way as "module=" I feel?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -648,19 +593,92 @@ def forward(_state, q, k, v, block_mask):
...
+        model = SimpleMLP().to(device=self.DEVICE, dtype=self.DTYPE)
+        container = OptimizersContainer(
+            OptimizersContainer.Config(
+                name="AdamW", lr=1e-3, implementation="for-loop"
+            ),
+            model_parts=[model],
+        )
+        inner = container.optimizers[0]
+
+        # Initialize Adam's lazy per-parameter state.
```
ydwu4, line 682: nit: Comments that doesn't carry meaningful infos makes it harder to review tbh.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -648,19 +593,92 @@ def forward(_state, q, k, v, block_mask):
...
+        optim_state_dict = inner.state_dict()
+        original_param_ids = [[id(p) for p in g["params"]] for g in inner.param_groups]
+        original_state_keys = list(inner.state.keys())
+        original_state_id = id(inner.state)
+
+        # Rebind to fake parameter tensors (zeros to make the swap obvious).
+        params_dict = dict(model.named_parameters(remove_duplicate=False))
+        fake_params = {name: torch.zeros_like(p) for name, p in params_dict.items()}
+
+        with _reparametrize_optimizer(inner, fake_params, optim_state_dict):
```
ydwu4: The _reparametrize_optimizer changes worth a separate PR

torchtitan/experiments/graph_trainer/precompile.py
```diff
@@ -309,7 +309,7 @@ class PrecompiledFxTraceArtifact:
     """
 
     serialized_gm: bytes
-    state_fqns: list[str]
+    module_state_fqns: list[str]
```
ydwu4: the renaming needs a separate PR.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -34,15 +34,15 @@
 from torchtitan.trainer import Trainer
 
 
-def make_fwd_bwd_step(loss_fn):
+def make_fwd_bwd_step(model, loss_fn):
```
ydwu4, line 39: The calling convention change of "module=" worth a separate PR since it has nothing to do with optimizer support.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -366,19 +619,29 @@ def minimal_fx_tracer(
...
     """
+    _check_optimizer_has_module(module, optimizer)
+
+    def _trace_with_args(*args: Any, **kwargs: Any) -> TracedResult:
+        model_state, optim_state = extract_train_state(module, optimizer)
+        module_state_fqns = list(model_state.keys())
+
+        # Skip optim_state when None: pytree would emit it as a leaf, which
+        # cudagraph rejects as a non-tensor input.
+        state_tree: Any = (
```
ydwu4: why do the branching? If it's None, we could just keep it as None. Later on when we unpacking it, we'll get None back automatically.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -336,25 +578,36 @@ def num_static_inputs(self) -> int:
 
 def minimal_fx_tracer(
     fn: Callable,
+    module: nn.Module | None = None,
```
ydwu4, line 388: Don't need to happen in this PR. We should probably support a pytree of modules.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: Yeah sure. I thought this name would be more clear haha.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: Yeah... but I feel we should still respect pytree input/output contract as long as we can.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: sure! It is important that we are initializing tho, i will just very very briefly mention it.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: Yep this is not needed. I am still waiting for pytorch nightly to pick up the change.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: This is because pytree treats None as leaf as a result we end up having one extra None argument in the make_fx graph

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: Yeah if we have use case, we should do it.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: yeah

torchtitan/experiments/graph_trainer/trainer.py
tugsbayasgalan: This function returns a callable fwd_bwd_step which can't take nn.Module as input argument now. Therefore we need to make model closure accessible to fwd_bwd_step

SherlockNoMad:
@ydwu4 can you help review this in details?

ydwu4 (approved):

xmfan (approved):

PR #3167, opened 2026-04-29
[MoE][4/n] deprecate expert tensor parallel (ETP)

acisseJZhong:
* __->__ #3167
* * __->__ #3125

Stacked on https://github.com/pytorch/torchtitan/pull/3125

## Summary

Removes expert tensor parallelism (ETP). With the trend toward fine-grained experts, sharding already-skinny expert weights with TP is impractical — the shapes become too small for efficient GEMMs. Additionally, when ETP>1 all TP ranks within the same EP group perform redundant all-to-all communication, wasting bandwidth (see [discussion](https://github.com/pytorch/torchtitan/pull/2842#discussion_r3066966471)). 

This removes the `ExpertTensorParallel` classes, the `etp` mesh axis, the `expert_tensor_parallel_degree` / `expert_parallel_comm_backend` config fields, and all ETP references across models and experiments.

## Test plan
- [ ] `pre-commit run --all-files` passes
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 1 --parallelism.expert_parallel_degree 2" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 1 --parallelism.expert_parallel_degree 2 --parallelism.expert_tensor_parallel_degree 1" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```


tests/integration_tests/models.py
tianyu-l: unrelated to this PR: this flag doesn't make sense, as there's no TP axis

tianyu-l: let's remove this test, as we don't aim to support dualpipe overlapping

torchtitan/config/configs.py
```diff
@@ -221,38 +221,10 @@ def __post_init__(self):
     """
     Expert parallelism degree. 1 means disabled. No effect for non-MoE models.
 
-    Currently, etp is either 1 or is the same as tp.
-
     Note that this is still an experimental feature. Some constraints will be
     relaxed soon when we have more flexible DeviceMesh support.
```
tianyu-l: remove this comment

torchtitan/config/configs.py
```diff
@@ -221,38 +221,10 @@ def __post_init__(self):
     """
     Expert parallelism degree. 1 means disabled. No effect for non-MoE models.
 
-    Currently, etp is either 1 or is the same as tp.
```
tianyu-l, line 224: let's add notes on the constraint, about
dense dp_shard x cp x tp == sparse efsdp (implicit) x ep

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -53,28 +53,11 @@ def parallelize_deepseekv3(
         model.parallelize(tp_mesh)
         maybe_enable_async_tp(parallelism, compile_config, tp_mesh)
 
-    # Check if using DeepEP/HybridEP for MoE communication
-    comm_backend = parallelism.expert_parallel_comm_backend
-    if comm_backend in ("deepep", "hybridep"):
-        if not parallel_dims.ep_enabled:
-            raise ValueError(
```
tianyu-l, line 60: we still need to error out right? Not sure if it was added in previous PR, but if not, we should error out in `update_from_configs` of MoE models model.py. It is where you get both model config and CLI override of parallelism configs.

acisseJZhong: I think it's safe to delete this assertion. If EP=1 and we are using EP dispatcher, it will error out here as EP group is not set. https://github.com/pytorch/torchtitan/blob/01a8b06843d33de428d807ddf9207a10273c7b8e/torchtitan/models/common/token_dispatcher.py#L603

tests/integration_tests/models.py
acisseJZhong: sounds good, I will just remove this flag for this EP test.

torchtitan/models/deepseek_v3/parallelize.py
tianyu-l: These are two levels of assertion. The one here should be ValueError (user mistake), that one should be assertion (system mistake). So we should still error out in config (which is cheaper).

acisseJZhong: yeah that makes sense. We should def error out invalid config before run starts.

torchtitan/models/deepseek_v3/model.py
```diff
@@ -223,6 +223,19 @@ def update_from_config(
                     layer_cfg.moe.router._debug_force_load_balance = (
                         debug.moe_force_load_balance
                     )
+                    comm_backend = getattr(
```
tianyu-l, line 226: Curious is there's any specific reason gpt-oss doesn't support deepep / hybridep. If not maybe it's low hanging fruit to enable.

tianyu-l (approved):

torchtitan/models/deepseek_v3/model.py
acisseJZhong: not very sure why there's no deepep/hybridep support. @wwwjn @shuhuayu in case you have more context. I can add one if no technical gaps.

PR #3176, opened 2026-04-30
Observability: structured logging + training instrumentation (#3176)

### LINKED ISSUE #3152: PG timeouts never adapted if state dict is loaded
### Bug description

On job resumptions, if a checkpoint is loaded and the `state_dict` of `Trainer` is loaded (sets `self.step`, c.f. https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L860), `self.step` in `Trainer` is  **not equal to 1** after the first training step. Thus the timeouts are never reduced by this line https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L841

We propose a fix with a local boolean variable (like `sharper_pg_timeouts_applied = False`) in `Trainer.train()`, so the change of the PG timeouts is not conditional on `self.step`, but on a local variable `sharper_pg_timeouts_applied`. Let me know if I should PR this solution.

### Versions

Current main branch.

felipemello1:
Summary:

**TLDR**: Enables structured-logging in torchtitan. Time spans, scalars and events can be logged, per rank, with metadata, to a jsonl or database, using python standard logger. This can then be converted into a gantt chart. It is cheap, runs on every step, and useful to find stragglers, high level bottlenecks, how asynchronous an RL run is, debug timeouts, weird behavior on some rank on some step, etc.

**For details of the APIs, please read the added readme in** torchtitan/observability/structured_loggger/README.md.

Examples:
 {F1989149827} 

## Other comments

To minimize the impact of lines changed in the trainer.py code:
- We use decorators where we can, instead of context managers
- We call tags and spans directly inside of some functions, e.g. ckpt, garbage collectors and profiler 
- **initialization**: We add it one time in the build call base class, so everything that calls build is automatically tracked.

## Is it safe?

The workhorse here is logger.info. So the same way you treat logger.info, you should treat the structured logger. 

It can be disabled with a global flag.

If the user forgets to setup `init_structured_logger`, we still call logger.info, but nothing is saved anywhere.

It detects `is_compiling` and skips if true.

## Profiler default paths

To align with expectations set by some internal tooling, we also changed the kineto and memory profilers default path.

Differential Revision: D101049878

### DIFF
--- a/torchtitan/components/checkpoint.py
+++ b/torchtitan/components/checkpoint.py
@@ -37,14 +37,15 @@ from torch.distributed.checkpoint.state_dict_saver import (
     AsyncSaveResponse,
 )
 from torch.distributed.checkpoint.stateful import Stateful
+
 from torchtitan.components.dataloader import BaseDataLoader
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
 from torchtitan.config import Configurable, TORCH_DTYPE_MAP
+from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
 from torchtitan.tools.logging import logger
 from torchtitan.tools.utils import GarbageCollection
-
 
 MODEL = "model"
 OPTIMIZER = "optimizer"
@@ -550,8 +551,9 @@ class CheckpointManager(Configurable):
             if MODEL in self.states:
                 self.states[MODEL].load_state_dict(state_dict)
 
+    @sl.log_trace_span("checkpoint_save")
     @torch.no_grad()
-    def save(self, curr_step: int, last_step: bool = False) -> None:
+    def save(self, curr_step: int, last_step: bool = False) -> bool:
         """Save the checkpoint for the current step.
 
         This function will save the checkpoint for the current step. If ``last_step`` is
@@ -564,11 +566,13 @@ class CheckpointManager(Configurable):
             last_step (bool, optional): Whether this is the last step of training.
 
         Returns:
-            None
+            bool: True if a checkpoint was written (or staged, for async modes) on
+            this step.
         """
         if not self._should_save(curr_step, last_step):
-            return
+            return False
 
+        sl.add_step_tag("checkpoint_save")
         begin = time.monotonic()
         logger.info("Saving the checkpoint (or staging if async is enabled).")
         checkpoint_id = self._create_checkpoint_id(curr_step)
@@ -578,7 +582,7 @@ class CheckpointManager(Configurable):
         # freed until _async_wait()
         if last_step:
             self._save_last_step(curr_step)
-            return
+            return True
 
         states = self._flattened_model_states_sd()
         if self.async_mode == AsyncMode.ASYNC_WITH_PINNED_MEM:
@@ -613,7 +617,9 @@ class CheckpointManager(Configurable):
             "Finished saving the checkpoint (or staging if async is enabled) "
             f"in {time.monotonic() - begin:.2f} seconds."
         )
+        return True
 
+    @sl.log_trace_span("checkpoint_load")
     @torch.no_grad()
     def load(self, step: int = -1) -> bool:
         """Load the checkpoint for the given step.
--- a/torchtitan/config/configurable.py
+++ b/torchtitan/config/configurable.py
@@ -10,6 +10,8 @@ from collections.abc import Iterator
 from dataclasses import dataclass, fields, replace
 from typing import ClassVar
 
+from torchtitan.observability import structured_logger as sl
+
 logger = logging.getLogger(__name__)
 
 
@@ -114,17 +116,18 @@ class Configurable:
                     f"{type(self).__name__} has no owner class. "
                     "Define Config inside a Configurable subclass."
                 )
-            if not kwargs:
-                return self._owner(config=replace(self))
+            with sl.log_trace_span(f"{type(self).__qualname__}.build"):
+                if not kwargs:
+                    return self._owner(config=replace(self))
 
-            config_fields = {f.name for f in fields(self)}
-            overlap = config_fields & kwargs.keys()
-            if overlap:
-                raise ValueError(
-                    f"build() kwargs {overlap} overlap with config fields. "
-                    "Put these values in the Config, not in build() kwargs."
-                )
-            return self._owner(config=replace(self), **kwargs)
+                config_fields = {f.name for f in fields(self)}
+                overlap = config_fields & kwargs.keys()
+                if overlap:
+                    raise ValueError(
+                        f"build() kwargs {overlap} overlap with config fields. "
+                        "Put these values in the Config, not in build() kwargs."
+                    )
+                return self._owner(config=replace(self), **kwargs)
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
--- a/torchtitan/observability/structured_logger/structured_logging.py
+++ b/torchtitan/observability/structured_logger/structured_logging.py
@@ -1,0 +1,438 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Structured-logging API: init_structured_logger, log_trace_span, log_trace_instant, log_trace_scalar.
+
+Emits structured JSONL events for phase timing, scalars, and diagnostics.
+Handler factories (JSONL, custom, etc.) are loaded dynamically via
+the ``TITAN_STRUCT_LOGGER_HANDLERS`` env var.
+"""
+
+import asyncio
+import enum
+import functools
+import importlib
+import inspect
+import logging
+import os
+from collections.abc import Callable
+from timeit import default_timer as timer
+from typing import Any, cast, TypeVar
+
+import torch
+
+from torchtitan.observability.structured_logger.step_state import get_step
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+# Used by this module for regular logging
+console_logger: logging.Logger = logging.getLogger(__name__)
+
+# Dedicated logger for structured events. Uses a name distinct from
+# ``__name__`` so we can set ``propagate = False`` here without also silencing
+# the console logger above.
+_structured_logger: logging.Logger = logging.getLogger("torchtitan.structured_logger")
+_structured_logger.propagate = False
+
+# Used to check if handler has been already initialized. If so, re-initializing
+# is a no-op
+_is_initialized: bool = False
+
+# Set by ``init_structured_logger(enable=False)`` to make all trace calls no-ops.
+_disabled: bool = False
+
+_DEFAULT_HANDLER_FACTORY = (
+    "torchtitan.observability.structured_logger.jsonl_handler.register_jsonl_handler"
+)
+
+
+def _structured_logger_disabled() -> bool:
+    """Whether structured logging is disabled.
+
+    Driven by the ``enable`` flag passed to :func:`init_structured_logger`
+    (sourced from ``DebugConfig.enable_structured_logging``).
+    """
+    return _disabled
+
+
+class StrEnum(enum.Enum):
+    """Stand-in for ``enum.StrEnum`` (added in Python 3.11)
+
+    Mimics it for our use case: ``str(member)`` returns the value (e.g.
+    ``"event"``), not ``"LogType.EVENT"``. Drop in favor of ``enum.StrEnum``
+    once Python 3.10 support is no longer needed.
+    """
+
+    def __str__(self) -> str:
+        return self.value
+
+
+class LogType(StrEnum):
+    """Record kind in the JSONL stream.
+
+    - ``EVENT``: paired span record (``*_start`` / ``*_end`` from ``log_trace_span``).
+    - ``INSTANT``: point-in-time record (``log_trace_instant``, ``log_trace_scalar``).
+    - ``TEXT``: free-text log record (filtered out by ``TraceEventsOnlyFilter``).
+    """
+
+    EVENT = "event"
+    INSTANT = "instant"
+    TEXT = "text"
+
+
+class ExtraFields(StrEnum):
+    """Keys for the ``extra`` dict passed to logging calls."""
+
+    LOG_TYPE = "log_type"
+    LOG_TYPE_NAME = "log_type_name"
+    EVENT_NAME = "event_name"
+    STEP = "step"
+    VALUE = "value"
+    RELATIVE_STEP = "relative_step"
+    TASK_NAME = "task_name"
+
+
+def event_extra(
+    event_type: str,
+    event_name: str | None = None,
+    step: int | None = None,
+    relative_step: int | None = None,
+    value: float | int | None = None,
+    task_name: str | None = None,
+    log_type: LogType = LogType.EVENT,
+) -> dict[str, Any]:
+    """Build the extra dict for a structured JSONL event record."""
+    return {
+        str(ExtraFields.LOG_TYPE): str(log_type),
+        str(ExtraFields.LOG_TYPE_NAME): str(event_type),
+        str(ExtraFields.EVENT_NAME): event_name,
+        str(ExtraFields.STEP): step,
+        str(ExtraFields.RELATIVE_STEP): relative_step,
+        str(ExtraFields.VALUE): value,
+        str(ExtraFields.TASK_NAME): task_name,
+    }
+
+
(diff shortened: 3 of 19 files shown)


rakkit:
questions: this merge `generate_gantt_trace("outputs/structured_logs/", "outputs/gantt.json")` seems gonna merge all files in this folder.  for use case, like we need to resume job multiple times (runs on  same dump folder) then the logs will come from differnt runs and it is werid if we merge them together.   


and do we by default expect this to runs on full ranks + every step? lts gonna be like O(1M) steps per 5T token on each rank.

felipemello1:
@rakkit, good questions!

>  like we need to resume job multiple times (runs on same dump folder) then the logs will come from differnt runs and it is werid if we merge them together.

If we are resuming, i assume it would be ok to have them together, no? But we could think about adding a run_id to the logs metadata, so users can filter by it. Another alternative is having the logs directed to a different dump_folder

>and do we by default expect this to runs on full ranks + every step? lts gonna be like O(1M) steps per 5T token on each rank.

We expect to have N rank files, yes. For the gantt, we could add flags so people can choose to only display N ranks, for example, or only the last N entries. But the main benefit here is being to query it with some database, e.g. if you have a nccl timeout, to find where/when it timedout, for example.

On the logger side, we could cycle the file, limiting to N max lines, or breaking into new files every N entries.

----

I thought that these refinements could be added as follow-up as users stress test it. Does it match your intuition?

rakkit:
>  i assume it would be ok to have them together. 

yes put them together is ok.  just feel weird, or at least we should allow filtering by runs.  on slurm clusters we usually get differnt nodes and usually when problem comes, we actually easily know and locate "which runs is borken" in that case checking the special runs log will be much easier and clean.

> I thought that these refinements could be added as follow-up as users stress test it

yeah it just we recently suffers a lot from scaling. everything on 1~2k gpu x GPFS is mess and i got PTSD on this

felipemello1:
> > i assume it would be ok to have them together.
> 
> yes put them together is ok. just feel weird, or at least we should allow filtering by runs. on slurm clusters we usually get differnt nodes and usually when problem comes, we actually easily know and locate "which runs is borken" in that case checking the special runs log will be much easier and clean.
> 
> > I thought that these refinements could be added as follow-up as users stress test it
> 
> yeah it just we recently suffers a lot from scaling. everything on 1~2k gpu x GPFS is mess and i got PTSD on this

makes sense! I can look into adding a 'run_id' metadata. But, as it is, it should be very easy for users to also add whatever handler/metadata they prefer (example in the readme)

>yeah it just we recently suffers a lot from scaling. everything on 1~2k gpu x GPFS is mess and i got PTSD on this

Makes sense! If/when you give it a try, let me know. I will think a bit more about remediating this in a follow up as well.

---

At first glance, do you see this type of logger being useful for your 1-2k gpu run?

rakkit:
> makes sense! I can look into adding a 'run_id' metadata. But, as it is, it should be very easy for users to also add whatever handler/metadata they prefer (example in the readme)

dumb solution is we take some magic hash or we  broadcast some ID. 

speaking of that, would be help if we can also log info like,   "i am [global-rank] 0,  [fsdp-rank] 0, [dp-rank] 0, [cp-rank] ....." etc from each parallel dims.


> At first glance, do you see this type of logger being useful for your 1-2k gpu run?

yes i think it gonna be help for both small and large scale for debug and found problem.    technically we could write an even short discussion of metadata structure and  ask codex/claude to vide code some magic view for diagnose.    for large scale the main concern is the log itsel, like it should not slow down training and be friendly to filesystem

felipemello1:
> would be help if we can also log info like, "i am [global-rank] 0, [fsdp-rank] 0, [dp-rank] 0, [cp-rank] ....." etc from each parallel dims.

We log the global rank. I think that this can be done in postprocessing if we can map rank to parallel dim.

>for large scale the main concern is the log itsel, like it should not slow down training and be friendly to filesystem

agreed. Internally I saw people doing it two ways:
1. Log to local host, and a side channel takes it and logs to database
2. Log directly to database

The jsonl handler is a naive way of doing it, saving directly to shared FS. For large scale, i assume one would want to change their handlers if writing to FS slows down the run. I would need to check if python's logger.info is blocking or not in this case, but thats basically all we are doing: logger.info with extra metadata.

torchtitan/components/checkpoint.py
```diff
@@ -564,56 +565,61 @@ def save(self, curr_step: int, last_step: bool = False) -> None:
...
-        # This GC is called for async checkpoint as it is useless to do
-        # GC right after async_save -- the CPU memory is not able to be
-        # freed until _async_wait()
-        if last_step:
-            self._save_last_step(curr_step)
-            return
+        sl.add_step_tag("checkpoint_save")
+        with sl.log_trace_span("checkpoint_save"):
+            begin = time.monotonic()
+            logger.info("Saving the checkpoint (or staging if async is enabled).")
```
fegin: Do we still need this if we have the event logger?

torchtitan/components/checkpoint.py
```diff
@@ -564,56 +565,61 @@ def save(self, curr_step: int, last_step: bool = False) -> None:
...
-        checkpoint_id = self._create_checkpoint_id(curr_step)
-        self._async_wait()
-        # This GC is called for async checkpoint as it is useless to do
-        # GC right after async_save -- the CPU memory is not able to be
-        # freed until _async_wait()
-        if last_step:
-            self._save_last_step(curr_step)
-            return
+        sl.add_step_tag("checkpoint_save")
+        with sl.log_trace_span("checkpoint_save"):
```
fegin: Is it bad that we actually use function decorator. I understand there will be a very small span for every step. Just don't know how bad if we actually do this.

fegin (commented):
Did a quick browse, mostly looks good to me as we already discussed internally. I have one high-level question, which I asked in the checkpointer too. If we have structured logging, should we trim some logger.info, which try to record the time span of that action?

Another comment, we don't have to do this in this PR. But I think we should review some code blocks. If the indentation is too deep and the block is too long, and it is actually put in a time span, this code block may deserve a helper function and use a decorator style, such as model init.

Finally, for @rakkit's scaling question, do you think we can add an option to only log in certain ranks? For example, sl gets the rank by using dist.get_rank() and only rank % X == 0 does the logging. For OSS users, distributed filesystem is sometimes expensive unlike that in Meta internally.

torchtitan/components/checkpoint.py
felipemello1: I was thinking that if someone later wants to filter "checkpoint_save" time, its easier to not worry about filtering no-ops. I think i could put the span decorator, but leave the tag inside.

felipemello1:
> If we have structured logging, should we trim some logger.info, which try to record the time span of that action?

our structured logger doesn't emit to console. We could change it, but it might be too spammy. I think that leaving the logger.info is helpful since the user can get a quick signal and checkpointing can take several seconds. Up to you.

> this code block may deserve a helper function and use a decorator style, such as model init.

Agreed. I can do it in a quick follow up.

felipemello1:
comment on CI: unrelated to this diff. Main is also broken.

fegin:
> our structured logger doesn't emit to console. We could change it, but it might be too spammy. I think that leaving the logger.info is helpful since the user can get a quick signal and checkpointing can take several seconds. Up to you.

Let's leave the logging then. I agree it will be spammy.

torchtitan/components/checkpoint.py
```diff
@@ -564,56 +565,61 @@ def save(self, curr_step: int, last_step: bool = False) -> None:
...
-            "Finished saving the checkpoint (or staging if async is enabled) "
-            f"in {time.monotonic() - begin:.2f} seconds."
-        )
+            logger.info(
+                "Finished saving the checkpoint (or staging if async is enabled) "
+                f"in {time.monotonic() - begin:.2f} seconds."
+            )
+        return True
 
+    @sl.log_trace_span("checkpoint_load")
```
tianyu-l, line 622: why making save and load asymmetric? using wrapper consistently sounds better

torchtitan/config/configurable.py
```diff
@@ -114,17 +116,18 @@ def build(self, **kwargs):
...
-
-            config_fields = {f.name for f in fields(self)}
-            overlap = config_fields & kwargs.keys()
-            if overlap:
-                raise ValueError(
-                    f"build() kwargs {overlap} overlap with config fields. "
-                    "Put these values in the Config, not in build() kwargs."
-                )
-            return self._owner(config=replace(self), **kwargs)
+            with sl.log_trace_span(f"{type(self).__qualname__}.build"):
```
tianyu-l, line 119: This will log every single module inside model, including every transformer layer / attention / wqkv / etc. I wonder how much overhead it brings. Also it feels unnecessary to be included in main branch. Would love to hear counter arguments.

torchtitan/tools/profiler.py
```diff
@@ -241,10 +259,13 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> bool:
 
     def step(self) -> None:
         """Advance all active profilers by one training step."""
+
         if self.torch_profiler is not None:
-            self.torch_profiler.step()
+            with sl.log_trace_span("kineto_profiler_step_call"):
```
tianyu-l, line 264: I think the step call itself is very cheap and not worth dedicated `log_trace_span`

torchtitan/tools/utils.py
```diff
@@ -55,15 +57,24 @@ def __init__(self, gc_freq: int = 1000, debug: bool = False):
...
+        collection actually ran, False otherwise."""
         if self.debug:
-            self.collect(
-                "Force GC to perform collection to obtain debug information",
-                generation=2,
-            )
-            gc.collect()
-        elif step_count > 1 and step_count % self.gc_freq == 0:
-            self.collect("Performing periodic GC collection")
+            with sl.log_trace_span("gc_collect"):
```
tianyu-l: curious why don't we add this as decorator? There are two styles
- we add `log_trace_span` regardless of performing the operation or not, and depend on `add_step_tag` to tell if it's really performed
- we add `log_trace_span` only when it's performed

Which one do you prefer? I think option 1 allows us to use decorator style more often.

torchtitan/observability/structured_logger/structured_logging.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+def _structured_logger_disabled() -> bool:
+    """Read ``TITAN_DISABLE_STRUCT_LOGGER`` on each call, not at import time.
+
+    Deferred so setting the env var after import (e.g. in a test
+    fixture) still takes effect, and so module import has no side
+    effect that depends on env state.
+    """
+    return os.environ.get("TITAN_DISABLE_STRUCT_LOGGER", "") == "1"
```
tianyu-l: torchtitan's convention is to prefer config over ENVVAR which is less cleaner in general.
Here I wouldn't say config can make it much cleaner than envvar, but from consistency perspective, is there any special justification for structure logging to use envvar? o/w we may put the switch in https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L345 and use a global var here.

torchtitan/trainer.py
```diff
@@ -816,34 +839,41 @@ def train(self):
...
+                        self.step
+                    ):
+                        self.validator.validate(self.model_parts, self.step)
+
+                    # signal the profiler that the next profiling step has started
+                    profiler.step()
+
+                    # reduce timeout after first train step for faster signal
+                    # (assuming lazy init and compilation are finished)
+                    if self.step == 1:
```
tianyu-l: could you change this to relative step which closes https://github.com/pytorch/torchtitan/issues/3152

torchtitan/trainer.py
```diff
@@ -729,7 +738,9 @@ def train_step(
 
         # Process each microbatch: move to GPU, forward/backward, then free
         accumulated_losses = []
-        for input_dict, labels in microbatches:
+        for microbatch_idx, (input_dict, labels) in enumerate(microbatches):
+            valid_tokens = local_valid_tokens_per_microbatch[microbatch_idx]
+            sl.log_trace_scalar({"local_valid_tokens": valid_tokens})
```
tianyu-l: This seems multiple logging on the same key, at the same step. Is it OK? How do you check the order?

torchtitan/trainer.py
```diff
@@ -712,11 +717,15 @@ def train_step(
...
+        local_valid_tokens_per_microbatch: list[int] = []
         local_valid_tokens = torch.tensor(0, dtype=torch.int64)
         for _microbatch in range(self.gradient_accumulation_steps):
-            input_dict, labels = next(data_iterator)
-            local_valid_tokens += (labels != IGNORE_INDEX).sum()
-            microbatches.append((input_dict, labels))
+            with sl.log_trace_span("fetching_batch"):
+                input_dict, labels = next(data_iterator)
+                microbatch_valid = (labels != IGNORE_INDEX).sum()
+                local_valid_tokens_per_microbatch.append(int(microbatch_valid.item()))
```
tianyu-l: nit: if it's not absolutely necessary, let's delay this fine-grained logging to later PRs when necessary. Logging `local_valid_tokens` over all microbatches seems less intrusive.

tianyu-l:
@claude review this PR

torchtitan/components/checkpoint.py
felipemello1: We call load once. We call save on every step. We can use the decorator, but then on every step we have this no-op `checkpoint_save`. Since both you and Chien-chin commented about it, I can just the decorator and only keep the tag inside, and users can filter by tag if they need to do some analysis.

torchtitan/config/configurable.py
felipemello1: > I wonder how much overhead it brings

It is a logger.info call. It didn't feel like a problem for llama8b (images below) -- check time scale

>Also it feels unnecessary to be included in main branch

llama4x and sixlib track multiple initializations and build tooling around it as well (the DAG critical path, for example). I think it is useful to find bottlenecks, if any, to reduce initialization time + log what is being initialized. For example, I remember an issue in Tune with RoPe initialization taking a long time.

Code overhead is also very small: an extra line in the configurable.py, vs multiple indented blocks in the trainer.py or in selected configs.

If there is ever a bug, its logged and claude can debug. It seems like a small cost for the benefits.

initialization section
image

zoomed in in the model build
image

torchtitan/tools/profiler.py
felipemello1: and thats why we need the logging :P, i added it after i saw a gap in the trace

First kineto step call
image

second one
image

torchtitan/tools/utils.py
felipemello1: sure, its the same as in the checkpointer. I will change.

torchtitan/observability/structured_logger/structured_logging.py
felipemello1: We could do something like this: 
```python
_ENABLED = True
def init_structured_logger(..., enabled=config.enable_structured_logger)
    global _ENABLED
    _ENABLE=enabled
```

The issue is that for RL, for example, we would have to pass this "enable" to every actor/subprocess. We already pass the output_dir, so it shouldn't be a problem, just one more arg.

torchtitan/trainer.py
felipemello1: It gets logged like a marker in the perfetto (image below). I ran it with multiple microbatches and it looked good (i dont have the profile now, but i can rerun it)

Order is not an issue, we also have a time stamp associated to it.

A query would be something like:
"Give me all 'local_valid_tokens' in step N ordered by timestamp"

image

torchtitan/trainer.py
felipemello1: sg, i will delete

torchtitan/observability/structured_logger/structured_logging.py
tianyu-l: oh I see. I have thought about moving output_dir to envvar for exactly this reason before. Let's just be consistent, I'm amenable to both.

felipemello1:
@tianyu-l , all of your comments/requests are addressed

**Changed:**                                                                                                             
 - ckpt save uses decorator                                                   
- GC uses decorator
- env var → config: replaced TITAN_DISABLE_STRUCT_LOGGER with DebugConfig.enable_structured_logging (default True).       
- relative step (closes #3152): if self.step == 1 → if self.step - loaded_step == 1, so the timeout reduction fires after the first step of this process even on resume.
- per microbatch local_valid_tokens: now logged once per step with the the total sum.

 **What i did **not** change:**
  - log_trace_span on all build() calls
  - log_trace_span on profiler step() 
  - per-microbatch fetching_batch span — kept the span, but removed the per-microbatch scalar log (not sure if you also wanted to remove the per-microbatch span)

I checked claude's review, and theres nothing substantial to add.

Let me know if that matches what you wanted

tianyu-l (approved):
lgtm, great work!

PR #3178, opened 2026-04-30
[GraphTrainer] Fix FlexAttention precompile bitwise deterministic tests

bobrenjc93:
* __->__ #3178

The FlexAttention precompile tests were skipped with "numerics mismatch
— under investigation". The root cause was that _run_steps_with_precompile
did not pass attention_masks to the model, while GraphTrainer's
post_dataloading_process generates a causal BlockMask for FlexAttention
models. This meant the two paths computed different attention patterns:
- GraphTrainer path: causal mask via BlockMask
- Precompile path: no mask (full attention)

Fix by adding _get_extra_kwargs() to the test base class, which generates
the same BlockMask attention masks that post_dataloading_process produces.
_run_steps_with_precompile now passes these masks and calls
maybe_register_blockmask_pytree_node() before tracing.

Depends on https://github.com/pytorch/pytorch/pull/182048


torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
```diff
@@ -110,6 +113,27 @@ def tearDown(self):
         FlexAttention.inductor_configs = self._orig_inductor_configs
         FlexAttention._compiled_flex_attn = self._orig_compiled_flex_attn
 
+    def _get_extra_kwargs(self, model: nn.Module) -> dict[str, object]:
```
SherlockNoMad, line 120: is this needed in the main precompile path?

SherlockNoMad (approved):
great, thanks for the quick fix!

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
bobrenjc93: Yes, _precompile_aot_fx_trace in precompile_main.py (lines 346–378) already builds extra_kwargs, including attention_masks from model.get_attention_masks(). The test’s _get_extra_kwargs simply mirrors that logic.

The test needs this because _run_steps_with_precompile performs its own trace, save, and load cycle instead of going through precompile_main.

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
```diff
@@ -456,8 +479,6 @@ def test_aot_fx_trace_vs_eager(self):
         run_traced = self._run_steps(copy.deepcopy(self.model), GraphTrainer)
         self._assert_runs_match(run_eager, run_traced, "eager vs aot_fx_trace: ")
 
-    # TODO: numerics mismatch between precompile and trace with FlexAttention
-    @unittest.skip("FlexAttention precompile numerics mismatch — under investigation")
     def test_precompile_vs_trace(self):
```
SherlockNoMad, line 486: FYI, if you rebase, you should see one more skipped precompile test for Qwen3Flex
would appreciate if you can also unskip that

bobrenjc93: will do

PR #3180, opened 2026-04-30
[CI] Use replace-imports-with-any to avoid missing packages causing CI failures

fegin:
* __->__ #3180

    Hope this fix the pyrefly painpoint caused by missing packages such as deep_ep and torchvision.

    `replace-imports-with-any` vs `ignore-missing-imports`:

    |                       | `ignore-missing-imports`                                              | `replace-imports-with-any`                                       |
    |-----------------------|-----------------------------------------------------------------------|------------------------------------------------------------------|
    | When it kicks in      | Only when the module can't be found on disk                           | Always, whether the module is installed or not                   |
    | What it does          | Suppresses the "missing-import" error and types the imported names as `Any` | Pretends the module is `Any` -- skips loading its real types entirely |
    | If the lib IS installed | Pyrefly type-checks against real types -> may surface errors -> contributors add ignores | Pyrefly still treats it as `Any` -> no errors -> behavior identical to CI |

    `ignore-missing-imports` only papers over the absence of a library. The moment a contributor installs it locally, pyrefly switches modes and starts type-checking it for real -- surfacing errors that contributors then silence with `# pyrefly: ignore [...]`. CI (without the lib) sees those ignores as useless
and tells contributors to remove them. That's the asymmetry.

    `replace-imports-with-any` removes the asymmetry by forcing everyone -- installed or not -- to see the module as `Any`. The module's real types are never consulted, so there are no environment-dependent errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,5 +64,5 @@ testpaths = ["tests"]
 
 [tool.pyrefly]
 project-excludes = ["torchtitan/experiments", "**/tests/**"]
-ignore-missing-imports = ["torchao.*", "torchft"]  # optional dependencies
+replace-imports-with-any = ["torchao.*", "torchft", "torchvision.*", "deep_ep.*", "jinja2.*"]  # optional dependencies
 search-path = ["../pytorch"]  # local built pytorch
--- a/torchtitan/components/tokenizer.py
+++ b/torchtitan/components/tokenizer.py
@@ -9,7 +9,7 @@ import json
 import os
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Any, cast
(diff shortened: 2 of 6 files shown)


tianyu-l (approved):

acisseJZhong:
thanks for addressing this!! huge life saver.

kinto0:
I'm curious - what is the reason these imports are skipped in deep_ep and torchvision? is it because some platforms can't install it? or do different platforms have different implementations?

acisseJZhong:
packages like deepep is not a hard dependency, and it takes time to build, so we didn't include/install in Titan CI. Therefore we added the missed import annotations to ignore pyrefly errors.

PR #3182, opened 2026-04-30
[ft] Forward attn_backend to llama3 config functions

d4l3k:
## Summary
- The upstream llama3 config functions (e.g. `_debugmodel`) now require an `attn_backend` parameter, but the FT experiment's `model_registry` was calling `llama3_configs[flavor]()` with no arguments
- This caused a `TypeError: _debugmodel() missing 1 required positional argument: 'attn_backend'` in CI ([job link](https://github.com/pytorch/torchtitan/actions/runs/25185478406/job/73841557782))
- Added `attn_backend` parameter to the FT experiment's `model_registry` with default `"sdpa"`, matching the upstream convention

## Test plan
- [ ] CI torchft integration test passes (previously failing with TypeError)

### DIFF
--- a/torchtitan/experiments/ft/llama3/__init__.py
+++ b/torchtitan/experiments/ft/llama3/__init__.py
@@ -14,8 +14,8 @@ from torchtitan.models.llama3 import (
 )
 
 
-def model_registry(flavor: str) -> FaultTolerantModelSpec:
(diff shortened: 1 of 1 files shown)


tianyu-l (approved):

acisseJZhong (approved):
thanks for catching!

PR #3188, opened 2026-04-30
[Checkpoint] BaseModel state dict pipeline and pure I/O CheckpointManager

mori360:
* #3189
* __->__ #3188

BaseModel owns the full state dict pipeline: filter (by requires_grad mode) → transform (HF key remapping) → cast (dtype, or any other requirements later). ModelWrapper is removed.

- get_sd(mode, external, dtype): single entry point. FULL/TRAINABLE/BASE modes filter by requires_grad via DCP ignore_frozen_params.
- to_external(sd, mode) / from_external(sd, mode): mode-based transform routing. BASE applies HF remapping, TRAINABLE reserved for converters.
- set_sd_transforms(sd_adapter): extracts to_hf/from_hf from sd_adapter. I/O concerns (storage reader/writer) stay on CheckpointManager.
- load_sd: strict=False for partial loads.

CheckpointManager becomes pure I/O:
- _load_from_hf(container, path) and _load_from_dcp(container, path) take a container dict and fill it. Caller builds container, converts, and applies — uniform pattern for all load paths.
- hf_io (renamed from sd_adapter) for HF storage reader/writer only.
- keys_match_mapping for adapter-only HF saves (consolidation mode).
- Initial load always uses BASE mode (== FULL when nothing frozen).

Trainer simplified:
- Removed _collect_adapter_key_patterns and freeze logic.
- One-line wiring: model_ref.set_sd_transforms(sd_adapter).

### DIFF
--- a/scripts/checkpoint_conversion/convert_from_hf.py
+++ b/scripts/checkpoint_conversion/convert_from_hf.py
@@ -11,7 +11,7 @@ from pathlib import Path
 import torch
 import torch.distributed.checkpoint as dcp
 from torch.distributed.checkpoint import HuggingFaceStorageReader
-from torchtitan.components.checkpoint import ModelWrapper
+from torch.distributed.checkpoint.state_dict import get_model_state_dict
 
 
 @torch.inference_mode()
@@ -23,14 +23,12 @@ def convert_from_hf(input_dir, output_dir, model_name, model_flavor):
 
     with torch.device("cpu"):
         model = model_config.build()
-    model = ModelWrapper(model)
-
     sd_adapter = model_spec.state_dict_adapter(model_config, None)
     assert (
         sd_adapter is not None
     ), "trying to convert checkpoint from HF to DCP safetensors format, but sd_adapter is not provided."
     # get state dict in tt format with allocated memory
-    state_dict = model._get_state_dict()
+    state_dict = get_model_state_dict(model)
     # convert empty state dict to hf format so that hf weights can be loaded into it
     hf_state_dict = sd_adapter.to_hf(state_dict)
     dcp.load(
--- a/torchtitan/components/checkpoint.py
+++ b/torchtitan/components/checkpoint.py
@@ -335,6 +335,7 @@ class CheckpointManager(Configurable):
         self.initial_load_in_hf_quantized = config.initial_load_in_hf_quantized
         self.last_save_model_only = config.last_save_model_only
         self.last_save_in_hf = config.last_save_in_hf
+        self.create_seed_checkpoint = config.create_seed_checkpoint
 
         # HF I/O config extracted from sd_adapter. CheckpointManager only
         # needs these for HF storage reader/writer — key transforms are on
@@ -816,9 +817,12 @@ class CheckpointManager(Configurable):
         )
 
     def _save_last_step(self, curr_step: int) -> None:
-        # Seed checkpoint (step 0) saves ALL params so it can be loaded
-        # under any parallelism. Training last-step saves only trainable.
-        save_mode = StateDictMode.FULL if curr_step == 0 else StateDictMode.TRAINABLE
+        # Seed checkpoints save ALL params so they can be loaded under any
+        # parallelism. Training last-step saves only trainable.
+        if self.create_seed_checkpoint:
+            save_mode = StateDictMode.FULL
+        else:
+            save_mode = StateDictMode.TRAINABLE
         if self.last_save_model_only:
             export_dtype = (
                 self.export_dtype
--- a/torchtitan/components/checkpoint.py
+++ b/torchtitan/components/checkpoint.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import enum
-import functools
 import os
 import queue
 import re
@@ -21,32 +20,26 @@ from typing import Any, cast, Literal
 import torch
 import torch.distributed as dist
 import torch.distributed.checkpoint as dcp
-import torch.nn as nn
 from torch.distributed.checkpoint import HuggingFaceStorageWriter
 from torch.distributed.checkpoint._consolidate_hf_safetensors import (
     consolidate_safetensors_files_on_every_rank,
 )
 from torch.distributed.checkpoint.staging import DefaultStager, StagingOptions
-from torch.distributed.checkpoint.state_dict import (
-    get_model_state_dict,
-    set_model_state_dict,
-    StateDictOptions,
-)
 from torch.distributed.checkpoint.state_dict_saver import (
     AsyncCheckpointerType,
     AsyncSaveResponse,
 )
-from torch.distributed.checkpoint.stateful import Stateful
 from torchtitan.components.dataloader import BaseDataLoader
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
 from torchtitan.config import Configurable, TORCH_DTYPE_MAP
+from torchtitan.protocols.model import BaseModel, StateDictMode
 from torchtitan.protocols.state_dict_adapter import BaseStateDictAdapter
 from torchtitan.tools.logging import logger
 from torchtitan.tools.utils import GarbageCollection
 
 
-MODEL = "model"
+MODEL = "model"  # Kept for backward compat with experiment code
 OPTIMIZER = "optimizer"
 LR_SCHEDULER = "lr_scheduler"
 DATALOADER = "dataloader"
@@ -57,32 +50,6 @@ class AsyncMode(str, enum.Enum):
     DISABLED = "disabled"
     ASYNC = "async"
     ASYNC_WITH_PINNED_MEM = "async_with_pinned_mem"
-
-
-class ModelWrapper(Stateful):
-    def __init__(self, model: nn.Module | list[nn.Module]) -> None:
-        self.model = [model] if isinstance(model, nn.Module) else model
-        self.cache_state_dict = self._get_state_dict()
-
-    def _get_state_dict(self) -> dict[str, Any]:
-        state_dict = {
-            k: v for sd in map(get_model_state_dict, self.model) for k, v in sd.items()
-        }
-        return state_dict
-
-    def state_dict(self) -> dict[str, Any]:
-        return self.cache_state_dict
-
-    def load_state_dict(self, state_dict: dict[str, Any]) -> None:
-        func = functools.partial(
-            set_model_state_dict,
-            model_state_dict=state_dict,
-            options=StateDictOptions(strict=False),
-        )
-        list(map(func, self.model))
-        # `set_model_state_dict()` does change the keys of the input state_dict,
-        # we will need to reinitialize the cache_state_dict.
-        self.cache_state_dict = self._get_state_dict()
 
 
 class Terminate:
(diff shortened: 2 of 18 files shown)


scripts/checkpoint_conversion/convert_from_hf.py
```diff
@@ -11,7 +11,7 @@
 import torch
 import torch.distributed.checkpoint as dcp
 from torch.distributed.checkpoint import HuggingFaceStorageReader
-from torchtitan.components.checkpoint import ModelWrapper
+from torch.distributed.checkpoint.state_dict import get_model_state_dict
```
tianyu-l, line 14: what's the plan with https://github.com/pytorch/torchtitan/pull/2441? after which we shouldn't need to call `get_model_state_dict`.

cc @fegin

torchtitan/trainer.py
```diff
@@ -173,7 +173,6 @@ def maybe_log(self) -> None:
     checkpointer: CheckpointManager
 
     # runtime utilities
-    device: torch.device
```
tianyu-l, line 176: why removing?

torchtitan/trainer.py
```diff
@@ -442,17 +441,26 @@ def __init__(self, config: Config):
...
+        if parallel_dims.pp_enabled:
+            model_ref._pp_parts = self.model_parts
+
+        sd_adapter = (
+            model_spec.state_dict_adapter(model_config, config.hf_assets_path)
+            if model_spec.state_dict_adapter
+            else None
+        )
+
+        model_ref.set_sd_transforms(sd_adapter)
```
tianyu-l: Let's not have this. Instead let's put `state_dict_adapter` as an attribute of `BaseModel` https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/model.py#L13

We need to move it out from `ModelSpec` https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/model_spec.py#L43

torchtitan/components/checkpoint.py
```diff
@@ -319,10 +301,10 @@ def __init__(
         self.enable = config.enable
         self.load_only = config.load_only
 
+        self.model = model
         self.states = states
         self.states.update(
             {
-                MODEL: ModelWrapper(model_parts),
```
tianyu-l, line 325: why removing instead of replacing with `model`?

torchtitan/components/checkpoint.py
```diff
@@ -428,15 +417,17 @@ def dcp_save(
         checkpoint_id: str,
         async_mode: AsyncMode,
         enable_garbage_collection: bool = False,
-        to_hf: bool = False,
+        hf_storage: bool = False,
     ) -> Future | AsyncSaveResponse | None:
         """Save the checkpoint with dcp.
         Args:
-            state_dict (dict): The state dict to save.
+            state_dict (dict): The state dict to save (key transforms already applied).
```
tianyu-l, line 499: what key transforms?

torchtitan/components/checkpoint.py
```diff
@@ -353,11 +335,18 @@ def __init__(
         self.initial_load_in_hf_quantized = config.initial_load_in_hf_quantized
         self.last_save_model_only = config.last_save_model_only
         self.last_save_in_hf = config.last_save_in_hf
+        self.create_seed_checkpoint = config.create_seed_checkpoint
+
+        # HF I/O config extracted from sd_adapter. CheckpointManager only
+        # needs these for HF storage reader/writer — key transforms are on
+        # the model via model.set_sd_transforms().
+        self.hf_io = sd_adapter
```
tianyu-l: why changing the name?

torchtitan/components/checkpoint.py
```diff
@@ -428,15 +417,17 @@ def dcp_save(
         checkpoint_id: str,
         async_mode: AsyncMode,
         enable_garbage_collection: bool = False,
-        to_hf: bool = False,
+        hf_storage: bool = False,
```
tianyu-l, line 495: why changing name?

torchtitan/components/checkpoint.py
```diff
@@ -447,14 +438,19 @@ def dcp_save(
...
-            ), "trying to save checkpoint in HF safetensors format, but sd_adapter is not provided."
-            state_dict = self.sd_adapter.to_hf(state_dict)
-
-            fqn_to_index_mapping = self.sd_adapter.fqn_to_index_mapping
-            if fqn_to_index_mapping:
+                self.hf_io is not None
+            ), "trying to save in HF safetensors format, but sd_adapter is not provided."
+            fqn_to_index_mapping = self.hf_io.fqn_to_index_mapping
+            # Only use sharded mapping when saved keys match the mapping
+            # (e.g. base model HF save). For adapter-only saves (PEFT),
```
tianyu-l, line 523: there is no adapter in this PR yet, can we delay the change to another PR?

torchtitan/components/checkpoint.py
```diff
@@ -631,86 +620,120 @@ def load(self, step: int = -1) -> bool:
...
-            if from_hf:
-                assert (
-                    model_only
-                ), "Only model can be loaded when loading from HF's safetensors checkpoint."
-
-            if from_quantized:
-                assert (
-                    from_hf
-                ), "Quantized checkpoint can only be loaded from HuggingFace format."
+        begin = time.monotonic()
```
tianyu-l, line 699: starting from here, the logic is too complicated and monolithic that I can't read

torchtitan/protocols/model.py
```diff
@@ -21,6 +45,138 @@ class BaseModel(Module):
...
+        CheckpointManager.
+
+        Called once by the trainer after model build.
+        """
+        if sd_adapter is not None:
+            self._to_hf = sd_adapter.to_hf
+            self._from_hf = sd_adapter.from_hf
+
+    def to_external(self, sd: dict[str, Any], mode: StateDictMode) -> dict[str, Any]:
+        """Convert native state dict keys to external format.
```
tianyu-l: confusing, why we have both hf and external?

torchtitan/trainer.py
```diff
@@ -442,17 +441,26 @@ def __init__(self, config: Config):
         self.step = 0
         self.ntokens_seen = 0
 
+        # Set up BaseModel reference for checkpoint (PP parts handled via _pp_parts)
+        model_ref = cast(BaseModel, self.model_parts[0])
+        if parallel_dims.pp_enabled:
+            model_ref._pp_parts = self.model_parts
```
tianyu-l: no, model_parts[0] shouldn't hold a reference to all model_parts

tianyu-l (changes requested):

tianyu-l:
discussed offline, two main feedback
- For the sake of PP, we would keep `ModelWrapper`, but it should be a thin wrapper to hold all model_parts. Each individual model part (`BaseModel`) should handle the state dict content (hf or not, trainable / frozen / all). In the future we may reconsider how we do PP (e.g. one model hosting all stages).
- Right now there seem to be complicated logic in save / load (e.g. load twice if saved a lora adapter before, firstly base model, then lora adapter). I would advocate for making it work e2e with a simple solution, namely -- provide two new configs (`initial_load_frozen_only`, `last_save_trainable_only`), which can be combined with existing hf options to achieve the following workflow: (1) create a model with lora, load an HF base model (2) save full DCP checkpoints (base + lora) (3) load full DCP checkpoints (base + lora) (4) save lora adapters only, in HF.

cc @fegin

scripts/checkpoint_conversion/convert_from_hf.py
fegin: We can land #2441

fegin (commented):
> Right now there seem to be complicated logic in save / load (e.g. load twice if saved a lora adapter before, firstly base model, then lora adapter). I would advocate for making it work e2e with a simple solution, namely -- provide two new configs (initial_load_frozen_only, last_save_trainable_only), which can be combined with existing hf options to achieve the following workflow: (1) create a model with lora, load an HF base model (2) save full DCP checkpoints (base + lora) (3) load full DCP checkpoints (base + lora) (4) save lora adapters only, in HF.

If we are doing this, we shouldn't make checkpointer more complicated. Instead this first two steps should be an offline process/script that lives outside of the main trainer - it can be a script that creates a customized checkpointer to do the offline process. If we are doing online conversion, it doesn't make sense to have the first two steps in the main checkpointer. 

Also avoid saving the `frozen model` was one request before because not every user has the luxury to have the storage to store two copies of the large models. I'm not sure if this is still the request.

torchtitan/components/checkpoint.py
```diff
@@ -53,36 +53,63 @@
...
+                trainable_sd.update(
+                    get_model_state_dict(
+                        part,
+                        options=StateDictOptions(ignore_frozen_params=True),
+                    )
+                )
+            if len(trainable_sd) == len(full_sd):
+                return full_sd
+            return {k: v for k, v in full_sd.items() if k not in trainable_sd}
+        return self._full_sd()
```
Andrei-Aksionov, line 98: It might be a noob question, but when we fine-tune a base model via LoRA, why do we might want to save base weights? These are frozen, don't change.

Or it is because weights are already sharded and if we need to restart training on a failed node, it already has it's shard of base weights so it should be faster to load them?

Even if it is so, why to do it on every checkpoint step? Just do it once and after save only trainable weights.

torchtitan/components/checkpoint.py
```diff
@@ -53,36 +53,63 @@
...
-        return state_dict
+    def _full_sd(self) -> dict[str, Any]:
+        sd: dict[str, Any] = {}
+        for part in self.model_parts:
+            sd.update(get_model_state_dict(part))
+        return sd
 
     def state_dict(self) -> dict[str, Any]:
-        return self.cache_state_dict
+        if self.mode == StateDictMode.TRAINABLE:
```
Andrei-Aksionov, line 75: Since state_dict doesn't return "cached" version of the state dict, it means that every call will return new tensors and thus it will force to reassign pinned memory to them, which effectively makes pinned version not-working (no benefits). 
Is it correct?

torchtitan/components/checkpoint.py
```diff
@@ -281,6 +308,21 @@ class Config(Configurable.Config):
...
+        from the checkpoint folder.
+        """
+
+        last_save_trainable_only: bool = False
+        """
+        When True, last step save only saves trainable params (e.g. LoRA
+        adapters). Combine with ``last_save_in_hf=True`` to export adapters
+        in PEFT safetensors format.
+        """
+
```
Andrei-Aksionov, line 325: As I understand it correctly, this state of the code doesn't support loading LoRA weights for training continuation (i.e. continue fine-tuning someone else's work).
But the code introduced complexity of loading weights into trainable parameters via ModelWrapper's mode.

torchtitan/components/checkpoint.py
```diff
@@ -53,36 +53,63 @@
...
 
 
-class AsyncMode(str, enum.Enum):
-    DISABLED = "disabled"
-    ASYNC = "async"
-    ASYNC_WITH_PINNED_MEM = "async_with_pinned_mem"
+class ModelWrapper(Stateful):
+    """Stateful wrapper that merges state dicts across PP stages.
 
+    Handles pipeline-parallel iteration and mode-based filtering.
```
Andrei-Aksionov, line 59: I don't think that it's a good idea to introduce mode-based filtering into this class.
Perhaps it's better to do such filtering in a dedicated method in the CheckpointManager class - a class that, em, manages such things.

PR #3189, opened 2026-04-30
[LoRA] Converter protocol and PEFT external format support

mori360:
* __->__ #3189
* #3188

Add converter protocol for external format transforms (e.g. PEFT key remapping). Converters are stored on ModelSpec.converters and wired via BaseModel.set_sd_transforms(sd_adapter, converters).

ModelSpec:
- Add converters: list[Any] field. Config registry appends converter instances after convert(). Trainer iterates via duck typing.

BaseModel:
- Extend set_sd_transforms to accept converters list. Each converter may implement build_external_transforms(sd_adapter) returning {to_external, from_external} callables.
- to_external/from_external now route by mode: BASE applies HF remapping, TRAINABLE applies converter transforms, FULL applies both.

LoRAConverter:
- Add build_external_transforms instance method returning PEFT key remapping (remap_lora_keys_to_hf / remap_lora_keys_from_hf).
- Remove _is_adapter / _key_pattern ClassVars (filtering is by requires_grad, not pattern matching).

Config registry:
- llama3_debugmodel_lora: append converter to model_spec.converters.
- llama3_debugmodel_lora_checkpoint: LoRA with HF save enabled.

LoRA TP Determinism with Seed Checkpoint:
With seed checkpoint and non-chunked CrossEntropyLoss, LoRA training produces bitwise identical loss and grad_norm across TP=1/2/4, DP=2 over 20 steps:
```
  LoRA + CrossEntropyLoss + seed checkpoint: TP=1 vs TP=2 vs TP=4 

  Loss:
  Step  1: TP1=8.02020  TP2=8.02021  TP4=8.02018
  Step  5: TP1=7.97558  TP2=7.97558  TP4=7.97557
  Step 10: TP1=7.72779  TP2=7.72779  TP4=7.72779
  Step 15: TP1=7.42778  TP2=7.42787  TP4=7.42784
  Step 20: TP1=7.22069  TP2=7.22079  TP4=7.22076

  Grad norm:
  Step  1: TP1=0.1591   TP2=0.1591   TP4=0.1592
  Step  5: TP1=0.2866   TP2=0.2866   TP4=0.2866
  Step 10: TP1=0.6532   TP2=0.6533   TP4=0.6533
  Step 15: TP1=1.1474   TP2=1.1473   TP4=1.1474
  Step 20: TP1=1.4419   TP2=1.4417   TP4=1.4417
```
Loss and grad_norm match to 5 significant digits (the stdout display precision) with no drift over 20 steps. Multiple steps show exact zero diff. This confirms that LoRA's TP parallelization — adapter sharding, frozen base weights, and gradient computation — is numerically correct across TP degrees.                                                                                                               
The seed checkpoint ensures identical initial weights (base + LoRA adapters) across all TP configurations, eliminating data RNG divergence as a variable. The non-chunked CrossEntropyLoss avoids a known ChunkedCELoss cross-TP determinism issue.

### DIFF
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -84,7 +84,7 @@ def _get_lora_cls(parent_cls: type) -> type:
         ) -> None:
             super().__init__(config)
             # Freeze base weight — only adapters should be trainable
-            for param in self.parameters():
+            for param in nn.Module.parameters(self):
                 param.requires_grad_(False)
             self._lora_scaling = _lora_alpha / _lora_rank
 
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -155,6 +155,7 @@ class LoRAConfig(Configurable.Config):
         from dataclasses import replace
 
         inner = self.inner
+        assert inner._owner is not None
         lora_cls = _get_lora_cls(inner._owner)
         instance = lora_cls(
             config=replace(inner),
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -1,0 +1,403 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import math
+import re
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+import torch.nn as nn
+
+from torch.distributed.tensor import Replicate
+
+from torchtitan.config import Configurable
+from torchtitan.models.common.decoder_sharding import dense_param_placement
+from torchtitan.models.common.linear import Linear
+from torchtitan.protocols.sharding import ShardingConfig
+from torchtitan.tools.logging import logger
+
+
+def _lora_adapter_sharding(
+    base_sharding: ShardingConfig | None,
+) -> tuple[ShardingConfig, ShardingConfig]:
+    """Derive LoRA adapter sharding from the base linear's TP sharding.
+
+    Both adapters are TP-sharded to match the base linear, ensuring all
+    LoRA gradients have the same TP placement — no mixed Replicate/Shard
+    norms in ``clip_grad_norm_``.
+
+    ``lora_b`` mirrors the base weight's TP shard so its output matches
+    ``base_out``'s placement (Shard(-1) for colwise, Partial for rowwise).
+
+    ``lora_a`` weight is Replicate (small rank dim, no benefit from TP
+    sharding).
+
+    Note: under TP, ``clip_grad_norm_`` reports a grad_norm that differs
+    from single-GPU by a constant factor (TP degree). This is a known
+    DTensor ``_NormPartial`` norm computation issue with mixed
+    Replicate/Shard placements, not a LoRA correctness bug — the loss
+    and convergence are identical.
+    """
+    replicate_weight = ShardingConfig(
+        state_shardings={"weight": dense_param_placement(tp=Replicate())},
+    )
+    if base_sharding is None:
+        return replicate_weight, replicate_weight
+
+    base_weight_sharding = base_sharding.state_shardings.get("weight")
+    if base_weight_sharding is None:
+        return replicate_weight, replicate_weight
+
+    # lora_b: same TP shard as base weight → output matches base_out
+    lora_b_sharding = ShardingConfig(
+        state_shardings={"weight": base_weight_sharding},
+    )
+    return replicate_weight, lora_b_sharding
+
+
+# Cache for dynamically created LoRA classes
+_lora_class_cache: dict[type, type] = {}
+
+
+def _get_lora_cls(parent_cls: type) -> type:
+    """Get or create a LoRA subclass for *parent_cls* (e.g. Linear, Float8Linear).
+
+    The returned class has a proper ``__init__`` that initialises both the
+    base linear and the LoRA adapters in a single construction — no
+    build-then-mutate.
+    """
+    if parent_cls in _lora_class_cache:
+        return _lora_class_cache[parent_cls]
+
+    class LoRALinear(parent_cls):  # type: ignore[valid-type, misc]
+        def __init__(
+            self,
+            config: Linear.Config,
+            *,
+            _lora_rank: int,
+            _lora_alpha: float,
+            _base_sharding: ShardingConfig | None = None,
+        ) -> None:
+            super().__init__(config)
+            # Freeze base weight — only adapters should be trainable
+            self.weight.requires_grad_(False)
+            if self.bias is not None:
+                self.bias.requires_grad_(False)
+            self._lora_scaling = _lora_alpha / _lora_rank
+
+            lora_a_sharding, lora_b_sharding = _lora_adapter_sharding(_base_sharding)
+            self.lora_a = Linear.Config(
+                in_features=config.in_features,
+                out_features=_lora_rank,
+                bias=False,
+                sharding_config=lora_a_sharding,
+                param_init={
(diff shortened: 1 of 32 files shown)


torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+    from single-GPU by a constant factor (TP degree). This is a known
+    DTensor ``_NormPartial`` norm computation issue with mixed
+    Replicate/Shard placements, not a LoRA correctness bug — the loss
+    and convergence are identical.
+    """
+    replicate_weight = ShardingConfig(
+        state_shardings={"weight": dense_param_placement(tp=Replicate())},
+    )
+    if base_sharding is None:
+        return replicate_weight, replicate_weight
```
tianyu-l: can we ban this case? When would we need LoRA sharding when base sharding is None?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+
+    base_weight_sharding = base_sharding.state_shardings.get("weight")
+    if base_weight_sharding is None:
+        return replicate_weight, replicate_weight
```
tianyu-l: same here

torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+    ``base_out``'s placement (Shard(-1) for colwise, Partial for rowwise).
+
+    ``lora_a`` weight is Replicate (small rank dim, no benefit from TP
+    sharding).
+
+    Note: under TP, ``clip_grad_norm_`` reports a grad_norm that differs
+    from single-GPU by a constant factor (TP degree). This is a known
+    DTensor ``_NormPartial`` norm computation issue with mixed
+    Replicate/Shard placements, not a LoRA correctness bug — the loss
+    and convergence are identical.
```
tianyu-l: Is there an issue tracking this?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+        return _lora_class_cache[parent_cls]
+
+    class LoRALinear(parent_cls):  # type: ignore[valid-type, misc]
+        def __init__(
+            self,
+            config: Linear.Config,
+            *,
+            _lora_rank: int,
+            _lora_alpha: float,
+            _base_sharding: ShardingConfig | None = None,
```
tianyu-l: this should be available in config above, so don't need to pass separately

torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+            else:
+                wrapped = FrozenConfig(inner=cfg)
+
+            if isinstance(parent, list):
+                parent[attr] = wrapped
+            else:
+                setattr(parent, attr, wrapped)
+
+        # Freeze non-Linear leaf modules (norms, embeddings)
+        for config_cls in (RMSNorm.Config, Embedding.Config):
```
tianyu-l, line 266: This doesn't sound robust. A better way is to traverse all configs in one pass and freeze anything that's not Linear.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+            config=replace(inner),
+            _lora_rank=self.rank,
+            _lora_alpha=self.alpha,
+            _base_sharding=inner.sharding_config,
+        )
+        # Apply Module.Config properties that inner.build() would set
+        if inner.param_init is not None:
+            instance._param_init = inner.param_init
+        if inner.sharding_config is not None:
+            instance._sharding_config = inner.sharding_config
```
tianyu-l: I think the reason you are using this is because you didn't have a Config of `lora_cls` so you are not using the `build` function here https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/module.py#L47 which does these for you.

We should have a Config inside `class LoRALinear(parent_cls)` and put rank and alpha in config.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,403 @@
...
+        if name in type(self).__slots__:
+            object.__setattr__(self, name, value)
+        else:
+            setattr(self.inner, name, value)
+
+    def build(self, **kwargs):
+        from dataclasses import replace
+
+        inner = self.inner
+        assert inner._owner is not None
```
tianyu-l, line 129: I don't think it would work with today's mxfp8 linear, whose config doesn't have an `_owner` https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/mx.py#L158

I think a more consistent way is to always dynamically create a class for mxfp8, similar to the LoRALinear in this PR.
cc @acisseJZhong

torchtitan/models/llama3/config_registry.py
```diff
@@ -104,6 +104,28 @@ def llama3_debugmodel_float8() -> Trainer.Config:
...
+def llama3_debugmodel_lora() -> Trainer.Config:
+    from torchtitan.components.lora import LoRAConverter
+
+    config = llama3_debugmodel()
+    lora_converter = LoRAConverter(
+        LoRAConverter.Config(rank=8, alpha=16.0, target_modules=["wq", "wkv", "wo"])
+    )
+    assert config.model_spec is not None
+    lora_converter.convert(config.model_spec.model)
+    config.model_spec.converters.append(lora_converter)
```
tianyu-l: we should extend the `quantization` field in `model_registry` from quantization converters to general converters, and apply converters in sequential order
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/config_registry.py#L100-L102

A safety check is that LoRAConverters have to be applied after all quantization converters.

torchtitan/components/lora.py
tianyu-l: also, with that I don't think you still need this `LoRAConfig` class.

tianyu-l (commented):
Only reviewed the LoRA logic. Will review state dict transformation later.

torchtitan/components/lora.py
mori360: Have the MXFP8GroupedExperts pr here: https://github.com/pytorch/torchtitan/pull/3251

torchtitan/models/qwen3_vl/__init__.py
```diff
@@ -591,4 +595,5 @@ def model_registry(
         pipelining_fn=None,
         post_optimizer_build_fn=None,
         state_dict_adapter=Qwen3VLStateDictAdapter,
+        converters=built_converters,
```
tianyu-l, line 598: don't need this in `ModelSpec`?

torchtitan/models/qwen3/__init__.py
```diff
@@ -619,15 +619,19 @@ def model_registry(
...
         kwargs["moe_comm_backend"] = moe_comm_backend
     config = qwen3_configs[flavor](**kwargs)
-    if quantization is not None:
-        for q in quantization:
-            q.build().convert(config)
+    built_converters: list = []
+    if converters is not None:
+        for c in converters:
+            built_converters.append(c.build())
+        validate_converter_order(built_converters)
```
tianyu-l, line 632: I think can validate before `c.build()` -- just need to validate on config, e.g. `if isinstance(converter, LoRAConverter.Config):`

torchtitan/protocols/model_spec.py
```diff
@@ -23,6 +23,28 @@
 PostOptimizerBuildFn: TypeAlias = Callable[..., None]
 
 
+def validate_converter_order(converters: list) -> None:
```
tianyu-l, line 26: let's put in torchtitan/models/utils.py for now

tianyu-l (commented):
Right now the lora PR is after checkpointing PR. I think it's easier to land 1. LoRA first, then 2. checkpointing, then 3. lora + checkpointing (can be combined with 2).

I.e. it should be fine to have a version of LoRA which doesn't work with checkpointing compatible.

PR #3160, opened 2026-04-29
[NOT READY FOR REVIEW][Module][MoE] Config-based MoE sharding (qwen3, llama4, gpt_oss, deepseek_v3)

fegin:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #3165
* #3161
* __->__ #3160
* #3159

Replace the imperative apply_moe_ep_tp with declarative ShardingConfig on
MoE submodules, mirroring the dense config-based sharding established by
PR1-PR3. Each model's update_from_config now populates sharding plans
on the MoE wrapper, router gate, shared experts, and routed experts;
``model.parallelize(parallel_dims)`` walks the tree and applies them.

Verified bit-identical loss/grad_norm to legacy at step 500 under EP=2
(``--debug.seed=42 --debug.deterministic``, 4xH100):

  - qwen3:       loss 2.02413, grad_norm 0.4862
  - llama4:      loss 1.80384, grad_norm 0.1891
  - deepseek_v3: loss 2.02331, grad_norm 0.2702
  - gpt_oss:     loss 1.90064, grad_norm 0.2658

Add ``local_input_grad_placements`` and ``local_output_grad_placements``
to ``ShardingConfig``, mirroring ``ColwiseParallelWithGradPlacement
(local_input_grad_placements=...)`` and ``NoParallel
(local_output_grad_placements=...)``. Plumbed through ``Module._needed_axes``,
``_shard_inputs`` (passes ``grad_placements`` to ``DTensor.from_local``),
and ``_shard_outputs`` (calls ``to_local(grad_placements=...)`` after the
output redistribute, switching the wrapper's effective return type to
local tensor when set).

These fields express what the imperative ``NoParallel`` /
``ColwiseParallelWithGradPlacement`` did for the legacy MoE plan; without
them, removing the per-linear annotations would silently over-count
gradients (Partial treated as Replicate downstream).

- ``expert_param_placement_sparse(etp_placement)`` -- ``{EP: Shard(0),
  ETP: ...}`` plus the canonical FSDP storage axes for routed experts
  under EP.
- ``expert_param_placement_dense(tp_placement)`` -- ``{TP: ...}`` for
  the EP=1+TP>1 fallback (today's ``TensorParallel``).
- ``set_moe_sharding_config(moe_cfg, *, tp_enabled, ep_enabled,
  etp_enabled, enable_sp, expert_param_layout)`` -- top-level helper
[...]

PR #3192, opened 2026-05-01
Refactor MoE to clean DTensor boundaries for shared/routed experts

acisseJZhong:
**Copying major part from @fegin's PR https://github.com/pytorch/torchtitan/pull/3160** 

* #3376
* __->__ #3192

image

https://excalidraw.com/#json=2abKr0m2s26fc6lyoF9Qq,MqMzUIoXWYJIfckHNOB7Sw

## Summary

Refactor MoE parallelization to establish clean DTensor boundaries:
- **Router**: gate weights Replicate, input all-gathered to Replicate (via `NoParallel`), compute on DTensor, output converted to local via `to_local(grad_placements=Partial)`
- **Shared experts**: weights TP-sharded (via `ColwiseParallel`/`RowwiseParallel`), input all-gathered to Replicate inside `ColwiseParallel`, compute on DTensor, output local via `to_local` (Partial placement, `use_local_output=True`)
- **Routed experts**: input converted to local via `to_local` at `GroupedExperts` boundary, EP-partitioned weights on local tensors, dispatch/compute/combine all local

Key changes:
- Move `shared_experts` computation from inside `TokenDispatcher.combine()` to `MoE.forward()`
- Move `to_local` from `MoE.forward()` into `GroupedExperts.forward()`, with proper `grad_placements` based on placement type (Shard vs Replicate)
- Replace `ColwiseParallelWithGradPlacement` with standard `ColwiseParallel` using `input_layouts`
- Remove `shared_experts` parameter from all token dispatchers, replace with `initial_output` buffer
- For EP+SP, keep `Shard(1)` as desired MoE input layout (avoid redundant all-gather at MoE boundary); for TP-only, keep `Replicate()` (required for TP-sharded expert weight reduction)

### Trade-offs
- Removes async overlap between `shared_experts` and the all-to-all combine. Previously `shared_experts` ran while the combine A2A was in flight on the NCCL stream. A follow-up can restore overlap using CUDA streams.

## Test plan
- [X] TP=1 EP=1 runs, losses different than main starting step1. 
- [X] TP=2 EP=1 runs, losses different than main starting step1.  
- [X] TP=1 EP=2 runs, losses different than main starting step1. 
- [X] TP=2 EP=2 runs, losses different than main starting step1. 

Loss is expected to diverge compared to main due to different reduction pattern, and shared expert computation changes place.

single run
```
NCCL_NVLS_ENABLE=0 NGPU=4 MODULE='llama4' CONFIG='llama4_debugmodel' ./run_train.sh \
    --parallelism.pipeline_parallel_degree 1 \
    --parallelism.data_parallel_shard_degree 2 \
    --parallelism.expert_parallel_degree 2 \
    --debug.deterministic --debug.seed=42
```
Compare loss
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```


torchtitan/models/llama4/parallelize.py
```diff
@@ -378,45 +379,37 @@ def apply_moe_ep_tp(
...
-                            local_input_grad_placements=(Partial(),)
-                        ),
-                        "moe.shared_experts.w2": RowwiseParallel(
-                            output_layouts=Partial(),
-                        ),
-                        "moe.shared_experts.w3": ColwiseParallelWithGradPlacement(
-                            local_input_grad_placements=(Partial(),)
-                        ),
+                        "moe.shared_experts.w1": ColwiseParallel(),
+                        "moe.shared_experts.w2": RowwiseParallel(),
```
tianyu-l: you can make the output Partial?

torchtitan/models/llama4/parallelize.py
```diff
@@ -378,45 +379,37 @@ def apply_moe_ep_tp(
...
-                # Without SP: input is already Replicate,
-                # all-reduce (Partial→Replicate) for output.
+                # slice (Replicate→Shard) for output.
+                # Without SP: input/output are both Replicate (no-op).
                 "moe": PrepareModuleInputOutput(
                     input_layouts=(sp_layout,),
                     desired_input_layouts=(Replicate(),),
                     use_local_input=False,
-                    output_layouts=(Partial(),),
+                    output_layouts=(Replicate(),),
```
tianyu-l: same here, can we make the output still Partial?

torchtitan/models/llama4/parallelize.py
```diff
@@ -378,45 +379,37 @@ def apply_moe_ep_tp(
...
                 # replicate computation for the router
                 "moe.router.gate": NoParallel(
                     local_output_grad_placements=(Partial(),),
                 ),
+                # All-reduce routed expert output from Partial to Replicate
+                # on the TP mesh. GroupedExperts returns (scored_output, scatter_indices);
+                # scored_output is Partial and needs all-reduce, scatter_indices passes through.
+                "moe.experts": PrepareModuleOutput(
+                    output_layouts=(Partial(), None),
+                    desired_output_layouts=(Replicate(), None),
```
tianyu-l: don't need to convert to Replicate here? we can delay to the end of MoE layer

tianyu-l:
> This PR performs two
separate all-reduces — one on routed experts (PrepareModuleOutput)
and one on shared experts (RowwiseParallel) — then combines the
Replicate results (reduce(shared_partial) + reduce(routed_partial)).

I didn't get why we have to do this. Ideally the output of routed experts should be Partial, and be added with the output of shared expert's Rowwise result which is also Partial.

acisseJZhong:
>  Ideally the output of routed experts should be Partial, and be added with the output of shared expert's Rowwise result which is also Partial.

yeah sorry comment was not up to date, the code already reflects this. Will update in a bit, the idea is exactly what you said: for shared expert input should be Shard(1) output partial. For routed experts, input Shard(1) and it will all gather to R internally, and outputs is partial. And we will sum up these two partials and reduce to Replicate at MoE boundary.

tianyu-l:
> for shared expert input should be Shard(1) output partial. For routed experts, input Shard(1) and it will all gather to R internally, and outputs is partial

should be the other way?
- shared expert: S(1) -> AG to Replicate -> col/row -> Partial
- routed expert: S(1) -> Partial

acisseJZhong:
> shared expert: S(1) -> AG to Replicate -> col/row -> Partial
- ahh yes this is correct, this AG is automatically handled by ColwiseParallel? input activation is S(1), and output activation after col&rowwise is Partial.

> routed expert: S(1) -> Partial
- Are you talking about activations or weights? For The weights should be Replicate/NoParallel(), activations/tokens input is S(1), output is Partial().

tianyu-l:
> Are you talking about activations or weights? For The weights should be Replicate/NoParallel()

I'm talking about activations.

For router weight, I think it should be `SequenceParallel`?
For routed experts weight, should be S(0).

torchtitan/models/llama4/parallelize.py
```diff
@@ -380,55 +378,62 @@ def apply_moe_ep_tp(
...
                 ),
-                # replicate computation for the router
+                # GroupedExperts output is local (SP-sharded tokens + TP
+                # partial sums). Kept local so MoE.forward() can reshape
+                # before the "moe" boundary wraps it as DTensor.
+                "moe.experts": PrepareModuleOutput(
+                    output_layouts=Partial(),
+                    desired_output_layouts=Partial(),
+                    use_local_output=False,
+                ),
```
tianyu-l, line 406: seems no-op, can we delete?

torchtitan/models/llama4/parallelize.py
```diff
@@ -380,55 +378,62 @@ def apply_moe_ep_tp(
...
+                # Replicate computation for the router. NoParallel keeps
+                # gate weights Replicate and all-gathers input if needed
+                # (Shard→Replicate for EP+SP, no-op for TP-only).
+                # Output converted to local for dispatch compatibility.
                 "moe.router.gate": NoParallel(
+                    input_layout=moe_desired_input_layouts,
                     local_output_grad_placements=(Partial(),),
```
tianyu-l: this is only needed when output need `to_local` https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/tensor_parallel.py#L87
but we should put local_map around grouped mm?

torchtitan/models/llama4/parallelize.py
```diff
@@ -380,55 +378,62 @@ def apply_moe_ep_tp(
...
+                "moe.experts": PrepareModuleOutput(
+                    output_layouts=Partial(),
+                    desired_output_layouts=Partial(),
+                    use_local_output=False,
+                ),
+                # Replicate computation for the router. NoParallel keeps
+                # gate weights Replicate and all-gathers input if needed
+                # (Shard→Replicate for EP+SP, no-op for TP-only).
+                # Output converted to local for dispatch compatibility.
                 "moe.router.gate": NoParallel(
```
tianyu-l: My guess is that you don't use `SequenceParallel` because in pure TP you wish router to do Replicate computation.

torchtitan/models/common/moe.py
```diff
@@ -80,18 +80,25 @@ def forward(
...
+        Args:
+            initial_output: buffer for scatter_add (e.g. shared_experts output).
+                If None, scatter_add accumulates into zeros.
         """
+        # Convert DTensor to local tensor for routed expert dispatch/compute.
+        # Replicate x → Partial grad (TP-sharded weights produce partial sums).
+        # Shard x → Shard grad (each rank has its token shard's gradient).
+        if isinstance(x, DTensor):
+            is_sharded = _is_shard_like(x.placements[0])
+            x = x.to_local(grad_placements=x.placements if is_sharded else (Partial(),))
```
tianyu-l: I think it's more readable if we do
if Replicate -> grad Partial; o/w stay the same

torchtitan/models/common/token_dispatcher.py
```diff
@@ -227,24 +219,17 @@ def dispatch(
...
-                x = F.pad(x, (0, 0, 0, pad))
-                top_scores = F.pad(top_scores, (0, 0, 0, pad))
-                selected_experts_indices = F.pad(
-                    selected_experts_indices, (0, 0, 0, pad)
-                )
-            x, top_scores, selected_experts_indices = self._split_along_sp(
-                x, top_scores, selected_experts_indices
+            # x is already the local SP shard (from DTensor Shard to_local
+            # in GroupedExperts.forward). Only split scores/indices to match,
+            # since the router (NoParallel) computed them on all tokens.
```
tianyu-l: I didn't get this, why top scores and indices are not sharded? They are all outputs of router, whose input is sharded.

Also, I think for DeepEP, previous impl of tp2ep might be wrong (we missed the split) which needs to be fixed.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -686,17 +649,10 @@ def combine(
...
-        # shared_experts runs in parallel with combine communication.
-        # This is the key optimization - we overlap compute with communication.
-        shared_out = shared_experts(x) if shared_experts is not None else None
-
-        # Sync the combine operation before using routed_output.
-        # This inserts a CUDA stream wait, ensuring combine is complete before
-        # the subsequent addition or reshape operations read routed_output.
         from torchtitan.distributed.deepep.deepep import sync_combine
 
         sync_combine()
```
tianyu-l: this becomes useless, I think

torchtitan/models/common/moe.py
```diff
@@ -361,12 +345,11 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-            x,
-            top_scores,
-            selected_experts_indices,
-            shared_experts=self.shared_experts,
-        )
+        shared_out = self.shared_experts(x) if self.shared_experts is not None else None
+
+        # Routed experts convert to local tensor at GroupedExperts boundary.
+        # shared_out is passed to combine() as the scatter_add initial buffer.
+        out = self.experts(x, top_scores, selected_experts_indices, shared_out)
```
tianyu-l: I'm not sure if we should put `shared_out` in the input
- I think you are trying to avoid allocating both scatter_add input and shared expert output.
- However, DeepEP path didn't benefit from it (no scatter_add)
- some people may prefer calling `sync_combine` after here (per https://github.com/pytorch/torchtitan/pull/3192/changes#r3237956653). I don't know, maybe we should still conditionally call `sync_combine` here when DeepEP is used.
- when using local_map to wrap GroupedExperts (maybe in next PR), you need to specific the sharding placement for this field, which sounds unnecessary.

torchtitan/models/llama4/parallelize.py
tianyu-l: should be Partial only when GroupedExperts is TP

acisseJZhong:
will change other MoE models after we finalize all the llama4 changes

PR #3194, opened 2026-05-01
[GraphTrainer] Annotate generated FX code with user source lines

aditvenk:
Stacked PRs:
 * #3198
 * #3197
 * #3199
 * #3195
 * __->__#3194

--- --- ---

### [GraphTrainer] Annotate generated FX code with user source lines

Pass verbose=True to graph.python_code() so the dumped Python carries
stack-trace + annotation comments on each node, matching the format
produced by gm.print_readable(). Widen the leading-identifier regex in
_generate_profiled_forward to tolerate the per-node type annotations
(``name: "type" = ...``) that verbose mode adds, so the profiler-wrapper
pass still matches each op.

Co-Authored with Claude

### DIFF
--- a/torchtitan/experiments/graph_trainer/custom_codegen.py
+++ b/torchtitan/experiments/graph_trainer/custom_codegen.py
@@ -87,7 +87,7 @@ class _CodegenGraphModule(torch.fx.GraphModule):
         Our custom codegen generates its own per-op profiling in
         _forward_profiled, so the base code must be clean.
         """
-        code = gm.graph.python_code("self").src
+        code = gm.graph.python_code("self", verbose=True).src
         assert "_RecordFunctionFast" not in code, (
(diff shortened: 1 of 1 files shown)


torchtitan/experiments/graph_trainer/custom_codegen.py
SherlockNoMad: you should be able to get stacktrace + annotation by setting some flags of ptyhon_code here. 

Upstream pytorch fx/graph.py has already handle this. 

See gm.print_readble() as ref

aditvenk: oh! let me check

aditvenk: Thanks for the suggestion! Much smaller diff now

SherlockNoMad (approved):

xmfan (approved):

PR #3195, opened 2026-05-01
[GraphTrainer] Skip identity-slice rewrite when start/end/step are dynamic Nodes

aditvenk:
Stacked PRs:
 * #3198
 * #3197
 * #3199
 * __->__#3195

--- --- ---

### [GraphTrainer] Skip identity-slice rewrite when start/end/step are dynamic Nodes

remove_identity_slice_pass compared `end >= dim_size` (and used start/step
as concrete ints) unconditionally. When any of start/end/step is itself an
FX Node (a dynamic value produced by a prior op, e.g. from a symbolic
shape query) this raised
`TypeError: '>=' not supported between instances of 'Node' and 'SymInt'`
or gave a wrong answer when comparing a Node to 0/1.

Skip such nodes: we can't prove a dynamic slice is identity at pass time,
so leave it alone. Adds unit tests that reproduce the failure for each of
start, end, and step being a dynamic Node.

Co-Authored with Claude

### DIFF
--- a/torchtitan/experiments/graph_trainer/remove_noop_passes.py
+++ b/torchtitan/experiments/graph_trainer/remove_noop_passes.py
@@ -132,6 +132,16 @@ def remove_identity_slice_pass(
         end = args[3] if len(args) > 3 else sys.maxsize
         step = args[4] if len(args) > 4 else 1
 
+        # Skip if start/end/step are dynamic graph values (FX Nodes) — we
+        # can't prove the slice is identity without concrete/symbolic values
+        # to compare.
+        if (
+            isinstance(start, torch.fx.Node)
+            or isinstance(end, torch.fx.Node)
+            or isinstance(step, torch.fx.Node)
+        ):
+            continue
+
         if start != 0 or step != 1:
             continue
 
--- a/torchtitan/experiments/graph_trainer/tests/test_passes.py
+++ b/torchtitan/experiments/graph_trainer/tests/test_passes.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import operator
+import sys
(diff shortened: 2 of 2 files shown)


torchtitan/experiments/graph_trainer/remove_noop_passes.py
```diff
@@ -144,6 +144,11 @@ def remove_identity_slice_pass(
         shape = val.shape
         dim_size = shape[dim]
 
+        # Skip if end is a dynamic graph value (FX Node) — we can't prove the
+        # slice is identity without a concrete/symbolic value to compare.
+        if isinstance(end, torch.fx.Node):
```
SherlockNoMad: I think the same logic applies to start, end, and step. 

if any of these are symints, we can't tell if it's noop

torchtitan/experiments/graph_trainer/tests/test_passes.py
```diff
@@ -1075,6 +1075,32 @@ def test_chained_identity_slices(self):
         remove_identity_slice_pass(gm)
         self.assertEqual(self._count_slice_nodes(gm), 0)
 
+    def test_dynamic_end_skipped(self):
```
SherlockNoMad: remember to also update test.

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/tests/test_passes.py
aditvenk: Done

torchtitan/experiments/graph_trainer/remove_noop_passes.py
aditvenk: Done

PR #3197, opened 2026-05-01
[graph_trainer] Compose simple_fsdp with torchao FSDP hooks

aditvenk:
Stacked PRs:
 * #3198
 * __->__#3197

--- --- ---

### [graph_trainer] Compose simple_fsdp with torchao FSDP hooks

simple_fsdp distributes parameters into DTensor storage, but torchao tensor
subclasses such as FP8/MXFP8 wrappers need their FSDP pre/post all-gather hooks
to prepare communication tensors and rebuild wrapper tensors after all-gather.
Bypassing those hooks can strip wrapper semantics before the value reaches the
linear or grouped MoE op.

Detect params whose tensor, or DTensor local tensor, defines both
fsdp_pre_all_gather and fsdp_post_all_gather. Preserve the wrapper in DTensor
storage. During ReplicateComputation, run the pre-hook through an autograd
Function before the existing redistribute path, then run the post-hook after
all-gather. The pre-hook bridge keeps the autograd edge from the wrapper param
to the hook-produced plain communication tensor; the post-hook bridge restores
wrapper outputs, including DTensor outputs when TP/EP placements are preserved.

Add SimpleFSDP unit coverage for hook composition with both a plain wrapper
param and a DTensor-local wrapper param, plus the existing MXFP8 graph capture
coverage.

### DIFF
--- a/torchtitan/experiments/graph_trainer/simple_fsdp.py
+++ b/torchtitan/experiments/graph_trainer/simple_fsdp.py
@@ -8,6 +8,7 @@ import sys
 from collections.abc import Generator, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass
+from typing import Any
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


aditvenk:
Options evaluated for this change in https://www.internalfb.com/phabricator/paste/view/P2298882490

SherlockNoMad:
Do you have a sample tlparse result I can take a look at the mxfp8 enabled graph? 

you can generate one by https://www.internalfb.com/phabricator/paste/view/P2298900998

aditvenk:
> Do you have a sample tlparse result I can take a look at the mxfp8 enabled graph?
> 
> you can generate one by https://www.internalfb.com/phabricator/paste/view/P2298900998

I am still iterating on this PR. Sorry for the noise -- forgot to push as Draft. 
Will get you a tlparse on the final commit. 
I am using OSDC, so no tlparse upload manifold :( Will figure it out

aditvenk:
> Do you have a sample tlparse result I can take a look at the mxfp8 enabled graph?
> 
> you can generate one by https://www.internalfb.com/phabricator/paste/view/P2298900998

@SherlockNoMad  -- here's the tlparse: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/avenkataraman/418452CB-F64F-4242-9FF7-17FA5032D4F3/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

SherlockNoMad:
I am following the tlparse result [here](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/avenkataraman/418452CB-F64F-4242-9FF7-17FA5032D4F3/custom/rank_0/-_-_-_-/make_fx_graph_traced_1.txt?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000)

and here's what I see. 

arg104_1, a raw fp32 weight. 
first casted to bf16, as _to_copy_112. 
fsdp allgather in bf16, into all_gather_into_tensor_42, and latter sliced and viewed as slice_36 and view_357. 
view_357 is now converted into mxfp8, as triton_to_mxfp8_dim0_55. 
finally computed by aten._scaled_mm, resulting as _scaled_mm_18. 

I am noob to mxfp8, so just checking 
1. is allgather expected to be carried out in bf16?  I can also imagine allgather in mxfp8, since it's just byte movement without compute.

SherlockNoMad:
Similarly, all of the reducescatter from tlparse is performing in fp32. 
is this expected?

SherlockNoMad:
MXFP8TrainingWeightWrapperTensor is a __torch_function__ tensor subclass, DTensor is a __torch_dispatch__ tensor subclass, the problem we are running into is the eager composability between these 2 category of tensor subclass. 

I would ask around to see if there are some precedence experience on handling this class of problem. 

Currently, we are enforcing an dispatch order in simple fsdp code, but it's not trivial to which subclass kicked in first. 

From graph's result, it looks like you have an outer dtensor, and inner MXFP8 Tensor, so allgather happens first, to_mxfp8 happens latter....

This is slightly counter intuitive, if I were to reason from the natural dispatch order: torch_function happens first, torch_dispatch happens second. (but I am not sure if the reasoning applied here... )

tianyu-l:
> torchao tensor subclasses such as FP8/MXFP8 wrappers need their FSDP pre/post all-gather hooks to prepare communication tensors and rebuild wrapper tensors after all-gather.

what exactly is done in the hooks? -- why do we need to prepare communication and rebuild tensor, if the comms are all in original precision? again, tensorwise fp8 shouldn't be considered

cc @danielvegamyhre

SherlockNoMad:
I think 2 questions can clarify my confusion, and worth documenting clearly. 

1. during model init, which happens first? MX8 module swapping, or simplefspd param swapping? 
2. IIIUC, MX8 module swapping happens first, so simple fsdp param swapping would see a param in mxfp8 tensor subclass. During DTensorization, it's ignoring it's mxfp8's subclassness, thus needing the fix. 
3. The fix is to record the mxfp8 metadata before DTensorization, and after allgather, you reapply the mxfp8 metadata to DTensor's inner local_tensor. 
4. I don't know how reduce_scatter is handled in the backward...

tianyu-l (changes requested):
.

aditvenk:
Closing based on internal discussions. Will re-open after FlexShard is landed

PR #3198, opened 2026-05-01
[graph_trainer] Add MXFP8 and Float8 debugmodel configs

aditvenk:
Stacked PRs:
 * __->__#3198
 * #3197

--- --- ---

### [graph_trainer] Add MXFP8 and Float8 debugmodel configs

Add graph_trainer debug-model configs for MXFP8 on Llama 3 and DeepSeek v3, plus Float8 graph_trainer debug-model configs for Llama 3 and DeepSeek v3 EP.

Keep the DeepSeek v3 Float8 EP config local to graph_trainer: it starts from deepseek_v3_debugmodel_ep(), swaps in Float8 linear and grouped-expert converters, then enables the graph_trainer compile config. Also add 4-GPU H100 graph_trainer Float8 integration cases for AOT FX trace with and without full Inductor, and install torchao nightly in the graph_trainer H100 workflow so those cases can run.

Co-Authored with Claude

PR #3199, opened 2026-05-01
[mxfp8] Fix MXFP8GroupedExpertsConverter to actually swap GroupedExperts params

aditvenk:
Stacked PRs:
 * #3198
 * #3197
 * __->__#3199

--- --- ---

### [mxfp8] Fix MXFP8GroupedExpertsConverter to actually swap GroupedExperts params

torchao's quantize_ defaults filter_fn to _is_linear, which only matches
nn.Linear modules. GroupedExperts is not an nn.Linear, so the quantize_
call inside MXFP8GroupedExpertsConfig.build() was a silent no-op — w1/w2/w3
stayed as plain tensors and torch._grouped_mm never saw an MXFP8 wrapper,
so the captured FX graph showed plain aten._grouped_mm instead of the
quantized scaled_grouped_mm path.

Pass a custom filter_fn that matches the built GroupedExperts instance so
its nn.Parameters get swapped to MXFP8 wrapper subclasses. Now the
__torch_function__ override on _grouped_mm fires correctly and routes to
_to_mxfp8_then_scaled_grouped_mm during forward.

Co-Authored with Claude

### DIFF
--- a/torchtitan/components/quantization/mx.py
+++ b/torchtitan/components/quantization/mx.py
@@ -162,7 +162,15 @@ class MXFP8GroupedExpertsConverter(QuantizationConverter):
                         instance = base_cls.build(self, **kwargs)
                         recipe = MXFP8TrainingRecipe(recipe_name)
                         mxfp8_op_config = MXFP8TrainingOpConfig.from_recipe(recipe)
-                        quantize_(instance, config=mxfp8_op_config)
+                        # torchao's quantize_ defaults filter_fn to _is_linear,
+                        # which skips GroupedExperts. Pass a filter that matches
+                        # the built GroupedExperts instance so its nn.Parameters
+                        # (w1/w2/w3) get swapped to MXFP8 wrapper subclasses.
+                        quantize_(
+                            instance,
+                            config=mxfp8_op_config,
+                            filter_fn=lambda mod, _fqn: isinstance(mod, GroupedExperts),
+                        )
                         return instance
 
                 _converted_config_cache[base_cls] = MXFP8GroupedExpertsConfig


danielvegamyhre (approved):

aditvenk:
Before: moe has grouped_mm
Screenshot 2026-05-01 at 12 21 53 PM


After: moe has scaled_grouped_mm
Screenshot 2026-05-01 at 12 22 25 PM


This appears to have been an unintentional regression from https://github.com/pytorch/torchtitan/pull/3127

torchtitan/components/quantization/mx.py
```diff
@@ -162,7 +162,15 @@ def build(self, **kwargs):
                         instance = base_cls.build(self, **kwargs)
                         recipe = MXFP8TrainingRecipe(recipe_name)
                         mxfp8_op_config = MXFP8TrainingOpConfig.from_recipe(recipe)
-                        quantize_(instance, config=mxfp8_op_config)
+                        # torchao's quantize_ defaults filter_fn to _is_linear,
```
acisseJZhong, line 165: do we need to fix https://github.com/pytorch/torchtitan/blob/a0d66052b4996058fd0aa4bf5f130502dc3da242/torchtitan/components/quantization/float8.py#L217 similarly?

torchtitan/components/quantization/mx.py
```diff
@@ -162,7 +162,15 @@ def build(self, **kwargs):
...
+                        # which skips GroupedExperts. Pass a filter that matches
+                        # the built GroupedExperts instance so its nn.Parameters
+                        # (w1/w2/w3) get swapped to MXFP8 wrapper subclasses.
+                        quantize_(
+                            instance,
+                            config=mxfp8_op_config,
+                            filter_fn=lambda mod, _fqn: isinstance(mod, GroupedExperts),
```
acisseJZhong, line 172: for ref: this is how it works before: https://github.com/pytorch/torchtitan/blob/e0a926f4f0a81d243aecc0a5a3507aa6a5940524/torchtitan/components/quantization/mx.py#L100

torchtitan/components/quantization/mx.py
aditvenk: Good catch. I don’t check the float8 as I was focusing on mxfp8. I can send a follow up for float8

tianyu-l: This is a miss of mine.

Before the recent refactor, there was only one mxfp8 converter for both linear and grouped mm https://github.com/pytorch/torchtitan/pull/3032/changes#diff-1a1b8aa2c436eb5db983e7327f90645990e906b786ec3b7ea96d77b7050416c8L25; the refactor creates two separate converters, which seems the source of discrepancy.

If possible, I think we should still unify both under the same converter, following the pattern of `MXFP8GroupedExpertsConverter` (but rename to `MXFP8Converter`).

cc @acisseJZhong

aditvenk: I will keep this PR isolated to MXFp8 only. Filed https://github.com/pytorch/torchtitan/issues/3219 for the follow-up work

tianyu-l (approved):

PR #3203, opened 2026-05-02
Enhance Lychee Link Checker (Resiliency & Performance)

Andrei-Aksionov:
Hey there 👋 

As explained in [pytorch/torchtitan#3158](https://github.com/pytorch/torchtitan/pull/3158#issuecomment-4353494369), this PR refactors the Lychee link-checking logic to eliminate flaky CI failures while significantly improving execution speed. 

<br>

## 1. Multi-Tier Check Strategy
The "Commit-time" experience is now separated from "Infrastructure monitoring" to ensure the development flow is not blocked by external outages.

*   **Pre-commit (Local & PR CI):** Configured to **fail only on 404**. Codes like `502` (e.g., "GitHub Unicorns") are accepted to prevent transient failures from stalling the workflow.
*   **Nightly CI:** Performs a strict check (accepting only `200`, `403`, `429`, `503`) with high persistence. It retries for up to **30 minutes**, outlasting most service outages. This run populates the cache with verified statuses that later can be reused in other workflow runs.

<br>

## 2. The Cache Lifecycle: Why `key` and `restore-keys` are Necessary
To understand the necessity of a dynamic `key` and the `restore-keys` fallback, one must first recognize that **GitHub Actions caches are immutable**. 

### The Problem with Static Keys
If a static key like `key: lychee-cache` is used without any dynamic parts, the workflow encounters a "Cache Hit" lock:
1.  **The First Run:** GitHub creates the `.lycheecache` file and saves it as `lychee-cache`.
2.  **Subsequent Runs:** GitHub finds an exact match for `lychee-cache` and downloads it. Because GitHub cannot update an existing cache, any new links or status updates discovered during the run are **discarded**.
3.  **The "Stale Cache" Effect:** Over time, the cache becomes frozen in the state of the first run. Fixed links remain marked as "broken," and new links are re-checked every single time, slowing down CI.

### The Solution: Dynamic Keys & Branch Isolation
GitHub restricts cache access based on branches: the **Default Branch** (`main`) is accessible by all, while **Feature Branches** can only access their own caches or the default branch's cache. 

By using a dynamic key (e.g., `cache-lychee-${{ github.sha }}`) combined with `restore-keys`, the process moves through two distinct phases:
- **Restore** (start of job)
- **Save** (end of job).

#### Step-by-Step Lifecycle (3-Commit Example)
| Phase | Commit 1: The "Inheritance" | Commit 2: The "Update" | Commit 3: The "Iteration" |
| :--- | :--- | :--- | :--- |
| **Restore** | Misses `cache-lychee-SHA1`. Falls back to `restore-keys` to pull the latest `main` cache that matches `cache-lychee-` pattern. | Misses `SHA2`. Pulls the most recent match from the branch scope (**SHA1**). | Misses `SHA3`. Pulls the latest available version from the branch (**SHA2**). |
| **Action** | Lychee runs, checks only new links, and uses the inherited cache for the rest. | Lychee updates results with any new findings. | Lychee uses the `SHA2` baseline, ensuring zero redundant checks. |
| **Save** | GitHub saves a **new** cache entry: `lychee-cache-SHA1`. | GitHub saves a new immutable entry: `lychee-cache-SHA2`. | GitHub saves the final state: `lychee-cache-SHA3`. |

> [!NOTE]
> This "chaining" effect ensures every commit builds upon the previous one, while keeping PR runs fast and the nightly "source of truth" accessible.

<br>

## 3. Optimization: Parallelism & Output
Previously, the configuration relied on `require_serial: true` and `verbose: true`.

*   **The Problem:** `verbose: true` was required to display the "Lychee not found" warning (since the script uses `exit 0`). However, because `pre-commit` spawns a process for every file, this caused the warning to print for every single file checked. `require_serial: true` was used to stop the log spam, but caused a **2x-3x slowdown**.
*   **The Fix:** An **Atomic Sentinel** is now used via `mkdir /tmp/lychee_lock`. Because `mkdir` is an atomic operation, only the first process successfully creates the directory and prints the warning. All other parallel processes fail to create the directory and remain silent.
*   **The Result:** `require_serial` and `verbose` are now **false**. The check runs in parallel (fast), and the warning prints exactly once (clean) by redirecting directly to the user's terminal screen via `> /dev/tty`.

<br>

## 4. Fix Lychee version to install
In **Lychee v0.24.1**, a change in the release archive structure broke dynamic installation scripts that fetched the "Latest" release.
*   **Change:** The installation now uses a **fixed, verified version** to prevent upstream changes from breaking the CI pipeline.

<br>

## 5. GITHUB_TOKEN
The `GITHUB_TOKEN` is explicitly passed to the Lychee action and pre-commit steps. This increases the rate limit for GitHub API requests, reducing `403 Forbidden` and `429 Too Many Requests` errors when checking internal repository links.

<br>

## Summary Overview
| Feature | Local / PR CI | Nightly CI |
| :--- | :--- | :--- |
| **Failure Condition** | Only `404` | Most non-200 codes |
| **Duration/Retries** | Fast (5 retries / 15 secs) | Patient (30 retries / 30 mins) |
| **Execution** | Parallel (via Sentinel) | Standard Action |
| **Cache Goal** | Consume & Increment | Refresh & Validate |


.gitignore
```diff
@@ -49,3 +49,6 @@ CLAUDE.local.md
 /debug_*.py
 CLAUDE_CONTEXT/
 /.claude/settings.local.json
+
+# Lychee (link checker) cache
+.lycheecache
```
Andrei-Aksionov, line 54: Lychee supports cache placement only in the root folder.

tianyu-l:
@fegin 
what do you think of https://github.com/pytorch/torchtitan/pull/3158#issuecomment-4353494369 and this PR?

fegin:
@Andrei-Aksionov If we actually fix the transit issue (I noticed that you accept some comment transient errors), should we move the check back to per-PR CI?

Andrei-Aksionov:
> ... should we move the check back to per-PR CI?

TL;DR: it does exactly that already.

---

This PR doesn’t fix transient network failures, but it makes the link-checking process significantly more resilient to them.

In this PR, Lychee checks both internal and external links during pre-commit (local/per-PR CI), but unlike the initial implementation, it only fails on `404`. The risk with this approach is that a `404` could technically be hidden behind a `502`: if a server is unresponsive, it cannot confirm whether a link is actually broken.

To mitigate this, I modified the nightly link check. It runs with a much higher retry limit to outlast most outages and will fail on 502s. Both the pre-commit (local/PR) and the nightly runs utilize `.lycheecache`. Because they share this cache, the nightly run "populates" the status of external links, which the per-PR CI then reuses. This ensures that the cache contains the correct URL status code.

Is this a slight over-complication? Probably 🙂 
While I like how these pieces work in symbiosis, we could potentially drop the nightly run without the experience degrading too much. 

(I didn't actually create the nightly action (it was added in a previous PR), I just updated it to function correctly within this new architecture.)

fegin:
Oh, yes, I understand the intention of the PR. I guess I didn't express it clearly. What I meant is that if already make it resilient to these errors, should we just drop the nightly check and **restore** the per-PR check?

Andrei-Aksionov:
As I wrote above there is a benefit of having a nightly check in the current setup, but overall, for simplicity, we can easily drop the nightly run.

@fegin Since there is a significant difference in timezones (I am in UTC+3), you can delete the nightly workflow. Everything else will work without problems.

Andrei-Aksionov:
@fegin I removed the nightly link-checking workflow.
Now we have only pre-commit link checking that runs every:
- `git commit` command (if pre-commit is installed)
- every PR via `lint` workflow

Basically the same as in the original implementation, but now:
- Fails only on `404`
- Uses `.lycheecache` to make link-checking faster and more resilient
- A couple of changes in the config of the pre-commit to make it a little bit faster

----

For a context: the current code in the main also runs per-PR CI check, but only kept hyperlinks check in CI and hyperlinks/external links check for a local pre-commit run.

fegin (approved):
Thanks for the fix!

PR #3159, opened 2026-04-29
[Full DTensor] Config-based Full DTensor for Llama3

fegin:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* __->__ #3159

Add full DTensor mode where all params, buffers, and inputs are DTensors on a multi-dimensional SPMD mesh (``dp_replicate`` x ``dp_shard`` x ``cp`` x ``tp``). Enabled via ``--training.full_dtensor``. CP is handled declaratively via ``LocalMapSpec`` (K/V ``Replicate`` on CP dim for all-gather) instead of ``apply_cp_to_forward`` hooks. FSDP2 uses ``DataParallelMeshDims`` to identify ``dp_replicate``,  ``dp_shard`` and ``cp`` as data-parallel dims on the multi-D mesh.

**Mesh Design:**

``ParallelDims`` precomputes one ``DeviceMesh`` per SPMD partition in ``_spmd_meshes`` (dense and sparse). Each ``Module`` self-resolves its mesh: ``_needed_axes(sharding_config)`` collects axes referenced by its ``NamedPlacement``s, then ``parallel_dims.get_module_mesh(axes)`` returns the cached submesh. Under ``full_dtensor``, all axes participate; under the legacy, non-full-dtensor path, axes are filtered to ``{tp, ep, etp}`` so DP and CP stay out-of-band. ``Module.parallelize`` asserts the resolved mesh is one of ``spmd_meshes()`` - out-of-order axes resolve to a different submesh and fail the check, enforcing the canonical axis order.

The same mechanism extends to sparse: a second SPMD partition  lives alongside dense in ``_spmd_meshes``, so MoE modules whose ``sharding_config`` references sparse axes resolve to the sparse submesh through the exact same ``get_module_mesh`` call - no model-side branching.

**Verified parallelisms:**

FSDP + TP
FSDP
HSDP
FSDP + TP
HSDP + TP
FSDP + CP + TP
PP + FSDP + TP
PP + FSDP + CP
PP + CP + TP

Note that: DP + CP + torch.compile fail.

**Issues**

  1. **`view()` rejects `Shard()` on a size-1 tensor dim:** Hit by PP-only and TP-only with bs=1 because we keep `dp_shard` in the mesh; this PR converts `Shard()` to `Replicate()` on size-1 mesh axes in `resolve_placements`.

[...]

PR #3204, opened 2026-05-04
fix(dist_reduce): run mesh all_reduce when DTensor mesh is orthogonal

saforem2:
Since #2963, `_dist_reduce` short-circuits all DTensor inputs with `return float(x.full_tensor().item())`, skipping the requested mesh all_reduce entirely. The justification in the comment ("avoid double-counting") only holds when the DTensor's own mesh covers the requested reduction mesh; when they are orthogonal — as in the trainer's loss reduction where the loss is a Replicated DTensor on the TP mesh and the requested mesh is `loss_mesh = batch x cp` — the cross-batch sum is silently dropped.

Symptom: every TP > 1 run reports `loss = true_loss / dp_world_size`. For a 2B run at TP=2 with 24 ranks, step-1 loss is 1.08 instead of the ~12.95 ln(vocab) baseline observed at TP=1. Gradients and optimizer steps are unaffected because backward goes through `loss.backward()` not `_dist_reduce` — only the printed/logged loss is wrong.

Fix: detect mesh axis overlap via `mesh_dim_names`. If the DTensor's mesh shares any axis with the requested mesh, `full_tensor()` already covers that axis and we skip the explicit all_reduce as before. If the meshes are disjoint, we materialize via `full_tensor()` and fall through to the plain-tensor path so the requested-mesh all_reduce runs.

Adds a unit test covering both the orthogonal case (was broken) and the same-mesh case (must not double-count).

### DIFF
--- a/torchtitan/distributed/utils.py
+++ b/torchtitan/distributed/utils.py
@@ -45,23 +45,37 @@ def _dist_reduce(
             process group, and then the result will be all_reduced for the mesh.
     """
     if isinstance(x, DTensor):
-        # DTensor path: ``full_tensor()`` already performs the mesh reduction
-        # for Partial placements and is a no-op for Replicate. Skipping the
-        # subsequent mesh all-reduce is required to avoid double-counting.
+        # DTensor path. ``full_tensor()`` reduces across the DTensor's own
+        # mesh (Partial → reduced; Replicate → no-op) and returns a plain
+        # tensor with the full value on every rank.
+        #
         # Shard placements are not supported — semantics are undefined since
         # the reduction target is ambiguous.
         assert all(p.is_replicate() or p.is_partial() for p in x.placements), (
             f"_dist_reduce received a DTensor with unsupported placements "
             f"{x.placements}; only Replicate/Partial are supported."
         )
-        if extra_pg is not None:
-            raise ValueError(
-                "_dist_reduce does not support DTensor input combined with "
-                "extra_pg: ``full_tensor()`` already reduces over the DTensor's "
-                "mesh, and extra_pg (e.g. the FT replica group) is orthogonal "
-                "to that mesh. Pass a plain tensor when using extra_pg."
-            )
-        return float(x.full_tensor().item())
+
+        # If the requested reduction mesh is the same mesh the DTensor
+        # already lives on, ``full_tensor()`` has done the reduction and
+        # we must skip the subsequent all_reduce to avoid double-counting.
+        # Otherwise (the requested mesh is orthogonal to the DTensor's mesh,
+        # e.g. reducing a TP-Replicated loss across batch_mesh), we still
+        # need to reduce on the requested mesh — fall through to the
+        # plain-tensor path with the materialized full tensor.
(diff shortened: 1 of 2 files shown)


torchtitan/distributed/utils.py
```diff
@@ -45,23 +45,37 @@ def _dist_reduce(
...
+        # If the requested reduction mesh is the same mesh the DTensor
+        # already lives on, ``full_tensor()`` has done the reduction and
+        # we must skip the subsequent all_reduce to avoid double-counting.
+        # Otherwise (the requested mesh is orthogonal to the DTensor's mesh,
+        # e.g. reducing a TP-Replicated loss across batch_mesh), we still
+        # need to reduce on the requested mesh — fall through to the
+        # plain-tensor path with the materialized full tensor.
+        x_mesh = x.device_mesh
+        x = x.full_tensor()
+        if mesh is None or _meshes_share_axes(x_mesh, mesh):
```
tianyu-l, line 68: 1. do we need to handle the `mesh is None` specially here?
2. when `_meshes_share_axes` is True, it doesn't mean the result is correct? but you would just return. I think a workaround is to just error out when two meshes are not identical.

fegin, line 68: I want to argue that the only clean workaround is to_local(). Then we don't need to infer whether it is full dtensor or non-full dtensor + TP. The actual fix is to deprecate the non-full dtensor case. Then we will unconditionally apply full_tensor() here. But given that it is hard to do deprecation without full dtensor being ready, we should just make it a plain tensor, which was what we always did for the TP case before this April -- we changed the TP-enabled model to always use DTensor for activation and loss this April.

I fixed this issue in https://github.com/pytorch/torchtitan/pull/3159. If you want to land with this PR, I prefer landing `to_local()`.

cc., @tianyu-l

fegin (changes requested):
I prefer using `to_local()` for now to fix the issue: https://github.com/pytorch/torchtitan/pull/3159.

saforem2:
Closing as superseded.

The orthogonal-mesh under-reporting was fixed by @fegin in #3159
(commit `d64eabcce`, merged 2026-05-18) — switching the DTensor
branch of `_dist_reduce` from `full_tensor()` to `to_local()` + the
explicit mesh `all_reduce`. From that PR's issue list:

> Our `_dist_reduce` under-reports loss for non full tensor:
> `full_tensor()` reduces only over the DTensor's own mesh (TP-only
> in legacy), skipping DP. Loss was under-reported by
> `dp_shard × cp`. Switched to `to_local()` then mesh all_reduce
> over loss_mesh.

That's the same case the #3204 tests covered (Replicate-on-TP loss
reduced across an orthogonal batch mesh). Verified on current
`upstream/main`: `_dist_reduce` now does `x = x.to_local()` for the
DTensor path, then runs the requested `funcol.all_reduce` —
no double-count, no skipped reduction.

@fegin / @tianyu-l — thanks for the review and pointer to #3159.
The `to_local()` shape is cleaner than the `full_tensor()` +
mesh-overlap helper I had here, agreed.

PR #3215, opened 2026-05-04
[GraphTrainer] Add memory policy registry and extra graph passes hook

syed-ahmed:
* #3218
* #3217
* #3216
* __->__ #3215

- Add AVAILABLE_MEMORY_POLICIES registry to passes.py so experiments can
  register custom memory policies without modifying graph_trainer code.
- Refactor tag_with_memory_policy_pass to dispatch via the registry.
- Change memory_policy config from Literal to str for extensibility.
- Add _get_extra_graph_passes hook to GraphTrainer for subclasses to
  inject additional graph passes into the aot_fx_trace pipeline.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -44,14 +44,19 @@ class GraphTrainerCompileConfig(CompileConfig):
...
+    Memory optimization policy for activation management (SAC, offload, etc.).
+
+    Built-in policies:
         default: save all compute-intensive ops and FSDP all_gathers.
         eager: alternate mm ops between save/recompute, matching the eager
             AC policy in torchtitan.distributed.activation_checkpoint.
         cpu_offload_all: offload all eligible activations to CPU.
             Work in progress — for development and testing only.
+
+    Experiments can register additional policies via
```
SherlockNoMad: Users can

SherlockNoMad: since we are not adding an experiment.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -1017,3 +1041,14 @@ def tlparse_log_graph_pass(
     "fsdp_reshard_after_fwd": fsdp_reshard_after_fwd_pass,
     "apply_sac": apply_sac_pass,
 }
+
+# Registry for memory policies dispatched by tag_with_memory_policy_pass.
+# Experiments can register custom policies without modifying this file:
+#   from torchtitan.experiments.graph_trainer.passes import AVAILABLE_MEMORY_POLICIES
+#   AVAILABLE_MEMORY_POLICIES["my_policy"] = my_policy_fn
+# Each policy is a callable: (gm, *, config) -> gm
+AVAILABLE_MEMORY_POLICIES: dict[str, Callable] = {
```
SherlockNoMad: Nice. 

Now I think all memory policy related pass can be its own memory_policy.py ? 

Also a nice pattern for registry is 

```
@register_memory_policy(name="default")
def _default_memory_policy_pass()
··· 

Neither comments are blocking tho, we can do it as follow up

File: torchtitan/experiments/graph_trainer/trainer.py [hunk 3]
```diff
@@ -132,6 +133,15 @@ def _load_precompiled_fx_trace(self, model: nn.Module) -> None:
             expected_fingerprint=config_fingerprint,
         )
 
+    def _get_extra_graph_passes(self, model: nn.Module) -> list[tuple]:
```
SherlockNoMad: now that we are not doing this as an experiment, I wonder if we still need this?

I think we will eventually need an API to customize graph passes. 

insert_before works, but kinda hacky... so we would prefer we don't do this. 

A more proper way to customize graph pass is to just directly construct the passes, and register it via registry. 

e.g.

@register_pass_pipeline("default") 
def compile_time_passes(...)

SherlockNoMad (changes requested):
not fan of _get_extra_graph_passes

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -72,6 +77,35 @@
...
 
 
+# ---------------------------------------------------------------------------
+# Registries — keyed by memory policy config type
+# ---------------------------------------------------------------------------
+
+MEMORY_POLICY_REGISTRY: dict[type, Callable] = {}
+PASS_PIPELINE_REGISTRY: dict[type, Callable] = {}
+POST_INIT_HOOKS: dict[type, Callable] = {}
+PRE_TRAIN_STEP_HOOKS: dict[type, Callable] = {}
```
SherlockNoMad: I suppose these hooks are used in the latter PRs.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -72,6 +77,35 @@
...
+
+
+def _make_registry_decorator(registry: dict):
```
SherlockNoMad, line 80: nice

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -15,6 +17,63 @@
...
+    """Eager SAC: alternate mm ops between save/recompute."""
+
+    pass
+
+
+@dataclass(kw_only=True, slots=True)
+class CpuOffloadAllMemoryPolicy:
+    """Offload all eligible activations to CPU (WIP)."""
+
+    pass
```
SherlockNoMad: Wondering what are the benefit of declaring a class over "plain string"？ is it for tyro?

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -341,9 +341,9 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
                     "--module graph_trainer.llama3",
                     "--config graph_trainer_llama3_debugmodel",
                     "--compile.mode aot_fx_trace",
-                    "--compile.memory_policy cpu_offload_all",
                     "--parallelism.data_parallel_shard_degree 4",
                     "--parallelism.tensor_parallel_degree 2",
+                    "compile.memory-policy:cpu-offload-all",
```
SherlockNoMad: hmm? 

is this the new format for comment line arg? don't need -- anymore?

you might need to do a global replace for all use of compile.memory_policy in graph trainer

torchtitan/experiments/graph_trainer/configs.py
```diff
@@ -15,6 +17,63 @@
...
+    def decorator(cls: type) -> type:
+        _MEMORY_POLICY_TYPES.append(Annotated[cls, tyro.conf.subcommand(name)])
+        # Patch the annotation on GraphTrainerCompileConfig so tyro sees the
+        # updated Union when it parses CLI args.
+        GraphTrainerCompileConfig.__annotations__["memory_policy"] = Union[
+            tuple(_MEMORY_POLICY_TYPES)
+        ]
+        return cls
+
+    return decorator
```
SherlockNoMad: hmm... seems complicated just for CLI support.... 
anything wrong with just plain string?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -1072,3 +1137,9 @@ def tlparse_log_graph_pass(
     "fsdp_reshard_after_fwd": fsdp_reshard_after_fwd_pass,
     "apply_sac": apply_sac_pass,
 }
+
+# Import paged stash registrations (registers additional policies + pipelines + hooks)
+try:
+    import torchtitan.experiments.graph_trainer.paged_stash_memory_policy  # noqa: F401, E402
```
SherlockNoMad: i suppose these are added in the future PRs.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -169,10 +177,12 @@ def _make_fx_forward_backward_step(
                     )
 
             if self.config.compile.enable_passes:
-                passes = construct_default_graph_passes(
-                    self._traced_step,
-                    self.config,
+                policy_type = type(self.config.compile.memory_policy)
+                pipeline_fn = PASS_PIPELINE_REGISTRY.get(
+                    policy_type, construct_default_graph_passes
                 )
```
SherlockNoMad: the key here won't be memory_policy, it should be something general, and up for user to define.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -200,6 +210,13 @@ def _make_fx_forward_backward_step(
 
         return loss
 
+    def train_step(
+        self, data_iterator: Iterator[tuple[dict[str, torch.Tensor], torch.Tensor]]
+    ):
+        policy_type = type(self.config.compile.memory_policy)
```
SherlockNoMad: same here, it's limiting to use memory_policy as key here. 
This should allow arbitrary used-define type, or just plain string.

SherlockNoMad (commented):
I think this mostly looks good, and I am signing this off so that you don't need another round of back-and-forth. 

1. I am not sure if we need to used type as the key for registry, str is simple is probably good enough. 
2. keys for pass_pipeline, and hooks shouldn't be memory_policy. It should support general user defined keys, i.e. string.

SherlockNoMad (approved):
I think this mostly looks good, and I am signing this off so that you don't need another round of back-and-forth. 

1. I am not sure if we need to used type as the key for registry, str is simple is probably good enough. 
2. keys for pass_pipeline, and hooks shouldn't be memory_policy. It should support general user defined keys, i.e. string.

torchtitan/experiments/graph_trainer/passes.py
syed-ahmed: Yes.

torchtitan/experiments/graph_trainer/trainer.py
syed-ahmed: Add a pass_pipeline key instead if that sounds ok?

torchtitan/experiments/graph_trainer/trainer.py
syed-ahmed: Same here, is a pass_pipeline key ok instead?

SherlockNoMad (approved):
Thanks!

PR #3216, opened 2026-05-04
[Paged Stash] CUDA-graphable MoE with paged activation stashing

syed-ahmed:
* #3218
* #3217
* __->__ #3216

Paged stash SAC for MoE expert activations: pre-allocated paged buffers
store dynamic-shaped _grouped_mm outputs for the backward pass, avoiding
both recomputation cost and memory fragmentation from variable token counts.

- Joint-graph pass (paged_stash_graph_pass.py): inserts paged_stash.copy/pop
  ops using ao's stream infrastructure, redirecting backward consumers through
  paged buffers so SAC+remat naturally ignores stashed activations.
- PagedStashTrainer (train.py): extends GraphTrainer with 3-level overflow
  defense (host spillover, cross-rank detection, retry with buffer growth).
- Memory policy registration: paged_stash and paged_stash_save_only policies.
- Fine-grained annotation via annotate_fn on _run_experts_grouped_mm.
- Comparison with Megatron's implementation here: https://gist.github.com/syed-ahmed/f82d8485007eef91e90f08f24607d940

AI assistant used: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


torchtitan/distributed/deepep/hybridep.py
```diff
@@ -507,9 +514,25 @@ def combine_tokens(
     )
 
 
+def check_hybridep_over_budget() -> torch.Tensor:
+    """Check if HybridEP dispatch exceeded capacity budget (tokens dropped).
+
+    Returns a GPU bool tensor.  Safe to call after the training step completes
+    (the .item() D2H sync is acceptable post-step).
```
xmfan, line 557: for graph trainer, training step includes optimizer, so weights are already updated by the time this is checked, maybe we can do the checks on device for then

torchtitan/experiments/cuda_graphable_moe/configs.py
```diff
@@ -0,0 +1,94 @@
...
+    base_config: Trainer.Config,
+):
+    """Convert a base Trainer.Config to a PagedStashTrainer.Config.
+
+    Copies all fields from the base config. The compile and
+    activation_checkpoint fields are removed (they have different types
+    in PagedStashTrainer.Config) and left as defaults; callers should
+    explicitly set them via ``_apply_aot_fx_trace_defaults``.
+    """
+    from torchtitan.experiments.cuda_graphable_moe.train import PagedStashTrainer
```
xmfan: why don't we move the config class here to avoid the circular import

torchtitan/experiments/cuda_graphable_moe/configs.py
```diff
@@ -0,0 +1,94 @@
...
+
+    d = {f.name: getattr(base_config, f.name) for f in fields(base_config)}
+    d.pop("compile")
+    d.pop("activation_checkpoint")
+
+    # aot_fx_trace traces fwd+loss+bwd with torch.autograd.grad which requires
+    # all parameters to participate in forward. ChunkedCELoss sets
+    # _skip_lm_head=True making lm_head.weight unused in the traced graph.
+    if isinstance(d.get("loss"), ChunkedCELoss.Config):
+        d["loss"] = CrossEntropyLoss.Config()
```
xmfan: would recommend this to be configured explicitly or at least to fail loudly

SherlockNoMad:
I will review when this is in graph_trainer

torchtitan/experiments/cuda_graphable_moe/paged_stash_graph_pass.py
```diff
@@ -0,0 +1,580 @@
...
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _is_moe_fc1_grouped_mm(node: fx.Node) -> bool:
+    """Check if a node is a MoE fc1 (gate or up) ``_grouped_mm`` output.
+
+    Matches ``aten._grouped_mm.default`` nodes that are part of the SwiGLU
```
xmfan: i think it will be _scaled_grouped_mm for ao mxfp8

torchtitan/experiments/cuda_graphable_moe/paged_stash_graph_pass.py
```diff
@@ -0,0 +1,580 @@
...
+
+    Returns ``(bwd_consumers, buffer)`` if eligible, or ``None``.
+
+    A node is eligible when all of:
+    1. It is a forward ``call_function`` node.
+    2. It is an ``aten._grouped_mm`` whose consumer is ``silu`` or ``mul``
+       (MoE fc1 gate/up projection, identified by op-target + consumer pattern).
+    3. It has real (non-sym) backward consumers.
+    4. It has a dynamic SymInt first dimension.
+    5. Its ``(dtype, shape[-1])`` matches a pre-allocated paged buffer.
```
xmfan: I'm not sure i understand the need for pattern matching. Naively I would of thought to grab all nodes between dispatch/combine nodes, then to check which ones are activations (used outside of those nodes). And any activations having an unbacked shape due to received number of tokens should be paged

xmfan: we could rely on annotations instead too

syed-ahmed: Yeah, I'm totally open to what to do here. I was trying to match exactly the activations Megatron applies paged stashing to (expert_fc1, moe_act, expert_fc2). Note that, megatron uses Swiglu fused grouped mlp: https://github.com/NVIDIA/TransformerEngine/pull/2769, so the approach we use - pattern matching vs annotation - can affect how many activations get picked up for paged stashing. I guess The main filter is unbacked shape, which should decide whether it should be paged stashed or not.

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1383 @@
...
+            "aot_fx_trace mode: paged stash pass will be applied at trace time"
+        )
+
+
+@register_pre_train_step_hook(PagedStashMemoryPolicy)
+def reset_paged_stash_buffers(trainer):
+    """Check previous step's overflow flag, then reset buffers for this step."""
+    for model_part in trainer.model_parts:
+        overflow = getattr(model_part, "_paged_stash_overflow", None)
+        if overflow is not None and overflow.item() != 0:
```
IvanYashchuk: This check looks too late for correctness. The copy kernel can set overflow during the current forward/backward, and the comment above `page_record` says the data may be wrong and the step will be retried. But this hook runs at the start of the next `train_step`, after the previous step may have already run backward and updated optimizer state with corrupted gradients.

Can we move this check before optimizer update and make it collective across ranks? If the intended behavior is retry, I think we need an explicit skip/retry path before optimizer state changes. If this is fail fast, the comment about overflow gradients being thrown away should change.

torchtitan/distributed/deepep/hybridep.py
```diff
@@ -507,9 +514,25 @@ def combine_tokens(
     )
 
 
+def check_hybridep_over_budget() -> torch.Tensor:
```
IvanYashchuk, line 553: This helper records an important correctness condition, but I do not see it wired into GraphTrainer or the paged-stash policy. In non-blocking mode, HybridEP can silently drop tokens when the capacity budget is exceeded. The dispatcher docs say the overflow flag is set on GPU, but this PR does not appear to enforce it before optimizer update.

Can we check this in the same pre-optimizer path as paged-stash overflow? If reduced capacity factors are only meant for debugging, I think we should require an explicit "allow dropped tokens" flag rather than making silent token drops possible in a training run.

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1383 @@
...
+
+# ---------------------------------------------------------------------------
+# paged_stash::copy -- pack a tensor into the paged buffer (CUDA or host)
+# ---------------------------------------------------------------------------
+
+
+@torch.library.custom_op(
+    "paged_stash::copy",
+    mutates_args=(),
+)
```
IvanYashchuk, line 700: Do we need to make the free-list mutation explicit in the graph? `paged_stash.copy` and `paged_stash.pop` both mutate global buffer/free-list state through `_PAGED_STASH_REGISTRY`, but the custom op schema says `mutates_args=()`, and the returned head/tail state is not threaded into later copy/pop ops.

Page allocation order is correctness-critical. Can we either thread an explicit state token/free-list state through the graph, or mark/model these ops as side-effectful in a way AOT partitioning and later graph passes respect? At minimum, I think we need a test that inspects the transformed graph and replays multiple layers to prove copy/pop ordering survives the full pass pipeline. @kshitij12345, could you please review this part? What are the best practices for mutation and new custom_ops?

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1383 @@
...
+# ===========================================================================
+# Paged stash graph pass
+# ===========================================================================
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _is_moe_fc1_grouped_mm(node: fx.Node) -> bool:
```
IvanYashchuk, line 910: This matcher seems narrower than the PR description. It picks bf16 `aten._grouped_mm.default` fc1 gate/up patterns by consumer shape, but not the SiLU output, the post-SwiGLU product, fc2-related activations, or `_scaled_grouped_mm` for AO/MXFP8. The buffer sizing also looks broader than the two grouped-mm outputs that are currently matched.

Can we define the exact set this policy owns and add a graph-pass test that asserts which nodes are stashed for the current bf16 path? Longer term, annotations around the expert region may be easier to maintain than consumer-pattern matching, especially for fused TransformerEngine/AO paths.

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1383 @@
...
+# ---------------------------------------------------------------------------
+# Module-level buffer registry -- accessed by custom ops at runtime
+# ---------------------------------------------------------------------------
+# Following the ao::offload pattern (which uses module-level _transfer_streams
+# and _wait_registry dicts), paged stash buffers are accessed through this
+# registry keyed by buffer ID.  The graph only carries the buffer_id as an
+# integer constant -- no buffer tensors appear as graph arguments, get_attr
+# nodes, or placeholders.  This avoids deepcopy and DTensor issues entirely.
+
+_PAGED_STASH_REGISTRY: dict[int, "PagedStashBuffer"] = {}
```
IvanYashchuk, line 196: Can we add an explicit cleanup path for this registry? `_PAGED_STASH_REGISTRY` and `_PAGED_STASH_BUFFERS_DICT` hold large CUDA/host buffers in module-level global state, and `unregister_paged_stash_buffer()` exists but is not used. That is probably okay for one trainer per process, but brittle for tests, repeated trainer construction, or long-running processes that switch configs.

I would either clear the paged-stash registry from `GraphTrainer.close()` for this policy, or document the single-trainer lifetime assumption and add a small test that constructs/closes two paged-stash trainers in one process without stale buffers.

IvanYashchuk (changes requested):
I like the direction of this PR. It matches the production path we have been seeing for MoE graphability: avoid conditional graph nodes, make the dispatch/expert path graph-compatible, and use paged storage to avoid worst-case activation memory across layers.

The parts I would fix before relying on this for real training experiments are correctness-related. Paged-stash overflow and HybridEP non-blocking capacity overflow both need to be caught before optimizer state is updated, and probably the check should be collective across ranks. Also, the mutation should be explicit enough that later graph passes cannot reorder or drop it.

The integration test is useful, and top-of-stack CI being green is a good signal. I would still add focused tests for graph surgery, copy/pop ordering, overflow-before-optimizer behavior, and multi-rank overflow before treating this as more than an experimental path.

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
xmfan: so in the graph trainer, we will want to have the whole training step including the optimizer in a single fx/cuda graph. if there's no way to delay the cpu sync from the .item check, we could consider having a device-side check

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+    if _PAGED_STASH_BUFFERS_DICT is not None:
+        bound_pass = functools.partial(
+            paged_stash_pass, paged_buffers=_PAGED_STASH_BUFFERS_DICT
+        )
+        new_passes = []
+        for p in passes:
+            _fn = p.func if isinstance(p, functools.partial) else p
+            if _fn is selective_activation_remat_pass:
+                new_passes.append(bound_pass)
+            new_passes.append(p)
```
SherlockNoMad: How is the compatibility between paged_stash_pass and apply_cpu_offload_pass? 

Probably need some pre-caution here

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+) -> torch.fx.GraphModule:
+    """Paged stash pass for the ``aot_fx_trace`` compile-time pass list.
+
+    Inserts ``paged_stash.copy/pop`` + ``ao.wait_tensor`` ops for eligible
+    activations.  Must run **before** ``selective_activation_remat_pass`` so
+    that backward consumers are already redirected when SAC + remat runs --
+    the original activations have no backward users and are naturally ignored
+    by rematerialization.
+    """
+    return apply_paged_stash_pass(gm, paged_buffers)
```
SherlockNoMad, line 1363: This layer of wrapping doesn't seem necessary? 

you can directly make apply_paged_stash_pass comply with the graphpass signature

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+    def buffers_dict(self) -> dict | None:
+        """Return the buffers dict for use in the graph pass."""
+        return self.buffers
+
+
+# ---------------------------------------------------------------------------
+# Module-level state for pass pipeline access
+# ---------------------------------------------------------------------------
+
+_PAGED_STASH_BUFFERS_DICT: dict | None = None
```
SherlockNoMad, line 178: have a word about the basic structure of this dict? 
type annotation is also helpful

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+def apply_paged_stash_pass(
+    gm: torch.fx.GraphModule,
+    paged_buffers: dict[tuple[torch.dtype, int], "PagedStashBuffer"],
+) -> torch.fx.GraphModule:
+    """Insert paged_stash.copy/pop + ao.wait_tensor into the joint graph.
+
+    1. Classify forward vs backward nodes using ``autograd_backward``.
+    2. For each eligible forward node (fc1 _grouped_mm, dynamic, has bwd consumers):
+       a. Insert ``paged_stash.copy`` + ``ao.wait_tensor`` after the fwd node.
+       b. Insert ``paged_stash.pop`` + ``ao.wait_tensor`` before bwd consumers.
```
SherlockNoMad, line 1091: cc @mlazos 

FYI, this is using your ao.wait_tensor op.

SherlockNoMad: wait_tensor recently added some "free after last use" logic, I wonder if it would affect your impl here

SherlockNoMad: see https://github.com/pytorch/pytorch/pull/181938

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+       c. Redirect backward consumers to read from the pop output.
+    3. After this pass, SAC's remat sees ``page_record`` (small int64) crossing
+       the boundary, not the large activation.
```
SherlockNoMad, line 1094: ok, currently apply_cpu_offload_pass runs right before selective_activation_remat_pass. 

and this pass also is positioned to run before selective_activation_remat_pass. 

In theory, if apply_cpu_offload_pass and apply_paged_stash_pass are targeting different nodes, it's would be fine. 

I think somewhere in this pass, you want to assert that the node you are stashing is doesn't have MUST_CPU_OFFLOAD tag.

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+                    buf.page_size,
+                    buf.hidden_size,
+                    actual_num_tokens,
+                    buffer_id,
+                ),
+            )
+            copy_node.meta["val"] = (page_record_fake, new_head_fake)
+            # MUST_SAVE so the partitioner doesn't treat it as impure
+            copy_node.meta["recompute"] = CheckpointPolicy.MUST_SAVE
+            copy_node.meta["ac_graph_id"] = 0
```
SherlockNoMad, line 1244: ac_graph_id is deprecated.

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+                    fwd_node,
+                    buf.page_size,
+                    buf.hidden_size,
+                    actual_num_tokens,
+                    buffer_id,
+                ),
+            )
+            copy_node.meta["val"] = (page_record_fake, new_head_fake)
+            # MUST_SAVE so the partitioner doesn't treat it as impure
+            copy_node.meta["recompute"] = CheckpointPolicy.MUST_SAVE
```
SherlockNoMad, line 1243: partitioner is not used in the aot_fx_trace path, so not sure if you still need this workaround

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+            copy_node.meta["ac_graph_id"] = 0
+            copy_node.meta["autograd_backward"] = False
```
SherlockNoMad, line 1245: non existence of autograd_backward means forward node,

so this is not wrong, but you probably don't need this line

SherlockNoMad: so all occurance of node.meta["autograd_backward"] = False is not needed

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+    Args:
+        must_stash_action: What to do with MoE fc1 ``_grouped_mm`` nodes.
+            ``PREFER_RECOMPUTE`` when the paged stash pass will redirect
+            backward consumers (the activation has no backward users after
+            surgery, so remat DCEs it).
+            ``MUST_SAVE`` for the baseline experiment that saves expert
+            activations as regular tensors (no paged stash).
+    """
+    save_ops = _get_save_ops()
+    save_ops.add(torch.ops._c10d_functional.all_gather_into_tensor.default)
```
SherlockNoMad, line 1385: not sure if you wanna have this... 
this would save all AG outputs, but in some cases, you wanna recompute AG, e.g. fsdp_reshard_after_forward

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+
+    def policy_fn(node: fx.Node) -> CheckpointPolicy:
+        if _is_moe_fc1_grouped_mm(node):
+            return must_stash_action
```
SherlockNoMad, line 1389: ah... nice, so this will never conflict with MUST_CPU_OFFLOAD. 

Now I wonder if you wanna extend CheckpointPolicy to have your own enum for stash. you are currently using recompute as a placeholder?

torchtitan/distributed/deepep/hybridep.py
```diff
@@ -217,23 +227,20 @@ def _dispatch_fake(
     topk_idx: torch.Tensor,
     topk_weights: torch.Tensor,
     num_experts: int,
-    non_blocking: bool = False,
-    moe_expert_capacity_factor: float | None = None,
-    pad_multiple: int | None = None,
+    non_blocking: bool,
+    moe_expert_capacity_factor: float | None,
+    pad_multiple: int | None,
```
crcrpar: is it fine to remove these defaults? I mean wouldn't this break some callers out there?

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+        page_size: int = 64
+        """Number of tokens per page in the paged stash buffer."""
+
+    def __init__(
+        self,
+        config: "PagedStash.Config",
+        *,
+        model: torch.nn.Module,
+        max_tokens: int,
+        capacity_factor: float | None,
```
crcrpar, line 63: what's the behavior when `capacity_factor=None`?

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+
+        host_buffer_size_factor: float = 0.0
+        """Factor for host (pinned CPU) spillover buffer. 0 = no host buffer."""
+
+        page_size: int = 64
+        """Number of tokens per page in the paged stash buffer."""
+
+    def __init__(
+        self,
+        config: "PagedStash.Config",
```
crcrpar: I'm not 100% sure but `from __future__ import annotations` would allow 
```suggestion
        config: PagedStash.Config,
```

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+from dataclasses import dataclass
+
+import torch
+
+# Side-effect import: registers ao::wait_tensor custom op
+import torch._functorch._activation_offloading.offload_ops as offload_ops  # noqa: F401
+import torch.fx as fx
+import triton
+import triton.language as tl
+from torch import Tensor
```
crcrpar, line 20: If this import is solely for type annotations, then I think it'd be clearer by doing something like
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from torch import Tensor
```

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+
+@torch.library.custom_op(
+    "paged_stash::copy",
+    mutates_args=(),
+)
+def paged_stash_copy(
+    tensor: Tensor,
+    page_size: int,
+    hidden_size: int,
+    num_tokens_tensor: Tensor,
```
crcrpar, line 705: above we use `torch.Tensor` as type annotations but here just bare `Tensor` so a consolidation would be great

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+
+    return page_record, new_free_list_head
+
+
+@paged_stash_copy.register_fake
+def paged_stash_copy_fake(
+    tensor: Tensor,
+    page_size: int,
+    hidden_size: int,
+    num_tokens_tensor: Tensor,
```
crcrpar, line 793: ditto, as for `Tensor` or `torch.Tensor`

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+    "paged_stash::pop",
+    mutates_args=(),
+)
+def paged_stash_pop(
+    page_record: Tensor,
+    page_size: int,
+    hidden_size: int,
+    dtype: torch.dtype,
+    buffer_id: int,
+) -> Tensor:
```
crcrpar, line 820: ditto

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+
+
+@paged_stash_pop.register_fake
+def paged_stash_pop_fake(
+    page_record: Tensor,
+    page_size: int,
+    hidden_size: int,
+    dtype: torch.dtype,
+    buffer_id: int,
+) -> Tensor:
```
crcrpar, line 893: ditto

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+# Joint-graph pass: eligibility check
+# ---------------------------------------------------------------------------
+
+
+def _is_paged_stash_eligible(
+    node: fx.Node,
+    forward_nodes: set[fx.Node],
+    backward_nodes: set[fx.Node],
+    paged_buffers: dict[tuple[torch.dtype, int], "PagedStashBuffer"],
+) -> tuple[list[fx.Node], "PagedStashBuffer"] | None:
```
crcrpar, line 1035: ditto - `from __future__ import annotations` would allow just `PagedStashBuffer`

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Joint-graph pass: main entry point
+# ---------------------------------------------------------------------------
+
+
+def apply_paged_stash_pass(
+    gm: torch.fx.GraphModule,
+    paged_buffers: dict[tuple[torch.dtype, int], "PagedStashBuffer"],
```
crcrpar, line 1084: ditto - str or type

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
```diff
@@ -0,0 +1,1402 @@
...
+# ---------------------------------------------------------------------------
+# aot_fx_trace-compatible pass wrapper
+# ---------------------------------------------------------------------------
+
+
+def paged_stash_pass(
+    gm: torch.fx.GraphModule,
+    example_inputs: tuple | None = None,
+    *,
+    paged_buffers: dict[tuple[torch.dtype, int], "PagedStashBuffer"],
```
crcrpar, line 1353: ditto - str or type

crcrpar (commented):
mostly cosmetic comments

torchtitan/distributed/deepep/hybridep.py
syed-ahmed: Agreed. Currently have a pre-train step hook (register_pre_train_step_hook) that raises a runtime error after the full train step when overflow is detected (so weights are probably still updated). Will look into how to do a check on device.

torchtitan/experiments/cuda_graphable_moe/configs.py
syed-ahmed: This is now outdated. Paged Stash config is moved to: 
```
from torchtitan.experiments.graph_trainer.paged_stash_memory_policy import PagedStash
```
It's now imported in graph_trainer config.

torchtitan/experiments/cuda_graphable_moe/configs.py
syed-ahmed: This is now outdated and not part of this PR.

torchtitan/experiments/cuda_graphable_moe/paged_stash_graph_pass.py
syed-ahmed: Agreed, is it ok to leave a TODO and first integrate MXFP8 with graph_trainer and then update this pass to account for the scaled_grouped_mm?

torchtitan/experiments/graph_trainer/paged_stash_memory_policy.py
kshitij12345: Sorry, I missed the ping here.

> mutates_args (Iterable[str] or "unknown") – The names of args that the function mutates. This MUST be accurate, otherwise, the behavior is undefined. If “unknown”, it pessimistically assumes that all inputs to the operator are being mutated.

custom_op doc: https://docs.pytorch.org/docs/2.12/library.html#torch.library.custom_op

As Ivan pointed out, the compiler is free to reorder operations that are pure (which mutates_args=() implies). 

To fix this,
1. Thread the state tensors explicitly: As mentioned by Ivan, pass `free_list_head` and `overflow` as args to `copy`, `free_list_tail` and `overflow` to `pop`, with correct `mutates_args`. This gives the compiler clear picture about which tensors are mutated. The tradeoff is that these tensors need to appear in the graph (via `get_attr` / `register_buffer`), which may conflict with the DTensor/deepcopy concerns mentioned in the registry comment. Registering them as non-persistent buffers (persistent=False) may help.
2. mutates_args="unknown": Pessimistically tells the compiler all inputs may be mutated, preventing reordering past anything that could alias. Avoids the DTensor/deepcopy issue since no new graph tensors are introduced, but may inhibit other valid optimizations since the compiler treats every input as potentially mutated.

PR #3222, opened 2026-05-05
[graph_trainer] Add context parallelism support to CooR precompile

bobrenjc93:
## Summary

- Replace the `NotImplementedError` for context parallelism in the `_precompile_aot_fx_trace` path with an actual call to `prepare_context_parallel_input`, matching the trainer's `post_dataloading_process`
- Fix `dummy_global_valid_tokens` to exclude the CP dimension — the trainer computes this via `dist_sum` over the "batch" mesh (dp_shard × dp_replicate), which doesn't cover CP ranks
- Add comprehensive unit and integration tests for the new CP precompile support

### DIFF
--- a/torchtitan/experiments/graph_trainer/precompile_main.py
+++ b/torchtitan/experiments/graph_trainer/precompile_main.py
@@ -320,15 +320,14 @@ def _precompile_aot_fx_trace(
     dummy_labels = torch.randint(
         0, vocab_size, (local_batch_size, seq_len), device=device
     )
-    # The trainer computes global_valid_tokens via dist_sum (an
-    # all-reduce + .item()), which returns a Python float. Use the
-    # same type here so make_fx bakes it as a graph constant — not a
-    # graph input — identical to the non-precompile runtime trace.
+    # Must match Trainer.train_step (trainer.py): local_valid_tokens is
+    # counted from raw dataloader labels BEFORE CP sharding, then
+    # all-reduced over the "batch" mesh (dp_shard × dp_replicate).
+    # The batch mesh is defined as dp_replicate * dp_shard in
+    # ParallelDims.build_mesh and excludes CP, so CP must NOT be
+    # included here.
     global_batch_size = (
-        local_batch_size
-        * parallel_dims.dp_shard
-        * parallel_dims.dp_replicate
-        * parallel_dims.cp
+        local_batch_size * parallel_dims.dp_shard * parallel_dims.dp_replicate
     )
     dummy_global_valid_tokens = float(global_batch_size * seq_len)
     extra_inputs: dict[str, torch.Tensor] = {}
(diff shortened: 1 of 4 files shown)


torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -366,6 +366,35 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
             "aot_fx_trace_llama3_fsdp_tp_full_inductor",
             ngpu=8,
         ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module graph_trainer.llama3",
+                    "--config graph_trainer_llama3_debugmodel",
+                    "--compile.mode aot_fx_trace",
+                    "--parallelism.data_parallel_shard_degree 4",
```
SherlockNoMad: let's do fsdp2xtp2xcp2

SherlockNoMad: ah, ic you already have it below, so that should be sufficient, don't need fsdp4xcp2

torchtitan/experiments/graph_trainer/tests/run_precompile_tests.py
```diff
@@ -143,6 +147,52 @@ def _build_precompile_tests() -> list[PrecompileTestDefinition]:
...
+                "--config graph_trainer_llama3_debugmodel",
+                "--compile.mode aot_fx_trace",
+                f"--compile.precompile_artifact_dir {fx_trace_cp_precompile_dir}",
+                "--parallelism.data_parallel_shard_degree 4",
+                "--parallelism.context_parallel_degree 2",
+            ],
+            test_descr="aot_fx_trace llama3 precompile FSDP+CP",
+            test_name="aot_fx_trace_llama3_precompile_fsdp_cp",
+            ngpu=8,
+        ),
```
SherlockNoMad: same here, fsdp2xtp2xcp2 should be sufficient.

torchtitan/experiments/graph_trainer/precompile_main.py
```diff
@@ -321,15 +321,14 @@ def _precompile_aot_fx_trace(
...
+    # all-reduced over the "batch" mesh (dp_shard × dp_replicate).
+    # The batch mesh is defined as dp_replicate * dp_shard in
+    # ParallelDims.build_mesh and excludes CP, so CP must NOT be
+    # included here.
     global_batch_size = (
-        local_batch_size
-        * parallel_dims.dp_shard
-        * parallel_dims.dp_replicate
-        * parallel_dims.cp
+        local_batch_size * parallel_dims.dp_shard * parallel_dims.dp_replicate
```
SherlockNoMad, line 330: I didn't double check the eager implementation, and is trusting your comment here.

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -372,6 +372,21 @@ def _build_llama3_tests() -> list[OverrideDefinitions]:
             "aot_fx_trace_llama3_fsdp_tp_full_inductor",
             ngpu=8,
         ),
+        OverrideDefinitions(
```
aditvenk, line 375: hmm.. I don't think this actually enables CP as I don't see graph trainer's parallelize doing anything for CP:
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/graph_trainer/llama3/parallelize.py#L32

Contrasted with
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/parallelize.py#L61

bobrenjc93: Good catch, I'll hold off on this until graph trainer supports CP

PR #3224, opened 2026-05-05
[rl] Remove VLLMGenerator.__del__ cleanup method

daniellepintz:
The `__del__` method is unnecessary and fragile — Python's GC naturally cleans up self._engine, and module globals like torch may be None during interpreter shutdown. Removing it eliminates the shutdown error `AttributeError: 'NoneType' object has no attribute 'cuda'` that some are experiencing.

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -366,4 +366,5 @@ class VLLMGenerator(Actor, Configurable):
         """Cleanup vLLM engine."""
         if hasattr(self, "_engine"):
             del self._engine
-            torch.cuda.empty_cache()
+            if torch is not None:
+                torch.cuda.empty_cache()
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -361,10 +361,3 @@ class VLLMGenerator(Actor, Configurable):
         logger.debug(
             f"{os.getpid()=} Generator pulled model state dict for policy v{version}"
         )
-
-    def __del__(self):
-        """Cleanup vLLM engine."""
-        if hasattr(self, "_engine"):
-            del self._engine
-            if torch is not None:
-                torch.cuda.empty_cache()


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -366,4 +366,5 @@ def __del__(self):
         """Cleanup vLLM engine."""
         if hasattr(self, "_engine"):
             del self._engine
-            torch.cuda.empty_cache()
+            if torch is not None:
```
tianyu-l: when would torch be None? curious why it didn't happen for others / before.

Lucaskabela: I have seen this error myself - it is a race condition with teardown afaik

Lucaskabela (approved):

torchtitan/experiments/rl/actors/generator.py
daniellepintz: This is what I found for why torch would be None: 

During interpreter shutdown, Python starts cleaning up global state, and module globals can get set to None before all object destructors (`__del__` methods) have finished running. The order in which globals are nulled out during shutdown is not guaranteed.

As to why it didn't happen before, I am really not sure, but can look into it tmro if I have time - it is weird that it just started happening recently, I think after I updated my environment

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -366,4 +366,5 @@ def __del__(self):
         """Cleanup vLLM engine."""
         if hasattr(self, "_engine"):
             del self._engine
```
tianyu-l: then my question would be -- why do we need to do this `del self._engine` manually? what if we just remove this function in general and let interpreter handle it? cc @wwwjn

wwwjn: I think we can remove this , because all of our generator teardown happens at the end of the program, which means the interpreter will clean up everything.

daniellepintz:
@wwwjn @tianyu-l Removed `__del__` entirely!

tianyu-l (approved):

PR #3225, opened 2026-05-05
[codex] Fix weight tying parameter accounting

bruce2233:
## Summary
- Keep tied input/output embeddings counted once in model parameter totals.
- Keep tied lm_head weights included in FLOPs accounting because the output projection still runs.
- Add a regression test that tied model parameter reporting matches unique PyTorch parameters.

## Root Cause
model.parameters() already de-duplicates shared parameters when input embeddings and lm_head are tied, but the accounting helper subtracted the embedding parameters again. That under-reported tied model size and also removed the output projection from the FLOPs estimate.

## Validation
- /data/app/modded-nanogpt/.venv/bin/python -m pytest -q tests/unit_tests/test_weight_tying.py

### DIFF
--- a/torchtitan/models/utils.py
+++ b/torchtitan/models/utils.py
@@ -407,6 +407,8 @@ def get_dense_model_nparams_and_flops(
             nparams: Total number of model parameters.
             num_flops_per_token: Estimated number of floating point operations per token.
     """
+    # model.parameters() de-duplicates shared parameters, so tied input/output
+    # embeddings are counted once.
     nparams = sum(p.numel() for p in model.parameters())
     nparams_embedding = sum(
         sum(p.numel() for p in m.parameters())
@@ -425,9 +427,9 @@ def get_dense_model_nparams_and_flops(
     # With tied embeddings, PyTorch's parameter iterator already counts the
     # shared input/output parameter once. That parameter still participates in
     # the lm_head matmul, so do not subtract it for either size or FLOPs.
-    nparams_for_flops = nparams if enable_weight_tying else nparams - nparams_embedding
+    nparams_for_matmul = nparams if enable_weight_tying else nparams - nparams_embedding
     num_flops_per_token = (
-        6 * nparams_for_flops + 6 * n_layers * n_heads * head_dims * seq_len
+        6 * nparams_for_matmul + 6 * n_layers * n_heads * head_dims * seq_len
     )
 
     return nparams, num_flops_per_token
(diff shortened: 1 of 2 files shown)


torchtitan/models/utils.py
tianyu-l, line 412: add a comment that `model.parameters()` already does the dedup

torchtitan/models/utils.py
```diff
@@ -422,14 +422,14 @@ def get_dense_model_nparams_and_flops(
     #    but recomputation should not be counted in calculating MFU           (+0)
     # 3. each matmul performs 1 multiplication and 1 addition                 (*2)
     # 4. we follow the convention and do not account for sparsity in causal attention
+    # With tied embeddings, PyTorch's parameter iterator already counts the
+    # shared input/output parameter once. That parameter still participates in
+    # the lm_head matmul, so do not subtract it for either size or FLOPs.
+    nparams_for_flops = nparams if enable_weight_tying else nparams - nparams_embedding
```
tianyu-l:
```suggestion
    nparams_for_matmul = nparams if enable_weight_tying else nparams - nparams_embedding
```

torchtitan/models/utils.py
```diff
@@ -493,16 +493,29 @@ def get_moe_model_nparams_and_flops(
...
+    # shared input/output parameter once. That parameter still participates in
+    # the lm_head matmul, so do not subtract it for either size or FLOPs.
+    nparams_embedding_for_flops = (
+        0
+        if (
+            hasattr(model_config, "enable_weight_tying")
+            and model_config.enable_weight_tying
+        )
+        else nparams_embedding
+    )
```
tianyu-l: this is a bit hard to read, let's follow the same structure for dense:
- `if weight tying: nparams_for_matmul = ...`
- `else: nparams_for_matmul = ...`

and then later
`num_flops_per_token = 6 * nparams_for_matmul + ...`

torchtitan/models/utils.py
```diff
@@ -493,16 +493,29 @@ def get_moe_model_nparams_and_flops(
...
-    # If weight tying is enabled, subtract embedding parameters from total count
     if (
         hasattr(model_config, "enable_weight_tying")
         and model_config.enable_weight_tying
     ):
-        nparams = nparams - nparams_embedding
+        logger.info(
+            "Weight tying enabled; shared input/output embedding parameters "
+            "are counted once."
+        )
```
tianyu-l: I think it's fine to remove all this

tianyu-l (commented):
Thanks for identifying and fixing the bug! Left some comments

torchtitan/models/utils.py
bruce2233: Thanks for your review. I just update the code :)

tianyu-l:
pls fix lint

tianyu-l (approved):
sgtm, thanks!

PR #3228, opened 2026-05-05
[draft] Add blockwise FP8 linears

iamzainhuda:
## Summary

- add an experimental blockwise FP8 linear recipe backed by the torchao prototype blockwise FP8 training linear
- add DeepSeek V3 debug-model configs for blockwise FP8 batch sizes 1, 4, and 8
- add helper scripts for BF16 versus blockwise FP8 debug-model comparisons using the default chunked-loss path
  - can remove later, this is to do runs locally for now

## Testing

8 GPU, bs=8, 1k steps
image

image


iamzainhuda:
cc: @danielvegamyhre

PR #3229, opened 2026-05-05
Fix ChunkedCELoss final FSDP wrapping

iamzainhuda:
## Summary

- keep final `norm` and `lm_head` in separate FSDP units when `ChunkedCELoss` is active
- avoid forcing an immediate `lm_head` reshard at the end of the chunked loss forward
- replace the custom chunked-loss autograd bridge with an equivalent zero-value gradient bridge
- add unit coverage showing the zero-value bridge returns the same scalar and computes the same hidden-state gradients as the legacy custom autograd bridge

`ChunkedCELoss` runs backward through `lm_head` chunk-by-chunk before the outer decoder backward reaches final `norm`. When `norm` and `lm_head` share one FSDP unit, `lm_head` backward/reshard can affect storage that `norm` still needs later in the same backward pass. On the 8-GPU DeepSeek V3 debug model this fails during the first backward pass with an unallocated tensor error.

Splitting `norm` and `lm_head` only for the chunked-loss path isolates their FSDP storage lifecycle while preserving the existing grouped wrapping for the standard loss path.

The zero-value gradient bridge keeps the same forward value as the old custom autograd Function while attaching the accumulated chunk gradient to `hidden_states` through regular PyTorch ops. The added unit test recreates the legacy bridge locally and verifies both approaches produce identical scalar values and hidden-state gradients.

## Validation

- `PYTHONPATH=/home/dev/ao pytest -q tests/unit_tests/test_loss.py`
- latest `origin/main` fails with `NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --training.steps 3`
- this branch completes the same 8-GPU DeepSeek V3 debug-model smoke for 3 steps

### DIFF
--- a/tests/unit_tests/test_loss.py
+++ b/tests/unit_tests/test_loss.py
@@ -16,6 +16,20 @@ from torchtitan.components.loss import (
 )
 
 
+class _LegacyDecoderOutputGradientBackProp(torch.autograd.Function):
+    """Legacy ChunkedCELoss gradient bridge kept for equivalence testing."""
+
+    @staticmethod
+    def forward(ctx, hidden_states, accumulated_grad, loss):
+        ctx.save_for_backward(accumulated_grad)
+        return loss.detach().clone()
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        (accumulated_grad,) = ctx.saved_tensors
+        return accumulated_grad, None, None
+
+
 class TestLoss(unittest.TestCase):
     def test_ignore_index_equal_per_token_contribution(self):
         """Test that each valid token contributes equally to the loss.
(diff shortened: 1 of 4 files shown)


wwwjn:
> latest origin/main fails with NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --training.steps 3

Hi thanks for this. Did you run this test with latest pytorch nightly? We have CI to  guard this which is passing on ChunkedLoss change. ChunkedLoss depends on PyTorch changes https://github.com/pytorch/pytorch/pull/180428

iamzainhuda:
The custom autograd function change is an additional thing I added that isn't actually needed to make this work but thought we could simplify. I'm happy to remove it if you want to keep it in. There is a small transient memory tradeoff because `hidden_states * accumulated_grad`, but this isn't persistent mem.

iamzainhuda:
> > latest origin/main fails with NGPU=8 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --training.steps 3
> 
> Hi thanks for this. Did you run this test with latest pytorch nightly? We have CI to guard this which is passing on ChunkedLoss change. ChunkedLoss depends on PyTorch changes [pytorch/pytorch#180428](https://github.com/pytorch/pytorch/pull/180428)

Just ran it on the latest pytorch nightly, the issue is gone - so this PR is no longer needed. Thank you for the response!

PR #3230, opened 2026-05-05
[rl] Add a tag to trigger RL CI runs

wwwjn:
Mirrors #3210 Add a tag to trigger RL CI

### DIFF
--- a/.github/workflows/integration_test_4gpu_rl.yaml
+++ b/.github/workflows/integration_test_4gpu_rl.yaml
@@ -7,10 +7,8 @@ on:
       - 'torchtitan/experiments/rl/**'
       - '.github/workflows/integration_test_4gpu_rl.yaml'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
-    paths:
-      - 'torchtitan/experiments/rl/**'
-      - '.github/workflows/integration_test_4gpu_rl.yaml'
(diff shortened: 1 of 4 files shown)


.github/workflows/integration_test_8gpu_rl_h100.yaml
tianyu-l, line 1: could you remind me why we need h100 for rl?

.github/workflows/integration_test_4gpu_rl.yaml
tianyu-l, line 1: why 4 gpu instead of 8?

.github/workflows/integration_test_8gpu_rl_h100.yaml
wwwjn: Previously was becase FA2 is not batch invariant with paged kv. Now FA2 should be batch invariant. I have another PR remove this H100 test to A10G.

.github/workflows/integration_test_4gpu_rl.yaml
wwwjn: Historically we start with 4 GPU. Using TP=4 is planned in later CI refactor PRs

tianyu-l (approved):
didn't verify, stamp to unblock

wwwjn:
> didn't verify, stamp to unblock

I can follow up with verification, thank you

PR #3232, opened 2026-05-05
[rl] Fix CI loss=0 and logprob=NaN

wwwjn:
3 Fixes:
1. uv pip install torchmonarch==0.4.1 update to matching README
2. `export VLLM_USE_FLASHINFER_SAMPLER=0`. We will need VLLM_USE_FLASHINFER_SAMPLER=0 because https://github.com/vllm-project/vllm/pull/40376 landed Apr. 29. For our CI environment, we didn't install nvcc so it won't support FlashInfer to be JIT compiled.
3. temporarily set num_splits=1 for FAv2. If not set, it caused NaN results during vllm generation.

### DIFF
--- a/.github/workflows/integration_test_4gpu_rl.yaml
+++ b/.github/workflows/integration_test_4gpu_rl.yaml
@@ -66,7 +66,7 @@ jobs:
         pip install uv
 
         # 1. Install Monarch and TorchStore
-        uv pip install torchmonarch==0.3.0
+        uv pip install torchmonarch==0.4.1
         uv pip install --no-deps "git+https://github.com/meta-pytorch/torchstore.git@main"
         uv pip install pygtrie portpicker
 
@@ -90,6 +90,11 @@ jobs:
 
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
         sudo chown -R $(id -u):$(id -g) "$RUNNER_TEMP/artifacts-to-be-uploaded"
+
+        # Install nvcc so FlashInfer can JIT compile its CUDA kernels.
+        # vLLM uses FlashInfer sampler by default (vllm-project/vllm#40376)
+        # but the CI docker image only has CUDA runtime, not the toolkit.
+        uv pip install nvidia-cuda-nvcc
 
         # Run E2E RL integration tests (TP=2 on 4 GPUs)
         python -m torchtitan.experiments.rl.tests.integration_tests \
--- a/torchtitan/experiments/rl/models/attention.py
+++ b/torchtitan/experiments/rl/models/attention.py
@@ -168,14 +168,16 @@ class PyTorchVarlenAttentionImpl(FlashAttentionImpl):
                 num_seqs + 1, dtype=torch.int32, device=query.device
             )
             cu_seqlens_k[1:] = torch.cumsum(seqused_k, dim=0)
-        # FA3 + batch-invariant: fix num_splits=1 to prevent non-deterministic
-        # split-k reductions. FA2 is automatically batch-invariant and does
-        # not accept num_splits.
         extra_kwargs = {}
 
-        # Disable split_kv in Flash Attention to ensure bitwise identical output.
-        # see https://github.com/pytorch/pytorch/pull/176905
-        if is_in_batch_invariant_mode() and current_flash_attention_impl() == "FA3":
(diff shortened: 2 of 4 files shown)


torchtitan/experiments/rl/models/attention.py
```diff
@@ -168,14 +168,14 @@ def forward(
...
 
-        # Disable split_kv in Flash Attention to ensure bitwise identical output.
-        # see https://github.com/pytorch/pytorch/pull/176905
-        if is_in_batch_invariant_mode() and current_flash_attention_impl() == "FA3":
+        # Fix num_splits=1 to prevent non-deterministic split-k reductions
+        # and avoid NaN from the split-KV heuristic with paged KV.
+        # After pytorch/pytorch#179760, FA2 also accepts num_splits and
+        # auto-selects num_splits>1 for paged KV, which can produce NaN.
+        # Always set num_splits=1 for FA2 with paged KV.
+        if is_in_batch_invariant_mode() or current_flash_attention_impl() != "FA3":
```
tianyu-l: Always using `num_splits=1` doesn't sound a fix.
Do we know what's the root cause? Are we tracking / creating issues? At least add a TODO here?

.github/workflows/integration_test_4gpu_rl.yaml
```diff
@@ -91,6 +91,9 @@ jobs:
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
         sudo chown -R $(id -u):$(id -g) "$RUNNER_TEMP/artifacts-to-be-uploaded"
 
+        # Disable FlashInfer JIT sampler — CI image lacks nvcc (vllm-project/vllm#40376)
+        export VLLM_USE_FLASHINFER_SAMPLER=0
```
tianyu-l: This doesn't sound a fix either. Should we install nvcc?
What's the implication on perf -- does it put us in a worse position?

wwwjn: It will make the sampling slower, but I haven't benchmarking the slowness yet. I set it to 0 to match our previous behavior - Previously it's opt-in and by default we are not using FlashInfer. Now vllm set using FlashInfer as default, I will need to invest more about how to install nvcc in CI

torchtitan/experiments/rl/models/attention.py
wwwjn: > Do we know what's the root cause? 

Not yet, raise this error to @liangel-02 , let me add a issue to tracking this regression

.github/workflows/integration_test_4gpu_rl.yaml
wwwjn: Let me check if it's possible/easy to install nvcc in CI now, and test if JIT compile FlashInfer would make CI slow down a lot

.github/workflows/integration_test_4gpu_rl.yaml
```diff
@@ -91,6 +91,11 @@ jobs:
         sudo mkdir -p "$RUNNER_TEMP/artifacts-to-be-uploaded"
         sudo chown -R $(id -u):$(id -g) "$RUNNER_TEMP/artifacts-to-be-uploaded"
 
+        # Install nvcc so FlashInfer can JIT compile its CUDA kernels.
+        # vLLM uses FlashInfer sampler by default (vllm-project/vllm#40376)
+        # but the CI docker image only has CUDA runtime, not the toolkit.
+        conda install -y -c nvidia cuda-nvcc
```
wwwjn: Attempt to install nvcc in CI

torchtitan/experiments/rl/models/attention.py
wwwjn: https://github.com/pytorch/torchtitan/issues/3235 I created this issue to track.

tianyu-l: can we use something like `== FA2` instead of `!= FA3`? I don't know what's the set of options.

torchtitan/experiments/rl/models/attention.py
tianyu-l, line 1: should we also fix https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L136

tianyu-l (approved):
please address comments

torchtitan/experiments/rl/models/attention.py
wwwjn: Nice catch, fixed.

PR #3234, opened 2026-05-06
RL: observability spans across trainer/generator/controller (#3234)

felipemello1:
Summary:

TLDR: Adds to the RL experiment the structured-logging APIs. 

Time spans, scalars and events can be logged, per rank, with metadata, to a jsonl or database, using python standard logger. This can then be converted into a gantt chart. It is cheap, runs on every step, and useful to find stragglers, high level bottlenecks, how asynchronous an RL run is, debug timeouts, weird behavior on some rank on some step, etc.

For details of the APIs, please read the added readme in torchtitan/observability/structured_loggger/README.md.

Examples:
 {F1989343083} 

## Is it safe?

The workhorse here is logger.info. So the same way you treat logger.info, you should treat the structured logger.

It can be disabled with a config flag

If the user forgets to setup init_structured_logger, we still call logger.info, but nothing is saved anywhere.

It detects is_compiling and skips if true.

Tested with async coroutines and monarch as well.

Differential Revision: D101225148

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -11,7 +11,8 @@ from dataclasses import dataclass, field
 
 import torch
 import torchstore as ts
-from monarch.actor import Actor, endpoint
+from monarch.actor import Actor, current_rank, endpoint
+
 from torchtitan.config import (
     CompileConfig,
     Configurable,
@@ -24,7 +25,9 @@ from torchtitan.experiments.rl.models.vllm_registry import (
     TORCHTITAN_CONFIG_FORMAT,
 )
 from torchtitan.experiments.rl.types import Completion
+from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.model_spec import ModelSpec
+from torchtitan.tools.logging import init_logger
 from torchtitan.tools.utils import has_cuda_capability
 from vllm import EngineArgs, LLMEngine, SamplingParams
 from vllm.config import AttentionConfig, CompilationConfig
@@ -186,7 +189,17 @@ class VLLMGenerator(Actor, Configurable):
         model_path: str,
         compile_config: CompileConfig,
         max_num_seqs: int,
+        output_dir: str,
     ):
+        init_logger()
+        sl.init_structured_logger(
+            source="rl_generator",
+            output_dir=output_dir,
+            rank=current_rank().rank,
+            enable=config.debug.enable_structured_logging,
+        )
+        sl.log_trace_instant("structured_logger_started")
+
         self.config = config
         self.model_spec = model_spec
 
@@ -252,9 +265,10 @@ class VLLMGenerator(Actor, Configurable):
             engine_kwargs["seed"] = config.debug.seed
         engine_args = EngineArgs(**engine_kwargs)
 
-        logger.info("Initializing LLMEngine from EngineArgs...")
-        self._engine = LLMEngine.from_engine_args(engine_args)
-        logger.info("vLLM rollout engine initialized")
+        with sl.log_trace_span("vllm_init"):
+            logger.info("Initializing LLMEngine from EngineArgs...")
+            self._engine = LLMEngine.from_engine_args(engine_args)
+            logger.info("vLLM rollout engine initialized")
 
         self.policy_version = 0
 
@@ -285,6 +299,12 @@ class VLLMGenerator(Actor, Configurable):
         return self._engine.model_executor.driver_worker.get_model()
 
     @endpoint
+    async def sync_log_step(self, step: int, relative_step: int | None = None) -> None:
+        """Sync the structured-logger step counter from the controller."""
+        sl.set_step(step, relative_step=relative_step)
+
+    @endpoint
+    @sl.log_trace_span("generate")
     async def generate(
         self,
         prompts: list[str],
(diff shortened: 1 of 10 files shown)


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -258,6 +273,12 @@ def _get_model(self):
         return self._engine.model_executor.driver_worker.get_model()
 
     @endpoint
+    async def sync_step(self, step: int, relative_step: int | None = None) -> None:
```
wwwjn: Maybe rename to something related to observability / structured logging?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -88,9 +90,17 @@ def __init__(
...
     ):
-
         init_logger()
+        sl.init_structured_logger(
+            source="rl_trainer",
+            output_dir=output_dir,
+            rank=current_rank().rank,
+            enable=enable_structured_logging,
+        )
+        sl.log_trace_instant("binary_start")
```
wwwjn: What does `binary_start` mean?

torchtitan/experiments/rl/grpo.py
```diff
@@ -550,18 +572,27 @@ async def train(self):
                 self._collate_episodes(per_rank_episodes)
                 for per_rank_episodes in self._shard_episodes(episodes)
             ]
-            fwd_bwd_metrics = self._get_rank_0_value(
-                self.trainer.forward_backward.call(batches).get()
-            )
-            optim_metrics = self._get_rank_0_value(self.trainer.optim_step.call().get())
+            with sl.log_trace_span("fwd_bwd"):
```
wwwjn: We already have another with sl.log_trace_span("forward_backward") in trainer.py. Are they duplicated? Should we consolidate most of the sl.log_trace_span() call in controller?

torchtitan/experiments/rl/actors/trainer.py
felipemello1: you are the 3rd person that asked that. sixlib uses it, and its basically to indicate that init_structured_logger started at this point. I will rename.

torchtitan/experiments/rl/grpo.py
felipemello1: good question! I think we should keep both. This one is from the perspective of the controller (check the first and last channel in the picture). Let me know if you disagree. I am also learning how we should navigate this.

Image

torchtitan/experiments/rl/grpo.py
```diff
@@ -351,107 +354,115 @@ async def setup(
...
-            # Multi-node mode: dedicate whole nodes to trainer vs generator
-            if (
-                trainer_nodes is None
-                or generator_nodes is None
-                or gpus_per_node is None
-            ):
-                raise ValueError(
-                    "trainer_nodes, generator_nodes, and gpus_per_node are "
-                    "required when host_mesh is provided"
+        with sl.log_trace_span("mesh_spawn"):
```
tianyu-l, line 377: quite large span in terms of LoC -- seems worth a better provisioner abstraction (later)

torchtitan/experiments/rl/grpo.py
```diff
@@ -550,18 +572,27 @@ async def train(self):
...
+                )
             metrics = {**fwd_bwd_metrics, **optim_metrics}
 
             # --- Weight sync --- #
-            t0 = time.perf_counter()
-            self.trainer.push_model_state_dict.call().get()
-            t_push = time.perf_counter() - t0
-            self.generator.pull_model_state_dict.call(metrics["policy_version"]).get()
-            t_sync = time.perf_counter() - t0
+            with sl.log_trace_span("trainer_push_model_state_dict"):
```
tianyu-l, line 610: in the initial weight push above you used "trainer_weight_push". Let's be consistent. I prefer we use the original function name like here.

torchtitan/experiments/rl/grpo.py
```diff
@@ -592,6 +623,15 @@ async def train(self):
 async def main():
     """Run the distributed RL training loop using Monarch."""
     config = ConfigManager().parse_args()
+    sl.init_structured_logger(
+        source="rl_controller",
+        output_dir=config.dump_folder,
+        rank=0,
+        # pyrefly: ignore [missing-attribute]
+        enable=config.trainer.debug.enable_structured_logging,
```
tianyu-l, line 653: @wwwjn sorry for the back and forth, but I now think we probably should keep just one debug config and pass it to both trainer and generator, just like the compile config.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -314,7 +334,8 @@ async def forward_backward(self, train_data: list[TrainBatch]) -> dict:
 
         # Backward pass
         self.optimizers.zero_grad()
-        loss.backward()
+        with sl.log_trace_span("backward"):
```
tianyu-l: let's be consistent with "model_forward", either is fine

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -169,7 +172,18 @@ def __init__(
...
         compile_config: CompileConfig,
         max_num_seqs: int,
+        output_dir: str,
+        enable_structured_logging: bool,
     ):
+        init_logger()
+        sl.init_structured_logger(
+            source="rl_generator",
+            output_dir=output_dir,
+            rank=current_rank().rank,
```
tianyu-l, line 198: Is this rank among the whole world, or this spmd world? It looks the former, and for the latter we have _local_rank here https://github.com/pytorch/torchtitan/blob/main/torchtitan/observability/structured_logger/jsonl_handler.py#L63

tianyu-l (approved):
please address comments

PR #3236, opened 2026-05-06
Enable DP+EP for MoE inference in vLLM wrapper

wwwjn:
* __->__ #3236

## DP2EP in generator
**Important Trick**: Inference doesn't use FSDP, instead it use pure data parallel. So in DP2EP implementation, dp_replicate degree is translated to pure data parallel in dense region, to bypass paralell_dim build and check rule.

Long term solution: torchtitan should support converting `data_replicate` degree to EP degree too, not orthogonal. 
- https://github.com/pytorch/torchtitan/blob/a6815bb8c5b240fadcd05f0373e9e505bc3b7271/torchtitan/distributed/parallel_dims.py#L234

## Batch-invariant 
Supported bitwise parity **trainer FSDP2 TP2 EP4 == generator DP2 TP2 EP4 bitwise** — FSDP2 with bf16 all-gather matches the generator's replicated bf16 dense DP (FSDP all-gathers full bf16 params before the forward, so the forward numerics are identical)


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -760,6 +771,20 @@ async def _decide_next_action(self) -> LoopDecision:
             )
             return LoopDecision(action=LoopAction.STEP, requests=requests)
 
+    def _engine_has_local_unfinished_requests(self) -> bool:
```
wwwjn: Need to rebase to latest main

torchtitan/experiments/rl/tests/test_bitwise_parity.py
wwwjn, line 1: Changes to this file is to support bitwise unit test for MoE model

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -467,6 +474,10 @@ def __init__(
             config_format=TORCHTITAN_CONFIG_FORMAT,
             dtype=config.model_dtype,
             tensor_parallel_size=config.parallelism.tensor_parallel_degree,
+            # Generator-only convention: this config field is TorchTitan FSDP
+            # degree for trainers, but here it is intentionally mapped to vLLM's
+            # pure data-parallel degree. The vLLM wrapper skips FSDP/DDP.
+            data_parallel_size=config.parallelism.data_parallel_shard_degree,
```
tianyu-l: This will be too confusing for users. Let's create a dedicated `InferenceParallelismConfig` and just use `data_parallel_degree`, which is almost a subset of https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configs.py#L85

We can translate it to the training one, if any utils require that.

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -739,7 +750,10 @@ async def _decide_next_action(self) -> LoopDecision:
                 lambda: self._close_request is not None
                 or self._model_state_dict_pull_request is not None
                 or self._queued_generation_requests
-                or self._engine.has_unfinished_requests()
+                # rank-0-only decision: use the local (no DP all-reduce) check;
+                # ``engine.has_unfinished_requests()`` would do a DP collective
+                # that the follower ranks never join -> deadlock under vLLM DP>1.
```
tianyu-l: didn't understand the comment

torchtitan/experiments/rl/actors/generator.py
tianyu-l, line 637: as discussed offline, right now it seems all dp ranks of a generator receives the same input

torchtitan/experiments/rl/actors/generator.py
tianyu-l: unresolved yet

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -168,16 +152,20 @@ def __init__(
         self.config = dataclasses.replace(model_config, layers=new_layers)
         logger.debug(f"Creating model with config: {self.config.to_dict()}")
 
-        # Build ParallelDims from the TorchTitan ParallelismConfig so TP/EP
-        # sharding sees the same mesh shape as vLLM. In the generator,
-        # data_parallel_shard_degree means vLLM pure DP, not TorchTitan FSDP.
+        # Translate the inference parallelism into torchtitan's full
+        # ParallelismConfig that ParallelDims / parallelize_fn consume.
+        torchtitan_parallelism = parallelism.to_torchtitan_parallelism_config()
```
tianyu-l: both are torchtitan parallelism
- maybe rename to `training_parallelism`, compared with inference parallelism?

not perfect, I admit

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -42,6 +43,57 @@
 TORCHTITAN_CONFIG_FORMAT = "torchtitan"
 
 
+@dataclass(kw_only=True, slots=True)
+class InferenceParallelismConfig:
```
wwwjn, line 47: I decided to put this config in vllm_registry (lower level of generator.py)
My current rational is `InferenceParallelismConfig` is not specific to RL's generator, but also can be used in pure generate / inference - It should be at the similar level of `vllm_wrapper.py`

Alternative option is vllm_wrapper.py or a separate config.py under models/

torchtitan/experiments/rl/train.py
```diff
@@ -201,7 +201,9 @@ async def main():
     rl_trainer: RLTrainer = config.build()
     try:
         trainer_world_size = _compute_world_size(config.trainer.parallelism)
-        generator_world_size = _compute_world_size(config.generator.parallelism)
+        generator_world_size = _compute_world_size(
```
tianyu-l: introduce _compute_generator_world_size and _compute_trainer_world_size

torchtitan/experiments/rl/actors/generator.py
wwwjn: This is a rebase issue and trying to solved, and the comment makes no-sense to me too. Removed

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -42,6 +43,56 @@
...
+    """vLLM pure data-parallel degree; parameters are replicated across these
+    groups. 1 means disabled."""
+
+    tensor_parallel_degree: int = 1
+    """Tensor parallelism degree. 1 means disabled."""
+
+    expert_parallel_degree: int = 1
+    """Expert parallelism degree for MoE layers. 1 means disabled."""
+
+    def to_training_parallelism_config(self) -> ParallelismConfig:
```
tianyu-l:
```suggestion
    def to_training(self) -> ParallelismConfig:
```

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -128,7 +179,7 @@ def model_spec_to_hf_config_dict(spec: ModelSpec) -> dict[str, Any]:
 def registry_to_vllm(
```
tianyu-l:
```suggestion
def register_to_vllm(
```
o/w doesn't read

tianyu-l (approved):

wwwjn:
h100 RL integration test waiting time too long, would love to have a forward fix if it breaks H100 RL test

PR #3237, opened 2026-05-06
[RL] - Enable experiment metrics

felipemello1:
## Why

torchtitan RL has no metric layer today. The training loop ends with `logger.info(f"Step {step} | Loss: ... | Reward: ...")` — no W&B integration or clean place for actors to declare metrics.

This PR adds APIs to define metrics, aggregate and log them. 

**For details, please check the README.**

Note: For this RFC, SPMD training actor is out of scope — we assume that it can run dist.all_reduce and make its metrics available to the controller on rank 0.

## Some extra changes

1. Refactored loss to enable proper mean-of-means, using sum of local and global tokens.
2. Refactored bit-wise to simplify and vectorize
3. In both cases above, we pack values to reduce the number of collectives.

## Non-goals

- In previous discussion I proposed to use a subprocess to aggregate and log. Turns out that
a) aggregating is faster than serializing the message to sending it to the subprocess;
b) wandb logs using another thread, so it is non blocking

So, for our current needs, it doesn't make sense to have a subprocess to aggregate and log.

## Test plan

```
python torchtitan/experiments/rl/grpo.py \
  --module rl --config rl_grpo_qwen3_0_6b \
  --num_steps 20 \
  --trainer.parallelism.tensor-parallel-degree 1 \
  --generator.parallelism.tensor-parallel-degree 1 \
  --metrics.enable-wandb
```

image
image
image
image
image


torchtitan/experiments/rl/tests/test_grpo_metrics.py
tianyu-l, line 1: similar to my comment in the other PR: shall we set up a CI yaml for these unit tests?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -294,51 +356,45 @@ async def forward_backward(self, train_data: list[TrainBatch]) -> dict:
...
 
         Returns:
-            dict with grad_norm and policy_version.
+            Reduced metrics from optimization step.
         """
         # TODO: Accept optional optimizer params (e.g. learning rate)
         # to allow controller-owned schedules (see Tinker API).
 
+        # capture LR before step
+        current_lrs = self.lr_schedulers.schedulers[0].get_last_lr()
```
tianyu-l, line 416: trainer.py has https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L699
we should at least add an assert over there as well -- can be done in this PR or separate PR

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -357,8 +413,11 @@ async def optim_step(self) -> dict:
         )
 
         return {
-            "grad_norm": grad_norm.item() if hasattr(grad_norm, "item") else grad_norm,
-            "policy_version": self.policy_version,
+            "train/grad_norm/mean": (
+                grad_norm.item() if hasattr(grad_norm, "item") else float(grad_norm)
```
tianyu-l: .item() will incur D2H sync, which may degrade perf in the async setup. What's your thought?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -357,8 +413,11 @@ async def optim_step(self) -> dict:
...
+            ),
+            "train/lr": current_lr,
+            "train/policy_version/mean": float(self.policy_version),
```
tianyu-l: does everything need to be a float? Anyway, taking mean of policy_version sounds tricky.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
-    return result
+    # sanity check
+    if len(vllm_token_log_probs) != len(batch_token_log_probs):
+        raise ValueError(
+            f"verify_logprob_identity: sample count mismatch - "
+            f"vllm={len(vllm_token_log_probs)}, "
+            f"trainer={len(batch_token_log_probs)}"
+        )
+
+    if not vllm_token_log_probs:
```
tianyu-l: curious, when would this happen?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+
 def verify_logprob_identity(
     vllm_token_log_probs: list[list[float]],
     batch_token_log_probs: list[torch.Tensor],
-) -> dict:
-    """
-    Check if vLLM log probs and computed log probs are bit-wise identical,
-    and compute the log ratio (train/generator) between them.
+    *,
+    device: torch.device,
```
tianyu-l, line 81: add docstring including justification of why we need it

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+            logprob_diff_max=zero,
+            num_tokens_different=zero,
+        )
+
+    # Make everything a tensor
+    vllm_flat = torch.tensor(
+        [lp for sample in vllm_token_log_probs for lp in sample],
+        dtype=torch.float32,
+    )
+    titan_flat = torch.cat([t.detach().cpu().float() for t in batch_token_log_probs])
```
tianyu-l: Please use generator vs. trainer, instead of vllm vs. titan.

In general I find it unnecessary to use "titan" in any variable naming inside the titan repo.

tianyu-l: why putting to cpu? In general I feel we should be more careful in any D2H and H2D sync.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+    if not vllm_token_log_probs:
+        zero = torch.zeros((), dtype=torch.float32, device=device)
+        return LogprobVerificationOutput(
+            logprob_diff_sum=zero,
+            logprob_diff_max=zero,
+            num_tokens_different=zero,
+        )
+
+    # Make everything a tensor
+    vllm_flat = torch.tensor(
```
tianyu-l: device?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+        [lp for sample in vllm_token_log_probs for lp in sample],
```
tianyu-l: how should I read this?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+    )
+    titan_flat = torch.cat([t.detach().cpu().float() for t in batch_token_log_probs])
+
+    # sanity check
+    if vllm_flat.numel() != titan_flat.numel():
+        raise ValueError(
+            f"verify_logprob_identity: total token count mismatch - "
+            f"vllm={vllm_flat.numel()}, trainer={titan_flat.numel()}"
+        )
+    if vllm_flat.numel() == 0:
```
tianyu-l: is this not the same as `if not vllm_token_log_probs` above?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+
+    # Make everything a tensor
+    vllm_flat = torch.tensor(
+        [lp for sample in vllm_token_log_probs for lp in sample],
+        dtype=torch.float32,
+    )
+    titan_flat = torch.cat([t.detach().cpu().float() for t in batch_token_log_probs])
+
+    # sanity check
+    if vllm_flat.numel() != titan_flat.numel():
```
tianyu-l: this sounds redundant too

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -248,15 +252,73 @@ def _build_model(
...
+            sum_pack = funcol.all_reduce(
+                sum_pack, reduceOp=c10d.ReduceOp.SUM.name, group=loss_mesh
+            )
+            max_value = funcol.all_reduce(
+                max_value, reduceOp=c10d.ReduceOp.MAX.name, group=loss_mesh
+            )
+
+        # Unpack and normalize by global_valid_tokens.
+        # Metrics are expected to already be scaled by local_valid_tokens.
+        global_valid_tokens = sum_pack[-1].clamp(min=1.0)
```
tianyu-l: this should be obtained before training to obtain the correct grad scaler? similar to the refactor from @wwwjn for sft trainer

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -248,15 +252,73 @@ def _build_model(
...
+
+        # loss metrics
+        out: dict[str, float] = {}
+        for i, k in enumerate(loss_keys):
+            out[k] = float((sum_pack[i] / global_valid_tokens).item())
+
+        # logprob verification metrics
+        n_loss = len(loss_keys)
+        out["bit_wise/logprob_diff/mean"] = float(
+            (sum_pack[n_loss] / global_valid_tokens).item()
```
tianyu-l: relying on the index such as `n_loss` sounds VERY fragile

torchtitan/experiments/rl/loss/types.py
tianyu-l, line 1: considering loss is not defined here, can we just put the new dataclasses in https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/types.py?

torchtitan/experiments/rl/loss/types.py
```diff
@@ -0,0 +1,37 @@
...
+
+    `token_mean_metric_sums` is a local SUM over valid tokens,
+    meaning that it is already weighted by the number of valid
+    tokens in the batch. To calculate metrics correctly, they
+    need to be all-reduced and then divided by the global number
+    of valid tokens.
+    """
+
+    loss: torch.Tensor
+    token_mean_metric_sums: dict[str, torch.Tensor]
```
tianyu-l: why nesting a dict inside a dataclass?

torchtitan/experiments/rl/loss/types.py
```diff
@@ -0,0 +1,37 @@
...
+    num_local_valid_tokens: torch.Tensor
+
+
+def sequence_scalar_token_weighted_sum(
```
tianyu-l: This should be in utils file, not types.py.
If it's only used in grpo.py, let's put it there as private function for now.

torchtitan/experiments/rl/grpo.py
```diff
@@ -73,28 +79,41 @@ def __call__(
...
-    ) -> tuple[torch.Tensor, dict[str, float]]:
-        per_sample_mean_lps = []
-        for policy_lps in policy_logprobs:
-            per_sample_mean_lps.append(policy_lps.mean())
-
-        mean_log_ratio = torch.stack(per_sample_mean_lps)
-        ratio = torch.exp(mean_log_ratio)
+    ) -> LossOutput:
+        response_lens = torch.tensor(
+            [policy_lps.numel() for policy_lps in policy_logprobs],
```
tianyu-l: let's avoid using acronyms, can be done in separate PR

torchtitan/experiments/rl/grpo.py
```diff
@@ -73,28 +79,41 @@ def __call__(
...
-            "pg_loss": pg_loss.item(),
-            "ratio_mean": ratio.mean().item(),
-            "ratio_clipped_frac": (torch.abs(ratio - clipped_ratio) > 1e-6)
-            .float()
-            .mean()
-            .item(),
-        }
-        return pg_loss, metrics
+        sample_pg_losses = -torch.min(ratio * advantages, clipped_ratio * advantages)
+        pg_loss = sample_pg_losses.mean()
```
tianyu-l: same here for "pg"

torchtitan/experiments/rl/grpo.py
```diff
@@ -73,28 +79,41 @@ def __call__(
...
+                ),
+                "loss/ratio/clipped_frac": sequence_scalar_token_weighted_sum(
+                    clipped_frac, response_lens
+                ),
+            }
+            num_local_valid_tokens = response_lens.sum().detach()
+
+        return LossOutput(
+            loss=pg_loss,
+            token_mean_metric_sums=token_mean_metric_sums,
```
tianyu-l: In terms of metrics collecting, what should be done here vs. in trainer? Right now it's scattered.

torchtitan/experiments/rl/grpo.py
```diff
@@ -161,6 +164,56 @@ def _log_samples(items: list[Episode] | list[Completion]) -> None:
         logger.info(f"       A: {item.text[:300].replace(chr(10), ' ').strip()}")
 
 
+# Default console keys per context. Override with
+# `--metrics.train-console-allow-list=.*` or
+# `--metrics.validation-console-allow-list=.*` to print all console keys.
+_RL_TRAIN_HEADLINE_METRIC_PATTERNS = [
```
tianyu-l: I found this to be confusing and not clean. What are they for?

torchtitan/experiments/rl/grpo.py
```diff
@@ -161,6 +164,56 @@ def _log_samples(items: list[Episode] | list[Completion]) -> None:
...
+    r"^timing/step$",
+]
+_RL_VALIDATION_HEADLINE_METRIC_PATTERNS = [
+    r"^validation/reward/_mean$",
+    r"^validation/reward/_max$",
+    r"^validation/response_length/mean$",
+]
+
+
+def _population_std(values: list[float]) -> float:
```
tianyu-l: maybe should be in a util file

torchtitan/experiments/rl/grpo.py
```diff
@@ -494,11 +597,31 @@ def _build_episodes(trajectories: list[Trajectory]) -> list[Episode]:
                         advantage=t.total_reward - group_mean,
                     )
                 )
-        return episodes
 
-    async def validate(self) -> dict:
+        num_groups = len(group_stds)
```
tianyu-l: nit: should get `num_groups` from `groups`, instead of `group_stds`

torchtitan/experiments/rl/grpo.py
```diff
@@ -42,6 +43,11 @@
 from torchtitan.config.manager import ConfigManager
 from torchtitan.experiments.rl.actors.generator import SamplingConfig, VLLMGenerator
 from torchtitan.experiments.rl.actors.trainer import PolicyTrainer
+from torchtitan.experiments.rl.loss.types import (
+    LossOutput,
+    sequence_scalar_token_weighted_sum,
+)
+from torchtitan.experiments.rl.observability import metrics as m
```
tianyu-l, line 50: `m` sounds too short. I think we can import both
- `metrics` so that we can do `metrics.Mean`
- `MetricLogger`, `MetricsConfig`, etc. directly

torchtitan/experiments/rl/observability/metrics/README.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## API at a glance
+
+```
+Metric(key, reduction)
+    │
+    reduction is one of:
+    │
+    ├── Mean(value, count=1.0)   ──►  key/mean
```
tianyu-l: not clear to me what `key/mean` means

torchtitan/experiments/rl/observability/metrics/README.md
```diff
@@ -0,0 +1,114 @@
...
+    ├── Max(value)               ──►  key/max
+    ├── Min(value)               ──►  key/min
+    ├── Std(value)               ──►  key/std
+    ├── Stats(value)             ──►  key/_{max,mean,min,std,sum}
```
tianyu-l: what's this?
also I wonder why there is no "Sum" as a reduction

torchtitan/experiments/rl/observability/metrics/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+```
+Metric(key, reduction)
+    │
+    reduction is one of:
+    │
+    ├── Mean(value, count=1.0)   ──►  key/mean
+    ├── Max(value)               ──►  key/max
+    ├── Min(value)               ──►  key/min
+    ├── Std(value)               ──►  key/std
```
tianyu-l: how do you reduce this, without more info?

torchtitan/experiments/rl/observability/metrics/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+aggregate_metrics(records)  ──►  dict[str, float]
+    groups by (key, reduction type), filters NaN entries,
+    raises on duplicate output keys.
+
+MetricLogger(backends).log(
+    step, records,
+    *,
+    console_allow_list=None,   # None = all, [] = silent, list[regex] = filtered
+    console_prefix="",         # e.g. "validate " for the validation line
```
tianyu-l: what is "validation line"?

torchtitan/experiments/rl/observability/metrics/README.md
```diff
@@ -0,0 +1,114 @@
...
+Any class implementing `log(metrics, step)` and `close()` is a backend. Inherit `MetricBackend` for clarity:
+
+```python
+import json
+
+from torchtitan.experiments.rl.observability.metrics import MetricBackend
+
+class JsonlMetricBackend(MetricBackend):
+    def __init__(self, path):
+        self._fh = open(path, "a")
```
tianyu-l: what is fh short for?

torchtitan/experiments/rl/observability/metrics/README.md
```diff
@@ -0,0 +1,114 @@
...
+        self._fh.write(json.dumps({"step": step, **metrics}) + "\n")
+        self._fh.flush()
+
+    def close(self):
+        self._fh.close()
+```
+
+## Non-goals
+
+- Multi-axis metrics (`step_axis`/`step_value`).
```
tianyu-l: what is multi-axis?

torchtitan/experiments/rl/grpo.py
```diff
@@ -452,33 +521,67 @@ async def setup(
...
                 Trajectory(sample_idx=c.prompt_idx, transitions=[(c, step_result)])
             )
-        return trajectories
+
+        response_lens = [len(c.token_ids) for c in completions]
+        prompt_lens = [len(c.prompt_token_ids) for c in completions]
+        total_lens = [p + r for p, r in zip(prompt_lens, response_lens, strict=True)]
+        truncated = [c.finish_reason == "length" for c in completions]
+        rollout_metrics: list[m.Metric] = [
+            m.Metric("rollout/response_length", m.Mean.from_list(response_lens)),
```
tianyu-l, line 583: what if there are multiple generator actors -- should we just manually collect all results in a list and call `[Metric].from_list`?

torchtitan/experiments/rl/grpo.py
```diff
@@ -237,9 +299,16 @@ def __post_init__(self):
     def __init__(self, config: Config):
         self.config = config
         self._proc_meshes = []
+        self.metric_logger: m.MetricLogger | None = None
```
tianyu-l: should this be in the config, as users might want to customize backend.

torchtitan/experiments/rl/observability/metrics/metric_logger.py
```diff
@@ -0,0 +1,282 @@
...
+        matches = sorted(
+            key for key in metrics if pattern.search(key) and key not in seen_set
+        )
+        for key in matches:
+            seen.append(key)
+            seen_set.add(key)
+    return seen
+
+
+def _fmt_value(value: Any) -> str:
```
tianyu-l: should this be put in core and standardize everything?

torchtitan/experiments/rl/observability/metrics/processor.py
tianyu-l, line 1: Could you help unify the sft MetricLogger with this one? Feels not wise to have different implementation for the same thing. Honestly I think this file looks a bit messy.

torchtitan/experiments/rl/observability/metrics/metric_logger.py
```diff
@@ -0,0 +1,282 @@
...
+    # separator divides them).
+    logger.info("%s\n%s", _CONSOLE_SEPARATOR, "  ".join(parts))
+
+
+# ---------------------------------------------------------------------------
+# MetricLogger
+# ---------------------------------------------------------------------------
+
+
+class MetricLogger:
```
tianyu-l: inherit `Configurable`

torchtitan/experiments/rl/observability/metrics/metric_logger.py
```diff
@@ -0,0 +1,282 @@
...
+        for backend in self._backends:
+            try:
+                backend.close()
+            except Exception:
+                logger.exception(
+                    "metric backend %s failed to close", type(backend).__name__
+                )
+
+    @classmethod
+    def build(
```
tianyu-l: after inheriting Configurable, we shouldn't need this `build` -- just need to implement the Constructor

torchtitan/experiments/rl/observability/metrics/metric_logger.py
```diff
@@ -0,0 +1,282 @@
...
+        cls,
+        config: "MetricLogger.Config",
+        *,
+        log_dir: str | None = None,
+        config_dict: dict[str, Any] | None = None,
```
tianyu-l: I know this is tech debt, but we shouldn't expose this whole config dict to the constructor of a metric logger, although I'm OK to bear with it for a bit more.

torchtitan/experiments/rl/observability/metrics/metric_logger.py
```diff
@@ -0,0 +1,282 @@
...
+        """Reserved for future cadence control. Values other than 1 currently
+        warn; RL still logs every step."""
+
+        train_console_allow_list: list[str] | None = None
+        """Regex patterns selecting console keys for **training** steps.
+        ``None`` = print all keys (alphabetical). ``[]`` = silent."""
+
+        validation_console_allow_list: list[str] | None = None
+        """Regex patterns selecting console keys for **validation** steps.
+        ``None`` = print all keys (alphabetical). ``[]`` = silent."""
```
tianyu-l: they are not used in this class; also, do we really need to keep two different list?

torchtitan/experiments/rl/observability/metrics/types.py
tianyu-l, line 1: maybe metrics/types.py?

torchtitan/experiments/rl/observability/metrics/metric_types.py
```diff
@@ -0,0 +1,331 @@
...
+    - ``Std(value)`` - one observation.
+    - ``Std.from_list(values)`` - many observations.
+
+    Combining records equals (within FP tolerance) the std of the
+    concatenation.
+    """
+
+    output_suffix: ClassVar[str] = "std"
+
+    def __init__(self, value):
```
tianyu-l: I don't know when this is going to be used, but I'd imagine we should include `count` as an arg?

torchtitan/experiments/rl/observability/metrics/types.py
```diff
@@ -0,0 +1,331 @@
...
+    @classmethod
+    def from_list(cls, values: Sequence[Any]) -> Std:
+        coerced = [float(v) for v in values]
+        instance = cls.__new__(cls)
+        instance.value = sum(coerced)
+        instance.count = len(coerced)
+        instance.sum_squares = sum(observation * observation for observation in coerced)
+        return instance
+
+    @classmethod
```
tianyu-l, line 269: why a classmethod?

torchtitan/experiments/rl/observability/metrics/metric_types.py
```diff
@@ -0,0 +1,331 @@
...
+            return {cls.output_suffix: float("nan")}
+        total_value = sum(record.value for record in metrics)
+        total_sum_squares = sum(record.sum_squares for record in metrics)
+        mean_value = total_value / total_count
+        # Clamp tiny negatives that arise from floating-point rounding.
+        variance = max(0.0, total_sum_squares / total_count - mean_value * mean_value)
+        return {cls.output_suffix: math.sqrt(variance)}
+
+
+class Stats(MetricReduction):
```
tianyu-l: name sounds too general

torchtitan/experiments/rl/observability/metrics/metric_types.py
```diff
@@ -0,0 +1,331 @@
...
+    @classmethod
+    def reduce(cls, metrics: Sequence[NoReduce]) -> dict[str, float]:
+        if len(metrics) != 1:
+            raise ValueError(
+                f"NoReduce expects exactly one entry per key; got {len(metrics)}."
+            )
+        return {cls.output_suffix: metrics[0].value}
+
+
+def aggregate_metrics(metrics: Iterable[Metric]) -> dict[str, float]:
```
tianyu-l: can we put this as a method of MetricLogger?

torchtitan/experiments/rl/observability/metrics/metric_types.py
```diff
@@ -0,0 +1,331 @@
...
+    "aggregate_metrics",
+]
+
+
+# ---------------------------------------------------------------------------
+# Reductions
+# ---------------------------------------------------------------------------
+
+
+class MetricReduction(Protocol):
```
tianyu-l: MetricReduction sounds not proper because it has both metric and reduction method.

I would propose
- calling this "Metric"
- calling key+Metric "NamedMetric"

torchtitan/experiments/rl/actors/trainer.py
wwwjn: global_valid_tokens reference: [#here ](https://github.com/pytorch/torchtitan/commit/0cb743558a7aa96eabbc099d1712f85b197e1edb)

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -248,15 +252,73 @@ def _build_model(
...
+                verification.num_tokens_different.detach().to(torch.float64),
+                num_local_valid_tokens.detach().to(torch.float64),
+            ]
+        )
+        # max_value (only one value -- no need to pack)
+        max_value = verification.logprob_diff_max.detach().to(torch.float32)
+
+        # All reduce
+        if loss_mesh is not None:
+            sum_pack = funcol.all_reduce(
```
wwwjn: nit: name `sum_pack` is not very easy to understand.  Maybe `sum_reduced_metrics`

torchtitan/experiments/rl/actors/trainer.py
wwwjn: Agreed, using index to index from the pack is less readable

torchtitan/experiments/rl/actors/utils.py
wwwjn: Maybe we can raise error here. The only reason I could think of is vllm side config is wrong, it doesn't tell vllm engine to return log prob to us

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,73 @@ def extract_response_logprobs(
...
+        zero = torch.zeros((), dtype=torch.float32, device=device)
+        return LogprobVerificationOutput(
+            logprob_diff_sum=zero,
+            logprob_diff_max=zero,
+            num_tokens_different=zero,
+        )
+
+    # Compute the logprob difference
+    diff = titan_flat - vllm_flat
+    return LogprobVerificationOutput(
```
wwwjn: Why we need a new LogprobVerificationOutput class to host the output?

torchtitan/experiments/rl/loss/types.py
wwwjn: This function name is not indicative? and also `response_lens` is the weight? The naming is very confusing in this PR

torchtitan/experiments/rl/observability/metrics/metric_logger.py
wwwjn: Can this logger being cosolidated with main trainers? https://github.com/pytorch/torchtitan/blob/8599818f813ae43d73aeb142a879cbbff9b1d9ad/torchtitan/components/metrics.py#L261? 

For pre-trainning trainer, it would be a purely SPMD trainer, and reduction is handled by SPMD . Can it be mapped to our NoReduce type? Beside that, a lot of code logic  (eg, backends), can be consolidate with main trainer's MetricsLogger?

torchtitan/experiments/rl/actors/trainer.py
felipemello1: we definitely want to log this, but we would have to gate over a `log_every_N_steps` flag, like we do in trainer.py. I didn't add this flag in this PR, i think I will do it as a follow up, when i refactor the losses.

torchtitan/experiments/rl/actors/trainer.py
felipemello1: its a cheap check to confirm its a number, but this should happen upstream, not there.

I do think that we should take the mean -- if we have samples from multiple policies, mean would make that evident. Do you disagree?

 But i need to check if i am actually computing the mean of the policy version somewhere.

torchtitan/experiments/rl/actors/utils.py
felipemello1: its just too defensive. I will simplify these checks and if they are not what we expect, they will just error out when comparing shapes.

torchtitan/experiments/rl/actors/utils.py
felipemello1: 100%, sorry, i let this slip

torchtitan/experiments/rl/actors/utils.py
felipemello1: When we aggregate the metrics, we need to put them in different buckets for all_reduce sum and max. Having a dataclass makes it safer to manage them. 

We could:
a) use a dictionary
b) have some structure to say: (key, all_reduce_type)
c) reduce on the spot, instead in blocks later

(b) would be ideal, and will be possible when we have better metric system for SPMD. Is it fine to leave the Output class for now? if not, i can make it a dictionary again.

torchtitan/experiments/rl/loss/types.py
felipemello1: i plan a follow up adding losses shortly, which will use this folder. If you don't mind, lets leave it here for now and i will have a follow up next week.

torchtitan/experiments/rl/loss/types.py
felipemello1: its part of the loss PR that will follow up: different losses have different metrics associated to them. The dict can hold any arbitrary (key, value)

torchtitan/experiments/rl/loss/types.py
felipemello1: also part of the loss follow up. Sorry i conflated things. I should have been more strict in my review. I will move this to utils for now.

felipemello1: will fix

torchtitan/experiments/rl/observability/metrics/metric_logger.py
felipemello1: "Can this logger being cosolidated with main trainers?" --> I will take a look and check how hard it is to do so. The trainer.py has some different requirements for MetricsProcessor. 

i would prefer to just enable it for RL and leave this for another refactor.  But let me check.

torchtitan/experiments/rl/actors/trainer.py
tianyu-l: > if we have samples from multiple policies, mean would make that evident. Do you disagree?

This might hide potential bugs -- e.g. if users see policy number to be 3, they might think every sample is from version 3, but actually they can come from mixed 2 and 4. Maybe logging max / min would clear the doubt.

torchtitan/experiments/rl/observability/metrics/metric_logger.py
felipemello1: this is slop, sorry. I asked claude to do a last minute change. i will clean things up here before i mark it ready for review again.

torchtitan/experiments/rl/observability/metrics/metric_types.py
felipemello1: "NamedMetric" makes sense but its a bit long. Usually adding a metric takes the whole line. I want to avoid having it break into two lines. Wdyt?

torchtitan/experiments/rl/observability/metrics/metric_types.py
felipemello1: It is the same problem as mean-of-means in loss where you have to do value*local/global.

For this __init__, we assume the user has a single value. If they have multiple, they would use '.from_list', that updates all 3: value, count, sum_squares. This allows us to not have to store a whole list of numbers.

if the user passes a `count` in __init__, they would also have to pass the `sum_squares`. I guess i could expose them as args, default to None, and if None, attribute a value to them. I will do that.

torchtitan/experiments/rl/observability/metrics/types.py
felipemello1: it doesnt depend on any internal attribute, e.g. self.value.

User is expected to pass a List[MetricReduction]. Maybe its a bit misleading, because it is not reducing the values it is holding.

In the aggregator, we do
```python
metric_by_key: dict[str, List[MetricRuduction]] = defaultdict(list)
for named_metric in metrics:
    metric_by_key[named_metric.key].append(named_metric.reduction)

for key, list_reductions in metric_by_key:
    reduce_type = type(list_reductions[0]) # class Std
    reduced_value = reduce_type.reduce(list_reductions)
```

torchtitan/experiments/rl/observability/metrics/metric_types.py
felipemello1: do you think that its sufficient that its imported from `metric_types` and inherits `MetricReduction`?

Maybe `SummaryStats`? but i wanted to use short names due to how long defining a metric is.

torchtitan/experiments/rl/observability/metrics/README.md
felipemello1: will add

torchtitan/experiments/rl/observability/metrics/README.md
felipemello1: -.-` bad description lol

validation step

In the console, it appears as: "Validation | Step ...."
For training, it would be: "Train | Step ..."

So users can differentiate both. Same idea for `console_allow_list`: both stages would want to share different metrics, e.g. grad_norm only in training.

torchtitan/experiments/rl/observability/metrics/README.md
felipemello1: file_handle, but this is bad. Will update.

torchtitan/experiments/rl/observability/metrics/README.md
felipemello1: this should be in the readme. I defined a readme.md that i wrote. I asked claude to port over, and it made some changes and added this section. I should have done another pass.

torchtitan/experiments/rl/tests/test_grpo_metrics.py
felipemello1: yes, i will look into that

torchtitan/experiments/rl/grpo.py
felipemello1: i will fix it here

torchtitan/experiments/rl/grpo.py
felipemello1: policy gradient. Its a common term for rl losses. I will write the full name. The other option is to name it "surrogate", which i dislike more.

torchtitan/experiments/rl/grpo.py
felipemello1: when i did it, was to make it clear what is coming from structured_logger (sl), and what is coming from metrics (m). I wanted to avoid people confusing both systems. So, everything from sl is `sl.log...`, and for experiment metrics its `m.Metric...`.

How do you feel about this argument?

torchtitan/experiments/rl/grpo.py
felipemello1: In my loss design, every loss should compute its own metrics. The trainer still calls all_reduce, but the loss knows the metrics it carries. It has to be that way because every RL loss has a different set of metrics. Please check: https://github.com/meta-pytorch/torchforge/tree/main/src/forge/rl/loss

torchtitan/experiments/rl/grpo.py
felipemello1: When we log metrics, we >50 of them. How do we define which ones to display on console? Also, training and validation want to display different metrics.

Here i defined which metrics i wanted to display on console. It uses a regex to filter the metric keys. Someone could do "loss" for example, it would display any metric that has "loss" in it.

Ideally, this would live in the config. I will move it there.

torchtitan/experiments/rl/grpo.py
felipemello1: I also wasnt sure about where this should live. I don't mind putting it in utils, but it is used only in _build_episodes, which lives here.

torchtitan/experiments/rl/grpo.py
felipemello1: Lets say you have a metric `"rollout/response_length"`.

1. Within each generator, you would define:
```python
m.Metric("rollout/response_length", m.Max.from_list(response_lens))
```

2. In the controller, you would collect all of these Metric objects

3. In the MetricsProcessor you would aggregate them.

e.g.
```python
[
   m.Metric("response_length", 100), # from generator 1
   m.Metric("response_length", 200), # from generator 2
   m.Metric("response_length", 300), # from generator 3
]
```

felipemello1: I will add this example in the readme to make it clearer

torchtitan/experiments/rl/observability/metrics/processor.py
felipemello1: answered in msg below. I will take care of the messiness

felipemello1:
I addressed all comments. Some specific notes:

@tianyu-l

1. **CI hookup.** Will follow up with another PR. There is this one, but i think we need a test-focused PR: https://github.com/pytorch/torchtitan/pull/3314#pullrequestreview-4274936617
2. **Metrics API consolidated.**
    - `MetricLogger` renamed to `MetricsProcessor`; `MetricReduction` renamed to `MetricValue`; `Stats` renamed to `SummaryStats`; `Sum` added.
    - Moved into a new `metric_console_logging.py` console-rendering helpers (`log_to_console`, `fmt_metric_value`, `_filter_allow_list`, `_COLOR_CYCLE`, `_CONSOLE_SEPARATOR`)  with a TODO to unify with main trainer's console-logging utilities. I didn't want to touch the main trainer code in this PR.
3. **README rewrite.** Hopefully some details are easier to grasp now
4. **MetricsProcessor unification with core deferred.** Explicit `TODO` left on the class pointing at `torchtitan/components/metrics.py:MetricsProcessor`.

Did not change:
- `m` alias for `metrics`. We also have "sl" for structured logger, and I want to make it obvious what comes from `.m` and what comes from `.sl`. Let me know you disagree,

@wwwjn

5. **F1 loss-scaling pattern.** Adopted the SFT-trainer pattern you linked: compute `num_global_valid_tokens` before forward, pass it into the loss, and have the loss return per-rank shares pre-normalized so SUM-reducing reconstructs the global token-weighted mean.

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -46,10 +49,24 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass(frozen=True, slots=True)
+class OptimStepOutput:
```
tianyu-l: should we put in types.py? I know that this is trainer specific

torchtitan/experiments/rl/types.py
tianyu-l: tbh I like `TrainingBatch` (prime-rl) more

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -289,14 +363,19 @@ async def forward_backward(self, train_data: list[TrainBatch]) -> dict:
                 f"generation max_tokens."
             )
 
+        # Compute global valid tokens BEFORE forward, so the loss
+        # and metrics can be scaled with (n * local_tokens)/global_tokens.
+        num_local_valid_tokens = torch.tensor(
+            sum(response_lens), device=device, dtype=torch.float32
+        )
+        num_global_valid_tokens = self._reduce_valid_tokens(num_local_valid_tokens)
```
tianyu-l: This is unnecessary. We have all the batch, why shard and do all-reduce?
@wwwjn is refactoring batcher, so fine to leave whatever works (minimally).

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,56 @@ def extract_response_logprobs(
     return result
 
 
+@dataclass(frozen=True, slots=True)
+class LogprobVerificationOutput:
```
tianyu-l: I think it's worth having something like `...PendingReduction` or `Partial...` in the name

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,56 @@ def extract_response_logprobs(
...
+    """Generator vs trainer drift metrics normalized by `num_global_valid_tokens`,
+    so trainer can sum all-reduce them."""
+
+    logprob_diff_mean: torch.Tensor
```
tianyu-l, line 69: are these scalar tensor?

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -47,68 +51,56 @@ def extract_response_logprobs(
...
-    vllm_token_log_probs: list[list[float]],
-    batch_token_log_probs: list[torch.Tensor],
-) -> dict:
-    """
-    Check if vLLM log probs and computed log probs are bit-wise identical,
-    and compute the log ratio (train/generator) between them.
+    generator_token_logprobs: list[list[float]],
+    trainer_token_logprobs: list[torch.Tensor],
+    *,
+    num_global_valid_tokens: torch.Tensor,
```
tianyu-l, line 80: Would be good to note the shapes of these tensors / arrays in the docstring.

It's also not obvious why they are not all tensors, maybe worth noting somewhere.

torchtitan/experiments/rl/observability/metrics/console.py
tianyu-l, line 1: To me, metrics/console.py is better than metrics/metrics_console_logging.py

torchtitan/experiments/rl/grpo.py
```diff
@@ -76,37 +95,54 @@ def __call__(
...
-        metrics = {
-            "pg_loss": pg_loss.item(),
-            "ratio_mean": ratio.mean().item(),
-            "ratio_clipped_frac": (torch.abs(ratio - clipped_ratio) > 1e-6)
-            .float()
-            .mean()
-            .item(),
-        }
-        return pg_loss, metrics
+        pg_loss = _token_weighted_mean(
```
tianyu-l, line 114: cc @wwwjn 
we were discussing this, which seems fixed here

torchtitan/experiments/rl/grpo.py
```diff
@@ -242,9 +285,10 @@ def __init__(self, config: Config):
         self.trainer = None
         self.generator = None
         self._proc_meshes = []
+        self.metrics_processor: m.MetricsProcessor | None = None
```
tianyu-l: this was my question to @wwwjn :
why don't we do `setup()` in constructor?

torchtitan/experiments/rl/grpo.py
```diff
@@ -469,33 +520,67 @@ async def setup(
...
         ]
         completions = self._get_rank_0_value(
             self.generator.generate.call([env.prompt for env in envs]).get()
         )
+
         trajectories: list[Trajectory] = []
+        steps: list[Step] = []
         for c in completions:
             step_result = envs[c.prompt_idx].step(c.text)
+            steps.append(step_result)
```
tianyu-l: Let's be careful.

Here each element (a dict of rewards) is from a different prompt.

By calling it steps you seem to assume this is multi-turn rewards? It is not the case. Here it's only single-turn.

Thus the meaning of `_build_reward_metrics` on it is not clear to me -- you are averaging across prompts, is that what you want?

torchtitan/experiments/rl/grpo.py
```diff
@@ -559,56 +700,86 @@ async def train(self):
...
 
-            # --- Train step --- #
+            # --- train ---
+            t_train_start = time.perf_counter()
             batches = [
                 self._collate_episodes(per_rank_episodes)
                 for per_rank_episodes in self._shard_episodes(episodes)
             ]
-            fwd_bwd_metrics = self._get_rank_0_value(
+            fwd_metrics = self._get_rank_0_value(
```
tianyu-l: why change name?

torchtitan/experiments/rl/observability/metrics/processor.py
tianyu-l, line 1: based on the class in this file, maybe rename to metrics/processor.py?

torchtitan/experiments/rl/observability/metrics/processor.py
```diff
@@ -0,0 +1,218 @@
...
+__all__ = [
+    "MetricBackend",
+    "MetricsProcessor",
+]
+
+
+class MetricsProcessor(Configurable):
+    """Aggregates Metric records and dispatches to backends and console.
+
+    TODO: unify with torchtitan/components/metrics.py:MetricsProcessor.
```
tianyu-l, line 39: This tooling is so cool -- core trainer will be jealous before that!

torchtitan/experiments/rl/actors/trainer.py
wwwjn: Yes, batcher (in controller now) has all the information, so we don't need the extra shard.

torchtitan/experiments/rl/grpo.py
felipemello1: setup() does proc-mesh spawning, TorchStore init, and the initial weight push/pull, which are all async due to monarch. __init__ can't be async.

Monarch .get() is a way around it (call async functions inside of sync functions), but i think it makes sense to keep the `setup`, since we are heavily dependent on monarch and it will also be necessary for the async script.

felipemello1: but it does beg the question: should we move `sync` work to `__init__`, or just have everything in setup? I am not sure. Maybe for now keep it in the setup?

torchtitan/experiments/rl/grpo.py
felipemello1: i removed the `steps` and used the saved `trajectories` directly. 

> you are averaging across prompts, is that what you want?

yes, i want to get all rewards from all prompts and log the mean reward

```python
def _build_reward_metrics(
    prefix: str,
    trajectories: list[Trajectory],
) -> list[m.Metric]:
    """One ``Mean`` metric per observed reward component across trajectories.

    Example::

        trajectories = [
            Trajectory(sample_idx=0, transitions=[(c0, Step(rewards={"correctness": 1.0, "format": 0.5}, done=True))]),
            Trajectory(sample_idx=1, transitions=[(c1, Step(rewards={"correctness": 0.0},               done=True))]),
        ]
        _build_reward_metrics("reward/component", trajectories)
        # -> [
        #      Metric("reward/component/correctness", Mean(sum=1.0, count=2)),  # 0.5
        #      Metric("reward/component/format",      Mean(sum=0.5, count=1)),  # 0.5 - "format" only in trajectory 0
        #    ]
    """
```

torchtitan/experiments/rl/observability/metrics/console.py
felipemello1: renamed

torchtitan/experiments/rl/actors/utils.py
felipemello1: renamed to PartialLogprobDrift

torchtitan/experiments/rl/actors/trainer.py
felipemello1: deleted

torchtitan/experiments/rl/types.py
felipemello1: renamed

torchtitan/experiments/rl/actors/trainer.py
felipemello1: moved

torchtitan/experiments/rl/grpo.py
felipemello1: claude silliness. Thanks for catching.

torchtitan/experiments/rl/observability/metrics/processor.py
felipemello1: done

torchtitan/experiments/rl/actors/utils.py
felipemello1:
```
class PartialLogprobDrift:
    """Per-rank generator-vs-trainer logprob drift awaiting reduction across the loss-mesh.

    Args:
        logprob_diff_mean: Scalar tensor; To be sum-reduced.
        logprob_diff_max: Scalar tensor; To be max-reduced.
        ratio_tokens_different: Scalar tensor; To be sum-reduced.
    """
```

torchtitan/experiments/rl/grpo.py
tianyu-l: I would prefer to
- put sync work in constructor
- maybe give more accurate names than `setup` for the async work

torchtitan/experiments/rl/actors/utils.py
felipemello1:
```python
def verify_logprob_identity(
    generator_token_logprobs: list[list[float]],
    trainer_token_logprobs: list[torch.Tensor],
    *,
    num_global_valid_tokens: torch.Tensor,
    device: torch.device,
) -> PartialLogprobDrift:
    """Compute per-rank drift between generator and trainer logprobs.

    Args:
        generator_token_logprobs (list[list[float]]): generator-side per-token logprobs, shaped
            `[num_episodes_local][response_len_i]`.
        trainer_token_logprobs (list[torch.Tensor]): Trainer-side per-token logprobs, one
            GPU tensor per episode, each of shape `[response_len_i]`.
        num_global_valid_tokens (torch.Tensor): Scalar tensor holding global token count
             across DP ranks. Used to normalize the output metrics.
        device: Device to use for tensor allocation, so metrics are ready for
            reduction across loss_mesh.

    Returns:
        PartialLogprobDrift.
    """
```

@wwwjn 's PR will change this and `trainer_token_logprobs` should just be a single tensor + mask now. I guess whoever merges last can fix this.

torchtitan/experiments/rl/actors/utils.py
tianyu-l: should this also be in types.py then? I'm lacking a way to tell what to put in types.py in general

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -257,22 +261,75 @@ async def sync_log_step(self, step: int, relative_step: int | None = None) -> No
         """Sync the structured-logger step counter from the controller."""
         sl.set_step(step, relative_step=relative_step)
 
+    def reduce_forward_backward_metrics(
```
tianyu-l, line 264: Instead of using plain tensors. I think we should instead use DTensor / spmd_types to handle the reduction, which gives more correctness guarantee.

Probably should be done after / along with @wwwjn 's batcher change.

Please leave a TODO here.

tianyu-l (approved):

PR #3238, opened 2026-05-06
[rl] Graceful shutdown GRPO

felipemello1:
> @daniellepintz hey, i am checking if we can have a better shutdown overall — its still a draft. I will review it better later.

## Problem

`torchtitan/experiments/rl/grpo.py` exits noisily today. Three symptoms:

**1. Finalizer error at exit.** `VLLMGenerator.__del__` calls `torch.cuda.empty_cache()` during interpreter shutdown, when `torch` may already be `None`:

```
Exception ignored in: <function VLLMGenerator.__del__ at ...>
AttributeError: 'NoneType' object has no attribute 'cuda'
```

(Already addressed by #3224 in upstream/main; this PR no longer overlaps with that change.)

**2. `main()` never tears down.** No `try/finally`, so process meshes and actors are not stopped on normal exit, on exception, or on Ctrl-C.

**3. NCCL leak warning + unresponsive Ctrl-C.** Even with cleanup wired up, `[W ProcessGroupNCCL.cpp:1647] destroy_process_group() was not called` still prints because nothing inside the worker destroys the process group. And Ctrl-C does nothing — Python's default SIGINT cancellation is not reliably invoked while Monarch's blocking `endpoint.call(...).get()` calls hold the main thread.

### Per-actor `close` endpoint

`VLLMGenerator.__del__` is removed. Each actor gets a `@endpoint async def close` that releases its own resources:

| actor | what its `close` does |
|---|---|
| `VLLMGenerator` | `renderer.shutdown` → `engine_core.shutdown` → vLLM's `cleanup_dist_env_and_memory()` (destroys NCCL groups, runs GC, empties accelerator cache). Same order vLLM's own `AsyncLLM.shutdown` uses. |
| `PolicyTrainer` | `torch.distributed.destroy_process_group()` if initialized. |

Endpoints are named `close` because `Actor.stop` is reserved by Monarch. `RLTrainer.close()` calls each actor's `close` first, then stops the proc meshes.

### `await` at every Monarch RPC

The train loop previously used `endpoint.call(args).get()` everywhere — sync waits inside `async def`. That blocks the OS thread, freezing the asyncio loop, which is why Python's default SIGINT cancellation never landed. Replacing every `.call().get()` with `await endpoint.call(...)` (eight sites across `setup`, `_collect_rollouts`, `validate`, `train`) yields control to the loop on every RPC — and that's enough for Ctrl-C to work without any custom signal-handling code.

```python
# before (sync wait — freezes the loop)
self.trainer.push_model_state_dict.call().get()

# after (yields to the loop while the worker runs)
await self.trainer.push_model_state_dict.call()
```

### Ctrl-C flow

```
Ctrl-C
  → asyncio.run's default handler calls task.cancel()
  → CancelledError raised at the next await (= the next Monarch RPC, often <1s away)
  → main() catches it, logs "Interrupted; attempting graceful shutdown..."
  → main()'s finally runs RLTrainer.close()
       trainer.close → generator.close → mesh.stop for each mesh
  → exit 0
```

No custom signal handler. No explicit yield. The awaited RPCs *are* the cancellation points.

## Verification

`CUDA_VISIBLE_DEVICES=4,5`, TP=1/TP=1:

| probe | result |
| --- | --- |
| `pytest torchtitan/experiments/rl/tests/test_shutdown.py -v` | **8 passed** |
| Normal exit (2 steps + post-validation) | exit 0; clean log; `Closing` breadcrumb present |
| Ctrl-C mid-training (manual harness) | exit 0; cancel landed before the next step started; `Interrupted; attempting graceful shutdown...` and `Closing` breadcrumbs present |

"Clean log" = `grep -cE "Exception ignored\|VLLMGenerator.__del__\|AttributeError\|destroy_process_group\|Traceback"` returns 0.

Training dynamics are bit-identical to the prior `.call().get()` shape (same Loss, Reward, Logprob diff per step on the smoke config — `await` and `.get()` wait on the same Future, only the thread vs coroutine blocking differs).

`pre-commit run` (file-scoped on the four changed files): all hooks pass; pyrefly skipped (binary missing in env), lychee skipped.

## Out of scope (intentionally)

- **SIGTERM** (cluster-scheduler termination signal). Default OS behavior applies; the process is killed without `close()` running. One extra `loop.add_signal_handler(signal.SIGTERM, ...)` line to add later if a deployment requires it.
- **Cancel-mid-RPC race.** When SIGINT lands during an awaited RPC, the asyncio wrapper around that Future cancels but the underlying Monarch RPC may keep running on the worker for a moment until `mesh.stop()` kills the process. Memory-safe (workers are separate processes); ungraceful for the in-flight RPC. Forge accepts the same trade-off; could be addressed with `asyncio.shield` if needed.

## Test plan

- [x] CPU unit tests pass (8/8)
- [x] Normal-exit GPU run is clean (no `Exception ignored`, no NCCL warning)
- [x] Ctrl-C mid-training exits cleanly with close; visible log lines on interrupt and close start
- [x] file-scoped `pre-commit` clean
- [ ] full `pre-commit run --all-files` (deferred — pyrefly binary missing in this env)
- [ ] CI

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -24,6 +24,7 @@ from torchtitan.protocols.model_spec import ModelSpec
 from torchtitan.tools.utils import has_cuda_capability
 from vllm import EngineArgs, LLMEngine, SamplingParams
 from vllm.config import AttentionConfig, CompilationConfig
+from vllm.distributed.parallel_state import cleanup_dist_env_and_memory
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
 
@@ -361,3 +362,32 @@ class VLLMGenerator(Actor, Configurable):
         logger.debug(
             f"{os.getpid()=} Generator pulled model state dict for policy v{version}"
         )
+
+    @endpoint
+    async def close(self) -> None:
+        """Release the vLLM engine and distributed state.
+
+        vLLM's sync ``LLMEngine`` (what we use) has no public ``shutdown``
+        method; only the async ``AsyncLLM`` does. We tear it down by
+        plumbing through its components in the same order ``AsyncLLM``
+        uses internally:
+
+        1. ``renderer.shutdown()`` — closes thread pools and the
+           multimodal-processor cache.
+        2. ``engine_core.shutdown()`` — stops the model worker and the
+           scheduler.
+        3. ``cleanup_dist_env_and_memory()`` — destroys NCCL / model-
+           parallel process groups, runs ``gc.collect``, empties the
+           accelerator cache.
+
+        Each step runs in a ``try/finally`` so a failure in one step
+        does not skip the next.
+        """
+        if self._engine is not None:
+            renderer = getattr(self._engine, "renderer", None)
+            try:
+                if renderer is not None:
+                    renderer.shutdown()
+            finally:
+                self._engine.engine_core.shutdown()
+        cleanup_dist_env_and_memory()
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -160,6 +160,12 @@ class PolicyTrainer(Actor, Configurable):
             f"PolicyTrainer initialized (dp_rank={self.dp_rank}, dp_size={self.dp_size})"
         )
 
+    @endpoint
+    async def close(self) -> None:
+        """Destroy the worker's torch.distributed process group."""
+        if torch.distributed.is_initialized():
+            torch.distributed.destroy_process_group()
+
     def _load_initial_hf_weights(self, model, checkpoint_path: str) -> None:
         """Load model weights from HF checkpoint using DCP and state_dict_adapter.
 
--- a/torchtitan/experiments/rl/grpo.py
+++ b/torchtitan/experiments/rl/grpo.py
@@ -610,8 +610,7 @@ async def main():
         await rl_trainer.setup()
         await rl_trainer.train()
     except (KeyboardInterrupt, asyncio.CancelledError):
-        # Signal-driven cancel; close() runs in finally below, then exit cleanly.
-        pass
+        logger.info("Interrupted; attempting graceful shutdown...")
     finally:
         await rl_trainer.close()
 
--- a/torchtitan/experiments/rl/grpo.py
+++ b/torchtitan/experiments/rl/grpo.py
@@ -22,11 +22,13 @@ python3 torchtitan/experiments/rl/grpo.py \
 """
 
 import asyncio
+import contextlib
 import logging
 import math
 import os
+import signal
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from dataclasses import dataclass, field
 
 # must run before torch import
@@ -236,15 +238,29 @@ class RLTrainer(Configurable):
 
     def __init__(self, config: Config):
         self.config = config
+        self.trainer = None
+        self.generator = None
         self._proc_meshes = []
 
-    async def cleanup(self):
-        """Stop all proc meshes to release GPU memory."""
-        for mesh in self._proc_meshes:
+    async def close(self):
+        """Best-effort: tear down actors, then stop proc meshes."""
+        logger.info("Closing: tearing down actors and process meshes.")
+        for actor_name, actor in (
+            ("trainer", self.trainer),
+            ("generator", self.generator),
+        ):
+            if actor is None:
+                continue
+            try:
+                await actor.close.call()
+            except Exception:
+                logger.exception("%s.close failed", actor_name)
+
+        for i, mesh in enumerate(self._proc_meshes):
             try:
                 await mesh.stop()
             except Exception:
-                pass
+                logger.exception("mesh.stop[%d] failed", i)
         self._proc_meshes = []
 
     def _get_rank_0_value(self, result, has_gpus: bool = True):
@@ -536,6 +552,10 @@ class RLTrainer(Configurable):
         logger.info(f"Pre:  {_format_validation(pre_validation)}")
 
         for step in range(num_steps):
+            # Ctrl-C cancellation is only observed between blocking Monarch RPCs.
+            # Yield here before starting the next step's RPCs.
+            await asyncio.sleep(0)
+
             step_start = time.perf_counter()
 
             # --- Collect data and create episodes --- #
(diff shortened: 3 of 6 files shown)


torchtitan/experiments/rl/grpo.py
```diff
@@ -589,13 +609,51 @@ async def train(self):
...
+            await rl_trainer.train()
+    finally:
+        await rl_trainer.shutdown()
 
 
 if __name__ == "__main__":
-    asyncio.run(main())
+    try:
+        asyncio.run(main())
+    except (KeyboardInterrupt, asyncio.CancelledError):
```
wwwjn: If main already handles it , why we need this try-except?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -361,3 +362,36 @@ async def pull_model_state_dict(self, version: int) -> None:
...
+           scheduler.
+        3. ``cleanup_dist_env_and_memory()`` — destroys NCCL / model-
+           parallel process groups, runs ``gc.collect``, empties the
+           accelerator cache.
+
+        Each step runs in a ``try/finally`` so a failure in one step
+        does not skip the next.
+        """
+        engine = self._engine
+        self._engine = None
```
wwwjn: Why we remove `self._engine` 's reference to the engine first?

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -160,6 +160,12 @@ def __init__(
             f"PolicyTrainer initialized (dp_rank={self.dp_rank}, dp_size={self.dp_size})"
         )
 
+    @endpoint
+    async def shutdown(self) -> None:
```
tianyu-l: may be good to unify name with trainer.py which uses `close` https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L877

torchtitan/experiments/rl/grpo.py
felipemello1: we dont. Deleted.

torchtitan/experiments/rl/actors/generator.py
felipemello1: silliness of claude. Thanks for catching that. I replaced it with `if self._engine is not None:`

torchtitan/experiments/rl/actors/trainer.py
felipemello1: done

torchtitan/experiments/rl/tests/test_shutdown.py
tianyu-l, line 1: Is this CPU test? Should we create something like https://github.com/pytorch/torchtitan/blob/main/.github/workflows/unit_test_cpu.yaml to guard it in CI (e.g. under the "rl" tag @wwwjn is working on)

tianyu-l, line 1: test_close sounds ambiguous (I thought it was testing numerics being close on first sight) -- maybe use test_actor_close or test_teardown

torchtitan/experiments/rl/grpo.py
```diff
@@ -589,12 +609,49 @@ async def train(self):
         )
 
 
+@contextlib.contextmanager
+def _cancel_on_signal() -> Iterator[None]:
```
tianyu-l: Do we not need this in pretrain because torchrun handles it?

Is this blessed way of handling things if using monarch? @JenniferWang

torchtitan/experiments/rl/tests/test_shutdown.py
felipemello1: changed to shutdown, but it makes me think that we should use shutdown/cleanup. It seems to be most of the libraries also use. I am fine to keep it as "close" though

torchtitan/experiments/rl/grpo.py
felipemello1: tldr: deleted it and replaced .get() with await

-----

turns out, i was addressing a symptom, not a root cause. The script as is, if you try to ctrol+c, it doesnt get interrupted.

the root cause is because we were using .get() instead of await. It works, but cant be interrupted.

So, first I added this `loop = asyncio.get_running_loop()` so we could "enter" the loop, and also added the await sleep(0). I tested it, and I only needed the await sleep(0) to let the keyboard interruption jump in.

But, even better: i changed the .get() to await. Now, as soon as i press ctrol+c, it interrupts, instead of having to wait until await sleep(0).

We can delete both this handler and the sleep(0).

good catch!

tianyu-l: I'd like to get a bit more clarify.

It seems we are changing everything from get() to await, basically using `await` for sync semantics. Is this expected? I would like to have clear separation between
- we need async somewhere to make the job exit gracefully
- showcasing how we should do sync RL training (using `get()` most of the time)

JenniferWang: @dulinriley the expert in shutdown noises from Monarch.

JenniferWang: My understanding is that there's no real `sync` endpoint call in monarch (or Ray). So if you need `async` to make cleanup works, that smells wrong and I agree with @tianyu-l  on the separation of consideration.

torchtitan/experiments/rl/tests/test_shutdown.py
wwwjn: Right, RL CPU unit test is not added to CI in this PR yet. @felipemello1  Do you want to update it to be included? It can also triggered by `rl` tag

torchtitan/experiments/rl/grpo.py
```diff
@@ -448,17 +462,17 @@ async def setup(
         await ts.initialize(mesh=trainer_mesh, strategy=ts.LocalRankStrategy())
 
         # push weights from trainer
-        self.trainer.push_model_state_dict.call().get()
+        await self.trainer.push_model_state_dict.call()
```
wwwjn: Why we need to change it to Monarch Async API? I used `.get()` becuase this is a blocking API which is strictly Sync calls

torchtitan/experiments/rl/grpo.py
wwwjn: A little bit more context: previously I decided to using `.get()` to implement Sync RL learning, because https://meta-pytorch.org/monarch/stable/generated/examples/getting_started.html#sending-a-simple-message this is a blocking API call

torchtitan/experiments/rl/grpo.py
```diff
@@ -536,6 +550,10 @@ async def train(self):
         logger.info(f"Pre:  {_format_validation(pre_validation)}")
 
         for step in range(num_steps):
+            # Yield to the event loop once per step so a pending Ctrl-C
+            # cancellation can land here (between blocking Monarch RPCs).
+            await asyncio.sleep(0)
```
tianyu-l: please leave a TODO for proper fix

tianyu-l (approved):

torchtitan/experiments/rl/grpo.py
dulinriley: So when you call .get() on the main thread, you go here:
https://github.com/meta-pytorch/monarch/blob/main/monarch_hyperactor/src/runtime.rs#L201
That releases the GIL, and awaits for the Rust tokio future to complete.
Every 100 ms it tries to reacquire the GIL and call the interpreter signal checker.

So if you're not seeing the script be responsive to Ctrl-C, there are three main possibilities:
* Was your SIGINT signal handler overwritten? Monarch doesn't override this from the default, which is installed by the Python runtime normally.  In python you can do: `import signal; signal.getsignal(signal.SIGINT)` and see if it's the default or something else. This may not be sufficient to tell you because Python itself overwrites the default to give you KeyboardInterrupt
* The signal_safe_block_on either (a) didn't run the part of the future doing the loop or (b) was unable to acquire the GIL. Why would it not be able to acquire the GIL? Maybe if you are Ctrl-C'ing some very long running native code, it may not release the GIL in a timely fashion, especially if it is itself stuck in some endless loop. 
* Is this not running on the main thread? In that case our implementation just blocks forever

For other reference, here's the CPython default SIGINT handler: https://github.com/python/cpython/blob/main/Modules/signalmodule.c#L226C5-L226C18
It just sets a flag in the interpreter to raise an exception at its next opportunity.
So if you're stuck in native code (C++ or Rust) it won't normally do anything until you actually check for exceptions and raise.

My recommendation to debug: when this happens, attach lldb / gdb and dump the stacks of all threads. We need to see if any thread is holding the GIL for longer than it needs to.

dulinriley: As for the earlier question:

> showcasing how we should do sync RL training (using get() most of the time)

Monarch has no real way to do fully synchronous work. The basic actor communication mechanism is not call and response, it's just a 1-way "send". You don't even know that the send was able to complete and leave your machine yet by the time it returns. We implement call-response by adding an address that the recipient can send back to with a response (a `PortRef<T>`). So it's inherently async, you're waiting for a different actor to drive their own message handling loop until they can send something back

We simulate synchronous work by just blocking your thread until the future has completed. So other than avoiding `async` on your functions, you don't get any benefits from using `.get()` and might have some drawbacks if your main thread could be doing other async work at the same time.

cc @zdevito who might have suggestions about how best to show sync RL training

dulinriley:
One minor thing: you said you did this:
> VLLMGenerator.__del__ is removed. Each actor gets a @endpoint async def close that releases its own resources:

Actors already have a `__cleanup__` you can use that is called automatically when you do `ActorMesh.stop()`, or when the actor stops for any other reason (like if it's orphaned from its creator). I'd recommend using that paradigm over your own `stop` function.

One caveat: cleanup is called with a time-bound, so I wouldn't recommend doing anything that takes > 3 seconds. This can be overridden with the environment variable `HYPERACTOR_CLEANUP_TIMEOUT=3s` but the default is 3 seconds.
https://github.com/meta-pytorch/monarch/blob/main/hyperactor/src/config.rs#L176-L182

dulinriley:
This PR might fix your hang with ctrl-c:
https://github.com/meta-pytorch/monarch/pull/3801

I think our atexit handler was stopping the script from completing shutdown. We will also disallow calling Future.get() inside an asyncio loop maybe by next release

felipemello1:
> We will also disallow calling Future.get() inside an asyncio loop maybe by next release


Thanks @dulinriley ! So, to confirm my understanding: we should **NOT** call .get() inside of `async def` functions?

dulinriley:
> > We will also disallow calling Future.get() inside an asyncio loop maybe by next release
> 
> Thanks @dulinriley ! So, to confirm my understanding: we should **NOT** call .get() inside of `async def` functions?

yes confirmed. Only call .get() in a synchronous context. If you are in an async function just await it instead

PR #3239, opened 2026-05-06
Introduce FlexShard for flexible bucketed parameter sharding

weifengpy:
* #3502
* #3537
* #3603
* #3317
* __->__ #3239

Introduce experimental FlexShard, a placement-driven API for flexible bucketed parameter sharding. FlexShard lets users group parameters into explicit communication buckets and assign each parameter a pluggable placement implementation. The initial runtime is eager for now.

- flex_shard() API with explicit BucketSpecs

- Per-bucket flat byte storage for sharded parameters

- Pluggable Placement contract for local layout, unshard collectives, and gradient reduction

- Shard, Owned, and RaggedShard example placements to prove the placement contract

- RaggedShard supports uneven flattened-prefix sharding, bucket unshard/reduce-grad, and focused CPU/CUDA runtime coverage

- Eager parameter accessors backed by batched bucket unshard hooks

- Per-bucket MixedPrecisionPolicy

- reshard-after-forward path using selective activation checkpointing

- Runtime support is eager-only; will explore graph capture and torch.compile in next PR

- Each bucket currently requires a uniform parameter dtype and placement tuple.

- Mixed precision is supported per bucket.

- CPU offload is represented in the API but rejected until implemented

Test Plan:
- python -m pytest -q torchtitan/experiments/flex_shard/tests/test_flex_shard_ragged_shard.py


torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+    4. Replaces each parameter with a plain tensor annotated with placement metadata
+    5. Registers property-based accessors for eager parameter access
+    6. Stores DStorages on the module (accessible via module.dstorages)
+
+    Each bucket gets its own byte buffer and DStorage, enabling independent
+    all-gather operations per bucket.
+
+    Nested wrapping is supported: apply flex_shard to inner modules first,
+    then to outer modules. The outer module's storage will exclude parameters
+    from already-wrapped inner modules.
```
tianyu-l: IIRC last time we agreed on a fqn-based, root-level bucketing plan, and banning nested wrapping. Is there any obstacle to achieving that?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+        shard_placement_fn: Required callable that maps
+            ``(named_params, mesh)`` to per-parameter placements.
+            The minimal eager path expects one ``Placement`` per parameter.
+            Example ``Shard(0)`` placements are available under
+            ``torchtitan.experiments.flex_shard.example.shard``.
+        buckets: Required list of bucket specifications. A single
+            whole-module bucket can be expressed as ``[BucketSpec(["*"])]``.
+            When ``reshard_after_forward=True``, FlexShard raises if bucket
+            hooks cannot run in both the original forward and activation
+            checkpoint recomputation.
```
tianyu-l, line 85: when would "bucket hooks cannot run in both ..."?

torchtitan/experiments/flex_shard/flex_shard/utils.py
```diff
@@ -0,0 +1,294 @@
...
+        managed_params.append((fqn, param))
+
+    return managed_params
+
+
+def _validate_flex_shard_mesh(mesh: DeviceMesh) -> None:
+    """Validate mesh inputs for FlexShard eager mode."""
+    if mesh.ndim != 1:
+        raise ValueError(
+            f"flex_shard requires a 1D DeviceMesh, but got {mesh.ndim}D mesh"
```
tianyu-l, line 194: add a TODO here for nD mesh support (seems we don't have a design yet), or raise `NotImplementedError`

torchtitan/experiments/flex_shard/flex_shard/utils.py
```diff
@@ -0,0 +1,294 @@
...
+
+def _validate_eager_params(
+    named_params: list[tuple[str, nn.Parameter]],
+    expected_device: torch.device | None = None,
+) -> None:
+    """Validate parameters supported by the eager-only path."""
+    try:
+        from torch.distributed.tensor import DTensor
+    except ImportError:
+        DTensor = None
```
tianyu-l, line 221: why optional dependency?

torchtitan/experiments/flex_shard/flex_shard/utils.py
```diff
@@ -0,0 +1,294 @@
...
+            raise ValueError(
+                "FlexShard eager mode expects plain parameters; "
+                f"{fqn!r} is a DTensor. DTensor composition is not supported yet."
+            )
+        if (
+            expected_device is not None
+            and param.device.type != "meta"
+            and param.device != expected_device
+        ):
+            raise ValueError(
```
tianyu-l, line 234: why don't we require params to be either all on meta device, or all not on?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+    all_params_meta = all(param.device.type == "meta" for _, param in named_params)
+    device = (
+        torch.device("meta") if all_params_meta else _get_device_from_mesh(shard_mesh)
+    )
+    _validate_eager_params(
+        named_params,
+        expected_device=None if all_params_meta else device,
+    )
+
+    param_placements = shard_placement_fn(named_params, shard_mesh)
```
tianyu-l: What's the design for mixed dense and sparse mesh? Please leave a TODO here if there's no concrete design yet.

torchtitan/experiments/flex_shard/flex_shard/utils.py
```diff
@@ -0,0 +1,294 @@
...
+            if placements != reference_placements:
+                raise ValueError(
+                    f"Bucket {bucket_idx} "
+                    f"{buckets[bucket_idx].patterns} "
+                    f"has mixed placements: {fqns[0]!r} uses "
+                    f"{reference_placements!r} but {fqn!r} uses "
+                    f"{placements!r}. All params in a FlexShard bucket must "
+                    "share the same placement tuple because bucket collectives "
+                    "use one placement layout. Split parameters with different "
+                    "placements into separate buckets."
```
tianyu-l, line 307: oh, makes sense

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+        raise TypeError("flex_shard buckets must be a list of BucketSpec objects.")
+    for bucket in buckets:
+        if bucket.offload_policy is not None:
+            raise NotImplementedError(
+                "FlexShard eager mode does not yet support BucketSpec.offload_policy."
+            )
+
+    param_fqns = [fqn for fqn, _ in named_params]
+    bucket_assignments = _assign_params_to_buckets(param_fqns, buckets)
+    _validate_bucket_placements(
```
tianyu-l: there's already some validation in `_assign_params_to_buckets` -- does it make sense to put the checks in `_validate_bucket_placements` there, too?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+    device: torch.device
+    param_placements: dict[str, tuple[Placement, ...]]
+    bucket_assignments: list[list[str]]
+
+
+def _prepare_flex_shard_inputs(
+    module: nn.Module,
+    mesh: DeviceMesh,
+    shard_placement_fn: PlacementFn,
+    buckets: list[BucketSpec],
```
tianyu-l: curious what'd the implementation for `FlatShard` be where these two args have close relationship

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+
+@dataclass(frozen=True)
+class PreparedFlexShardInputs:
+    """Validated inputs and derived setup state for flex_shard()."""
+
+    named_params: list[tuple[str, nn.Parameter]]
+    shard_mesh: DeviceMesh
+    device: torch.device
+    param_placements: dict[str, tuple[Placement, ...]]
+    bucket_assignments: list[list[str]]
```
tianyu-l: does `param_placements` already cover `named_params`?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+
+    storages, fqn_to_bucket_spec = _materialize_bucket_storages(
+        module,
+        inputs.named_params,
+        inputs.bucket_assignments,
+        buckets,
+        inputs.param_placements,
+        inputs.shard_mesh,
+        inputs.device,
```
tianyu-l: why don't we just send `inputs`?

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+    storages: list[DStorage] = []
+    fqn_to_bucket_spec: dict[str, BucketSpec] = {}
+
+    for bucket_idx, bucket_fqns in enumerate(bucket_assignments):
+        if not bucket_fqns:
+            continue
+
+        bucket_spec = buckets[bucket_idx]
+        for fqn in bucket_fqns:
+            fqn_to_bucket_spec[fqn] = bucket_spec
```
tianyu-l: This doesn't need to be generated in this function? Can be generated in prepare_inputs

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+    storage_nbytes: int = 0  # bytes reserved for this param's local storage
+    global_numel: int = 0  # total elements in unsharded param
+
+    @property
+    def placement(self) -> Placement:
+        """The single placement supported by the minimal eager path."""
+        return _get_single_placement(self.placements)
+
+
+class DStorage:
```
tianyu-l: conceptually does one DStorage always correspond to one bucket? If so, is it proper that we name it `ShardedBucketStorage`?

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+        info = self._param_infos[fqn]
+        return info.placement.make_local_storage_view(self._byte_storage, info)
+
+
+def _materialize_bucket_storages(
+    module: nn.Module,
+    named_params: list[tuple[str, nn.Parameter]],
+    bucket_assignments: list[list[str]],
+    buckets: list[BucketSpec],
+    param_placements: dict[str, tuple[Placement, ...]],
```
tianyu-l: I somehow think these can be grouped more logically in a (maybe bigger) dataclass

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+            )
+            param_infos[fqn] = info
+
+        return param_infos, current_byte_offset
+
+    def copy_params_from(
+        self,
+        named_params: list[tuple[str, nn.Parameter]],
+    ) -> None:
+        """Pack original parameter data into byte storage."""
```
tianyu-l, line 438: Seems this function packs the local (sharded) part of param into byte storage. Let's make it clear in function name and docstring.

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+    def get_local_view(self, fqn: str) -> torch.Tensor:
+        """Get the local tensor view for a parameter by FQN (from sharded storage)."""
+        info = self._param_infos[fqn]
+        return info.placement.make_local_storage_view(self._byte_storage, info)
+
+
+def _materialize_bucket_storages(
+    module: nn.Module,
+    named_params: list[tuple[str, nn.Parameter]],
+    bucket_assignments: list[list[str]],
```
tianyu-l: let's use a more clear name, like `fqn_to_bucket_spec`

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,675 @@
...
+
+def _get_bucket_autograd_unshard_unsupported_reason(
+    storage: DStorage,
+) -> str | None:
+    """Return why a reshard-after-forward bucket cannot use the custom autograd path."""
+    infos = list(storage._param_infos.values())
+    if not infos:
+        return None
+    if storage.byte_storage.device.type != "cuda":
+        return f"storage is on {storage.byte_storage.device.type}"
```
tianyu-l: what is the fundamental reason when device is not cuda?

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,675 @@
...
+
+def _storage_requires_batched_unshard(storage: DStorage) -> bool:
+    """Return whether parameter access must use hook-provided tensors."""
+    return bool(storage._param_infos)
+
+
+def _storage_uses_bucket_autograd_unshard(storage: DStorage) -> bool:
+    """Return whether reshard-after-forward should use the custom bucket autograd path."""
+    if not storage._reshard_after_forward:
+        return False
```
tianyu-l: what's the consideration here? If we reshard_after_forward, can we not use autograd path or does it add too much overhead?

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,675 @@
...
+    if not storage._param_infos:
+        return False
```
tianyu-l: when would this happen?

torchtitan/experiments/flex_shard/flex_shard/param_access.py
```diff
@@ -0,0 +1,288 @@
...
+                    leaf_module = getattr(leaf_module, part)
+            else:
+                leaf_module = child
+        if hasattr(leaf_module, "_checkpoint_wrapped_module"):
+            leaf_module = leaf_module._checkpoint_wrapped_module
+        return cls(leaf_module, parts[-1])
+
+
+@dataclass
+class EagerParamAccessState:
```
tianyu-l: to myself: didn't quite get what this is doing -- skipping for now

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,675 @@
...
+    )
+
+
+def _create_eager_param_states(
+    module: nn.Module,
+    storages: list[DStorage],
+    fqn_to_bucket_spec: dict[str, BucketSpec],
+    device: torch.device,
+) -> dict[nn.Module, dict[str, EagerParamAccessState]]:
+    """Create eager parameter access state grouped by owning leaf module."""
```
tianyu-l: can benefit from a more detailed docstring

torchtitan/experiments/flex_shard/flex_shard/unsharded_param_access.py
```diff
@@ -0,0 +1,288 @@
...
+
+def _register_param_accessors(
+    module: nn.Module,
+    param_states: dict[str, EagerParamAccessState],
+) -> None:
+    """Register per-parameter property getters that read eager access state.
+
+    Uses dynamic subclass creation (not nn.utils.parametrize) to avoid
+    state_dict key mangling. state_dict reads self._parameters directly,
+    bypassing property getters.
```
tianyu-l: this seems the same trick as used by SimpleFSDP for state_dict? https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/graph_trainer/simple_fsdp.py#L135

For SimpleFSDP to access single param (during init, not in state_dict), we had https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/graph_trainer/simple_fsdp.py#L33

What's the mechanism here?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+        storages,
+        fqn_to_bucket_spec,
+        inputs.device,
+    )
+    _register_module_param_accessors(module_param_map)
+
+    # Reshard-after-forward: in eager mode, wrap each layer in checkpoint with
+    # a selective policy that recomputes only collective ops (all-gather,
+    # broadcast), saving compute ops to avoid redundant work.
+    reshard_storages = [s for s in storages if s._reshard_after_forward]
```
tianyu-l: does it make sense to specify MUST_SAVE when reshard_after_forward is False, which unifies the treatment regardless of the flag?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,239 @@
...
+    if reshard_storages:
+        _apply_reshard_after_forward(module, reshard_storages)
+
+    # Install batched all-gather hooks for eager mode when the storage layout
+    # supports one collective per bucket.
+    _install_batched_allgather_hooks(storages, module_param_map)
```
tianyu-l: didn't review this very carefully yet -- I'm assuming it handles prefetching / reordering in a similar fashion to FSDP2?

torchtitan/experiments/flex_shard/example/shard.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def __repr__(self) -> str:
+        return f"Shard({self.dim})"
+
+    @override
+    def compute_local_shape(
+        self, global_shape: torch.Size, rank: int, world_size: int
+    ) -> torch.Size:
+        dim_size = global_shape[self.dim]
+        chunk = (dim_size + world_size - 1) // world_size
```
aditvenk, line 76: This seems different than DTensor's uneven chunk distribution approach -- could this become an issue for inter-op with DTensor for checkpointing or something else?

torchtitan/experiments/flex_shard/example/shard.py
```diff
@@ -0,0 +1,158 @@
...
+        shape = list(global_shape)
+        shape[self.dim] = local_dim
+        return torch.Size(shape)
+
+    @override
+    def compute_local_numel(
+        self, global_shape: torch.Size, rank: int, world_size: int
+    ) -> int:
+        local_shape = self.compute_local_shape(global_shape, rank, world_size)
+        numel = 1
```
aditvenk: nit: can replace with math.prod(local_shape)

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        rank: int,
+        world_size: int,
+    ) -> LocalStorageLayout:
+        """Return the local storage layout for one parameter."""
+        local_shape = self.compute_local_shape(global_shape, rank, world_size)
+        local_numel = self.compute_local_numel(global_shape, rank, world_size)
+        return LocalStorageLayout(
+            local_shape=local_shape,
+            local_numel=local_numel,
+            storage_nbytes=local_numel * dtype.itemsize,
```
aditvenk, line 116: this assumes no padding; is that always true? Or could this be implementation-dependent based on the sub-class?

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Each subclass implements per-param sharding, storage layout, full-parameter
+    assembly, and gradient-reduction packing.
+    """
+
+    def compute_local_numel(
+        self, global_shape: torch.Size, rank: int, world_size: int
+    ) -> int:
+        """How many elements this rank holds for a param with global_shape."""
+        raise NotImplementedError
```
aditvenk: Instead of raising, can use abc.ABC & @abstractmethod instead. This will catch missing implementations earlier than runtime

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class Placement:
+    """Base class for FlexShard placement strategies.
+
+    Each subclass implements per-param sharding, storage layout, full-parameter
+    assembly, and gradient-reduction packing.
+    """
+
+    def compute_local_numel(
```
aditvenk, line 78: Are both compue_local_numel and compute_local_shape  needed? Can't we compute numel from local shape?

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+@dataclass(frozen=True)
+class LocalStorageLayout:
+    """Placement-owned local storage layout for one parameter."""
+
+    local_shape: torch.Size
+    local_numel: int
+    storage_nbytes: int
+
+
+class Placement:
```
aditvenk: __eq__ and __hash__ contracts missing from base class. We should add to ensure each subclass implements this.

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        info: ParamInfo,
+    ) -> torch.Tensor:
+        """Return the exposed local parameter view from bucket storage."""
+        nbytes = info.local_numel * info.dtype.itemsize
+        byte_view = byte_storage[info.byte_offset : info.byte_offset + nbytes]
+        return byte_view.view(info.dtype).view(info.local_shape)
+
+    def pack_reduce_grad(
+        self,
+        tensors: list[torch.Tensor],
```
aditvenk, line 171: Is it intentional that Placement layer is packing/unpacking on entire bucket of parameters in pack_reduce_graph / unpack_reduced_graph ? 

For other methods, the Placement class is operating on per-param basis; so we could implement pack/unpack one grad in this class and have the bucketing layer handle the multi-parameter concerns.

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        infos: list[ParamInfo],
+        world_size: int,
+    ) -> tuple[torch.Tensor, Any]:
+        """Pack full gradients into a flat reduce-scatter input buffer."""
+        raise NotImplementedError
+
+    def unpack_reduced_grad(
```
aditvenk: nit: API name is "reduced" here vs "reduce" in pack_reduce_grad

torchtitan/experiments/flex_shard/example/shard.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @override
+    def assemble_from_shards(
+        self,
+        per_rank_shards: list[torch.Tensor],
+        global_shape: torch.Size,
+        dtype: torch.dtype,
+    ) -> torch.Tensor:
+        non_empty = [s for s in per_rank_shards if s.numel() > 0]
+        if non_empty:
```
aditvenk: When would non_empty be False in realistic scenario?

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+        param_placements: dict[str, tuple[Placement, ...]],
+        mesh: DeviceMesh,
+        device: torch.device,
+        bucket_spec: BucketSpec,
+    ) -> DStorage:
+        """Create storage for one bucket and install its sharded parameters."""
+        param_infos, total_bytes = cls.create_param_infos(
+            named_params,
+            mesh,
+            param_placements,
```
tianyu-l, line 199: only one placement is enough? because all placements in the same bucket should be identical

torchtitan/experiments/flex_shard/flex_shard/utils.py
```diff
@@ -0,0 +1,305 @@
...
+                raise ValueError(
+                    f"Bucket {bucket_idx} "
+                    f"{buckets[bucket_idx].patterns} "
+                    f"has mixed parameter dtypes: {fqns[0]!r} uses "
+                    f"{reference_dtype} but {fqn!r} uses {dtype}. "
+                    "All params in a FlexShard storage must share the same "
+                    "dtype."
+                )
+            placements = param_placements[fqn]
+            if placements != reference_placements:
```
tianyu-l, line 298: curious how would this work for `FlatShard` (what equality means for it)

torchtitan/experiments/flex_shard/flex_shard/utils.py
```diff
@@ -0,0 +1,305 @@
...
+    if device_module is None:
+        return torch.device(mesh.device_type)
+    return torch.device(mesh.device_type, device_module.current_device())
+
+
+def _validate_eager_params(
+    named_params: list[tuple[str, nn.Parameter]],
+    expected_device: torch.device | None = None,
+) -> None:
+    """Validate parameters supported by the eager-only path."""
```
tianyu-l, line 217: what makes it eager-only?

torchtitan/experiments/flex_shard/flex_shard/param_access.py
```diff
@@ -0,0 +1,288 @@
...
+    placements: tuple[Placement, ...],
+    global_shape: torch.Size,
+    global_stride: tuple[int, ...],
+    mesh: DeviceMesh,
+) -> None:
+    """Annotate a tensor with FlexShard placement metadata."""
+    tensor._placements = placements
+    tensor._global_shape = global_shape
+    tensor._global_stride = global_stride
+    tensor._mesh = mesh
```
tianyu-l: does a parameter always need mesh?

torchtitan/experiments/flex_shard/flex_shard/unsharded_param_access.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+
+
+def _attach_flex_shard_module_state(
+    module: nn.Module,
+    storages: list[DStorage],
+) -> None:
+    """Attach FlexShard state and mixin accessors to a module."""
+    setattr(module, _DSTORAGES_ATTR, storages)
+    setattr(module, _DSTORAGE_ATTR, storages[0] if storages else None)
```
tianyu-l: what makes bucket 0 in a module special?

torchtitan/experiments/flex_shard/flex_shard/param_access.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+_parametrized_module_class_counter = 0
+
+
+def _register_param_accessors(
+    module: nn.Module,
+    param_states: dict[str, EagerParamAccessState],
+) -> None:
+    """Register per-parameter property getters that read eager access state.
```
tianyu-l: let's add a docstring on the states change conditions of the underlying state machine.

torchtitan/experiments/flex_shard/flex_shard/reshard_after_forward.py
```diff
@@ -0,0 +1,289 @@
...
+        with ctx:
+            yield
+    finally:
+        recompute_state.exit_recompute(token)
+
+
+# These produce the unsharded param tensors that we want freed per-layer.
+_FLEX_SHARD_COLLECTIVE_OPS = {
+    torch.ops._c10d_functional.all_gather_into_tensor.default,
+    torch.ops._c10d_functional.wait_tensor.default,
```
tianyu-l, line 49: this for sure interfere with non-FlexShard ops
https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/activation_checkpoint.py#L28

E.g. TP all-gather being recomputed or not can be impacted by `reshard_after_forward`?
Also, in the long term when we have multiple types of custom Placements, maintaining this dict is not sustainable. Maybe we should
- require custom ops in Placement classes
- use them in AC policy here

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,681 @@
...
+
+def _install_batched_allgather_hooks(
+    storages: list,
+    module_param_map: dict[nn.Module, dict[str, EagerParamAccessState]],
+) -> None:
+    """Install pre/post forward hooks for batched per-bucket all-gather.
+
+    In eager mode, each DStorage's pre-forward hook starts a single batched
+    all-gather unshard call (one collective per bucket), then sets
+    _pre_gathered on each parameter access state so the property getter can
```
tianyu-l: why it's called pre-gathered, if it's also used for module in after-all-gather states?

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,681 @@
...
+    _PARAM_FQN_ATTR,
+    EagerParamAccessState,
+    ParamModuleInfo,
+)
+from .utils import _get_storage_debug_fqn, _record_function_if_eager
+
+
+logger = logging.getLogger(__name__)
+
+ParamEntry = tuple[ParamModuleInfo, EagerParamAccessState, ParamInfo]
```
tianyu-l: would be good to clarify when would each one be used

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,681 @@
...
+            contexts = {}
+            setattr(root_module, _EAGER_COMM_CONTEXTS_ATTR, contexts)
+
+        context = contexts.get(device)
+        if context is None:
+            device_handle = _get_device_handle(device.type)
+            context = cls(
+                device_handle=device_handle,
+                all_gather_stream=device_handle.Stream(priority=-1),
+                reduce_scatter_stream=device_handle.Stream(priority=-1),
```
tianyu-l: Not sure how to read this -- would each bucket create a new pair of Stream?

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,681 @@
...
+            and target is self.storage._module
+            and any(
+                hasattr(child, "_checkpoint_wrapped_module")
+                for child in self.storage._module.modules()
+                if child is not self.storage._module
+            )
+        ):
+            logger.debug(
+                "Skipping root-level batched all-gather hook because child "
+                "checkpoint recomputation would not replay the root hook.",
```
tianyu-l: oh didn't understand this

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,681 @@
...
+        elif common and common not in prefixes:
+            common = ""
+        if not common:
+            return self.storage._module
+        mod = self.storage._module
+        for part in common.split("."):
+            mod = getattr(mod, part)
+        return mod
+
+    def hook_target_module(self) -> nn.Module | None:
```
tianyu-l: didn't understand this function

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
tianyu-l: The orders between the two seems:
- first apply reshard_after_forward
- then add allgather hooks

but logically it should be first having the hooks -> then reshard and replay the hooks during backward

tianyu-l (commented):
Oh what a profound PR! I'm still trying to learn the basics so apologies if my questions are only scratching the surface.

torchtitan/experiments/flex_shard/example/shard.py
weifengpy: good catch! I updated `extract_local_shard` to match DTensor uneven sharding

this particularly R52 line is same to pytorch dtensor, `full_chunk_size = (curr_local_size + num_chunks - 1) // num_chunks`

torchtitan/experiments/flex_shard/example/shard.py
weifengpy: realistically we won't empty case with 0 numel becuase model.parameters() should hav numel > 0

I asserted on non-empty per_rank_shards to make the code simplier

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: good point. I added code comment and changed it to dataclass for better readability

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: all buckets under the root_module share AG/RS stream. `get_or_create` might be confusing. I updated PR to split into 2 methods
```
BucketCommContext.get(...)
BucketCommContext.create(...)
```

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: for all-gather, we need `_local_shards` to get `[local shard of p0, local shard of p1, local shard of p2]`

use_autograd=True/False is tricky though. I am checking whether I can use use_autograd=True regardless of reshard_after_forward=True/False

weifengpy: renamed to `resolve_bucket_forward_hook_module` to better readability

for a bucket with `layers.0.attn.weight` and `layers.0.mlp.weight`, we should hook `layers.0`

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: I added code comment to help understanding. 

Suppose the model has `layers.1` wrapped with activation checkpointing, but `layers.0` is not:

`BucketSpec(["embed.*", "layers.1.*", "output.*"], reshard_after_forward=True)`

During backward, activation checkpointing recomputes only the checkpointed region for `layers.1`; it does not call `model.__call__` again:

```
  checkpoint backward
    layers.1 recompute forward
```

The fix is to split the bucket so params used by `layers.1` are in a bucket whose hook lands on `layers.1`

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
ezyang: Do we have to call this Placement lol, DTensor already took that name

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
```
ezyang, line 5: Consider asking an LLM to make the documentation in this file more elaborate, and then edit it yourself, it will help.

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+
+    def copy_param_to_storage(
+        self,
+        byte_storage: torch.Tensor,
+        info: ParamInfo,
+        param: torch.Tensor,
+        rank: int,
+        world_size: int,
+    ) -> None:
```
ezyang, line 139: Is use of this function mandatory? What if I have a checkpoint that is already the correct local shard (or, more horribly, the *wrong* shard), do I have to gather the entire parameter before copying it to storage?

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+import torch.nn as nn
+from torch._prims_common import make_contiguous_strides_for
+
+from .param_access import set_sharding_info
+from .utils import _get_single_placement, _set_param_on_module
+
+if TYPE_CHECKING:
+    from torch.distributed.device_mesh import DeviceMesh
+
+    from .placement_contract import Placement
```
ezyang: The circularity here is annoying. See if an LLM can construct a module hierarchy that isn't circular

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
ezyang: It's not obvious if I call this multiple times to write multiple parameters into a single byte storage? So I actually end up running this on literally *every* parameter, even if most of them don't copy into this storage, is that right?

torchtitan/experiments/flex_shard/example/ragged_shard.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# TODO: Add a ragged shard placement example once uneven shard semantics are finalized.
```
ezyang: Nah, you should just implement exactly the veScale semantics, to show the abstraction is strong enough.

torchtitan/experiments/flex_shard/example/flat_shard.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# TODO: Add a flat shard placement example once the placement contract is finalized.
```
ezyang, line 7: I think it is important to do these alternate placements sooner rather than later, so as to flush out kinks in the placement contract (e.g., it being awkward or not general/specific enough in some situations.)

torchtitan/experiments/flex_shard/example/shard.py
```diff
@@ -0,0 +1,157 @@
...
+        shape = list(global_shape)
+        shape[self.dim] = local_dim
+        return torch.Size(shape)
+
+    @override
+    def compute_local_numel(
+        self, global_shape: torch.Size, rank: int, world_size: int
+    ) -> int:
+        local_shape = self.compute_local_shape(global_shape, rank, world_size)
+        return math.prod(local_shape)
```
ezyang: Why is this not *always* the implementation of `compute_local_numel`?

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        self, global_shape: torch.Size, rank: int, world_size: int
+    ) -> int:
+        """How many elements this rank holds for a param with global_shape."""
+        raise NotImplementedError
+
+    def compute_local_shape(
+        self, global_shape: torch.Size, rank: int, world_size: int
+    ) -> torch.Size:
+        """Local shape this rank holds for a param with global_shape."""
+        raise NotImplementedError
```
ezyang, line 88: It's kind of annoying that you're doing inheritance here. Why can't Shard/RaggedShard etc be a lightweight Protocol and we do composition rather than inheritance? There's no state on this superclass so inheritance isn't buying you anything, and doing inheritance makes it unclear which methods are fair game to override versus not.

ezyang: Similar energy to https://github.com/pytorch/torchtitan/pull/3239/changes#r3231107051

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+    Args:
+        param_dtype: Dtype for forward compute. Parameters are all-gathered
+            in storage dtype, then cast to param_dtype. If None, no cast.
+        reduce_dtype: Dtype for gradient reduction. Gradients are cast to
+            this dtype before reduce-scatter. If None, uses param_dtype
+            (or storage dtype if param_dtype is also None).
+    """
+
+    param_dtype: torch.dtype | None = None
+    reduce_dtype: torch.dtype | None = None
```
ezyang, line 49: No, this is too inflexible. We need to support things like custom all-reduce algorithms / stochastic rounding / etc. We need an open registration mechanism here.

There is a compositionality problem with Placement here. Naively, you might hope the mixed precision policy can compose with the Placement orthogonally. But this isn't always true. For example, `reduce_scatter` goes to two different functions depending if you're doing an even or uneven scatter. It is easy to imagine someone writing a stochastic rounding kernel that *only* works if things are even, but not if they're uneven.  At the end of the day, if you want efficient comms, you need comms kernels. We can provide default less efficient implementations but people need to be able to swap them out with optimal ones.

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def unpack_reduced_grad(
+        self,
+        recv_buf: torch.Tensor,
+        infos: list[ParamInfo],
+        layout: Any,
+        rank: int,
+        world_size: int,
+    ) -> list[torch.Tensor]:
+        """Unpack a flat reduce-scatter output buffer into local grad shards."""
```
ezyang: It is good to explicitly point out if dtype conversion happens here or not.

torchtitan/experiments/flex_shard/example/ragged_shard.py
ezyang: cc @aorenste, let's be thinking about how to make compile-on-one-rank work with things like ragged shard here.

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        """Local shape this rank holds for a param with global_shape."""
+        raise NotImplementedError
+
+    def extract_local_shard(
+        self,
+        param: torch.Tensor,
+        rank: int,
+        world_size: int,
+    ) -> torch.Tensor:
+        """Extract this rank's local shard from the full parameter."""
```
ezyang: Does the output tensor here have to be contiguous?

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        raise NotImplementedError
+
+    def assemble_from_shards(
+        self,
+        per_rank_shards: list[torch.Tensor],
+        global_shape: torch.Size,
+        dtype: torch.dtype,
+    ) -> torch.Tensor:
+        """Reconstruct the full parameter from typed per-rank shards."""
+        raise NotImplementedError
```
ezyang, line 101: I feel like this forces a situation where FSDP must all gather the bucketed parameters (which were bucked per parameter, as opposed to flat) to then do a separate kernel to put them together. But of course, one can always imagine writing a collective to do this directly in one go, and it would be good for this to be expressible in the API.

torchtitan/experiments/flex_shard/flex_shard/bucket_collectives.py
```diff
@@ -0,0 +1,456 @@
...
+    recv_buf = torch.empty(
+        send_buf.numel() // world_size,
+        dtype=send_buf.dtype,
+        device=send_buf.device,
+    )
+    with _record_comm_if_eager("FlexShard::reduce_scatter", debug_fqn):
+        dist.reduce_scatter_tensor(
+            output=recv_buf,
+            input=send_buf,
+            op=dist.ReduceOp.AVG,
```
ezyang: The choice to continue to do AVG here is quite dangerous and while it is a helpful default for many users, we should expect more out torchtitan. There is some relationship to spmd types and delayed grad scaling here. (cc @pianpwk )

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+    FlexShard path currently rejects non-None offload policies before storage
+    materialization.
+
+    Args:
+        pin_memory: Whether to pin CPU memory for faster H2D/D2H
+            transfers via DMA. Set to False if insufficient CPU memory.
+            Default True.
+    """
+
+    pin_memory: bool = True
```
ezyang, line 66: I'd probably just not include this at all, if you aren't going to implement it right now.

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
ezyang: Maybe the situation I have described is unrealistic, since it's still quite hard to beat NCCL on inter-node comms, and that's likely what one will be doing here.

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,693 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
```
ezyang, line 5: I assume this is largely modeled off of the FSDP2 runtime

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+        offload_policy: CPU offload policy for this bucket. TODO: implement
+            and test CPU offload before allowing this in flex_shard().
+        reshard_after_forward: Whether to free this bucket's unsharded
+            parameters after forward and recompute them in backward. This
+            defaults to True. Buckets that reshard after forward must have
+            hooks that run in both the original forward and activation
+            checkpoint recomputation.
+    """
+
+    patterns: list[str]
```
ezyang, line 101: I hope to see this normalized away once we get to internal data structures, maybe.

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,246 @@
...
```
ezyang: It would actually be cool if this was optional. The idea is that you can use FlexShard to get your module into the correct form that your bucketing says it should be, but then you manually call the all gathers / reduce scatters in your code. @gchanan was telling me about a FAIR person who wanted to write their code this way.

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+
+from .param_access import set_sharding_info
+from .utils import _get_single_placement, _set_param_on_module
+
+if TYPE_CHECKING:
+    from torch.distributed.device_mesh import DeviceMesh
+
+    from .placement_contract import Placement
+    from .reshard_after_forward import _ReshardAfterForwardRecomputeState
+
```
ezyang, line 26: It's pretty unclear which of these structs are intended to be user facing API, versus internal bookkeeping.

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
```diff
@@ -0,0 +1,415 @@
...
+    patterns: list[str]
+    mp_policy: MixedPrecisionPolicy | None = None
+    offload_policy: OffloadPolicy | None = None
+    reshard_after_forward: bool = True
+
+
+def _assign_params_to_buckets(
+    param_fqns: list[str],
+    buckets: list[BucketSpec],
+) -> list[list[str]]:
```
ezyang: Consider using type aliases to make it clear what the position in the list means (which bucket?)

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
tianyu-l: IIUC right now this implementation does not rely on backward hooks to achieve prefetching in backward, thanks to the pre-determined linear fetching order. The linear order is specified by user. Is it enough for the FAIR person or not?

If instead we allow user to call gather in code, then it'd be hard to do backward prefetching at proper place, unless we use backward hooks, or require user to call reduce-scatter in autograd functions.

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
```diff
@@ -0,0 +1,246 @@
...
+        param_placements=param_placements,
+        bucket_assignments=bucket_assignments,
+    )
+
+
+def flex_shard(
+    module: nn.Module,
+    mesh: DeviceMesh,
+    shard_placement_fn: PlacementFn,
+    buckets: list[BucketSpec],
```
tianyu-l: chatted offline: maybe we should provide bucket-centric API and internal dataclasses

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,693 @@
...
+        )
+
+    def prefetch_next(self) -> None:
+        """Start the next bucket's all-gather if no prefetch is in flight."""
+        if self.context.pending is not None:
+            return
+        prefetch_order = self.context.buckets
+        is_recompute = self.in_reshard_after_forward_recompute()
+        if is_recompute:
+            prefetch_order = prefetch_order[::-1]
```
tianyu-l, line 491: oh, here is the conditional prefetching order for fwd vs. bwd

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,693 @@
...
+            return
+        local_shards = self._local_shards(use_autograd=self.use_autograd_unshard)
+        if self.use_autograd_unshard:
+            prefetched_result = self.take_pending()
+            runtime = BucketAllGatherRuntime(
+                bucket=self,
+                prefetched_result=prefetched_result,
+            )
+            full_params = list(_BucketAllGather.apply(runtime, *local_shards))
+            self.prefetch_next()
```
tianyu-l: chatted offline: 
Let's say this is pre-forward of layer 2. It's not clear if this reduce-scatter (say of layer 2 after layer 2 bwd) would compete with all-gather (correspondingly, layer 0).

IIRC, in FSDP2, backward comms are arranged linearly and all-gather (on critical path) always happens before the competing reduce-scatter (can be delayed).

Let me try breakdown the ordering here:
- in fwd, `_BucketAllGather` attaches reduce-scatter to layer 2 params -> `prefetch_next` initiates all-gather on layer 3 param -> run fwd of layer 2
- in bwd
    - run pre-forward hook of layer 2: `_BucketAllGather` is no-op -> `prefetch_next` initiates all-gather on layer 1 param -> run backward of layer 2 -> after which, the **reduce-scatter of layer 2 grad** runs
    - run pre-forward hook of layer 1: `_BucketAllGather` is no-op -> `prefetch_next` initiates **all-gather on layer 0 param** -> run backward of layer 1

If it's correct, it means reduce-scatter of layer 2 grad would delay all-gather of layer 0 param, which can potentially expose comms that could've been hidden.

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,693 @@
...
+                    self.collected_grads.clear()
+                    for idx, grad in zip(indices, grads, strict=True):
+                        if grad is not None:
+                            self.collected_grads[idx] = grad
+                    self._reduce_collected_grads()
+
+                torch.autograd.graph.register_multi_grad_hook(
+                    leaves,
+                    _on_grads,
+                )
```
tianyu-l: what's this and how is it different from backward of `_BucketAllGather`?

torchtitan/experiments/flex_shard/example/shard.py
weifengpy: good point. I update the PR to define  `compute_local_numel` at base Placement

Separately, will follow up on  whether we should abandon inheritance

torchtitan/experiments/flex_shard/flex_shard/bucket_collectives.py
weifengpy: good call. I marked this as TODO. indeed titan/mtia/nccl copy-engine kernels almost all use SUM

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: the difference is mainly in backward

For flex shard, autograd have full visibility to AG:
local sharded params -> bucket all-gather -> full params -> module compute -> loss

backward naturally goes through the reverse edge:
full-param grads -> _BucketAllGather.backward -> reduce-scatter -> local sharded grads

ezyang: Ah. I think you do need to think about how you are scheduling the reduce scatters though?

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
ezyang: If you let a user do that, they'd be responsible for calling backward hooks at appropriate spots where they want backward prefetching. Not really any way around that lol.

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: good point. i just modified code to do async RS to overlap with compute

full-param grads -> _BucketAllGather.backward -> async reduce-scatter

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
weifengpy: this reminds me of the recent AC proposal

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: it was for cpu unit test with cpu mesh. I changed those test to use cuda mesh. now all buckets are on cuda for sure at runtime. (still cuda after H2D in cpu offloading)

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: we modifed the code to use autograd path regardless of reshard_after_forward=True/False

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: I changed this to assert - realistically it shouldn't happen

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: good question. I modifed code to always use autograd code path, aka `_BucketAllGather`. No more `register_multi_grad_hook`

I was using register_multi_grad_hook for reshard_after_forward=False, because backward start with full params

torchtitan/experiments/flex_shard/flex_shard/reshard_after_forward.py
```diff
@@ -0,0 +1,289 @@
...
+    """Activation recompute policy for per-layer reshard-after-forward.
+
+    Marks collective ops (all-gather, broadcast, wait_tensor) for
+    recomputation; activation checkpointing discards their outputs after each layer's
+    forward. All other ops (matmul, attention, etc.) are saved, avoiding
+    redundant compute recomputation in backward.
+    """
+    from torch.utils.checkpoint import CheckpointPolicy
+
+    if func in _FLEX_SHARD_COLLECTIVE_OPS:
```
wconstab, line 103: we'd expect people to insert their own collectives for some placements, right?  So we'd need this thing to be extensible

torchtitan/experiments/flex_shard/example/flat_shard.py
wconstab: yea, not only the impl, but i want to convince myself that the combination of param_shard_fn + buckets gives us what we need.  (since, this separation at least restricts telling the placement additional info about each param in the bucket, such as how to order them).

speaking of which, reduction order matters- do we provide a way to control the order of params in a bucket even in simple cases?

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: Good catch. This was a real issue!

I updated the PR to achieve ideal AG-RS launching order: all-gather prefetch happens before RS
```
  layers.2 recompute hook
    all-gather/consume layers.2
    prefetch backward next bucket: layers.1

  layers.2 backward
    prepare layers.2 reduce-scatter inputs
    queue reduce-scatter launch, but do not launch NCCL yet

  layers.1 recompute hook
    consume layers.1
    prefetch backward next bucket: layers.0
    then flush one queued reduce-scatter launch: layers.2
```

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
weifengpy: good question. I updated `flex_shard/__init__.py` to be accurate what's public

```
  BucketSpec
  flex_shard
  LocalStorageLayout
  MixedPrecisionPolicy
  OffloadPolicy
  Placement
```

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
```diff
@@ -0,0 +1,686 @@
...
+            strict=True,
+        ):
+            if grad is None:
+                continue
+            grads.append(grad.contiguous())
+            valid_infos.append(entry.param_info)
+            valid_param_refs.append(entry.module_info)
+
+        if grads:
+            if bucket.storage._reshard_after_forward:
```
tianyu-l: I might be reading it wrong:
Why queuing or not depend on this bucket's reshard_after_forward decision? My intuition is that it should depend on next all-gather (typically previous layer)'s reshard_after_forward decision.

torchtitan/experiments/flex_shard/flex_shard/reshard_after_forward.py
weifengpy: critical finding! totally agree with the gap. I was thinking of flex_shard custom op namespace, eg `torch.ops.flex_shard.all_gather_into_tensor`

once having all placements + 2D DP+TP, I should be able to figure out about solution for custom op registration

torchtitan/experiments/flex_shard/example/flat_shard.py
weifengpy: total agree on the landing placements sooner. cc @anshul-si who are working on placements 

I am polishing full graph capture to guard the placement development, so flex shard + torch.compile works on day 1:  https://github.com/pytorch/torchtitan/pull/3317

> do we provide a way to control the order of params in a bucket even in simple cases?

the numeric is deterministic today because named_parameters() traversal happens in flex shard core, not placements under example/ folder

torchtitan/experiments/flex_shard/flex_shard/reshard_after_forward.py
weifengpy: true - I can address this as we add more placements. this is also needed in another discussion about custom op like `torch.ops.flex_shard.all_gather_into_tensor` https://github.com/pytorch/torchtitan/pull/3239/changes/BASE..d5ca612942cd920c53a1277756eb2aa933880dd0#r3239690057

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
weifengpy: I investigated this - this PR assume linear execution order from bucket0 -> bucket 1 -> ... -> bucket N

we can potentialy swap the AC - flex_shard order after supporting arbitrary wrapping: https://github.com/pytorch/torchtitan/pull/3348

In this PR, AC happens first, hook installation happens after, it's validate checkpoint_wrapper to raise error for unsupported bucketing like this
```
  [
      BucketSpec(["layers.0.*"]),
      BucketSpec(["layers.1.*"]),
      BucketSpec(["tok_embeddings.*", "pos_embeddings.*", "norm.*", "output.*"]),
  ]
```

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
weifengpy: > I have a checkpoint that is already the correct local shard, do I have to gather the entire parameter before copying it to storage?

for loading checkpoint, we would need meta init -> flex_shard (skip copy_param_to_storage becuase of meta device) -> load checkpoint into model sharded params

weifengpy: > It's not obvious if I call this multiple times to write multiple parameters into a single byte storage? 

for one bucket containing p0, p1, p2, yes, we call it three times to write those three parameter shards into one byte storage. it is init-time only, not per iteration

> I actually end up running this on literally every parameter, even if most of them don't copy into this storage, is that right?

the callchain is from `bucket_named_params = [(fqn, named_params_dict[fqn]) for fqn in bucket_fqns]`. 
each call actually copies one parameter into the bucket. no call for parameters outside the bucket

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
weifengpy: no obstacles. this docstring is staled. I will do some cleaning to remove anything related to nested wrapping

torchtitan/experiments/flex_shard/flex_shard/bucket_runtime.py
weifengpy: You are right. I updated the PR with `should_defer_reduce_scatter_for_backward_prefetch`. queueing RS  depends on next all-gather

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
weifengpy: good catch - I moved `__eq__` and `__hash__` into base

weifengpy: I actually intentionally choose to be the same placement naming to DTensor, because they are placement without torch dispatch. If we ended with Protocol, we might have a different name, say PlacementProtocol

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
weifengpy: good point. compute_local_numel is a thin wrapper of compute_local_shape now

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
weifengpy: It does not have to be contiguous

say NCCL does not support contiguous input, we only need `.contiguous()` at copy-in time
say NCCL support contiguous input, we would need the customized copy kernel to support non-contiguous layout

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
weifengpy: agreed - `assemble_from_shards` is a Shard specific helper function. I removed it from placement contract

torchtitan/experiments/flex_shard/flex_shard/placement_contract.py
weifengpy: subclass can override this to support padding. for example, we have `storage_nbytes=layout.storage_nbytes + self.padding_nbytes` in `_PaddedShard`  test_flex_shard_buckets.py

torchtitan/experiments/flex_shard/flex_shard/bucket_storage.py
weifengpy: updated with type alias

torchtitan/experiments/flex_shard/flex_shard/unsharded_param_access.py
weifengpy: Good question. Bucket 0 is not special; this was a stale assumption from nested wrapping. I removed _DSTORAGE_ATTR entirely

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
weifengpy: one example is `BucketSpec(["embed.*", "layers.1.*", "output.*"], reshard_after_forward=True)`

The deepest common ancestor for that bucket is the root model, so FlexShard would install the bucket unshard hook on model

the situition might be mitigated after enabling arbitrary wrapping: https://github.com/pytorch/torchtitan/pull/3348 . But I want to be more cautious of the code complexity

torchtitan/experiments/flex_shard/flex_shard/flex_shard.py
weifengpy: done - I passed input object directly now. indeed cleaner

torchtitan/experiments/flex_shard/example/owned.py
```diff
@@ -0,0 +1,260 @@
...
+        for full_param in prepared.buffers:
+            with _record_comm_if_eager(
+                "FlexShard::broadcast",
+                prepared.placement_state.debug_fqn,
+            ):
+                dist.broadcast(
+                    full_param,
+                    src=self.owner_rank,
+                    group=prepared.placement_state.pg,
+                )
```
tianyu-l: Does this work with bucketing? If so, how?

torchtitan/experiments/flex_shard/example/shard.py
tianyu-l, line 1: Do we expect non-advanced users to write their own custom placements, or do we expect flex_shard to unblock power users (including us)?

torchtitan/experiments/flex_shard/example/owned.py
weifengpy: good catch. GroupedOwned from [dsv3](https://github.com/pytorch/torchtitan/pull/3603/changes#diff-df6e6c963aaad665070a2c4b0d0d3d87367a991c75133979ebbdbe64589b9e6eR1471-R1475)  ensures single collective per bucket, packing owned parameters into a flat buffer

this plain Owned placement won't be public facing after consolidating dsv3 changes into this core PR

tianyu-l: If we believe this won't be useful in practice, maybe we can even delete?

PR #3241, opened 2026-05-06
[graph_trainer] Gate overlap_fsdp_ag_rs_pass behind a config flag

yiming0416:
- Adds `compile.enable_fsdp_ag_rs_overlap` (default False) to `GraphTrainerConfig`.
- Removes `overlap_fsdp_ag_rs_pass` from the unconditional default pass list. Instead, `joint_transformer_block_bucketing_reordering_pass` now optionally runs the overlap pass first when the new flag is set.

### DIFF
--- a/torchtitan/experiments/graph_trainer/fsdp_passes.py
+++ b/torchtitan/experiments/graph_trainer/fsdp_passes.py
@@ -361,6 +361,7 @@ def joint_transformer_block_bucketing_reordering_pass(
     module_bucket_plans: list[list[str] | str],
     insert_overlap_deps: bool = False,
     bucket_mode: BucketMode | None = None,
+    enable_fsdp_ag_rs_overlap: bool = False,
 ) -> torch.fx.GraphModule:
     """Run joint-graph manual bucketing and reordering.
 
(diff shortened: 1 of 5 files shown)


torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -243,7 +244,16 @@ def joint_transformer_block_bucketing_reordering_pass(
             ``preserve_node_ordering`` after the topological sort.
         bucket_mode: bucket mode forwarded to the underlying bucketer;
             defaults to ``"custom_ops"`` via the parent class.
+        enable_fsdp_ag_rs_overlap: when ``True``, run ``overlap_fsdp_ag_rs_pass``
```
aditvenk, line 383: Why a new flag instead of running it unconditionally?

SherlockNoMad: feedback from WeiFeng that overlapping ag/rs does't necessarily give perf win, given ag and rs are sharing the network resource. (same ranks participating the communication) 
which make sense to me, so we need to benchmark in scaling run to validate this.

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/fsdp_passes.py
yiming0416: More context here: https://github.com/pytorch/torchtitan/pull/3156?fbclid=IwY2xjawRop4dleHRuA2FlbQIxMQBicmlkETFyZkFWck9mS1RpS21pc1JJc3J0YwZhcHBfaWQBMAABHg21cckVfcoK-j0JGxgYq6vXQggs1qPDTy5eFhEVJovkaswAUcFtgVttPggo_aem_74LE5LyrXBf7rC1JdvBypQ#issuecomment-4383398078

aditvenk (approved):

PR #3242, opened 2026-05-06
[rl] Register customized config parser to vllm + less vllm config dependency

wwwjn:
* #3236
* #3142
* __->__ #3242

vllm has this customized config parser registry support so we can plug in TorchTitan's config parser. Why we need this:
- get rid of dependency on a HF format checkpoint folder when initializing. Don't implicitly depend on `config.json` as config source of truth

Another changes in this PR:
- remove the round-trip translation from torchtitan config -> vllm config -> torchtitan config. Using closure to bypass.


torchtitan/experiments/rl/actors/generator.py
```diff
@@ -206,7 +203,14 @@ def __init__(
             distributed_executor_backend="external_launcher",
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=not config.cudagraph.enable,
-            hf_overrides={"architectures": [VLLM_MODEL_NAME]},
+            # Tell wrapper to skip the HF weight load when running with random
+            # init: the trainer pushes its random-init weights to TorchStore
+            # right after engine init, so reading the on-disk checkpoint would
+            # be wasted work AND clobber the trainer's weights.
+            hf_overrides={
+                "architectures": [VLLM_MODEL_NAME],
```
wwwjn: This line can be removed.

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -38,29 +59,61 @@ def register_model_to_vllm_model_registry(
...
     TorchTitanVLLMModelFromSpec.__name__ = VLLM_MODEL_NAME
     TorchTitanVLLMModelFromSpec.__qualname__ = VLLM_MODEL_NAME
-
-    # Register with vLLM
     ModelRegistry.register_model(VLLM_MODEL_NAME, TorchTitanVLLMModelFromSpec)
 
+    # Dynamic config parser class capturing ModelSpec (and any registration-
+    # time custom fields) in the closure.
+    @register_config_parser(TORCHTITAN_CONFIG_FORMAT)
+    class TorchTitanConfigParserForSpec(ConfigParserBase):
```
wwwjn: Nested in the registry() function because we need to access model_spec when registering config_parser via closure

torchtitan/hf_datasets/text_datasets.py
wwwjn, line 1: Not related to this PR. should be cleaned.

torchtitan/experiments/rl/models/vllm_config_parser.py
```diff
@@ -0,0 +1,115 @@
...
+        "hidden_size": cfg.dim,
+        "num_hidden_layers": len(cfg.layers),
+        "num_attention_heads": n_heads,
+        "num_key_value_heads": n_kv_heads,
+        "head_dim": head_dim,
+        "max_position_embeddings": cfg.rope.max_seq_len,
+        "rope_theta": cfg.rope.theta,
+        "rms_norm_eps": cfg.norm.eps,
+        "tie_word_embeddings": getattr(cfg, "enable_weight_tying", False),
+        # torch_dtype is intentionally omitted: EngineArgs(dtype=...) always
```
wwwjn: There's a field named "torch_dtype" in HF format which is required by vllm. I ignored it because EngineArgs(dtype=...) always override it

wwwjn: Will remove this comment to clean up.

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -38,29 +59,61 @@ def register_model_to_vllm_model_registry(
             during model construction.
     """
     from torchtitan.experiments.rl.models.vllm_wrapper import TorchTitanVLLMModelWrapper
+    from transformers import PretrainedConfig
```
wwwjn: We / RL side have to depend on PretrainedConfig definition from `transformers` as it's required to as return-ed type of ConfigParser. I think ConfigParser is a clean abstraction, but we would need to depend on `transformers` unfortunately

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -38,29 +59,61 @@ def register_model_to_vllm_model_registry(
...
+    # ``PretrainedConfig``. For per-engine runtime flags, callers should use
+    # ``EngineArgs(hf_overrides={...})`` instead.
+    #
+    # CompileConfig is converted to a dict here (rather than stored as a
+    # dataclass instance) so the resulting ``PretrainedConfig`` stays
+    # JSON-serializable — vLLM's telemetry / caching paths sometimes call
+    # ``hf_config.to_json()`` and a dataclass attribute would crash that.
+    # Wrapper reconstructs via ``CompileConfig(**hf_config.compile_config)``.
+    custom_hf_config_fields: dict[str, Any] = {
+        "compile_config": dataclasses.asdict(compile_config),
```
wwwjn: I moved compile_config from  function signature to the customized config field. Convert to dict for json-serializability.  This method requires the Config class needs to be json-serializiable, but I think it might be tricky sometimes if the data class is not serializable. Want to post this and discuss

Most of fields from config system can be passed and reconstructed in the same way, except ShardingConfig, model_spec

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -272,8 +277,13 @@ def __init__(
                 self.model.freqs_cis, max_model_len
             )
 
-        # Initial load model weights from HuggingFace checkpoint path
-        self._initial_load_weights(checkpoint_path=vllm_config.model_config.model)
+        # Initial load model weights from HuggingFace checkpoint path.
+        if not getattr(
+            vllm_config.model_config.hf_config, "skip_init_weights_load", False
+        ):
+            self._initial_load_weights(checkpoint_path=vllm_config.model_config.model)
```
tianyu-l: making this comment again https://github.com/pytorch/torchtitan/pull/3142#discussion_r3196747704, why generator would need to load this weight

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -206,7 +203,14 @@ def __init__(
...
+                "skip_init_weights_load": config.debug.random_init,
```
tianyu-l: why vllm needs this

torchtitan/experiments/rl/models/vllm_registry.py
tianyu-l: what's the benefit of this change?

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -38,29 +59,61 @@ def register_model_to_vllm_model_registry(
...
     class TorchTitanVLLMModelFromSpec(TorchTitanVLLMModelWrapper):
         def __init__(self, *, vllm_config, prefix=""):
             super().__init__(
                 model_spec=model_spec,
                 vllm_config=vllm_config,
                 prefix=prefix,
-                compile_config=compile_config,
             )
 
-    # Set the class name so vLLM can identify it
```
tianyu-l, line 56: why removing these comments

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -199,14 +195,26 @@ def __init__(
...
             # tells vLLM to run one worker per process (no subprocess spawning)
             distributed_executor_backend="external_launcher",
             gpu_memory_utilization=config.gpu_memory_limit,
             enforce_eager=not config.cudagraph.enable,
-            hf_overrides={"architectures": [VLLM_MODEL_NAME]},
+            # Per-engine torchtitan config sections forwarded to the wrapper
+            # via the custom ConfigParser. Keys are validated against
+            # ``_ALLOWED_TORCHTITAN_CONFIG_OVERRIDES`` at engine init: every
+            # value here must originate from a torchtitan config dataclass
+            hf_overrides={
```
wwwjn: I moved compile_config from function signature to the customized config field. Convert to dict for json-serializability. This method requires the Config class needs to be json-serializiable, but I think it might be tricky sometimes if the data class is not serializable. Want to post this and discuss. 

Most of fields from config system can be passed and reconstructed in the same way, except ShardingConfig, model_spec. For integrity, each field being passed via `hf_overrides` are checked in `vllm_config_parser.py`

torchtitan/experiments/rl/actors/generator.py
wwwjn: This was used to decided whether we should skip initial weight loading in generator. But I updated this PR to make this part more clear

Also reply to https://github.com/pytorch/torchtitan/pull/3142#discussion_r3196747704, When using TorchtianVLLMWrapper to run purely inference (and eval), we still need to `load_init_weight`.

torchtitan/experiments/rl/models/vllm_registry.py
wwwjn: 1. Easily passing more config from torchtitan's config system to torchtitan vllm wrapper. This is needed because vllm wrapper is not only doing model initialization, but also load weights/ parallelism /etc
2. We don't need to have a fake config.json (file based config) in CI testing. All config are directly derived from model_spec

I think it's a more standard way to passing torchtitan's configs fields into vllm wrapper, instead of passing via closure

torchtitan/distributed/utils.py
```diff
@@ -417,7 +417,7 @@ def _get_distributed_backend(enable_cpu_backend):
     device_id: torch.device | None = None
     if comm_config.mode == "torchcomms":
         try:
-            import torchcomms  # noqa: F401  # pyrefly: ignore [missing-import]
```
wwwjn, line 420: Will be reverted, not related to this PR

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -20,12 +20,14 @@
 )
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
-from torchtitan.config import CommConfig, Configurable, TORCH_DTYPE_MAP
-from torchtitan.config.configs import (
+from torchtitan.config import (
```
wwwjn, line 23: This change is just consolidate the import path

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -160,6 +164,16 @@ def __post_init__(self):
                     f"Generator does not support expert parallelism, "
                     f"got ep={p.expert_parallel_degree}"
                 )
+            if p.enable_sequence_parallel:
+                logger.warning(
+                    "Generator enable_sequence_parallel=True hurts inference "
+                    "throughput; prefer SP=False."
+                )
```
tianyu-l, line 173: this won't be supported by spmd_types erasure mode I think, so I don't mind we ban it
cc @pianpwk

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -199,14 +214,17 @@ def __init__(
         engine_kwargs = dict(
             model=model_path,
```
tianyu-l, line 222: what is this path for?

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -169,14 +118,28 @@ def __init__(
...
+        compile_config: CompileConfig,
         vllm_config: VllmConfig,
         prefix: str = "",
-        compile_config: CompileConfig,
     ):
         super().__init__()
 
         assert vllm_config is not None, "vllm_config is required"
 
+        # PP and CP are not supported on this inference path
```
tianyu-l: this "raise ValueError" may better happen at grpo trainer post_init, to be consistent

here we only need `assert`

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
+        model_spec: TorchTitan ModelSpec containing model config and components.
+        parallelism: Authoritative parallelism configuration. The wrapper
+            uses this directly to build ``ParallelDims``; the caller is
+            responsible for translating the relevant fields (TP, EP) to
+            ``EngineArgs`` so vLLM's own world layout matches.
+        compile_config: torch.compile config applied per-layer by the
+            wrapper's parallelize step.
     """
     from torchtitan.experiments.rl.models.vllm_wrapper import TorchTitanVLLMModelWrapper
+    from transformers import PretrainedConfig
```
tianyu-l: oh, can we not depend on this? I know it's already implicit via vllm dependency, but trying to see if we can avoid explicit dependency (someday we can remove)

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
     TorchTitanVLLMModelFromSpec.__qualname__ = VLLM_MODEL_NAME
-
-    # Register with vLLM
     ModelRegistry.register_model(VLLM_MODEL_NAME, TorchTitanVLLMModelFromSpec)
 
+    # Dynamic config parser class capturing ModelSpec in the closure. This
+    # parser only produces HF-shaped fields; torchtitan-specific config is
+    # delivered through the model-class closure above.
+    @register_config_parser(TORCHTITAN_CONFIG_FORMAT)
+    class TorchTitanConfigParserForSpec(ConfigParserBase):
```
tianyu-l:
```suggestion
    class TorchTitanConfigParser(ConfigParserBase):
```

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
     from vllm.logger import init_logger
     from vllm.model_executor.models.registry import ModelRegistry
+    from vllm.transformers_utils.config import register_config_parser
+    from vllm.transformers_utils.config_parser_base import ConfigParserBase
 
     logger = init_logger(__name__)
 
-    # Create dynamic model class capturing ModelSpec in the closure
+    # Dynamic model class capturing torchtitan config in the closure.
     class TorchTitanVLLMModelFromSpec(TorchTitanVLLMModelWrapper):
```
tianyu-l: maybe simplify, from a titan-centric view
```suggestion
    class VLLMModelFromSpec(VLLMModelWrapper):
```

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
+        def parse(
+            self,
+            model,
+            trust_remote_code,
+            revision=None,
+            code_revision=None,
+            **kwargs,
+        ):
+            config_dict = model_spec_to_hf_config_dict(model_spec)
+            return config_dict, PretrainedConfig.from_dict(config_dict)
```
tianyu-l, line 204: It's actually very weird that the contract is both a config_dict and a cls(config_dict), sounds redundant to me

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
+    ffn = getattr(layer0, "feed_forward", None)
+    if ffn is not None:
+        # Unused: only v1/metrics/perf.py reads it (off by default). SwiGLU hidden == w1.out_features.
+        hf["intermediate_size"] = ffn.w1.out_features
+
+    moe = getattr(layer0, "moe", None)
+    if moe is not None:
+        hf[
+            "num_experts"
+        ] = moe.experts.num_experts  # presence required: >0 toggles MoE/EP branches
```
tianyu-l: nit: put `# presence required` above this field

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
+        "rms_norm_eps": cfg.norm.eps,  # only minimax-qk-norm fusion reads it; wrapper builds RMSNorm
+        "tie_word_embeddings": getattr(
+            cfg, "enable_weight_tying", False
+        ),  # multimodal/GGUF only; wrapper ties weights
+        "bos_token_id": 0,  # Fuyu-only; engine reads tokenizer/sampling tokens
+        "eos_token_id": 1,  # per-model files only; engine reads tokenizer/sampling tokens
+    }
 
-def register_model_to_vllm_model_registry(
+    ffn = getattr(layer0, "feed_forward", None)
```
tianyu-l: using layer0 is not robust? what if a transformer has 1st layer MoE, 2nd layer FFN?

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,189 @@
...
+    if ffn is not None:
+        # Unused: only v1/metrics/perf.py reads it (off by default). SwiGLU hidden == w1.out_features.
+        hf["intermediate_size"] = ffn.w1.out_features
+
+    moe = getattr(layer0, "moe", None)
```
tianyu-l: vice versa

torchtitan/experiments/rl/actors/generator.py
pianpwk: hmm are we not supporting SP in spmd_types

tianyu-l: we support SP in spmd_types when sequence length is evenly divisible; we don't support SP when sequence length is not evenly divisible.
- spmd_types doesn't handle padding & unpadding like DTensor.
- uneven usually show up only in inference
- for inference we always use non-SP (vanilla TP)

wwwjn: Sure,  we can ban SP directly.

torchtitan/experiments/rl/actors/generator.py
wwwjn: Now it serves 2 purpose:
1. Loading tokenizer. This can be removed by passing  `tokenizer=tokenizer_path` to EngineArgs. 
2.Initial_weight_loading: Will sort out the weight loading part for both trainer and generator in next PR. 

After lifting both, we can pass some fake path, say "torchtitan", to vllm

torchtitan/experiments/rl/models/vllm_registry.py
wwwjn: We can do it via `vllm.transformers_utils.config import PretrainedConfig`, as we already depend on vllm

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -5,62 +5,203 @@
...
+    ffn = None
+    for layer in cfg.layers:
+        ffn = getattr(layer, "feed_forward", None)
+        if ffn is not None:
+            break
+    moe = None
+    for layer in cfg.layers:
+        moe = getattr(layer, "moe", None)
+        if moe is not None:
+            break
```
tianyu-l: let's use `next()` to simplify this for-loop
https://stackoverflow.com/questions/9014058/creating-a-python-list-comprehension-with-an-if-and-break

torchtitan/experiments/rl/grpo.py
```diff
@@ -234,6 +237,41 @@ def __post_init__(self):
                         "and has not been validated for determinism."
                     )
 
+            # VLLMGenerator only supports TP. vLLM handles its own parallelism;
```
tianyu-l: sorry my earlier comment introduced some noise
https://github.com/pytorch/torchtitan/pull/3242#discussion_r3211802882

I meant to say that we should put some check in post_init, but not ParallelismConfig's post_init because it'll be shared by core trainer, policy trainer, and generator. I think the proper place to put these checks is generator config's post_init.

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: nit
```suggestion
class VLLMModelWrapper(Module):
```
given it's pretty unambiguous that we are wrapping torchtitan model with vllm wrapper

tianyu-l (approved):

torchtitan/experiments/rl/grpo.py
wwwjn: Sorry my bad , moved to Generator as it's required by RL generator auto-regressive nature only

PR #3243, opened 2026-05-06
Enable SFT for HF Transformer backed trainer

HosseinKaviani-H:
## Summary

  Enables supervised fine-tuning (SFT) for any HuggingFace model in the
  transformers_modeling_backend experiment, building on the native SFT infrastructure from #2556.

  - Uses FlexAttention with BlockMask for packed sequences. Builds the BlockMask in
    `post_dataloading_process` (before parallelism sharding) using TorchTitan's existing
    `get_causal_mask_mod` + `get_document_mask_mod` — same mask infrastructure as native TorchTitan SFT.
  - `HFFlexAttention` subclasses TorchTitan's `FlexAttention` kernel module, handling HF's tensor
    layout via transposes. Registered as a submodule on each HF attention layer so
    `apply_cp_to_forward` works naturally — no globals needed.
  - Registers `flex_torchtitan` as a custom attention implementation that bypasses HF's internal
    mask creation (which builds at the wrong point for TP+SP) and routes through a compile-friendly
    `flex_attention` call.
  - `SFTTrainer` subclass overrides `post_dataloading_process` to build the BlockMask at full
    sequence length before any parallelism touches the data.
  - `HFTransformerStateDictAdapter` enables loading pretrained HF weights (trivial `model.` prefix
    mapping, handles weight tying).
  - `HFBackendTokenizer` passes `bos_token`/`eos_token` to chat templates and fixes an edge case
    where BOS and EOS share the same token string.
  - TitanDenseModelConfig defaults changed to None so HF config values aren't overridden when using
    full-size models.
  - No dynamo `cache_size_limit` workaround needed: relies on the upstream nested-`AsyncCollectiveTensor`
    fix (pytorch#186442), so the flex path stays under the default recompile limit across all parallelisms.
  - All changes within `torchtitan/experiments/transformers_modeling_backend/` — no core modifications.

  ## Files changed

  | File | Description |
  |---|---|
  | trainer.py (new) | `SFTTrainer` builds BlockMask before parallelism sharding, handles CP load balancer |
  | state_dict_adapter.py (new) | `model.` prefix strip/add for HF weight loading, handles weight tying |
  | tokenizer.py (new) | Passes special tokens to chat templates, fixes shared BOS/EOS edge case |
  | model.py | `HFFlexAttention` subclass + `_flex_torchtitan_attention_forward`; `forward()` accepts positions/attention_masks; guard 
  `intermediate_size`/`head_dim` recalculation; disable `attention_dropout` for FlexAttention |
  | configs.py | `build()` override to use `SFTTrainer` |
  | parallelize.py | Collects `HFFlexAttention` modules and calls `apply_cp_to_forward` for CP |
  | __init__.py | `sft_debugmodel`/`sft_full` flavors with `attn_implementation="flex_torchtitan"`; wire `state_dict_adapter`; None defaults for
  TitanDenseModelConfig |
  | config_registry.py | SFT config functions (random init + pretrained weight loading) using `ChatDataLoader` and `HFBackendTokenizer` |
  | tests/integration_tests.py | 2-GPU SFT integration test |

  ## Parallelism support

  | Parallelism | Status |
  |---|---|
  | FSDP | Works |
  | FSDP + TP | Works |
  | FSDP + TP + PP | Works |
  | FSDP + AC (selective/full) | Works |
  | FSDP + compile | Works |
  | FSDP + TP + compile | Works |
  | FSDP + CP | Works |
  | CP + compile | Works |
  | TP + CP | Works |
  | FSDP + TP + CP | Works |
  | FSDP + TP + CP + compile | Works |

  ## Test plan

  - [x] Pretraining regression: `transformers_modeling_backend_debugmodel` identical loss before/after
  - [x] SFT debugmodel (Qwen3 arch): loss 11.0 → 3.3, 2 GPUs
  - [x] SFT debugmodel (Mistral arch): loss 10.9 → 8.3 — architecture-agnostic
  - [x] SFT with pretrained Qwen3-0.6B: loss 13.4 → 6.1
  - [x] SFT with pretrained Qwen2.5-0.5B-Instruct: loss 5.3 → 1.7
  - [x] SFT with pretrained Llama-3.2-1B-Instruct: loss 7.6 → 0.0001
  - [x] SFT with pretrained Seed-Coder-8B-Instruct: loss 6.8 → 2.1
  - [x] SFT with pretrained Granite-3.1-2B-Instruct: loss 50.2 → 12.2
  - [x] FSDP + TP (4 GPU): works
  - [x] FSDP + TP + PP (8 GPU): works
  - [x] FSDP + compile (2 GPU): works
  - [x] FSDP + TP + compile (4 GPU): works
  - [x] FSDP + CP (2 GPU): works
  - [x] CP + compile (2 GPU): works
  - [x] TP + CP (4 GPU): works
  - [x] FSDP + TP + CP (8 GPU): works
  - [x] FSDP + TP + CP + compile (8 GPU, full 28-layer qwen3_0.6b): works — 0 cache-limit hits at default recompile limit
  - [x] Reproducibility: compare loss curves against native SFT on same model/data
  - [x] SFT loss/KL parity vs native (#2556), qwen3_0.6b, same weights + packed batch: max |loss diff| 3.4e-06, max KL 5.0e-09

  Numeric verification results:

  ### Native TorchTitan vs HuggingFace — logit parity (qwen3_0.6b, same weights + input, fp32, 161 positions)

  | metric | value |
  |---|---|
  | top1 match | 100.0000% |
  | top5 overlap | 100.0000% |
  | KL(HF ‖ native) | 1.60e-06 |
  | KL(native ‖ HF) | 6.78e-07 |
  | max \|logit diff\| | 9.06e-05 |
  | mean \|logit diff\| | 5.14e-06 |
  | cosine similarity | 1.000000 |

### DIFF
--- a/torchtitan/experiments/transformers_modeling_backend/config_registry.py
+++ b/torchtitan/experiments/transformers_modeling_backend/config_registry.py
@@ -99,7 +99,7 @@ def transformers_modeling_backend_sft_full() -> TransformersBackendConfig:
         hf_model="Qwen/Qwen3-0.6B",
         model_spec=model_registry("sft_full"),
         tokenizer=HFBackendTokenizer.Config(),
-        optimizer=OptimizersContainer.Config(lr=2e-5),
+        optimizer=default_adamw(lr=2e-5),
         lr_scheduler=LRSchedulersContainer.Config(
             warmup_steps=2,
             decay_ratio=0.8,
@@ -127,9 +127,7 @@ def transformers_modeling_backend_sft_full() -> TransformersBackendConfig:
             interval=10,
             last_save_model_only=False,
         ),
-        activation_checkpoint=ActivationCheckpointConfig(
-            mode="selective",
-        ),
+        activation_checkpoint=SelectiveAC.Config(),
     )
 
 
@@ -148,7 +146,7 @@ def transformers_modeling_backend_sft_debugmodel() -> TransformersBackendConfig:
         hf_model="Qwen/Qwen3-4B-Instruct-2507",
         model_spec=model_registry("sft_debugmodel"),
         tokenizer=HFBackendTokenizer.Config(),
-        optimizer=OptimizersContainer.Config(lr=8e-4),
+        optimizer=default_adamw(lr=8e-4),
         lr_scheduler=LRSchedulersContainer.Config(
             warmup_steps=2,
             decay_ratio=0.8,
@@ -173,7 +171,5 @@ def transformers_modeling_backend_sft_debugmodel() -> TransformersBackendConfig:
             interval=10,
             last_save_model_only=False,
         ),
-        activation_checkpoint=ActivationCheckpointConfig(
-            mode="selective",
-        ),
+        activation_checkpoint=SelectiveAC.Config(),
     )
--- a/torchtitan/experiments/transformers_modeling_backend/config_registry.py
+++ b/torchtitan/experiments/transformers_modeling_backend/config_registry.py
@@ -14,7 +14,10 @@ from torchtitan.distributed.activation_checkpoint import SelectiveAC
 from torchtitan.experiments.transformers_modeling_backend.configs import (
     TransformersBackendConfig,
 )
-from torchtitan.hf_datasets.text_datasets import ChatDataLoader, HuggingFaceTextDataLoader
+from torchtitan.hf_datasets.text_datasets import (
+    ChatDataLoader,
+    HuggingFaceTextDataLoader,
+)
 from torchtitan.tools.profiler import Profiler
 
 from . import model_registry
--- a/torchtitan/experiments/transformers_modeling_backend/config_registry.py
+++ b/torchtitan/experiments/transformers_modeling_backend/config_registry.py
@@ -14,10 +14,11 @@ from torchtitan.distributed.activation_checkpoint import SelectiveAC
 from torchtitan.experiments.transformers_modeling_backend.configs import (
     TransformersBackendConfig,
 )
-from torchtitan.hf_datasets.text_datasets import HuggingFaceTextDataLoader
+from torchtitan.hf_datasets.text_datasets import ChatDataLoader, HuggingFaceTextDataLoader
 from torchtitan.tools.profiler import Profiler
 
 from . import model_registry
+from .tokenizer import HFBackendTokenizer
 
 
 def transformers_modeling_backend_debugmodel() -> TransformersBackendConfig:
@@ -80,4 +81,99 @@ def transformers_modeling_backend_full() -> TransformersBackendConfig:
             last_save_model_only=False,
         ),
         activation_checkpoint=SelectiveAC.Config(),
+    )
+
+
+def transformers_modeling_backend_sft_full() -> TransformersBackendConfig:
+    """SFT config with real HF pretrained weights loaded via initial_load_in_hf."""
+
+    def process_sample(sample):
+        return [
+            {"role": "user", "content": sample["question"]},
+            {"role": "assistant", "content": sample["answer"]},
+        ]
+
+    return TransformersBackendConfig(
+        loss=CrossEntropyLoss.Config(),
+        hf_assets_path="./tests/assets/qwen3_0.6b",
+        hf_model="Qwen/Qwen3-0.6B",
+        model_spec=model_registry("sft_full"),
+        tokenizer=HFBackendTokenizer.Config(),
+        optimizer=OptimizersContainer.Config(lr=2e-5),
+        lr_scheduler=LRSchedulersContainer.Config(
+            warmup_steps=2,
+            decay_ratio=0.8,
+            decay_type="linear",
+            min_lr_factor=0.0,
+        ),
+        training=TrainingConfig(
+            local_batch_size=2,
+            seq_len=2048,
+            steps=10,
+        ),
+        dataloader=ChatDataLoader.Config(
+            dataset_path="json",
+            load_dataset_kwargs={
+                "data_files": "tests/assets/sft_test/data.json",
+                "split": "train",
+            },
+            sample_processor=process_sample,
+        ),
+        metrics=MetricsProcessor.Config(log_freq=1),
+        checkpoint=CheckpointManager.Config(
+            enable=True,
+            initial_load_in_hf=True,
+            initial_load_model_only=True,
+            interval=10,
+            last_save_model_only=False,
+        ),
+        activation_checkpoint=ActivationCheckpointConfig(
+            mode="selective",
+        ),
+    )
+
+
+def transformers_modeling_backend_sft_debugmodel() -> TransformersBackendConfig:
+    """SFT debug config for the transformers backend using ChatDataLoader."""
+
+    def process_sample(sample):
+        return [
+            {"role": "user", "content": sample["question"]},
+            {"role": "assistant", "content": sample["answer"]},
+        ]
+
+    return TransformersBackendConfig(
+        loss=CrossEntropyLoss.Config(),
+        hf_assets_path="./tests/assets/tokenizer",
+        hf_model="Qwen/Qwen3-4B-Instruct-2507",
+        model_spec=model_registry("sft_debugmodel"),
+        tokenizer=HFBackendTokenizer.Config(),
+        optimizer=OptimizersContainer.Config(lr=8e-4),
+        lr_scheduler=LRSchedulersContainer.Config(
+            warmup_steps=2,
+            decay_ratio=0.8,
+            decay_type="linear",
+            min_lr_factor=0.0,
+        ),
+        training=TrainingConfig(
+            local_batch_size=8,
+            seq_len=2048,
+            steps=10,
+        ),
+        dataloader=ChatDataLoader.Config(
+            dataset_path="json",
+            load_dataset_kwargs={
+                "data_files": "tests/assets/sft_test/data.json",
+                "split": "train",
+            },
+            sample_processor=process_sample,
+        ),
+        metrics=MetricsProcessor.Config(log_freq=1),
+        checkpoint=CheckpointManager.Config(
+            interval=10,
+            last_save_model_only=False,
+        ),
+        activation_checkpoint=ActivationCheckpointConfig(
+            mode="selective",
+        ),
     )
--- a/torchtitan/experiments/transformers_modeling_backend/config_registry.py
+++ b/torchtitan/experiments/transformers_modeling_backend/config_registry.py
@@ -157,8 +157,12 @@ def transformers_modeling_backend_sft_debugmodel() -> TransformersBackendConfig:
             min_lr_factor=0.0,
         ),
         training=TrainingConfig(
-            local_batch_size=8,
-            seq_len=2048,
+            # Keep this small: this debug model uses the full Qwen3 vocab
+            # (~152k), so cross-entropy materializes a
+            # local_batch_size * seq_len * vocab logits tensor. batch=8,
+            # seq=2048 is ~9GB in fp32 and OOMs the 22GB CI GPUs.
+            local_batch_size=1,
+            seq_len=1024,
             steps=10,
         ),
         dataloader=ChatDataLoader.Config(
--- a/torchtitan/experiments/transformers_modeling_backend/configs.py
+++ b/torchtitan/experiments/transformers_modeling_backend/configs.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from dataclasses import dataclass
+from dataclasses import dataclass, fields, replace
 
 from torchtitan.trainer import Trainer
 
@@ -13,3 +13,20 @@ from torchtitan.trainer import Trainer
 class TransformersBackendConfig(Trainer.Config):
     hf_model: str = ""
     """HuggingFace model ID (e.g., 'Qwen/Qwen2.5-7B')"""
+
+    def build(self, **kwargs):
+        from torchtitan.experiments.transformers_modeling_backend.trainer import (
+            SFTTrainer,
+        )
+
+        if not kwargs:
+            return SFTTrainer(config=replace(self))
+
+        config_fields = {f.name for f in fields(self)}
+        overlap = config_fields & kwargs.keys()
+        if overlap:
+            raise ValueError(
+                f"build() kwargs {overlap} overlap with config fields. "
+                "Put these values in the Config, not in build() kwargs."
+            )
+        return SFTTrainer(config=replace(self), **kwargs)
--- a/torchtitan/experiments/transformers_modeling_backend/configs.py
+++ b/torchtitan/experiments/transformers_modeling_backend/configs.py
@@ -4,29 +4,10 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from dataclasses import dataclass, fields, replace
+from torchtitan.experiments.transformers_modeling_backend.trainer import SFTTrainer
 
-from torchtitan.trainer import Trainer
-
-
-@dataclass(kw_only=True, slots=True)
-class TransformersBackendConfig(Trainer.Config):
-    hf_model: str = ""
-    """HuggingFace model ID (e.g., 'Qwen/Qwen2.5-7B')"""
-
-    def build(self, **kwargs):
-        from torchtitan.experiments.transformers_modeling_backend.trainer import (
-            SFTTrainer,
-        )
-
-        if not kwargs:
-            return SFTTrainer(config=replace(self))
-
-        config_fields = {f.name for f in fields(self)}
-        overlap = config_fields & kwargs.keys()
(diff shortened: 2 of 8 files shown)


torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -177,10 +257,18 @@ def _configure_hf_attention(self, attn_implementation: str):
             """Configure HuggingFace attention settings."""
             self._titan_injected_model_args["attn_implementation"] = attn_implementation
             self.attn_implementation = attn_implementation
-            # NOTE:(3outeille):This will force create_causal_mask to return None
-            AttentionInterface._global_mapping[
-                attn_implementation
-            ] = sdpa_attention_forward
+            if attn_implementation == "flex_torchtitan":
```
acisseJZhong, line 263: maybe align the naming with core titan?

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -177,10 +257,18 @@ def _configure_hf_attention(self, attn_implementation: str):
...
+                # HF model code has more dynamo guards than native TorchTitan,
```
acisseJZhong: could you explain more on why HF model code has more dynamo guards? and why setting size to 64?

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -177,10 +257,18 @@ def _configure_hf_attention(self, attn_implementation: str):
...
+                # causing extra recompilations during TP warmup.
+                import torch._dynamo.config
+                torch._dynamo.config.cache_size_limit = 64
+                AttentionInterface._global_mapping[
+                    attn_implementation
+                ] = _flex_torchtitan_attention_forward
+            elif attn_implementation not in AttentionInterface._global_mapping:
+                AttentionInterface._global_mapping[
```
acisseJZhong, line 272: add a log saying attn_implementation not available and default to sdpa?

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -258,8 +346,17 @@ def update_from_config(
             self.mlp_bias = False
             self.use_cache = False
             self.initializer_range = 1.0  # use as std for normal init in embedding
-
-            if not hasattr(self, "inter_dim"):  # Only for llama model
+            # FlexAttention does not support dropout
+            if hasattr(self, "attention_dropout"):
+                self.attention_dropout = 0.0
```
acisseJZhong, line 364: add a warning? what's the gap to support dropout in flex attention?

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: This was one of my ambiguities as well. HF's attention forward has a bunch of if checks (KV cache, output_attentions, use_cache, dict dispatch) that native TorchTitan doesn't have. Each one becomes a dynamo guard. When TP's async collectives produce different tensor types during the first few steps, the combination of these extra guards causes more recompilations than the default limit of 8 allows. It stabilizes after 3-5 steps. So 64 just gives enough room to get there. What are your thoughts?

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: Done.

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: Added. Checking flex_attention, it doesn't have the dropout parameter in it. Now I am not sure about the gap. Dropout could potentially be implemented through score_mod but that's my best guess. HF chose to block rather than implement it.

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: So HF couples two things to the attn_implementation name: which attention function to call and whether to build a mask. We need FlexAttention but can't use HF's mask creation - need to build it before the parallelism  sharding. The custom name lets us split these and register our function, opt out of HF's mask building. My understanding is that this is the same approach pretraining already uses with "sdpa_torchtitan".

torchtitan/experiments/transformers_modeling_backend/configs.py
```diff
@@ -13,3 +13,20 @@
...
+        from torchtitan.experiments.transformers_modeling_backend.trainer import (
+            SFTTrainer,
+        )
+
+        if not kwargs:
+            return SFTTrainer(config=replace(self))
+
+        config_fields = {f.name for f in fields(self)}
+        overlap = config_fields & kwargs.keys()
+        if overlap:
```
acisseJZhong: what is this overlap logic for?

torchtitan/experiments/transformers_modeling_backend/model.py
acisseJZhong: @xmfan would appreciate your opinion.

torchtitan/experiments/transformers_modeling_backend/tokenizer.py
```diff
@@ -0,0 +1,38 @@
...
+        if self.eos_id is None and self._hf_config:
+            eos_str = self._get_token_from_config(self._hf_config, "eos_token")
+            if eos_str is not None:
+                self.eos_token = eos_str
+                self.eos_id = self.tokenizer.token_to_id(eos_str)
+
+    def apply_chat_template(self, messages, **kwargs):
+        kwargs.setdefault("bos_token", self.bos_token or "")
+        kwargs.setdefault("eos_token", self.eos_token or "")
+        kwargs.setdefault("add_generation_prompt", True)
```
acisseJZhong, line 37: why can't we pass these kwargs in when calling this function?  In general, why do we need to create this subclass?

torchtitan/experiments/transformers_modeling_backend/configs.py
HosseinKaviani-H: It's copied from the base Configurable.Config.build(). It prevents accidentally passing a config field as a build() kwarg ([torchtitan/config/configurable.py)](https://github.com/pytorch/torchtitan/blob/main/torchtitan/config/configurable.py#L124))

torchtitan/experiments/transformers_modeling_backend/tokenizer.py
HosseinKaviani-H: The core calls apply_chat_template with only messages and no kwargs. Some HF models chat templates reference bos_token/eos_token in their Jinja template (like Seed-Coder) and the base tokenizer doesn't pass them. The subclass automatically injects these so all HF chat templates work without changing the call site. 

There's also an EOS edge case. Granite uses the same string for both bos_token and eos_token. The base tokenizer has an if/elif that sets BOS but skips EOS when they match. ChatDataset requires eos_id for padding packed sequences so it crashes. The subclass detects and fixes this.

torchtitan/experiments/transformers_modeling_backend/configs.py
acisseJZhong: I think this class should inherit base config? can we call the super class method instead of redoing in subclass?

HosseinKaviani-H: I don't think we can call super().build() directly. This uses self._owner which resolves to Trainer and we need SFTTrainer. One cleaner option could be to just override _owner on the config class instead of reimplementing build():

```
class TransformersBackendConfig(Trainer.Config):
    hf_model: str = ""

TransformersBackendConfig._owner = SFTTrainer
```
Then super().build() would naturally return SFTTrainer. No duplication. Thoughts?

torchtitan/experiments/transformers_modeling_backend/model.py
xmfan: > TP's async collectives produce different tensor types 

What do we mean by tensor "types"? Limit to 64 is quite high, but I think it's fine since it stabilizes after 5 steps.

HosseinKaviani-H: @xmfan By "types" I mean the Python type() - AsyncCollectiveTensor vs regular Tensor. First few steps FSDP's allgather returns AsyncCollectiveTensor, later steps it's a resolved Tensor. Dynamo guards on type() so each variation triggers a recompile. Combined with HF's extra if checks, this goes past 8. Stabilizes after ~10-15 recompiles. I can decrease it to 32 if that makes more sense here?

xmfan: I think this is a new compile guards issue, so the weights are sometimes AsyncCollectiveTensor, iirc the compiler will always await them before running the compiled region and guards were always created on the tensor class. 

It sounds like it doesn't block you, so the increased recompile count sounds good. But if you have time, please file an issue on pytorch repo.

HosseinKaviani-H: @xmfan Sg, I have created a repro and submitted an issue here: https://github.com/pytorch/pytorch/issues/187073

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -12,17 +12,97 @@
...
+        *,
+        attention_masks: BlockMask | None = None,
+        scale: float | None = None,
+        enable_gqa: bool = True,
+        **kwargs,
+    ) -> torch.Tensor:
+        from torch.nn.attention.flex_attention import flex_attention
+
+        # Transpose to (batch, heads, seq, dim) for flex_attention kernel
+        q, k, v = q.transpose(1, 2), k.transpose(1, 2), v.transpose(1, 2)
```
acisseJZhong: could you follow the naming convention in core torchtitan? https://github.com/pytorch/torchtitan/pull/3443/changes

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -12,17 +12,97 @@
...
+        q = query.transpose(1, 2)
+        k = key.transpose(1, 2)
+        v = value.transpose(1, 2)
+        out = flex_module(
+            q, k, v, attention_masks=block_mask, scale=scaling
+        )
+        # Native layout → HF layout
+        return out.transpose(1, 2), None
+
+    # Fallback: call flex_attention directly (no CP support)
```
acisseJZhong: under what case will this trigger? since we always register `HFFlexAttention` under `flex_torchtitan`

torchtitan/experiments/transformers_modeling_backend/trainer.py
```diff
@@ -0,0 +1,73 @@
...
+        extra_kwargs["positions"] = positions
+
+        if self.parallel_dims.cp_enabled:
+            from torchtitan.distributed.context_parallel import (
+                prepare_context_parallel_input,
+            )
+
+            load_balancer = self.config.parallelism.context_parallel_load_balancer
+            if attn_mask_type == "block_causal":
+                load_balancer = "ptrr"
```
acisseJZhong, line 78: why hardcoded to this load balancer?

torchtitan/experiments/transformers_modeling_backend/trainer.py
```diff
@@ -0,0 +1,73 @@
...
+    sharding) so it works correctly with TP + sequence parallel.
+    """
+
+    @sl.log_trace_span("post_dataloading_process")
+    def post_dataloading_process(
+        self, input_dict: dict[str, torch.Tensor], labels: torch.Tensor
+    ) -> tuple[torch.Tensor, torch.Tensor, dict[str, torch.Tensor], dict[str, Any]]:
+        inputs = input_dict["input"]
+        extra_inputs = {k: v for k, v in input_dict.items() if k != "input"}
+        extra_kwargs: dict[str, Any] = {}
```
acisseJZhong: align with https://github.com/pytorch/torchtitan/blob/073221a0a3f64880457cf954d136a73491d39042/torchtitan/trainer.py#L640?

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: Done

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: Good catch - it's dead code. _flex_kernel is attached to every self_attn whenever flex_torchtitan is selected, which is exactly when this forward runs, so it's always present. Removed the fallback and access module._flex_kernel directly. Verified the flex path still trains.

torchtitan/experiments/transformers_modeling_backend/trainer.py
HosseinKaviani-H: Aligned now.

torchtitan/experiments/transformers_modeling_backend/trainer.py
HosseinKaviani-H: ptrr is required, not arbitrary: block_causal builds a BlockMask and runs FlexAttention and _PTRRLoadBalancer is the FlexAttention balancer (it requires a BlockMask). The default headtail is SDPA-only and would mis-balance packed per-document sequences so ptrr is the only valid choice here.

torchtitan/experiments/transformers_modeling_backend/configs.py
acisseJZhong: Got it, why don't we inherit from SFTTrainer.Config?

HosseinKaviani-H: @acisseJZhong Done. Gave SFTTrainer its own nested Config(Trainer.Config) so _owner auto-wires to SFTTrainer, and configs.py is now just  TransformersBackendConfig = SFTTrainer.Config. Dropped the duplicated build()/overlap check. Verified _owner=SFTTrainer and the overlap check still fires.

HosseinKaviani-H:
@fegin would appreciate your comments on this PR.

torchtitan/experiments/transformers_modeling_backend/trainer.py
acisseJZhong: I see https://github.com/pytorch/torchtitan/blob/badf21a1382a31810ffc843ed47166a1b397d2ad/torchtitan/config/configs.py#L213, what if user specified `None`? you would be overriding users' input?

acisseJZhong (approved):

torchtitan/experiments/transformers_modeling_backend/model.py
```diff
@@ -178,10 +260,23 @@ def _configure_hf_attention(self, attn_implementation: str):
...
-            # NOTE:(3outeille):This will force create_causal_mask to return None
-            AttentionInterface._global_mapping[
-                attn_implementation
-            ] = sdpa_attention_forward
+            if attn_implementation == "flex_torchtitan":
+                # HF model code has more dynamo guards than native TorchTitan,
+                # causing extra recompilations during TP warmup.
+                import torch._dynamo.config
+
+                torch._dynamo.config.cache_size_limit = 64
```
acisseJZhong: can you try removing this? I saw there's a fix landed in https://github.com/pytorch/pytorch/issues/187073

HosseinKaviani-H: @acisseJZhong I don't think that fix applies here. #3737 focuses on gpt-oss. Is there another fix I don't know about it?

acisseJZhong: can you follow up on the issue to see if there's a fix?

torchtitan/experiments/transformers_modeling_backend/trainer.py
HosseinKaviani-H: Good point. It was overriding unconditionally. block_causal uses a FlexAttention BlockMask and "headtail" is SDPA-only and can't shard it, so "ptrr" is needed there. Fixed to only override the default and warn - explicit None (disable) and "ptrr" are now respected.

torchtitan/experiments/transformers_modeling_backend/model.py
HosseinKaviani-H: Removed the workaround as it works without it now. Updated the PR description.

fegin (approved):
Attention part looks good to me.

torchtitan/experiments/transformers_modeling_backend/config_registry.py
```diff
@@ -157,8 +157,12 @@ def process_sample(sample):
...
         ),
         training=TrainingConfig(
-            local_batch_size=8,
-            seq_len=2048,
+            # Keep this small: this debug model uses the full Qwen3 vocab
+            # (~152k), so cross-entropy materializes a
+            # local_batch_size * seq_len * vocab logits tensor. batch=8,
+            # seq=2048 is ~9GB in fp32 and OOMs the 22GB CI GPUs.
+            local_batch_size=1,
+            seq_len=1024,
```
HosseinKaviani-H, line 165: @acisseJZhong Just want to make sure you are aligned before I merged this. Changed this to pass the CI test and prevent OOM.

PR #3244, opened 2026-05-06
[graph_trainer] Pass global_valid_tokens into loss_fn instead of dividing externally

ydwu4:
* #3250
* #3248
* #3247
* #3246
* #3249
* __->__ #3244

graph_trainer's compute_loss previously did
``loss_fn(pred, labels) / global_valid_tokens`` — calling the loss without
gvt, then dividing the result. This breaks the "loss is the autograd
endpoint" mental model and forces every loss-side autograd.Function backward
to chain-rule the external division through grad_output, which interacts
poorly with custom Functions whose returned grads live on a different
DeviceMesh than the loss output (e.g. lm_head params on (fsdp, tp) vs loss
on (tp), as we'd hit with chunked CE plumbing).

global_valid_tokens is already part of the BaseLoss.__call__ contract —
``CrossEntropyLoss``, ``MSELoss``, and ``ChunkedCELoss`` all accept it and
divide internally if provided. Passing gvt through restores "loss is the
endpoint": no external scaling on the autograd path, downstream grad_output
is trivially 1.0, and the loss tensor lands on a single mesh determined by
the loss computation rather than the gvt operand's mesh.

Numerically equivalent: BaseLoss.__call__ does
``loss = self.fn(...); if gvt: loss = loss / gvt`` — identical to the prior
external divide for the existing CrossEntropyLoss path that
to_graph_trainer_config swaps to.


torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -41,7 +41,7 @@ def make_fwd_bwd_step(loss_fn):
     # attribute loss nodes in the traced graph.
     @annotate_fn({_MODULE_FQN: "loss"})
     def compute_loss(pred, labels, global_valid_tokens):
-        return loss_fn(pred, labels) / global_valid_tokens
+        return loss_fn(pred, labels, global_valid_tokens)
```
SherlockNoMad: lol, in such case, we can just do 

loss_fn = annotate_fn({_MODULE_FQN: "loss"})(loss_fn)

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/trainer.py
ydwu4: Sure

PR #3245, opened 2026-05-06
Make ChunkedCELoss support torch.autograd.grad

ydwu4:
* #3248
* #3247
* #3246
* __->__ #3245
* #3244

Add a ``support_autograd_grad`` opt-in flag to ``ChunkedCELoss.Config``
that exposes the lm_head parameter gradients as explicit autograd
outputs of the returned loss tensor. Designed for FX-tracing flows
(e.g. graph_trainer's aot_fx_trace) where ``param.grad`` side-effect
writes from ``chunk_loss.backward()`` inside the chunk loop don't
survive into the captured graph and replay therefore produces
all-zero param grads.

Mechanism, when the flag is True:

  - The chunk loop runs unchanged: per-chunk ``chunk_loss.backward()``
    populates ``lm_head.weight.grad`` with the correctly sharded value
    (FSDP last-chunk reduce-scatter still handles the actual reduction).
  - After the loop, the sharded ``param.grad`` is captured via
    ``p.grad.detach()`` and ``p.grad`` is cleared.
  - The captured grads (plus the existing accumulated hidden_states
    grad) are plumbed through a new ``_ChunkedLossWithParamGrads``
    autograd Function as saved tensors. Its backward returns them as
    grads for the corresponding lm_head parameter inputs, so outer
    ``torch.autograd.grad(loss, lm_head.parameters())`` resolves to
    real gradients instead of None / zeros.
  - Under FSDP, ``set_requires_gradient_sync(False)`` is set on lm_head
    and restored after the outer backward via a callback queued on the
    autograd engine. Without this, outer ``loss.backward()`` would
    re-fire the post-accumulate-grad hook on already-sharded grads and
    either error or produce wrong values.

Both ``_DecoderOutputGradientBackProp`` (existing) and the new
``_ChunkedLossWithParamGrads`` multiply the saved grad by
``grad_output`` before returning so external scaling of the loss
(e.g. ``loss / global_valid_tokens``) chain-rules correctly. Saved
grads can live on a different ``DeviceMesh`` than ``grad_output``
(e.g. params on ``(fsdp, tp)`` vs scalar grad on ``(tp)``), so
``grad_output`` is unwrapped via ``to_local()`` when it's a DTensor
to avoid mesh-mismatch errors in the multiply.

Tests:

  - tests/unit_tests/test_loss.py: parametrized ``support_autograd_grad
    in {False, True}`` equivalence check against the unchunked CE
    standard path; bitwise (rtol=0, atol=0) check that True and False
    paths produce identical grads; side-effect contract check that the
    True path doesn't populate ``param.grad``.
  - graph_trainer/tests/test_trace_module.py: end-to-end test that
    traces a small ``lm_head + ChunkedCELoss(support_autograd_grad=
    True)`` train_step via ``trace_train_step`` and verifies
    ``torch.equal`` between eager and replayed loss + h grad + lm_head
    grad on a CUDA model.

The flag defaults to False so existing callers (the eager torchtitan
trainer) are unaffected; consumers that want explicit param grads
opt in. graph_trainer wires this in a separate commit.

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -229,6 +229,14 @@ class ChunkedCELoss(BaseLoss):
     class Config(BaseLoss.Config):
         num_chunks: int = 8
         """Number of chunks to split the sequence into."""
+        support_autograd_grad: bool = False
+        """If True, plumb the sharded lm_head param grads (already produced
+        by the chunk loop's reduce-scatter) out as explicit autograd outputs
+        so outer ``torch.autograd.grad(loss, [hidden_states,
+        *lm_head.parameters()])`` returns them directly — no reliance on
+        ``param.grad`` side effects. Required for FX tracing / replay where
+        side-effecting ``.grad`` writes don't survive. Compatible with both
+        outer ``loss.backward()`` and ``torch.autograd.grad`` consumers."""
 
     def __init__(
         self,
@@ -239,6 +247,7 @@ class ChunkedCELoss(BaseLoss):
         self.fn: LossFunction = cross_entropy_loss
         self._maybe_compile(compile_config)
         self.num_chunks = config.num_chunks
+        self.support_autograd_grad = config.support_autograd_grad
         self.lm_head: nn.Module | None = None
 
     def set_lm_head(self, lm_head: nn.Module) -> None:
@@ -345,6 +354,16 @@ class ChunkedCELoss(BaseLoss):
 
         accumulated_grad = grad_accumulator.result().to(hidden_states.dtype)
 
+        if self.support_autograd_grad:
+            return _ChunkedLossWithParamGrads.apply(
+                hidden_states,
+                accumulated_grad,
+                total_loss,
+                lm_head,
+                fsdp_enabled,
+                *lm_head.parameters(),
+            )
+
         # Return a differentiable loss via _DecoderOutputGradientBackProp. When
         # .backward() is called (by the trainer or PP schedule), autograd
         # calls _DecoderOutputGradientBackProp.backward which returns accumulated_grad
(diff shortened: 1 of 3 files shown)


torchtitan/components/loss.py
```diff
@@ -383,10 +402,92 @@ def forward(
...
         # Return accumulated_grad as the gradient for hidden_states.
         # Autograd then propagates this through hidden_states' existing
         # decoder graph — equivalent to hidden_states.backward(accumulated_grad)
         # but expressed as a return value so autograd handles the traversal
         # in a single pass (no "backward through graph twice" error).
-        return accumulated_grad, None, None
+        # Multiply by grad_output so external scaling of the loss (e.g.
+        # ``loss / global_valid_tokens``) chain-rules into the grad. See
+        # _ChunkedLossWithParamGrads.backward for the to_local() rationale.
+        if isinstance(grad_output, DTensor):
```
ydwu4, line 416: The more robust of way doing bacwkard is to apply chain rule and do grad_output *  accumulated_grad.

However, in order to do this, we need to to_local because the device mesh is different: 

```python
RuntimeError: Cannot run aten.mul.Tensor on inputs with different meshes:
  got DeviceMesh((fsdp=4, tp=2)) and DeviceMesh((tp=2))

Why 
- grad_output = torch.ones_like(loss) from autograd's start. loss lives on hidden_states' mesh = (tp,). So grad_output is a Replicate() scalar DTensor on (tp,).
- Saved param_grad (e.g. lm_head.weight.grad) lives on the param's mesh = (fsdp, tp).
- Even with no external scaling, the multiply at backward time crosses meshes.
```

Maybe we should just keep the original implementation and add a note to users so they don't use the loss further?

PR #3246, opened 2026-05-06
[graph_trainer] Add preserve_ops_order utility for shielding regions from graph passes

ydwu4:
* #3250
* #3248
* #3247
* __->__ #3246
* #3249

**Problem:**
We don't functionalize the graph so **the implicit read-after-write dependency is
invisible to graph passes**. 

The **joint_transformer_block_bucketing_reordering_pass** moving readers of ChunkedCELoss's 
grad-accumulator buffer past the in-place ``aten.copy_`` writes that fill it.

**Implementation:**
Add skip_module(fqns) decorator that wraps a graph pass so it **cannot**
reorder or restructure nodes whose ``module_fqn`` annotation is in
``fqns``. Mechanically:

  1. Before the wrapped pass runs, ``_pin_ops_order`` walks the
     graph and pairs consecutive nodes in the listed modules with
     ``preserve_node_ordering`` from
     ``torch._inductor.fx_passes.control_dependencies``. Each anchored
     node ends up wrapped in a ``control_deps`` HOP whose first arg
     points to the previous anchored node, so any pass that does
     topological scheduling against arg edges sees the region as a
     single linear chain that cannot be split.
  2. The wrapped pass runs against the shielded graph.
  3. ``_unpin_ops_order`` unwraps every ``control_deps`` HOP back
     to the original op so subsequent passes see the unmodified graph.


xmfan:
what if we did a graph pass to insert node dependencies instead?

ydwu4:
That would also work, If we want the dependencies to be always in the graph, could call the _pin_ops_order without _unpin_ops_order. 

 I want to do the dependency preserving in a local per-pass manner (kind of like a pass modifier, higher order pass lol). Many passes don't reorder things anyway. Keeping the control deps hop in graph seems a bit overshooting for them.

check the usage in https://github.com/pytorch/torchtitan/pull/3247/changes!

torchtitan/experiments/graph_trainer/preserve_ops_order.py
```diff
@@ -0,0 +1,157 @@
...
+        if not get_attr_node.users:
+            gm.graph.erase_node(get_attr_node)
+            delattr(gm, str(get_attr_node.target))
+
+    if control_deps_nodes:
+        gm.graph.lint()
+        gm.recompile()
+
+
+def preserve_ops_order(fqns: list[str]) -> Callable[[Callable], Callable]:
```
SherlockNoMad, line 127: I think this is an interesting generic solution to prevent nodes from moving around... 

but I see this as a stop-bleed solution that would come handy when someone need a quick and dirty fix. 

My worry is that its existence will encourage irresponsible use of topo_sort, and passes should still try best to avoid global manipulation of the graph. 

so we should still land this, but as an stop-gap quick fix pass.

ydwu4: Agree global transformation is hard to get right for a non-functional graph. But I also don't want to restrict us to only local transformations: isn't global transformation the main magic we get out of full graph tracing (compared with eager/ partial graph)? Topo-sorting and partial ordering are very generic concepts in graph algorithms and we could provide utilities for users to easily do that IMO.

ydwu4: Having said that, I'm not sure what pass would actually benefit from a global topo-sort!

SherlockNoMad (approved):
didn't review line-by-line, but I think this is useful... 

I would suggest put this under an experimental/ dir.

PR #3247, opened 2026-05-06
[graph_trainer] Fix ChunkedCELoss interactions with SAC and bucketing passes

ydwu4:
* #3250
* #3248
* __->__ #3247

Two pass-level fixes needed for ChunkedCELoss to compose correctly
with graph_trainer's compile pipeline. Neither fix exercises a
ChunkedCELoss code path on its own (graph_trainer's
``to_graph_trainer_config`` still swaps ChunkedCELoss to
CrossEntropyLoss until the next commit flips the toggle); the fixes
land here so the toggle commit doesn't bundle pass-pipeline changes
with config flips.

1. apply_sac_pass: skip nodes whose module is outside the transformer
   layer stack (``layer_id == _NOT_IN_LAYERS``). Matches eager
   ``apply_ac`` (torchtitan/distributed/activation_checkpoint.py),
   which iterates ``model.layers.named_children()`` and wraps each
   transformer block — never the loss / lm_head / norm at the top
   level.

   Without this skip, ChunkedCELoss's chunked region (lm_head +
   per-chunk ce_loss) gets tagged ``PREFER_RECOMPUTE``. Per-chunk
   ``chunk_loss.backward()`` dispatches autograd ops that get tagged
   ``meta["autograd_backward"] = True`` (one tagged region per
   chunk, plus the outer training-loop backward). When the upstream
   ``remat_using_tags_for_fwd_loss_bwd_graph`` pass then walks the
   joint graph and finds N+1 disjoint backward regions where
   PREFER_RECOMPUTE forward nodes feed in, it errors with: "Detected
   N disjoint backward regions that require recomputation, but
   remat only supports one such region in a forward-loss-backward
   graph."

2. joint_transformer_block_bucketing_reordering_pass: wrap with
   skip_module(["loss", "lm_head"]). The bucketing pass topologically
   reorders ops to overlap collectives with compute but uses arg-edge
   topology only, so it doesn't track tensor aliasing/mutation
   dependencies. ChunkedCELoss's GradAccumulator populates a buffer
   via ``aten.copy_(slice_view, src)`` in-place writes; the
   downstream ``view + to`` that read the buffer have no arg-edge
   dependence on the writes (the dependency is via shared storage
   between ``slice_view`` and the buffer, not via FX node args).

   The bucketing pass therefore sees the readers as freely movable
   compute and pulls them into an overlap window of a downstream
   collective — leaving the readers seeing the unwritten zeros
   buffer. Empirically: ``accumulated_grad`` returned for
   ``hidden_states`` is all-zero at runtime (verified by forcing
   ``torch.ones_like`` and watching grad_norm jump), so decoder
   backward propagates zero and every decoder gradient is zero.
   Loss matches eager step 1 (forward path is unaffected) but
   diverges step 2+ because the optimizer only updates the lm_head.

   Wrapping the bucketing entry in ``skip_module`` shields the
   loss/lm_head region behind ``control_deps`` HOPs so the bucketing
   pass treats it as an opaque ordered chain and respects the
   original write-then-read sequence.

End-to-end verification of the cumulative stack (this commit + the
flip in the next commit) is documented in the next commit's message.


torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -170,9 +174,20 @@ def compile_time_passes(
...
+        # ChunkedCELoss's per-chunk in-place writes into a grad accumulator
+        # buffer (and the downstream readers of that buffer) live under
+        # ``module_fqn`` ``loss``/``lm_head``; without pinning their order,
+        # the bucketing pass moves the readers past the writes and decoder
+        # grads come out all-zero. ``preserve_ops_order`` chains the loss
+        # region through ``control_deps`` HOPs so the bucketing pass cannot
+        # break the read-after-write ordering.
+        preserve_ops_order(["loss", "lm_head"])(
+            functools.partial(
+                joint_transformer_block_bucketing_reordering_pass,
```
SherlockNoMad, line 182: I had previously intended to ban the use of topo_sort in graph pass to prevent such case from happening. 

This preserve_ops_order is giving me some new ideas.

ydwu4: Yeah, we could do this partial-order thing through annotations/graph passes.

PR #3248, opened 2026-05-06
[graph_trainer] Enable ChunkedCELoss for deepseek_v3 and qwen3

ydwu4:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):

* __->__ #3248

Previously ``to_graph_trainer_config`` swapped any ``ChunkedCELoss``
config to ``CrossEntropyLoss`` because graph_trainer's tracer relied
on ``param.grad`` side effects that ChunkedCELoss's manual chunked
backward didn't preserve through the FX trace.

With ``ChunkedCELoss(support_autograd_grad=True)`` (commit a) plumbing
sharded lm_head grads as explicit autograd outputs, and the SAC + bucketing
pass fixes (commit d) that prevent loss-region nodes from being
recomputed/reordered, ChunkedCELoss now traces correctly. Flip
``support_autograd_grad=True`` instead of swapping to CrossEntropyLoss.

Cleanup that follows the swap removal:

  - ``deepseek_v3_debugmodel_ep_ce_loss``, ``qwen3_debugmodel_ce_loss``,
    ``qwen3_moe_debug_ep_ce_loss``: graph_trainer-only baselines that
    existed solely so loss_compare's eager side matched graph_trainer's
    forced CE. Now obsolete; deleted. (``llama3_debugmodel_ce_loss``
    in the main models registry stays — separately used by TP/SP
    integration tests.)
  - ``test_numerics.py``: baselines for llama3, qwen3 (dense + MoE),
    and deepseek_v3 flip from ``*_ce_loss`` to the standard
    chunked-CE configs. Both eager and graph_trainer now use
    ChunkedCELoss for these comparisons.

Verified end-to-end (FSDP=4 x TP=2 [x EP=2 for MoE], deterministic
seed=42, 5 steps, bitwise loss + grad_norm match against eager):

  - graph_trainer.qwen3_debugmodel                       (dense)
  - graph_trainer.llama3_debugmodel                      (dense)
  - graph_trainer.deepseek_v3_debugmodel                 (dense)
  - graph_trainer.deepseek_v3_debugmodel_ep              (MoE)
  - graph_trainer.deepseek_v3_debugmodel_flex_attn_ep    (MoE+flex)
  - graph_trainer.qwen3_debugmodel_moe_ep                (MoE)

PR #3249, opened 2026-05-06
Make ChunkedCELoss support torch.autograd.grad

ydwu4:
* #3250
* #3248
* #3247
* #3246
* __->__ #3249

Add a ``support_autograd_grad`` opt-in flag to ``ChunkedCELoss.Config``
that exposes the lm_head parameter gradients as explicit autograd
outputs of the returned loss tensor. Designed for FX-tracing flows
(e.g. graph_trainer's aot_fx_trace) where ``param.grad`` side-effect
writes from ``chunk_loss.backward()`` inside the chunk loop don't
survive into the captured graph and replay therefore produces
all-zero param grads.

Mechanism, when the flag is True:

  - The chunk loop runs unchanged: per-chunk ``chunk_loss.backward()``
    populates ``lm_head.weight.grad`` with the correctly sharded value
    (FSDP last-chunk reduce-scatter still handles the actual reduction).
  - After the loop, the sharded ``param.grad`` is captured via
    ``p.grad.detach()`` and ``p.grad`` is cleared.
  - The captured grads (plus the existing accumulated hidden_states
    grad) are plumbed through a new ``_ChunkedLossWithParamGrads``
    autograd Function as saved tensors. Its backward returns them as
    grads for the corresponding lm_head parameter inputs, so outer
    ``torch.autograd.grad(loss, lm_head.parameters())`` resolves to
    real gradients instead of None / zeros.
  - Under FSDP, ``set_requires_gradient_sync(False)`` is set on lm_head
    and restored after the outer backward via a callback queued on the
    autograd engine. Without this, outer ``loss.backward()`` would
    re-fire the post-accumulate-grad hook on already-sharded grads and
    either error or produce wrong values.

Both autograd Functions (the existing ``_DecoderOutputGradientBackProp``
and the new ``_ChunkedLossWithParamGrads``) return saved grads as-is
without chain-ruling through ``grad_output``. The contract is that the
loss returned by these Functions is the autograd endpoint — callers
must pass any scaling factor (e.g. ``global_valid_tokens``) into
``loss_fn`` rather than dividing the returned loss externally. See
graph_trainer's ``compute_loss`` for the canonical pattern. This
matches the pre-existing behavior of ``_DecoderOutputGradientBackProp``
and avoids a structural FSDP+TP problem: ``grad_output`` is a
``Replicate()`` scalar DTensor on the loss's mesh (typically ``(tp,)``)
while saved param grads live on the params' mesh (e.g. ``(fsdp, tp)``);
DTensor refuses cross-mesh ``aten.mul.Tensor``, so any chain-rule
multiply would crash at runtime.

Tests:

  - tests/unit_tests/test_loss.py: parametrized ``support_autograd_grad
    in {False, True}`` equivalence check against the unchunked CE
    standard path; bitwise (rtol=0, atol=0) check that True and False
    paths produce identical grads; side-effect contract check that the
    True path doesn't populate ``param.grad``.
  - graph_trainer/tests/test_trace_module.py: end-to-end test that
    traces a small ``lm_head + ChunkedCELoss(support_autograd_grad=
    True)`` train_step via ``trace_train_step`` and verifies
    ``torch.equal`` between eager and replayed loss + h grad + lm_head
    grad on a CUDA model.

The flag defaults to False so existing callers (the eager torchtitan
trainer) are unaffected; consumers that want explicit param grads
opt in. graph_trainer wires this in a separate commit.


torchtitan/components/loss.py
```diff
@@ -229,6 +229,16 @@ class ChunkedCELoss(BaseLoss):
...
         num_chunks: int = 8
         """Number of chunks to split the sequence into."""
+        support_autograd_grad: bool = False
+        """If True, plumb the sharded lm_head param grads (already produced
+        by the chunk loop's reduce-scatter) out as explicit autograd outputs
+        so outer ``torch.autograd.grad(loss, [hidden_states,
+        *lm_head.parameters()])`` returns them directly — no reliance on
+        ``param.grad`` side effects. Required for FX tracing / replay where
+        side-effecting ``.grad`` writes don't survive. Compatible with both
+        outer ``loss.backward()`` and ``torch.autograd.grad`` consumers.
```
SherlockNoMad: this is a smart fix. 

do you have a tlparse on the graph of chunked loss I can take a look?

ydwu4: Sure here is one: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/yidi-graph-trainer-moe-deepseek-v3-debugmodel-no-chainrule/rank_0/-_-_-_-/make_fx_graph_traced_1.txt?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000


For a smaller model: check https://www.internalfb.com/phabricator/paste/view/P2311018461

torchtitan/components/loss.py
```diff
@@ -229,6 +229,16 @@ class ChunkedCELoss(BaseLoss):
     class Config(BaseLoss.Config):
         num_chunks: int = 8
         """Number of chunks to split the sequence into."""
+        support_autograd_grad: bool = False
```
tianyu-l: sounds that a cleaner, non-intrusive way would be
- making a subclass of chunked loss class, e.g. by turning the autograd function a class variable and override it
- swapping to use it for graph trainer configs

ydwu4: Sure, refactored accordingly!

torchtitan/components/loss.py
```diff
@@ -345,15 +345,64 @@ def __call__(
...
+        returns ``accumulated_grad`` as the gradient for ``hidden_states``,
+        propagating through the decoder. Subclasses override to swap in a
+        different autograd Function.
+        """
         return _DecoderOutputGradientBackProp.apply(
             hidden_states, accumulated_grad, total_loss
         )
 
 
+class ChunkedCELossWithParamGrads(ChunkedCELoss):
```
tianyu-l: please define this in graph trainer, as it's not required as part of core

torchtitan/components/loss.py
```diff
@@ -389,4 +438,86 @@ def backward(  # pyrefly: ignore[bad-override]
...
         # but expressed as a return value so autograd handles the traversal
         # in a single pass (no "backward through graph twice" error).
+        # Note: this is not safe if downstream accidentally runs tensor ops after
+        # the loss returns, which would produce a non-trivial grad_output that we need
+        # to properly handle. The complicated part is that grad_output might not be
+        # on the same device mesh as accumlated_grad.
         return accumulated_grad, None, None
+
+
+class _ChunkedLossWithParamGrads(torch.autograd.Function):
```
tianyu-l: same here

torchtitan/components/loss.py
ydwu4: Moved into graph trainer.

tianyu-l (approved):
lgtm on the core side

torchtitan/components/loss.py
SherlockNoMad: nice, the graph looks good to me.

torchtitan/experiments/graph_trainer/chunked_loss.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Designed for graph_trainer, where the chunk loop's per-chunk
+    ``param.grad`` side-effect writes don't survive the captured graph and
+    replay therefore produces all-zero param grads. Compatible with both
+    outer ``loss.backward()`` and ``torch.autograd.grad`` consumers.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(ChunkedCELoss.Config):
+        pass
```
SherlockNoMad: not sure if this required or not?

SherlockNoMad (approved):

SherlockNoMad:
in the latter PR, we should have a variant that uses chunked loss in the test_bitwise_deterministic.py

torchtitan/experiments/graph_trainer/chunked_loss.py
ydwu4: Probably not. Can just remove it I feel. If we need, we could add it back later.

PR #3250, opened 2026-05-06
[loss] Chain-rule grad_output through ChunkedCELoss autograd Function backwards

ydwu4:
* __->__ #3250
* #3248
* #3247

Both ``_DecoderOutputGradientBackProp`` (eager + ``support_autograd_grad
= False``) and ``_ChunkedLossWithParamGrads`` (``support_autograd_grad =
True``) previously returned saved grads as-is, ignoring ``grad_output``.
This violated the autograd contract — a Function backward owes the
caller ``dL/dinput = dout/dinput * grad_output`` — but worked in
practice because every in-tree caller follows the convention of passing
scaling factors (e.g. ``global_valid_tokens``) into ``loss_fn`` rather
than scaling the returned loss externally, so ``grad_output`` is
always the autograd-start ``ones_like(loss)``.

Multiply through ``grad_output`` so the contract holds for any
consumer, including future ones that scale externally. The naive
``g * grad_output`` is rejected by DTensor on FSDP+TP layouts because
``grad_output`` lives on the loss output's mesh (typically the
activation mesh ``(tp,)``) while saved param grads live on the params'
mesh (e.g. ``(fsdp, tp)``); add a ``_maybe_redistribute_multiply``
helper that redistributes ``grad_output`` to ``g``'s mesh first when
both are DTensors on different meshes. For the only ``grad_output``
shape we actually encounter — a ``Replicate()`` scalar (what
``torch.autograd.grad`` constructs for a scalar loss output) — the
redistribute is comm-free at runtime; it just rewrites placement
metadata.

The autograd-correctness change is independent of the ChunkedCELoss
enablement work: when ``grad_output`` is the autograd-start
``ones_like(loss)`` (the case for every in-tree caller after the
preceding commit "[graph_trainer] Pass global_valid_tokens into
loss_fn instead of dividing externally"), this commit is a runtime
no-op. It lands as a separate commit to keep the contract-correctness
discussion separate from the enablement diff. The risk of *not*
shipping this is silent wrong gradients for any future consumer that
scales the loss externally, which would be near-impossible to detect
from training logs.

Existing tests (``tests/unit_tests/test_loss.py``,
``test_skip_module.py``, ``test_trace_module.py``) still pass.
End-to-end equivalence: verified that the previous commit ("Make
ChunkedCELoss support torch.autograd.grad") bitwise-matches eager
across 6 graph_trainer configs (3 dense + 3 MoE) with
``support_autograd_grad=True``; this commit is mathematically a no-op
on top of that since ``grad_output = ones_like(loss)`` in all those
configs.


torchtitan/components/loss.py
```diff
@@ -375,6 +375,33 @@ def __call__(
...
+    redistribute is a no-op at runtime — no collective fires.
+    """
+    from torch.distributed.tensor import DTensor, Replicate
+
+    if (
+        isinstance(g, DTensor)
+        and isinstance(grad_output, DTensor)
+        and g.device_mesh != grad_output.device_mesh
+    ):
+        grad_output = grad_output.redistribute(
```
ydwu4: ugh.. this is vibecoded. Not sure it's the correct way of doing this. The idea is to allow grad_output ( a sclalar tensor) to mulfiply with gradients (with a different device mesh).

torchtitan/components/loss.py
```diff
@@ -375,6 +375,31 @@ def __call__(
...
+    tensor cannot be sharded, and ``torch.autograd.grad`` seeds the
+    backward with ``ones_like(loss)``. We extract the local scalar via
+    ``to_local`` rather than ``redistribute``: DTensor's redistribute
+    does not implement cross-mesh comm (raises
+    ``NotImplementedError("Cross device mesh comm not supported yet!")``),
+    so a naive ``g * grad_output`` crashes whenever ``g`` (saved param
+    grad on, e.g., ``(fsdp, tp)``) and ``grad_output`` (loss on, e.g.,
+    ``(tp,)``) live on different meshes. ``to_local`` on a Replicate
+    scalar is local-only — no collective fires — and ``g * scalar`` then
+    broadcasts cleanly regardless of ``g``'s mesh.
```
fegin, line 388: This is not a problem under full dtensor mode as we always operate on a full SPMD mesh. So I prefer not to land this PR.

ydwu4: I think the real problem is that people might call ChunkedCELoss then do some more tensor ops on the output and get a new loss. We need to figure out a way to back propagate the grad_out somehow.

fegin: Do you have a concrete example in TorchTitan?  If not now and this is a forward looking fix, then it may not be an issue once we migrate to spmd_types.

ydwu4: https://github.com/pytorch/torchtitan/pull/3244 is an example. Can you elaborate a bit more why spmd_types can help on this? 

To be clear, it's fine we're not landing this.  If ChunkedCELoss is a user-facing API, not back propagating grad_out is a footgun. We probably should warn user about it.

fegin: Sorry my statement was too vague and not accurate. What I wanted to say is that we should avoid `to_local()` because it will not work with full_dtensor case. And The reason why you need to do `to_local` is the cross mesh issue, which is going to be fixed with full dtensor and spmd types.

Back propgating grad_out is not a concern.

ydwu4: Ok! Do we want to land a temporary fix to stop bleeding? If so, any recommendation on the proper implementation before full dtensor and spmd types support land?

fegin: Can this be moved to `torchtitan/experiments/graph_trainer/chunked_loss.py`? I saw that in the early PR, you already create a subclass. This way you won't be blocked by the full dtensor and spmd dtypes adoption.

ydwu4: sure i'll just do it for the graph_trainer chunked loss then

PR #3253, opened 2026-05-06
[spmd_types] spmd infra

pianpwk:
Main infra PR for spmd_types + titan.

`spmd_types.py`:
- Adds a thread-local DeviceMesh stack, and current_mesh(), set_current_mesh() helpers, to be set and accessed at model init (weight parallelize) and runtime (PG access for collectives)
- Various helpers for spmd_types : module-boundary redistributions, spmd -> DTensor placement translation (for full_dtensor backend), `mesh_size(axis_name)` helper that returns > 1 when mesh is set & axis is active. Some of this will be moved to spmd_types in near-term, see comments.

`parallel_dims.py`:
- For spmd backend, we need 2 views over the world mesh: [pp, dp, cp, tp] for typechecking, and [pp, dp_replicate, dp_shard, cp, tp] to pass the unfolded fsdp axes to fully_shard, via DataParallelMeshDims. So we hold both the full-DTensor-style dense mesh, as well as the "typechecking" mesh.

`module.py`: module.parallelize() paths for spmd backend: weight init, local SPMD drop in, input/output redistribution.

`trainer.py/utils.py`: typechecking context, input annotation, spmd set_current_spmd_mesh, PP + typechecking raises a hard error.

Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #3278
* #3472
* #3632
* #3631
* #3468
* #3471
* __->__ #3253

PR #3254, opened 2026-05-06
[spmd_llama3] FSDP/HSDP

pianpwk:
* #3366
* #3344
* #3255
* __->__ #3254
* #3278
* #3337
* #3253


torchtitan/models/common/decoder_sharding.py
```diff
@@ -16,39 +19,45 @@
...
+
+    SPMD: DP axes are ``S(0)``, CP defaults to ``S(1)``.
     """
+    if cp is None:
+        cp = shard.S(1)
     return {
-        DP_REPLICATE: Shard(0),
-        DP_SHARD: Shard(0),
+        DP_REPLICATE: shard.S(0),
+        DP_SHARD: shard.S(0),
```
ezyang: You might consider doing this mechanical rewrite in its entirely own PR for ease of reviewing

torchtitan/trainer.py
```diff
@@ -682,15 +707,49 @@ def forward_backward_step(
...
+            current_mesh = (
+                spmd.set_current_mesh(spmd_state().all_axes)
+                if is_spmd_active()
+                else contextlib.nullcontext()
+            )
+            typechecker = (
+                spmd.typecheck(local=(self.spmd_typechecking == "local"))
+                if is_spmd_active() and self.spmd_typechecking in ("local", "global")
+                else contextlib.nullcontext()
+            )
```
ezyang: OK, so, are you sure you want to do it this way? There are some benefits to having typechecking always be on, but the typechecker is designed without efficiency in mind and so if you are running it on every training step and you don't have something like CUDA graphs going, I think this is just an unnecessary tax on training QPS. Most of the other internal examples setup the spmd.typecheck context manager specifically in a test rather than directly in trainer.

torchtitan/trainer.py
```diff
@@ -461,9 +478,12 @@ def __init__(self, config: Config):
         )
 
         loss_parallel_enabled = (
-            parallel_dims.tp_enabled and not config.parallelism.disable_loss_parallel
+            parallel_dims.tp_enabled
+            and not config.parallelism.disable_loss_parallel
+            and not is_spmd_active()
```
ezyang: Fixed in the next PR I guess? Good to have comments discussing TODO in this case

torchtitan/trainer.py
```diff
@@ -48,9 +53,19 @@
...
+from torchtitan.protocols.types import MeshAxisName
 from torchtitan.tools.logging import logger
 from torchtitan.tools.profiler import Profiler
 
+# Input annotation for SPMD path: (batch, seq) — DP on batch, CP on seq.
+_INPUT_SPMD_TYPES: NamedPlacement = {
+    MeshAxisName.DP_REPLICATE: spmd.S(0),
+    MeshAxisName.DP_SHARD: spmd.S(0),
+    MeshAxisName.CP: spmd.S(1),
+    MeshAxisName.TP: spmd.R,
```
ezyang: Is this actually known? Isn't the exact location of batch vs sequence a model specific thing? Or can you /only/ do that in the torchtitan trainer?

ezyang: Claude says it's sus these are applied to `extra_inputs`

torchtitan/components/loss.py
```diff
@@ -366,6 +379,20 @@ class _DecoderOutputGradientBackProp(torch.autograd.Function):
...
+        accumulated_grad: torch.Tensor,
+        loss: torch.Tensor,
+    ) -> torch.Tensor:
+        """Output has the same type as ``loss`` (3rd arg)."""
+        result = _DecoderOutputGradientBackProp.apply(
+            hidden_states, accumulated_grad, loss
+        )
+        if spmd.has_local_type(loss):
+            spmd.assert_type(result, spmd.get_local_type(loss))
+        return result
```
ezyang, line 591: Are you sure this isn't just a `register_local_autograd_function`?

torchtitan/models/llama3/parallelize.py
```diff
@@ -89,8 +88,25 @@ def parallelize_llama(
...
+            mesh_names.append("tp")
+        dp_mesh = parallel_dims.get_mesh(mesh_names)
+        dp_mesh_dims = DataParallelMeshDims(
+            shard="dp_shard",
+            replicate="dp_replicate" if parallel_dims.dp_replicate_enabled else None,
+        )
+    else:
+        names = ["dp_replicate", "fsdp"] if parallel_dims.dp_replicate_enabled else ["fsdp"]
+        dp_mesh = parallel_dims.get_mesh(names)
+        dp_mesh_dims = None
```
ezyang: I'm still pretty unsure about the difference here.

torchtitan/components/loss.py
pianpwk: hmm on DP, we have `hidden_states: V, accumulated_grad: V, loss: P` and even though they don't interact, the typechecker complains they don't mix.. maybe there's a smart workaround?

PR #3255, opened 2026-05-06
[spmd_llama3] TP/SP/LP

pianpwk:
standalone TP, TP+SP, TP+LP, TP+SP+LP bitwise compared to main

* #3366
* #3344
* __->__ #3255
* #3254
* #3278
* #3337
* #3253


torchtitan/trainer.py
```diff
@@ -67,6 +68,39 @@
 }
 
 
+def convert_tp_states_for_compute(model: torch.nn.Module, tp_pg) -> None:
```
pianpwk: was told FSDP2 handles TP AR, and we can just init as R@tp ?

PR #3257, opened 2026-05-07
Wrong token count in validation when CP>1

francesco-bertolotti:
## Wrong token count in validation

When `CP > 1`, `post_dataloading_process` shards inputs across ranks. During validation, this causes token counts to be incorrect because counting happens after sharding. This is not an issue during training, where tokens are counted before `post_dataloading_process`.

### Fix

Move `post_dataloading_process` to occur after token counting in validation.

## `post_dataloading_process` not receiving the correct `attn_config`

At some point, the validation and trainer implementations of `post_dataloading_process` became misaligned. In validation, `attn_config` was not being retrieved correctly and defaulted to `None`. As a consequence, the mask type defaulted to `causal` even when `block_causal` was configured.

### Fix

As a temporary fix, I copied the relevant logic from the trainer into the validation `post_dataloading_process`.

This comment highlights the issue:
https://github.com/pytorch/torchtitan/blob/ae5ff104999a937ce0d34d8a720969ae23110140/torchtitan/components/validate.py#L181-L185

I can open a follow-up PR to address this more cleanly. A few possible approaches:

* Make `Validator` inherit from the trainer so it reuses `post_dataloading_process` directly. This feels somewhat heavy-handed since that is the only functionality we need to share.
* Pass `post_dataloading_process` into `Validator.__init__`. We could also make the trainer implementation a `@staticmethod` to reduce coupling.

### DIFF
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -256,11 +256,6 @@ class Validator(BaseValidator):
                 input_dict[k] = v.to(device_type)
             labels = labels.to(device_type)
 
-            # Process data (extract inputs, handle attention masks, CP sharding)
-            inputs, labels, extra_inputs, extra_kwargs = self.post_dataloading_process(
-                input_dict, labels, model_parts
-            )
-
             # Count valid tokens for this batch
             local_valid_tokens = torch.tensor(0, dtype=torch.int64, device=device_type)
             local_valid_tokens += (labels != IGNORE_INDEX).sum()
@@ -273,6 +268,11 @@ class Validator(BaseValidator):
                 )
             else:
                 global_valid_tokens = local_valid_tokens.float()
+
+           # Process data (extract inputs, handle attention masks, CP sharding)
+            inputs, labels, extra_inputs, extra_kwargs = self.post_dataloading_process(
+                input_dict, labels, model_parts
+            )
 
             if parallel_dims.pp_enabled:
                 assert self.pp_schedule is not None
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -21,6 +21,7 @@ from torchtitan.config import Configurable, ParallelismConfig
 from torchtitan.distributed import ParallelDims, utils as dist_utils
 from torchtitan.distributed.context_parallel import prepare_context_parallel_input
 from torchtitan.hf_datasets.text_datasets import HuggingFaceTextDataLoader
+from torchtitan.models.common.attention import FlexAttention, VarlenAttention
 from torchtitan.models.common.decoder import Decoder
 from torchtitan.observability import structured_logger as sl
 from torchtitan.tools import utils
@@ -188,6 +189,7 @@ class Validator(BaseValidator):
         positions = extra_inputs.pop("positions", None)
         if isinstance(model_config, Decoder.Config):
             attn_config = model_config.layers[0].attention
+            inner_attention = attn_config.inner_attention
 
             if attn_config.mask_type == "block_causal":
                 assert (
@@ -198,14 +200,13 @@ class Validator(BaseValidator):
                     inputs.shape[1], dtype=torch.int32, device=inputs.device
                 ).repeat(inputs.shape[0], 1)
 
-        try:
-            extra_kwargs["attention_masks"] = cast(
-                Decoder, model_parts[0]
-            ).get_attention_masks(
-                positions=positions,  # pyrefly: ignore [bad-argument-type]
-            )
-        except TypeError:
-            pass
+            if isinstance(
+                inner_attention, (FlexAttention.Config, VarlenAttention.Config)
+            ):
+                model = cast(Decoder, model_parts[0])
+                extra_kwargs["attention_masks"] = model.get_attention_masks(
+                    positions=positions,
+                )
 
         extra_kwargs["positions"] = positions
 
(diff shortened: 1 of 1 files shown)


torchtitan/components/validate.py
```diff
@@ -184,30 +184,31 @@ def post_dataloading_process(
...
+                positions = torch.arange(
+                    inputs.shape[1], dtype=torch.int32, device=inputs.device
+                ).repeat(inputs.shape[0], 1)
 
         try:
-            # pyrefly: ignore [not-callable]
             extra_kwargs["attention_masks"] = cast(
-                BaseModel, model_parts[0]
+                Decoder, model_parts[0]
             ).get_attention_masks(
```
wwwjn: Why we don't put get_attention_mask() under `if isinstance(model_config, Decoder.Config):`

Mirroring: https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/trainer.py#L611

wwwjn (commented):
Nice catch, thanks! Can you attach an actual run log to show grad_norm difference before v.s. after this change?

In long term, validate.py shares a lot of duplicate code with trainer.py

torchtitan/components/validate.py
francesco-bertolotti: Good point, fixing it now!

francesco-bertolotti:
Hi @wwwjn,

This bug does not affect training behavior in any way. The issue was only in validation reporting: the validation loss appeared roughly 4X (bc I am running with 4 ranks) higher than it actually was because `global_valid_tokens` was miscalculated. There is no impact on gradient norms or optimization.

Here are two comparison runs:

This is the validation loss before (`test`) and after (`fixed`) my commits: valid

This is the gradient norm before and after the commits (unchanged): grads

Configuration used:

```python
torchtitan.trainer.Trainer.Config(
    model_spec=torchtitan.models.llama3.model_registry("debugmodel", attn_backend="flex"),
    dataloader = torchtitan.hf_datasets.text_datasets.HuggingFaceTextDataLoader.Config(
        dataset = "c4_test",
        num_workers = 1,
    ),
    parallelism = torchtitan.trainer.ParallelismConfig(
        data_parallel_replicate_degree=1, 
        data_parallel_shard_degree=-1,
        context_parallel_degree=4,
        context_parallel_load_balancer=None,
    ),
    compile = torchtitan.trainer.CompileConfig(
        enable=False,
    ),
    optimizer = torchtitan.components.optimizer.OptimizersContainer.Config(
        lr=1e-4,
    ),
    training = torchtitan.config.TrainingConfig(
        local_batch_size=1,
        seq_len=512,
        steps=100,
        dtype="float32",
        mixed_precision_param="float32"
    ),
    loss = torchtitan.components.loss.CrossEntropyLoss.Config(),
    lr_scheduler=torchtitan.components.lr_scheduler.LRSchedulersContainer.Config(
        warmup_steps=10,
        decay_ratio=0,
        decay_type="linear",
        min_lr_factor=0.0,
    ),
    checkpoint=torchtitan.components.checkpoint.CheckpointManager.Config(
        interval=1000,
        last_save_model_only=False,
        export_dtype="bfloat16",
        enable=True,
    ),
    activation_checkpoint=torchtitan.config.ActivationCheckpointConfig(
        mode="selective",
    ),
    metrics=torchtitan.components.metrics.MetricsProcessor.Config(
        enable_tensorboard=True,
        enable_wandb=False,
        log_freq=1
    ),
    debug=torchtitan.config.configs.DebugConfig(
        seed=42,
        deterministic=True,
    ),
    validator=torchtitan.components.validate.Validator.Config(
        enable=True,
        freq=10,
        dataloader=torchtitan.hf_datasets.text_datasets.HuggingFaceTextDataLoader.Config(
            dataset = "c4_test",
            dataset_path = None,
            infinite = False,
        ),
        steps=3,
    ),
    hf_assets_path="./tests/assets/tokenizer",
    dump_folder=".data/experiments/test",
)
```

torchtitan/components/validate.py
```diff
@@ -184,29 +185,30 @@ def post_dataloading_process(
...
-        try:
-            # pyrefly: ignore [not-callable]
-            extra_kwargs["attention_masks"] = cast(
-                BaseModel, model_parts[0]
-            ).get_attention_masks(
-                positions=extra_kwargs.get("positions"),
-            )
-        except TypeError:
-            pass
+        if isinstance(model_config, Decoder.Config):
```
fegin, line 190: Is this a logic change or simply a typing improvement? I think it is the later but I want to confirm.

tianyu-l (approved):
oh, still an issue caused by validator not in sync with trainer.
In the long term, we should think about what refactor makes the most sense.
In the short term, we can add validator loss to be part of the CI loss guard. cc @wwwjn

torchtitan/components/validate.py
francesco-bertolotti: I am not sure either. I just aligned the code with the one from the trainer:
https://github.com/pytorch/torchtitan/blob/8599818f813ae43d73aeb142a879cbbff9b1d9ad/torchtitan/trainer.py#L594-L614

If this is a problem, I can change it/

fegin: That's fine. It is a typing fix. Thanks!

fegin (approved):

PR #3259, opened 2026-05-07
[Do not review][rl] Move RL H100 CI to A10G

wwwjn:
Follow up on the RL CI fix PR:  This is originally part of #3041 . 
-  Move H100 RL test to A10G, as FAv2 can also achieve batch invariant with paged kv cache now.


torchtitan/experiments/rl/config_registry.py
```diff
@@ -252,3 +252,54 @@ def rl_grpo_qwen3_0_6b_batch_invariant() -> RLTrainer.Config:
             debug=batch_invariant_config,
         ),
     )
+
+
+def rl_grpo_qwen3_debug_batch_invariant() -> RLTrainer.Config:
+    """Batch-invariant GRPO config using debug model with random init (4 GPUs: 2 gen + 2 train).
+
+    Uses the tiny debugmodel (dim=256, 8 layers, vocab=2048) with random
+    initialization — no HF checkpoint needed. Fits on A10G (22 GB) GPUs.
```
wwwjn: This is because when using batch-invariant kernels, the memory usage will be higher than non-batch-invariant, because these kernels has extra intermediate tensor allocation. 

But the cons for using random init is we don't know if the model are actually learning something.

tianyu-l: that's fine, it's not the job of this config

tianyu-l (approved):

wwwjn:
The CI failing is because vllm nightly build is still on May14, which haven't pickup latest pytorch fix from Anshul

torchtitan/experiments/rl/sum_digits.py
```diff
@@ -57,13 +60,13 @@ def __init__(self, config: Config, *, step: int = 0, group_idx: int = 0):
...
-                "correctness": self._correctness_reward(completion),
-                "format": self._format_reward(completion),
-            },
-            done=True,
-        )
+        rewards = {
+            "correctness": self._correctness_reward(completion),
+            "format": self._format_reward(completion),
+        }
+        if self._config.debug_random_reward > 0.0:
```
wwwjn: Update on this PR: Update some findings which is blocking us landing this PR:

1. Once we move H100 to A10G, we will got OOM if still use 0.6B model with TP=4 for both trainer(4GPUs) and generator(4GPUs). The OOM only happens when batch-invariant mode is on, because the batch-invariant kernel consume more GPU memory. 
2. If we use debug model to check batch invariant integration run, the RL loss will be all zero because the reward is always 0.0. This is because model generates non-sense generation and the rewards are always 0.0 so we are not testing anything, except for the first forward pass's bitwise parity. 

So I ends up adding this random , hash-based rewards for debugging model

PR #3262, opened 2026-05-07
Support kwargs in minimal_fx_tracer

tugsbayasgalan:
The traced fn can now take keyword arguments, mirroring torch.export's pattern: (args, kwargs) is bundled as a single pytree and the kwarg key list is stored on TracedResult so run_traced reorders runtime kwargs to trace-time order (with a clear error on key mismatch).

Authored with Claude.

### DIFF
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -7,7 +7,7 @@
 import copy
 from collections.abc import Callable, Generator
 from contextlib import contextmanager, nullcontext
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any
 
 import torch
@@ -272,6 +272,18 @@ def extract_module_state(mod: nn.Module) -> dict[str, torch.Tensor]:
     }
 
 
+def _check_no_extra_module_in_user_inputs(
+    args: tuple, kwargs: dict, fn_name: str
+) -> None:
+    """Reject extra nn.Module instances in user args/kwargs (args[0] is the
+    module-rooted entry point)."""
+    if any(isinstance(v, nn.Module) for v in (*args, *kwargs.values())):
+        raise ValueError(
+            f"{fn_name} supports exactly one nn.Module at args[0]. "
+            "Additional nn.Module instances found in args[1:] or kwargs."
+        )
+
+
 @dataclass
 class TracedResult:
     """Execution metadata returned by :func:`minimal_fx_tracer`.
@@ -279,23 +291,31 @@ class TracedResult:
     Attributes:
         gm: The traced FX graph as a pure function of flat tensors.
         example_inputs: Trace-time fake flat inputs used by downstream graph passes.
-        state_fqns: Trace-time state keys.
         num_flat_inputs: Number of flat graph inputs before subclass unwrapping.
         input_subclass_layouts: Subclass unwrap/rewrap metadata for inputs.
+        user_inputs_spec: Trace-time pytree spec for ``(args, kwargs)``.
         num_flat_outputs: Number of flat graph outputs before subclass rewrapping.
         output_subclass_layouts: Subclass unwrap/rewrap metadata for outputs.
         output_spec: Original output pytree spec used during reconstruction.
+        state_fqns: Trace-time state keys.
     """
 
     gm: torch.fx.GraphModule
+
+    # input related
     example_inputs: tuple[Any, ...]
-    state_fqns: list[str]
     num_flat_inputs: int
     input_subclass_layouts: dict[int, SubclassLayout]
+    user_inputs_spec: pytree.TreeSpec
+    tensor_input_indices: list[int]
+
+    # output related
     num_flat_outputs: int
     output_subclass_layouts: dict[int, SubclassLayout]
     output_spec: pytree.TreeSpec
-    tensor_input_indices: list[int] = field(default_factory=list)
+
+    # state related
+    state_fqns: list[str]
 
     @property
     def num_static_inputs(self) -> int:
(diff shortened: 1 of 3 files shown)


torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -503,7 +509,13 @@ def run_traced(
...
     """
     state_flat = list(state.values())
-    user_args_flat, _ = pytree.tree_flatten(list(args))
+    if set(kwargs) != set(traced_result.user_kwarg_keys):
+        raise ValueError(
+            f"kwarg key mismatch: got {list(kwargs)}, "
+            f"expected {list(traced_result.user_kwarg_keys)}"
+        )
+    kwargs = {k: kwargs[k] for k in traced_result.user_kwarg_keys}
+    user_args_flat, _ = pytree.tree_flatten((args, kwargs))
```
ydwu4: We could probably directly compare the stored in_spec vs in_spec at runtime so we could save the user_kwarg_keys in TracedResult.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -551,18 +569,23 @@ def run_traced_train_step(
...
+    of that name must be passed via the underlying ``fn``'s positional args.
+    """
 
     if not isinstance(module, nn.Module):
         raise ValueError(
             "run_traced_train_step requires args[0] to be an nn.Module, "
             f"got {type(module).__name__}."
         )
-    if any(isinstance(arg, nn.Module) for arg in args):
+    if any(isinstance(arg, nn.Module) for arg in (*args, *kwargs.values())):
```
ydwu4: Nit:the nn module check appears twice. seems can be extracted out as a helper.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
SherlockNoMad: +1, input_spec seems more general. 

also, I would consider runtime overhead. This is in the critical perf path.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
SherlockNoMad: +1

SherlockNoMad (approved):
you can land when you collect Yidi's stamp.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -501,9 +520,23 @@ def run_traced(
...
+
+    When ``_validate_runtime=True``, kwargs are validated and reordered against
+    the trace-time spec (raises on missing/extra keys). Off by default to keep
+    the per-step path overhead-free; the caller must pass kwargs in trace-time
+    order.
     """
     state_flat = list(state.values())
-    user_args_flat, _ = pytree.tree_flatten(list(args))
+    if _validate_runtime:
+        expected_kwarg_keys = traced_result.user_inputs_spec.child(1).context
```
ydwu4: ughh... i remember user_inputs_spec == user_args_spec will check the dictionary keys. We don't need to actually unroll the context i think?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -295,6 +307,7 @@ class TracedResult:
     num_flat_outputs: int
     output_subclass_layouts: dict[int, SubclassLayout]
     output_spec: pytree.TreeSpec
+    user_inputs_spec: pytree.TreeSpec
```
ydwu4: Can re-order the fields into the following in this PR?

```python
    gm: torch.fx.GraphModule

   # input related
    example_inputs: tuple[Any, ...]
    num_flat_inputs: int
    input_subclass_layouts: dict[int, SubclassLayout]
    user_inputs_spec: pytree.TreeSpec

    # output related
    num_flat_outputs: int
    output_subclass_layouts: dict[int, SubclassLayout]
    output_spec: pytree.TreeSpec


    # state related
    state_fqns: list[str]
```

Need a follow up PR: 

Can we do a full auditing of the fields in this dataclass because i'm not sure how these are related to each other and why they're useful from the documentation.

- example_inputs: should it actually be flat_example_inputs? Can they be accessed from ph.node.meta["val"]? 
- num_flat_inputs: how is this related to example_inputs?
- input_subclass_layout: is it on flat input?
- user_inputs_spec: pytree.TreeSpec

There seems to be a clear structure for both inputs and outputs: num, subclass_layouts, tree_spec. Might worth adding a level of abstraction.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
ydwu4: I see the problem. we could have valid call but with different pytree spec:

```python
fn(a, b, c)
fn(a, b, c=c)
```
will have different input_spec for fn but they're both valid.

To address this, i think we should do an args, kwargs binding/normalization first before pytree flatten.  In this way, if the binding fails that means the function is invoked in a way that's not supported.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: Yeah sounds good. For this PR, i will just reorder the fields.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -272,29 +272,50 @@ def extract_module_state(mod: nn.Module) -> dict[str, torch.Tensor]:
...
+
+    # output related
     num_flat_outputs: int
     output_subclass_layouts: dict[int, SubclassLayout]
     output_spec: pytree.TreeSpec
+
+    # state related
+    state_fqns: list[str]
+
     tensor_input_indices: list[int] = field(default_factory=list)
```
ydwu4: lol, what about this one! I didn't notice we have a tensor_input_indices.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
ydwu4, line 592: Doesn't need to happen in this PR. 


Just noticed that this might cause kwarg collision e.g imagine kwargs consists of validate_module_fqns.

The correct way of handling this is just like minimal_fx_tracer, return a callable instead.
```python
run_traced(traced_result, module, validate_module_fqns)(*args, **kwargs)
```
, in that case, might as well as change run_traced to be replay():
```python
replay(traced_result, module, validate_module_fqns)(*args, **kwargs)
```

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -486,6 +486,92 @@ def test_core_explicit_state_train_step(self):
...
+            with torch.nn.utils.stateless._reparametrize_module(model, state):
+                return model(tokens) * scale
+
+        state = extract_module_state(model)
+        traced = minimal_fx_tracer(forward)(state, tokens, scale=scale)
+        out_ref = forward(state, tokens, scale=scale)
+        out_traced = run_traced(traced, state, tokens, scale=scale)
+        self.assertTrue(torch.equal(out_ref, out_traced))
+
+    def test_kwargs_runtime_reorder_raises(self):
```
ydwu4, line 504: Add a TODO here so when upstream pytree dict key sorting issue is fixed, the author could be comfortable to removes this.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: Yeah hahaha. I will do this in https://github.com/pytorch/torchtitan/pull/3164/changes#diff-8fe284997e4eaba06224a6516c1222b5bc1c4e68dac9489de2bf91de4306b574

torchtitan/experiments/graph_trainer/make_fx_tracer.py
tugsbayasgalan: It is default. But yeah i moved it by not making it default. This seems precompile related.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
tugsbayasgalan: added in the make_fx_tracer.py

ydwu4 (approved):

PR #3263, opened 2026-05-07
[Checkpoint] BaseModel state dict pipeline and pure I/O CheckpointManager

mori360:
Move state dict logic into BaseModel:
- StateDictMode enum (FULL/TRAINABLE/BASE) for mode-based save/load
- get_sd(mode) returns filtered state dict via DCP StateDictOptions
- sd_adapter property and set_sd_adapter() for HF format transforms
- to_hf/from_hf/adapter_to_hf delegate to stored adapter

Refactor CheckpointManager:
- Remove sd_adapter constructor parameter; access via model.sd_adapter
- ModelWrapper uses part.get_sd(mode) instead of get_model_state_dict
- Add initial_load_frozen and last_save_trainable_only config fields
- Extract _load_initial, _load_from_dcp, _load_from_hf helpers
- _save_last_step supports TRAINABLE mode for adapter-only export

Update all callsites (trainer, experiments, scripts) to wire sd_adapter onto model before checkpoint build.

### DIFF
--- a/scripts/checkpoint_conversion/convert_from_hf.py
+++ b/scripts/checkpoint_conversion/convert_from_hf.py
@@ -11,7 +11,7 @@ from pathlib import Path
 import torch
 import torch.distributed.checkpoint as dcp
 from torch.distributed.checkpoint import HuggingFaceStorageReader
-from torchtitan.components.checkpoint import ModelWrapper
+from torch.distributed.checkpoint.state_dict import get_model_state_dict
 
 
 @torch.inference_mode()
@@ -23,14 +23,12 @@ def convert_from_hf(input_dir, output_dir, model_name, model_flavor):
 
     with torch.device("cpu"):
         model = model_config.build()
-    model = ModelWrapper(model)
-
     sd_adapter = model_spec.state_dict_adapter(model_config, None)
     assert (
         sd_adapter is not None
     ), "trying to convert checkpoint from HF to DCP safetensors format, but sd_adapter is not provided."
     # get state dict in tt format with allocated memory
-    state_dict = model._get_state_dict()
+    state_dict = get_model_state_dict(model)
     # convert empty state dict to hf format so that hf weights can be loaded into it
     hf_state_dict = sd_adapter.to_hf(state_dict)
     dcp.load(
--- a/scripts/checkpoint_conversion/convert_from_hf.py
+++ b/scripts/checkpoint_conversion/convert_from_hf.py
@@ -11,7 +11,7 @@ from pathlib import Path
 import torch
 import torch.distributed.checkpoint as dcp
 from torch.distributed.checkpoint import HuggingFaceStorageReader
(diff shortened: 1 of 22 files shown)


mori360:
This pr is a new one based from https://github.com/pytorch/torchtitan/pull/3188 to remove the dependency between lora and checkpoint.

> For the sake of PP, we would keep ModelWrapper, but it should be a thin wrapper to hold all model_parts. Each individual model part (BaseModel) should handle the state dict content (hf or not, trainable / frozen / all). In the future we may reconsider how we do PP (e.g. one model hosting all stages).
Right now there seem to be complicated logic in save / load (e.g. load twice if saved a lora adapter before, firstly base model, then lora adapter). I would advocate for making it work e2e with a simple solution, namely -- provide two new configs (initial_load_frozen_only, last_save_trainable_only), which can be combined with existing hf options to achieve the following workflow: (1) create a model with lora, load an HF base model (2) save full DCP checkpoints (base + lora) (3) load full DCP checkpoints (base + lora) (4) save lora adapters only, in HF.

@tianyu-l This pr addresses what we discussed before.

> If we are doing this, we shouldn't make checkpointer more complicated. Instead this first two steps should be an offline process/script that lives outside of the main trainer - it can be a script that creates a customized checkpointer to do the offline process. If we are doing online conversion, it doesn't make sense to have the first two steps in the main checkpointer.
Also avoid saving the frozen model was one request before because not every user has the luxury to have the storage to store two copies of the large models. I'm not sure if this is still the request.

@fegin 
The checkpointer itself is not that complicated — dcp_load and dcp_save are unchanged. The new logic is just a mode flag (StateDictMode) that filters which keys are included, using existing DCP StateDictOptions(ignore_frozen_params=True).

Steps 1-4 all run inline during training (not offline) because they're the natural workflow for adapter finetuning: load base -> train -> save. The existing initial_load_in_hf already handled step 1 on main — we only added the frozen filter on top.

For the storage concern: last_save_trainable_only=True saves only adapter weights at the final step. The interval DCP checkpoints do save everything (base + LoRA) for simplicity — this keeps the resume path straightforward (single dcp.load with mode=FULL). We can revisit interval-level trainable-only saves if storage becomes a blocker, but for now we keep it simple and assume users primarily need full DCP for resume and trainable-only HF for export.

fegin:
I think my point is fairly simple. Users generally don't expect frozen parameters to be saved multiple times during training just for fault tolerance purpose. The storage usage can be huge. But I agree the logic will be complicated. If we don't want to support this, just flag this as a limitation.

mori360:
> I think my point is fairly simple. Users generally don't expect frozen parameters to be saved multiple times during training for fault tolerance purpose. The storage usage can be huge. But I agree the logic will be complicated. If we don't to support this, just flag this as a limitation.

This PR is for checkpoint only, later in the lora pr we can mention it

PR #3264, opened 2026-05-07
[LoRA] Add LoRA converter and rename quantization to converters

mori360:
Add config-level LoRA converter infrastructure:
- LoRAConfig wraps any Linear.Config, builds LoRA subclass with frozen base and trainable lora_a/lora_b adapters
- FrozenConfig wraps any Module.Config, freezes all params at build time
- LoRAConverter walks model config tree, wraps targets with LoRAConfig and non-targets with FrozenConfig
- Dynamic LoRA class factory (_get_lora_cls) composes with quantization (Float8, MXFP8)

Rename model_registry parameter from 'quantization' to 'converters' across all models (llama3, llama4, deepseek_v3, qwen3, qwen3_vl, gpt_oss, flux) to support both quantization and LoRA converters.

Add validate_converter_order to ensure quantization precedes LoRA. Preserve requires_grad through DTensor distribution in Module.

### DIFF
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -588,10 +588,12 @@ def build_features_test_list() -> list[OverrideDefinitions]:
                 [
                     "--module llama3 --config llama3_debugmodel_lora",
                     "--parallelism.tensor_parallel_degree 2",
+                    "--parallelism.pipeline_parallel_degree 2",
                 ],
             ],
             "LoRA training test",
             "lora",
+            ngpu=8,
         ),
         OverrideDefinitions(
             [
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -587,6 +587,7 @@ def build_features_test_list() -> list[OverrideDefinitions]:
             [
                 [
                     "--module llama3 --config llama3_debugmodel_lora",
+                    "--parallelism.tensor_parallel_degree 2",
                 ],
             ],
             "LoRA training test",
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -586,6 +586,15 @@ def build_features_test_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
+                    "--module llama3 --config llama3_debugmodel_lora",
+                ],
+            ],
+            "LoRA training test",
+            "lora",
+        ),
+        OverrideDefinitions(
+            [
+                [
                     "--comm.mode torchcomms",
                     "--parallelism.context_parallel_degree 2",
                     "--parallelism.pipeline_parallel_degree 2",
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -60,7 +60,7 @@ def _get_lora_cls(parent_cls: type) -> type:
     if parent_cls in _lora_class_cache:
         return _lora_class_cache[parent_cls]
 
-    parent_config_cls = parent_cls.Config
+    parent_config_cls = parent_cls.Config  # pyrefly: ignore [missing-attribute]
 
     class LoRALinear(parent_cls):  # type: ignore[valid-type, misc]
         @dataclass(kw_only=True, slots=True)
@@ -182,7 +182,7 @@ class LoRAConverter(ModelConfigConverter):
         """Create a LoRALinear.Config from a base Linear.Config."""
         assert cfg._owner is not None
         lora_cls = _get_lora_cls(cfg._owner)
-        return lora_cls.Config(
+        return lora_cls.Config(  # pyrefly: ignore [missing-attribute]
             **{f.name: getattr(cfg, f.name) for f in fields(cfg)},
             rank=self.rank,
             alpha=self.alpha,
--- a/torchtitan/components/lora.py
+++ b/torchtitan/components/lora.py
@@ -54,29 +54,44 @@ def _get_lora_cls(parent_cls: type) -> type:
     """Get or create a LoRA subclass for *parent_cls* (e.g. Linear, Float8Linear).
 
     The returned class has a proper ``Config`` that extends the parent's Config
-    with LoRA adapter configs (``lora_a``, ``lora_b``) and ``lora_scaling``.
-    Adapters are built in ``__init__`` from their stored configs, following
-    the standard torchtitan Config → build pattern.
+    with ``rank`` and ``alpha``.  Adapters are built in ``__init__`` from the
+    base config's dimensions and sharding.
     """
     if parent_cls in _lora_class_cache:
         return _lora_class_cache[parent_cls]
 
-    parent_config_cls = parent_cls.Config  # pyrefly: ignore [missing-attribute]
+    parent_config_cls = parent_cls.Config
 
     class LoRALinear(parent_cls):  # type: ignore[valid-type, misc]
         @dataclass(kw_only=True, slots=True)
         class Config(parent_config_cls):  # type: ignore[misc]
-            lora_a: Linear.Config
-            lora_b: Linear.Config
-            lora_scaling: float
+            rank: int
+            alpha: float
 
         def __init__(self, config: Config) -> None:
             super().__init__(config)
             for param in nn.Module.parameters(self):
                 param.requires_grad_(False)
-            self._lora_scaling = config.lora_scaling
-            self.lora_a = config.lora_a.build()
-            self.lora_b = config.lora_b.build()
+            self._lora_scaling = config.alpha / config.rank
+            lora_a_sharding, lora_b_sharding = _lora_adapter_sharding(
+                config.sharding_config
+            )
+            self.lora_a = Linear.Config(
+                in_features=config.in_features,
+                out_features=config.rank,
+                bias=False,
+                sharding_config=lora_a_sharding,
+                param_init={
+                    "weight": lambda w: nn.init.kaiming_uniform_(w, a=math.sqrt(5)),
+                },
+            ).build()
+            self.lora_b = Linear.Config(
+                in_features=config.rank,
+                out_features=config.out_features,
+                bias=False,
+                sharding_config=lora_b_sharding,
+                param_init={"weight": nn.init.zeros_},
+            ).build()
 
         def forward(self, input: torch.Tensor) -> torch.Tensor:
             base_out = super().forward(input)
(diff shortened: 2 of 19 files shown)


mori360:
This pr comes from https://github.com/pytorch/torchtitan/pull/3189, without dependency on the checkpoint part.

`validate_converter_order` is moved to torchtitan/models/utils.py and is validated before .build().

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            setattr(self.inner, name, value)
+
+    def build(self, **kwargs):
+        instance = self.inner.build(**kwargs)
+        instance.requires_grad_(False)
+        return instance
+
+
+class LoRAConverter(Configurable):
```
tianyu-l: This should inherit `ModelConfigConverter` -- let's rename `QuantizationConverter` https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/__init__.py#L35 and put it in protocols/model.py so it can be inherited by both lora and fp8/mxfp8 converters.

torchtitan/models/llama3/config_registry.py
```diff
@@ -97,13 +97,28 @@ def llama3_debugmodel_float8() -> Trainer.Config:
...
         "debugmodel",
-        quantization=[
+        converters=[
             Float8LinearConverter.Config(model_compile_enabled=model_compile_enabled),
         ],
     )
     return config
 
 
+def llama3_debugmodel_lora() -> Trainer.Config:
```
tianyu-l, line 107: please add a test in CI features.py

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+            return base_out + self._lora_scaling * lora_out
+
+    LoRALinear.__name__ = f"LoRA{parent_cls.__name__}"
+    LoRALinear.__qualname__ = f"LoRA{parent_cls.__name__}"
+    _lora_class_cache[parent_cls] = LoRALinear
+    return LoRALinear
+
+
+@dataclass(kw_only=True, slots=True)
+class LoRAConfig(Configurable.Config):
```
tianyu-l: Can we put this inside `LoRALinear` (to be `LoRALinear.Config`)?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+            raise AttributeError(name) from None
+        return getattr(inner, name)
+
+    def __setattr__(self, name: str, value) -> None:
+        if name in type(self).__slots__:
+            object.__setattr__(self, name, value)
+        else:
+            setattr(self.inner, name, value)
+
+    def build(self, **kwargs):
```
tianyu-l: In particular, we shouldn't need to define this `build` function.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,276 @@
...
+            else:
+                wrapped = FrozenConfig(inner=cfg)
+
+            if isinstance(parent, list):
+                parent[attr] = wrapped
+            else:
+                setattr(parent, attr, wrapped)
+
+        # Freeze non-Linear leaf modules (norms, embeddings)
+        for config_cls in (RMSNorm.Config, Embedding.Config):
```
tianyu-l: https://github.com/pytorch/torchtitan/pull/3189#discussion_r3186696257 not addressed
I don't think we should special cases here.

tianyu-l (changes requested):

torchtitan/models/llama3/__init__.py
```diff
@@ -377,12 +378,13 @@ def _405b(attn_backend: str) -> Llama3Model.Config:
 def model_registry(
     flavor: str,
     attn_backend: str = "sdpa",
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```
here and in other files

torchtitan/components/lora.py
```diff
@@ -0,0 +1,223 @@
...
+                    "weight": lambda w: nn.init.kaiming_uniform_(w, a=math.sqrt(5)),
+                },
+            ),
+            lora_b=Linear.Config(
+                in_features=self.rank,
+                out_features=cfg.out_features,
+                bias=False,
+                sharding_config=lora_b_sharding,
+                param_init={"weight": nn.init.zeros_},
+            ),
```
tianyu-l: let's put these in `LoRALinear.__init__()`. We just need to pass basic configs to it, such as those defined at 
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/linear.py#L26-L28
and alpha, rank

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -534,16 +535,17 @@ def model_registry(
     attn_backend: str = "sdpa",
     moe_comm_backend: str = "standard",
     non_blocking_capacity_factor: float | None = None,
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/models/flux/__init__.py
```diff
@@ -545,12 +546,13 @@ def _flux_debug() -> FluxModel.Config:
 
 def model_registry(
     flavor: str,
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -347,14 +348,15 @@ def _120b(
 def model_registry(
     flavor: str,
     moe_comm_backend: str = "standard",
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/models/llama3/__init__.py
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/models/llama4/__init__.py
```diff
@@ -349,15 +350,16 @@ def model_registry(
     flavor: str,
     attn_backend: str = "flex",
     moe_comm_backend: str = "standard",
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/models/qwen3/__init__.py
```diff
@@ -619,15 +620,16 @@ def model_registry(
     flavor: str,
     attn_backend: str = "sdpa",
     moe_comm_backend: str | None = None,
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/models/qwen3_vl/__init__.py
```diff
@@ -574,15 +575,16 @@ def model_registry(
     flavor: str,
     attn_backend: str = "sdpa",
     moe_comm_backend: str | None = None,
-    quantization: list[QuantizationConverter.Config] | None = None,
+    converters: list[Configurable.Config] | None = None,
```
tianyu-l:
```suggestion
    converters: list[ModelConfigConverter.Config] | None = None,
```

torchtitan/protocols/model.py
```diff
@@ -7,9 +7,22 @@
 from abc import abstractmethod
 from dataclasses import dataclass
 
+from torchtitan.config import Configurable
+
 from .module import Module
 
 
+class ModelConfigConverter(Configurable):
```
tianyu-l, line 15: should define its own Config class

tests/integration_tests/features.py
```diff
@@ -583,6 +583,16 @@ def build_features_test_list() -> list[OverrideDefinitions]:
             "Float8 emulation test",
             "float8_emulation",
         ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_lora",
+                    "--parallelism.tensor_parallel_degree 2",
+                ],
```
tianyu-l, line 592: only need 4 gpu?

torchtitan/components/lora.py
```diff
@@ -0,0 +1,222 @@
...
+        """Module names to apply LoRA to (matched against the last segment of the FQN).
+        None means all Linear layers. An empty list means no layers."""
+
+    def __init__(self, config: Config, **kwargs):
+        if config.rank <= 0:
+            raise ValueError(f"LoRA rank must be positive, got {config.rank}")
+        self.config = config
+        self.rank = config.rank
+        self.alpha = config.alpha
+        self.target_modules = (
```
tianyu-l, line 167: do we know that this is a required feature by most people? If not I'd recommend we start simple, without this selective support.

torchtitan/components/lora.py
```diff
@@ -0,0 +1,222 @@
...
+            alpha: float
+
+        def __init__(self, config: Config) -> None:
+            super().__init__(config)
+            for param in nn.Module.parameters(self):
+                param.requires_grad_(False)
+            self._lora_scaling = config.alpha / config.rank
+            lora_a_sharding, lora_b_sharding = _lora_adapter_sharding(
+                config.sharding_config
+            )
```
tianyu-l, line 78: this sounds problematic:
Current flow is
- at config time we call `model_registry` in which `config.sharding_config` is not set
- at run time, in `update_from_config` calls we set `config.sharding_config`
- however, here you are referring to `config.sharding_config` before it's usually set, so should always get None results. Have you tested that it works in real training job?

I think a workaround might be hacking the setter function of `LoRALinear.Config` (similar to what you did for `FrozenConfig`) so that when sharding_config is set for the entire module e.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder_sharding.py#L192, it also updates the a/b sharding config accordingly.

tianyu-l (changes requested):

torchtitan/components/lora.py
mori360: It's common feature supported by most LoRAs. In moe-lora, gate is excluded from lora, and in rl-lora, target selective layers to control parameter budget and avoid reward hacking

torchtitan/components/lora.py
mori360: `sharding_config` is not set at config time and gets set in update_from_config. But `_lora_adapter_sharding(config.sharding_config)` is called in `LoRALinear.__init__`, which runs at build time              
, after `update_from_config has already` set the sharding. The converter only swaps configs in the tree — it never calls build(). 

The current pr pass the dp=2 tp=2 feature test, and also verify by adding a debug log.
```
  [rank0]: LoRA training active with rank=8, alpha=16.0, target_modules=['wkv', 'wo', 'wq']   ← config time (model_registry)                                                                                                       
  ...                                                                                                                    
  [rank0]: Building llama3 debugmodel                                                          ← build time starts       
  ...                                                                                                                    
  [rank0]: LoRALinear.__init__: sharding_config=ShardingConfig(...)                            ← __init__ runs here, during build 
```

torchtitan/models/llama3/config_registry.py
```diff
@@ -97,13 +97,28 @@ def llama3_debugmodel_float8() -> Trainer.Config:
...
+    from torchtitan.components.lora import LoRAConverter
+
+    config = llama3_debugmodel()
+    config.model_spec = model_registry(
+        "debugmodel",
+        converters=[
```
tianyu-l, line 113: instead of creating a new one, could you modify https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/config_registry.py#L116 to compose lora and float8, and test this one with parallelisms in CI?

torchtitan/components/lora.py
tianyu-l: makes sense, thanks!

tianyu-l (approved):
one last nit on testing

PR #3269, opened 2026-05-07
[optimizer] support mixed optimizers

shuhuayu:
Summary:
1. Add mixed optimizer support to `OptimizersContainer`: different parameter groups can now use different optimizers (e.g., SGD for MoE router gates, AdamW for everything else)
2. `ParamGroupConfig` redesigned with two override modes: same optimizer with kwargs overrides, or switch to a different optimizer entirely via optimizer_name + optimizer_kwargs                                                                                       
3. Move Adam-specific fields (beta1, beta2, eps, weight_decay) from default `OptimizersContainer.Config` into `optimizer_kwargs` dict for optimizer-agnostic config                                                                                                            
4. Adapt LR scheduler checkpointing for mixed optimizers: save only last_epoch (plus full state for observability), restore by recomputing lr from each scheduler's own `base_lr`s. Correct for mixed optimizers and resharding                                                                
5. Add per-optimizer-group LR logging to TensorBoard/WandB (e.g., lr/AdamW_default, lr/SGD_.router.gate.)
6. Add init-time optimizer summary log showing param counts, labels, and key kwargs per group                                         
7. Add `_validate_params` to verify every trainable parameter is assigned to exactly one optimizer                                      
8. Add `qwen3_debugmodel_moe_param_groups` config as a runnable example with AdamW + SGD

### DIFF
--- a/torchtitan/components/lr_scheduler.py
+++ b/torchtitan/components/lr_scheduler.py
@@ -38,8 +38,7 @@ class LRSchedulersContainer(Stateful, Configurable):
 
     **Checkpoint behavior**
     All schedulers share the same lambda and step together. On save, only
-    ``last_epoch`` is needed for restore; full per-scheduler state is saved
-    alongside for observability. On load, ``last_epoch`` is restored and each
+    ``last_epoch`` is saved. On load, ``last_epoch`` is restored and each
     scheduler recomputes its lr from its own optimizer's ``base_lrs``. This
     handles mixed optimizers (different base lrs) and resharding (different
     number of schedulers between save and load).
@@ -211,25 +210,24 @@ class LRSchedulersContainer(Stateful, Configurable):
 
     def get_metrics(self) -> dict[str, float]:
         """Return per-param-group learning rates keyed for logging."""
-        lr_metrics = {}
+        metrics = {}
         for scheduler in self.schedulers:
             opt = scheduler.optimizer
             opt_name = type(opt).__name__
             for group, lr_val in zip(opt.param_groups, scheduler.get_last_lr()):
-                lr_metrics[f"lr/{opt_name}_{group['_label']}"] = float(lr_val)
-        return lr_metrics
+                metrics[f"lr/{opt_name}_{group['pattern']}"] = float(lr_val)
+        return metrics
 
     def step(self) -> None:
         for scheduler in self.schedulers:
             scheduler.step()
 
     def state_dict(self) -> dict[str, Any]:
-        # Save last_epoch (the step counter, shared across all schedulers) for
-        # loading, plus full per-scheduler state for observability/debugging.
-        return {
-            "last_epoch": self.schedulers[0].last_epoch,
-            "schedulers": [s.state_dict() for s in self.schedulers],
-        }
+        # Only last_epoch is needed — each scheduler recomputes its lr from
+        # its own optimizer's base_lrs on load. Per-scheduler state (base_lrs,
+        # _last_lr) is not saved because it's reconstructed from the optimizer
+        # config at construction time.
+        return {"last_epoch": self.schedulers[0].last_epoch}
 
     def load_state_dict(self, state_dict: dict[str, Any]) -> None:
         # Only restore last_epoch. Each scheduler recomputes _last_lr from its
--- a/torchtitan/components/lr_scheduler.py
+++ b/torchtitan/components/lr_scheduler.py
@@ -209,7 +209,7 @@ class LRSchedulersContainer(Stateful, Configurable):
     def __len__(self) -> int:
         return len(self.schedulers)
 
-    def get_lr_metrics(self) -> dict[str, float]:
+    def get_metrics(self) -> dict[str, float]:
         """Return per-param-group learning rates keyed for logging."""
         lr_metrics = {}
         for scheduler in self.schedulers:
@@ -236,6 +236,11 @@ class LRSchedulersContainer(Stateful, Configurable):
         # own optimizer's base_lrs and the shared lambda. This is correct for
         # mixed optimizers (different base_lrs) and resharding (different number
         # of schedulers between save and load).
+        #
+        # NOTE: torchtitan's LR schedules are stateless functions
+        # of (last_epoch, base_lr) — LambdaLR with a pure lambda. If a stateful
+        # scheduler (e.g. ReduceLROnPlateau) is added, this method must be updated
+        # to restore additional state.
         last_epoch = state_dict["last_epoch"]
         for scheduler in self.schedulers:
             scheduler.last_epoch = last_epoch
--- a/torchtitan/components/lr_scheduler.py
+++ b/torchtitan/components/lr_scheduler.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import copy
 import functools
 import math
 from collections.abc import Callable, Iterator
@@ -37,11 +36,13 @@ class LRSchedulersContainer(Stateful, Configurable):
     signature as ``torch.optim.lr_scheduler.LRScheduler`` class: ``step()``, ``state_dict()``,
     ``load_state_dict()``.
 
-    **Limitations**
-    This class assumes all the lr schedulers are the same. There is no easy way to support
-    resharding for multiple different LRSchedulers because LRScheduler.state_dict() is not
-    resharding friendly. Therefore, the limitation is used to allow TorchTitan to support
-    lr scheduler resharding.
+    **Checkpoint behavior**
+    All schedulers share the same lambda and step together. On save, only
+    ``last_epoch`` is needed for restore; full per-scheduler state is saved
+    alongside for observability. On load, ``last_epoch`` is restored and each
+    scheduler recomputes its lr from its own optimizer's ``base_lrs``. This
+    handles mixed optimizers (different base lrs) and resharding (different
+    number of schedulers between save and load).
 
     Args:
         optimizers (OptimizersContainer): The corresponding optimizers for the lr_schedulers.
@@ -208,21 +209,35 @@ class LRSchedulersContainer(Stateful, Configurable):
     def __len__(self) -> int:
         return len(self.schedulers)
 
+    def get_lr_metrics(self) -> dict[str, float]:
+        """Return per-param-group learning rates keyed for logging."""
+        lr_metrics = {}
+        for scheduler in self.schedulers:
+            opt = scheduler.optimizer
+            opt_name = type(opt).__name__
+            for group, lr_val in zip(opt.param_groups, scheduler.get_last_lr()):
+                lr_metrics[f"lr/{opt_name}_{group['_label']}"] = float(lr_val)
+        return lr_metrics
+
     def step(self) -> None:
         for scheduler in self.schedulers:
             scheduler.step()
 
     def state_dict(self) -> dict[str, Any]:
-        # While there may be multiple schedulers, we only save the first one because
-        # the state_dict is the same for all. See the limitations section in the
-        # docstring.
-        return self.schedulers[0].state_dict()
+        # Save last_epoch (the step counter, shared across all schedulers) for
+        # loading, plus full per-scheduler state for observability/debugging.
+        return {
+            "last_epoch": self.schedulers[0].last_epoch,
+            "schedulers": [s.state_dict() for s in self.schedulers],
+        }
 
     def load_state_dict(self, state_dict: dict[str, Any]) -> None:
-        # Load the same state_dict for all schedulers. The key value we're concerned
-        # within ``LRScheduler.state_dict()`` is ``last_epoch``, which is an integer
-        # that is immutable. As long as ``training.steps`` and ``lr_scheduler.warmup_steps``
-        # in the config remain unchanged when resuming from a checkpoint, this
-        # approach is safe. We call ``copy()`` here to ensure extra safety.
+        # Only restore last_epoch. Each scheduler recomputes _last_lr from its
+        # own optimizer's base_lrs and the shared lambda. This is correct for
+        # mixed optimizers (different base_lrs) and resharding (different number
+        # of schedulers between save and load).
+        last_epoch = state_dict["last_epoch"]
         for scheduler in self.schedulers:
-            scheduler.load_state_dict(copy.deepcopy(state_dict))
+            scheduler.last_epoch = last_epoch
+            scheduler._step_count = last_epoch + 1
+            scheduler._last_lr = scheduler.get_lr()
--- a/torchtitan/components/optimizer.py
+++ b/torchtitan/components/optimizer.py
@@ -67,28 +67,6 @@ class ParamGroupConfig:
     Must include all required kwargs (e.g. ``lr``). No implicit defaults."""
 
 
-def default_adamw(lr: float = 8e-4, **kwargs: Any) -> list[ParamGroupConfig]:
-    """Create a catch-all AdamW param group with standard defaults.
-
-    Use as a convenience for the common case::
-
-        OptimizersContainer.Config(param_groups=default_adamw(lr=3e-4))
-    """
-    return [
-        ParamGroupConfig(
-            pattern=r".*",
-            optimizer_name="AdamW",
-            optimizer_kwargs={
-                "lr": lr,
-                "betas": (0.9, 0.95),
-                "eps": 1e-8,
-                "weight_decay": 0.1,
-                **kwargs,
-            },
-        )
-    ]
-
-
 T = TypeVar("T", bound=Optimizer)
 
 
@@ -117,13 +95,18 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
 
     @dataclass(kw_only=True, slots=True)
     class Config(Configurable.Config):
+        param_groups: list[ParamGroupConfig] = field(default_factory=list)
+        """Per-parameter-group optimizer configurations. Each entry specifies a
+        regex pattern and a self-contained optimizer setup.
+        Patterns are checked in order; first match wins."""
+
         implementation: Literal[
             "for-loop", "foreach", "fused", "fused_opt_states_bf16"
         ] = "fused"
         """
         Optimizer implementation mode applied to all optimizer instances.
         Per-param-group ``optimizer_kwargs`` can override this (e.g.
-        ``"fused": False`` for optimizers that don't support fused with DTensor).
+        ``"fused": False`` for optimizers that don't support fused).
 
         - 'fused': Use fused implementation (CUDA only) for best performance.
         - 'foreach': Use some horizontal fusion of tensors for better performance.
@@ -135,12 +118,6 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
           OptimizersInBackwardContainer). See docs/bf16_optimizer_states.md.
         - more info: https://pytorch.org/docs/stable/optim.html
         """
-
-        param_groups: list[ParamGroupConfig] = field(default_factory=default_adamw)
-        """Per-parameter-group optimizer configurations. Each entry specifies a
-        regex pattern and a self-contained optimizer setup.
-        Patterns are checked in order; first match wins.
-        Defaults to ``default_adamw()``."""
 
     optimizers: list[T]
     model_parts: list[nn.Module]
@@ -173,7 +150,7 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
     @staticmethod
     def _build_param_groups(
         model: nn.Module,
-        param_groups: list[ParamGroupConfig],
+        param_group_configs: list[ParamGroupConfig],
         impl_kwargs: dict[str, Any],
     ) -> dict[str, list[dict[str, Any]]]:
         """Build PyTorch param groups from model parameters, partitioned by optimizer.
@@ -181,13 +158,13 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
         Each parameter is assigned to the first matching ParamGroupConfig pattern.
         Returns a dict mapping optimizer name to a list of param group dicts.
 
-        Each param group dict includes a ``_label`` key with the regex pattern
+        Each param group dict includes a ``pattern`` key with the regex pattern
         string for logging.
         """
         result: dict[str, list[dict[str, Any]]] = defaultdict(list)
         claimed: set[str] = set()  # first-match-wins
 
-        for pg in param_groups:
+        for pg in param_group_configs:
             pattern = re.compile(pg.pattern)
             matched = []
             for name, param in model.named_parameters():
@@ -205,17 +182,17 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
             result[pg.optimizer_name].append(
                 {
                     "params": matched,
-                    "_label": pg.pattern,
+                    "pattern": pg.pattern,
                     **impl_kwargs,
                     **pg.optimizer_kwargs,
                 }
             )
 
-        return dict(result)
+        return result
 
     def __init__(self, config: Config, *, model_parts: list[nn.Module]) -> None:
         impl_kwargs = self._build_impl_kwargs(config)
-        param_groups = config.param_groups
+        param_group_configs = config.param_groups
         all_params = []
         self.optimizers = []
         self.model_parts = model_parts
@@ -224,7 +201,7 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
 
         for part_idx, model in enumerate(self.model_parts):
             groups_by_opt_name = self._build_param_groups(
-                model, param_groups, impl_kwargs
+                model, param_group_configs, impl_kwargs
             )
             for opt_name, opt_param_groups in groups_by_opt_name.items():
                 opt_cls = self._resolve_optimizer_cls(opt_name)
@@ -258,10 +235,10 @@ class OptimizersContainer(Optimizer, Stateful, Configurable, Generic[T]):
             for group in opt.param_groups:
                 num_params = len(group["params"])
                 kwargs = {k: v for k, v in group.items() if k in _KEY_KWARGS}
-                label = group["_label"]
+                pattern = group["pattern"]
                 logger.info(
                     f"Optimizer {opt_name} (model_part={part_idx}): "
-                    f"{num_params} params [{label}] {kwargs}"
+                    f"{num_params} params [{pattern}] {kwargs}"
                 )
 
     def _validate_params(self, all_params: list[nn.Parameter]) -> None:
@@ -393,7 +370,7 @@ class OptimizersInBackwardContainer(OptimizersContainer):
 
     def __init__(self, config: Config, *, model_parts: list[nn.Module]) -> None:
         impl_kwargs = self._build_impl_kwargs(config)
-        param_groups = config.param_groups
+        param_group_configs = config.param_groups
         all_params = []
         self.model_parts = model_parts
         # Maps each optimizer to its model part (for state_dict/load_state_dict)
@@ -402,18 +379,18 @@ class OptimizersInBackwardContainer(OptimizersContainer):
         optim_dict: dict[nn.Parameter, Optimizer] = {}
         for part_idx, model in enumerate(self.model_parts):
             groups_by_opt_name = self._build_param_groups(
-                model, param_groups, impl_kwargs
+                model, param_group_configs, impl_kwargs
             )
             for opt_name, opt_param_groups in groups_by_opt_name.items():
                 opt_cls = self._resolve_optimizer_cls(opt_name)
                 for group in opt_param_groups:
-                    label = group["_label"]
+                    pattern = group["pattern"]
                     kwargs = {
-                        k: v for k, v in group.items() if k not in ("params", "_label")
+                        k: v for k, v in group.items() if k not in ("params", "pattern")
                     }
                     for p in group["params"]:
                         optim_dict[p] = opt_cls(
-                            [{"params": [p], "_label": label, **kwargs}]
+                            [{"params": [p], "pattern": pattern, **kwargs}]
                         )
                         self._model_part_indices.append(part_idx)
             all_params.extend(p for p in model.parameters())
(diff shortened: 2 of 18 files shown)


torchtitan/components/optimizer.py
```diff
@@ -39,80 +39,87 @@
...
-    derived by multiplying the global optimizer values with the specified multipliers.
+    Each entry specifies a regex pattern matching parameter FQNs. Parameters
+    matching the pattern can override optimizer hyperparameters or switch to a
+    different optimizer entirely.
+
+    There are two override modes:
+
+    1. **Same optimizer, different hyperparams** (``optimizer_name`` is None):
+       ``optimizer_kwargs`` are merged on top of the global optimizer defaults.
+       Unspecified kwargs inherit from the global config.
```
fegin: Now that we support different optimizers, can we internally just merge the two logic? We already create a separate optimizer object regardless it is mode 1 (different parameter groups) or mode 2 (different optimizer classes). If that will make the optimizer container logic simpler, we should go with this route. I don't see obvious downside of creating multiple optimizer objects.

shuhuayu: No, we do not create separate optimizer objects if it is mode 1. Pytorch natively supports it if there are several `param_groups` with their own kwargs, see below. 
```
for opt_name, param_groups in groups_by_opt.items():
    opt_cls = self._resolve_optimizer_cls(opt_name)
    self.optimizers.append(opt_cls(param_groups))
```
But I agree that we can simplify the logic. We can remove the default optimizer, and config all optimizer setups use `param_groups`, which is basically using mode 2 for all parameters. What do you think?

fegin: I know PyTorch natively supports mode 1.

My point is that if we always use mode 2 to implement the logic regardless it is mode 1 or mode 2, will it be simpler? 

We can assume there is only one parameter group per optimizer.. And it sounds like your response align with the point:`We can remove the default optimizer, and config all optimizer setups use param_groups, which is basically using mode 2 for all parameters. What do you think?`. But I will leave it to you to decide.

shuhuayu: Yeah, we should go with full mode 2 to make it simpler, thanks. Also synced with Tianyu offline.

shuhuayu:
Linearized the parameter group optimizer configs for mixed optimizer case as discussed. Now support two parallel entry points, same internal path: 
- Base case. Set `name`, `lr`, `optimizer_kwargs` on `Config` for all parameters. A catch-all param group is generated automatically. No `param_groups` needed. `OptimizersContainer.Config(lr=8e-4)`set lr for default AdamW optimizer.  `OptimizersContainer.Config(name="SGD", lr=1e-2, optimizer_kwargs={"momentum": 0.9})` uses SGD for parameters and set its parameters.
- Per param group case. Specify `param_groups` explicitly for per-parameter-group optimizer control. Config's `name`, `lr`, `optimizer_kwargs` are ignored. `OptimizersContainer.Config(param_groups=[ ParamGroupConfig(pattern=r"\.router\.", optimizer_name="SGD", optimizer_kwargs={"lr": 1e-4, ...}), ParamGroupConfig(pattern=r".*", optimizer_name="AdamW", optimizer_kwargs={"lr": 8e-4, ...}), ])` 

The base case is kept for code simplicity for the case of one optimizer for all. Both cases produce a `list[ParamGroupConfig]` that feeds into the same linearized `_build_param_groups` code path.

As an exampled in `torchtitan/models/qwen3/config_registry.py`, the wandb log shows: 
image

shuhuayu:
side comment: ci seemed to be including unchanged files in current pr into lint checks.

torchtitan/components/lr_scheduler.py
```diff
@@ -208,21 +209,35 @@ def __iter__(self) -> Iterator[LRScheduler]:
     def __len__(self) -> int:
         return len(self.schedulers)
 
+    def get_lr_metrics(self) -> dict[str, float]:
```
tianyu-l:
```suggestion
    def get_metrics(self) -> dict[str, float]:
```

torchtitan/components/lr_scheduler.py
```diff
@@ -208,21 +209,35 @@ def __iter__(self) -> Iterator[LRScheduler]:
...
+        # Only restore last_epoch. Each scheduler recomputes _last_lr from its
+        # own optimizer's base_lrs and the shared lambda. This is correct for
+        # mixed optimizers (different base_lrs) and resharding (different number
+        # of schedulers between save and load).
+        last_epoch = state_dict["last_epoch"]
         for scheduler in self.schedulers:
-            scheduler.load_state_dict(copy.deepcopy(state_dict))
+            scheduler.last_epoch = last_epoch
+            scheduler._step_count = last_epoch + 1
+            scheduler._last_lr = scheduler.get_lr()
```
tianyu-l, line 246: I feel this has too much assumptions on the LR scheduler internal, e.g. last_epoch is enough to determine the current lr. This sounds fine if torchtitan lr scheduler is simple enough, but let's document that?

torchtitan/components/lr_scheduler.py
```diff
@@ -208,21 +209,35 @@ def __iter__(self) -> Iterator[LRScheduler]:
...
-        # that is immutable. As long as ``training.steps`` and ``lr_scheduler.warmup_steps``
-        # in the config remain unchanged when resuming from a checkpoint, this
-        # approach is safe. We call ``copy()`` here to ensure extra safety.
+        # Only restore last_epoch. Each scheduler recomputes _last_lr from its
+        # own optimizer's base_lrs and the shared lambda. This is correct for
+        # mixed optimizers (different base_lrs) and resharding (different number
+        # of schedulers between save and load).
+        last_epoch = state_dict["last_epoch"]
         for scheduler in self.schedulers:
-            scheduler.load_state_dict(copy.deepcopy(state_dict))
```
tianyu-l, line 228: Not sure if it's at least worth it to still call this first, then do the overrides you are doing.

torchtitan/components/lr_scheduler.py
```diff
@@ -208,21 +209,35 @@ def __iter__(self) -> Iterator[LRScheduler]:
...
     def state_dict(self) -> dict[str, Any]:
-        # While there may be multiple schedulers, we only save the first one because
-        # the state_dict is the same for all. See the limitations section in the
-        # docstring.
-        return self.schedulers[0].state_dict()
+        # Save last_epoch (the step counter, shared across all schedulers) for
+        # loading, plus full per-scheduler state for observability/debugging.
+        return {
+            "last_epoch": self.schedulers[0].last_epoch,
+            "schedulers": [s.state_dict() for s in self.schedulers],
```
tianyu-l: this seems not used

torchtitan/components/optimizer.py
```diff
@@ -37,82 +37,87 @@
...
         model_parts (List[nn.Module]): List of model parts to be optimized.
-        optimizer_kwargs (Dict[str, Any]): Keyword arguments for the optimizers.
-        name (str): Name of the optimizers.
     """
 
     @dataclass(kw_only=True, slots=True)
     class Config(Configurable.Config):
         name: str = "AdamW"
-        """Optimizer to use"""
+        """Default optimizer type, used when param_groups is not specified."""
```
tianyu-l: I personally dislike having such branches. I don't think it's terrible to always let users specify at least one param group.

torchtitan/components/optimizer.py
```diff
@@ -125,31 +130,28 @@ class Config(Configurable.Config):
...
     optimizers: list[T]
     model_parts: list[nn.Module]
 
     @staticmethod
     def _resolve_optimizer_cls(name: str) -> type:
-        optimizer_classes = {"Adam": torch.optim.Adam, "AdamW": torch.optim.AdamW}
+        optimizer_classes = {
+            "Adam": torch.optim.Adam,
+            "AdamW": torch.optim.AdamW,
+            "SGD": torch.optim.SGD,
```
tianyu-l: why adding this?

torchtitan/components/optimizer.py
```diff
@@ -158,92 +160,127 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-        for model in self.model_parts:
-            param_groups = self._build_param_groups(model, config, optimizer_kwargs)
-            self.optimizers.append(optimizer_cls(param_groups))
-            for group in param_groups:
-                all_params.extend(group["params"])
+        # Maps each optimizer to its model part (for state_dict/load_state_dict)
+        self._model_part_indices: list[int] = []
+
+        for part_idx, model in enumerate(self.model_parts):
+            groups_by_opt = self._build_param_groups(model, param_groups, impl_kwargs)
```
tianyu-l: `groups_by_opt` sounds a bad name -- different groups can have the same optimizer cls.

torchtitan/components/optimizer.py
```diff
@@ -158,92 +160,127 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
+                for group in opt_param_groups:
+                    all_params.extend(group["params"])
+
+        self._validate_params(all_params)
+
         if config.implementation == "fused_opt_states_bf16":
             self._register_bf16_optimizer_state_hook()
-        self._validate_length(len(self.model_parts))
-        self._post_init(all_params, optimizer_kwargs)
+        self._post_init(all_params, impl_kwargs)
```
tianyu-l: this doesn't sound right, you are losing all kwargs but the fused / foreach ones

torchtitan/components/optimizer.py
```diff
@@ -158,92 +160,127 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-            group_kwargs = {**default_kwargs}
-            group_kwargs["lr"] = default_kwargs["lr"] * pg.lr_multiplier
-            group_kwargs["weight_decay"] = (
-                default_kwargs["weight_decay"] * pg.weight_decay_multiplier
+
+            label = re.sub(r"[^a-zA-Z0-9._]", "", pg.pattern.replace("|", "_or_"))
+            result[pg.optimizer_name].append(
+                {
+                    "params": matched,
+                    "_label": label,
```
tianyu-l: what's this for?

torchtitan/components/optimizer.py
shuhuayu: ideally yes. I kept this only for simplicity. I did not figure out how users who use the default optimizer adamw for all, can change its config conveniently. Like keeping this base case we can use `optimizer=OptimizersContainer.Config(lr=8e-4)`, but if users need to specify this param group, which requires explicit rewrite
```
OptimizersContainer.Config(
    param_groups=[
        ParamGroupConfig(
            pattern=r".*",
            optimizer_name="AdamW",
            optimizer_kwargs={"lr": 8e-4, "betas": (0.9, 0.95), "eps": 1e-8, "weight_decay": 0.1},
        ),
    ],
)
```
Another way i can think of is to use a helper function like `default_adamw(lr=xxx)` that can return this `ParamGroupConfig`? but that means we need other functions like `default_muon(lr=xxx)`, or we can create a function `default_optimizer(name=xxx, lr=xxx)`. What do you think?

torchtitan/components/optimizer.py
shuhuayu: This is for logging, like in wandb, one can log the lr values of each optimizer-param-group pair.

torchtitan/components/optimizer.py
```diff
@@ -158,92 +160,127 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-        # Then each matched group in pattern order
-        for i, pg in enumerate(config.param_groups):
-            if i not in grouped_params:
                 continue
-            group_kwargs = {**default_kwargs}
-            group_kwargs["lr"] = default_kwargs["lr"] * pg.lr_multiplier
-            group_kwargs["weight_decay"] = (
-                default_kwargs["weight_decay"] * pg.weight_decay_multiplier
+
+            label = re.sub(r"[^a-zA-Z0-9._]", "", pg.pattern.replace("|", "_or_"))
```
tianyu-l: what is this line doing? why can't we just log the pattern?

torchtitan/components/optimizer.py
tianyu-l: > Another way i can think of is to use a helper function like default_adamw(lr=xxx) that can return this ParamGroupConfig?

Feel free to do that. Users can do whatever in the config space, just don't have to have the complexity in actual optimizer class.

E.g. model_registry() functions is programming in config space, not actual runtime space.

torchtitan/components/optimizer.py
shuhuayu: SGD is not necessary, removed.

torchtitan/components/optimizer.py
shuhuayu: Switched to `groups_by_opt_name`. here we are building a dict `{optimizer_name: [param_group_dicts]}` to group parameters by its optimizer name. When config mixed optimizers, we config optimizers for each param_groups; when build optimizers, we build for each each optimizer for efficiency, and each optimizer can be used in several param_groups.

torchtitan/components/optimizer.py
shuhuayu: Yes, we lost some kwargs compared to the previous use. I realize we may not need to pass real kwargs to it at all. Actual optimizers are added here `self.optimizers.append(opt_cls(opt_param_groups))`, and `Optimizer.__init__` is called to support `register_step_pre_hook`, no kwargs needed at all.

torchtitan/components/lr_scheduler.py
shuhuayu: Previously `state_dict` is saved from the `schedulers[0]`, if we stick to that and call `load_state_dict` on `schedulers[1]` which has different learning rate, this would be a bug. `load_state_dict` overwrites `base_lrs` with the saved values, so `schedulers[1]` would get `schedulers[0]`'s learning rate. Setting `last_epoch` directly avoids this, since each scheduler keeps its own `base_lrs` from construction, and recomputes `_last_lr` from there.

torchtitan/components/lr_scheduler.py
shuhuayu: You are right, `schedulers` are never used, and we recompute it from `base_lrs` in construction. As we commented above, this is for the info completeness of a checkpoint.

torchtitan/components/lr_scheduler.py
shuhuayu: Yeah it works since the current lr scheduler in torchtitan are pure lambda functions. Added a note on this.

torchtitan/components/lr_scheduler.py
tianyu-l: "Info completeness" is not convincing -- you got complete info to restore anyway.

I would suggest removing, and adding a note to explain.

cc @fegin

torchtitan/components/lr_scheduler.py
tianyu-l: cc @fegin

torchtitan/components/optimizer.py
```diff
@@ -31,88 +31,100 @@
...
-    beta1: float | None = None
-    beta2: float | None = None
-    """Override betas for this group. None means use the global optimizer betas.
-    Each can be overridden independently."""
+def default_adamw(lr: float = 8e-4, **kwargs: Any) -> list[ParamGroupConfig]:
+    """Create a catch-all AdamW param group with standard defaults.
+
+    Use as a convenience for the common case::
+
+        OptimizersContainer.Config(param_groups=default_adamw(lr=3e-4))
```
tianyu-l: why don't we make it directly return the whole `OptimizerContainer.Config(...)`?

torchtitan/components/optimizer.py
```diff
@@ -124,32 +136,28 @@ class Config(Configurable.Config):
...
-        Patterns are checked in order; first match wins."""
-
-        def __post_init__(self):
-            if self.implementation == "fused_opt_states_bf16":
-                if self.name not in ("Adam", "AdamW"):
-                    raise ValueError(
-                        "implementation='fused_opt_states_bf16' is only supported "
-                        f"for Adam/AdamW, got optimizer '{self.name}'"
-                    )
+        param_groups: list[ParamGroupConfig] = field(default_factory=default_adamw)
```
tianyu-l: nit: list `param_groups` before `implementation`

torchtitan/components/optimizer.py
```diff
@@ -124,32 +136,28 @@ class Config(Configurable.Config):
...
+        """Per-parameter-group optimizer configurations. Each entry specifies a
+        regex pattern and a self-contained optimizer setup.
+        Patterns are checked in order; first match wins.
+        Defaults to ``default_adamw()``."""
```
tianyu-l: No, why?

torchtitan/components/optimizer.py
```diff
@@ -158,92 +166,117 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-        Each parameter is assigned to the first matching ParamGroupConfig pattern,
-        or to the default group if no pattern matches. Returns a list of dicts
-        with "params" key and optimizer kwargs, suitable for passing to an optimizer.
-        """
-        if not config.param_groups:
-            params = [p for p in model.parameters() if p.requires_grad]
-            return [{"params": params, **default_kwargs}]
-
-        compiled_patterns = [re.compile(pg.pattern) for pg in config.param_groups]
+        param_groups: list[ParamGroupConfig],
```
tianyu-l:
```suggestion
        param_group_configs: list[ParamGroupConfig],
```

torchtitan/components/optimizer.py
```diff
@@ -31,88 +31,100 @@
...
-        """Weight decay to use"""
-
         implementation: Literal[
             "for-loop", "foreach", "fused", "fused_opt_states_bf16"
         ] = "fused"
         """
-        Specify which optimizer implementation to use:
+        Optimizer implementation mode applied to all optimizer instances.
+        Per-param-group ``optimizer_kwargs`` can override this (e.g.
+        ``"fused": False`` for optimizers that don't support fused with DTensor).
```
tianyu-l: don't mention DTensor, as many optimizer don't have fused impl regardless

torchtitan/components/optimizer.py
```diff
@@ -158,92 +166,117 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
                 continue
-            group_kwargs = {**default_kwargs}
-            group_kwargs["lr"] = default_kwargs["lr"] * pg.lr_multiplier
-            group_kwargs["weight_decay"] = (
-                default_kwargs["weight_decay"] * pg.weight_decay_multiplier
+
+            result[pg.optimizer_name].append(
+                {
+                    "params": matched,
+                    "_label": pg.pattern,
```
tianyu-l: can we just call it pattern or _pattern?

torchtitan/components/optimizer.py
```diff
@@ -158,92 +166,117 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-            if pg.beta1 is not None or pg.beta2 is not None:
-                default_beta1, default_beta2 = default_kwargs["betas"]
-                group_kwargs["betas"] = (
-                    pg.beta1 if pg.beta1 is not None else default_beta1,
-                    pg.beta2 if pg.beta2 is not None else default_beta2,
-                )
-            result.append({"params": grouped_params[i], **group_kwargs})
 
-        return result
+        return dict(result)
```
tianyu-l: why can't we return `result`?

torchtitan/components/optimizer.py
```diff
@@ -158,92 +166,117 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-            grouped_params[matched_index].append(param)
-
-        # Warn for patterns that matched nothing
-        for i, pg in enumerate(config.param_groups):
-            if i not in grouped_params:
+        Each param group dict includes a ``_label`` key with the regex pattern
+        string for logging.
+        """
+        result: dict[str, list[dict[str, Any]]] = defaultdict(list)
+        claimed: set[str] = set()  # first-match-wins
```
tianyu-l, line 163: I feel this also helps dedup weight-tied params

torchtitan/components/optimizer.py
```diff
@@ -274,32 +307,25 @@ def state_dict(self) -> dict[str, Any]:
...
             options=StateDictOptions(flatten_optimizer_state_dict=True),
         )
-        return {
-            k: v
-            for sd in map(func, self.model_parts, self.optimizers)
-            for k, v in sd.items()
-        }
+        result = {}
+        for opt, part_idx in zip(self.optimizers, self._model_part_indices):
+            sd = func(self.model_parts[part_idx], opt)
```
tianyu-l, line 286: other than potentially making things looking worse here, why don't we extend this optimization https://github.com/pytorch/torchtitan/pull/3269#discussion_r3331049295 across model_parts?

torchtitan/components/lr_scheduler.py
shuhuayu: Removed. What i meant is from checkpoint itself we can recover all `lr`s. Currently we can get complete info in `load_state_dict` because we used info from our construction code plus checkpoint. But I think perhaps I was overthinking, let's remove it.

torchtitan/components/optimizer.py
shuhuayu: Agreed. That's a cleaner separation.

torchtitan/components/optimizer.py
shuhuayu: Was thinking the patterns in `re` pattern may affect logging strings, and it was wrong to assume that. Switched to use plain `re` pattern.

torchtitan/components/optimizer.py
shuhuayu: That's cleaner for the default standalone case. Switched to that. Users can define their own per parameter group helper function like the current one to compose with each other.

torchtitan/components/optimizer.py
shuhuayu: Previously we used `default_factory=default_adamw ` on `param_groups`, but removed it after we make `default_adamw` return a complete `OptimizersContainer.Config`. As for the part of "first match wins`, this is because in `_build_param_groups` we used the `claimed` set, and only add new param group configs when there is no existing matches.

torchtitan/components/optimizer.py
shuhuayu: Accurate.

torchtitan/components/optimizer.py
shuhuayu: Switched to `pattern`.

torchtitan/components/optimizer.py
shuhuayu: We can. `dict(result)` is a defensive step to avoid the case there is an access of non-existing keys. Since there is no such access, we can directly use `result`.

torchtitan/components/optimizer.py
shuhuayu: Conceptually yes, even the tie-parameter appear twice, it will only receive configs once. But here we used `for name, param in model.named_parameters()`, and `named_parameters()` will do the deduplication silently since pytorch uses `id(param)` for that.

torchtitan/components/optimizer.py
shuhuayu: not sure iiuc, if you are referring to one optimizer per type across all model parts, here are some reasons why we could not do that: 1) `get_optimizer_state_dict` needs a single model part to resolve parameter
  FQNs, and 2) in pp, different model parts may be on different devices.

tianyu-l: (1) is legit, but I'm sure it can be worked around by not using `get_optimizer_state_dict` (maybe not worth it)
(2) is not real, different model_parts refer to different PP stages on the same local device (in interleaved schedules). E.g. layer 0, 1, 4, 5 on PP rank 0; layer 2, 3, 6, 7 on PP rank 1. On rank 0, model_parts[0] has 0, 1; model_parts[1] has 4, 5.

shuhuayu: Thx for the correction on 2). I was wrong on that. For 1) it is doable by changing checkpointing logic and can be complex. I am leaning to keep the current per-model-part grouping, and optimize it later if multiple optimizer instances matters.

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +201,48 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
     )
 
 
+def rl_grpo_qwen3_debug() -> RLTrainer.Config:
```
tianyu-l: why adding this?

torchtitan/components/optimizer.py
```diff
@@ -31,88 +31,83 @@
...
-        beta1: float = 0.9
-        beta2: float = 0.95
-        """Exponential moving average hyperparameters to use"""
-
-        eps: float = 1e-8
-        """Epsilon value to use"""
-
-        weight_decay: float = 0.1
-        """Weight decay to use"""
+        param_group_configs: list[ParamGroupConfig] = field(default_factory=list)
```
tianyu-l:
```suggestion
        param_groups: list[ParamGroupConfig] = field(default_factory=list)
```
I guess my point is
- if it's already inside a `Config`, there's no need to call it config explicitly
- if it's mixed with actual objects, e.g. in `build_param_groups()`, then you have to make it explicit

torchtitan/components/optimizer.py
```diff
@@ -158,92 +143,117 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
-        """Build PyTorch param groups from model parameters and config.
-
-        Each parameter is assigned to the first matching ParamGroupConfig pattern,
-        or to the default group if no pattern matches. Returns a list of dicts
-        with "params" key and optimizer kwargs, suitable for passing to an optimizer.
-        """
-        if not config.param_groups:
-            params = [p for p in model.parameters() if p.requires_grad]
-            return [{"params": params, **default_kwargs}]
+        param_group_configs: list[ParamGroupConfig],
```
tianyu-l, line 151: yes, here we should use `..._configs`

torchtitan/components/optimizer.py
```diff
@@ -274,32 +284,25 @@ def state_dict(self) -> dict[str, Any]:
...
-
-    def _post_init(
-        self, all_params: list[nn.Parameter], optimizer_kwargs: dict[str, Any]
-    ) -> None:
+    def _post_init(self, all_params: list[nn.Parameter]) -> None:
         # We need to call Optimizer.__init__() to initialize some necessary optimizer
-        # functionality such as hooks.
-        Optimizer.__init__(self, all_params, optimizer_kwargs)
+        # functionality such as hooks (e.g. register_step_pre_hook for MoE load balancing).
+        Optimizer.__init__(self, all_params, {})
```
tianyu-l, line 302: @fegin please review this

tianyu-l (commented):
Looks good to me, despite a couple of nits.

Please ask @fegin for final stamp.

torchtitan/experiments/rl/config_registry.py
shuhuayu: rebase accident, removed.

torchtitan/components/optimizer.py
shuhuayu: make sense, updated.

torchtitan/components/lr_scheduler.py
fegin: > I feel this has too much assumptions on the LR scheduler internal

IMHO, this is not avoidable. The LR scheduler state_dict doesn't have a good design to support resharding. An alternative is that we don't assume resharding is supported for optimizer and lr_scheduler if we want to keep this simple.

torchtitan/components/optimizer.py
```diff
@@ -31,88 +31,83 @@
...
     """Regex pattern matched against parameter fully qualified names (FQNs).
-    E.g. '.*bias$', '.*norm.*', '.*\\.embed_tokens\\..*'"""
-
-    lr_multiplier: float = 1.0
-    """Multiplied with the global optimizer lr to get this group's lr."""
+    E.g. '.*bias$', '.*norm.*', '.*\\.embed_tokens\\..*', '.*' (catch-all)"""
 
-    weight_decay_multiplier: float = 1.0
-    """Multiplied with the global optimizer weight_decay to get this group's weight_decay."""
+    optimizer_name: str = "AdamW"
```
fegin: Do we want both optimizer_name and optimizer_kwargs to have no default values

torchtitan/components/optimizer.py
```diff
@@ -158,92 +143,117 @@ def _build_optimizer_kwargs(config: Config) -> dict[str, Any]:
...
+                if param.requires_grad and name not in claimed and pattern.search(name):
+                    matched.append(param)
+                    claimed.add(name)
+
+            if not matched:
                 logger.warning(
                     f"Optimizer param_groups pattern '{pg.pattern}' "
                     f"matched no parameters"
                 )
-
```
fegin, line 207: Should this be a warning or hard error? Can we clarify the expectation of this, whether it is acceptable to not have a match?

torchtitan/components/optimizer.py
fegin: this should be a safe change.

torchtitan/components/optimizer.py
fegin: Do we still want to support Optimizer In Backward? Maybe easier to just delete if no one is using this feature. cc., @tianyu-l

tianyu-l: Yes, let's delete it. If complaints show up later, we reconsider.

torchtitan/components/optimizer.py
shuhuayu: thx for the catch. both should have no defaults, updated.

torchtitan/components/optimizer.py
shuhuayu: i think an error is more informative since a warning can be silently ignored in training logs.

torchtitan/components/optimizer.py
shuhuayu: sounds good, deleted.

torchtitan/components/optimizer.py
```diff
@@ -339,83 +339,36 @@ def _bf16_state_init_hook(
...
-            optim.register_step_pre_hook(_bf16_state_init_hook)
+            if isinstance(optim, (torch.optim.Adam, torch.optim.AdamW)):
+                optim.register_step_pre_hook(_bf16_state_init_hook)
 
     def init_cache_state_dict(self) -> None:
         """Initialize cached state dict for TorchFT. No-op for base class."""
         pass
 
 
-class OptimizersInBackwardContainer(OptimizersContainer):
```
tianyu-l, line 349: IIRC there was a test in features.py in CI for this, but I'm not seeing it deleted.

tianyu-l (approved):

fegin (approved):
Good refactor! Some CI failures are real, please fix them.

torchtitan/components/optimizer.py
shuhuayu: Thanks for the catch, removed.

PR #3270, opened 2026-05-07
[graph_trainer] Refactor selective activation remat to in-place

tugsbayasgalan:
Summary:

The remat pass previously rebuilt the graph wholesale (fx.Graph() + node_copy of every node) and relied on whole-graph DCE to remove dead must_recompute originals. Refactor to mutate gm.graph in place: dups are inserted in front of their first backward consumer, backward args are redirected to the dups, and only originals whose users became empty are erased. Original node identities and names are preserved, the topological-order assumption is explicit (input graph order drives insertion, validated by gm.graph.lint() at the end), and the underlying function takes the standard (gm, example_inputs) graph pass signature.

CPU-offload reload chains are handled by hoisting the chain in front of the earliest dup that needs it - the in-place equivalent of upstream's "eagerly copy reload chain into the new graph" trick. 

Authored with Claude.

Test Plan:
Unit tests:
  pytest torchtitan/experiments/graph_trainer/tests/test_passes.py -x
  -> 68 passed, 1 skipped (3 new in TestSelectiveActivationRematPass)

End-to-end on 8xH100 / Llama3 8B / FSDP=4 + TP=2 / aot_fx_trace / no cudagraph, with --debug.seed=42 --debug.deterministic:

Screenshot 2026-05-07 at 5 16 08 PM

CPU offload using upstream remat pass 

Screenshot 2026-05-07 at 5 16 23 PM
CPU offload using our refactor 

 Before: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/.tmpfmqvbv/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000
                                                                                      
After:  https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/.tmp9ImSg9/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

### DIFF
--- a/torchtitan/experiments/graph_trainer/.claude/CLAUDE.md
+++ b/torchtitan/experiments/graph_trainer/.claude/CLAUDE.md
@@ -66,7 +66,8 @@ two-step process:
    - `apply_cpu_offload_pass` — inserts offload/reload/wait ops for
      `MUST_CPU_OFFLOAD` nodes.
    - `selective_activation_remat_pass` — duplicates `MUST_RECOMPUTE`
-     ops before backward and DCEs the originals.
+     ops in front of their backward consumers and erases originals whose
+     consumers were all backward.
 
 The `--compile.memory_policy` config selects the tagging strategy.
 New policies (e.g. budget-aware mixed SAC + offload) should be added
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -28,14 +28,6 @@ def _is_backward_node(node: torch.fx.Node) -> bool:
     return node.meta.get("autograd_backward", False)
 
 
-def _is_recomputed_node(node: torch.fx.Node) -> bool:
-    # TODO: Workaround — recomputed nodes (from SAC) should carry
-    # autograd_backward=True but remat_using_tags_for_fwd_loss_bwd_graph
-    # copies metadata from the original forward node. Fix upstream to
-    # tag recomputed nodes with autograd_backward=True.
-    return node.name.endswith("_recomputed")
-
-
 def _get_layer_id(node: torch.fx.Node) -> int:
     """Extract the layer index from the node's module_fqn metadata.
 
--- a/torchtitan/experiments/graph_trainer/fsdp_passes.py
+++ b/torchtitan/experiments/graph_trainer/fsdp_passes.py
@@ -39,7 +39,6 @@ from torch.utils.checkpoint import CheckpointPolicy
 
 from torchtitan.experiments.graph_trainer.common_utils import (
     _is_backward_node,
-    _is_recomputed_node,
     _MODULE_FQN,
 )
 from torchtitan.tools.logging import logger
@@ -70,6 +69,13 @@ def annotate_fsdp_all_gather(
     graph = gm.graph
 
     def force_recompute_node(node):
+        # Respect MUST_CPU_OFFLOAD set by ``tag_all_offloadable_activations``:
+        # the offload chain already keeps the activation off-GPU between
+        # forward and backward, so re-tagging as MUST_RECOMPUTE/MUST_SAVE
+        # would either undo the offload selection or re-save GPU memory we
+        # just freed.
+        if node.meta.get("recompute") == CheckpointPolicy.MUST_CPU_OFFLOAD:
+            return
         if reshard_after_forward:
             node.meta["recompute"] = CheckpointPolicy.MUST_RECOMPUTE
         else:
@@ -448,9 +454,6 @@ def joint_transformer_block_bucketing_reordering_pass(
             defaults to ``"custom_ops"`` via the parent class.
     """
 
-    def _is_backward(node: torch.fx.Node) -> bool:
-        return _is_backward_node(node) or _is_recomputed_node(node)
-
     def _stack_fn(node: torch.fx.Node) -> list[tuple[str, type]]:
         fqn = node.meta.get("custom", {}).get(_MODULE_FQN)
         if not fqn:
@@ -461,7 +464,7 @@ def joint_transformer_block_bucketing_reordering_pass(
         gm,
         module_bucket_plans,
         insert_overlap_deps,
-        is_backward_fn=_is_backward,
+        is_backward_fn=_is_backward_node,
         module_stack_fn=_stack_fn,
         bucket_mode=bucket_mode,
     ).run()
--- a/torchtitan/experiments/graph_trainer/passes.py
+++ b/torchtitan/experiments/graph_trainer/passes.py
@@ -34,6 +34,7 @@ from torchtitan.distributed.fsdp import get_fsdp_reshard_after_forward_policy
 from torchtitan.experiments.graph_trainer.common_utils import (
     _get_layer_id,
     _is_backward_node,
+    _MODULE_FQN,
     _NOT_IN_LAYERS,
 )
 from torchtitan.experiments.graph_trainer.cpu_offload import (
@@ -60,6 +61,9 @@ from torchtitan.experiments.graph_trainer.remove_noop_passes import (
     remove_detach_pass,
     remove_identity_slice_pass,
     remove_identity_view_pass,
+)
+from torchtitan.experiments.graph_trainer.selective_activation_remat import (
+    selective_activation_remat_pass,
 )
 from torchtitan.tools.logging import logger
 
@@ -775,6 +779,13 @@ def apply_sac_pass(
         if _is_backward_node(node):
             continue
 
+        # Skip the post-layer epilogue (lm_head + loss). Chunked-loss
+        # regions split backward into multiple disjoint regions, and the
+        # remat pass only supports one region with must_recompute deps.
+        fqn = node.meta.get("custom", {}).get(_MODULE_FQN, "")
+        if fqn.startswith(("lm_head", "loss")):
+            continue
+
         if node.target in (
             operator.getitem,
             torch.ops._c10d_functional.wait_tensor.default,
@@ -907,22 +918,6 @@ def tag_with_memory_policy_pass(
     gm = MEMORY_POLICY_REGISTRY[memory_policy](gm, config=config)
     log_activation_memory_policy(gm)
     return gm
-
-
-def selective_activation_remat_pass(
-    gm: torch.fx.GraphModule,
-    example_inputs: tuple | None = None,
-) -> torch.fx.GraphModule:
-    """Duplicate recompute nodes for backward use, then DCE unused forward versions.
-
-    Wraps ``remat_using_tags_for_fwd_loss_bwd_graph`` with the graph pass
-    signature ``(gm, example_inputs)``.
-    """
-    from torchtitan.experiments.graph_trainer.selective_activation_remat import (
-        remat_using_tags_for_fwd_loss_bwd_graph,
-    )
-
-    return remat_using_tags_for_fwd_loss_bwd_graph(gm)
 
 
 def full_inductor_compilation_pass(
--- a/torchtitan/experiments/graph_trainer/selective_activation_remat.py
+++ b/torchtitan/experiments/graph_trainer/selective_activation_remat.py
@@ -4,79 +4,43 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-"""AC rematerialize pass: Duplicates recompute nodes for backward, then DCE removes unused forward versions."""
+"""AC rematerialize pass: in-place duplicate recompute nodes for backward."""
 
-import itertools
 import logging
-from typing import Any, overload
+from typing import Any
 
 import torch
 import torch.fx as fx
 from torch._functorch.compile_utils import raise_getitems
 from torch._functorch.partitioners import (
-    cleanup_recompute_tags,
-    force_save_bw_mutation_src,
     has_recomputable_ops,
     has_recomputable_rng_ops,
-    is_not_collective,
     must_recompute,
 )
 
+from torchtitan.experiments.graph_trainer.common_utils import _is_backward_node
+
 
 log = logging.getLogger(__name__)
-_EMPTY_CUSTOM_META: dict[str, object] = {}
-
-
-def is_impure_node_for_dce(node: fx.Node) -> bool:
-    # Check for special collectives that should be treated as pure
-    if not is_not_collective(node):
-        # It's a collective (wait_tensor, all_gather_into_tensor, etc.)
-        # Treat as pure - can be eliminated if unused
-        return False
-
-    # For everything else, fall back to the DEFAULT logic
-    # This is what eliminate_dead_code() calls when is_impure_node=None
-    impure_random = True
-    if torch._guards.TracingContext.try_get():
-        impure_random = torch._inductor.config.fallback_random
-    return node.is_impure(impure_random)
-
-
-def _is_backward_node(node: fx.Node, use_phase: bool = False) -> bool:
-    """Check if node is in backward region.
-
-    If use_phase is True, only checks custom["phase"] == "backward"
-    (user annotation). Otherwise falls back to node.meta["autograd_backward"],
-    which Dynamo adds when tracing torch.autograd.grad.
-    """
-    custom = node.meta.get("custom", _EMPTY_CUSTOM_META)
-    if use_phase:
-        return custom.get("phase") == "backward"
-    return node.meta.get("autograd_backward", False)
-
-
-def _has_user_phase_annotation(gm: fx.GraphModule) -> bool:
-    """Check if any node has the user-level phase: backward annotation."""
-    return any(
-        node.meta.get("custom", _EMPTY_CUSTOM_META).get("phase") == "backward"
-        for node in gm.graph.nodes
-    )
 
 
 def _collect_backward_regions(
-    gm: fx.GraphModule, use_phase: bool
+    gm: fx.GraphModule,
 ) -> list[tuple[int, int, bool]]:
     """Returns (bwd_start, bwd_end, needs_remat) for each backward region.
 
     Regions are maximal contiguous runs of backward nodes, as [start, end)
-    indices into the graph node list.
+    indices into the graph node list. This is still kind of OK for chunked
+    loss because: (1) we would have errored earlier if there were multiple
+    regions that need recompute, and (2) we only ever do recompute on the
+    last backward.
     """
     regions: list[tuple[int, int, bool]] = []
     bwd_start: int | None = None
     needs_remat = False
 
     for idx, node in enumerate(gm.graph.nodes):
-        if _is_backward_node(node, use_phase=use_phase):
+        if _is_backward_node(node):
             if bwd_start is None:
                 bwd_start = idx
                 needs_remat = False
@@ -94,19 +58,29 @@ def _collect_backward_regions(
     return regions
 
 
-def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModule:
-    """
-    Duplicate recompute nodes for backward use. DCE removes unused forward versions.
+def selective_activation_remat_pass(
+    gm: fx.GraphModule,
+    example_inputs: Any = None,
+) -> fx.GraphModule:
+    """In-place remat: insert recompute duplicates before backward consumers.
 
-    Backward regions are identified by custom["phase"] == "backward" (user
-    annotation) or node.meta["autograd_backward"] == True (set automatically when
-    Dynamo traces torch.autograd.grad). When the user provides phase
-    annotations, only those annotated regions are used.
+    For each ``must_recompute`` forward node consumed by a backward node, a
+    duplicate is inserted just before the backward consumer and that
+    consumer's args are redirected to the duplicate. Original forward nodes
+    whose consumers were all backward become dead and are erased; originals
+    with remaining forward consumers stay.
+
+    The graph is mutated in place: original node identities and names are
+    preserved, only the duplicates carry a ``_recomputed`` suffix. No
+    whole-graph DCE or topological reorder is performed.
+
+    Backward regions are identified by
+    ``node.meta["autograd_backward"] == True``, set by both Dynamo and
+    non-strict ``make_fx`` tracing when tracing ``torch.autograd.grad``.
 
     The graph may contain multiple disjoint backward regions (e.g. chunked
     loss). Regions that do not depend on recomputable forward nodes are
-    skipped. Only one region may require remat; if multiple do, we error
-    and ask the user to annotate which region to rematerialize.
+    skipped. Only one region may require remat; if multiple do, we error.
     """
     if not has_recomputable_ops(gm):
         return gm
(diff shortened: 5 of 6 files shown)


tugsbayasgalan:
@claude Pls review this PR

torchtitan/experiments/graph_trainer/.claude/CLAUDE.md
```diff
@@ -65,8 +65,9 @@ two-step process:
    are no-ops when no nodes carry the relevant tag):
    - `apply_cpu_offload_pass` — inserts offload/reload/wait ops for
      `MUST_CPU_OFFLOAD` nodes.
-   - `selective_activation_remat_pass` — duplicates `MUST_RECOMPUTE`
-     ops before backward and DCEs the originals.
+   - `remat_using_tags_for_fwd_loss_bwd_graph` — duplicates `MUST_RECOMPUTE`
```
SherlockNoMad: selective_activation_remat_pass is a better name

torchtitan/experiments/graph_trainer/tests/test_passes.py
```diff
@@ -1403,6 +1403,213 @@ def test_mm_rs_becomes_fused_op(self):
         self.assertTrue(any(n.target == fused for n in gm.graph.nodes))
 
 
+class TestSelectiveActivationRematPass(TestCase):
```
SherlockNoMad, line 1426: thanks for porting over the test

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -94,20 +77,34 @@ def _collect_backward_regions(
...
+
+    The graph is mutated in place: original node identities and names are
+    preserved, only the duplicates carry a ``_recomputed`` suffix. No
+    whole-graph DCE or topological reorder is performed.
 
-    Backward regions are identified by custom["phase"] == "backward" (user
-    annotation) or node.meta["autograd_backward"] == True (set automatically when
-    Dynamo traces torch.autograd.grad). When the user provides phase
-    annotations, only those annotated regions are used.
+    Backward regions are identified by ``custom["phase"] == "backward"``
```
SherlockNoMad: I don't think custom["phase"] is set anywhere in titan?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
SherlockNoMad: I think we can remove this line. 

this is tagging pass's job to ensure end of layer is tagged as SAVED.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -94,20 +77,34 @@ def _collect_backward_regions(
...
+    (user annotation) or ``node.meta["autograd_backward"] == True`` (set by
+    Dynamo when tracing ``torch.autograd.grad``). When the user provides
+    phase annotations, only those annotated regions are used.
 
     The graph may contain multiple disjoint backward regions (e.g. chunked
     loss). Regions that do not depend on recomputable forward nodes are
     skipped. Only one region may require remat; if multiple do, we error
     and ask the user to annotate which region to rematerialize.
     """
+    del example_inputs  # unused; accepted for graph pass signature compatibility
```
SherlockNoMad: don't need?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
SherlockNoMad: deprecate use_phase

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,79 +148,128 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
-    order = {node: idx for idx, node in enumerate(gm.graph.nodes)}
-    new_graph = fx.Graph()
-    env: dict[fx.Node, fx.Node] = {}
-    recomputed_nodes: dict[fx.Node, fx.Node] = {}
-
-    # Insert forward nodes
-    for node in itertools.islice(gm.graph.nodes, 0, bwd_start):
-        env[node] = new_graph.node_copy(node, lambda x: env[x])
+    all_nodes = list(gm.graph.nodes)
+    bwd_nodes = all_nodes[bwd_start:bwd_end]
```
SherlockNoMad, line 118: in case of chunked loss, I don't think this heuristic is robust?

SherlockNoMad:
Show me a tlparse for diffing the graph before and after, sample script https://www.internalfb.com/intern/paste/P2312893118/

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: here the assumption is that chunked region doesn't do AC

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,79 +148,128 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+    # remaining forward consumers stay in place. Iterate in reverse graph
+    # order so downstream originals are erased first, freeing their upstream
+    # originals' user lists for erase in the same pass.
+    for orig in reversed(list(recomputed_nodes)):
+        if not orig.users:
+            gm.graph.erase_node(orig)
 
-    new_gm.recompile()
+    # raise_getitems pass for better memory (like default_partition)
+    gm = raise_getitems(gm)
```
SherlockNoMad, line 324: Is this needed?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: so we really end up with one backward region we want to do remat. I think apply_sac shouldn't tag the lm_head layer with AC.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,79 +148,128 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
 
-    return new_gm
+    gm.graph.lint()
```
SherlockNoMad: gm.recompile

torchtitan/experiments/graph_trainer/selective_activation_remat.py
SherlockNoMad: do we still need this line?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
SherlockNoMad: yea, make sense. Let's make the assumption explicit as a comment, so that it can be adjusted accordingly if assumption changes.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: i will do it as followup

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: ok

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: yeah i think this seemed like net positive thing that moves getitem calls closer to the producer.

tugsbayasgalan: claude response:
  Some ops return more than one tensor. For example var_mean returns a tuple of (var, 
  mean). In fx that looks like:
                                                                                      
  parent = aten.var_mean(x)    # produces a tuple of 2 tensors    
  var    = parent[0]           # getitem extracts var                                 
  mean   = parent[1]           # getitem extracts mean                                
                                                                                      
  parent (the tuple holding pointers to var-storage and mean-storage) stays alive in  
  GPU memory until both parent[0] and parent[1] have run. That's a refcount thing —
  the tuple has two consumers (the two getitems), and only after the last consumer    
  runs can it be freed.                                           

  The problem: sometimes those two getitems end up far apart in the graph:            
   
  parent = aten.var_mean(x)                                                           
  var    = parent[0]                                                                  
   
  ... lots of other ops ...     # parent is still alive!                              
  ... 100 more lines ...        # parent is still alive!          
                                                                                      
  mean   = parent[1]            # NOW parent can finally be freed                     
                                                                                      
  The tuple holding both var-storage and mean-storage sits in GPU memory across all   
  those "other ops" — wasted memory.                              
                                                                                      
  What raise_getitems does: moves every getitem to immediately after its parent.      
   
  parent = aten.var_mean(x)                                                           
  var    = parent[0]                                              
  mean   = parent[1]            # moved up
                                                                                      
  ... lots of other ops ...     # parent already freed
                                                                                      
  Now both extractions happen right away, the tuple's refcount drops to zero, and it  
  gets freed immediately. The individual var and mean tensors are still alive (they
  have downstream consumers), but the tuple wrapper is gone.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: Yeah this is the ultimate fallback. I can do cleanup as follow up. DOn't want to make this PR too big.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,79 +118,143 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+            # Re-hoisting in either case would collapse the prefetch gap the
+            # offload pass set up, killing async H2D overlap.
+            prev = moved_offload.get(n)
+            anchor_pos = order[prev] if prev is not None else order[n]
+            if anchor_pos <= order[target]:
+                continue
+            bwd_reload_chain.add(n)
+            stack.extend(n.all_input_nodes)
+
+        # TODO: when we DO move (chain is currently behind target), all
```
tugsbayasgalan, line 248: cc: @mlazos Claude tells me this is an issue even in previous implementation. Can this case ever happen?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: lets just deleteit

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -28,14 +28,6 @@ def _is_backward_node(node: torch.fx.Node) -> bool:
     return node.meta.get("autograd_backward", False)
 
 
-def _is_recomputed_node(node: torch.fx.Node) -> bool:
```
SherlockNoMad, line 31: nice clean up

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -449,7 +455,7 @@ def joint_transformer_block_bucketing_reordering_pass(
     """
 
     def _is_backward(node: torch.fx.Node) -> bool:
```
SherlockNoMad: this layer of wrapper can be removed.

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -775,6 +779,13 @@ def apply_sac_pass(
         if _is_backward_node(node):
             continue
 
+        # Skip the post-layer epilogue (lm_head + loss). Chunked-loss
+        # regions split backward into multiple disjoint regions, and the
+        # remat pass only supports one region with must_recompute deps.
+        fqn = node.meta.get("custom", {}).get(_MODULE_FQN, "")
+        if fqn.startswith(("lm_head", "loss")):
+            continue
```
SherlockNoMad, line 787: we should treat this as an defensive programming practice, 
tagging pass should skip these region when tagging 

Not sure how reliable it's to count on startwith xyz

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -4,67 +4,36 @@
...
-    # For everything else, fall back to the DEFAULT logic
-    # This is what eliminate_dead_code() calls when is_impure_node=None
-    impure_random = True
-    if torch._guards.TracingContext.try_get():
-        impure_random = torch._inductor.config.fallback_random
-    return node.is_impure(impure_random)
-
-
-def _is_backward_node(node: fx.Node, use_phase: bool = False) -> bool:
+def _is_backward_node(node: fx.Node) -> bool:
```
SherlockNoMad: reuse the one from common_utils

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -94,19 +63,29 @@ def _collect_backward_regions(
...
-    annotations, only those annotated regions are used.
+    Backward regions are identified by
+    ``node.meta["autograd_backward"] == True``, set by both Dynamo and
+    non-strict ``make_fx`` tracing when tracing ``torch.autograd.grad``.
 
     The graph may contain multiple disjoint backward regions (e.g. chunked
     loss). Regions that do not depend on recomputable forward nodes are
-    skipped. Only one region may require remat; if multiple do, we error
-    and ask the user to annotate which region to rematerialize.
+    skipped. Only one region may require remat; if multiple do, we error.
```
SherlockNoMad, line 83: > Only one region may require remat; if multiple do, we error.

what does this mean?

tugsbayasgalan: For some reason if we end up with multiple backward region each requiring AC recompute, we error for now. We can later extend it to support multiple backward each doing their own AC recompute

SherlockNoMad: I see, so this pass is assuming the entire backward is contiguous. 

and backward of chunkloss is not included?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
SherlockNoMad, line 53: hmm... would this be broken by chunked loss?

SherlockNoMad: we can land this first without supporting chunkedloss, just raising the q

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -118,26 +97,15 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
 
-    # Use partitioner pass to normalize AC node tags.
-    gm = cleanup_recompute_tags(gm, is_default_partition=True)
-
-    force_save_bw_mutation_src(gm)
-
-    # must_recompute (used inside _collect_backward_regions) requires
-    # cleanup_recompute_tags to have run first.
-    use_phase = _has_user_phase_annotation(gm)
-    regions = _collect_backward_regions(gm, use_phase)
```
SherlockNoMad, line 129: nice cleanup and simplification

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: The expectation is that we will find all backward regions and skip all regions that don't need AC.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: They are included. but we won't be doing any recompute in those chunked regions so we just skip them in the pass.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
-
-    def remat_input(x: object) -> object:
-        # fx.Node can have args that are primitive types (e.g. int, float, bool)
-        if not isinstance(x, fx.Node):
-            return x
-        return recomputed_nodes.get(x, env[x])
-
-    def gather_recompute_deps(node: fx.Node) -> set[fx.Node]:
-        deps: set[fx.Node] = set()
+    def claim_for(bwd_node: fx.Node) -> None:
```
SherlockNoMad: we can have a better name for this . 

def collect_fw_nodes_to_recompute_for(bwd_node)

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+        # chain members get prepended adjacent to ``target`` — collapsing
+        # the prefetch gap between ``reload`` and ``wait_tensor`` to ~0
+        # and serializing the H2D against compute. If this becomes a
+        # measurable regression we should place ``reload`` early in the
+        # bwd region and ``wait_tensor`` just before ``target`` to restore
+        # overlap.
+        # Prepend in graph (topological) order so deps land before dependents.
+        for n in sorted(bwd_reload_chain, key=order.__getitem__):
+            target.prepend(n)
+            moved_offload[n] = target
```
SherlockNoMad, line 258: log.debug(moved x before y)

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+        recomputed_nodes[fwd_node] = dup
+        log.debug(
+            "Recomputing %s before backward node %s", fwd_node.name, bwd_target.name
+        )
+
+    # Redirect each backward node's inputs to the recomputed duplicates.
+    # Backward consumers of offloaded forward nodes were already redirected
+    # to their reload chain by the CPU offload pass, so the offload branch
+    # of remat_input is a no-op here.
+    for bwd_node in bwd_nodes:
```
SherlockNoMad: nit, I think you can just it over the unique value of remat_targets? 

i.e. the bw node that triggered the recompute

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+        bwd_node.kwargs = torch.fx.map_arg(bwd_node.kwargs, remat_input)
+
+    # Targeted erase: original forward must_recompute nodes whose consumers
+    # were all backward now have no users and can be removed. Originals with
+    # remaining forward consumers stay in place. Iterate in reverse graph
+    # order so downstream originals are erased first, freeing their upstream
+    # originals' user lists for erase in the same pass.
+    for orig in reversed(list(recomputed_nodes)):
+        if not orig.users:
+            gm.graph.erase_node(orig)
```
SherlockNoMad, line 321: log.debug(erased x, in replace of x_recomputed)

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+        if reload_node is not None:
+            return reload_node
+        return x
+
+    # Walk the existing graph in forward order. Each claimed must_recompute
+    # node is duplicated immediately before its target backward node; topology
+    # is satisfied by the graph's own order — upstream deps appear earlier in
+    # the walk, so their duplicates are inserted (and visible via
+    # ``recomputed_nodes``) before any downstream duplicate references them.
+    for fwd_node in all_nodes[:bwd_start]:
```
SherlockNoMad: all_nodes is kinda of slow, I think you can simplify this by 
walking in the original order of remat_targets's keys, sort by `order` first

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
-    new_gm.recompile()
+    def remat_input(x: object) -> object:
+        """Arg-transform: redirect must_recompute originals to their dups, and
+        offloaded forward nodes to their CPU-reload chain. Hoisting of the
+        reload chain happens separately in the dup-creation loop."""
+        if not isinstance(x, fx.Node):
+            return x
+        if x in recomputed_nodes:
+            return recomputed_nodes[x]
+        reload_node = x.meta.get("cpu_offload_reload_node")
```
SherlockNoMad: it's more reliable to just rely on the target, 
ao.reload, ao.wait are the reload chain

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +119,173 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+        bwd_target = remat_targets.get(fwd_node)
+        if bwd_target is None:
+            continue
+        # Pre-hoist offload reload chains for any args referencing offloaded
+        # forward nodes, so the chain executes before the dup we're about to
+        # create. Mirrors upstream's eager-copy-into-new-graph trick.
+        for arg in fwd_node.all_input_nodes:
+            reload_node = arg.meta.get("cpu_offload_reload_node")
+            if reload_node is not None:
+                ensure_offload_chain_before(reload_node, bwd_target)
```
SherlockNoMad: same here, don't rely on meta. use target directly.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: we only track forward to first backward that needs this forward. So we will miss some backward nodes that need to be redirected.

tugsbayasgalan: oh well i guess we can go from  remat_targets.keys() and derive all backward nodes.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: Done. Although now we have one more whole graph scan to populate the targets.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: ditto

torchtitan/experiments/graph_trainer/selective_activation_remat.py
tugsbayasgalan: Done.

torchtitan/experiments/graph_trainer/passes.py
tugsbayasgalan: discussed offline

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -151,80 +114,214 @@ def remat_using_tags_for_fwd_loss_bwd_graph(gm: fx.GraphModule) -> fx.GraphModul
...
+    recomputed_nodes: dict[fx.Node, fx.Node] = {}
+    # CPU offload: track which bwd target each reload-chain node was last
+    # hoisted before, so we can re-hoist if an earlier dup needs it later.
+    moved_offload: dict[fx.Node, fx.Node] = {}
+
+    # Build offloaded_fwd -> bwd_wait map by walking the offload op pattern
+    # (apply_cpu_offload_pass emits: F -> ao.offload -> ao.wait_tensor ->
+    # ao.reload -> ao.wait_tensor). Used to redirect a recompute dup that
+    # consumes an offloaded fwd to read from the bwd-region GPU value.
+    offloaded_fwd_to_bwd_wait: dict[fx.Node, fx.Node] = {}
```
SherlockNoMad, line 154: turn this into a util function

offloaded_fwd_to_bwd_wait = collect_x_and_y_pair()

SherlockNoMad (approved):

PR #3278, opened 2026-05-08
[spmd_types] llama3 enablement

pianpwk:
* #3740
* #3690
* #3673
* #3655
* #3741
* #3664
* #3654
* __->__ #3278


torchtitan/components/loss.py
```diff
@@ -32,6 +35,56 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+    assert tp_pg is not None
+    tp_rank = dist.get_rank(tp_pg)
+    vocab_chunk_size = logits.shape[-1]
+    logits_2d = logits.flatten(0, 1).float()
+    labels_1d = labels.flatten(0, 1)
+
+    # Distributed log-sum-exp: 2 all-reduces for numerical stability
+    with torch.no_grad():
+        local_max = logits_2d.max(dim=-1).values
+        global_max = funcol.all_reduce(local_max, "MAX", tp_pg)
```
pianpwk: this needs {tp: R} annot

torchtitan/components/loss.py
```diff
@@ -353,10 +388,108 @@ def __call__(
...
+        grad_leaf = (grad_types, grad_spec) if grad_spec else grad_types
+
+        return (loss_type, grad_leaf)
+
+    def __call__(
+        self,
+        pred: torch.Tensor,
+        labels: torch.Tensor,
+        global_valid_tokens: torch.Tensor | None = None,
+    ) -> torch.Tensor:
```
pianpwk, line 524: handling for TP/SP/LP isn't quite right, still WIP

torchtitan/models/common/attention.py
```diff
@@ -208,6 +209,36 @@ def __init__(self, config: Config) -> None:
...
+        k: torch.Tensor,
+        v: torch.Tensor,
+        *,
+        attention_masks: BlockMask | None,
+        scale: float | None,
+        return_lse: bool,
+        enable_gqa: bool,
+    ):
+        """Runs compiled FlexAttention under no-typecheck, re-annotates following q."""
+        with spmd.no_typecheck():
```
ezyang, line 267: It would be good to have a more streamlined recipe for flex specifically

ezyang:
Suggested review prompt:

> We're reviewing HEAD. In all of the code here, to understand if the spmd_types
rewrite is correct, we need to be able to infer what the original DTensor
placements of the code were. This is nontrivial because DTensor is a runtime
concept, but we can often figure things out by context clues as well as the
conventional sharding we expect to see in llama3. First, review the modified
code regions per the HEAD~ contents and make your best guess on what the
placements are. Then review spmd_types change in HEAD for correctness. If you
need more context on spmd_types, a checkout lives at ~/Dev/spmd_types

torchtitan/components/loss.py
```diff
@@ -25,9 +29,6 @@
 
 def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
     """Cross-entropy loss with sum reduction for token-based normalization."""
-    if isinstance(pred, DTensor) and isinstance(labels, DTensor):
-        return _cross_entropy_via_local_map(pred, labels)
```
ezyang, line 38: This seems sus, as it seems the local map codepath has gone completely dead

torchtitan/components/loss.py
```diff
@@ -518,7 +614,7 @@ def forward(
         loss: torch.Tensor,
     ) -> torch.Tensor:
         ctx.save_for_backward(accumulated_grad)
-        return loss.detach()
+        return loss.detach().clone()
```
ezyang: Sus

torchtitan/components/loss.py
```diff
@@ -36,77 +37,109 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
-        Replicate() if isinstance(p, Shard) and p.dim == 2 else p
-        for p in pred.placements
-    )
-    if labels.placements != expected_labels_placements:
-        raise ValueError(
-            f"cross_entropy_loss: expected labels placements {expected_labels_placements}, "
-            f"got {labels.placements}"
-        )
+@spmd.register_autograd_function
+class _LossParallelCrossEntropy(torch.autograd.Function):
```
ezyang, line 66: This could have easily been its own PR, all by itself

torchtitan/components/loss.py
```diff
@@ -36,77 +37,109 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
-        flat_pred = pred_local.flatten(0, 1).float()
-        flat_labels = labels_local.flatten(0, 1)
-        if not vocab_sharded:
-            return torch.nn.functional.cross_entropy(
-                flat_pred,
-                flat_labels,
-                reduction="sum",
-                ignore_index=IGNORE_INDEX,
-            )
+        result = _LossParallelCrossEntropy.apply(logits, labels)
```
ezyang: Does this really support arbitrary input sharding on logits/labels? Seems unlikely...

torchtitan/components/loss.py
ezyang: Reader: this is a custom autograd function because we've eliminated the loss parallel context manager and now must do it by hand.

torchtitan/components/loss.py
```diff
@@ -36,77 +37,109 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        tp_rank = dist.get_rank(tp_pg)
+
+        logits_shape = logits.shape
+        vocab_chunk_size = logits_shape[-1]
+        logits_2d = logits.flatten(0, -2).float()
+        labels_1d = labels.flatten()
+
+        # allreduce max for max logit value
+        local_max = logits_2d.max(dim=-1).values
+        global_max = local_max.clone()
```
ezyang: This clone seems unnecessary

torchtitan/components/loss.py
```diff
@@ -36,77 +37,109 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
-        flat_pred_placements = tuple(_flatten_placement(p) for p in pred.placements)
-        flat_labels_placements = tuple(_flatten_placement(p) for p in labels.placements)
-        pred_dtensor = DTensor.from_local(
-            flat_pred, mesh, flat_pred_placements, run_check=False
-        )
-        labels_dtensor = DTensor.from_local(
-            flat_labels, mesh, flat_labels_placements, run_check=False
+    @staticmethod
+    # pyrefly: ignore [bad-override]
+    def forward(
```
ezyang, line 108: Does this manage to torch.compile?

torchtitan/components/loss.py
```diff
@@ -36,77 +37,109 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        local_target_logit = logits_2d[
+            torch.arange(labels_1d.shape[0], device=logits.device), local_idx
+        ]
+        local_target_logit = local_target_logit * in_range.to(logits_2d.dtype)
+        target_logit = local_target_logit.clone()
+        dist.all_reduce(target_logit, op=dist.ReduceOp.SUM, group=tp_pg)
+
+        # compute loss
+        log_global_exp_sum = global_exp_sum.log()
+        per_token_loss = -target_logit + log_global_exp_sum + global_max
```
ezyang: Claude says:

```
   2. Loss‑scalar op ordering → not bitwise‑identical (grad is). This is the one worth flagging against the
  CLAUDE.md bitwise requirement:
    - PyTorch gathers from log_probs: loss_t = -log_probs[target] = -((logit−max) − logsumexp).
    - New gathers the raw logit and recombines: loss_t = -logit[target] + logsumexp + max.

  Same value in exact arithmetic, but different float32 association → the loss scalar can differ by ~1
  ULP. The gradient is bitwise identical, because both exp() the same log_probs = shifted −
  log_global_exp_sum (shifted, max, and the all‑reduced sumexp are computed identically). Since grad_norm
  comes from grads it'll match; only the logged loss may drift in the low digits. If exact loss parity
  matters, gather log_probs[target] instead of the raw logit to match PyTorch's ordering.
```

torchtitan/components/loss.py
```diff
@@ -36,77 +37,109 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        dist.all_reduce(global_max, op=dist.ReduceOp.MAX, group=tp_pg)
+
+        # allreduce sum over offset logits for global denominator
+        shifted = logits_2d - global_max.unsqueeze(-1)
+        exp_shifted = shifted.exp()
+        global_exp_sum = exp_shifted.sum(dim=-1)
+        dist.all_reduce(global_exp_sum, op=dist.ReduceOp.SUM, group=tp_pg)
+
+        #  Mask labels outside this vocab shard; the TP all-reduce selects the owner rank's logit.
+        vocab_start = tp_rank * vocab_chunk_size
```
ezyang: So this only works with even sharding right?

torchtitan/components/loss.py
ezyang: Concretely, the PR accepts DTensor, unwraps them, and then calls this autograd function, and then rewraps them into DTensor, and goes on. This can be done without any of the other infra.

torchtitan/components/loss.py
```diff
@@ -350,89 +383,66 @@ def __init__(
         self._maybe_compile(compile_config)
         self.num_chunks = config.num_chunks
         self.lm_head: nn.Module | None = None
+        self.enable_sp: bool = False
+        self.loss_parallel: bool = False
```
ezyang: @tianyu-l We knew ahead of time we would have to do this to remove the DTensor runtime tracking, but it would be good to see if you like this new form.

torchtitan/trainer.py
```diff
@@ -469,14 +473,20 @@ def __init__(self, config: Config):
             base_folder=config.dump_folder,
         )
 
-        loss_parallel_enabled = (
-            parallel_dims.tp_enabled and not config.parallelism.disable_loss_parallel
-        )
         self.train_context = dist_utils.get_train_context(
-            loss_parallel_enabled,
+            False,
```
ezyang: ?

torchtitan/trainer.py
```diff
@@ -469,14 +473,20 @@ def __init__(self, config: Config):
...
             spmd_typechecking=config.debug.spmd_typechecking,
         )
 
+        if isinstance(self.loss_fn, ChunkedCELoss):
```
ezyang: Kinda ugly, not sure what a better way is though (@tianyu-l)

torchtitan/trainer.py
```diff
@@ -784,7 +794,12 @@ def train_step(
             global_valid_tokens = local_valid_tokens.float()
 
         if is_spmd_active() and isinstance(global_valid_tokens, torch.Tensor):
-            spmd.assert_type(global_valid_tokens, spmd.I)
+            spmd.assert_type(
+                global_valid_tokens,
+                named_placement_to_spmd(
```
ezyang: oh, how come this doesn't work. Do we need to cut a `spmd_types` release lmao

torchtitan/models/llama3/model.py
```diff
@@ -114,11 +109,12 @@ def update_from_config(
             from torchtitan.models.llama3.sharding import set_llama3_sharding_config
             from torchtitan.components.loss import ChunkedCELoss
 
+            chunked_loss = isinstance(trainer_config.loss, ChunkedCELoss.Config)
             set_llama3_sharding_config(
                 self,
-                loss_parallel=not parallelism.disable_loss_parallel,
+                loss_parallel=chunked_loss and not parallelism.disable_loss_parallel,
                 enable_sp=parallelism.enable_sequence_parallel,
-                chunked_loss=isinstance(trainer_config.loss, ChunkedCELoss.Config),
```
ezyang: oh I guess testing the isinstance is so bad then lol

torchtitan/models/llama3/model.py
```diff
@@ -85,13 +85,8 @@ def update_from_config(
             # Sync rope max_seq_len
             self.rope = dataclasses.replace(self.rope, max_seq_len=seq_len)
 
-            if parallelism.context_parallel_degree > 1 and isinstance(
-                self.layers[0].attention.inner_attention, VarlenAttention.Config
-            ):
-                raise NotImplementedError(
-                    "Context Parallel only supports SDPA and FlexAttention. "
-                    "Varlen attention is not supported with CP."
-                )
```
ezyang, line 81: What happened here?

torchtitan/components/loss.py
```diff
@@ -321,10 +354,10 @@ class ChunkedCELoss(BaseLoss):
...
 
     TP / SP composability:
-        Hidden states are redistributed to ``Replicate()`` on the TP mesh
-        before chunking, so each chunk enters the lm_head as ``Replicate()``
-        input regardless of whether SP is enabled. With SP, this is an
-        all-gather from ``Shard(1)``; without SP, it's a no-op.
+        Hidden states are redistributed to ``R`` on the TP mesh before
+        chunking, so each chunk enters the lm_head as ``R`` input regardless
+        of whether SP is enabled. With SP, this is an all-gather from
+        ``S(1)``; without SP, it's a no-op.
```
ezyang: I think you gotta talk about the new configs here lmao

ezyang:
A big mega question: should we block on global SPMD types being "good" enough before landing this? Because we are losing correctness guarantees relative to the old torchtitan code by doing DTensor to local spmd types.

ezyang (changes requested):
bumping from inbox

torchtitan/components/loss.py
tianyu-l: Instead of keeping this bool, we should just config `self.fn` to use `_LossParallelCrossEntropy`.

torchtitan/models/common/nn_modules.py
tianyu-l, line 1: nn_modules is for wrapping existing nn.modules. If you are inventing new ones, please put in their dedicated files, e.g. embedding.py

torchtitan/trainer.py
tianyu-l: Since we have dedicated `_LossParallelCrossEntropy`, `loss_fn` shouldn't have any thing depending on "loss_parallel".

torchtitan/components/loss.py
```diff
@@ -451,52 +461,125 @@ def _chunk(t):
...
+        # With SP, we all-gather S(1)->R. without SP, we allreduce gradients across vocab shards.
+        mesh = current_mesh() if is_spmd_active() else None
+        if mesh is not None and "tp" in (mesh.mesh_dim_names or ()):
+            hidden_states = spmd.redistribute(
+                hidden_states,
+                mesh.get_group("tp"),
+                src=spmd.S(1) if self.enable_sp else spmd.I,
+                dst=spmd.R,
+                backward_options={"op_dtype": hidden_states.dtype},
+            )
```
tianyu-l: this should happen in sharding config, not here depending on `enable_sp`

torchtitan/trainer.py
tianyu-l: similarly, `enable_sp` shoudn't be a property of `loss_fn`

torchtitan/components/loss.py
pianpwk: the pain point is ChunkedCELoss is not a module, so it isn't affected by with_redistribution. I think either we'd extend parallelize() to non-modules, or convert this to one?

torchtitan/components/loss.py
pianpwk: moving to https://github.com/pytorch/torchtitan/pull/3467, will move the rest and this PR's ghstack metadata in a bit. cc @tianyu-l

torchtitan/components/loss.py
tianyu-l: what's the key thing that makes `ChunkedCELoss` not a Module?

Also, you could insert this redistribute as out_src -> out_dst for e.g. the norm before loss?

pianpwk: not sure, lack of weights, unclear ownership of lm_head? cc @wwwjn . but would be a good follow up. last norm seems a good place for now.

torchtitan/models/llama3/model.py
```diff
@@ -86,12 +86,26 @@ def update_from_config(
...
+                and not chunked_loss
+            ):
+                raise ValueError(
+                    "Llama3 local SPMD loss parallel requires ChunkedCELoss. "
+                    "Use ChunkedCELoss or set parallelism.disable_loss_parallel=True."
+                )
             set_llama3_sharding_config(
                 self,
-                loss_parallel=not parallelism.disable_loss_parallel,
+                loss_parallel=loss_parallel,
```
tianyu-l: After megatron-style loss parallel is supported, this shouldn't be in the parallelism config any more -- it should be part of "model code"

torchtitan/models/common/rope.py
```diff
@@ -349,6 +353,9 @@ def _rotate_half(x: torch.Tensor) -> torch.Tensor:
         return torch.cat((-x2, x1), dim=-1)
 
 
+@spmd.local_map(
+    out_types={"dp": spmd.S(0), "cp": spmd.S(1), "tp": spmd.R}
+)
```
pianpwk: @tianyu-l I want to argue this is a local SPMD region, and we need spmd.local_map ?

DTensor works because the `rope_cache_expanded = rope_cache[None, :, None, :].expand(bsz, -1, -1, -1)` expands to the global shape.

So before torch.gather, rope_cache_expanded is [global_bs, global_seq, ...]: Rep@DP, Rep@CP, while positions is S(0)@DP, S(1)@CP. The implicit redistribution is Rep->S(0)@DP, making global shape expand -> local shape.

spmd_types doesn't have global shape access, so rope_cache_expanded is [local_bs, local_seq, ...]. We could mutate_type("dp", R->V), but I don't think that's quite right.

torchtitan/models/common/rope.py
```diff
@@ -368,16 +375,7 @@ def _reshape_for_broadcast(
...
-        assert positions.shape == (1, seqlen)
-        rope_cache = rope_cache[positions.squeeze(0)]
-        assert rope_cache.shape == (seqlen, cache_width)
-        shape = [
-            d if i == 1 else cache_width if i == ndim - 1 else 1
-            for i, d in enumerate(query_shape)
-        ]
-        return rope_cache.view(*shape)
-    else:
+    elif positions.size(0) == bsz:
```
pianpwk: similarly, this reorder is because previous code relied on DTensor global shape comparisons (if local/micro batch size is 1, we'd take the local-broadcasting branch.

I don't remember the exact error but global typechecking would not like the ops we were doing

torchtitan/models/llama3/model.py
pianpwk: I think you still need this flag to set whether lm_head immediately allgathers vocab dim or not? But working on moving non-spmd_types backends to _LossParallelCrossEntropy too

torchtitan/models/llama3/model.py
```diff
@@ -80,11 +80,23 @@ def update_from_config(
...
+            loss_parallel = not parallelism.disable_loss_parallel
+            if (
+                parallelism.spmd_backend == "spmd_types"
+                and loss_parallel
+                and not chunked_loss
+            ):
+                raise ValueError(
+                    "Llama3 local SPMD loss parallel requires ChunkedCELoss. "
+                    "Use ChunkedCELoss or set parallelism.disable_loss_parallel=True."
+                )
```
pianpwk: @tianyu-l how important is this? WIP and not a big change, but in case it has breakage

what is the difficulty though? it feels without chunked loss it should be easier

torchtitan/models/common/attention.py
```diff
@@ -207,6 +209,36 @@ def __init__(self, config: Config) -> None:
...
+                k,
+                v,
+                block_mask=block_mask,
+                scale=scale,
+                enable_gqa=enable_gqa,
+                return_aux=return_aux,
+                kernel_options=kernel_options,
+            )
+        if get_spmd_backend() == "spmd_types":
+            spmd.assert_type(out, spmd.V, spmd.PartitionSpec("dp", "tp", "cp", None))
```
tianyu-l, line 279: Earlier you said some thing like "we can't assert str types on the axes that don't exist in current mesh, because sharding prop will error out reporting no such axes". But here you are annotating full axes without knowing what mesh axes are available.

Also, when do we consider using `spmd.local_map` vs. not -- does it only provide local typechecking mode, but not no_typechecking mode?

It'd be good / consistent to make it a closure with in one API call? Doesn't sound blocking though.

torchtitan/models/common/attention.py
```diff
@@ -207,6 +209,36 @@ def __init__(self, config: Config) -> None:
...
+        k: torch.Tensor,
+        v: torch.Tensor,
+        *,
+        block_mask: BlockMask | None,
+        scale: float | None,
+        enable_gqa: bool,
+        return_aux: AuxRequest,
+        kernel_options: dict,
+    ):
+        """Run compiled FlexAttention outside SPMD typechecking, and propagate types."""
```
tianyu-l: say something like "because spmd_types cannot propagate into Flex"

torchtitan/models/llama3/model.py
tianyu-l: sounds OK, let's add a note on the "WIP" part

torchtitan/models/llama3/model.py
tianyu-l: > you still need this flag to set whether lm_head immediately allgathers vocab dim or not?

hmm could you say more? I was thinking
1. we should always enable loss parallel (via the autograd fn), similar to VocabParallelEmbedding
2. I can imagine using chunked loss or not affecting the decision of gathering or not at the norm output

torchtitan/models/common/rope.py
tianyu-l: hmm, need to check if we could pass in proper shape to make it remain global

torchtitan/models/common/attention.py
pianpwk: > Earlier you said some thing like "we can't assert str types on the axes that don't exist in current mesh, because sharding prop will error out reporting no such axes". But here you are annotating full axes without knowing what mesh axes are available.

Sorry, I want to say that was much earlier, before I learned spmd_types will do size-1 axis filtering. So we can set full [dp,cp,tp] mesh context, call assert types, and spmd_types decides whether to annotate each axis or not.

> spmd.local_map vs. not -- does it only provide local typechecking mode, but not no_typechecking mode?

Yes. Added that to API improvements.

torchtitan/models/llama3/model.py
pianpwk: I think if your model uses plain cross_entropy_loss (no LP), the lm_head matmul is sharded, but you need to allgather logits right after, unlike in LP where you can do distributed softmax.

But yes if we delete non-LP altogether, no problem

> I can imagine using chunked loss or not affecting the decision of gathering or not at the norm output

also cc https://github.com/pytorch/torchtitan/pull/3682 should move that unconditionally

.github/workflows/integration_test_8gpu_features.yaml
```diff
@@ -139,6 +139,17 @@ jobs:
...
+        echo "Checking Llama3 FlexAttention FSDP+TP+CP SPMD loss parity against full_dtensor"
+        export spmd_dense_options="--activation_checkpoint.mode=none --parallelism.tensor_parallel_degree=2 --parallelism.context_parallel_degree=2"
+        python3 scripts/loss_compare.py . . \
+          --baseline-config=llama3_debugmodel_flex_attn \
+          --test-config=llama3_debugmodel_flex_attn \
+          --baseline-options="${spmd_dense_options} --parallelism.spmd_backend=full_dtensor" \
+          --test-options="${spmd_dense_options} --parallelism.spmd_backend=spmd_types --debug.spmd_typechecking" \
+          --job-dump-folder="${RUNNER_TEMP}/artifacts-to-be-uploaded/spmd_loss_comparison" \
+          --no-seed-checkpoint --assert-equal --steps=10
+        rm -rf $RUNNER_TEMP/artifacts-to-be-uploaded/*
```
tianyu-l: unrelated to this PR: it seems this is called multiple times, maybe just need last time purge?

torchtitan/models/common/attention.py
```diff
@@ -599,9 +636,23 @@ def forward(
         xq, xk, xv = self.wq(x), self.wk(x), self.wv(x)
         # Use -1 instead of n_heads (or n_kv_heads) to infer the
         # actual local heads from sizes as TP may have sharded them.
-        xq = xq.view(bs, seqlen, -1, self.head_dim)
-        xk = xk.view(bs, seqlen, -1, self.head_dim)
-        xv = xv.view(bs, seqlen, -1, self.head_dim)
+
+        def local_qkv_head_split(x):
```
tianyu-l, line 693: I guess we need to do similar things to https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L628

torchtitan/models/common/attention.py
```diff
@@ -207,6 +209,41 @@ def __init__(self, config: Config) -> None:
         super().__init__()
         self.kernel_options = config.kernel_options
 
+    @staticmethod
+    def compiled_flex_attn(
```
tianyu-l, line 250: Have we enabled all tests in features.py on spmd?

Would you need to do something for Varlen attention as well? https://github.com/pytorch/torchtitan/blob/main/tests/integration_tests/features.py#L527

tianyu-l (approved):

torchtitan/components/lora.py
```diff
@@ -25,26 +25,33 @@ def _lora_adapter_sharding(
...
+        lora_b_sharding = ShardingConfig(
+            state_shardings={"weight": base_weight_sharding},
+        )
+        return replicated_weight, lora_b_sharding
+    else:
+        assert base_weight_sharding == dense_param_placement(tp=spmd.S(1))
+        lora_a_sharding = ShardingConfig(
+            state_shardings={"weight": dense_param_placement(tp=spmd.S(1))},
+        )
+        return lora_a_sharding, replicated_weight
```
pianpwk, line 54: @tianyu-l lora feature test has implicit weight redistribution (for rowwise vs. colwise), checking if this makes sense

torchtitan/models/common/attention.py
```diff
@@ -160,29 +162,43 @@ def forward(
...
-        out_BLNH = out_TNH.view(B, L, -1, H).to(q_BLNH.dtype)
+        if get_spmd_backend() == "spmd_types" and spmd.is_type_checking():
+            spmd.assert_type(
+                out_TNH, spmd.V, spmd.PartitionSpec(("dp", "cp"), "tp", None)
+            )
+            spmd.assert_type(
+                lse_NT, spmd.V, spmd.PartitionSpec("tp", ("dp", "cp"))
+            )
 
+        out_BLNH = out_TNH.view(B, L, -1, H).to(q_BLNH.dtype)
```
pianpwk, line 199: @tianyu-l for varlen

torchtitan/models/common/attention.py
pianpwk: added spmd_types backend to features.py . typechecking has known incompatibilities (compile, SAC, PP, etc.), but we test without typecheck

I may need to land something in spmd_types to prop P(sum) / scalar -> P(sum) for non-chunked CE loss though.

tianyu-l: maybe skip spmd on the single non-chunked CE loss for now

torchtitan/models/common/attention.py
```diff
@@ -160,29 +162,43 @@ def forward(
...
+                **varlen_kwargs,
+            )
 
         # varlen_attn returns the packed output (T, N, H), plus the LSE when an
         # out_transform epilogue was requested.
         if out_transform is None:
             assert isinstance(result, torch.Tensor)
+            if get_spmd_backend() == "spmd_types" and spmd.is_type_checking():
+                spmd.assert_type(
+                    result, spmd.V, spmd.PartitionSpec(("dp", "cp"), "tp", None)
```
tianyu-l: Whenever there's any tensor dim sharded as `(dp, cp)`, that's likely incorrect, because it must be coming from folding `[S(0), S(1)]`, which should've resulted in `StridedShard`.

Maybe `no_typecheck()` the entire region for now and leave a TODO.

torchtitan/components/lora.py
tianyu-l: hmm, this makes sense to me. I must have missed the point when I was reviewing it. Curious what redistribution happens.

torchtitan/models/common/attention.py
pianpwk: the folding is needed for VarLen's signature (in the code above), but I just realized VarLen + CP is supported yet, so removing CP from assert_type for now, adding a comment.

torchtitan/protocols/module.py
```diff
@@ -564,11 +564,12 @@ def _redistribute_inputs(
 
                 # SPMD source placements are part of the config contract: assert
                 # before redistributing so typechecking catches placement mismatch.
-                spmd.assert_type(
-                    value,
-                    src_spmd_layout.axis_types,
-                    src_spmd_layout.partition_spec,
-                )
+                if spmd.is_type_checking():
```
tianyu-l, line 568: maybe mention why it doesn't work with compile during erasure, if not having this if/else

PR #3279, opened 2026-05-08
[codex] Fix shared BOS and EOS token inference

bruce2233:
## Summary
- allow a tokenizer special token to be recorded as both BOS and EOS
- add a regression test for tokenizers that use one token for both roles, such as GPT-2-style `<|endoftext|>`

## Root cause
`HuggingFaceTokenizer._process_special_token` used `if/elif` when assigning BOS and EOS roles. If `bos_token` and `eos_token` were the same string, the BOS branch won and `eos_id` stayed unset. Calls to `encode(..., add_bos=True, add_eos=True)` then omitted the EOS token.

## Validation
- `/data/app/torchtitan/.venv/bin/python -m pytest -q tests/unit_tests/test_tokenizer.py::TestTokenizerIntegration::test_same_bos_eos_token_is_inferred_for_both_ids`
- `/data/app/torchtitan/.venv/bin/python -m pytest -q tests/unit_tests/test_dataloader.py`

### DIFF
--- a/tests/unit_tests/test_tokenizer.py
+++ b/tests/unit_tests/test_tokenizer.py
@@ -90,6 +90,30 @@ class TestTokenizerIntegration(unittest.TestCase):
             # No chat template should be set
             self.assertIsNone(tok._chat_template)
 
+    def test_same_bos_eos_token_is_inferred_for_both_ids(self):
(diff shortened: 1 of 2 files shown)


tianyu-l (approved):
sgtm

PR #3281, opened 2026-05-08
[graph_trainer] Fix cpuoff init hang

IvanKobzarev:
● ## Fix CPU offloading init hang: bypass DTensor dispatch for param init

  ### Problem

  When running with `--training.enable_cpu_offload`, model initialization hangs indefinitely during `init_weights`. No error, no traceback — just a silent deadlock. Observed on 8B model (10+ min hang), 70B would be proportionally worse.

  ### Root Cause

  When `enable_cpu_offload=True`, parameters are materialized on CPU as DTensors whose device mesh uses NCCL (CUDA). Two issues compound to cause the deadlock:

  **1. SimpleFSDP parametrization triggers allgather during init**

  `simple_fsdp` wraps each module with a `ReplicateComputation` parametrization. During `_init_self_parameters`, `named_parameters(recurse=False)` triggers the parametrization getter → `replicate_compute()` → allgather on every parameter
  access. For CPU tensors with an NCCL-backed mesh, this allgather deadlocks because NCCL expects CUDA tensors.

  **2. DTensor dispatch hangs for in-place random ops on CPU with CUDA mesh**

  Even with the parametrization disabled, `nn.init.normal_(dtensor)` routes through DTensor's `__torch_dispatch__`, which attempts to coordinate the random initialization across ranks via NCCL collectives. NCCL cannot operate on CPU tensors →
   deadlock.

  ### Investigation

  Added timing instrumentation to isolate the bottleneck:

  - `parallelize_fn`: 0.12s ✓
  - `to_empty(cpu)`: 0.02s ✓
  - `init_weights` layers (32 transformer blocks): 4.91s ✓
  - `init_weights` lm_head (SimpleFSDPLinear, 525M params): **HANG**

  Notably, `tok_embeddings` (same 525M params) succeeded — it runs first before NCCL is fully initialized. By the time `lm_head` runs, NCCL strictly requires CUDA tensors.

  Benchmarked plain CPU `normal_()`: ~42s for the full 8B model, ~5s with tp=8 sharding. The hang is not CPU being slow — it is a deadlock in DTensor/NCCL dispatch.

  ### Fix

  **`protocols/module.py` — init local tensor directly:**

  ```python
  # Before (hangs — DTensor dispatch tries NCCL collectives on CPU):
  self._param_init[name](param)

  # After (bypasses DTensor dispatch entirely):
  from torch.distributed._tensor import DTensor
  init_target = param._local_tensor if isinstance(param, DTensor) else param
  self._param_init[name](init_target)

  Each rank initializes its own local shard independently, which is correct for random initialization of sharded parameters.

  trainer.py — disable parametrization during init:

  # Before (allgather on every param access during init):
  cast(BaseModel, model).init_weights(buffer_device=buffer_device)

  # After (skips allgather):
  from torchtitan.experiments.graph_trainer.simple_fsdp import disable_active_parametrization
  with disable_active_parametrization():
      cast(BaseModel, model).init_weights(buffer_device=buffer_device)

  Result

  8B model init_weights with CPU offloading: 10+ min hang → 9.25s

  Test Plan

  - Verified 8B graph_trainer with --training.enable_cpu_offload completes init_weights in ~9s (was hanging indefinitely)
  - Verified 8B graph_trainer without CPU offloading still works (no regression, init_weights ~0.25s)
  - CI integration tests

### DIFF
--- a/torchtitan/protocols/module.py
+++ b/torchtitan/protocols/module.py
@@ -130,7 +130,19 @@ class Module(nn.Module, Configurable):
                 f"{type(self).__name__}. "
                 f"Available: {list(self._param_init.keys())}"
             )
-        self._param_init[name](param)
+        # When CPU offloading is active, params are CPU DTensors on a CUDA
+        # (NCCL) mesh. DTensor dispatch for in-place random ops (normal_,
+        # uniform_) tries NCCL collectives on CPU tensors, causing a deadlock.
+        # Init the local shard directly to bypass DTensor dispatch.
+        from torch.distributed._tensor import DTensor
+        if (
+            isinstance(param, DTensor)
+            and param.device.type == "cpu"
+            and param.device_mesh.device_type == "cuda"
+        ):
+            self._param_init[name](param._local_tensor)
+        else:
+            self._param_init[name](param)
 
     def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> None:
         """Initialize this module's own buffers.
--- a/torchtitan/trainer.py
+++ b/torchtitan/trainer.py
@@ -385,9 +385,22 @@ class Trainer(torch.distributed.checkpoint.stateful.Stateful, Configurable):
 
                 model.to_empty(device=init_device)
                 with torch.no_grad():
-                    # TODO: Change this back to init_weights once
-                    # autoparallel contains the wrap_init_states
-                    cast(BaseModel, model).init_weights(buffer_device=buffer_device)
+                    if config.training.enable_cpu_offload:
+                        # Disable simple_fsdp parametrization during init to
+                        # avoid triggering allgather on every parameter access.
+                        # With CPU offloading, params are CPU tensors on an NCCL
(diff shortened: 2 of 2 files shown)


torchtitan/protocols/module.py
```diff
@@ -130,7 +130,19 @@ def _init_param(self, name: str, param: nn.Parameter) -> None:
                 f"{type(self).__name__}. "
                 f"Available: {list(self._param_init.keys())}"
             )
-        self._param_init[name](param)
+        # When CPU offloading is active, params are CPU DTensors on a CUDA
+        # (NCCL) mesh. DTensor dispatch for in-place random ops (normal_,
+        # uniform_) tries NCCL collectives on CPU tensors, causing a deadlock.
```
SherlockNoMad, line 135: NCCL collectives on CPU tensors 

this sounds like a bug on cpu_offload_pass, 

do you have a tlparse for the hanging run, we can look at the before and after graph to debug

PR #3304, opened 2026-05-08
[RFC] Multi-turn chat dataset

### LINKED ISSUE #2757: [RFC] Multi-turn SFT

felipemello1:
https://github.com/pytorch/torchtitan/pull/2769 raised some questions on how titan should handle multi-run chat dataset. I did some research with claude/codex to see how other libraries handle it and propose a path for us.

This RFC has two docs:
1) RFC: Bulk of the discussion
2) Appendix: Some extra info that I didn't really spend time on

Please take a look at the rfc.md and check if it makes sense, so we can follow through with PR 2769

# TLDR

### Axis A — how libraries build the loss mask
- **A1 — Span recovery:** render the full conversation; recover per-message token spans by re-rendering prefixes (with a fallback for templates that rewrite history).
- **A2 — HF generation-mask:** Jinja's `{% generation %}` tag plus `apply_chat_template(return_assistant_tokens_mask=True)` returns a per-token mask directly.
- **A3 — Final-only prompt subtraction:** tokenize prompt vs full conversation; supervise only the suffix.
- **A4 — Renderer-emitted spans:** a Python class returns tokens with per-message attribution as it writes them.

###  Axis B — who owns the chat format?
- **B1 — User edits the model's stock chat template** (manual Jinja editing).
- **B2 — Library ships a training template** (one or more Jinja files maintained by the framework).
- **B3 — Library bypasses Jinja entirely** (Python renderer/tokenizer code IS the format).

###  C - one row or branched rows?
- **C1 SINGLE:** one conversation → one row.
- **C2 PER_ASSISTANT:** one conversation → one row per supervised assistant turn.

###  Decision:
**V1 default path = A1 + B1 + C1 + AS_RENDERED.** Render with the model's stock HF chat template, recover assistant spans and keep one row per conversation.

**Long-term renderer path = A4 + B3 + C1.** A Python renderer emits tokens and token ownership together. It can run as-rendered, preserve reasoning, or branch per assistant turn as an explicit policy choice.


docs/rfcs/multi_turn_chat_sft/1-rfc.md
```diff
@@ -0,0 +1,542 @@
...
+The same renderer with strip_thinking_from_history=True would match A1's
+inference-shape output above.
+```
+### Axis A recap
+
+```text
+A1 span recovery       — works with any HF template; pays O(n_turns) extra renders.
+A2 HF generation-mask  — fast when usable; 0/8 stock templates ship {% generation %}.
+A3 final-only          — simplest; only trains the last assistant.
+A4 renderer-emitted    — most precise; requires per-model Python code.
```
tianyu-l, line 338: For A4, where is the per-model work coming from?

docs/rfcs/multi_turn_chat_sft/1-rfc.md
```diff
@@ -0,0 +1,542 @@
+# Multi-turn SFT chat datasets
+## TL;DR
+
+This RFC decides how torchtitan should turn a multi-turn `messages` list into `(input_ids, labels)` for SFT.
+
+- **Data contract:** `messages + keep_loss`, where `keep_loss[i]` says whether message `i` should get cross-entropy loss.
+- **V1 algorithm:** render with the model's stock chat template, recover per-message token spans by incremental rendering, and fall back to message-payload diff when the template rewrites history.
```
tianyu-l, line 7: What is "stock inference" / "stock chat template"?

During multi-turn inference, is the default behavior stripping off <think> traces? If so, is it to save context length?

docs/rfcs/multi_turn_chat_sft/1-rfc.md
```diff
@@ -0,0 +1,542 @@
...
+
+- **Label decision:** which token positions get cross-entropy loss?
+- **Context decision:** which prior tokens are visible during the forward pass?
+
+Concretely, when computing loss on ASST_2, does the forward pass see ASST_1's `<think>` or not? The answer depends on the chat template:
+
+```text
+AS_RENDERED  (Qwen3 / DeepSeek-R1 / Nemotron 3 / gpt-oss stock templates):
+  input  = [USER_1] [ASST_1_WITHOUT_THINK] [USER_2] [ASST_2_WITH_THINK]
+  loss   =    0              1                0              1
```
tianyu-l, line 68: Let's say if I'm at turn N, and the history is
`<user> <assistant-without-think> <user> <assistant-without-think> ... <user>`
which seems what the model can see before replying -- why would the last assistant having think trace? Is thinking not the job of SFT? I think again, it depends on what would happen in inference time.

tianyu-l (commented):
Thanks a lot for the comprehensive analysis and proposal! I found it to be really, really helpful to understand the context.

I think the decision should be made based on
- most importantly, matching distribution between training and inference. From current RFC, it's not entirely clear to me what happens in real inference scenario, and I've asked some inline questions. (That said, I'm totally onboard with the idea of gradually getting there with easy-to-execution short-term plan.)
- secondly, sharing the infrastructure between SFT and RL, including features we want to enable in the future (tool-calling, etc.)

docs/rfcs/multi_turn_chat_sft/1-rfc.md
slimfrkha: stock chat template is the one found in the HF hub.
Because HF hub artefact is destined for inference, its jinja usually strips out previous turn thinking traces to save KV cache (each thinking trace is usually extremely long).

docs/rfcs/multi_turn_chat_sft/1-rfc.md
felipemello1: > why would the last assistant having think trace?

The last assistant have thinking trace because it was the last produced (not stripped yet). The stock Jinja (HF standard) finds the latest <user> input and strips every thinking before that.

> Is thinking not the job of SFT?

It is -- SFT should train on thinking traces. Thats why this is a footgun for our users to just let them user the jinja as it is (AS_RENDERED). Ideally we would have a dedicated renderer or jinja per model to handle it. There are two major ways to handle it:

a) PER_ASSISTANT: branch on every assistant turn -- O(k) turns
b) PRESERVE_WITH_TEMPLATE: modify the stripping behavior and leave the thinking trace on every turn -- train <> inference mismatch

It seem that most of the libraries prefer (b) for the efficiency gains.

haydn-jones:
@felipemello1 New library from primeintellect might be interesting to look at: https://www.primeintellect.ai/blog/renderers

PR #3308, opened 2026-05-08
Remove MoE expert for-loop fallback

sanketpurandare:
torch._grouped_mm already provides a CUDA fallback path when the fused grouped GEMM kernel is unavailable, including on pre-SM90 hardware. Keeping a separate Python for-loop expert implementation duplicates that fallback, carries an extra configuration branch, and makes MoE behavior diverge across models.

Use the grouped-mm path unconditionally and rely on PyTorch to choose either the fused kernel or its built-in loopy fallback.

### DIFF
--- a/torchtitan/experiments/graph_trainer/graph_utils.py
+++ b/torchtitan/experiments/graph_trainer/graph_utils.py
@@ -75,8 +75,6 @@ def export_joint(
     )
     with coor_ctx:
         with (
-            # TODO Investigate error on MOE model with use_grouped_mm=False.
-            # For repro, see: https://gist.github.com/zhxchen17/d794ff58236243d9faddf713b9fc6a61
             torch._dynamo.config.patch(fake_tensor_cache_enabled=False),
             torch.fx.traceback.preserve_node_meta(),
         ):
--- a/torchtitan/models/common/moe.py
+++ b/torchtitan/models/common/moe.py
@@ -20,66 +20,12 @@ from torchtitan.protocols.module import Module
 from .token_dispatcher import LocalTokenDispatcher
 
 
-# NOTE: keeping this for-loop implementation for comparison
-#       and readability, may remove later
-def _run_experts_for_loop(
-    w1: torch.Tensor,
-    w2: torch.Tensor,
-    w3: torch.Tensor,
-    x: torch.Tensor,
-    num_tokens_per_expert: torch.Tensor,
-) -> torch.Tensor:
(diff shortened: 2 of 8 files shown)


torchtitan/models/common/moe.py
```diff
@@ -20,39 +20,6 @@
...
-        h = h * torch.matmul(x_expert, w3[expert_idx].transpose(-2, -1))
-        h = torch.matmul(h, w2[expert_idx].transpose(-2, -1))
-        # h shape (tokens_per_expert(varying), dim)
-        out_experts_splits.append(h)
-    out = torch.cat(out_experts_splits, dim=0)
-
-    return out
-
-
 def _run_experts_grouped_mm(
```
tianyu-l: let's put this back inline, no need to have a separate function any more

torchtitan/models/gpt_oss/moe.py
```diff
@@ -50,45 +50,6 @@ def swiglu(x, alpha: float = 1.702, limit: float = 7.0):
...
-        # Apply custom autograd function to scale bias in forward but not in backward
-        b2 = ScaleBiasForward.apply(mlp2_bias[expert_idx], tp_degree)
-        h = torch.matmul(h, mlp2_weight[expert_idx].transpose(-2, -1)) + b2
-        out_experts_splits.append(h)
-    out = torch.cat(out_experts_splits, dim=0)
-
-    return out
-
-
 def _run_experts_grouped_mm(
```
tianyu-l: same

sanketpurandare: done

torchtitan/models/common/moe.py
sanketpurandare: done

torchtitan/experiments/graph_trainer/graph_utils.py
```diff
@@ -75,8 +75,6 @@ def export_joint(
     )
     with coor_ctx:
         with (
-            # TODO Investigate error on MOE model with use_grouped_mm=False.
-            # For repro, see: https://gist.github.com/zhxchen17/d794ff58236243d9faddf713b9fc6a61
```
SherlockNoMad, line 79: sign off for graph trainer change.

SherlockNoMad (approved):
sign off for graph trainer change, didn't review others.

tianyu-l (approved):

PR #3313, opened 2026-05-10
Fix loss logging when TP + CP enabled

### LINKED ISSUE #3280: Qwen3Moe: TP+CP loss discrepancy with TP and CP only.
I am currently playing with Qwen3Moe. 
I see that running on 4 gpus, DP=1, EP=1, SP=False,
the loss CrossEntropyLoss (not the most recent Chunked one) 
with TP=4 or CP=4 (with or without loss compile) are reasonably compatible. 
While TP=2 AND CP=2 activate together produce a loss that's about half of the previous one. 

I was wondering if there might be some issue with parallel loss aggregation due to the two parallelism levels. 
Attached image (highest losses are TP=4 and CP=4, the lower one is TP=2 and CP=2). 

Image

wwwjn:
Fix #3280 

## Root cause
When TP + CP is enabled CP is the not in DTensor's mesh and DTensor mesh only have "tp". So `_dist_reduce` in `torchtitan/distributed/utils.py` directly return without reduction on CP mesh. 
```
if isinstance(x, DTensor):
      return float(x.full_tensor().item())  # ← drops the caller's `mesh`
 return float(funcol.all_reduce(x, group=mesh).item())  # plain branch correctly uses mesh
```

## Fix
- Add a explicit reduction on CP  mesh 
- The meaning of function `_dist_reduce` is vague: When CP is enabled, it doesn't being represented in DTensor mesh. 
    -  I tried alternative method: eg, don't return from the DTensor path, continue reduction on mesh. This is very error-prone because the `mesh` has assumption: `mesh` is flattened to be 1D. It's very easy to double-reduce on DTensor's mesh and `mesh` parameter

## Why it only happen when TP x CP?
- When TP enabled, the `loss` is a DTensor, which will hit the DTensor path

### DIFF
--- a/torchtitan/components/validate.py
+++ b/torchtitan/components/validate.py
@@ -324,6 +324,14 @@ class Validator(BaseValidator):
             global_avg_loss = dist_utils.dist_sum(
                 loss, parallel_dims.get_optional_mesh("loss")
             )
+            # When TP is enabled, ``loss`` is a TP DTensor and ``dist_sum``
+            # only reduces over the DTensor's TP mesh; the CP axis of
+            # ``loss_mesh`` was dropped. Reduce over CP explicitly here.
+            if parallel_dims.tp_enabled and parallel_dims.cp_enabled:
+                cp_mesh = parallel_dims.get_mesh("cp")
+                global_avg_loss = dist_utils.dist_sum(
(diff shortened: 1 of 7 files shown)


wwwjn:
Close because of: https://github.com/pytorch/torchtitan/pull/3159/changes#r3214339904

PR #3314, opened 2026-05-11
[rl] Add optional chunked logprobs to reduce peak memory

DamianSzwichtenberg:
`compute_logprobs` now uses `F.cross_entropy(reduction="none")`] (replacing the previous `log_softmax + gather`), following the same pattern as [forge's compute_logprobs](https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L79-L112). It can optionally process in chunks along the sequence dimension, reducing the peak float32 intermediate from `O(B * S * V)` to `O(B * chunk_size * V)`. Opt-in via the `logprobs_chunk_size` config field (default: None = no chunking).

The chunked path is bit-for-bit identical to the non-chunked path because cross-entropy reduces over vocab independently per position — chunking along the sequence axis does not change the computation. Loss, reward, and logprob diffs are identical across all 10 training steps + pre/post validation, with and without chunking.

Usage: set `logprobs_chunk_size` in PolicyTrainer.Config:

    trainer=PolicyTrainer.Config(
        logprobs_chunk_size=256,
        ...
    )


.github/workflows/integration_test_8gpu_rl_h100.yaml
```diff
@@ -92,6 +92,10 @@ jobs:
         HF_ASSETS_PATH="$MODEL_PATH" torchrun --nproc-per-node=2 -m pytest \
           torchtitan/experiments/rl/tests/test_bitwise_parity.py -v
 
+        # Run compute_logprobs unit tests (CPU-only, no torchrun needed)
+        python -m pytest \
```
wwwjn: This yaml file is for GPU test, I think it's worth adding an separate CPU unit test yaml file to run RL CPU unit tests

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -21,10 +33,32 @@ def compute_logprobs(logits: torch.Tensor, token_ids: torch.Tensor) -> torch.Ten
...
+        log_probs = F.log_softmax(shift_logits.float(), dim=-1)
+        return log_probs.gather(2, shift_targets.unsqueeze(-1)).squeeze(-1)
+
+    # Chunked log_softmax + gather to avoid materializing full [seq, vocab] fp32
+    seq_len = shift_logits.shape[1]
+    chunks = []
+    for start in range(0, seq_len, chunk_size):
+        end = min(start + chunk_size, seq_len)
+        chunk_logits = shift_logits[:, start:end, :].float()
+        chunk_lp = F.log_softmax(chunk_logits, dim=-1)
```
wwwjn: Chunking might be problematic when loss_parallel is enabled and logits are shard on vocab dimension, calling log_softmax might issue one collective per chunk. In RL, now we don't enable loss_parallel which is good, and we can probably adding a guard asserting if logits are DTensor, it has Replicate() placement?

wwwjn (commented):
Thanks, this is a nice optimization

torchtitan/experiments/rl/actors/utils.py
DamianSzwichtenberg: Good catch! Added a placement guard (just for chunking path).

.github/workflows/integration_test_8gpu_rl_h100.yaml
DamianSzwichtenberg: Done.

torchtitan/experiments/rl/actors/utils.py
```diff
@@ -8,23 +8,68 @@
...
+        # keeping grad_fn intact.
+        log_probs = F.log_softmax(shift_logits.float(), dim=-1)
+        return log_probs.gather(2, shift_targets.unsqueeze(-1)).squeeze(-1)
+
+    # Chunked log_softmax + gather to avoid materializing full [seq, vocab] fp32
+    # Guard: chunking with vocab-sharded logits would issue one collective per chunk
+    if dtensor_placements is not None:
+        from torch.distributed.tensor.placement_types import Replicate
+
+        if not all(isinstance(p, Replicate) for p in dtensor_placements):
```
wwwjn: nit: you can use p.is_replicate() or p.is_shard() instead, which is simpler and clearer

.github/workflows/unit_test_cpu_rl.yaml
```diff
@@ -0,0 +1,35 @@
+name: RL CPU Unit Test
+
+on:
+  push:
+    branches: [ main ]
```
wwwjn: https://github.com/pytorch/torchtitan/commit/8599818f813ae43d73aeb142a879cbbff9b1d9ad#diff-5352d6248ca962635fea389024f074b8c3b06dc252ff809fcb841ae9f34477c4R10-R38 Can you switch to triggered by `rl` label here?

.github/workflows/unit_test_cpu_rl.yaml
wwwjn, line 1: cc @felipemello1  about RL CPU tests, we can put it here

felipemello1 (commented):
two high level notes, please see if you guys agree:
1. I think that we should leverage the same CrossEntropy loss used in Torchtitan. We probably just need to pass reduce=None, to get per token logprobs, instead of reduce="sum".

2. For the CI, it seems that for every new unit test file, we have to append it to the yaml. Is that true? Cant we just run it on all tests under /tests instead, and maybe have a special folder for GPU/integration tests or a decorator for GPU tests?

.github/workflows/unit_test_cpu_rl.yaml
DamianSzwichtenberg: Done.

torchtitan/experiments/rl/actors/utils.py
DamianSzwichtenberg: Fixed.

DamianSzwichtenberg:
> two high level notes, please see if you guys agree:
> 
> 1. I think that we should leverage the same CrossEntropy loss used in Torchtitan. We probably just need to pass reduce=None, to get per token logprobs, instead of reduce="sum".
> 2. For the CI, it seems that for every new unit test file, we have to append it to the yaml. Is that true? Cant we just run it on all tests under /tests instead, and maybe have a special folder for GPU/integration tests or a decorator for GPU tests?

1. Good point! In my opinion, `cross_entropy`s flattened `[batch*seq, vocab]` shape requirements would make sequence-based chunking more complex and less intuitive compared to the current `log_softmax` approach on `[batch, seq, vocab]`.
2. You're absolutely right - we currently have to add each test file explicitly. I followed the pattern from `unit_test_cpu_torchft.yaml` which lists specific files. There's an alternative approach used in the main repo: `unit_test_cpu.yaml` runs the full [unit_tests](https://github.com/pytorch/torchtitan/tree/main/tests/unit_tests) directory, and GPU tests self-filter using `@unittest.skipUnless(torch.cuda.is_available())` (example: [test_compile_moe.py#L45](https://github.com/pytorch/torchtitan/blob/main/tests/unit_tests/test_compile_moe.py#L45)).

felipemello1:
1. Check how i did it in forge
compute_logprobs basically calls cross_entropy: https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L79-L112

and then the loss calls it: https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/grpo.py#L78

You should NOT have to refactor the entire loss to look like forge's, but the point is: we can and should call `cross_entropy` directly, because its highly optimized when compiled

2.
> There's an alternative approach used in the main repo: unit_test_cpu.yaml runs the full [unit_tests](https://github.com/pytorch/torchtitan/tree/main/tests/unit_tests) directory, and GPU tests self-filter using @unittest.skipUnless(torch.cuda.is_available()) (example: [test_compile_moe.py#L45](https://github.com/pytorch/torchtitan/blob/main/tests/unit_tests/test_compile_moe.py#L45)).

This sounds good, i think we can create a decorator so this check is shorter, e.g. @gpu_test(n) that are used for tests that need gpu. Check torchtune: https://github.com/meta-pytorch/torchtune/blob/bd2a0fc7c31430972728494fa01aaeeb0ebf1ba1/tests/recipes/test_eleuther_eval.py#L58

Thats probably better, since most tests do NOT need gpu

---

Let me know if you feel comfortable tackling them as separate PRs -- unit tests first would be better

DamianSzwichtenberg:
> 1. Check how i did it in forge
>    compute_logprobs basically calls cross_entropy: https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L79-L112
> 
> and then the loss calls it: https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/grpo.py#L78
> 
> You should NOT have to refactor the entire loss to look like forge's, but the point is: we can and should call `cross_entropy` directly, because its highly optimized when compiled
> 
> 
> > There's an alternative approach used in the main repo: unit_test_cpu.yaml runs the full [unit_tests](https://github.com/pytorch/torchtitan/tree/main/tests/unit_tests) directory, and GPU tests self-filter using @unittest.skipUnless(torch.cuda.is_available()) (example: [test_compile_moe.py#L45](https://github.com/pytorch/torchtitan/blob/main/tests/unit_tests/test_compile_moe.py#L45)).
> 
> This sounds good, i think we can create a decorator so this check is shorter, e.g. @gpu_test(n) that are used for tests that need gpu. Check torchtune: https://github.com/meta-pytorch/torchtune/blob/bd2a0fc7c31430972728494fa01aaeeb0ebf1ba1/tests/recipes/test_eleuther_eval.py#L58
> 
> Thats probably better, since most tests do NOT need gpu
> 
> Let me know if you feel comfortable tackling them as separate PRs -- unit tests first would be better

@felipemello1 Thanks for the feedback! Just to clarify the scope:

Should I:

- include test improvements in this PR - Add the @gpu_test decorator and auto-discovery to the workflow alongside the chunked logprobs feature, then tackle cross_entropy in a follow-up PR?
- split into separate PRs - First PR just adds test infrastructure improvements to existing RL tests, then this chunked logprobs with cross_entropy (with my new tests) benefits from that infrastructure?

Happy to work on this either way.

felipemello1:
@DamianSzwichtenberg , i will open a PR today to address the unit test infrastructure. You should **not** have to touch any yaml after that. Do you want to focus on the chunked CE portion? 

To summarize, we would need to delete our "compute_logprobs" and leverage ChunkedCrossEntropy in torchtitan with reduce=None to get per-token logprobs

Let me know if that makes sense / you agree.

DamianSzwichtenberg:
@felipemello1 Got it, will implement chunked CE. Thanks 😉

DamianSzwichtenberg:
I've updated `compute_logprobs` to use `F.cross_entropy(reduction="none")` directly, following the same pattern as [forge's compute_logprobs](https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L79-L112).

I investigated reusing `ChunkedCELoss`, in my opinion it does not fit here; it takes hidden states (not logits), returns a scalar loss (not per-token logprobs), and its per-chunk backward + custom autograd reconnection assume the loss is computed inside the chunk loop — while the RL loss pipeline extracts variable-length response slices before computing the loss, making per-chunk backward non-trivial to align. Adapting it would require a new code path sharing no logic with the existing one beyond the `cross_entropy` call.

Waiting on the PR that fixes RL tests.

wwwjn:
For the tests, I agree we should add GPU unit tests for core and RL. In experiement folder we have some examples: https://github.com/pytorch/torchtitan/blob/cee49826b9a071b2254c1573588d35bfb9a1668a/.github/workflows/integration_test_8gpu_graph_trainer.yaml#L7, looking for the refactor.  cc @felipemello1 

For ChunkedCELoss, I think this overall idea is the same, but need more refactor of PolicyTrainer, Concretely:
>  it takes hidden states (not logits), returns a scalar loss (not per-token logprobs)
https://github.com/pytorch/torchtitan/blob/d4a3c63499dbeb7442a29e2bc5a0a3590af3bbdb/torchtitan/trainer.py#L410 We should set the skip_lm_head in PolicyTrainer.model,  which will make

Then remove compute_logits() and replace it with ChunkedCELoss to only compute logits (which is not doable today), and then call GRPOLoss.

> while the RL loss pipeline extracts variable-length response slices before computing the loss,

This is avoidable. We need this "slicing" because we used seq_lens, prompt_lens, response_lens in our dataflow, and we use the "length" to slicing out the response. Alternatively we can use `reponse_masks` which contains the same information.

The real difficulty I think is CE loss is baked into ChunkedCELoss today, and it will call forward then **backward**  per chunk. We don't want to backward using CEloss in RL, but we want to use the chunked mechanism + GRPOloss. We'll need to either make ChunkedCELoss more flexible, or have a ChunkedGRPOLoss

DamianSzwichtenberg:
@wwwjn Thanks for the feedback on ChunkedCELoss—I’ll definitely dig into that extension.

Since the current PR is already validated and self-contained, I’d like to merge this first to keep the changes atomic and avoid further scope creep.

Regarding the ChunkedCELoss follow-up, would you prefer I start with an RFC to align on the design, or should I go straight to an implementation PR for discussion?"

tianyu-l:
> The real difficulty I think is CE loss is baked into ChunkedCELoss today, and it will call forward then backward per chunk. We don't want to backward using CEloss in RL, but we want to use the chunked mechanism + GRPOloss.

@wwwjn it should already work with `requires_grad=False`. I didn't follow closely how backward is done in current RL loop, but maybe we could call `compute_logprobs` in `torch.no_grad` mode?

https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/loss.py#L288

wwwjn:
> but maybe we could call `compute_logprobs` in `torch.no_grad` mode?

I think we can use the ChunkedCELoss class with torch.no_grad to achieve the same functionality as this PR (chunked forward of`compute_logprobs()`)

However, I was worried about the the backward. Today's RL loss flow is:
- log_probs = compute_logprobs(logits)
- loss = GRPOLoss(log_probs)

If we pass hidden_states.requires_grad = False, then the log_probs will not have a proper gradient. Then the loss won't backprop properly?

felipemello1:
> If we pass hidden_states.requires_grad = False, then the log_probs will not have a proper gradient. Then the loss won't backprop properly?

thats my understanding as well

I think we would first need to understand why the backward is called within the loss. Is it because we need to release the memory after each chunk, or is it because of the fsdp hooks related to the `lm_head`?

If its because of the `lm_head`, maybe we can have `ChunkedCE` (without lm_head) and `FusedChunkedCE`.

If its because of the memory, then its more complicated. Perhaps do something like:

`ChunkLoss(fn=my_lr_loss)`

But i wonder if we reinventing the wheel and this is already a solved problem in other RL frameworks/Liger, and we can just take a look at how they do it, lol.

tianyu-l (changes requested):
It seems to me that the GRPO loss for RL and cross-entropy loss for SFT both decompose into two steps: cross-entropy (reduction=None) -> reduction. They are different in that
- In SFT loss, we can do sequence-dim chunked forward+backward, because the reduction step is independent across sequence dimension.
- In GRPO, the reduction step has ratio computation which needs info across individual sequences (also seen in https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L193)

Not surprisingly, this PR is trying to do chunked fwd on the cross-entropy without reduction, and is not doing bwd.
- Not sure if there's a clean way to merge the implementation of `ChunkedCELoss`, which really focuses on the handling of both fwd and bwd. One way is to add if-else branching to omit the bwd part, but IMO it's not proper to call it a "Loss" any more.
- If instead of seq-dim, we do chunking on the batch-dim, then I do see the implementation being much closer. However, for small local batch size (e.g. 1), there could be nothing to chunk.

There is room to improve this PR though.
- Right now we still materialize the full logits (B, S, vocab_size).
- I feel we can call `model` fwd with `skip_lm_head=True` and do the `lm_head` computation inside in chunked fashion, to save more peak memory.

I'm putting a hold on this PR for more discussions.

felipemello1:
I spent the day on it yesterday. My tested solution is to pass a `chunk_loss_fn` to the ChunkedCrossEntropy that 
a) does the chunking (so it knows how to compute the masks, weights, normalization, etc)
b) call the loss on the chunk AFTER the CE and BEFORE the backwards.

```
chunked_token_loss(
    hidden_states,
    lm_head,
    labels,
    chunk_loss_fn,
) -> scalar
```

so something like:
```
class TitanCoreChunkedCELoss(..., chunk_loss_fn)
	chunks = chunk_loss_fn.get_chunks(inputs)
	for chunk in chunks:
		logprobs = do_CE_loss()
	    loss = chunk_loss_fn.loss_chunk(logprobs)
	    loss.backwards()
```

For the pure pre-training case, `chunk_loss_fn.loss_chunk(logprobs)` is a dummy `sum`.

That way, the titan CE loss stays the exact same, but it works with any RL loss.

I will have a PR to showcase it. Perf is the same, and the design is not intrusive.

tianyu-l:
@felipemello1 
> call the loss on the chunk AFTER the CE and BEFORE the backwards

Since we will do packing in the batch, and a single original sequence could be spanning multiple chunks, I'm not sure if you can always compute the sequence-wise GRPO loss within a chunk, per my comment
> In GRPO, the reduction step has ratio computation which needs info across individual sequences (also seen in https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L193)

wwwjn:
> @felipemello1
> 
> > call the loss on the chunk AFTER the CE and BEFORE the backwards
> 
> Since we will do packing in the batch, and a single original sequence could be spanning multiple chunks, I'm not sure if you can always compute the sequence-wise GRPO loss within a chunk, per my comment
> 
> > In GRPO, the reduction step has ratio computation which needs info across individual sequences (also seen in https://github.com/meta-pytorch/torchforge/blob/main/src/forge/rl/loss/ops.py#L193)

The math makes sense to me, we will need the trainer's forward logprobs of the whole sequence. So it makes it even harder to reuse ChunkedCELoss code which assume the loss is calculated on per-token level. We should invent a new chunked loss class for these kind of Forward-only-chunk

The best chunk-style loss in RL I could think is: 
 1. Run trainer model forward in chunked manner (chunked lm_head) to get per-token logprob
2. Don't backward and just collect the per-token logprobs
3. Sum per-sequence to get sequence level ratio R
4. Compute the GRPO loss from R and advantage
6. Then backward


On caveat today: in torchtitan RL today, we are not using sequence level ratio which is not strictly GRPO.

tianyu-l:
@wwwjn 

> On caveat today: in torchtitan RL today, we are not using sequence level ratio which is not strictly GRPO.

I think we are
https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/grpo.py#L106

felipemello1:
> nce level ratio which is not strictly GRPO.

This is wrong -.-`. We should be doing exp(trainer_logprobs - generator_logprobs), and if they are the same (on-policy -- which is our case), then we have exp(0)=1. But instead, we just do log(trainer_logprobs), which is something that is not 1 or a constant per tokens. But it will get fixed. Let me see if i can have a small-ish RFC/PR tomorrow to share for you guys to review

DamianSzwichtenberg:
Based on your feedback, I think we all agree: the GRPO loss requires computing the mean over the full sequence, which means we cannot decompose the loss computation per chunk and call backward immediately. This is a fundamental difference from ChunkedCELoss.

In my experiments with ChunkedCELoss, moving backward outside of the chunked loop led to memory corruption (RuntimeError: storage of size 0). In my understanding, the implementation balances communication and memory usage as follows:

We disable resharding before the chunk loop, which reduces the number of forward all-gathers from N to 1
We call backward per chunk, which frees the intermediate logits immediately
After all chunks complete, we manually reshard the weights
The key issue is: if we move backward outside the loop and call reshard() before it, autograd will have a reference to tensors that were freed (due to resharding), causing the memory corruption error.
(I'm talking about this [code](https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/loss.py#L410))

For RL, I merged the LM head with logprob computation by checkpointing the LM head using torch.checkpoint.checkpoint(..., use_reentrant=False). This breaks the autograd graph's reference to freed storage, allowing us to reshard safely. The trade-off is that checkpointing costs N backward all-gathers during recomputation. 

Any thoughts on the best path forward?

felipemello1:
@DamianSzwichtenberg , thanks for helping us with this!!

i let claude/opus do some work and i have a couple of versions:
a) One that calls .backward on every chunk
b) One that concatenates chunks (like discussed), do RL loss, then run backwards

Most RL losses work with independent chunks, as they are averaged per token. GSPO averages per sequence and doesn't work with (a). For maximum perf, we would support both ways, but we could simplify and go just with (b).

The thing is: the code is ugly,  but i know that it runs with multiple parallelisms and has parity with grad_norm, loss, tok/s.

I don't have bandwidth at the moment to go much deeper into it, so, if you are willing to help, I am thinking about these steps:
1. I do a quick review myself and clean it up a bit
2. Ask you to take a look, salvage the ideas that are good and work towards bringing us to a great PR

Would you feel comfortable with that?

DamianSzwichtenberg:
@felipemello1 Sounds good, happy to help. Let me know when it's ready for review.

DamianSzwichtenberg:
@felipemello1 Any updates 😉 ?

felipemello1:
@DamianSzwichtenberg 

here it is: https://github.com/pytorch/torchtitan/pull/3477

I am sorry for the delay, and thanks for staying interested in it!

PR #3318, opened 2026-05-11
[rl] Better initial weight loading

wwwjn:
As titled, update the trainer and generator to use bot checkpointManager

### DIFF
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -12,13 +12,6 @@ from typing import Any
 import torch
 import torchstore as ts
 from monarch.actor import Actor, current_rank, endpoint
-from torch.distributed.checkpoint.state_dict import (
-    get_model_state_dict,
-    set_model_state_dict,
-    StateDictOptions,
-)
-
-from monarch.actor import Actor, endpoint
 from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
--- a/torchtitan/experiments/rl/actors/trainer.py
+++ b/torchtitan/experiments/rl/actors/trainer.py
@@ -159,6 +159,8 @@ class PolicyTrainer(Actor, Configurable):
             training_steps=config.training.steps,
         )
 
+        self.policy_version = 0
+
         # Always build CheckpointManager; enable is a field on the config.
         # When enable=False (CI/debug), load() is a no-op and random init stands.
         self.checkpointer = config.checkpoint.build(
@@ -166,7 +168,7 @@ class PolicyTrainer(Actor, Configurable):
             model_parts=self.model_parts,
             optimizers=self.optimizers,
             lr_schedulers=self.lr_schedulers,
-            states={},
+            states={"train_state": self},
             sd_adapter=self.sd_adapter,
             base_folder=config.dump_folder,
         )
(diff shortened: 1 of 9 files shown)


torchtitan/experiments/rl/config_registry.py
```diff
@@ -165,50 +187,6 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
     )
 
 
-def rl_grpo_qwen3_debug() -> RLTrainer.Config:
```
wwwjn, line 168: Remove as it's not useful. Random init a model can not learn anything

torchtitan/experiments/rl/models/vllm_registry.py
```diff
@@ -129,6 +130,7 @@ def registry_to_vllm(
     *,
     parallelism: ParallelismConfig,
     compile_config: CompileConfig,
+    checkpoint_config: CheckpointManager.Config | None = None,
```
tianyu-l: I personally prefer this to not be optional and be explicit

torchtitan/experiments/rl/actors/trainer.py
```diff
@@ -145,6 +143,24 @@ def __init__(
...
         )
 
+        # Always build CheckpointManager; enable is a field on the config.
+        # When enable=False (CI/debug), load() is a no-op and random init stands.
+        self.checkpointer = config.checkpoint.build(
+            dataloader=None,
+            model_parts=self.model_parts,
+            optimizers=self.optimizers,
+            lr_schedulers=self.lr_schedulers,
+            states={},
```
tianyu-l: should save `self.policy_version`?

wwwjn: Now our usage of Checkpoint manager is minimal - just use checkpointManager to load initial weight. Policy version should be used for save and resume RL training, which we haven't dig into the plan yet. 

I think to resume the RL training, PolicyTrainer should just save policy_version which is enough (We can add more field if needed later)

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -370,62 +368,37 @@ def compute_logits(
...
-            hf_assets_path=None,
+        # Model-only CheckpointManager: initial_load_model_only=True (default)
+        # ensures only MODEL state is loaded, so None optimizer/lr_scheduler
+        # are never accessed.
+        checkpointer = cfg.build(
+            dataloader=None,
+            model_parts=[self.model],
+            optimizers=None,
+            lr_schedulers=None,
+            states={},
```
tianyu-l, line 398: I think you need `self` to utilize what you put in `state_dict`
https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L454

tianyu-l (approved):

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: I'm a little bit confused here because I haven't design the Generator's checkpoint & resume. 

The trainer put https://github.com/pytorch/torchtitan/blob/main/torchtitan/trainer.py#L870 these fields in the trainer's checkpoint. We didn't put anything in generator's checkpoint yet IIUC
- Moreover, should generator be stateless? By "stateless", I mean does generator needs to know current train_steps? Or what information should generator get when it resume from a checkpoint

tianyu-l: sorry I said nonesense.

I meant trainer should have it, which you already did https://github.com/pytorch/torchtitan/pull/3318/changes#diff-73ec9f285034fd742cd8e14135f7e3bbfd22f091d9e36452dd41eb210f060220R155

generator shouldn't be checkpointed at all I think

PR #3321, opened 2026-05-12
[graph_trainer] Improve SAC tagging and CPU offload pass metadata

SherlockNoMad:
## Summary
- Rename `apply_sac_pass` to `tag_sac_policy` to clarify it only tags nodes (not the remat transform)
- Skip `torch.ops.device_mesh._get_submesh` in SAC tagging (metadata-only op, no tensor output)
- Propagate source node metadata (stacktrace, module_fqn, seq_nr) to CPU offload chain nodes for better tlparse/graph dump readability
- Remove unused `cpu_offload_reload_node` metadata (remat pass discovers offload chain via graph structure)
- Fix `defer_offload_waits` to defer from production site instead of latest storage-chain consumer — avoids pushing waits too far (e.g. layers.1's wait landing at layers.3)
- Fix deferred waits landing before offload ops by including `ao.offload` in region anchors

## Test plan
- [ ] `pytest torchtitan/experiments/graph_trainer/tests/test_passes.py -x`
- [ ] `NGPU=8` run with `--compile.debug_graph_passes` and verify tlparse graph dumps show correct metadata and ordering
- [ ] Verify bitwise deterministic test still passes

**Following are desirable, but didn't' land.** 

Before the fix, we have both waits from layer0 and 1 at the end of layer 2
image

after the fix
image

Before the fix
https://www.internalfb.com/intern/diffing/?before_paste_number=2321471471&after_paste_number=2321471510&regex_remove_pattern=&enable_regex_remove=0&strip_empty_lines=0&line_wrap=0&selected_tab=plain_diff

After the fix (removing these few line)
https://www.internalfb.com/intern/diffing/?before_paste_number=2320391950&after_paste_number=2320391970&regex_remove_pattern=&enable_regex_remove=0&strip_empty_lines=0&line_wrap=0&selected_tab=plain_diff


sanketpurandare (approved):

torchtitan/experiments/graph_trainer/cpu_offload.py
```diff
@@ -518,8 +521,6 @@ def defer_offload_waits(
             break
         if n.op != "call_function":
             continue
-        if n.target in _AO_OPS:
-            continue
```
SherlockNoMad, line 522: cc @mlazos 

Removing these 2 lines, so that we have deferred waits after the ao.offload (for the output of layer)

torchtitan/experiments/graph_trainer/cpu_offload.py
```diff
@@ -538,16 +539,9 @@ def defer_offload_waits(
...
             continue
 
-        # If storage chain extends to a later region (cross-layer views),
-        # use the latest consumer region.
-        chain_nodes, _ = _get_storage_chain(node)
-        real_consumers = {n for n in chain_nodes if n.target not in _AO_OPS}
-        for c in real_consumers:
-            idx = node_to_region_idx.get(c)
-            if idx is not None:
-                consumer_idx = max(consumer_idx, idx)
```
SherlockNoMad, line 548: this was deferring layer0's wait to layer 2. 

Removing this fixed it

torchtitan/experiments/graph_trainer/cpu_offload.py
```diff
@@ -538,16 +539,9 @@ def defer_offload_waits(
         if consumer_idx is None:
             continue
 
-        # If storage chain extends to a later region (cross-layer views),
-        # use the latest consumer region.
-        chain_nodes, _ = _get_storage_chain(node)
```
mlazos, line 543: I'm not sure if we can remove this part - is the reason you removed because we don't handle views? This deferred it N regions after the last consumer

torchtitan/experiments/graph_trainer/cpu_offload.py
```diff
@@ -423,25 +429,22 @@ def apply_cpu_offload_pass(
...
+            wait_node.meta.update(src_meta)
             wait_node.meta["val"] = val
             wait_node.meta["autograd_backward"] = True
 
         for user in bwd_users:
             user.replace_input_with(node, wait_node)
 
-        # Store mapping so the remat pass can redirect recomputed nodes
-        # to the reloaded tensor instead of the freed forward tensor.
-        node.meta["cpu_offload_reload_node"] = wait_node
```
mlazos, line 443: this piece was also important iirc, did we handle this some other way?

SherlockNoMad: yes, remat can walk the graph to find the upstream paired offload. 
don't need to rely on node.meta

torchtitan/experiments/graph_trainer/cpu_offload.py
SherlockNoMad: let me show you 

Before the fix
https://www.internalfb.com/intern/diffing/?before_paste_number=2321471471&after_paste_number=2321471510&regex_remove_pattern=&enable_regex_remove=0&strip_empty_lines=0&line_wrap=0&selected_tab=plain_diff

After the fix (removing these few line) 
https://www.internalfb.com/intern/diffing/?before_paste_number=2320391950&after_paste_number=2320391970&regex_remove_pattern=&enable_regex_remove=0&strip_empty_lines=0&line_wrap=0&selected_tab=plain_diff

torchtitan/experiments/graph_trainer/cpu_offload.py
mlazos: yes this seems good, at first I was skeptical but this is why you propagate the metadata to the offload nodes right?

torchtitan/experiments/graph_trainer/cpu_offload.py
SherlockNoMad: discussed offline, we will do it by using the dep argument from ao.wait node instead. 

This change is reverted in the PR.

PR #3323, opened 2026-05-12
Debug Numerics with DebugMode

yushangdi:
# PR Summary: DebugMode-Based Numerics Debugger

## Overview

Example output: 

llama3:  https://www.internalfb.com/intern/px/p/9N1sk/
qwen3: https://www.internalfb.com/intern/px/p/9N2ph/

Screenshot 2026-05-12 at 2 33 05 PM

This PR adds a DebugMode-backed numerics debugger for TorchTitan training runs. It introduces a `--profiler.dump_numerics` flag that captures per-op activation metadata during a selected training step and a `torchtitan.tools.compare_numerics` CLI that compares two capture logs and writes an interactive HTML diff.

It can be used to compare bit-wise equivalent numerics

1) between different eager runs
2) between different traced runs
3) between eager and traced runs

under the same seed, deterministic settings, precision, and parallelism.

An alternative implementation that uses a standalone activation tracker is at https://github.com/pytorch/torchtitan/pull/3190. However, it has some problem with eager code source line tracking. It also doesn't track input hashes and input sources (i.e. which op produced the input). As I add those features, I feel like I'm re-inventing DebugMode, so I just switched back to DebugMode implementation.

**I do not expect the default and deterministic matching algorithm to perfectly match all the ops, especially when comparing eager vs traced runs. But agent should do a pretty good job and use manual override to update the visualization.**

### Per-op numerics capture

- Added `torchtitan/tools/activation_tracer.py`.
- Captures per-op output stats:
  - shape and dtype
  - output DebugMode norm hash, L2 norm, min, max, mean
  - input DebugMode norm hashes
  - input producer attribution where available
  - module FQN, source location, and forward/backward phase
- Uses `torch.utils._debug_mode.DebugMode` with:
  - `record_nn_module=True` for eager module attribution
  - `record_stack_trace=True` for source locations and autograd forward traces
  - `record_ids=True` for tensor ID based producer tracking
  - `DebugMode.log_tensor_hashes(hash_fn="norm", hash_inputs=True)` for output and pre-op input hashes
- Computes tensor stats inline during dispatch instead of retaining tensor clones, keeping memory bounded during capture.
- For tuple/list outputs, records all output tensor hashes from DebugMode; shape/dtype and secondary stats are taken from the first captureable tensor output.
- Tracks skipped excluded ops and writes them into the log header so the HTML report can show what was intentionally filtered.

### Profiler integration

- Added `Profiler.Config.dump_numerics`, exposed as `--profiler.dump_numerics`.
- The profiler now builds a `NumericsDebugger` when numerics dumping is enabled.
- Capture is aligned with `--profiler.profile_freq`; for example, `--training.steps 2 --profiler.profile_freq 2` captures step 2.
- `Trainer` now passes `model_parts[0]` into the profiler so the debugger can install module hooks needed for FQN attribution.
- Output path:

```text
{dump_folder}/numerics/rank_{rank}_activations.log
```
Example log format:

```

For tuple/list outputs, `Output hash` is comma-separated and includes every
captureable tensor output.
[layers.0._checkpoint_wrapped_module.attention.qkv_linear.wq/op_0_mm]
  Shape: torch.Size([16384, 256]), Dtype: torch.bfloat16
  L2 norm:  6.348642e+02
  Min:      -1.140625e+00
  Max:      1.242188e+00
  Mean:     -2.894286e-03
  Output hash: 1.038582e+06
  Input hashes: 3.125000e+05, 8.441234e+04
  Input producers: layers.0.attention_norm/op_0__fused_rms_norm;
  Location: torchtitan/train.py:66
```

### Graph trainer support

- Added `FQNInterpreter` in `torchtitan/experiments/graph_trainer/debug_utils.py`.
- `GraphTrainer` uses `FQNInterpreter` only while numerics capture is active.
- Normal graph-trainer execution stays on the direct `gm(*inputs)` path when capture is not active.
- During traced replay, `FQNInterpreter` restores per-node metadata into context vars so traced captures carry comparable module FQNs, stack locations, and forward/backward phase.

### HTML comparison tool

- Added `torchtitan/tools/compare_numerics.py`.
- Parses two activation logs.
- Matches ops across runs using:
  - manual overrides from `--override`
  - exact normalized op keys
  - fuzzy op keys with per-module counters stripped
  - stats-based fallback matching
- Produces a self-contained HTML report with:
  - forward/backward sections
  - match, diff, run1-only, and run2-only filters
  - match-strategy chips
  - output and input hash comparison
  - per-cell relative-diff coloring
  - input-producer hover tooltips
  - skipped-op chips for excluded dispatched ops

### Tests

- Added `tests/unit_tests/test_activation_tracer.py`.
- Added `tests/unit_tests/test_compare_numerics.py`.
- Added `torchtitan/experiments/graph_trainer/tests/test_numerics_debugger_e2e.py` and wired it into the 8-GPU graph-trainer integration workflow.
- Coverage includes capture filtering, FQN cleanup, stack parsing, DebugMode capture behavior, tuple-output hash capture, profiler-driven dumping, traced replay context propagation, log parsing, matching strategies, override handling, input producer rendering, skipped-op rendering, HTML escaping, and an eager-vs-`aot_fx_trace` Llama3 E2E check with minimum match-quality assertions.

## How Phase, Stack Location, and Module FQN Are Determined

The capture path combines DebugMode metadata, graph-trainer FX metadata, and an autograd-node-to-module map. The source depends on whether the op runs in eager or traced replay, and whether it is a forward or backward op.

| Mode | Phase | Phase source | Stack location source | Module FQN source |
| --- | --- | --- | --- | --- |
| Eager | Forward | No active autograd node, so `record_hook` labels it `forward`. | DebugMode live dispatch traceback from `record_stack_trace=True`, filtered to remove torch internals, DebugMode frames, checkpoint plumbing, and interpreter frames. | DebugMode `current_nn_module_stack` from `record_nn_module=True`; the root model-class prefix is stripped so keys match `named_modules()` style. |
| Eager | Backward | `torch._C._current_autograd_node()` is non-null, so `record_hook` labels it `backward`. | DebugMode autograd-saved forward traceback, parsed and filtered. This is available because stack recording enables anomaly metadata without enabling nan checking. | `_grad_fn_to_module[autograd_node]`, populated during forward by global module hooks and `_record_grad_fn`. |
| `aot_fx_trace` replay | Forward or backward | `FQNInterpreter` reads `node.meta["autograd_backward"]` and sets `_current_phase_override`. | `FQNInterpreter` reads `node.meta["stack_trace"]`, parses it, and sets `_current_stack_frames`. | `FQNInterpreter` reads `node.meta["custom"]["module_fqn"]` and sets `_current_module_name`. |

The priority order inside `DebugModeTracer.record_hook` is:

1. Use context vars set by `FQNInterpreter` when replaying traced graphs.
2. Fall back to DebugMode `ModTracker` for eager forward ops.
3. Fall back to `_grad_fn_to_module` for eager backward ops.
4. Use `"<none>"` only if no attribution source is available.

## How Matching Works in `compare_numerics`

`torchtitan.tools.compare_numerics` parses both activation logs into ordered `OpEntry` lists, then `match_entries` aligns them with progressively more relaxed passes. Earlier passes claim entries first, and the final HTML keeps first-log order for paired and first-only rows.

1. Manual overrides: if `--override <csv>` is provided, each `<log1_key>, <log2_key>` pair is forced before automatic matching. This handles known structural attribution differences such as activation-checkpoint recompute FQNs, FSDP collective renaming, and layer counter shifts.
2. Exact key match: compares normalized `module_fqn/op_N_opname` keys. Normalization strips a leading root class segment like `FSDPLlama3Model.` so eager DebugMode keys match traced metadata keys.
3. Fuzzy key match: strips the per-module op counter, so `layers.0.feed_forward/op_7_mul` can match another `layers.0.feed_forward/mul` when counters drift between modes.
4. Stats match: falls back to `(op_type, Shape, output_hash)`, which catches semantically identical ops whose FQN differs across backends.

Once a pair is found, `_compute_match_status` decides whether it is a match or diff using only shape, output hash, and input hashes. L2/min/max/mean are still shown and colored in the HTML, but they do not drive the match/diff classification because they are secondary summaries of the same tensor.

## How `_record_grad_fn` Attributes Backward Ops to Modules

Backward ops run through the C++ autograd engine, so eager module forward stacks are no longer active when those ops dispatch. To recover module attribution, `DebugModeTracer` installs global module forward pre/post hooks while capture is armed.

The pre-hook records input `grad_fn`s for the module. Those input nodes define the boundary to the caller or sibling modules.

The post-hook calls `_record_grad_fn(output, module_fqn, input_grad_fns)`. `_record_grad_fn` walks backward from each output tensor's `grad_fn` through `next_functions`:

```python
stack = [output.grad_fn]
visited = set()
while stack:
    fn = stack.pop()
    if fn in visited or fn in input_grad_fns:
        continue
    visited.add(fn)
    if fn not in _grad_fn_to_module:
        _grad_fn_to_module[fn] = module_fqn
    for parent, _ in fn.next_functions:
        if parent is not None and parent not in visited:
            stack.append(parent)
```

Two details matter:

- It does not overwrite an existing mapping. Child modules finish their post-hooks before parent modules, so child-owned ops keep the more specific FQN.
- It still walks through already-mapped nodes. This lets a parent module claim "glue" ops that happen between child modules, such as `silu` and `mul` inside a feed-forward block, without taking ownership of the child linear ops.

During backward capture, `record_hook` looks up the current autograd node in `_grad_fn_to_module` and uses the mapped FQN for the captured op.

## Expected Workflow

```text
+-------------------------------------------------------------+
| 1. Run two comparable training jobs                         |
|                                                             |
| - same seed                                                 |
| - deterministic debug mode                                  |
| - same precision and parallelism                            |
| - `--profiler.dump_numerics`                                |
| - same capture step via `--profiler.profile_freq`           |
+-------------------------------------------------------------+
                              |
                              v
+-------------------------------------------------------------+
| 2. Generate the comparison HTML                             |
|                                                             |
| - run `python -m torchtitan.tools.compare_numerics`          |
| - inspect matched rows, diffs, run-only rows, input hashes, |
|   and producer links                                        |
+-------------------------------------------------------------+
                              |
                              v
+-------------------------------------------------------------+
| 3. Triage the comparison                                    |
+-------------------------------------------------------------+
        |                         |                         |
        v                         v                         v
+----------------------+  +----------------------+  +----------------------+
| 3a. Real discrepancy |  | 3b. Bad match        |  | 3c. Numerics match   |
|                      |  |                      |  |                      |
| - fix implementation |  | - ask agent for      |  | - done              |
| - recapture runs     |  |   override pairs     |  |                      |
|                      |  | - regenerate HTML    |  |                      |
+----------------------+  +----------------------+  +----------------------+
        |                         |
        |                         |
        v                         v
   back to step 1            back to step 2
```

Manual overrides use this format:

```csv
# Format: <first_log_key>, <second_log_key>
feed_forward/op_0_silu, layers.5.feed_forward/op_2_silu
layers.0/op_1_add, layers.0/op_0_add
tok_embeddings/op_0__allgather_base_, tok_embeddings/op_1_all_gather_into_tensor_out
layers.0/op_1__allgather_base_, layers.0.attention_norm/op_1_wait_tensor
```

An example override CSV is available at [P2321486470](https://www.internalfb.com/phabricator/paste/view/P2321486470). It covers activation-checkpoint recompute pairings, layer-level counter shifts, and FSDP collective renaming/lowering, including traced `wait_tensor` rows for synchronized collective results.

This workflow is agent-friendly. In practice, asking an agent to explain discrepancies or batch-generate candidate manual overrides from the two logs usually works well; the reviewer still needs to sanity-check that each override pairs semantically equivalent ops before accepting the regenerated comparison.

## Example Usage

The full example script is:

```bash
./agent_space/run_numerics_debug_mode.sh
```

It runs eager and `aot_fx_trace` captures with the same seed and deterministic settings, then generates an HTML diff.

### Eager capture

```bash
NGPU=${NGPU:-2}
DP_SHARD=${DP_SHARD:-2}
SEED=42
DUMP_FOLDER=./outputs_numerics_debug_mode

NGPU=$NGPU MODULE=llama3 CONFIG=llama3_debugmodel \
    ./run_train.sh \
    --dump_folder $DUMP_FOLDER/eager \
    --training.steps 2 \
    --profiler.dump_numerics \
    --profiler.profile_freq 2 \
    --debug.seed $SEED \
    --debug.deterministic \
    --parallelism.data_parallel_shard_degree $DP_SHARD \
    --training.mixed_precision_param float32
```

### `aot_fx_trace` capture

```bash
NGPU=$NGPU MODULE=graph_trainer.llama3 CONFIG=graph_trainer_llama3_debugmodel \
    ./run_train.sh \
    --dump_folder $DUMP_FOLDER/traced \
    --training.steps 2 \
    --profiler.dump_numerics \
    --profiler.profile_freq 2 \
    --compile.mode aot_fx_trace \
    --debug.seed $SEED \
    --debug.deterministic \
    --parallelism.data_parallel_shard_degree $DP_SHARD \
    --training.mixed_precision_param float32
```

For `aot_fx_trace`, step 1 traces and step 2 replays. Capturing step 2 makes the log represent replay execution, with `FQNInterpreter` restoring node-level metadata while the traced graph runs.

### Generate the HTML diff

```bash
python3 -m torchtitan.tools.compare_numerics \
    $DUMP_FOLDER/eager/numerics/rank_0_activations.log \
    $DUMP_FOLDER/traced/numerics/rank_0_activations.log \
    --name1 eager_debug_mode \
    --name2 traced_debug_mode \
    --override agent_space/overrides.csv \
    --output $DUMP_FOLDER/diff.html
```

The override file is optional. It is useful when eager and traced runs produce semantically equivalent ops with different attribution, such as activation-checkpoint recompute FQNs, FSDP collective renaming, or layer-level op counter shifts.

## Expected Outputs

After running the example script:

```text
outputs_numerics_debug_mode/
  eager/numerics/rank_0_activations.log
  traced/numerics/rank_0_activations.log
  diff.html
```

Open `diff.html` in a browser to inspect matched ops, diffs, run-only ops, input hashes, and producer links.

## Notes for Reviewers

- The core training path is unchanged unless `--profiler.dump_numerics` is set.
- Graph-trainer replay uses `FQNInterpreter` only while numerics capture is active.
- The capture uses deterministic settings in the example script and avoids `--debug.deterministic_warn_only`.
- The implementation intentionally keeps comparison tooling under `torchtitan/tools/` and graph-trainer-specific replay metadata bridging under `torchtitan/experiments/graph_trainer/`.


torchtitan/trainer.py
```diff
@@ -818,6 +818,7 @@ def train(self):
         with config.profiler.build(
             global_step=self.step,
             base_folder=config.dump_folder,
+            model=self.model_parts[0],
```
fegin: Does this mean that PP is not supported because there will be multiple model parts?

yushangdi: for now yes. the PR is already kinda long, so I don't want to further complicate it. I can try to extend it to support PP in a follow up PR.

agent_tooling/numerics_debugging/compare_numerics.py
```diff
@@ -0,0 +1,1151 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Compare two numerics activation logs and generate an interactive HTML diff.
```
SherlockNoMad, line 8: didn't read.

this kinda of tool is clearly maintained by agent, and I think titan can have some convention about this. 

cc @tianyu-l

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -554,7 +555,10 @@ def run_traced(
     flat_inputs, _ = _unwrap_subclasses(all_args)
 
     with torch.no_grad():
-        flat_outputs = traced_result.gm(*flat_inputs)
+        if interpreter_cls is not None:
+            flat_outputs = interpreter_cls(traced_result.gm).run(*flat_inputs)
```
SherlockNoMad: interesting... I don't hate this

SherlockNoMad: what's the interpreter you end up using?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -85,6 +85,16 @@ def __init__(self, config):
         # Run post-init hook for the active pass pipeline
         POST_INIT_HOOKS.get(self.config.compile.pass_pipeline, lambda _: None)(self)
 
+    def _get_interpreter_cls(self) -> type | None:
```
SherlockNoMad: _mabye_get_fqn_interpreter

agent_tooling/numerics_debugging/compare_numerics.py
yushangdi: my 2 cents about this: indeed this is maintained by agent. But I think still we should commit this code, as it serves as a better starting point than letting the agent start from scratch. 

Some features that I iterated on and resulted in this .py file:
- a not-too-terrible matching algorithms that respects manual override and does some fuzzy matching
- collapsable/expandable columns when some output tuples are too long
- regrex search/filter by op name 
- show match method
- show excluded ops

One proposal: we could create an agent_playground/ folder to put these things. files in this folder cannot be used by any other folder. They must be standalone.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
yushangdi: The `FQNInterpreter` that I added to debug_utils.py.

torchtitan/experiments/graph_trainer/trainer.py
yushangdi: updated.

.github/workflows/integration_test_8gpu_graph_trainer.yaml
```diff
@@ -5,11 +5,17 @@ on:
     branches: [ main ]
     paths:
       - 'torchtitan/experiments/graph_trainer/**'
+      - 'torchtitan/tools/activation_tracer.py'
+      - 'torchtitan/tools/compare_numerics.py'
+      - 'torchtitan/tools/profiler.py'
```
tianyu-l: According to the `experiments` folder contract, let's not auto-trigger graph trainer CI. Instead, we recommend tagging `ci/graph_trainer` to jobs that might affect graph trainer.

torchtitan/trainer.py
tianyu-l: Please add a TODO in code

agent_tooling/numerics_debugging/activation_tracer.py
```diff
@@ -0,0 +1,909 @@
...
+        # Install global forward hooks so backward ops can recover
+        # their owning module's FQN via the autograd graph.  DebugMode's
+        # ModTracker only fires on Python forward calls, so during
+        # backward (driven by the C++ autograd engine) its module stack
+        # is empty.  These hooks fire on every nn.Module.__call__:
+        # walk the autograd graph from each module's output during the
+        # forward post-hook and stash grad_fn -> fqn in _grad_fn_to_module.
+        # _maybe_record_module_fqn in record_hook reads it during backward.
+        _grad_fn_to_module.clear()
+        module_to_fqn = {id(m): name for name, m in self.model.named_modules()}
```
tianyu-l, line 580: I see that you are sending model reference all the way here for getting this dict to insert hooks. Is there a nicer way?

agent_tooling/numerics_debugging/activation_tracer.py
tianyu-l, line 1: I can certainly tell how useful it is, but it is certainly not torchtitan's interest to maintain this file.
- How would you convince people that this shouldn't be in core pytorch or in annex?
- How much of this file is for graph trainer specific? Maybe there's chance to abstract it out and put in `graph_trainer/`.

agent_tooling/numerics_debugging/compare_numerics.py
tianyu-l: Maybe we can put it in `scripts/` with a README.

torchtitan/trainer.py
tianyu-l: If you just want to get a union of `module_to_fqns`, then it's easy to do it here?

agent_tooling/numerics_debugging/activation_tracer.py
yushangdi: none of the file is graph trainer specific (well, maybe one could argue that some of the override hooks are graph_trainer specific).

I'm ok with putting it in `graph_trainer`, but I still want eager to be able use these logging (e.g. so we can compare eager and traced runs). Then we'd need to import this file from `graph_trainer` folder. Is this ok? 

Alternatively, maybe this can live in an annex repo or pytorch core, but it would be less hackable if someone wants to customize the activation tracer while debugging titan.

tianyu-l, line 1: conclusion 1: put these in a subfolder (e.g. `numerics_debugging`) under `agent_tooling` (which probably should be at root level)

torchtitan/tools/profiler.py
tianyu-l, line 1: conclusion 3: don't change core; instead, have skills to teach agent to use the debug tool

tests/unit_tests/test_activation_tracer.py
tianyu-l, line 1: conclusion 2: don't need to have test to guard on it; agents can fix them -- just submit new PR for the fix

agent_tooling/numerics_debugging/activation_tracer.py
tianyu-l: conclusion 4: Even code under `agent_tooling` is not our "selling point", compared with other core code (thus we shouldn't need to review and maintain every single line), we should still put efforts in de-slop. In particular, the code under this file can be improved.

tianyu-l (commented):
Inlined some conclusions we had during meeting. Overall the idea is to give it a try and see how it goes.

yushangdi:
moved code to `agent_tooling` now. Agent is able to generate a numerics report. See example dialogue at https://www.internalfb.com/phabricator/paste/view/P2343644790?view=markdown and output at https://pxl.cl/9SSP9

agent_tooling/numerics_debugging/activation_tracer.py
```diff
@@ -0,0 +1,909 @@
...
+                op_name=op_name,
+            )
+
+            # Register this op as the producer of its output tensors so
+            # downstream consumers can find it.
+            for tid in record.get("output_ids") or []:
+                producer_map[tid] = key
+
+
+class NumericsDebugger:
```
tianyu-l: nit: does it make sense to rename this to
```suggestion
class ActivationTracer:
```
- it only captures activation trace, but doesn't do any "debug" work
- the file name is activation_tracer.py
- (actually, I didn't review carefully, and didn't understand why there's two layer structure of this class and an attribute `DebugModeTracer`)

I don't have strong opinion though.

tianyu-l (approved):

felipemello1:
sorry, i took a quick look at how the files are organized. Would it be better to organize it as a skill? take a look at how OpenEnv has it: https://github.com/meta-pytorch/OpenEnv/tree/main/.claude/skills/generate-openenv-env

- The skill has a short description
- The skill itself is short and points to a larger doc

This is how anthropic recommends building it: https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#skill-structure

agent_tooling/numerics_debugging/activation_tracer.py
yushangdi: Renamed to `ActivationCaptureProfiler`. It wraps DebugModeTracer to plug into torchtitan's Profiler; the split keeps the core tracer decoupled from the profiler integration.

and I also updated to move some .md files to the `.claude/skills` folder as @felipemello1  suggests.

PR #3325, opened 2026-05-12
[graph_trainer] Add graph EP chunking pass

### LINKED ISSUE #3434: RFC: Graph-Based EP Overlap For GraphTrainer
# EP Overlap Graph Chunking RFC

## Status And Scope

This document is the current design contract for the TorchTitan GraphTrainer
EP-overlap stack. It describes the implemented contracts, APIs, scheduling
algorithms, validation strategy, and known limitations.

The system is intentionally split into narrow passes:

```text
trace GraphTrainer train step
  -> cleanup / canonicalization
  -> memory policy / CPU offload / remat
  -> eager or graph EP chunk metadata
  -> EP process-group isolation
  -> FSDP bucket ordering
  -> EP token-exchange scheduling
  -> EP chunk symbolic-shape concretization
  -> optional FSDP dense-region scheduling
  -> optional async TP pass
  -> regional or full Inductor
```

Current supported EP-overlap modes:

```text
+------------------------+-----------+-----------------+----------------------+
| module root            | chunk dim | chunk strategy  | status               |
+------------------------+-----------+-----------------+----------------------+
| layers.*               | batch     | eager, graph    | supported            |
| layers.*.moe           | batch     | eager, graph    | supported            |
| layers.*.moe           | seq       | eager, graph    | supported            |
| layers.*               | seq       | any             | rejected             |
| any graph chunking     | any       | graph + tp > 1  | rejected for now     |
+------------------------+-----------+-----------------+----------------------+
```

Eager chunking is the semantic oracle. Graph chunking is the performant traced
implementation that constructs the same two-chunk program directly in FX and
then schedules EP communication.

## TLDR

For expert readers, the design is:

```text
Goal:
  Split selected EP regions into two chunks, launch EP token exchanges for one
  chunk while useful compute from the peer chunk is available, and keep FSDP
  AG/RS away from MoE token-exchange windows when the explicit dense scheduler
  is enabled.

Semantic oracle:
  eager chunking:

    y = module(x)

  becomes:

    x0, x1 = split(x)
    y0 = module(x0)
    y1 = module(x1)
    y  = cat(y0, y1)

Graph implementation:
  trace unchunked model -> find selected symbolic dim -> copy selected
  dataflow body twice -> materialize live-outs -> schedule token exchanges.

Core invariant:
  A graph EP chunk body must contain EP collectives and local compute only.
  Non-EP collectives inside a graph chunk body are rejected unless a future
  layout-boundary proof explicitly supports them.

Current graph TP rule:
  graph EP chunking rejects tensor_parallel_degree > 1. After DTensor lowering,
  the FX graph contains physical rank-local tensors and TP/SP layout helpers;
  blindly chunking those paths is not proven eager-equivalent.
```

Current high-level pass order:

```text
cleanup
  -> memory/offload/remat tagging
  -> populate eager chunk metadata              # eager strategy only
     OR populate chunk symbol metadata          # graph strategy only
        -> ep_overlap_chunk_pass                # graph strategy only
  -> isolate_ep_process_group_pass
  -> optional reassign_collective_pgs_pass      # FSDP AG/RS overlap flag only
  -> joint_transformer_block_bucketing_reordering_pass
  -> ep_overlap_schedule_pass
  -> concretize_ep_chunk_symbolic_shapes_pass
  -> schedule_fsdp_comms_to_dense_regions_pass  # dense-region flag only
  -> optional async_tensor_parallel_pass
  -> regional/full Inductor
```

MoE-root forward intuitive target schedule:

```text
chunk0 router
chunk0 token-count A2A
chunk0 token-count sync setup
chunk1 router
chunk1 token-count A2A
chunk1 token-count sync setup
chunk0 dispatch token_exchange launch
chunk1 dispatch token_exchange launch
chunk0 shared_expert
chunk1 shared_expert
chunk0 dispatch wait
chunk0 groupedMM
chunk0 combine token_exchange launch
chunk1 dispatch wait
chunk1 groupedMM
chunk1 combine token_exchange launch
chunk0 combine wait
chunk1 combine wait
```

MoE-root backward intuitive target schedule:

```text
chunk1 combine token_exchange launch
chunk0 combine token_exchange launch
chunk1 remat/groupedMM prefix
chunk0 remat/groupedMM prefix
chunk1 combine wait
chunk1 groupedMM input_grad
chunk1 dispatch token_exchange launch
chunk0 combine wait
chunk0 groupedMM input_grad
chunk0 dispatch token_exchange launch
chunk1 groupedMM wgrad
chunk0 groupedMM wgrad
chunk1 dispatch wait
chunk0 dispatch wait
```

These target schedules are not hardcoded node lists. They are design targets
that explain the overlap intuition: launch both chunks' next useful EP
exchanges as early as their dependencies allow, run independent compute while
the exchange is in flight, and delay waits until the payload is needed. The
implemented scheduler derives a legal schedule from graph metadata, dependency
closures, marker labels, and validation; the goal is to arrive automatically at
this schedule or a similarly useful one by encoding the decisions behind the
intuition, not by matching node names or fixed positions.

Dense FSDP scheduling, when explicitly enabled:

```text
forward execution:  layer i-1 -> layer i
  launch AG(i) inside dense_fwd(i-1)
  keep AG wait just before layer i param use

backward execution: layer i+1 -> layer i -> layer i-1
  launch AG(i) inside dense_bwd(i+1)
  launch RS(i) inside dense_bwd(i-1)
  sink RS wait/output-unpack closure to graph tail
```

Numerics stance:

```text
strict bitwise tests:
  compare eager chunking vs graph chunking
  disable early param-grad accumulation
  do not enable dense FSDP scheduling unless the test targets it

performance path:
  early param-grad accumulation enabled by default
  dense FSDP scheduling enabled only by explicit flag
```

Debug knobs:

```text
--compile.debug_graph_passes
--compile.disable_passes pass_name[,pass_name...]

debug_graph_passes logs pass-before/pass-after graphs to TLParse.
disable_passes skips named passes exactly and is intended for local ablation.
```

## Mental Model

DeepSeek-style MoE execution has expensive expert-parallel token exchanges and
expert compute. Without chunking, a layer tends to serialize the token exchange
and compute for one large activation extent:

```text
router/topk
  -> token-count exchange/sync
  -> dispatch all_to_all
  -> dispatch wait
  -> groupedMM
  -> combine all_to_all
  -> combine wait
  -> next work
```

EP overlap creates two equal chunks of the selected logical extent:

```text
full input x
    |
    +-- x0 -> selected region -> y0 --+
    |                                 +--> full output y
    +-- x1 -> selected region -> y1 --+
```

The important scheduling opportunity is that an async collective launch is not
useful unless there is independent compute before the corresponding wait:

```text
chunk0 launch A2A -----------------------------+
chunk1 useful compute                          | A2A in flight
chunk1 launch A2A -----------------------------+
chunk0 useful compute / wait when needed <-----+
```

Two module roots are supported because they have different scheduling meaning:

```text
Transformer-root chunking: layers.*

  chunk0 attention + moe + residual
  chunk1 attention + moe + residual

  The scheduler keeps the generic wait-gated marker order. It does not apply
  MoE-specific first-marker pairing because the chunk body includes attention
  and other dense work before the MoE token setup.

MoE-root chunking: layers.*.moe

  chunk0 router/token exchange/expert/shared expert
  chunk1 router/token exchange/expert/shared expert

  The scheduler pairs both chunks' first marker setup before launching either
  first token exchange. This is the mode that enables the clean MoE schedule
  shown in the TLDR.
```

Terms used throughout the passes:

```text
root:
  A concrete module FQN matched by the configured pattern, for example
  layers.7 or layers.7.moe.

body:
  Nodes inside one root and one direction that are influenced by the selected
  chunk symbol or by existing per-chunk provenance.

live-in:
  A value produced outside the body and consumed inside it.

live-out:
  A value produced inside the body and consumed outside it.

provenance:
  Internal mapping from a full graph value to its chunk0/chunk1 values.

marker:
  A true EP token-exchange all_to_all launch used as a scheduling anchor.

token-count exchange:
  The all_to_all that exchanges per-expert token counts. It is setup for
  split sizes, not a scheduling marker.

token exchange:
  The dispatch or combine all_to_all that moves token payloads.

wait:
  The wait_tensor node for an async c10d collective.

dense region:
  Dense attention/FFN compute used as an insertion window for FSDP AG/RS
  launches. In MoE layers this intentionally excludes MoE token exchange.
```

## Configuration And APIs

Primary user-facing config fields live on `GraphTrainerCompileConfig`. EP
overlap is controlled by structured config, not by `compile.passes`:

```python
compile.ep_overlap.enabled: bool
compile.ep_overlap.strategy: Literal["eager", "graph"]
compile.ep_overlap.chunk_dim: Literal["batch", "seq"]
compile.ep_overlap.module_fqn: Literal["layers.*", "layers.*.moe"]
compile.ep_overlap.disable_early_grad_accumulation: bool
compile.enable_fsdp_ag_rs_overlap: bool
compile.enable_fsdp_dense_region_overlap: bool
compile.inductor_compilation: Literal["regional", "full"]
compile.debug_graph_passes: bool
compile.disable_passes: list[str]
```

Typical CLI shapes:

```bash
# Transformer-batch EP overlap.
--compile.mode aot_fx_trace
--compile.inductor_compilation regional
--compile.ep_overlap.enabled
--compile.ep_overlap.strategy graph
--compile.ep_overlap.chunk_dim batch
--compile.ep_overlap.module_fqn layers.*

# MoE-batch EP overlap.
--compile.mode aot_fx_trace
--compile.inductor_compilation regional
--compile.ep_overlap.enabled
--compile.ep_overlap.strategy graph
--compile.ep_overlap.chunk_dim batch
--compile.ep_overlap.module_fqn layers.*.moe

# MoE-seq EP overlap.
--compile.mode aot_fx_trace
--compile.inductor_compilation regional
--compile.ep_overlap.enabled
--compile.ep_overlap.strategy graph
--compile.ep_overlap.chunk_dim seq
--compile.ep_overlap.module_fqn layers.*.moe

# Add dense-region FSDP AG/RS scheduling.
--compile.enable_fsdp_dense_region_overlap
```

Validation rules:

```text
ep_overlap.enabled:
  appends the ep_overlap trace-input preparer and enables the EP chunking /
  scheduling pass family

ep_overlap.module_fqn:
  must be exactly layers.* or layers.*.moe

seq chunking:
  only supported with layers.*.moe

graph chunking + tp > 1:
  rejected until the graph can prove non-EP layout paths are outside the chunk
  body

enable_fsdp_dense_region_overlap:
  explicit opt-in. If EP overlap is also enabled, it only composes with graph
  chunking rooted at layers.*.moe. Otherwise the request is skipped with a
  warning. Without EP overlap, the pass can run as a standalone FSDP scheduling
  ablation.

disable_passes:
  debug only. Pass names match exactly.
```

Important pass APIs:

```python
populate_chunk_dim_metadata_pass(
    gm,
    *,
    mode: Literal["batch", "seq"],
) -> fx.GraphModule

ep_overlap_chunk_pass(
    gm,
    *,
    mode: Literal["batch", "seq"],
    module_pattern: str,
    num_static_inputs: int,
    optimize_grad_live_out: bool,
    require_all_to_all: bool,
) -> fx.GraphModule

populate_eager_chunk_metadata_pass(gm) -> fx.GraphModule

isolate_ep_process_group_pass(gm) -> fx.GraphModule

ep_overlap_schedule_pass(
    gm,
    *,
    module_pattern: str,
    require_all_to_all: bool,
    pair_first_token_exchange: bool,
) -> fx.GraphModule

concretize_ep_chunk_symbolic_shapes_pass(gm, example_inputs=None)
    -> fx.GraphModule

schedule_fsdp_comms_to_dense_regions_pass(
    gm,
    *,
    moe_layer_ids: frozenset[int],
    n_layers: int,
    transformer_bucket_counts_by_layer: dict[int, int] | None,
    strict: bool,
) -> fx.GraphModule
```

## Pass Pipeline

The current pass builder constructs the EP-overlap pipeline from explicit
config and model/parallelism facts.

Simplified pseudocode:

```python
passes = [
    eliminate_dead_code_pass,
    canonicalize_graph_pass,
]

ep_overlap_chunk_passes = []
if compile.ep_overlap.enabled:
    chunk_dim, strategy, module_fqn = validate_ep_overlap_config(
        compile.ep_overlap
    )
    if strategy == "graph" and tensor_parallel_degree > 1:
        raise ValueError

    if strategy == "eager":
        ep_overlap_chunk_passes += [populate_eager_chunk_metadata_pass]

    if strategy == "graph":
        ep_overlap_chunk_passes += [
            populate_chunk_dim_metadata_pass(mode=chunk_dim),
            ep_overlap_chunk_pass(
                mode=chunk_dim,
                module_pattern=module_fqn,
                num_static_inputs=traced_result.num_static_inputs,
                optimize_grad_live_out=not (
                    compile.ep_overlap.disable_early_grad_accumulation
                ),
                require_all_to_all=ep_degree > 1,
            ),
        ]

passes += [
    tag_with_memory_policy_pass(config=config),
    apply_cpu_offload_pass(...),
    selective_activation_remat_pass,
]

if compile.ep_overlap.enabled:
    passes += ep_overlap_chunk_passes
    passes += [isolate_ep_process_group_pass, eliminate_dead_code_pass]

if compile.enable_fsdp_ag_rs_overlap:
    passes += [reassign_collective_pgs_pass]

passes += [joint_transformer_block_bucketing_reordering_pass(...)]

if compile.ep_overlap.enabled:
    passes += [
        ep_overlap_schedule_pass(
            module_pattern=module_fqn,
            require_all_to_all=ep_degree > 1,
            pair_first_token_exchange=(module_fqn == "layers.*.moe"),
        ),
        concretize_ep_chunk_symbolic_shapes_pass,
    ]

enable_dense_scheduler = compile.enable_fsdp_dense_region_overlap
if enable_dense_scheduler and compile.ep_overlap.enabled and not (
    strategy == "graph" and module_fqn == "layers.*.moe"
):
    warn_and_disable_dense_scheduler()
    enable_dense_scheduler = False

if enable_dense_scheduler:
    passes += [schedule_fsdp_comms_to_dense_regions_pass(strict=True)]

if parallelism.enable_async_tensor_parallel:
    passes += [async_tensor_parallel_pass]

if inductor_compilation == "full":
    passes += [full_inductor_compilation_pass]
else:
    passes += [
        regional_inductor_pass,
    ]
```

Why the order matters:

```text
memory/remat before chunking:
  The chunk planner sees the actual graph that will run, including recompute
  nodes that can become backward overlap filler.

chunking before bucketing/scheduling:
  Bucketing and scheduling must see the final two-chunk body shape and metadata.

EP PG isolation before scheduling:
  EP all-to-alls should be on their isolated high-priority PG before the final
  graph order is fixed.

FSDP bucketing before EP scheduling:
  FSDP AG/RS launches and waits should be present when EP scheduling and dense
  scheduling make placement decisions.

EP scheduling before concretization:
  The scheduler consumes chunk metadata and symbolic body structure. Backend
  codegen does not need those temporary chunk symbols.

dense FSDP scheduling after EP scheduling:
  Dense scheduling moves AG/RS launches after EP regions have reached their
  intended order, so it does not accidentally place FSDP comms inside MoE
  token-exchange windows.

async TP after EP/FSDP scheduling:
  TP micro-pipelining rewrites TP collectives into fused symmetric-memory
  kernels, so it runs after EP and FSDP placement decisions have used the
  original c10d graph structure.
```

Precompile constraint:

```text
precompile fingerprints EP-overlap config fields that affect compiled output:
enabled, chunk_dim, strategy, module_fqn, and disable_early_grad_accumulation.
Training rejects a mismatched artifact unless fingerprint validation is
explicitly skipped.
```

## Chunked Loss Boundary

GraphTrainer traces a full forward + loss + backward step with
`torch.autograd.grad(loss, params)`. The regular `ChunkedCELoss` accumulates
`lm_head` parameter gradients through `.grad` side effects inside its chunk
loop, which is correct for eager training but not sufficient as a traced graph
output.

Current GraphTrainer config conversion therefore replaces
`ChunkedCELoss.Config` with `ChunkedCELossWithParamGrads.Config`:

```text
base Trainer config:
  loss = ChunkedCELoss.Config(...)

GraphTrainer config:
  loss = ChunkedCELossWithParamGrads.Config(...)
```

The graph-trainer loss variant captures the sharded `lm_head` parameter grads
created by the chunk loop, clears the eager `.grad` side effect, and exposes
those grads as explicit autograd outputs. This keeps chunked loss semantically
equivalent to the Trainer path while making the fwd/loss/bwd graph replayable.

## Graph Chunking Contract

Graph chunking implements the eager two-call wrapper directly on the traced FX
graph. It does not replay module code. It does not run broad shape propagation
to paper over stale fake values. It uses selected-symbol dataflow, schema
metadata, and explicit materialization proofs.

### Selected Symbol Metadata

Before tracing, selected logical input dimensions are marked dynamic:

```text
batch mode:
  selected dim = tensor dim 0

seq mode:
  selected dim = tensor dim 1
```

After tracing, `populate_chunk_dim_metadata_pass` records symbol hints on
placeholders:

```python
CHUNK_SYMBOL_HINTS_META = {
    symbol_u0: 128,
}
```

Rules:

```text
1. The selected symbolic extent remains symbolic during chunking/scheduling.
2. Hints are used for evenness checks and final concretization, not as the
   semantic source of truth.
3. Conflicting hints for the same symbol are errors.
4. The selected extent must be even for pairwise two-chunk scheduling.
```

### Root Discovery

Production graph chunking finds roots that contain real EP all-to-all nodes:

```text
all_to_all_single node
  -> module FQN metadata
  -> pattern root match, for example layers.7.moe
```

Single-rank tests may not have real all-to-all. In that case, when
`require_all_to_all=False`, the pass can find roots from EP annotations.

```text
production EP config:
  no real matching EP all-to-all -> error

single-rank test config:
  no real all-to-all but EP annotations exist -> allowed
```

### Body Planning

The module FQN only bounds the search. The actual duplicated body is chosen by
selected-symbol influence and provenance.

```text
candidate root nodes:
  nodes inside selected root and direction

seed nodes:
  nodes whose tensor metadata, symbolic args, or per-chunk provenance depends
  on selected chunk symbols

body:
  selected-symbol dataflow closure, bounded by root and direction
```

ASCII view:

```text
outside root                 selected root                  outside root
------------                 -------------                  ------------
param/buffer --------------> shared/static use

x: [B, S, H] ---> split ---> chunk body node ---> live-out ----> full user
                         \
                          -> peer chunk body

shape scalar u0 ----------> chunked symbolic scalar consumer

unrelated tensor ---------> shared full input
```

Graph chunking validates that selected graph bodies do not contain unsupported
non-EP collectives:

```text
allowed in MoE graph chunk body:
  EP token-count exchange
  EP dispatch/combine token exchange
  EP waits
  router/top-k/permutation
  groupedMM
  shared expert compute
  local view/index/scatter plumbing

not allowed today:
  TP/SP collectives
  CP collectives
  FSDP/EFSDP activation layout collectives
  unknown c10d process groups
```

This is graph-native because after DTensor lowering the FX graph no longer has
DTensor placement objects. The pass can only trust observable graph evidence.

### Live-Ins

Live-in behavior is semantic, not syntactic:

```text
+--------------------------------------------+-----------------------------+
| live-in kind                               | chunk behavior              |
+--------------------------------------------+-----------------------------+
| tensor carrying selected symbol             | split along selected dim    |
| value with existing provenance              | reuse chunk0/chunk1 values  |
| symbolic scalar derived from selected dim   | rewrite symbol expression   |
| static placeholder / param / buffer         | share full value            |
| unrelated tensor or scalar                  | share full value            |
+--------------------------------------------+-----------------------------+
```

Tensor live-in split:

```text
x: [u0, H]
     |
     v
split_with_sizes(x, [u0//2, u0//2], dim=0)
     |
     +--> x0: [u0//2, H]
     +--> x1: [u0//2, H]
```

Sequence split for MoE roots:

```text
x: [B, u1, H]
     |
     v
split_with_sizes(x, [u1//2, u1//2], dim=1)
     |
     +--> x0: [B, u1//2, H]
     +--> x1: [B, u1//2, H]
```

Scalar rewrite:

```text
full scalar:
  u0

chunk scalar:
  u0 // 2

derived expression:
  u0 * hidden -> (u0 // 2) * hidden
```

### Body Copying

One chunk body is rewritten in place. The peer chunk body is copied:

```text
original node N
  chunk0: N rewritten with chunk0 args
  chunk1: N_copy created with chunk1 args
```

Argument mapping:

```python
def map_arg(arg, chunk_id):
    if arg in per_chunk_provenance:
        return per_chunk_provenance[arg][chunk_id]
    if arg in split_live_ins:
        return split_live_ins[arg][chunk_id]
    if arg is selected_symbolic_scalar:
        return rewritten_chunk_scalar(arg)
    if arg is symbolic_expression_containing_selected_symbol:
        return rewrite_selected_symbols(arg)
    return arg
```

Copied body nodes receive stable metadata:

```text
chunked_region_fqn
chunked_region_is_backward
chunked_region_producer = graph
chunked_region_role = body
chunk_id = 0 or 1
```

Parameter communication chains are not copied merely because they are adjacent
to the selected body. Parameters and buffers are shared live-ins; their FSDP
all-gather chains are layout/parameter-availability plumbing, not per-token MoE
semantics. Duplicating them inside each chunk would turn one parameter all-gather
into two all-gathers and would move FSDP communication into the EP overlap
window. Instead, graph chunking keeps parameter all-gather chains outside the
body, lets both chunks consume the same gathered parameter value, and leaves the
AG wait at the parameter user.

The backward reduction side follows the same principle. Per-chunk parameter
gradient values are semantic body results, but the FSDP reduce-scatter chain is
distributed-gradient plumbing. Graph chunking should materialize the full
parameter gradient once, then feed one reduction chain, rather than copying a
separate reduction chain for each chunk. Early param-grad accumulation is the
performance form of that rule.

```text
desired:

  param shard
      |
      v
  AG launch/wait -> full param
                    /        \
                   v          v
              chunk0 use   chunk1 use

not copied:

  chunk0: AG launch/wait -> chunk0 use
  chunk1: AG launch/wait -> chunk1 use

backward reduction shape:

  chunk0 grad --+
                +--> accumulated grad -> one reduce_scatter
  chunk1 grad --+
```

### Mutation Ordering

FX-visible mutations are preserved using operator schema alias/write metadata.
If a shared mutable live-in is updated by both chunks, chunk 1 observes chunk 0's
mutation result:

```text
eager order:
  buf1 = update_(buf0, chunk0)
  buf2 = update_(buf1, chunk1)

graph chunking order:
  update__chunk0(buf, update0)
  update__chunk1(update__chunk0, update1)
```

This matters for MoE bookkeeping tensors too. If `num_tokens_per_expert` is
updated in-place while computing dispatch metadata, or if an expert-bias buffer
is updated by load-balancing logic, chunk 1 must see the post-chunk0 value. The
pass relies on FX-visible alias/write metadata for this ordering; it does not
guess from names.

Python side effects that do not appear in FX are unsupported.

### Live-Outs

A body value consumed outside the body is classified by consumer:

```text
chunked consumer:
  same selected root + same direction

full consumer:
  outside selected root
  graph output
  root/direction boundary that needs eager-equivalent full value
```

If only chunked consumers exist, no full value is materialized:

```text
live_out_chunk0 -> chunked_user_chunk0
live_out_chunk1 -> chunked_user_chunk1
```

If full users exist, materialization must be proven:

```text
tensor with selected chunk dim:

  y0: [u0//2, H] --+
                   +--> cat(dim=0) -> y: [u0, H] -> full user
  y1: [u0//2, H] --+

backward param-gradient partial:

  dw0 --+
        +--> add(dw0, dw1) -> dw -> grad user / graph output
  dw1 --+

scalar shape expression:

  full scalar users keep original full scalar.
  chunked scalar users receive rewritten chunk-local scalar.
```

Unsupported live-outs fail loudly. The pass does not use materialization as a
generic fallback for unknown values.

### Early Param-Grad Accumulation

The default performance path materializes proven backward parameter-gradient
partials after the required cast but before distributed gradient reduction. The
goal is to hoist chunk accumulation before reduction, converting two equivalent
per-chunk reductions into one reduction over the accumulated gradient. This is
the `optimize_grad_live_out` path.

```text
without early accumulation:

  chunk0 grad partial
      -> cast/reduce-dtype suffix
      -> reduce_scatter ------------------+
                                           +--> later grad handling
  chunk1 grad partial                      |
      -> cast/reduce-dtype suffix          |
      -> reduce_scatter ------------------+

with early accumulation:

  chunk0 grad partial -> cast/reduce-dtype suffix --+
                                                    +--> add
  chunk1 grad partial -> cast/reduce-dtype suffix --+     |
                                                          v
                                                   single reduce_scatter

disabled for strict bitwise:
  preserve eager chunking's cast/reduction order
```

The optimization must not move the cast after the add. The copy/materialization
chain stops at the cast so both chunk partials are first converted to the same
intended reduction/accumulation dtype, then accumulated, then reduced once. That
preserves the precision behavior while removing the duplicated distributed
reduction.

Rules:

```text
1. Only proven backward parameter-gradient partials can use add materialization.
2. The proof is recorded in metadata.
3. Strict eager-vs-graph bitwise tests pass
   --compile.ep_overlap.disable_early_grad_accumulation.
```

## Eager Chunking Contract

Eager chunking installs thin wrappers around selected module forwards before
tracing. The wrapper defines the semantic oracle:

```python
def chunked_forward(*args, **kwargs):
    full_extent = first_matching_tensor.shape[dim]
    torch._check(full_extent % 2 == 0)
    chunk_extent = full_extent // 2

    split_args = split_matching_tensors(args, chunk_extent, dim)
    split_kwargs = split_matching_tensors(kwargs, chunk_extent, dim)

    y0 = original_forward(chunk0_args, chunk0_kwargs)
    y1 = original_forward(chunk1_args, chunk1_kwargs)

    return tree_cat([y0, y1], dim)
```

Eager wrappers annotate:

```text
split boundaries
chunk body nodes
materialization nodes
root FQN
chunk id
```

`populate_eager_chunk_metadata_pass` promotes traceback annotations into normal
FX node metadata so `ep_overlap_schedule_pass` can schedule eager-produced and
graph-produced chunked regions with the same metadata contract.

Important distinction:

```text
eager chunking:
  creates a correct chunked program before tracing

graph chunking:
  traces the unchunked program, then transforms FX into an equivalent chunked
  program

future eager optimization:
  may trace eager chunking, then hoist/sink duplicated non-EP layout chains
  only when a formal equivalence proof exists
```

## EP Process-Group Isolation

EP process-group isolation prevents EP all-to-alls from sharing the same
process-group object with non-EP collectives. Sharing the same PG object can
couple ordering and stream behavior across unrelated communication.

Current algorithm:

```python
for node in graph:
    pg = collective_pg_name(node)
    if node is all_to_all_single with EP PG slot:
        ep_pg_names.add(pg)
    else if node is supported c10d collective:
        non_ep_pg_names.add(pg)

for ep_pg in ep_pg_names:
    if ep_pg resolves to the same PG object as any non_ep_pg:
        extra_ep_pg = new_group(
            ranks=same_ranks,
            backend="nccl",
            high_priority=True,
            group_desc="ep_extra",
        )
        rewrite EP all_to_all pg operand to extra_ep_pg
```

Before:

```text
PG_A object
  |
  +--> FSDP AG/RS
  +--> EP dispatch/combine all_to_all
```

After:

```text
PG_A object
  |
  +--> FSDP AG/RS

PG_EP_EXTRA object, same ranks, high-priority
  |
  +--> EP dispatch/combine all_to_all
```

Rank equality alone is not enough to trigger rewriting. The pass checks
resolved process-group object identity. Distinct PGs with the same ranks are
already isolated.

## EP Scheduling Algorithm

The EP scheduler is intentionally a scheduler only. It consumes chunked-region
metadata from eager or graph chunking. It must not change tensor values,
materialization, or provenance.

### Scheduling Markers

True scheduling markers are dispatch/combine token-exchange all-to-all launches:

```text
target:
  torch.ops._c10d_functional.all_to_all_single.default

metadata:
  custom[EP_token_exchange] in {"dispatch", "combine"}
```

The token-count all-to-all is not a marker:

```text
token-count exchange:
  setup for CPU split-size lists
  belongs in the first marker closure
  not a payload token-exchange scheduling anchor
```

Waits may inherit token-exchange metadata through traceback. The scheduler
normalizes waits to:

```text
custom[EP_token_exchange_wait] = "dispatch" or "combine"
```

### Region Planning Pseudocode

```python
def schedule_ep_overlap(gm, module_pattern):
    regions = collect_chunked_regions(gm, module_pattern)
    owner_by_node = build_owner_lookup(regions)

    scheduled_regions = []
    for region in regions:
        assert chunks(region) == {0, 1}

        exchanges = {
            0: collect_token_exchange_markers(region.chunk0),
            1: collect_token_exchange_markers(region.chunk1),
        }

        if no markers and require_all_to_all:
            error
        if only one chunk has markers:
            error
        if marker counts or labels differ:
            error

        phases = build_region_phases(
            region,
            exchanges,
            chunk_order=(1, 0) if backward else (0, 1),
            pair_first_token_exchange=(root pattern is layers.*.moe),
        )
        scheduled_regions.append((region, phases))

    apply_global_topological_schedule(scheduled_regions)
    validate_phase_order_materialized()
```

### Marker Closure

For each token-exchange launch, the scheduler builds the same-chunk dependency
closure required to launch it:

```text
closure(launch):
  all same-region body ancestors needed by launch
  dependencies hidden behind unowned graph plumbing
  no peer-chunk dependency unless explicitly allowed by dependency traversal
  no dependency on a later same-chunk marker
```

Example:

```text
router/topk
  -> local reorder
  -> token-count exchange
  -> token-count sync
  -> dispatch all_to_all launch

closure(dispatch launch) =
  router/topk, local reorder, token-count exchange, sync, dispatch launch
```

### Wait-Gated Phase Emission

After each marker pair, the scheduler emits ready filler work that does not
require token-exchange waits:

```python
for marker_index in markers:
    emit marker closures for both chunks in target order
    emit ready non-c10d filler from future marker closures and body filler

emit remaining body tails with waits allowed
```

This preserves producer-before-consumer ordering because final application uses
a global topological emitter:

```text
requested region phase order
       |
       v
global emitter recursively emits dependencies
       |
       v
lint + recompile + validate phase order
```

If a requested phase requires a later phase node as an input, the pass errors
with a dependency path instead of silently changing the schedule.

### MoE-Root Forward Intuitive Target Schedule

For `layers.*.moe`, the first marker setup is paired across chunks:

```text
chunk0 router
chunk0 token-count A2A
chunk0 sync setup
chunk1 router
chunk1 token-count A2A
chunk1 sync setup
chunk0 dispatch token_exchange launch
chunk1 dispatch token_exchange launch

chunk0 shared_expert --------------------------+
chunk1 shared_expert                           | dispatch A2A in flight
                                                |
chunk0 dispatch wait <-------------------------+
chunk0 groupedMM
chunk0 combine token_exchange launch ----------+
                                                |
chunk1 dispatch wait <-------------------------+
chunk1 groupedMM
chunk1 combine token_exchange launch ----------+

chunk0 combine wait <--------------------------+
chunk1 combine wait <--------------------------+
```

The core idea is:

```text
launch both dispatch exchanges early
use shared expert as filler
wait only when groupedMM needs routed tokens
launch combine as soon as each chunk's expert output is ready
```

The scheduler does not hardcode this textual order. It builds marker closures
and emits legal ready work with a global topological emitter. This diagram only
documents the target intuition that the generic algorithm is trying to recover:
pair the first exchanges, use independent compute as filler, and keep waits as
late as correctness allows.

### MoE-Root Backward Intuitive Target Schedule

Backward chunk order is reversed:

```text
chunk1 combine token_exchange launch
chunk0 combine token_exchange launch

chunk1 remat/groupedMM prefix -----------------+
chunk0 remat/groupedMM prefix                  | combine A2A in flight
                                                |
chunk1 combine wait <--------------------------+
chunk1 groupedMM input_grad
chunk1 dispatch token_exchange launch ---------+
                                                |
chunk0 combine wait <--------------------------+
chunk0 groupedMM input_grad
chunk0 dispatch token_exchange launch ---------+

chunk1 groupedMM wgrad
chunk0 groupedMM wgrad

chunk1 dispatch wait <-------------------------+
chunk0 dispatch wait <-------------------------+
```

Activation-checkpoint remat nodes are not special side-channel nodes. If they
belong to the selected backward body and their inputs are ready, they are
ordinary filler work and can run while EP communication is in flight.

As in forward, this is not a hardcoded schedule. The algorithm derives the
legal order from chunk metadata, marker labels, and dependencies; the backward
diagram explains why reversed chunk order and paired combine launches expose
useful remat/input-grad/wgrad filler.

### Transformer-Root Schedule

For `layers.*`, the chunk body includes attention and full transformer block
work around the MoE. The scheduler uses the regular wait-gated per-marker
closure schedule. It does not apply MoE first-marker pairing:

```text
chunk0 attention
chunk0 router/count/sync
chunk0 dispatch launch
chunk1 attention
chunk1 router/count/sync
chunk1 dispatch launch
ready filler
waits when needed
```

This is intentional. The MoE-specific sync and first-marker pairing only
matches the structure of `layers.*.moe` roots.

### Scheduler Validation

The scheduler validates:

```text
both chunks are present
marker counts match
marker labels match
each marker has exactly one wait in the same body
wait appears after launch in original graph
all body nodes are emitted once
no c10d wait is used as filler before waits are allowed
requested phase order materializes after global topological emission
graph lints and recompiles
```

## Token-Count Sync Copy Scheduling

The token dispatcher exposes private helper methods so tracing annotations can
mark the token-count and token-exchange pieces precisely:

```python
AllToAllTokenDispatcher._token_count_exchange
AllToAllTokenDispatcher._sync_token_count_exchange
AllToAllTokenDispatcher._dispatch_token_exchange
AllToAllTokenDispatcher._combine_token_exchange
```

Annotation mapping:

```text
_token_count_exchange:
  custom[EP_token_count_exchange] = dispatch

_sync_token_count_exchange:
  custom[EP_token_count_sync] = dispatch

_dispatch_token_exchange:
  custom[EP_token_exchange] = dispatch

_combine_token_exchange:
  custom[EP_token_exchange] = combine
```

Token-count sync exists because EP all-to-all split sizes are CPU lists:

```text
num_tokens_per_expert
  -> token-count all_to_all
  -> wait_tensor
  -> input_splits.to(cpu)
  -> output_splits.to(cpu)
  -> .tolist()
  -> dispatch all_to_all split sizes
```

In traced graphs, the CPU copies appear as `_to_copy` nodes. The dispatcher
keeps these copies synchronous during tracing so list reads cannot race stale
CPU data. The scheduler can then make a more global decision because it sees
both chunks' sync closures.

For MoE-root first-marker pairing:

```text
chunk0 pre-copy deps
chunk1 pre-copy deps
chunk0 input_splits _to_copy(non_blocking=True)
chunk0 output_splits _to_copy(non_blocking=True)
chunk1 input_splits _to_copy(non_blocking=True)
chunk1 output_splits _to_copy(non_blocking=False)
chunk0 post-copy list/scalar consumers
chunk1 post-copy list/scalar consumers
chunk0 first token_exchange launch
chunk1 first token_exchange launch
```

The final copy blocks. Earlier copies can be non-blocking because the sync
region is emitted together and CPU list/scalar consumers are emitted after the
copy group.

Validation:

```text
1. If sync optimization applies, both chunks must have sync parts.
2. Each chunk must have exactly two token-count sync CPU copies.
3. The copies must be CPU-destination _to_copy nodes.
4. Token-count sync copies must be inside the first marker closure.
5. Stray sync copies outside the first closure are errors.
```

## FSDP Bucketing And Dense-Region Scheduling

FSDP AG/RS scheduling is separate from EP token-exchange scheduling. It is
controlled by `--compile.enable_fsdp_dense_region_overlap`.

Motivation:

```text
AG/RS and EP token exchanges contend for communication resources.
When possible, AG/RS launches should be moved into neighboring dense attention
regions instead of landing inside MoE token-exchange windows.
```

### Mesh-Aware Bucket Counts

The effective expert-FSDP degree is:

```text
efsdp = dp_shard * cp * tp // ep
```

Bucket plan shape:

```text
MoE transformer layer, efsdp == 1:
  one bucket per transformer block

MoE transformer layer, efsdp > 1:
  dense/shared bucket:
    attention
    attention_norm
    ffn_norm
    moe.router
    moe.shared_experts

  expert bucket:
    moe.experts

dense-only transformer layer:
  one bucket per transformer block

top-level:
  tok_embeddings
  norm
  lm_head
```

The scheduler validates transformer-block bucket counts when expected counts are
provided:

```text
for each layer:
  fwd AG count == expected bucket count
  bwd AG count == expected bucket count
  bwd RS count == expected bucket count
```

### Dense Region Definition

For MoE layers:

```text
dense region =
  attention / dense compute nodes

excluded =
  MoE token exchange
  router/shared/expert nodes unless intentionally classified as dense bucket
  FSDP plumbing
  waits
  c10d collectives
  view/copy/slice/sym_size plumbing
```

For dense-only layers, the whole transformer block can be considered dense
compute.

### Launch Placement

Forward order is `layer i-1 -> layer i`:

```text
dense_fwd(i-1)
  AG(i) launch -------------------------+
  dense compute                         | AG in flight
                                        |
layer i param use
  AG(i) wait <--------------------------+
```

Backward order is `layer i+1 -> layer i -> layer i-1`:

```text
dense_bwd(i+1)
  AG(i) launch -------------------------+
  dense compute                         | AG in flight
                                        |
layer i backward param use
  AG(i) wait <--------------------------+

dense_bwd(i-1)
  RS(i) launch -------------------------+
  dense compute                         | RS in flight
                                        |
graph tail/output
  RS(i) wait/output-unpack <------------+
```

Top-level edge buckets:

```text
top-level norm/lm_head forward AG:
  can be placed in dense_fwd(last transformer block)

top-level norm/lm_head backward RS:
  can be placed in dense_bwd(last transformer block)

last transformer block backward AG:
  can be placed in top-level backward dense compute when present
```

Embedding edge buckets are left to the upstream bucketer.

### Dependency-Aware Target Selection

The pass does not blindly move a launch to the first dense node. It first
collects the FSDP launch chain:

```text
pre_bucket op(s)
all_gather / reduce_scatter launch
slice plumbing that belongs to FSDP infra
```

Then it picks the first dense target after all external inputs to that chain:

```python
min_pos = max(order[input] + 1 for input in external_inputs(chain))
target = first_dense_node_with_order_at_least(min_pos)
```

This matters for RS. The RS gradient input may be produced after the naive first
dense node, so moving the RS launch too early would violate data dependencies.

### Wait Sinking

AG waits stay where the parameter user needs them.

RS waits and pure output-unpack users are sunk to graph tail:

```text
reduce_scatter launch
  -> wait_tensor
  -> getitem/view/reshape/slice/copy output-unpack
  -> graph output
```

The pass allows pure output-unpack closure nodes while sinking. If a non-output
semantic user is found and strict mode is enabled, the pass errors.

### Strict Mode

With `strict=True`, the dense scheduler errors if:

```text
expected transformer bucket counts do not match
an interior AG/RS launch has no valid dense target
a launch chain cannot be moved without violating dependencies
an RS wait closure cannot be safely sunk
```

This is deliberate. Silent no-ops would make trace evidence ambiguous.

## Symbolic Shapes And Concretization

Graph chunking preserves selected unbacked symbols through chunking and EP
scheduling:

```text
before chunking:
  x shape: [u0, H]

inside chunk body:
  x0 shape: [u0//2, H]
  x1 shape: [u0//2, H]

after materialization:
  y shape: [u0, H]
```

Backend codegen is not assumed to handle every temporary unbacked chunk symbol,
so `concretize_ep_chunk_symbolic_shapes_pass` runs after EP scheduling.

Concretization contract:

```text
1. Collect selected chunk symbols and hints.
2. Rewrite graph args containing selected symbols to hinted concrete values
   when possible.
3. Rewrite node metadata (`val`, `example_value`) similarly.
4. Replace scalar SymInt/SymBool nodes with constants when their hints evaluate.
5. Validate no concretized guard became false.
6. Remove dead scalar/assert plumbing.
7. Clear chunk dynamic marks from example inputs.
8. Validate no selected chunk symbols remain in graph args or metadata.
```

Guard validation example:

```text
original guard:
  torch._assert(u0 == 8)

after chunk-symbol substitution with u0 -> 4:
  torch._assert(False)

result:
  ValueError, not silent cleanup
```

Scalar plumbing that can be removed when dead:

```text
sym_size
floordiv
mod
comparison ops
sym_ite
assert_scalar
```

Concretization is not a semantic pass. It is a backend-readiness pass after all
chunk/schedule reasoning has already consumed the symbolic contract.

## Numerics And Correctness Validation

Correctness is layered. No single test proves the whole stack.

### Static/Unit Contracts

Unit tests should cover:

```text
config validation
selected-symbol metadata
root discovery
live-in splitting and scalar rewrite
live-out cat materialization
backward grad add materialization
mutation alias ordering
non-EP collective rejection
token-exchange marker validation
sync-copy scheduling validation
FSDP dense target selection
pass ordering and selective disablement
false guard detection during concretization
```

### Single-GPU Bitwise

Single-GPU eager-vs-graph bitwise tests validate chunking semantics without real
distributed EP communication:

```text
cases:
  transformer_batch: layers.* batch
  moe_batch:         layers.*.moe batch
  moe_seq:           layers.*.moe seq

test mode:
  graph chunking
  early param-grad accumulation disabled
  compare against eager chunked model
```

This proves the chunking transform and pass composition, not real all-to-all
overlap.

### Distributed Numerics

Distributed DeepSeek-v3 loss comparisons validate real EP all-to-all and
distributed sharding composition:

```text
baseline:
  eager chunking

test:
  graph chunking
  early param-grad accumulation disabled for strict bitwise comparison

covered modes:
  transformer_batch
  moe_batch
  moe_seq
```

Dense FSDP scheduling is not part of the bitwise equivalence path unless a test
is explicitly about dense scheduling. It changes collective placement and is a
performance scheduling feature, not the semantic oracle.

### Integration

Integration tests exercise FlexAttention DSV3 debug models with both regional
and full Inductor:

```text
FlexAttention:
  graph transformer_batch
  graph moe_batch + dense FSDP scheduling
  graph moe_seq + dense FSDP scheduling

Parallelism:
  graph EP overlap uses dp_shard=8, ep=4, tp=1 today
  eager can cover TP compositions where graph chunking is not yet proven
```

Regional Inductor annotates FlexAttention HOPs for compilation while leaving
the rest of the graph interpreted. Full Inductor compiles the entire graph and
is covered as a separate integration mode.

### Trace Contract Validation

Trace/TLParse validation should check:

```text
EP schedule:
  dispatch/combine launches appear in intended chunk order
  waits appear only when their users require payloads
  shared expert/remat/groupedMM filler appears between launch and wait

sync copies:
  MoE first-marker sync copies are grouped across chunks
  all but final sync copy in the group are non-blocking

FSDP dense scheduling:
  AG/RS launches are in neighboring dense regions
  AG waits stay near parameter users
  RS waits/output-unpack closure sinks to graph tail
  AG/RS launches do not land in MoE token-exchange windows

chunk body:
  MoE graph chunk body has EP collectives only
```

### What Performance Evidence Must Show Later

Performance validation should separate these effects:

```text
EP chunking/scheduling:
  token-exchange launch order
  token-exchange wait placement
  amount of EP A2A overlapped with useful compute

sync-copy scheduling:
  number and blocking mode of CPU _to_copy nodes
  whether sync copies are grouped as intended

dense FSDP scheduling:
  AG/RS launch placement
  AG/RS wait placement
  whether AG/RS avoids MoE token-exchange windows

runtime metric:
  TPS and MFU by mode
  EP A2A time as percent of step span
  rank-aggregated behavior, not just rank 0
```

### Graph Debugging Workflow

Preferred workflow:

```text
1. Enable --compile.debug_graph_passes.
2. Inspect pass-before/pass-after graphs.
3. Verify chunked_region metadata and chunk ids.
4. Verify token-exchange marker labels.
5. Verify final node order in TLParse before profiler interpretation.
6. Use --compile.disable_passes for narrow local ablations when a pass boundary
   needs to be isolated.
7. Remove temporary instrumentation before landing.
```

Avoid relying on pass logs alone for overlap claims. The profiler trace and
TLParse graph must show actual launch and wait placement.

### Graph Chunking With TP

Current behavior:

```text
if compile.ep_overlap.strategy == "graph" and tensor_parallel_degree > 1:
    raise ValueError
```

Reason:

```text
Graph chunking runs after DTensor lowering.

At that point, the graph contains physical rank-local tensors plus TP/SP layout
helpers, collectives, waits, views, index maps, split factors, and scalar
plumbing. There is no placement object left that says which symbolic use is a
logical activation extent and which symbolic use is layout machinery.

Blindly substituting selected_symbol -> selected_symbol // 2 can therefore
chunk TP/SP helper paths, not just model activations.
```

This is not fundamentally TP-specific. The general unsafe condition is:

```text
The chunked logical dimension participates in a non-EP layout-changing path
inside the proposed graph chunk body, or the pass blindly rewrites layout
helper expressions derived from that dimension.
```

### Future Safe Graph TP Support

The future graph-native rule should be:

```text
chunk body must contain no non-EP collectives
```

To support TP/SP, the planner should move boundaries through pure layout chains:

```text
live-in path:

  candidate live-in: [B, u0 // tp, H]
      |
      v
  TP/SP all_gather + wait + view/cat/slice
      |
      v
  better live-in:   [B, u0, H]
      |
      v
  split logical full sequence into chunks

live-out path:

  chunk outputs:    [B, u0//2, H] per chunk
      |
      v
  cat full output:  [B, u0, H]
      |
      v
  TP/SP reduce_scatter + wait + view/slice
      |
      v
  final local shard [B, u0 // tp, H]
```

Guardrails:

```text
only cross allowlisted pure layout/view/scalar ops
only cross known non-EP layout collectives as boundary prologue/epilogue
do not cross router/top-k/groupedMM/shared-expert/dense compute
require a unique dominance-safe chain
require token/order preservation, not shape equality alone
validate final chunk body contains EP collectives only
reject if any proof is missing
```

Desired MoE TP/SP contract:

```text
MoE prologue:
  Shard(1) -> Replicate

MoE body:
  logical full-sequence chunks
  EP collectives only
  local router/groupedMM/shared-expert compute

MoE epilogue:
  Partial/Replicate -> Shard(1)
```

### Future Eager-Chunking Optimization

Another future direction is to use eager chunking as the always-correct program
and optimize duplicated boundary chains after tracing.

Live-in hoisting:

```text
current eager-lowered shape:

  split
    -> per-chunk layout all_gather/wait/view
    -> chunk body

optimized:

  layout all_gather/wait/view
    -> split logical full tensor
    -> chunk body
```

Live-out sinking:

```text
current eager-lowered shape:

  chunk body
    -> per-chunk reduce_scatter/wait/view
    -> cat

optimized:

  chunk body
    -> cat logical full tensor
    -> reduce_scatter/wait/view
```

These rewrites require formal equivalence:

```text
L_i(S_i(x)) == S_i(L(x))
Cat_i(R_i(y_i)) == R(Cat_i(y_i))
```

The chain must be pure layout or a proven linear reduction, process groups must
match, token/order mapping must be preserved, and the optimized body must keep
non-EP collectives outside the EP chunk loop.

## Error Model

Use `ValueError` for user-facing config or graph-contract failures. Use
assertions only for impossible internal states.

Errors should include as much actionable context as available:

```text
node name
target overload
module FQN
root FQN
chunk id
direction
shape argument
fake shape
process group / mesh axis
blocking dependency
```

Important errors:

```text
invalid EP overlap config
seq chunking outside layers.*.moe
graph chunking with tp > 1
no selected roots
missing real EP all-to-all in production EP config
odd selected extent hint
unsupported non-EP collective in graph chunk body
unsupported live-out materialization
token-exchange marker mismatch
token-count sync copy mismatch
false guard after concretization
FSDP dense scheduling blocker in strict mode
```

sanketpurandare:
Stacked PRs:
 * #3328
 * __->__#3325
 * #3363

--- --- ---

### [graph_trainer] Add graph EP chunking pass

Add the FX graph chunking implementation for EP overlap, including chunk metadata propagation, trace-call input preparation, symbolic chunk shape handling, and graph chunk pass tests. This builds on the eager metadata contract and keeps scheduling as a later pass.

Concretize graph-chunk symbolic shapes before Inductor so production graph chunking does not rely on regional Inductor handling raw chunk expressions such as u0 // 2.

Test Plan:\n- conda run -n pt-dev python -m pytest -q torchtitan/experiments/graph_trainer/tests/test_passes.py::TestChunkPasses\n- conda run -n pt-dev python -m pytest -q torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py::TestDSv3FlexAttnBitwiseDeterministic::test_ep_chunk_matches_eager_chunking_bitwise

### DIFF
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -66,6 +66,7 @@ def build_decoder_config_for_backend(
 
 
 _MODULE_FQN = "module_fqn"
+_EP_TOKEN_EXCHANGE = "EP_token_exchange"
 _NOT_IN_LAYERS = -1
 
 
--- a/torchtitan/experiments/graph_trainer/ep_chunk_pass.py
+++ b/torchtitan/experiments/graph_trainer/ep_chunk_pass.py
@@ -1,0 +1,2449 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Graph chunking for EP overlap.
+
+The pass is a graph implementation of the eager wrapper:
+
+    split selected live-ins -> run selected module on each chunk -> materialize
+    full live-outs when a non-chunked consumer needs them.
+
+The correctness target is numerical equivalence to eager chunking, not an
+identical lowered graph. ``module_fqn`` seeds the candidate region, but body
+ownership is dataflow based: only nodes reached by chunk-influenced values are
+duplicated. Boundary nodes that the pass creates (split/getitem/scalar
+halves/cat/add) use the parent-wrapper metadata, while copied body nodes keep
+the selected module metadata and receive ``chunk_id``.
+
+Terms:
+- live-in: a value consumed by the selected body and produced outside it.
+  Tensor live-ins whose fake shape contains the selected unbacked chunk symbol
+  are split. Per-chunk provenance from an earlier region is reused directly.
+  Static state, params, buffers, and other full values are shared by both chunks.
+- live-out: a value produced by the selected body and consumed outside it.
+  Chunked consumers receive the matching per-chunk value through provenance.
+  Full consumers receive a materialized value: ``cat`` for values carrying the
+  chunk dimension and ``add`` for backward parameter-gradient reductions.
+- provenance: the internal ``full_value -> (chunk0, chunk1)`` table. It is not
+  materialized as tuple nodes because no downstream pass needs tuple plumbing.
+
+The implementation deliberately stays local. It does not special-case SAC,
+activation offload, or bucketing nodes; those interactions are expressed as
+ordinary dataflow through live-ins and live-outs. The selected unbacked SymInt is
+the only source of truth for chunk influence; graph metadata records only
+symbol->hint pairs so cleanup can erase the symbols after scheduling. The
+selected dimension is symbolic while this pass and scheduling run, then
+``concretize_ep_chunk_symbolic_shapes_pass`` replaces chunk-derived symbolic
+shapes with their config hints before backend codegen.
+
+V1 creates two equal chunks because the EP schedule is pairwise. The data
+structures use per-chunk maps, so extending to N equal chunks is mechanical, but
+the overlap scheduler and tests must be generalized at the same time.
+
+Pseudo-code
+===========
+1. Mark the requested input dimension dynamic before tracing, then import the
+   selected symbol hints onto placeholders after tracing.
+2. Discover selected module roots and plan forward/backward body nodes by
+   following only values influenced by the selected chunk symbol.
+3. Split chunked live-ins, reuse provenance from earlier chunked regions, and
+   create half-size symbolic scalars used by copied body nodes.
+4. Rewrite one chunk body in place and copy the peer chunk body, preserving
+   eager-like metadata, fake values, and mutation alias dependencies.
+5. For every live-out, either route per-chunk provenance to chunked consumers or
+   materialize a full value with cat/add for full consumers.
+6. Erase unused copied nodes, validate selected-symbol shapes, lint, recompile,
+   and leave final symbol concretization to the cleanup pass after scheduling.
+"""
+
+from __future__ import annotations
+
+import operator
+from dataclasses import dataclass
+from typing import Any, Literal
+
+import sympy
+import torch
+import torch.fx as fx
+from torch.utils._pytree import tree_leaves, tree_map
+from torch.utils._sympy.functions import FloorDiv
+
+from torchtitan.experiments.graph_trainer.common_utils import (
+    _EP_TOKEN_EXCHANGE,
+    _get_module_fqn,
+    _is_backward_node,
+    _MODULE_FQN,
+)
+from torchtitan.experiments.graph_trainer.configs import validate_ep_overlap_config
+from torchtitan.experiments.graph_trainer.ep_pass_utils import (
+    _eval_hint,
+    chunk_symbol_hints_for_mode,
+    CHUNK_SYMBOL_HINTS_META,
+    depends_on_chunk_symbol,
+    expr_key,
+    free_symbols,
+    is_c10d_functional_node,
+    is_module_fqn_inside_root,
+    is_view_target,
+    ordered_nodes,
+    same_shape_with_hints,
+    same_symbolic_expr,
+    same_tensor_domain,
+    statically_equals_hint,
+    symbolic_expr,
+    tensor_meta,
+    view_shape_arg_index,
+)
+from torchtitan.experiments.graph_trainer.registry import (
+    register_trace_call_input_preparer,
+    register_trace_input_preparer,
+)
+from torchtitan.tools.logging import logger
+
+aten = torch.ops.aten
+ChunkMode = Literal["batch", "seq"]
+
+
+# Step 0: Metadata and small graph utilities used by the lowering steps below.
+
+
+@dataclass(frozen=True)
+class _Region:
+    root_fqn: str
+    synthetic_parent_fqn: str
+    is_backward: bool
+    nodes: tuple[fx.Node, ...]
+
+
+@dataclass
+class _Plan:
+    region: _Region
+    body: frozenset[fx.Node]
+    live_ins: frozenset[fx.Node]
+    live_out_users: dict[fx.Node, tuple[fx.Node, ...]]
+    grad_escape_boundaries: frozenset[fx.Node]
+
+
+@dataclass(frozen=True)
+class _ChunkDim:
+    dim: int
+    hint: int | None
+
+
+@dataclass(frozen=True)
+class _MaterializationBoundary:
+    node: fx.Node
+    suffix: tuple[fx.Node, ...]
+
+
+@dataclass(frozen=True)
+class _MatchedRoot:
+    root_fqn: str
+    synthetic_parent_fqn: str
+
+
+def _custom_meta(node: fx.Node) -> dict[str, Any]:
+    """Return mutable custom metadata when present, otherwise an empty dict."""
+    custom = node.meta.get("custom")
+    return custom if isinstance(custom, dict) else {}
+
+
+def _mutates_input(node: fx.Node) -> bool:
+    """Return whether a call_function schema writes through an aliased input."""
+    return bool(_mutable_arg_indices(node))
+
+
+def _mutable_arg_indices(node: fx.Node) -> tuple[int, ...]:
+    """Return positional schema arg indices written by ``node``."""
+    schema = getattr(node.target, "_schema", None)
+    if node.op != "call_function" or schema is None:
+        return ()
+    return tuple(
+        idx
+        for idx, arg in enumerate(schema.arguments)
+        if arg.alias_info is not None and arg.alias_info.is_write
+    )
+
+
+def _is_semantic_user(node: fx.Node) -> bool:
+    """Return whether a user must be preserved when crossing a chunk boundary."""
+    return node.op == "output" or bool(node.users) or _mutates_input(node)
+
+
+def _copy_meta(meta: dict[str, Any]) -> dict[str, Any]:
+    copied = dict(meta)
+    if "custom" in copied:
+        copied["custom"] = dict(copied["custom"])
+    if "unbacked_bindings" in copied:
+        copied["unbacked_bindings"] = dict(copied["unbacked_bindings"])
+    return copied
+
+
+def _set_direction_meta(node: fx.Node, *, is_backward: bool) -> None:
+    if is_backward:
+        node.meta["autograd_backward"] = True
+    else:
+        node.meta.pop("autograd_backward", None)
+
+
+def _set_chunk_meta(
+    node: fx.Node,
+    *,
+    region: _Region,
+    role: str,
+    chunk_id: int | None = None,
+) -> None:
+    _set_direction_meta(node, is_backward=region.is_backward)
+    node.meta["chunked_region_fqn"] = region.root_fqn
+    node.meta["chunked_region_is_backward"] = region.is_backward
+    node.meta["chunked_region_producer"] = "graph"
+    node.meta["chunked_region_role"] = role
+    if chunk_id is None:
+        node.meta.pop("chunk_id", None)
+    else:
+        node.meta["chunk_id"] = chunk_id
+
+
+def _set_synthetic_meta(
+    node: fx.Node,
+    *,
+    region: _Region,
+    role: str,
+    chunk_id: int | None = None,
+) -> None:
+    """Attach parent-wrapper metadata to nodes introduced by graph chunking."""
+    custom = dict(_custom_meta(node))
+    parent_fqn = region.synthetic_parent_fqn
+    if parent_fqn:
+        custom[_MODULE_FQN] = parent_fqn
+        node.meta["custom"] = custom
+    elif custom:
+        custom.pop(_MODULE_FQN, None)
+        node.meta["custom"] = custom
+    else:
+        node.meta.pop("custom", None)
+    _set_chunk_meta(node, region=region, role=role, chunk_id=chunk_id)
+
+
+def _pattern_root(pattern: str, fqn: str) -> _MatchedRoot | None:
+    """Return the matched root and synthetic-parent FQN for a module pattern."""
+    pattern_parts = pattern.split(".")
+    fqn_parts = fqn.split(".")
+    if len(fqn_parts) < len(pattern_parts):
+        return None
+    for pattern_part, fqn_part in zip(pattern_parts, fqn_parts):
+        if pattern_part != "*" and pattern_part != fqn_part:
+            return None
+    root_parts = fqn_parts[: len(pattern_parts)]
+    return _MatchedRoot(
+        root_fqn=".".join(root_parts),
+        synthetic_parent_fqn=".".join(root_parts[:-1]),
+    )
+
+
+def _region_roots(gm: fx.GraphModule, patterns: list[str]) -> list[_MatchedRoot]:
+    """Find concrete module roots matching the configured chunk patterns."""
+    roots: dict[str, _MatchedRoot] = {}
+    for node in gm.graph.nodes:
+        fqn = _get_module_fqn(node)
+        if not fqn:
+            continue
+        for pattern in patterns:
+            if (matched := _pattern_root(pattern, fqn)) is not None:
+                roots.setdefault(matched.root_fqn, matched)
+    return sorted(
+        roots.values(),
+        key=lambda matched: (matched.root_fqn.count("."), matched.root_fqn),
+    )
+
+
+def _region_nodes(
+    gm: fx.GraphModule,
+    root: str,
+    *,
+    is_backward: bool,
+    symbol_hints: dict[object, int],
+    grad_escape_boundaries: set[fx.Node],
+    per_chunk_sources: set[fx.Node],
+) -> tuple[fx.Node, ...]:
+    """Return selected-symbol-influenced nodes inside one root/direction."""
+    candidates = {
+        node
+        for node in gm.graph.nodes
+        if node.op == "call_function"
+        and node.meta.get("chunked_region_role") is None
+        and _is_backward_node(node) == is_backward
+        and (fqn := _get_module_fqn(node))
+        and is_module_fqn_inside_root(fqn, root)
+    }
+    if is_backward:
+        changed = True
+        while changed:
+            changed = False
+            for node in gm.graph.nodes:
+                neighbors = (*node.all_input_nodes, *node.users)
+                if (
+                    node in candidates
+                    or node.op != "call_function"
+                    or _get_module_fqn(node)
+                    or not _is_backward_node(node)
+                    or is_c10d_functional_node(node)
+                ):
+                    continue
+                anchored_to_root = any(neighbor in candidates for neighbor in neighbors)
+                if not anchored_to_root:
+                    anchored_to_root = any(
+                        neighbor in per_chunk_sources
+                        and is_module_fqn_inside_root(_get_module_fqn(neighbor), root)
+                        for neighbor in neighbors
+                    )
+                if not anchored_to_root:
+                    continue
+                neighbor_fqns = [
+                    fqn for neighbor in neighbors if (fqn := _get_module_fqn(neighbor))
+                ]
+                if not neighbor_fqns or all(
+                    is_module_fqn_inside_root(fqn, root) for fqn in neighbor_fqns
+                ):
+                    candidates.add(node)
+                    changed = True
+    return _chunk_influenced_region_nodes(
+        gm,
+        candidates,
+        is_backward,
+        symbol_hints,
+        grad_escape_boundaries,
+        per_chunk_sources,
+    )
+
+
+def _node_depends_on_chunk_symbol(
+    node: fx.Node, symbol_hints: dict[object, int]
+) -> bool:
+    """Return whether a node's fake value carries the selected chunk symbol."""
+    return any(
+        depends_on_chunk_symbol(value, symbol_hints)
+        for value in tree_leaves(node.meta.get("val"))
+    )
+
+
+def _is_ep_collective_node(node: fx.Node, body: frozenset[fx.Node]) -> bool:
+    """Return whether a c10d node belongs to EP token exchange/count plumbing."""
+    custom = _custom_meta(node)
+    if "EP" in custom or _EP_TOKEN_EXCHANGE in custom:
+        return True
+    if (
+        node.target == torch.ops._c10d_functional.wait_tensor.default
+        and node.all_input_nodes
+        and node.all_input_nodes[0] in body
+    ):
+        waited_custom = _custom_meta(node.all_input_nodes[0])
+        return "EP" in waited_custom or _EP_TOKEN_EXCHANGE in waited_custom
+    return False
+
+
+def _validate_no_non_ep_collectives(plan: _Plan) -> None:
+    """Graph chunking may only duplicate EP collectives inside chunk bodies."""
+    offenders = [
+        node
+        for node in plan.body
+        if is_c10d_functional_node(node) and not _is_ep_collective_node(node, plan.body)
+    ]
+    if offenders:
+        order = {node: idx for idx, node in enumerate(plan.region.nodes)}
+        details = ", ".join(
+            f"{node.name}:{getattr(node.target, '__name__', node.target)}"
+            for node in sorted(offenders, key=order.get)
+        )
+        raise ValueError(
+            "Graph EP chunking requires chunk bodies to contain only EP "
+            f"collectives. Found non-EP c10d node(s) in {plan.region.root_fqn}: "
+            f"{details}."
+        )
+
+
+def _owned_chunk_dependencies(
+    body: set[fx.Node],
+    candidates: set[fx.Node],
+    symbol_hints: dict[object, int],
+) -> set[fx.Node]:
+    """Return selected-symbol producers owned by an eager chunk body.
+
+    The selected module root defines the eager chunk body. In flattened FX,
+    Python shape computations and shape-derived tensor values can appear as
+    ancestors of body nodes even though they do not data-depend on an activation
+    live-in. These are part of invoking the module on a chunk, not live-ins to
+    split. Ownership is still bounded: a producer is included only when it is in
+    the same candidate root and every semantic user is also inside this closure.
+    """
+
+    deps: set[fx.Node] = set()
+    work = [inp for node in body for inp in node.all_input_nodes]
+    while work:
+        node = work.pop()
+        if (
+            node in body
+            or node in deps
+            or not _node_depends_on_chunk_symbol(node, symbol_hints)
+        ):
+            continue
+        # Shape scalar helpers do not always carry module FQN metadata, but
+        # they are still part of invoking the selected body when every escaping
+        # user remains in the chunk closure.
+        if node not in candidates and not (
+            node.op == "call_function" and tensor_meta(node) is None
+        ):
+            continue
+        deps.add(node)
+        work.extend(node.all_input_nodes)
+
+    changed = True
+    while changed:
+        changed = False
+        owned = body | deps
+        for dep in tuple(deps):
+            if any(user not in owned and _is_semantic_user(user) for user in dep.users):
+                deps.remove(dep)
+                changed = True
+    return deps
+
+
+def _grad_output_suffix_nodes(
+    gm: fx.GraphModule, grad_escape_boundaries: set[fx.Node]
+) -> set[fx.Node]:
+    """Return post-materialization grad-output plumbing outside chunk bodies."""
+    suffix_nodes: set[fx.Node] = set()
+    for output in _graph_output_nodes(gm)[1:]:
+        node = output
+        seen: set[fx.Node] = set()
+        while (
+            node not in seen
+            and node not in grad_escape_boundaries
+            and tensor_meta(node) is not None
+        ):
+            seen.add(node)
+            suffix_nodes.add(node)
+            inputs = _tensor_inputs(node)
+            if len(inputs) != 1:
+                break
+            node = inputs[0]
+    return suffix_nodes
+
+
+def _semantic_body_closure(body: set[fx.Node]) -> set[fx.Node]:
+    """Keep only nodes needed to produce chunked tensor outputs.
+
+    Symbolic shape helpers are part of the body only when they feed real tensor
+    compute inside the body. Shape-only template chains whose only semantic
+    consumer is a full parent-region allocation should remain full; otherwise
+    the pass would need to "materialize" scalar metadata that has no meaningful
+    chunk reconstruction.
+    """
+    seeds = {
+        node
+        for node in body
+        if _mutates_input(node)
+        or any(user.op == "output" for user in node.users)
+        or (
+            tensor_meta(node) is not None
+            and any(user not in body and _is_semantic_user(user) for user in node.users)
+        )
+    }
+    required: set[fx.Node] = set()
+    work = list(seeds)
+    while work:
+        node = work.pop()
+        if node in required:
+            continue
+        required.add(node)
+        work.extend(inp for inp in node.all_input_nodes if inp in body)
+    return required
+
+
+def _chunk_influenced_region_nodes(
+    gm: fx.GraphModule,
+    candidates: set[fx.Node],
+    is_backward: bool,
+    symbol_hints: dict[object, int],
+    grad_escape_boundaries: set[fx.Node],
+    per_chunk_sources: set[fx.Node],
+) -> tuple[fx.Node, ...]:
+    """Shrink module candidates to the body that must be duplicated per chunk."""
+    influenced: set[fx.Node] = set()
+    for node in gm.graph.nodes:
+        if node not in candidates:
+            continue
+        if not is_backward and any(
+            inp not in candidates
+            and inp.op == "call_function"
+            and tensor_meta(inp) is not None
+            and _is_backward_node(inp)
+            for inp in node.all_input_nodes
+        ):
+            raise ValueError(
+                f"Chunk pass crossed into the opposite graph direction through {node.name}."
+            )
+        if any(
+            inp in influenced
+            or inp in per_chunk_sources
+            or (inp not in candidates and _is_tensor_with_chunk_dim(inp, symbol_hints))
+            for inp in node.all_input_nodes
+        ):
+            influenced.add(node)
+
+    body: set[fx.Node] = set()
+    work = list(influenced)
+    while work:
+        node = work.pop()
+        if node in body:
+            continue
+        body.add(node)
+        work.extend(
+            inp
+            for inp in node.all_input_nodes
+            if inp in candidates and inp in influenced
+        )
+    body |= _owned_chunk_dependencies(body, candidates, symbol_hints)
+    body = _semantic_body_closure(body)
+    if is_backward:
+        body -= _grad_output_suffix_nodes(gm, grad_escape_boundaries)
+    return tuple(node for node in gm.graph.nodes if node in body)
+
+
+def _find_regions(
+    gm: fx.GraphModule,
+    patterns: list[str],
+    *,
+    symbol_hints: dict[object, int],
+    grad_escape_boundaries: set[fx.Node],
+    per_chunk_sources: set[fx.Node],
+) -> list[_Region]:
+    """Build ordered forward/backward regions for all matched roots."""
+    regions = []
+    for matched in _region_roots(gm, patterns):
+        for is_backward in (False, True):
+            nodes = _region_nodes(
+                gm,
+                matched.root_fqn,
+                is_backward=is_backward,
+                symbol_hints=symbol_hints,
+                grad_escape_boundaries=grad_escape_boundaries,
+                per_chunk_sources=per_chunk_sources,
+            )
+            if nodes:
+                regions.append(
+                    _Region(
+                        matched.root_fqn,
+                        matched.synthetic_parent_fqn,
+                        is_backward,
+                        nodes,
+                    )
+                )
+    order = ordered_nodes(gm)
+    return sorted(regions, key=lambda region: order[region.nodes[0]])
+
+
+# Step 1: Populate/prepare the selected dynamic dimension used as chunk truth.
+
+
+def populate_chunk_dim_metadata_pass(
+    gm: fx.GraphModule,
+    example_inputs: tuple[Any, ...] | None = None,
+    *,
+    mode: ChunkMode,
+) -> fx.GraphModule:
+    """Step 1: record selected dynamic symbol hints on placeholders."""
+    del example_inputs
+    if mode not in ("batch", "seq"):
+        raise ValueError(f"Unknown chunk mode: {mode!r}")
+    # The chunk pass uses the selected unbacked SymInt as its only source of
+    # truth. This populate pass records only symbol->hint pairs on placeholders
+    # so later cleanup can concretize exactly the symbols we introduced; it does
+    # not propagate per-node chunk-dimension metadata.
+    symbol_hints: dict[object, int] = chunk_symbol_hints_for_mode(gm, mode)
+    if not symbol_hints:
+        raise ValueError(
+            "ep_overlap graph chunking expected at least one selected dynamic "
+            "placeholder dimension, but none was found."
+        )
+    for node in gm.graph.nodes:
+        if node.op != "placeholder":
+            continue
+        val = tensor_meta(node)
+        node_symbols = (
+            {
+                symbol: hint
+                for symbol, hint in symbol_hints.items()
+                if any(symbol in free_symbols(extent) for extent in val.shape)
+            }
+            if val is not None
+            else {}
+        )
+        if node_symbols:
+            node.meta[CHUNK_SYMBOL_HINTS_META] = node_symbols
+    return gm
+
+
+def mark_chunk_dynamic_dims(tensor: torch.Tensor, *, mode: ChunkMode) -> None:
+    """Step 1: mark the user-selected tensor dimension unbacked with a hint."""
+    from torch._dynamo.decorators import mark_unbacked
+
+    dim = {"batch": 0, "seq": 1}.get(mode)
+    if dim is None:
+        raise ValueError(f"Unknown chunk mode: {mode!r}")
+    if tensor.dim() <= dim:
+        raise ValueError(
+            f"Cannot mark {mode} dim {dim} for shape {tuple(tensor.shape)}."
+        )
+    chunk_dim_shape = int(tensor.shape[dim])
+    if chunk_dim_shape < 2 or chunk_dim_shape % 2:
+        raise ValueError(
+            "EP overlap graph chunking requires an even selected dimension, "
+            f"got {mode} size {chunk_dim_shape} for shape {tuple(tensor.shape)}."
+        )
+    mark_unbacked(
+        tensor,
+        dim,
+        hint_override=chunk_dim_shape,
+        # Graph chunking splits the selected dimension into two equal pieces.
+        # Use the original half-size as the lower bound so downstream shape
+        # checks can still prove non-singleton subchunks after further local
+        # chunking (for example ChunkedCELoss over sequence).
+        min=chunk_dim_shape // 2,
+        max=chunk_dim_shape,
+        specialize_on=[
+            lambda extent, chunk_dim_shape=chunk_dim_shape: extent == chunk_dim_shape
+        ],
+        shape_id=f"torchtitan_chunk_{mode}",
+    )
+
+
+@register_trace_input_preparer("ep_overlap")
+def prepare_ep_overlap_trace_inputs(
+    compile_config: Any,
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
+) -> None:
+    """Step 1: prepare trace inputs for graph chunking when EP overlap is enabled."""
+    if not compile_config.ep_overlap.enabled:
+        return
+    mode, chunk_strategy, _ = validate_ep_overlap_config(compile_config.ep_overlap)
+    if chunk_strategy == "eager":
+        return
+    dim = {"batch": 0, "seq": 1}[mode]
+    if not args or not isinstance(args[0], torch.Tensor):
+        raise ValueError("ep_overlap tracing expects first user input to be a Tensor")
+    hint = int(args[0].shape[dim])
+
+    def mark_leaf(value: object) -> None:
+        if (
+            isinstance(value, torch.Tensor)
+            and value.dim() > dim
+            and int(value.shape[dim]) == hint
+        ):
+            # TODO(sanketpurandare): replace this bounded shape-equality
+            # heuristic with explicit token-grid input role metadata.
+            mark_chunk_dynamic_dims(value, mode=mode)
+
+    # The traced training step is `(inputs, labels, global_valid_tokens,
+    # extra_inputs, extra_kwargs)`. Mark only tensors with semantic training
+    # roles tied to the model token grid; skip scalar loss bookkeeping.
+    model_inputs = [args[0]]
+    if len(args) > 1:
+        model_inputs.append(args[1])
+    if len(args) > 3:
+        model_inputs.append(args[3])
+    if len(args) > 4:
+        model_inputs.append(args[4])
+    model_inputs.append(kwargs)
+
+    for leaf in tree_leaves(model_inputs):
+        mark_leaf(leaf)
+
+
+@register_trace_call_input_preparer("ep_overlap")
+def prepare_ep_overlap_trace_call_inputs(
+    compile_config: Any,
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
+) -> tuple[tuple[Any, ...], dict[str, Any]] | None:
+    """Bind FlexAttention mask lengths to fake token-grid dims during tracing."""
+    if not compile_config.ep_overlap.enabled:
+        return None
+    _, chunk_strategy, _ = validate_ep_overlap_config(compile_config.ep_overlap)
+    if chunk_strategy == "eager" or len(args) <= 3 or not isinstance(args[3], dict):
+        return None
+
+    extra_kwargs = args[3]
+    positions = extra_kwargs.get("positions")
+    attention_masks = extra_kwargs.get("attention_masks")
+    if not isinstance(positions, torch.Tensor) or attention_masks is None:
+        return None
+    if positions.dim() < 2:
+        return None
+
+    from torch.nn.attention.flex_attention import BlockMask
+
+    seq_len = positions.shape[1]
+
+    def rebind(mask: object) -> object:
+        if not isinstance(mask, BlockMask):
+            return mask
+        return BlockMask(
+            seq_lengths=(seq_len, seq_len),
+            kv_num_blocks=mask.kv_num_blocks,
+            kv_indices=mask.kv_indices,
+            full_kv_num_blocks=mask.full_kv_num_blocks,
+            full_kv_indices=mask.full_kv_indices,
+            q_num_blocks=mask.q_num_blocks,
+            q_indices=mask.q_indices,
+            full_q_num_blocks=mask.full_q_num_blocks,
+            full_q_indices=mask.full_q_indices,
+            BLOCK_SIZE=mask.BLOCK_SIZE,
+            mask_mod=mask.mask_mod,
+            dq_write_order=mask.dq_write_order,
+            dq_write_order_full=mask.dq_write_order_full,
+            dq_kv_order=mask.dq_kv_order,
+            dq_kv_order_spt=mask.dq_kv_order_spt,
+        )
+
+    if isinstance(attention_masks, dict):
+        rebound_masks = {key: rebind(mask) for key, mask in attention_masks.items()}
+    else:
+        rebound_masks = rebind(attention_masks)
+    if rebound_masks is attention_masks:
+        return None
+
+    rebound_extra_kwargs = dict(extra_kwargs)
+    rebound_extra_kwargs["attention_masks"] = rebound_masks
+    rebound_args = list(args)
+    rebound_args[3] = rebound_extra_kwargs
+    return tuple(rebound_args), kwargs
+
+
+# Step 2: Discover selected-symbol body regions and classify their boundaries.
+
+
+def _chunk_dim_from_symbols(
+    val: torch.Tensor | None, symbol_hints: dict[object, int]
+) -> _ChunkDim | None:
+    """Map the selected unbacked symbol to the tensor dimension it appears in."""
+    if val is None or not symbol_hints:
+        return None
+    matches: list[_ChunkDim] = []
+    for dim, size in enumerate(val.shape):
+        if not (free_symbols(size) & symbol_hints.keys()):
+            continue
+        hint = _eval_hint(size, symbol_hints)
+        matches.append(_ChunkDim(dim, int(hint) if hint is not None else None))
+    if len(matches) > 1:
+        raise ValueError(
+            f"Chunk pass found selected chunk symbols in multiple dims: {tuple(val.shape)}."
+        )
+    return matches[0] if matches else None
+
+
+def _validate_selected_symbol_shapes(gm: fx.GraphModule, *, mode: ChunkMode) -> None:
+    """Check every chunk-created selected-symbol tensor is still classifiable."""
+    symbol_hints = chunk_symbol_hints_for_mode(gm, mode)
+    if not symbol_hints:
+        return
+    for node in gm.graph.nodes:
+        if node.meta.get("chunked_region_role") is None:
+            continue
+        val = tensor_meta(node)
+        if val is None:
+            continue
+        if not any(free_symbols(size) & symbol_hints.keys() for size in val.shape):
+            continue
+        if _chunk_dim_from_symbols(val, symbol_hints) is None:
+            raise ValueError(
+                "Chunk pass could not classify selected-symbol shape for "
+                f"{node.name}: {tuple(val.shape)}."
+            )
+
+
+def _chunk_dim_for_live_in(node: fx.Node, symbol_hints: dict[object, int]) -> int:
+    dim_hint = _chunk_dim_from_symbols(tensor_meta(node), symbol_hints)
+    val = tensor_meta(node)
+    if dim_hint is None or val is None:
+        raise ValueError(f"{node.name} does not carry the selected chunk symbol.")
+    dim, hint = dim_hint.dim, dim_hint.hint
+    if dim >= val.dim():
+        raise ValueError(f"Selected chunk dim {dim} is invalid for {node.name}.")
+    if hint is not None and hint % 2:
+        raise ValueError(
+            f"Cannot split selected chunk dimension of {node.name}: {hint}."
+        )
+    return dim
+
+
+def _split_view_meta(val: torch.Tensor, dim: int) -> torch.Tensor:
+    shape = list(val.shape)
+    shape[dim] = shape[dim] // 2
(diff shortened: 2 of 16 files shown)


SherlockNoMad:
would be easier for me to review if you can show a tlp, so I can diff the graph before and after the pass.

torchtitan/models/deepseek_v3/model.py
```diff
@@ -101,7 +101,8 @@ def forward(
         attention_masks: AttentionMasksType | None,
         positions: torch.Tensor | None = None,
     ):
-        bsz, seqlen, _ = x.size()
+        bsz = x.size(0)
+        seqlen = x.size(1)
```
SherlockNoMad: hmm... I am surprised this rewrite is needed?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -487,6 +487,14 @@ def fn_with_subclass_handling(*plain_args: Any) -> list:
...
         # Must run before DCE so that forward nodes used for matching aren't removed.
         _copy_fwd_metadata_to_bw_nodes(traced)
+        for node, arg in zip(
+            (n for n in traced.graph.nodes if n.op == "placeholder"),
+            unwrapped_args,
+            strict=True,
+        ):
+            chunk_dims = getattr(arg, "_torchtitan_chunk_dims", None)
+            if chunk_dims is not None:
+                node.meta["torchtitan_chunk_dims"] = dict(chunk_dims)
```
SherlockNoMad: model specific or pass specific optimization shouldn't live in tracer. 
we need to keep tracer general and portable.

SherlockNoMad: can this be a preprocessing pass?

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -168,13 +172,33 @@ def compile_time_passes(
...
     ]
+    if config.compile.chunk_batch_modules:
+        passes.append(
+            functools.partial(
+                chunk_batch_pass,
+                module_patterns=config.compile.chunk_batch_modules,
+                num_static_inputs=traced_result.num_static_inputs,
+            )
+        )
+    if config.compile.chunk_seq_modules:
```
SherlockNoMad: are these two mutually exclusive? 
if so, use elif

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -70,6 +73,57 @@ def annotate_module_fqns(model: nn.Module) -> None:
...
+# ---------------------------------------------------------------------------
+# These helpers are shared by chunk passes and tests around graph_trainer's
+# traced FX metadata. Keep chunk-specific rewrite policy in chunk_passes.py.
+
+
+def _is_module_fqn_inside_root(fqn: str, root_fqn: str) -> bool:
+    return fqn == root_fqn or fqn.startswith(root_fqn + ".")
+
+
+def _node_index(node: torch.fx.Node) -> int | None:
```
SherlockNoMad: this is inefficient if we need to call it multiple times. 
better build the node to index once, and read from cache.

torchtitan/experiments/graph_trainer/ep_chunk_pass.py
```diff
@@ -0,0 +1,982 @@
...
+
+    gm.graph.lint()
+    gm.recompile()
+    logger.info(
+        "Applied chunk_%s to %d regions (%d materialized live-outs): %s",
+        mode,
+        len(regions),
+        transformed,
+        module_patterns,
+    )
```
SherlockNoMad, line 2420: I would add 10x more logger.debug(...) to make this pass more maintainable.

sanketpurandare:
Added a Manifold-backed tlparse artifact here:

https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/a00c7cc7-93ab-41cb-a09d-c7de1e0f0d97/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

The useful entries to compare are `before_chunk_batch_pass` and `after_chunk_batch_pass`; this is a minimal traced graph that shows the pass inserting split, duplicated chunk compute, and cat materialization.

torchtitan/experiments/graph_trainer/ep_chunk_pass.py
sanketpurandare: Added targeted `logger.debug(...)` coverage around placeholder metadata import, matched regions, per-region plan sizes, and live-out materialization counts. I also cleaned up the pass structure and comments so the logs map to the main pass phases without dumping whole graphs.

torchtitan/models/deepseek_v3/model.py
sanketpurandare: Rechecked with the current PyTorch fixes and reverted this back to `bsz, seqlen, _ = x.size()`. The focused chunk tests, including the DeepSeekV3 bitwise chunking subtests, still pass.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
sanketpurandare: Moved this out of the generic tracer. The chunk-specific placeholder metadata import now lives in `import_chunk_dim_metadata_pass`, which is inserted immediately before the selected chunk pass in `compile_time_passes`.

torchtitan/experiments/graph_trainer/passes.py
sanketpurandare: Simplified this by replacing the two independently enabled lists with one `chunk_modules` list plus one `chunk_mode` selector. Pass construction now dispatches to either batch or seq chunking from that single mode, and it validates that selected chunk roots are compatible with downstream transformer-block scheduling boundaries.

torchtitan/experiments/graph_trainer/common_utils.py
sanketpurandare: Fixed. I removed the repeated `_node_index` / `_is_static_placeholder` scan from `common_utils.py`; `chunk_passes.py` now computes the static placeholder set once and reuses it while deriving static nodes and chunkable live-ins.

sanketpurandare:
Added two DeepSeekV3 debug-model tlparse artifacts as well:

Transformer block batch chunking (`layers.*`):
https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/2ccf18d7-c6e7-417f-bc45-55bb2517d7a5/dsv3_transformer_chunk_batch/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

MoE block batch chunking (`layers.*.moe`):
https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/1fc9547b-f670-42ac-bb9b-50fcb930df78/dsv3_moe_chunk_batch/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

I checked the local trace logs before posting these. Both contain `before_chunk_batch_pass` / `after_chunk_batch_pass`. The transformer trace chunks `layers.0` through `layers.5`; the MoE trace chunks `layers.1.moe` through `layers.5.moe` (`layers.0` has `feed_forward`, not `moe`). The after graphs show the expected dynamic batch split with `floor(u0/2)`, duplicated `_chunk0` / `_chunk1` compute, `_chunk_cat` full-shape materialization, and `_chunk_tuple` provenance.

SherlockNoMad:
> Added two DeepSeekV3 debug-model tlparse artifacts as well:
> 
> Transformer block batch chunking (`layers.*`): https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/2ccf18d7-c6e7-417f-bc45-55bb2517d7a5/dsv3_transformer_chunk_batch/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000
> 
> MoE block batch chunking (`layers.*.moe`): https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/1fc9547b-f670-42ac-bb9b-50fcb930df78/dsv3_moe_chunk_batch/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000
> 
> I checked the local trace logs before posting these. Both contain `before_chunk_batch_pass` / `after_chunk_batch_pass`. The transformer trace chunks `layers.0` through `layers.5`; the MoE trace chunks `layers.1.moe` through `layers.5.moe` (`layers.0` has `feed_forward`, not `moe`). The after graphs show the expected dynamic batch split with `floor(u0/2)`, duplicated `_chunk0` / `_chunk1` compute, `_chunk_cat` full-shape materialization, and `_chunk_tuple` provenance.

FYI, you can add `--compile.debug_graph_passes` to see before vs. after for all passes. 
by diffing the before and after, I got https://www.internalfb.com/intern/diffing/?paste_number=2321706137

I see the chunking (every node is duplicated twice) is happening as expected, but didn't see the overlap going?

sanketpurandare:
That was correct for this PR: #3325 only inserted the chunking transform, so the graph diff showed duplication/materialization but no EP launch/wait reordering.

The overlap scheduling is implemented in the stacked EP-overlap PR (#3328). The public `ep_overlap` pipeline is split into two phases: chunking runs after SAC/remat, and EP scheduling runs after bucketing so it can reorder the already chunked MoE dispatch/combine regions. I posted the EP-overlap tlparse artifacts on #3328 showing the dispatch/combine launch and delayed wait ordering.

EP-overlap artifacts:
- Transformer-block batch overlap: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/f43be333-f0f6-4028-91ba-6a58e22de8f8/ep-overlap-transformer-block-batch-skip-dense-current/rank_0/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000
- MoE-only batch overlap: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/75fa8de5-a9d6-4170-9279-7d441d73c235/ep-overlap-moe-batch-current/rank_0/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

For the overlap artifact, the useful entries are `before_ep_overlap_schedule_pass` / `after_ep_overlap_schedule_pass` and the generated rank code under `custom_codegen_pass`. The scheduling evidence to look for is chunked dispatch/combine `all_to_all_single` launches followed by delayed wait-suffix nodes annotated with `EP_wait=True`.

I verified the generated scheduled rank-0 code for `layers.1.moe.experts`: dispatch launches appear as `all_to_all_single_1_chunk0` then `all_to_all_single_1_chunk1`, dispatch wait/use starts after both launches, combine launches appear as `all_to_all_single_2_chunk0` then `all_to_all_single_2_chunk1`, and combine waits appear after both combine launches. The delayed wait suffix nodes are annotated with `EP_wait=True`.

torchtitan/experiments/graph_trainer/remove_noop_passes.py
```diff
@@ -177,3 +177,14 @@ def remove_identity_slice_pass(
     gm.graph.lint()
     gm.recompile()
     return gm
+
+
+def remove_dead_code_pass(
```
SherlockNoMad: already in main.

torchtitan/experiments/graph_trainer/tests/test_numerics.py
```diff
@@ -343,6 +414,27 @@ def test_moe_dsv3_aot_fx_trace_vs_eager(self):
...
+    @unittest.skip("upstream DTensor mixed-dtype sharding propagation regression")
+    def test_moe_dsv3_ep_overlap_moe_seq_aot_fx_trace_vs_eager_chunked(self):
+        self.assertTrue(_run_deepseek_v3_ep_overlap_moe_seq_loss_compare())
+
+    # TODO: Disabled due to upstream PyTorch nightly DTensor regression.
+    # Sharding propagation fails for aten.mm.default with mixed dtypes
+    # (bf16 activations, f32 weights) on the TP mesh. Re-enable once fixed.
+    @unittest.skip("upstream DTensor mixed-dtype sharding propagation regression")
+    def test_moe_dsv3_ep_overlap_moe_batch_aot_fx_trace_vs_eager_chunked(self):
+        self.assertTrue(_run_deepseek_v3_ep_overlap_moe_batch_loss_compare())
```
SherlockNoMad: are these addressed already?

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -98,6 +98,42 @@ def annotate_moe_ep_regions() -> None:
...
+# ---------------------------------------------------------------------------
+# These helpers are shared by chunk passes and tests around graph_trainer's
+# traced FX metadata. Keep chunk-specific rewrite policy in ep_chunk_pass.py.
+
+
+def _is_module_fqn_inside_root(fqn: str, root_fqn: str) -> bool:
+    return fqn == root_fqn or fqn.startswith(root_fqn + ".")
+
+
+def _tensor_meta(node: torch.fx.Node) -> torch.Tensor | None:
```
SherlockNoMad: val can be list/tuple/sym_int...

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -98,6 +98,42 @@ def annotate_moe_ep_regions() -> None:
...
+    val = node.meta.get("val")
+    return val if isinstance(val, torch.Tensor) else None
+
+
+def _free_symbols(value: object) -> frozenset[object]:
```
SherlockNoMad: docstring, whats' free symbols? u1, u2... ?

torchtitan/experiments/graph_trainer/common_utils.py
```diff
@@ -98,6 +98,42 @@ def annotate_moe_ep_regions() -> None:
...
+    from torch.fx.experimental.symbolic_shapes import free_symbols
+
+    return frozenset(free_symbols(value))
+
+
+def _dynamic_dim_symbols(val: torch.Tensor, dim: int) -> frozenset[object]:
```
SherlockNoMad: doc string

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -155,18 +166,85 @@ def compile_time_passes(
...
 
     n_layers = len(config.model_spec.model.layers)
+    module_bucket_plans = get_default_transformer_block_buckets(n_layers)
     passes: list[Callable] = [
         remove_detach_pass,
         remove_identity_view_pass,
         remove_identity_slice_pass,
         normalize_view_ops_as_reshape,
     ]
     if "ep_overlap" in config.compile.passes:
```
SherlockNoMad: hmm... I originally intend to deprecate config.compile.passes.. 
but this is a new way to use it. 

I would rename config.compile.passes to config.compile.extra_passes 

nit, fine as follow up

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
```diff
@@ -143,6 +146,10 @@ def _run_steps(
         trainer_cls: type,
         *,
         enable_passes: bool = True,
+        compile_passes: list[str] | None = None,
+        compile_ep_overlap_chunk_dim: str = "batch",
+        compile_ep_overlap_module_fqn: str = "layers.*",
+        compile_ep_overlap_disable_early_grad_accumulation: bool = False,
```
SherlockNoMad, line 177: this is probably a sign for ep_overlap to have its own Config. 

compile.ep_overlap.*

SherlockNoMad:
Need some case in integration_tests.py

torchtitan/experiments/graph_trainer/ep_chunk_pass.py
```diff
@@ -0,0 +1,2101 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Graph chunking for EP overlap.
```
SherlockNoMad, line 7: This pass is very complicated... and is only maintainable by you / claude. 

I am fine with landing this, coz it's a standalone graph pass, and it's complexity is well encapsulated, and if thing doesn't work, ppl can and will turn it off. 

but my point is, I have seen your profiler trace, and I trust this would work. 

so I am approving this as a blackbox owned by you.

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/passes.py
```diff
@@ -136,16 +150,84 @@ def compile_time_passes(
...
+            # router/top-k/groupedMM/shared-expert/dense compute; and validate
+            # that the final chunk body contains no non-EP collectives.
+            #
+            # This also avoids redundant inner MoE TP collectives: router can
+            # slice the replicated boundary input to its owned tokens, while
+            # shared experts can reuse that replicated input instead of issuing
+            # their own input all-gathers and leave the final reduction to the
+            # MoE boundary epilogue.
+            #
+            # Batch graph chunking with TP is temporarily allowed so we can
```
IvanKobzarev, line 197: Looks like stale comment about support of TP

torchtitan/experiments/graph_trainer/tests/test_numerics.py
sanketpurandare: removed them

torchtitan/experiments/graph_trainer/passes.py
sanketpurandare: no there are upstream problems on how TP uses some global indexing when composed with Ep, need to fix that first

IvanKobzarev (approved):

torchtitan/experiments/graph_trainer/remove_noop_passes.py
sanketpurandare: removed

sanketpurandare:
> Need some case in integration_tests.py

@SherlockNoMad Current stack does have EP-overlap integration coverage, but it lives in the scheduling commit: six      
DeepSeekV3 FlexAttention EP-overlap integration variants for regional/full x transformer_batch/moe_batch/
moe_seq. We also have 3 distributed bitwise equivalence tests with DeepSeekV3 sdpa regional transformer_batch/moe_batch/moe_seq

torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py
sanketpurandare: We now have EpOverlapConfig under GraphTrainerCompileConfig.ep_overlap.

torchtitan/experiments/graph_trainer/common_utils.py
sanketpurandare: fixed

torchtitan/experiments/graph_trainer/common_utils.py
sanketpurandare: This is stale

torchtitan/experiments/graph_trainer/common_utils.py
sanketpurandare: addresed

PR #3328, opened 2026-05-12
[graph_trainer] Add EP overlap scheduling pass

### LINKED ISSUE #3434: RFC: Graph-Based EP Overlap For GraphTrainer

sanketpurandare:
Stacked PRs:
 * __->__#3328
 * #3325
 * #3363

--- --- ---

### [graph_trainer] Add EP overlap scheduling pass

Add the EP overlap scheduler, process-group isolation, FSDP dense-region scheduling, and the scheduling metadata needed to pair EP token exchanges with chunked compute regions.

Split MoE token-dispatcher collectives into annotated helper methods so the graph scheduler can distinguish token-count sync, dispatch, and combine boundaries while preserving the existing non-blocking input split copy behavior.

Test Plan:
- conda run -n pt-dev python -m pytest -q torchtitan/experiments/graph_trainer/tests/test_passes.py::TestFsdpDenseSchedulerPass torchtitan/experiments/graph_trainer/tests/test_passes.py::TestChunkPasses torchtitan/experiments/graph_trainer/tests/test_precompile.py::TestConfigFingerprint torchtitan/experiments/graph_trainer/tests/test_passes.py::TestEagerChunking
- conda run -n pt-dev python -m pytest -q torchtitan/experiments/graph_trainer/tests/test_chunked_loss.py torchtitan/experiments/graph_trainer/tests/test_precompile.py::TestPrecompileLossSetup torchtitan/experiments/graph_trainer/tests/test_trace_module.py::TestTraceModule::test_chunked_ce_loss_train_step tests/unit_tests/test_loss.py::TestGradAccumulator tests/unit_tests/test_loss.py::TestGradAccumulatorDTensor tests/unit_tests/test_loss.py::TestChunkedCELoss tests/unit_tests/test_loss.py::TestChunkedCELossSPMD
- conda run -n pt-dev python -m pytest -q torchtitan/experiments/graph_trainer/tests/test_bitwise_deterministic.py::TestDSv3FlexAttnBitwiseDeterministic::test_ep_chunk_matches_eager_chunking_bitwise
- conda run -n pt-dev python -m pytest -q torchtitan/experiments/graph_trainer/tests/test_numerics.py::TestGraphTrainerNumerics::test_moe_dsv3_ep_overlap_aot_fx_trace_vs_eager_chunked torchtitan/experiments/graph_trainer/tests/test_numerics.py::TestGraphTrainerNumerics::test_moe_dsv3_ep_overlap_moe_seq_aot_fx_trace_vs_eager_chunked torchtitan/experiments/graph_trainer/tests/test_numerics.py::TestGraphTrainerNumerics::test_moe_dsv3_ep_overlap_moe_batch_aot_fx_trace_vs_eager_chunked
- python torchtitan/experiments/graph_trainer/tests/integration_tests.py <out> --test_suite graph_trainer_h100 --test_name <each ep_overlap entry> --ngpu 8
- SKIP=pyrefly-check conda run -n pt-dev pre-commit run --files <stack files>
- conda run -n pt-dev pyrefly check --project-excludes=agent_space --project-excludes=torchtitan/experiments --project-excludes='**/tests/**' --output-format omit-errors

### DIFF
--- a/torchtitan/models/common/token_dispatcher.py
+++ b/torchtitan/models/common/token_dispatcher.py
@@ -254,6 +254,130 @@ class AllToAllTokenDispatcher(BaseEPTokenDispatcher):
     def __init__(self, config: Config):
         super().__init__(config)
 
+    def _token_count_exchange(
+        self,
+        num_local_tokens_per_expert_E: torch.Tensor,
+        pg,
+        ep_size: int,
+    ) -> torch.Tensor:
+        """Exchange per-rank expert token counts before the data all-to-all.
+
+        This method is separate from ``dispatch`` so graph passes can annotate
+        the count exchange independently from the true token-exchange
+        scheduling markers.
+        """
+        if (
+            torch.compiler.is_compiling() or torch.compiler._is_non_strict_tracing()
+        ) and get_spmd_backend() != "spmd_types":
+            return all_to_all_single(
+                num_local_tokens_per_expert_E.view(ep_size, -1),
+                None,
+                None,
+                group=self.ep_mesh,
+            )
+
+        return spmd.all_to_all(
+            num_local_tokens_per_expert_E.view(ep_size, -1),
+            pg,
+            src=spmd.V,
+            dst=spmd.V,
+        )
+
+    def _sync_token_count_exchange(
+        self,
+        num_local_tokens_per_expert_E: torch.Tensor,
+        num_global_tokens_per_local_expert_EP_e: torch.Tensor,
+        ep_size: int,
+    ) -> tuple[torch.Tensor, list[int], list[int]]:
+        """Wait for token counts and materialize CPU split lists.
+
+        Local input splits can copy to CPU non-blocking; remote output splits
+        must be ready before launching the variable-size data all-to-all.
+        """
+        # Need to wait explicitly because it is used by a triton kernel later
+        # which doesn't realize that AsyncCollectiveTensor needs unwrapping
+        num_global_tokens_per_local_expert_EP_e = (
+            torch.ops._c10d_functional.wait_tensor(
+                num_global_tokens_per_local_expert_EP_e
+            )
+        )
+        num_global_tokens_per_local_expert_E = (
+            num_global_tokens_per_local_expert_EP_e.reshape(-1)
+        )
+        input_splits = (
+            num_local_tokens_per_expert_E.view(ep_size, -1)
+            .sum(dim=1)
+            .to(torch.device("cpu"), non_blocking=True)
+        )
+        # NOTE: this would incur a device-to-host sync
+        output_splits = (
+            num_global_tokens_per_local_expert_E.view(ep_size, -1)
+            .sum(dim=1)
+            .to(torch.device("cpu"), non_blocking=False)
+        )
+        input_splits_list = input_splits.tolist()
+        output_splits_list = output_splits.tolist()
+
+        return (
+            num_global_tokens_per_local_expert_E,
+            input_splits_list,
+            output_splits_list,
+        )
+
+    def _dispatch_token_exchange(
+        self,
+        routed_input_ND: torch.Tensor,
+        pg,
+        output_splits: list[int],
+        input_splits: list[int],
+    ) -> torch.Tensor:
+        """Launch the dispatch all-to-all that moves routed tokens to experts."""
+        if (
+            torch.compiler.is_compiling() or torch.compiler._is_non_strict_tracing()
+        ) and get_spmd_backend() != "spmd_types":
+            return all_to_all_single(
+                routed_input_ND,
+                output_splits,
+                input_splits,
+                self.ep_mesh,
+            )
+
+        return spmd.all_to_all(
+            routed_input_ND,
+            pg,
+            src=spmd.V,
+            dst=spmd.V,
+            output_split_sizes=output_splits,
+            input_split_sizes=input_splits,
+        )
+
+    def _combine_token_exchange(
+        self,
+        routed_output_RD: torch.Tensor,
+        pg,
+        input_splits: list[int],
+        output_splits: list[int],
+    ) -> torch.Tensor:
+        """Launch the combine all-to-all that returns expert outputs to tokens."""
+        if (
+            torch.compiler.is_compiling() or torch.compiler._is_non_strict_tracing()
+        ) and get_spmd_backend() != "spmd_types":
+            return all_to_all_single(
+                routed_output_RD,
+                input_splits,
+                output_splits,
+                self.ep_mesh,
+            )
+
+        return spmd.all_to_all(
+            routed_output_RD,
+            pg,
+            src=spmd.V,
+            dst=spmd.V,
+            output_split_sizes=input_splits,
+            input_split_sizes=output_splits,
+        )
+
     # pyrefly: ignore [bad-override]
     def dispatch(
         self,
@@ -330,60 +454,27 @@ class AllToAllTokenDispatcher(BaseEPTokenDispatcher):
                 )
 
             with torch.no_grad():
-                if torch.compiler.is_compiling() and get_spmd_backend() != "spmd_types":
-                    num_global_tokens_per_local_expert_EP_e = all_to_all_single(
-                        num_local_tokens_per_expert_E.view(ep_size, -1),
-                        None,
-                        None,
-                        group=self.ep_mesh,
-                    )
-                else:
-                    num_global_tokens_per_local_expert_EP_e = spmd.all_to_all(
-                        num_local_tokens_per_expert_E.view(ep_size, -1),
-                        pg,
-                        src=spmd.V,
-                        dst=spmd.V,
-                    )
-                # Need to wait explicitly because it is used by a triton kernel later
-                # which doesn't realize that AsyncCollectiveTensor needs unwrapping
-                num_global_tokens_per_local_expert_EP_e = (
-                    torch.ops._c10d_functional.wait_tensor(
-                        num_global_tokens_per_local_expert_EP_e
-                    )
-                )
-                num_global_tokens_per_local_expert_E = (
-                    num_global_tokens_per_local_expert_EP_e.reshape(-1)
-                )
-                input_splits = (
-                    num_local_tokens_per_expert_E.view(ep_size, -1)
-                    .sum(dim=1)
-                    .to(torch.device("cpu"), non_blocking=True)
-                )
-                # NOTE: this would incur a device-to-host sync
-                output_splits = (
-                    num_global_tokens_per_local_expert_E.view(ep_size, -1)
-                    .sum(dim=1)
-                    .to(torch.device("cpu"), non_blocking=False)
-                )
-                input_splits_list = input_splits.tolist()
-                output_splits_list = output_splits.tolist()
-
-            if torch.compiler.is_compiling() and get_spmd_backend() != "spmd_types":
-                routed_input_RD = all_to_all_single(
-                    routed_input_ND,
-                    output_splits_list,
-                    input_splits_list,
-                    self.ep_mesh,
-                )
-            else:
-                routed_input_RD = spmd.all_to_all(
-                    routed_input_ND,
+                num_global_tokens_per_local_expert_EP_e = self._token_count_exchange(
+                    num_local_tokens_per_expert_E,
                     pg,
-                    src=spmd.V,
-                    dst=spmd.V,
-                    output_split_sizes=output_splits_list,
-                    input_split_sizes=input_splits_list,
+                    ep_size,
                 )
+                (
+                    num_global_tokens_per_local_expert_E,
+                    input_splits_list,
+                    output_splits_list,
+                ) = self._sync_token_count_exchange(
+                    num_local_tokens_per_expert_E,
+                    num_global_tokens_per_local_expert_EP_e,
+                    ep_size,
+                )
+
+            routed_input_RD = self._dispatch_token_exchange(
+                routed_input_ND,
+                pg,
+                output_splits_list,
+                input_splits_list,
+            )
             # Reorder from rank-major to expert-major via _permute.
             #
             # num_global_tokens_per_local_expert_E layout after all-to-all
@@ -530,22 +621,12 @@ class AllToAllTokenDispatcher(BaseEPTokenDispatcher):
             )
             # All-to-all combine: returns AsyncCollectiveTensor — the a2a runs
             # on the NCCL stream and won't block until the tensor is accessed.
-            if torch.compiler.is_compiling() and get_spmd_backend() != "spmd_types":
-                routed_output_RD = all_to_all_single(
-                    routed_output_RD,
-                    metadata.input_splits,
-                    metadata.output_splits,
-                    self.ep_mesh,
-                )
-            else:
-                routed_output_RD = spmd.all_to_all(
-                    routed_output_RD,
-                    pg,
-                    src=spmd.V,
-                    dst=spmd.V,
-                    output_split_sizes=metadata.input_splits,
-                    input_split_sizes=metadata.output_splits,
-                )
+            routed_output_RD = self._combine_token_exchange(
+                routed_output_RD,
+                pg,
+                metadata.input_splits,
+                metadata.output_splits,
+            )
 
         if get_spmd_backend() == "spmd_types":
             if spmd.is_type_checking():  # dense mesh reinterpret
--- a/torchtitan/experiments/graph_trainer/common_utils.py
+++ b/torchtitan/experiments/graph_trainer/common_utils.py
@@ -65,8 +65,35 @@ def build_decoder_config_for_backend(
     return config
 
 
+def _local_stride(tensor: torch.Tensor) -> tuple[int, ...]:
+    return (
+        tensor.to_local().stride() if isinstance(tensor, DTensor) else tensor.stride()
+    )
+
+
+def _maybe_materialize_grad_for_param_layout(
+    param: torch.Tensor, grad: torch.Tensor
+) -> torch.Tensor:
+    """Match eager autograd's ``param.grad`` layout contract after graph replay.
+
+    Graph replay assigns ``torch.autograd.grad`` outputs manually, bypassing
+    AccumulateGrad's normal stride materialization. Copying through
+    ``empty_like(param)`` restores the param's global and DTensor-local layout
+    when Inductor returns an equivalent but differently-strided grad.
+    """
+    if grad.stride() == param.stride() and _local_stride(grad) == _local_stride(param):
+        return grad
+
+    materialized_grad = torch.empty_like(param)
+    materialized_grad.copy_(grad)
+    return materialized_grad
+
+
 _MODULE_FQN = "module_fqn"
+_EP_TOKEN_COUNT_EXCHANGE = "EP_token_count_exchange"
+_EP_TOKEN_COUNT_SYNC = "EP_token_count_sync"
 _EP_TOKEN_EXCHANGE = "EP_token_exchange"
+_EP_TOKEN_EXCHANGE_WAIT = "EP_token_exchange_wait"
 _NOT_IN_LAYERS = -1
 
 
@@ -145,6 +172,18 @@ def annotate_moe_ep_regions() -> None:
     AllToAllTokenDispatcher.combine = annotate_fn({"EP": "combine"})(
         AllToAllTokenDispatcher.combine
     )
+    AllToAllTokenDispatcher._token_count_exchange = annotate_fn(
+        {_EP_TOKEN_COUNT_EXCHANGE: "dispatch"}
+    )(AllToAllTokenDispatcher._token_count_exchange)
+    AllToAllTokenDispatcher._sync_token_count_exchange = annotate_fn(
+        {_EP_TOKEN_COUNT_SYNC: "dispatch"}
+    )(AllToAllTokenDispatcher._sync_token_count_exchange)
+    AllToAllTokenDispatcher._dispatch_token_exchange = annotate_fn(
+        {_EP_TOKEN_EXCHANGE: "dispatch"}
+    )(AllToAllTokenDispatcher._dispatch_token_exchange)
+    AllToAllTokenDispatcher._combine_token_exchange = annotate_fn(
+        {_EP_TOKEN_EXCHANGE: "combine"}
+    )(AllToAllTokenDispatcher._combine_token_exchange)
     MoE.forward = annotate_fn({"EP": "compute"})(MoE.forward)
     _MOE_EP_REGIONS_ANNOTATED = True
 
@@ -207,12 +246,31 @@ def get_default_transformer_block_buckets(
     n_layers: int,
     *,
     chunked_loss_enabled: bool = False,
+    moe_layer_ids: frozenset[int] = frozenset(),
+    split_moe_expert_buckets: bool = False,
 ) -> list[list[str] | str]:
     """Get default transformer block buckets for manual bucketing passes.
 
     Assumes the standard Decoder layout: tok_embeddings, layers.0..N-1,
     norm, and output (e.g., Llama3, DeepSeekV3, Qwen3).
     """
+    layer_buckets: list[list[str] | str] = []
+    for layer_id in range(n_layers):
+        if layer_id in moe_layer_ids and split_moe_expert_buckets:
+            layer_buckets.extend(
+                [
+                    [
+                        f"layers.{layer_id}.attention_norm",
+                        f"layers.{layer_id}.attention",
+                        f"layers.{layer_id}.ffn_norm",
+                        f"layers.{layer_id}.moe.router",
+                        f"layers.{layer_id}.moe.shared_experts",
+                    ],
+                    f"layers.{layer_id}.moe.experts",
+                ]
+            )
+        else:
+            layer_buckets.append(f"layers.{layer_id}")
     final_bucket = ["norm", "lm_head"]
     if chunked_loss_enabled:
         # Chunked loss moves the lm_head weight use under module_fqn "loss".
@@ -220,7 +278,7 @@ def get_default_transformer_block_buckets(
 
     return [
         "tok_embeddings",
-        *[f"layers.{i}" for i in range(n_layers)],
+        *layer_buckets,
         final_bucket,
     ]
 
--- a/torchtitan/experiments/graph_trainer/configs.py
+++ b/torchtitan/experiments/graph_trainer/configs.py
@@ -139,6 +139,17 @@ class GraphTrainerCompileConfig(CompileConfig):
     two collectives can overlap. It is a no-op when the graph contains no
     FSDP all-gathers."""
 
+    enable_fsdp_dense_region_overlap: bool = False
+    """When True, schedule FSDP AG/RS buckets into neighboring dense regions.
+
+    This is disabled by default because it changes FSDP collective placement
+    and is intended for performance/integration validation, not
+    bitwise-equivalence tests. When EP overlap is also enabled, this scheduler
+    only composes with graph chunking rooted at ``layers.*.moe``; otherwise the
+    explicit request is skipped with a warning. Without EP overlap, it can run
+    as a standalone FSDP scheduling ablation.
+    """
+
     ep_overlap: EpOverlapConfig = field(default_factory=EpOverlapConfig)
     """Configuration for EP-overlap chunking and scheduling."""
 
--- a/torchtitan/experiments/graph_trainer/ep_overlap_pass.py
+++ b/torchtitan/experiments/graph_trainer/ep_overlap_pass.py
@@ -1,0 +1,998 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Schedule already chunked EP token-exchange regions.
+
+Contract
+========
+This pass is intentionally a scheduler only.  It consumes a graph that has
+already been chunked by either eager chunking or ``ep_chunk_pass`` and must not
+change tensor values, live-in/live-out materialization, or provenance.  The only
+semantic input it relies on is chunk-body metadata collected by
+``collect_chunked_regions``.
+
+For each selected forward/backward region:
+
+* exactly two chunk bodies, chunk 0 and chunk 1, must be present;
+* true EP scheduling markers are
+  ``_c10d_functional.all_to_all_single.default`` launches inside the selected
+  chunk body;
+* ``custom[_EP_TOKEN_EXCHANGE]`` marks true token-exchange all-to-all launches;
+  waits may inherit the annotation from traceback and are normalized to
+  ``custom[_EP_TOKEN_EXCHANGE_WAIT]``;
+* marker counts and labels must match across chunks;
+* forward emits marker pairs in chunk order 0 then 1, backward emits 1 then 0;
+* MoE-root chunking pairs both chunks' first-marker setup before launching the
+  first marker pair; wider transformer-root chunking keeps the regular
+  wait-gated per-chunk closure schedule for every marker;
+* token-count sync CPU copies, when present in the first marker closure, are
+  launched before their CPU scalar/list consumers;
+* after each marker pair, ready non-collective body work is emitted as filler
+  before advancing to the next marker pair;
+* all graph nodes remain in the sorted graph exactly once and the final graph
+  must lint.
+
+The same contract covers eager and graph chunking.  If a chunked region violates
+the contract, the pass errors rather than producing a silent schedule change.
+
+Pseudo-code
+===========
+1. Collect chunked regions and build node -> chunk-owner lookup from shared EP
+   pass metadata.
+2. For each region, collect true token-exchange all-to-all markers per chunk
+   and normalize wait annotations inherited from traceback.
+3. Validate both chunks have the same marker signature, then build dependency
+   closures needed to launch each marker.
+4. Emit wait-gated phases: marker pair in chunk order, ready filler work that
+   does not need token-exchange waits, then final tail work with waits allowed.
+5. Apply the requested region phases through a stable topological sort, lint,
+   recompile, and validate that phase order materialized.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+import torch.fx as fx
+from torch._dynamo.graph_deduplication import _stable_topological_sort
+from torch.utils._ordered_set import OrderedSet
+
+from torchtitan.experiments.graph_trainer.common_utils import (
+    _EP_TOKEN_COUNT_SYNC,
+    _EP_TOKEN_EXCHANGE,
+    _EP_TOKEN_EXCHANGE_WAIT,
+)
+from torchtitan.experiments.graph_trainer.ep_pass_utils import (
+    ChunkBody,
+    ChunkedRegion,
+    ChunkOwner,
+    collect_chunked_regions,
+    is_c10d_functional_node,
+    ordered_nodes,
+)
+from torchtitan.tools.logging import logger
+
+
+_GRAPH_BOUNDARY_OPS = {"placeholder", "get_attr"}
+_EP_PHASES = {"dispatch", "combine"}
+
+
+# Step 0: Small metadata helpers and local scheduling records.
+
+
+@dataclass(frozen=True)
+class _TokenExchange:
+    label: str
+    launch: fx.Node
+
+
+@dataclass(frozen=True)
+class _ScheduledRegion:
+    region: ChunkedRegion
+    phases: tuple[tuple[fx.Node, ...], ...]
+
+
+@dataclass(frozen=True)
+class _SyncClosureParts:
+    pre_copy: tuple[fx.Node, ...]
+    copies: tuple[fx.Node, ...]
+    post_copy: tuple[fx.Node, ...]
+    launches: tuple[fx.Node, ...]
+
+
+def _custom_meta(node: fx.Node) -> dict[str, Any]:
+    """Return mutable custom metadata when present, otherwise an empty dict."""
+    custom = node.meta.get("custom")
+    return custom if isinstance(custom, dict) else {}
+
+
+def _ep_label(node: fx.Node) -> str:
+    """Return the optional EP phase label for logs/wait metadata."""
+    phase = _custom_meta(node).get(_EP_TOKEN_EXCHANGE)
+    return phase if phase in _EP_PHASES else "all_to_all"
+
+
+def _is_token_exchange_launch(node: fx.Node) -> bool:
+    """Return whether a node is a token-exchange scheduling marker.
+
+    Only all-to-all ops annotated with ``EP_token_exchange`` (dispatch or
+    combine) are scheduling markers. The counts all-to-all (which only has
+    ``EP: dispatch`` without ``EP_token_exchange``) is NOT a marker; it
+    stays in the closure with shared expert + router compute so that
+    compute is available as filler during the data dispatch window.
+    """
+    return (
+        node.op == "call_function"
+        and node.target == torch.ops._c10d_functional.all_to_all_single.default
+        and _custom_meta(node).get(_EP_TOKEN_EXCHANGE) in _EP_PHASES
+    )
+
+
+def _is_c10d_functional_node(node: fx.Node) -> bool:
+    """Return whether a node is a distributed functional op."""
+    return is_c10d_functional_node(node)
+
+
+def _is_token_count_sync_copy(node: fx.Node) -> bool:
+    return (
+        node.op == "call_function"
+        and node.target == torch.ops.aten._to_copy.default
+        and _custom_meta(node).get(_EP_TOKEN_COUNT_SYNC) == "dispatch"
+    )
+
+
+def _is_cpu_destination_copy(node: fx.Node) -> bool:
+    device = node.kwargs.get("device")
+    if device is not None:
+        return torch.device(device).type == "cpu"
+
+    val = node.meta.get("val")
+    val_device = getattr(val, "device", None)
+    return val_device is not None and val_device.type == "cpu"
+
+
+def _set_copy_non_blocking(node: fx.Node, non_blocking: bool) -> None:
+    kwargs = dict(node.kwargs)
+    kwargs["non_blocking"] = non_blocking
+    node.kwargs = kwargs
+
+
+def _same_region_owner(
+    node: fx.Node,
+    *,
+    owner_by_node: dict[fx.Node, ChunkOwner],
+    root_fqn: str,
+    is_backward: bool,
+) -> ChunkOwner | None:
+    """Return chunk ownership only when it belongs to the same selected region."""
+    owner = owner_by_node.get(node)
+    if (
+        owner is not None
+        and owner.root_fqn == root_fqn
+        and owner.is_backward == is_backward
+    ):
+        return owner
+    return None
+
+
+def _is_wait_for_token_exchange(node: fx.Node, node_set: set[fx.Node]) -> bool:
+    return (
+        node.op == "call_function"
+        and node.target == torch.ops._c10d_functional.wait_tensor.default
+        and node.all_input_nodes
+        and node.all_input_nodes[0] in node_set
+        and _is_token_exchange_launch(node.all_input_nodes[0])
+    )
+
+
+def _collect_token_exchanges(
+    body: ChunkBody,
+    *,
+    order: dict[fx.Node, int],
+) -> tuple[_TokenExchange, ...]:
+    """Step 2: collect true token-exchange launches for one chunk body."""
+    node_set = set(body.nodes)
+    exchanges: list[_TokenExchange] = []
+    for node in body.nodes:
+        if not _is_token_exchange_launch(node):
+            phase = _custom_meta(node).get(_EP_TOKEN_EXCHANGE)
+            if phase is None:
+                continue
+            if _is_wait_for_token_exchange(node, node_set):
+                custom = dict(_custom_meta(node))
+                custom.pop(_EP_TOKEN_EXCHANGE, None)
+                custom[_EP_TOKEN_EXCHANGE_WAIT] = phase
+                node.meta["custom"] = custom
+                continue
+            raise ValueError(
+                "ep_overlap found EP token-exchange metadata on non-marker "
+                f"node {node.name} ({node.target}). Only all_to_all launches "
+                "and their wait_tensor nodes may carry this annotation."
+            )
+            continue
+
+        label = _ep_label(node)
+        waits = [
+            user
+            for user in node.users
+            if user in node_set
+            and user.op == "call_function"
+            and user.target == torch.ops._c10d_functional.wait_tensor.default
+        ]
+        if len(waits) != 1:
+            raise ValueError(
+                f"ep_overlap expected one token-exchange wait for {node.name}, "
+                f"found {len(waits)}."
+            )
+        wait = waits[0]
+        if order[wait] <= order[node]:
+            raise ValueError(
+                "ep_overlap expected token-exchange wait to appear after its "
+                f"launch, found launch={node.name} wait={wait.name} for "
+                f"{body.owner}."
+            )
+        custom = dict(_custom_meta(wait))
+        custom.pop(_EP_TOKEN_EXCHANGE, None)
+        custom[_EP_TOKEN_EXCHANGE_WAIT] = label
+        wait.meta["custom"] = custom
+        exchanges.append(_TokenExchange(label=label, launch=node))
+    return tuple(exchanges)
+
+
+def _exchange_signature(exchanges: tuple[_TokenExchange, ...]) -> tuple[str, ...]:
+    """Return the semantic marker sequence used to match chunk pairs."""
+    return ("all_to_all",) * len(exchanges)
+
+
+def _exchange_labels(exchanges: tuple[_TokenExchange, ...]) -> tuple[str, ...]:
+    """Return optional marker labels for diagnostics."""
+    return tuple(exchange.label for exchange in exchanges)
+
+
+def _validate_exchange_labels(
+    root: str,
+    direction: str,
+    exchanges_by_chunk: dict[int, tuple[_TokenExchange, ...]],
+) -> None:
+    labels0 = _exchange_labels(exchanges_by_chunk[0])
+    labels1 = _exchange_labels(exchanges_by_chunk[1])
+    if labels0 != labels1:
+        raise ValueError(
+            f"ep_overlap expected matching token-exchange labels for "
+            f"{root!r} ({direction}), found chunk0={labels0} chunk1={labels1}."
+        )
+
+
+# Step 3: Build marker dependency closures and identify ready filler work.
+
+
+def _hidden_body_deps(
+    node: fx.Node,
+    *,
+    owner_by_node: dict[fx.Node, ChunkOwner],
+    root_fqn: str,
+    is_backward: bool,
+) -> tuple[fx.Node, ...]:
+    """Find same-region body deps behind unowned graph plumbing."""
+    deps: list[fx.Node] = []
+    seen: set[fx.Node] = set()
+    stack = list(node.all_input_nodes)
+    while stack:
+        dep = stack.pop()
+        if dep in seen:
+            continue
+        seen.add(dep)
+        owner = _same_region_owner(
+            dep,
+            owner_by_node=owner_by_node,
+            root_fqn=root_fqn,
+            is_backward=is_backward,
+        )
+        if owner is not None:
+            deps.append(dep)
+        elif owner_by_node.get(dep) is None and dep.op not in _GRAPH_BOUNDARY_OPS:
+            stack.extend(dep.all_input_nodes)
+    return tuple(deps)
+
+
+def _body_deps(
+    node: fx.Node,
+    *,
+    body: ChunkBody,
+    owner_by_node: dict[fx.Node, ChunkOwner],
+) -> tuple[fx.Node, ...]:
+    """Return same-region body deps, including deps behind unowned plumbing."""
+    deps: list[fx.Node] = []
+    for dep in node.all_input_nodes:
+        owner = _same_region_owner(
+            dep,
+            owner_by_node=owner_by_node,
+            root_fqn=body.owner.root_fqn,
+            is_backward=body.owner.is_backward,
+        )
+        if owner is not None:
+            deps.append(dep)
+        elif owner_by_node.get(dep) is None and dep.op not in _GRAPH_BOUNDARY_OPS:
+            deps.extend(
+                _hidden_body_deps(
+                    dep,
+                    owner_by_node=owner_by_node,
+                    root_fqn=body.owner.root_fqn,
+                    is_backward=body.owner.is_backward,
+                )
+            )
+    return tuple(dict.fromkeys(deps))
+
+
+def _body_ancestors(
+    nodes: tuple[fx.Node, ...],
+    *,
+    body: ChunkBody,
+    owner_by_node: dict[fx.Node, ChunkOwner],
+    node_set: set[fx.Node],
+) -> set[fx.Node]:
+    """Return same-body ancestors needed before the requested body nodes."""
+    ancestors: set[fx.Node] = set()
+    stack = [
+        dep
+        for node in nodes
+        for dep in _body_deps(node, body=body, owner_by_node=owner_by_node)
+        if dep in node_set
+    ]
+    while stack:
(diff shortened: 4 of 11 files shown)


sanketpurandare:
TL parse artifacts from the EP overlap graph-pass validation runs:

- Transformer-block batch overlap: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/1f5771e2-e435-4368-817d-10380c676e71/ep-overlap-transformer-block-batch-pass-completed-oom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000
- MoE-only batch overlap: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/sanketpurandare/b0e361c5-5ce4-41a6-8b5f-ffceea7dc641/ep-overlap-moe-batch-pass-completed-nccl/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

Verification notes:

- Transformer-block default mode applied `chunk_batch` to 12 layer regions and `ep_overlap` to 10 EP chunked regions. That matches the debug model shape: dense/non-MoE layers can be chunked but do not have EP all-to-all scheduling phases.
- MoE-only mode applied `chunk_batch` to 10 MoE regions and `ep_overlap` to 10 EP chunked regions.
- The MoE-only graph diff shows the expected EP duplication from chunking: `all_to_all_single.default: 25 -> 50`, `histc.default: 10 -> 20`, and `wait_tensor.default: 291 -> 366`.
- Generated rank0 code shows the intended launch/wait ordering for a MoE layer: chunk 0 dispatch launch, chunk 1 dispatch launch, then dispatch waits/compute, followed by chunk 0 combine launch, chunk 1 combine launch, then combine waits. The delayed wait-suffix nodes are tagged with `EP_wait=True` after the relevant all-to-all boundaries.

These artifacts validate the graph transformation and scheduling structure. They are not performance numbers; the performance/profile runs need to be collected separately on exclusively available GPUs.

xmfan (commented):
please include both a tlparse and a perfetto trace of before vs after

torchtitan/experiments/graph_trainer/ep_chunk_pass.py
```diff
@@ -1058,11 +1085,26 @@ def _scalar_sum_matches(
         return False
 
 
+def _scalar_halves_match_guarded_even_original(
+    chunk0_val: object, chunk1_val: object, original_val: object
+) -> bool:
+    try:
+        half_original = _expr(original_val) // 2
```
xmfan: configurable chunk size?

torchtitan/experiments/graph_trainer/chunk_passes.py
```diff
@@ -1776,3 +1846,219 @@ def chunk_seq_pass(
...
+
+
+def _phase_nodes(
+    chunk_nodes: list[fx.Node],
+    *,
+    is_backward: bool,
+    order: dict[fx.Node, int],
+) -> tuple[list[fx.Node], list[fx.Node], list[fx.Node]] | None:
+    dispatch_nodes = [node for node in chunk_nodes if _ep_region(node) == "dispatch"]
+    combine_nodes = [node for node in chunk_nodes if _ep_region(node) == "combine"]
```
xmfan: 1 loop

torchtitan/experiments/graph_trainer/chunk_passes.py
```diff
@@ -1776,3 +1846,219 @@ def chunk_seq_pass(
...
+    if not dispatch_nodes and not combine_nodes:
+        return None
```
xmfan: we should probably split up the dense vs moe paths, they can share a lot of logic, but it's a bit hard to reason about both at the same time

torchtitan/experiments/graph_trainer/chunk_passes.py
```diff
@@ -1776,3 +1846,219 @@ def chunk_seq_pass(
...
+        for node in ep_nodes:
+            if order[node] > last_idx:
+                custom = dict(_custom_meta(node))
+                custom["EP_wait"] = True
+                node.meta["custom"] = custom
+
+    return (
+        [node for node in chunk_nodes if order[node] <= first_idx],
+        [node for node in chunk_nodes if first_idx < order[node] <= second_idx],
+        [node for node in chunk_nodes if order[node] > second_idx],
```
xmfan: please piggy back off the same loops everywhere

torchtitan/experiments/graph_trainer/ep_chunk_pass.py
sanketpurandare: The chunking dataflow is generic over module granularity and the selected logical dimension, but this PR intentionally keeps the chunk count fixed at two because the EP overlap schedule is pairwise. I added a TODO in the pass contract calling out what N-way support would require: N-way provenance, materialization, scalar handling, eager references, and launch/wait scheduling.

torchtitan/experiments/graph_trainer/chunk_passes.py
sanketpurandare: Refactored the EP phase helper so it collects dispatch/combine nodes and last all-to-all launches in one pass over the chunk body, then does one boundary-aware pass to mark wait suffixes and partition phases. This removes the repeated comprehensions while keeping the validation checks local.

torchtitan/experiments/graph_trainer/chunk_passes.py
sanketpurandare: The user-facing control for this is `ep_overlap_modules`: `layers.*` selects transformer-block roots, while `layers.*.moe` selects only MoE roots. I also made the EP-overlap chunk phase skip matched roots with no EP dispatch/combine metadata, so transformer-block mode no longer chunks dense/non-EP layers unnecessarily. Regions with MoE dispatch/combine metadata go through EP scheduling.

torchtitan/experiments/graph_trainer/chunk_passes.py
sanketpurandare: Folded the repeated phase scans into the same collection/partition flow used by the EP scheduler. The helper still needs two conceptual passes because the wait boundaries are known only after seeing the last all-to-all in dispatch/combine, but it no longer rebuilds the same filtered lists multiple times.

torchtitan/models/common/token_dispatcher.py
IvanKobzarev, line 508: This signature change is failing the tests python -m pytest -q tests/unit_tests/test_expert_parallel.py

IvanKobzarev (approved):

aditvenk (approved):

torchtitan/models/common/token_dispatcher.py
sanketpurandare: Fixed the test

torchtitan/models/common/token_dispatcher.py
```diff
@@ -381,20 +449,17 @@ def _permute(
         Collapses token count matrix ``t_mat`` from ``(EP, e)`` to
         ``num_global_tokens_per_local_expert_e`` ``(e,)`` by summing across ranks.
         """
-        # pyrefly: ignore [missing-attribute]
-        ep_size = self.ep_mesh.size()
-        e = num_global_tokens_per_local_expert_E.shape[0] // ep_size
```
tianyu-l, line 432: why removing these and putting into args?

tianyu-l (approved):
unblock, but please answer my question

wwwjn (approved):
Stamp to unblock

torchtitan/models/common/token_dispatcher.py
sanketpurandare: The original motivation was:
  - dispatch() had already proven self.ep_mesh is not None and computed ep_size.
  - _permute() and TorchAOTokenDispatcher._permute() both needed ep_size and local expert count.
  - Passing them in avoided another self.ep_mesh.size() inside _permute() and removed the old pyrefly missing-attribute ignore caused by self.ep_mesh being typed as optional.

sanketpurandare: I can revert back the original behavior of recomputing it, all I want are separate functions for token count sync, dispatch and combine.

tianyu-l: sounds fine

torchtitan/models/common/token_dispatcher.py
```diff
@@ -530,22 +615,12 @@ def combine(
...
-                    dst=spmd.V,
-                    output_split_sizes=metadata.input_splits,
-                    input_split_sizes=metadata.output_splits,
-                )
+            routed_output_RD = self._combine_token_exchange(
+                routed_output_RD,
+                pg,
+                metadata.input_splits,
+                metadata.output_splits,
+            )
```
pianpwk, line 629: this looks like a plain refactor , but in case you're testing spmd_types breakage , it should be reasonable to run this test + spmd backend + typechecking (it's gated in integration_tests/features.py) : https://github.com/pytorch/torchtitan/blob/622c1acaa80b7ad7b3fcc4988fdb28375322375e/tests/integration_tests/features.py#L572-L585

maybe for qwen3 as well to be safe

after https://github.com/pytorch/torchtitan/pull/3673 landed

cc @tianyu-l

fegin (approved):

torchtitan/models/common/token_dispatcher.py
sanketpurandare: @pianpwk why were unit tests not added when making changes to token dispatcher? @tianyu-l when committing code changes we should add at-least one test that exercises new path. Adding tests in upward stack for a downward PR is only okay when you plan to land the stack all at once.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -254,6 +254,130 @@ class Config(BaseEPTokenDispatcher.Config):
...
+        ep_size: int,
+    ) -> torch.Tensor:
+        """Exchange per-rank expert token counts before the data all-to-all.
+
+        This method is separate from ``dispatch`` so graph passes can annotate
+        the count exchange independently from the true token-exchange
+        scheduling markers.
+        """
+        if (
+            torch.compiler.is_compiling() or torch.compiler._is_non_strict_tracing()
```
tianyu-l, line 270: just curious, does the two flag ever overlap -- can they both be true under any circumstance?

tianyu-l (approved):
sg

torchtitan/models/common/token_dispatcher.py
sanketpurandare: yes during nested compile, for example using flex attention under non-strict tracing

PR #3341, opened 2026-05-13
[graph_trainer] Fix full_inductor_compilation_pass on graphs with c10d collectives

tugsbayasgalan:
Summary:
``make_fx`` traces ``dist.*`` collectives as raw ``c10d.{op}_`` inplace ops with a torchbind ``ProcessGroup`` baked in as a graph attr. Inductor only recognizes the ``_c10d_functional.{op}`` + ``wait_tensor`` form, and the FakeTensor / compile_fx_inner cache key calls ``__eq__`` on the PG which is not implemented on the torchbind. Mirror ``standalone_compile``'s prep: functionalize the inplace collectives and unbox the PG before retracing with decompositions.

Test Plan:
New test ``test_full_inductor_pass_on_collective`` in test_trace_module.py runs the pass on a graph containing ``dist.all_reduce`` and asserts the compiled output matches eager.

### DIFF
--- a/torchtitan/experiments/graph_trainer/inductor_passes.py
+++ b/torchtitan/experiments/graph_trainer/inductor_passes.py
@@ -14,6 +14,10 @@ regional_inductor.
 from __future__ import annotations
 
 import torch
+from torch._inductor._functionalize_collectives import (
+    _functionalize_inplace_collectives,
+    _unbox_process_group_torchbinds,
+)
 from torch._inductor.compile_fx import compile_fx_inner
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/graph_trainer/inductor_passes.py
SherlockNoMad, line 206: I would attempt to rewrite full inductor to the following workflow 

tag all nodes in graph, us regional_inductor_pass, effective it should be the same as full_inductor.

PR #3343, opened 2026-05-13
[graph_trainer] Fix SAC remat to produce fresh FakeTensor storage for recomputed nodes

IvanKobzarev:
`selective_activation_remat_pass` uses `graph.node_copy` to duplicate forward nodes into the backward region.  `node_copy` shallow-copies `meta`, so `meta["val"]` (the FakeTensor) is the exact same object as the original -- both share the same storage `_cdata`.

The overlap scheduler's `MemoryTracker` (in `memory_estimator.py`) tracks memory via storage identity.  `GraphAliasTracker` attributes each storage to the first node that outputs it (the forward node). When the scheduler reorders the graph so that a recomputed node is scheduled before the forward allocator, the storage is never added to `current_live_storages`, and downstream consumers hit:

    AssertionError: all input storages should be currently allocated

This was observed in MAST job `SWEEPAP7_sfsdp_8b_NOCG-tp8-bs4-*` (Llama 8B, TP=8, SFSDP, `auto_bucketing` pass) where `_fused_rms_norm_48_recomputed` shared storage with `_fused_rms_norm_48`, and the scheduler placed the recomputed backward chain before the forward chain.

Fix: after `node_copy`, re-execute the duplicated node's target with its inputs' `meta["val"]` to produce a fresh FakeTensor with its own storage.  This is safe because recomputation by definition re-runs the same op on the same logical inputs.

Verified on 64-GPU MAST job: the original config that crashed at step 1 now completes all steps with conveyor PyTorch (no memory estimator changes needed).

### DIFF
--- a/torchtitan/experiments/graph_trainer/selective_activation_remat.py
+++ b/torchtitan/experiments/graph_trainer/selective_activation_remat.py
@@ -24,6 +24,41 @@ from torchtitan.experiments.graph_trainer.common_utils import _is_backward_node
 log = logging.getLogger(__name__)
 
 
+def _retrace_node_val(
+    node: fx.Node, fake_mode: torch._subclasses.fake_tensor.FakeTensorMode
+) -> None:
+    """Re-execute a node's op to produce a fresh ``meta["val"]`` with
+    independent storage.
+
+    ``graph.node_copy`` shallow-copies ``meta``, so the duplicate's
+    ``meta["val"]`` is the exact same FakeTensor (same storage ``_cdata``)
+    as the original.  Downstream passes (e.g., ``MemoryTracker`` in
+    overlap scheduling) rely on distinct storages to track allocations;
+    shared storages between forward and recomputed nodes cause assertion
+    failures when the scheduler reorders across them.
+
+    Re-executing the op (rather than ``empty_like``) correctly preserves
+    aliasing for view ops (reshape, transpose, split, etc.) whose output
+    must share storage with their input.
+    """
+    if node.op != "call_function" or "val" not in node.meta:
+        return
+
+    from torch._dispatch.python import enable_python_dispatcher
+
+    def get_val(x: object) -> object:
+        if isinstance(x, fx.Node):
+            return x.meta.get("val")
+        return x
+
+    fake_args = torch.fx.map_arg(node.args, get_val)
+    fake_kwargs = torch.fx.map_arg(node.kwargs, get_val)
+
+    with fake_mode, enable_python_dispatcher():
+        result = node.target(*fake_args, **fake_kwargs)
+    node.meta["val"] = result
+
+
 def _collect_backward_regions(
     gm: fx.GraphModule,
 ) -> list[tuple[int, int, bool]]:
@@ -91,6 +126,12 @@ def selective_activation_remat_pass(
             "in recompute regions. Please move RNG operations outside "
             "of recompute regions, or use joint graph mode (where partitioner handles RNG)."
         )
+
+    from torch._dynamo.utils import detect_fake_mode
+
+    fake_mode = detect_fake_mode(
+        tuple(node.meta.get("val") for node in gm.graph.find_nodes(op="placeholder"))
+    )
 
     regions = _collect_backward_regions(gm)
     if not regions:
@@ -287,6 +328,7 @@ def selective_activation_remat_pass(
             dup = gm.graph.node_copy(fwd_node, remat_input)
         dup.name = fwd_node.name + "_recomputed"
         dup.meta["autograd_backward"] = True
+        _retrace_node_val(dup, fake_mode)
         recomputed_nodes[fwd_node] = dup
         log.debug(
             "Recomputing %s before backward node %s", fwd_node.name, bwd_target.name


torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -24,6 +24,48 @@
...
+
+    fake_mode = detect_fake_mode(fake_args)
+    if fake_mode is None:
+        fake_mode = detect_fake_mode(fake_kwargs)
+
+    if fake_mode is not None:
+        with fake_mode, enable_python_dispatcher():
+            result = node.target(*fake_args, **fake_kwargs)
+    else:
+        result = node.target(*fake_args, **fake_kwargs)
```
SherlockNoMad: do we really need to recompute to create a new fake tensor? 

aten.empty_like should do ?

IvanKobzarev: Yes, for this particular bug this will create a new storage and that should be fine.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -24,6 +24,48 @@
...
+def _retrace_node_val(node: fx.Node) -> None:
+    """Re-execute a node's op to produce a fresh ``meta["val"]`` with
+    independent storage.
+
+    ``graph.node_copy`` shallow-copies ``meta``, so the duplicate's
+    ``meta["val"]`` is the exact same FakeTensor (same storage ``_cdata``)
+    as the original.  Downstream passes (e.g., ``MemoryTracker`` in
+    overlap scheduling) rely on distinct storages to track allocations;
+    shared storages between forward and recomputed nodes cause assertion
+    failures when the scheduler reorders across them.
```
SherlockNoMad, line 38: tbh, the root problem is that 

ppl shouldn't relying on faketensor's aliasing relationship to infer storage.

IvanKobzarev: But we need to predict aliasing for correct memory estimations, and we do not have any other information. 
aot_inductor and inductor also rely on this for aliasing analysis

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -24,6 +24,48 @@
...
+    fake_mode = detect_fake_mode(fake_args)
+    if fake_mode is None:
+        fake_mode = detect_fake_mode(fake_kwargs)
+
+    if fake_mode is not None:
+        with fake_mode, enable_python_dispatcher():
+            result = node.target(*fake_args, **fake_kwargs)
+    else:
+        result = node.target(*fake_args, **fake_kwargs)
+    node.meta["val"] = result
```
SherlockNoMad, line 59: can we just do new_val = torch.empty_like(node.meta["val"]) ?

SherlockNoMad (changes requested):

torchtitan/experiments/graph_trainer/selective_activation_remat.py
eellison: That is how we analysis all across the code base. reinplacing, memory, etc. what is a better way ?

sanketpurandare: I have used FakeTensor storage before for memory estimation and has worked quite reliably. There only an edge case in wait_tensor, where FakeTensor returns a tensor with new storage while in real execution wait_tensor is just a sync call.

SherlockNoMad: This approach is relying on fake impl correctly preserving the aliasing info, which I find it brittle.

maybe we already have good test coverage to guard this for aten ops, but for custom ops, and user provided fake impl, I think it's brittle.

SherlockNoMad: but I am not going to flight this, if this has been the convention used in many repos...

eellison: cc @tugsbayasgalan , @zou3519 on on-by-default fake tensor impl aliasing correctness checks. 

Nonetheless - that is, IMO, a separate question from the fix here, which is addressing the graph trainer breaking fx graph invariants.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
SherlockNoMad, line 58: another dump q, if all the inputs are already fake, do we still need to run it under fake mode?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -287,6 +329,7 @@ def remat_input(x: object) -> object:
             dup = gm.graph.node_copy(fwd_node, remat_input)
         dup.name = fwd_node.name + "_recomputed"
         dup.meta["autograd_backward"] = True
+        _retrace_node_val(dup)
```
SherlockNoMad: I also see some other passes manually manipulating node.meta[val] 

do you think it worth having a faketensor_propagation util function that runs at the end of pass ? 

sth just like gm.lint()

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -24,6 +24,48 @@
...
+        if isinstance(x, fx.Node):
+            return x.meta.get("val")
+        return x
+
+    fake_args = torch.fx.map_arg(node.args, get_val)
+    fake_kwargs = torch.fx.map_arg(node.kwargs, get_val)
+
+    fake_mode = detect_fake_mode(fake_args)
+    if fake_mode is None:
+        fake_mode = detect_fake_mode(fake_kwargs)
```
SherlockNoMad: this is getting fake_mode repeatly for many node, 

would it be better to detect_fake_mode_from_gm and pass the fake mode in ?

SherlockNoMad (approved):

torchtitan/experiments/graph_trainer/selective_activation_remat.py
IvanKobzarev: I think remat regions should be much smaller, than doing full retrace, which can be costly for large models that unroll all the layers . 10K nodes level. IMHO iterational correct FakeTensor retracing should be better for compile time.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
IvanKobzarev: yes, fixed it

PR #3356, opened 2026-05-14
Fix optimizer state and module state coupling

tugsbayasgalan:
When we pass in the optimizer that was initialized via param groups, the flat order of optimizer param states don't align with module param orders. This PR fixes it by reordering the user supplied param order according to optimizer param order. 

We don't want to make swapin_optim_state too complicated, hence doing the reordering outside of it.

### DIFF
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -310,21 +310,26 @@ def _reparametrize_train_state(
 ):
     """Reparametrize module and optimizer with explicit tensor state for tracing."""
     with contextlib.ExitStack() as stack:
-        if module is not None:
-            stack.enter_context(stateless._reparametrize_module(module, model_state))
         if optimizer is not None:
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -310,21 +310,26 @@ def _reparametrize_train_state(
...
+            # param_groups. Walk param_groups and resolve names by id() against
+            # the originals; must run before _reparametrize_module rebinds them.
+            id_to_name = {
+                id(p): n for n, p in module.named_parameters(remove_duplicate=False)
+            }
             params_for_optim = {
-                name: model_state[name]
-                for name, _ in module.named_parameters(remove_duplicate=False)
+                id_to_name[id(p)]: model_state[id_to_name[id(p)]]
+                for group in optimizer.param_groups
```
ydwu4, line 323: Changes looks good but this double loop looks suspicious. Is this the only way?

ydwu4 (approved):

SherlockNoMad (approved):

PR #3357, opened 2026-05-14
[rl] Route RL CI with pytest markers

felipemello1:
TLDR:
- Added decorators "@gpu_test" and "@h100_test". 
- Cpu tests are automatically picked up by the new cpu CI. 
- The gpu marked tests are routed to the proper machines. 
- yaml doesn't have custom torchrun commands to run specific test files. It just checks all tests with the proper marks (or the lack of them)

------

Adding a new RL test currently is brittle. The expected behavior is that when you write a test file, the CI would just pick it up. However, we currently have 2 yaml files: one for h100 and another for a cheaper gpu. In both of them, we have a torchrun command to run a **specific file**.

This means that, if a dev writes a new file with tests, CI won't automatically pick it up, unless the yaml is edited as well.

This PR refactors the CI and the current unit tests so that tests are routed by flags: @gpu_test and @h100_test. It also shifts the responsability of who executes torchrun: the test does, not the yaml.

## Writing RL tests going forward

- CPU test: add a normal `test_*.py` test with no marker.
- CUDA test: add `@gpu_test(num_gpus=N)`.
- H100 test: add `@h100_test(num_gpus=N)`.
- GRPO integration config: append to `test_integration.py`.
- Distributed check needing `torchrun`: add a marked pytest wrapper and put the torchrun payload in `tests/scripts/`, writing rank-0 JSON to `--result-path`.

## Downstream changes
**Problem 1**: GRPO integration tests used a custom runner launched with torchrun by the yaml
**Solution 1**: Make GRPO configs pytest cases

`test_integration.py` now parametrizes `OverrideDefinitions` as normal pytest cases:

- `build_rl_test_list()` cases use marker `gpu_test(num_gpus=case.ngpu)`
- `build_rl_h100_test_list()` cases use marker `h100_test(num_gpus=case.ngpu)`
- subprocess launch uses `sys.executable` and argv lists

**Problem 2**: Bitwise parity had similar issues as above
**Solution 2**: Keep pytest as the router, use a script as the worker

`test_bitwise_parity.py` is a small H100-marked pytest wrapper. It launches `tests/scripts/bitwise_parity.py` under `torchrun`, then loads the JSON result written by rank 0 and runs assertions on it.

**Problem 3**: No CPU CI
**Solution 3**: added it

### Other cleanup

- Deleted the old `integration_tests.py` custom runner.
- Removed stale RL package re-exports from `torchtitan/experiments/rl/__init__.py`.
- Added RL paths/workflows to the `ciflow/rl` labeler rule (necessary to avoid warnings).

### For the future

We have to install torchvision to avoid import errors. We should check how to protect these imports in titan, so setup is faster in CI.

### Test

GPU tests: tested locally
CPU tests from CI 
image


torchtitan/experiments/rl/tests/integration_tests.py
felipemello1, line 1: this is now `test_integration.py`

torchtitan/experiments/rl/tests/scripts/bitwise_parity.py
felipemello1, line 1: previously `test_bitwise_parity.py`, which was executed with `torchrun` by the yaml. The code was moved here, and the `test_bitwise_parity.py` is a regular pytest that calls this script with torchrun. i.e. calling torchrun is the test's responsability, not the yaml's, so the yaml is unaware of what the test is doing or needs

torchtitan/experiments/rl/tests/test_bitwise_parity.py
felipemello1, line 1: launches with torchrun `tests/scripts/bitwise_parity.py` and loads the json produced to assert results

torchtitan/experiments/rl/tests/test_integration.py
felipemello1, line 1: same as `integration_tests.py`, just:
a. renamed, to be a test (file has to start with "test_")
b. used parametrize + @gpu_test and @h100_test markers

torchtitan/experiments/rl/tests/integration_tests.py
felipemello1, line 1: renamed to `test_integration.py`, so its picked up by pytest

.github/workflows/integration_test_4gpu_rl.yaml
```diff
@@ -95,7 +95,8 @@ jobs:
...
 
-        # Run E2E RL integration tests (TP=2 on 4 GPUs)
-        python -m torchtitan.experiments.rl.tests.integration_tests \
-          $RUNNER_TEMP/artifacts-to-be-uploaded --ngpu 4 \
-          --hf_assets_path "$MODEL_PATH"
+        # runs tests tagged with @gpu and NOT @h100_gpu
+        RL_TEST_OUTPUT_DIR="$RUNNER_TEMP/artifacts-to-be-uploaded" \
+        HF_ASSETS_PATH="$MODEL_PATH" \
+          python -m pytest torchtitan/experiments/rl/tests \
+          -m "gpu and not h100" --durations=20 -v
```
wwwjn, line 102: Does pytest support using this language to filter the markers? This sounds cool

torchtitan/experiments/rl/tests/README.md
```diff
@@ -0,0 +1,17 @@
+# RL tests
+
+## Routing
+
+| Test type | Location | Marker |
+|---|---|---|
+| CPU unit test | `tests/test_*.py` | none |
+| CUDA unit test | `tests/test_*.py` | `@gpu_test(num_gpus=N)` |
```
wwwjn, line 8: The name is confusing, today `@gpu_test` and `@h100_test` are used to label "integration test", not unit test. Technically we don't have GPU unit test today. 

To be clear on those terminology:
- GPU unit test: For example I want to test `class MoE()` using vllm inference, I still need `torchrun` to launch SPMD runs, only mark a pytest with @gpu_test is not enough. It's more like your "Distributed check that needs `torchrun`" 
- Integration test: It must on GPU, we just need to 

We should consider following categorization: 
- CPU unit test
- GPU unit test: similar to your "Distributed check that needs `torchrun`" , we need to make pytest + torchrun launch shared across GPU unit tests
- H100 Integration tests
- Integration tests

torchtitan/experiments/rl/tests/test_bitwise_parity.py
```diff
@@ -1,563 +1,114 @@
...
+
+class TestBitwiseParityIntegration:
+    @h100_test(num_gpus=2)
+    def test_bitwise_parity(self, tmp_path):
+        result_path = tmp_path / "bitwise_parity_result.json"
+        result = subprocess.run(
+            [
+                sys.executable,
+                "-m",
+                "torch.distributed.run",
```
wwwjn, line 79: Is it same as torchrun?

torchtitan/experiments/rl/tests/test_bitwise_parity.py
```diff
@@ -1,563 +1,114 @@
...
-        registry_to_vllm(
-            config.model_spec,
-            parallelism=config.generator.parallelism,
-            compile_config=config.compile,
+    return "\n".join(parts)
+
+
+class TestBitwiseParityIntegration:
+    @h100_test(num_gpus=2)
+    def test_bitwise_parity(self, tmp_path):
```
wwwjn, line 73: And can we bake `bitswise_parity.py` into this file? No need to have 2 separate file.

wwwjn (commented):
Thanks, I want to discuss more around a shared / common infra for launching GPU unit tests. We can make the way of launching torchtitan/experiments/rl/tests/test_bitwise_parity.py shared and easier for adding more GPU unit tests

Using pytest mechanism to distinguish between normal integration test and H100 integration test makes sense to me.

.github/workflows/integration_test_4gpu_rl.yaml
felipemello1: yes, because they are registered. We can also install pytest.integration_test (or something like that), and it enables some flags and markers as well.

torchtitan/experiments/rl/tests/README.md
felipemello1: I feel like there are two axis:
1. type of hardware that defines which CI will run it (cpu, h100, other gpu)
2. some category of test that we may want to run at difference cadence (e.g. integration tests are slow and we may not want it in CI)

I would **not** say that integration_test == torchrun or integration_test==gpu.

We though about adding a tag @integration_test, but it was not clear to me what this would give us atm

>we need to make pytest + torchrun launch shared across GPU unit tests

wdym by shared?

> I still need torchrun to launch SPMD runs, only mark a pytest with @gpu_test is not enough

The test launches it with torchrun, it doesnt need a special category other than `@ gpu`, unless we want to flag it as slow or something like that

---

Please let me know if you disagree and what actions do you think i should take. Are you saying that we should add @integration_test decorator?

torchtitan/experiments/rl/tests/test_bitwise_parity.py
felipemello1: yes!  torchrun is the console-script wrapper for python -m torch.distributed.run.

These are equivalent:

torchrun --nproc_per_node=4 script.py

python -m torch.distributed.run --nproc_per_node=4 script.py

torchtitan/experiments/rl/tests/test_bitwise_parity.py
felipemello1: hmmm, we could, idk if we should. The other file is so long that it made sense to me to be its own script, so its easy to distinguish the actual pytest from the script. This is similar to how integrations tests also runs other scripts.

I dont have a strong opinion though, happy to put them together if you think its better.

.github/workflows/integration_test_8gpu_rl_h100.yaml
tianyu-l, line 1: @wwwjn we no longer need this yaml after your refactor, right? Because non-h100 can also test bitwise parity

.github/workflows/unit_test_cpu_rl.yaml
```diff
@@ -0,0 +1,56 @@
...
+        # TODO: Drop torchvision once CPU collection no longer imports vLLM.
+        # vLLM imports torchvision today, so keep it ABI-matched with torch.
+        uv pip install torch torchvision vllm torchcomms --pre \
+          --extra-index-url https://download.pytorch.org/whl/nightly/cu130 \
+          --index-strategy unsafe-best-match
+
+        uv pip install -e .
+
+        python -m pytest torchtitan/experiments/rl/tests \
+          -m "not gpu" --durations=20 -v
```
tianyu-l, line 56: this mean "not gpu and not h100"?

torchtitan/experiments/rl/tests/integration_tests.py
tianyu-l: why this refactor only touches RL? This is "violating" global pattern.

torchtitan/experiments/rl/tests/README.md
tianyu-l: Today's integration tests almost always mean E2E tests running the train script with some config override.

I think we should still clearly separate them clearly
- CPU unit test
- GPU unit test (distributed), which doesn't exist today.
- GPU integration / E2E test

pyproject.toml
```diff
@@ -61,6 +61,10 @@ include = ["torchtitan*"]
 [tool.pytest.ini_options]
 addopts = ["--showlocals"]  # show local variables in tracebacks
 testpaths = ["tests"]
+markers = [
+    "gpu(num_gpus): RL test requiring CUDA GPUs; num_gpus records the minimum GPU count",
+    "h100: RL test requiring H100 GPUs; use h100_test so the test is also marked gpu",
+]
```
tianyu-l, line 67: what does this unlock -- e.g. do we have to put it in pyproject.toml for `pytest` to discover and allow `pytest.mark.gpu`

torchtitan/experiments/rl/tests/README.md
felipemello1: so

1. add a @integration_test decorator
2. apply it to test_integration.py and test_bitwise_parity
3. 
Have 3 cis: cpu, integration, gpu? 
or 
Have 2 CIs: just cpu + [integration+gpu]?

torchtitan/experiments/rl/tests/integration_tests.py
felipemello1: Yeah, i agree that its different from other integration tests in titan. I dont think that refactoring all of titan is worth it. I think i will revert it and let the yaml decide which tests to call for integration.

We just need to decide about the integration test tag.

torchtitan/experiments/rl/tests/README.md
tianyu-l: we should have
- CPU unit
- GPU unit (missing)
- GPU e2e h100
- GPU e2e

each might be separated by jobs
- features, models, rl, graph_trainer, etc.

or would you propose something new?

.github/workflows/integration_test_8gpu_rl_h100.yaml
wwwjn: Yes I was rebasing and trying to land my PR to remove this file

torchtitan/experiments/rl/tests/README.md
wwwjn: > I would **not** say that integration_test == torchrun or integration_test==gpu.

Yes I agree.  integration test today in torchtitan core is mostly running torchrun (because of SPMD) but RL is using `python`. Today all integration tests runs on GPU.  By "integration test", I was referring to E2E test. 

I agree on Tianyu's classification, and we should go with the categories. I was suggesting we are should setup GPU unit test infra first (figure out the code we can share to launch GPU unit tests, and the way for others to add more GPU unit tests)

> add a @integration_test decorator
The decorator is used by `pytest` to "classify" tests , today we are using each yaml files to classify them. What's the advantage of using a decorator than using yaml files?

And put the launching command (eg, `torchrun` or `pytorch` within the pytest itself), I'm supportive of this and I think it's more clear.

torchtitan/experiments/rl/tests/test_bitwise_parity.py
wwwjn: I know it's quiet long but the bitwise_parity.py now itself it's very specific to a model and test scenario, and it's not flexible enough to be reused / used by other users.  

I'm actually ok with both, and I think we can put it as a separate script and make it more flexible by passing some kwargs, so user can reuse this script to run locally on other models / parallelism settings (in the future)

PR #3317, opened 2026-05-11
Make FlexShard traceable by torch.compile

weifengpy:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #3502
* #3537
* #3603
* __->__ #3317
* #3239

PR #3359, opened 2026-05-14
Add Owned placement and placement-owned collectives

weifengpy:
* __->__ #3359
* #3317
* #3239

Move bucket unshard and gradient-reduction collective selection into the placement contract so Shard owns all-gather/reduce-scatter and Owned can own broadcast/reduce-to-owner. Rename bucket runtime handles to placement-neutral unshard/reduce-grad terminology and add Owned placement tests plus a support plan.


weifengpy:
Folded into #3239 so the first FlexShard PR proves the placement contract with both Shard and Owned examples.

Issue #2956, opened 2026-04-14
[MoE] Token dispatcher and MoE cleanup TODOs


acisseJZhong:
## MoE / Token Dispatcher Cleanup TODOs

A collection of cleanup tasks for the MoE token dispatcher code following https://github.com/pytorch/torchtitan/pull/2842

- [ ] **Change tensor shape suffix namings**
Standardize tensor variable naming to use shape suffixes consistently (e.g. `_NxK`, `_ED`) across token dispatchers and MoE modules for readability. See more in https://medium.com/@NoamShazeer/shape-suffixes-good-coding-style-f836e72e24fd

- [ ] **Extract shared local reordering helper in token dispatchers**
The local reordering block (histc, argsort, score application, token gathering) is duplicated between `LocalTokenDispatcher.dispatch()` and `AllToAllTokenDispatcher.dispatch()` (~20 lines each). Extract into a `_local_reorder(x, top_scores, selected_experts_indices)` method on the base class that returns `(routed_input, num_tokens_per_expert, token_indices_experts_sorted, top_scores_experts_sorted)`.

- [ ] **Deduplicate `num_tokens_per_expert` computation between router and dispatchers**
`torch.histc(selected_experts_indices...)` is computed in both `TokenChoiceTopKRouter.forward()` (for load balancing) and inside each dispatcher's `dispatch()` (for expert computation). Consider having the dispatcher accept the router's `num_tokens_per_expert` as input to avoid the redundant histc call.

- [ ] **Remove duplicated `forward()` in `GptOssGroupedExperts`**
`GptOssGroupedExperts.forward()` is identical to `GroupedExperts.forward()` (dispatch → `_experts_forward` → combine). Since `GptOssGroupedExperts` inherits from `GroupedExperts`, the override can be removed — only `_experts_forward()` needs to differ.

- [ ] **Clean up logging in parallelize.py**
Consolidate and clean up the logging statements in `parallelize.py` for MoE parallelization (EP, ETP, TP). Also remove `comm_backend: str = "standard"` from `apply_moe_ep_tp` — the comm backend is now encoded in the token dispatcher config and no longer needed as a separate parameter.

- [x] **Unify MoE and DeepEPMoE**
`DeepEPMoE` overrides `MoE.forward()` to handle async combine + shared_experts overlap. Explore unifying them so the base `MoE.forward()` handles both NCCL and DeepEP paths, reducing code duplication.

[...]

PR #3360, opened 2026-05-14
Enables HybridEP torch.compile support and adds eager integration tests

syed-ahmed:
Make the HybridEP dispatch custom op compatible with torch.compile by moving buffer initialization inside the op impl (opaque to dynamo) and passing ep_size/group_name through the op schema instead of relying on global state accessed during tracing.

Key changes:
- hybridep.py: Uses custom_op to resolve schema. Resolve the ProcessGroup from the name inside _dispatch_impl and call get_buffer there, avoiding dynamo tracing issues with _disable_current_modes and pybind11 objects. Fix buffer_config access for the refactored HybridEPBuffer API (configurer.buffer_config). Add initialization logging.
- config_registry.py: Add deepseek_v3_debugmodel_hybridep config with non-blocking dispatch (capacity_factor=1.0).
- models.py: Add FSDP+HybridEP and FSDP+HybridEP+compile integration tests.
- CI: Move DeepEP installation from graph_trainer CI yaml into install_deepep.sh.

Assistant used: Claude Opus 4.7 (1M context)

### DIFF
--- a/tests/integration_tests/h100.py
+++ b/tests/integration_tests/h100.py
@@ -4,12 +4,15 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import importlib
 import logging
 
 from tests.integration_tests import OverrideDefinitions
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+
+_DEEPEP_AVAILABLE = importlib.util.find_spec("deep_ep") is not None
 
 
 def build_h100_tests_list() -> list[OverrideDefinitions]:
@@ -82,6 +85,21 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             "HSDP+CP+torch.compile+Float8",
             "hsdp+cp+compile+float8",
             ngpu=8,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module deepseek_v3 --config deepseek_v3_debugmodel_hybridep",
+                    "--parallelism.data_parallel_shard_degree 4",
+                    "--parallelism.expert_parallel_degree 2",
+                    "--compile.enable",
+                    "--compile.components model,loss",
+                ],
+            ],
+            "DeepSeek V3 FSDP+HybridEP+compile",
+            "deepseek_v3_fsdp+hybridep+compile",
+            ngpu=4,
+            disabled=not _DEEPEP_AVAILABLE,
         ),
     ]
     return integration_tests_flavors
--- a/tests/integration_tests/h100.py
+++ b/tests/integration_tests/h100.py
@@ -4,15 +4,12 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import importlib
 import logging
 
 from tests.integration_tests import OverrideDefinitions
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
-
-_DEEPEP_AVAILABLE = importlib.util.find_spec("deep_ep") is not None
 
 
 def build_h100_tests_list() -> list[OverrideDefinitions]:
@@ -99,7 +96,6 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             "DeepSeek V3 FSDP+HybridEP+compile",
             "deepseek_v3_fsdp+hybridep+compile",
             ngpu=4,
-            disabled=not _DEEPEP_AVAILABLE,
         ),
     ]
     return integration_tests_flavors
--- a/tests/integration_tests/h100.py
+++ b/tests/integration_tests/h100.py
@@ -19,7 +19,6 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
     same root config file.
     """
     integration_tests_flavors = [
-        # TODO: re-enable this test once the async TP issue is fixed
         OverrideDefinitions(
             [
                 [
@@ -30,7 +29,6 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             ],
             "2D async TP compile",
             "2d_asynctp_compile",
-            disabled=True,
         ),
         OverrideDefinitions(
             [
@@ -52,7 +50,6 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             "fsdp_symm_mem",
             ngpu=2,
         ),
-        # TODO: re-enable this test once the async TP issue is fixed
         OverrideDefinitions(
             [
                 [
@@ -67,7 +64,6 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             "FSDP+async TP+PP+torch.compile+Float8",
             "fsdp+tp+cp+compile+float8",
             ngpu=8,
-            disabled=True,
         ),
         OverrideDefinitions(
             [
--- a/tests/integration_tests/models.py
+++ b/tests/integration_tests/models.py
@@ -5,10 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
-import importlib
 from tests.integration_tests import OverrideDefinitions
-
-_DEEPEP_AVAILABLE = importlib.util.find_spec("deep_ep") is not None
 
 
 def build_model_tests_list() -> list[OverrideDefinitions]:
@@ -34,21 +31,6 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
             "DeepSeek V3 FSDP+EP+compile",
             "deepseek_v3_fsdp+ep+compile",
             ngpu=4,
-        ),
-        OverrideDefinitions(
-            [
-                [
-                    "--module deepseek_v3 --config deepseek_v3_debugmodel_hybridep",
-                    "--parallelism.data_parallel_shard_degree 4",
-                    "--parallelism.expert_parallel_degree 2",
-                    "--compile.enable",
-                    "--compile.components model,loss",
-                ],
-            ],
-            "DeepSeek V3 FSDP+HybridEP+compile",
-            "deepseek_v3_fsdp+hybridep+compile",
-            ngpu=4,
-            disabled=not _DEEPEP_AVAILABLE,
         ),
         OverrideDefinitions(
             [
--- a/tests/integration_tests/models.py
+++ b/tests/integration_tests/models.py
@@ -5,7 +5,10 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import importlib
 from tests.integration_tests import OverrideDefinitions
+
+_DEEPEP_AVAILABLE = importlib.util.find_spec("deep_ep") is not None
 
 
 def build_model_tests_list() -> list[OverrideDefinitions]:
@@ -31,6 +34,21 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
             "DeepSeek V3 FSDP+EP+compile",
             "deepseek_v3_fsdp+ep+compile",
             ngpu=4,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module deepseek_v3 --config deepseek_v3_debugmodel_hybridep",
+                    "--parallelism.data_parallel_shard_degree 4",
+                    "--parallelism.expert_parallel_degree 2",
+                    "--compile.enable",
+                    "--compile.components model,loss",
+                ],
+            ],
+            "DeepSeek V3 FSDP+HybridEP+compile",
+            "deepseek_v3_fsdp+hybridep+compile",
+            ngpu=4,
+            disabled=not _DEEPEP_AVAILABLE,
         ),
         OverrideDefinitions(
             [
--- a/tests/integration_tests/models.py
+++ b/tests/integration_tests/models.py
@@ -22,7 +22,7 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
-                    "--module deepseek_v3 --config deepseek_v3_debugmodel_ep",
+                    "--module deepseek_v3 --config deepseek_v3_debugmodel",
                     "--parallelism.data_parallel_shard_degree 4",
                     "--parallelism.expert_parallel_degree 2",
                     "--compile.enable",
@@ -35,7 +35,7 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
-                    "--module deepseek_v3 --config deepseek_v3_debugmodel_ep",
+                    "--module deepseek_v3 --config deepseek_v3_debugmodel",
                     "--parallelism.pipeline_parallel_degree 2",
                     "--parallelism.pipeline_parallel_schedule Interleaved1F1B",
                     "--parallelism.data_parallel_shard_degree 2",
@@ -50,7 +50,7 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
         OverrideDefinitions(
             [
                 [
-                    "--module deepseek_v3 --config deepseek_v3_debugmodel_ep",
+                    "--module deepseek_v3 --config deepseek_v3_debugmodel",
                     "--parallelism.data_parallel_replicate_degree 2",
                     "--parallelism.data_parallel_shard_degree 2",
                     "--parallelism.expert_parallel_degree 2",
--- a/torchtitan/components/quantization/mx.py
+++ b/torchtitan/components/quantization/mx.py
@@ -144,9 +144,6 @@ def _get_mxfp8_grouped_experts_cls(parent_cls: type) -> type:
 class MXFP8GroupedExpertsConverter(QuantizationConverter):
     """Apply MXFP8 quantization to MoE expert grouped GEMMs."""
 
-    # MXFP8: scaling block size is (1 x 32), so contracting dim must be divisible by 32.
-    PAD_MULTIPLE = 32
-
     @dataclass(kw_only=True, slots=True)
     class Config(QuantizationConverter.Config):
         recipe_name: Literal["mxfp8_rceil"] = "mxfp8_rceil"
@@ -154,6 +151,11 @@ class MXFP8GroupedExpertsConverter(QuantizationConverter):
         Quantization recipe name for grouped GEMMs. Options: ["mxfp8_rceil"]
 
         - mxfp8_rceil: MXFP8 dynamic quantization with RCEIL rounding mode when computing the e8m0 scale factors.
+        """
+        pad_multiple: int = 32
+        """
+        Pad per-expert token groups to this multiple for MXFP8 grouped GEMM alignment.
+        The CuTeDSL quantization kernel on sm_100 requires multiples of 128.
         """
 
     def __init__(self, config: Config):
@@ -176,7 +178,7 @@ class MXFP8GroupedExpertsConverter(QuantizationConverter):
 
     def convert(self, model_config) -> None:
         for _fqn, config, parent, attr in model_config.traverse(GroupedExperts.Config):
-            swap_token_dispatcher(config, self.PAD_MULTIPLE)
+            swap_token_dispatcher(config, self.config.pad_multiple)
             base_module_cls = type(config)._owner
             quantized_cls = _get_mxfp8_grouped_experts_cls(base_module_cls)
             config_cls = quantized_cls.Config  # type: ignore[attr-defined]
--- a/torchtitan/distributed/deepep/hybridep.py
+++ b/torchtitan/distributed/deepep/hybridep.py
@@ -22,12 +22,12 @@ from dataclasses import dataclass
 from typing import Any
 
 import torch
+import torch.distributed as dist
 from torch._library.opaque_object import (
     get_opaque_type_name,
     OpaqueBase,
     register_opaque_type,
 )
-import torch.distributed as dist
 
 from torchtitan.tools.logging import logger
 
@@ -176,6 +176,7 @@ def _dispatch_impl(
     """
     num_local_experts = num_experts // ep_size
 
+    # pyrefly: ignore [bad-argument-type]
     group = dist.distributed_c10d._resolve_process_group(group_name)
     get_buffer(
         group=group,
@@ -347,7 +348,9 @@ def _combine_bwd_impl(
         handle=handle.value,
     )
     if grad_probs_dense is None:
-        grad_probs_dense = torch.empty(0, device=grad_hidden.device, dtype=torch.float32)
+        grad_probs_dense = torch.empty(
+            0, device=grad_hidden.device, dtype=torch.float32
+        )
     return grad_x, grad_probs_dense
 
 
--- a/torchtitan/distributed/deepep/hybridep.py
+++ b/torchtitan/distributed/deepep/hybridep.py
@@ -27,8 +27,9 @@ from torch._library.opaque_object import (
     OpaqueBase,
     register_opaque_type,
 )
-from torch.distributed import ProcessGroup
-from torch.utils._python_dispatch import _disable_current_modes
+import torch.distributed as dist
+
+from torchtitan.tools.logging import logger
 
 _buffer: Any = None  # Global buffer instance
 
@@ -96,12 +97,27 @@ _handle_type = get_opaque_type_name(DispatchHandle)
 torch.library.define(
     "hybridep::dispatch",
     f"(Tensor x, Tensor topk_idx, Tensor topk_weights, int num_experts, "
-    f"bool non_blocking, float? moe_expert_capacity_factor, int? pad_multiple) -> (Tensor, Tensor, Tensor, {_handle_type})",
+    f"int ep_size, str group_name, bool non_blocking, "
+    f"float? moe_expert_capacity_factor, int? pad_multiple) "
+    f"-> (Tensor, Tensor, Tensor, {_handle_type})",
 )
 
 torch.library.define(
     "hybridep::combine",
     f"(Tensor x, {_handle_type} handle, int num_tokens, int? pad_multiple) -> Tensor",
+)
+
+# Backward ops: these wrap the communication calls so AOT Autograd can trace them
+# dispatch_bwd: used in combine's backward (scatters gradients via dispatch_with_permute)
+torch.library.define(
+    "hybridep::dispatch_bwd",
+    f"(Tensor grad_combined, {_handle_type} handle, int num_permuted_tokens, int? pad_multiple) -> Tensor",
+)
+
+# combine_bwd: used in dispatch's backward (gathers gradients via combine_with_unpermute)
+torch.library.define(
+    "hybridep::combine_bwd",
+    f"(Tensor grad_hidden, Tensor grad_scores, {_handle_type} handle, int num_tokens, int num_experts) -> (Tensor, Tensor)",
 )
 
 
@@ -142,6 +158,8 @@ def _dispatch_impl(
     topk_idx: torch.Tensor,
     topk_weights: torch.Tensor,
     num_experts: int,
+    ep_size: int,
+    group_name: str,
     non_blocking: bool = False,
     moe_expert_capacity_factor: float | None = None,
     pad_multiple: int | None = None,
@@ -156,13 +174,16 @@ def _dispatch_impl(
       then reads tokens_per_expert from pinned CPU memory to compute the
       exact num_permuted_tokens on the host.
     """
-    global _buffer
-    if _buffer is None:
-        raise RuntimeError(
-            "HybridEP buffer not initialized. Call dispatch_tokens() first."
-        )
+    num_local_experts = num_experts // ep_size
 
-    num_local_experts = num_experts // _buffer.group_size
+    group = dist.distributed_c10d._resolve_process_group(group_name)
+    get_buffer(
+        group=group,
+        hidden_dim=x.shape[1],
+        num_tokens=x.shape[0],
+        num_local_experts=num_local_experts,
+    )
+
     from deep_ep.hybrid_ep_buffer import indices_to_map
 
     routing_map, probs = indices_to_map(
@@ -176,7 +197,7 @@ def _dispatch_impl(
         ), "moe_expert_capacity_factor is required for non_blocking dispatch"
         num_permuted_tokens = _num_permuted_tokens_for_non_blocking(
             x.shape[0],
-            _buffer.group_size,
+            ep_size,
             num_local_experts,
             topk_idx.shape[1],
             moe_expert_capacity_factor,
@@ -217,16 +238,18 @@ def _dispatch_fake(
     topk_idx: torch.Tensor,
     topk_weights: torch.Tensor,
     num_experts: int,
+    ep_size: int,
+    group_name: str,
     non_blocking: bool = False,
     moe_expert_capacity_factor: float | None = None,
     pad_multiple: int | None = None,
 ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, DispatchHandle]:
     """Fake dispatch for torch.compile tracing."""
-    num_local_experts = num_experts // _buffer.group_size
+    num_local_experts = num_experts // ep_size
     if non_blocking:
         out_tokens = _num_permuted_tokens_for_non_blocking(
             x.shape[0],
-            _buffer.group_size,
+            ep_size,
             num_local_experts,
             topk_idx.shape[1],
             moe_expert_capacity_factor,  # pyrefly: ignore [bad-argument-type]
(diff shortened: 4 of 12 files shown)


syed-ahmed:
There are few things to unpack here:
- When @elfiegg 's HybridEP landed in , torch.compile was not being used for expert computation: https://github.com/pytorch/torchtitan/blob/156db2eadada6ad25324bbf9f4ee396a79faa914/torchtitan/models/llama4/parallelize.py#L661-L669
- Fast forward, the full transformer block is now compiled with fullgraph=True. Without the changes in this PR, HybridEP initialization causes graph breaks.
- https://github.com/pytorch/torchtitan/commit/a5a60cc2adae61069705e5e0facb1cab6a5ec1dd added `_disable_current_modes` for DeepEP `Buffer.__init__ ` calling `all_gather_object()`. However, HybridEP buffer object doesn't do that: https://github.com/deepseek-ai/DeepEP/blob/hybrid-ep/deep_ep/hybrid_ep_buffer.py. So I believe adding _disable_current_modes for HybridEP was wrong to begin with. We need to remove it first because it's not torch.compilable. Also remove the `_buffer is None` condition because `get_buffer` already checks that, and I needed that originally to get around _disable_current_modes in graph trainer as well (https://github.com/pytorch/torchtitan/commit/115f4c9cfce6cffde4052c5a9cabf463df709660).
- After that, we need to move get_buffer into the custom op, so that dynamo doesn't trace it causing graph breaks. That brings in all the `ep_size` and ProcessGroup string plumbing.
- Now that things are compilable, grouped_mm backward fails when running with compile, most likely because Inductor materializes the transposed X^T as an intermediate buffer with column-major strides (I'm not the expert here). It then passes this non-contiguous buffer directly to extern_kernels._grouped_mm. On GB200, the CUTLASS _grouped_mm kernel fails to initialize with these column-major strides. Note this doesn't happen on eager. Adding .contiguous() before _grouped_mm in the forward ensures the weight tensors are row-major. This propagates through autograd - inductor's backward intermediate buffers for the grad-weight computation are then also contiguous, so CUTLASS initializes successfully.

tests/integration_tests/models.py
```diff
@@ -32,6 +32,60 @@ def build_model_tests_list() -> list[OverrideDefinitions]:
             "deepseek_v3_fsdp+ep+compile",
             ngpu=4,
         ),
+        OverrideDefinitions(
```
tianyu-l: The additions will add too much burden to current models.py CI. I'd recommend adding a single test, or better, repurposing an existing deepseek test to use hybridep.

torchtitan/distributed/deepep/hybridep.py
tianyu-l, line 1: @shuhuayu could you help review this file

torchtitan/models/common/moe.py
```diff
@@ -63,17 +63,13 @@ def _experts_forward(
...
-        )
-        h = h * torch._grouped_mm(
-            x.bfloat16(), w3.bfloat16().transpose(-2, -1), offs=offsets
-        )
-        return torch._grouped_mm(
-            h, w2.bfloat16().transpose(-2, -1), offs=offsets
-        ).type_as(x)
+        w1t = w1.bfloat16().transpose(-2, -1).contiguous()
+        w2t = w2.bfloat16().transpose(-2, -1).contiguous()
+        w3t = w3.bfloat16().transpose(-2, -1).contiguous()
```
tianyu-l: > moe.py: Make transposed weights contiguous before _grouped_mm to fix CUTLASS initialization failures on sm_100 (GB200) in the compiled backward pass.

Hmm didn't get it. I think we deliberately make it so that after transpose it's column major.

syed-ahmed: I actually don't know much about the grouped_mm path that inductor takes. Without these .contingous() calls, this is exact error I'm seeing. The other workaround is, if I set pad_multiple=8 for the DeepEPTokenDispatcher. I wonder if the hybrid ep path was lacking some improvements from: https://github.com/pytorch/torchtitan/issues/2956.

```
[rank0]:[titan] 2026-05-14 16:00:12,548 - root - INFO - Initializing HybridEP buffer: hidden_dim=256, max_tokens_per_rank=16384, num_local_experts=4, ep_size=2                                                                                                                                                                                                                                                                                                                                                                                                
[rank0]:/opt/pytorch/pytorch/aten/src/ATen/native/cuda/GroupMMCommon.cuh:64: prepare_grouped_gemm_data: block: [0,0,0], thread: [1,0,0] Assertion `delta % align == 0 && "expected input tensor dynamic dimension byte size to be non-negative multiple of 16\n"` failed.                                                                                                                                                                                                                                                                                      
[rank0]:/opt/pytorch/pytorch/aten/src/ATen/native/cuda/GroupMMCommon.cuh:64: prepare_grouped_gemm_data: block: [0,0,0], thread: [2,0,0] Assertion `delta % align == 0 && "expected input tensor dynamic dimension byte size to be non-negative multiple of 16\n"` failed.                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]: Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
[rank0]:[rank0]:   File "<frozen runpy>", line 198, in _run_module_as_main                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
[rank0]:[rank0]:   File "<frozen runpy>", line 88, in _run_code                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
[rank0]:[rank0]:   File "/workspace/torchtitan/torchtitan/train.py", line 79, in <module>                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:     main()                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
[rank0]:[rank0]:   File "/workspace/torchtitan/torchtitan/train.py", line 65, in main                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
[rank0]:[rank0]:     trainer.train()                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 367, in wrapper                                                                                                                                                                                                                                                                                                                                                                                                                              
[rank0]:[rank0]:     return f(*args, **kwargs)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
[rank0]:[rank0]:   File "/workspace/torchtitan/torchtitan/trainer.py", line 829, in train                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:     self.train_step(data_iterator)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
[rank0]:[rank0]:   File "/workspace/torchtitan/torchtitan/trainer.py", line 732, in train_step                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
[rank0]:[rank0]:     loss = self.forward_backward_step(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
[rank0]:[rank0]:   File "/workspace/torchtitan/torchtitan/observability/structured_logger/structured_logging.py", line 436, in sync_wrapper                                                                                                                                                                                                                                                                                                                                                                                                                    
[rank0]:[rank0]:     return func(*args, **kwargs)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
[rank0]:[rank0]:   File "/workspace/torchtitan/torchtitan/trainer.py", line 688, in forward_backward_step                                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:     loss.backward()                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_tensor.py", line 633, in backward                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
[rank0]:[rank0]:     torch.autograd.backward(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/autograd/__init__.py", line 395, in backward                                                                                                                                                                                                                                                                                                                                                                                                                                                               
[rank0]:[rank0]:     _engine_run_backward(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/autograd/graph.py", line 913, in _engine_run_backward                                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass                                                                                                                                                                                                                                                                                                                                                                                                                                     
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                     
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/autograd/function.py", line 333, in apply_boxed                                                                                                                                                                                                                                                                                                                                                                                                                                                            
[rank0]:[rank0]:     return self._get_user_fn()(self, *args)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_functorch/_aot_autograd/runtime_wrappers.py", line 3402, in backward                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:     return CompiledFunction._bwd_fn(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_functorch/_aot_autograd/subclass_codegen.py:codegen(compiled_function_backward)", line 23, in _compiled_backward                                                                                                                                                                                                                                                                                                                                                                                          
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_functorch/_aot_autograd/subclass_codegen.py:codegen(compiled_function_backward)", line 14, in impl_fn                                                                                                                                                                                                                                                                                                                                                                                                     
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_functorch/_aot_autograd/runtime_wrappers.py", line 3472, in _backward_impl                                                                                                                                                                                                                                                                                                                                                                                                                                
[rank0]:[rank0]:     return call_func_at_runtime_with_args(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_functorch/_aot_autograd/utils.py", line 126, in call_func_at_runtime_with_args                                                                                                                                                                                                                                                                                                                                                                                                                            
[rank0]:[rank0]:     out = normalize_as_list(f(args))                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
[rank0]:[rank0]:                             ^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_dynamo/eval_frame.py", line 1382, in _fn                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
[rank0]:[rank0]:     return fn(*args, **kwargs)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
[rank0]:[rank0]:   File "/opt/pytorch/pytorch/torch/_inductor/output_code.py", line 753, in __call__                                                                                                                                                                                                                                                                                                                                                                                                                                                           
[rank0]:[rank0]:     return self.current_callable(inputs)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
[rank0]:[rank0]:   File "/tmp/torchinductor_root/tz/ctz7rncet4c3rb6t7rkxub4qo6akzvp3y3szkrwoj2vvlc5zoyee.py", line 2206, in call                                                                                                                                                                                                                                                                                                                                                                                                                               
[rank0]:[rank0]:     buf22 = extern_kernels._grouped_mm(buf21, getitem_17, buf13, out_dtype=None)                                                                                                                                                                                                                                                                                                                                                                                                                                                              
[rank0]:[rank0]:             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                                                                              
[rank0]:[rank0]: RuntimeError: cutlass cannot initialize                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
[rank0]:CUDA error encountered at: file=/workspace/DeepEP/csrc/hybrid_ep/hybrid_ep.cu, line=81, call='cudaDeviceSynchronize()', Reason=cudaErrorAssert:device-side assert triggeredFatal Python error: Aborted                                                                                                                                                                                                                                                                                                                                                 
[rank0]:                                                                                                                                                                                                               
```

tianyu-l: Does this error only show up when compile is enabled? If so probably should ask @xmfan

syed-ahmed: Yes, only when compile is enabled with --compile.enable --compile.components model,loss

torchtitan/distributed/deepep/hybridep.py
```diff
@@ -156,13 +161,17 @@ def _dispatch_impl(
...
     """
-    global _buffer
-    if _buffer is None:
-        raise RuntimeError(
-            "HybridEP buffer not initialized. Call dispatch_tokens() first."
-        )
+    num_local_experts = num_experts // ep_size
+
+    # pyrefly: ignore [implicit-import, bad-argument-type]
+    group = torch._C._distributed_c10d._resolve_process_group(group_name)
```
shuhuayu: Any reason to prefer the c++ binding here? a previous similar usage (https://github.com/pytorch/torchtitan/blob/cf3c4312e59988fb00d1720c53a95143379dcc06/torchtitan/experiments/graph_trainer/fsdp_passes.py#L78) used a pure python wrapper. We can remove the implicit import pyrefly error suppression if using a python wrapper.

torchtitan/models/common/moe.py
shuhuayu: Is this a pytorch bug to be upstreamed? If this error only happens on `sm_100`, should we make these operations (memory heavy) conditional and leave a TODO until a fix is in pytorch?

xmfan: Does this happen only with HybridEP? If the code path before runs fine with compile, it suggests the HybridEP path has different metadata of inputs to the grouped mm

syed-ahmed: Yes, only happens with HybridEP (verified by running `NGPU=4 MODULE=deepseek_v3 CONFIG=deepseek_v3_debugmodel ./run_train.sh --parallelism.expert_parallel_degree=4 --compile.enable --compile.components model,loss`). Thanks for this pointer! Let me investigate a bit more.

syed-ahmed: Ok @xmfan is right I believe. The fake backward for dispatch and combine in hybridep uses new_zeros:
https://github.com/pytorch/torchtitan/blob/52a292d2977690d407bd81781de932f7f7dc56c5/torchtitan/distributed/deepep/hybridep.py#L284

https://github.com/pytorch/torchtitan/blob/52a292d2977690d407bd81781de932f7f7dc56c5/torchtitan/distributed/deepep/hybridep.py#L327-L328

Inductor sees these as plain allocations it fully controls, so when the downstream graph does grad * scores -> transpose -> _grouped_mm, Inductor fuses the chain and materializes the result as a row-major (hidden_dim, tokens) buffer, triggering the CUTLASS TMA alignment assertion on sm_100 because the 2D/2D _grouped_mm backward requires per-expert token counts to be 8-aligned when A is row-major.

The standard EP path doesn't hit this because its combine backward uses native PyTorch ops (scatter_add, all_to_all) - Inductor traces through them fully and reuses existing buffers via reinterpret_tensor (column-major view), never allocating a fresh row-major transposed buffer.

I have tried replacing the new_zeros calls in the backward tracing path with dedicated custom ops (hybridep::combine_bwd, hybridep::dispatch_bwd) that have register_fake implementations. Inductor cannot optimize through custom op boundaries, so it must accept the output in standard (tokens, hidden_dim) contiguous layout and use reinterpret_tensor for the subsequent transpose - producing column-major A, which bypasses the alignment check entirely. Do you think that's the correct approach?

torchtitan/components/quantization/mx.py
tianyu-l: should we delete this?

syed-ahmed: Yes.

tests/integration_tests/models.py
syed-ahmed: Addressed. Only have a single test now.

torchtitan/models/common/moe.py
syed-ahmed: Ok I consulted:
- https://docs.google.com/document/d/1_W62p8WJOQQUzPsJYa7s701JXt0qf2OfLub2sbkHOaU/edit?tab=t.0#heading=h.ptttacy8y1u9
- https://docs.pytorch.org/docs/2.12/library.html#torch.library.custom_op

and looks like hybridep should have declared the backward formulas for combine and dispatch to be custom ops per this comment from the manual: "If your operator's backward pass is a custom kernel, then it should be invoked through a custom operator.". By doing that, we don't need to have workarounds like checking if the untraceable dispatch handle is none (`if dispatch_handle is None or dispatch_handle.value is None:`) and based on that use `new_zeros` to have fake implementation. Now the backwards have their own custom ops, and inductor doesn't try to trace into it and cause the alignment problem.

torchtitan/distributed/deepep/hybridep.py
syed-ahmed: Thanks! Addressed.

syed-ahmed:
> Is this a pytorch bug to be upstreamed? If this error only happens on sm_100, should we make these operations (memory heavy) conditional and leave a TODO until a fix is in pytorch?

@shuhuayu addressed the comment. See thread above.

torchtitan/components/quantization/mx.py
```diff
@@ -155,6 +152,11 @@ class Config(QuantizationConverter.Config):
 
         - mxfp8_rceil: MXFP8 dynamic quantization with RCEIL rounding mode when computing the e8m0 scale factors.
         """
+        pad_multiple: int = 32
+        """
+        Pad per-expert token groups to this multiple for MXFP8 grouped GEMM alignment.
+        The CuTeDSL quantization kernel on sm_100 requires multiples of 128.
+        """
```
syed-ahmed, line 159: Note this change is needed to run with MXFP8 on GB200. Tested using the following config:
```
def deepseek_v3_671b() -> Trainer.Config:
    compile_config = CompileConfig(enable=True, components=["loss"])
    model_compile_enabled = (
        compile_config.enable and "model" in compile_config.components
    )
    return Trainer.Config(
        loss=ChunkedCELoss.Config(),
        hf_assets_path="./assets/hf/DeepSeek-V3.1-Base",
        model_spec=model_registry(
            "671B",
            attn_backend="flex",
            converters=[
                MXFP8LinearConverter.Config(
                    recipe_name="mxfp8_rceil",
                    fqns=[
                        "attention.wq",
                        "attention.wo",
                        "feed_forward",
                        "shared_experts",
                    ],
                    model_compile_enabled=model_compile_enabled,
                ),
                MXFP8GroupedExpertsConverter.Config(
                    recipe_name="mxfp8_rceil",
                    model_compile_enabled=model_compile_enabled,
                    pad_multiple=128,
                ),
            ],
            moe_comm_backend="hybridep",
            non_blocking_capacity_factor=0.0375
        ),
        dataloader=HuggingFaceTextDataLoader.Config(
            dataset="c4",
        ),
        optimizer=OptimizersContainer.Config(lr=2.2e-4),
        lr_scheduler=LRSchedulersContainer.Config(
            warmup_steps=2000,
            decay_ratio=0.8,
            decay_type="cosine",
            min_lr_factor=0.1,
        ),
        training=TrainingConfig(
            local_batch_size=4,
            seq_len=4096,
            steps=10000,
        ),
            pipeline_parallel_schedule="Interleaved1F1B",
            expert_parallel_degree=2,
        ),
        checkpoint=CheckpointManager.Config(interval=500),
        activation_checkpoint=ActivationCheckpointConfig(
            mode="selective",
        ),
        compile=compile_config,
    )
```

syed-ahmed:
DSV3 671B run (PP=1, EP=64, DP=256, BS=8, --compile.components model,loss, full AC) on 64x4 GB200:
```
--rdzv_conf read_timeout=3600 --local-ranks-filter 0 --role rank --tee 3 -m torchtitan.train --module torchtitan_configs.deepseek_v3 --config deepseek_v3_671b --profiler.no-enable-profiling --metrics.enable-tensorboard --metrics.save_for_all_ranks --training.steps 50 --training.local_batch_size 8 --training.seq_len 4096 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 1 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 64 --dataloader.dataset c4_test --debug.seed 42 --compile.enable --compile.components model,loss --activation_checkpoint.mode=full --parallelism.data_parallel_shard_degree=-1 --comm.init_timeout_seconds=3600 --comm.train_timeout_seconds=3600
[rank0]:[titan] 2026-05-21 13:41:40,784 - root - INFO - torchtitan version: 0.2.2+git28c986b (0.0.0 means __version__ is not defined correctly).
[rank0]:[titan] 2026-05-21 13:41:40,994 - torchtitan.observability.structured_logger.jsonl_handler - INFO - Structured logging -> JSONL: /workspace/torchtitan/outputs/structured_logs/training.global_rank_0.20260521-134140-81UE0N.jsonl
[rank0]:[titan] 2026-05-21 13:41:41,422 - root - WARNING - ENV[TORCH_NCCL_ASYNC_ERROR_HANDLING] = 1 will be overridden to 3 based on job config
[rank0]:[titan] 2026-05-21 13:41:41,424 - root - INFO - Building device mesh with parallelism: pp=1, dp_replicate=1, dp_shard=256, cp=1, tp=1, ep=64
[rank0]:[titan] 2026-05-21 13:41:41,469 - root - INFO - Successfully created meshes with active dimensions: ['batch', 'loss', 'ep', 'efsdp', 'fsdp']
[rank0]:[titan] 2026-05-21 13:41:41,470 - root - INFO - [GC] Initial GC collection took 0.00 seconds
[rank0]:[titan] 2026-05-21 13:41:41,472 - root - INFO - Loading tokenizer from tokenizer.json
[rank0]:[titan] 2026-05-21 13:41:41,651 - root - INFO - Preparing c4_test dataset from /jet_art_root/unverified/text/core-models/c4/en_train_3gb
[rank0]:[titan] 2026-05-21 13:42:24,485 - root - INFO - Building deepseek_v3 671B
[rank0]:[titan] 2026-05-21 13:42:24,689 - root - INFO - TensorBoard logging enabled. Logs will be saved at ./outputs/tb/20260521-1342/rank_0
[rank0]:[titan] 2026-05-21 13:42:24,689 - root - INFO - CUDA capacity: NVIDIA GB200 with 184.31GiB memory
[rank0]:[titan] 2026-05-21 13:42:24,825 - root - INFO - Total parameter count: dense 14,456,871,936, sparse 656,569,532,416, active 37,552,282,624
[rank0]:[titan] 2026-05-21 13:42:24,825 - root - INFO - Model deepseek_v3 671B size: 671,026,404,352 total parameters
[rank0]:[titan] 2026-05-21 13:42:24,825 - root - INFO - Compiling the loss function with torch.compile
[rank0]:[titan] 2026-05-21 13:42:24,900 - root - INFO - Applied full activation checkpointing to the model
[rank0]:[titan] 2026-05-21 13:42:24,908 - root - INFO - Compiling each TransformerBlock with torch.compile
[rank0]:[titan] 2026-05-21 13:42:25,485 - root - INFO - Applied fully_shard to the model
[rank0]:[titan] 2026-05-21 13:42:26,501 - root - INFO - Peak FLOPS used for computing MFU: 2.500e+15
[rank0]:[titan] 2026-05-21 13:42:26,502 - root - INFO - CUDA memory usage for model: 9.85GiB(5.35%)
[rank0]:[titan] 2026-05-21 13:42:26,505 - root - WARNING - Warmup steps (2000) exceed total steps (50). Adjusting warmup steps to 50.
[rank0]:[titan] 2026-05-21 13:42:26,505 - root - WARNING - Warmup (50) + decay (40) steps exceed total steps (50). Adjusting decay steps to 0.
[rank0]:[titan] 2026-05-21 13:42:26,506 - root - WARNING - model.safetensors.index.json not found at hf_assets_path: /jet_art_root/unverified/model/core-models/deepseek-v3.1-base_tokenizer/hf-d3d4eaf/model.safetensors.index.json.                     Defaulting to saving a single safetensors file if checkpoint is saved in HF format
[rank0]:[titan] 2026-05-21 13:42:26,553 - root - INFO - Trainer is initialized with local batch size 8, global batch size 2048, gradient accumulation steps 1, sequence length 4096, total steps 50 (warmup 2000)
[rank0]:[titan] 2026-05-21 13:42:26,553 - root - INFO - Training starts at step 1
[rank0]:[titan] 2026-05-21 13:42:48,136 - root - INFO - Initializing HybridEP buffer: hidden_dim=7168, max_tokens_per_rank=32768, num_local_experts=4, ep_size=64
[rank0]:[titan] 2026-05-21 13:43:42,739 - root - INFO - step:  1  loss: 12.27126  grad_norm:  2.3152  memory: 87.19GiB(47.30%)  tps: 421  tflops: 118.24  mfu: 4.73%
[rank0]:[titan] 2026-05-21 13:43:42,740 - root - INFO - Synchronizing and adjusting timeout for all ProcessGroups to 1:00:00
[rank0]:[titan] 2026-05-21 13:46:27,278 - root - INFO - step: 10  loss:  9.35680  grad_norm: 10.0919  memory: 108.15GiB(58.68%)  tps: 1,792  tflops: 503.94  mfu: 20.16%
[rank0]:[titan] 2026-05-21 13:50:01,018 - root - INFO - step: 20  loss:  7.53822  grad_norm:  5.1168  memory: 108.15GiB(58.68%)  tps: 1,533  tflops: 431.04  mfu: 17.24%
[rank0]:[titan] 2026-05-21 13:53:45,236 - root - INFO - step: 30  loss:  7.00716  grad_norm:  3.7247  memory: 108.15GiB(58.68%)  tps: 1,461  tflops: 410.89  mfu: 16.44%
[rank0]:[titan] 2026-05-21 13:57:27,142 - root - INFO - step: 40  loss:  6.92056  grad_norm:  8.6678  memory: 108.15GiB(58.68%)  tps: 1,477  tflops: 415.18  mfu: 16.61%
[rank0]:[titan] 2026-05-21 14:00:47,618 - root - INFO - [GC] Performing periodic GC collection took 0.14 seconds
[rank0]:[titan] 2026-05-21 14:01:10,962 - root - INFO - step: 50  loss:  6.53257  grad_norm:  3.0405  memory: 108.15GiB(58.68%)  tps: 1,464  tflops: 411.63  mfu: 16.47%
[rank0]:[titan] 2026-05-21 14:01:10,963 - root - INFO - Sleeping 2 seconds for other ranks to complete
[rank0]:[titan] 2026-05-21 14:01:12,963 - root - INFO - Training completed
[rank0]:[titan] 2026-05-21 14:01:15,633 - root - INFO - Process group destroyed
```

torchtitan/distributed/deepep/hybridep.py
```diff
@@ -96,14 +97,29 @@ def _apply_scores(
...
     "hybridep::combine",
     f"(Tensor x, {_handle_type} handle, int num_tokens, int? pad_multiple) -> Tensor",
 )
 
+# Backward ops: these wrap the communication calls so AOT Autograd can trace them
+# dispatch_bwd: used in combine's backward (scatters gradients via dispatch_with_permute)
+torch.library.define(
+    "hybridep::dispatch_bwd",
+    f"(Tensor grad_combined, {_handle_type} handle, int num_permuted_tokens, int? pad_multiple) -> Tensor",
+)
```
xmfan: any reason to use torch.library.define instead of torch.library.custom_op?

xmfan (approved):
looks good, please use torch.library.custom_op if possible: https://docs.pytorch.org/docs/2.12/library.html#torch.library.custom_op

tests/integration_tests/h100.py
```diff
@@ -72,5 +75,20 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
...
+                    "--parallelism.data_parallel_shard_degree 4",
+                    "--parallelism.expert_parallel_degree 2",
+                    "--compile.enable",
+                    "--compile.components model,loss",
+                ],
+            ],
+            "DeepSeek V3 FSDP+HybridEP+compile",
+            "deepseek_v3_fsdp+hybridep+compile",
+            ngpu=4,
+            disabled=not _DEEPEP_AVAILABLE,
```
tianyu-l: This may hide deepep install issue on H100? Can we use https://github.com/pytorch/torchtitan/blob/main/tests/integration_tests/__init__.py#L26

tianyu-l: also, is deepep only available on H?

torchtitan/components/quantization/mx.py
tianyu-l: - compared with making it configurable, I wonder if we should detect hardware capability and make it automatically adapting
- at least we should error out when incompatible, similar to https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/config_utils.py#L48

tests/integration_tests/h100.py
syed-ahmed: I followed the pattern from graph_trainer: https://github.com/pytorch/torchtitan/pull/3390 for this. It looks like https://github.com/pytorch/pytorch/pull/180786 was responsible for the breakage and not DeepEP. Is it okay for the CI to break if we ever run into this scenario again or until the DeepEP packaging is figured out?

syed-ahmed:
CI failures are now just the `RuntimeError: "histogram_cpu" not implemented for 'torch.int64'` failures, which are not related to this PR (possibly has been fixed by https://github.com/pytorch/pytorch/pull/183970 ?)

tests/integration_tests/h100.py
tianyu-l: I would prefer the error is exposed by CI rather than silently skipped.

tianyu-l:
> CI failures are now just the RuntimeError: "histogram_cpu" not implemented for 'torch.int64' failures, which are not related to this PR (possibly has been fixed by https://github.com/pytorch/pytorch/pull/183970 ?)

@acisseJZhong it seems the PR is merged 3 days ago. Is pytorch nightly lagged behind?

acisseJZhong:
For cuda test: titan CI runs on x86_64 Linux machines, and the Linux x86_64 builds stopped after dev20260521: https://download.pytorch.org/whl/nightly/torch/, search for torch-2.13.0.dev20260521+cu130-cp310-cp310-manylinux_2_28_x86_64.whl. 

For rocm test Claude reasons: 
```
ROCm has x86_64 wheels all the way through dev20260523. But the
  CI log showed it still installed dev20260521:

  It downloaded dev20260523 first, then fell back to dev20260521.
  This is likely a dependency conflict — torchvision or another
  package pinned a max torch version, so pip's resolver rejected
  dev20260523 and dev20260522 and settled on dev20260521.

The latest torchvision for cu130 is
  dev20260522, which pins torch==2.13.0.dev20260521. 

looks like the whole chain is bottlenecked: the CUDA x86_64 nightly
  build infra broke after May 21, which means no new torch or
  torchvision wheels are being published for that platform.
  Everything is stuck at dev20260521.
```

anyone we could ask to unblock this?

syed-ahmed:
@atalman @nWEIdia @tinglvv Would you know why there are no manylinux wheels after dev20260521?

nWEIdia:
> @atalman @nWEIdia @tinglvv Would you know why there are no manylinux wheels after dev20260521?

Looks like the nightly build jobs exceeded 4 hours and timed out. 
e.g. https://github.com/pytorch/pytorch/actions/runs/26277452493/job/77411806220

tests/integration_tests/h100.py
syed-ahmed: > also, is deepep only available on H?

Yes, looks like it's Hopper and above only: https://github.com/deepseek-ai/DeepEP#requirements

syed-ahmed:
> - compared with making it configurable, I wonder if we should detect hardware capability and make it automatically adapting
> - at least we should error out when incompatible, similar to https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/config_utils.py#L48

@tianyu-l This is a tile size requirement set by the kernel that TorchAO selects: 

https://github.com/pytorch/ao/blob/c81eb798a5256f8807690a05cdca6843a22f6527/torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_2d_32x1.py#L1011-L1012

I'm not sure if there is a clean way to detect this. Handling this in TorchTitan might add maintenance burden if TorchAO updates the kernel in the future, and changes the block size. I think keeping it just configurable is the most flexible here, until there is a better way to detect what TorchAO is using.

tests/integration_tests/h100.py
syed-ahmed: Ok, the tests will now run and fail visibly if deep_ep is missing, without blocking other tests.

torchtitan/components/quantization/mx.py
syed-ahmed: @tianyu-l This is a tile size requirement set by the kernel that TorchAO selects:

https://github.com/pytorch/ao/blob/c81eb798a5256f8807690a05cdca6843a22f6527/torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_2d_32x1.py#L1011-L1012

I'm not sure if there is a clean way to detect this. Handling this in TorchTitan might add maintenance burden if TorchAO updates the kernel in the future, and changes the block size. I think keeping it just configurable is the most flexible here, until there is a better way to detect what TorchAO is using.

tests/integration_tests/h100.py
tianyu-l: could you help remove this outdated comment?

tianyu-l: similar

torchtitan/models/deepseek_v3/config_registry.py
```diff
@@ -63,6 +63,16 @@ def deepseek_v3_debugmodel_ep() -> Trainer.Config:
     return config
```
tianyu-l: it seems we no longer need this, could you help clean up?

tests/integration_tests/h100.py
```diff
@@ -72,5 +72,19 @@ def build_h100_tests_list() -> list[OverrideDefinitions]:
             "hsdp+cp+compile+float8",
             ngpu=8,
         ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module deepseek_v3 --config deepseek_v3_debugmodel_hybridep",
```
tianyu-l, line 85: oh I probably didn't understand why we are testing this on H100?
- Does it not work on A10 CI machines? If it works, why don't we test on them, as H is more limited resource.
- I thought the value of HybridEP is more for GB. Is testing on H100 a good proxy?
- Also I don't think we have a test for DeepEP on A/H CI. What would you suggest?

tinglvv:
> @atalman @nWEIdia @tinglvv Would you know why there are no manylinux wheels after dev20260521?

The nightly x86 cuda wheel builds have been broken after 5/22. Prepared (updated link with a new PR:https://github.com/pytorch/pytorch/pull/185560) to increase the default timeout for x86_64 nightly cuda wheel builds to restore the build health. Once this gets merged, we should have the manylinux wheels back again. 

Feel free to ping us (@nWEIdia @tinglvv) for CI/CD related issues in the future.

tests/integration_tests/h100.py
syed-ahmed: - I haven't tried DeepEP on ampere machines. From the readme and existing issues (https://github.com/deepseek-ai/DeepEP/issues/481), it doesn't look like it's actively maintained or there are environment variable workarounds to disable h100 features to make it work. I think it might be easier to just follow the minimum requirement of hopper for this feature.
- Since we don't have GB in the CI _yet_, I think testing on H100 is a good proxy to maintain the integration. 
- TBH, I think we should deprecate/cleanup deepep backend since it doesn't compose with CUDAGraph or figure out how to upgrade to the new DeepEP V2 backend (and see if it composes with cudagraphs? Not sure if it does.).

torchtitan/models/deepseek_v3/config_registry.py
syed-ahmed: Seems like this config is used by graph_trainer and other tests: https://github.com/search?q=repo%3Apytorch%2Ftorchtitan%20deepseek_v3_debugmodel_ep&type=code exercising the standard ep path.

tests/integration_tests/h100.py
syed-ahmed: Addressed.

syed-ahmed: Addressed.

syed-ahmed:
@tianyu-l summary so far:
- deepep test failures should now be exposed in upstream CI without blocking other tests.
- the PAD_MULTIPLE factor should just remain configurable since it's a torchao specific variable and there isn't a good way to handle it without adding maintenance burden.
- the histogram_cpu failures should be resolved in torchtitan when @tinglvv's https://github.com/pytorch/pytorch/pull/185289 lands in pytorch/pytorch and nightly wheel is available.
- async TP tests are now enabled in CI.

tests/integration_tests/h100.py
tianyu-l: Does HybridEP completely cover DeepEP on H100, in terms of features and efficiency?

torchtitan/models/deepseek_v3/config_registry.py
tianyu-l: Maybe I missed something, but it looks identical to `deepseek_v3_debugmodel`. Can we do a global switch?

tianyu-l (approved):

syed-ahmed:
Cleaned up deepseek_v3_debugmodel_ep config globally. Will merge after verifying CI status tomorrow.

syed-ahmed:
Verified that https://github.com/pytorch/pytorch/pull/185560 is pulling latest nightly pytorch and existing 8 GPU model tests are passing. Merging based on this signal.

PR #3362, opened 2026-05-15
[graph_trainer] Support marked symbolic input dims in tracing

sanketpurandare:
Stacked PRs:
 * #3328
 * #3325
 * #3363
 * __->__#3362

--- --- ---

### [graph_trainer] Support marked symbolic input dims in tracing

Add the graph_trainer tracing prerequisites needed by later EP chunking work without introducing the chunking passes themselves. minimal_fx_tracer now accepts mark_unbacked and mark_dynamic metadata on plain tensor inputs, preserves Dynamo dynamic-range annotations during fakeification, and keeps wrapper-subclass inputs rejected so DTensor-style layouts do not silently lose their metadata.

Make RoPE singleton-position handling symbolic-shape friendly without adding tracer-specific runtime checks, make no-op cleanup passes guard symbolic comparisons, and materialize traced gradients to the parameter layout before assigning param.grad.

Move AutoParallel CI checks after the graph_trainer checks in the default and H100 workflows so a stale AutoParallel package cannot hide the graph_trainer signal from this stack.

Test Plan:\n- conda run -n pt-dev pytest -q torchtitan/experiments/graph_trainer/tests/test_trace_module.py -k 'fakeify_input_copies_only_shape_annotations or mark_dynamic_min_max_preserves_range or mark_dynamic_batch_and_seq_dims_with_rope or mark_unbacked_positions_batch_dim_with_rope or rope_symbolic_positions_compile_fullgraph or materialize_grad_for_param_layout_restores_param_strides or nested_wrapper_subclass_marked_inner_rejected or mark_dynamic_wrapper_subclass_rejected or dtensor_mark_unbacked_rejected'\n- conda run -n pt-dev python -m pytest -q tests/unit_tests/test_loss.py::TestChunkedCELoss::test_equal_split_chunks_match_torch_chunk_for_supported_shapes tests/unit_tests/test_loss.py::TestChunkedCELoss::test_equal_split_chunks_backward_matches_torch_chunk tests/unit_tests/test_loss.py::TestChunkedCELoss::test_symbolic_seq_len_traces_chunking tests/unit_tests/test_loss.py::TestChunkedCELoss::test_rejects_non_divisible_sequence_length tests/unit_tests/test_loss.py::TestChunkedCELoss::test_single_token_chunks_match_torch_chunk_reference torchtitan/experiments/graph_trainer/tests/test_chunked_loss.py::TestChunkedCELossWithParamGrads::test_bitwise_equal_with_chunked_celoss

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -559,9 +559,18 @@ class ChunkedCELoss(BaseLoss):
         # local seq=0 and breaking GradAccumulator's slice writes.
         # ``local_map`` runs the chunking body on plain tensors; under the
         # non-DTensor (eager) path we call ``_chunk_local`` directly.
-        # ``.contiguous()`` breaks shared storage from ``torch.chunk``.
+        # Equal chunk sizes also match GradAccumulator's sequential slice
+        # writes, which use one chunk length for each write offset.
         def _chunk_local(t):
-            return tuple(c.contiguous() for c in torch.chunk(t, num_chunks, dim=1))
+            seq_len = t.shape[1]
+            torch._check(
+                seq_len % num_chunks == 0,
+                lambda: "ChunkedCELoss sequence length must be divisible by num_chunks",
+            )
+            chunk_len = seq_len // num_chunks
+            return tuple(
+                c.contiguous() for c in torch.split(t, [chunk_len] * num_chunks, dim=1)
+            )
 
         def _chunk(t):
             if not isinstance(t, DTensor):
--- a/torchtitan/experiments/graph_trainer/dynamic_shapes.py
+++ b/torchtitan/experiments/graph_trainer/dynamic_shapes.py
@@ -7,10 +7,21 @@
 """Dynamic-shape helpers used by minimal_fx_tracer.
 
 Covers:
-- detection of mark_unbacked() metadata on plain and wrapper-subclass tensors,
+- detection of mark_unbacked() and mark_dynamic() metadata on plain and
+  wrapper-subclass tensors,
 - input fakeification with a StatelessSymbolicContext built from that metadata,
 - post-trace materialization of deferred ShapeEnv runtime asserts and
   symbolic-input guards.
+
+Contract:
+- Users express input dynamic shapes with PyTorch's Dynamo tensor-marking APIs:
+  ``torch._dynamo.mark_dynamic`` and
+  ``torch._dynamo.decorators.mark_unbacked``.
+- Marked dims are supported only on plain tensor inputs. Wrapper subclasses are
+  rejected before unwrapping so markings cannot be silently dropped or copied to
+  the wrong inner tensor.
+- This module is the only place where minimal_fx_tracer interprets or copies
+  raw ``_dynamo_*`` tensor annotations.
 """
 
 from typing import Any
@@ -20,17 +31,50 @@ from torch._subclasses import FakeTensorMode
 from torch.utils._python_dispatch import is_traceable_wrapper_subclass
 
 
+_DYNAMO_SHAPE_ANNOTATION_NAMES = (
+    "_dynamo_unbacked_indices",
+    "_dynamo_strict_unbacked_indices",
+    "_dynamo_dynamic_indices",
+    "_dynamo_dynamic_range",
+    "_dynamo_shape_ids",
+    "_dynamo_unbacked_bounds",
+    "_dynamo_hint_overrides",
+    "_specialize_on",
+    "_has_dynamo_dim_marking",
+)
+"""Dynamo tensor attributes that describe user-requested symbolic shapes.
+
+This is an explicit allowlist, not a general tensor-attribute copy. The tracer
+reads a subset of these attrs to build a SymbolicContext, then copies the full
+allowlist onto the fake tensor so later Dynamo/FakeTensor consumers see the
+same shape annotations they would have seen on the original input.
+"""
+
+
 def _tensor_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+    """Return whether ``tensor`` itself carries mark_unbacked() metadata."""
     return bool(
         getattr(tensor, "_dynamo_unbacked_indices", None)
         or getattr(tensor, "_dynamo_strict_unbacked_indices", None)
     )
 
 
-def _wrapper_subclass_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+def _tensor_has_mark_dynamic(tensor: torch.Tensor) -> bool:
+    """Return whether ``tensor`` itself carries mark_dynamic() metadata."""
+    return bool(getattr(tensor, "_dynamo_dynamic_indices", None))
+
+
+def _wrapper_subclass_has_marked_dynamic_dims(tensor: torch.Tensor) -> bool:
+    """Detect marked dims anywhere inside a traceable wrapper subclass.
+
+    minimal_fx_tracer unwraps subclasses such as DTensor before fakeification
+    and rewraps them after execution. Marked dynamic dims on wrapper inputs are
+    therefore rejected rather than silently dropped or copied onto the wrong
+    plain inner tensor.
+    """
     if not is_traceable_wrapper_subclass(tensor):
         return False
-    if _tensor_has_mark_unbacked(tensor):
+    if _tensor_has_mark_unbacked(tensor) or _tensor_has_mark_dynamic(tensor):
         return True
 
     attrs, _ = tensor.__tensor_flatten__()
@@ -38,32 +82,54 @@ def _wrapper_subclass_has_mark_unbacked(tensor: torch.Tensor) -> bool:
         inner_value = getattr(tensor, attr)
         if not isinstance(inner_value, torch.Tensor):
             continue
-        if _tensor_has_mark_unbacked(
-            inner_value
-        ) or _wrapper_subclass_has_mark_unbacked(inner_value):
+        if (
+            _tensor_has_mark_unbacked(inner_value)
+            or _tensor_has_mark_dynamic(inner_value)
+            or _wrapper_subclass_has_marked_dynamic_dims(inner_value)
+        ):
             return True
     return False
 
 
 def _symbolic_context_for_marked_dims(tensor: torch.Tensor) -> Any | None:
+    """Convert Dynamo shape annotations into FakeTensor symbolic context.
+
+    This function semantically interprets the original tensor's annotations:
+    which dims are unbacked, dynamic, range-constrained, or specialization
+    candidates. It deliberately uses direct ``getattr`` reads instead of the
+    copy allowlist because each attribute affects the SymbolicContext in a
+    different way.
+    """
     from torch.fx.experimental.symbolic_shapes import (
         DimDynamic,
         RelaxedUnspecConstraint,
         StatelessSymbolicContext,
+        StrictMinMaxConstraint,
     )
+    from torch.utils._sympy.numbers import int_oo
+    from torch.utils._sympy.value_ranges import ValueRanges
 
     marked_unbacked_indices = getattr(tensor, "_dynamo_unbacked_indices", set())
     marked_strict_unbacked_indices = getattr(
         tensor, "_dynamo_strict_unbacked_indices", set()
     )
+    marked_dynamic_indices = getattr(tensor, "_dynamo_dynamic_indices", set())
+    # mark_dynamic() records ranges separately from the marked dim set. The
+    # dim set says a dimension is dynamic; the range refines its constraints.
+    dynamic_ranges = {
+        dim_range.dim: dim_range
+        for dim_range in getattr(tensor, "_dynamo_dynamic_range", set())
+    }
     has_outer_marked_dims = bool(
-        marked_unbacked_indices or marked_strict_unbacked_indices
+        marked_unbacked_indices
+        or marked_strict_unbacked_indices
+        or marked_dynamic_indices
     )
     if not has_outer_marked_dims:
         return None
 
     dynamic_sizes = [DimDynamic.STATIC] * tensor.dim()
-    constraint_sizes = [None] * tensor.dim()
+    constraint_sizes: list[Any] = [None] * tensor.dim()
     specialize_on = [
         list(getattr(tensor, "_specialize_on", {}).get(i, []))
         for i in range(tensor.dim())
@@ -75,6 +141,22 @@ def _symbolic_context_for_marked_dims(tensor: torch.Tensor) -> Any | None:
         elif dim in marked_strict_unbacked_indices:
             dynamic_sizes[dim] = DimDynamic.UNBACKED
             constraint_sizes[dim] = RelaxedUnspecConstraint(warn_only=False)
+        elif dim in marked_dynamic_indices:
+            dynamic_sizes[dim] = DimDynamic.DYNAMIC
+            dim_range = dynamic_ranges.get(dim)
+            if dim_range is None or (dim_range.min is None and dim_range.max is None):
+                constraint_sizes[dim] = RelaxedUnspecConstraint(warn_only=False)
+            else:
+                # mark_dynamic() stores one-sided ranges with None, while
+                # ValueRanges requires concrete sentinels. ShapeEnv still
+                # intersects this with its normal dynamic-size invariant later.
+                constraint_sizes[dim] = StrictMinMaxConstraint(
+                    vr=ValueRanges(
+                        lower=0 if dim_range.min is None else dim_range.min,
+                        upper=int_oo if dim_range.max is None else dim_range.max,
+                    ),
+                    warn_only=False,
+                )
 
     return StatelessSymbolicContext(
         dynamic_sizes=dynamic_sizes,
@@ -88,17 +170,35 @@ def _symbolic_context_for_marked_dims(tensor: torch.Tensor) -> Any | None:
 def _fakeify_input(
     fake_mode: FakeTensorMode, arg: torch.Tensor, *, input_name: str
 ) -> torch.Tensor:
+    """Fakeify one tracer input while preserving user shape annotations.
+
+    ``FakeTensorMode.from_tensor`` consumes the SymbolicContext but does not
+    generally preserve Python attributes from ``arg``. After fakeification, copy
+    only Dynamo's shape-annotation allowlist so later graph passes and Dynamo
+    helpers can still inspect the fake tensor consistently.
+    """
     from torch._dynamo.source import LocalSource
+
+    def copy_tensor_annotations(fake_arg: torch.Tensor) -> torch.Tensor:
+        for name in _DYNAMO_SHAPE_ANNOTATION_NAMES:
+            if hasattr(arg, name):
+                setattr(fake_arg, name, getattr(arg, name))
+        return fake_arg
 
     symbolic_context = _symbolic_context_for_marked_dims(arg)
     if symbolic_context is None:
-        return fake_mode.from_tensor(arg, static_shapes=True)
+        return copy_tensor_annotations(fake_mode.from_tensor(arg, static_shapes=True))
 
     source = LocalSource(input_name, is_input=True)
-    return fake_mode.from_tensor(
-        arg,
-        source=source,
-        symbolic_context=symbolic_context,
+    # Do not wrap this in ShapeEnv.ignore_fresh_unbacked_symbols(): the
+    # StatelessSymbolicContext binds marked input dims to placeholders, so they
+    # are not unowned fresh symbols produced by tracing graph operations.
+    return copy_tensor_annotations(
+        fake_mode.from_tensor(
+            arg,
+            source=source,
+            symbolic_context=symbolic_context,
+        )
     )
 
 
--- a/torchtitan/experiments/graph_trainer/make_fx_tracer.py
+++ b/torchtitan/experiments/graph_trainer/make_fx_tracer.py
@@ -7,7 +7,7 @@
 import contextlib
 import copy
 from collections.abc import Callable, Generator
-from contextlib import contextmanager, nullcontext
+from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import Any
 
@@ -24,8 +24,7 @@ from torch.utils._python_dispatch import is_traceable_wrapper_subclass
 from torchtitan.experiments.graph_trainer.dynamic_shapes import (
     _fakeify_input,
     _insert_runtime_asserts as _insert_runtime_asserts_pass,
-    _tensor_has_mark_unbacked,
-    _wrapper_subclass_has_mark_unbacked,
+    _wrapper_subclass_has_marked_dynamic_dims,
 )
 
 # Tensors and make_fx-safe primitives are allowed as pytree leaves in args.
@@ -410,40 +409,23 @@ def minimal_fx_tracer(
         for arg in full_args:
             if not isinstance(arg, torch.Tensor):
                 continue
-            if getattr(arg, "_dynamo_dynamic_indices", None) or getattr(
-                arg, "_dynamo_dynamic_range", None
-            ):
-                raise ValueError("minimal_fx_tracer only supports mark_unbacked()")
-            if _wrapper_subclass_has_mark_unbacked(arg):
+            if _wrapper_subclass_has_marked_dynamic_dims(arg):
                 raise ValueError(
-                    "minimal_fx_tracer only supports mark_unbacked() on plain tensor "
-                    "inputs; wrapper subclasses such as DTensor are not supported"
+                    "minimal_fx_tracer only supports marked dynamic dims on plain "
+                    "tensor inputs; wrapper subclasses such as DTensor are not "
+                    "supported"
                 )
         unwrapped_args, input_layouts = _unwrap_subclasses(full_args)
-        has_mark_unbacked = any(
-            isinstance(a, torch.Tensor) and _tensor_has_mark_unbacked(a)
-            for a in unwrapped_args
-        )
-
         fake_mode = FakeTensorMode(
             allow_non_fake_inputs=True,
             shape_env=torch.fx.experimental.symbolic_shapes.ShapeEnv(),
         )
-        # Fresh unbacked symbols allocated when fakeifying mark_unbacked() inputs
-        # are wired to placeholders via the symbolic context, so they must not
-        # land in the ShapeEnv's pending_fresh_unbacked_symbols list.
-        ignore_ctx = (
-            fake_mode.shape_env.ignore_fresh_unbacked_symbols()
-            if has_mark_unbacked
-            else nullcontext()
+        fake_args = tuple(
+            _fakeify_input(fake_mode, a, input_name=f"input_{i}")
+            if isinstance(a, torch.Tensor)
+            else a
+            for i, a in enumerate(unwrapped_args)
         )
-        with ignore_ctx:
-            fake_args = tuple(
-                _fakeify_input(fake_mode, a, input_name=f"input_{i}")
-                if isinstance(a, torch.Tensor)
-                else a
-                for i, a in enumerate(unwrapped_args)
-            )
 
         output_layouts: dict[int, SubclassLayout] = {}
         num_flat_outputs: int = 0
--- a/torchtitan/experiments/graph_trainer/remove_noop_passes.py
+++ b/torchtitan/experiments/graph_trainer/remove_noop_passes.py
@@ -20,6 +20,7 @@ single pass-list entry; the sub-passes remain public so they can be tested
 import sys
 
 import torch
+from torch.fx.experimental.symbolic_shapes import guard_or_false
 
 from torchtitan.tools.logging import logger
 
@@ -138,6 +139,12 @@ _IDENTITY_VIEW_TARGETS = {
 }
 
 
+def _same_shape(shape1: torch.Size, shape2: torch.Size) -> bool:
+    return len(shape1) == len(shape2) and all(
+        guard_or_false(dim1 == dim2) for dim1, dim2 in zip(shape1, shape2)
+    )
+
+
 def remove_identity_view_pass(
     gm: torch.fx.GraphModule, example_inputs=None
 ) -> torch.fx.GraphModule:
@@ -170,7 +177,7 @@ def remove_identity_view_pass(
         ):
             continue
 
-        if inp_val.shape == out_val.shape:
+        if _same_shape(inp_val.shape, out_val.shape):
             node.replace_all_uses_with(inp)
             gm.graph.erase_node(node)
             count += 1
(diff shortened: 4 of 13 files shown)


torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -27,20 +27,26 @@ def _tensor_has_mark_unbacked(tensor: torch.Tensor) -> bool:
     )
 
 
-def _wrapper_subclass_has_mark_unbacked(tensor: torch.Tensor) -> bool:
+def _tensor_has_mark_dynamic(tensor: torch.Tensor) -> bool:
```
aditvenk, line 62: Should this also consider _dynamo_dynamic_range ? Previously, minimal tracer seems to have considered this too as "mark_dynamic"

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -75,6 +84,8 @@ def _symbolic_context_for_marked_dims(tensor: torch.Tensor) -> Any | None:
         elif dim in marked_strict_unbacked_indices:
             dynamic_sizes[dim] = DimDynamic.UNBACKED
             constraint_sizes[dim] = RelaxedUnspecConstraint(warn_only=False)
+        elif dim in marked_dynamic_indices:
+            dynamic_sizes[dim] = DimDynamic.DYNAMIC
```
aditvenk, line 145: Should we update constraint_sizes if dynamic has max/min hints?

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -117,6 +117,42 @@ def forward(self, x):
...
+        rope_cache = torch.randn(seq * 2, head_dim * 2)
+
+        for tensor in (x, xq, xk, positions):
+            mark_dynamic(tensor, 0)
+            mark_dynamic(tensor, 1)
+
+        traced = minimal_fx_tracer(forward)(
+            x, xq, xk, freqs_cis, rope_cache, positions
+        )
+        self.assertTrue(
```
aditvenk, line 333: Test with a shape different than what was traced?

torchtitan/models/common/rope.py
```diff
@@ -21,6 +21,16 @@
 ]
 
 
+def _check_shape_equal(actual, expected, context: str) -> None:
```
aditvenk: Can we restrict this change to graph_trainer only in some way? It seems necessary only for the minimal fx tracer?

torchtitan/experiments/graph_trainer/dynamic_shapes.py
sanketpurandare: Added support for both `mark_dynamic` and `mark_unbacked` metadata paths in the minimal tracer. The current helper imports the Dynamo dynamic/unbacked annotations into the `StatelessSymbolicContext`, including the unbacked bounds used by hinted symbolic dimensions.

torchtitan/experiments/graph_trainer/dynamic_shapes.py
sanketpurandare: The implementation keeps strict unbacked dimensions constrained with `RelaxedUnspecConstraint` and passes the Dynamo-provided unbacked bounds into the symbolic context. The min/max hint path is preserved through `unbacked_bounds` rather than specializing the traced graph to concrete sizes.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
sanketpurandare: Added tracing coverage for marked dynamic inputs and runtime assertion materialization. The traced graph keeps the input dimension symbolic and emits ShapeEnv runtime checks, so the test now verifies behavior beyond the exact concrete size used during capture.

torchtitan/models/common/rope.py
sanketpurandare: The RoPE change is limited to replacing Python shape equality assertions with `torch._check`, which is the PyTorch-native way to let symbolic sizes flow without weakening eager/runtime validation.

torchtitan/experiments/graph_trainer/dynamic_shapes.py
aditvenk: My question was pertaining to the backed case (elif dim in marked_dynamic_indices)

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -90,15 +101,33 @@ def _fakeify_input(
 ) -> torch.Tensor:
     from torch._dynamo.source import LocalSource
 
+    tensor_annotation_names = (
```
aditvenk: How is this list generated? How do we know if it is exhaustive?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -155,74 +156,77 @@ def _wrap_subclasses(
...
-    chain that partially feeds real computation keeps its live portion.
-    Restricting removal to ``empty_strided`` descendants keeps this narrower
-    than a whole-graph ``eliminate_dead_code``.
-
-    TODO: this is a workaround for the dead shadow nodes described above. The
-    upstream fix wraps shard propagation's fake-tensor metadata inference in
-    ``disable_proxy_modes_tracing`` so it never enters the make_fx graph (see
-    ``ShardingPropagator._propagate_tensor_meta_non_cached`` in PyTorch).
-    Remove this pass once that fix lands in the graph_trainer CI nightly.
+def remove_dead_dtensor_metadata_chains_pass(
```
aditvenk: Is this still required given @SherlockNoMad 's recent change to disable DTensor shard prop from tracing?

torchtitan/experiments/graph_trainer/selective_activation_remat.py
```diff
@@ -96,11 +96,6 @@ def selective_activation_remat_pass(
     if not regions:
         return gm
 
-    # Assumption: chunked-loss regions (e.g. lm_head) do not carry AC, so
```
aditvenk, line 112: why remove?

SherlockNoMad:
There are some coupling between tracer and EP_overlap passes... 

This needs to be called out in readme.md /  claude.md ... together with clear instruction on how to use ep overlap, and document it's composability and limitations.

torchtitan/distributed/activation_checkpoint.py
```diff
@@ -61,6 +61,10 @@ def _get_default_save_ops() -> set:
     comm_ops = [
         torch.ops._c10d_functional.reduce_scatter_tensor.default,
         torch.ops._c10d_functional.all_to_all_single.default,
+        # D2H scalar reads (token dispatcher split sizes). Each
+        # recomputation triggers a CUDA synchronize; saving the tiny
+        # CPU scalar avoids re-executing the counts all-to-all chain.
+        torch.ops.aten._local_scalar_dense.default,
```
aditvenk: If this is a perf optimization, can we defer to a separate PR?

torchtitan/models/common/rope.py
```diff
@@ -362,20 +372,32 @@ def _reshape_for_broadcast(
     cache_width = rope_cache.shape[-1]
     if positions is None:
         rope_cache = rope_cache[0:seqlen]
-        assert rope_cache.shape == (seqlen, cache_width)
+        _check_shape_equal(rope_cache.shape, (seqlen, cache_width), "rope_cache")
```
aditvenk: cc @tianyu-l -- could you help review the changes in models/common and provide code owner signoff?

aditvenk: question for @sanketpurandare  -- can these changes to models/common be separated to its own PR? I don't see dedicated testing in this PR that requires this change here, but maybe i am mistaken.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -251,6 +251,76 @@ class Config(BaseEPTokenDispatcher.Config):
...
+        num_tokens_per_expert_group = torch.ops._c10d_functional.wait_tensor(
+            num_tokens_per_expert_group
+        )
+        # Graph capture records .tolist() as split-size scalar reads. Keep
+        # the copy synchronous there so split sizes cannot race stale CPU data.
+        non_blocking = not (
+            torch.compiler.is_compiling() or torch.compiler._is_non_strict_tracing()
+        )
+        input_splits = num_tokens_per_expert.view(ep_size, -1).sum(dim=1)
+        input_splits = input_splits.to(torch.device("cpu"), non_blocking=non_blocking)
```
aditvenk: This is flipping to Blocking which is a change from prior. 

Given https://github.com/pytorch/pytorch/pull/185252, can't we keep this as non_blocking=True and have the compiler to insert the sync?

torchtitan/distributed/activation_checkpoint.py
sanketpurandare: It's more than a perf optimization because when we are using full inductor, local_scalar_dense is saved by default by the partitioner, this makes it consistent with regional inductor in aot_fx_trace mode, added the comment of why this is important for our use case. So it's a correctness + perf optimization. Further local_scalar_dense is a sym-int when tracing with mark_unbacked.

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
```diff
@@ -116,7 +119,166 @@ def forward(self, x):
...
+                marker(inner, 0)
+                wrapper = _TraceableWrapper(_TraceableWrapper(inner))
+
+                with self.assertRaisesRegex(
+                    ValueError,
+                    "only supports marked dynamic dims on plain tensor inputs",
+                ):
+                    minimal_fx_tracer(forward)(wrapper)
+
+    def test_mark_dynamic_batch_and_seq_dims_with_rope(self):
```
sanketpurandare, line 254: @aditvenk this is the test for rope.

torchtitan/models/common/rope.py
sanketpurandare: These changes do belong here since we need to support symbolic tracing for torchtitan models/common. If I move them into separate PR how do I justify those changes? See the tests above, I have tested it.

torchtitan/models/common/token_dispatcher.py
sanketpurandare: Yeah forgot to update this one.

torchtitan/models/common/rope.py
aditvenk: Ack, I missed it, sorry

torchtitan/experiments/graph_trainer/tests/test_trace_module.py
aditvenk: Thanks! sorry i missed it

torchtitan/experiments/graph_trainer/dynamic_shapes.py
sanketpurandare: Yes, I will update the constraint sizes for mark_dynamic case

torchtitan/experiments/graph_trainer/dynamic_shapes.py
sanketpurandare: Adding an inline comment explaining it.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
sanketpurandare: Removed it.

torchtitan/experiments/graph_trainer/selective_activation_remat.py
sanketpurandare: left as is

torchtitan/components/loss.py
```diff
@@ -516,6 +516,37 @@ def set_lm_head(self, lm_head: nn.Module) -> None:
...
+        torch._check(
+            seq_len % num_chunks == 0,
+            lambda: (
+                "ChunkedCELoss sequence length must be divisible by "
+                f"num_chunks={num_chunks}, got {seq_len}"
+            ),
+        )
+        chunk_len = seq_len // num_chunks
+        torch._check(
+            chunk_len > 1,
```
IvanKobzarev: Why chunk_len can not be 1 ?

torchtitan/experiments/graph_trainer/dynamic_shapes.py
```diff
@@ -75,6 +130,16 @@ def _symbolic_context_for_marked_dims(tensor: torch.Tensor) -> Any | None:
...
             dynamic_sizes[dim] = DimDynamic.UNBACKED
             constraint_sizes[dim] = RelaxedUnspecConstraint(warn_only=False)
+        elif dim in marked_dynamic_indices:
+            dynamic_sizes[dim] = DimDynamic.DYNAMIC
+            dim_range = dynamic_ranges.get(dim)
+            if dim_range is None or (dim_range.min is None and dim_range.max is None):
+                constraint_sizes[dim] = RelaxedUnspecConstraint(warn_only=False)
+            else:
+                constraint_sizes[dim] = StrictMinMaxConstraint(
+                    vr=ValueRanges(lower=dim_range.min, upper=dim_range.max),
```
IvanKobzarev: For   For mark_dynamic(x, 0, min=2), the metadata is:

  dim_range.min == 2
  dim_range.max == None

(one sided range) it will pass into ValueRanges, that does not support None's

sanketpurandare: Great catch, fixed properly!

torchtitan/components/loss.py
sanketpurandare: Fixed

IvanKobzarev (approved):

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -410,40 +409,23 @@ def _trace_with_args(*args: Any, **kwargs: Any) -> TracedResult:
...
-        )
-
         fake_mode = FakeTensorMode(
             allow_non_fake_inputs=True,
             shape_env=torch.fx.experimental.symbolic_shapes.ShapeEnv(),
         )
-        # Fresh unbacked symbols allocated when fakeifying mark_unbacked() inputs
-        # are wired to placeholders via the symbolic context, so they must not
-        # land in the ShapeEnv's pending_fresh_unbacked_symbols list.
-        ignore_ctx = (
```
aditvenk, line 435: Why remove the ignore_fresh_unbacked_symbols() ? 

Could this cause issues for inputs with mark_unbacked?

torchtitan/distributed/activation_checkpoint.py
ydwu4: putting _local_scalar_dense in comm_ops is weird. Is there other alternative?

torchtitan/experiments/graph_trainer/make_fx_tracer.py
```diff
@@ -410,40 +409,23 @@ def _trace_with_args(*args: Any, **kwargs: Any) -> TracedResult:
         for arg in full_args:
             if not isinstance(arg, torch.Tensor):
                 continue
-            if getattr(arg, "_dynamo_dynamic_indices", None) or getattr(
-                arg, "_dynamo_dynamic_range", None
-            ):
-                raise ValueError("minimal_fx_tracer only supports mark_unbacked()")
-            if _wrapper_subclass_has_mark_unbacked(arg):
+            if _wrapper_subclass_has_marked_dynamic_dims(arg):
```
ydwu4, line 412: hm.. the dynamic shape implentation is getting increasingly complicated. What's the one and only one correct way of doing dynamic shape?

torchtitan/experiments/graph_trainer/remove_noop_passes.py
```diff
@@ -138,6 +138,18 @@ def remove_detach_pass(
 }
 
 
+def _can_guard_true(expr) -> bool:
```
ydwu4: nit: why not just use guard_or_false?

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -90,6 +90,32 @@ def fwd_bwd_step(inputs, labels, global_valid_tokens, extra_kwargs):
     return fwd_bwd_step
 
 
+def _materialize_grad_for_param_layout(
```
ydwu4, line 93: is there a test?

sanketpurandare: Existing coverage is `TestMinimalFXTracerDynamicShapes.test_materialize_grad_for_param_layout_restores_param_strides`.

sanketpurandare:
> There are some coupling between tracer and EP_overlap passes...
> 
> This needs to be called out in readme.md / claude.md ... together with clear instruction on how to use ep overlap, and document it's composability and limitations.

@SherlockNoMad Added an EP-overlap section to the GraphTrainer README and a shorter developer
note in the graph_trainer guidance. The docs now make the intended coupling explicit: EP-overlap graph chunking has a trace-input preparer that marks token-grid dims before `minimal_fx_tracer` fakeifies inputs, and the graph passes consume those symbolic dims. The docs also list the supported flags, supported `chunk_dim`/`module_fqn` combinations, and current composability
limits.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
sanketpurandare: @aditvenk This was actually a hack, using `ignore_fresh_unbacked_symbols()`. This path no longer needs `ignore_fresh_unbacked_symbols()` because `_fakeify_input()` converts the input marking into a `StatelessSymbolicContext`, so the unbacked SymInt is owned by the placeholder symbolic context instead of being a fresh unowned symbol produced by an op. The regression test asserts that a `mark_unbacked` placeholder leaves no pending or ignorable fresh unbacked symbols after tracing.

torchtitan/experiments/graph_trainer/make_fx_tracer.py
sanketpurandare: Tightened the dynamic-shape contract and documented it in `dynamic_shapes.py`: the single supported user-facing mechanism is PyTorch's Dynamo tensor marking APIs on plain input tensors (`mark_dynamic` and `mark_unbacked`). The tracer imports those markings through `StatelessSymbolicContext`; wrapper subclasses with marked dims are rejected before unwrapping; and all raw `_dynamo_*` interpretation stays in `dynamic_shapes.py`. This makes the implementation rule explicit instead of letting EP-overlap tracing look like a separate dynamic-shape system.

torchtitan/experiments/graph_trainer/remove_noop_passes.py
sanketpurandare: Removed the `_can_guard_true()` wrapper and used `guard_or_false` directly.

torchtitan/experiments/graph_trainer/trainer.py
sanketpurandare: The test checks that a dense grad with a different stride is materialized to the parameter stride and that the helper is a no-op when the grad already satisfies the parameter layout contract.

aditvenk (approved):

torchtitan/distributed/activation_checkpoint.py
tianyu-l: where is this op called?

torchtitan/components/loss.py
tianyu-l, line 1: @pianpwk could you help review

torchtitan/models/common/rope.py
```diff
@@ -364,30 +379,45 @@ def _reshape_for_broadcast(
     cache_width = rope_cache.shape[-1]
     if positions is None:
         rope_cache = rope_cache[0:seqlen]
-        assert rope_cache.shape == (seqlen, cache_width)
```
tianyu-l, line 368: I'm lost in the changes, definitely not readable / mergeable.

Would you be able to move forward, if we turn these explicit checks into comments?

tianyu-l (changes requested):

wwwjn (approved):
Stamp to unblock

torchtitan/models/common/rope.py
sanketpurandare: The change is not that complex, we are asserting shape equalities in the original code, this means we need concrete shape at trace time to prove the equality. When you use `torch._check`, if the equality can be determined at trace time, it is evaluated and works like an assert. If you can't determine the equality due to a dynamic shape you emit a runtime assertion that will be graph node which will error out if the runtime actual shape doesn't satisfy the equality. I will write a more interpretable version that may have a bit of code redundancy but will read more easily. If you are still on satisfied we can turn them into comments.

torchtitan/models/common/rope.py
```diff
@@ -362,33 +363,47 @@ def _reshape_for_broadcast(
...
-            index=positions.view(bsz, seqlen, 1, 1).expand(bsz, seqlen, 1, cache_width),
-        )
-        return rope_cache
+
+    # Per-batch positions, plus singleton positions whose first dimension was
+    # not statically provable above, use the general gather path.
+    positions_bsz = positions.shape[0]
+    torch._check(positions.shape[1] == seqlen)
+    torch._check((positions_bsz == 1) | (positions_bsz == bsz))
+    positions = positions.expand(bsz, seqlen)
```
tianyu-l: is this new, why?

torchtitan/distributed/activation_checkpoint.py
tianyu-l: not answered

torchtitan/models/common/rope.py
```diff
@@ -362,33 +363,47 @@ def _reshape_for_broadcast(
...
+    torch._check(positions.ndim == 2)
+    # Concrete/provable singleton positions can use the cheaper prefix-shaped
+    # view path. If singleton-ness is symbolic, fall through to gather below.
+    if guard_or_false(positions.size(0) == 1):
+        torch._check(positions.shape[0] == 1)
+        torch._check(positions.shape[1] == seqlen)
         rope_cache = rope_cache[positions.squeeze(0)]
-        assert rope_cache.shape == (seqlen, cache_width)
+        torch._check(rope_cache.shape[0] == seqlen)
+        torch._check(rope_cache.shape[1] == cache_width)
```
tianyu-l: my general feeling, representing myself:
- originally people would like to have `assert` to have confidence in code correctness
- now we already know the code is correct, and adding these checks to satisfy graph tracing. So it sounds to me "user doesn't need this, graph tracer need them"
- This is a programming model change (tiny or big). In order for it to be the default, graph tracer needs to have proper education to the world, e.g. via documentation.

I would like us to be gradually getting there, so suggesting removal for now.

sanketpurandare: I am okay with removing all the asserts. I had to add torch._check because of the assert. Coming to the if-else case, if the condition based on a shape value is satisfied then we take the fast path and if not then we are taking a safe path that works for all cases. In such cases we need to use `gurad_or_false`. There is no way around it. If you are okay taking the default path that works for all cases then we can remove the `guard_or_false`.

torchtitan/distributed/activation_checkpoint.py
sanketpurandare: https://github.com/pytorch/torchtitan/pull/3362/changes#r3456002294

sanketpurandare: But anyways, now that Simon has already added routed input shape to be the output size of repeat_interleave and torch.arange I no longer need it so removing it.

sanketpurandare: Also FYI, when you use to_list() on a integer tensor these ops are generated or when you create a cpu scalar from gpu scalar

torchtitan/models/common/rope.py
sanketpurandare: It is not new check the logic, if I am not able to prove `positions.size(0) == 1` then I need to first expand them in this case since guard_or_false will fallback to this path. It is the same logic as above.

torchtitan/components/loss.py
pianpwk: as of recently, we already assert this chunk divisibility in init time right ? otherwise checks look fine to me

torchtitan/components/loss.py
```diff
@@ -523,6 +523,37 @@ def set_lm_head(self, lm_head: nn.Module) -> None:
...
+        chunk_len = seq_len // num_chunks
+        torch._check(
+            chunk_len > 0,
+            lambda: (
+                "ChunkedCELoss sequence chunks must have positive length, "
+                f"got seq_len={seq_len} and num_chunks={num_chunks}"
+            ),
+        )
+        return tuple(
+            c.contiguous() for c in torch.split(t, [chunk_len] * num_chunks, dim=1)
```
pianpwk: chunk->split is because chunk is unbacked-symint unfriendly? could also slice/narrow but nbd

torchtitan/models/common/rope.py
```diff
@@ -362,33 +363,47 @@ def _reshape_for_broadcast(
     assert ndim > 1
     bsz, seqlen = query_shape[:2]
     # cache_width is `head_dim * 2` for CosSinRoPE, and `head_dim // 2` for ComplexRoPE
+    torch._check(rope_cache.ndim == 2)
```
pianpwk: should this be a plain assertion, ndim should never be symbolic ?

torchtitan/models/common/rope.py
```diff
@@ -362,33 +363,47 @@ def _reshape_for_broadcast(
...
     cache_width = rope_cache.shape[-1]
     if positions is None:
+        # No explicit positions: use the prefix cache and broadcast it over batch.
         rope_cache = rope_cache[0:seqlen]
-        assert rope_cache.shape == (seqlen, cache_width)
+        torch._check(rope_cache.shape[0] == seqlen)
+        torch._check(rope_cache.shape[1] == cache_width)
```
pianpwk: nit: is nice to do `torch._check(..., lambda: error msg)` to be user friendly

torchtitan/components/loss.py
sanketpurandare: yeah chunk implementation does an internal assert which is not unbacked friendly

torchtitan/models/common/rope.py
sanketpurandare: I tried to do that Tianyu wanted minimal change

sanketpurandare: The earlier asserts didn't have error message either

torchtitan/components/loss.py
sanketpurandare: init time asserts are fine for me they don't get traced

torchtitan/models/common/rope.py
pianpwk: should we just custom-opify this in future

pianpwk: but also all this should go away if we figure out deleting the rope broadcasting branch ? unless graph trainer actually encounters it. most of the pain is just from deciding to expand or not

sanketpurandare: graph trainer doesn't want the broadcast branch, rope had it originally, so i just tried to make it tracer friendly. all the asserts should anyway be during init time, runtime paths having asserts is also bad for perf

pianpwk: I tried deleting in https://github.com/pytorch/torchtitan/pull/3693 but had to revert https://github.com/pytorch/torchtitan/pull/3750

@tianyu-l if it helps, after deletion the only delta should be `assert ... -> torch._check(..., lambda: some nice error message)`

torchtitan/models/common/rope.py
```diff
@@ -362,33 +363,47 @@ def _reshape_for_broadcast(
...
-            dim=1,
-            index=positions.view(bsz, seqlen, 1, 1).expand(bsz, seqlen, 1, cache_width),
-        )
-        return rope_cache
+
+    # Per-batch positions, plus singleton positions whose first dimension was
+    # not statically provable above, use the general gather path.
+    positions_bsz = positions.shape[0]
+    torch._check(positions.shape[1] == seqlen)
+    torch._check((positions_bsz == 1) | (positions_bsz == bsz))
```
pianpwk: sorry for lots of comments - was this one needed ? I thought expand will just implicitly add its own checks: https://github.com/pytorch/pytorch/blob/5b88630d363bf79e0dc9263588cf5f8054af54cf/torch/_refs/__init__.py#L3197-L3200

pianpwk (approved):

torchtitan/models/common/rope.py
sanketpurandare: @tianyu-l and @pianpwk I removed the torch._check.

torchtitan/models/common/rope.py
```diff
@@ -364,24 +365,27 @@ def _reshape_for_broadcast(
     # cache_width is `head_dim * 2` for CosSinRoPE, and `head_dim // 2` for ComplexRoPE
     cache_width = rope_cache.shape[-1]
     if positions is None:
+        # No explicit positions: use the prefix cache and broadcast it over batch.
         rope_cache = rope_cache[0:seqlen]
-        assert rope_cache.shape == (seqlen, cache_width)
```
tianyu-l, line 368: please move to comment, not delete

torchtitan/models/common/rope.py
```diff
@@ -364,24 +365,27 @@ def _reshape_for_broadcast(
...
             d if i == 1 else cache_width if i == ndim - 1 else 1
             for i, d in enumerate(query_shape)
         ]
         return rope_cache.view(*shape)
-    elif positions.size(0) == 1:
-        assert positions.shape == (1, seqlen)
+
+    # Concrete/provable singleton positions can use the cheaper prefix-shaped
+    # view path. If singleton-ness is symbolic, fall through to gather below.
+    if guard_or_false(positions.size(0) == 1):
```
tianyu-l, line 381: This branch was removed, but added back in https://github.com/pytorch/torchtitan/pull/3750 due to not compatible with vllm during inference

Add a TODO for @pianpwk to remove it

torchtitan/components/loss.py
```diff
@@ -559,9 +583,10 @@ def __call__(
         # local seq=0 and breaking GradAccumulator's slice writes.
         # ``local_map`` runs the chunking body on plain tensors; under the
         # non-DTensor (eager) path we call ``_chunk_local`` directly.
-        # ``.contiguous()`` breaks shared storage from ``torch.chunk``.
+        # Equal chunk sizes also match GradAccumulator's sequential slice
+        # writes, which use one chunk length for each write offset.
         def _chunk_local(t):
-            return tuple(c.contiguous() for c in torch.chunk(t, num_chunks, dim=1))
+            return ChunkedCELoss._split_equal_chunks(t, num_chunks)
```
tianyu-l: using a static method inside an inline function sounds strange -- can we also inline `_split_equal_chunks`?

tianyu-l (approved):
please address remaining comments

PR #3364, opened 2026-05-15
to_local() for ChunkedCELoss + no LP

pianpwk:
Fixes mixed tensor-DTensor error when running TP w/ no SP, no loss parallel: `NGPU=4 ./run_train.sh --module llama3 --config llama3_debugmodel --parallelism.tensor_parallel_degree=4 --activation_checkpoint.mode=none --training.steps=10 --parallelism.no_enable_sequence_parallel --parallelism.disable_loss_parallel --debug.deterministic --debug.seed 42`

I believe https://github.com/pytorch/torchtitan/pull/2937 calls lm_head with a R@TP sharding config, and calls plain `cross_entropy_loss` with plain tensor labels, if loss parallel is off. However the logits remain in DTensor form (a `use_local_output=True` flag is appropriate but we don't have that).

The fix is asserting they're R@TP (sharded on DP/CP axes should be fine?) and extracting local tensor for loss calculation.

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -11,6 +11,7 @@ from typing import TypeAlias
 
 import torch
 import torch.nn as nn
+from torch.distributed.tensor import DTensor, Replicate
 from torchtitan.config import CompileConfig, Configurable
 from torchtitan.tools.logging import logger
 
@@ -22,6 +23,13 @@ LossFunction: TypeAlias = Callable[..., torch.Tensor]
 
 def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
     """Cross-entropy loss with sum reduction for token-based normalization."""
+    if isinstance(pred, DTensor) and not isinstance(labels, DTensor):
+        mesh_axis_names = pred.device_mesh.mesh_dim_names
+        if mesh_axis_names is not None and "tp" in mesh_axis_names:
+            tp_axis = mesh_axis_names.index("tp")
+            if isinstance(pred.placements[tp_axis], Replicate):
+                pred = pred.to_local()
+
     return torch.nn.functional.cross_entropy(
         pred.flatten(0, 1).float(),
         labels.flatten(0, 1),


torchtitan/components/loss.py
```diff
@@ -22,6 +23,13 @@
 
 def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
     """Cross-entropy loss with sum reduction for token-based normalization."""
+    if isinstance(pred, DTensor) and not isinstance(labels, DTensor):
+        mesh_axis_names = pred.device_mesh.mesh_dim_names
+        if mesh_axis_names is not None and "tp" in mesh_axis_names:
```
fegin, line 28: `labels` is not DTensor, only `tp` should be in `mesh_axis_names`. Let's convert this to an assert.

torchtitan/components/loss.py
```diff
@@ -22,6 +23,13 @@
...
+            tp_axis = mesh_axis_names.index("tp")
+            if isinstance(pred.placements[tp_axis], Replicate):
+                pred = pred.to_local()
```
fegin, line 31: ~~If SP is disabled, but loss is enabled, does this still work?  loss parallel requires DTensor.~~

fegin (commented):
Can you verify PP + no sp + no loss_parall ~~and loss_parall + no sp~~?

torchtitan/components/loss.py
pianpwk: Hmm I think LP goes through the `loss_parallel` handling, which handles mixed types: https://github.com/pytorch/pytorch/blob/2f4ebdb928b9f8043d7103151e57a06fcfa061b4/torch/distributed/tensor/parallel/loss.py#L282-L303

torchtitan/components/loss.py
pianpwk: ah I guess we won't hit this in full DTensor/spmd_types right, sounds good

torchtitan/components/loss.py
fegin: No, my comment doesn't make sense. Just ignore my comment. Just check if PP works

wwwjn:
Hmm I think the error should be hit for both chunked CE loss and non-chunked CE loss? If TP outputs a DTensor, and there's no loss_parallel() to wrap nn.cross_entropy and convert `labels ` to be DTensor  , it will hit the error. Did I understand correctly?

torchtitan/components/loss.py
wwwjn, line 31: I think what loss parallel do is wrapping `labels` to DTensor. As we are moving to full DTensor, `labels` should also be DTensor. Why we don't convert labels to be DTensor, but using another way around?

PR #3365, opened 2026-05-15
[rl] fix CI timeout issue by properly tear down for vllm engine V2

wwwjn:
RL integration test failure because of timeout.
- The training finished, but stuck in tear down step and finally timeout
- It starts when we switch to vLLM engine v2

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -32,7 +32,6 @@ from torchtitan.tools.logging import init_logger
 from torchtitan.tools.utils import has_cuda_capability
 from vllm import EngineArgs, LLMEngine, SamplingParams
 from vllm.config import AttentionConfig, CompilationConfig
-from vllm.distributed.parallel_state import cleanup_dist_env_and_memory
 from vllm.outputs import RequestOutput
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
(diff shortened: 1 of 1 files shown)


wwwjn:
Current RL CI failure is the same as main CI failure, we can wait for next PyTorch nightly which should contain the fix

felipemello1 (approved):

felipemello1:
the error i got was that `engine_core` doesnt have the method `shutdown()`. Can you confirm that the error you have is truly because of `cleanup_dist_env_and_memory`, and not the `shutdown` method?

wwwjn:
> the error i got was that `engine_core` doesnt have the method `shutdown()`. 

Is the error you got before this PR or after this PR? 

I'm postpone landing this PR and waiting for the CI to turn green. But now we are blocked by https://download.pytorch.org/whl/nightly/vllm/  vllm nightly is not built recently

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -424,12 +423,11 @@ async def close(self) -> None:
...
-        3. ``cleanup_dist_env_and_memory()`` — destroys NCCL / model-
-           parallel process groups, runs ``gc.collect``, empties the
-           accelerator cache.
 
-        Each step runs in a ``try/finally`` so a failure in one step
-        does not skip the next.
+        We intentionally skip ``cleanup_dist_env_and_memory()`` here:
+        with ``external_launcher``, vLLM reuses the process group that
+        Monarch created. Destroying it here would prevent Monarch's
+        ``mesh.stop()`` from completing its own collective teardown.
```
pzhan9, line 531: In my observation, the hang happened at this `cleanup_dist_env_and_memory` step. The script never reached `mesh.stop()`.

wwwjn: Thanks, yes, so the solution is skip cleanup_dist_env_and_memory() , and let mesh.stop() handles shutdown

PR #3367, opened 2026-05-15
ci: declare workflow-level `contents: read` on 2 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 2 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

The following files were left implicit because they reference `GITHUB_TOKEN` / use a write-scope action / trigger on `pull_request_target`. Those scopes are best declared by maintainers: `docker-builds.yml`, `lint.yaml`.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/unit_test_cpu.yaml
+++ b/.github/workflows/unit_test_cpu.yaml
@@ -13,6 +13,9 @@ concurrency:
   group: unit-test${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   build-test:
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
--- a/.github/workflows/unit_test_cpu_torchft.yaml
+++ b/.github/workflows/unit_test_cpu_torchft.yaml
@@ -15,6 +15,9 @@ concurrency:
   group: unit-test${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   build-test:
     uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main


tianyu-l:
@claude review this PR

tianyu-l:
> set-matrix.yaml also lacks a permissions: block, but since it's a workflow_call (reusable workflow), it inherits permissions from the calling workflow. Not adding it there is reasonable — though a permissions: {} (empty/no permissions) could further harden it since it doesn't need any token access at all. This is purely optional and not a concern for this PR.

should we do it or not?

arpitjain099:
Good question @tianyu-l. Reading `set-matrix.yaml` end to end:

- Takes a few `workflow_call` inputs.
- Branches on `github.event_name` / `github.ref` evaluated by Actions before the step.
- Writes a matrix JSON to `$GITHUB_OUTPUT`.

There is no checkout step, no `secrets.GITHUB_TOKEN` reference, and no external action that uses a token. So the workflow does not actually need any scopes on `GITHUB_TOKEN`.

On the inheritance semantics: when a reusable workflow does not declare `permissions:`, the called job inherits whatever the caller has issued at the call site. When the reusable declares `permissions:` explicitly, those become the scopes for the called job regardless of what the caller grants (it does not get to widen them past what the caller has). So adding `permissions: {}` to `set-matrix.yaml`:

- has zero functional impact today since the job never touches the GitHub API,
- protects future-you if someone adds a third-party action to this file later and the caller has a wide default token,
- doesn't risk breaking any caller because the callers don't rely on `set-matrix.yaml` having token scope.

Happy to roll it into this PR if you'd like the defense-in-depth, or leave it for a follow-up. Your call.

tianyu-l (approved):

PR #3369, opened 2026-05-15
[graph_trainer] Use separate EP process groups for overlap

sanketpurandare:
Stacked PRs:
 * #3328
 * #3325
 * #3363
 * #3362
 * __->__#3369

--- --- ---

### [graph_trainer] Use separate EP process groups for overlap

When EP all-to-alls use the same process group object as FSDP or TP collectives, they share a NCCL communicator/stream and can serialize communication that the EP overlap schedule is trying to expose.

Add isolate_ep_process_group_pass as a dedicated EP pass instead of mixing this policy into FSDP bucketing. The pass scans traced collectives, identifies EP process-group users by _c10d_functional.all_to_all_single.default nodes, and compares the resolved process-group object identity against non-EP collective process groups. If an EP process group is object-identical to a non-EP collective group, the pass creates one high-priority EP-only process group over the same EP ranks and rewrites all traced EP all-to-all calls that used the shared source group.

The matching contract is intentionally process-group based because graph-level metadata does not carry DeviceMesh objects. Rank-set equality is only a validator: two distinct process groups may have the same ranks but already have separate NCCL communicators, so rank-set equality alone is not a sufficient reason to rewrite EP. The replacement communicator is always constructed from the EP process group itself, preserving the EP group rank order used by all-to-all split semantics.

FSDP all-gathers continue to use the existing extra FSDP group from overlap_fsdp_ag_rs_pass. Generic MoE dispatch/combine traceback metadata and EP_token_exchange scheduling metadata are not required for EP process-group isolation; the scheduling pass still owns token-exchange wait ordering.

Test Plan:
- pytest -q torchtitan/experiments/graph_trainer/tests/test_passes.py::TestOverlapFsdpAgRsPass

### DIFF
--- a/torchtitan/experiments/graph_trainer/fsdp_passes.py
+++ b/torchtitan/experiments/graph_trainer/fsdp_passes.py
@@ -64,28 +64,47 @@ def is_wait_tensor_from_fsdp(node: torch.fx.Node) -> bool:
 _EXTRA_FSDP_PG_REGISTRY: dict[str, str] = {}
 
 
-def _get_or_create_extra_fsdp_pg(source_pg_name: str) -> str:
-    """Return the extra PG name for ``source_pg_name``, creating it once.
-
-    The extra PG is a new NCCL process group with the same ranks as the source
-    FSDP PG but a different communicator (and therefore a different CUDA stream).
-    """
+def _get_or_create_extra_pg(
+    source_pg_name: str,
+    registry: dict[str, str],
+    *,
+    group_desc: str,
+    high_priority: bool = False,
+) -> str:
     import torch.distributed as dist
 
-    if source_pg_name in _EXTRA_FSDP_PG_REGISTRY:
-        return _EXTRA_FSDP_PG_REGISTRY[source_pg_name]
+    if source_pg_name in registry:
+        return registry[source_pg_name]
 
     source_pg = dist.distributed_c10d._resolve_process_group(source_pg_name)
     ranks = dist.get_process_group_ranks(source_pg)
-    extra_pg = dist.new_group(
-        ranks=ranks, group_desc="fsdp_extra", use_local_synchronization=True
+    pg_options = (
+        dist.ProcessGroupNCCL.Options(is_high_priority_stream=True)
+        if high_priority and hasattr(dist, "ProcessGroupNCCL")
+        else None
     )
-    _EXTRA_FSDP_PG_REGISTRY[source_pg_name] = extra_pg.group_name
+    extra_pg = dist.new_group(
+        ranks=ranks,
+        backend="nccl" if pg_options is not None else None,
+        pg_options=pg_options,
+        group_desc=group_desc,
+        use_local_synchronization=True,
+    )
+    registry[source_pg_name] = extra_pg.group_name
     logger.info(
-        f"Created extra FSDP PG (source: {source_pg_name}, "
-        f"extra: {extra_pg.group_name})"
+        f"Created extra {group_desc} PG (source: {source_pg_name}, "
+        f"extra: {extra_pg.group_name}, high_priority={high_priority})"
     )
     return extra_pg.group_name
+
+
+def _get_or_create_extra_fsdp_pg(source_pg_name: str) -> str:
+    """Return an extra FSDP PG with the same ranks and a distinct NCCL stream."""
+    return _get_or_create_extra_pg(
+        source_pg_name,
+        _EXTRA_FSDP_PG_REGISTRY,
+        group_desc="fsdp_extra",
+    )
 
 
 def overlap_fsdp_ag_rs_pass(
@@ -93,15 +112,15 @@ def overlap_fsdp_ag_rs_pass(
     example_inputs: tuple | None = None,
 ) -> torch.fx.GraphModule:
     """
-    Reassign FSDP all-gather nodes to extra NCCL process groups for
-    AG/RS overlap in backward.
+    Reassign FSDP all-gathers to extra NCCL process groups.
 
     Discovers all distinct FSDP PGs by inspecting the graph (e.g. one for
     FSDP, another for expert-FSDP), creates an extra NCCL PG over the same
     ranks for each (giving it a separate CUDA stream), and rewrites every
     all-gather to the corresponding extra PG. This separates all-gathers
     from reduce-scatters onto different streams, enabling AG/RS overlap in
-    backward.
+    backward. EP communicator isolation is handled by
+    ``isolate_ep_process_group_pass``.
 
     No-op when the graph has no FSDP all-gathers. Must be applied BEFORE
     bucketing passes so bucketed all-gathers inherit the new PG name.
@@ -119,13 +138,13 @@ def overlap_fsdp_ag_rs_pass(
         pg: _get_or_create_extra_fsdp_pg(pg) for pg in source_pg_names
     }
 
-    count = 0
+    ag_count = 0
     for node in gm.graph.nodes:
         if is_all_gather(node) and node.args[2] in pg_mapping:
             # AG args: (input_tensor, group_size, group_name)
             node.args = (node.args[0], node.args[1], pg_mapping[node.args[2]])
-            count += 1
-    if count > 0:
+            ag_count += 1
+    if ag_count > 0:
         for source, target in pg_mapping.items():
             logger.info(f"Rewrote all-gather node(s) from PG {source} to PG {target}")
     gm.recompile()
--- a/torchtitan/experiments/graph_trainer/ep_process_group_pass.py
+++ b/torchtitan/experiments/graph_trainer/ep_process_group_pass.py
@@ -1,0 +1,165 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""EP process-group isolation for graph_trainer.
+
+Algorithm:
+1. Scan collectives and separate EP all-to-alls from all other collectives.
+2. If an EP all-to-all resolves to the same process-group object as a non-EP
+   collective, create one extra PG with the same ranks for EP traffic.
+3. Rewrite the EP all-to-all PG argument. Other collectives keep their original
+   PGs and ordering.
+"""
+
+from __future__ import annotations
+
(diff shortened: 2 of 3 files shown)


torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+    return _get_or_create_extra_pg(
+        source_pg_name,
+        _EXTRA_EP_PG_REGISTRY,
+        group_desc="ep_extra",
+        high_priority=True,
+    )
+
+
+def _pg_rank_set(pg_name: str) -> frozenset[int]:
+    import torch.distributed as dist
```
aditvenk: nit: looks like we importing locally in 3+ functions, consider moving to the top of the file.

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+    return False
+
+
+def _is_ep_all_to_all(node: fx.Node) -> bool:
+    custom = node.meta.get("custom", {})
+    return (
+        node.op == "call_function"
+        and "all_to_all_single" in str(node.target)
+        and isinstance(custom, dict)
+        and custom.get("EP") in ("dispatch", "combine")
```
aditvenk: Who adds this metadata?

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+    Reassign FSDP all-gathers and overlapping EP all-to-alls to extra NCCL
+    process groups.
 
     Discovers all distinct FSDP PGs by inspecting the graph (e.g. one for
     FSDP, another for expert-FSDP), creates an extra NCCL PG over the same
     ranks for each (giving it a separate CUDA stream), and rewrites every
     all-gather to the corresponding extra PG. This separates all-gathers
     from reduce-scatters onto different streams, enabling AG/RS overlap in
-    backward.
+    backward. If an EP all-to-all resolves to the same process group as an FSDP
```
aditvenk: Would we ever want to enable EP A2A in a separate stream but not a separate stream for AG/RS ? Should this be a separate pass with its own enable toggle?

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+        if is_fsdp_reduce_scatter(node):
+            fsdp_rs_pg_names.add(node.args[3])
 
     if not source_pg_names:
         return gm
 
     pg_mapping: dict[str, str] = {
         pg: _get_or_create_extra_fsdp_pg(pg) for pg in source_pg_names
     }
+    ep_source_pg_names = OrderedSet(
```
aditvenk: nit: Can avoid extra O(N) scan of graph nodes by populating ep_source_pg_names inside previous for loop

aditvenk (approved):

torchtitan/experiments/graph_trainer/fsdp_passes.py
SherlockNoMad: `custom.get("EP") in ("dispatch", "combine")`  this is not reliable. 

In some come case, I see the entire EP is annotated as "compute"

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+                )
+            return True
+    return False
+
+
+def _is_ep_all_to_all(node: fx.Node) -> bool:
+    custom = node.meta.get("custom", {})
+    return (
+        node.op == "call_function"
+        and "all_to_all_single" in str(node.target)
```
SherlockNoMad: let's do direct target match. 

node.target == torch.ops.c10d.all_to_all_single.default

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+                    f"{pg_name} and {candidate_pg_name}"
+                )
+            return True
+    return False
+
+
+def _is_ep_all_to_all(node: fx.Node) -> bool:
+    custom = node.meta.get("custom", {})
+    return (
+        node.op == "call_function"
```
SherlockNoMad: node.op == "call_function" is redundant.

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
-    backward.
+    backward. If an EP all-to-all resolves to the same process group as an FSDP
+    reduce-scatter, it receives its own extra high-priority PG constructed from
+    the EP group. Rank-set equality alone is not enough: two different process
+    groups can have the same ranks but already use separate NCCL communicators.
 
     No-op when the graph has no FSDP all-gathers. Must be applied BEFORE
     bucketing passes so bucketed all-gathers inherit the new PG name.
     """
     source_pg_names: OrderedSet[str] = OrderedSet()
```
SherlockNoMad, line 128: are these fsdp_ag_pg_names? rename if so

SherlockNoMad: these looks like a mixed of fsdp ag and rs pgs?

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
...
+    custom = node.meta.get("custom", {})
+    return (
+        node.op == "call_function"
+        and "all_to_all_single" in str(node.target)
+        and isinstance(custom, dict)
+        and custom.get("EP") in ("dispatch", "combine")
+    )
+
+
 def overlap_fsdp_ag_rs_pass(
```
SherlockNoMad, line 110: update pass name to reflect EP change. 

overlap_collectives_pass? 
reassign_pg_for_overlap_collectives_pass?

torchtitan/experiments/graph_trainer/fsdp_passes.py
```diff
@@ -135,72 +136,171 @@ def fsdp_reshard_after_fwd_pass(
 # Each NCCL PG gets its own CUDA stream, so the extra PG is what enables
 # AG/RS overlap in backward.
 _EXTRA_FSDP_PG_REGISTRY: dict[str, str] = {}
+_EXTRA_EP_PG_REGISTRY: dict[str, str] = {}
```
SherlockNoMad: feels like they can be merged?

SherlockNoMad (approved):
thanks! lgtm overall, just minor comments.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Moved the EP process-group logic into its own graph-trainer pass, so the local imports are no longer spread through the FSDP pass. The remaining distributed imports are contained in the PG resolution/creation helpers that need them.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Reworked this to identify EP process-group users by the traced `_c10d_functional.all_to_all_single.default` calls themselves, instead of relying on generic `custom["EP"]` phase metadata. That keeps PG isolation tied to the collectives issued through the EP mesh rather than to scheduler-only traceback labels.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Split this out into `isolate_ep_process_group_pass` with its own graph-trainer pass hook. It rewrites EP all-to-all collectives when their process group object is shared with non-EP collectives.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: The standalone pass now performs one graph scan to classify EP and non-EP process groups, then rewrites only the matching EP all-to-alls. There is no extra FSDP-specific second scan for EP source PG discovery.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Updated the predicate to match the actual `_c10d_functional.all_to_all_single.default` target directly. The pass no longer requires the `EP_token_exchange` scheduler annotation for PG isolation.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Kept the `node.op` guard only as an FX safety check before reading call-function arguments. The semantic selection is the direct c10d all-to-all target; PG isolation no longer depends on the token-exchange annotation.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Renamed and generalized the collection to distinguish EP all-to-all PGs from non-EP collective PGs. The pass now handles FSDP AG/RS, TP collectives, and other non-EP collectives uniformly.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Renamed the pass to `isolate_ep_process_group_pass`, which reflects the actual behavior: EP all-to-alls are moved to an EP-only process group when they otherwise share a PG with non-EP collectives.

torchtitan/experiments/graph_trainer/fsdp_passes.py
sanketpurandare: Kept the EP PG isolation separate from FSDP bucketing because it is not FSDP-specific. It can apply when EP shares a PG with FSDP, TP, or another non-EP collective source.

PR #2545, opened 2026-03-11
[Qwen3.5 MoE] Add hybrid decoder model (GatedDeltaNet + full attention + MoE)

gali-leilei:
## Summary

Upstreams Qwen3.5 MoE support from the internal fork. Qwen3.5 MoE is a hybrid decoder: **GatedDeltaNet** (linear attention) for most layers + **full attention** every N layers, with **MoE + gated shared expert FFN**. Validated on real 35B-A3B checkpoints (HF↔DCP roundtrip: 19/19 tensors exact) and L20X GPUs (debugmodel + 35B-A3B, 10 steps, no errors).

## New model: `torchtitan/models/qwen3_5_moe/`

- **model.py**: `OffsetRMSNorm`, `RMSNormGated`, `GatedDeltaNet` (FLA optional, `torch_naive` fallback), `Attention` (output gating + partial RoPE), `TransformerBlock`, `Model`
- **parallelize.py**: TP/FSDP/CP/EP support. DTensor-safe wrappers for depthwise conv1d (`_DTensorSafeConv1d`) and FLA kernel dispatch (`_install_dtensor_safe_dispatch`). Registers `softplus` as DTensor pointwise op.
- **state_dict_adapter.py**: HF↔TorchTitan conversion. Handles expert `gate_up_proj [E,2I,D]` ↔ `w1`+`w3` split/concat.
- **config_registry.py**: `debugmodel`, `35b-a3b`, `35b-a3b-sdpa`, `35b-a3b-varlen` training configs
- **`__init__.py`**: model configs (debugmodel, 35b-a3b, 122b-a10b, 397b-a17b, 397B_A19B), `build()`, `model_registry()`

## Related changes

- `torchtitan/models/__init__.py`: add `"qwen3_5_moe"` to `_supported_models`
- `tests/integration_tests/models.py`: `qwen3_5_moe` FSDP+TP+EP integration test (4 GPUs)

## Test plan

- [ ] `pre-commit run --all-files` — flake8, ufmt, pydoclint, codespell all pass
- [ ] Integration test on 4 GPUs: `--module qwen3_5_moe --config qwen3_5_moe_debugmodel` with FSDP+TP+EP

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3371, opened 2026-05-15
[qwen3_5] evolve qwen3_vl to qwen3_5

shuhuayu:
Qwen3.5 supersedes Qwen3-VL with a hybrid attention architecture: 75% GatedDeltaNet (linear attention) + 25% full attention with output gating and partial RoPE.

Model changes:                                            
- Hybrid decoder with GatedDeltaNet and Qwen35Attention
- Head-sharded TP on GatedDeltaNet projections (`ColwiseParallel`/`RowwiseParallel`)                                            
- OffsetRMSNorm, RMSNormGated, MoE with shared expert                                                                       
- Removed DeepStack                                                                                                         

Parallelisms: fsdp, tp+sp, ep, pp, verified identical logits via numerical tests (`scripts/checkpoint_conversion/numerical_tests_qwen3_5_shard.py`). 

Numerical parity: kl ~3e-7 against hf models (4b, multimodal) and 100% top-1/top-5 match (`scripts/checkpoint_conversion/numerical_tests_qwen3_5.py`).

Many thanks to @gali-leilei for initiating the effort of enabling qwen3.5 decoder in torchtitan in https://github.com/pytorch/torchtitan/pull/2545, some components are reused in this pr.


torchtitan/models/utils.py
```diff
@@ -463,6 +463,8 @@ def get_moe_model_nparams_and_flops(
     n_heads: int,
     head_dims: int,
     seq_len: int,
+    *,
+    num_full_attn: int | None = None,
```
tianyu-l: can compute this from `model_config` right?

torchtitan/models/qwen3_5/README.md
```diff
@@ -0,0 +1,88 @@
...
+| Expert Parallelism (EP) | For MoE variants |
+| Context Parallel (CP) | Text-only; full attention uses ring attention, GatedDeltaNet allgathers full sequence |
+| Pipeline Parallel (PP) | Vision encoder assigned to first stage; 1F1B and Interleaved1F1B schedules |
+| Sample Packing | Configurable via `packing_buffer_size` in dataloader config |
+
+## Numerical Parity
+
+End-to-end KL divergence against HuggingFace Transformers (4B, multimodal inputs): **~3e-7** average, with **100% top-1 and top-5 match**.
+
+Parallelism correctness: bitwise identical logits across no-parallel, FSDP, FSDP+EP, FSDP+EP+TP, and FSDP+EP+TP+CP configs.
```
tianyu-l: hmm, how could this be true? Different parallelisms have different reductions

torchtitan/models/qwen3_5/sharding.py
```diff
@@ -0,0 +1,371 @@
...
+    TODO: Remove once DTensor Conv1d dispatch handles sharded groups.
+    """
+    original_forward = conv1d_module.forward.__func__
+
+    def safe_forward(self, x):
+        if isinstance(x, DTensor):
+            mesh, plc = x.device_mesh, x.placements
+            w = self.weight
+            if isinstance(w, DTensor):
+                w = w.to_local()
```
tianyu-l: With spmd_types, hopefully we don't need to do this manual conversion.

For now, let's do to_local in the module, similar to GroupedExperts, and use LocalMapConfig to convert inputs, instead of patching forward.

torchtitan/models/qwen3_5/sharding.py
```diff
@@ -0,0 +1,371 @@
...
+    conv1d_module.forward = types.MethodType(safe_forward, conv1d_module)
+
+
+def _wrap_pos_embed_for_interpolation(vision_encoder) -> None:
+    """Wrap compute_position_embeddings to convert pos_embed to local.
+
+    F.interpolate's decomposition uses _unsafe_index which doesn't support
+    DTensor. Since pos_embed is Replicate, to_local is a no-op for data.
+
+    TODO: Remove once F.interpolate on FSDP2-managed DTensors is fixed upstream.
```
tianyu-l: If this can be fixed soon, let's wait.

torchtitan/models/qwen3_5/parallelize.py
```diff
@@ -0,0 +1,200 @@
...
+    edp_mesh = None
+    if parallel_dims.ep_enabled:
+        edp_mesh_names = (
+            ["dp_replicate", "efsdp"]
+            if parallel_dims.dp_replicate_enabled
+            else ["efsdp"]
+        )
+        edp_mesh = parallel_dims.get_optional_mesh(edp_mesh_names)
+
+    apply_fsdp(
```
tianyu-l: do we not need things like `_apply_fsdp_to_vision_encoder` any more?

torchtitan/models/qwen3_5/parallelize.py
```diff
@@ -0,0 +1,200 @@
...
+from torchtitan.distributed.tensor_parallel import maybe_enable_async_tp
+from torchtitan.models.llama4.parallelize import apply_fsdp, apply_moe_ep_tp
+from torchtitan.models.qwen3_5.sharding import (
+    set_deltanet_sub_module_sharding,
+    set_vision_encoder_sub_module_sharding,
+)
+from torchtitan.tools.logging import logger
+
+
+def parallelize_qwen3_5(
```
tianyu-l, line 63: We should use config-based parallelization.

It might be easier if you wait until full_dtensor PR lands and only support full_dtensor (error not if not full_dtensor). cc @fegin 

(Or wait further until spmd_types is done.)

torchtitan/models/qwen3_5/vision_encoder.py
```diff
@@ -14,22 +15,21 @@
...
-from torchtitan.protocols.module import Module, ModuleDict, ModuleList
+from torchtitan.protocols.module import Module, ModuleDict
 
 LayerNorm = Module.from_nn_module(nn.LayerNorm)
 GELU = Module.from_nn_module(nn.GELU)
 
 _compiled_create_block_mask = torch.compile(create_block_mask)
 
 
-def get_vision_block_mask_mod(num_patch: torch.Tensor, max_num_patch: int):
```
tianyu-l, line 23: Is `max_num_patch` not used prior to this PR (and thus a bug)?

torchtitan/models/qwen3_5/vision_encoder.py
```diff
@@ -356,48 +299,44 @@ class VisionAttention(Module):
...
     def __init__(
-        self, dim: int, n_heads: int, *, qkv: Linear.Config, proj: Linear.Config
+        self,
+        dim: int,
+        num_heads: int,
+        *,
+        wq: Linear.Config,
+        wk: Linear.Config,
+        wv: Linear.Config,
+        proj: Linear.Config,
```
tianyu-l: why they are passed in, instead of being put in config?

torchtitan/models/qwen3_5/vision_encoder.py
```diff
@@ -419,50 +358,52 @@ class VisionTransformerBlock(Module):
...
+        num_heads: int,
         layer_norm_eps: float,
         *,
-        attn_qkv: Linear.Config,
+        attn_wq: Linear.Config,
+        attn_wk: Linear.Config,
+        attn_wv: Linear.Config,
         attn_proj: Linear.Config,
         mlp_fc1: Linear.Config,
         mlp_fc2: Linear.Config,
```
tianyu-l: similar

torchtitan/models/qwen3_5/vision_encoder.py
```diff
@@ -419,50 +358,52 @@ class VisionTransformerBlock(Module):
...
     ):
         super().__init__()
         self.norm1 = LayerNorm(dim, eps=layer_norm_eps)
         self.norm2 = LayerNorm(dim, eps=layer_norm_eps)
-        self.attn = VisionAttention(dim, n_heads, qkv=attn_qkv, proj=attn_proj)
+        self.attn = VisionAttention(
```
tianyu-l: why not build from config?

torchtitan/models/qwen3_5/vision_encoder.py
```diff
@@ -616,40 +538,28 @@ def forward(
...
         attention_mask = _compiled_create_block_mask(
             mask_mod,
             num_vision,
             None,
             max_num_patch,
             max_num_patch,
-            device=hidden_states.device,
+            device=x.device,
         )
-        deepstack_features = []
```
tianyu-l, line 612: oh does Qwen3.5 not use deepstack anymore?

torchtitan/models/qwen3_5/__init__.py
```diff
@@ -0,0 +1,1056 @@
...
+    "122B-A10B": _122b_a10b_qwen35,
+    "397B-A17B": _397b_a17b_qwen35,
+}
+
+
+def model_registry(
+    flavor: str,
+    attn_backend: str = "sdpa",
+    moe_comm_backend: str | None = None,
+    quantization: list[QuantizationConverter.Config] | None = None,
```
tianyu-l: need to update: we updated this to `converters` to support LoRA

torchtitan/models/qwen3_5/__init__.py
```diff
@@ -0,0 +1,1056 @@
...
+            temporal_patch_size=2,
+            spatial_merge_size=2,
+            out_hidden_size=3072,
+            num_position_embeddings=2304,
+        ),
+        mrope_section=[11, 11, 10],
+    )
+
+
+def _397b_a17b_qwen35(
```
tianyu-l: the `_qwen35` suffix sounds redundant

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1045 @@
...
+    head expansion for grouped linear attention internally so that
+    repeat_interleave runs on local tensors under TP.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        # "fla_chunked": parallel within chunks, fast for training (default)
+        # "fla_fused_recurrent": token-by-token, lower memory for long sequences
+        # "torch_naive": pure-Python reference, for numerical testing only
+        backend: str = "fla_chunked"
```
tianyu-l: type should be `Literal[...]`

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1045 @@
...
+            dim=config.value_head_dim,
+            eps=config.norm_eps,
+            param_init=config.norm_init,
+        ).build()
+        self.out_proj = Linear.Config(
+            in_features=value_dim,
+            out_features=config.dim,
+            bias=False,
+            param_init=config.out_proj_init,
+        ).build()
```
tianyu-l: The config should be created in this class's Config, not in `__init__`. In particular, the 
GatedDeltaNet.Config doesn't need to host `fla_backend`.

benefit is 
- to be able to swap things like kernel easily
- can specify sharding of each module at config time

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1045 @@
...
+    Each layer uses either full attention (Qwen35Attention) or linear
+    attention (GatedDeltaNet), determined by ``layer_type`` in config.
+    Both types share the same FFN/MoE structure.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        layer_type: str  # "full_attn" or "linear_attn"
+        attention: Qwen35Attention.Config | None = None
+        deltanet: GatedDeltaNet.Config | None = None
```
tianyu-l:
```suggestion
        delta_net: GatedDeltaNet.Config | None = None
```

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1045 @@
...
+            h = self.attn(h, freqs_cis, attention_masks, positions)
+        else:
+            h = self.attn(h)
+        x = x + h
+
+        h = self.ffn_norm(x)
+        if self.moe_enabled:
+            moe_out = self.moe(h)
+            if self.shared_expert_enabled:
+                shared_out = torch.sigmoid(self.shared_gate(h)) * self.shared_ffn(h)
```
tianyu-l: instead of doing this, can we extend https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/config_utils.py#L153 and use existing `shared_expert` inside MoE module?

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1045 @@
...
+    """Hybrid transformer block for Qwen3.5.
+
+    Each layer uses either full attention (Qwen35Attention) or linear
+    attention (GatedDeltaNet), determined by ``layer_type`` in config.
+    Both types share the same FFN/MoE structure.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        layer_type: str  # "full_attn" or "linear_attn"
```
tianyu-l: don't really need this? The config can be built that this block either has attention / deltanet. Refer to how feed_forward vs. moe is selected.

torchtitan/models/qwen3_5/parallelize.py
```diff
@@ -0,0 +1,200 @@
...
+    # Applies to full attention layers only — GatedDeltaNet is recurrent
+    # and allgathers the full sequence via cp=Replicate() in sharding.
+    if parallel_dims.cp_enabled:
+        cp_mesh = parallel_dims.get_mesh("cp")
+        full_attn_inner_modules = [
+            block.attn.inner_attention  # pyrefly: ignore [missing-attribute]
+            for block in model.layers.values()  # pyrefly: ignore [not-callable]
+            if block.layer_type == "full_attn"  # pyrefly: ignore [missing-attribute]
+        ]
+        if full_attn_inner_modules:
```
tianyu-l: I didn't see an "else" here -- how are you handling sharded activation on linear attention layers?

torchtitan/models/qwen3_5/parallelize.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    model_compile_enabled = (
+        compile_config.enable and "model" in compile_config.components
+    )
+
+    # Context Parallel: wrap inner attention forward BEFORE TP so CP logic
+    # runs inside the local_map boundary on local tensors.
+    # Applies to full attention layers only — GatedDeltaNet is recurrent
+    # and allgathers the full sequence via cp=Replicate() in sharding.
+    if parallel_dims.cp_enabled:
```
tianyu-l, line 87: Since CP is non-trivial, let's just raise NotImplementedError
https://www.internalfb.com/metamate/M4978C

torchtitan/models/qwen3_5/model.py
shuhuayu: make sense, submodule configs are moved to module.config.

torchtitan/models/qwen3_5/model.py
shuhuayu: good point, this is redundant since `attention` and `delta_net` already indicates this.

torchtitan/models/qwen3_5/model.py
shuhuayu: applied.

torchtitan/models/qwen3_5/model.py
shuhuayu: extended in the `common/config_utils.py`. Currently only `qwen3_5` uses this `sigmoid` gate, but this is a simple extension can be used later.

torchtitan/models/qwen3_5/vision_encoder.py
shuhuayu: yes, this was a bug.

torchtitan/models/qwen3_5/parallelize.py
shuhuayu: Ok, let me use the non full-dtensor approach for now.

torchtitan/models/qwen3_5/parallelize.py
shuhuayu: agreed.

torchtitan/models/qwen3_5/parallelize.py
shuhuayu: We used `Replicate()` for that. but as discussed in a previous thread, current cp is inefficient and beat the purpose of supporting it. cp is removed for now.

torchtitan/models/qwen3_5/parallelize.py
shuhuayu: This was previously handled in `fully_shard(model, **fsdp_config)`, but as you said we should separate it.  Apply fsdp to vision encoder and treat vit as a single unit.

torchtitan/models/qwen3_5/README.md
shuhuayu: you are right. what the script did is just near identical numerically.

torchtitan/models/qwen3_5/sharding.py
shuhuayu: sounds great, refactored to the style used in groupedexperts.

torchtitan/models/qwen3_5/vision_encoder.py
shuhuayu: refactored.

torchtitan/models/qwen3_5/vision_encoder.py
shuhuayu: refactored.

torchtitan/models/utils.py
shuhuayu: correct, removed.

shuhuayu (commented):
Left a TODO on `conv1d` waiting for dtensor support in https://github.com/pytorch/pytorch/pull/186129

torchtitan/models/common/moe.py
```diff
@@ -320,6 +320,7 @@ class Config(Module.Config):
         router: TokenChoiceTopKRouter.Config
         load_balance_coeff: float | None = 1e-3
         shared_experts: FeedForward.Config | None = None
+        shared_expert_gate: Module.Config | None = None
```
tianyu-l:
```suggestion
        shared_experts_gate: Module.Config | None = None
```

torchtitan/models/common/moe_sharding.py
```diff
@@ -225,6 +225,14 @@ def set_moe_sharding_config(
             enable_ep=enable_ep, enable_sp=enable_sp
         )
 
+    if getattr(moe_cfg, "shared_expert_gate", None) is not None:
```
tianyu-l: why do we need getattr? It seems always existing (could be None)

torchtitan/models/common/moe.py
```diff
@@ -447,6 +453,10 @@ def _generate_routing_map(
 
         out_BLD = out_TD.view(B, L, D)
         if shared_out_BLD is not None:
+            if self.shared_expert_gate is not None:
+                shared_out_BLD = (
+                    torch.sigmoid(self.shared_expert_gate(x_BLD)) * shared_out_BLD
+                )
```
tianyu-l: What's the behavior under TP?
We used to assume on TP mesh shared_out_BLD is Partial, now there will be more collectives??
If TP is not supposed to be used (DP, EP only) as it's not efficient, then in sharding annotation, don't annotate / support TP.

torchtitan/models/qwen3_5/sharding.py
```diff
@@ -0,0 +1,272 @@
...
+        in_src_shardings={"x": dense_activation_placement(tp=Shard(1))},
+        in_dst_shardings={"x": dense_activation_placement(tp=Replicate())},
+        out_dst_shardings=dense_activation_placement(tp=Shard(1)),
+    )
+
+
+def set_deltanet_conv1d_sharding(deltanet_module) -> None:
+    """Set sharding on GatedDeltaNet sub-modules built inline.
+
+    Conv1d modules don't have Config fields, so their sharding must be
```
tianyu-l: Could you do similar things as in https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/nn_modules.py

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1003 @@
...
+from torchtitan.models.common.rope import apply_rotary_emb_cos_sin
+from torchtitan.models.utils import get_moe_model_nparams_and_flops
+from torchtitan.protocols.module import Module
+
+from .sharding import set_qwen35_sharding_config
+from .vision_encoder import Qwen35VisionEncoder
+
+
+class _Conv1d(nn.Conv1d, Module):
+    pass
```
tianyu-l: put in https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/nn_modules.py

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1003 @@
...
+
+try:
+    from fla.ops.gated_delta_rule import (
+        chunk_gated_delta_rule as _fla_chunk_gated_delta_rule,
+        fused_recurrent_gated_delta_rule as _fla_fused_recurrent_gated_delta_rule,
+    )
+
+    _HAS_FLA = True
+except ImportError:
+    _HAS_FLA = False
```
tianyu-l: I think it doesn't make sense to run this model with FLA. Let's put this in model specific requirements.txt and in CI.

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1003 @@
...
+
+
+def _l2norm(x: torch.Tensor, dim: int = -1, eps: float = 1e-6) -> torch.Tensor:
+    """L2 norm using rsqrt(sum(x²) + eps), not x/max(norm, eps) like F.normalize, to match FLA kernel."""
+    return x * torch.rsqrt((x * x).sum(dim=dim, keepdim=True) + eps)
+
+
+def _torch_naive_gated_delta(
```
tianyu-l: maybe native, not naive

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,1003 @@
...
+            out = F.conv1d(
+                x.to_local(),
+                w,
+                None,
+                conv.stride,
+                conv.padding,
+                conv.dilation,
+                local_groups,
+            )
+            x = DTensor.from_local(out, mesh, plc, run_check=False)
```
tianyu-l: use local_map, not to_local / from_local
specify gradient placement

torchtitan/trainer.py
```diff
@@ -600,10 +599,19 @@ def post_dataloading_process(
         positions = extra_inputs.pop("positions", None)
 
         if isinstance(self.model_config, Decoder.Config):
-            attn_config = self.model_config.layers[0].attention
-            inner_attention = attn_config.inner_attention
+            attn_config = next(
+                (
+                    l.attention
+                    for l in self.model_config.layers
+                    if getattr(l, "attention", None) is not None
```
tianyu-l: can be simplified... to just use `getattr`

torchtitan/models/common/decoder.py
```diff
@@ -109,7 +109,18 @@ def update_from_config(
 
             tp = parallelism.tensor_parallel_degree
             if tp > 1:
-                attention = self.layers[0].attention
+                attention = next(
+                    (
+                        l.attention
+                        for l in self.layers
+                        if getattr(l, "attention", None) is not None
```
tianyu-l: same

tianyu-l: given how frequent this is used, we probably should create a `property` in Decoder config to compute this.

torchtitan/models/common/moe_sharding.py
shuhuayu: indeed. this one and a pre-existing `getattr` are removed.

torchtitan/models/common/moe.py
shuhuayu: more accurate. the hf keys remain unchanged as `shared_expert_gate`.

torchtitan/models/common/moe.py
shuhuayu: you are right, when tp is on and shared experts are used, `Dtensor` does not know we have already gathered from `Shard(1)` for the experts computation itself so it will do it twice and thus waste one collection. I redesigned the `shared_experts` module which now inherits from `FeedForward`.

shuhuayu: actually when tp is on, there are two duplicated all-gather for `w1` and `w3`, which seems to me unnecessary. i rewrite it so one all gather for three: `w1`, `w3`, and optional gate.

torchtitan/models/qwen3_5/model.py
shuhuayu: maybe you are saying it doesn't make sense to run it without `FLA`? added the dependency in `.ci/docker/requirements-vlm.txt`.

torchtitan/models/common/moe.py
```diff
@@ -292,6 +292,30 @@ def forward(
         )
 
 
+class SharedExperts(FeedForward):
```
tianyu-l: Given the gate thing is very much qwen3_5 specific, I would put this in qwen3_5 folder for now, and all other models still use FeedForward.

torchtitan/models/common/moe.py
```diff
@@ -319,7 +343,7 @@ class Config(Module.Config):
         experts: GroupedExperts.Config
         router: TokenChoiceTopKRouter.Config
         load_balance_coeff: float | None = 1e-3
-        shared_experts: FeedForward.Config | None = None
+        shared_experts: SharedExperts.Config | None = None
```
tianyu-l: and since it inherits FeedForward, we can keep it unchanged.

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -231,7 +232,7 @@ def _build_dsv3_layers(
                     comm_backend=moe_comm_backend,
                     non_blocking_capacity_factor=non_blocking_capacity_factor,
                 ),
-                shared_experts=make_ffn_config(
+                shared_experts=make_shared_experts_config(
```
tianyu-l: only do this to qwen3_5 shared experts

torchtitan/models/qwen3_5/sharding.py
```diff
@@ -0,0 +1,264 @@
...
+        Qwen35Attention,
+        Qwen35Model,
+        Qwen35TransformerBlock,
+    )
+
+_REPLICATE_PARAM = dense_param_placement(tp=Replicate())
+_REPLICATE_STATE = ShardingConfig(
+    state_shardings={"weight": _REPLICATE_PARAM, "bias": _REPLICATE_PARAM}
+)
+_REPLICATE_ACT = dense_activation_placement(tp=Replicate())
```
tianyu-l: not sure if we should share reference among all usages

torchtitan/models/common/moe.py
```diff
@@ -292,6 +292,30 @@ def forward(
...
+
+    def __init__(self, config: Config):
+        super().__init__(config)
+        self.gate = config.gate.build() if config.gate is not None else None
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        out = super().forward(x)
+        if self.gate is not None:
+            # TODO: make the gate activation configurable (e.g. softmax, silu)
+            out = torch.sigmoid(self.gate(x)) * out
```
tianyu-l: self.gate is Replicate
x is sharded
self.gate(x) is sharded -> replicate
out is partial -> final outcome is Partial

sounds correct.

torchtitan/models/common/moe_sharding.py
```diff
@@ -214,17 +214,39 @@ def set_moe_sharding_config(
...
+    # it (their per-linear input redistributions become no-ops). w2 (rowwise)
+    # keeps the output Partial; the Partial->sp_layout reduce happens once at
+    # the MoE boundary.
+    shared = moe_cfg.shared_experts
+    if shared is not None:
+        sp_layout = Shard(1) if enable_sp else Replicate()
+        shared_input_layout = Replicate() if not enable_ep else sp_layout
+        shared.sharding_config = ShardingConfig(
+            in_src_shardings={"x": dense_activation_placement(tp=shared_input_layout)},
+            in_dst_shardings={"x": dense_activation_placement(tp=Replicate())},
```
tianyu-l: This is worth fixing even if we split up FeedForward and SharedExperts

tianyu-l (commented):
not sure how popular the shared experts gate would be, so would like to stay conservative

torchtitan/models/qwen3_5/model.py
tianyu-l: since it's in the requirements, can we remove such check, or put the raise here -- if one wants to run qwen3_5, they need to install fla, regardless of if they intend to use native impl or not

shuhuayu:
Thanks for all the comments/suggestions! Some updates: 1) refactored `mrope` for per layer rope and moved its position building into dataloader, and add `mrope_positions` to trainer's `input_dict` in the `extra_kwargs` part for all pp stages. 2) refactored to spmd types. 3) refactored `sharedexperts` so it is now only in qwen3_5. 4) redid numerical tests which still passed.

torchtitan/trainer.py
```diff
@@ -586,12 +586,13 @@ def post_dataloading_process(
         extra_kwargs: dict[str, Any] = {}
 
         positions = extra_inputs.pop("positions", None)
+        mrope_positions = extra_inputs.pop("mrope_positions", None)
```
tianyu-l: No, this is model detail, shouldn't be exposed in trainer.

torchtitan/models/utils.py
tianyu-l: remove

torchtitan/models/common/decoder.py
```diff
@@ -79,7 +79,31 @@ class Config(BaseModel.Config):
...
+            # first layer that carries RoPE to expose the model context length.
+            for layer_cfg in self.layers:
+                attention_cfg = getattr(layer_cfg, "attention", None)
+                rope_cfg = getattr(attention_cfg, "rope", None)
+                if rope_cfg is not None:
+                    return rope_cfg.max_seq_len
+            raise ValueError("Decoder config does not define RoPE max_seq_len.")
+
+        @property
+        def first_attn_config(self) -> BaseAttention.Config | None:
```
tianyu-l:
```suggestion
        def first_attention(self) -> BaseAttention.Config | None:
```

torchtitan/models/common/decoder.py
```diff
@@ -116,7 +140,11 @@ def update_from_config(
 
             tp = parallelism.tensor_parallel_degree
             if tp > 1:
-                attention = self.layers[0].attention
+                attention = self.first_attn_config
+                if attention is None:
+                    raise ValueError(
+                        "No layer with attention config found for TP validation."
+                    )
```
tianyu-l: why raise, no Attention means all-good?

torchtitan/models/common/decoder.py
```diff
@@ -280,7 +308,10 @@ def get_attention_masks(
         self,
         positions: torch.Tensor,
     ) -> AttentionMasksType:
-        attn_config = self.config.layers[0].attention
+        attn_config = self.config.first_attn_config
+        assert (
+            attn_config is not None
+        ), "get_attention_masks requires an attention layer"
```
tianyu-l: similar, no attention sounds fine? E.g. some single pipeline stage only has DeltaNet module

torchtitan/models/qwen3_5/parallelize.py
```diff
@@ -0,0 +1,216 @@
...
+        cpu_offload=training.enable_cpu_offload,
+        reshard_after_forward_policy=parallelism.fsdp_reshard_after_forward,
+        ep_degree=parallel_dims.ep,
+        edp_mesh=edp_mesh,
+    )
+
+    logger.info("Applied fully_shard to the Qwen3.5 model")
+
+    if training.enable_cpu_offload:
+        logger.info("Applied CPU Offloading to the Qwen3.5 model")
```
tianyu-l: remove

torchtitan/models/qwen3_5/parallelize.py
```diff
@@ -0,0 +1,216 @@
...
+            num_layers,
+            input_weight,
+            output_weight,
+        ) = _get_pipeline_metadata(parallel_dims, parallelism, model_config)
+        fqn_per_part = _generate_llm_fqn_per_model_part(
+            num_virtual_stages, num_layers, input_weight, output_weight
+        )
+        # Vision encoder lives on the first stage alongside tok_embeddings
+        if hasattr(model, "vision_encoder") and model.vision_encoder is not None:
+            fqn_per_part[0].insert(0, "vision_encoder")
```
tianyu-l, line 202: not sure how heavy `vision_encoder` it is, maybe worth investigating if we should adjust `parallelism.pipeline_parallel_first_stage_less_layers` later

torchtitan/models/qwen3_5/model.py
```diff
@@ -0,0 +1,782 @@
...
+    class Config(Decoder.Config):
+        vision_encoder: Qwen35VisionEncoder.Config
+
+        def update_from_config(
+            self,
+            *,
+            config,
+            **kwargs,
+        ) -> None:
+            Decoder.Config.update_from_config(self, config=config, **kwargs)
```
tianyu-l, line 552: move this to bottom after https://github.com/pytorch/torchtitan/pull/3595, whichever lands first @wwwjn

torchtitan/models/qwen3_5/parallelize.py
shuhuayu: added comments to reflect this forward-looking point.

torchtitan/experiments/graph_trainer/trainer.py
```diff
@@ -146,7 +144,7 @@ def forward_backward_step(
             labels,
             global_valid_tokens,
             params,
-            extra_inputs,
+            {},
```
tianyu-l: maybe kill this field

torchtitan/trainer.py
```diff
@@ -561,40 +561,32 @@ def post_dataloading_process(
...
-        positions = extra_inputs.pop("positions", None)
+        positions = extra_kwargs.get("positions", None)
 
         # positions and attention_masks are optional (Decoder.forward defaults
         # both to None). Build attention masks only for the masked backends
         # (Flex/Varlen), which is where get_attention_masks is defined. A
         # maskless backend (e.g. the SDPA config used by the graph_trainer
         # tests) still receives positions for RoPE but no masks — it relies on
         # is_causal instead.
-        if isinstance(self.model_config, Decoder.Config) and positions is not None:
```
tianyu-l, line 596: `positions is not None` lost

torchtitan/trainer.py
tianyu-l: don't need this line

torchtitan/trainer.py
```diff
@@ -561,40 +561,32 @@ def post_dataloading_process(
...
-        Note:
-            The distinction between extra_inputs and extra_kwargs is important for
-            pipeline parallelism: extra_kwargs are forwarded to all pipeline stages,
-            while extra_inputs are only available to the first stage. Positions
-            always go into extra_kwargs so every stage can apply RoPE correctly.
+                - extra_kwargs: Additional keyword arguments for the model forward
+                    (e.g. positions, attention_masks), forwarded to every
+                    pipeline-parallel stage.
         """
         inputs = input_dict["input"]
```
tianyu-l, line 571: `inputs` and `labels` are really not special and IMO not worth special handling, except for how `labels` is involved in loss computation. Can delay the general refactor to later.

torchtitan/components/validate.py
tianyu-l: same

tianyu-l (approved):

shuhuayu:
thanks for the careful and sharp reviews! let's merge it to avoid more refactors and iterate later for bugs/features.

PR #3384, opened 2026-05-17
[MoE] Move shared_experts out of combine() for clean DTensor boundaries

acisseJZhong:
* (to be filled)

Move shared_experts computation from inside TokenDispatcher.combine()
to MoE.forward(), so that:
- Router stays on DTensor (via NoParallel)
- Shared experts stay on DTensor (via ColwiseParallel/RowwiseParallel)
- Only routed experts convert DTensor→local at GroupedExperts boundary

This removes the DTensor→local→DTensor round-trip that shared experts
previously went through, and simplifies the combine() interface by
removing the shared_experts parameter from all token dispatchers.

Trade-off: this removes the async overlap between shared_experts and
the all-to-all combine communication. Previously shared_experts ran
while the combine a2a was in flight on the NCCL stream. A follow-up
can restore overlap using CUDA streams.

### DIFF
--- a/tests/unit_tests/test_expert_parallel.py
+++ b/tests/unit_tests/test_expert_parallel.py
@@ -7,7 +7,6 @@
 import unittest
 
 import torch
-import torch.nn.functional as F
 
 from torchtitan.models.common.token_dispatcher import AllToAllTokenDispatcher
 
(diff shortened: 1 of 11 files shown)


acisseJZhong:
Superseded by #3386

PR #3386, opened 2026-05-17
[MoE][5/n] Refactor MoE to clean DTensor boundaries for shared/routed experts

acisseJZhong:
**Copying major part from @fegin's PR https://github.com/pytorch/torchtitan/pull/3160** 

* #3389
* __->__ #3386
* __->__ #3167

image

https://excalidraw.com/#json=2abKr0m2s26fc6lyoF9Qq,MqMzUIoXWYJIfckHNOB7Sw

## Summary

Refactor MoE parallelization to establish clean DTensor boundaries:
- **Router**: gate weights Replicate, input all-gathered to Replicate (via `NoParallel`), compute on DTensor, output converted to local via `to_local(grad_placements=Partial)`
- **Shared experts**: weights TP-sharded (via `ColwiseParallel`/`RowwiseParallel`), input all-gathered to Replicate inside `ColwiseParallel`, compute on DTensor, output local via `to_local` (Partial placement, `use_local_output=True`)
- **Routed experts**: input converted to local via `to_local` at `GroupedExperts` boundary, EP-partitioned weights on local tensors, dispatch/compute/combine all local

Key changes:
- Move `shared_experts` computation from inside `TokenDispatcher.combine()` to `MoE.forward()`
- Move `to_local` from `MoE.forward()` into `GroupedExperts.forward()`, with proper `grad_placements` based on placement type (Shard vs Replicate)
- Replace `ColwiseParallelWithGradPlacement` with standard `ColwiseParallel` using `input_layouts`
- Remove `shared_experts` parameter from all token dispatchers, replace with `initial_output` buffer
- For EP+SP, keep `Shard(1)` as desired MoE input layout (avoid redundant all-gather at MoE boundary); for TP-only, keep `Replicate()` (required for TP-sharded expert weight reduction)

### Trade-offs
- Removes async overlap between `shared_experts` and the all-to-all combine. Previously `shared_experts` ran while the combine A2A was in flight on the NCCL stream. A follow-up can restore overlap using CUDA streams.

## Test plan
single run
```
NCCL_NVLS_ENABLE=0 NGPU=4 MODULE='llama4' CONFIG='llama4_debugmodel' ./run_train.sh \
    --parallelism.pipeline_parallel_degree 1 \
    --parallelism.data_parallel_shard_degree 2 \
    --parallelism.expert_parallel_degree 2 \
    --debug.deterministic --debug.seed=42
```
Loss exactly match for dp4tp1ep1. 
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 4 --parallelism.tensor_parallel_degree 1 --
parallelism.expert_parallel_degree 1 --parallelism.no-ena
ble_sequence_parallel" --baseline-ngpus=4 --test-module='
llama4' --test-config='llama4_debugmodel' --test-options=
"--parallelism.pipeline_parallel_degree 1 --parallelism.d
ata_parallel_shard_degree 4 --parallelism.tensor_parallel
_degree 1 --parallelism.expert_parallel_degree 1 --parall
elism.no-enable_sequence_parallel" --test-ngpus=4 --asser
t-equal --no-seed-checkpoint
```

For other settings, loss is expected to diverge compared to main due to different reduction pattern, and shared expert computation changes place.

DP2TP2EP2, dense part default SP2
image

DP2TP2EP4, dense part SP off
image

tested deepep with both SP on and off
image


torchtitan/models/gpt_oss/parallelize.py
```diff
@@ -51,9 +38,6 @@ def parallelize_gptoss(
         ({parallel_dims.tp}) and 2 * CP degree ({parallel_dims.cp}).
         """
 
-    if parallelism.full_dtensor:
-        raise NotImplementedError("full_dtensor is not supported yet.")
```
tianyu-l, line 55: I don't think we can delete this

torchtitan/protocols/sharding.py
```diff
@@ -132,7 +122,6 @@ class ShardingConfig:
     in_dst_shardings: dict[str, NamedPlacement] | None = None
     out_src_shardings: NamedPlacement | tuple[NamedPlacement, ...] | None = None
     out_dst_shardings: NamedPlacement | None = None
-    local_input_grad_placements: dict[str, NamedPlacement] | None = None
     local_output_grad_placements: NamedPlacement | None = None
```
tianyu-l: hope we can remove this as well; please see if my comment makes sense to you 
https://github.com/pytorch/torchtitan/pull/3159#discussion_r3253875475

torchtitan/models/gpt_oss/expert_parallel.py
tianyu-l, line 1: can we also delete https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/expert_parallel.py

torchtitan/models/common/moe.py
```diff
@@ -72,26 +71,35 @@ def _experts_forward(
             x.bfloat16(), w3.bfloat16().transpose(-2, -1), offs=offsets
         )
         return torch._grouped_mm(
-            h, w2.bfloat16().transpose(-2, -1), offs=offsets
+            +h, w2.bfloat16().transpose(-2, -1), offs=offsets
```
tianyu-l: accidental?

torchtitan/models/common/moe.py
```diff
@@ -72,26 +71,35 @@ def _experts_forward(
...
+        return self.token_dispatcher.combine(routed_output, metadata, x)
+
+    def parallelize(self, parallel_dims) -> None:
+        """Parallelize expert weights, then wire EP/TP meshes on the dispatcher
+        so dispatch/combine see the right meshes at runtime."""
+        super().parallelize(parallel_dims)
+        self.token_dispatcher.wire_meshes(
+            ep_mesh=parallel_dims.get_optional_mesh("ep"),
+            tp_mesh=parallel_dims.get_optional_mesh("tp"),
+        )
```
tianyu-l, line 105: Looks OK for now. Add a TODO for @pianpwk :

With spmd_types and `set_current_mesh`, we should move them to `current_mesh` calls inside `AllToAllTokenDispatcher` and `DeepEPTokenDispatcher`

torchtitan/models/gpt_oss/sharding.py
```diff
@@ -17,38 +17,63 @@
...
 
+# Routed-expert layout for ``GptOssGroupedExperts`` (mlp1/mlp2 fused
+# weights + biases). Mirrors ``GptossExpertTensorParallel`` /
+# ``GptossTensorParallel`` partition_fns: mlp1_weight/bias colwise,
+# mlp2_weight rowwise, mlp2_bias replicated.
+_GPT_OSS_EXPERTS_PARAM_LAYOUT: dict[str, Placement] = {
+    "mlp1_weight": Shard(1),
+    "mlp1_bias": Shard(1),
+    "mlp2_weight": Shard(2),
+    "mlp2_bias": Replicate(),
```
tianyu-l, line 35: IIRC this should be init to `Partial` / `spmd.P` to avoid implicit redistribution. Add a TODO for @pianpwk

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+    redistributed to sp_layout. MoE.forward() operates on DTensors —
+    the DTensor→local conversion happens at GroupedExperts boundary.
+    """
+    sp_layout: Placement = Shard(1) if enable_sp else Replicate()
+    moe_desired_input_layouts = Replicate() if not enable_ep else sp_layout
+
+    return ShardingConfig(
+        state_shardings={
+            "expert_bias": dense_param_placement(tp=Replicate()),
+            "tokens_per_expert": dense_param_placement(
```
tianyu-l: DP / CP should always be `Partial`, regardless of EP enabled or not?

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+            out_dst_shardings=dense_activation_placement(tp=Shard(0)),
+        )
+    else:
+        return ShardingConfig(
+            state_shardings=state,
+            out_dst_shardings=dense_activation_placement(tp=Replicate()),
+        )
+
+
+def _moe_wrapper_sharding_config(*, enable_ep: bool, enable_sp: bool) -> ShardingConfig:
```
tianyu-l:
```suggestion
def _moe_sharding_config(*, enable_ep: bool, enable_sp: bool) -> ShardingConfig:
```
or `_moe_module_sharding_config`

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+    """``ShardingConfig`` at the MoE wrapper boundary.
+
+    Mirrors ``PrepareModuleInputOutput``: input arrives at sp_layout,
```
tianyu-l: remove reference to `PrepareModuleInputOutput`

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+        ep_enabled: Whether expert parallelism is enabled.
+        enable_sp: Whether sequence parallelism is enabled (affects the
+            wrapper's enter/exit TP layout).
+        expert_param_layout: ``{param_name: tp_placement}`` for the
+            routed experts' weight params (used on the EP-disabled +
+            TP-enabled path).
+    """
+    # MoE wrapper boundary. Set whenever TP is enabled (with or without
+    # SP) -- the wrapper still needs to convert DTensor inputs to local.
+    if enable_tp:
```
tianyu-l: There shouldn't be this flag? TP applied or not shouldn't affect the `sharding_config`; TP annotations will be stripped out during mesh resolution at runtime.

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+                "selected_experts_indices": experts_in_layout,
+            },
+            out_src_shardings=experts_out_layout,
+            out_dst_shardings=experts_out_layout,
+            local_map=LocalMapConfig(
+                in_grad_placements=(
+                    experts_in_grad_layout,
+                    experts_in_grad_layout,
+                    experts_in_grad_layout,
+                ),
```
tianyu-l: nit: it seems here the grad placements are always the same as input placements, so maybe we can let that be default behavior to simplify the code. With spmd_types, we shouldn't need to worry about it any way.

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+            ),
+        )
+    elif enable_tp:
```
tianyu-l: similarly, this should be `else:`

O/w when TP or EP not applied, it doesn't have any sharding.

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+            )
+            moe_cfg.shared_experts.w2.sharding_config = _shared_expert_rowwise_config()
+            moe_cfg.shared_experts.w3.sharding_config = _shared_expert_colwise_config(
+                enable_ep=enable_ep, enable_sp=enable_sp
+            )
+
+    # Routed experts: local_map converts DTensor inputs to local for
+    # dispatch/compute/combine, then wraps local output as DTensor(Partial).
+    experts_out_layout = dense_activation_placement(tp=Partial())
+    if enable_ep:
```
tianyu-l, line 229: I think we can share most of the code in two branches below

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,267 @@
...
+        },
+        out_src_shardings=dense_activation_placement(tp=Partial()),
+        out_dst_shardings=dense_activation_placement(tp=sp_layout),
+    )
+
+
+def set_moe_sharding_config(
+    moe_cfg,
+    *,
+    enable_tp: bool,
```
tianyu-l: I don't think we need this flag

tianyu-l (commented):
Makes a lot of sense to me now.

In addition to inline comments:
1. Please share numerics comparison. In particular, the numerics should be bitwise at least when we apply TP only (no SP, no EP).
2. Your pictures are super cool. I think it's worth putting them in the repo, to help others understand and to remind ourselves. E.g. we can have a `models/common/MOE_SHARDING.md` referencing those pics. WDYT?

torchtitan/models/gpt_oss/sharding.py
acisseJZhong: for gptoss, this happens in experts forward(local region) and it is initialized to Replicate because we scale bias here https://github.com/pytorch/torchtitan/blob/cb7bf09aba8625f0ee873daa17e842879cb9a981/torchtitan/models/gpt_oss/moe.py#L20

for normal moe, we also annotated bias to be Replicate and the comp happens for shared experts(dtensor region). https://fburl.com/30eaoyic, I assume you mean we should change to Partial bias to avoid the implicit redistribution in addmm?

torchtitan/models/common/moe_sharding.py
acisseJZhong: yes nice catch. I think otherwise the redistribute https://github.com/pytorch/torchtitan/blob/cb7bf09aba8625f0ee873daa17e842879cb9a981/torchtitan/components/optimizer.py#L483 will miss the reduction / noop.

also need to fix in optimizer.py to maintain the non full dtensor path where `tokens_per_expert` becomes a DTensor on the TP-only mesh.

torchtitan/models/common/moe_sharding.py
acisseJZhong: without this flag, how should we tell if use TP on shared expert or not?

torchtitan/models/common/moe_sharding.py
acisseJZhong: I thought when neither EP nor TP is on, we should have no sharding? 
as indicated by the comment
```
    # else: EP and TP both disabled -- experts stay plain tensors,
    # no sharding_config.
```

torchtitan/models/common/moe_sharding.py
acisseJZhong: it's not the same below? Shard <-> Shard, Replicate <-> Partial

torchtitan/models/gpt_oss/sharding.py
tianyu-l: I think so. Note that the gradient of it should be Replicate, so there will be implicit redistribute in optimizer step, which is fine as it's not in fwd/bwd region.

torchtitan/models/common/moe_sharding.py
tianyu-l: Always use in config, regardless of TP is in the mesh or not, to provide the default. Later at runtime `resolve_mesh` will help filter out the TP sharding, if TP is not enabled.

torchtitan/models/common/moe_sharding.py
tianyu-l: the comment is not correct, see my comment above

torchtitan/models/common/moe_sharding.py
tianyu-l: oh you are right, sorry I misread

acisseJZhong:
> Please share numerics comparison. In particular, the numerics should be bitwise at least when we apply TP only (no SP, no EP).

Tested dp4tp1 loss is same on main. For dp2tp2 with no sp: loss is expected to be different since we change the routed output scatter add and removed shared expert overlap. 

```
 Main: 
  all_reduce(scatter_add(shared_out_partial, routed_partial))

 Your branch:
  all_reduce(scatter_add(zeros_partial, routed_partial) + shared_out_partial)
```

torchtitan/models/gpt_oss/sharding.py
acisseJZhong: do we have the forward Partial backward Replicate guarantee in Dtensor today?

torchtitan/models/common/decoder_sharding.py
```diff
@@ -16,21 +16,28 @@
 TP = MeshAxisName.TP
 
 
-def dense_param_placement(*, tp: Placement) -> NamedPlacement:
+def dense_param_placement(
```
tianyu-l: Maybe we should separate the dense param from tokens_per_expert buffer
- buffer is not a "param"
- only this buffer needs to be special cased

so it's better not to compromise the entire function for just one use case

torchtitan/models/common/moe.py
```diff
@@ -308,7 +319,7 @@ def __init__(self, config: Config):
         self.register_buffer(
             "tokens_per_expert",
             torch.zeros(num_experts, dtype=torch.float32),
-            persistent=False,
+            persistent=True,
```
tianyu-l: no, we shouldn't change this, it will break DCP

torchtitan/models/gpt_oss/sharding.py
tianyu-l: No we don't, so can wait for spmd_types

torchtitan/models/llama4/sharding.py
```diff
@@ -15,37 +15,58 @@
...
+            layer_cfg, enable_sp=enable_sp, enable_tp=enable_tp, enable_ep=enable_ep
+        )
 
 
 def _set_llama4_layer_sharding(
-    layer_cfg: "Llama4TransformerBlock.Config", *, enable_sp: bool
+    layer_cfg: "Llama4TransformerBlock.Config",
+    *,
+    enable_sp: bool,
+    enable_tp: bool,
```
tianyu-l: remove

torchtitan/models/gpt_oss/sharding.py
```diff
@@ -17,38 +17,62 @@
...
+}
+
+
 def set_gpt_oss_sharding_config(
     config: "GptOssModel.Config",
     *,
     loss_parallel: bool,
     enable_sp: bool,
+    enable_tp: bool,
```
tianyu-l: remove

torchtitan/models/gpt_oss/model.py
```diff
@@ -219,6 +219,8 @@ def update_from_config(
                 self,
                 loss_parallel=not parallelism.disable_loss_parallel,
                 enable_sp=parallelism.enable_sequence_parallel,
+                enable_tp=parallelism.tensor_parallel_degree > 1,
```
tianyu-l: remove

torchtitan/models/llama4/model.py
```diff
@@ -178,6 +178,8 @@ def update_from_config(
                 self,
                 loss_parallel=not parallelism.disable_loss_parallel,
                 enable_sp=parallelism.enable_sequence_parallel,
+                enable_tp=parallelism.tensor_parallel_degree > 1,
```
tianyu-l: remove, similar for qwen

torchtitan/components/optimizer.py
```diff
@@ -479,19 +479,38 @@ def _update_expert_bias(
...
+            # so redistribute reduces across all ranks and dp_cp_covered
+            # is True. Under non-full_dtensor only TP/EP are in the mesh,
+            # so DP/CP partial counts need a separate all-reduce.
+            is_dtensor = isinstance(
+                tokens_per_expert_by_layer, torch.distributed.tensor.DTensor
+            )
+            if is_dtensor:
+                dt = tokens_per_expert_by_layer
+                # Reduce across TP/EP (and DP/CP under full_dtensor).
+                dp_cp_covered = dt.device_mesh.ndim > len({"tp", "ep"})
```
tianyu-l: we can use this flag
https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/parallel_dims.py#L35

torchtitan/components/optimizer.py
```diff
@@ -479,19 +479,38 @@ def _update_expert_bias(
...
+                tokens_per_expert_by_layer = dt.redistribute(
+                    placements=[Replicate()] * dt.device_mesh.ndim
+                ).full_tensor()
```
tianyu-l: just `.full_tensor()`, which is redistribute to Replicate + to_local

torchtitan/components/optimizer.py
```diff
@@ -479,19 +479,38 @@ def _update_expert_bias(
         tokens_per_expert_by_layer = torch.vstack(tokens_per_expert_list)
 
         if loss_mesh is not None:
```
tianyu-l: This doesn't sound correct any more.

In the past, we always have replicate buffer per TP mesh (because we were not doing sequence parallel on router). But now, on TP mesh it can also be Partial, so we shouldn't condition on loss_mesh (which consists of DP and CP).

I think we should
- if full_dtensor, redistribute to Replicate
- if not full dtensor, first redistribute to Replicate on TP (if existing), then explicit call all-reduce on loss_mesh.

torchtitan/models/common/moe.py
acisseJZhong: oops yeah forget to change this back during debugging

torchtitan/components/optimizer.py
```diff
@@ -478,20 +478,41 @@ def _update_expert_bias(
...
+            tokens_per_expert_by_layer, torch.distributed.tensor.DTensor
+        )
+        if parallel_dims.full_dtensor:
+            # full_dtensor: DTensor mesh includes all axes (DP/CP/TP/EP).
+            # A single redistribute covers everything.
+            assert is_dtensor
+            dtensor_mesh = tokens_per_expert_by_layer.device_mesh
+            tokens_per_expert_by_layer = tokens_per_expert_by_layer.redistribute(
+                placements=[Replicate()] * dtensor_mesh.ndim
+            )
```
tianyu-l, line 493: This will incur multiple all-reduce, one on each of the SPMD mesh axes. In the long term, we should provide utils to do one all-reduce on the global spmd_mesh. Please add a TODO here.

tianyu-l (approved):
LGTM, great work!

acisseJZhong:
~kept the original `apply_moe_ep_tp`, `ExpertParallel`, `TensorParallel`, `ColwiseParallelWithGradPlacement` since qwen3vl hasn't been migrated to the new config sharding. We could delete them as soon as we migrated qwen3 vl to config based sharding~. cc @shuhuayu @fegin 

updated qwen3vl to use config based sharding, except I kept the original parallelize function for vision encoder `_apply_tp_to_vision_encoder`  https://github.com/pytorch/torchtitan/blob/0f5cd801eac7c95ec42eed40cee8720e50e3ac18/torchtitan/models/qwen3_vl/parallelize.py#L174. I thought it should be better to make each submodule under `Qwen3VLVisionEncoder` configurable and apply config based sharding as a separate PR.

torchtitan/components/optimizer.py
```diff
@@ -478,20 +478,43 @@ def _update_expert_bias(
...
-                    * tokens_per_expert_by_layer.device_mesh.ndim
-                )
-            else:
-                # Perform single all-reduce to get global statistics across all processes
-                pg = loss_mesh.get_group()
+        if parallel_dims.full_dtensor:
+            # full_dtensor: DTensor mesh includes all axes (DP/CP/TP/EP).
+            # redistribute Partial→Replicate covers everything.
+            assert isinstance(
+                tokens_per_expert_by_layer, torch.distributed.tensor.DTensor
```
tianyu-l, line 485: we can simplify this (and below) by importing `DTensor`

torchtitan/distributed/tensor_parallel.py
tianyu-l, line 23: I think we can deprecate this together with qwen3_vl and HF backend refactor.

tianyu-l (commented):
Would be good to show loss compare with EP.

Right now you are comparing
before FSDP2 TP2 vs. after FSDP2 TP2

We can e.g. add one more
after FSDP2 SP2 EP4
which would require loading from the same seed checkpoint (and reshard)

torchtitan/distributed/tensor_parallel.py
acisseJZhong: it's still used in qwen3_vl vision encoder parallelization. 

also used in torchtitan/experiments/rl/models/parallelize.py and torchtitan/experiments/transformers_modeling_backend/parallelize.py, which we can refactor.

tianyu-l: > also used in torchtitan/experiments/rl/models/parallelize.py

Not sure why it's still there; we should delete. @wwwjn

acisseJZhong:
added test for DP2TP2EP2, dense SP2 and DP2TP2EP4, dense SP off.

torchtitan/components/optimizer.py
fegin: Are you able to verify this code path? We don't have any full_dtensor path that will trap into `_update_expert_bias` as we only enable full dtensor for llama3.

fegin (approved):
LGTM, the PR is cleaner than original prototype. 

1. The llama3 full dtensor PR is landed. You can rebase to origin/main and ditch the previous update PR.

2. I like you removing `local_input_grad_placements` and `local_output_grad_placements`. But can you remind me why we can remove them now?

3. Great work to make qwen3_vl EP config based. My original intention is to have another PR targeting on qwen3_vl as well. But this PR works well. I'll take care of the vision parts from my PR.

torchtitan/models/qwen3_vl/parallelize.py
```diff
@@ -280,44 +158,28 @@ def parallelize_qwen3_vl(
...
-        if parallelism.enable_async_tensor_parallel and not model_compile_enabled:
-            raise RuntimeError("Async TP requires torch.compile")
-
-        tp_mesh = parallel_dims.get_mesh("tp")
+    if parallel_dims.cp_enabled:
+        raise NotImplementedError("Context Parallel is not yet supported for Qwen3-VL.")
 
-        # Apply TP to vision encoder
+    if parallel_dims.tp_enabled:
+        # Apply TP to vision encoder (still uses parallelize_module path)
```
fegin: You can add a TODO for me.

acisseJZhong:
> removing local_input_grad_placements

We had local_input_grad_placements for [`ColwiseParallelWithGradPlacement`](https://github.com/pytorch/torchtitan/blob/cf3c4312e59988fb00d1720c53a95143379dcc06/torchtitan/distributed/tensor_parallel.py#L109) used in [shared expert](https://github.com/pytorch/torchtitan/blob/8aa96acb282e959f8ceea5fe4342ebbddc33b745/torchtitan/models/llama4/parallelize.py#L425) to keep d_x partial.

Now it should be covered by [_shared_expert_colwise_config](https://github.com/pytorch/torchtitan/blob/2cd5b38eed51a3ec2100ca80fe602bdcd5b6d437/torchtitan/models/common/moe_sharding.py#L64) and _shared_expert_rowwise_config? (if the incoming gradient is Replicate, and weight is Shard, then d_x is automatically Partial, etc). See the graph [here](https://excalidraw.com/#json=2Km7coiWFw4D27--Hwf1j,aX9InmT4R-ghizMWpZpuCw) for more details.   

> removing local_output_grad_placements

We had local_output_grad_placements for [`NoParallel`](https://github.com/pytorch/torchtitan/blob/cf3c4312e59988fb00d1720c53a95143379dcc06/torchtitan/distributed/tensor_parallel.py#L42 ) to pass in specific output gradient placements, and we used it on [router gate](https://github.com/pytorch/torchtitan/blob/cf3c4312e59988fb00d1720c53a95143379dcc06/torchtitan/models/llama4/parallelize.py#L411 ) to make sure we have forward Replicate(weight) backward Partial (weight gradient). 

Now it should all be covered in the dtensor config based sharding: for example, if we specify the router weight to be [Replicate](https://github.com/pytorch/torchtitan/blob/2cd5b38eed51a3ec2100ca80fe602bdcd5b6d437/torchtitan/models/common/moe_sharding.py#L109), and the router output is Replicate. If the incoming gradient is Partial, then weight gradient is also Partial, etc. The correctness can be further guaranteed through spmd_types.

acisseJZhong:
> Are you able to verify this code path? We don't have any full_dtensor path that will trap into _update_expert_bias as we only enable full dtensor for llama3.

No I haven't run full dtensor mode. This code is only verified in theory. I can work on enabling full dtensor for MoE models if you haven't have a PR for it!

fegin:
@acisseJZhong I have a PR for full dtensor the rest with the some changes to qwen3-vl to make sure everything is config-based. I'll rebase to origin/main once your PR is landed. Thanks!

torchtitan/models/common/moe_sharding.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+    state = {
+        "weight": dense_param_placement(tp=Replicate()),
+        "bias": dense_param_placement(tp=Replicate()),
+    }
+    if enable_ep:
+        return ShardingConfig(
+            state_shardings=state,
+            in_dst_shardings={"input": dense_activation_placement(tp=Shard(0))},
+            out_dst_shardings=dense_activation_placement(tp=Shard(0)),
```
tianyu-l, line 116: This feels wrong in full_dtensor. I think there should be StridedShard because of https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/moe.py#L346

The solution should be moving the `.view(-1, dim)` inside `GroupedExperts`, so that the here it's `tp=Shard(1)`.

acisseJZhong: yeah good point, right now it's fine because in non full dtensor mode resolve_mesh filters to only the TP axis. Let's have a follow up PR to move `.view(-1, dim)` inside `GroupedExperts`, as right now router assumes `(bs*slen, dim)` input shape.

torchtitan/models/qwen3_vl/model.py
```diff
@@ -306,6 +315,8 @@ def _compute_mrope_freqs(
         # --- Compute interleaved MRoPE cos/sin from position IDs ---
 
         freqs_cis = self.freqs_cis
+        if isinstance(freqs_cis, DTensor):
```
acisseJZhong, line 318: the qwen3 vl test is failing https://github.com/pytorch/torchtitan/actions/runs/26136643022/job/76873206317?pr=3386 due to mix of dTensor and tensor because we access the Dtensor freqs_cis with tensor index. I hacked here, but not sure if you have a better workaround. @shuhuayu 

Image

acisseJZhong: aslo when reproducing qwen3vl test locally, I am hitting 
```
RuntimeError: cuDNN Frontend error: [cudnn_frontend] Error: No valid execution plans built.
``` 

wondering do you know how to fix? I am already using cuda 13.0 with pytorch nightly.

shuhuayu: > https://github.com/pytorch/torchtitan/actions/runs/26136643022/job/76873206317?pr=3386 

This is rocm test, is cuda test also failing? i checked locally it works well with latest torch nightly.

acisseJZhong: somehow I wasn't able to find the 8 GPU Model Tests on GPU in CI for this PR, only rocm. I would image this issue being device agnostic, failing across cuda and rocm. However, I failed to repro locally because of the cuDNN frontend error.

shuhuayu: I see. I tested the main locally on 8 h100s and it works. i think in general we should avoid checking `dtensor` in model code, and try our best to move it to `sharding.py/parallelize.py`.

acisseJZhong: now with this hack the issue seems gone. https://github.com/pytorch/torchtitan/actions/runs/26148781436/job/76910702209?pr=3386, but I am not sure if this is the best fix.

acisseJZhong:
the failing test with `RuntimeError: "histogram_cpu" not implemented for 'torch.int64'` is expected to be fixed by https://github.com/pytorch/pytorch/pull/183970 which is already landed in torch.
image

torchtitan/models/qwen3_vl/model.py
tianyu-l, line 325: I think this is because `position_ids` is not converted to DTensor when `full_dtensor=False`?

I think now you can do whatever to unblock, as long as correctness is guaranteed. `freqs_cis` should be Replicate when it's DTensor, so calling to_local() sounds a reasonable thing to do.

acisseJZhong: @shuhuayu l'm gonna leave this formal fix to you! I think we could also keep freqs_cis as a plain tensor during comp, and sharding config will take care of redistributing it to Replicate.

ydwu4:
I did some local verification and it seems this PR causes a CI failure for graph trainer: https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_graph_trainer_h100.yaml?query=branch%3Amain

Can verify if this is the case and if so, revert then fix it?

acisseJZhong:
@ydwu4 are you referring to the failure on `  RuntimeError: "histogram_cpu" not implemented for 'torch.int64'
`? see https://github.com/pytorch/torchtitan/pull/3386#issuecomment-4500009588, I believe it should be fixed after using the new pytorch nightly

fegin:
I would prefer a forward fix (if we need a fix) rather than reverting. I think that's the conclusion of the core change v.s. graph trainer breakage.  We should do a better job to find out the breakage earlier though.

PR #3388, opened 2026-05-17
Add xpugraph pass for graph trainer 

jemitche1:
This PR adds xpugraph support to graph trainer while preserving existing cuda behavior. Also, the related integration test was updated from CUDA-specific description to graph_capture since the test now represents backend-specific graph capture behavior.


SherlockNoMad:
Hi thanks for the contribution. 

But I am afraid we don't have the CI machine with XPU to test this...

SherlockNoMad (changes requested):
No CI machine to test this...

PR #3390, opened 2026-05-17
[graph_trainer] Fix H100 CI failure from DeepEP compilation break

SherlockNoMad:
## Summary
- DeepEP (pinned at commit `1b8f467`) fails to compile against the latest PyTorch nightly due to an ABI break in `ATen/core/stack.h` (`torch::jit::push` and `TuplePacker` removed/renamed). Since the H100 CI script uses `set -eux`, the `pip install` failure aborts the entire workflow — no tests run at all.
- Made the DeepEP installation non-fatal (`|| echo WARNING`) so the remaining H100 tests (DeepSeek-v3 EP, Qwen3, async TP, numerics, precompile, bitwise deterministic) still run.
- Disabled the `aot_fx_trace_deepseek_v3_hybridep` integration test until DeepEP is updated to be compatible with the latest PyTorch nightly.

## Test plan
- [ ] Verify H100 CI workflow passes with DeepEP install failure (HybridEP test skipped, all other tests run)
- [ ] Re-enable HybridEP test once DeepEP publishes a fix for the `ATen/core/stack.h` ABI change

### DIFF
--- a/torchtitan/experiments/graph_trainer/tests/integration_tests.py
+++ b/torchtitan/experiments/graph_trainer/tests/integration_tests.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import argparse
+import importlib
 import os
 
 from tests.integration_tests import OverrideDefinitions
@@ -15,6 +16,8 @@ from tests.integration_tests.run_tests import run_tests
 # triggered by the full DTensor change (#2149). Re-enable once the
 # partitioner issue is resolved.
 _JIT_DISABLED = True
+
+_DEEPEP_AVAILABLE = importlib.util.find_spec("deep_ep") is not None
 
 
 def _build_llama3_tests() -> list[OverrideDefinitions]:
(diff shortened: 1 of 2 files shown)


torchtitan/experiments/graph_trainer/tests/integration_tests.py
```diff
@@ -359,6 +359,9 @@ def _build_deepseek_v3_tests() -> list[OverrideDefinitions]:
             "aot_fx_trace_deepseek_v3_fsdp_tp_ep_full_inductor",
             ngpu=8,
         ),
+        # TODO: Disabled due to DeepEP compilation failure against PyTorch
```
aditvenk: is there any way to disable only if DeepEP installation fails? e.g., the install could set an env var and the disable gates on it?

aditvenk (approved):

torchtitan/experiments/graph_trainer/tests/integration_tests.py
SherlockNoMad: Good idea — updated. The HybridEP test now gates on `_DEEPEP_AVAILABLE = importlib.util.find_spec("deep_ep") is not None`, so it auto-disables when DeepEP fails to install and auto-enables once it compiles again. No env var needed.

tianyu-l (approved):

syed-ahmed:
Looks like https://github.com/pytorch/pytorch/pull/180786 was responsible for the breakage. I have manually tested installation on latest nightly pytorch and DeepEP installs. Also checked latest graph trainer pipeline and looks like deepep is available and graph_trainer + hybridep tests run. The changes in this PR are good though and should stay.

PR #3391, opened 2026-05-18
[rl] Add TITO generator and gen metrics

felipemello1:
This PR does three things:
- Changes generator to receive tokens instead of text (token-in/token-out) by encoding prompts in the controller and passing `tokenized_prompts` into the generator. This is the standard in RL training and avoids retokenization of texts, which leads to all sorts of issues.
- Adds generation_metrics to the generation output.
- Sets wandb=True in our configs

### test
image

[actor=<root>] ============================================================
[actor=<root>] Validation summary (pre / post):
[actor=<root>]   validation/reward/_max:  +1.300  /  +1.300
[actor=<root>]   validation/reward/_mean:  +0.385  /  +0.700
[actor=<root>]   validation/reward/_min:  +0.000  /  +0.300
[actor=<root>]   validation/reward/_std:  +0.541  /  +0.490
[actor=<root>]   validation/reward/_sum:  +7.700  /  +14.000
[actor=<root>]   validation/reward/component/correctness/mean:  +0.250  /  +0.400
[actor=<root>]   validation/reward/component/format/mean:  +0.135  /  +0.300
[actor=<root>] ============================================================

image

### DIFF
--- a/torchtitan/experiments/rl/actors/generator.py
+++ b/torchtitan/experiments/rl/actors/generator.py
@@ -24,6 +24,7 @@ from torchtitan.experiments.rl.models.vllm_registry import (
     registry_to_vllm,
     TORCHTITAN_CONFIG_FORMAT,
 )
+from torchtitan.experiments.rl.observability import metrics as m
 from torchtitan.experiments.rl.types import Completion
 from torchtitan.observability import structured_logger as sl
 from torchtitan.protocols.model_spec import ModelSpec
@@ -32,10 +33,69 @@ from torchtitan.tools.utils import has_cuda_capability
 from vllm import EngineArgs, LLMEngine, SamplingParams
 from vllm.config import AttentionConfig, CompilationConfig
 from vllm.distributed.parallel_state import cleanup_dist_env_and_memory
+from vllm.outputs import RequestOutput
 from vllm.sampling_params import RequestOutputKind
 from vllm.v1.attention.backends.registry import AttentionBackendEnum
 
 logger = logging.getLogger(__name__)
+
+
+def _prepare_generation_request_metrics(
+    output: RequestOutput, *, prefix: str
+) -> list[m.Metric]:
+    """Prepare vLLM metrics from a RequestOutput.
+
+    For `[num_prompts]` submitted prompts, vLLM returns `[num_prompts]`
+    per parent `RequestOutput`s (one per `add_request` call), each carrying
+    a single `RequestStateStats` on `.metrics`.
+
+    Caveat under `SamplingParams.n > 1`: vLLM stores one `RequestStateStats`
+    per child request; the parent output exposes the **last-finishing**
+    child's timeline. `arrival_time` is shared across siblings, but
+    [`queued_ts`, `scheduled_ts`, `first_token_ts`, `last_token_ts`,
+    `num_generation_tokens`] describe one specific child — not an aggregate,
+    not the first sibling's. The other `n-1` siblings' stats are dropped by
+    vLLM at ``output_processor._finish_request``.
+    """
+
+    # TODO: Per-request fields here come from RequestOutput.metrics
+    # (RequestStateStats). Engine-aggregate stats, such as KV-cache usage,
+    # prefix-cache hit rate, preemptions, and batch occupancy, live in
+    # SchedulerStats / IterationStats and require registering a
+    # vllm.v1.metrics.loggers.StatLoggerBase via
+    # LLMEngine.from_engine_args(..., stat_loggers=[...]).
+
+    metric_values: dict[str, float] = {}
+    if output.num_cached_tokens is not None:
+        metric_values[f"{prefix}/num_cached_tokens"] = output.num_cached_tokens
+
+    stats = output.metrics
+    if stats is not None:
+        metric_values[f"{prefix}/queue_time_ms"] = (
+            stats.scheduled_ts - stats.queued_ts
+        ) * 1000
+
+        if stats.num_generation_tokens > 0:
+            metric_values[f"{prefix}/time_to_first_token_ms"] = (
+                stats.first_token_latency * 1000
+            )
+            metric_values[f"{prefix}/prefill_time_ms"] = (
+                stats.first_token_ts - stats.scheduled_ts
+            ) * 1000
+
+        if stats.num_generation_tokens > 1:
+            first_to_last_token_ms = (stats.last_token_ts - stats.first_token_ts) * 1000
+            metric_values[f"{prefix}/decode_time_ms"] = first_to_last_token_ms
+            metric_values[
+                f"{prefix}/inter_token_latency_ms"
+            ] = first_to_last_token_ms / (stats.num_generation_tokens - 1)
+
+    # Emit each value with both Mean and Max aggregators.
+    return [
+        metric
+        for key, value in metric_values.items()
+        for metric in (m.Metric(key, m.Mean(value)), m.Metric(key, m.Max(value)))
+    ]
 
 
 @dataclass(kw_only=True, slots=True)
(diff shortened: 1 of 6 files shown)


torchtitan/experiments/rl/config_registry.py
```diff
@@ -23,6 +23,7 @@
 from torchtitan.experiments.rl.actors.generator import SamplingConfig, VLLMGenerator
 from torchtitan.experiments.rl.actors.trainer import PolicyTrainer
 from torchtitan.experiments.rl.grpo import GRPOLoss, RLTrainer
+from torchtitan.experiments.rl.observability import metrics as m
```
tianyu-l: Here maybe import `MetricsProcessor` is better, as there's no sl / other m.metrics

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -364,30 +431,50 @@ async def generate(
...
 
             completions: list[Completion] = []
+            generation_metrics: list[m.Metric] = []
+            output_token_counts: list[int] = []
             for output in all_outputs:
                 prompt_idx = int(output.request_id)
-                prompt_token_ids = output.prompt_token_ids
+                # TODO: When renderer-owned prompt construction lands, check
+                # if we should return the output prompt token IDs
+                # or keep the input prompt token IDs
```
tianyu-l: when would they be different?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -32,12 +33,71 @@
...
+
+    For `[num_prompts]` submitted prompts, vLLM returns `[num_prompts]`
+    per parent `RequestOutput`s (one per `add_request` call), each carrying
+    a single `RequestStateStats` on `.metrics`.
+
+    Caveat under `SamplingParams.n > 1`: vLLM stores one `RequestStateStats`
+    per child request; the parent output exposes the **last-finishing**
+    child's timeline. `arrival_time` is shared across siblings, but
+    [`queued_ts`, `scheduled_ts`, `first_token_ts`, `last_token_ts`,
+    `num_generation_tokens`] describe one specific child — not an aggregate,
```
tianyu-l, line 56: how about `num_cached_tokens`

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -32,12 +33,71 @@
...
+def _prepare_generation_request_metrics(
+    output: RequestOutput, *, prefix: str
+) -> list[m.Metric]:
+    """Prepare vLLM metrics from a RequestOutput.
+
+    For `[num_prompts]` submitted prompts, vLLM returns `[num_prompts]`
+    per parent `RequestOutput`s (one per `add_request` call), each carrying
+    a single `RequestStateStats` on `.metrics`.
+
+    Caveat under `SamplingParams.n > 1`: vLLM stores one `RequestStateStats`
```
tianyu-l, line 52: Given this "caveat", I assume the metrics is "lossy". Do we plan to fix / workaround it, e.g. by `SchedulerStats / IterationStats`?

tianyu-l (approved):
LGTM, had some questions.

torchtitan/experiments/rl/actors/generator.py
felipemello1: No, the cached_tokens is correct, the prompt is prefilled once and shared across siblings. 

That said, i will be modifying this so instead of doing n=8, we would do 8 generations of n=1, which is necessary for multiturn and also works for single turn. We will have the "caveat" fixed by default.

torchtitan/experiments/rl/actors/generator.py
felipemello1: answered above

torchtitan/experiments/rl/actors/generator.py
tianyu-l: hmm, I actually don't know how it works.
- In general, "prefix caching" shouldn't work only for prompt tokens? So I'd imagine that with prefix caching on, more things than prefix caching could be cached.
- When we have multiple requests, if they are all put in prefill stage together, would the cache be used at all?
- Is there other tradeoff between sending n=8 vs. sending 8 n=1 requests? Intuitively the former might be more efficient?

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -316,21 +377,25 @@ async def sync_log_step(self, step: int, relative_step: int | None = None) -> No
     @sl.log_trace_span("generate")
     async def generate(
         self,
-        prompts: list[str],
+        tokenized_prompts: list[list[int]],
         *,
         sampling_config: SamplingConfig | None = None,
-    ) -> list[Completion]:
-        """Generate completions for the given prompts.
+        metrics_prefix: str = "generator",
```
tianyu-l, line 383: what's the consideration of not hardcoding `is_validation` as in https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/observability/metrics/processor.py#L162

felipemello1: People may have multiple generators, e.g. 2 models talking to each other, teacher/student, llm as a judge, etc. All sorts of ideas in RL around generators.

tianyu-l: btw should we update the default here https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/observability/metrics/processor.py#L54

torchtitan/experiments/rl/grpo.py
```diff
@@ -306,6 +307,8 @@ def __init__(self, config: Config):
             log_dir=config.dump_folder,
             job_config=config.to_dict(),
         )
+        # TODO: Replace this single-turn tokenizer with renderer
+        self.tokenizer = HuggingFaceTokenizer(tokenizer_path=config.hf_assets_path)
```
wwwjn, line 319: In core titan, the tokenization is done during the data loader. Should we consider introduce a dataloader for RL as well (in the following PRs)?

torchtitan/experiments/rl/actors/generator.py
wwwjn: I have read the similar behavior of sending n 1-request to vllm before: https://fburl.com/code/ztfbbbew

Sending 8 n=1 requests helps better manage the unique ids in vllm requests (needs to dig into vllm for more details). However, these no information about the impact on perf

felipemello1: I did a bit of digging:

1. Internally, vllm still breaks down the requests into single ones (per link that @wwwjn sent), and appends them to a queue. The scheduler batches multiple queued requests into a forward pass (continuous batching).
	
2. There's a scheduler that loops over requests and checks if their hash is a cache hit or miss. The first miss schedules the full prefill (e.g., 100 prompt tokens); the others register a cache hit — a reservation on the prefiller's blocks (the KV values themselves aren't computed yet). All 8 then run in one forward pass: within that pass, the prefiller writes the prompt KV, and the cache-hit siblings wait for it and read when ready (the kernel writes K, V before the attention step, so there's no race). After this forward, all 8 start decoding.
	
3. So if we do 8×n=1 or 1×n=8, we still calculate the full prompt prefill only once. One nice thing: vLLM does chunked prefill by default — new incoming prefill tokens get mixed into the same forward pass as ongoing decode tokens, so a new request doesn't stall in-flight ones. There's also prefill-decode disaggregation, which goes further by running prefill and decode on separate machines (each phase has a different compute/memory profile).
	
4. Prefix caching isn't prompt-only — any full block (default=16 tokens) of finalized tokens (prompt or generated) gets hashed and cached.

torchtitan/experiments/rl/actors/generator.py
felipemello1: i dont think that we want any of the generator metrics in the console log. We already have tokens/s. Let me know if you disagree

torchtitan/experiments/rl/grpo.py
felipemello1: Hmmm, we do have a `dataset` and a `make_env` that associates a sample to an env. I dont think we need a dataloader because there is no batching. For each dataset sample, we launch an async rollout request.

torchtitan/experiments/rl/actors/generator.py
felipemello1: They shouldnt be (I tested it). I got rid of it -- I removed it from "Completions". Now the `Trajectory` will just hold the `prompt_ids`, instead of the `Completion.prompt_ids`. We do: "token prompts in, token response out".

Having the generator output prompt tokens only makes sense if the input is text.

PR #3392, opened 2026-05-18
[MTP] Add MTP module for deepseek_v3 model

### LINKED ISSUE #933: [Feature] Add Multi-Token Prediction module
We need to create a multi-token prediction module (same as what DeepSeek v3 used) that can plug into architectures such as DeepSeek v3 etc. and allow for multi-token prediction training with an arbitrary level of blocks.  (default =4). 
This should be a modular architecture in that it clones the existing block and appends to the neck of the model and auto-integrates for training.

Matrix-Z97:
## Summary

Integrate shared Multi-Token Prediction (MTP) support for DeepSeek V3 into the current upstream codebase, using the common decoder module path, declarative config build mechanism, class-based loss framework, and current parallel/sharding APIs.

## Model Structure
`DeepSeekV3Model` inherits `MTPDecoder`. If MTP is disabled,
`MTPDecoder.forward()` falls back to `Decoder.forward()`.

`MTPDecoder` owns the normal decoder fields (`tok_embeddings`, `layers`, `norm`,
`lm_head`) plus `mtp_layers: ModuleList[MTPTransformerBlock] | None`.

When MTP is enabled, output is a list:
```python
[main_logits, mtp_logits_1, mtp_logits_2, ...]
```

## Config
MTP fields are directly part of `MTPDecoder.Config`:

```python
class MTPDecoder.Config(Decoder.Config):
    num_mtp_layers: int = 0
    mtp_layer_config: TransformerBlock.Config | None = None
    detach_heads: bool = False
```

`mtp_layer_config` may be provided by a model flavor. If absent,
`MTPDecoder.materialize_mtp_config()` derives it from `config.layers[-1]` and
creates an `MTPTransformerBlock.Config` during `Config.update_from_config()`,
before model build and before model-specific sharding config is applied.

Example config:
```python
config.model_spec.model.num_mtp_layers = 1
config.loss = MTPLoss.Config(
    global_vocab_size=decoder_vocab_size(config.model_spec),
    loss_scaling_factor=0.3,
)
```

## MTP Block
`MTPTransformerBlock` inherits `TransformerBlock`. Its config extends the normal
block config with MTP-only modules:
```python
class MTPTransformerBlock.Config(TransformerBlock.Config):
    enorm: RMSNorm.Config
    hnorm: RMSNorm.Config
    eh_proj: Linear.Config
    mtp_norm: RMSNorm.Config
```

Each MTP layer follows:
```text
h = eh_proj(cat(enorm(shifted_embedding), hnorm(previous_hidden)))
h = h + attention(attention_norm(h), attention_masks, positions)
h = h + moe_or_ffn(ffn_norm(h))
h = mtp_norm(h)
```

## Input Organization
The dataloader still provides normal decoder inputs: `tokens: [B, S]`,
`labels: [B, S]`, and optional `positions: [B, S]`.

MTP shifted tokens are generated inside the model with `roll_mtp_sequence()`.
The operation preserves length and uses reset-style positions to avoid crossing
packed-document boundaries:
```text
tokens:    [A0, A1, A2, B0, B1, B2]
positions: [ 0,  1,  2,  0,  1,  2]
shift=1:  [A1, A2,  0, B1, B2,  0]
```

`MTPLoss` uses the same helper to generate shifted labels, filling invalid
document-tail targets with `IGNORE_INDEX`.

## Forward And Loss
`MTPDecoder.forward()` runs the main decoder, keeps the pre-final-norm hidden
state for MTP, applies final norm for the main output, then runs each MTP depth.

`MTPLoss` consumes the list output and computes
`main_loss + loss_scaling_factor * average(mtp_losses)`.

## Parallelism Status
FSDP includes MTP layers in decoder transformer block traversal. TP/EP use
model-specific sharding rules. PP, CP, and Chunked CE are unsupported here.

## Limitations
- MTP relies on positions to avoid packed-document boundary crossing; without
  positions, it falls back to sequence-wide shift and is only correct for
  unpacked samples.
- PP and CP are explicitly unsupported in the current DeepSeek-V3 path.
- Chunked CE is not supported for MTP in this PR.
- DeepSeek-V3 still needs model-specific sharding and checkpoint key mapping.

## Open Questions
- Is deriving `MTPTransformerBlock.Config` from `config.layers[-1]` appropriate,
  or should model flavors explicitly provide `mtp_layer_config`?
- Should packed-document boundaries come from explicit `cu_seqlens` /
  `PackedSeqParams` instead of reset-style positions?
- What is the preferred PP policy for placing or partitioning MTP layers?


Matrix-Z97:
@tianyu-l Could you help review this code?

Matrix-Z97:
@tianyu-l Hi, could you help review this MTP implementation when you have time?

The MTP module is implemented as a common MTPBlock / MTPTransformerBlock, configured through Trainer.Config.mtp, and wired into Decoder.forward() so model-specific code only needs to provide sharding/state-dict details. The CP path uses dataloader-prepared [S + K] sequences and keeps the extra shifted tokens local after CP sharding so each rank can compute the MTP depths.

There are still a few limitations in the current version: MTPLoss is implemented as a weighted auxiliary CE loss rather than a more general loss composition mechanism, and PP is not supported for MTP yet.  If needed, I’d be happy to further improve and polish these parts based on your feedback.

Do you have any plans or preferred direction for supporting MTP? I’d appreciate any guidance on whether this common-module approach matches the direction you’d like TorchTitan to take.

tianyu-l:
@Matrix-Z97 sorry I apologize for the delay, will definitely get to it before end of week

Matrix-Z97:
> @Matrix-Z97 sorry I apologize for the delay, will definitely get to it before end of week
No worries! Sounds good. Let me know if any questions or suggestions come up.

torchtitan/trainer.py
```diff
@@ -105,6 +106,9 @@ class Config(Configurable.Config):
         debug: DebugConfig = field(default_factory=DebugConfig)
         override: OverrideConfig = field(default_factory=OverrideConfig)
         loss: BaseLoss.Config = field(default_factory=BaseLoss.Config)
+        # Multi-Token Prediction training config. Trainer wires this shared
+        # config into both the model and MTPLoss before either component builds.
+        mtp: MTPBlock.Config | None = None
```
tianyu-l: Trainer shouldn't be aware of MTP. In general let's try to make this file oblivious to MTP.

torchtitan/distributed/fsdp.py
tianyu-l, line 110: This is a function that applies FSDP to decoder. I feel Decoder + MTP form it's own model structure which is worth its own FSDP application logic, potentially calling this one as a sub-function.

torchtitan/distributed/context_parallel.py
tianyu-l, line 1: Given the complexity, I would worry about CP in a separate PR.

torchtitan/components/loss.py
```diff
@@ -377,6 +379,78 @@ def __init__(self, config: Config, *, compile_config: CompileConfig | None = Non
         self._maybe_compile(compile_config)
 
 
+class MTPLoss(BaseLoss):
```
tianyu-l: I imagine it'd be trickier to apply Chunked CE loss with MTP, which is the current default in torchtitan.

In the first PR, I agree that let's not worry about that.

torchtitan/models/common/mtp.py
tianyu-l, line 1: I have to admit that I haven't tried to implement MTP in torchtitan by myself, so the direction I'm giving below may have problems, but here's the high level idea I would propose based on my limited understanding looking at the Figure 3 in DeepSeek V3 paper:
- Decoder + MTP should be the model itself, instead of putting MTP into trainer.
- the new high level module (e.g. MTPDecoder) should inherit Decoder and add MTP blocks, or compose with Decoder by owning Decoder and MTP blocks as attributes.
- The output of this new module should be a tuple/list from [Main Decoder, MTP Module 1, MTP Module 2, etc.].
- The loss function (e.g. MTPCrossEntropyLoss) should consume one labels and the tuple above, with MTP weights being in a config.

tianyu-l (changes requested):
Sorry for the delay! Left some initial reviews. 

Also fyi I'll be afk for two days, during which I'd have limited access for further review.

torchtitan/models/common/mtp.py
Matrix-Z97: Thanks for the suggestion. I agree with this direction, and I updated the implementation to follow this model-centric structure.

In the current version:

- `MTPDecoder` inherits from `Decoder` and owns the `MTPBlock`.
- DeepSeek-V3 now uses `MTPDecoder`, so `Decoder + MTP` is represented as the model itself.
- The forward output is a list:

  ```python
  [main_logits, mtp_logits_1, mtp_logits_2, ...]
  ```

- `MTPLoss` consumes the single labels tensor together with that list of logits, and applies the configured auxiliary MTP loss weight.
- Trainer no longer owns or wires an `mtp_config`. It only uses the model config’s `extra_mtp_tokens` to request the additional future tokens needed by prepared-input MTP.

torchtitan/distributed/context_parallel.py
Matrix-Z97: OK, the CP-related parts have already been removed. If needed, I can further improve this in a follow-up.

torchtitan/distributed/fsdp.py
Matrix-Z97: That makes sense. The reason I reused the decoder FSDP path here is that mtp_block is also composed of transformer-block-like modules, so it can share the same FSDP treatment as the main decoder layers.

This keeps the change minimal while still covering the MTP parameters correctly. If preferred, I can split this into an explicit apply_fsdp_to_mtp_decoder() wrapper that prepares the combined block list and then reuses the existing decoder FSDP logic as a sub-function.

rakkit:
what about PP?  for PP it will be tricky.  
And it will be good to make MTP as torchtitan "component" to be general for qwen, mimo, glm, etc models 

we are also very interested in having MTP for our model training, i will be back to work next week and can help review more.

Matrix-Z97:
> we are also very interested in having MTP for our model training, i will be back to work next week and can help review more.

Good question. A straightforward implementation for PP would be to ensure that the MTP block is placed on the last pipeline stage together with the final decoder output, though there may be better designs.

Given the complexity, I’d suggest handling PP support in a follow-up PR after the current MTP architecture becomes more mature.

I also fully agree with your suggestion to make MTP a common component. The current MTP implementation is already moving in that direction, and we can keep refining it to make it more reusable across models. Looking forward to your help and feedback.

torchtitan/models/common/decoder.py
```diff
@@ -330,3 +329,102 @@ def get_attention_masks(
                 f"Only VarlenAttention and FlexAttention support attention masks, "
                 f"got {type(inner_attn).__name__}"
             )
+
+
+class MTPDecoder(Decoder):
```
tianyu-l: let's put this in mtp.py and leave decoder.py intact

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,240 @@
...
+        if self.detach_heads:
+            input_offset = input_offset.detach()
+        h = self.eh_proj(
+            torch.cat([self.enorm(input_offset), self.hnorm(prev_embed)], dim=-1)
+        )
+        h = self.inner(h, attention_masks, positions)
+        return self.mtp_norm(h)
+
+
+class MTPBlock(Module):
```
tianyu-l: I don't feel we need this abstraction. A MTPDecoder can directly own the list of MTP modules?

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,240 @@
...
+    return MTPSequenceSlices(
+        main=sequence[:, :seq_len],
+        offsets=tuple(
+            sequence[:, offset : offset + seq_len]
+            for offset in range(1, num_mtp_modules + 1)
+        ),
+    )
+
+
+def mtp_labels_for_depth(
```
tianyu-l: Is this used? Let's try to make the PR minimal

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,240 @@
...
+            outputs.append(h)
+        return outputs
+
+
+def split_mtp_sequence(
+    sequence: torch.Tensor,
+    *,
+    num_mtp_modules: int,
+) -> MTPSequenceSlices:
+    """Split ``[B, S + K]`` into main ``[B, S]`` and K shifted ``[B, S]`` views."""
```
tianyu-l: Does this work with document packing? Seems not

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,240 @@
...
+            inner_block_config: TransformerBlock.Config,
+            *,
+            dim: int,
+        ) -> "MTPTransformerBlock.Config":
+            if isinstance(inner_block_config, cls):
+                return inner_block_config
+            # Keep the inner transformer block intact and wrap only the MTP
+            # fusion/projection modules around it. This lets model-specific
+            # block construction stay in the original block config.
+            return cls(
```
tianyu-l: It's confusing for this cls to accept attention, feed_forward, etc. **and** inner_block, **and** enorm, etc.

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,240 @@
...
+    The block implements the DeepSeek-V3 style fusion:
+
+    ``eh_proj(cat(enorm(shifted_embedding), hnorm(previous_hidden)))``
+
+    followed by one regular transformer block.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(TransformerBlock.Config):
+        inner_block_config: TransformerBlock.Config
```
tianyu-l: would inheriting `TransformerBlock` (instead of composition) makes it easier?

torchtitan/models/deepseek_v3/state_dict_adapter.py
```diff
@@ -51,6 +51,11 @@ def __init__(
             "model.layers.{}.mlp.shared_experts.up_proj.weight": "layers.{}.moe.shared_experts.w3.weight",
             "model.layers.{}.mlp.shared_experts.down_proj.weight": "layers.{}.moe.shared_experts.w2.weight",
             "model.layers.{}.mlp.gate.e_score_correction_bias": "layers.{}.moe.expert_bias_E",
+            # MTP Module
+            "model.layers.{}.enorm.weight": "layers.{}.enorm.weight",
+            "model.layers.{}.hnorm.weight": "layers.{}.hnorm.weight",
+            "model.layers.{}.eh_proj.weight": "layers.{}.eh_proj.weight",
+            "model.layers.{}.shared_head.norm.weight": "layers.{}.mtp_norm.weight",
```
tianyu-l, line 58: This means DeepSeek weights contain these fields to be loaded, right?

Given it's DSv3 specific, can we restrict the modification to be within deepseek_v3 folder, including the changes in loss.py, and the creation of mtp.py

torchtitan/trainer.py
```diff
@@ -496,11 +497,12 @@ def __init__(self, config: Config):
         self.tokenizer = config.tokenizer.build(tokenizer_path=config.hf_assets_path)
 
         # build dataloader
+        dataloader_seq_len = config.training.seq_len + self.extra_mtp_tokens
```
tianyu-l: I'm not comfortable with this change
- MTP is modeling behavior
- dataloader seq len is infra behavior

Let's say the dataloader previous loads a single row with seq_len=8 and two documents
[A0, A1, A2, B0, B1, B2, B3, B4]

How are you handling MTP for each document? Would TP / CP (which would shard on seq dim at some point) behavior change?

Maybe should have an RFC before code change?

tianyu-l (changes requested):

torchtitan/trainer.py
Matrix-Z97: Thanks for pointing this out! It's important to keep the MTP module decoupled from the dataloader for TP/CP correctness, and the sequence sharding you mentioned is an excellent point.

I’ll rethink how to properly decouple the design. Regarding the RFC, do you prefer I post the draft directly under this PR, or should I open a new issue/doc for it?

tianyu-l: post under this PR is fine, but however you prefer

Matrix-Z97: I updated the implementation so the dataloader no longer needs to change seq_len or load extra MTP tokens. The model still receives the normal [B, S] input, and MTP shifted inputs are generated inside the model with a length-preserving shift. For packed documents, the shift uses reset-style positions to avoid crossing document boundaries. For example:
```
tokens:    [A0, A1, A2, B0, B1, B2, B3, B4]
positions: [ 0,  1,  2,  0,  1,  2,  3,  4]
shift=1:  [A1, A2,  0, B1, B2, B3, B4, 0]
```
The loss side uses the same shifting logic for MTP labels, with invalid document-tail positions filled as ignore labels, so A2 will not be trained to predict B0.

For CP/SP, this likely needs extra communication, such as exchanging boundary tokens between neighboring ranks or all-gathering the sequence before shifting and re-sharding.

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,252 @@
...
+
+    MTP is kept as model behavior: the main decoder consumes the normal input
+    sequence, and each MTP layer predicts one extra depth from internally shifted
+    token embeddings.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Decoder.Config):
+        num_mtp_layers: int = 0
+        mtp_layer_config: TransformerBlock.Config | None = None
```
tianyu-l: each mtp layer should have its own config, like https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder.py#L82

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,252 @@
...
+        detach_heads: bool = False
```
tianyu-l: Interesting flag. In reality is this always False? If so can we not introduce this flag right now?

torchtitan/distributed/context_parallel.py
```diff
@@ -152,8 +152,7 @@ def prepare_context_parallel_input(
         load_balancer_type,
     )
     extra_kwargs["positions"] = positions
-    if attention_masks is not None:
```
tianyu-l, line 160: why changing

torchtitan/components/loss.py
```diff
@@ -195,6 +196,8 @@ def backward(  # pyrefly: ignore[bad-override]
         ctx,
         grad_output: torch.Tensor,
     ) -> tuple[torch.Tensor, None, None, None, None]:
+        if isinstance(grad_output, DTensor):
+            grad_output = grad_output.to_local()
```
tianyu-l: why changing?

torchtitan/components/loss.py
```diff
@@ -377,6 +382,85 @@ def __init__(self, config: Config, *, compile_config: CompileConfig | None = Non
         self._maybe_compile(compile_config)
 
 
+class MTPLoss(BaseLoss):
```
tianyu-l: let's put this in mtp.py to isolate the change for now

torchtitan/models/common/mtp.py
```diff
@@ -0,0 +1,252 @@
...
+            eh_proj=Linear.Config(
+                in_features=config.dim * 2,
+                out_features=config.dim,
+                bias=False,
+            ),
+            mtp_norm=RMSNorm.Config(normalized_shape=config.dim),
+        )
+
+    def __init__(self, config: Config):
+        self.materialize_mtp_config(config)
```
tianyu-l: The model architecture should have been set in config time, not in constructor. From config, we should know how many MTP layers are there, if each is using MoE, etc.

torchtitan/models/deepseek_v3/config_registry.py
```diff
@@ -65,29 +63,12 @@ def deepseek_v3_debugmodel() -> Trainer.Config:
...
-            ),
-            MXFP8GroupedExpertsConverter.Config(
-                model_compile_enabled=model_compile_enabled,
-                pad_multiple=128,
-            ),
-        ],
+    config.model_spec.model.num_mtp_layers = 1
+    config.loss = MTPLoss.Config(
+        global_vocab_size=decoder_vocab_size(config.model_spec),
+        loss_scaling_factor=0.3,
```
tianyu-l: Is there literature / evidence from DeepSeek V3 that this value should be set as 0.3?

torchtitan/components/loss.py
```diff
@@ -377,6 +382,85 @@ def __init__(self, config: Config, *, compile_config: CompileConfig | None = Non
...
+    """DeepSeek-V3 multi-token prediction loss."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(BaseLoss.Config):
+        loss_scaling_factor: float = 0.3
```
tianyu-l: maybe
```suggestion
        mtp_scale: float = 0.3
```
to be less ambiguous

torchtitan/distributed/fsdp.py
tianyu-l: let's be conservative and create a separate `apply_fsdp_to_mtp_decoder` used by deepseek_v3 only for now

torchtitan/models/common/mtp.py
tianyu-l, line 1: please put this under deepseek_v3/ folder for now

torchtitan/components/loss.py
```diff
@@ -377,6 +382,85 @@ def __init__(self, config: Config, *, compile_config: CompileConfig | None = Non
...
+                    f"prediction, got {len(pred)} predictions."
+                )
+
+            main_loss = self.fn(
+                pred[0],
+                labels[:, : pred[0].shape[1]],
+                global_vocab_size=self.global_vocab_size,
+            )
+            mtp_loss = pred[0].new_zeros((), dtype=torch.float32)
+            positions = loss_inputs.get("positions", None)
```
tianyu-l: positions should never be None

torchtitan/components/loss.py
```diff
@@ -320,6 +323,7 @@ def __call__(
         pred: torch.Tensor,
         labels: torch.Tensor,
         global_valid_tokens: float | None = None,
+        **loss_inputs: Any,
```
tianyu-l:
```suggestion
        **kwargs: Any,
```

torchtitan/components/loss.py
```diff
@@ -352,6 +356,7 @@ def __call__(
...
```
tianyu-l:
```suggestion
        **kwargs: Any,
```

torchtitan/models/deepseek_v3/config_registry.py
```diff
@@ -65,29 +63,12 @@ def deepseek_v3_debugmodel() -> Trainer.Config:
     )
 
 
-def deepseek_v3_debugmodel_mxfp8() -> Trainer.Config:
+def deepseek_v3_debugmodel_mtp() -> Trainer.Config:
```
tianyu-l, line 74: maybe update https://github.com/pytorch/torchtitan/blob/main/tests/integration_tests/models.py#L97 to use this mtp version, AND make sure it works with `--debug.spmd_typechecking`

torchtitan/models/deepseek_v3/model.py
```diff
@@ -200,8 +200,22 @@ def update_from_config(
...
-            Decoder.Config.update_from_config(self, config=config, **kwargs)
+            MTPDecoder.Config.update_from_config(self, config=config, **kwargs)
             parallelism = config.parallelism
+            if (
+                self.num_mtp_layers > 0
+                and parallelism.pipeline_parallel_degree > 1
+            ):
+                raise NotImplementedError(
+                    "DeepSeek-V3 MTP does not support pipeline parallelism yet."
+            )
```
tianyu-l: What's the difficulty? In dsv3, there's only 1 mtp layer, so we can always put it on layer PP rank, similar to `lm_head`.

That said, I'm OK with supporting it in another PR.

torchtitan/models/deepseek_v3/model.py
```diff
@@ -200,8 +200,22 @@ def update_from_config(
             config,
             **kwargs,
         ) -> None:
-            Decoder.Config.update_from_config(self, config=config, **kwargs)
+            MTPDecoder.Config.update_from_config(self, config=config, **kwargs)
```
tianyu-l, line 204: The PP / CP check below should be put in `MTPDecoder`'s `update_from_config`, as they are not dsv3-specific

torchtitan/trainer.py
tianyu-l: For SP (on TP axis) it should work out-of-box, because the labels are never sharded, assuming we don't use Chunked CE loss. Is this right?

For CP it indeed sounds tricky, especially considering we would like to push CP input sharding into dataloader. cc @fegin @jinsooihm

torchtitan/trainer.py
```diff
@@ -746,7 +744,15 @@ def forward_backward_step(
             assert len(model_parts) == 1
             with self.train_context():
                 pred = model_parts[0](inputs, **extra_kwargs)
-                loss, _ = self.loss_fn(pred, labels, global_valid_tokens)
+                loss_inputs = {}
```
tianyu-l: no need to introduce this variable

torchtitan/trainer.py
```diff
@@ -746,7 +744,15 @@ def forward_backward_step(
...
+                if "positions" in extra_kwargs:
+                    loss_inputs["positions"] = extra_kwargs["positions"]
+                loss, _ = self.loss_fn(
+                    pred,
+                    labels,
+                    global_valid_tokens,
+                    **loss_inputs,
```
tianyu-l:
```suggestion
                    positions=extra_kwargs["positions"],
```
might need to adjust the if-else a bit

torchtitan/components/loss.py
tianyu-l: and then `del kwargs` after entering the function

tianyu-l (changes requested):

torchtitan/models/deepseek_v3/config_registry.py
Matrix-Z97: We found it in [DeepSeek-V3 Technical Report](https://arxiv.org/pdf/2412.19437)
image

torchtitan/models/common/mtp.py
Matrix-Z97: Do you mean that the MTP layer configs should be fully materialized in the model config, for example in `deepseek_v3/__init__.py`, instead of being derived or expanded inside `MTPDecoder.__init__()`?

I agree that this would better match TorchTitan’s config-time architecture definition. My only concern is that defining the full MTP config in the model flavor may tightly couple MTP to each model config. The current approach is a bit more pluggable: the config registry only needs to set num_mtp_layers, and update_from_config() materializes per-layer MTP configs by reusing the main model’s transformer block config, including attention / FFN / MoE choices.

torchtitan/trainer.py
Matrix-Z97: In the current MTP path, the main decoder hidden states become SP-sharded after the transformer layers, while the shifted-token validity mask/labels are produced from unsharded tokens / positions. That caused layout mismatches in practice, for example when multiplying mtp_prev_embed with the valid-token mask and when using torch.where to build packed-document-correct MTP labels.

torchtitan/models/common/mtp.py
Matrix-Z97: will remove it

torchtitan/distributed/context_parallel.py
Matrix-Z97: will remove it

torchtitan/models/common/mtp.py
tianyu-l: > My only concern is that defining the full MTP config in the model flavor may tightly couple MTP to each model config.

This is not necessarily true? You can still do "programming" in config time, e.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/deepseek_v3/__init__.py#L344

You can built mtp_layers based on what you have in regular layers, or based on some common input configs.

`update_from_config()` gives one a way to program in the config space AFTER command-line input changes python config, but in modeling you don't need to worry about that (one shouldn't change model on CLI). What ever you could do in `update_from_config()` related to model only, you could do in `__init__.py`.

torchtitan/trainer.py
Matrix-Z97: Updated!

torchtitan/trainer.py
Matrix-Z97: Updated!

torchtitan/models/deepseek_v3/model.py
Matrix-Z97: Updated!

torchtitan/components/loss.py
Matrix-Z97: Updated!

torchtitan/models/common/mtp.py
Matrix-Z97: Updated!

torchtitan/components/loss.py
Matrix-Z97: Updated!

torchtitan/components/loss.py
Matrix-Z97: Updated!

torchtitan/models/deepseek_v3/config_registry.py
Matrix-Z97: Updated the related integration test to use the MTP debug config. I also verified it with --debug.spmd_typechecking; the SPMD backend now runs through with the minimal spmd.no_typecheck() scopes needed for scalar loss composition and mask placement handling.

torchtitan/models/common/mtp.py
Matrix-Z97: Thanks. I updated the current MTP path accordingly: `mtp_layers` are now built in `deepseek_v3/__init__.py` from the regular DeepSeek-V3 layer config, similar to the existing flavor-config construction pattern. `MTPDecoder.Config.update_from_config()` no longer derives or materializes MTP layer configs; it only temporarily includes the already-defined `mtp_layers` so the existing decoder config update/sharding logic can be reused, then removes them from the main layer list again.

torchtitan/distributed/fsdp.py
Matrix-Z97: Done. I kept the generic decoder FSDP path unchanged and added a separate `apply_fsdp_to_mtp_decoder()` in `mtp.py`

The MTP-specific wrapper temporarily exposes mtp_layers through the decoder-style model.layers interface, calls the existing `apply_fsdp_to_decoder()` to reuse the battle-tested FSDP/MoE handling, and then removes the temporary entries. This keeps the change conservative and scoped to DeepSeek-V3 MTP for now.

torchtitan/models/deepseek_v3/model.py
Matrix-Z97: Good point. For the current DeepSeek-V3 setting with a single MTP layer, a simple policy could place the MTP layer on the last PP stage together with lm_head, which should be much more manageable than a fully general PP design.

The main difficulty I had in mind is the more general case: multiple MTP layers, how to partition them across PP stages, how to pass the main decoder hidden state into the MTP path, and how to aggregate the list of logits/losses cleanly through the PP schedule.

I agree this can be supported in a follow-up PR.

torchtitan/models/deepseek_v3/mtp.py
rakkit, line 1: i wondering why dont we put "MTP" just into torchtitan comments? this file seems does not rely on any of "deepseek_v3", and merge mtp loss into torchtitan's loss and merge "apply_fsdp_to_mtp_decoder" into  "apply_fsdp_to_decoder"  instead of creating more branches

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+            )
+            mtp_outputs.append(mtp_prev_embed)
+
+        outputs = [h] + mtp_outputs
+        if self._skip_lm_head:
+            return outputs
+        return [
+            self.lm_head(item) if self.lm_head is not None else item
+            for item in outputs
+        ]
```
rakkit, line 278: i wonder actually if we want to cat the logits together?  return single tensor will make PP work easier (but we will need tell MTP loss how many mtp blocks we have) 

also if new version pytorch compiler is smart, it should be able to optimize the lm-head + ce loss on stacked giant logits without OOM

torchtitan/models/deepseek_v3/mtp.py
tianyu-l: I think this is just staying on the conservative side in terms of new features -- Is there any tech reports showing MTP is used beyond DeepSeek (v3) (in pretraining)?

rakkit: [mimo](https://arxiv.org/abs/2601.02780)
[GLM-4&5](https://arxiv.org/abs/2602.15763v2)
[nemotron-3-super](https://arxiv.org/abs/2604.12374v1)
[step-flash-3.5](https://arxiv.org/abs/2602.10604)
qwen (no report but it does support MTP)

i think it will be suprise if any frontier model is not using MTP

tianyu-l: oh then it's my ignorance -- sorry about that

But it indeed makes it awkward to support chunked loss at least with our current `ChunkedLossWrapper`. I guess we can start with no chunked loss.

rakkit: i think we can actually take advantages of using `ChunkedLossWrapper`? just cat all logits to let MTP return single tesnor.  and on MTP loss wrapper side to chunk or re-sumup loss.

rakkit:
it might be good to seprealy logging out the main loss and mtp loss per mtp block and sum up loss. 

do you actually have test run on it already? if we are doing it correctly, train with MTP should make model slightly converge faster..

torchtitan/components/loss.py
```diff
@@ -195,6 +195,8 @@ def backward(  # pyrefly: ignore[bad-override]
         ctx,
         grad_output: torch.Tensor,
     ) -> tuple[torch.Tensor, None, None, None, None]:
+        if isinstance(grad_output, DTensor):
+            grad_output = grad_output.to_local()
```
tianyu-l: is this necessary

torchtitan/models/deepseek_v3/parallelize.py
```diff
@@ -42,8 +46,7 @@ def parallelize_deepseekv3(
         # runs inside the local_map boundary on local tensors.
         if parallel_dims.cp_enabled:
             apply_cp_to_forward(
-                # pyrefly: ignore [missing-attribute]
-                [block.attention.inner_attention for block in model.layers.values()],
+                _get_inner_attentions_for_context_parallel(model),
```
tianyu-l: is this necessary? can we revert?

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+    """Decoder variant that owns MTP layers.
+
+    MTP is kept as model behavior: the main decoder consumes the normal input
+    sequence, and each MTP layer predicts one extra depth from internally shifted
+    token embeddings.
+    """
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Decoder.Config):
+        num_mtp_layers: int = 0
```
tianyu-l: this field can be removed, it can be derived from `mtp_layers`

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+                    config=config,
+                    **kwargs,
+                )
+
+            num_main_layers = len(self.layers)
+            self.layers.extend(self.mtp_layers)
+            try:
+                Decoder.Config.update_from_config(self, config=config, **kwargs)
+            finally:
+                del self.layers[num_main_layers:]
```
tianyu-l, line 197: this is too hacky

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+            reshard_after_forward_policy=reshard_after_forward_policy,
+            ep_degree=ep_degree,
+            edp_mesh=edp_mesh,
+            dp_mesh_dims=dp_mesh_dims,
+            edp_mesh_dims=edp_mesh_dims,
+            enable_symm_mem=enable_symm_mem,
+        )
+    finally:
+        for key in mtp_layer_keys:
+            del model.layers[key]
```
tianyu-l, line 320: similar, this is hacky

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+                input_offset,
+                mtp_prev_embed,
+                attention_masks,
+                positions,
+            )
+            mtp_outputs.append(mtp_prev_embed)
+
+        outputs = [h] + mtp_outputs
+        if self._skip_lm_head:
+            return outputs
```
tianyu-l: We should `raise ValueError` claiming `skip_lm_head` is not supported with MTP decoder for now, before Chunked loss is supported.

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+    """Left-roll an MTP sequence while preserving packed-document boundaries.
+
+    Without ``positions``, this is a regular left shift with the tail filled.
+    With reset-style positions, each document slice is shifted independently:
+    ``[A0, A1, A2, B0, B1]`` becomes ``[A1, A2, fill, B1, fill]``.
+    """
+    if shift < 0:
+        raise ValueError(f"MTP roll shift must be non-negative, got {shift}.")
+    if shift == 0:
```
tianyu-l: do we need this branch? can we delete?

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+        valid_mask,
+        current_spmd_mesh(),
+        dense_activation_placement(tp=spmd.R, cp=spmd.R).per_axis_spmd_types(),
+        dense_activation_placement(tp=spmd.S(1), cp=spmd.R).per_axis_spmd_types(),
+    )
+
+
+def _maybe_redistribute_valid_mask_to_target(
+    valid_mask: torch.Tensor,
+    target: torch.Tensor | None,
```
tianyu-l: confusing what is arg is for

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+    fill_value: int = 0,
+    return_valid_mask: bool = False,
+    valid_mask_target: torch.Tensor | None = None,
+) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+    """Left-roll an MTP sequence while preserving packed-document boundaries.
+
+    Without ``positions``, this is a regular left shift with the tail filled.
+    With reset-style positions, each document slice is shifted independently:
+    ``[A0, A1, A2, B0, B1]`` becomes ``[A1, A2, fill, B1, fill]``.
+    """
```
tianyu-l, line 72: It's hard to understand what this function is doing. Please add comments, including explanation of args and return values to the docstring

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+        h = self.tok_embeddings(tokens)
+        for layer in self.layers.values():
+            h = layer(h, attention_masks, positions)
+
+        mtp_prev_embed = h
+        h = self.norm(h) if self.norm is not None else h
+
+        mtp_outputs = []
+        for depth, layer in enumerate(self.mtp_layers, 1):
+            offset_tokens, valid_offset_tokens = roll_mtp_sequence(
```
tianyu-l: why they are called `offset_tokens, valid_offset_tokens`? The names are confusing enough that I could not understand what's going on.

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+                tokens,
+                shift=depth,
+                positions=positions,
+                fill_value=0,
+                return_valid_mask=True,
+                valid_mask_target=mtp_prev_embed,
+            )
+            input_offset = self.tok_embeddings(offset_tokens)
```
tianyu-l: these should've been computed in `h`?

torchtitan/models/deepseek_v3/mtp.py
tianyu-l: Image

Could you map the name to the notation in the picture?

In their main module, only t1 to t4 are used as inputs. Are we doing the same or are we including t1 to t5 as inputs for the main module?

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+def _redistribute_valid_mask_to_sp(valid_mask: torch.Tensor) -> torch.Tensor:
+    return spmd_redistribute_per_axis(
+        valid_mask,
+        current_spmd_mesh(),
+        dense_activation_placement(tp=spmd.R, cp=spmd.R).per_axis_spmd_types(),
+        dense_activation_placement(tp=spmd.S(1), cp=spmd.R).per_axis_spmd_types(),
+    )
+
+
+def _maybe_redistribute_valid_mask_to_target(
```
tianyu-l: do you really need to call redistribute explicitly in the code? Could you utilize the `ShardingConfig`?

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,420 @@
...
+from torchtitan.models.common.nn_modules import Linear, RMSNorm
+from torchtitan.protocols.module import ModuleList
+
+
+def _redistribute_valid_mask_to_sp(valid_mask: torch.Tensor) -> torch.Tensor:
+    return spmd_redistribute_per_axis(
+        valid_mask,
+        current_spmd_mesh(),
+        dense_activation_placement(tp=spmd.R, cp=spmd.R).per_axis_spmd_types(),
+        dense_activation_placement(tp=spmd.S(1), cp=spmd.R).per_axis_spmd_types(),
```
tianyu-l: If SP is turned off, would you still do this??

tianyu-l (changes requested):
There's still too much slop in the code. This is a complicated feature, at least the integration into torchtitan seems complex.

If you couldn't make the code clear at one time, consider making a design doc for alignment first, listing what you are trying to solve, and what remain as TODO items.

Matrix-Z97:
> There's still too much slop in the code. This is a complicated feature, at least the integration into torchtitan seems complex.
> 
> If you couldn't make the code clear at one time, consider making a design doc for alignment first, listing what you are trying to solve, and what remain as TODO items.

Thanks for the feedback. I agree this feature needs clearer alignment before adding more integration complexity.

I have added a design note under `torchtitan/models/deepseek_v3/mtp.md` to document the current MTP scope: configuration and model integration, TP/SP input handling, FSDP/TP/EP sharding, and MTP loss behavior. I also listed the remaining unsupported pieces explicitly, including CP/PP integration and ChunkLoss support.

And some unclear parts in the implementation based on the review comments  were cleaned up.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Yeah, that sounds reasonable to me. Returning one concatenated logits tensor would probably make the PP interface much simpler than passing around a list.

The only extra thing is that MTPLoss would need to know how many MTP blocks there are, so it can split main logits vs MTP logits and apply the right shifted labels. I think this is a good direction, but maybe better as a follow-up once the current MTP behavior is clearly documented and aligned.

Matrix-Z97:
> it might be good to seprealy logging out the main loss and mtp loss per mtp block and sum up loss.
> 
> do you actually have test run on it already? if we are doing it correctly, train with MTP should make model slightly converge faster..

I agree that logging main_loss and per-block mtp_loss separately would make this easier to diagnose, and I can add that in the next update.  And we ran experiments based on the deepseek_v3_16B model. Unfortunately, we did not observe a clear faster loss decrease with MTP enabled so far.

torchtitan/components/loss.py
Matrix-Z97: Sorry, this was from an earlier change for the `default` SPMD backend. This issue no longer exists under the current `spmd_types` backend, so I have removed it.

torchtitan/models/deepseek_v3/parallelize.py
Matrix-Z97: This was from an earlier attempt to make CP work with MTP. Since CP is not supported by MTP yet, I removed this change and reverted the path back to the simpler form.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Thanks, I have removed it.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Yeah, agree this is hacky. I added it as a compromise to reuse the existing torchtitan decoder config path for MTP layers.

I’m thinking about a cleaner way to share that logic. Open to suggestions if you have a preferred structure.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: I have removed it.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Sorry about that. I’ve added a more detailed docstring now, including what the function does, the args, and the return values.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Thanks for the suggestion. I integrated this into the MTP transformer block as `mtp_input_valid_mask`, so it can now reuse `ShardingConfig` instead of calling redistribute explicitly in the forward code.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: ow this is handled through `ShardingConfig`, and the behavior depends on `enable_sp`.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Yes, there is a small difference from the figure.

In the figure, the MTP examples are shown as clean continuous sequences, like main input `t1...t4`, MTP1 input `t2...t5`, and MTP2 input `t3...t6`.

In our implementation, the main model still consumes the original packed input row, for example:

`[A0, A1, B1, B2, B3, C1, C2]`

Then MTP inputs are derived by shifting inside each packed document boundary:

- MTP1 input: `[A1, fill, B2, B3, fill, C2, fill]`
- MTP2 input: `[fill, fill, B3, fill, fill, fill, fill]`

So we follow the same idea of using `t_{i+k}` for MTP depth `k`, but we also mask out positions that would cross packed document boundaries. The main module is not extended with extra future tokens.

torchtitan/models/deepseek_v3/config_registry.py
tianyu-l: does numerics match the DTensor version you used to have?

torchtitan/models/deepseek_v3/sharding.py
```diff
@@ -60,6 +61,12 @@ def set_deepseek_v3_sharding_config(
         _set_deepseek_v3_layer_sharding(
             layer_cfg, enable_sp=enable_sp, enable_ep=enable_ep
         )
+    if len(getattr(config, "mtp_layers", [])) > 0:
```
tianyu-l:
```suggestion
    if len(config.mtp_layers) > 0:
```

torchtitan/trainer.py
```diff
@@ -746,7 +744,19 @@ def forward_backward_step(
             assert len(model_parts) == 1
             with self.train_context():
                 pred = model_parts[0](inputs, **extra_kwargs)
-                loss, _ = self.loss_fn(pred, labels, global_valid_tokens)
+                if "positions" in extra_kwargs:
```
tianyu-l: nit: let's do
```
loss_kwargs = {}
if "positions" in extra_kwargs:
  loss_kwargs["positions"] = extra_kwargs["positions"]
loss, _ = self.loss_fn(pred, labels, global_valid_tokens, **loss_kwargs)
```

torchtitan/models/deepseek_v3/mtp.py
tianyu-l: not addressed

torchtitan/models/deepseek_v3/MTP.md
tianyu-l, line 1: let's rename to MTP.md

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,405 @@
...
+from torchtitan.models.common.nn_modules import Linear, RMSNorm
+from torchtitan.protocols.module import ModuleList
+
+
+def roll_mtp_sequence(
+    sequence: torch.Tensor,
+    *,
+    shift: int,
+    positions: torch.Tensor | None = None,
+    fill_value: int = 0,
```
tianyu-l: don't need default value here

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,405 @@
...
+
+    def __init__(self, config: Config, *, compile_config: CompileConfig | None = None):
+        self.fn: LossFunction = cross_entropy_loss
+        self._maybe_compile(compile_config)
+        self.mtp_scale = config.mtp_scale
+        self.global_vocab_size = config.global_vocab_size
+
+    def __call__(
+        self,
+        pred: torch.Tensor | list[torch.Tensor],
```
tianyu-l: can we only support list, not single tensor? Your MTPDecoder is outputing a list anyway.

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,405 @@
...
+                raise ValueError(
+                    "MTPLoss expects main prediction plus at least one MTP "
+                    f"prediction, got {len(pred)} predictions."
+                )
+
+            main_loss = self.fn(
+                pred[0],
+                labels[:, : pred[0].shape[1]],
+                global_vocab_size=self.global_vocab_size,
+            )
```
tianyu-l: > Yes, there is a small difference from the figure.

> In the figure, the MTP examples are shown as clean continuous sequences, like main input t1...t4, MTP1 input t2...t5, and MTP2 input t3...t6.

> In our implementation, the main model still consumes the original packed input row

I can see the trade-off, as the paper way is wasting signals (but maintain the same amount of signal for main module and MTP modules?)

In the MTP.md, could you
- clarify this PR's behavior 
- elaborate this difference from the original paper
- compare with what other mainstream OSS repo is doing

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,405 @@
...
+            h = layer(h, attention_masks, positions)
+
+        prev_depth_hidden = h
+        h = self.norm(h) if self.norm is not None else h
+
+        mtp_outputs = []
+        for depth, layer in enumerate(self.mtp_layers, 1):
+            # Shift token ids before embedding. Reusing the main embedding
+            # output would locally shift a sequence-parallel shard.
+            mtp_input_tokens, mtp_input_valid_mask = roll_mtp_sequence(
```
tianyu-l, line 253: I'm not convinced that you need to roll the inputs and compute embedding for every depth.

Why couldn't we NOT roll the inputs at all, just get the outputs from each of main, MTP1, MTP2, etc. with exactly the same inputs, and then only worry about rolling when computing loss??

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,405 @@
...
+                fill_value=0,
+                return_valid_mask=True,
+            )
+            mtp_input_embed = self.tok_embeddings(mtp_input_tokens)
+            prev_depth_hidden = layer(
+                mtp_input_embed,
+                prev_depth_hidden,
+                mtp_input_valid_mask,
+                attention_masks,
+                positions,
```
tianyu-l, line 266: sending this `positions` which has not been rolled seems wrong???

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,405 @@
...
+            where the shifted value came from a valid source position.
+
+    Returns:
+        The shifted tensor. If ``return_valid_mask`` is true, returns
+        ``(shifted, valid_mask)``.
+
+    Example:
+        ``sequence=[A0, A1, A2, B0, B1]`` and
+        ``positions=[0, 1, 2, 0, 1]`` with ``shift=1`` returns
+        ``[A1, A2, fill, B1, fill]``.
```
tianyu-l, line 71: If I originally have tokens [t0, t1, ..., t6] and labels [l0, ... l6], I understand that our main module is seeing tokens [t0, t1, ..., t5] and labels [l1, ..., l6] (dataloader behavior) as model inputs.
- how about MTP1, does it see tokens [t1, ..., t5, FILL] and labels [l1, ..., l6]?? Could you write in the .md how you are aligning inputs with labels for different MTP depths?

torchtitan/models/deepseek_v3/sharding.py
Matrix-Z97: Modified

torchtitan/trainer.py
Matrix-Z97: modified

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: modified

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: modified

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: I think we still need to roll the MTP input embeddings.

For example, with input tokens `[t0, t1, t2, t3]`, the main model produces hidden states `[h0, h1, h2]`, where each `h_i` is used to predict the next token.

For MTP1, the paper-style module should concatenate shifted token embeddings with the main hidden states:

- input embedding branch: `[Emb(t1), Emb(t2), Emb(t3)]`
- previous hidden branch: `[h0, h1, h2]`
- target: `[t2, t3, t4]`

So each position uses `h_i` together with `Emb(t_{i+1})`.

If we do not roll the input tokens, MTP1 would instead concatenate:

- input embedding branch: `[Emb(t0), Emb(t1), Emb(t2)]`
- previous hidden branch: `[h0, h1, h2]`

That misaligns the embedding branch by one step. Another way to view it is that the main hidden states are already shifted for next-token prediction, so the MTP embedding branch needs the same shift to stay aligned with the MTP depth.

torchtitan/models/deepseek_v3/config_registry.py
Matrix-Z97: Yes, numerics look matched. I compared `spmd_types` with the previous `default` backend run step by step, and the losses are almost identical.
| Step | spmd_types Loss | default Loss | Diff (spmd_types - default) |
| ---: | ---: | ---: | ---: |
| 1 | 7.48780 | 7.48780 | 0.00000 |
| 2 | 6.37492 | 6.37489 | +0.00003 |
| 3 | 5.83008 | 5.83002 | +0.00006 |
| 4 | 5.57289 | 5.57282 | +0.00007 |
| 5 | 5.22176 | 5.22176 | 0.00000 |
| 6 | 5.03322 | 5.03319 | +0.00003 |
| 7 | 4.85717 | 4.85712 | +0.00005 |
| 8 | 4.71992 | 4.71979 | +0.00013 |
| 9 | 5.04609 | 5.04603 | +0.00006 |
| 10 | 4.57552 | 4.57556 | -0.00004 |

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: I think keeping `positions` unrolled is intentional here.
The MTP hidden state is still anchored at position `i`; we only shift the token embedding branch to provide `Emb(t_{i+k})`. So for MTP1 we combine:
```
[h0, h1, h2] + [Emb(t1), Emb(t2), Emb(t3)]
```
and the output is still produced at positions `[0, 1, 2]`, then supervised with shifted labels `[t2, t3, t4]`.
If we also `roll` positions, the block would treat those outputs as if they were located at `[1, 2, 3]`, which would change the RoPE/attention semantics and break the alignment with `h_i`.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: I've updated MTP.md

torchtitan/models/deepseek_v3/MTP.md
Matrix-Z97: Done

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: I've updated the MTP.md and observed a similar `roll_tensor` operation in Megatron-LM's MTP implementation, which shifts the token input for each MTP depth instead of only shifting labels in the loss.

References:

- [multi_token_prediction.py](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/transformer/multi_token_prediction.py#L137)

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: I considered reusing the first `self.tok_embeddings(tokens)` output, but I think it is unsafe under SP.

If we roll the embedding output directly, the roll happens after the embedding has already followed the configured activation placement. With sequence parallel enabled, that output is sequence-sharded, so a local roll would only shift within the local shard and would be wrong at shard boundaries.

Rolling token ids first keeps the shift in the logical token sequence, then calling `tok_embeddings(mtp_input_tokens)` lets the embedding module and its `ShardingConfig` produce the correct TP/SP placement. So this costs one extra embedding call per MTP depth, but avoids making MTP depend on manual cross-shard activation shifting.

Matrix-Z97:
@tianyu-l Hi, are you still reviewing this code?

torchtitan/models/deepseek_v3/MTP.md
```diff
@@ -0,0 +1,246 @@
...
+
+### 2.3 Input and Label Alignment by Depth
+
+The trainer still provides the normal next-token-prediction batch. For an unpacked row:
+
+| Path | Model input tokens | Loss labels |
+| --- | --- | --- |
+| Main | `[t0, t1, t2, t3, t4, t5]` | `[t1, t2, t3, t4, t5, t6]` |
+| MTP depth 1 | `[t1, t2, t3, t4, t5, FILL]` | `[t2, t3, t4, t5, t6, IGNORE]` |
+| MTP depth 2 | `[t2, t3, t4, t5, FILL, FILL]` | `[t3, t4, t5, t6, IGNORE, IGNORE]` |
```
tianyu-l, line 83: Please comment that this is different from the paper figures?

I also have a maybe noob question on how MTP helps inference:
During inference if I only have [t1, t2, t3, t4], I could use main module to predict [t5] but I couldn't predict anything beyond using [t1 -- t4], with all the MTP heads. Is the idea that I could feed [t1 -- t5] back to MTP1 to get a drafted [t6] and [t1 -- t6] to MTP2 to get a drafted [t7]?

torchtitan/models/deepseek_v3/mtp.py
```diff
@@ -0,0 +1,407 @@
...
+            # Shift token ids before embedding. Reusing the main embedding
+            # output would locally shift a sequence-parallel shard.
+            mtp_input_tokens, mtp_input_valid_mask = roll_mtp_sequence(
+                tokens,
+                shift=depth,
+                positions=positions,
+                fill_value=0,
+                return_valid_mask=True,
+            )
+            mtp_input_embed = self.tok_embeddings(mtp_input_tokens)
```
tianyu-l, line 260: I still feel the embeddings can be reused. I guess you are not rolling the embeddings because with SP on the output of embedding would be sharded in sequence dim, thus not rollable.

I think your solution is fine for now. Could you add a NOTE on this?

torchtitan/models/deepseek_v3/config_registry.py
tianyu-l: > the losses are almost identical.

It would be great if we could explain the difference.

tianyu-l (approved):
LGTM. Thanks for the work, I learned a lot!

Please rebase and address remaining comments.

torchtitan/models/deepseek_v3/MTP.md
Matrix-Z97: Thanks, I added a note clarifying how our fixed-length input alignment differs from the paper figure.
Your inference understanding is close, but we do not need to feed the full prefix back into each MTP module. The main model predicts `t5`; MTP1 then combines the main model's last hidden state with `Emb(t5)` to draft `t6`; MTP2 combines MTP1's hidden state with `Emb(t6)` to draft `t7`. These drafted tokens can then be verified by the main model through speculative decoding.

torchtitan/models/deepseek_v3/mtp.py
Matrix-Z97: Thanks! I took another look at the embedding path. When SP is off, the embedding output is [batch, seq_len, hidden_dim], so we could reuse and shift it directly. This is also true with TP, since the output is replicated after the vocab-parallel reduction.
With SP on, each rank only has part of the sequence, so a local roll would be wrong at shard boundaries. Reusing the embeddings there would need extra cross-shard communication. I added a NOTE to explain this.

torchtitan/models/deepseek_v3/config_registry.py
Matrix-Z97: Have you seen similar small numerical differences between `spmd_types` and `DTensor` in other models? That might help us understand whether this is specific to MTP or expected backend-level numerical variation.

PR #3395, opened 2026-05-19
RoPE refactor: Using model's max_sequence_length as the upper bound of Training.sequence_length

wwwjn:
- Make seq_len > rope.max_seq_len a hard error: All models' update_from_config now raise ValueError instead of logging a warning when the training sequence length exceeds the model's intrinsic maximum. The model cannot produce valid RoPE embeddings for positions beyond this limit.
- Fix vLLM wrapper RoPE cache sizing: The wrapper was passing a default TrainingConfig() (with seq_len=2048) to update_from_config, which shrank the RoPE cache from the model's intrinsic max (e.g., 4096 for Qwen3) to 2048. It then had to extend the cache back via _extend_rope_cache because vLLM's max_model_len (derived from the original max_position_embeddings=4096) exceeded the shrunken cache.
    - Fixed by passing TrainingConfig(seq_len=rope.max_seq_len) so the cache stays at the model's intrinsic maximum. Removed _extend_rope_cache and self.rope_config which are no longer needed. Added a max_model_len <= rope.max_seq_len enforcement.

NOTE: In next PR, I plan to change current implementation of all layers share one RoPE cache, to be each layer has one RoPE cache.

### DIFF
--- a/torchtitan/experiments/rl/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/models/vllm_wrapper.py
@@ -22,7 +22,6 @@ from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.config import (
     ActivationCheckpointConfig,
     CompileConfig,
-    DebugConfig,
     ParallelismConfig,
     TrainingConfig,
 )
@@ -184,21 +183,13 @@ class VLLMModelWrapper(Module):
         # Fill sharding configs on the config BEFORE build so every sub-module
         # is constructed with its ShardingConfig attached (required by the
         # declarative model.parallelize() API).
-        # NOTE: We pass seq_len=rope.max_seq_len so the RoPE cache stays at
-        # the model's intrinsic maximum. Using the default TrainingConfig()
-        # would shrink it to 2048, which is wrong for inference — vLLM can
-        # generate positions up to max_model_len (derived from the model's
-        # max_position_embeddings).
-        # TODO: Refactor update_from_config to separate parallelism setup
-        # from training-specific config (rope resize, debug flags).
+        # Only ``parallelism`` is passed; ``training`` is omitted so the
+        # model's intrinsic ``rope.max_seq_len`` is preserved (the inference
+        # path must not shrink the RoPE cache to a training seq_len).
         from types import SimpleNamespace
 
         self.config.update_from_config(
-            trainer_config=SimpleNamespace(
-                training=TrainingConfig(seq_len=self.config.rope.max_seq_len),
-                parallelism=parallelism,
-                debug=DebugConfig(),
-            )
+            trainer_config=SimpleNamespace(parallelism=parallelism)
         )
 
         # Enforce that vLLM's max_model_len does not exceed the model's
--- a/torchtitan/experiments/rl/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/models/vllm_wrapper.py
@@ -22,6 +22,7 @@ from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.config import (
     ActivationCheckpointConfig,
     CompileConfig,
+    DebugConfig,
     ParallelismConfig,
     TrainingConfig,
 )
@@ -183,14 +184,21 @@ class VLLMModelWrapper(Module):
         # Fill sharding configs on the config BEFORE build so every sub-module
         # is constructed with its ShardingConfig attached (required by the
         # declarative model.parallelize() API).
-        # Only ``parallelism`` is needed here; ``training`` is intentionally
-        # omitted so the model's intrinsic ``rope.max_seq_len`` is preserved
-        # (the vLLM inference path must not shrink the RoPE cache to the
-        # trainer's default seq_len).
+        # NOTE: We pass seq_len=rope.max_seq_len so the RoPE cache stays at
+        # the model's intrinsic maximum. Using the default TrainingConfig()
+        # would shrink it to 2048, which is wrong for inference — vLLM can
+        # generate positions up to max_model_len (derived from the model's
+        # max_position_embeddings).
+        # TODO: Refactor update_from_config to separate parallelism setup
+        # from training-specific config (rope resize, debug flags).
         from types import SimpleNamespace
 
         self.config.update_from_config(
-            trainer_config=SimpleNamespace(parallelism=parallelism)
+            trainer_config=SimpleNamespace(
+                training=TrainingConfig(seq_len=self.config.rope.max_seq_len),
+                parallelism=parallelism,
+                debug=DebugConfig(),
+            )
         )
 
         # Enforce that vLLM's max_model_len does not exceed the model's
--- a/torchtitan/experiments/rl/models/vllm_wrapper.py
+++ b/torchtitan/experiments/rl/models/vllm_wrapper.py
@@ -22,7 +22,6 @@ from torchtitan.components.checkpoint import CheckpointManager
 from torchtitan.config import (
     ActivationCheckpointConfig,
     CompileConfig,
-    DebugConfig,
     ParallelismConfig,
     TrainingConfig,
 )
@@ -184,23 +183,30 @@ class VLLMModelWrapper(Module):
         # Fill sharding configs on the config BEFORE build so every sub-module
         # is constructed with its ShardingConfig attached (required by the
         # declarative model.parallelize() API).
-        # TODO: Refactor update_from_config to accept ParallelismConfig
-        # directly instead of requiring a trainer_config wrapper.
+        # Only ``parallelism`` is needed here; ``training`` is intentionally
+        # omitted so the model's intrinsic ``rope.max_seq_len`` is preserved
+        # (the vLLM inference path must not shrink the RoPE cache to the
+        # trainer's default seq_len).
         from types import SimpleNamespace
 
         self.config.update_from_config(
-            trainer_config=SimpleNamespace(
-                training=TrainingConfig(),
-                parallelism=parallelism,
-                debug=DebugConfig(),
-            )
+            trainer_config=SimpleNamespace(parallelism=parallelism)
         )
+
+        # Enforce that vLLM's max_model_len does not exceed the model's
+        # intrinsic max_seq_len -- positions beyond this are invalid.
+        max_model_len = vllm_config.model_config.max_model_len
+        rope_max_seq_len = self.config.rope.max_seq_len
+        if max_model_len > rope_max_seq_len:
+            raise ValueError(
+                f"vLLM max_model_len ({max_model_len}) exceeds the model's "
+                f"maximum supported sequence length ({rope_max_seq_len}). "
+                f"Set max_model_len <= {rope_max_seq_len} in EngineArgs or "
+                f"use a model with a larger rope.max_seq_len."
+            )
 
         # TODO: Check if it's possible to apply meta init
         self.model = self.config.build()
-
-        # RoPE config from model for cache extension
-        self.rope_config = self.config.rope
 
         # With TP, collectives may return AsyncCollectiveTensor (overlap
         # path) or plain Tensor (sync path) depending on timing.  Dynamo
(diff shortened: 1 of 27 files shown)


torchtitan/models/deepseek_v3/model.py
```diff
@@ -196,8 +195,11 @@ def update_from_config(
             debug = trainer_config.debug
             seq_len = training.seq_len
             if seq_len > self.rope.max_seq_len:
-                logger.warning(
-                    f"Sequence length {seq_len} exceeds original maximum {self.rope.max_seq_len}."
+                raise ValueError(
+                    f"Training sequence length {seq_len} exceeds model's "
+                    f"maximum supported sequence length "
+                    f"{self.rope.max_seq_len}. The model cannot produce "
+                    f"valid RoPE embeddings for positions beyond this limit."
```
tianyu-l:
```suggestion
                raise ValueError(
                    f"Training sequence length {seq_len} exceeds model's "
                    f"maximum supported sequence length "
                    f"{self.rope.max_seq_len}."
```
The last sentence sounds inaccurate so just remove.

tianyu-l: also, given it's used in every decoder model, maybe we should put common things in `Decoder.Config.update_from_config()`, then call it using super() in subclasses.

torchtitan/models/llama4/model.py
```diff
@@ -128,14 +127,24 @@ def update_from_config(
...
             for layer_cfg in self.layers:
                 if layer_cfg.moe is not None:
+                    if (
+                        layer_cfg.moe.experts.use_grouped_mm
+                        and not has_cuda_capability(9, 0)
+                    ):
+                        logger.warning(
+                            "Failed to use grouped mm, which is only supported on SM90 or later",
+                        )
+                        layer_cfg.moe.experts.use_grouped_mm = False
```
tianyu-l: I think you need to rebase

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -184,24 +187,38 @@ def __init__(
...
-                training=TrainingConfig(),
+                training=TrainingConfig(seq_len=self.config.rope.max_seq_len),
                 parallelism=parallelism,
                 debug=DebugConfig(),
             )
         )
 
+        # Enforce that vLLM's max_model_len does not exceed the model's
+        # intrinsic max_seq_len -- positions beyond this are invalid.
+        max_model_len = vllm_config.model_config.max_model_len
```
tianyu-l: Where is this coming from, not the new config parser? If so it's not possible to raise below??

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -184,24 +187,38 @@ def __init__(
...
+        # the model's intrinsic maximum. Using the default TrainingConfig()
+        # would shrink it to 2048, which is wrong for inference — vLLM can
+        # generate positions up to max_model_len (derived from the model's
+        # max_position_embeddings).
+        # TODO: Refactor update_from_config to separate parallelism setup
+        # from training-specific config (rope resize, debug flags).
         from types import SimpleNamespace
 
         self.config.update_from_config(
             trainer_config=SimpleNamespace(
```
tianyu-l: this is only a workaround; the proper fix is to let `update_from_config` not tie to "trainer config"

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -184,24 +187,38 @@ def __init__(
...
-                training=TrainingConfig(),
+                training=TrainingConfig(seq_len=self.config.rope.max_seq_len),
```
tianyu-l: We shouldn't create this training config ad hoc -- that's the source of bug.

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: Now I also wonder what's the default behavior in native vllm -- does it always init a rope cache of max_model_len, regardless of whether it'll be all used? E.g. what if max_model_len=1M, but typical generation length is < 10k, would there just be waste?

I ask because we can dynamically adjust in sft trainer / rl trainer, but seems not in our generator.

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: I was thinking to make trainer_config optional in update_from_config.  In inference path, we should just not passing the training config. Let me update

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: No it's not from the new config parser, it was there way before that. 

The check following is for our correct version and prevent vllm passing any max_model_len from `vllm.EngineArgs(max_model_len=xxx)`: we should correct the update_from_config during inference path, don't let trainconfig.seq_length to override the model's max_sequence length.

wwwjn: https://github.com/vllm-project/vllm/blob/257af77bc2b612d5ebd0aecea777139036543af3/vllm/model_executor/layers/rotary_embedding/base.py#L58-L63 In vllm's native Qwen3 model, they do init rope cache of max_model_len, not incrementally increase the length of RoPE cache.

torchtitan/models/deepseek_v3/model.py
wwwjn: Updated, super() call fails with slots=True dataclasses in subclasses, so I used Decoder.Config.update_from_config() directly

torchtitan/models/llama3/model.py
```diff
@@ -75,15 +74,10 @@ def update_from_config(
             trainer_config,
             **kwargs,
         ) -> None:
-            training = trainer_config.training
+            Decoder.Config.update_from_config(
```
tianyu-l: hmm can we not do `super().update_from_config` because of `slots=True`?

torchtitan/models/llama3/model.py
tianyu-l, line 91: can we not put things above in `Decoder.update_from_config()`?

torchtitan/models/llama4/model.py
tianyu-l: need to rebase, this field is gone

torchtitan/models/qwen3_vl/model.py
```diff
@@ -64,22 +63,18 @@ def update_from_config(
...
-                if layer_cfg.moe is not None:
-                    layer_cfg.moe.router._debug_force_load_balance = (
-                        debug.moe_force_load_balance
-                    )
+            if debug is not None:
+                for layer_cfg in self.layers:
+                    if layer_cfg.moe is not None:
+                        layer_cfg.moe.router._debug_force_load_balance = (
+                            debug.moe_force_load_balance
+                        )
```
tianyu-l: I even think this can also go to Decoder parent

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: > prevent vllm passing any max_model_len from `vllm.EngineArgs(max_model_len=xxx)`

`vllm.EngineArgs(max_model_len=xxx)` is called by us, we should set it there instead of raise here (like, we are not raising on other fields mismatch) https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L336
- curious what's the default for `max_model_len` if we don't set?
- how's `max_model_len` different from https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/models/vllm_registry.py#L94

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -184,24 +183,30 @@ def __init__(
...
+        # path must not shrink the RoPE cache to a training seq_len).
         from types import SimpleNamespace
 
         self.config.update_from_config(
-            trainer_config=SimpleNamespace(
-                training=TrainingConfig(),
-                parallelism=parallelism,
-                debug=DebugConfig(),
-            )
+            trainer_config=SimpleNamespace(parallelism=parallelism)
```
tianyu-l: > I was thinking to make trainer_config optional in update_from_config. In inference path, we should just not passing the training config. Let me update

This is not done yet, right?

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: it doesn't have a default number: 

- https://docs.vllm.ai/en/latest/configuration/engine_args/#-max-model-len It doesn't have a default, seems it will find the largest max_model_len that fit into memory, starting from the model's `max_position_embeddings`
- vllm use max_position_embeddings to initialize RoPE, and use `max_model_len` to specify max prompt length.

So the bug is:
1. Engine initialize first, set max_model_len to be max_position_embeddings ,which is 4096.
2. Then model's max_sequence_length  is updated by `update_from_config` because it passed TrainConfig() which has default seq_len=2048. 

I think a clear way of doing it: We can pass the `EngineArgs(max_model_len=xxx)` explicitly, and we translate vllm_config.max_model_len to be Torchtian's TrainingConfig.seq_length. This would be compatible with our current update_from_config check. Wdyt?

torchtitan/models/llama3/model.py
wwwjn: Yes, super().update_from_config error out because of slots=True

torchtitan/experiments/rl/models/vllm_wrapper.py
wwwjn: This is not done yet, as we discussed offline yesterday, I will update the type hint of  `trainer_config: Configurable` in Trainer and perform check when `isinstance(trainer_config, Trainer.Config)`

torchtitan/experiments/rl/models/vllm_wrapper.py
tianyu-l: > We can pass the EngineArgs(max_model_len=xxx) explicitly

Yes we should do this.

> we translate vllm_config.max_model_len to be Torchtian's TrainingConfig.seq_length

why do we need this? In generator we shouldn't specify `seq_len` at all.

torchtitan/models/llama3/model.py
wwwjn: Do you mean we should put all these check in common `Deocer.update_from_config()`?

The TP head divisibility check (n_heads % tp, n_kv_heads % tp) is repeated across  all 6 models  so we can move. The CP and weight-tying checks are different across model (llama4 bans CP entirely, deepseek_v3 checks for SDPA specifically, other models don't have weight tying), so those are harder to unify.

torchtitan/experiments/rl/models/vllm_wrapper.py
```diff
@@ -183,25 +183,21 @@ def __init__(
...
+        class _InferenceConfig:
+            parallelism: ParallelismConfig
 
         self.config.update_from_config(
-            trainer_config=SimpleNamespace(
-                training=TrainingConfig(),
-                parallelism=parallelism,
-                debug=DebugConfig(),
-            )
+            trainer_config=_InferenceConfig(parallelism=parallelism)
```
wwwjn: I made a separate InfereceConfig class to align with `Decoder.update_from_config` function signature, instead of directly passing parallelism

torchtitan/models/deepseek_v3/model.py
tianyu-l, line 215: I feel this is outdated. FlexCP should just work for every model?

torchtitan/models/common/decoder.py
```diff
@@ -78,6 +81,83 @@ class Config(BaseModel.Config):
         # and fix the typing here
         layers: list  # list[TransformerBlock.Config] or subclass configs
 
+        def update_from_config(
+            self,
+            *,
+            trainer_config,
```
tianyu-l:
```suggestion
            config,
```
as it's not specific to trainer config any more

torchtitan/models/common/decoder.py
```diff
@@ -78,6 +81,83 @@ class Config(BaseModel.Config):
...
+            When *trainer_config* is a ``Trainer.Config``, validates
+            ``training.seq_len`` against the model's intrinsic
+            ``rope.max_seq_len``, resizes the RoPE cache, and propagates
+            debug flags.  Non-trainer callers (e.g. the vLLM inference
+            wrapper) may pass a ``ParallelismConfig`` directly; in that
+            case the training/debug setup is skipped.
+            """
+            from torchtitan.trainer import Trainer
+
+            parallelism = trainer_config.parallelism
```
tianyu-l: assert existence and isinstance before usage

torchtitan/models/common/decoder.py
```diff
@@ -78,6 +81,83 @@ class Config(BaseModel.Config):
...
+                        )
+                        and parallelism.expert_parallel_degree == 1
+                    ):
+                        raise ValueError(
+                            f"{type(token_dispatcher_cfg).__qualname__} "
+                            "requires expert parallelism "
+                            "(expert_parallel_degree > 1)."
+                        )
+
+            if isinstance(trainer_config, Trainer.Config):
```
tianyu-l: add a NOTE on why RL generator doesn't need this -- I think essentially it's not possible to predict how long the generated sequence would be, so we allocate according to model's max seq len capability

tianyu-l (approved):
some final comments

torchtitan/models/deepseek_v3/model.py
wwwjn: cc @fegin  do you know if  FlexCP works for all the models?

wwwjn:
The model test keeps failing, so I patched https://github.com/pytorch/pytorch/pull/183970/changes locally and run the 8GPU integration test locally, ~they passed~. They need #3448 to fix

PR #3396, opened 2026-05-19
[RFC] a registry mechanism for in-repo / out-of-repo Module-level overrides

### LINKED ISSUE #2183: [RFC] Introduce a Plugin-Mechanism for hardware-specific optimazations
**Summary**
This RFC proposes a modular plugin architecture for torchtitan. Inspired by successful extension patterns in vLLM, this mechanism aims to decouple core training logic from hardware-specific optimizations and custom feature extensions (e.g., fused kernel, specialized optimizers, quantization). Currently torchtitan provides some experimental extension points (e.g., in the experimental/ directory or through certain hook interfaces). However, these are insufficient for deep hardware integration

**Proposed Changes**
Firstly, we want to add Kernel-Plugin to enable the integration of platform-specific fused kernel. Including 
1、Dynamic Loading via Entry Points: Utilize Python's entry_points (similar to vLLM) to allow external packages to register themselves automatically.
2、Define the base class CustomOp: Support dispatching and registration (similar to vLLM) 
```
class CustomOp(nn.Module):
    """
    Base class for custom ops.
    Dispatches the forward method to the appropriate backend.
    """

    def __new__(cls, *args, **kwargs):
        try:
            op_name = cls.__name__
        except AttributeError:
            raise TypeError(
                f"Cannot instantiate '{cls.__name__}': its 'name' attribute "
                f"was not set, possibly because it was not decorated with "
                f"@CustomOp.register, or it's the CustomOp base class itself."
            ) from None

        if op_name not in cls.op_registry_oot:
            op_cls_to_instantiate = cls
        else:
            op_cls_to_instantiate = cls.op_registry_oot[op_name]
            logger.debug(
                "Instantiating custom op: %s using %s",
                op_name,
                str(op_cls_to_instantiate),
            )
        return super().__new__(op_cls_to_instantiate)
     ... 
    # Decorator to register custom ops.
    @classmethod
    def register(cls, name: str):
        def decorator(op_cls):
            assert name not in cls.op_registry, f"Duplicate op name: {name}"
            op_cls.name = name
            cls.op_registry[name] = op_cls
            return op_cls

        return decorator
```
3、Modify the model defination using CustomOp, but we can refine this incrementally through subsequent PRs. Alternatively, to minimize the impact on torchtitan's core design, we can just remain current implementation and dynamiclly override by monkey-patch

Feedback is highly welcomed. I’m keen to refine this proposal based on your comments as soon as possible, ensuring we find the best path for hardware extensibility in torchtitan

tianyu-l:
2026/06/07 update:

Addressed most comments, and aimed for landing, with latest design updated in `torchtitan/config/OVERRIDE.md`

In particular, I created a `FusedSwiGLU` override, but admitted that checkpoint incompatibility between fused and non-fused version. Issue tracked in https://github.com/pytorch/torchtitan/issues/3569.

---------------------------------

With the following main goals

**[GOAL 1]**
Provide users a way to use efficient kernels that are not "native implementations".
- main goal is to provide decent performance on ops that don't have efficient implementations in pytorch core, e.g. [RoPE](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/rope.py) with various context-length extension methods.
- lower bar on code maintainability for kernel code; can iterate fast. Similar gist to how we look at agent tooling https://github.com/pytorch/torchtitan/pull/3323#pullrequestreview-4301364966
- doesn't aim to support all hardware, may start with Triton for H100 (available in CI)
- if the `torchtitan/registry` folder grows too quick / becomes messy, we probably should create a separate repo to host overrides

**[GOAL 2]**
Provide team a way to experiment with and showcase non-trivial implementations, without making intrusive changes to the rest of repo
- e.g. fused quantized MoE (cc @vkuzo @danielvegamyhre)

**[GOAL 3]**
Provide hardware vendors a way to override modules, without touching the main repo.
- addresses https://github.com/pytorch/torchtitan/issues/2183
- e.g. https://github.com/pytorch/torchtitan/pull/3266 could be applied without upstreaming, optionally with its own pip package

**[NO GOAL]**
- op-level overrides: one should consider pytorch native ways, e.g. via `torch.library`
- the proposal is orthogonal to graph capture -- it is override author's responsibility to make sure the registered module can be traced (cc @SherlockNoMad @xmfan @aditvenk)

**Please read torchtitan/registry/DESIGN.md for details.** Feedback is welcome.

Additional notes
- [alternative proposal](https://gist.github.com/syed-ahmed/169d3bf96eedefd51e4a001a02a617b8) from @syed-ahmed

### DIFF
--- a/torchtitan/config/OVERRIDE.md
+++ b/torchtitan/config/OVERRIDE.md
@@ -1,0 +1,515 @@
+# Configurable Override Mechanism
+
+Swap any `Configurable` in a torchtitan training config — a model component,
+optimizer, loss, dataloader, and so on — for an alternative implementation
+without editing in-repo code, by importing a module that registers an override.
+This is the reference for how the mechanism works and how to write overrides.
+
+## Motivation
+
+Torchtitan has a config-driven build system: every component defines a nested
+`Config` dataclass and `config.build()` constructs the owning object. Swapping a
+component (e.g. Float8 quantization, LoRA) currently works via
+`ModelConfigConverter`, which traverses the model config tree and replaces
+`Config` nodes during config construction, inside `config_registry.py` functions.
+
+That works well for in-repo, first-class features, but it requires editing the
+repo for *every* alternative implementation. The override mechanism removes that
+requirement for three classes of work:
+
+1. **Hardware-specific kernels.** Efficient Triton/CUDA kernels for rotary
+   embeddings, fused attention, or custom norms that target specific hardware.
+   These are valuable for performance but are (a) hardware-specific, (b) harder
+   to maintain, and (c) reasonably held to a lower code-quality bar than core.
+   We want to offer vendors a clean extension point rather than either lowering
+   the core bar or saying "no".
+
+2. **Larger fused regions.** An efficient MoE may fuse dispatch–compute–combine
+   with custom routing, communication, and quantization. This spans several
+   `Config` nodes and is best expressed as a single replacement at a higher
+   scope (e.g. replace `MoE.Config` wholesale).
+
+3. **Team experimentation.** Trying a non-trivial implementation without
+   threading intrusive changes through core, then keeping or discarding it
+   cheaply.
+
+### In-repo vs. external
+
+The same mechanism serves both in-repo examples (`torchtitan/overrides/`) and
+external packages, and the decision of where an implementation lives maps onto
+the goals above:
+
+- **In-repo (`torchtitan/overrides/`)** — illustrative examples and broadly
+  useful kernels we are willing to maintain at a lower bar than core. They are
+  opt-in and never run unless requested, so they cannot regress default
+  behavior.
+- **External package** — vendor- or hardware-specific code, or anything we do
+  not want to maintain in-tree. It lives in the vendor's namespace and is
+  activated by listing its module path.
+
+**Non-goals.** An operator-override API in TorchTitan. To replace or add an op,
+use PyTorch's custom-operator path and wrap the op in a `Module` — see
+[Custom kernels and `torch.compile`](#custom-kernels-and-torchcompile).
+TorchTitan does not provide an op-level override surface.
+
+## Design Overview
+
+Override authors register a factory against a `Configurable.Config` class. After
+config construction and before any `build()`, the mechanism traverses the config
+tree and replaces matching nodes with the factory's output.
+
+### Core principles
+
+- **Explicit opt-in.** Nothing happens unless the user lists modules in
+  `override.imports`. No auto-detection from hardware or installed packages.
+- **Strictly scoped to the request.** Only overrides *registered by the listed
+  modules* (or their submodules) are applied — not the whole global table. This
+  is enforced by provenance (the module a factory is defined in).
+- **Per-instance targeting, per-node conflicts.** An override selects *which*
+  matched nodes it claims via `fqns` (FQN globs). Two overrides conflict only
+  when they claim the *same node* (or one claims an ancestor of the other's
+  node) — not merely the same Config class. So two vendors' norm kernels on
+  different layers, or A/B-ing a kernel per layer, compose cleanly; genuine
+  collisions error out.
+- **Any `Configurable`.** Overrides apply across the entire `Trainer.Config`
+  tree — model components, optimizer, loss, dataloader, etc. — not just the
+  model.
+- **Minimal surface.** The whole mechanism lives in
+  `torchtitan/config/override.py`, reusing the `Configurable.Config.traverse()` +
+  replace pattern that the Float8/LoRA converters already use.
+
+## How It Works
+
+### Registration
+
+An override module defines a replacement and registers a factory with
+`@override`:
+
+`TritonRoPE` below is **illustrative** (a compact subclass-adds-one-field
+example); no `triton_rope.py` is shipped. It is a valid override: each attention
+module owns a `rope` submodule (a `RoPE.Config`), so swapping in a custom RoPE is
+an ordinary component override.
+
+```python
+from dataclasses import dataclass
+
+from torchtitan.config import derive, override
+from torchtitan.models.common.rope import RoPE
+
+
+class TritonRoPE(RoPE):
+    """RoPE applied with a Triton kernel (illustrative)."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(RoPE.Config):
+        block_size: int = 128
+
+    def __init__(self, config: Config): ...
+    def forward(self, query, key, positions=None): ...
+
+
+@override("triton_rope", target=RoPE.Config,
+          description="Triton rotary embedding, ~2x faster on A100+")
+def triton_rope(cfg: RoPE.Config) -> TritonRoPE.Config:
+    # `derive` copies every field shared with RoPE.Config; the factory states
+    # only its deltas. See "Deriving the replacement config" below.
+    return derive(cfg, TritonRoPE.Config, block_size=128)
+```
+
+The factory receives the matched config and returns its replacement. *Which*
+instances it applies to is controlled declaratively by `fqns` (see
+[per-instance targeting](#per-instance-targeting)), not by inspecting fields
+inside the factory.
+
+### Deriving the replacement config
+
+Prefer `derive(cfg, NewConfig, **deltas)` over hand-copying each field. It builds
+`NewConfig` by copying fields shared by name with `cfg`, applying `deltas` on
+top; the factory states only what it actually changes.
+
+This matters for version-robustness. Hand-copying couples a factory to the
+target's field set at the time it was written:
+
+```python
+# Fragile: RoPE.Config has many fields (theta, scaling, rope_factor, beta_fast,
+# ...); hand-copying only some silently drops the rest to their defaults — and a
+# field added later would be dropped too, with no error.
+return TritonRoPE.Config(dim=cfg.dim, max_seq_len=cfg.max_seq_len,
+                         theta=cfg.theta, scaling=cfg.scaling)
+
+# Robust: every RoPE.Config field (including ones added later, inherited by the
+# subclass TritonRoPE.Config) is copied automatically; only block_size is a delta.
+return derive(cfg, TritonRoPE.Config, block_size=128)
+```
+
+Semantics: fields on both `cfg` and `NewConfig` are copied from `cfg`; fields
+only on `NewConfig` come from `deltas` (else their declared default); fields only
+on `cfg` are dropped; a delta naming a field that doesn't exist on `NewConfig`
+raises. Copying is shallow (like `dataclasses.replace`), which is fine because
+the matched node is being replaced and detached. `derive` is most valuable when
+the replacement subclasses the target (the common drop-in case), where new core
+fields are inherited and thus carried over for free.
+
+**Configuration is code.** An override takes no CLI flags (it runs after CLI
+parsing), so the override module *is* its configuration: bake sensible defaults
+into the factory (above, `block_size=128`). A user who wants different values
+imports their own module that registers a factory with their deltas — e.g.
+`derive(cfg, TritonRoPE.Config, block_size=256)`. We deliberately don't add a
+CLI-settable `kwargs` dict (untyped, error-prone) or a typed override-`Config`
+parsed at config-construction time (which would re-couple the config registry to
+the override package, defeating the no-touch goal).
+
+### Activation
+
+```bash
+torchtitan_train --module llama3 --config llama3_8b \
+    --override.imports torchtitan.overrides.fused_swiglu
+```
+
+### Application
+
+In `Trainer.__init__`, after `model_config.update_from_config()` (which sets
+sharding config on the pre-override modules) and before any component is built:
+
+1. Import the listed modules — this triggers their `@override` decorators.
+2. Resolve the *active* set: overrides whose defining module is one of the
+   listed imports (or a submodule of one).
+3. Collect claims: traverse the original tree and record every `(override,
+   node)` pair, filtered by each override's `fqns` selector.
+4. Check for per-node conflicts (two overrides claiming the same node, or one
+   claiming an ancestor of another's node).
+5. Apply: build each replacement and swap it in. Because claims are gathered
+   before any mutation, application is order-independent (replacements are never
+   re-traversed).
+6. Log every replacement.
+
+```
+INFO: [Override] fused_swiglu: model_spec.model.layers.0.feed_forward FeedForward.Config -> FusedSwiGLU.Config
+INFO: [Override] fused_swiglu: model_spec.model.layers.1.feed_forward FeedForward.Config -> FusedSwiGLU.Config
+...
+INFO: Applied 32 override(s)
+```
+
+The model config is reached even though it is nested under a non-`Configurable`
+`ModelSpec`: `ModelSpec.traverse` exposes its `model` entry to the traversal.
+FQNs are kept as full paths from the `Trainer.Config` root — the model config is
+`model_spec.model` and a component is `model_spec.model.layers.0.feed_forward`.
+Preserving the full path (rather than resetting to bare model names) is what lets
+per-node conflict detection recognize a whole-model override as an ancestor of a
+component override. (This differs from converter `filter_fqns`, which are bare
+because converters traverse the model config directly.)
+
+### External packages
+
+```python
+# vendor_x/overrides.py
+from torchtitan.config import override
+from torchtitan.models.common.moe import MoE
+
+
+@override("vendor_x_fused_moe", target=MoE.Config,
+          description="Vendor X fused MoE for XPU")
+def vendor_x_moe(cfg: MoE.Config) -> "VendorXFusedMoE.Config":
+    return VendorXFusedMoE.Config(num_experts=cfg.num_experts, ...)
+```
+
+```bash
+pip install torchtitan-vendor-x
+torchtitan_train --module deepseek_v3 --config dsv3_671B \
+    --override.imports vendor_x.overrides
+```
+
+### Version compatibility
+
+An override depends on the internal `Config` shapes it reads and the modules it
+imports, so external packages should:
+
+- **Pin a torchtitan version range** they have validated against. The override
+  contract is the public `Configurable.Config` fields of the targets; these can
+  change across releases.
+- **Align their pytorch-nightly expectation** with the torchtitan version they
+  pin, since hardware kernels are typically nightly-sensitive. Document the
+  tested torchtitan × pytorch pair in the package.
+
+## Per-instance Targeting
+
+Targeting is first-class, not an afterthought: many motivating cases collide by
+design (two vendors with norm kernels, or A/B-ing a kernel per layer). The
+`fqns` selector on `@override` says *which* matched nodes an override claims,
+declaratively — no field-sniffing inside the factory.
+
+`fqns` is one of:
+
+- `None` (default) — every instance of `target`.
+- a list of FQN globs — `fnmatch` style, where `*` also crosses `.`; a node is
+  claimed if any glob matches its FQN.
+
+The FQN is the full path from the `Trainer.Config` root, e.g. a model component
+is `model_spec.model.layers.0.feed_forward` and the optimizer is `optimizer`.
+Globs with `*` (which crosses `.`) keep selectors readable.
+
+```python
+# NVFP4 on every in-layer linear, leaving the top-level LM head ("...output")
+# untouched (required for NVFP4; useful for MXFP8)
+@override("nvfp4_linear", target=Linear.Config,
+          fqns=["*.layers.*.attention.*", "*.layers.*.feed_forward.*"])
+def nvfp4_linear(cfg: Linear.Config) -> "NVFP4Linear.Config":
+    return NVFP4Linear.Config(in_features=cfg.in_features, ...)
+
+# Fused MoE only on the later layers
+@override("vendor_moe", target=MoE.Config, fqns=["*.layers.1[0-9].moe"])
+def vendor_moe(cfg: MoE.Config) -> "VendorMoE.Config":
+    return VendorMoE.Config(num_experts=cfg.num_experts, ...)
+```
+
+`fqns` accepts globs only. A general predicate selector (`(fqn, cfg) -> bool`),
+which would also enable field/type-based selection (e.g. "skip already-quantized
+linears"), is listed under [Future work](#future-work).
+
+### Per-node conflict detection
+
+Conflicts are detected per *node*, not per *class*. Two overrides may target the
+same Config class freely as long as their claimed nodes are disjoint — so one
+norm kernel on `*.layers.0.*` and another on `*.layers.1.*` coexist, and A/B-ing
+a kernel across layers works. An error is raised only when two different overrides
+claim the **same node**, or when one claims an **ancestor** of another's node
+(nested replacement, which would be order-dependent). The fix in both cases is
+to narrow the `fqns` selectors so the claims are disjoint.
+
+Claims are gathered from the **original** tree before any mutation, and a
+replacement is never re-traversed, so application is order-independent — one
+override can never silently affect what another matches.
+
+#### Parent- and child-class overrides
+
+This same rule answers the parent/child question directly. Say override A
+targets a parent Config (e.g. `MoE.Config`) and override B targets a Config
+nested inside it (e.g. `GroupedExperts.Config`):
+
+- **Disjoint subtrees** (A on `...layers.0.moe`, B on `...layers.1.moe.experts`)
+  — the claimed nodes are unrelated, so both apply.
+- **Overlapping** (A on `...layers.0.moe`, B on `...layers.0.moe.experts`) — B's
+  node is inside A's, the ancestor case above, so we **error**.
+
+We error rather than pick one of the two plausible behaviors implicitly:
+
+1. *Apply the parent, then override a child in the new parent.* Only meaningful
+   if A's replacement still exposes a matching child, and it reintroduces
+   ordering (parent-before-child) and re-traversal — which we deliberately avoid.
+2. *Silently drop the child override* (when A's replacement has a different
+   structure with no such child). A silent no-op is exactly the kind of
+   surprise that hides bugs.
+
+Because claims are collected on the original tree, we can't even tell at
+check time which case a given pair would fall into — so instead of guessing, we
+ask the author to disambiguate: narrow `fqns` so the claims are disjoint, or drop
+one override (e.g. let the parent override subsume the child).
+
+## Scope: any `Configurable`
+
+Overrides traverse the whole `Trainer.Config`, so the optimizer, loss,
+dataloader, and validator configs are overridable too — not only model
+components. For example, an emerging or mixed-precision optimizer can be swapped
+in by targeting `OptimizersContainer.Config` without editing a config registry
+function. The model config is reached via `ModelSpec.traverse` (above); the model
+config itself is a valid target (whole-model swap), while `ModelSpec` is not — a
+`target` must be a `Configurable.Config` subclass, so a plain class like
+`ModelSpec` is rejected at registration.
+
+## Interaction with Converters
+
+In-repo converters (Float8, LoRA via `ModelConfigConverter`) run *first*, inside
+`model_registry()` during config construction; overrides run later in
+`Trainer.__init__` and see the post-converter tree. The order is deliberate: an
+in-repo converter cannot be expected to understand arbitrary external overrides,
+so it runs against the known core configs, and overrides layer on top.
+
+Conversely, converter-style transforms *can* be expressed as overrides (a
+factory that rewrites a `Config`), so external code does not need the converter
+machinery.
+
+| Override Target | Conflicts with a converter? | Notes |
+|-----------------|------------------------------|-------|
+| RoPE / FeedForward / MoE / RMSNorm / inner attention | No | Converters don't touch these |
+| GroupedExperts.Config | Possibly | `Float8GroupedExpertsConverter` rewrites this |
+| Linear.Config | Yes | Float8/LoRA replace these |
+
+Where a converter already rewrote a node, target that node by location with
+`fqns` so the override only claims the instances you intend (e.g. specific
+layers the converter left as plain `Linear.Config`). Field/type-based exclusion
+— "skip instances already turned into `Float8Linear.Config`" — is not expressible
+with FQN globs alone and motivates the future predicate selector.
+
+### Override vs. converter: which to use
+
+Converters are the right home for **core, broadly-applicable transforms that
+preserve a property we guarantee** (e.g. batch-invariance, which is largely
+op-level and must hold across the model). Overrides exist for an *unblocking*
+purpose: swapping in an alternative implementation. By design we make **no
+promise that any core property survives an arbitrary override** — that is the
+override author's responsibility. So features like batch-invariant mode stay
+converters, not overrides.
+
+## Checkpoint Compatibility
+
+An override that changes a module's parameter layout changes its checkpoint
+FQNs. An override checkpoints whatever real parameters it defines, and is
+interoperable with another checkpoint only when the FQNs match. The fused
+example (`fused_swiglu.py`) defines one `w13` parameter, so it saves/loads
+`...feed_forward.w13` — **not** the stock `w1.weight` / `w3.weight`. A fused
+checkpoint round-trips with another fused run; it is *not* a drop-in for stock
+checkpoints (and vice-versa).
+
+**Module-level `state_dict` hooks do NOT bridge this** — neither the low-level
+`_save_to_state_dict` / `_load_from_state_dict` nor the public `state_dict()` /
+`load_state_dict()`. It is tempting to present `w13` as `w1.weight` / `w3.weight`
+that way. It works under plain `nn.Module.load_state_dict`, but torchtitan
+checkpoints via DCP `get_model_state_dict` / `set_model_state_dict`, which read
+the keys from `state_dict()` and then run `_get_fqns()` on every key — resolving
+it back to a *real* module attribute via `getattr` (DCP needs the actual param
+object for DTensor/FSDP metadata and in-place load). A fabricated `w1` key has no
+backing attribute, so both paths raise `'FusedSwiGLU' object has no attribute
+'w1'` regardless of which method produced the key. The one escape `_get_fqns`
+offers, `_fqn_modifiers`, only renames/skips an intermediate attribute name; it
+cannot split one parameter into two keys. Do not use this approach.
+
+**The sanctioned mechanism is a model-level `BaseStateDictAdapter`** — the same
+hook used for HF conversion (`Llama3StateDictAdapter`). An adapter transforms the
+flat key→tensor dict *after* `get_model_state_dict`, so it can split `w13` into
+`w1.weight`/`w3.weight` (and recombine on load) without any attribute-resolution
+problem. The catch: the adapter is set on `ModelSpec` (per model), and the
+override mechanism does not expose a hook for an override to contribute one. So
+layout-changing overrides are self-consistent but not stock-compatible — until
+an override-contributed adapter hook exists (see [Future work](#future-work)),
+treat them as checkpoint-incompatible with stock, or pair them with a model whose
+`state_dict_adapter` performs the mapping.
+
+A candidate design (tracked in
+[#3569](https://github.com/pytorch/torchtitan/issues/3569)) is to add an optional
+`state_dict_translator` argument to `@override` that is plugged into the model's
+existing `from_hf` / `to_hf` conversion. The override would declare the
+fused↔canonical FQN mapping (e.g. `w13` ↔ `w1.weight`/`w3.weight`) alongside its
+factory, and the adapter would apply it on save/load — letting an override use
+its own internal FQNs while still reading/writing checkpoints in the canonical
+format.
+
+## Parallelism
+
+Parallelism is expressed entirely through the `Module` protocol: a replacement
+satisfies `init_states` and declares a `ShardingConfig` for the states and
+activations it wants sharded. `Module.parallelize()` reads that `ShardingConfig`
+exactly as it does for core modules, so an override composes with TP/FSDP by
+declaring its own sharding — nothing model-specific is required, and the
+mechanism deliberately stays config-driven rather than depending on imperative
+per-model parallelization code.
+
+One thing worth stating plainly:
+
+- **Fusion under TP.** Fusing weights can interact subtly with tensor
+  parallelism — the fused tensor's layout must admit a correct shard. A *flat*
+  fused gate+up weight `(2*hidden, dim)` has none (`spmd.S(0)` would hand one rank
+  all of `w1` and another all of `w3`). The fix is a layout whose TP-sharded axis
+  gives each rank matching slices: `fused_swiglu` stores `w13` as
+  `(hidden, 2, dim)` and shards `spmd.S(0)` on `hidden`, so each rank holds
+  `(hidden/tp, 2, dim)` — a slice of both gate and up (the Megatron
+  column-parallel layout). `hidden` is also dim 0, so FSDP shards it cleanly at
+  any degree. The single GEMM is an `einsum` that keeps `hidden` sharded, so it
+  never reshapes across the sharded axis. This composes with FSDP and TP through
+  the ordinary `ShardingConfig`; no model-specific code.
+
+## Custom kernels and `torch.compile`
+
+An override that wraps a custom CUDA or Triton kernel must stay compatible with
+`torch.compile`, because torchtitan compiles the transformer blocks by default
+(`compile.components` includes `"model"`). A raw kernel call is opaque to Dynamo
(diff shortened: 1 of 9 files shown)


torchtitan/registry/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    by_target: dict[type, list[str]] = defaultdict(list)
+    for override in _REGISTRY.values():
+        by_target[override.target_cls].append(override.name)
+
+    conflicts = {target: names for target, names in by_target.items() if len(names) > 1}
+    if conflicts:
+        lines: list[str] = []
+        for target, names in conflicts.items():
+            lines.append("  " + target.__qualname__ + ": " + ", ".join(sorted(names)))
+        raise ValueError(
```
drisspg: if you have two overrides 1 targeting a parent of another do you think ti should also warn?

tianyu-l: Great point. In that case, I'd just error out (in first version). WDYT?

tests/unit_tests/test_registry.py
```diff
@@ -0,0 +1,161 @@
...
+        self.assertIsInstance(
+            parent_cfg.children[1], ComponentA.Config
+        )  # dim=48, skipped
+
+    def test_duplicate_name_raises(self):
+        @register("dup", target=ComponentA.Config)
+        def first(cfg):
+            return cfg
+
+        with self.assertRaises(ValueError, msg="already registered"):
```
drisspg: I think you want rasiesRegex

torchtitan/registry/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+            importlib.import_module(mod_path)
+        except ImportError as e:
+            raise ImportError(
+                f"Failed to import override module '{mod_path}': {e}"
+            ) from e
+
+    _check_conflicts()
+
+    replacements: list[str] = []
+    for override in _REGISTRY.values():
```
drisspg: nit: youare potentially applying all mods in the registry even if not requested in override config

torchtitan/trainer.py
```diff
@@ -250,6 +252,11 @@ def __init__(self, config: Config):
         )
         self.model_config = model_config
 
+        if config.override.modules:
+            from torchtitan.registry import apply_overrides
+
+            apply_overrides(model_config, config.override)
```
drisspg: had a long convo with codex about the odering of:

> The intended contract is that factories receive the post-`update_from_config` config and are responsible for translating/preserving relevant fields into the replacement config. I agree this should be documented explicitly, especially for common Module.Config fields like param_init and sharding_config, and we can add a regression test to lock that expectation down. 

And I kind of agree that if we can have some programmatic way to enforce the overrides makes a config matching copy that would be good, or maybe we just re-order?

torchtitan/registry/__init__.py
drisspg: Yeah I think that policy makes sense

torchtitan/registry/__init__.py
tianyu-l: Good catch. In the current version, if a mod appears in the import chain, it'll be applied regardless if it's in override config or not. I think my intention was to strictly limit to things in the override config, to be explicit.

torchtitan/trainer.py
tianyu-l: > And I kind of agree that if we can have some programmtic way to enforce the overrides makes a config matchign copy that would be good, or maybe we just re-order?

By re-order I assume you mean put `apply_overrides` earlier, say before `update_from_config`? I don't think we could do that, because `update_from_config` will set sharding config that's specific to the pre-override module.

I deliberately put `apply_overrides` after `update_from_config` before model build, to give override a chance to do a complete replacement, including `param_init` and `sharding_config`. I feel that even if we don't add additional check, the existing code will error out at the places that consume these fields, but I agree that providing helpful error messages at proper place would be good, for both humans and agents.

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+4. Log every replacement with FQN and type change.
+
+```
+INFO: [Override] triton_rope: layers.0.rope RoPE.Config -> TritonRoPE.Config
+INFO: [Override] triton_rope: layers.1.rope RoPE.Config -> TritonRoPE.Config
+...
+INFO: Applied 32 module overrides
+```
+
+### External Packages
```
xmfan: Do you have a criteria for landing in-repo vs keeping as external in mind?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+def vendor_x_moe(cfg: MoE.Config) -> VendorXFusedMoE.Config:
+    return VendorXFusedMoE.Config(
+        num_experts=cfg.num_experts,
+        top_k=cfg.router.top_k,
+        ...
+    )
+```
+
+```bash
+pip install torchtitan-vendor-x
```
xmfan: what's the thinking on version compatibility? should we recommend`torchtitan-vendor-x` to pin against a torchtitan version? the external packages depend on titan's module signatures.

tianyu-l: There are two layers
1. Being able to modify torchtitan config without touching torchtitan code. This is fundamental capability, unblocking development inside/outside the team, which is what I wanted to focus on in this RFC.
2. Above that, there is this version compatibility, which can be (and I think has to be) handled in the way you suggested. I agree it's good to clarify it in the RFC itself.

syed-ahmed: How about pytorch version? Should vendor package also walk along nightly pytorch similar to TorchTitan? Wondering how to avoid this scenario in the future and reduce maintenance burden from TorchTitan: https://github.com/pytorch/torchtitan/pull/3390.

torchtitan/registry/DESIGN.md
tianyu-l: Right now I can think of three layers, corresponding to the three goals.
- [GOAL 1] If we don't land it, the perf will look awkward (whether it's pretrain fwd-bwd / RL generator fwd-only) and thus hurt torchtitan. Then let's land it in-repo, which also serves as an example of how to use this plug-in mechanism. It can be for specific common hardware we have access to in CI.
- [GOAL 2] Our team may develop experimental features in torchtitan, in the `experiments` folder. They can choose to use existing `ModelConfigConverters` (more explicit), or to use this registration mechanism (less explicit, but doesn't need to create extra `config_registry`). Not a huge difference here.
- [GOAL 3] For external community to showcase their advanced technologies which we are not ready to depend on in-repo (including hardware that we don't have access in CI), we recommend them being out-of-repo.

torchtitan/registry/DESIGN.md
tianyu-l: torchtitan should publish stable versions that align with pytorch stable versions. Vendor packages probably only needs to care these versions backed by pytorch stable releases.

torchtitan could publish more frequently as we are doing active (BC-breaking) development, or working closely with vendor to unblock new features that vendor would like to showcase. But in this case, vendor package has to be OK with some pytorch nightly being pinned.

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+embedding application is done by free functions (`apply_rotary_emb_complex`, etc.)
+called inside `GQAttention.forward()`. To make RoPE replaceable at the Module level,
+the application step must be part of the RoPE Module.
+
+**Plan:** For v1, pass the RoPE Module (not just its cache tensor) through the forward
+chain. `GQAttention.forward()` calls `rope(xq, xk, positions)` instead of
+`apply_fn(xq, xk, cache, positions)`. In a follow-up, each attention layer will own
+its own RoPE module as a submodule, making it naturally replaceable via the config tree.
+
+## Parallelism Compatibility
```
SherlockNoMad: IIUC, CP is applied on the Attention module, and it's composability interface to consider.

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+## Open Questions for Feedback
+
+1. **Naming:** "Module Override Registry", "Kernel Registry", "Implementation Registry"?
+   Current proposal uses "override" throughout.
+
+2. **Checkpoint compatibility:** Override authors are responsible for state_dict key
+   compatibility. Should the registry enforce or validate this?
+
+3. **Scope of traverse:** Currently limited to `model_config` (the model's Config tree).
+   Should it extend to the full `Trainer.Config` tree (optimizer, dataloader, etc.)?
```
syed-ahmed: Here are couple use cases for overriding optimizers:
- mixed precision optimizers that we had deeply integrated with torchtitan/pytorch before: https://github.com/pytorch/torchtitan/pull/2732, https://github.com/pytorch/pytorch/issues/146542
- emerging optimizers: https://github.com/NVIDIA-NeMo/Emerging-Optimizers.

torchtitan/registry/DESIGN.md
tianyu-l: With recent refactor, all SPMD parallelism (other than FSDP) would be encoded in sharding config; in the near future (e.g. when FlexShard is ready), the storage placement would also be put in config.

So it's the author's responsibility to make sure correct sharding config is provided during config swapping.

A concrete example, if someone would like to implement fused ring-based CP, they can do
- assuming inputs and outputs are sharded on sequence dimension
- use `LocalMapConfig` as the contract for above
- explicit call send/recv/all-to-all collectives in the CP Attention implementation

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+No changes needed to any model's `config_registry.py` or `__init__.py` — that's the
+point of the design.
+
+## Open Questions for Feedback
+
+1. **Naming:** "Module Override Registry", "Kernel Registry", "Implementation Registry"?
+   Current proposal uses "override" throughout.
+
+2. **Checkpoint compatibility:** Override authors are responsible for state_dict key
+   compatibility. Should the registry enforce or validate this?
```
syed-ahmed: Could you please point to the state_dict key this referring to? Is this mainly to make sure e2e workflows like loading model into vllm after training in torchtitan has a good UX when using custom modules from registry?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+The `@register` decorator stores metadata + factory in a global dict. The factory
+function takes the original Config and returns a replacement Config.
+
+### Activation
+
+The user specifies override modules via config (CLI or config file):
+
+```bash
+torchtitan_train --module llama3 --config llama3_8b \
+    --override.modules='["torchtitan.registry.triton_rope"]'
```
SherlockNoMad: For agentic hillclimb, I think we need to make the system much more pluggable and overridable, so agent can swap out any components. E.g. swap module impl, optimizer, kernels... (and more). 

GraphTrainer currently have 4 open registry, see torchtitan/experiments/graph_trainer/registry.py
User can customize pass_pipeline, memory policy, some hooks to run in the trainer.step()... 

I am also further extending the registry to "fusion region extractor" and "kernel generator" in this prototype
https://github.com/pytorch/torchtitan/compare/kernel-fusion-workflow?expand=1

Agent can bring its own "extractor", and try various kernel provider (e.g. KernelAgent, torch.compile, Helion codegen...) to optimize kernel.

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+
+2. **Checkpoint compatibility:** Override authors are responsible for state_dict key
+   compatibility. Should the registry enforce or validate this?
+
+3. **Scope of traverse:** Currently limited to `model_config` (the model's Config tree).
+   Should it extend to the full `Trainer.Config` tree (optimizer, dataloader, etc.)?
+
+4. **Filter functions:** v1 replaces ALL instances of a target Config class. Should we
+   support per-instance filtering (e.g., "only replace RoPE in layers 0-10") in the
+   `@register` API, or is that a v2 feature?
```
syed-ahmed: Supporting per-instance filtering is required for NVFP4 training (check layer sensitivity section here: https://arxiv.org/pdf/2509.25149). Also seems like MXFP8 training also cares about this:

https://github.com/pytorch/torchtitan/blob/52a292d2977690d407bd81781de932f7f7dc56c5/torchtitan/components/quantization/mxfp8.md?plain=1#L200

syed-ahmed (commented):
Thank you @tianyu-l ! Love this! Let's ship it. Soliciting more comments from colleagues.

torchtitan/registry/DESIGN.md
tianyu-l: > I think we need to make the system much more pluggable and overridable, so agent can swap out any components. E.g. swap module impl, optimizer, kernels... (and more).

Totally, we can make it support any `Configurable` swapping in the first version. I think that'd be enough, because you can swap anything in the config tree, all the way up to trainer.

torchtitan/registry/DESIGN.md
tianyu-l: Sure, I can add that in the first version. What would be good enough, e.g. fqn / regex / etc.?

some existing examples:
https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/quantization/mx.py#L54
https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/lora.py#L157

torchtitan/registry/DESIGN.md
tianyu-l: This is talking about the checkpoint compatibility before vs. after override.

It may or may not be necessary depending on what we want to do:
- In pretraining, if we just want to swap to efficient implementation and train from scratch, it doesn't matter if the checkpoint from the original (slow) impl can be loaded to new (efficient) impl at all.
- For continued training / sft when loading from a HF checkpoint, torchtitan only provides state dict adapter from HF checkpoint fqn to torchtitan native fqn. If the new impl doesn't work torchtitan native fqn, then it needs to override the state dict adapter in the config as well. It is not obvious why the new impl wouldn't follow the native fqn.
- For RL between trainer and generator, existing effort is using the same model definition, so checkpoint would be naturally compatible. In the future, I can imagine that people do different optimizations on trainer model and generator model. In that case, it should be desired that the state dict is still compatible, to avoid any fqn remapping, which itself is error-prone (we will still need to do resharding).

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+
+```python
+# torchtitan/registry/triton_rope.py (in-repo example)
+from torchtitan.registry import register
+from torchtitan.models.common.rope import RoPE
+from torchtitan.protocols.module import Module
+
+class TritonRoPE(Module):
+    """RoPE implementation using Triton kernels."""
+
```
xmfan: I'm interested in adding a Fused SwiGLU module, ideally we may want to be able to also customize the  `__init__` parametrization so that a kernel can access w1 and w3 as a single contiguous buffer:

```
Unfused (current): 
x1 = x @ w1.T      # w1: [H, D]
x3 = x @ w3.T      # w3: [H, D]
out = (F.silu(x1) * x3) @ w2.T  # w2: [D, H]

Fused:
y13 = x @ w13.T    # w13: [2H, D]
x1, x3 = y13.chunk(2, dim=-1)
out = (F.silu(x1) * x3) @ w2.T
```

I think `GroupedExperts.__init__` gets called before loading weights though, so the module needs to keep the same FQNs, unless we expose some post_init hook that can further customize parametrization? I'm not sure how it would interact with parallelization though.

There might be still a way to implement w1 and w3 as chunks of a larger contiguous allocation, and have the kernel read from them as if they were 1.

Any concerns with allowing modules to do this?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+### External Packages
+
+A vendor ships a pip package:
+
+```python
+# In package "torchtitan-vendor-x", file: vendor_x/overrides.py
+from torchtitan.registry import register
+from torchtitan.models.common.moe import MoE
+
+class VendorXFusedMoE(Module):
```
shuhuayu: In this design, can we in addition fuse communication? like, can i have a kernel that does allreduce + rmsnorm?

torchtitan/registry/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    description: str
+
+
+_REGISTRY: dict[str, ModuleOverride] = {}
+
+
+def register(
+    name: str,
+    *,
+    target: type,
```
crcrpar: better if it's `Type[Configurable.Config]` or `type[Configurable.Config]`?

torchtitan/registry/DESIGN.md
tianyu-l: Of course we can, as long as the `ShardingConfig` at the module boundary is set correct.

torchtitan/registry/DESIGN.md
tianyu-l: This is interesting case!

> I think GroupedExperts.__init__ gets called before loading weights though, so the module needs to keep the same FQNs, unless we expose some post_init hook that can further customize parametrization?

Sounds the right approach to me, if we want to load a checkpoint saved from unfused (and not introduce any more complicated state dict adapter in tree).

> I'm not sure how it would interact with parallelization though.

~~For TP, if `w13` is of shape [2*D, H], making it colwise sharded, the placement would still be Shard(1), which is expressible by spmd_types.~~

I was wrong: the colwise sharding should happen on D, not H, thus couldn't be expressed today.

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+**Plan:** For v1, pass the RoPE Module (not just its cache tensor) through the forward
+chain. `GQAttention.forward()` calls `rope(xq, xk, positions)` instead of
+`apply_fn(xq, xk, cache, positions)`. In a follow-up, each attention layer will own
+its own RoPE module as a submodule, making it naturally replaceable via the config tree.
+
+## Parallelism Compatibility
+
+Replacement modules must satisfy the `Module` protocol (`ShardingConfig`, `init_states`,
+etc.) for TP/FSDP to work. Model-specific `parallelize_fn` functions may reference
+specific module types.
```
acisseJZhong: what does reference specific module types mean? right now each model's parallelize_fn have access to only model?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+
+**v1:** Document that overrides must be drop-in compatible (same interface, same
+submodule structure).
+
+**Long-term:** Push parallelism fully into the Module protocol so `parallelize_fn`
+doesn't need to know module types. This aligns with torchtitan's ongoing direction
```
acisseJZhong: what will this look like?

torchtitan/registry/DESIGN.md
acisseJZhong: how could we tell if users' override will cause ckpt compatability issues? For customized kernel it probably won't change model weights, but it will for fused module?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+| Untyped kwargs dict | No | Yes | ~50 lines | CLI-friendly but error-prone, awkward nested dict syntax. |
+| Typed override Config (like ModelConfigConverter) | Yes | Yes | ~100 lines | Couples config_registry to override package, defeating the "no-touch" goal. |
+
+The custom module approach requires zero additional infrastructure and handles all
+use cases. Most overrides have few tuning knobs with sensible defaults.
+
+## Interaction with Existing Converters
+
+Existing converters (Float8, LoRA) run inside `model_registry()` during config
+construction. Overrides run later in `Trainer.__init__`. The override sees the
```
acisseJZhong: so the order is first model config convertor, and then overrides. wondering is this better then the other way?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+torchtitan_train --override.modules='["my_overrides"]' ...
+```
+
+**Why this over other approaches:**
+
+| Approach | Typed? | CLI? | Extra infra? | Tradeoff |
+|----------|--------|------|--------------|----------|
+| Custom module file (recommended) | Yes | No (edit file) | None | Simple, fully typed, no infra. Changing params requires file edit. |
+| Untyped kwargs dict | No | Yes | ~50 lines | CLI-friendly but error-prone, awkward nested dict syntax. |
+| Typed override Config (like ModelConfigConverter) | Yes | Yes | ~100 lines | Couples config_registry to override package, defeating the "no-touch" goal. |
```
acisseJZhong: seems model config convertor can be covered by `Custom module file (recommended)`?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+
+1. **Hardware-specific kernels.** Efficient Triton kernels for rotary embeddings, fused
+   attention, or custom norms that only work on specific hardware. These are valuable for
+   performance but (a) hardware-specific, (b) harder to maintain, (c) held to a lower
+   code quality bar than core. We don't want to gate landing these on the same review bar
+   as core modules, and we don't want to say "no" to vendors when we can instead provide
+   a clean extension mechanism.
+
+2. **Fused module implementations.** An efficient MoE implementation may fuse the
+   dispatcher-compute-combine pipeline with custom routing, communication, and
```
acisseJZhong: can't wait to see megamoe in titan!!

torchtitan/registry/DESIGN.md
tianyu-l: Our direction is to remove model-specific parallelize_fn and push parallelism into config. So I wouldn't read hard into this (AI-generated) sentence.

torchtitan/registry/DESIGN.md
tianyu-l: Like your MoE refactor, TP/CP/EP works with config-based module boundary redistribute.

For FSDP / FlexShard, we will put bucketing and storage placement into config as well.

torchtitan/registry/DESIGN.md
tianyu-l: It depends on user purpose when swapping modules / configs. It is possible that user may not care about ckpt compatibility at all (e.g. train from scratch with efficient impls, or experimenting with a complete new module).

When they care:
- obvious errors should be caught by DCP or torchstore
- authors who implement the fused module should be responsible for ckpt compatibility, e.g. guard by unit test. See related discussion https://github.com/pytorch/torchtitan/pull/3396/changes/BASE..f3bfee46b599269483e972f8320c530876684f03#r3270109556

torchtitan/registry/DESIGN.md
tianyu-l: I think so.

In-repo config converter may not understand overrides at all. Overrides author knows everything, so if they believe a in-repo converter could work with their override, they could
- swap inner module of a converted module, e.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/components/lora.py#L115
- swap the entire converted module, and then apply in-repo converter again inside the override

torchtitan/registry/DESIGN.md
acisseJZhong: Yeah that makes more sense. Maybe ask ai to polish this as community users might not get this if we commit this md?

torchtitan/registry/DESIGN.md
tianyu-l: That's correct. The override is actually very powerful, but it doesn't mean we should use it whenever we can; in fact, I would suggest we use it as little as possible for in-repo development.

Dynamic import and dependency injection would make code less straightforward in general. Even the `ModelConfigConverter` makes it somewhat worse than building everything you need bottom-up from scratch (remember we had this quantization "as converter" or "from direct build" discussion). Override registration is taking a step further, but comes with its own tradeoff.

torchtitan/registry/DESIGN.md
tianyu-l: Yeah I'll polish everything; this is more like an RFC with real code.

torchtitan/registry/DESIGN.md
syed-ahmed: I think fqns should be good enough for now.

torchtitan/registry/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+            factory=fn,
+            description=description,
+        )
+        return fn
+
+    return decorator
+
+
+def _check_conflicts() -> None:
+    """Raise if two overrides target the same Config class."""
```
wconstab: haven't reviewed the rest yet, but do you deal with the situation where one override targets Parent and another override targets Child?   I guess there are a few scenarios
1) new parent still has a version of child: we could expect the child in the new parent to also be overridden.  but wed have to make sure to do the parent override first and then then the child override
2) new parent doesn't even have the child anymore (it has a different structure). The child override now needs to be discarded, which I think is ok as a silent no-op?

tianyu-l: It seems this is NOT the same question as https://github.com/pytorch/torchtitan/pull/3396#discussion_r3269092726, and you are talking about A owns B, and our target is A (and B).

If so, my answer would be:
it's all up to the new implementation of A
- if new A still has child B, it can extract old A's child B and plug in new A; or it can create its new B (if B only appears in A, one doesn't need to override B)
- if new A doesn't have B any more, that's totally fine, and the Bs which belong to old A just disappear

torchtitan/registry/DESIGN.md
sdmyzlp: +1 for full `Trainer.Config` scope override — this adds flexibility given torchtitan's clean config-driven build system. Some examples where it helps:

- **Profiler.Config**: Vendors beyond CUDA/XPU currently monkey-patch `build_torch_profiler` to enable custom `ProfilerActivity` and `experimental_config`, an override would avoid that.
- **OptimizerContainer.Config**: With vendor-specific interfaces like [torch_npu.empty_with_swapped_memory](https://www.hiascend.com/document/detail/zh/Pytorch/2600/apiref/torchnpuCustomsapi/docs/zh/custom_APIs/torch_npu/torch_npu-empty_with_swapped_memory.md), optimizer tensors can be allocated directly on accelerator-accessible DRAM. Being able to override `OptimizerContainer.Config` would make this kind of integration seamless.

sdmyzlp: +1 for full `Trainer.Config` scope override — this adds flexibility given torchtitan's clean config-driven build system. Some examples where it helps:

- **Profiler.Config**: Vendors beyond CUDA/XPU currently monkey-patch `build_torch_profiler` to enable custom `ProfilerActivity` and `experimental_config`, an override would avoid that.
- **OptimizerContainer.Config**: With vendor-specific interfaces like [torch_npu.empty_with_swapped_memory](https://www.hiascend.com/document/detail/zh/Pytorch/2600/apiref/torchnpuCustomsapi/docs/zh/custom_APIs/torch_npu/torch_npu-empty_with_swapped_memory.md), optimizer tensors can be allocated directly on accelerator-accessible DRAM. Being able to override `OptimizerContainer.Config` would make this kind of integration seamless.

tbqh:
I setup the [TransformerEngine NVFP4 integration](https://github.com/tbqh/torchtitan_te_plugin) to work off of this. Seems to work pretty well:

```
@register("te_nvfp4_linear", target=Linear.Config)
def te_nvfp4_linear_override(cfg: Linear.Config) -> TENVFP4Linear.Config | None:
    if cfg.in_features % 16 != 0 or cfg.out_features % 16 != 0:
        return None
    return TENVFP4Linear.Config(
        in_features=cfg.in_features,
        out_features=cfg.out_features,
        bias=cfg.bias,
        param_init=cfg.param_init,
    )
```

torchtitan/registry/DESIGN.md
tbqh: I'll update the [NVFP4 integration](https://github.com/tbqh/torchtitan_te_plugin) to use this once it's ready. Right now it's just replacing every linear layer.

torchtitan/registry/DESIGN.md
sdmyzlp: > If the new impl doesn't work torchtitan native fqn, then it needs to override the state dict adapter in the config as well.

Could we address this by allowing an optional `state_dict_translator` argument to `@register`? It would be propagated to the existing `from_hf` / `to_hf` functions, eventually enabling things like:
```python
@register("fused_ffn", target=FeedForward.Config,
          state_dict_translator=ffn_translator,
          description="Fused FFN")
def fused_ffn(cfg):
    return FusedFFN.Config(
        dim=cfg.w1.in_features,
        hidden_dim=cfg.w1.out_features,
        param_init={"fused_gate_up.weight": ..., "fused_down.weight": ...},
        ...
    )
```
This way, override authors can freely use their own internal FQN while still loading/saving checkpoints in a compatible format. Does this approach make sense to you?

tianyu-l: How is that different from the checkpointing hook idea in https://github.com/pytorch/torchtitan/pull/3396#discussion_r3270109556?

I'm not seeing the full proposal on `state_dict_translator`, but it seems requiring inventing one more mechanism in the current `Module` class? If it can be directly covered by subclassing and overriding `state_dict` / `load_state_dict`, I tend to not introduce more mechanisms.

sdmyzlp: I think there is no difference, both seek to bridge the canonical and override internal fqn, but if `state_dict` itself can eventually be overriden (seems it doesn't have a `Config` for the time being?), that would be the simplest approach.

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+**Long-term:** Push parallelism fully into the Module protocol so `parallelize_fn`
+doesn't need to know module types. This aligns with torchtitan's ongoing direction
+and will naturally make overrides fully composable with parallelism.
+
+## Scope of Changes
+
+| File | Change |
+|------|--------|
+| `torchtitan/registry/__init__.py` | **New.** Registry mechanism (~100 lines) |
+| `torchtitan/config/configs.py` | Add `OverrideConfig` dataclass (~10 lines) |
```
IvanYashchuk: This scope table seems stale relative to the patch.

`OverrideConfig` is implemented in `torchtitan.registry`, while the table says `torchtitan/config/configs.py`. The table also lists RoPE/attention refactors and `torchtitan/registry/triton_rope.py`, which are not in this PR. Can we split this into current patch vs follow-up work so readers know what is being proposed now?

torchtitan/registry/DESIGN.md
```diff
@@ -0,0 +1,273 @@
...
+### Core Principles
+
+- **Explicit opt-in.** No auto-detection. The user specifies which override modules to
+  import via `--override.modules`. If you don't ask for it, nothing changes.
+- **Error on conflict.** If two imported modules register overrides for the same Config
+  class, fail loudly. No priority system — the user resolves conflicts by importing
+  fewer modules.
+- **Module-level scope.** Overrides replace `Configurable.Config` nodes in the config
+  tree. For op-level replacement (e.g., a single function), either wrap it in a Module
+  first, or use PyTorch's `TORCH_LIBRARY` mechanism.
```
IvanYashchuk: I agree with keeping TorchTitan's v1 registry at module/config level.

For custom CUDA/Triton kernels inside an override module, can we point authors to the PyTorch custom op paths (`torch.library.custom_op` or `torch.library.triton_op`, fake/meta kernels, autograd registration, and `opcheck`)? That makes the `torch.compile` responsibility concrete without adding a TorchTitan operator override API in this PR.

torchtitan/registry/DESIGN.md
IvanYashchuk: If we expect external packages to depend on this API, I think the follow-up question is which target configs are stable extension points.

The existing versioning thread covers pinning, so I would keep this narrower: are `Linear.Config`, `RoPE.Config`, `MoE.Config`, etc. intended as stable hook names, or are they best-effort Python class targets for now?

IvanYashchuk (commented):
I think this is a useful v1 if we keep it scoped to config-tree module overrides.

Most of the registry semantics are already being covered in other threads: global registry scoping, checkpoint compatibility, version pinning, etc.

From the `torch.compile` side, I would keep the first operator-level guidance outside TorchTitan's registry. For lower-level CUDA/Triton kernels, can we point override authors to PyTorch custom op paths (`torch.library.custom_op` / `torch.library.triton_op`, fake/meta kernels, autograd registration, `opcheck`) and ask them to own an eager plus `torch.compile` smoke test? That keeps TorchTitan's v1 contract at the module/config level without creating a second operator override API here.

torchtitan/registry/DESIGN.md
wwwjn: For FusedMLP, if I want to use the new registry system (take used w13, w2), I would need to do following:
1. Set correct ShardingConfig for w13 and w2 then registry
2. Add load_state_dict(), state_dict() function in my registry (We should use hooks for that?)

Is there anything I missed? Or current registry is not able to handle FusedMLP?

torchtitan/registry/DESIGN.md
wwwjn: I want to patch to batch invariant kernels, so seems it might be further override by other kernel overrides. Should we consider batch-invariant mode as a converter, because it's mostly op-level override, right?

zstreet87:
Read this through as a hardware-vendor lens (GOAL 3) and tried it end-to-end with a simulated out-of-repo vendor package. The injection seam (post-config /
  pre-build) is the right call. Two things stood out:

  1. Manual field-copy in factories silently breaks on core changes. The factory hand-copies each field (as in the DESIGN examples). When a target Config
  later gains a field, every vendor factory drops it and gets the default — no error. A small derive(cfg, NewConfig, **deltas) helper that copies shared
  fields structurally (factory specifies only the deltas) would remove this whole class of version-coupling bugs.

  2. Per-instance targeting (open question #4) feels central, not v2. The motivating cases collide by design — two vendors with norm kernels, or A/B-ing a
  kernel per layer. A selector on @register (FQN glob or predicate) plus per-node conflict detection (error only when two overrides claim the same node, not
  the same class) would cover "fused MoE only on these layers" without sniffing fields inside the factory.

torchtitan/registry/DESIGN.md
tianyu-l: We have one for flex https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/config_registry.py#L40

I think in general 
- batch invariance enablement is not only done at Module level, but also has such implicit behavior https://github.com/pytorch/torchtitan/blob/main/torchtitan/distributed/utils.py#L259, https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L140. It's not clear to me if we should make batch invariance application a single config converter
- batch invariance mode right now is a core feature; overrides, on the other hand, is more for unblocking purpose, and I don't want to promise any property in core to be still available on arbitrary overrides. E.g. if user enable quantization in titan config (not a proper example because we are moving away from such design), but swaps in a non-quantized linear through override, it's not titan's responsibility to quantize whatever they plugged in. Similar for batch invariance.

torchtitan/registry/DESIGN.md
tianyu-l: See related discussion https://github.com/pytorch/torchtitan/pull/3396#discussion_r3269378704

In particular, other than publishing stable versions, there won't be any guarantee. So it's closer to
> best-effort Python class targets for now

torchtitan/registry/DESIGN.md
tianyu-l: added a section in the design md

torchtitan/registry/__init__.py
tianyu-l: changing the design a bit -- both scenarios are handled by an explicit error

tianyu-l:
@zstreet87 Thanks for the very thoughtful feedback! I've addressed them in latest commit, please take a look.

torchtitan/registry/DESIGN.md
tianyu-l: Today it is not supported, issue summarized in https://github.com/pytorch/torchtitan/issues/3569

We probably need something as @sdmyzlp mentioned in https://github.com/pytorch/torchtitan/pull/3396#discussion_r3290485359

torchtitan/registry/DESIGN.md
tianyu-l: I seem to be wrong after playing with the idea for a bit.

Created https://github.com/pytorch/torchtitan/issues/3569 to track the issues. Wish to land this version first.

wwwjn (approved):

torchtitan/protocols/model_spec.py
```diff
@@ -41,3 +46,26 @@ class ModelSpec:
...
+        The incoming ``_prefix`` (the path to this ModelSpec, e.g.
+        ``"model_spec"``) is intentionally dropped: model component FQNs use the
+        bare module names (e.g. ``"layers.0.feed_forward.w1"``) that
+        ``model.traverse`` produces and that converter ``filter_fqns`` already
+        match, so they stay stable regardless of nesting.
+        """
+        model = self.model
+        if model is None:
+            return
+        if isinstance(model, config_cls):
```
wwwjn: So this if will only be hit when a user override the whole model?
eg, target=Decoder.Config. Is it valid? Should we directly ban this behavior?

torchtitan/config/OVERRIDE.md
```diff
@@ -0,0 +1,538 @@
...
+
+The `TritonRoPE` snippets used above are illustrative only; a real RoPE override
+depends on the refactor below.
+
+### RoPE example (planned)
+
+Today `RoPE.forward()` only returns a precomputed cache; the application is done
+by free functions inside `GQAttention.forward()`. To make RoPE replaceable as a
+Module, the application must move into the RoPE Module — attention would call
+`rope(xq, xk, positions)`, and a follow-up would give each attention layer its
```
wwwjn: Should already be done, we can add another issue to have a TritonRoPE override

torchtitan/config/override.py
```diff
@@ -0,0 +1,334 @@
...
+    """
+    # (a) exact same node, identified by container + key.
+    by_node: dict[tuple[int, str | int], str] = {}
+    for c in claims:
+        key = (id(c.parent), c.attr)
+        owner = by_node.get(key)
+        if owner is not None and owner != c.ov.name:
+            raise ValueError(
+                f"Overrides '{owner}' and '{c.ov.name}' both claim node "
+                f"'{c.fqn}'. Narrow their `where` selectors so they target "
```
fegin: What does the `where` mean?

torchtitan/protocols/model_spec.py
```diff
@@ -41,3 +46,26 @@ class ModelSpec:
...
+            yield "model", model, self, "model"
```
fegin: It seems that we will be able to repalce the entire model using override. Not quite sure if my understanding is correct.

torchtitan/config/override.py
```diff
@@ -0,0 +1,334 @@
...
+            setattr(c.parent, c.attr, new_cfg)
+        replacements.append(
+            f"[Override] {c.ov.name}: {c.fqn} "
+            f"{type(c.cfg).__qualname__} -> {type(new_cfg).__qualname__}"
+        )
+
+    if replacements:
+        for line in replacements:
+            logger.info(line)
+        logger.info(f"Applied {len(replacements)} override(s)")
```
fegin, line 337: Should we have a warning if there is no replacements but the user claims override?

torchtitan/overrides/fused_swiglu.py
```diff
@@ -0,0 +1,143 @@
...
+        "FeedForward checkpoints ('w1.weight'/'w3.weight'). A checkpoint saved "
+        "with this override only loads into a run that also uses it. See "
+        "torchtitan/config/OVERRIDE.md 'Checkpoint Compatibility'.",
+    )
+
+    # The fused weight reuses the gate projection's initializer; keyed by the
+    # fused parameter name "w13" for the Module init protocol.
+    param_init = None
+    if cfg.w1.param_init is not None and "weight" in cfg.w1.param_init:
+        param_init = {"w13": cfg.w1.param_init["weight"]}
```
fegin: we actually make `w2w3` share the same init function. You code here is not wrong. But should we consider to avoid making w2 and w3 have the same init function?

fegin (approved):
Left some questions and nits.

drisspg:
@claude Please do a deep review focused on subtle correctness bugs and regressions that are easy for human reviewers to miss. Prioritize findings over summary. I dont care about anything that isnt a real bug drop all the noise but that, if you there isnt anything to report just say so with a small comment

torchtitan/registry/DESIGN.md
xmfan: I corrected the sample, the w13 would concat on H dim, not D dim

xmfan (approved):

torchtitan/protocols/model_spec.py
tianyu-l: I ended up allowing that. E.g. a multimodal model may have Decoder as a child module -- I don't feel we have strong reasons to not allow swapping that.

torchtitan/protocols/model_spec.py
tianyu-l: Yes, I ended up allowing that. See my response to wwwjn above.

torchtitan/overrides/fused_swiglu.py
tianyu-l: I ended up faithfully following what's specified today, i.e. w1 and w2 having different init fns.

PR #3398, opened 2026-05-19
[Module] Replace from_nn_module with native Module subclasses

fegin:
* #3401
* #3399
* __->__ #3398

**Motivation**
`from_nn_module` is designed to be a drop-in replacement of nn.Module. However it causes several issues including not easy to do sharding and may violate the `build()` constraint. So instead of continuing hacking it, this PR replaces the usage in TorchTitan with the explicit class definitions.

**Summary**
Replace `Module.from_nn_module` dynamic class generation with explicit diamond-inheritance classes in nn_modules.py, following the same pattern as Linear, RMSNorm, and Embedding.

**Detail**
- Add Conv2d, GroupNorm, LayerNorm (parameterized, with typed Config fields and _init_self_parameters override delegating to reset_parameters)
- Add GELU, SiLU, Identity (lightweight, Config with minimal or no extra fields)
- Delete from_nn_module classmethod and _created_classes cache from module.py (~80 lines removed)
- Update all flux callsites (layers.py, autoencoder.py) from direct construction to Config-based construction
- Rewrite TestFromNnModule as TestNnModuleWrappers testing native classes


torchtitan/models/common/nn_modules.py
tianyu-l, line 1: let's put all "trivial" common modules here, including linear, rmsnorm, embedding, etc. and only keep non-trivial modules in separate files.

tianyu-l (approved):
one comment

review.html
tianyu-l, line 1: what's this, lol

xmfan: lol maybe we can also have some gitignored agent scratch space folder

xmfan (approved):

review.html
fegin: wow, the agent went crazy to add too many things that I locally own. That's interesting.

PR #3410, opened 2026-05-20
[NOT READY][Varlen CP] Enable Varlen attention under Context Parallel + Full DTensor

fegin:
* #3411
* __->__ #3410

Add varlen-attention support under CP with full_dtensor mode. Q stays sharded on the CP axis while K and V use DTensor Replicate placement so DTensor all-gathers them automatically.

Key design:
- VarlenMetadata is a frozen dataclass in torchtitan/models/common/attention.py with fields cu_seq_q, cu_seq_k, max_q, max_k, plus an optional k_local_indices.
- CPVarlenMetadata in torchtitan/distributed/varlen_cp.py extends VarlenMetadata with k_local_indices always populated and a from_global builder that produces per-rank metadata. For each (doc, rank) contiguous-Q chunk, seqlen_k = (global doc-relative offset at chunk end) + 1.
- cp_shard calls CPVarlenMetadata.from_global directly for varlen masks. Q/K/V tensors and BlockMask still use the upstream _context_parallel_shard dispatcher.
- VarlenAttention.forward re-packs K/V via k_local_indices when present, reading the field directly on VarlenMetadata so the attention module never imports CP infrastructure.

Varlen+CP only works under full_dtensor because without DTensor Replicate placement, K and V are plain local tensors and nothing all-gathers them across CP. The legacy apply_cp_to_forward path raises with a message pointing users to parallelism.full_dtensor=True.

Other changes:
- full_dtensor.validate_config no longer rejects flex/varlen.
- llama3.model.update_from_config requires parallelism.full_dtensor for varlen+CP instead of rejecting the combo.
- qwen3.model.update_from_config error message updated for consistency with the new wording.
- VarlenAttention.forward now rejects non-causal window_size when k_local_indices is set, since the CP builder constructs cu_seq_k as a right-aligned causal window.
- Integration tests: full_dtensor_cp_varlen, full_dtensor_cp_fsdp_varlen, full_dtensor_tp_cp_fsdp_varlen.
- CPU unit tests: multi-doc straddling, headtail rearrangement, multi-batch, CP=4, divisibility (with and without load balancer), cross-attention validation, double-shard rejection.

### DIFF
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -55,17 +55,19 @@ def _enable_full_dtensor(t: OverrideDefinitions) -> OverrideDefinitions:
     All features.py tests run under full_dtensor except PP-only variants
     (see ``_is_pp_only``) and CP + compile variants (upstream symint
     limitation); legacy non-full_dtensor coverage lives in models.py.
-    CP variants also switch to FlexAttention config because
-    full_dtensor + CP is not supported with SDPA / Varlen.
+    CP variants default to FlexAttention (full_dtensor + CP under SDPA is
+    not supported), but variants that already pick their own ``--module``
+    or ``--config`` (e.g. the varlen+CP tests) keep that selection.
     """
     new_args = []
     for variant in t.override_args:
         prefix: list[str] = []
         has_cp = any("context_parallel_degree" in arg for arg in variant)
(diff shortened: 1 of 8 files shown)


fegin:
The ghstack cause some issues. Close this PR.

PR #3412, opened 2026-05-21
[loss] Fix ChunkedCELoss + TP gradient placement mismatch

fegin:
* __->__ #3412

**TLDR**
When verifying CP8 v.s TP8, TP posts a significant higher loss than CP. After comparing with single GPU run, it is clear that TP + ChunkedCELoss doesn't work correctly. This PR fixes the issue.

**Summary**
GradAccumulator was wrapping the buffer with the reference activation's placement (Replicate), but the buffer actually held the local values of chunk_grad, whose placements can be different than activations' placement. With TP and a loss-parallel ColwiseParallel lm_head, chunk_grad comes back as Partial(sum) on the TP axis, and to_local() returns each rank's partial slice without an all-reduce. The Replicate label then made downstream autograd treat the partial values as already summed, poisoning the decoder backward and pushing the loss curve well above the single-GPU baseline.

**This PR fixes the issue by capturing _placements from the first added chunk_grad instead of from the reference activation.**

Validated via scripts/loss_compare.py on llama3 debugmodel + global batch size 8 + 500 steps + seed checkpoint + --debug.seed=42 --debug.deterministic.

| Config              | Step 1 loss | Step 500 loss |
|---------------------|-------------|---------------|
| Single GPU          |    8.054873 |      2.342949 |
| TP8 before this fix |    8.054862 |      2.727362 |
| TP8 after this fix  |    8.054862 |      2.353253 |

**We can also infer the placement if doing this in backward causes tracing issues**

### DIFF
--- a/torchtitan/components/loss.py
+++ b/torchtitan/components/loss.py
@@ -186,8 +186,12 @@ class GradAccumulator:
     this uses a pre-allocated buffer with in-place copies for better memory efficiency.
 
     Args:
-        reference: Reference tensor to derive shape, device, and DTensor metadata.
-            If a DTensor, result() returns a DTensor with matching placements.
+        reference: Reference tensor for shape, device, and DTensor-ness. If a
+            DTensor, only its device mesh is reused; the placement of the
+            returned DTensor is taken from the first added chunk (see add()),
+            not from this reference, so the buffer is labeled with the actual
+            gradient placement (e.g. Partial(sum) for a loss-parallel
+            ColwiseParallel lm_head) rather than the activation placement.
         num_chunks: Number of chunks that will be added.
         seq_dim: The sequence dimension along which chunks are accumulated.
         dtype: Dtype for the buffer.
@@ -214,12 +218,11 @@ class GradAccumulator:
         self.seq_dim = seq_dim
         self._next_idx = 0
         self._device_mesh: DeviceMesh | None = None
+        # Captured from the first added chunk; see __init__ docstring.
         self._placements: tuple[Placement, ...] | None = None
 
-        # Track DTensor metadata for transparent wrap-back in result()
         if isinstance(reference, DTensor):
             self._device_mesh = reference.device_mesh
-            self._placements = reference.placements
             local = reference.to_local()
         else:
             local = reference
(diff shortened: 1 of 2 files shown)


ezyang:
@pianpwk I'm pretty sure that spmd types would have caught this

acisseJZhong:
> GradAccumulator was wrapping the buffer with the reference activation's placement (Replicate), but the buffer actually held the local values of chunk_grad.

maybe I didn't have much context, but wondering why we don't wrap the buffer with Partial if this is the case?

fegin:
@acisseJZhong 

> maybe I didn't have much context, but wondering why we don't wrap the buffer with Partial if this is the case?

That's what this PR does. The statement you refer to just states the status quo.

torchtitan/components/loss.py
```diff
@@ -186,8 +186,12 @@ class GradAccumulator:
     this uses a pre-allocated buffer with in-place copies for better memory efficiency.
 
     Args:
-        reference: Reference tensor to derive shape, device, and DTensor metadata.
-            If a DTensor, result() returns a DTensor with matching placements.
+        reference: Reference tensor for shape, device, and DTensor-ness. If a
+            DTensor, only its device mesh is reused; the placement of the
+            returned DTensor is taken from the first added chunk (see add()),
+            not from this reference, so the buffer is labeled with the actual
+            gradient placement (e.g. Partial(sum) for a loss-parallel
```
tianyu-l: I think it has nothing to do with loss parallel.
It happens as long as input_dst is Replicate, weight is Shard, per
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder_sharding.py#L250-L255

torchtitan/components/loss.py
tianyu-l, line 430: This PR implicitly assumes that
the forward placement of `h_chunk` is the same as forward placement of `hidden_states`

This is true, e.g. on TP they are both `Replicate`, but it adds one layer of indirection. It is still possible to write buggy code.

tianyu-l (approved):
Thanks for fixing! Agree that spmd_types could help, as spmd.R means backward should be spmd.P.

Some nit comments.

tianyu-l (approved):
Thanks for fixing! Agree that spmd_types could help, as spmd.R means backward should be spmd.P.

Some nit comments.

torchtitan/components/loss.py
fegin: ye, you are right, it is just we enable loss parallel and Claude thinks this is the root cause. Sharded weight will cause this problem.

wwwjn:
Thanks for the fix! The fix looks good to me, and the gradient's placements were previously wrongly labeled as the activation's placement

PR #3414, opened 2026-05-21
Fix imports for latest DeepEP

hirschsn:
With DeepEP HEAD, the imports from `deep_ep.utils` fail. According to the current DeepEP Readme (https://github.com/deepseek-ai/deepep#example-use-in-model-training-or-inference-prefilling), all three classes can be imported from the main `deep_ep` package. This PR fixes the imports accordingly.

### DIFF
--- a/torchtitan/distributed/deepep/deepep.py
+++ b/torchtitan/distributed/deepep/deepep.py
@@ -18,9 +18,7 @@ from torch.distributed import ProcessGroup
 from torch.utils._python_dispatch import _disable_current_modes
 
 try:
-    from deep_ep import Buffer
-
-    from deep_ep.utils import EventHandle, EventOverlap
+    from deep_ep import Buffer, EventHandle, EventOverlap
 except ImportError as e:
     raise ImportError(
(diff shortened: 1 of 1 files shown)


torchtitan/distributed/deepep/deepep.py
```diff
@@ -18,9 +18,7 @@
 from torch.utils._python_dispatch import _disable_current_modes
 
 try:
-    from deep_ep import Buffer
-
-    from deep_ep.utils import EventHandle, EventOverlap
+    from deep_ep import Buffer, EventHandle, EventOverlap
```
tianyu-l, line 21: @acisseJZhong did you not hit this issue with latest DeepEP?

acisseJZhong: Yeah, I had a similar issue and unblocked by changing the DeepEP init https://fburl.com/1sg5q99m. This should be a better fix.

acisseJZhong (approved):
Thanks for catching!

tianyu-l (approved):

PR #3415, opened 2026-05-21
[RFC] Replace HF Datasets / StatefulDataLoader with a Grain-based Pipeline

slimfrkha:
This RFC proposes replacing torchtitan's current data stack (`hf_datasets/`, `components/dataloader.py`) with a Grain-based pipeline where every processing step is a typed, composable `Op` in a dataclass config. The motivation is three structural problems with the current stack: processing behavior is not configurable (text field, packing algorithm, BOS/EOS all hardcoded), the text and chat paths duplicate the same stateful iteration skeleton independently, and resume logic is hand-rolled complexity on top of HF's iterator state. The proposed design expresses the full pipeline — per-source and post-mix — as ordered `Op` lists validated at config-parse time, with mixing, DP sharding, the Map→Iter crossover, and checkpointing handled automatically by the framework.

### DIFF
--- a/docs/rfcs/grain.md
+++ b/docs/rfcs/grain.md
@@ -1,0 +1,899 @@
+# RFC: Replace HF Datasets / StatefulDataLoader with a Grain-based Pipeline
+
+## Summary
+
+torchtitan's data configs cover source selection well — dataset name, path,
+per-source weights, seed — but processing behavior is not configurable at all:
+the text field, tokenization options, packing algorithm, and any filtering are
+hardcoded inside the dataset classes. Changing any of them means editing Python. This RFC replaces torchtitan's current data stack with a Grain-based pipeline where both source selection
+and processing behavior are expressed as composable `Op`s in a dataclass
+config. The old stack is deleted entirely.
+
+## Motivation
+
+Four concrete problems drive this change.
+
+**Processing behavior is not configurable.** `HuggingFaceTextDataLoader.Config`
+and `InterleavedHuggingFaceTextDataLoader.Config` expose source selection
+(dataset name, path, weights, seed) but nothing about how records are
+processed. The text field is hardcoded to `sample["text"]`, `add_bos` and
+`add_eos` are always `True`, the packing algorithm is fixed per class, and
+there is no concept of filtering. `ChatDataLoader.Config.sample_processor` is
+`tyro.conf.Suppress` — it cannot be set from CLI or a config file at all.
+Adding any processing capability (a length filter, a different tokenization
+strategy, a new packing algorithm) means editing the dataset classes or adding
+a new one.
+
+**Text and chat paths duplicate the same skeleton independently.** Both
+`HuggingFaceTextDataset.__iter__` and `ChatDataset._iter_greedy_packed` share
+the same structure: iterate over data, tokenize, accumulate into a buffer, slice
+when full, handle infinite re-loop with per-epoch shuffle. Each maintains its
+own state dict with different keys, its own backward-compat fallbacks (e.g. the
+`positions_buffer` warning in `HuggingFaceTextDataset.load_state_dict`), and its
+own bugs.
+
+**Resume logic is genuinely complex.** `HuggingFaceTextDataset` and
+`ChatDataset` each carry separate code paths for map-style vs. iterable
+datasets, with manual `set_epoch` synchronization on restore. `ChatDataset`
+adds `_pending_input_ids`/`_pending_label_ids` to survive the edge case where a
+sample straddles a packing boundary at checkpoint time. This complexity is
+inherent to hand-rolling stateful iteration on top of HuggingFace's own
+iterator state.
+
+## The Core Idea
+
+Grain's API is already a good data pipeline description language. Here is a
+complete multi-source pretrain pipeline in raw Grain:
+
+```python
+def make_corpus(file_glob, seed):
+    src = grain.sources.ArrayRecordDataSource(sorted(glob.glob(file_glob)))
+    ds  = grain.MapDataset.source(src).seed(seed).shuffle().repeat()
+    ds  = ds.map(parse_and_tokenize)
+    return ds
+
+web   = make_corpus("/prep_ar/web/*.array_record",   seed=11)
+code  = make_corpus("/prep_ar/code/*.array_record",  seed=22)
+books = make_corpus("/prep_ar/books/*.array_record", seed=33)
+
+mixed = grain.MapDataset.mix([web, code, books], weights=[0.70, 0.20, 0.10])
+mixed = mixed[dp_rank :: dp_world_size]
+it    = mixed.to_iter_dataset(grain.ReadOptions(num_threads=8))
+it    = FirstFitPackIterDataset(it, length_struct={"input_ids": seq_len}, num_packing_bins=64)
+it    = it.batch(local_batch_size, drop_remainder=True)
+it    = it.mp_prefetch(grain.MultiprocessingOptions(num_workers=16))
+```
+
+This code already tells the whole story. The problem is that it is not
+configurable: `parse_and_tokenize`, the packing bin count, the weights, the
+seed — all hardcoded. Adding a source or swapping the packing algorithm means
+editing this function. Sharing it across training recipes means copy-paste.
+
+The proposed design makes each step a named, typed `Op` in a dataclass config.
+`make_corpus` becomes `Source(pipeline=[ArrayRecordReader(...), Shuffle(),
+Repeat(), TextTokenize()])`. The post-mix chain becomes
+`pipeline=[FirstFitPack(num_packing_bins=64), Batch(), Prefetch(num_workers=16)]`.
+The mix, DP slice, and Map→Iter crossover are injected automatically by the
+orchestrator — they are framework concerns, not user concerns. Every parameter
+that was hardcoded is now a config field, overridable from CLI.
+
+## Proposed Design
+
+The design has three interlocking concepts. An **`Op`** is the unit of
+composition — a dataclass that wraps one Grain operation and declares two
+independent contracts: a **Stage** contract (`accepts`/`produces`) tracking
+the shape and readiness of the records flowing through it, and a **Phase**
+contract tracking the Grain backing it requires (`MapDataset` vs
+`IterDataset`). The two axes are independent by design: Stage catches
+record-shape errors like tokenize-after-pack at config-parse time; Phase drives
+the Map→Iter crossover automatically at build time. A **Pipeline** is an
+ordered list of Ops with those contracts validated end-to-end: each source
+runs its own pipeline up to the mixing boundary, the framework mixes the
+sources, and a shared post-mix pipeline carries the mixed stream through to a
+training batch.
+
+### The stage ladder
+
+Every record flowing through the pipeline is in one of five stages:
+
+```
+SOURCE  →  RAW  →  ENCODED  →  TRAINSAMPLE  →  TRAINBATCH
+```
+
+This tracks *granularity and readiness*, not modality:
+
+| Stage | Meaning |
+|---|---|
+| `SOURCE` | Empty sentinel before any data exists. Only a reader op accepts it. |
+| `RAW` | One undecoded record as storage emits it — a dict of bytes or strings. |
+| `ENCODED` | Processed model-input features for one source item. Variable shape, any modality. think tokens for text |
+| `TRAINSAMPLE` | One training-ready item conformed to the model's per-step shape (`seq_len` for text tokens). |
+| `TRAINBATCH` | Stacked tensors in the `(input_dict, labels)` shape the trainer consumes. |
+
+The pipeline validates this sequence at `__post_init__`. A missing reader, a
+reader in the wrong position, or tokenize-after-pack all surface as
+`ValueError`s before any data is touched.
+
+### The `Op` abstraction
+
+`Op` is the single extension point. Every op declares two independent contracts
+as class-level metadata:
+
+```python
+class Op(Configurable.Config):
+    accepts:  ClassVar[frozenset[Stage] | Literal["any"]]      # record shape accepted
+    produces: ClassVar[Stage | Literal["same"]]                 # record shape emitted
+    phase:    ClassVar[Literal["map", "iter", "any"]] = "any"  # grain backing required
+
+    def apply(self, ds, ctx: PipelineContext): ...
+```
+
+`accepts`/`produces` is the **stage contract** — it catches record-shape errors
+like tokenize-after-pack. `phase` is the **grain backing contract** — it tells
+the orchestrator whether an op requires a `MapDataset`, an `IterDataset`, or
+works on either. The two axes are genuinely independent: `Prefetch` is
+stage-agnostic (`"any"`/`"same"`) but `"iter"`-only, because `mp_prefetch`
+does not exist on a `MapDataset`.
+
+A concrete op is typically a few lines:
+
+```python
+@dataclass
+class Repeat(Op):
+    accepts, produces = "any", "same"
+    num_epochs: int | None = None   # None → infinite
+
+    def apply(self, ds, ctx):
+        return ds.repeat(self.num_epochs)
+```
+
+`Op.__init_subclass__` auto-registers every subclass by class name. The
+registry is what will allow a YAML loader (see [What This Unlocks](#what-this-unlocks))
+to construct an arbitrary pipeline from a file with no additional wiring.
+
+### Pipeline config shape
+
+Two dataclasses describe a complete data run:
+
+```python
+@dataclass
+class Source:
+    pipeline: list   # [Reader, ...ops...]; SOURCE → ENCODED
+    weight: float = 1.0
+
+@dataclass
+class DataPipeline.Config:
+    sources: list    # list[Source]
+    pipeline: list   # list[Op]; ENCODED → TRAINBATCH
+    seed: int = 42
+```
+
+A `Source` is a reader-headed op list plus its mixing weight. The post-mix
+`pipeline` runs over the combined stream after all sources are mixed. Both are
+validated at `__post_init__`: per-source pipelines must run `SOURCE → ENCODED`;
+the post-mix pipeline must run `ENCODED → TRAINBATCH`.
+
+The fields are typed as bare `list` — not `list[Op]` — matching torchtitan's
+existing tyro pattern for heterogeneous lists. Pipeline structure is built
+before tyro runs (in `config_registry.py` or from a future YAML file), and
+tyro layers leaf overrides on top. CLI overrides stay clean:
+
+```
+--sources.0.weight=0.65 --pipeline.0.num_packing_bins=128
+```
+
+### Framework-internal concerns
+
+These are handled inside `DataPipeline.Config.build(ctx)` and never appear in
+user configs.
+
+**Source tagging.** Each record gets `{"source_idx": i}` injected immediately
+after its reader, so per-source observability is available throughout the
+entire pipeline.
+
+**DP sharding.** After mixing, `mixed[dp_rank :: dp_world_size]` is applied
+before the Map→Iter crossover. Sharding is a free index-slice on the mixed
+`MapDataset` — no extra state, perfectly balanced.
+
+**Map→Iter crossover.** The orchestrator tracks the current grain backing. The
+first time it meets an `"iter"` op while still on a `MapDataset`, it injects
+`.to_iter_dataset(...)` automatically. A `"map"` op encountered after the
+crossover raises. `"any"` ops before the crossover stay on `MapDataset`, which
+is cheaper.
+
+**Mixing.** `MapDataset.mix(sources, weights, seed)` is structural — it is
+many-to-one and does not fit the `list[Op]` model, so it is not an op.
+
+**Checkpointing.** Delegated entirely to Grain's `IterDataset` state dict.
+
+The runtime object is a `DataPipeline` shim (~15 lines) that wraps the
+terminal `IterDataset`, exposes `__iter__` and `Stateful`, and refuses to
+load a checkpoint whose `dp_world_size` differs from the current run. Grain
+does not support resharding, and silently restoring the wrong iterator state
+would be worse than a clean failure.
+
+### Text instantiation
+
+Text is the simplest path through the ladder:
+
+- **`ENCODED`:** `{input_ids, positions}` — one tokenized document, variable
+  length. No labels yet.
+- **`TRAINSAMPLE`:** those fields packed or padded to `seq_len + 1`. After
+  `NextTokenShift`: `{input_ids, positions, label_ids}`.
+- **`TRAINBATCH`:** those fields stacked into `(input_dict, labels)`.
+
+Labels are derived late, by `NextTokenShift` (`label_ids = input_ids[1:]`),
+because plain causal-LM labels are a pure mechanical shift — there is nothing
+to compute at encode time.
+
+The concrete text ops are: `ArrayRecordReader`, `Shuffle`, `Repeat`,
+`TextTokenize`, `MinLengthFilter`, `FirstFitPack`, `GreedyPack`,
+`NextTokenShift`, `Batch`, `Prefetch`.
+
+Canonical pretrain pipeline:
+
+```python
+DataPipeline.Config(
+    sources=[Source(pipeline=[ArrayRecordReader(...), Shuffle(), Repeat(), TextTokenize()])],
+    pipeline=[FirstFitPack(num_packing_bins=64), NextTokenShift(), Batch(), Prefetch(num_workers=4)],
+)
+```
+
+See the [Appendix](#appendix-code-sketch) for multi-source, SFT, and
+validation examples.
+
+### Multimodal instantiation
+
+Vision-text reuses the same five-stage ladder unchanged. Two things differ.
+
+**Labels are produced at encode time, not shifted later.** For vision-text,
+`ENCODED` carries `{input_ids, positions, pixel_values, label_ids}`. Masking
+vision and placeholder tokens requires knowing which tokens they are — masking
+is intrinsic to encoding, not a late mechanical shift. Vision-text therefore
+**skips `NextTokenShift`**.
+
+**Two ops specialize:**
+
+- `VisionTextEncode` (`RAW → ENCODED`): decodes images, inserts
+  vision-placeholder tokens, tokenizes, and masks vision-token labels in one
+  step. Processing parameters (patch size, image normalization, etc.) live on
+  the op, not on `ctx` — context stays lean.
+- `MMPack` (`ENCODED → TRAINSAMPLE`): payload-aware bin-packing that keeps
+  each sample's `pixel_values` attached.
+- `MMBatch` (`TRAINSAMPLE → TRAINBATCH`): a collator, not a stack — it
+  patchifies images, pads ragged image counts, builds `grid_thw`, and caps
+  images per batch.
+
+**Payloads survive shared ops transparently.** `MapOp` re-attaches a reserved
+set of framework keys (`_PASSTHROUGH = {"source_idx"}`) after every
+`transform`, so framework metadata can never be accidentally dropped. For
+payload fields like `pixel_values`, the contract is simpler: they survive
+because no op between producer and consumer rewrites the record. A custom op
+that must carry a payload is responsible for doing so explicitly.
+
+The full `MMBatch.combine` internals are deferred to the multimodal PR.
+
+### Customization
+
+The standard ops cover the common cases, but the design is not a closed box.
+Four family bases make custom ops easy to slot in:
+
+- **`MapOp`** — 1:1 record transform. Implement `transform(sample, ctx) → dict`;
+  the base handles `ds.map(...)` wiring and re-attaches framework metadata.
+- **`FilterOp`** — record predicate. Implement `keep(sample, ctx) → bool`.
+- **`PackOp`** — length-normalize `ENCODED → TRAINSAMPLE`. Implement `apply(ds, ctx)`.
+- **`BatchOp`** — collate `TRAINSAMPLE → TRAINBATCH`. Implement `combine(samples, ctx) → dict`;
+  the base handles batching and the universal `(input_dict, labels)` split.
+
+For cases that don't fit any family — a new modality, bespoke Grain code, a
+stage jump — subclass `Op` directly and write raw Grain in `apply`. The minimal
+custom pipeline is two ops:
+
+```python
+class MyEncode(MapOp):            # RAW → ENCODED: custom processing
+    accepts, produces = frozenset({Stage.RAW}), Stage.ENCODED
+    def transform(self, sample, ctx): ...
+
+class MyPost(Op):                 # ENCODED → TRAINBATCH in one op
+    accepts, produces, phase = frozenset({Stage.ENCODED}), Stage.TRAINBATCH, "iter"
+    def apply(self, ds, ctx): ...
+
+DataPipeline.Config(
+    sources=[Source(pipeline=[ArrayRecordReader(file_glob=...), MyEncode()])],
+    pipeline=[MyPost()],
+)
+```
+
+Mixing, DP sharding, the Map→Iter crossover, and checkpointing are all
+injected by the framework around your ops — you write only the parts that are actually custom.
+
+For the ultimate escape hatch — an exotic topology, a custom sharding strategy, iterable sources that don't fit the Op model at all — subclass `DataPipeline.Config` and redefine `build`. The trainer calls `config.data.build(ctx)` and gets a `DataPipeline` back; it does not care how it was constructed:
+
+```python
+@dataclass
+class MyPipeline(DataPipeline.Config):
+    my_param: int = 32
+
+    def build(self, ctx: PipelineContext) -> DataPipeline:
+        # full raw Grain — any topology, any source format
+        src = grain.sources.ArrayRecordDataSource(...)
+        ds  = grain.MapDataset.source(src).seed(self.seed).shuffle().repeat()
+        ds  = ds.map(lambda s: my_custom_processing(s, ctx, self.my_param))
+        ds  = ds[ctx.dp_rank :: ctx.dp_world_size]
+        it  = ds.to_iter_dataset(grain.ReadOptions(num_threads=8))
+        it  = my_custom_pack(it, ctx.seq_len)
+        it  = it.batch(ctx.local_batch_size, drop_remainder=True)
+        it  = it.mp_prefetch(grain.MultiprocessingOptions(num_workers=8))
+        return DataPipeline(it, dp_world_size=ctx.dp_world_size)
+```
+
+The only contract: `build` must return a `DataPipeline` — the thin shim
+described under *Framework-internal concerns* above, which wraps a terminal `IterDataset`, exposes `__iter__` and `Stateful`, and guards against resharding on restore. The full implementation is in the Appendix. Stage validation, the Op registry, and framework concerns are all bypassed — full power, full responsibility.
+
+## Key Design Decisions
+
+**MapDataset-only sources.** All sources are expressed as `grain.MapDataset`
+— indexable and deterministic. Streaming-only formats go through an offline
+preprocessing step that emits ArrayRecord shards. The payoff: global shuffle,
+exact DP sharding, and exact resume. An opt-in iterable-sources mode for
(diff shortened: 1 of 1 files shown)


tianyu-l:
cc @rakkit @haydn-jones

rakkit:
thx. needs take a time to read.

if we are gonna to bring big refactor. should we consider to bring all "data" & "mix" related function on tables? for example, 

- data-mix scheduler (allow multiple phase training) 
- proper logger & monitor (e.g. i count actual sampled documents, tokens, sampled doc/token ratio, expected doc/token ratio etc)
- evals on multiple docs (composer & olmo's code can easily have val loss on differnt dataset)
- for sft/RL conversation length/round etc.
- proper/better data config system

(i am not saying we should have them at one PR, just have a todo list and we can can make sure the dataset can have proper api for future work)

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+2. `to_dict` emits `op = type(self).__name__` per step, so any config is
+   serializable and round-trippable.
+3. Bare-`list` field typing already accepts externally-constructed entries.
+
+A future loader reads a YAML file, looks up each `op` key in the registry,
+pops it, and instantiates the dataclass — running `__post_init__` stage
+validation on the way. A malformed file-defined pipeline fails the same way a
+code-defined one does. The target shape:
+
+```yaml
```
tianyu-l, line 419: Not ready to take this. It sounds a global change which is worth its own RFC.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+**Full deletion, no wrapping.** A compatibility shim would preserve the config
+duplication problem this design exists to solve. The old classes are deleted;
+behaviors are reimplemented as ops only where they fit the pipeline model.
+
+## What Gets Deleted / What Changes
+
+**Files deleted:**
+
+- `torchtitan/hf_datasets/*`
+- `torchtitan/components/dataloader.py`
```
tianyu-l, line 360: curious where to put the new file(s)?

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+per-source, before the mix, forfeits `MapDataset.mix`: mixing must then happen
+at the `IterDataset` level, losing the global shuffle and the free index-slice
+DP sharding that the base design relies on. Is document-frequency weighting
+acceptable as the default? If so, should the config emit an explicit warning?
+
+**3. Full deletion vs. a deprecation period.**
+
+There is no migration path for existing data configs. Names like
+`HuggingFaceTextDataset`, `ChatDataset`, and `StatefulDataLoader` disappear.
+Is a clean break the right call?
```
tianyu-l, line 504: It's fine to break BC and migrate at a single shot. A less risky plan is to set new dataloader as default, but keep the old one in the repo for a while.

Anyway, let's avoid putting energy in doing shims that compromise the new design.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+described under *Framework-internal concerns* above, which wraps a terminal `IterDataset`, exposes `__iter__` and `Stateful`, and guards against resharding on restore. The full implementation is in the Appendix. Stage validation, the Op registry, and framework concerns are all bypassed — full power, full responsibility.
+
+## Key Design Decisions
+
+**MapDataset-only sources.** All sources are expressed as `grain.MapDataset`
+— indexable and deterministic. Streaming-only formats go through an offline
+preprocessing step that emits ArrayRecord shards. The payoff: global shuffle,
+exact DP sharding, and exact resume. An opt-in iterable-sources mode for
+streaming sources is supported by the phase system, with documented tradeoffs
+(see [What This Unlocks](#what-this-unlocks)).
```
tianyu-l, line 340: Agreed that serious runs should only do map-style, but it'd be good to still support running stream datasets from HF `datasets`, which lowers the bar of trying out torchtitan (without downloading the entire dataset to disk).

docs/rfcs/grain.md
slimfrkha: totally agree. i was just trying to say that the proposed design buys easy integration to YAML as input if wanna go that route and I show what would that be like.

docs/rfcs/grain.md
slimfrkha: So you would include PR 4 in PR 2 from day one ?

docs/rfcs/grain.md
slimfrkha: I would go for:

```
  torchtitan/components/data/
    __init__.py          # public API re-exports
    stage.py             # Stage enum + _validate_stages
    context.py           # PipelineContext
    pipeline.py          # Source, DataPipeline
    ops/
      __init__.py        # for registry population
      base.py            # Op, MapOp/FilterOp/PackOp/BatchOp, Reader
      text.py            # TextTokenize, NextTokenShift, MinLengthFilter, FirstFitPack, GreedyPack
      multimodal.py      # VisionTextEncode, MMPack, MMBatch
```

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+
+The only contract: `build` must return a `DataPipeline` — the thin shim
+described under *Framework-internal concerns* above, which wraps a terminal `IterDataset`, exposes `__iter__` and `Stateful`, and guards against resharding on restore. The full implementation is in the Appendix. Stage validation, the Op registry, and framework concerns are all bypassed — full power, full responsibility.
+
+## Key Design Decisions
+
+**MapDataset-only sources.** All sources are expressed as `grain.MapDataset`
+— indexable and deterministic. Streaming-only formats go through an offline
+preprocessing step that emits ArrayRecord shards. The payoff: global shuffle,
+exact DP sharding, and exact resume. An opt-in iterable-sources mode for
```
rakkit, line 338: would be good if we can have some, even temporal solution, to support checkpoint resume after DP world size changed

slimfrkha: Grain's data-loader checkpoint saves progress per shard (tied to the exact shard_count/shard layout that produced it). Restore into a different DP shard count and there's no automatic re-sharding. you'll either get a validation failure or silently wrong positions. Exact resume is only guaranteed when sharding is identical.

docs/rfcs/grain.md
tianyu-l: That is detail that we can work out with @felipemello1 , I'm more talking about eventual state.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+that must carry a payload is responsible for doing so explicitly.
+
+The full `MMBatch.combine` internals are deferred to the multimodal PR.
+
+### Customization
+
+The standard ops cover the common cases, but the design is not a closed box.
+Four family bases make custom ops easy to slot in:
+
+- **`MapOp`** — 1:1 record transform. Implement `transform(sample, ctx) → dict`;
```
haydn-jones, line 281: From a user perspective: how would per-record random augmentation work here (e.g., for image augmentation, span masking, whatever)? `ctx` doesn't seem to carry a generator / seed. 

Will it be something like `worker_init_fn` (which personally I am still confused about to this day, maybe a self-report...), or something more JAX / grain inspired where you have a `RandomMapOp.transform(sample, rng, ctx)` signature?

slimfrkha: yes. the principle is lean on grain's primitives. Per-record randomness is already a grain feature: alongside .map(fn), grain has .random_map(fn) where fn takes (element, rng) and the rng is a numpy.Generator deterministically seeded from (pipeline seed, element index). That gives randomness that's resume-safe, DP-shard invariant, and worker-count invariant.

In the proposed design it's a sibling base to MapOp:

```python
  class RandomMapOp(Op):
      """Like MapOp but transform gets a per-record rng (numpy.Generator),
      seeded deterministically from pipeline seed + element index."""
      def apply(self, ds, ctx):
          return ds.random_map(lambda s, rng: self.transform(s, rng, ctx))
      def transform(self, sample, rng, ctx): ...

  class SpanMask(RandomMapOp):
      accepts, produces = frozenset({Stage.ENCODED}), "same"
      mask_prob: float = 0.15
      def transform(self, sample, rng, ctx):
          ids = list(sample["input_ids"])
          for i in range(len(ids)):
              if rng.random() < self.mask_prob:
                  ids[i] = ctx.tokenizer.mask_id
          return {**sample, "input_ids": ids}
```
No rng on ctx, no worker_init_fn. same record always gets the same rng across resume, any DP rank, and any worker count.

docs/rfcs/grain.md
tianyu-l: From poolside tech report https://poolside.ai/assets/laguna/laguna-m1-xs2-technical-report.pdf

> Data Loading and Checkpointing. We stream training data rather than reading it from local disk. Streaming is handled by Blender, our internal data service, which mixes multiple sources according to specified mixture weights and curriculum, and exposes a gRPC API for fetching batches with consistent global composition and synchronization across workers. On the training side, a sidecar process prefetches batches from Blender and hands them to the main process through shared memory; the main process then copies them to the GPU via pinned memory. Checkpoints are persisted primarily to S3, with writes distributed across all nodes within a single model replica to make use of the available write bandwidth. Reliable S3 interaction at this scale required substantial engineering effort to handle request throttling and transient errors gracefully on both the read and write paths. On reads, to avoid throttling at restart, only a single replica fetches the checkpoint from S3 and broadcasts the weights to the remaining ranks via RDMA.

slimfrkha: Thanks for the pointer. This setup (gRPC sidecar, shared-memory transport, distributed S3 checkpointing with throttle handling) is pure infra engineering that lives outside the training framework. it's a data service that the trainer consumes, not something the trainer implements. 
What we should make easy is the low-friction case: load_dataset(..., streaming=True) for experimentation without a preprocessing step. It can be close to this code
```python
@dataclass
class IterableReader(Op):
    """Generic iterable source. source_fn is any callable returning an iterable."""
    accepts  = frozenset({Stage.SOURCE})
    produces = Stage.RAW
    phase    = "iter"
    source_fn: Callable[[], Iterable]

    def apply(self, ds, ctx: PipelineContext):
        return grain.IterDataset.source(self.source_fn)

# stream directly from HF - dummy example
DataPipeline.Config(
    sources=[Source(pipeline=[
        IterableReader(source_fn=lambda: load_dataset("allenai/c4", split="train", streaming=True)),
        TextTokenize(),
    ])],
    pipeline=[FirstFitPack(num_packing_bins=64), Batch(), Prefetch(num_workers=4)],
)

# needed changes sketch in DataPipeline.build to cater 
# for iterable case (dp sharding, shuffling, mixing)
def build(self, ctx: PipelineContext) -> DataPipeline:
    # 1. Build per-source pipelines
    sources = []
    for src in self.sources:
        ds = None
        for op in src.pipeline:       # first op is always a Reader
            ds = op.apply(ds, ctx)
        sources.append(ds)

    iterable_mode = isinstance(sources[0], grain.IterDataset)

    if not iterable_mode:
        # 2a. MapDataset path
        mixed = grain.MapDataset.mix(sources, weights=[s.weight for s in self.sources])
        mixed = mixed[ctx.dp_rank :: ctx.dp_world_size]       # free index-slice
        it = mixed.to_iter_dataset(grain.ReadOptions(...))  # Map→Iter crossover
    else:
        # 2b. IterDataset path
        mixed = grain.IterDataset.mix(sources, weights=[s.weight for s in self.sources])
        mixed = mixed.shuffle(buffer_size=10_000)              # WindowShuffle, not global
        it = mixed.filter(lambda _, i: i % ctx.dp_world_size == ctx.dp_rank)  # read amplification

    # 3. Post-mix pipeline
    for op in self.pipeline:
        it = op.apply(it, ctx)

    return DataPipeline(it, dp_world_size=ctx.dp_world_size)
```

rakkit: FYI, mosaic have this [streaming](https://github.com/mosaicml/streaming) times ago. we have been use it years ago (composer + streaming + llm-foundry)

slimfrkha:
Hey @felipemello1  , just bumping this PR in case it got buried. happy to address any feedback.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+
+This code already tells the whole story. The problem is that it is not
+configurable: `parse_and_tokenize`, the packing bin count, the weights, the
+seed — all hardcoded. Adding a source or swapping the packing algorithm means
+editing this function. Sharing it across training recipes means copy-paste.
+
+The proposed design makes each step a named, typed `Op` in a dataclass config.
+`make_corpus` becomes `Source(pipeline=[ArrayRecordReader(...), Shuffle(),
+Repeat(), TextTokenize()])`. The post-mix chain becomes
+`pipeline=[FirstFitPack(num_packing_bins=64), Batch(), Prefetch(num_workers=16)]`.
```
felipemello1, line 75: i am not against it, but in the past i have heard complaints about this type of operation being hard to debug. That said, it is the pattern that I see huggingface and torchdata following. For the sake of brainstorming, another option would be to discretize it into functions that call these methods, but I guess this would be an anti-pattern for grain.

felipemello1:
> Hey @felipemello1 , just bumping this PR in case it got buried. happy to address any feedback.

hey @slimfrkha , thanks for the ping! I started reviewing it but I am a bit tight on time today. Let me try to finish it tomorrow.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+as class-level metadata:
+
+```python
+class Op(Configurable.Config):
+    accepts:  ClassVar[frozenset[Stage] | Literal["any"]]      # record shape accepted
+    produces: ClassVar[Stage | Literal["same"]]                 # record shape emitted
+    phase:    ClassVar[Literal["map", "iter", "any"]] = "any"  # grain backing required
+
+    def apply(self, ds, ctx: PipelineContext): ...
+```
```
felipemello1, line 129: just to be 100% clear, is this a `grain` convention or something you are proposing ?

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+```python
+@dataclass
+class Source:
+    pipeline: list   # [Reader, ...ops...]; SOURCE → ENCODED
+    weight: float = 1.0
+
+@dataclass
+class DataPipeline.Config:
+    sources: list    # list[Source]
+    pipeline: list   # list[Op]; ENCODED → TRAINBATCH
```
felipemello1, line 167: For this pipelien abstraction, is this something you saw in their tutorials? If so, would you have a link? It would make me more confident.

i saw them getting a [transform function and using .map](https://github.com/google/grain/blob/af382532a32422bed58e959ff91f423c28667820/docs/tutorials/data_sources/huggingface_dataset_tutorial.md?plain=1#L243-L254)

It goes a bit back to my point. We could do:
```python
dataset = (
    grain.MapDataset.source(hf_train)
    .shuffle(seed=42)
    .filter(lambda x: x["star"] > 3)  # filters samples
    .map(process_date)
    .map(process_transformer)
    .map(process_sample_to_np)
    .to_iter_dataset()
    .batch(batch_size=2)  # batches consecutive elements
)
```

I hear you saying that we should do:
```python
class DataPipeline.Config:
    sources: [hf_train]
    pipeline: [shuffle(seed=42), filter(lambda x: x["star"] > 3) , transformer_op, to_iter_dataset, batch(batch_size=2)]
```

but there is also another option, something like, which is more opinionated, but people can add their own builders for specific cases:

```python
class TextDataset()
	def __init__(self, source, process_fn: callable, filter: str,, seed:int = 42, shuffle:bool = True,  , batch_size:int = 2):
		ds = load_dataset(source)
		if shuffle:
			ds = ds.shuffle(seed)
		
		 if filter:
		      ds = apply_filter(filter, ds)
		
		if process_fn:
			ds = ds.map(process_fn)
		
		ds = ds.to_iter_dataset.batch(batch_size)
		
		return ds
```

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+        ),
+        Source(
+            pipeline=[ArrayRecordReader(file_glob="/data/code/*.array_record"),
+                      Shuffle(), Repeat(), TextTokenize(text_field="content", add_eos=False)],
+            weight=0.20,
+        ),
+        Source(
+            pipeline=[ArrayRecordReader(file_glob="/data/books/*.array_record"),
+                      Shuffle(), Repeat(), TextTokenize()],
+            weight=0.10,
```
felipemello1, line 556: nit: is there a way to do weight normalization, so that if a dataset gets added/removed, its not necessary to update all weights to sum to 1

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+        inputs = {("input" if k == "input_ids" else k): v for k, v in batched.items()}
+        return inputs, labels
+
+
+class Reader(Op):
+    """Root op: no input, produces RAW, creates the MapDataset."""
+    accepts, produces, phase = frozenset({Stage.SOURCE}), Stage.RAW, "map"
+
+
+# ── structural ops ────────────────────────────────────────────────────────────
```
felipemello1, line 697: this works, but i am a bit skeptical that we have to reimplement ops like `shuffle` and `repeat` in titan, if those are stablished concepts in grain

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+class Batch(BatchOp):
+    def combine(self, samples, ctx):
+        return {k: stack([s[k] for s in samples]) for k in samples[0]}
+
+
+# ── text ops ──────────────────────────────────────────────────────────────────
+
+@dataclass
+class TextTokenize(MapOp):
+    accepts, produces = frozenset({Stage.RAW}), Stage.ENCODED
```
felipemello1, line 749: this is not rocket science, but it is a new paradigm, more complex than just thinking: "I have a function, and i will apply it using `ds.map(tokenize_text)`.

felipemello1 (commented):
thanks for the RFC! The abstractions are nice and elegant, but i want to be conservative before committing to them.

My understanding is that the pipeline concept was created by you and not something suggested in their docs. Is this correct? have you seen some other library using `grain` this way?

I can see how we add a lot of flexibility and give power to the users. That said, it comes at a cost:
a) user has to understand how the `ops` machinery work;
b) we have to "reimplement" simple ops like `shuffle`
c) pipeline might be hard to debug, since it is harder to inspect in between ops.

I also understand that it is a popular way of doing things in libraries like langchain, but it is also a bit controversial. I know developers that dislike this type of abstraction.

---

Another option would be to be more opinionated, e.g. parsing a dataset has some key steps: loading, shuffling, filtering, applying a transform, packing, interleaving, dataloader.

We could discretize these steps and offer knobs.
```python
if shuffle:
    ds = ds.shuffle()
```

---

To be clear, i am not rejecting the pipeline abstraction, but i would be more comfortable accepting it if I could check how other people implement grain this way and battle tested it.

Thank you!!!

docs/rfcs/grain.md
slimfrkha: yes super easy to integrate a weight normalization step before this line.
```python
mixed = grain.MapDataset.mix(sources, weights=[s.weight for s in self.sources])
```

slimfrkha:
> thanks for the RFC! The abstractions are nice and elegant, but i want to be conservative before committing to them.
> 
> My understanding is that the pipeline concept was created by you and not something suggested in their docs. Is this correct? have you seen some other library using `grain` this way?
> 
> I can see how we add a lot of flexibility and give power to the users. That said, it comes at a cost: a) user has to understand how the `ops` machinery work; b) we have to "reimplement" simple ops like `shuffle` c) pipeline might be hard to debug, since it is harder to inspect in between ops.
> 
> I also understand that it is a popular way of doing things in libraries like langchain, but it is also a bit controversial. I know developers that dislike this type of abstraction.
> 
> Another option would be to be more opinionated, e.g. parsing a dataset has some key steps: loading, shuffling, filtering, applying a transform, packing, interleaving, dataloader.
> 
> We could discretize these steps and offer knobs.
> 
> ```python
> if shuffle:
>     ds = ds.shuffle()
> ```
> 
> To be clear, i am not rejecting the pipeline abstraction, but i would be more comfortable accepting it if I could check how other people implement grain this way and battle tested it.
> 
> Thank you!!!

Thanks @felipemello1  for the pushback. I am moving away from a big part of the current design.

To answer directly: yes, the Op/pipeline layer was my invention, not something Grain prescribes. Grain gives you MapDataset/IterDataset and chainable methods; the composable-op engine on top was mine. I don't have a battle-tested precedent to point you to. MaxText (which does use Grain) deliberately hardcodes a few fixed pipelines and exposes knobs (similar to the current torchtitan code).

Your three costs all land:
- steep learning curve for users
- wrapping one-liners of Grain calls (Shuffle, Repeat)
- difficulty to debug

The one thing the Op engine uniquely buys (compared to an opiniated pipeline with knobs) is reordering/inserting steps from a config file without touching Python. But this is rather rare user case that probably doesn't justify all the introduced machinery and its cons

---
New design 

What's gone: Op, Stage, Phase — and with them the op registry/__init_subclass__, the family bases (MapOp/FilterOp/PackOp/BatchOp), the Source wrapper, the auto-injecting orchestrator, and the YAML-loader idea. No engine.

What stays: 
- DataPipeline: wrapping the terminal Grain IterDataset:exposes __iter__ + Stateful, and refuses to restore a checkpoint whose dp_world_size changed (Grain can't reshard).
- DataPipeline.Config: abstract base
- PipelineContext: runtime objects, built once in the trainer and passed to build()

NEW:
- A small set of explicit named pipeline classes: PretrainPipeline, SFTPipeline; each a single build() that writes the raw Grain chain top-to-bottom (shuffle/repeat/pack/etc. are just lines, not ops). Users pick a class in config registry and turn flat knobs
- Anything custom (span masking, a new modality) is a DataPipeline.Config subclass with its own build() in raw Grain — the escape hatch is just Grain, which is documented and battle-tested.

Sketch code
```python
@dataclass
class PipelineContext:
    tokenizer: BaseTokenizer
    seq_len: int
    dp_rank: int
    dp_world_size: int
    local_batch_size: int

class DataPipeline:
    def __init__(self, ds, dp_world_size):
      self._ds, self._dp = ds, dp_world_size
    def __iter__(self):
      return iter(self._ds)
    def state_dict(self):
      return {"ds": self._ds.state_dict(), "dp_world_size": self._dp}
    def load_state_dict(self, sd):
        if sd["dp_world_size"] != self._dp:
            raise ValueError("dp_world_size changed; resharding not supported")
        self._ds.load_state_dict(sd["ds"])

class DataPipeline.Config:   # abstract base (nested)
    seed: int = 42
    def build(self, ctx) -> DataPipeline: raise NotImplementedError

class PretrainPipeline(DataPipeline.Config):
    def build(self, ctx): ...   # full inline
    # use something like the below link to detect streaming or map
    # https://github.com/pytorch/torchtitan/pull/3415#discussion_r3331885345

class SFTPipeline(DataPipeline.Config):
    def build(self, ctx): ...   # full inline
```

slimfrkha:
> > thanks for the RFC! The abstractions are nice and elegant, but i want to be conservative before committing to them.
> > My understanding is that the pipeline concept was created by you and not something suggested in their docs. Is this correct? have you seen some other library using `grain` this way?
> > I can see how we add a lot of flexibility and give power to the users. That said, it comes at a cost: a) user has to understand how the `ops` machinery work; b) we have to "reimplement" simple ops like `shuffle` c) pipeline might be hard to debug, since it is harder to inspect in between ops.
> > I also understand that it is a popular way of doing things in libraries like langchain, but it is also a bit controversial. I know developers that dislike this type of abstraction.
> > Another option would be to be more opinionated, e.g. parsing a dataset has some key steps: loading, shuffling, filtering, applying a transform, packing, interleaving, dataloader.
> > We could discretize these steps and offer knobs.
> > ```python
> > if shuffle:
> >     ds = ds.shuffle()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To be clear, i am not rejecting the pipeline abstraction, but i would be more comfortable accepting it if I could check how other people implement grain this way and battle tested it.
> > Thank you!!!
> 
> Thanks @felipemello1 for the pushback. I am moving away from a big part of the current design.
> 
> To answer directly: yes, the Op/pipeline layer was my invention, not something Grain prescribes. Grain gives you MapDataset/IterDataset and chainable methods; the composable-op engine on top was mine. I don't have a battle-tested precedent to point you to. MaxText (which does use Grain) deliberately hardcodes a few fixed pipelines and exposes knobs (similar to the current torchtitan code).
> 
> Your three costs all land:
> 
> * steep learning curve for users
> * wrapping one-liners of Grain calls (Shuffle, Repeat)
> * difficulty to debug
> 
> The one thing the Op engine uniquely buys (compared to an opiniated pipeline with knobs) is reordering/inserting steps from a config file without touching Python. But this is rather rare user case that probably doesn't justify all the introduced machinery and its cons
> 
> New design
> 
> What's gone: Op, Stage, Phase — and with them the op registry/**init_subclass**, the family bases (MapOp/FilterOp/PackOp/BatchOp), the Source wrapper, the auto-injecting orchestrator, and the YAML-loader idea. No engine.
> 
> What stays:
> 
> * DataPipeline: wrapping the terminal Grain IterDataset:exposes **iter** + Stateful, and refuses to restore a checkpoint whose dp_world_size changed (Grain can't reshard).
> * DataPipeline.Config: abstract base
> * PipelineContext: runtime objects, built once in the trainer and passed to build()
> 
> NEW:
> 
> * A small set of explicit named pipeline classes: PretrainPipeline, SFTPipeline; each a single build() that writes the raw Grain chain top-to-bottom (shuffle/repeat/pack/etc. are just lines, not ops). Users pick a class in config registry and turn flat knobs
> * Anything custom (span masking, a new modality) is a DataPipeline.Config subclass with its own build() in raw Grain — the escape hatch is just Grain, which is documented and battle-tested.
> 
> Sketch code
> 
> ```python
> @dataclass
> class PipelineContext:
>     tokenizer: BaseTokenizer
>     seq_len: int
>     dp_rank: int
>     dp_world_size: int
>     local_batch_size: int
> 
> class DataPipeline:
>     def __init__(self, ds, dp_world_size):
>       self._ds, self._dp = ds, dp_world_size
>     def __iter__(self):
>       return iter(self._ds)
>     def state_dict(self):
>       return {"ds": self._ds.state_dict(), "dp_world_size": self._dp}
>     def load_state_dict(self, sd):
>         if sd["dp_world_size"] != self._dp:
>             raise ValueError("dp_world_size changed; resharding not supported")
>         self._ds.load_state_dict(sd["ds"])
> 
> class DataPipeline.Config:   # abstract base (nested)
>     seed: int = 42
>     def build(self, ctx) -> DataPipeline: raise NotImplementedError
> 
> class PretrainPipeline(DataPipeline.Config):
>     def build(self, ctx): ...   # full inline
>     # use something like the below link to detect streaming or map
>     # https://github.com/pytorch/torchtitan/pull/3415#discussion_r3331885345
> 
> class SFTPipeline(DataPipeline.Config):
>     def build(self, ctx): ...   # full inline
> ```

Hey @felipemello1, can you please have a look at the above ? Thanks 🙏

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+
+## Open Questions
+
+**1. Should PRs 2–5 follow the sequencing above, or is a different ordering preferred?**
+
+The plan above bundles the preprocessing tool with the deletion (PR 2) to
+avoid a gap where the old stack is gone but users have no path to prepare
+their data. If the preprocessing tool needs more design time, an alternative
+is to land the framework and text ops first while keeping the old stack, then
+delete once the tool ships.
```
fegin, line 484: The design is pretty self-contained and can be verified without integrating to the trainer. I prefer that we replace the existing dataloader after PR4 is finished.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+torchtitan's data configs cover source selection well — dataset name, path,
+per-source weights, seed — but processing behavior is not configurable at all:
+the text field, tokenization options, packing algorithm, and any filtering are
+hardcoded inside the dataset classes. Changing any of them means editing Python. This RFC replaces torchtitan's current data stack with a Grain-based pipeline where both source selection
+and processing behavior are expressed as composable `Op`s in a dataclass
+config. The old stack is deleted entirely.
+
+## Motivation
+
+Four concrete problems drive this change.
```
fegin, line 14: There are only 3 problems listed below. Anything you want to add?

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+
+A `Source` is a reader-headed op list plus its mixing weight. The post-mix
+`pipeline` runs over the combined stream after all sources are mixed. Both are
+validated at `__post_init__`: per-source pipelines must run `SOURCE → ENCODED`;
+the post-mix pipeline must run `ENCODED → TRAINBATCH`.
+
+The fields are typed as bare `list` — not `list[Op]` — matching torchtitan's
+existing tyro pattern for heterogeneous lists. Pipeline structure is built
+before tyro runs (in `config_registry.py` or from a future YAML file), and
+tyro layers leaf overrides on top. CLI overrides stay clean:
```
fegin, line 179: > The fields are typed as bare `list` — not `list[Op]` — matching torchtitan's
existing tyro pattern for heterogeneous lists.

Is this True? Or just some inconsistencies in the code? It seems to me that most configs are fully typed.

> CLI overrides stay clean:

We need to think about how many configurations we want to expose to to CLI. All components are config-based, which mean that users can adjust the need via configurations without exposing to CLI.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+
+```
+--sources.0.weight=0.65 --pipeline.0.num_packing_bins=128
+```
+
+### Framework-internal concerns
+
+These are handled inside `DataPipeline.Config.build(ctx)` and never appear in
+user configs.
```
fegin, line 188: Can't the context be part of the config of DataPipeline? We didn't enforce this for non-module components. But asking to understand the limitation as that is the pattern we would like to achieve.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+**Mixing.** `MapDataset.mix(sources, weights, seed)` is structural — it is
+many-to-one and does not fit the `list[Op]` model, so it is not an op.
+
+**Checkpointing.** Delegated entirely to Grain's `IterDataset` state dict.
+
+The runtime object is a `DataPipeline` shim (~15 lines) that wraps the
+terminal `IterDataset`, exposes `__iter__` and `Stateful`, and refuses to
+load a checkpoint whose `dp_world_size` differs from the current run. Grain
+does not support resharding, and silently restoring the wrong iterator state
+would be worse than a clean failure.
```
fegin, line 213: If users want to implement their own resharding algorithm, is it possible?

docs/rfcs/grain.md
fegin, line 281: Can this and/or other custom ops be used for customized data shuffling for load balancing? But the op needs to know the data parallel world size.

docs/rfcs/grain.md
```diff
@@ -0,0 +1,899 @@
...
+- **Mixing** dispatches to `IterDataset.mix`. All sources must share the same
+  backing — a mix of `MapDataset` and `IterDataset` sources is a config error.
+- **DP sharding** becomes a stateful strided op: each rank keeps element `i`
+  where `i % dp_world_size == dp_rank`, rather than an index-slice.
+- **Shuffle** degrades to `WindowShuffle` — an approximate, buffer-based
+  shuffle instead of a global permutation.
+
+The cost: each rank reads the full stream and discards `(N-1)/N` of it
+(read amplification), and shuffle quality is approximate. The gain: streaming
+sources without an offline preprocessing step.
```
fegin, line 398: > The cost: each rank reads the full stream and discards `(N-1)/N` of it
(read amplification), and shuffle quality is approximate. 

N is the DP world size?

docs/rfcs/grain.md
slimfrkha: That's just a draft typo. can ignore.

docs/rfcs/grain.md
slimfrkha: > Is this True? Or just some inconsistencies in the code? It seems to me that most configs are fully typed.

1. This citation is stale. It is about ModelConvertersContainer that no longer exists (only in torchtitan/models/flux/README.md). It became model_registry(converters=list[ModelConfigConverter.Config]). heterogeneous but not tyro exposed.
2. Our case is different: DataPipeline.Config/Source lists are heterogeneous and on the tyro path.
3. Bare list breaks the CLI; list[Op]/list[Source] fixes it. Tested (tyro 1.0.6, real bases). I had it backwards.

Conclusion: type list[Op] / list[Source]; drop the bare-list rec.

slimfrkha: > > CLI overrides stay clean:
> 
> We need to think about how many configurations we want to expose to to CLI. All components are config-based, which mean that users can adjust the need via configurations without exposing to CLI.

I'd lean towards this design: one composable `DataPipeline.Config` + some thin factory functions:
 - Standard use: factory functions returning a `DataPipeline.Config`: `build_pretraining_map_pipeline`,
  `build_pretraining_stream_pipeline`, `build_sft_map_pipeline`, `build_sft_stream_pipeline`. Same pattern as `model_registry`.
  - Custom use : construct `DataPipeline.Config` directly in `config_registry`
  - CLI like --data.pipeline.3.num_workers=16, Handy for power users that wanna grid search some pipeline related fields. CLI exposes all fields to the user as we dont know in advance what the user wants to tweak without touching py code.

docs/rfcs/grain.md
slimfrkha: `PipelineContext` isn't part of `DataPipeline.Config`. Config holds what the user declares once and what's serializable; ctx carries runtime objects/derived values threaded in at `build()`. The five fields are excluded for three reasons:
  - `dp_rank`, `dp_world_size`: runtime-derived from the mesh after distributed init; don't exist at parse time, and exposing them would corrupt sharding.
  - `seq_len`, `local_batch_size`: single-source-of-truth; already canonical in training.
  - `tokenizer`: single-source-of-truth (declared at config.tokenizer, built once, shared to dataloader + validator) and non-serializable (holds callables).

This is the established pattern, not a data-specific limitation: `Configurable.build(**kwargs)` exists precisely to pass runtime objects (tokenizer, ranks, etc.) that don't belong in config. used the same way by tokenizer, optimizer, checkpoint, and validator.

docs/rfcs/grain.md
slimfrkha: My initial investigations:
- Map (indexable) sources: resharding is possible: either out-of-the-box via Grain's ElasticIterator (pre-pack region) or with a user-written script.
- Streaming sources: hard to not possible. you'd need to replay the whole source stream. stream needs to be perfectly deterministic.

docs/rfcs/grain.md
slimfrkha: 1. `dp_world_size` is already accessible. `ctx` is passed to every `Op.apply()` and to `DataPipeline.Config.build().` No new wiring needed.
2. You can't do it with a plain `Op`. DP sharding is framework-injected (the `[rank::world]` slice), not an exposed `Op`, and it sits between mix and the post-mix pipeline. No user-op slot sees the global pre-shard stream. So to control sharding/balancing you must subclass `DataPipeline.Config` and override `build()` (the escape hatch).
3. You may need to inject some offline metadata in the data beforehand (like maybe records lengths) which Grain can't give lazily. So inject metadata then use in custom `build()`

docs/rfcs/grain.md
slimfrkha: yes `dp_world_size`

slimfrkha:
Hi folks,
Bumping this up. @tianyu-l @fegin @felipemello1

felipemello1:
hey @slimfrkha , i am very sorry for the delay. We are sprinting for a deadline for this month (end of H1). After that, i will make sure to prioritize this RFC and work with you to land the changes in titan, which will be impactful. Meanwhile, maybe tianyu or fegin can comment further. Otherwise i will do it in a few days, after our deadline. Thanks for keeping pushing for it!

slimfrkha:
Sure I understand ! Just wanted to put this on your radar guys and not get forgotten.
Happy to address any comment when you're available again.

felipemello1:
hey @slimfrkha , fyi i am working on a reply/some ideas.

felipemello1:
hey @haydn-jones @slimfrkha @rakkit , we merged the grain PR support in titan https://github.com/pytorch/torchtitan/pull/4088

Please let us know if you guys have any feedback. We can address it in follow ups. Thanks for the work/review in this RFC!!!

PR #3416, opened 2026-05-21
[graph_trainer] Fix fsdp_passes compat with BitsetAncestors from pytorch passes

IvanKobzarev:
…rch#183559

pytorch#183559 changed `node_ancestors` from `dict[Node, OrderedSet[Node]]` to `BitsetAncestors` in the overlap scheduler. The `_apply_trailing_block` method in `JointManualOverlapScheduler` used `OrderedSet(self.node_ancestors[node])` which breaks with the new API.

Use `is_ancestor()` when available, with an `OrderedSet` fallback for pytorch builds that predate the change.

Authored by Claude.


SherlockNoMad (approved):

SherlockNoMad:
tbh, I suspect we will end up cloning the manual overlap pass into titan for agility. 

we have done so for selective_remat_pass...

PR #3420, opened 2026-05-22
[Don't reveiw][GraphTrainer] Kernel fusion workflow

SherlockNoMad:
experimental only. 

Draft PR for code share and discussion. 

Summary

  This branch implements an offline fused kernel generation and acceleration pipeline for graph_trainer. It's a 3-step workflow that discovers fusible operator regions in the FX graph, generates optimized Triton
   kernels offline, and conditionally replaces those regions at training time.

  The 3-Step Workflow

  1. Extract (FUSED_KERNEL_DIR=/tmp/kernels ./run_train.sh) — During training, the fused_kernel_pass scans the FX graph for fusible regions (groups of pointwise, reduction, norm ops connected by data flow),
  computes a stable hash for each unique region (ops + shapes), and writes a problem.py file describing each region. No graph modification — training runs unmodified.
  2. Generate + Benchmark (python -m ...kernel_gen.generate --dir /tmp/kernels) — Offline step. For each problem.py, calls KernelAgent (an LLM-based Triton kernel generator) to produce a kernel.py, then
  benchmarks it against eager PyTorch and torch.compile. Writes benchmark.json recording which backend is fastest. Includes correctness validation (5 random trials, auto-detected tolerance tiers).
  3. Replace (same command as step 1) — On the next training run, the same fused_kernel_pass finds the same regions by hash, checks benchmark.json, and only replaces regions where a non-eager backend proved
  faster. Replacement inserts a direct call_function(kernel_fn) — zero dispatch overhead. Regions without kernels or where eager won stay untouched.

  Key Design Decisions

  - Conditional replacement: no benchmark.json = no replacement. Safety-first — unbenchmarked kernels are never used.
  - Pluggable extractors: fqn (default, segments at module boundaries) and inductor (uses inductor's decomposition + is_fusible_node classification, maps back via from_node provenance).
  - Op filtering: collectives (_c10d_functional, bucketing), flash attention, embeddings, and bare matmuls are excluded. Matmul is only kept if it has epilog compute (e.g., mm->cast).
  - Backend auto-selection: per-region, picks triton > torch.compile > eager based on benchmark timings.
  - Hash-based directory layout: each region gets a directory named by its content hash, making the pipeline idempotent and cache-friendly.


torchtitan/experiments/graph_trainer/kernel_gen/generate.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3
+"""Run KernelAgent on all problem files: generate + NCU-guided optimize.
+
+For each problem in generated/<name>/:
+  1. Generate a correct Triton kernel (if kernel.py doesn't exist)
+  2. Optimize with NCU profiling (if optimized_kernel.py doesn't exist)
```
SherlockNoMad, line 6: @kaiming-cheng 

this is currently doing kernel_gen, and kernel_optimize in 2 steps. 
I think they can be merge as one step?

torchtitan/experiments/graph_trainer/kernel_gen/generate.py
```diff
@@ -0,0 +1,300 @@
...
+  python -m torchtitan.experiments.graph_trainer.kernel_gen.generate --skip-optimize  # generation only
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from pathlib import Path
+
```
SherlockNoMad, line 20: @kaiming-cheng 

Can you help review the use of KernelAgent in this file to ensure best practice?

torchtitan/experiments/graph_trainer/kernel_gen/kernelagent_bridge.py
```diff
@@ -0,0 +1,319 @@
...
+    try:
+        return agent.generate_kernel(
+            problem_description,
+            test_code=test_code,
+        )
+    finally:
+        agent.cleanup()
+
+
+_DEFAULT_MODEL = "claude-opus-4-7"
```
SherlockNoMad, line 172: you mention better model ? which one should I use?

torchtitan/experiments/graph_trainer/kernel_gen/generate.py
SherlockNoMad: Also, what are the knobs I can use to ensure iterative improvement for generated kernels?

torchtitan/experiments/graph_trainer/kernel_gen/kernelagent_bridge.py
kaiming-cheng: We found gpt-5.5 with thinking gives us the best performance. For KA's previous experiment, we use internal plugboard as the endpoint to access these models. I can share the doc with you to see how we can integrate the latest model here.

torchtitan/experiments/graph_trainer/kernel_gen/generate.py
kaiming-cheng: It should be fine to keep both steps since generation focused on coverage and optimization focused on performance. I suspect we will do more of an optimize only run after the first pass (i.e., we achieve 100% coverage on the kernels)

torchtitan/experiments/graph_trainer/kernel_gen/generate.py
```diff
@@ -0,0 +1,300 @@
...
+        return {"name": name, "success": False, "message": "no problem.py"}
+
+    from torchtitan.experiments.graph_trainer.kernel_gen.kernelagent_bridge import optimize_kernel
+
+    kernel_code = kernel_path.read_text()
+    opt_dir = str(problem_dir / "opt_logs")
+
+    print(f"[{name}] Optimizing with NCU ({opt_rounds} rounds)...", flush=True)
+    try:
+        result = optimize_kernel(
```
kaiming-cheng, line 102: We can create an `OptimizationManager` class to handle the optimization here. For example, here is how we run the [optimization example](https://github.com/meta-pytorch/KernelAgent/blob/main/examples/run_opt_manager.py#L48) :

```

#Use this to get the optimization result 
result = _run_strategy(
            args.strategy (default to beam_search),
            kernel_code,
            problem_file,
            test_code,
            log_dir,
            max_rounds=args.max_rounds,
        )

def _run_strategy(
    strategy: str,
    kernel_code: str,
    problem_file: Path,
    test_code: str,
    log_dir: Path,
    max_rounds: int | None = None,
) -> dict:
    """Run a single strategy using its config file."""
    config_path = _CONFIGS_DIR / f"{strategy}.yaml"
    if not config_path.exists():
        print(f"ERROR: config not found: {config_path}")
        sys.exit(1)

    print("\n" + "=" * 80)
    print(f"{strategy.upper()} OPTIMIZATION")
    print("=" * 80)
    print(f"Config: {config_path}")

    manager = OptimizationManager(
        config=str(config_path),
        log_dir=log_dir / strategy,
        database_path=log_dir / strategy / "program_db.json",
        openai_model="gpt-5.5"
    )

    return manager.run_optimization(
        initial_kernel=kernel_code,
        problem_file=problem_file,
        test_code=test_code,
        max_rounds=max_rounds,
    )
```
Where the `OptimizationManager` will apply a more powerful optimization strategy, i.e., 

```
strategy: str = "beam_search",
        num_workers: int = 4,
        max_rounds: int = 10,

```

This will yield performance a lot stronger than single worker greedy based approach.

torchtitan/experiments/graph_trainer/kernel_gen/generate.py
kaiming-cheng: mostly these knobs: 
1. num_workers (higher for test-time scaling) 
2. max_rounds (higher for test-time scaling) 
3. openai_model (we can try several frontier models and pick the best one) 
4. high_reasoning_effort = True 

For say, after an optimization loop with a given model (opus-4-7 for say), I will feed it to another optimization loop with different models

PR #3423, opened 2026-05-23
[MoE][7/n] Keep 3D tensors through MoE, flatten inside GroupedExperts

acisseJZhong:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #3425
* __->__ #3423

## Summary
Address https://github.com/pytorch/torchtitan/pull/3386#discussion_r3269104636

Move `.view(-1, dim)` from `MoE.forward()` into `GroupedExperts.forward()` / `GptOssGroupedExperts.forward()`, so the router and shared experts see 3D `(bs, slen, dim)` tensors. This avoids `StridedShard` when `.view(-1, dim)` is called on a `Shard(1)` 3D DTensor at the MoE boundary.

- `GroupedExperts.forward` / `GptOssGroupedExperts.forward`: flatten to 2D at entry
- `TokenChoiceTopKRouter`: use `dim=-1` and relative dim ops to handle any number of leading dims
- `moe_sharding.py`: update placements for 3D shapes (`Shard(1)` for slen, `Shard(2)` for feature dim)

## Test
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2 --training.local_batch_size=1" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2 --training.local_batch_size=1" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```

PR #3389, opened 2026-05-17
[MoE][6/n] Extract local_reorder, split DeepEP/HybridEP dispatchers

acisseJZhong:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* __->__ #3389
* __->__ #3386
* #3385

## Summary
MoE cleanup from https://github.com/pytorch/torchtitan/issues/2956:

- **Extract `local_reorder` helper** on `LocalTokenDispatcher`: deduplicates the histc/argsort/score-weighting block between `LocalTokenDispatcher.dispatch()` and `AllToAllTokenDispatcher.dispatch()`
- **Split `DeepEPTokenDispatcher` into `DeepEPTokenDispatcher` + `HybridEPTokenDispatcher`**: eliminates `comm_backend` string branching; config fields (`non_blocking_capacity_factor`, `pad_multiple`) are now class-specific. Callsites use `isinstance` instead of `getattr(..., "comm_backend", ...)`
- **Fix DeepEP CUDA stream race** (#3413): added `torch.cuda.synchronize()` before `get_dispatch_layout` in `dispatch_tokens()` to ensure `masked_fill` results are visible on DeepEP's comm stream. Without this, `get_dispatch_layout` reads stale (pre-masked) expert indices, causing `num_tokens_per_expert` to overcount and `_grouped_mm` to assert.

## Test plan
Test loss comparions 
```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel_ep' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2 --parallelism.no-enable_sequence_parallel" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel_ep' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2 --parallelism.no-enable_sequence_parallel" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```

PR #3425, opened 2026-05-24
[MoE][8/n] Use shape suffix for MoE

acisseJZhong:
* __->__ #3425
* __->__ #3423
* __->__ #3389 
* __->__ #3386
* __->__ #3167 
* __->__ #3125
* __->__ #2960
* __->__ #2842

Change MoE shape notation to https://medium.com/@NoamShazeer/shape-suffixes-good-coding-style-f836e72e24fd

```
NCCL_NVLS_ENABLE=0 python scripts/loss_compare.py . main --baseline-module='llama4' --baseline-config='llama4_debugmodel' --baseline-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2" --baseline-ngpus=4 --test-module='llama4' --test-config='llama4_debugmodel' --test-options="--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2 --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2" --test-ngpus=4 --assert-equal --no-seed-checkpoint
```


acisseJZhong:
@claude review

acisseJZhong:
@claude review this PR, especially if there are any typos, try to be concise.

acisseJZhong:
@claude review again

torchtitan/models/common/moe.py
```diff
@@ -144,34 +157,34 @@ def __init__(self, config: Config):
...
     ) -> tuple[torch.Tensor, torch.Tensor]:
         """Balanced round-robin expert assignment.
-        Returns (selected_experts_indices (bs, slen, top_k) LongTensor, top_scores (bs, slen, top_k) FloatTensor).
+        Returns (topk_ids_BLK ``(B, L, K)`` LongTensor, topk_scores_BLK ``(B, L, K)`` FloatTensor).
         """
-        bs, slen, _ = scores.shape
+        bs, slen, _ = scores_BLE.shape
         # Round-robin indices with exact balance
-        selected_experts_indices = (
+        topk_ids_BLK = (
```
tianyu-l: why renaming?

torchtitan/models/common/moe.py
tianyu-l: how about this? should be `expert_bias_E`

torchtitan/models/common/token_dispatcher.py
```diff
@@ -16,13 +16,21 @@
 from torchtitan.config import Configurable
 from torchtitan.ops.scatter_add import deterministic_scatter_add
 
+# Shape suffix legend
+# (https://medium.com/@NoamShazeer/shape-suffixes-good-coding-style-f836e72e24fd):
+#   B = batch, L = sequence length, D = model dimension,
+#   F = hidden (FFN intermediate) dimension, E = num experts,
+#   e = num local experts (E / EP), K = top-k,
+#   T = num tokens (B*L flattened), N = routed tokens (T*K),
+#   R = routed tokens assigned to local experts
```
tianyu-l: Keep the one in moe.py is good enough, let's not maintain two copies

torchtitan/models/common/moe.py
```diff
@@ -18,6 +18,14 @@
 
 from .token_dispatcher import DeepEPTokenDispatcher, LocalTokenDispatcher
 
+# Shape suffix legend
+# (https://medium.com/@NoamShazeer/shape-suffixes-good-coding-style-f836e72e24fd):
+#   B = batch, L = sequence length, D = model dimension,
+#   F = hidden (FFN intermediate) dimension, E = num experts,
+#   e = num local experts (E / EP), K = top-k,
```
tianyu-l: Did you use `e` anywhere? I'm not sure if we want to differentiate local vs. global.

`e` is EP infra details; you didn't differentiate B, D, F, etc. for DP / TP. Given how torchtitan aims to decouple infra and modeling, I'd suggest we keep this Noam notation to only worry about global semantics (not sharded ones).

torchtitan/models/common/moe.py
```diff
@@ -18,6 +18,14 @@
...
+#   T = num tokens (B*L flattened), N = routed tokens (T*K),
+#   R = routed tokens assigned to local experts
```
tianyu-l: It's very unclear what the differences between T and R are.
- In current code T->R transition happens at "local reorder", which doesn't match the description here. It matches better if T->R transition happens at "all-to-all"
- Also here the description sounds we should use `r` instead of `R`.

torchtitan/models/common/moe.py
tianyu-l, line 1: we should also do this to attention layer, given different attention backends order dimensions differently, e.g.
https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/attention.py#L121-L123

worth a separate PR

torchtitan/models/common/moe.py
acisseJZhong: I think `topk_ids_BLK` and `topk_scores_BLK` align better? wdyt about `topk_expert_ids_BLK`

torchtitan/models/common/moe.py
acisseJZhong: yeah i used `num_tokens_per_local_expert_e` in token dispatcher. Ideally it should be `num_tokens_per_local_expert_e` to indicate it's num of tokens per local experts, differentiating from `E`. 

So in moe.py there's only `num_tokens_per_expert_E`, but in token_dispatcher.py, there is both `num_tokens_per_expert_E `and `num_tokens_per_local_expert_e`. Do you think this is fine?

torchtitan/models/common/moe.py
acisseJZhong: I added one line comment to all2all token dispatcher where `R = sum(num_tokens_per_local_expert_e)` and local dispatcher where `R = sum(num_tokens_per_local_expert_E)`. 

T->R happens inside the dispatcher dispatch. 

Regarding the need to have r/R to mirror e/E, Claude gives this argument which I think makes sense:

> The E/e distinction is useful because both appear as
>   explicit tensor dimensions in the same scope (weights are (E, F, D),
>   token counts are (e,)). You need to tell them apart.
> 
>   R doesn't have that problem — there's no place where global-R and
>   local-R coexist in the same scope. Before all-to-all it's N = T*K, after
>    all-to-all it's the local share, and the variable just gets reassigned.
>    The current legend definition ("routed tokens assigned to local
>   experts") already covers it.

torchtitan/models/common/token_dispatcher.py
```diff
@@ -245,120 +245,132 @@ def dispatch(
...
+            num_local_tokens_per_expert_E,
+            token_indices_experts_sorted_N,
+            topk_scores_experts_sorted_N,
+        ) = self._local_reorder(x_TD, topk_scores_TK, topk_expert_ids_TK)
 
         # generate the input splits and output splits for all-to-all
         with torch.no_grad():
-            num_tokens_per_expert_group = all_to_all_single(
-                num_tokens_per_expert,
+            num_global_tokens_per_local_expert_eEP = all_to_all_single(
```
tianyu-l: change back to E as discussed offline -- one letter one dim

torchtitan/models/common/moe.py
```diff
@@ -18,6 +18,14 @@
 
 from .token_dispatcher import DeepEPTokenDispatcher, LocalTokenDispatcher
 
+# Shape suffix legend
+# (https://medium.com/@NoamShazeer/shape-suffixes-good-coding-style-f836e72e24fd):
+#   B = batch, L = sequence length, D = model dimension,
+#   F = hidden (FFN intermediate) dimension, E = num experts,
+#   e = num local experts (E / EP), EP = expert-parallel ranks,
```
tianyu-l: update accordingly -- remove EP
also it might be worth adding a note why we special case `e` in tokens_per_expert, and it's not used in EP-sharded grouped experts

torchtitan/models/gpt_oss/moe.py
```diff
@@ -106,55 +106,66 @@ def _experts_forward(
...
+            .to(num_tokens_per_expert_E.dtype)
         )
         num_tokens_per_expert_long = torch.cat(
-            [num_tokens_per_expert, tail_slack]
+            [num_tokens_per_expert_E, tail_slack]
         ).long()
 
-        h = torch._grouped_mm(
-            x.bfloat16(), mlp1_weight.transpose(-2, -1).bfloat16(), offs=offsets
+        h_R2F = torch._grouped_mm(
```
tianyu-l: Note that this is also violating 1 ascii 1 dim principle. E.g. how I can differentiate a 2 dim tensor from a 3 dim tensor where 2nd dim is of size "2"?

Maybe the obvious way is to introduce another letter which mathematically == 2F.

I'm also OK to live with this, but please add comment here.

tianyu-l (approved):

PR #3426, opened 2026-05-25
[forge] bugfix: ForgeEngine broken with `ChunkedCELoss`

zTonyZhao:
# Summary

`ForgeEngine` is currently broken on main branch. Constructing a Trainer from torchtitan/experiments/forge/example_train.py fails with:

```
    File "/[venv]/lib/python3.12/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 367, in wrapper
      return f(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^
    File "/[workspace]/torchtitan/torchtitan/experiments/forge/example_train.py", line 47, in __init__
      super().__init__(config)
    File "/[venv]/lib/python3.12/site-packages/torch/distributed/elastic/multiprocessing/errors/__init__.py", line 367, in wrapper
      return f(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^
    File "/[workspace]/torchtitan/torchtitan/experiments/forge/engine.py", line 164, in __init__
      self.loss_fn = self.train_spec.loss.build(
                     ^^^^^^^^^^^^^^^^^^^^
  AttributeError: 'ModelSpec' object has no attribute 'loss'
```

This was introduced by #2937. That PR moved the loss config from `ModelSpec.build_loss_fn` to the top-level `Trainer.Config.loss` field and updated torchtitan/trainer.py accordingly, but the sibling change to ForgeEngine was wrong: it became `self.train_spec.loss.build(...)` even though `ModelSpec` no longer has a loss field at all.

# Changes

- Build the loss from the top-level config, matching Trainer: `self.loss_fn = config.loss.build(compile_config=config.compile)`.
- Add a `loss: BaseLoss.Config` field on ForgeEngine.Config so that ForgeEngine can be used standalone (the existing example_train.py path keeps working because it passes the wider Trainer.Config whose loss is populated by the model registry).
- Port the ChunkedCELoss `lm_head` wiring from Trainer. Without this, models whose registry defaults to ChunkedCELoss.Config() (e.g. llama3, qwen3) would crash again during the first forward.
- Fix an unrelated `NameError` introduced by #2822: `parallelism_config.disable_loss_parallel` -> `config.parallelism.disable_loss_parallel`.

No behavior change for torchtitan/trainer.py; this only realigns ForgeEngine with the post-#2937 loss-building contract.

### DIFF
--- a/torchtitan/experiments/forge/engine.py
+++ b/torchtitan/experiments/forge/engine.py
@@ -13,7 +13,7 @@ import torch
 from torch.distributed.elastic.multiprocessing.errors import record
 
 from torchtitan.components.checkpoint import CheckpointManager
-from torchtitan.components.loss import LossFunction
+from torchtitan.components.loss import BaseLoss, ChunkedCELoss, LossFunction
 from torchtitan.components.lr_scheduler import LRSchedulersContainer
 from torchtitan.components.optimizer import OptimizersContainer
 from torchtitan.config import Configurable, TORCH_DTYPE_MAP
@@ -54,6 +54,7 @@ class ForgeEngine(torch.distributed.checkpoint.stateful.Stateful, Configurable):
         compile: CompileConfig = field(default_factory=CompileConfig)
         comm: CommConfig = field(default_factory=CommConfig)
         debug: DebugConfig = field(default_factory=DebugConfig)
+        loss: BaseLoss.Config = field(default_factory=BaseLoss.Config)
 
         def to_dict(self) -> dict[str, Any]:
             return asdict(self)
@@ -161,9 +162,7 @@ class ForgeEngine(torch.distributed.checkpoint.stateful.Stateful, Configurable):
             init_device = device_type
             buffer_device = None
 
-        self.loss_fn = self.train_spec.loss.build(
-            config.compile, parallel_dims=parallel_dims
-        )
+        self.loss_fn = config.loss.build(compile_config=config.compile)
 
         # verify batch sizes
         global_batch_size = config.training.global_batch_size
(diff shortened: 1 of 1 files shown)


felipemello1:
hey @zTonyZhao, i am happy to approve it, but could you sanity check what would happen for the RL script in forge? This seems to work well for SFT in forge using titan, but for RL, would everything still work fine?

for example, rl losses dont have the attr "set_lm_head"

zTonyZhao:
> hey @zTonyZhao, i am happy to approve it, but could you sanity check what would happen for the RL script in forge? This seems to work well for SFT in forge using titan, but for RL, would everything still work fine?
> 
> for example, rl losses dont have the attr "set_lm_head"

Thank you so much for your review.

There seems to be only a SFT script in module `torchtitan.experiments.forge`. No RL scripts were found in `forge` module, while module `torchtitan.experiments.rl` does not use ForgeEngine but defines its own actors. In fact, this module was not used anywhere but the module itself.

Maybe I'm looking in the wrong place. Could you please further point me to the specific RL script or module you have in mind? I’d be happy to double-check and make sure it aligns with this PR.

felipemello1:
@zTonyZhao , in the forge GRPO reciep, it imports a TrainerActor https://github.com/meta-pytorch/torchforge/blob/b32f06b6417a6d3e7ed50355036a1133e000d932/apps/grpo/main.py#L21

which also uses ForgeEngine: https://github.com/meta-pytorch/torchforge/blob/b32f06b6417a6d3e7ed50355036a1133e000d932/src/forge/actors/trainer/titan.py#L41

I assume that you were looking in Forge, is that correct? we stopped developing in Forge in favor of TitanRL (which doesnt use ForgeEngine), but we kept the ForgeEngine here for now.

PR #3428, opened 2026-05-26
[graph_trainer] Fix precompile pickle failure with FlexAttention BlockMask

bobrenjc93:
## Summary

Fix `TypeError: cannot pickle code objects` when precompiling models with FlexAttention (e.g. Llama3-70B with `attn_backend="flex"`).

### Root cause

Two PRs combined to create this bug:

1. **#2916** ("Add CooR precompile support for DeepSeek V3") hoisted `BlockMask` into `extra_kwargs["attention_masks"]` so FlexAttention masks flow through `minimal_fx_tracer` as kwargs. This was fine at the time because `user_inputs_spec` was not serialized.

2. **#3262** ("Support kwargs in minimal_fx_tracer") added `user_inputs_spec` to `PrecompiledFxTraceArtifact` for optional runtime validation. This caused `pickle.dumps(artifact)` to serialize the `TreeSpec`, which now contained a `_MaskModWrapper` from the `BlockMask` pytree context. `_MaskModWrapper` holds the `mask_mod` closure (a local function from `get_causal_mask_mod`), which Python's pickle cannot serialize.

Neither PR broke things on its own — #2916 put `BlockMask` in the kwargs, and #3262 started serializing the kwargs' `TreeSpec`.

### Fix

Remove `user_inputs_spec` from `PrecompiledFxTraceArtifact`. The `mask_mod` is already compiled into standalone Inductor HOPs (`AOTCompiledArtifact`) baked into `serialized_gm` by the regional inductor pass, so it does not need to be preserved in the artifact metadata. `user_inputs_spec` is only used for an optional runtime validation check in `run_traced()` (disabled by default), so omitting it is safe.

## Test plan

- [x] New unit test `test_artifact_pickle_with_blockmask_treespec` — creates a `TracedResult` with an unpicklable BlockMask `TreeSpec`, verifies `from_traced_result` + `pickle.dumps` succeeds
- [x] `pytest torchtitan/experiments/graph_trainer/tests/test_precompile.py` — all 20 tests pass

### DIFF
--- a/torchtitan/experiments/graph_trainer/precompile.py
+++ b/torchtitan/experiments/graph_trainer/precompile.py
@@ -145,7 +145,11 @@ class PrecompiledFxTraceArtifact:
     output_subclass_layouts: dict[int, SubclassLayout]
     output_spec: pytree.TreeSpec
     tensor_input_indices: list[int]
-    user_inputs_spec: pytree.TreeSpec
+    # user_inputs_spec is intentionally omitted: it can contain
+    # FlexAttention _MaskModWrapper objects that are not picklable,
+    # and the mask_mod is already compiled into standalone Inductor
+    # HOPs (AOTCompiledArtifact) baked into serialized_gm. The spec
+    # is only used for optional runtime validation in run_traced().
     config_fingerprint: ConfigFingerprint = ConfigFingerprint("")
 
     @classmethod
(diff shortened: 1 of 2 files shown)


IvanKobzarev (approved):
Thanks. Verified on single node runs with precompile

SherlockNoMad (approved):

SherlockNoMad:
can you check if the disabled precompile test in test_bitwise_numercis.py can pass now?

PR #3430, opened 2026-05-26
[Varlen CP] Enable Varlen attention under Context Parallel + Full DTensor

fegin:
* #3431
* __->__ #3430

## TLDR

Similar to FlexCP that only support allgather-based CP, this PR extends CP support to Varlen. The major difference is that we rely on DTensor to allgather instead of manual allgather.

## Summary
Add varlen-attention support under CP with full_dtensor mode. Q stays sharded on the CP axis while K and V use DTensor Replicate placement so DTensor all-gathers them automatically.

### Key design
- `CPVarlenMetadata` in torchtitan/distributed/varlen_cp.py extends `VarlenMetadata` with `k_local_indices`  and a `from_global` builder that produces per-rank metadata. For each (doc, rank) contiguous-Q chunk, `seqlen_k = (global doc-relative offset at chunk end) + 1`.
- `cp_shard()` calls `CPVarlenMetadata.from_global` directly for varlen masks. Q/K/V tensors and BlockMask still use the upstream `_context_parallel_shard dispatcher`.
- `VarlenAttention.forward` re-packs K/V via `k_local_indices` when present.

Varlen+CP only works under full_dtensor because without DTensor Replicate placement, K and V are plain local tensors and nothing all-gathers them across CP. The legacy `apply_cp_to_forward` path raises with a message pointing users to parallelism.full_dtensor=True.

## Loss curve and performance
Loss curve verification and performance comparison are in the next PR

## Flow
from_global

### DIFF
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -415,6 +415,19 @@ def build_features_test_list() -> list[OverrideDefinitions]:
                 [
                     "--module llama3 --config llama3_debugmodel_varlen_attn",
                     "--parallelism.context_parallel_degree=2",
+                    "--parallelism.context_parallel_load_balancer=None",
+                ]
+            ],
+            "Full DTensor CP (Varlen, no load balancer)",
+            "full_dtensor_cp_varlen_no_lb",
+            ngpu=2,
+            skip_rocm_test=True,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_varlen_attn",
+                    "--parallelism.context_parallel_degree=2",
                 ]
             ],
             "Full DTensor CP (Varlen)",
--- a/tests/integration_tests/features.py
+++ b/tests/integration_tests/features.py
@@ -55,17 +55,19 @@ def _enable_full_dtensor(t: OverrideDefinitions) -> OverrideDefinitions:
     All features.py tests run under full_dtensor except PP-only variants
     (see ``_is_pp_only``) and CP + compile variants (upstream symint
     limitation); legacy non-full_dtensor coverage lives in models.py.
-    CP variants also switch to FlexAttention config because
-    full_dtensor + CP is not supported with SDPA / Varlen.
+    CP variants default to FlexAttention (full_dtensor + CP under SDPA is
+    not supported), but variants that already pick their own ``--module``
+    or ``--config`` (e.g. the varlen+CP tests) keep that selection.
     """
     new_args = []
     for variant in t.override_args:
         prefix: list[str] = []
         has_cp = any("context_parallel_degree" in arg for arg in variant)
         has_compile = any("compile.enable" in arg for arg in variant)
+        has_module = any("--module" in arg or "--config" in arg for arg in variant)
         if not _is_pp_only(variant, t.ngpu) and not (has_cp and has_compile):
             prefix.append("--parallelism.full_dtensor")
-        if has_cp:
+        if has_cp and not has_module:
             prefix.append("--module llama3 --config llama3_debugmodel_flex_attn")
         new_args.append(tuple(prefix) + tuple(variant))
     return dataclasses.replace(t, override_args=tuple(new_args))
@@ -407,6 +409,45 @@ def build_features_test_list() -> list[OverrideDefinitions]:
             "HSDP",
             "hsdp",
             ngpu=4,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_varlen_attn",
+                    "--parallelism.context_parallel_degree=2",
+                ]
+            ],
+            "Full DTensor CP (Varlen)",
+            "full_dtensor_cp_varlen",
+            ngpu=2,
+            skip_rocm_test=True,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_varlen_attn",
+                    "--parallelism.context_parallel_degree=2",
+                    "--parallelism.data_parallel_shard_degree=2",
+                ]
+            ],
+            "Full DTensor CP + FSDP (Varlen)",
+            "full_dtensor_cp_fsdp_varlen",
+            ngpu=4,
+            skip_rocm_test=True,
+        ),
+        OverrideDefinitions(
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_varlen_attn",
+                    "--parallelism.tensor_parallel_degree=2",
+                    "--parallelism.context_parallel_degree=2",
+                    "--parallelism.data_parallel_shard_degree=2",
+                ]
+            ],
+            "Full DTensor TP + CP + FSDP (Varlen)",
+            "full_dtensor_tp_cp_fsdp_varlen",
+            ngpu=8,
+            skip_rocm_test=True,
         ),
         OverrideDefinitions(
             [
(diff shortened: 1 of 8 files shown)


torchtitan/models/common/attention.py
```diff
@@ -55,10 +55,16 @@
 ]
 
 
-class VarlenMetadata(NamedTuple):
-    """
-    Cumulative sequence positions for queries and keys/values.
+@dataclass(frozen=True, eq=False)
```
tianyu-l, line 58: can we also have `kw_only` and `slots`, to be consistent with the rest of repo?

torchtitan/models/common/attention.py
```diff
@@ -106,25 +112,48 @@ def forward(
...
+                raise ValueError(
+                    "Varlen attention under CP only supports causal masking "
+                    f"(window_size=(-1, 0)); got window_size={self.window_size}."
+                )
+            xk_packed = xk_packed.index_select(
+                0, attention_masks.k_global_gather_indices
+            )
+            xv_packed = xv_packed.index_select(
+                0, attention_masks.k_global_gather_indices
+            )
```
tianyu-l, line 152: To clarify:
- this will keep only the relevant k,v, which technically is the minimal information to compute with thus having efficient gains over computing with full-lengths kv
- How critical is it? If we allow full Replicate k, v, is it possible to massage the metadata e.g. `cu_seq_q` to make it still work? My understanding is NO but just to confirm.

torchtitan/models/common/attention.py
```diff
@@ -106,25 +112,48 @@ def forward(
...
+        # Function-local import to avoid cyclic dep with varlen_cp.py.
+        from torchtitan.distributed.varlen_cp import CPVarlenMetadata
+
+        if isinstance(attention_masks, CPVarlenMetadata):
+            # CP metadata's right-aligned causal trick only works for causal masks.
+            if self.window_size != (-1, 0):
+                raise ValueError(
+                    "Varlen attention under CP only supports causal masking "
+                    f"(window_size=(-1, 0)); got window_size={self.window_size}."
+                )
```
tianyu-l, line 146: Any thoughts on other case? I would worry if this is too narrow.

On the other hand, for anything that's beyond causal LLM, e.g. VLM where on pictures we have full bidirectional masks, Varlen attention could not do the job and we have to use Flex attention.

Is there anything that's in between which would use e.g. blockwise bidirectional attention?

cc @francesco-bertolotti

tests/integration_tests/features.py
```diff
@@ -408,6 +410,58 @@ def build_features_test_list() -> list[OverrideDefinitions]:
...
+                    "--module llama3 --config llama3_debugmodel_varlen_attn",
+                    "--parallelism.tensor_parallel_degree=2",
+                    "--parallelism.context_parallel_degree=2",
+                    "--parallelism.data_parallel_shard_degree=2",
+                ]
+            ],
+            "Full DTensor TP + CP + FSDP (Varlen)",
+            "full_dtensor_tp_cp_fsdp_varlen",
+            ngpu=8,
+            skip_rocm_test=True,
```
tianyu-l, line 463: oh, does RoCM not support varlen attention? @akashveramd

tests/integration_tests/features.py
```diff
@@ -408,6 +410,58 @@ def build_features_test_list() -> list[OverrideDefinitions]:
...
+            [
+                [
+                    "--module llama3 --config llama3_debugmodel_varlen_attn",
+                    "--parallelism.tensor_parallel_degree=2",
+                    "--parallelism.context_parallel_degree=2",
+                    "--parallelism.data_parallel_shard_degree=2",
+                ]
+            ],
+            "Full DTensor TP + CP + FSDP (Varlen)",
+            "full_dtensor_tp_cp_fsdp_varlen",
```
tianyu-l, line 461: Can we only add one new test? features.py is already under pressure. Or consider piggyback on other existing tests.

torchtitan/distributed/varlen_cp.py
```diff
@@ -0,0 +1,248 @@
...
+import torch
+from torch.distributed.device_mesh import DeviceMesh
+from torch.distributed.tensor.experimental._context_parallel._load_balancer import (
+    _LoadBalancer,
+)
+
+from torchtitan.models.common.attention import VarlenMetadata
+
+
+@dataclass(frozen=True, eq=False)
```
tianyu-l, line 27: similar

torchtitan/distributed/varlen_cp.py
```diff
@@ -0,0 +1,248 @@
...
+    packed K can have unused gaps between per-segment visible regions when ``B > 1``;
+    ``k_global_gather_indices`` picks out just those regions. When a load balancer is active
+    the indices also compose with the per-batch inverse permutation so the gather hits
+    the correct entries in the rearranged K/V.
+    """
+
+    k_global_gather_indices: torch.Tensor
+
+    @classmethod
+    def from_global(
```
tianyu-l, line 46: maybe `from_replicate` to be consistent with DTensor / spmd_types language?

torchtitan/distributed/varlen_cp.py
```diff
@@ -0,0 +1,248 @@
...
+        # Fuse max_q / max_k / total_k into a single D2H transfer to keep only one D2H.
+        max_q, max_k, total_k = (
+            torch.stack([seqlen_q.max(), seqlen_k.max(), seqlen_k.sum()]).cpu().tolist()
+        )
+
+        # Flat gather index (length total_k) over original K coords; per segment
+        # covers [doc_global_start, last_global], built via repeat_interleave +
+        # arange offset.
+        bases = torch.repeat_interleave(doc_global_start, seqlen_k)
+        seg_starts_repeated = torch.repeat_interleave(cu_seq_k[:-1], seqlen_k)
```
tianyu-l, line 228: I learned that `repeat_interleave` will cause D2H sync when
- input is on GPU
- `output_size` is not given 

It should be fine here as `post_dataloading_process` happens on CPU. Maybe good to add a note somewhere.

torchtitan/distributed/varlen_cp.py
```diff
@@ -0,0 +1,248 @@
...
+        seqlen_k = last_global - doc_global_start + 1
+
+        cu_seq_q = torch.cat(
+            [torch.zeros(1, dtype=dtype, device=device), seqlen_q.cumsum(0).to(dtype)]
+        )
+        cu_seq_k = torch.cat(
+            [torch.zeros(1, dtype=dtype, device=device), seqlen_k.cumsum(0).to(dtype)]
+        )
+
+        # Fuse max_q / max_k / total_k into a single D2H transfer to keep only one D2H.
```
tianyu-l, line 219: should already be on CPU?

torchtitan/distributed/varlen_cp.py
tianyu-l, line 1: Didn't review this file too carefully, as the summary picture is already very helpful.

Do you consider landing the picture for the community as well? similar to @acisseJZhong 's https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/MOE_SHARDING.md

tianyu-l, line 1: Do you consider putting this into context_parallel.py? Sounds very related to me.

Or we can have
- `context_parallel/varlen.py`

tianyu-l (commented):
looks quite good, my main question is about whether the requirement on causal being too strong

tianyu-l:
Actually could you also show illustrations when load balancer is used?
- Is head tail load balancing per-document, or per sequence?
- If head & tail belong to the same document, would head & tail parts of q share the same kv or different kv?
- Does the causal trick still matter much (e.g. the tail would require seeing everything)?

PR #3433, opened 2026-05-27
[RFC] Collect grouped GEMM shapes and MoE runtime metrics

githubsgi:
## Summary

Adds an opt-in MoE debug/metrics subsystem that captures grouped-GEMM shapes and per-expert runtime load statistics for Mixture-of-Experts models. The goal is to make expert load imbalance, token routing, and grouped-GEMM problem sizes observable during real training without perturbing the hot path when disabled.

Fully **disabled by default**; enabled and configured through `--metrics.moe.*` flags.

## What it collects

- **Grouped-GEMM shapes** (`w1`/`w3`/`w2` `(M, K, N)` tuples) and the per-expert token counts (the runtime `M` dimension), with `layer_id` and `ep_rank`/`ep_size` attribution sourced from the EP mesh.
- **Load-imbalance metrics**, both intra-rank (min/max/mean/std/CV over a rank's local experts) and cross-rank/global (gathered on rank 0, per step/layer).
- **Dense GEMM shapes** for the non-MoE `nn.Linear` layers, captured into the run manifest.
- A **run manifest** (model/flavor, seq_len, batch sizes, parallel degrees, grouped + dense GEMM templates) rendered into TensorBoard as a markdown table.

## Sinks

`--metrics.moe.sinks` selects one or more of:

- `jsonl` / `csv` — raw per-record export.
- `histogram` — per-expert token-count histogram keyed by layer (compact for long runs).
- `tb` — logs into TorchTitan's existing TensorBoard run (requires `--metrics.enable_tensorboard`):
  - `moe_expert_load/layer_{L}` — per-layer rank/expert x step heatmap (+ z-score), per-step histograms and spread scalars.
  - `moe_layer_load/expert_{E}` — per-expert layer x step heatmap.
  - `moe_layer_spread/{tokens_max,tokens_min,tokens_range,tokens_mean}` — one line chart per statistic, overlaying every MoE layer.
  - `run_metadata` — the manifest table.

Offline plotting utilities (`agent_tooling/moe_metrics_plots/`) render the same heatmaps from exported artifacts.

## Config (`--metrics.moe.*`)

| Flag | Default | Purpose |
| --- | --- | --- |
| `enabled` | `False` | Master switch. |
| `sample_every` | `1` | Record once every N train steps. |
| `ranks` | `rank0` | `all`, `rank0`, or comma-separated rank list. |
| `sinks` | `[jsonl]` | Any of `jsonl`, `csv`, `histogram`, `tb`. |
| `collect_shapes` | `True` | Include grouped-GEMM shape tuples. |
| `collect_imbalance` | `True` | Emit per-(step, layer) imbalance stats. |
| `output_dir` | `moe_metrics` | Artifact dir under the dump folder. |
| `max_records_per_rank` | `100000` | Per-rank buffer cap before drops. |

The cross-rank gather (global imbalance + TB heatmap cube) only runs when `ranks == 'all'`, so it never introduces a collective on configs that didn't ask for one.

## Files

- `torchtitan/components/moe_metrics.py` — collector, records, sinks, imbalance accumulation, manifest.
- `torchtitan/components/metrics.py` — wires the collector into `MetricsProcessor`; shares the TensorBoard `log_dir` across ranks.
- `torchtitan/models/common/moe.py`, `token_dispatcher.py`, `decoder.py` — emit grouped-GEMM records from the routing/dispatch path.
- `torchtitan/trainer.py` — collector lifecycle (flush/close).
- `agent_tooling/moe_metrics_plots/` — offline heatmap plotting.
- `tests/unit_tests/test_moe_metrics.py`, `test_expert_parallel.py` — unit tests.
- `docs/rfc-moe-metrics.md` — design notes.

## Validation

Validated end-to-end on DeepSeek-V3 16B, 8x B200, TP=8 / EP=8, 100 steps:

- Clean run (0 errors, all 8 ranks tear down cleanly); loss 12.02 -> 6.44.
- `tb` + `histogram` sinks both produce artifacts; TensorBoard shows all three image groups plus the run-metadata table.
- Global imbalance CSV populated through step 100.

Rebased onto current `main`.

## Example output

Sample TensorBoard images from the DeepSeek-V3 16B run above (8x B200, TP=8 / EP=8, 100 steps).

Per-layer expert load (`moe_expert_load/layer_{L}`) -- rank/local-expert x step heatmap with a per-step z-score companion:

Per-expert token load

Per-expert layer load (`moe_layer_load/expert_{E}`) -- MoE-layer x step heatmap for a single expert row:

Per-layer token load

Per-layer spread curves (`moe_layer_spread/{tokens_max,...}`) -- one line per MoE layer vs step for a chosen statistic:

Per-layer spread curves

## Notes / open questions for reviewers

- Default `ranks=rank0` keeps the feature single-rank and collective-free unless the user opts into `all`. Right default for the heatmap cube, or should `tb` imply `all`?
- The `tb` sink opens its own `SummaryWriter` on the shared `log_dir` at close, producing a second event file. Alternative is to reuse the training `SummaryWriter`. Happy to switch if preferred.


githubsgi:
@tianyu-l , @wconstab , would appreciate feedback.

githubsgi:
@felipemello1, when you get a chance , please take a look.

felipemello1:
hey @githubsgi , thanks! We are finishing a sprint with deadline end of June. This metric/obsevability topic is dear to me, so i want to properly review it. I will try to do it this week! But, worst case i will do it next week after our sprint ends, if thats ok with you.

Thanks for posting the RFC!

githubsgi:
@felipemello1, looking forward to your comments.

felipemello1:
> @felipemello1, looking forward to your comments.

hey @githubsgi , i am on PTO. I will be back on Monday and review it. Sorry about that.

felipemello1:
hey @githubsgi , fyi i started the review. I will have it finished by tomorrow.

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+        offsets=offsets,
+    )
+    h = F.silu(torch._grouped_mm(x.bfloat16(), w1.bfloat16().transpose(-2, -1), offs=offsets))
+    h = h * torch._grouped_mm(x.bfloat16(), w3.bfloat16().transpose(-2, -1), offs=offsets)
+    return torch._grouped_mm(h, w2.bfloat16().transpose(-2, -1), offs=offsets).type_as(x)
+```
+
+`maybe_record_grouped_gemm` is a single `if not enabled: return` check on the
+hot path. When enabled it issues exactly one `.cpu().tolist()` on the
+already-on-device `num_tokens_per_expert` tensor; no extra D2H sync occurs
```
felipemello1, line 170: sorry, its not clear to me: does the .cpu require cpu<->gpu synchronization? I wonder if we can avoid that, somehow. Maybe some non blocking api, or doing it in bulk later.

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+`padded_tokens_per_expert` field to the `metadata` dict already returned by
+`*TokenDispatcher.dispatch(...)` (see
+[token_dispatcher.py](torchtitan/models/common/token_dispatcher.py)). All
+existing dispatchers fall back to `num_tokens_per_expert` when no padding is
+applied, so the field is purely additive.
+
+### 4.3 The Collector
+
+```python
+# torchtitan/components/moe_metrics.py
```
felipemello1, line 183: we should probably move it to torchtitan/observability. Currently the observability is scattered between components and observability, but eventually, anything related to that should be in the same `torchtitan/observability` folder.

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+### 4.4 Config
+
+Extend `torchtitan/config/configs.py`:
+
+```toml
+[metrics.moe]
+enabled        = false      # master switch
+sample_every   = 1          # record every N steps once enabled
+collect_shapes = true       # the grouped-GEMM (M,K,N) tuples
+collect_imbalance = true    # load-balance scalars to TB
```
felipemello1: is there a considerable extra cost to record those? if not, i would just delete the knobs and simplify, hardcoding them to true.

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+sinks          = ["jsonl"]  # any of {"jsonl","csv","tb"}
+output_dir     = "outputs/moe_metrics"
+max_records_per_rank = 100000
+ranks          = "all"      # "all" | "rank0" | "<comma-sep rank list>"
```
felipemello1: What does rank refers to here? global rank, local rank, dp rank, expert rank?

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+
+When disabled: one Python-level `if collector is None` per
+`_experts_forward` call (≈ tens of ns), no GPU sync.
+
+When enabled with default settings:
+
+- One `.cpu().tolist()` on a tensor of shape `(num_local_experts,)`
+  (typically ≤ 256 elements). Triggers one D2H, but `num_tokens_per_expert`
+  is already needed CPU-side for the existing `tokens_per_expert.add_` path,
+  so we reuse that synchronization point.
```
felipemello1, line 293: since this seems to be the major issue with this PR, i would just double check on this. Would torch profiler be the best way to confirm it?

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+- One dataclass construction + one `sink.write` per layer per step
+  (~O(N_layers) Python work per step).
+- Total measured overhead target: **< 0.5 %** of step time on Llama4-17Bx16E
+  at 8×H100.
```
felipemello1: nice. Do you see any reason for it to be different at larger scale and/or for other models? Do you know how <0.5% was measured? i.e. was it the avg time per step before/after?

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+    def is_enabled(self) -> bool: ...
+    def record(self, rec: GroupedGemmRecord) -> None: ...
+    def flush(self) -> None: ...           # invoked from train loop end-of-step
+    def aggregate(self) -> dict[str, float]: ...  # for TB scalars
+```
+
+Sinks (each optional, each behind a small protocol):
+
+- `JsonlSink(path)` — appends one record per line; cheapest, rank-local.
+- `CsvSink(path)` — flat schema, convenient for pandas / the existing
```
felipemello1: I am skeptical that we need a csv and json version.

docs/rfc-moe-metrics.md
```diff
@@ -0,0 +1,299 @@
...
+4. Announce in release notes; keep default off for one release before
+   flipping rank-0 JSONL on by default in dev configs.
+
+## 9. Open Questions
+
+- Should we also record the **router** logits histogram (top-k margins)?
+  Useful for collapse diagnostics, but ~`(num_experts,)` floats per token
+  is larger; leaning toward a separate opt-in flag.
+- Do we want a Parquet sink for production-scale runs, or is JSONL+offline
+  conversion sufficient? Inclined to start with JSONL.
```
felipemello1, line 368: starting with jsonl is fine. Ideally we/users could enable logging this to something like graphana/prometheus. Maybe tensorboard backend is good enough as well?

felipemello1:
@claude please review this PR

felipemello1 (commented):
I did a first pass on the rfc. It is very well written, thank you.

The moe_metrics.py is too dense and i need more time to digest it.

docs/rfc-moe-metrics.md
githubsgi: global .

docs/rfc-moe-metrics.md
githubsgi: Good catch. Yes, .to("cpu") on a device tensor forces a device->host sync, so the original code paid one sync per token/padded tensor (up to ~2N per flush for N buffered records).

The tensor being copied is the per-expert routing histogram - a small 1-D int tensor of length num_local_experts (typically 8-64) giving the number of tokens routed to each local expert. It's the raw signal we need for the grouped-GEMM shapes and the load-imbalance stats.

I addressed it two ways:

**1. Cheaper copy.**

(a) Deferred off the forward path. The copy no longer happens in the (activation-checkpointed) expert forward. During forward we only stash the on-device count tensors into a pending buffer; the D2H copy runs later in flush() at end-of-step.
(b) Batched into a single sync. flush() concatenates every pending record's token (and padded) count tensor into one 1-D tensor, does a single .to("cpu").tolist(), then splits the result host-side via a cursor. This collapses the ~2N syncs per flush into exactly 1.
(c) non_blocking wouldn't help. The payload is tiny (a few ints per record), so this is sync-latency bound, not bandwidth bound. We also immediately call .tolist(), which needs the data on host, so an async copy would just have to synchronize right after anyway. Batching is the win.

**2. Gated, so default training pays nothing.** The sync only occurs when the feature is enabled and the rank is selected and it's a sampled step (sample_every). Default runs never touch this path at all - this is the main reason the cost is a non-issue in practice.

Validated on a live 8-rank deepseek_v3_16b run (TP4/EP8, 20 steps): records_seen: 1040, records_dropped: 0, imbalance/manifest output correct, and the 23 unit tests pass.

docs/rfc-moe-metrics.md
githubsgi: Good point, and I agree MoE metrics are conceptually pure observability. I looked at moving it and want to lay out the tradeoff.

Today the observability package holds only the structured_logger subpackage, while all the swappable training building blocks (*metrics.py*, *checkpoint.py*, *dataloader.py*, *loss.py*, *optimizer.py*, *validate.py*, etc.) live in the components package. *moe_metrics.py* landed in components next to *metrics.py* mainly because it's tightly coupled to it: the collector is driven by MetricsProcessor (a Configurable component), and the grouped-GEMM hook and TB sink hang off that processor.

So the placement is a consequence of that coupling, not a deliberate "it's not observability" decision. The clean end state is to move both *metrics.py* and *moe_metrics.py* into observability together, so all metrics/TB logging lives next to structured logging. Moving *moe_metrics.py* alone doesn't make sense - since it depends on MetricsProcessor, it would just create a messy observability -> components back-import and split a tightly-coupled pair across two packages.

On the other hand, moving *metrics.py* is a larger, disruptive refactor: it touches the spec registry, every "from torchtitan.components.metrics import ..." callsite, and the tests - i.e. it churns a battle-tested path unrelated to this PR's feature. Realistically, relocating *metrics.py* to observability would be a bigger PR than this one, and would mix an unrelated refactor into this smaller feature review.

My preference is to keep *moe_metrics.py* in components alongside *metrics.py* for this PR, and do the *metrics.py* + *moe_metrics.py* move to observability as a dedicated follow-up that can be reviewed on its own.

docs/rfc-moe-metrics.md
githubsgi: Agreed, there's no meaningful extra cost. Both are cheap: shape recording is a couple of tensor `.shape` reads (pure Python, no device work), and imbalance is a small streaming accumulation over the per-expert token-count histogram that we're already copying to host. Neither adds a sync or a collective on the hot path - the only real cost (the D2H copy of the routing counts) is already paid once per sampled step and is now batched, and the whole thing is gated behind `enabled` + `ranks` + `sample_every` so default runs pay nothing.

So I've deleted both knobs and hardcoded the behavior on: `collect_shapes` and `collect_imbalance` are removed from the config, shape collection is unconditional, and imbalance is now gated only by `enabled`. Updated the tests and the RFC config block to match.

docs/rfc-moe-metrics.md
githubsgi: Good point — to clarify, they aren't two serializations of the same records. They're different granularities of the shape data, and both are opt-in through `metrics.moe.sinks`. I've switched the default to `histogram` so the common case is the compact one:

- **`histogram`** (*HistogramMoESink*, now the default): a compact CSV of `(layer_id, M, count)` plus a per-layer summary (min/max/mean/median). Storage is O(unique `(layer_id, M)` bins), so it stays bounded over arbitrarily long runs and never trips `max_records_per_rank`. This is what you want for a normal long training run where you just care about the distribution of per-expert token counts.

- **`jsonl`** (*JsonlSink*): one line per grouped-GEMM record, full fidelity, storage is O(records). This is only for debugging a specific shape sequence — when you need to see the exact per-step order of shapes rather than an aggregate. It's the one bounded by `max_records_per_rank`.

So it's not JSON-vs-CSV of the same data; it's "bounded summary histogram" (default) vs "full per-record trace for shape-sequence debugging" (opt-in). The invariant N/K dims live once in the manifest rather than on every line, which is why the histogram can afford to be so compact.

docs/rfc-moe-metrics.md
githubsgi: On the backends: the `tb` sink already covers a lot of the "dashboard" need without any new infra. It folds the offline heatmap post-processing into the training run and logs into TorchTitan's existing TensorBoard `log_dir`:

- per-expert load heatmaps per layer (`moe_expert_load/layer_{L}`),
- per-step spread scalars (`tokens_max/min/range/mean`, `max_over_mean`) that you can watch live to catch routing collapse,
- per-layer spread line charts overlaying all layers, and
- the run manifest (grouped-GEMM shape templates) as a markdown table.

So the intended story is: `histogram`/`jsonl` for local artifacts, `tb` when you want to eyeball it live in the TensorBoard you already have running — no extra service to stand up.

Prometheus/Grafana would be a nice follow-up PR for people already running that stack.

docs/rfc-moe-metrics.md
githubsgi: Great questions — the original "< 0.5%" was an analytical estimate, not a measured number, so I ran a proper A/B to back it up.

**How it's now measured.** Seed-matched (`--debug.seed 42`) 60-step runs on DeepSeek-V3 16B, 8xB200, TP=8/EP=8, identical config except the collector toggle (`sample_every=10`, `tb`+`histogram` sinks). I compared steady-state (step >= 20, discarding warmup) mean TPS reported per step:

| | steady-state mean TPS |
| --- | --- |
| metrics **off** | 1473 |
| metrics **on** | 1495 |

So yes — exactly your guess: avg per-step throughput before vs. after. The "on" run came out ~1.5% *faster*. As a bonus the per-step loss/grad_norm were **bitwise-identical** between the two runs (e.g. step 60: loss 6.73730, grad_norm 1.3461 in both), confirming the collector doesn't perturb the compute path.

Updated section 4.6 of RFC   with this measured A/B test also.

**Why it should stay negligible at larger scale / other models.** The design keeps the cost O(1) per (layer, step), independent of tokens or GPU count:

- The only device work is one `.cpu().tolist()` on a `(num_local_experts,)` tensor (<= 256 elems), and it **piggybacks on a D2H sync that already happens** for the aux-loss-free load-balancing (`tokens_per_expert.add_`) — so no *new* sync is introduced regardless of scale.
- Everything else is per-layer Python: one dataclass construction + one `sink.write`. That scales with `num_layers`, not with batch/seq/world size, so as GPUs, tokens, or model size grow, the fixed per-step Python cost shrinks as a fraction of step time.
- With `ranks=all` there is one `gather_object` of small per-rank dicts. That's the one piece that grows with world size, but it's off the hot path (flush/close) after completion of  training, the payload is tiny (per-(step,layer) stats, not tensors), and `sample_every` throttles it. If it ever mattered at very large EP, we'd bound it further or restrict to a subset of ranks.

Directionally the *fraction* should get **smaller**, not larger, at bigger scale, because step time grows while the collector's per-step Python cost stays fixed. The main model-dependent knob is `num_layers` (more MoE layers = a bit more per-step Python), so a very deep MoE would see marginally more, but still in the same negligible regime.

docs/rfc-moe-metrics.md
githubsgi: The end-to-end A/B TPS I posted is an illuminating measure: same config, --debug.seed 42, steady-state steps only, on/off TPS is within noise level, and loss/grad_norm are bitwise-identical between off and on (which also rules out any hidden numeric or control-flow issues)

For the disabled default path, the hook is just a global-boolean early return that Dynamo constant-folds out (0 graph breaks, op absent from the FX graph), which is below the profiler's reliable resolution.

githubsgi:
Thanks for the thorough review @claude. Pushed a follow-up commit (`90cda4a3`) that addresses the blocking and correctness items; point-by-point below.

### Blocking

**1. `FusedGroupedExperts` signature mismatch.** Fixed. `FusedGroupedExperts._experts_forward` now takes `dispatch_metadata: object | None = None`, matching the base `MoE.forward` call. Added a CUDA/accelerator-gated forward regression test (`test_fused_experts_forward_matches_base_grouped_experts` in *test_fused_swiglu_override.py*) that builds a `FusedGroupedExperts`, runs `_experts_forward(x, num_tokens_per_expert, None)` with the 3-arg signature, and checks fwd/bwd parity against the base within bf16 tolerance -- this would have caught the regression.

**2. `torch.compile(fullgraph=True)`.** Confirmed and handled. The validation run was eager (compile off), and you're right that the hook is not fullgraph-traceable when a collector is active. The disabled default path is now made compile-safe: `maybe_record_grouped_gemm` first checks a module-global `_COLLECTOR_INSTALLED` boolean and returns immediately, which Dynamo constant-folds away (verified 0 graph breaks, op absent from the FX graph on `aot_eager` and `inductor`). The enabled + model-compile combination is genuinely incompatible, so `MetricsProcessor.__init__` now raises a clear `ValueError` when `metrics.moe.enabled=true` is combined with `compile.enable=true` (`model` in `compile.components`), rather than failing opaquely at trace time. The RFC (section 4.6) documents this as eager-only.

### Design / architecture

**3. Reuse `tokens_per_expert_E`.** RFC section 4.5 ("Why Not Reuse MoE.tokens_per_expert_E") now explains this: that buffer is a running accumulator for aux-loss-free balancing (not an instantaneous per-step value), it double-counts under activation-checkpoint recompute, and it is at padded/microbatch granularity rather than the active-token granularity the imbalance metrics need. So the separate dispatch-metadata path is intentional.

**4. Model-file footprint.** The `getattr`-heavy metadata extraction now lives in *moe_metrics.py* (in the hook), not in model code. What remains in *moe.py* is the single `layer_id` thread plus one hook call, which as you note is essentially unavoidable for attribution.

**5. `gpt_oss` uncovered.** Intentional and now documented -- RFC section 4.2 and the module docstring note gpt_oss is not supported at this point.

### Correctness details

**6. Cross-rank gather under `ranks='all'`.** Confirmed: there is no early-return path where one rank skips the gather while others block. `_enabled` is uniform across ranks under `ranks='all'`, and the gather sends whole dicts, so divergent per-(step, layer) keys are fine (as you noted).

**7. Unconditional `broadcast_object_list`.** Fixed. `_resolve_tb_log_dir` now returns early when `config.moe.enabled` is False, so `MetricsProcessor.__init__` stays collective-free by default. `config.moe.enabled` is identical on every rank, so ranks skip or broadcast together.

**8. `top_k`/`layer_id == -1` sentinel.** Handled: the manifest renders `-1` as "unknown" (`_fmt_manifest_value`), and no plot path divides by `top_k` when it is `-1`.

### Minor
- *metrics.py* blank-line churn: reverted.
- `_extract_model_metadata`: kept the best-effort/manifest-only comment.
- Second `SummaryWriter` / `tb` implying `ranks='all'`: agree on both -- keeping the explicit opt-in for the WORLD collective, and noted the writer-reuse preference (overlaps a separate thread with @felipemello1).

### Testing / process
- Added the `FusedGroupedExperts` forward test (#1). For the compiled-forward smoke coverage, the disabled-path compile-safety is asserted by `test_hook_compiles_fullgraph_when_no_collector` (backend eager).
- `pre-commit` passes on all changed files.
- Numerics: with `metrics.moe.enabled=False`, loss and grad_norm are bitwise-identical vs. pre-PR (`--debug.seed 42`, steady-state steps); the end-to-end A/B TPS is within noise. This also rules out any hidden numeric or control-flow drift from the disabled hook sitting in the compiled/AC region.

Happy to attach a profiler trace of the enabled path if that's the form of overhead confirmation you'd prefer.

githubsgi:
@felipemello1, I  have gone through each of the  comments. Please take a look.

githubsgi:
@felipemello1 , could you please review ?

githubsgi:
@felipemello1 , a gentle reminder.

githubsgi:
@felipemello1 , @tianyu-l and @wconstab   FYI,   rebased and hence had to force push.  Please review.

tianyu-l:
could you take a look at https://github.com/pytorch/torchtitan/pull/4202 and see if it fits your workflow?

githubsgi:
@tianyu-l Thanks — I rebuilt this on top of #4202 rather than as a parallel path. 

Branch: https://github.com/githubsgi/torchtitan/tree/sg/moe-gemm-shapes

**What did and didn't carry over:** 

#4202's `tensor_logging` runtime, statistics kernels and hook plumbing do all the heavy lifting, and I'm reusing them unchanged. 

What it doesn't have is a **non-scalar sink** — `MetricLogger` can only emit scalars, so a length-`num_experts` load vector has to be reduced to summary statistics or fanned out to one series per element. For a 26-layer, 64-expert model the latter is 1664 series. (The 3D layer-by-step surfaces in the #4202 README were rendered by hand in matplotlib from a wandb export, not by the framework.)

So this adds the missing sink layer and the MoE views on top:

**`torchtitan/components/metrics.py`** — `log_histogram`, `log_histogram_bins`, `log_image`, `log_text` across `BaseLogger` / `TensorBoardLogger` / `WandBLogger` / `LoggerContainer`. Generic; nothing MoE-specific.

**`torchtitan/observability/tensor_logging/vector_metrics.py`** — a vector metric slot (one fixed-width FP32 row per metric, padded to a single dense slab so the cross-rank combine is one collective), plus `publish_vector_views()` emitting `.hist`, `.hist_by_element`, `.timeline`, `.heatmap` and a `.heatmap.legend`.

**`torchtitan/observability/moe_shapes.py`** — static grouped-GEMM shapes, as a JSON manifest and a Markdown table.

**Four details worth noting:**

1. `log_histogram_bins` wraps `add_histogram_raw`, not `add_histogram`. The latter bins a *sample population* by magnitude, which throws away expert identity — exactly the thing we want to see. Passing buckets directly keeps element index on the x axis. Moments are computed over the bin-index distribution so TB's summary stats stay coherent with the buckets.

2. `collect_with_vectors()` returns scalars and vectors from one reduction. Calling `collect()` and then reading the vectors would issue two collectives *and* the second would read zeroed buffers.

3. **Grouping.** Everything MoE is namespaced under a single `metrics_tensor_shapes/` prefix — histograms, heatmaps, timelines, the legend and the shape manifest — so it collapses into one TensorBoard section instead of scattering across the sidebar; `metrics_tensor_values/` keeps #4202's activation and gradient statistics. The namespace is a parameter on `publish_vector_views()`, not baked into the tag names, so history stays keyed by the bare metric name and renaming a section can't orphan an accumulated timeline. Two further groupings happen *within* that section: sibling layers matching on `layers.N.` collapse into one stacked heatmap rather than N separate images, and `.hist` / `.hist_by_element` / `.timeline` share the same x axis (element index) so a layer's histogram and its timeline can be read against each other. Worth noting TB only groups on the text before the *first* slash — deeper paths don't nest — so one prefix per feature is the most structure available. I put expert load under shapes rather than values because the load vector *is* the grouped-GEMM M dimension, but say the word if you'd rather see it split.

4. Images are colorized via a literal viridis LUT and ruled every eighth cell, keeping this PyTorch-only per the core dependency rule. The tradeoff is no real axis labels, which the legend text compensates for. If an optional matplotlib import that warns-and-skips is acceptable, I'd rather render properly labeled figures — happy to switch.

CPU unit tests cover element-wise reduction and buffer reset, the bin layout, sibling stacking, timeline bounding and ragged-group rejection. I've left this on my fork rather than opening a PR — happy to retarget `main` once #4202 lands, and to follow whatever the API settles into (I saw the `TensorLoggingState` → `TensorLogger` and `Config.build()` requests).

PR #3445, opened 2026-05-28
[Module][Flux] Convert Flux modules to fully config-based

fegin:
* #3447
* #3446
* __->__ #3445

**Summary**
As title, no logic change, just convert all the module classes to be config-based.

Verified the loss bit-wise parity with 100 steps.

### DIFF
--- a/torchtitan/models/flux/trainer.py
+++ b/torchtitan/models/flux/trainer.py
@@ -6,7 +6,7 @@
 
 import time
 from collections.abc import Iterable, Iterator
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 
 import torch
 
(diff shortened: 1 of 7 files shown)


torchtitan/models/flux/trainer.py
```diff
@@ -82,23 +82,30 @@ def __init__(self, config: Config):
...
+            .to(device=self.device, dtype=self._dtype)
+        )
+        self.t5_encoder = (
+            FluxEmbedder.Config(
+                version=config.encoder.t5_encoder,
+                random_init=config.encoder.random_init,
+            )
+            .build()
+            .to(device=self.device, dtype=self._dtype)
+        )
```
tianyu-l, line 112: nit: the config themselves should be in model_registry

tianyu-l (approved):
please make sure numerics don't change -- would be good to include a loss_compare test for flux (and MoE) in model CI.

PR #3447, opened 2026-05-28
[Full DTensor] Enable full_dtensor for all MoE models

fegin:
* #3503
* __->__ #3447

**TLDR**

Enable --training.full_dtensor for qwen3 / llama4 / gpt_oss / deepseek_v3

**Detail**

- Drop the EP gate and wire the sparse SPMD mesh for routed-expert FSDP: dense params use (dp_replicate, dp_shard, cp, tp), routed experts use (dp_replicate, efsdp, ep). MoE sharding_config is set unconditionally so EP-only (TP=1) still satisfies FSDP2's all-params-are-DTensor invariant.
- Fix CP + MoE: the experts return a 3D output so dp_shard and cp stay on separate dims instead of folding into a single _StridedShard that DTensor mishandles (correct forward but wrong gradients under CP). gpt_oss gets the same fix in its own GroupedExperts.
- Two smaller CP fixes: the router gate gathers TP before its F.linear (avoids a multi-axis _StridedShard wall under CP+SP), and the FSDP mesh-info is taken from FSDP2 instead of hand-built (avoids an overlapping-mesh crash under CP+EP).

**Verification**

Bit-identical loss and grad_norm vs the non-full_dtensor path across all parallelism combos including CP, for qwen3 / llama4 / gpt_oss. deepseek_v3 matches except a ~1e-5 difference on FSDP+TP+EP, where DTensor reductions sum in a different float32 order than legacy (MLA / node-limited routing) -- a reduction-order difference, not a correctness regression. 

CP combos use --no-seed-checkpoint because the loss_compare seed checkpoint hit a separate FSDP2 checkpoint-reshard bug, fixed upstream in [pytorch#186126](https://github.com/pytorch/pytorch/pull/186126/).

To get FSDP + CP + TP + EP work, we also need https://github.com/pytorch/pytorch/pull/186282


torchtitan/models/llama4/parallelize.py
```diff
@@ -136,13 +142,17 @@ def apply_fsdp(
     ep_degree: int = 1,
     edp_mesh: DeviceMesh | None = None,
     enable_symm_mem: bool = False,
+    dp_mesh_dims: "Any" = None,
+    edp_mesh_dims: "Any" = None,
```
tianyu-l: type should be `DataParallelMeshDims | None`?

torchtitan/distributed/full_dtensor.py
```diff
@@ -86,6 +80,17 @@ def _get_dp_mesh_axes(parallel_dims: ParallelDims) -> DataParallelMeshDims:
 
 
 _DENSE_SPMD_AXES = ["dp_replicate", "dp_shard", "cp", "tp"]
+_SPARSE_SPMD_AXES = ["dp_replicate", "efsdp", "ep"]
```
tianyu-l: To be honest, for perfect symmetry, we should call `efsdp` -> `edp_shard`. Maybe you don't want to create two fields always with the same meaning.

Since eventually we are going to remove `"fsdp"` for non-full_dtensor, I don't mind doing this to `"efsdp"` now / later.

torchtitan/distributed/full_dtensor.py
```diff
@@ -86,6 +80,17 @@ def _get_dp_mesh_axes(parallel_dims: ParallelDims) -> DataParallelMeshDims:
 
 
 _DENSE_SPMD_AXES = ["dp_replicate", "dp_shard", "cp", "tp"]
```
tianyu-l: Since we already have `DataParallelMeshDims`, I somehow feel we have enough metadata to feed [dp, cp] computation mesh to FSDP2, instead of [dp_replicate, dp_shard, cp].

The current design sounds half-way, i.e. keeping dp_replicate vs. other separation (impl details), but hiding [dp_shard, cp] flattening into `DataParallelMeshDims`.

tianyu-l: Fine for now, because I think @pianpwk 's change https://github.com/pytorch/pytorch/pull/181519 probably can work with that, despite the spmd annotation is on DP, not dp_replicate + dp_shard.

fegin: The current `DataParallelMeshDims` is unfortunately not enough. Assuming that TorchTitan only create `dp`, and `DataParallelMeshDims` has `dp_replicate` and `dp_shard`. The fully shard need to know how to split `dp` into `dp_replicate` and `dp_shard`. And `fully_shard` current logic assumes `dp_replicate` dim is the first dimension of the parameter's mesh.

A good design is to completely replace the current fully_shard frontend with the new one (flex_shard) that assume `dp` is going to be "split" into two, `dp_replicate` and `dp_shard`, and the later will be flattened into `cp`. We can of course making `fully_shard` to accept the `dp` only. However, it is unclear to me that if a temporary solution like this is enough or it worths to introduce such a complexity to an API that is not originally designed with the  idea in mind and we are going to move away.

torchtitan/distributed/full_dtensor.py
fegin: that's the only reason, I don't want to create another one `edp_shard` which is completely identical to `efsdp`.

tianyu-l (approved):

PR #3448, opened 2026-05-28
Fix model test failure in #3395: rope refactor

wwwjn:
Fixs needed in #3395.

This is by running model test on a 8GPU devmachine, as our main branch CI keeps failing.
- decoder.py: shared TP validation assumed every attention config has n_kv_heads, but DeepSeek V3 MLA only has n_heads, so we fall back to n_heads when KV heads are not separate.
 
- rope.py: Qwen3-VL MRoPE passes a pre-broadcast 4D RoPE cache where shape[0] is batch size, not max sequence length, so the position bounds check must only run for non-4D indexed caches. cc @shuhuayu

### DIFF
--- a/torchtitan/models/common/rope.py
+++ b/torchtitan/models/common/rope.py
@@ -408,6 +408,9 @@ def apply_rotary_emb_cos_sin(
     positions = _maybe_wrap_positions(positions, xq)
     head_dim = xq.shape[-1]
     if rope_cache.ndim != 4:
+        # Qwen3-VL MRoPE passes a pre-broadcast 4D RoPE cache where shape[0]
+        # is batch size, not max sequence length, so only validate positions
+        # for non-4D indexed caches.
         if positions is not None:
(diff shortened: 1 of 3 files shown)


torchtitan/models/common/rope.py
```diff
@@ -406,10 +406,10 @@ def apply_rotary_emb_cos_sin(
         positions: optional position indices
     """
     positions = _maybe_wrap_positions(positions, xq)
-    if positions is not None:
-        _maybe_check_max_pos(positions, max_valid_pos=rope_cache.shape[0] - 1)
     head_dim = xq.shape[-1]
     if rope_cache.ndim != 4:
+        if positions is not None:
+            _maybe_check_max_pos(positions, max_valid_pos=rope_cache.shape[0] - 1)
```
tianyu-l, line 414: add comments

tianyu-l (approved):

PR #3347, opened 2026-05-14
[rl] Add Batcher in RL Loop

wwwjn:
### Code Path
SFT:  DataLoader( pack (iterable_dataset))
RL: Batcher(pack (iterable_dataest))

### Key changes
- Batcher(Configurable) with microbatch_size and max_seq_length config
- batch(episodes) yields TrainBatch one at a time for gradient accumulation
- _episode_samples() converts Episode to per-token fields: input_ids, ref_logprobs, response_mask, advantages — all same length, all packed by pack()
- GRPOLoss operates on [B, L] tensors element-wise: (response_mask * advantages * logprobs).sum() / response_mask.sum()

## Batching algorithm
Batcher 
```
per_step_batches, global_valid_tokens = self._batcher.batch(
    episodes, dp_degree=self.trainer_dp_degree
)   
```

Inside batch():
1. Pack — _pack_episodes() converts episodes to per-token dicts (input_ids, ref_logprobs, response_mask,
advantages), calls shared pack() to greedily bin-pack into fixed [1, seq_length] rows util consumed **ALL** episodes
2. Pad or Truncate:
    - if `global_batch_size == -1`: Rounds the packed row count up to the nearest multiple of `local_batch_size * dp_degree`. 
    - If `global_batch_size != -1`, faithfully **truncate or pad** to user passed `global_batch_size`.
4. Calculate the `gradient_accumulation_steps`:
    - if `global_batch_size == -1`,  gradient_accumulation_steps = total_rows / (local_batch_size * dp_degree)
    - If `global_batch_size != -1`,  gradient_accumulation_steps  = global_batch_size / (local_batch_size * dp_degree)
5. Count global_valid_tokens = total response tokens (sum of response_mask across all rows)
6. Split arrange into [grad_accum_steps][dp_degree] of TrainBatch[B, L] where B = local_batch_size

Then Gradient accumulation loop 
```
for step_batches in per_step_batches:       # iterate grad_accum_steps
    fwd_bwd_metrics = trainer.forward_backward(
        step_batches,          # list[TrainBatch], one per DP rank
        global_valid_tokens,   # shared denominator across all steps
    )
```

PR #3449, opened 2026-05-28
For Monarch Future, use `await fut` instead of `fut.get()` when applicable

pzhan9:
when Monarch's `Future.get()` is called from within an active event loop, it blocks synchronously and does not yield control. As a result, it may degrade performance by preventing other tasks from running, and can potentially cause deadlocks if this future depends on them. [Monarch encourages](https://github.com/meta-pytorch/monarch/blob/017244c8a608f8ccf22f94c2f542cee00f1fecc4/python/monarch/_src/actor/future.py#L114-L115) its users to use 'await' instead in this situation.

This PR updates `grpo.py` to be consistent with this recommendation.

## Test Plan

I run the smoke test from the README to make sure the smoke test still works with both `torchmonarch==0.4.1` and `--pre monarch`. I also verified I could cancel the test in the middle of training by `ctrl+c`.

### DIFF
--- a/torchtitan/experiments/rl/grpo.py
+++ b/torchtitan/experiments/rl/grpo.py
@@ -534,12 +534,12 @@ class RLTrainer(Configurable):
 
         # Initial weight sync from trainer to generator
         with sl.log_trace_span("trainer_push_model_state_dict"):
-            self.trainer.push_model_state_dict.call().get()
+            await self.trainer.push_model_state_dict.call()
         with sl.log_trace_span("generator_pull_model_state_dict"):
-            self.generator.pull_model_state_dict.call(0).get()
+            await self.generator.pull_model_state_dict.call(0)
 
     @sl.log_trace_span("_collect_rollouts")
-    def _collect_rollouts(
+    async def _collect_rollouts(
         self,
         num_groups: int,
         step: int,
@@ -565,7 +565,7 @@ class RLTrainer(Configurable):
             for env in envs
         ]
         completions, generation_metrics = self._get_rank_0_value(
-            self.generator.generate.call(tokenized_prompts).get()
+            await self.generator.generate.call(tokenized_prompts)
         )
 
         trajectories: list[Trajectory] = []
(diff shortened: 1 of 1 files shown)


felipemello1:
thanks @pzhan9 , it looks great. Lets just wait https://github.com/pytorch/torchtitan/pull/3347 to land, since Jiani is almost done with it. Then maybe you can rebase and we land this as well?

torchtitan/experiments/rl/grpo.py
tianyu-l, line 1: two questions:
- are you saying even if the semantics is synchronous (not async), we should use `await` + async function call to implement, instead of `get`
- @felipemello1 did https://github.com/pytorch/torchtitan/pull/3238 to error out gracefully when `get()` is used. Are they still necessary after moving to `async` -- should we undo some changes for simplicity, if Monarch already handles them?

felipemello1: in my PR, i had to introduce this hack os await sleep(0), but its bad. It doesn't interrupt immediately -- a few steps still keep running. Doing the await, however, interrupts immediately.

felipemello1 (approved):
please rebase, since jiani's pr landed. I am approving to unblock, but Tianyu has a few questions.

Congrats on your first github PR in a while! hahaha

torchtitan/experiments/rl/grpo.py
pzhan9: @tianyu-l 

About 1, yes. `fut.get()` should be only used when there is no event loop available. See the doc string I just added to monarch [for details](https://github.com/meta-pytorch/monarch/commit/0695faf6a5217e262baef291f2ecb4d2f76053d9).

About 2, I removed `await sleep(0)` @felipemello1 mentioned above.  Other changes in that PR however I think is still necessary. Specifically, monarch's `proc_mesh.stop()` does not know and will not run any logic encapsulated in the `def close`s added by @felipemello1 . So you need something like `def stop` to clean up resource before monarch forcefully shutdown the mesh.

tianyu-l (approved):

PR #3450, opened 2026-05-28
[MoE] Using routing map instead of histc to count number_tokens_per_expert

acisseJZhong:
## What changed
 1. Router.forward() now builds and returns an explicit boolean routing_map_BLE with shape (B, L, E), using scatter_() (using a local map to wrap around scatter_ becuase scatter_ doesn't support DTensor yet) from topk_expert_ids_BLK.
 2. Per-expert token counts are computed from `routing_map.sum(...)` instead of `torch.histc(...)`.

## Test plan

Loss comparison: The loss is bit-wise identical across 100 stesps
```
  NCCL_NVLS_ENABLE=0 \
  CUDA_VISIBLE_DEVICES=4,5,6,7 \
  python scripts/loss_compare.py . main \
    --baseline-module='llama4' \
    --baseline-config='llama4_debugmodel' \
    --baseline-options='--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2
  --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2' \
    --baseline-ngpus=4 \
    --test-module='llama4' \
    --test-config='llama4_debugmodel' \
    --test-options='--parallelism.pipeline_parallel_degree 1 --parallelism.data_parallel_shard_degree 2
  --parallelism.tensor_parallel_degree 2 --parallelism.expert_parallel_degree 2' \
    --test-ngpus=4 \
    --no-seed-checkpoint \
    --output-folder=agent_space/loss_compare_current_vs_main_dp2tp2ep2
```

### DIFF
--- a/torchtitan/models/common/moe.py
+++ b/torchtitan/models/common/moe.py
@@ -28,23 +28,6 @@ from .token_dispatcher import DeepEPTokenDispatcher, LocalTokenDispatcher
 #   K = top-k, T = num tokens (B*L flattened),
 #   N = routed tokens (T*K), R = routed tokens assigned to local experts
 
-
-def _generate_routing_map_ble(
-    scores_BLE: torch.Tensor,
-    topk_expert_ids_BLK: torch.Tensor,
-) -> torch.Tensor:
-    """Build a boolean routing map from expert ids.
-
-    ``scores_BLE`` supplies the ``(B, L, E)`` shape and DTensor placement when
-    this helper is wrapped by ``local_map``.
-    """
-    return torch.zeros_like(scores_BLE, dtype=torch.bool).scatter_(
-        -1,
-        topk_expert_ids_BLK,
-        True,
-    )
-
-
 class GroupedExperts(Module):
     @dataclass(kw_only=True, slots=True)
     class Config(Module.Config):
@@ -109,7 +92,7 @@ class GroupedExperts(Module):
         x_BLD: torch.Tensor,
         topk_scores_BLK: torch.Tensor,
         topk_expert_ids_BLK: torch.Tensor,
-        routing_map_BLE: torch.Tensor,
+        num_tokens_per_expert_E: torch.Tensor,
     ) -> torch.Tensor:
         """Dispatch tokens to experts, compute, combine, and scatter_add.
 
@@ -119,18 +102,16 @@ class GroupedExperts(Module):
         """
         B, L, D = x_BLD.shape
         K = topk_scores_BLK.size(-1)
-        E = routing_map_BLE.size(-1)
         T = B * L
         x_TD = x_BLD.view(T, D)
         topk_scores_TK = topk_scores_BLK.view(T, K)
         topk_expert_ids_TK = topk_expert_ids_BLK.view(T, K)
-        routing_map_TE = routing_map_BLE.view(T, E)
         (
             routed_input_RD,
             num_tokens_per_expert_E,
             metadata,
         ) = self.token_dispatcher.dispatch(
-            x_TD, topk_scores_TK, topk_expert_ids_TK, routing_map_TE
+            x_TD, topk_scores_TK, topk_expert_ids_TK, num_tokens_per_expert_E
         )
         routed_output_RD = self._experts_forward(
             routed_input_RD, num_tokens_per_expert_E
@@ -244,7 +225,7 @@ class TokenChoiceTopKRouter(Module):
 
     def forward(
         self, x_BLD: torch.Tensor, expert_bias_E: torch.Tensor | None = None
-    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
         """
         Args:
             x_BLD: Input ``(B, L, D)``.
@@ -253,8 +234,7 @@ class TokenChoiceTopKRouter(Module):
         Returns:
             topk_scores_BLK: Routing scores ``(B, L, K)``.
             topk_expert_ids_BLK: Expert indices ``(B, L, K)``.
-            routing_map_BLE: Boolean routing map ``(B, L, E)``.
-            num_tokens_per_expert_E: Token counts per expert ``(E,)``.
+            scores_BLE: Full routing scores ``(B, L, E)``.
         """
         # Compute gate in float32 to help stability of expert load balancing.
         with torch.autocast(device_type=x_BLD.device.type, dtype=torch.float32):
@@ -297,35 +277,10 @@ class TokenChoiceTopKRouter(Module):
             topk_scores_BLK = topk_scores_BLK / denominator
         topk_scores_BLK = topk_scores_BLK * self.route_scale
 
-        # Build routing map with scatter. scatter_ does not support mixed
-        # local Tensor / DTensor arguments, so run the scatter on local tensors
-        # under local_map when router outputs are DTensors.
-        if isinstance(topk_expert_ids_BLK, DTensor):
-            assert isinstance(
-                scores_BLE, DTensor
-            ), "scores_BLE and topk_expert_ids_BLK should both be DTensors"
-            generate_routing_map = local_map(
-                _generate_routing_map_ble,
-                in_placements=(
-                    scores_BLE.placements,
-                    topk_expert_ids_BLK.placements,
-                ),
-                out_placements=(scores_BLE.placements,),
-                device_mesh=scores_BLE.device_mesh,
-            )
-        else:
-            generate_routing_map = _generate_routing_map_ble
-
-        routing_map_BLE = generate_routing_map(  # pyrefly: ignore [bad-argument-count]
-            scores_BLE, topk_expert_ids_BLK
-        )
-        num_tokens_per_expert_E = routing_map_BLE.sum(dim=(0, 1))
-
         return (
             topk_scores_BLK,
             topk_expert_ids_BLK,
-            routing_map_BLE,
-            num_tokens_per_expert_E,
+            scores_BLE,
         )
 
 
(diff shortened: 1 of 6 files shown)


torchtitan/models/common/moe.py
```diff
@@ -92,6 +109,7 @@ def forward(
         x_BLD: torch.Tensor,
         topk_scores_BLK: torch.Tensor,
         topk_expert_ids_BLK: torch.Tensor,
+        routing_map_BLE: torch.Tensor,
```
tianyu-l: In fact, I think you can just pass in `num_tokens_per_expert_E`

torchtitan/models/common/moe.py
```diff
@@ -274,15 +297,36 @@ def forward(
...
+                out_placements=(scores_BLE.placements,),
+                device_mesh=scores_BLE.device_mesh,
+            )
+        else:
+            generate_routing_map = _generate_routing_map_ble
+
+        routing_map_BLE = generate_routing_map(  # pyrefly: ignore [bad-argument-count]
+            scores_BLE, topk_expert_ids_BLK
         )
+        num_tokens_per_expert_E = routing_map_BLE.sum(dim=(0, 1))
```
tianyu-l: Let's
- return `scores_BLE` (which will be needed by https://github.com/pytorch/torchtitan/pull/3000/changes#diff-9c3e17fd22b0e249ef06806990a6a33b74d45f9ce009100d3a040e27ef074a3eR306 anyways), not `routing_map_BLE` or `num_tokens_per_expert_E`
- compute them in MoE fwd, not in router

torchtitan/models/common/moe.py
```diff
@@ -378,7 +423,9 @@ def forward(self, x_BLD: torch.Tensor) -> torch.Tensor:
         with torch.no_grad():
             self.tokens_per_expert_E.add_(num_tokens_per_expert_E)
 
-        out_TD = self.experts(x_BLD, topk_scores_BLK, topk_expert_ids_BLK)
+        out_TD = self.experts(
+            x_BLD, topk_scores_BLK, topk_expert_ids_BLK, routing_map_BLE
```
tianyu-l: let's pass `num_tokens_per_expert_E` directly.

torchtitan/models/common/moe_sharding.py
```diff
@@ -250,6 +250,7 @@ def set_moe_sharding_config(
             "x_BLD": experts_in_layout,
             "topk_scores_BLK": experts_in_layout,
             "topk_expert_ids_BLK": experts_in_layout,
+            "routing_map_BLE": experts_in_layout,
```
tianyu-l: Let's say we pass `num_tokens_per_expert_E` here. It will be used in `offsets` of `grouped_mm`. I think this field won't have gradient. Thus the gradient placement will have no effect. Let's specify None here and see if the code runs and numerics still match.

torchtitan/models/common/moe.py
```diff
@@ -274,15 +297,36 @@ def forward(
...
-            max=self.num_experts,
+        # Build routing map with scatter. scatter_ does not support mixed
+        # local Tensor / DTensor arguments, so run the scatter on local tensors
+        # under local_map when router outputs are DTensors.
+        if isinstance(topk_expert_ids_BLK, DTensor):
+            assert isinstance(
+                scores_BLE, DTensor
+            ), "scores_BLE and topk_expert_ids_BLK should both be DTensors"
+            generate_routing_map = local_map(
+                _generate_routing_map_ble,
```
tianyu-l: maybe define this function inline here

torchtitan/models/common/moe.py
```diff
@@ -274,15 +297,36 @@ def forward(
...
 
-        # group tokens together by expert indices from 0 to num_experts and pass that to experts forward
-        num_tokens_per_expert_E = torch.histc(
-            topk_expert_ids_BLK.view(-1),
-            bins=self.num_experts,
-            min=0,
-            max=self.num_experts,
+        # Build routing map with scatter. scatter_ does not support mixed
+        # local Tensor / DTensor arguments, so run the scatter on local tensors
+        # under local_map when router outputs are DTensors.
```
tianyu-l: leave a TODO to support sharding prop for scatter

torchtitan/models/common/moe_sharding.py
wwwjn: I see, the `num_tokens_per_expert_E` shouldn't have gradient because it's just routing information.

torchtitan/models/common/moe.py
```diff
@@ -109,7 +110,9 @@ def forward(
             routed_input_RD,
             num_tokens_per_expert_E,
             metadata,
-        ) = self.token_dispatcher.dispatch(x_TD, topk_scores_TK, topk_expert_ids_TK)
+        ) = self.token_dispatcher.dispatch(
+            x_TD, topk_scores_TK, topk_expert_ids_TK, num_tokens_per_expert_E
+        )
```
tianyu-l, line 123: why sending `num_tokens_per_expert_E` to dispatch() at all?

torchtitan/models/common/token_dispatcher.py
```diff
@@ -595,7 +604,9 @@ def dispatch(
         x_TD: torch.Tensor,
         topk_scores_TK: torch.Tensor,
         topk_expert_ids_TK: torch.Tensor,
+        num_tokens_per_expert_E: torch.Tensor,
```
tianyu-l: can we not pass it in?

torchtitan/models/common/moe.py
wwwjn: Oh i see, the issue. I was trying to remove `histc` usage in LocalTokenDispatcher by passing the `num_tokens_per_expert_E` as well, but acutally is not necessary. We can still use `histc` in LocalTokenDispatcher. Updated

tianyu-l: No. Where is num_tokens_per_expert_E / histc used in dispatcher?

torchtitan/models/common/token_dispatcher.py
```diff
@@ -88,13 +88,12 @@ def _local_reorder(
...
         """
-        # group tokens together by expert indices from 0 to num_experts and pass that to experts forward
         num_tokens_per_expert_E = torch.histc(
-            topk_expert_ids_TK.view(-1),
+            topk_expert_ids_TK.float(),
             bins=self.num_experts,
             min=0,
             max=self.num_experts,
-        )
+        ).to(torch.int64)
```
tianyu-l: maybe I'm missing something -- could you explain why we need to compute `num_tokens_per_expert_E` in dispatcher and return it AT ALL?

tianyu-l (changes requested):

torchtitan/models/common/token_dispatcher.py
wwwjn: mb, I've already moved the calculation in `MoE` level and we don't need to calculate in dispatcher level. I should clean up this part cleanly

But we still need to pass `num_tokens_per_expert_E` into dispatch() API and return from it. This is because for `num_tokens_per_expert_E` in DeepEPDispatcher path actually comes from DeepEP kernel's return value. If DeepEP is applied, we still need to return. To be more clear, the current implementation is: 
1. For AllToAllTokenDispatcher we pass `num_tokens_per_expert_E` into dispatch() call and return it as is.
2. For DeepEPTokenDispatcher, we use the DeepEP returned `num_tokens_per_expert_E`, instead of what we passed into dispatch().

An alternative is don't pass `num_tokens_per_expert_E` into dispatch() at all,  just drop the`num_tokens_per_expert_E`  from DeepEP and use the value calculated in `MoE`.  Can we take the alternative way? Is it possible that DeepEP returns a different value than what we calculated from routing_map?

torchtitan/models/common/moe.py
```diff
@@ -363,13 +363,50 @@ def forward(self, x_BLD: torch.Tensor) -> torch.Tensor:
...
-            num_tokens_per_expert_E,
+            scores_BLE,
         ) = self.router(x_BLD, self.expert_bias_E)
 
+        # Build routing map with scatter. scatter_ does not support mixed
+        # local Tensor / DTensor arguments, so run the scatter on local tensors
+        # under local_map when router outputs are DTensors.
+        # TODO: Remove this local_map workaround once DTensor sharding
+        # propagation supports scatter with mixed Tensor / DTensor arguments.
+        def generate_routing_map_ble(
```
tianyu-l:
```suggestion
        def _generate_routing_map(
```

torchtitan/models/common/token_dispatcher.py
tianyu-l: change `tokens_per_expert_E` variable name
same for DeepEP

tianyu-l, line 326: add a TODO that we might want to use this as buffer to compute `expert_bias_E`, but HybridEP token dropping is not clear.

tianyu-l (approved):

torchtitan/models/common/moe.py
```diff
@@ -53,45 +54,48 @@ def __init__(self, config: Config):
     def _experts_forward(
         self,
         x_RD: torch.Tensor,
-        num_tokens_per_expert_E: torch.Tensor,
+        num_global_tokens_per_local_expert_e: torch.Tensor,
```
wwwjn: @tianyu-l  Can I get a final review of the rename in `_experts_forward` function? Basically I renamed `E` -> `e`

tianyu-l: Good point. I discussed with @acisseJZhong before.

In grouped experts:
- if EP is used, then this notation is correct (e is sharded version of E)
- if TP is used, then we should do `w1_EfD` instead because TP shards on different things
- maybe we should use x_rD instead of x_RD depending on batch/seq dim sharding

Given all these complications, let's just use capital letters (such as `E`) and `num_tokens_per_expert_E`.

tianyu-l: good to add a NOTE on notation

torchtitan/models/common/moe_sharding.py
```diff
@@ -258,6 +261,9 @@ def set_moe_sharding_config(
                 experts_in_grad_layout,
                 experts_in_grad_layout,
                 experts_in_grad_layout,
+                # num_local_tokens_per_expert_E is routing metadata, but it is
+                # still a DTensor input to local_map and must have placements.
+                _tokens_per_expert_placement(enable_ep=enable_ep),
```
wwwjn, line 266: I have to put a placement here, because local_map doesn't take None as a input, cc @tianyu-l

tianyu-l: it takes?
https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/experimental/_func_map.py#L20-L29

torchtitan/models/common/moe.py
wwwjn: Sounds good, 

> let's just use capital letters (such as E) and `num_tokens_per_expert_E`.

I updated `num_tokens_per_expert_E` to be the parameter name of _expert_forwarrd(), but in the callsite, I still pass `num_tokens_per_expert_E=num_global_tokens_per_local_expert_e` to distinguish: 
```
(
    routed_input_RD,
    num_global_tokens_per_local_expert_e,
    metadata,
) = self.token_dispatcher.dispatch(
    x_TD, topk_scores_TK, topk_expert_ids_TK, num_local_tokens_per_expert_E
)
routed_output_RD = self._experts_forward(
    routed_input_RD, num_global_tokens_per_local_expert_e
)
```

tianyu-l: that sounds good

PR #3451, opened 2026-05-28
Refactor gypo.py into trainer.py and train.py

pzhan9:
`grpo.py` logically contains two parts:
* The runner part that parses the config, build Trainer, and kick off training
* The actual Trainer.

This is similar to how pretraining is set up, and provides a cleaner structure. This PR does the refactoring.

### DIFF
--- a/torchtitan/experiments/rl/train.py
+++ b/torchtitan/experiments/rl/train.py
@@ -1,0 +1,208 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+RL training loop using Monarch Actors.
+
+This demonstrates:
+1. Distributed actor architecture with VLLMGenerator (vLLM) and PolicyTrainer (TorchTitan)
+   running on separate GPU meshes
+2. Weight synchronization across meshes via TorchStore: the trainer publishes its
+   model state dict and the generator pulls it into its own parallelism layout,
+   with direct GPU-to-GPU RDMA transfer when available
+3. Envs driven rollouts; reward and advantage computation live inline
+   in the controller.
+
+Command to run:
+python3 torchtitan/experiments/rl/train.py \
+    --module rl --config rl_grpo_qwen3_0_6b \
+    --hf_assets_path=<path_to_model_checkpoint>
+"""
+
+import asyncio
+import logging
+import os
+from collections.abc import Callable
+from dataclasses import dataclass
+
+# must run before torch import. Set it as early as possible to avoid other
+# imports transitively importing torch.
+os.environ.setdefault("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:True")
+
+from monarch.actor import HostMesh, ProcMesh, this_host
+
+from torchtitan.config import ConfigManager, ParallelismConfig
+from torchtitan.experiments.rl.trainer import RLTrainer
+from torchtitan.observability import structured_logger as sl
+
+
+logger = logging.getLogger(__name__)
+
+
+class PerHostProvisioner:
+    """Allocates non-overlapping GPU ranges within a single host.
+
+    On the same host, the trainer and generator run on separate GPU
+    meshes (e.g. GPUs 0-3 for training, GPUs 4-7 for generation). Each
+    call to `allocate(n)` reserves the next *n* GPUs and returns a
+    bootstrap callable that sets `CUDA_VISIBLE_DEVICES` before CUDA
+    initializes in the spawned process, ensuring each mesh only sees its
+    own devices.
+    """
+
+    def __init__(self, total_gpus: int = 8):
+        self.total_gpus = total_gpus
+        self.next_gpu = 0
+
+    @property
+    def available(self) -> int:
+        return self.total_gpus - self.next_gpu
+
+    def allocate(self, num_gpus: int) -> Callable[[], None]:
+        if num_gpus > self.available:
+            raise RuntimeError(
+                f"Requested {num_gpus} GPUs but only {self.available} "
+                f"available (total={self.total_gpus}, allocated={self.next_gpu})"
+            )
+        gpu_ids = list(range(self.next_gpu, self.next_gpu + num_gpus))
+        self.next_gpu += num_gpus
+
+        def _bootstrap():
+            os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(str(g) for g in gpu_ids)
+            # TODO: Remove once Monarch/PyTorch fixes concurrent import during unpickling.
+            import torch  # noqa: F401
+
+        return _bootstrap
+
+
+@dataclass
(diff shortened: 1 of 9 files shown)


torchtitan/experiments/rl/grpo.py
```diff
@@ -891,8 +881,15 @@ async def train(self):
...
-    config = ConfigManager().parse_args()
+async def train_main(
+    *,
+    config: "RLTrainer.Config | None" = None,
+    trainer_host_mesh=None,
+    generator_host_mesh=None,
+    gpus_per_node: int | None = None,
+):
+    if config is None:
+        config = ConfigManager().parse_args()
```
tianyu-l: My mental model is that `RLTrainer.Config` should be self-contained for the entire training job.

What's the motivation of separating mesh config from trainer config?

torchtitan/experiments/rl/grpo.py
```diff
@@ -407,9 +407,8 @@ def _collate_episodes(episodes: list[Episode]) -> TrainingBatch:
     async def setup_async(
         self,
         *,
-        host_mesh=None,
-        trainer_nodes: int | None = None,
-        generator_nodes: int | None = None,
+        trainer_host_mesh=None,
+        generator_host_mesh=None,
```
tianyu-l: Can we make this required even for single-node? Ideally single-node vs. multi-node should be transparent.

torchtitan/experiments/rl/grpo.py
```diff
@@ -453,23 +451,20 @@ async def setup_async(
...
         # shrink this span to a single call.
         with sl.log_trace_span("mesh_spawn"):
-            if host_mesh is not None:
-                # Multi-node mode: dedicate whole nodes to trainer vs generator
-                if (
-                    trainer_nodes is None
-                    or generator_nodes is None
-                    or gpus_per_node is None
-                ):
+            if self._multi_node:
```
tianyu-l: hoping to remove this if-else

torchtitan/experiments/rl/grpo.py
```diff
@@ -891,8 +881,15 @@ async def train(self):
         logger.info("=" * 60)
 
 
-async def main():
-    config = ConfigManager().parse_args()
+async def train_main(
```
tianyu-l: Do you think this is trying to do something we did in pretraining?
- train.py parses the config, build Trainer, and kick off training https://github.com/pytorch/torchtitan/blob/main/torchtitan/train.py
- trainer.py contains the actual Trainer logic

I feel we should also separate this file into two.

torchtitan/experiments/rl/train.py
tianyu-l, line 1: I don't think this is associated with GRPO. Can we call it train.py, and the other file trainer.py, to mimic what torchtitan pretraining is organized.

torchtitan/experiments/rl/grpo.py
tianyu-l: This seems outdated. Could you help update it to mention torchstore?

torchtitan/experiments/rl/grpo.py
```diff
@@ -167,743 +76,94 @@ def _bootstrap():
...
-                PolicyTrainer,
-                config.trainer,
-                model_spec=config.model_spec,
-                hf_assets_path=config.hf_assets_path,
-                generator_dtype=config.generator.model_dtype,
-                compile_config=config.compile,
-                output_dir=config.dump_folder,
-            )
+def create_proc_mesh(
+    config: RLTrainer.Config,
```
tianyu-l: for encapsulation, let's only send in `trainer_world_size` and `generator_world_size`

torchtitan/experiments/rl/grpo.py
```diff
@@ -167,743 +76,94 @@ def _bootstrap():
...
+    gpus_per_node: int
 
-        generator: VLLMGenerator.Config = field(default_factory=VLLMGenerator.Config)
-        """VLLMGenerator actor configuration (vLLM engine, sampling)."""
 
-        metrics: m.MetricsProcessor.Config = field(
-            default_factory=m.MetricsProcessor.Config
-        )
+@dataclass
+class ProcMeshes:
```
tianyu-l: This seems very thin and used within this file, maybe don't need this abstraction right now?

torchtitan/experiments/rl/grpo.py
```diff
@@ -167,743 +76,94 @@ def _bootstrap():
...
-                "trainer",
-                PolicyTrainer,
-                config.trainer,
-                model_spec=config.model_spec,
-                hf_assets_path=config.hf_assets_path,
-                generator_dtype=config.generator.model_dtype,
-                compile_config=config.compile,
-                output_dir=config.dump_folder,
-            )
+def create_proc_mesh(
```
tianyu-l: please add docstring for this method, especially the different behavior between `host_meshes` being None or not.

torchtitan/experiments/rl/grpo.py
```diff
@@ -23,112 +23,21 @@
...
-                    (torch.abs(ratio - clipped_ratio) > 1e-6).float() * loss_mask
-                ).sum()
-                / loss_denominator,
-            }
-
-        return loss, metrics
+logger = logging.getLogger(__name__)
 
 
 class Provisioner:
```
tianyu-l: Can we rename this to `PerHostProvisioner` or `IntraNodeProvisioner`? Ideally a real `Provisioner` should be in charge of the whole world.

tianyu-l (approved):
nice

torchtitan/experiments/rl/train.py
pzhan9: I renamed the other file to `trainer.py`. But I want to point out there is another file named as trainer. Not sure if that is a concern.

torchtitan/experiments/rl/actors/trainer.py

tianyu-l: Thanks for reminding. I think it's fine for now as they live in different folder. Let's have discussion later if we want to rename them further.

torchtitan/experiments/rl/train.py
```diff
@@ -0,0 +1,208 @@
...
+    assert isinstance(config, RLTrainer.Config)
+    sl.init_structured_logger(
+        source="rl_controller",
+        output_dir=config.dump_folder,
+        rank=0,
+        enable=config.trainer.debug.enable_structured_logging,
+    )
+    sl.log_trace_instant("structured_logger_started")
+
+    rl_trainer: RLTrainer = config.build()
```
pzhan9, line 187: note: technically we can have different `RLTrainer`s, e.g. the one we have now is `SyncTrainer`, and later add an `AsyncTrainer`. Then we can use config to decide which one to use.

tianyu-l: I think SyncTrainer makes little sense as production runs if we only support non-collocated.

My wish is that we support async only, but with a parameter to control the lag, so sync debugging is possible (instead of maintaining two trainers).

cc @felipemello1

felipemello1: agreed

PR #3456, opened 2026-05-29
Fix LoRA freezing for non-linear modules

### LINKED ISSUE #3438: `LoRAConverter` silently leaves non-Linear parameters trainable
### Summary

`torchtitan.components.lora.LoRAConverter` only walks `Linear.Config`
nodes in the model config tree — target Linears become `LoRALinear`
(base frozen), non-target Linears get wrapped with `FrozenConfig`.
Every **non-Linear** trainable parameter (RMSNorm `weight`, embedding
`weight`) keeps the default `requires_grad=True`, so the default
optimizer (which filters by `if p.requires_grad`) starts training them
alongside the LoRA adapters.

The API is named `LoRAConverter` and is the only obvious LoRA on-ramp,
so users (and the project's own integration test
`llama3_debugmodel_float8_emulate_lora`) end up training norms and
embeddings without realising.

### Why it matters

For each step, the leaked non-Linear params:

1. Get **optimizer steps + Adam state** — wasted memory and compute.
2. Get **FSDP reduce-scatter** for their (effectively zero) gradient
   buckets — wasted communication.
3. **Diverge from the pretrained values** silently — there's no
   user-visible signal that "LoRA training" is also drifting
   normalisation statistics and token embeddings.

For LoRA-only checkpoint flows (downstream code that stores just
`.lora_a.*` / `.lora_b.*`), the accidentally-trained
norms/embeddings **aren't saved**, so every resume reloads the HF
base and wipes the accumulated updates — the wasted training was
strictly negative.

### Reproduction

```python
"""Uses the same flavor + target_modules as torchtitan's own LoRA
integration test ``llama3_debugmodel_float8_emulate_lora`` in
``torchtitan/models/llama3/config_registry.py``.
"""
import torch
from torchtitan.components.lora import LoRAConverter
from torchtitan.models.llama3 import llama3_configs

cfg = llama3_configs["debugmodel"](attn_backend="flex")
LoRAConverter.Config(
    rank=8, alpha=16.0, target_modules=["wq", "wkv", "wo"]
).build().convert(cfg)

with torch.device("meta"):
    model = cfg.build()
model.to_empty(device="cpu")

non_lora_trainable = [
    (n, tuple(p.shape))
    for n, p in model.named_parameters()
    if p.requires_grad and ".lora_a." not in n and ".lora_b." not in n
]
lora_trainable = sum(
    1 for n, _ in model.named_parameters()
    if ".lora_a." in n or ".lora_b." in n
)

print(f"LoRA tensors (expected trainable):     {lora_trainable}")
print(f"Non-LoRA params still trainable (BUG): {len(non_lora_trainable)}")
for name, shape in non_lora_trainable:
    print(f"  {name}  {shape}")
```

#### Output

```
LoRA tensors (expected trainable):     48
Non-LoRA params still trainable (BUG): 14

  tok_embeddings.weight  (2048, 256)
  layers.0.attention_norm.weight  (256,)
  layers.0.ffn_norm.weight  (256,)
  layers.1.attention_norm.weight  (256,)
  layers.1.ffn_norm.weight  (256,)
  layers.2.attention_norm.weight  (256,)
  layers.2.ffn_norm.weight  (256,)
  layers.3.attention_norm.weight  (256,)
  layers.3.ffn_norm.weight  (256,)
  layers.4.attention_norm.weight  (256,)
  layers.4.ffn_norm.weight  (256,)
  layers.5.attention_norm.weight  (256,)
  layers.5.ffn_norm.weight  (256,)
  norm.weight  (256,)
```

Every layer's two RMSNorm `weight`s plus the top-level `norm.weight`
plus the (2048×256) `tok_embeddings.weight` are all marked
`requires_grad=True`. Production flavors leak proportionally more.

### Expected behaviour

After `LoRAConverter` runs and the model is built, **only**
`.lora_a.*` / `.lora_b.*` parameters should be trainable.

### Environment

- torch: `2.12.0+cu126`
- Python: 3.12

yinli-systems:
Fixes #3438.

## Summary
- add an optional `recurse` mode to `Configurable.Config.traverse` while preserving the default stop-at-match behavior
- make `FrozenConfig` a `Module.Config` wrapper so frozen configs still satisfy module-config type checks
- freeze non-Linear leaf module configs during LoRA conversion, preventing embeddings/norms from remaining trainable
- extend LoRA coverage to assert that only adapter parameters are trainable

## Test plan
- `python3 -m compileall -q torchtitan/config/configurable.py torchtitan/components/lora.py tests/unit_tests/test_lora.py tests/unit_tests/test_configurable.py`
- `git diff --check`
- `.venv/bin/python -m ruff check torchtitan/config/configurable.py torchtitan/components/lora.py tests/unit_tests/test_lora.py tests/unit_tests/test_configurable.py`
- `PYTHONPATH=. /tmp/torchtitan-py311-nightly/bin/python -m pytest -q tests/unit_tests/test_lora.py tests/unit_tests/test_configurable.py`


torchtitan/components/lora.py
```diff
@@ -122,7 +123,7 @@ def __getattr__(self, name: str):
         return getattr(inner, name)
 
     def __setattr__(self, name: str, value) -> None:
-        if name in type(self).__slots__:
+        if name in {f.name for f in fields(type(self))}:
```
tianyu-l: curious what's the motivation of this change

torchtitan/components/lora.py
```diff
@@ -105,7 +106,7 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
 
 
 @dataclass(kw_only=True, slots=True)
-class FrozenConfig(Configurable.Config):
+class FrozenConfig(Module.Config):
```
tianyu-l: Do we need this change? I understand that `FrozenConfig` always wraps a `Module`, but `Module` has many metadata which is probably not needed by this class.

torchtitan/components/lora.py
tianyu-l: probably should make this `Module.Config`?

torchtitan/components/lora.py
```diff
@@ -214,6 +228,20 @@ def convert(self, model_config) -> None:
             else:
                 setattr(parent, attr, new_cfg)
 
+        # Second pass: freeze non-Linear leaf modules such as embeddings and
```
tianyu-l: Why do we need two passes? I had thought one pass should suffice.

yinli-systems:
Thanks for the questions.

The motivation for this PR is to close a LoRA correctness gap: the converter currently only replaces/freezes Linear.Config nodes. If target_modules is a subset, non-Linear leaf modules that own parameters, such as embeddings and norms, can still remain trainable, so the run is no longer adapter-only training.

On the specific points:

- The fields(type(self)) change was only needed because I temporarily made FrozenConfig inherit Module.Config. With slotted dataclass inheritance, __slots__ only contains slots declared on that class and can miss inherited dataclass fields such as param_init/sharding_config, while fields() includes inherited dataclass fields. That said, I agree this complexity is unnecessary if FrozenConfig does not inherit Module.Config.

- I agree FrozenConfig does not need Module.Config metadata. I will revert it back to Configurable.Config and make inner: Module.Config instead, which captures the actual invariant without adding unused Module fields to the wrapper.

- The two passes were because the original Linear.Config traversal handles only Linear nodes, while the second pass was added to freeze non-Linear parameter-owning leaf modules without freezing composite modules that contain LoRA adapters. But I agree this can be cleaner. I will collapse this into one recurse=True traversal over Module.Config, handling Linear configs and selected parameter-owning non-Linear leaves in the same pass.

I’ll push a follow-up that narrows the wrapper type, removes the unnecessary Module.Config inheritance, and simplifies the traversal.
________________________________

torchtitan/components/lora.py
```diff
@@ -188,26 +196,62 @@ def _make_lora_config(self, cfg: Linear.Config):
...
+        return isinstance(
+            cfg,
+            (
+                Conv2d.Config,
+                Embedding.Config,
+                GroupNorm.Config,
+                LayerNorm.Config,
+                RMSNorm.Config,
+            ),
+        )
```
tianyu-l: This only cover the existing cases today, which won't extend well when a model has new types of submodules.

In fact, I don't think we have to test if something is parametrized at all. Can we just apply `FrozenConfig` wrapping regardless?

torchtitan/components/lora.py
```diff
@@ -188,26 +196,62 @@ def _make_lora_config(self, cfg: Linear.Config):
...
+                )
+                if is_target:
+                    new_cfg = self._make_lora_config(cfg)
+                    matched.add(last_segment)
+                else:
+                    new_cfg = FrozenConfig(inner=cfg)
+            elif (
+                self._is_parameterized_non_linear_config(cfg)
+                and not self._has_nested_module_config(cfg)
+            ):
```
tianyu-l: I don't think this is necessary / sufficient either.

What if a Module has submodule and trainable parameters -- would you skip setting `requires_grad=False`?

I think we just need to have two cases:
- if Linear.Config and matching target, do lora wrapping
- else do FrozenConfig wrapping

WDYT?

Overall there might be the tricky case where a Module has Linear submodule, and a trainable parameter, which we need to make sure traverse order and wrapping order is compatible.

yinli-systems: Thanks, this makes sense. I updated the PR to use the two-case logic:

- matching `Linear.Config` -> LoRA wrapping
- every other `Module.Config` -> `FrozenConfig` wrapping

To handle the composite-module case, `FrozenConfig` now freezes only direct parameters via `parameters(recurse=False)`. Traversal still recurses into the original config tree, so child configs are converted/wrapped separately; this avoids recursively freezing nested LoRA adapters while still freezing direct parameters on a parent module that owns both parameters and submodules.

I also added a regression test with a composite module that has a direct `Parameter` plus a child `Linear` target, asserting the direct param and base weight are frozen and only LoRA adapter weights remain trainable.

Updated commit: 9d2553033

torchtitan/components/lora.py
```diff
@@ -122,14 +122,19 @@ def __getattr__(self, name: str):
...
+        if name in {f.name for f in fields(type(self))}:
             object.__setattr__(self, name, value)
         else:
             setattr(self.inner, name, value)
 
     def build(self, **kwargs):
+        if self.param_init is not None:
+            self.inner.param_init = self.param_init
+        if self.sharding_config is not None:
+            self.inner.sharding_config = self.sharding_config
```
tianyu-l: I'm assuming you are adding this because the order follows:
- create a non-lora model config, which doesn't have sharding config. E.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/__init__.py#L213
- call lora config converter to freeze non-lora parameters. E.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/config_registry.py#L118
- set sharding_config before actual build, e.g. https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/common/decoder_sharding.py#L239-L241

However, 3 should already be covered by the `__setattr__` override above. Thus
- it is not necessary to do it here one more time
- actually I feel this could be wrong, because `self.sharding_config` is not really set due to the setattr override (unless the override itself has bugs)

torchtitan/components/lora.py
```diff
@@ -105,14 +105,14 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
 
 
 @dataclass(kw_only=True, slots=True)
-class FrozenConfig(Configurable.Config):
-    """Config wrapper that freezes all parameters at build time.
+class FrozenConfig(Module.Config):
```
tianyu-l: For the reason mentioned below, this should still not be a `Module.Config` subclass?

yh8899:
What should we do if we have nn.Parameter？

tianyu-l:
> What should we do if we have nn.Parameter？

what's the context of this question?

yh8899:
> what's the context of this question?

Like this 'self.scale_shift_table = nn.Parameter(torch.empty(video_mod_param_num, config.dim))', bare nn.Parameter, not Module.

tianyu-l:
> Like this 'self.scale_shift_table = nn.Parameter(torch.empty(video_mod_param_num, config.dim))', bare nn.Parameter, not Module.

What is "self", can we not wrap its config by `FrozenConfig`?

yh8899:
> What is "self", can we not wrap its config by `FrozenConfig`?

Here:
https://github.com/huggingface/diffusers/blob/b003a47354ca3a28b4b21fceb4d649656d8cee99/src/diffusers/models/transformers/transformer_ltx2.py#L571

like this, it is not wraped by Config. After convert, parameter still trainable.

tianyu-l:
It's in `LTX2VideoTransformerBlock`, which itself can be wrapped by `FrozenConfig` https://github.com/huggingface/diffusers/blob/b003a47354ca3a28b4b21fceb4d649656d8cee99/src/diffusers/models/transformers/transformer_ltx2.py#L412

and latest change would freeze it?
https://github.com/pytorch/torchtitan/pull/3456/changes#diff-998ac1924c6dc1231fb40f979abb1400fe4d7fa7c9a7158f65bae0117efc7fe1R136-R137

yh8899:
emm,i haven't test the latest change, is this can freeze its parameters?

yh8899:
Yeah, it can solve the ltx block, but it can't solve parameter in top model,like the WanTransformer3DModel:
https://github.com/huggingface/diffusers/blob/b003a47354ca3a28b4b21fceb4d649656d8cee99/src/diffusers/models/transformers/transformer_wan.py#L624

FrozenConfig is not a BaseModel.Config

tianyu-l:
> FrozenConfig is not a BaseModel.Config

oh, I think that was caused by 
```
model_config.traverse(Module.Config, recurse=True)
```
not searching the `model_config` itself. Should be an easy fix. @Kevin-Li-2025

yinli-systems:
Thanks for pointing out the top-level bare `nn.Parameter` case. I pushed a follow-up in `a67de64c0` that handles the root `Module.Config` explicitly.

What changed:
- `Configurable.Config.traverse(..., include_self=True)` can now opt into yielding the root config. The root yield uses `parent=None` / `attr=None` because it cannot be replaced through a parent container.
- `LoRAConverter` uses that root yield to mark the root `Module.Config` for direct-parameter freezing instead of trying to wrap/replace the root with `FrozenConfig`.
- `Module.Config.build()` freezes only direct parameters when that internal flag is set, so nested LoRA adapters remain trainable.
- `_make_lora_config` now copies only dataclass init fields, so the private freeze flag does not leak into generated LoRA configs.

I also added regression coverage for:
- `traverse(..., include_self=True)` root traversal semantics.
- A root module with a direct bare `nn.Parameter` plus a child Linear converted to LoRA, verifying that only LoRA adapter params remain trainable.

Validation run locally in a Python 3.10 + PyTorch nightly CPU environment matching the unit-test workflow setup:
- `pytest tests/unit_tests/test_lora.py tests/unit_tests/test_configurable.py -q`: 20 passed.
- `pytest tests/unit_tests -q -k 'not TestTokenizerIntegration'`: 372 passed, 2 skipped, 5 deselected.
- `git diff --check`: passed.
- pre-commit hooks on modified files, excluding pyrefly: passed.
- targeted pyrefly on `torchtitan/components/lora.py`, `torchtitan/config/configurable.py`, and `torchtitan/protocols/module.py`: 0 errors.

The full `tests/unit_tests` run had two unrelated tokenizer integration failures: one HuggingFace gated Llama repo 401 and one DeepSeek tokenizer/transformers behavior mismatch. The PR is now rebased on upstream/main and GitHub reports it as mergeable, pending review.

torchtitan/config/configurable.py
```diff
@@ -73,8 +73,15 @@ def _convert(val):
             }
 
         def traverse(
-            self, config_cls: type, *, _prefix: str = ""
-        ) -> Iterator[tuple[str, "Configurable.Config", object, str | int]]:
+            self,
+            config_cls: type,
+            *,
+            recurse: bool = False,
+            include_self: bool = False,
```
tianyu-l: why providing False option? Should always include self.

torchtitan/components/lora.py
```diff
@@ -183,26 +188,38 @@ def _make_lora_config(self, cfg: Linear.Config):
...
             )
+
+            if parent is None:
+                if is_target:
+                    raise ValueError(
+                        "LoRAConverter cannot replace the root Linear.Config "
+                        "in-place. Pass a model config that owns the target "
+                        "Linear.Config as a field."
+                    )
+                cfg._freeze_direct_params = True
```
tianyu-l: I don't see why we have to introduce this mechanism. Couldn't we make it self contained here?

torchtitan/protocols/module.py
```diff
@@ -41,6 +41,7 @@ class Module(nn.Module, Configurable):
     class Config(Configurable.Config):
         param_init: dict | None = None
         sharding_config: ShardingConfig | None = None
+        _freeze_direct_params: bool = field(default=False, init=False, repr=False)
```
tianyu-l: shouldn't introduce this, I need to be convinced o/w

tianyu-l (changes requested):
plus some of my earlier comments not addressed

yinli-systems:
Thanks, this is fair. I tried to cover the top-level bare `nn.Parameter` case by adding `_freeze_direct_params` to `Module.Config`, but I agree that makes a LoRA-specific concern leak into the base module protocol. I don't want to defend that direction.

I'll rework the patch along these lines:

- make root traversal the normal behavior instead of adding an `include_self=False` option;
- move `FrozenConfig` back to a small `Configurable.Config` wrapper with `inner: Module.Config`;
- remove the explicit `param_init` / `sharding_config` copying in `FrozenConfig.build()` since `__setattr__` already forwards those updates to `inner`;
- remove `_freeze_direct_params` from `Module.Config` entirely;
- keep the LoRA converter self-contained: matching `Linear.Config` gets LoRA wrapping, other replaceable `Module.Config`s get `FrozenConfig` wrapping.

For the root bare-parameter case, I think the current in-place `convert(config) -> None` API is the awkward part: we cannot actually replace the root config through a parent container. Instead of hiding that by mutating `Module.Config`, I'll either make that limitation explicit in this PR or handle it with an explicit converter API change if that still seems worth it after simplifying the patch.

I also see the earlier `FrozenConfig` / `Module.Config` and sharding comments were still effectively unresolved. I'll address those directly in the next push rather than stacking more special cases on top of this version.

tianyu-l:
> handle it with an explicit converter API change if that still seems worth it after simplifying the patch.

I think this is fair, we can change ModelConfigConverter to return a modified model_config, not necessarily inplace.
https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/model.py#L27

torchtitan/config/configurable.py
```diff
@@ -85,21 +92,45 @@ def traverse(
...
+                                    config_cls, recurse=recurse, _prefix=item_fqn
+                                )
                         elif hasattr(item, "traverse"):
-                            yield from item.traverse(config_cls, _prefix=item_fqn)
+                            yield from item.traverse(
+                                config_cls, recurse=recurse, _prefix=item_fqn
+                            )
                 elif hasattr(val, "traverse"):
-                    yield from val.traverse(config_cls, _prefix=fqn)
+                    yield from val.traverse(config_cls, recurse=recurse, _prefix=fqn)
```
Andrei-Aksionov: I know it's a nit, but could you simplify this method a bit by removing duplicated code?

As I see it we just need to yield if there is a match and maybe continue descending; if there is no match - maybe continue descending if possible.
And apply this logic to every item in a list consecutively.

So that logic could be formed as a function (e.g. _visit or something like that):
```text
def _visit(item):
    if matches:
        yield
        maybe descend
    elif is_config:
        descend
```

And apply it:
```text
if field is list:
    apply _visit() to every element
else:
    apply _visit() once
```

This one passes updated `tests/unit_tests/test_configurable.py`:
```python
def traverse(
    self,
    config_cls: type,
    *,
    recurse: bool = False,
    include_self: bool = False,
    _prefix: str = "",
) -> Iterator[
    tuple[str, "Configurable.Config", object | None, str | int | None]
]:
    if include_self and isinstance(self, config_cls):
        yield _prefix, self, None, None
        if not recurse:
            return

    def _visit(val, fqn, parent, key):
        if isinstance(val, config_cls):
            yield fqn, val, parent, key
            if recurse:
                yield from val.traverse(
                    config_cls, recurse=recurse, _prefix=fqn
                )
        elif isinstance(val, Configurable.Config):
            yield from val.traverse(config_cls, recurse=recurse, _prefix=fqn)

    for f in fields(self):
        val = getattr(self, f.name)
        fqn = f"{_prefix}.{f.name}" if _prefix else f.name

        if isinstance(val, list):
            for i, item in enumerate(val):
                yield from _visit(item, f"{fqn}.{i}", val, i)
        else:
            yield from _visit(val, fqn, self, f.name)
```

And thanks to fewer indentations should be easier to read.

tianyu-l:
any updates? @Kevin-Li-2025

tianyu-l:
we'll take over if no followup

yinli-systems:
Pushed an updated version at 338a6f05. This should address the earlier review comments and the top-level bare parameter case without keeping the LoRA-specific flag in Module.Config.

What changed:
- ModelConfigConverter.convert now returns the possibly modified model config, and the model registries use that return value. This lets a converter replace/wrap the root config explicitly instead of mutating base Module.Config state.
- Configurable.Config.traverse now includes the root config by default. There is no include_self option anymore; the root yield uses parent=None / attr=None, and nested recursive yields preserve the actual replaceable parent/attr.
- FrozenConfig is back to a small Configurable.Config wrapper with inner: Module.Config. It no longer inherits Module.Config, no longer carries param_init/sharding_config metadata, and no longer needs _freeze_direct_params.
- LoRAConverter is self-contained now: matching Linear.Config nodes become LoRA configs, every other Module.Config is wrapped with FrozenConfig, including the root via the returned config. FrozenConfig freezes only direct params, so nested LoRA adapters remain trainable.
- The quantization converters now also return model_config so the converter protocol remains consistent.

I added regression coverage for recursive traversal parent/attr semantics, composite modules with direct parameters, and root modules with direct parameters.

Local validation:
- py_compile on the touched Python files: passed
- git diff --check: passed
- ruff format --check on touched files: passed
- pytest tests/unit_tests/test_configurable.py -q: 13 passed

I could not run test_lora.py end-to-end on my local macOS/Python 3.14 environment because the import path requires triton and there is no triton wheel for this environment, but the LoRA changes are covered by the added tests and should run in the Linux CI environment.

torchtitan/components/lora.py
```diff
@@ -106,13 +107,13 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
 
 @dataclass(kw_only=True, slots=True)
 class FrozenConfig(Configurable.Config):
```
tianyu-l: CPU unit test failed. It seems caused by instance of this `FrozenConfig` would fail conditions like
```
if isinstance(x, FlexAttention.Config):
```

I'm not sure what's the best solution -- I guess we have to introduce dynamic subclassing?

yinli-systems:
Pushed an update at 1255e9a7 to address the CPU unit test failure from the frozen config wrapper.

The previous `FrozenConfig(inner=...)` wrapper no longer preserved checks like `isinstance(inner_attention, FlexAttention.Config)`. I replaced it with per-config dynamic frozen subclasses, so non-LoRA configs still subclass their original config class while only overriding `build()` to freeze direct parameters.

One important detail: because the new frozen config is a copy rather than a forwarding wrapper, the converter now applies replacements from leaves to root. That keeps child LoRA replacements visible when the root/composite config is copied into its frozen subclass.

I also added a regression test covering the `FlexAttention.Config` type-check case.

Local validation:
- `python3 -m py_compile torchtitan/components/lora.py tests/unit_tests/test_lora.py`: passed
- `git diff --check -- torchtitan/components/lora.py tests/unit_tests/test_lora.py`: passed
- minimal dynamic subclass behavior check: passed

I still cannot run `tests/unit_tests/test_lora.py` end-to-end on my local macOS/Python 3.14 setup because collection imports Triton-only modules and there is no usable Triton package for this environment. The added regression should run in the Linux CI environment.

torchtitan/components/lora.py
```diff
@@ -104,42 +105,43 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
...
+    if getattr(config_cls, "_lora_frozen_config", False):
+        return config_cls
+    if config_cls in _frozen_config_class_cache:
+        return _frozen_config_class_cache[config_cls]
 
-    Works with any ``Module.Config`` — used by ``LoRAConverter`` to freeze
-    non-target Linear modules.
-    """
+    class FrozenConfig(config_cls):  # type: ignore[valid-type, misc]
+        _lora_frozen_config = True
```
tianyu-l: This feels unnecessary, can we remove?

tianyu-l (approved):

yinli-systems:
Thanks for the review. I removed the redundant `_lora_frozen_config` marker and pushed the update. The PR is currently approved and mergeable; I only see the Meta checks on the latest commit, so GitHub Actions may need maintainer approval to run on the fork.

torchtitan/components/lora.py
```diff
@@ -109,14 +109,15 @@ def _get_frozen_config_cls(
     config_cls: type[Module.Config],
 ) -> type[Module.Config]:
     """Get or create a config subclass that freezes direct build parameters."""
-    if getattr(config_cls, "_lora_frozen_config", False):
+    if any(
+        issubclass(config_cls, frozen_cls)
+        for frozen_cls in _frozen_config_class_cache.values()
+    ):
         return config_cls
```
tianyu-l: why do we need this? Is the two lines below not enough?

yinli-systems:
I updated the branch to address the latest CI state: removed the unrelated fake-backend timeout change from this PR and merged current upstream/main (the branch was 17 commits behind; the latest scheduled RL workflow on main passed). The prior RL failure occurred before LoRA conversion, while Tyro parsed a stale ModelSpec default. The PR diff is back to the LoRA/config-converter scope, and py_compile plus git diff --check pass locally. Could a maintainer approve/rerun the fork workflows on commit 280c352e?

tianyu-l:
The RL test is healthy on main https://github.com/pytorch/torchtitan/actions/workflows/integration_test_8gpu_rl.yaml?query=branch%3Amain
so it's likely that the regression is introduced from this PR

yinli-systems:
I found the RL H100 failure root cause and pushed `639bb4f1`.

The PR changed `ModelConfigConverter.convert()` to return the possibly replaced root config, and the model registries now assign that return value. `LMHeadCastConverter` was still an in-place converter returning `None`, so the RL Qwen3 registry path ended up replacing `ModelSpec.model` with `None`; that is why the bitwise parity setup failed at `model_spec.model.update_from_config(...)`.

The fix makes `LMHeadCastConverter.convert()` return the original model config after swapping `lm_head`, and adds a focused regression test for that contract.

Local checks run:
- `py_compile torchtitan/experiments/rl/models/cast_linear.py torchtitan/experiments/rl/tests/test_cast_linear.py`
- `git diff --check`

I tried to run `torchtitan/experiments/rl/tests/test_cast_linear.py` locally, but this Mac environment is missing the full RL/Triton dependency stack; the test collection reaches `torchtitan.distributed.minimal_async_ep.kernels` and fails on `import triton`. The H100 CI should now cover the failing RL path.

yinli-systems:
CI triage update after `639bb4f1`:

I was able to fetch the failing job logs through the job-level logs API (`/actions/jobs/{job_id}/logs`), even though `gh run view --log` did not show useful output.

Current failing checks on this PR:

- `RL H100 Integration Tests` (`28022867130`, job `83055788717`) failed before running tests while `uv pip install torch torchvision vllm xformers ...` was fetching `numpy` metadata from PyPI: `Connection reset by peer (os error 104)`. This looks like a transient runner/network/PyPI failure and should be rerun.
- `8 GPU Model Tests` CUDA/ROCm only failed the `gpt_oss` flavors. The failing commands do not pass any LoRA/model converter overrides, so these are not exercising the code path changed by this PR. Recent `main` 8GPU model runs are also failing (`28081841013` failed and `28099543230` is still in progress at the time of writing), so I do not see a PR-scoped LoRA/config-converter fix to make safely here.
- `RL Integration Tests` on this PR failed in `RL GRPO TP=2 no compile` during generator/vLLM setup (`Cannot set version_counter for inference tensor` in RoPE cache reshape). I checked recent `main` RL workflows as well, and they are currently red too (`28099497831`, `28100485308`), though with a different GPT-OSS RL failure signature. Given that the RL workflow is not green on `main` right now, I cannot safely attribute this to the LoRA/config-converter diff without a rerun after `main` recovers.

I do not want to add unrelated fixes to this already-approved LoRA PR. Could a maintainer rerun the H100 job (network failure), and rerun/compare the 8GPU model + RL jobs once the corresponding `main` workflows are green? If a rerun still leaves a PR-only failure, I can follow up on that specific remaining failure.

tianyu-l:
I believe several issues have been fixed. Would you please rebase to latest main so that we can test this PR again? Thanks!

yinli-systems:
Rebased onto latest `main` and force-pushed the branch. New head: `e494cffb`.

Local checks I ran after the rebase:
- `git diff --check upstream/main...HEAD`
- `PYTHONPYCACHEPREFIX=/tmp/torchtitan-pycache /usr/bin/python3 -m py_compile torchtitan/components/lora.py`
- `PYTHONPATH=$PWD PYTHONPYCACHEPREFIX=/tmp/torchtitan-pycache .venv-torchtitan-codex/bin/python -m pytest tests/unit_tests/test_configurable.py -q` (`13 passed`)

I also tried `tests/unit_tests/test_lora.py`, but local collection is blocked on this Mac venv by missing `triton`; CI should cover it on the Linux GPU environment. Thanks!

PR #3458, opened 2026-05-29
[RoPE] Move centralized freqs_cis to each transform layer

wwwjn:
- For language model: move the RoPE cache to each layer's attention
- For qwen3-vl: the RoPE is related to the input.
    -  Qwen3-VL computes input-dependent MRoPE position IDs with shape [3, batch, seq] from tokens, image/video grids, and packed-sequence positions, then passes those IDs through the normal layer positions argument.
    - Each attention layer keeps its own static RoPE cache; when RoPE.forward() sees 3D positions and mrope_section, it locally builds the interleaved 4D MRoPE cos/sin cache for that forward without mutating persistent layer state.

Why we make this change:
- Allow registry kernels: Make RoPE replaceable as a module level.
- For PP: every stage owns the RoPE modules for its local layers

## Test 
Qwen3-VL loss comparision:
```
  python scripts/loss_compare.py main <temp_patch_commit> \
    --baseline-module=qwen3_vl --baseline-config=qwen3_vl_debugmodel \
    --test-module=qwen3_vl --test-config=qwen3_vl_debugmodel \
    --baseline-options='--dataloader.dataset=cc12m-test' \
    --test-options='--dataloader.dataset=cc12m-test' \
    --steps=2 --baseline-ngpus=1 --test-ngpus=1 \
    --assert-equal
```

  Result: passed. Losses were bitwise equal from TensorBoard:

  step 1: baseline 12.634057998657227, test 12.634057998657227
  step 2: baseline 12.479432106018066, test 12.479432106018066

DeepSeek-v3 loss comparision: 
```
  python scripts/loss_compare.py main <temp_patch_commit> \
    --baseline-module=deepseek_v3 --baseline-config=deepseek_v3_debugmodel \
    --test-module=deepseek_v3 --test-config=deepseek_v3_debugmodel \
    --steps=2 --baseline-ngpus=1 --test-ngpus=1 \
    --assert-equal
````
  Result: passed. TensorBoard losses were exactly equal:

  step 1: baseline 8.083154678344727, test 8.083154678344727
  step 2: baseline 6.283287048339844, test 6.283287048339844

### CODE REVIEW

File: torchtitan/models/llama3/__init__.py [hunk 1]
```diff
@@ -120,18 +125,14 @@ def _debugmodel(attn_backend: str) -> Llama3Model.Config:
...
-            scaling="llama",
-        ),
         layers=_build_llama3_layers(
             n_layers=n_layers,
             dim=dim,
             n_heads=n_heads,
             hidden_dim=compute_ffn_hidden_dim(dim, multiple_of=256),
+            rope_max_seq_len=131072,
+            rope_theta=500000,
+            rope_scaling="llama",
```
wwwjn: Passing these rope parameters from ``_build_llama3_layers`` -> `_make_gqa` config, then create a RoPE.Config() instance for each layer's attention module.

torchtitan/models/gpt_oss/sharding.py
```diff
@@ -77,19 +77,21 @@ def _set_gpt_oss_layer_sharding(
     layer_cfg.ffn_norm.sharding_config = norm
     attn_x_placement: Placement = Shard(1) if enable_sp else Replicate()
 
-    # Attention: input x gathered to Replicate, freqs_cis always Replicate.
+    # Attention: input x gathered to Replicate. RoPE is read from the attention
```
wwwjn: This coment will be reverted

torchtitan/models/llama3/__init__.py
```diff
@@ -70,6 +69,9 @@ def _build_llama3_layers(
     dim: int,
     n_heads: int,
     hidden_dim: int,
+    rope_max_seq_len: int,
+    rope_theta: float,
+    rope_scaling: str,
```
wwwjn: Passing these rope parameters from `_build_llama3_layers() -> _make_gqa()`, then create a RoPE.Config() instance for each layer's attention module.

tests/unit_tests/test_fsdp_moe_sharding.py
```diff
@@ -32,35 +32,34 @@ def _build_llama4_model(num_experts: int = 8) -> Llama4Model:
...
             attn_backend="flex",
+            shared_experts_hidden_dim=None,
             moe_comm_backend="standard",
+            non_blocking_capacity_factor=None,
+            rope_max_seq_len=2048,
+            rope_theta=500000,
+            rope_backend="complex",
+            rope_scaling="llama",
+            rope_scaling_factor=16.0,
+            rope_high_freq_factor=1.0,
```
tianyu-l: no, rope should have its own config

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -321,7 +321,9 @@ def __init__(
             # Enables RequestOutput.metrics, so generator metrics can be returned
             disable_log_stats=False,
         )
-        engine_kwargs["max_model_len"] = model_spec.model.rope.max_seq_len
+        engine_kwargs["max_model_len"] = model_spec.model.layers[
+            0
+        ].attention.rope.max_seq_len
```
tianyu-l: we probably should have a property in decoder that provide a shortcut to `layers[0].attention.rope.max_seq_len`

torchtitan/models/common/config_utils.py
```diff
@@ -67,12 +68,25 @@ def make_gqa_config(
...
+    rope_scaling: Literal["none", "llama", "yarn"] = "none",
+    rope_scaling_factor: float = 8.0,
+    rope_low_freq_factor: float = 1.0,
+    rope_high_freq_factor: float = 4.0,
+    rope_original_max_position_embeddings: int = 8192,
+    rope_factor: float = 1.0,
+    rope_beta_fast: float = 32.0,
+    rope_beta_slow: float = 1.0,
+    rope_original_seq_len: int = 4096,
+    rope_mscale: float = 0.0,
```
tianyu-l: No...

torchtitan/models/common/decoder_sharding.py
```diff
@@ -138,13 +138,15 @@ def set_gqa_attention_sharding(attention_cfg, *, enable_sp: bool) -> None:
...
         in_src_shardings={
             "x": dense_activation_placement(tp=attn_x_placement),
-            "rope_cache": dense_param_placement(tp=Replicate()),
         },
         in_dst_shardings={
             "x": dense_activation_placement(tp=Replicate()),
-            "rope_cache": dense_param_placement(tp=Replicate()),
         },
     )
+    if attention_cfg.use_rope:
```
tianyu-l: I think we should deprecate this field, and depend on if rope field is None / not-None

torchtitan/models/common/rope.py
```diff
@@ -211,18 +212,87 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
+        key: torch.Tensor,
+        positions: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Apply rotary embeddings to query and key tensors."""
+        rope_cache = self._cache_for_positions(positions)
+        if self.config.backend == "cos_sin":
+            return apply_rotary_emb_cos_sin(query, key, rope_cache, positions)
+        return apply_rotary_emb_complex(query, key, rope_cache, positions)
+
+    def _compute_mrope_cache(self, position_ids: torch.Tensor) -> torch.Tensor:
```
tianyu-l: This doesn't look right -- we shouldn't special case mrope for qwen3_vl inside default RoPE.

It seems worth it's own RoPE class.

cc @shuhuayu

tianyu-l (changes requested):

torchtitan/models/common/config_utils.py
wwwjn: I tried directly passing a RoPE.Config() directly to make_gqa_config() and then I find all layers are sharing a mutable config state across layers, then I have to make a copy each time build a attention. So I ends up following how to build Linear / QKVLinear. 

If these config fields are too many, the make a copy path would look like:
```
def make_gqa_config(xxx, rope: RoPE.Config)
    copied_rope = dataclasses.replace(rope) # make a copy
    return GQAttention(
        rope=copied_rope
    )
```
Does this looks better?

torchtitan/models/common/rope.py
wwwjn: Let me try to subclass it. My understanding is Qwen3-VL's final mrope is related to input data (`position_ids`), so we can not create them ahead of time, so we can just create base frequency table like normal rope

torchtitan/models/common/config_utils.py
tianyu-l: yeah that's better

torchtitan/models/common/rope.py
shuhuayu: I would prefer a separate `MRoPE` class. A basic RoPE class serves as the backbone for text models. MRoPE, and similarly 2D RoPE,  are specially for models with vision input, and uses components in basic `RoPE` class.

torchtitan/models/llama3/__init__.py
```diff
@@ -120,18 +121,20 @@ def _debugmodel(attn_backend: str) -> Llama3Model.Config:
...
-            backend="complex",
-            scaling="llama",
-        ),
         layers=_build_llama3_layers(
             n_layers=n_layers,
             dim=dim,
             n_heads=n_heads,
+            n_kv_heads=None,
             hidden_dim=compute_ffn_hidden_dim(dim, multiple_of=256),
+            fuse_qkv=False,
```
tianyu-l: why False?

torchtitan/models/llama3/__init__.py
```diff
@@ -120,18 +121,20 @@ def _debugmodel(attn_backend: str) -> Llama3Model.Config:
...
-            max_seq_len=131072,
-            theta=500000,
-            backend="complex",
-            scaling="llama",
-        ),
         layers=_build_llama3_layers(
             n_layers=n_layers,
             dim=dim,
             n_heads=n_heads,
+            n_kv_heads=None,
```
tianyu-l: seems orthogonal change

torchtitan/models/deepseek_v3/__init__.py
```diff
@@ -174,6 +192,13 @@ def _build_dsv3_layers(
     attn_backend: str,
     moe_comm_backend: str,
     non_blocking_capacity_factor: float | None,
+    rope_max_seq_len: int,
+    rope_theta: float,
+    rope_scaling: str,
+    rope_factor: float,
+    rope_beta_fast: float,
+    rope_beta_slow: float,
+    rope_original_seq_len: int,
```
tianyu-l: not fixed

torchtitan/models/common/config_utils.py
```diff
@@ -67,17 +69,23 @@ def make_gqa_config(
...
     qk_norm: RMSNorm.Config | None = None,
 ) -> GQAttention.Config:
     """Build a fully-specified GQAttention.Config."""
     n_kv = n_kv_heads if n_kv_heads is not None else n_heads
     per_head_dim = head_dim if head_dim is not None else dim // n_heads
+    if rope.dim != per_head_dim:
+        raise ValueError(
+            f"RoPE dim must match attention head dim, got {rope.dim=} and "
+            f"{per_head_dim=}."
+        )
```
tianyu-l: do we not have such check before?

torchtitan/models/qwen3_vl/model.py
tianyu-l: why do we still need this at top level; we already put it in each layer's mrope

torchtitan/models/qwen3_vl/model.py
```diff
@@ -95,12 +94,21 @@ def __init__(self, config: Config):
         self.vision_encoder = config.vision_encoder.build()
 
         self.mrope_section = config.mrope_section
+        for layer in self.layers.values():
+            if not isinstance(layer.attention.rope, MRoPE):
+                raise ValueError("Qwen3-VL attention layers must use MRoPE.")
```
tianyu-l: If you really want to do this check, it should be in `Qwen3Model.Config`'s post_init, couldn't change during CLI so perhaps not in `update_from_config`

torchtitan/models/common/rope.py
```diff
@@ -211,18 +211,28 @@ def _precompute_cos_sin(self) -> torch.Tensor:
         sin = theta.sin() * mscale
         return torch.cat([cos, sin], dim=-1)
 
-    def forward(
-        self, seq_len: int, positions: torch.Tensor | None = None
+    def _cache_for_positions(
```
tianyu-l: the necessity of this function is not clear

torchtitan/models/common/rope.py
```diff
@@ -211,18 +211,28 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
+            return apply_rotary_emb_cos_sin(query, key, rope_cache, positions)
+        return apply_rotary_emb_complex(query, key, rope_cache, positions)
+
     def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> None:
-        if buffer_device is not None:
-            with torch.device(buffer_device):
-                self.cache = self._precompute()
-        else:
-            self.cache = self._precompute()
+        cache = self.cache.to_local() if isinstance(self.cache, DTensor) else self.cache
```
tianyu-l: why to_local??

torchtitan/models/common/rope.py
```diff
@@ -402,14 +412,12 @@ def apply_rotary_emb_cos_sin(
         xq: (bsz, seqlen, n_heads, head_dim)
         xk: (bsz, seqlen, n_kv_heads, head_dim)
         rope_cache: (max_seqlen, head_dim * 2) with cos and sin concatenated,
-            or (bsz, seqlen, 1, head_dim * 2) if already broadcast-shaped (MRoPE)
+            or (bsz, seqlen, 1, head_dim * 2) if already broadcast-shaped
         positions: optional position indices
     """
-    positions = _maybe_wrap_positions(positions, xq)
     head_dim = xq.shape[-1]
     if rope_cache.ndim != 4:
```
tianyu-l: since we refactor mrope out, should we remove this special case == 4?

torchtitan/models/common/rope.py
```diff
@@ -211,18 +211,28 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        positions: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Apply rotary embeddings to query and key tensors."""
+        rope_cache = self._cache_for_positions(positions)
+        if self.config.backend == "cos_sin":
+            return apply_rotary_emb_cos_sin(query, key, rope_cache, positions)
+        return apply_rotary_emb_complex(query, key, rope_cache, positions)
```
tianyu-l: these functions should become instance methods, which has `cache` access

torchtitan/experiments/rl/actors/generator.py
wwwjn: Good point, I added a property in Decoder to do so.

torchtitan/models/llama3/__init__.py
wwwjn: Removed, also checked and reverted all the unrelated changes in other models' config

torchtitan/models/common/config_utils.py
wwwjn: Should b e removed, this a unfortunately a slop

torchtitan/models/common/rope.py
wwwjn: For normal RoPE, it directly returns cache,
For MRoPE, it returns the cache calculated with `position`

torchtitan/models/common/rope.py
wwwjn: I check the usage of these functions, and one thing stops us make it an instance methods is in Qwen3-VL vision encoder,  we need to pass rope_cache (which is calculated outside, not `self`)

torchtitan/models/common/rope.py
wwwjn: Currently `apply_rotary_emb_cos_sin` is still shared across rope and MRoPE. 

An alternative way is to make `apply_rotary_emb_cos_sin` an instance method in `class RoPE` and let `MRoPE(RoPE)` override it when inheritating from RoPE, and only handles 4D situation in MRoPE. The disadvantage is we need to maintain 2 copy of code `apply_rotary_emb_cos_sin`. So I'd prefer current solution with some notes here, wdyt?

torchtitan/models/common/rope.py
```diff
@@ -211,18 +211,29 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
+
     def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> None:
```
wwwjn: This function needs to be restored exactly to main's version

torchtitan/models/common/attention.py
```diff
@@ -614,8 +610,8 @@ def __init__(self, config: Config):
             else config.dim // config.n_heads
         )
         self.enable_gqa = self.n_heads > self.n_kv_heads
-        self.use_rope = config.use_rope
-        self.rope_backend = config.rope_backend
+        self.use_rope = config.rope is not None
```
tianyu-l: can we delete this field?

torchtitan/models/common/decoder.py
```diff
@@ -194,25 +203,13 @@ def init_states(
         *,
         buffer_device: torch.device | None = None,
     ) -> None:
-        # Compute buffer_device before recursion so children (RoPE) get
-        # the correct device when buffer_device is not explicitly provided.
-        if buffer_device is None:
-            buffer_device = self.freqs_cis.device
```
tianyu-l, line 201: This logic is gone. Is it valid change?

torchtitan/models/common/rope.py
```diff
@@ -402,14 +413,12 @@ def apply_rotary_emb_cos_sin(
...
     """
-    positions = _maybe_wrap_positions(positions, xq)
     head_dim = xq.shape[-1]
     if rope_cache.ndim != 4:
-        # Qwen3-VL MRoPE validates position IDs before building its 4D cache.
-        # Here shape[0] is max sequence length only for non-4D indexed caches.
+        positions = _maybe_wrap_positions(positions, xq)
         if positions is not None:
             _maybe_check_max_pos(positions, max_valid_pos=rope_cache.shape[0] - 1)
         rope_cache = _reshape_for_broadcast_cos_sin(rope_cache, xq, positions)
```
tianyu-l: - Can we extract this part out as `_prepare_rope_cache`, so that MRoPE can override and put `_compute_mrope_freqs` logic in it?
- sin_cos and complex should define two RoPE subclasses, each provide their own defintion of `_precompute_cache`, `_prepare_rope_cache`, `apply_rotary_emb`; the selection of backend should be done at config time, not runtime -- similar to how attention backend works now
- in theory MRoPE can also have two variants one for sin_cos, the other for complex. According to https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/qwen3_vl/model.py#L129, today's qwen3_vl only supports cos_sin (did we error out if backend is complex?).

tianyu-l: This will make it easier for kernel override work to be done.

tianyu-l (changes requested):
I think we need to make bigger changes

torchtitan/models/common/decoder.py
wwwjn: Sorry I didn't mean to remove this, just want to update this logic to not access the `self.freqs_cis.device` after move rope to each layer's attention

torchtitan/models/qwen3_vl/vision_encoder.py
```diff
@@ -364,6 +364,7 @@ class Config(Module.Config):
         n_heads: int
         qkv: Linear.Config
         proj: Linear.Config
+        rope: RoPE.Config
```
wwwjn: I added a RoPE field in vision_encoder, but the usage is slightly different with other callsite: 
- other callsite calls self.rope.forward()
- Qwen3-vl's vision encoder call self.rope.apply_rotary_emb(q, k, rope_cache) and take the rope_cache calculated outside.

The advantage of doing this is we don't need to maintain an extra helper function `apply_rotary_emb` which is only called in vision encoder now. cc @shuhuayu

torchtitan/models/qwen3/__init__.py
```diff
@@ -213,6 +215,11 @@ def _debugmodel(attn_backend: str) -> Qwen3Model.Config:
             head_dim=head_dim,
             hidden_dim=3072,
             attn_backend=attn_backend,
+            rope=CosSinRoPE.Config(
```
tianyu-l, line 218: Is our state dict adapter working with both cos-sin & complex (then need to change adapter code), or is it assuming one of them (we should error out if otherwise)?

torchtitan/models/qwen3_vl/model.py
```diff
@@ -109,11 +102,10 @@ def _compute_mrope_freqs(
         special_tokens: dict[str, int],
         positions: torch.Tensor | None = None,
     ) -> torch.Tensor:
-        """Build 3D position IDs and compute interleaved MRoPE cos/sin frequencies.
+        """Build 3D position IDs for Qwen3-VL MRoPE.
 
-        Constructs (temporal, height, width) position IDs for each token, then
-        looks up cos/sin from the 1D RoPE table and overwrites H/W-assigned dims
-        with their own position lookups.
```
tianyu-l, line 116: some docstring seems gone

torchtitan/models/common/rope.py
```diff
@@ -211,100 +267,83 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        rope_cache: torch.Tensor,
+        positions: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        if rope_cache.ndim != 4:
+            raise ValueError(
+                f"CosSinRoPE expects a prepared 4D RoPE cache, got {rope_cache.ndim}D."
+            )
```
tianyu-l: qwen3_vl would use 2 paths, are you sure erroring out here is valid?

torchtitan/models/common/rope.py
```diff
@@ -211,100 +267,83 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
-def _reshape_for_broadcast_cos_sin(
+    def _prepare_rope_cache(
+        self,
+        query: torch.Tensor,
+        positions: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
+        positions = _maybe_wrap_positions(positions, query)
+        if positions is not None:
+            _maybe_check_max_pos(positions, max_valid_pos=self.cache.shape[0] - 1)
+        return _reshape_for_broadcast(self.cache, query, positions), None
```
tianyu-l: For complex this is noop, but for cos_sin positions is embedded in cache.
Ideally we should avoid such inconsistency. Can we unify the structure a bit more? I'm OK with collapsing to a larger function overall (but providing some utils functions to share between impls).

tianyu-l (commented):
haven't reviewed the device part, will review later.

torchtitan/models/common/decoder.py
```diff
@@ -194,25 +204,25 @@ def init_states(
...
+        # to infer the post-to_empty device from state.
         if buffer_device is None:
-            buffer_device = self.freqs_cis.device
+            for tensor in itertools.chain(
+                self.parameters(recurse=True),
+                self.buffers(recurse=True),
+            ):
+                if tensor.device.type != "meta":
+                    buffer_device = tensor.device
+                    break
```
tianyu-l: why do we need this? The logic should be moved in RoPE buffer init code.

torchtitan/models/common/rope.py
```diff
@@ -74,16 +65,51 @@ class Config(Module.Config):
...
+        return self.apply_rotary_emb(query, key, rope_cache, positions)
+
+    def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> None:
+        if buffer_device is None:
+            self.cache = self._precompute_cache()
         else:
-            return self._precompute_cos_sin()
+            with torch.device(buffer_device):
+                self.cache = self._precompute_cache()
+
```
tianyu-l, line 132: This looks good enough.

tianyu-l (changes requested):

torchtitan/models/common/rope.py
wwwjn: Not quiet,  previously in Decoder._init_self_buffers, we infer the buffer_device from rope cache when the input `buffer_device=None`.   
```
init_device = device_type
buffer_device = None  # when not using CPU offload
model.to_empty(device=init_device)
model.init_weights(buffer_device=buffer_device)
```

So I was trying to keep the same behavior of "inferring the buffer_device" , although I didn't fully understand why `buffer_device` is set to `None`.  

I think we'd better infer the buffer_device in Decoder before it recursively call children's `_init_self_buffers(buffer_deivce=None)` cause MoE also have buffer which might need to infer the device buffer again in MoE module. wdyt?

tianyu-l: sorry yes, the inferring part should be kept
you can get it here from self.cache.device

wwwjn: > sorry yes, the inferring part should be kept you can get it here from self.cache.device

Thank you, there's one more decision to make: Do we want the infer happen in Decoder, or in each Child Module? If in each child module, we'd need to change MoE module as well

tianyu-l: For now put in each child's module, for encapsulation.

In the long term, may need to have a better abstraction in Module class. Please add a TODO

cc @fegin

torchtitan/models/qwen3/__init__.py
wwwjn: In state_dict_adapter, we don't load or save RoPE cache so the it doesn't matter if we change name from `freqs_cis` to `rope.cache`. 

The only potential issue I could think of is Qwen3 model's HF weights are corresponding to `CosSinRoPE` (which we know after the discussion with HF), and DeepSeek's HF weights are corresponding to `ComplexRoPE`. We can add a check in loading from HF path if the RoPE type is not matching, .

tianyu-l: In terms of impl difference, we can have if-else in sd_adapter to do mapping on tensor, depending on the assumption in checkpoint.

But not a priority, we can also just error out if it doesn't match checkpoint.

torchtitan/models/common/rope.py
```diff
@@ -74,16 +65,55 @@ class Config(Module.Config):
...
+        return self.apply_rotary_emb(query, key, rope_cache, positions)
+
+    def _init_self_buffers(self, *, buffer_device: torch.device | None = None) -> None:
+        if buffer_device is None:
+            # After ``to_empty()``, the existing cache records the target device.
+            # Recompute there when the caller does not pass an explicit buffer device.
+            buffer_device = self.cache.device
+        if buffer_device.type == "meta":
+            self.cache = self._precompute_cache()
         else:
```
tianyu-l: why do we need this if-else?

orginal code doesn't have

torchtitan/models/common/rope.py
tianyu-l: ping me after you fixed this.

torchtitan/models/qwen3/__init__.py
wwwjn: Sounds good, I added check for llama3(Complex) , llama4 (Complex for RoPE layer), qwen3 (CosSin) , DeepSeek-v3(Complex), GPT-OSS (CosSinRoPE) , 

The check is performed at `from_hf()`. Putting it here is a safe because `from_hf() ` is only called when loading from HF. But it could still be a little bit overkill: a user might pre-train a model with different rope from scratch, save it in HF format, then reload it from the saved checkpoint with the different RoPE. But this is a super corner case so it's fine.

torchtitan/models/common/rope.py
wwwjn: The updated contract is:
- `_prepare_rope_cache()` will take `positions` as input, and return a 4-D cache already aligned with `query`'s shape, and already sliced by `position`. So this step baked `position` into the cache.
    - MRoPE rewrite this function by calling `_compute_mrope_cache()` and bake the position information as well
- `apply_rotary_emb()` will take the 4-D cache returned from `_prepare_rope_cache()` and apply it on query and key. So it will reject a 2D cache which is not yet ready to apply.

torchtitan/models/common/rope.py
wwwjn: Based on my understanding,. Qwen3-VL used it in:
1. In language encoder: where we used MRoPE, MRoPE's `_prepare_rope_cache` will return a 4D RoPE cache. 
2. In vision encoder: Vision encoder computes it's 4D RoPE ([here](https://github.com/pytorch/torchtitan/blob/4b226872f6ecd1de05e58f4f4e458dc7c0f962cc/torchtitan/models/qwen3_vl/vision_encoder.py#L543)) and passed the 4D rope to VisionAttention.Forward(), then we can call `apply_ropo_emb()` with this 4D input

torchtitan/models/common/rope.py
wwwjn: Nice catch, I don't think this is needed. Removed. 

We will only do meta init in the following way:
```
with torch.device("meta"):
    model = model_config.build()
# parallelize, etc
model.to_empty(cuda)
model.init_states(buffer_device=None)
```
So _init_self_buffers function will not see "meta" device in our current code path

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -60,6 +59,13 @@ def _make_gptoss_attn_config(
     head_dim: int = 64,
     sliding_window_size: int = 128,
     fuse_qkv: bool = False,
+    rope_max_seq_len: int,
+    rope_theta: float,
+    rope_scaling: Literal["llama", "none", "yarn"],
+    rope_factor: float,
+    rope_beta_slow: float,
+    rope_beta_fast: float,
+    rope_original_seq_len: int,
```
tianyu-l: fix

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -163,7 +179,14 @@ def _build_gptoss_layers(
...
     non_blocking_capacity_factor: float | None = None,
-) -> list[TransformerBlock.Config]:
+    rope_max_seq_len: int,
+    rope_theta: float,
+    rope_scaling: Literal["llama", "none", "yarn"],
+    rope_factor: float,
+    rope_beta_slow: float,
+    rope_beta_fast: float,
+    rope_original_seq_len: int,
+) -> list[GptOssTransformerBlock.Config]:
```
tianyu-l: fix

torchtitan/models/qwen3/state_dict_adapter.py
```diff
@@ -68,6 +69,15 @@ def __init__(self, model_config: Qwen3Model.Config, hf_assets_path: str | None):
...
         }
 
+    def _validate_hf_rope_config(self) -> None:
+        for layer in self.model_config.layers:
+            rope = layer.attention.rope
+            if not isinstance(rope, CosSinRoPE.Config):
+                raise ValueError(
+                    "Qwen3 HF checkpoint conversion assumes CosSinRoPE.Config; "
+                    f"got {type(rope).__name__}."
+                )
```
tianyu-l: If we don't want to define this for every model, could put in https://github.com/pytorch/torchtitan/blob/main/torchtitan/protocols/state_dict_adapter.py, with the RoPE cls as an arg.

torchtitan/models/common/rope.py
```diff
@@ -211,100 +308,86 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
+        return _reshape_for_broadcast(self.cache, query, positions)
+
+    def apply_rotary_emb(
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        rope_cache: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Apply cos/sin RoPE using the rotate-half convention."""
+        if rope_cache.ndim != 4:
```
tianyu-l: would this ever be hit? can we remove

torchtitan/models/common/rope.py
```diff
@@ -211,100 +308,86 @@ def _precompute_cos_sin(self) -> torch.Tensor:
...
-
-    rope_cache: (max_seqlen, head_dim * 2) where first half is cos, second is sin
-    x: (bsz, seqlen, n_heads, head_dim) real
-    """
+    """Reshape a RoPE cache for broadcasting with query/key tensors."""
     ndim = x.ndim
     assert ndim > 1
-    bz, seqlen, _, head_dim = x.shape
+    bsz, seqlen = x.shape[:2]
+    cache_width = rope_cache.shape[-1]
```
tianyu-l, line 362: comment that this is head_dim * 2 for cos_sin, and head_dim for complex

torchtitan/models/common/rope.py
```diff
@@ -161,7 +218,47 @@ def linear_ramp_factor(
...
+    ) -> torch.Tensor:
+        """Return complex cache shaped for query/key broadcast.
+
+        Returns:
+            Cache of shape ``(1 or batch, seq_len, 1, dim / 2)``.
+        """
+        positions = _maybe_wrap_positions(positions, query)
+        if positions is not None:
+            _maybe_check_max_pos(positions, max_valid_pos=self.cache.shape[0] - 1)
+        xq_ = torch.view_as_complex(query.float().reshape(*query.shape[:-1], -1, 2))
```
tianyu-l: This will be wasted. I don't think you need to materialize this tensor, all you need in `_reshape_for_broadcast` is the input shape

torchtitan/models/common/rope.py
```diff
@@ -74,16 +65,82 @@ class Config(Module.Config):
...
+
+    def _precompute_cache(self) -> torch.Tensor:
+        """Build the reusable cache for all positions up to ``max_seq_len``.
+
+        Returns:
+            RoPE cache for all valid positions.
+        """
+        raise NotImplementedError
+
+    def _prepare_rope_cache(
```
tianyu-l: remove `rope` in the name; maybe there's a more accurate name

torchtitan/models/qwen3_vl/vision_encoder.py
```diff
@@ -402,7 +404,9 @@ def forward(
         )
         # Each: (num_vision, max_num_patch, n_heads, head_dim)
         q, k, v = qkv.permute(2, 0, 1, 3, 4).unbind(0)
-        q, k = apply_rotary_emb_cos_sin(q, k, rope_cache)
+        # Vision RoPE cache is already position-specific from grid_thw, so
+        # bypass RoPE.forward() and apply the prepared cache directly.
```
tianyu-l: You'll still create rope.cache regardless, which is a waste anyway. (you only save `prepare_cache` computation)

You can make `apply_rotary_emb` a static function, and call `CosSinRoPE.apply_rotary_emb` without hosting rope in config.

This is creating one cache and use for all layers, similar to what default llm model does before this PR. But I guess it's fine as you don't split `VisionEncoder` across multiple PP ranks.

tianyu-l (changes requested):

torchtitan/models/qwen3_vl/vision_encoder.py
wwwjn: > You'll still create rope.cache regardless, which is a waste anyway.

Agreed, this is wasting calculation but avoiding having another helper function, but static function is a better idea and this function also doesn't need to access `self`

torchtitan/models/common/rope.py
wwwjn: What about `_build_positioned_cache()` or `_apply_positions_to_cache()` , I want to emphasis the cache is already slices by position (and also reshaped) in this function

tianyu-l: yeah, maybe
```
reshaped_cache = self._reshape_cache(query, positions)
```

torchtitan/models/common/rope.py
wwwjn: I left it as a defensive guard, as we also change `apply_rotary_emb()` a static function, it can be called with rope inputs. I want to guard the use case like Qwen3-VL vision encoder. But yes we can also remove that if the callsite is all controlled by us

torchtitan/models/common/rope.py
wwwjn: Added! it should be `head_dim * 2` for cos_sin, and `head_dim // 2` for complex.

For complex we have:
```
 freqs = 1.0 / (theta ** (torch.arange(0, dim, 2)[: (dim // 2)].float() /
  dim))
... # freqs shape doesn't change
freqs_cis = torch.polar(torch.ones_like(freqs), freqs)  # complex64

```
Since torch.arange(0, dim, 2) has dim // 2 items, freqs_cis has shape (max_seq_len, dim // 2)

torchtitan/models/flux/model/state_dict_adapter.py
```diff
@@ -8,7 +8,6 @@
 import logging
 import os
 import re
-
```
wwwjn, line 11: This empty line change is local linter issue, can be removed

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -281,17 +285,16 @@ def _20b(
...
-            max_seq_len=131072,
-            theta=150000.0,
-            backend="cos_sin",
-            scaling="yarn",
-            rope_factor=32,
-            beta_slow=32.0,
-            beta_fast=1.0,
-            original_seq_len=4096,
+            rope=CosSinRoPE.Config(
+                dim=dim // 64,
```
wwwjn: This line needs to be updated to align with orignial config.

torchtitan/models/llama3/state_dict_adapter.py
```diff
@@ -158,6 +158,7 @@ def to_hf(self, state_dict: dict[str, Any]) -> dict[str, Any]:
         return hf_state_dict
 
     def from_hf(self, hf_state_dict: dict[str, Any]) -> dict[str, Any]:
```
wwwjn, line 161: I didn't further change `from_hf()` function signature because the callsite is generic `checkpoint.py`

torchtitan/models/qwen3/__init__.py
```diff
@@ -105,7 +112,7 @@ def _build_qwen3_layers(
                     inner_attention=inner_attention,
                     fuse_qkv=fuse_qkv,
                     mask_type=mask_type,
-                    rope_backend="cos_sin",
+                    rope=rope,
```
wwwjn, line 115: To preserve the same semantic, we check the RoPE type in state_dict_adapter

torchtitan/models/gpt_oss/__init__.py
```diff
@@ -163,7 +164,8 @@ def _build_gptoss_layers(
     fuse_qkv: bool = False,
     moe_comm_backend: str,
     non_blocking_capacity_factor: float | None = None,
-) -> list[TransformerBlock.Config]:
+    rope: RoPE.Config,
+) -> list[GptOssTransformerBlock.Config]:
```
wwwjn: Will be reverted

torchtitan/models/qwen3/state_dict_adapter.py
```diff
@@ -103,9 +106,9 @@ def to_hf(self, state_dict: dict[str, Any]) -> dict[str, Any]:
 
                 # Store the GroupedExperts Weight metadata for from_hf()
                 if isinstance(value, DTensor):
-                    self.grouped_expert_weight_placements[
-                        abstract_key
-                    ] = value.placements
+                    self.grouped_expert_weight_placements[abstract_key] = (
```
wwwjn: This change is also linter, can be removed

torchtitan/protocols/state_dict_adapter.py
```diff
@@ -87,6 +86,7 @@ def __init__(
         model_config: BaseModel.Config,
         hf_assets_path: str | None,
     ):
+        self.model_config = model_config
```
wwwjn, line 89: This is needed because we will access model_config when validation hf rope config

torchtitan/protocols/state_dict_adapter.py
```diff
@@ -111,6 +111,23 @@ def __init__(
         else:
             self.fqn_to_index_mapping = None
 
+    def _validate_hf_rope_config(
+        self,
+        expected_rope_cls: type,
+        *,
+        allow_none: bool = False,
```
tianyu-l, line 118: add a comment why we have this -- for llama4 which combines rope and nope

I'll delete llama4 soon, so worth adding a TODO

tianyu-l (approved):
very nice refactor!!!

Please make sure numerics are bitwise.

wwwjn:
The RL test failure is unrelated to this PR, which will be fixed #3541 and @felipemello1

PR #3335, opened 2026-05-13
Revert "Avoid repeat_interleave output-size sync (#3274)"

xmfan:
Stacked PRs:
 * __->__#3335

--- --- ---

Revert "Avoid repeat_interleave output-size sync (#3274)"

This reverts commit e917ff4cf993e44f7e2abfcfbf30e0d5e7cc9ae9.

cc @syed-ahmed, who reported it breaks

```
-m torchtitan.train --module torchtitan_configs.deepseek_v3 --config deepseek_v3_671b --profiler.no-enable-profiling --metrics.enable-tensorboard --metrics.save_for_all_ranks --training.steps 50 --training.local_batch_size 16 --training.seq_len 4096 --parallelism.tensor_parallel_degree 1 --parallelism.pipeline_parallel_degree 4 --parallelism.context_parallel_degree 1 --parallelism.expert_parallel_degree 32 --compile.components model,loss --parallelism.data_parallel_shard_degree=64 --activation_checkpoint.mode=full
```

```
[rank0]:[CUDA_KERNEL_ASSERT] /opt/pytorch/pytorch/aten/src/ATen/native/cuda/Repeat.cu:20: compute_cuda_kernel: block: [3,0,0], thread: [416,0,0]: Assertion failed: `result_size == cumsum_ptr[size - 1]`: Invalid input! In `repeat_interleave`, the `output_size` argument (30364) must be the same as the sum of the elements in the `repeats` tensor (30365).
```

Strange issue
routed_input.shape[0] == 30364 # number of 
num_tokens_per_expert_group.sum() == 30365

This would imply the CPU splits used to allocate the all-to-all output disagree with the CUDA count tensor later used for permutation

PR #3459, opened 2026-05-29
[EP] Make input_splits D2H non-blocking and simplify _permute

aditvenk:
- Always use non_blocking=True for the input_splits device-to-host copy
- Reuse the repeat_interleave unbacked symint (seg_ids.shape[0]) instead of recomputing the token total in _permute

Verifies fix made for https://github.com/pytorch/pytorch/issues/182190

### DIFF
--- a/torchtitan/models/common/token_dispatcher.py
+++ b/torchtitan/models/common/token_dispatcher.py
@@ -283,14 +283,10 @@ class AllToAllTokenDispatcher(LocalTokenDispatcher):
                     num_global_tokens_per_local_expert_E
                 )
             )
-            # non_blocking=True is safe in eager, but under torch.compile the
-            # async D2H transfer can race with the subsequent .tolist()/.item()
-            # calls, producing stale values and failing unbacked-symint guards.
-            non_blocking = not torch.compiler.is_compiling()
             input_splits = (
                 num_local_tokens_per_expert_E.view(ep_size, -1)
                 .sum(dim=1)
-                .to(torch.device("cpu"), non_blocking=non_blocking)
+                .to(torch.device("cpu"), non_blocking=True)
             )
             # NOTE: this would incur a device-to-host sync
             output_splits = (
@@ -356,7 +352,6 @@ class AllToAllTokenDispatcher(LocalTokenDispatcher):
         ep_size = self.ep_mesh.size()
         e = num_global_tokens_per_local_expert_E.shape[0] // ep_size
         device = num_global_tokens_per_local_expert_E.device
-        total = num_global_tokens_per_local_expert_E.sum()
 
         # (EP, e) matrix of token counts per (rank, local_expert)
         t_mat = num_global_tokens_per_local_expert_E.view(ep_size, e)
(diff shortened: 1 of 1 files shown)


aditvenk:
TLParse: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/avenkataraman/4e3798e5-bd05-4f04-ab26-4d60a90a3eea/custom/rank_0/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

sync after copy_(.., True):
```
        buf64.copy_(buf63, True)
        _d2h_event_buf64 = torch.Event()
        _d2h_event_buf64.record()
        _d2h_event_buf64.synchronize()
        with torch.cuda._DeviceGuard(0):
            torch.cuda.set_device(0)
            buf65 = buf63; del buf63  # reuse
            # Topologically Sorted Source Nodes: [view_15, sum_2, output_splits], Original ATen: [aten.view, aten.sum, aten._to_copy]
            raw_stream0 = get_raw_stream(0)
            triton_poi_fused__to_copy_sum_view_12.run(buf58, buf65, 4, stream=raw_stream0)
```

tianyu-l (approved):
@sanketpurandare mentioned that there are multiple unnecessary D2H syncs in this code.

There was an attempt by @xmfan to remove some, but that got reverted https://github.com/pytorch/torchtitan/pull/3335

xmfan (approved):

torchtitan/models/common/token_dispatcher.py
sanketpurandare, line 371: This line itself creates a cpu-gpu sync since we have not specified output_size to repeat_interleave

sanketpurandare: Check this: https://docs.pytorch.org/docs/2.12/generated/torch.repeat_interleave.html

torchtitan/models/common/token_dispatcher.py
```diff
@@ -380,9 +375,13 @@ def _permute(
...
         output_starts = segment_lens.cumsum(0) - segment_lens
+        # seg_ids.shape[0] == segment_lens.sum(), but reuses the unbacked symint
+        # already created by repeat_interleave above. Computing the total token
+        # count separately would introduce a second symint for the same value,
+        # producing an Eq(u1, u2) constraint that inductor cannot lower.
         permuted_indices = (
             input_starts[seg_ids]
-            + torch.arange(total, device=device)
+            + torch.arange(seg_ids.shape[0], device=device)
             - output_starts[seg_ids]
```
sanketpurandare, line 380: This is correct, using total here would introduce another cpu-gpu sync since total is a gpu scalar

torchtitan/models/common/token_dispatcher.py
sanketpurandare, line 296: We should add here: `output_size = output_splits.sum().item()`. Since this is a cpu tensor, calling item here is not harmful and the benefit is, it will only create a single sym-int with correct range and not a single sym-int expression with a large range.

torchtitan/models/common/token_dispatcher.py
sanketpurandare: We can use the output_size as defined above here instead.

torchtitan/models/common/token_dispatcher.py
sanketpurandare: We can now pass the the output_size defined above to repeat_interleave.

sanketpurandare (changes requested):

torchtitan/models/common/token_dispatcher.py
aditvenk: I thought @xmfan tried this already but had to be reverted as Syed ran into some issues with 671B. https://github.com/pytorch/torchtitan/pull/3335

sanketpurandare (approved):
I tried for quite a while I am unable to repro Syed's error but I think that can be addressed in future PR. So approving this!

PR #3460, opened 2026-05-29
Use torchmonarch stable

pzhan9:
monarch's latest stable release is 0.5.0:
https://pypi.org/project/torchmonarch/0.5.0/

Instead of pinning a version, it is better to install the latest stable directly, so we do use a version that is too old.

### DIFF
--- a/.github/workflows/integration_test_4gpu_rl.yaml
+++ b/.github/workflows/integration_test_4gpu_rl.yaml
@@ -65,7 +65,7 @@ jobs:
         pip install uv
 
         # 1. Install Monarch and TorchStore
-        uv pip install torchmonarch==0.4.1
+        uv pip install torchmonarch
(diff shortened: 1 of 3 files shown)


tianyu-l (commented):
shall we also update here ? https://github.com/pytorch/torchtitan/blob/main/.github/workflows/integration_test_4gpu_rl.yaml#L68

> (actually do we need to pin to a version? can we just install the latest stable?)

Using latest sounds OK with me, is there any risk?

pzhan9:
> shall we also update here ? https://github.com/pytorch/torchtitan/blob/main/.github/workflows/integration_test_4gpu_rl.yaml#L68
> 
> > (actually do we need to pin to a version? can we just install the latest stable?)
> 
> Using latest sounds OK with me, is there any risk?

@tianyu-l using latest stable is okay, I do not expect any risk. I updated the 3 places I found in the codebase. Please take one more look.

tianyu-l (approved):

PR #3461, opened 2026-05-30
[MoE] Add TP+EP config to Qwen3 MoE deterministic loss test (FSDP4+TP2+EP4)

yichuan-w:
Add a loss_compare.py invocation in integration_test_8gpu_models.yaml to verify DeepSeek V3 FSDP+EP produces deterministic losses, following the pattern from integration_test_8gpu_features.yaml. This ensures EP does not silently break numerical correctness.

Golden loss file has a placeholder value — CI will fail on first run and print the actual losses, which should be pasted back as the golden reference.


.github/workflows/integration_test_8gpu_models.yaml
tianyu-l, line 1: sorry, I forgot that we have qwen MoE test already in https://github.com/pytorch/torchtitan/blob/main/.github/workflows/integration_test_8gpu_features.yaml#L125-L137

The problem is that it's not running on EP. Instead of adding DSv3, could you modify the config there to test EP? Specifically, let's test "dense FSDP 4, TP 2" -> "sparse EFSDP 2, EP 4".

yichuan-w: i was confused by this too yesterday since there is already QwenMOE there, will add --parallelism.tensor_parallel_degree 2                           
    --parallelism.expert_parallel_degree 4 soon.

yichuan-w:
CI found a bug: Qwen3 MoE with TP+EP crashes due to `aten.histc`
  not supporting int64 in DTensor's fake tensor propagation. Filed
  as https://github.com/pytorch/torchtitan/actions/runs/26697144371/job/78683492198?pr=3461.

  Options:
  1. Drop TP for now (test FSDP+EP only), add TP back after the
  histc issue is fixed
  2. Wait for the fix before merging

  @tianyu-l what do you prefer?

tianyu-l:
https://github.com/pytorch/torchtitan/pull/3450 should fix it. Let's rebase after that.

tianyu-l:
https://github.com/pytorch/torchtitan/pull/3604 fixes another EP bug introduced recently. Please rebase after that, we should see loss change.

wwwjn:
#3604 just merged, you can rebase and get the training loss curve. Thank you!

tianyu-l:
oh gosh, now you need to wait for https://github.com/pytorch/torchtitan/pull/3648 @pianpwk

yichuan-w:
Thanks @wwwjn! Rebased onto latest `main` (past #3604) and regenerated the golden loss curve for the new FSDP+TP+EP config — the Qwen3 MoE deterministic loss check passes now ✅ (I also switched the `loss_compare.py` call to baseline-only mode by matching `--test-options` to `--baseline-options`, so it runs the TP2+EP4 config once and asserts against the reference curve).

Everything on my end is fixed and **8 GPU Feature Tests is green**. The only remaining red check is **8 GPU Model Tests**, which fails on `qwen3_5_moe_fsdp+tp+ep+pp` (`aten.embedding.default got mixed torch.Tensor and DTensor` in qwen3_5's `_prepare_multimodal_embeds`). That's a pre-existing failure on `main` and unrelated to this PR — this PR only touches the qwen3 MoE loss asset and the features workflow, not qwen3_5. Looks like #3648 (qwen3.5 tok_embeddings LocalMap region) should fix it; once that lands and I rebase, everything should be green and this is ready to merge.

tianyu-l:
https://github.com/pytorch/torchtitan/pull/3648 merged

yichuan-w:
@tianyu-l @wwwjn CI is fully green now (13 passing / 8 skipped / 0 failing) — ready for review/merge 🙏

For transparency, the earlier red runs were **not** caused by the substance of this PR:

1. **`qwen3_5_moe_fsdp+tp+ep+pp` (Model Tests)** — a Qwen3.5 `mixed torch.Tensor and DTensor` embedding error, unrelated to this change. Fixed upstream by #3648; resolved after rebasing onto latest `main`.
2. **ROCm MoE loss mismatch (Feature Tests)** — legitimately on me: this PR switches the Qwen3 MoE loss test from FSDP8 to FSDP4+TP2+EP4 for both arches, but I had only regenerated the CUDA golden. I've now updated the ROCm golden `tests/assets/losses/qwen3_moe_rocm_mi350x.txt` to the new EP-config baseline, and ROCm Feature Tests passes.
3. **Flux `HSDP+CP+Validation+Inference` (Model Tests, both CUDA & ROCm)** — `huggingface_hub 408 Client Error: Request Time-out` while downloading the validation parquet from the HF xet CDN. Pure infra/network flake, cleared on re-run.

The core change — Qwen3 MoE FSDP4+TP2+EP4 deterministic loss parity — passes on both CUDA and ROCm. Thanks!

tianyu-l (approved):

PR #3462, opened 2026-05-30
Faster doc mask for flex flash

drisspg:
Faster doc mask for flex flash

### What this does

Changing how we right `block_causal` to more capitalize no better codegen. `Block_causal` does not need generic document boundaries so we use this to produce much faster code in the fwd. We make use of the fact that things are packed. Feel free to look at pytorch code for specifics on why we are faster but this lets us do vectorization / vectorized loads. Which can be up to 2x perf on doc mask in general as we >> more work.  

One tangential  perf tweak I make is that I cap offset to a safe upper bound. This allows for static shapes instead of dynamic shapes which is produces generically faster codegen regardless of impl.

Local testing: 
Qwen3-32B, seq_len=8192, lbs=1, bf16, DP shard=-1 on 2x GB200, `--compile.enable`,
`--training.steps=100`. Steady-state averaged over the last 4 logged samples
(steps 70-100). `kv_docend` helper applied for FlexFlash packed.

W&B project (single-click compare):
<https://meta.wandb.io/drisspg/torchtitan-pr3462-doc-mask>

| backend                | doc mask        | tps    | tflops | mfu    | W&B run |
|---|---|---:|---:|---:|---|
| Triton FlexAttention   | standard         | 3221.3 | 784.3  | 31.37% | https://meta.wandb.io/drisspg/torchtitan-pr3462-doc-mask/runs/3ayg2gea |
| Triton FlexAttention   | packed_kv_docend | 3297.3 | 802.7  | 32.10% | https://meta.wandb.io/drisspg/torchtitan-pr3462-doc-mask/runs/tnp985eq |
| FlexFlash (cute_dsl)   | standard         | 3499.0 | 851.7  | 34.07% | https://meta.wandb.io/drisspg/torchtitan-pr3462-doc-mask/runs/vs9ucha3 |
| FlexFlash (cute_dsl)   | packed_kv_docend | 3533.0 | 860.1  | 34.40% | https://meta.wandb.io/drisspg/torchtitan-pr3462-doc-mask/runs/8yatbatx |

Deltas (packed_kv_docend vs standard):
- Triton FlexAttention: **+2.36% tps**, +0.73 pp MFU
- FlexFlash (cute_dsl, padded): **+0.97% tps**, +0.33 pp MFU

Note:
I need to land below before this pr to have a perf impact-> ttps://github.com/pytorch/pytorch/pull/184438

| backend | variant | mask create | fwd | bwd |
|---|---|---:|---:|---:|
| Triton FlexAttention | `standard_doc_ids` | 4.668 ms | 1.116 ms / 392.57 TF/s | 3.473 ms / 315.39 TF/s |
| Triton FlexAttention | `packed_kv_docend` | 3.814 ms (22.4% faster) | 1.086 ms / 403.52 TF/s (+10.95 TF/s, +2.8%) | 3.263 ms / 335.66 TF/s (+20.27 TF/s, +6.4%) |
| FlexFlash | `standard_doc_ids` | 4.912 ms | 0.867 ms / 505.25 TF/s | 1.799 ms / 608.90 TF/s |
| FlexFlash | `packed_kv_docend` | 3.439 ms (42.8% faster) | 0.772 ms / 567.19 TF/s (+61.94 TF/s, +12.3%) | 1.624 ms / 674.59 TF/s (+65.69 TF/s, +10.8%) |

^ on gb200

##  E2E traces:

| run | rank | local file | perfetto URL (overlap-split) |
|---|---:|---|---|
| FlexFlash standard (profile50)       | 0 | `outputs/profiling/pr3462_flexflash_standard_profile50/iteration_50/rank0_trace.pftrace`       | https://fburl.com/perfetto/f9q2ogpi |
| FlexFlash standard (profile50)       | 1 | `outputs/profiling/pr3462_flexflash_standard_profile50/iteration_50/rank1_trace.pftrace`       | https://fburl.com/perfetto/ssyu56ve |
| FlexFlash packed_kv_docend (profile50) | 0 | `outputs/profiling/pr3462_flexflash_kvdocend_profile50/iteration_50/rank0_trace.pftrace` | https://fburl.com/perfetto/s0k3av7n |
| FlexFlash packed_kv_docend (profile50) | 1 | `outputs/profiling/pr3462_flexflash_kvdocend_profile50/iteration_50/rank1_trace.pftrace` | https://fburl.com/perfetto/qkejg63k |


torchtitan/models/common/attention.py
```diff
@@ -337,6 +338,24 @@ def document_mask(
...
+
+    This assumes per-token positions reset to 0 at each packed document boundary
+    and then increase by 1 within the document.
+    """
+
+    def packed_causal_document_mask(
+        b: torch.Tensor, h: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
+    ) -> torch.Tensor:
+        doc_start = q_idx - positions[b, q_idx]
+        return (kv_idx >= doc_start) & (q_idx >= kv_idx)
```
fegin: Do we have do both document and causal here? Or can we still keep the composition way to the mask like document mask + causal mask?

drisspg: yeah we can mixin the other, and mask ccausal, I wrote this way https://github.com/pytorch/torchtitan/pull/3462/changes/BASE..efdbce31e23b83f569f1c19c8bc63d54bb3f77fe#diff-157976c1efe4e4728b2c0e370f5fc6e898cac6e84543976131555398f7d94749L251 because it seems we are mathcign against `block_causal` so why make it more generic than needed

drisspg: up to you I dont have a strong preference, the main thing is that this assumes a packing format which (what we already have)

drisspg: sorry ddint see we do and_masks later on will udpate

.gitignore
```diff
@@ -52,3 +52,5 @@ CLAUDE_CONTEXT/
 
 # Lychee (link checker) cache
 .lycheecache
+
+experiment_notes/
```
tianyu-l: not sure how generic this looks

torchtitan/models/common/attention.py
```diff
@@ -337,6 +338,23 @@ def document_mask(
...
+    """Creates a packed document mask to compose with a causal mask.
+
+    This assumes per-token positions reset to 0 at each packed document boundary
+    and then increase by 1 within the document.
+    """
+
+    def packed_document_mask(
+        b: torch.Tensor, h: torch.Tensor, q_idx: torch.Tensor, kv_idx: torch.Tensor
+    ) -> torch.Tensor:
+        return kv_idx >= q_idx - positions[b, q_idx]
```
tianyu-l: It seems this alone doesn't give per-doc bidirectional masking.
- Are you saying that this, combined with causal, gives the same things as per-document causal?
- I guess I understand the benefit is on perf side. But in general, what's the guidance to write efficient mask_mod than not?
- I think it's better to couple it with causal in one mask_mod; o/w itself is not really useful. "packed" is not accurate name, I think.
- It seems we are getting the same mask, thus same numerics, can we deprecate `get_document_mask_mod`.

.gitignore
drisspg: this was helpful for me, if you dont want this in gitignore I can remove but IMO i think this is harmless

torchtitan/models/common/attention.py
drisspg: 1.  yes
2. his is very tied to the implementation and which arch. For now the higher order bit is that using q_idx is cheaper than kv_idx for forward masking. The bwd is more robustt to this - and currently doesnt have the vectorized path enabled in flash that would let us do this more efficiently. The largest perf indicator is avoiding memory spills.
3. I had that and Chien wanted the opposite. Ragged or packed or varlen whatever name (I dont think this is user facing so I dont have a preference)
4. Forsure but  `get_document_mask_mod` is an impl detail and nice for testing `block-causal` is the user facing API

drisspg:
I did one mroe sanity check at bsz > 1 and the packed positons or offsets (prebake in the q_idx offset once at creation) both have big register spills at backward. We increase livenss of b_idx and we run over by like 1 register :(

Variants:
   1. OG / standard doc-id mask
   2. packed via positions: `kv_idx >= q_idx - positions[b, q_idx]`
   3. packed via offsets: precompute document-start offsets, then `kv_idx >= offsets[doc_id[q_idx]]`

 For `bsz=1`:

   | variant | mask creation | fwd | bwd |
   |---|---:|---:|---:|
   | standard doc ids | 0.55 ms | 0.62 ms | 1.75 ms |
   | packed positions | 0.47 ms | 0.40 ms | 1.98 ms |
   | packed static offsets | 0.50 ms | 0.40 ms | 1.78 ms |

   For `bsz=2`:

   | variant | mask creation | fwd | bwd |
   |---|---:|---:|---:|
   | standard doc ids | 0.88 ms | 1.08 ms | 3.51 ms |
   | packed positions | 0.78 ms | 0.81 ms | 3.97 ms |
   | packed 2D offsets | 0.83 ms | 0.78 ms | 4.50 ms |
   
   
   therefore I only enabled when bsz = 1 which IMO should be more standard for block causal

drisspg:
~Alright did some more tuning and wrote again and instead of getting faster fwd I balanced and got faster fwd (new with vectorized loads  pr landed in core) and indexing on kv for backward which is lane uniform. Re-runnin experiements~

Update the main summary with the results

fegin:
> I had that and Chien wanted the opposite. Ragged or packed or varlen whatever name (I dont think this is user facing so I dont have a preference)

I'm okay either way. It's just that the motivation to design a composition masks goes away. If we prefer just one single mask, let's just remove the usage of `and_mask` completely.

.gitignore
tianyu-l: what's the source of this name? If it's personal let's remove; o/w please add comment where it's coming from.

torchtitan/models/common/attention.py
```diff
@@ -337,6 +342,76 @@ def document_mask(
...
+) -> _mask_mod_signature:
+    """Creates a document mask to compose with a causal mask.
+
+    This uses the same convention as get_document_mask_mod: per-token positions
+    reset to 0 at each packed document boundary and then increase by 1 within the
+    document. For single-sample batches, the causal mask supplies the upper
+    bound, ``kv_idx <= q_idx``, and this mask supplies the lower bound,
+    ``doc_start[q_idx] <= kv_idx``. For multi-sample batches, the causal mask
+    supplies the lower bound, ``kv_idx <= q_idx``, and this mask supplies the
+    upper bound, ``q_idx <= doc_end[kv_idx]``. For example, a query in the next
```
tianyu-l: 1. What is single-sample vs. multi-sample, is single-sample batch size 1?
2. Is the difference of using `doc_start` vs. `doc_end` crucial?

torchtitan/models/common/attention.py
```diff
@@ -337,6 +342,76 @@ def document_mask(
     return document_mask
 
 
+def get_packed_document_mask_mod(
```
tianyu-l: Even the existing slow version of document mask_mod is designed to work with "packing" -- aka putting multiple document samples in a row. 
- `document_mask` didn't but `doc_ids = torch.cumsum((positions == 0).int(), dim=1) - 1` assumes.

Can we name it to something like
```suggestion
def get_efficient_causal_mask_mod_for_packed_document(
```

torchtitan/models/common/attention.py
```diff
@@ -337,6 +342,76 @@ def document_mask(
...
+    positions: torch.Tensor,
+) -> _mask_mod_signature:
+    """Creates a document mask to compose with a causal mask.
```
tianyu-l: Please mention that this is efficient (thus worth more effort), so that it is clear to reader why we have two document related masks.

torchtitan/models/common/attention.py
```diff
@@ -337,6 +342,76 @@ def document_mask(
...
+    bound, ``kv_idx <= q_idx``, and this mask supplies the lower bound,
+    ``doc_start[q_idx] <= kv_idx``. For multi-sample batches, the causal mask
+    supplies the lower bound, ``kv_idx <= q_idx``, and this mask supplies the
+    upper bound, ``q_idx <= doc_end[kv_idx]``. For example, a query in the next
+    document cannot attend to a key in the previous document because that key's
+    ``doc_end`` is before the query. The result is same-document causal masking.
+    This mask is not intended for non-causal use.
+    """
+    batch_size, seq_len = positions.shape
+    if batch_size == 1:
```
tianyu-l, line 365: mention that the if-else is a result of manual tuning

torchtitan/models/common/attention.py
```diff
@@ -37,6 +37,10 @@
 from torchtitan.protocols.module import Module
 
 
+def _round_up(value: int, multiple: int) -> int:
+    return ((value + multiple - 1) // multiple) * multiple
```
tianyu-l: seems a common util, e.g. https://github.com/pytorch/torchtitan/pull/3077 also wants to use
So maybe put in torchtitan/tools/utils.py (or inline it to where it's needed)

tianyu-l (approved):

torchtitan/models/common/attention.py
drisspg: 1. yeah clarified just bsz >1 
2. Yeah this basically determines if we index with q_idx (fast for the forward) or kv_idx (fast for the backward)

PR #3463, opened 2026-05-30
[Draft][CI] Migrate docker-builds to OSDC (ARC)

huydhn:
## What

Migrates torchtitan's CI image build (`docker-builds.yml`) to OSDC (ARC), mirroring pytorch/pytorch's `docker-builds.yml`:
- builds via a **remote BuildKit** running on OSDC (`driver: remote`),
- runs on the **release-runners** pool (exempt from the cache-enforcer ghcr.io egress block) inside `ghcr.io/actions/actions-runner:latest`,
- pushes to ECR and **mirrors to GHCR**, hash-tagged by `git rev-parse HEAD:.ci/docker`.

This is the foundation for moving the 8-GPU H100/A100 (and the shared `set-matrix.yaml`) consumers to `linux_job_v3` — they'll reference the GHCR-mirrored image this produces. That consumer migration is a **follow-up PR** (intentionally not included here, since its image reference depends on the registry/tag scheme settled here).

## ⚠️ Assumptions to verify (torchtitan cluster) — please confirm before merge
1. BuildKit endpoint `tcp://buildkitd-amd64.buildkit:1234` is reachable from torchtitan's runners.
2. The `release-runners` group / `mt-rel-l-x86iavx512-8-64` label exists for torchtitan.
3. A `GHCR_PAT` secret with push access exists.
4. `ECR_REPOSITORY` (`torchtitan/<name>`) and `GHCR_REPOSITORY` (`ghcr.io/pytorch/torchtitan-ci-image`) names are correct.
5. `role/arc` can push to the torchtitan ECR repo.
6. Cross-repo actions `pytorch/pytorch/.github/actions/{setup-linux,ecr-login}@main` are acceptable (they support `use-arc`; test-infra's `setup-linux` does not).

Kept as **draft** for review of these infra assumptions.

### DIFF
--- a/.github/workflows/docker-builds.yml
+++ b/.github/workflows/docker-builds.yml
@@ -1,5 +1,19 @@
 name: docker-builds
 
+# OSDC (ARC) migration of the torchtitan CI image build, mirroring
+# pytorch/pytorch's .github/workflows/docker-builds.yml:
+#   - builds via a *remote* BuildKit running on OSDC
+#   - runs on the release-runners pool (exempt from the cache-enforcer ghcr.io
+#     egress block) so the GHCR mirror step can push
+#   - pushes the image to ECR and mirrors it to GHCR (hash-tagged by .ci/docker)
+#
+# ASSUMPTIONS TO VERIFY for torchtitan's cluster (flagged for review):
+#   1. BuildKit endpoint `tcp://buildkitd-amd64.buildkit:1234` is reachable.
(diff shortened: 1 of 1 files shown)


huydhn:
Stale — superseded by #4875, which does the same migration against the current BuildKit setup.

#4875 also covers what this one did not: the `REMOTE_BUILDKIT` switch in `.ci/docker/build.sh`, `id-token: write` for the OIDC push to ECR, and `ciflow/docker` gating so the build is opt-in rather than running on every PR.

PR #3466, opened 2026-05-31
[spmd_types] add parallelism.spmd_backend config

pianpwk:
The initial thought for spmd_types + titan was to kill the legacy & full_dtensor paths at once. As managing that became harder (lots of bitwise configs, ongoing model/full_dtensor work), we settled on having it as a 3rd path in the short-term, under `parallelism.spmd_backend: Literal["default", "full_dtensor", "spmd"] = "default"`

This PR just adds the flag and modifies existing uses of `.full_dtensor`, banning spmd backend until it's enabled.

* #3278
* #3472
* #3468
* #3471
* #3253
* #3501
* __->__ #3466


torchtitan/config/configs.py
```diff
@@ -139,8 +139,14 @@ class ParallelismConfig:
     enable_sequence_parallel: bool = True
     """Whether to use SequenceParallel as part of tensor parallelism. Enabled by default."""
 
-    full_dtensor: bool = False
-    """Whether to use full DTensor mode."""
+    spmd_backend: Literal["default", "full_dtensor", "spmd"] = "default"
```
tianyu-l:
```suggestion
    spmd_backend: Literal["default", "full_dtensor", "spmd_types"] = "default"
```

torchtitan/config/configs.py
```diff
@@ -139,8 +139,14 @@ class ParallelismConfig:
...
+    """
+    SPMD backend selector.
+
+    - "default": use the existing TorchTitan parallelism paths.
+    - "full_dtensor": use the existing full DTensor path.
+    - "spmd": use the new local-tensor SPMD path.
```
tianyu-l: just say spmd_types

it's not only about "local-tensor"

torchtitan/config/configs.py
```diff
@@ -366,6 +377,9 @@ class DebugConfig:
     seed: int | None = None
     """Choose the base RNG seed used for training"""
 
+    spmd_typechecking: bool = False
+    """Enable global SPMD type checking for the local-tensor SPMD backend."""
```
tianyu-l: say something like
- only effective under spmd_backend=="spmd_types"

tianyu-l (approved):

fegin (approved):
LGTM, but I think we should consider to use StrEnum in the future.

PR #3467, opened 2026-05-31
[spmd_types] manual loss parallel CE

pianpwk:
Adds the spmd_types integration to titan part-by-part, starting from replacing the `torch.distributed.tensor.parallel.loss_parallel()` DTensor-based context manager with a manually written `_LossParallelCrossEntropy` custom autograd function.

Written to support uneven sharding (on last rank), IGNORE_INDEX, verified bitwise against loss_parallel() in test cases. The PR adds spmd_types requirement to titan, as it includes typecheck_forward & testing.

Not yet wired into the training path - the later stack will wire it for the spmd backend path, but likely leaves DTensor & plain-tensor on the loss_parallel() path for now.

* #3551
* #3550
* #3278
* #3472
* #3468
* #3471
* #3253
* #3501
* #3466
* __->__ #3467


torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+
+    Supports uneven vocab sharding (last TP rank may hold fewer classes) and
+    ``IGNORE_INDEX`` labels.  Forward uses three TP all-reduces (max, sumexp,
+    gather); backward is fused (NLL + log-softmax) with zero collectives.
+
+    All inputs and outputs are plain ``torch.Tensor`` (not DTensor).
+    """
+
+    @staticmethod
+    def typecheck_forward(
```
tianyu-l, line 59: Is `typecheck_forward` an spmd types native solution for autograd function?

This sounds a partial solution.
- In torchtitan we shouldn't do partial mesh check, instead we should do full mesh check.
- There are more than one way to handle things, including `local_spmd` field in Module. Is there way to consolidate?
 
Is it worth it to introduce a `Node(Configurable)` which
- `Module` inherits
- can still have in / out placements

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        spmd.assert_type(labels, {tp_group: spmd.I})
+        result = _LossParallelCrossEntropy.apply(
+            logits,
+            labels,
+            tp_group,
+            global_vocab_size,
+        )
+        output_type = dict(spmd.get_local_type(logits))
+        output_type[tp_group] = spmd.I
+        spmd.assert_type(result, output_type)
```
tianyu-l, line 79: My worry is that such autograd functions would not enforce typing on all mesh axes

torchtitan/components/loss.py
pianpwk: > Is typecheck_forward an spmd types native solution for autograd function?

yes, see spmd_types docs, the typechecker will look for this method
```
# Set of autograd.Function subclasses registered with a typecheck_forward
# staticmethod.  When .apply() is intercepted by __torch_function__,
# typecheck_forward is called INSTEAD of .apply().  It should assert_type()
# on inputs, call .apply() to execute, and assert_type() on the output.
```

> In torchtitan we shouldn't do partial mesh check, instead we should do full mesh check.

hmm, I feel this should be handled by earlier module boundaries where we declare activation placements, or the later lm_head sharding? (I should ensure that's explicit and asserted). This fn in theory shouldn't touch the DP/CP axes, so re-checking if DP/CP is on seems a bit messy.

> There are more than one way to handle things, including local_spmd field in Module. Is there way to consolidate?

They should serve different purposes here - local_spmd regions are still transparent to typechecking, we just switch to local typecheck. With custom autograd typecheck the type propagation is opaque, so this is a black-box where `typecheck_forward` is trusted by the typechecker (no typecheck).

> Node(Configurable)

this seems... useful for chunked loss, and maybe other contexts outside spmd types too. But I'm not sure about the scope of this - in MoE & VL there are some forward sections & functions that require local/no-typecheck - would the plan be to refactor all of them to Node

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
     )
 
 
+def _vocab_shard_size_and_offset(
```
tianyu-l: the file is already big, maybe make this inline if it's only used for loss parallel

torchtitan/components/loss.py
tianyu-l: OK that sounds convincing!

> in MoE & VL there are some forward sections & functions that require local/no-typecheck - would the plan be to refactor all of them to Node

that seems a different topic, I think we should kill the non-DTensor code as much as possible, outside the modules that we apply local_spmd on.
@shuhuayu can also help on this.

tianyu-l (approved):
The only concern is that this code not being actually integrated, but I'll stamp.

.ci/docker/requirements.txt
```diff
@@ -8,3 +8,4 @@ tokenizers >= 0.15.0
 safetensors
 einops
 pillow
+spmd_types==0.2.0
```
fegin, line 11: The README.md still uses `pip install -r requirements.txt`. So we should also add this to the general requirements.txt

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        spmd.assert_type(result, output_type)
+        return result
+
+    @staticmethod
+    # pyrefly: ignore [bad-override]
+    def forward(
+        ctx,
+        logits: torch.Tensor,
+        labels: torch.Tensor,
+        tp_group: dist.ProcessGroup,
```
fegin, line 88: Can we take a DeviceMesh? Most places in trainer still take DeviceMesh. So it is better to be consistent. Or is there any blocker?

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+    """
+    Vocab-parallel cross-entropy on plain (non-DTensor) local tensors.
+
+    Replaces ``torch.distributed.tensor.parallel.loss_parallel()`` with an
+    explicit autograd Function so that SPMD code can operate on local tensors
+    and process groups directly, without the DTensor-based context manager.
+
+    Supports uneven vocab sharding (last TP rank may hold fewer classes) and
+    ``IGNORE_INDEX`` labels.  Forward uses three TP all-reduces (max, sumexp,
+    gather); backward is fused (NLL + log-softmax) with zero collectives.
```
fegin: nit: use three TP all-reduce to reduce the results of three oerators.

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        log_probs = shifted - torch.log(shifted_sumexp)
+
+        # Mask labels outside this vocab shard; the TP all-reduce below selects
+        # the owner rank's log probability for each target token.
+        safe_labels = torch.where(labels_1d != IGNORE_INDEX, labels_1d, 0)
+        out_of_range = (safe_labels < vocab_start) | (
+            safe_labels >= vocab_start + local_vocab_size
+        )
+        local_labels = safe_labels - vocab_start
+        local_labels = local_labels.clone()
```
fegin: Is this clone redundant? `safe_labels - vocab_start` creates a new tensor. So `clone()` seems to be redundant.

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        local_labels[out_of_range] = 0
+
+        local_result = torch.gather(log_probs, -1, local_labels.unsqueeze(-1))
+        local_result = local_result.clone()
```
fegin: ditto

torchtitan/components/loss.py
```diff
@@ -36,6 +38,151 @@ def cross_entropy_loss(pred: torch.Tensor, labels: torch.Tensor) -> torch.Tensor
...
+        ctx,
+        grad_output: torch.Tensor,
+    ) -> tuple[torch.Tensor, None, None, None]:
+        log_probs, labels_1d = ctx.saved_tensors
+        safe_labels = torch.where(labels_1d != IGNORE_INDEX, labels_1d, 0)
+        out_of_range = (safe_labels < ctx.vocab_start) | (
+            safe_labels >= ctx.vocab_start + ctx.local_vocab_size
+        )
+        local_labels = safe_labels - ctx.vocab_start
+        local_labels = local_labels.clone()
```
fegin: same redundant clone()?

fegin (approved):
LGTM, some change requests but not critical.

.ci/docker/requirements.txt
pianpwk: I think it's just a symlink to this one?

torchtitan/components/loss.py
pianpwk: I think once spmd infra lands we can just call current_mesh() and stop threading this. I wasn't planning to move full DTensor to this as part of the stack yet

PR #3468, opened 2026-05-31
[spmd_types] embedding vocab parallel

pianpwk:
Adds a custom vocab-parallel Embedding module, to be wired for spmd_types and DTensor backend. Uses a local SPMD / local_map region, avoiding DTensor dependence on MaskPartial.

* #3278
* #3472
* #3632
* #3631
* __->__ #3468


torchtitan/models/common/decoder_sharding.py
```diff
@@ -302,14 +302,19 @@ def set_decoder_sharding_config(
         if enable_sp
         else dense_activation_placement(tp=spmd.I)
     )
+    embed_src = dense_activation_placement(tp=spmd.P)
```
tianyu-l: a quick in-person chat would convince me this

pianpwk: my offline attempt is:

we shard along vocab dimension, so each rank offsets indices to be between [0, shard_size] and creating a mask for out-of-bound indices (the index values are clamped before F.embedding)

The output is then (embed_out_with_offset_indices * mask_for_inbound_indices), which is correct value if inbound, 0 if outbound.

The results can be allreduced, getting 0 embedding values for all ranks except 1 ?

torchtitan/models/common/decoder_sharding.py
tianyu-l:
```suggestion
    embed_out_dst = (
```

torchtitan/models/common/decoder_sharding.py
```diff
@@ -293,14 +293,19 @@ def set_decoder_sharding_config(
...
```
tianyu-l:
```suggestion
    embed_out_src = dense_activation_placement(tp=spmd.P)
```

torchtitan/models/common/decoder_sharding.py
tianyu-l: sounds good

torchtitan/models/common/embedding.py
```diff
@@ -0,0 +1,64 @@
...
+import torch.distributed as dist
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.distributed.tensor import DTensor
+
+from torchtitan.distributed.spmd_types import current_mesh, mesh_size
+from torchtitan.protocols.module import Module
+
+
+class Embedding(nn.Embedding, Module):
```
tianyu-l, line 24:
```suggestion
class VocabParallelEmbedding(nn.Embedding, Module):
```

torchtitan/models/common/decoder_sharding.py
```diff
@@ -293,14 +293,19 @@ def set_decoder_sharding_config(
...
         },
         in_src_shardings={"input": embed_input},
         in_dst_shardings={"input": embed_input},
-        out_src_shardings=embed_out,
+        out_src_shardings=embed_src,
         out_dst_shardings=embed_out,
+        # spmd_types backend relies on local SPMD + manual vocab-parallel embedding impl.
+        local_map=LocalMapConfig(in_grad_placements=(None,))
+        if spmd_backend == "spmd_types"
+        else None,
```
tianyu-l: what happens to DTensor path? I felt the VocabParallelEmbedding should be local map region?

pianpwk: DTensor path should still use MaskPartial, we did this because spmd types doesn't have it

torchtitan/models/common/embedding.py
pianpwk: is this a bit misleading? it seems we'd use it for all paths, including non-parallel, or DTensor paths that don't use the custom impl and rely on MaskPartial.

There's also the option of restoring nn_modules.py Embedding wrapper and swapping between this/that, but that also seems messy.

tianyu-l: I think we should use DTensor local_map for this `VocabParallelEmbedding`, and migrate away from `MaskPartial` stuff.

torchtitan/models/common/decoder_sharding.py
tianyu-l: see above comment

pianpwk:
updated DTensor to use local, non-MaskPartial path

pianpwk:
switched to use local region for all DTensor / legacy / spmd_types

torchtitan/models/common/decoder_sharding.py
```diff
@@ -256,12 +256,21 @@ def set_decoder_sharding_config(
     )
     loss_tp = spmd.S(-1) if loss_parallel else spmd.I
 
+    embed_out_src = dense_activation_placement(tp=spmd.P)
+    embed_input = SpmdLayout(
```
tianyu-l: curious why not reuse `dense_activation_placement`?

torchtitan/models/common/embedding.py
tianyu-l: please add a TODO if you don't want to do renaming in this PR.

torchtitan/models/common/embedding.py
```diff
@@ -0,0 +1,77 @@
...
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        num_embeddings: int
+        embedding_dim: int
+
+    def __init__(self, config: Config):
+        super().__init__(config.num_embeddings, config.embedding_dim)
+        self.tp_group: dist.ProcessGroup | None = None
+
+    def parallelize(self, parallel_dims: "ParallelDims") -> None:
```
tianyu-l, line 39: add a TODO to delete this?

torchtitan/models/common/embedding.py
```diff
@@ -0,0 +1,77 @@
...
+    """Configurable embedding with optional local vocab-parallel execution."""
+
+    @dataclass(kw_only=True, slots=True)
+    class Config(Module.Config):
+        num_embeddings: int
+        embedding_dim: int
+
+    def __init__(self, config: Config):
+        super().__init__(config.num_embeddings, config.embedding_dim)
+        self.tp_group: dist.ProcessGroup | None = None
```
tianyu-l, line 37: I think this might be a case where you might want global spmd on outer axes (DP, CP) and local spmd on inner axis (TP)? Not a strong motivation though.

torchtitan/models/common/nn_modules.py
```diff
@@ -75,18 +75,6 @@ def __init__(self, config: Config):
         )
 
 
-class Embedding(nn.Embedding, Module):
```
tianyu-l, line 78: This means all torchtitan models are using the new VocabParallelEmbedding regardless of parallelism config, right?

tianyu-l (approved):

PR #3472, opened 2026-05-31
[spmd_types] SPMD support for ChunkedCELoss

pianpwk:
Moves the seq-dim allgather onto the pre-lm-head norm, making it ShardingConfig-based. Annotates types in ChunkedCELoss impl.

* #3690
* #3673
* #3655
* #3664
* #3654
* #3278
* __->__ #3472


torchtitan/components/loss.py
tianyu-l, line 1: +229 -80

didn't read carefully, where does the added complexity mainly come from?

pianpwk: cut it down a bit, but the remaining is:
- refactor chunk loss process under chunked_loss_and_grad, as a local SPMD region
- typechecking for
  - total_loss and per-chunk-loss values to be P@DP/CP; total_loss on allocation (I think we write like this for compile?), and per-chunk-loss needs reinterpret(V->P), as local SPMD doesn't infer this
  - grad_buffer after exiting local region (just copy hidden states)
  - _DecoderOutputGradientBackProp (~passthrough)

pianpwk: though the churn is mostly refactor for readability as a standalone file. But maybe you think it got worse, happy to go either way.

torchtitan/components/loss.py
```diff
@@ -634,6 +738,7 @@ def _gradient_backprop(
         )
 
 
+@spmd.register_autograd_function
```
tianyu-l: There won't be anything fwd happening after this autograd's fwd (as loss is already computed), so with or without typechecking it doesn't give any more safety.

Are we adding this because
- it's still has a fwd function
- and spmd_typechecking would fail if we do nothing

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+                logits = lm_head(h_chunk)
+                if current_mesh() is not None and self.loss_parallel:
+                    mesh = current_mesh()
+                    assert mesh is not None
+                    chunk_loss = _LossParallelCrossEntropy.apply(
+                        logits,
+                        label_chunk,
+                        mesh.get_group("tp"),
+                        getattr(lm_head, "out_features"),
+                    )
```
tianyu-l: Let's
- not have this if-else, and delete all `loss_parallel` flag in torchtitan
- change self.fn to use the autograd version of LossParallel
- mesh.get_group should be put inside `_LossParallelCrossEntropy` and it should work even when TP is not enabled.

FYI we are moving towards making both ChunkedLossWrapper and the inner loss function to be configurable https://github.com/pytorch/torchtitan/pull/3477#discussion_r3337716804
- the vocab size should come from config of LossParallel, to decouple from ChunkedLossWrapper
- for now I don't mind you hardcode this class into chunked loss + tp loss parallel to unblock

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+        return chunk_loss
+
+    def chunked_loss_and_grad(
+        self,
+        hidden_states: torch.Tensor,
+        labels: torch.Tensor,
+        *,
+        global_valid_tokens: torch.Tensor | None,
+        fsdp_enabled: bool,
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
```
tianyu-l: add docstring for return values

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+        to bridge backward into the decoder graph.
+        """
+        lm_head = self.lm_head
+        assert lm_head is not None, "Set lm_head before calling ChunkedCELoss"
+
+        requires_grad = hidden_states.requires_grad
+        with spmd.local():
+            h_detached = hidden_states.detach().requires_grad_(requires_grad)
+            h_chunks, label_chunks = self.chunk_states_and_labels(h_detached, labels)
+            grad_accumulator = GradAccumulator(
```
tianyu-l: do we need to instantiate this GradAccumulator even when `requires_grad=False`?

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+                zip(h_chunks, label_chunks, strict=True)
+            ):
+                if fsdp_enabled and i == last_idx:
+                    lm_head.set_requires_gradient_sync(  # pyrefly: ignore[not-callable]
+                        True, recurse=False
+                    )
+
+                logits = lm_head(h_chunk)
+                if current_mesh() is not None and self.loss_parallel:
+                    mesh = current_mesh()
```
tianyu-l: nit: first assign then check, instead of call the function twice

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+                    chunk_loss = self.fn(logits, label_chunk)
+                if global_valid_tokens is not None:
+                    chunk_loss = chunk_loss / global_valid_tokens
+                chunk_loss = self.reinterpret_data_axes(chunk_loss)
+                total_loss = total_loss + chunk_loss.detach()
+
+                if requires_grad:
+                    backward_context = (
+                        spmd.no_typecheck()
+                        if current_mesh() is not None
```
tianyu-l: shouldn't use current_mesh() being None or not to tell if we should do type checking

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+                        else contextlib.nullcontext()
+                    )
+                    with backward_context:
+                        chunk_loss.backward()
+                        assert h_chunk.grad is not None
+                        grad_accumulator.add(h_chunk.grad)
+                        h_chunk.grad = None
+
+            grad_buffer = (
```
tianyu-l: let's not invent multiple names for the same thing
```suggestion
            accumulated_grad = (
```

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+        This is the memory-saving inner loop. It detaches hidden states, runs
+        lm_head and CE one local sequence chunk at a time, accumulates the
+        detached scalar loss, and records the per-chunk hidden gradients needed
+        to bridge backward into the decoder graph.
+        """
+        lm_head = self.lm_head
+        assert lm_head is not None, "Set lm_head before calling ChunkedCELoss"
+
+        requires_grad = hidden_states.requires_grad
+        with spmd.local():
```
tianyu-l, line 571: > refactor chunk loss process under chunked_loss_and_grad, as a local SPMD region

I guess this is where you mentioned that this class doesn't support Module configs such as in / out sharding and local_spmd.

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
             return wrapped(t)
 
-        # ``detach`` + ``requires_grad_`` makes each chunk a leaf so it
-        # accumulates ``.grad`` for ``GradAccumulator``.
+        requires_grad = hidden_states.requires_grad
         h_chunks = [
             c.detach().requires_grad_(requires_grad) for c in _chunk(hidden_states)
         ]
         label_chunks = list(_chunk(labels))
+        return h_chunks, label_chunks
```
tianyu-l: what's the local spmd types of these values?

The local_map was not used with enable_spmdtype flag.

torchtitan/components/loss.py
```diff
@@ -634,6 +738,7 @@ def _gradient_backprop(
         )
 
 
+@spmd.register_autograd_function
 class _DecoderOutputGradientBackProp(torch.autograd.Function):
```
tianyu-l, line 674: also need to type check https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/graph_trainer/chunked_loss.py for graph_trainer

torchtitan/components/loss.py
tianyu-l: I read all the code. I can see what values are added in this PR. It's just becoming a monster .. which sounds OK for now.

torchtitan/components/loss.py
ezyang, line 571: You should use `local_map` here right?

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
-                    True, recurse=False
-                )
+        The chunk loop accumulates a scalar sum over the local batch/sequence
+        shard. Under local SPMD typecheck this is Partial on data axes,
+        invariant on TP, matching the eventual ChunkedCELoss output.
+        """
+        total_loss = reference.new_zeros((), dtype=torch.float32)
+        mesh = current_mesh()
+        if mesh is None or not spmd.is_type_checking():
+            return total_loss
```
ezyang: I get very nervous about code like this because it is easy to end up with divergent behavior with/without spmd type checking

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
-        for i, (h_chunk, label_chunk) in enumerate(zip(h_chunks, label_chunks)):
-            if fsdp_enabled and i == last_idx:
-                lm_head.set_requires_gradient_sync(  # pyrefly: ignore[not-callable]
-                    True, recurse=False
-                )
+        The chunk loop accumulates a scalar sum over the local batch/sequence
+        shard. Under local SPMD typecheck this is Partial on data axes,
+        invariant on TP, matching the eventual ChunkedCELoss output.
+        """
+        total_loss = reference.new_zeros((), dtype=torch.float32)
```
ezyang: use `mutate_type` instead

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+            if mesh_size(axis_name) == 1:
+                continue
+            chunk_loss = spmd.reinterpret(
+                chunk_loss,
+                mesh.get_group(axis_name),
+                src=spmd.V,
+                dst=spmd.P,
+                expert_mode=True,
+            )
+        return chunk_loss
```
ezyang: This looks wrong

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+        for axis_name in ("dp", "cp"):
+            if mesh_size(axis_name) == 1:
+                continue
+            chunk_loss = spmd.reinterpret(
+                chunk_loss,
+                mesh.get_group(axis_name),
+                src=spmd.V,
+                dst=spmd.P,
+                expert_mode=True,
+            )
```
ezyang: I'd deeply prefer if we adjust the SPMD types API to be nicer here. Something like

```
spmd.assert_type(chunk_loss, {"dp": spmd.P, "cp": spmd.P})
```

I think this should correctly work when you have V and do the conversion. (I could be misremembering though!)

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
+                else:
+                    chunk_loss = self.fn(logits, label_chunk)
+                if global_valid_tokens is not None:
+                    chunk_loss = chunk_loss / global_valid_tokens
+                chunk_loss = self.reinterpret_data_axes(chunk_loss)
+                total_loss = total_loss + chunk_loss.detach()
+
+                if requires_grad:
+                    backward_context = (
+                        spmd.no_typecheck()
```
ezyang: You can just unconditionally no typecheck right

torchtitan/components/loss.py
```diff
@@ -556,60 +530,190 @@ def _chunk(t):
...
-            logits = lm_head(h_chunk)
+            if fsdp_enabled:
+                lm_head.set_reshard_after_forward(True)
+                lm_head.set_reshard_after_backward(True)
+                lm_head.set_requires_gradient_sync(True, recurse=False)
+                lm_head.reshard()
+
+        if grad_buffer is not None and current_mesh() is not None:
+            spmd.assert_type(
+                grad_buffer,
```
ezyang: probably would have been better to have gotten this to propagate somehow (e.g., always assert types when coming out of no typecheck regions)

torchtitan/components/loss.py
```diff
@@ -646,6 +751,21 @@ class _DecoderOutputGradientBackProp(torch.autograd.Function):
...
+        accumulated_grad: torch.Tensor,
+        loss: torch.Tensor,
+    ) -> torch.Tensor:
+        # The hidden activation type depends on the model's sharding config;
+        # we'll defer hidden_states typechecking to previous module boundaries.
+        result = _DecoderOutputGradientBackProp.apply(
+            hidden_states, accumulated_grad, loss
+        )
+        if has_local_type(loss):
+            spmd.assert_type(result, spmd.get_local_type(loss))
```
ezyang: not needed with `local_map`

ezyang (dismissed):
.

torchtitan/components/loss.py
pianpwk: > spmd_typechecking would fail if we do nothing

this, and I remember the input type interaction is weird (I, non-I) so local spmd doesn't pass

torchtitan/components/loss.py
pianpwk: cleaned it up, removed the refactor

torchtitan/components/loss.py
pianpwk: I think @fegin will look at spmd_types + graph trainer, it is still on legacy backend IIRC

torchtitan/components/loss.py
pianpwk: I feel like for uses like this, the local() + assert_type_like is cleaner, than to refactor the body as a helper, and write out the full mesh output placements to feed to local map.

I'm pleasantly surprised spmd.local_map works with DTensor inputs though?

pianpwk: But I think longer-term cleanup (for DTensor backend too) is to make most of the loss region local SPMD, there's so many to_local/local_map in this file

torchtitan/components/loss.py
```diff
@@ -583,6 +580,12 @@ def _chunk(t):
                 lm_head.set_reshard_after_backward(False)
                 lm_head.set_requires_gradient_sync(False, recurse=False)
 
+            # TODO(pianpwk): migrate DTensor backend off pytorch's loss_parallel to simplify this.
+            use_loss_parallel_autograd_fn = (
```
tianyu-l: why introducing this flag at all?

The order of this PR and next seems reversed. We should first replace DTensor `loss_parallel` by autograd function.

pianpwk: Cleaned this up, moved that PR to bottom. Should be a bit more in one place now (DTensor & spmd_types), under cross_entropy_loss method

torchtitan/components/loss.py
```diff
@@ -677,3 +711,17 @@ def backward(  # pyrefly: ignore[bad-override]
...
         return accumulated_grad, None, None
+
+    @staticmethod
+    def typecheck_forward(
+        hidden_states: torch.Tensor,
+        accumulated_grad: torch.Tensor,
+        loss: torch.Tensor,
+    ) -> torch.Tensor:
+        # The hidden activation type depends on the model's sharding config;
+        # we'll defer hidden_states typechecking to previous module boundaries.
```
tianyu-l: what are the possibilities?

torchtitan/components/loss.py
```diff
@@ -615,6 +642,12 @@ def _chunk(t):
             assert grad_accumulator is not None
             accumulated_grad = grad_accumulator.result().to(hidden_states.dtype)
 
+        if get_spmd_backend() == "spmd_types":
+            spmd.assert_type(
+                accumulated_grad,
+                {"dp": spmd.V, "cp": spmd.V, "tp": spmd.R},
+                spmd.PartitionSpec("dp", "cp", None),
+            )
```
tianyu-l: This is gradient, why would it need a type?
Also, why "tp" dim is spmd.R?

torchtitan/trainer.py
```diff
@@ -441,6 +441,8 @@ def __init__(self, config: Config):
         # Non-PP: single model part always has lm_head.
         # PP: only the last stage has lm_head; non-last stages skip this.
         if isinstance(self.loss_fn, ChunkedCELoss):
+            assert isinstance(self.model_config, Decoder.Config)
+            self.loss_fn.global_vocab_size = self.model_config.vocab_size
```
tianyu-l: global_vocab_size should be a config of `ChunkedCELoss`, set in config_registry.py's not rather than patched in trainer

torchtitan/components/loss.py
pianpwk: hidden_states is directly from decoder, so we'd have to condition this on SP (`I or S(1)`)

torchtitan/components/loss.py
pianpwk: fair point, removed typechecking for this

torchtitan/trainer.py
pianpwk: hardcoded values in model registries

torchtitan/models/qwen3/config_registry.py
```diff
@@ -26,7 +26,7 @@
 
 def qwen3_debugmodel() -> Trainer.Config:
     return Trainer.Config(
-        loss=ChunkedCELoss.Config(),
+        loss=ChunkedCELoss.Config(global_vocab_size=2048),
```
tianyu-l: should come from model_spec rather than hardcoding

tianyu-l (approved):
see https://github.com/pytorch/torchtitan/pull/3472#discussion_r3437750625
let's not hardcode vocab_size

torchtitan/models/common/decoder.py
```diff
@@ -142,6 +143,10 @@ def update_from_config(
                     "Weight tying is not supported with Pipeline Parallel."
                 )
 
+            loss_config = getattr(config, "loss", None)
+            if isinstance(loss_config, ChunkedCELoss.Config):
+                loss_config.global_vocab_size = self.vocab_size
```
tianyu-l, line 150: This function is to give CLI config a chance to override python file config. The vocab size shouldn't be adjusted by CLI, so we could've set it entirely in config_registry files.

If you think the radius is too large, I'm fine with this for now but please add a NOTE / TODO.

PR #3474, opened 2026-06-01
  [RL] Add deterministic loss guard for GRPO training

### LINKED ISSUE #2691: [RL] Add CI to guard on numerics change
We should aim for adding the numerics using the same pattern with pre-training side CI:

1. Locally check our model numerics matches vllm's native model
2. Enable deterministic mode. Run with specific task and getting training metrics , eg reward , loss. Record these metrics in a file.
3. Run CI to compare against the loss in the file.

yichuan-w:
[RL] Add deterministic loss guard for GRPO training

  PR body：

  ## Summary
  - Add `rl_grpo_qwen3_0_6b_loss_guard` config with seed=42, deterministic mode, and
  TensorBoard enabled
  - Add RL-specific `loss_compare.py` script that runs training, extracts `loss/mean` from
  TensorBoard, and compares against a committed reference file
  - Add reference losses verified bit-identical across two independent H100 runs
  - Add loss guard CI step to the H100 RL workflow

  Follows the same pattern as pre-training loss guard (`scripts/loss_compare.py` +
  `tests/assets/losses/`).

  Closes https://github.com/pytorch/torchtitan/issues/2691

  ## Test plan
  - [x] Ran `rl_grpo_qwen3_0_6b_loss_guard` config twice on 4x H100, losses are
  bit-identical
  - [x] `--assert-equal` passes against the committed reference file


yichuan-w:
The reference loss file currently contains losses from a local H100 run (verified
  deterministic across two runs). The CI loss values still need to be updated once the
  H100 CI runners become available — the RL CI job has been stuck in pending. Will update
  the reference file with CI-environment losses once it runs.
  
  
  The RL H100 CI (integration_test_8gpu_rl_h100) has been stuck
  in pending — it requires the ciflow/rl label and H100 runners which seem to have limited
  availability.

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +191,72 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
...
 
 
+def rl_grpo_qwen3_0_6b_loss_guard() -> RLTrainer.Config:
+    """Deterministic GRPO config for loss guard CI (4 GPUs: 2 gen + 2 train).
+
+    Runs with fixed seed, deterministic ops, and TensorBoard logging so that
+    training losses can be extracted and compared against a committed reference
+    file for regression detection.
+    """
+    debug = DebugConfig(batch_invariant=True, deterministic=True, seed=42)
```
shuhuayu: why do we need `batch_invariant` for this task?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +191,72 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
...
+    group_size = 8
+    return RLTrainer.Config(
+        model_spec=model_registry("0.6B", attn_backend="varlen"),
+        hf_assets_path="torchtitan/experiments/rl/example_checkpoint/Qwen3-0.6B",
+        num_steps=10,
```
shuhuayu, line 277: We can do 100 steps to be consistent with the pre-train side.

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +191,72 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
...
+        num_prompts_per_step=5,
+        num_validation_samples=20,
```
shuhuayu, line 279: For this particular loss guard, we do not need validation? setting it as 0 can save some time.

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,282 @@
...
+        "-m", "torchtitan.experiments.rl.grpo",
+        "--module", "rl",
+        "--config", config,
+        f"--hf_assets_path={hf_assets_path}",
+        f"--dump_folder={dump_folder}",
+    ]
+
+    log_print(f"Running: {' '.join(cmd)}")
+
+    repo_root = os.path.dirname(
```
shuhuayu: is `repo_root` necessary? can we drop the `cwd` option?

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,282 @@
...
+    args = parser.parse_args()
+
+    export_result = args.export_result or None
+    import_result = args.import_result or None
+
+    if args.assert_equal and not import_result:
+        parser.error("--assert-equal requires --import-result")
+
+    if import_result and not args.assert_equal:
+        parser.error("--import-result requires --assert-equal")
```
shuhuayu: this seems unnecessary.

shuhuayu (changes requested):

yichuan-w:
The RL loss guard CI job fails during uv pip install -e . with:

  error: could not create 'torchtitan.egg-info': Permission denied

  The Docker container runs under a user that doesn't have write
  permission to the repo directory (/pytorch/torchtitan/). This is
  an infra issue, not related to the code changes.

  cc @tianyu-l @wconstab @huydhn

torchtitan/experiments/rl/tests/assets/losses/rl_grpo_cuda.txt
wwwjn, line 1: Put this under `experiments/rl/tests/assets/losses/rl_grpo_cuda.txt`, ideally we should put rl related files under rl folder, and avoid touch main

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,277 @@
...
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Compare RL training losses against a committed reference file for regression
+detection.  Follows the same pattern as ``scripts/loss_compare.py`` for
```
wwwjn, line 11: Maybe I missed something, why we don't slightly modify the scripts/loss_compare.py to make it support launching `python grpo.py`, will reuse same us a lot of lines of code?

torchtitan/experiments/rl/config_registry.py
```diff
@@ -201,6 +191,69 @@ def rl_grpo_qwen3_14b() -> RLTrainer.Config:
     )
 
 
+def rl_grpo_qwen3_0_6b_loss_guard() -> RLTrainer.Config:
```
wwwjn:
```suggestion
def rl_grpo_qwen3_0_6b_batch_invariant() -> RLTrainer.Config:
```

This is a normal config and can be run not only in CI, we can give it a general name by the config content itself

wwwjn: I just realize there's a `rl_grpo_qwen3_0_6b_batch_invariant`, why adding a new one?

wwwjn (commented):
Thanks for working on this!

yichuan-w:
Verified locally on H100x4: two runs of rl_grpo_qwen3_0_6b_loss_guard (100 steps, deterministic=True, seed=42, no batch_invariant) produce bit-identical losses (full-precision repr() comparison). Removed batch_invariant as suggested.

Note: this works for the current single-turn SumDigits workload where vLLM batching is deterministic given a fixed seed. For more dynamic workloads (e.g., multi-turn with variable sequence lengths), batch_invariant may still be needed; we can revisit if the loss guard becomes flaky.

torchtitan/experiments/rl/scripts/loss_compare.py
yichuan-w: The existing loss comparison utility is specific to the core torchrun workflow, while RL training uses a different launcher and logging setup; I'd prefer not to introduce RL-specific logic into core, but can help extract shared utilities if useful.

yichuan-w:
also verified locally on H100x4: two independent runs using rl_grpo_qwen3_0_6b_batch_invariant (CLI overrides: seed=42, 100 steps, TensorBoard enabled, validation disabled) produce bit-identical losses (repr() full-precision comparison, 100 steps, 0 diffs).

This is trivial since w/o batch_invariant is working well, just document here

torchtitan/experiments/rl/scripts/loss_compare.py
shuhuayu: yeah, i think we can use a rl specific one for now, and move it to core once rl graduates.

torchtitan/experiments/rl/grpo.py
```diff
@@ -674,6 +691,8 @@ async def validate(self) -> list[m.Metric]:
         """
         t_validate_start = time.perf_counter()
         num_samples = self.config.num_validation_samples
+        if num_samples == 0:
```
shuhuayu: maybe add an inline comment on this new use case where `num_samples = 0` is not a bug.

shuhuayu (approved):

torchtitan/experiments/rl/scripts/loss_compare.py
wwwjn: > not to introduce RL-specific logic into core 

That sounds fair to me, thanks

.github/workflows/integration_test_8gpu_rl_h100.yaml
```diff
@@ -96,3 +96,11 @@ jobs:
         python -m torchtitan.experiments.rl.tests.integration_tests \
           $RUNNER_TEMP/artifacts-to-be-uploaded --test_suite h100 --ngpu 4 \
           --hf_assets_path "$MODEL_PATH"
+
+        # RL Loss Guard: deterministic loss regression check
+        LOSS_FILE="torchtitan/experiments/rl/tests/assets/losses/rl_grpo_cuda.txt"
+        python torchtitan/experiments/rl/scripts/loss_compare.py \
+          --hf-assets-path "$MODEL_PATH" \
```
wwwjn, line 119: Worth explicitly set  `--config rl_grpo_qwen3_0_6b_batch_invariant` in the command, otherwise if it's using  `rl_grpo_qwen3_0_6b` the loss won't match

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,283 @@
...
+) -> None:
+    """Run RL GRPO training as a subprocess."""
+    cmd = [
+        sys.executable,
+        "-m", "torchtitan.experiments.rl.grpo",
+        "--module", "rl",
+        "--config", config,
+        f"--hf_assets_path={hf_assets_path}",
+        f"--dump_folder={dump_folder}",
+        "--trainer.debug.seed=42",
```
wwwjn, line 130: If it's for loss comparison purpose, do we need `deterministic` flag here?  https://github.com/pytorch/torchtitan/blob/refs/heads/main/torchtitan/experiments/rl/config_registry.py#L209

Only set seed here is not enough to make any config run-to-run determinism, we'll need a determinism flag to make it run to run determinism. I would suggest `determinism` is needed in loss comparision

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,285 @@
...
+       --export-result tests/assets/losses/rl_grpo_cuda.txt
+
+2. Assert losses match reference (CI mode):
+   python torchtitan/experiments/rl/scripts/loss_compare.py \\
+       --hf-assets-path /path/to/Qwen3-0.6B \\
+       --dump-folder /tmp/rl_loss_guard \\
+       --import-result tests/assets/losses/rl_grpo_cuda.txt \\
+       --assert-equal
+
+3. Run training only (no comparison):
```
wwwjn: Why we need to use this script to run training only?

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,285 @@
...
+    else:
+        print(LOG_PREFIX)
+
+
+# ---------------------------------------------------------------------------
+# TensorBoard extraction
+# ---------------------------------------------------------------------------
+
+
+def extract_losses_from_tensorboard(tb_dir: str) -> dict[int, float]:
```
wwwjn, line 63: Can we import this from main loss_compare.py?

torchtitan/experiments/rl/scripts/loss_compare.py
```diff
@@ -0,0 +1,285 @@
...
+    log_print(f"Extracted {len(losses)} steps from TensorBoard events")
+    return losses
+
+
+# ---------------------------------------------------------------------------
+# Reference file I/O
+# ---------------------------------------------------------------------------
+
+
+def read_losses_from_file(path: str) -> dict[int, float]:
```
wwwjn: Same helper function - Can we import from main to avoid maintain 2 copies?

torchtitan/experiments/rl/scripts/loss_compare.py
yichuan-w: Good catch — removed.

yichuan-w:
cc @tianyu-l @shuhuayu @wwwjn — This PR is currently blocked by a CI installation issue (see failing job: https://github.com/pytorch/torchtitan/actions/runs/27046478004/job/79833428103?pr=3474). Because of this, I'm unable to obtain the actual loss values. That said, the code itself should be ready for review.

wwwjn (approved):
The structure looks much clear! Can you rebase to latest main and potentially update the loss file? #3582  merged which changed the task from SumDigits to AlphaBetaSort

yichuan-w:
## Loss guard on AlphabetSort is degenerate (all-zero) — generator-seeding root cause

**TL;DR:** After rebasing onto latest main (with #3582 SumDigits → AlphabetSort), the RL loss-guard run produces `loss/mean: 0` for **every** step (100/100), with `train/grad_norm/mean: 0` and `rollout_reward/zero_std_frac: 1.0`. So I don't want to commit an all-zero reference — it wouldn't guard anything. The cause is in the generator's determinism handling, not in this PR. Details + a proposed fix below.

### Why the loss is exactly 0

GRPO's advantage is `(reward - group_mean) / group_std`. If all `group_size` samples in a group are identical, `group_std == 0` → advantage `== 0` → per-token PG loss `== 0` → `loss/mean == 0` and `grad_norm == 0` (the policy never updates). The metric `rollout_reward/zero_std_frac: 1.0` confirms **every** group has zero reward std, i.e. the samples within each group are identical.

### Why the samples are identical now but weren't before

Both the old and new setups are "deterministic, seed=42", but they achieve determinism differently:

**Old** — `unified/actors/generator.py` (this is what produced the previous *non-zero* SumDigits reference):
```python
    ...
    n=self.num_samples_per_prompt,   # n == group_size, in ONE request
    seed=self.config.seed,           # seed=42
)
```
A single request asks vLLM for `n = group_size` samples; the seeded RNG **advances across the n samples**, so the group is *both* reproducible *and* diverse → non-zero reward variance → non-zero loss.

**New** — `actors/generator.py` (current main, continuous-batching generator):
```python
def _build_sampling_params(self, sampling: SamplingConfig) -> SamplingParams:
    return SamplingParams(
        ...
        n=1,                          # n=1, issued group_size times
        seed=self.config.debug.seed,  # SAME seed=42 on every request
    )
```
Each of the `group_size` samples is a separate `n=1` request seeded with the **same** `seed=42`. For a fixed prompt that yields `group_size` **identical** completions → zero reward std → zero advantage → zero loss.

So the change from `n=num_samples` (one request) to `n=1` (one request per sample, same seed) is what collapses intra-group diversity under determinism.

### Proposed fix

Offset the per-request seed by the sample index, so the group is diverse again while remaining reproducible run-to-run:

```python
# actors/generator.py — when building the per-sample request
seed = None if cfg.debug.seed is None else cfg.debug.seed + sample_idx
return SamplingParams(..., n=1, seed=seed)
```

`sample_idx` is already threaded through the rollout path (rollout ids are `.../sample={sample_idx}`). This restores the old `n>1` diversity under the new n=1 continuous-batching architecture, and the run stays bitwise-reproducible across runs for a fixed base seed.

This is a core-generator change (outside the loss-guard scope), so flagging before I update `rl_grpo_cuda.txt`. I'm happy to send the per-sample-seed fix as a separate PR (or fold it in here) and then regenerate the reference. Let me know which you prefer.

yichuan-w:
cc @wwwjn @tianyu-l @felipemello1

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -791,15 +735,28 @@ def _fail_outstanding_futures(self, exc: BaseException) -> None:
...
+        seed = base_seed
+        if base_seed is not None:
+            match = re.search(r"sample=(\d+)", request.request_id)
+            seed = base_seed + (int(match.group(1)) if match else 0)
         return SamplingParams(
             temperature=sampling.temperature,
             top_p=sampling.top_p,
             max_tokens=sampling.max_tokens,
             n=1,  # always expects a single sample per request. Caller can call N times.
             stop_token_ids=self._stop_token_ids or None,
```
tianyu-l: We shouldn't pass in `stop_token_ids` at all https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/actors/generator.py#L377

We should fill `stop_token_ids` in `sampling` before https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/trainer.py#L511

We should put sample-specific seed in `sampling` before https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/rollout/rollouter.py#L189

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -791,15 +735,28 @@ def _fail_outstanding_futures(self, exc: BaseException) -> None:
...
+        ``request_id`` (``.../sample=<idx>/...``). Each sample in a group is a
+        separate ``n=1`` request, so a single fixed seed would make all samples
+        in a group identical -> zero reward std -> zero GRPO advantage. Offsetting
+        keeps runs bitwise-reproducible while restoring intra-group diversity.
+        """
+        sampling = request.sampling
+        base_seed = self.config.debug.seed
+        seed = base_seed
+        if base_seed is not None:
+            match = re.search(r"sample=(\d+)", request.request_id)
```
tianyu-l: This is error-prone: if this code change https://github.com/pytorch/torchtitan/blob/main/torchtitan/experiments/rl/rollout/rollouter.py#L191, it will silently break here. See how to fix in the other comment.

tianyu-l (changes requested):

torchtitan/experiments/rl/actors/generator.py
```diff
@@ -791,15 +735,22 @@ def _fail_outstanding_futures(self, exc: BaseException) -> None:
             self._pull_model_state_dict_future = None
             self._model_state_dict_pull_request = None
 
-    def _build_sampling_params(self, sampling: SamplingConfig) -> SamplingParams:
-        """Translate a `SamplingConfig` into vLLM `SamplingParams` (n=1; seed from debug)."""
+    def _build_sampling_params(self, request: GenerationRequest) -> SamplingParams:
```
tianyu-l: no need to change function signature

tianyu-l (approved):
nice, one more fix

